PR #1071, opened 2026-06-17
fix(wrap): percent-encode non-ASCII cwd names in X-Headroom-Project header

### LINKED ISSUE #1069: [BUG] `X-Headroom-Project` header rejects non-ASCII cwd names (e.g. Chinese / Japanese / Korean)
### What happened

When the current working directory contains non-ASCII characters (e.g. Chinese) or spaces, `headroom wrap claude` fails immediately on the first API call.

### Repro

```bash
mkdir -p /tmp/test-中文-项目
cd /tmp/test-中文-项目
source ~/.local/venv/headroom/bin/activate
headroom wrap claude -- -p "hi"
```

### Actual output

```
Starting Headroom proxy on port 8787...
Proxy ready on http://127.0.0.1:8787
...
Launching Claude Code (API routed through Headroom)...
ANTHROPIC_BASE_URL=http://127.0.0.1:8787
ENABLE_TOOL_SEARCH=true (on-demand tool loading kept on; issue #746)
API Error: Header 'X-Headroom-Project' has invalid value: '第二大脑共享'
```

(I also reproduced this with a folder called `第二大脑共享` — i.e. a project path that contains only non-ASCII characters, no spaces, so spaces alone are not the trigger.)

### Expected

Either:
- The project name should fall back to a sanitized ASCII slug (e.g. `second-brain-shared`), or
- The header value should be URL-encoded / percent-encoded before being sent, or
- The proxy should accept any RFC 7230 token (which permits non-ASCII via UTF-8 + `token68` / `quoted-string`)

### Environment

- headroom version: **0.26.0**
- OS: macOS 26.6 (aarch64)
- Install: `uv pip install "headroom-ai[all]"` into a fresh venv (Python 3.11.15)
- CWD: `/Volumes/ai系统盘/第二大脑共享/` and `/tmp/test-中文-项目/` both reproduce
- CWD with pure ASCII (e.g. `/tmp/`) works fine

### Notes

I searched existing issues and didn't find a duplicate. The error fires before the first token of the user prompt is sent, so the proxy never gets a chance to compress anything — the whole wrap session is unusable in non-ASCII project directories. This blocks adoption in any team with Chinese / Japanese / Korean / Cyrillic project names.

Thanks for looking into this!

MrAshRhodes:
## Description

Non-ASCII directory names (Chinese, Japanese, Korean, Cyrillic) caused an immediate API error when using `headroom wrap claude`:

```
API Error: Header 'X-Headroom-Project' has invalid value: '第二大脑共享'
```

RFC 7230 requires HTTP header values to be visible ASCII only. The raw cwd basename was being sent directly, breaking the entire session before the first token.

Closes #1069

## Type of Change

## Changes Made

- `headroom/cli/wrap.py` — `_project_name_from_cwd()`: percent-encode non-ASCII chars via `urllib.parse.quote(name, safe="-_.() ")` so the header value is always ASCII-safe
- `headroom/proxy/savings_tracker.py` — `sanitize_project_name()`: `urllib.parse.unquote()` before cleanup so the stored/displayed project name is the original Unicode directory name

ASCII-only project names are unaffected (quote/unquote is a no-op for them).

## Testing

### Test Output

```text
tests/test_cli/test_wrap_helpers.py::TestApplyProjectHeaderEnv::test_non_ascii_cwd_name_is_percent_encoded PASSED
tests/test_cli/test_wrap_helpers.py::TestApplyProjectHeaderEnv::test_non_ascii_cwd_header_is_ascii_safe PASSED
tests/test_proxy_project_savings.py::test_sanitize_project_name_decodes_percent_encoded_non_ascii PASSED
======================== 15 passed, 1 warning in 0.42s =========================
```

## Real Behavior Proof

- Environment: macOS 15, Python 3.11.9, headroom dev install from source
- Exact command / steps: `mkdir /tmp/test-中文-项目 && cd /tmp/test-中文-项目`, then run `.venv/bin/pytest tests/test_cli/test_wrap_helpers.py::TestApplyProjectHeaderEnv::test_non_ascii_cwd_header_is_ascii_safe -v` — header_value.encode("ascii") passes without UnicodeEncodeError
- Observed result: `X-Headroom-Project` header contains percent-encoded ASCII (`test-%E4%B8%AD%E6%96%87-%E9%A1%B9%E7%9B%AE`); proxy decodes back to `test-中文-项目` for storage
- Not tested: live end-to-end wrap session with a real Claude API key

## Review Readiness

## Checklist


JerrettDavis (approved):
LGTM!

PR #1078, opened 2026-06-17
fix(wrap): write env.ANTHROPIC_BASE_URL to settings.json so daemon-spawned conversations inherit proxy (#951)

### LINKED ISSUE #951: [BUG] headroom wrap claude: daemon child sessions don't inherit ANTHROPIC_BASE_URL, bypassing proxy after first conversation
## Description
When using `headroom wrap claude`, only the first conversation routes traffic 
through the Headroom proxy. Any subsequent conversation created via the Claude 
Code conversation list bypasses the proxy entirely and hits api.anthropic.com 
directly, meaning no compression or token savings are applied.

## To Reproduce
1. Install headroom with `pip install "headroom-ai[all]"`
2. Run `headroom wrap claude`
3. Send a message in the first conversation — proxy receives the request ✅
4. Press left arrow → conversation list → create a new conversation
5. Send a message — proxy receives nothing ❌

## Expected Behavior
All conversations, including those created after initial launch, should route 
through the Headroom proxy at `http://127.0.0.1:8787`.

## Actual Behavior
Only the first conversation routes through the proxy. New conversations spawn 
from Claude Code's pre-forked daemon process pool, which does not have 
`ANTHROPIC_BASE_URL` set. Those sessions talk directly to api.anthropic.com.

Confirmed via environment inspection of running processes: the daemon process 
correctly has `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`, but its spawned 
child session processes do not inherit it.

Setting `apiBaseUrl` in `~/.claude/settings.json` does not fix this — child 
sessions still bypass the proxy.

## Code Sample
N/A — this is a CLI/daemon issue, not a library issue.

## Error Output
No explicit error. New conversations silently bypass the proxy with zero 
entries appearing in `curl http://127.0.0.1:8787/stats` after sending messages.

## Environment
- **Headroom version**: 0.25
- **Python version**: 3.13.3
- **OS**: macOS
- **LLM Provider**: Anthropic (Claude subscription, OAuth bearer token auth)

## Additional Context
Claude Code runs a persistent background daemon (`cc-daemon-501`) that 
pre-forks spare processes for new conversations. `headroom wrap claude` sets 
`ANTHROPIC_BASE_URL` on the top-level process, but the daemon's child spawn 
path drops this env var. The fix likely requires killing any existing daemon 
before launching, so the daemon itself is started under Headroom's environment 
and all children inherit it correctly.

MrAshRhodes:
## Description

Claude Code pre-forks conversation workers via spawn (not fork) on macOS. Those workers read settings files fresh on each new session rather than inheriting the daemon process's environment. `headroom wrap claude` was passing `ANTHROPIC_BASE_URL` only via `subprocess.run`'s `env` dict, which reaches the initial Claude Code process and the daemon — but not conversation workers spawned later from the daemon pool. New conversations silently bypassed the proxy and hit `api.anthropic.com` directly.

### Design decision: why project-local settings

Three approaches were considered:

**1. Global `~/.claude/settings.json`** — rejected. This file is shared across every Claude Code session on the machine. A user who runs `headroom wrap claude` in one terminal but opens an unwrapped session elsewhere would have their global settings rewritten to point at the Headroom proxy. If the proxy dies and cleanup does not run (SIGKILL, crash), the stale URL breaks all future sessions until the user manually edits the global file.

**2. Kill cc-daemon before launch** — rejected. The issue itself suggests this, but killing the daemon is disruptive: it destroys the pre-forked worker pool shared by any other open Claude Code windows. Active conversations may lose their parent process. This is a hard-to-reverse side-effect of a command the user expects to be safe.

**3. Project-local `<cwd>/.claude/settings.local.json`** — chosen. Claude Code applies `env` keys from project-local settings per its documented precedence order (Local > Project > User), and reloads them per-conversation. Scoping to the project means: other projects and unwrapped sessions are unaffected; the file is git-ignored by default so it won't be committed; and the worst-case stale URL (proxy crash without cleanup) affects only that one project's local settings and is trivially recoverable by re-running `headroom wrap claude` or deleting the file.

Closes #951

## Type of Change

## Changes Made

- Added `_write_claude_wrap_base_url(proxy_url, *, foundry_mode, settings_path)` in `headroom/cli/wrap.py`: merges `ANTHROPIC_BASE_URL` (or `ANTHROPIC_FOUNDRY_BASE_URL` in foundry mode) into `<cwd>/.claude/settings.local.json` under the `env` key. Returns the previous value for restore.
- Added `_restore_claude_wrap_base_url(previous, *, foundry_mode, settings_path)`: called in the `wrap claude` `finally` block and in `unwrap_claude` to remove or restore the key so a stale proxy URL is never left behind.
- `unwrap_claude` calls restore for both standard and foundry keys.
- New test file `tests/test_cli/test_wrap_claude_base_url.py` (12 tests covering write, restore, roundtrip, foundry mode, sibling key preservation, and noop on absent file).

## Testing

### Test Output

```text
pytest tests/test_cli/test_wrap_claude_base_url.py -v
12 passed in 0.21s

ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_claude_base_url.py
All checks passed!
```

## Real Behavior Proof

- Environment: macOS, branch `fix/daemon-base-url-inheritance`, Python 3.11.9.
- Exact command / steps: Ran `pytest tests/test_cli/test_wrap_claude_base_url.py -v` and `ruff check` on the modified files from the PR branch.
- Observed result: 12 new unit tests pass; ruff reports no issues.
- Not tested: Live end-to-end verification (opening a second conversation via the daemon pool and confirming proxy receives traffic) — not safe to test inside the current wrapped session on port 8787.

## Review Readiness

## Checklist

## Additional Notes

The issue reporter tried `apiBaseUrl` in settings.json and found it ineffective. That key configures the API endpoint at the CC UI layer, not the process environment. `env.ANTHROPIC_BASE_URL` is the correct mechanism for propagating an environment variable to CC worker processes.

### DIFF
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -575,6 +575,85 @@ def _remove_claude_rtk_hooks(settings_path: Path | None = None) -> bool:
     return True
 
 
+def _write_claude_wrap_base_url(
+    proxy_url: str,
+    *,
+    foundry_mode: bool = False,
+    settings_path: Path | None = None,
+) -> str | None:
+    """Persist proxy URL into project-local settings env key for daemon child inheritance.
+
+    Claude Code's cc-daemon pre-forks conversation workers using spawn (not
+    fork), so those workers read settings.json fresh rather than inheriting
+    the daemon's environment.  Writing env.ANTHROPIC_BASE_URL into the
+    project-local settings file (.claude/settings.local.json in cwd) ensures
+    every new conversation — including those started after the initial launch —
+    routes through the Headroom proxy without touching the global user settings
+    file or affecting sessions in other projects.  Returns the previous value
+    so the caller can restore it on exit (issue #951).
+    """
+    path = settings_path or (Path.cwd() / ".claude" / "settings.local.json")
+    payload: dict[str, Any] = {}
+    if path.exists():
+        try:
+            payload = json.loads(path.read_text(encoding="utf-8"))
+        except (OSError, json.JSONDecodeError):
+            payload = {}
+    if not isinstance(payload, dict):
+        payload = {}
+    env_map = dict(payload.get("env") or {}) if isinstance(payload.get("env"), dict) else {}
+    key = "ANTHROPIC_FOUNDRY_BASE_URL" if foundry_mode else "ANTHROPIC_BASE_URL"
+    previous = env_map.get(key)
+    env_map[key] = proxy_url
+    payload["env"] = env_map
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(json.dumps(payload, indent=2) + "\n", encoding="utf-8")
+    return previous
+
+
+def _restore_claude_wrap_base_url(
+    previous: str | None,
+    *,
+    foundry_mode: bool = False,
+    settings_path: Path | None = None,
+) -> None:
+    """Restore (or remove) the env key written by _write_claude_wrap_base_url.
+
+    Called in both the wrap-session finally block and unwrap_claude so the
+    project-local settings entry is never left pointing at a dead proxy.  When
+    ``previous`` is None the key is removed; when it has a value it is
+    restored — preserving any URL the project already had set.
+    """
+    path = settings_path or (Path.cwd() / ".claude" / "settings.local.json")
+    if not path.exists():
+        return
+    try:
+        payload = json.loads(path.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError):
+        return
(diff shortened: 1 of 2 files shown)


MrAshRhodes:
CI note:  failed on  — a 10s timeout concurrency/deadlock test that is unrelated to this PR's changes (only  and  were touched). Main branch last CI run passed this same test. This is a known-flaky runner-timing failure.

JerrettDavis (approved):

PR #1084, opened 2026-06-17
fix: slugify project header values

ketpatil77:
## Description
Fix `X-Headroom-Project` so it stays ASCII-safe across Windows paths and non-ASCII current working directories.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation only

## Changes Made
- Slugify the project name derived from the cwd before sending it in the header.
- Add a stable fallback when cwd contains no ASCII-safe characters.
- Add tests for spaced paths and non-ASCII cwd names.

## Testing
- Added unit tests for the helper behavior.
- Verified the header derivation no longer depends on raw non-ASCII cwd text.

## Real Behavior Proof
Before this change, a cwd like `C:\Users\जयेश\Documents\New project` could produce a header value with unsupported characters. After this change, the header is normalized to an ASCII-safe slug or fallback value.

## Review Readiness


JerrettDavis (approved):
Reviewed the project-header slugification. The client-side ASCII slug/fallback prevents invalid HTTP header values before proxy-side sanitization, and the new tests cover spaces and non-ASCII cwd names. CI is green.

JerrettDavis:
Thank you for fixing the project header safety issue here. I refreshed the branch against current main and confirmed there is no remaining diff: main now uses an ASCII-safe, reversible project header encoding for non-ASCII cwd names. Closing this PR as superseded by main while preserving the contribution history in the branch. Local validation: direct `_project_name_from_cwd()` / `_apply_project_header_env()` check passed for a non-ASCII project name, and `git diff --check` passed.

PR #1086, opened 2026-06-17
ci: add Windows wheel build job (win_amd64)

sgfdhgfnjgmhgmnfd657-sys:
### Summary

Adds `build-wheel-windows` job to the CI pipeline that compiles the Rust extension on `windows-latest` and uploads the resulting `.whl` as a separate artifact (`headroom-wheel-windows`).

This addresses the long-standing missing Windows wheel.

### Changes

- New job `build-wheel-windows`: mirrors the existing `build-wheel` (Linux) job
- Uses `dtolnay/rust-toolchain@stable` for Rust setup on Windows
- Uses `Swatinem/rust-cache` for dependency caching
- Builds with CI cargo profile for speed
- 45-minute timeout (Windows Rust builds are slower)
- Uploads wheel as `headroom-wheel-windows` artifact

### Testing

✅ **Local compilation verified**: built v0.26.0 from source on Windows 10 (Python 3.12.10, Rust 1.96.0, MSVC Build Tools 2022). The wheel installed and ran successfully.

### Notes

Only the CI-profile build is added here. The release-wheel publish (`release.yml`) can be updated in a follow-up PR once this basic Windows build is proven in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,8 @@ name: CI
 #
 # Notes: CPU-only torch everywhere (no CUDA stack); test shards run HF_HUB_OFFLINE.
 # Multi-version (3.10/3.11/3.13) coverage on main is a planned follow-up.
+# Windows wheel (win_amd64) built separately — builds the Rust ext just like the
+# Linux wheel, then uploads as a separate artifact for downstream consumption.
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


JerrettDavis (approved):
Reviewed the added Windows wheel job against the existing CI workflow. It is isolated behind the same code-change gate, mirrors the Linux wheel build path, uploads a distinct artifact, and the PR checks are green.

PR #1102, opened 2026-06-18
fix(parser): whitespace waste signal always reported zero

nyxst4ck:
## Description

`detect_waste_signals` in `headroom/parser.py` computed the `whitespace_tokens` waste signal as `0` for every input, so collapsible whitespace never contributed to `WasteSignals.total()`.

The root cause was this normalization step:

```python
normalized_text = " ".join(ws_matches)
```

That keeps each matched whitespace run verbatim and only inserts one extra space between runs. The normalized text is therefore never shorter than the original matched whitespace text. This PR collapses each matched run to one space instead:

```python
normalized_text = " "  len(ws_matches)
```

## Type of Change

- [x] Bug fix
- [x] Test coverage
- [x] Changelog entry

## Changes Made

- Fixed the whitespace waste calculation in `headroom/parser.py`.
- Tightened the existing excessive-whitespace test so it asserts real savings instead of allowing `0`.
- Added a regression test proving the signal scales with long whitespace runs.

### Test Output

```text
88 parser tests passed
ruff check: passed
ruff format --check: passed
```

- [x] `uv run pytest tests/test_parser.py` -> `88 passed`
- [x] `ruff check`
- [x] `ruff format --check`

## Real Behavior Proof

- Environment: current PR head, pure-Python parser path.
- Exact command / steps: ran `uv run pytest tests/test_parser.py`, Ruff, and formatting checks with long whitespace-run coverage.
- Observed result: 88 parser tests pass; fixture waste is 2 tokens and a realistic padded log reports 65 rather than zero.
- Not tested: compiled Rust, hnswlib, torch, or live proxy paths because the changed path is pure Python.

Before this fix, the existing whitespace fixture and a padded-log case both returned `signals.whitespace_tokens == 0`, even though the matched whitespace was collapsible.

After this fix:

```text
fixture whitespace_tokens: 2
realistic padded log whitespace_tokens: 65
```

The new tests fail against `main` and pass with this branch.

## Review Readiness

## Scope Notes

I did not retest compiled Rust, `hnswlib`, torch, or live proxy paths on this Windows machine. The changed path is pure Python and covered by `tests/test_parser.py`.


JerrettDavis (approved):
Reviewed the whitespace waste fix. Collapsing each matched run to one space matches the parser comment and the new tests pin the previously-zero signal. CI is green.

JerrettDavis:
Maintainer refresh: merged current main and resolved the stale `CHANGELOG.md` conflict to main while preserving this PR's parser fix and tests. Local validation: `tests/test_parser.py` passed (`88 passed`) and `git diff --check` passed.

JerrettDavis (approved):
Re-reviewed after the whitespace waste-signal refresh. The fix now correctly models each excessive whitespace run as one normalized space, and the regression tests cover both the fixture case and larger runs.

Local validation:
- `uv run --with pytest --with pytest-asyncio python -m pytest tests/test_parser.py -q` (88 passed)
- `uv run ruff check headroom/parser.py tests/test_parser.py`

Approval stands.

nyxst4ck:
Thanks for refreshing the branch and re-running validation. I checked the PR body against the governance note: the Testing section now includes concrete command output, the Real Behavior Proof section includes environment/steps/observed result/not-tested scope, and the review readiness boxes are checked. The current PR Governance checks are also passing (	emplate and label).

nyxst4ck:
I inspected the two failing CI matrix jobs. Both fail before exercising this PR's parser change because the workflow sets `TRANSFORMERS_OFFLINE=1` while the `all-MiniLM` Hugging Face cache is absent; `SentenceTransformer` then raises `LocalEntryNotFoundError` / `OSError` while memory tests load the model.

The PR changes only `headroom/parser.py` and `tests/test_parser.py`, so I am not changing it to mask an unrelated offline-model cache failure. The parser-specific local test passed during the prior refresh. A workflow/cache repair or rerun with the model artifact available is needed to clear these two jobs.

nyxst4ck:
Rebased onto current main to trigger a fresh CI run. Parser-only validation is green: 94/94 tests, Ruff check, Ruff format check, and git diff --check. The full local dev environment still requires native MSVC tooling, so I did not alter the parser to work around unrelated infrastructure.

JerrettDavis (approved):
Re-reviewed current head cdc475f5. The whitespace waste calculation now models the intended normalization correctly: each excessive whitespace run collapses to a single space, so long runs produce nonzero savings instead of cancelling themselves out.

Focused local checks passed:
- `uv run python -m pytest tests/test_parser.py -q`
- `uv run ruff check headroom/parser.py tests/test_parser.py`

GitHub's merge-state label is still `UNSTABLE`, but the status rollup for this head shows no failing or incomplete checks.

PR #1089, opened 2026-06-17
feat(opencode): native OpenCode support — provider discovery, wrap/unwrap, MCP, install

hareeshkar:
## Description

Adds native OpenCode support to Headroom: \`headroom wrap opencode\` and \`headroom unwrap opencode\` with **transparent fetch-level interception**, full provider discovery, MCP registration, install infrastructure, and centralized port management.

**Why:** Headroom has no OpenCode integration today. Users must manually configure \`headroom proxy\` and set \`OPENAI_BASE_URL\` / \`ANTHROPIC_BASE_URL\`. This PR makes OpenCode a first-class target and covers **all providers — including those added mid-session via \`/connect\`** — without touching the user's on-disk config.

Closes #1193

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Architecture: How It Works

```
headroom wrap opencode
│
├── build_launch_env() injects three env vars:
│   ├── OPENCODE_CONFIG_CONTENT  — known providers at launch (model picker shows real names)
│   ├── HEADROOM_PROXY_URL       — http://127.0.0.1:{port}
│   └── NODE_OPTIONS             — --import=/path/to/shim.mjs
│
└── OpenCode process starts
    ├── shim.mjs loaded via NODE_OPTIONS (before any userland code — ESM preload)
    │   └── patches globalThis.fetch (the only transport Vercel AI SDK uses)
    │       ├── External AI call? → set x-headroom-base-url: <origin>
    │       │                     → rewrite URL to HEADROOM_PROXY_URL
    │       │                     → retry 3× at 75ms if proxy not ready (cold-start grace)
    │       ├── Loopback / localhost? → pass through (no proxy→proxy loops)
    │       └── Already proxied? → pass through
    │
    └── Headroom proxy receives request
        └── x-headroom-base-url → normalize_upstream_base() → forward to real provider
            (proxy_routes.py lines 428, 482, 585, 884 — no proxy changes needed)
```

**Mid-session `/connect` provider:** User adds a new provider mid-session. The shim already intercepts all outbound HTTPS calls — no config change, no restart required.

[...]

PR #1055, opened 2026-06-16
feat: add headroom wrap opencode command

Epicism:
## Description

Add `headroom wrap opencode` and `headroom unwrap opencode` commands to configure OpenCode to use the Headroom proxy, similar to how Claude and Codex are configured.

Closes #1018

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Created `headroom/providers/opencode/__init__.py` and `runtime.py`
- Created `headroom/mcp_registry/opencode.py` with `OpenCodeRegistrar` class
- Created `tests/test_opencode_registrar.py` with 24 unit tests
- Modified `headroom/cli/wrap.py` with `wrap opencode` and `unwrap opencode` commands
- Modified `headroom/mcp_registry/__init__.py` to export `OpenCodeRegistrar`
- Modified `headroom/mcp_registry/install.py` to include OpenCode in registrars
- Modified `headroom/install/models.py` to add `OPENCODE` to `ToolTarget`

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ uv run pytest tests/test_opencode_registrar.py -v
24 passed in 0.08s
```

## Real Behavior Proof

- Environment: macOS, Python 3.14.6, OpenCode 1.17.7, headroom from source
- Exact command / steps: uv run headroom wrap opencode --prepare-only --no-context-tool --no-serena
- Observed result: Deepseek provider baseURL overridden to http://127.0.0.1:8787/v1
- Not tested: Memory MCP, code graph, Serena MCP

## Review Readiness

- [x] I have performed a self-review
- [x] This PR is ready for human review

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have updated the CHANGELOG.md if applicable

## Additional Notes

[...]

PR #1105, opened 2026-06-18
feat: headroom wrap opencode / unwrap opencode CLI

rudironsoni:
## Summary

This PR implements transparent `headroom wrap opencode` support without asking users to edit OpenCode provider URLs, choose an extra CLI flag, or maintain a static provider list.

The wrapper now lives at the runtime transport boundary: OpenCode keeps its user/provider config, while Headroom intercepts outbound provider traffic in-process and routes it through the local Headroom proxy.

### Transparent OpenCode wrapping

- `headroom wrap opencode` injects the `headroom-opencode` plugin through `OPENCODE_CONFIG_CONTENT`.
- Existing OpenCode provider URLs are preserved. We do not rewrite user config URLs to point at Headroom.
- Existing `OPENAI_BASE_URL` and `ANTHROPIC_BASE_URL` env vars are preserved.
- Local OpenCode traffic, localhost traffic, and Headroom proxy traffic bypass the shim to avoid loops.

### Runtime transport interception

- Added an OpenCode plugin transport shim that wraps:
  - `globalThis.fetch`
  - `http.request` / `http.get`
  - `https.request` / `https.get`
- External provider calls are routed to the local Headroom proxy.
- The original upstream origin is passed through `x-headroom-base-url`, so the proxy can forward to the real provider without changing OpenCode config.
- External `http2.connect` is blocked loudly instead of allowing direct provider traffic to leak outside Headroom.

### Live provider additions

Provider coverage is no longer based on a static config scan. Because routing happens at outbound request time, providers added mid-session are routed through Headroom automatically as long as they use the covered Node transport paths.

### Subagent and child-process coverage

- The parent OpenCode plugin sets a packaged Node preload shim through `NODE_OPTIONS=--import=.../hook-shim/handler.js`.
- The transport shim patches `child_process.spawn`, `exec`, `execFile`, and `fork` so child Node processes receive the Headroom preload even when OpenCode passes a custom `env`.
- The child-process shim fails closed if it loads without `HEADROOM_OPENCODE_TRANSPORT_PROXY_URL`.
- This closes the subagent leak path where a child Node process could otherwise start without Headroom transport interception.

## Why this goes beyond PR #1089

PR #1089 improves OpenCode provider registration, but it still focuses on provider config shape. This PR moves the enforcement boundary to runtime transport interception.

This PR goes further because:

- No provider URL rewriting is required.
- New providers added mid-session are covered automatically.
- Subagents and child Node processes inherit the Headroom transport shim.
- Direct external HTTP/2 paths fail loudly instead of leaking.
- The wrap remains transparent to the user's OpenCode provider config.
- The wrapper is fail-closed for unsupported child-process preload state.

## Additional robustness fixes

While validating the change in Docker, the full Python suite exposed unrelated Linux/container robustness issues. These are fixed in this PR so the suite is green:

- Binary cache handling now treats cache paths under a non-writable existing parent as unavailable, including when tests run as root in Docker.
- `release_version.py` honors `MANUAL_VER` before git calls so direct script execution works outside a `.git` checkout.
- Test logger isolation now resets relevant Headroom child loggers so proxy logging setup cannot poison later `caplog` tests.
- The scanner missing-path test now uses a guaranteed missing `tmp_path` child instead of relying on `/nonexistent/path`.

## Validation

All implementation validation was run inside Docker.

- Full Python suite from a fresh Docker copy: `6605 passed, 523 skipped`.
- Ruff on changed Python/OpenCode paths: passed.
- OpenCode plugin typecheck: passed.
- OpenCode plugin tests: `9 passed`.
- OpenCode plugin build: passed.
- Hook shim preload smoke test: passed.

## Notes

This PR intentionally does not add a CLI option. `headroom wrap opencode` means full wrap. Either Headroom wraps OpenCode transparently, or the path fails loudly instead of silently leaking provider traffic.


chopratejas:
Hi - how is this different than https://github.com/chopratejas/headroom/pull/1055 ?

rudironsoni:
> Hi - how is this different than #1055 ?

@chopratejas 

Thanks for the review.

**How this is different from #1055**: that PR just adds a wrap.py command that sets OPENAI_BASE_URL and calls subprocess.run("opencode"). 

**This one has the full provider layer**. OpencodeRegistrar handles idempotent MCP registration with backup and restore. Config injection writes a @ai-sdk/openai-compatible provider into opencode.json via OPENCODE_CONFIG_CONTENT. Persistent install works through ToolTarget.OPENCODE in the planner. There are 72 unit tests and an e2e test that all pass inside Docker. 

**The plugins/opencode/ npm package provides createHeadroomProvider, buildOpencodeConfigContent, and a CCR retrieve tool that users can install separately.**

---

I will update the PR body to match the template and address the low code coverage.

vuthanhtrung2010:
Cool PR, I would like to see it merged.

erickvieira:
@rudironsoni, this is an awesome feature, thanks so much! I was actually messing around with custom providers yesterday using the DeepSeek subscription plan via the /connect auth method inside opencode itself. I ended up getting better results by using just OPENCODE_CONFIG_CONTENT and leaving OPENAI_BASE_URL untouched. Happy to share my scripts if it helps give you the full picture, no pressure at all.

One thing I noticed: forcing DeepSeek through OpenAI's "compatibility" endpoints by setting OPENAI_BASE_URL seems to disable the Ctrl+T / variant_cycle system in opencode.

Quick question (and full disclosure, since I don't have the full context to judge the backend correctness myself): did you happen to run this through some less conventional edge cases, especially outside the typical OpenAI/Anthropic circuits? Just wondering if DeepSeek V4 got a spin too, mainly to confirm broad coverage.

Again, huge thanks for this!!! <3

Harishwarrior:
Seems like https://github.com/chopratejas/headroom/pull/1055 got approved. I think the author should wait and check which PR is more suitable to merge.

rudironsoni:
@erickvieira thanks a lot for the detailed notes and for sharing the `/connect` context. That was useful feedback.

I implemented the suggestions in this PR:

- Headroom no longer forces `OPENAI_BASE_URL` or `ANTHROPIC_BASE_URL` over user-provided values. Existing env vars are preserved, so OpenCode provider state from `/connect`, including custom providers such as DeepSeek, is not clobbered by the wrapper.
- The Headroom OpenCode integration now uses `OPENCODE_CONFIG_CONTENT` for the injected Headroom provider config instead of relying on global provider env mutation. That keeps the Headroom provider scoped to the wrapped OpenCode process.
- The injected provider is transparent and does not ship a hardcoded model list, so OpenCode keeps handling model and variant behavior instead of Headroom constraining it.
- I added optional plugin support through `HEADROOM_OPENCODE_PLUGIN` and `append_headroom_plugin()`, so users can opt into the Headroom OpenCode plugin path without disturbing their existing OpenCode provider setup.
- Tests were added around preserving `OPENAI_BASE_URL`, `ANTHROPIC_BASE_URL`, and `HEADROOM_PROJECT`, plus config/plugin behavior and the Docker wrap e2e flow.

I did not run a live DeepSeek V4 credential test, so I do not want to claim that specific provider was fully exercised. The fix is aimed at the exact class of issue you described: avoid forcing DeepSeek or other `/connect` providers through OpenAI compatibility env vars, and let OpenCode keep owning those provider-specific paths.

JerrettDavis (changes requested):

rudironsoni:
Thanks for the review. I pushed a follow-up commit: `62cfba54 test: isolate CI and local state failures`.

What changed in this update:

- Fixed the stale provider-scope assertion so `opencode` is included in the install CLI error message.
- Isolated ML model registry state between lifecycle tests, which fixed the local crash reproduced by running ML registry lifecycle tests before OAuth routing tests.
- Removed local-machine assumptions from tests: the persistent wrap test no longer probes a real `127.0.0.1:8787` service, and traffic learner tests now use resolved temp project paths.
- Fixed traffic learner path anchoring for project paths containing `@`, which was dropping valid anchored patterns as unanchored.
- Fixed telemetry `/stats` reporting so OTEL/Langfuse env configuration is reflected even before FastAPI lifespan startup, while Langfuse still reports unconfigured when no env is present.

Validation was rerun in Docker only, isolated from my local Headroom install:

- Full Python suite: `7045 items / 6 skipped`, exit `0`.
- Focused OpenCode Python suite: `144 passed`.
- OpenCode plugin: typecheck, tests, and build passed.
- Ruff on touched files passed.

Current GitHub state after the push: the old failing checks are no longer on the head commit. The new workflow runs for `62cfba5457f152be5727b008b0b6c020898547b6` are currently `action_required` with no jobs started yet, while PR Governance and GitGuardian passed. So there is nothing failing on the new head yet, but the workflows need maintainer approval before they can run.

DevEstacion:
can we get this merged pls

Newer1107:
Merge needed!

rudironsoni:
@JerrettDavis Review, please?

JerrettDavis (approved):

Newer1107:
Documentation needs to be updated

rudironsoni:
Follow-up docs PR opened for this comment: https://github.com/headroomlabs-ai/headroom/pull/1317

PR #1124, opened 2026-06-18
feat: add HEADROOM_KEEPALIVE_EXPIRY to keep upstream connections warm

akb4q:
## Description

The Python proxy's `httpx.AsyncClient` (in `server.py`) sets `max_connections` and `max_keepalive_connections` but never `keepalive_expiry`, so httpx's default of **5 seconds** applies. Idle upstream connections are dropped after 5s, and any request after a >5s gap pays a fresh TCP + TLS handshake — costly on high-RTT upstream paths. The **Rust** `crates/headroom-proxy` reqwest client already hardcodes `pool_idle_timeout(Duration::from_secs(90))`; the Python path silently differs at 5s. This PR closes that gap.

Closes #

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `ProxyConfig.keepalive_expiry: float = 90.0` (`headroom/proxy/models.py`)
- Wired into `httpx.Limits(keepalive_expiry=...)` (`headroom/proxy/server.py`)
- `HEADROOM_KEEPALIVE_EXPIRY` env in both env-based config builders (`headroom/proxy/server.py`)
- CLI `--keepalive-expiry` (env `HEADROOM_KEEPALIVE_EXPIRY`) following the existing `--max-keepalive` option pattern (`headroom/cli/proxy.py`)
- Docs row in `configuration.mdx` + a CLI env test in `tests/test_cli_proxy_env.py`
- Default of 90s matches the Rust path; operators can override (e.g. back to `5`).

## Testing

### Test Output

```text
$ ruff check headroom/proxy/models.py headroom/proxy/server.py headroom/cli/proxy.py tests/test_cli_proxy_env.py
All checks passed!
$ ruff format --check (same files)
4 files already formatted
```

I did not run the full `pytest` suite locally (it requires a maturin build + heavy optional deps). The added test mirrors the existing `test_cli_proxy_env.py` patterns and the CLI option follows the adjacent `--max-keepalive` exactly.

## Real Behavior Proof

- Environment: a live headroom deployment (installed `headroom-ai`, Python 3.11) reaching an upstream over a high-RTT tunnel.
- Exact command / steps: applied the same field change, restarted the proxy, then inspected the live config.
- Observed result: `ProxyConfig.keepalive_expiry == 90.0` at runtime; proxy serves normally; sparse upstream requests no longer re-handshake within the 90s window (the ~300ms cold-handshake penalty that previously recurred after the 5s default expiry is gone).
- Not tested: full `pytest`/`mypy` suite locally (maturin build).

## Review Readiness

## Additional Notes

Default changes from httpx's implicit 5s to 90s to reach parity with the Rust `pool_idle_timeout(90s)`; this is the intended behavior alignment rather than a silent regression. CHANGELOG not touched (no entry pattern for proxy knobs observed); happy to add one if preferred.

### DIFF
--- a/docs/content/docs/configuration.mdx
+++ b/docs/content/docs/configuration.mdx
@@ -223,6 +223,7 @@ headroom proxy --learn --min-evidence 3
 | `HEADROOM_LIMIT_CONCURRENCY` | Maximum concurrent connections before 503 | `1000` |
 | `HEADROOM_MAX_CONNECTIONS` | Maximum upstream HTTP connections | `500` |
 | `HEADROOM_MAX_KEEPALIVE` | Maximum upstream keep-alive connections | `100` |
+| `HEADROOM_KEEPALIVE_EXPIRY` | Seconds an idle upstream keep-alive connection is kept open | `90` |
 | `HEADROOM_BUDGET` | Daily budget limit in USD | -- |
 | `HEADROOM_TELEMETRY` | Set to `off` to disable anonymous telemetry | enabled |
 | `HEADROOM_STATELESS` | Set to `true` to disable filesystem writes | `false` |
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -138,6 +138,14 @@ def _selected_context_tool() -> str:
     help="Maximum upstream keep-alive connections (default: 100, env: HEADROOM_MAX_KEEPALIVE)",
 )
 @click.option(
+    "--keepalive-expiry",
+    "keepalive_expiry",
+    default=90.0,
+    type=click.FloatRange(min=0),
+    envvar="HEADROOM_KEEPALIVE_EXPIRY",
+    help="Seconds an idle upstream keep-alive connection is kept open (default: 90, env: HEADROOM_KEEPALIVE_EXPIRY)",
+)
+@click.option(
     "--mode",
     default=None,
     metavar="[token|cache]",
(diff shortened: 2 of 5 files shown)


JerrettDavis (changes requested):
The keepalive setting is not fully wired through the direct server CLI. This PR adds the Click headroom proxy --keepalive-expiry option and env handling, but the python -m headroom.proxy.server argparse path still has no --keepalive-expiry argument and builds ProxyConfig.keepalive_expiry from HEADROOM_KEEPALIVE_EXPIRY with a hardcoded 90.0 fallback instead of an argparse value. Please add the argparse flag next to --max-keepalive and use args.keepalive_expiry as the fallback, with a focused test if possible.

akb4q:
Thanks @JerrettDavis, you're right — the flag wasn't wired through the direct `python -m headroom.proxy.server` argparse path. Fixed in 656027c:

- Added the `--keepalive-expiry` argparse flag next to `--max-keepalive` (type `float`, default `90.0`).
- The argparse-build path now uses `_get_env_float("HEADROOM_KEEPALIVE_EXPIRY", args.keepalive_expiry)` instead of the hardcoded `90.0`, matching the env+args pattern used by `--max-connections`/`--max-keepalive`.
- Left `_proxy_config_from_env()` as-is — it's the pure-env path with no `args`, so the `90.0` literal is correct there.
- Added a focused subprocess test that runs the real server parser and asserts `--keepalive-expiry` is registered as a float (a bad value exits argparse before the server boots, so it stays fast).

PTAL.

JerrettDavis (approved):

PR #1129, opened 2026-06-18
Marker-free lossless_only mode + gate opaque-blob CCR markers behind enable_ccr_marker

### LINKED ISSUE #1091: [BUG] `enable_ccr_marker=False` does not suppress opaque-blob CCR markers — no config yields guaranteed-lossless SmartCrusher output
## Description

SmartCrusher's compaction path is **lossless-first** for arrays (good — it
attempts CSV+schema compaction and only falls through to the lossy row-drop
path when lossless savings are below `lossless_min_savings_ratio`). But there
is **no configuration that produces marker-free output**, because the
opaque-blob cell substitution path emits `<<ccr:HASH,string,KB>>` markers
**unconditionally** — it does not honor `enable_ccr_marker` /
`inject_retrieval_marker`, which gate only the row-drop sentinel path.

The consequence: an array that compacts *losslessly in structure* still emits a
CCR marker if any single string cell exceeds `opaque_min_bytes` (256 by
default). For any consumer that needs guaranteed-lossless output (e.g. an agent
that re-reads its own tool results, where a `<<ccr:...>>` marker forces a
retrieval round-trip and lands the content in context twice), there is no way
to opt out.

## The asymmetry (code references)

The row-drop sentinel path **is** gated by the flag —
`crates/headroom-core/src/transforms/smart_crusher/crusher.rs:735`:

```rust
let (ccr_hash, dropped_summary) = if dropped_count > 0 && self.config.enable_ccr_marker {
    // ... emit <> + store write
```

The opaque-blob path is **not**. A string cell longer than `opaque_min_bytes`
is classified `Opaque` purely on byte length, with no reference to the flag —
`compaction/classifier.rs:97-98`:

```rust
// Opaque-blob check — only for strings above the byte threshold.
if s.len() <= cfg.opaque_min_bytes {
    return CellClass::Scalar;
}
// → CellClass::Opaque(...) otherwise
```

…and is rendered to a marker unconditionally: `CellClass::Opaque` →
`CellValue::OpaqueRef` in `compaction/compactor.rs:220`, the `OpaqueRef` arm in
`compaction/formatter.rs:307` calls `format_ccr_marker` (`:311`) with no flag
check, and the walker's `emit_opaque_ccr_marker` (`compaction/walker.rs:142-144`,
definition at `:171`) is likewise ungated.

The code documents this as the intended scope of the gate
(`headroom/transforms/smart_crusher.py:270-271`):

> Scope: gates ONLY the row-drop sentinel path. Stage-3c.2 opaque-string CCR
> substitutions still emit always […]

…and the `enable_ccr_marker` docstring itself repeats the limited scope
(`config.rs:68-82`). So this is by-design today — but it leaves no path to
lossless-only output.

## Why this matters (and where the eval numbers come from)

The public eval in NousResearch/hermes-agent#47866 measured Headroom's
compression path against ~3,000 real agent tool outputs under a *reject-on-any-
marker* policy (lossless-only). It reported ~0.34% net savings and only ~2.5%
on `search_files`.

That number is dominated by this asymmetry, not by densification failing:

- A `search_files` content result is an array of `{path, line, content}`. It
  compacts losslessly in structure (path elision, key elision).
- But if **one** matched line exceeds 256 bytes, that single cell becomes an
  opaque `<<ccr:...>>` marker — which can't be disabled — so the **entire
  output** is rejected by a lossless-only gate, scoring ~0.
- A native densifier with no opaque-substitution step hit **57.8%** lossless on
  the same shape. The gap is the ungated opaque path, not the compactor.

In other words: SmartCrusher's structural densification is genuinely lossless
and competitive; it's the inability to suppress opaque-cell offloading that
makes "lossless-only" unusable today.

```python
import json
from headroom.transforms.smart_crusher import SmartCrusher
from headroom.config import CCRConfig

# CCR fully "disabled"
crusher = SmartCrusher(ccr_config=CCRConfig(enabled=False, inject_retrieval_marker=False))

rows = [{"path": "a.py", "line": i, "content": "x" * 300} for i in range(10)]
out = crusher.crush(json.dumps(rows))

print(out.compressed)  # contains <> per >256B content cell

# Expected (lossless-only intent): no <> in output
# Actual: markers present despite enable_ccr_marker=False
assert "<<ccr:" not in out.compressed   # fails
```

## Proposed fix

Make opaque-blob substitution respect the existing flag, and expose a clear
lossless-only contract:

1. **Gate `OpaqueRef` emission on `enable_ccr_marker`.** When the flag is off,
   leave long string cells inline (verbatim) instead of substituting a marker.
   This makes `enable_ccr_marker=False` mean what it says: no CCR markers, from
   *any* path.
2. **Add an explicit `lossless_only` / `no_markers` mode** (config + `compress()`
   kwarg) that guarantees marker-free output: lossless densification applies,
   anything that would require a marker (row-drop *or* opaque offload) is left
   uncompressed rather than offloaded.
3. Keep current behavior as the default so existing proxy/CCR users are
   unaffected.

## Actual Behavior

The public eval in NousResearch/hermes-agent#47866 measured Headroom's
compression path against ~3,000 real agent tool outputs under a *reject-on-any-
marker* policy (lossless-only). It reported ~0.34% net savings and only ~2.5%
on `search_files`.

That number is dominated by this asymmetry, not by densification failing:

- A `search_files` content result is an array of `{path, line, content}`. It
  compacts losslessly in structure (path elision, key elision).
- But if **one** matched line exceeds 256 bytes, that single cell becomes an
  opaque `<<ccr:...>>` marker — which can't be disabled — so the **entire
  output** is rejected by a lossless-only gate, scoring ~0.
- A native densifier with no opaque-substitution step hit **57.8%** lossless on
  the same shape. The gap is the ungated opaque path, not the compactor.

In other words: SmartCrusher's structural densification is genuinely lossless
and competitive; it's the inability to suppress opaque-cell offloading that
makes "lossless-only" unusable today.

## The asymmetry (code references)

The row-drop sentinel path **is** gated by the flag —
`crates/headroom-core/src/transforms/smart_crusher/crusher.rs:735`:

```rust
let (ccr_hash, dropped_summary) = if dropped_count > 0 && self.config.enable_ccr_marker {
    // ... emit <> + store write
```

The opaque-blob path is **not**. A string cell longer than `opaque_min_bytes`
is classified `Opaque` purely on byte length, with no reference to the flag —
`compaction/classifier.rs:97-98`:

```rust
// Opaque-blob check — only for strings above the byte threshold.
if s.len() <= cfg.opaque_min_bytes {
    return CellClass::Scalar;
}
// → CellClass::Opaque(...) otherwise
```

…and is rendered to a marker unconditionally: `CellClass::Opaque` →
`CellValue::OpaqueRef` in `compaction/compactor.rs:220`, the `OpaqueRef` arm in
`compaction/formatter.rs:307` calls `format_ccr_marker` (`:311`) with no flag
check, and the walker's `emit_opaque_ccr_marker` (`compaction/walker.rs:142-144`,
definition at `:171`) is likewise ungated.

The code documents this as the intended scope of the gate
(`headroom/transforms/smart_crusher.py:270-271`):

> Scope: gates ONLY the row-drop sentinel path. Stage-3c.2 opaque-string CCR
> substitutions still emit always […]

…and the `enable_ccr_marker` docstring itself repeats the limited scope
(`config.rs:68-82`). So this is by-design today — but it leaves no path to
lossless-only output.

### Acceptance criteria

- [ ] With `enable_ccr_marker=False`, no `<<ccr:...>>` marker appears in output
      from the **opaque-blob path** (it is already suppressed for the row-drop
      path). The repro above passes.
- [ ] Long string cells (> `opaque_min_bytes`) are preserved inline when
      `enable_ccr_marker=False`, not substituted.
- [ ] A separate `lossless_only` mode exists and is documented; output is
      byte-recoverable with zero markers from **either** path.
- [ ] Existing default behavior (markers on) is unchanged; covered by a
      regression test mirroring `enable_ccr_marker_true_is_default_behavior`.

## Additional Context

- NousResearch/hermes-agent#47866 — independent eval; the ~0.34% / 2.5% figures
  trace to the asymmetry above.
- Supersedes the `headroom wrap hermes` framing in #712: the actionable upstream
  change isn't a Hermes-specific wrapper, it's a lossless-only contract usable by
  any agent that re-reads its tool outputs.
- A downstream consumer ([NousResearch/hermes-agent#39691](https://github.com/NousResearch/hermes-agent/issues/39691), comment thread) has articulated demand for exactly this contract — a "headroom_lossless" mode that rejects CCR/retrieval-marker outputs. Without it, conservative integrators have no usable path.

nujovich:
## Description

`enable_ccr_marker` only gated the **row-drop sentinel** path. The **opaque-blob** path still emitted `<<ccr:HASH,kind,size>>` markers unconditionally whenever a string cell exceeded `opaque_min_bytes` (256), so **no configuration could produce a fully marker-free prompt**. Any `<<ccr:>>` marker is a promise that the full payload lives in the CCR store and must be fetched back via a retrieval tool call — there was no way to get compression without that round-trip dependency.

**Rebased on #1130 (merged).** That PR fixed the opaque-blob gate at the classifier (`ClassifyConfig.emit_opaque_markers`, driven by `enable_ccr_marker`) and closed #1091. This branch originally carried its own equivalent gating commit; that commit is now **redundant and has been dropped** — `classifier.rs` here is identical to upstream. What remains is the **net-new** work that is **not** in #1130:

- **Strict `lossless_only` mode** — keeps lossless tabular compaction, but routes every path that would need a CCR marker (row-drop sentinel **and** opaque-blob offload) to leave content uncompacted instead, so output is always marker-free **and** byte-recoverable.
- **Python parity** — `lossless_only` exposed across both config dataclasses, a `SmartCrusher` kwarg, and a per-call `crush(..., lossless_only=)` override.
- **`HEADROOM_LOSSLESS_ONLY` env var** — wires the mode through to the proxy runtime so real agents can use it.

The #1130 opaque gate is consumed here through a single centralized helper (`opaque_markers_enabled() = enable_ccr_marker && !lossless_only`) used by **all four** `ClassifyConfig` construction sites.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- **`feat(smart_crusher)`** — Add `lossless_only` (default `false`): keeps lossless tabular compaction but routes every marker-requiring path (row-drop sentinel + opaque-blob offload) to leave content uncompacted instead. Exposed across the Rust core, PyO3 bridge, both Python config dataclasses, a `SmartCrusher` kwarg, a per-call `crush(..., lossless_only=)` override, and `smart_crush_tool_output`. Includes a `debug_assert` documenting the load-bearing invariant (a `lossless_only` crusher must never reach the CCR store write).
- **`refactor(smart_crusher)`** — Extract `SmartCrusherConfig::opaque_markers_enabled()` as the single source of truth for `enable_ccr_marker && !lossless_only`, consumed by **all four** `ClassifyConfig` sites: the compaction-stage builder, `with_compaction_format`, the top-level `process_string` path (Rust core), and the PyO3 `compact_document_json` document-compactor path. No site derives the gate inline anymore, so they cannot drift.
- **`feat(proxy)`** — Expose the mode via `HEADROOM_LOSSLESS_ONLY`: `ContentRouterConfig.smart_crusher_lossless_only` → `_get_smart_crusher`; the proxy reads the env var and sets it on the live router config. Previously reachable only via the Python API, never through the proxy.

## Testing

- [ ] Type checking passes (`mypy headroom`) — not run (see Additional Notes)
- [x] Manual testing performed (proxy env-var seam, end-to-end — see Real Behavior Proof)

### Test Output

```text
### RUST  (cargo test -p headroom-core --lib smart_crusher)
test result: ok. 323 passed; 0 failed; 0 ignored; 0 measured; 516 filtered out

### PYTEST  (test_smart_crusher_bugs.py + test_agent_savings.py + test_smart_crusher_toin_attachment.py)
45 passed

### RUFF  (changed files)
All checks passed!

### FMT + CLIPPY  (cargo fmt --check && cargo clippy --workspace --lib)
clean — no warnings
```

New/affected tests: `enable_ccr_marker_false_suppresses_opaque_markers`, `lossless_only_leaves_array_uncompacted_instead_of_dropping`, `lossless_only_inlines_opaque_blobs_when_table_ships`, `lossless_only_never_writes_to_ccr_store` (Rust); `TestLosslessOnlyMode`, `test_router_lossless_only_flag_reaches_crusher`, `test_router_lossless_only_defaults_off` (Python). Coexists green with #1130's `long_string_stays_scalar_when_opaque_markers_disabled` (Rust) and `test_smart_crusher_toin_attachment.py` (Python). The Python `TestOpaqueMarkerGate` from the dropped gating commit was removed as redundant with #1130's coverage.

### Proxy env-var seam — end-to-end (this revision)

The one path with no automated coverage was `server.py` reading `HEADROOM_LOSSLESS_ONLY` from the environment and threading it into the live router config. Verified end-to-end by instantiating the **real** `HeadroomProxy`, pulling the `ContentRouter` out of its pipeline, and crushing a 50-row array with >256B opaque cells through the real Rust crusher:

| | `HEADROOM_LOSSLESS_ONLY=1` | env unset (default) |
|---|---|---|
| `crusher._lossless_only` | **True** | **False** |
| output contains `<<ccr:` | **No** (marker-free) | Yes (normal lossy) |
| byte-recoverable (round-trips to original JSON) | **Yes** | No (rows offloaded) |

This confirms the full chain `os.environ["HEADROOM_LOSSLESS_ONLY"]` → `server.py` → `ContentRouterConfig.smart_crusher_lossless_only` → `content_router.py` → `crusher_config.lossless_only` → Rust crusher. The default column proves strict mode genuinely changes behavior (not a no-op) and that the default path is unchanged.

### Prior live-traffic run

- Environment: Headroom proxy in front of a real agent (Hermes) routed to NVIDIA NIM (OpenAI-compatible upstream). Isolated config dir; `OPENAI_TARGET_API_URL=https://integrate.api.nvidia.com/v1`, `HEADROOM_LOSSLESS_ONLY=1`. Confirmed with `ss` that all LLM traffic flowed agent → proxy → upstream with no direct bypass.
- Exact command / steps: Start the proxy with `python -m headroom.proxy.server --host 127.0.0.1 --port 8787`; point the agent's LLM base_url at `http://127.0.0.1:8787/v1`; run a real chat plus a `search_files`-style task; read `/stats` and `/v1/retrieve/stats`; then toggle `HEADROOM_LOSSLESS_ONLY` and repeat for the comparison.
- Observed result: With 150K+ tokens of real traffic processed, `lossless_only` kept the CCR store empty (`entry_count: 0`) and emitted zero markers. A synthetic before/after with opaque (>256B) cells produced 12 `<<ccr:>>` markers in default mode and 0 under `lossless_only`, with output round-tripping to the original JSON structure.
- Not tested: A live `lossless_only`-vs-markers contrast on real agent traffic. The SmartCrusher offload path never engaged on this agent's tool outputs (`total_compressions: 0`; CCR store stayed at `entry_count: 0` even after a broad codebase search), and compression stayed marginal (~0.2–0.4%) in both modes. The agent's tool results don't match the crushable-array profile the offload paths target, so the marker path is never exercised in that integration. Why SmartCrusher barely engages with this agent's outputs is a separate integration question (output format / routing / size thresholds), out of scope for this change.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A (config docstrings updated in-tree; no separate docs)
- [ ] I have updated the CHANGELOG.md if applicable — N/A

## Additional Notes

- Rebased on top of merged #1130; the now-redundant opaque-blob gating commit was dropped, so this PR is purely the `lossless_only` feature + proxy wiring on top of #1130's gate.
- `mypy headroom` was not run in this environment; happy to add the result if CI requires it.
- Default behavior is fully preserved: `enable_ccr_marker` defaults to `true`, `lossless_only` defaults to `false`, and `HEADROOM_LOSSLESS_ONLY` unset is a no-op.


JerrettDavis (approved):
Reviewed the marker-free/lossless_only SmartCrusher work. The opaque-marker gate is centralized, strict mode avoids row-drop and CCR-store writes, Python/PyO3/proxy configuration paths are wired, and regression tests cover marker suppression, per-call override, and router propagation. CI is green.

JerrettDavis (approved):

JerrettDavis (approved):

nujovich:
### CI status: only failure is a known-flaky upstream perf test

After merging latest `main`, all checks pass **except `test (1)`**, which failed on a
single timing-sensitive test — unrelated to this PR's changes:

`tests/test_codex_ws_compression_scheduler.py::test_concurrent_compression_has_no_semaphore_tail`

**Why this isn't caused by this PR:**

1. This PR's commits never touch that file — it came in via the merge from `main`
   (`git log <merge-base>..<pr-head> -- tests/test_codex_ws_compression_scheduler.py`
   returns nothing).
2. It's a stress/latency test that the repo has repeatedly tried to de-flake on shared
   runners — e.g. `e94a36cb test(codex): de-flake semaphore-tail ratio check on fast runners`,
   `c22342e1 fix(tests): make stress test CI-robust`, `f82d61b1 fix(tests): make codex
   scheduler stress test machine-independent`.
3. It failed by ~1 ms over the noise floor: assertion is `ratio < 4.0 OR p99 < 25 ms`;
   the run measured `p99 = 29 ms` (`ratio = 22.2x`, `p50 = 1 ms`) — i.e. runner CPU
   contention, not a regression.

**Everything else is green:** `lint`, `test (2)`, `test (3)`, `test (4)`, `test (ubuntu)`.
The earlier `test (2)/(3)` failures were a HuggingFace model-cache timing issue on
cross-repo PR runs (the shared `Linux-models-allMiniLM-v2` cache wasn't warm yet); they
now pass.

Could you please **re-run the failed `test (1)` job**? No code change is needed — the test
should pass on a less-contended runner. Happy to rebase again if you'd prefer.

PR #1139, opened 2026-06-18
feat(proxy): HEADROOM_ORT_EP for OpenVINO/CUDA EP selection; fix Phase E skips when auth-mode-policy-enforcement disabled

RubenAAA:
## Description

Two related changes to the Rust proxy:

**1. `HEADROOM_ORT_EP` — runtime execution provider selection for ONNX Runtime**

Adds `init_ort_ep()` to `headroom-core`, called once at proxy startup before any ONNX session is created. The fastembed `EmbeddingScorer` and magika `Session` share the same ORT singleton, so a single env var covers both.

| Variable | Values | Default |
|---|---|---|
| `HEADROOM_ORT_EP` | `cpu` / `openvino` / `cuda` | `cpu` (no-op) |
| `HEADROOM_ORT_OPENVINO_DEVICE` | `NPU`, `CPU`, `GPU`, `GPU.0`, `HETERO:NPU,GPU` | `NPU` |
| `HEADROOM_ORT_OPENVINO_CACHE` | path to blob cache dir | unset |

ort 2.x uses `libloading` for dynamic EP dispatch — all EP types are always compiled in; no compile-time feature flags are needed. The direct `ort` dep is pinned to `=2.0.0-rc.12` to unify with the singleton already in the tree via fastembed/magika. Failures are non-fatal: a `WARN` is logged and ORT falls back to CPU automatically.

Windows note: `fastembed` on Windows uses `ort-load-dynamic` (already the case in `Cargo.toml`) rather than `ort-download-binaries-*` to avoid DirectML link-time deps. OpenVINO is a Windows-native runtime; setting `HEADROOM_ORT_EP=openvino` on Windows targets the Intel NPU via the installed OpenVINO redistributable.

**2. Fix Phase E passes skipped for non-PAYG auth with `--auth-mode-policy-enforcement disabled`**

`CompressionPolicy` at request entry was already forced to PAYG when `--auth-mode-policy-enforcement disabled` is set, but `compress_anthropic_request` still received the raw classifier output. This caused the E1 tool-array sort, E2 schema-key sort, and E3 `cache_control` auto-placement guards to keep skipping even though the operator explicitly opted into the PAYG pipeline.

Token reduction matters for subscription users too — rate limits are token-based, context windows are finite, and smaller payloads transfer faster. The billing-only framing of the original guard was too narrow.

Fix: derive `effective_auth_mode` at the call site using the same enforcement-flag override already applied to `CompressionPolicy`, and pass that into `compress_anthropic_request` instead of the raw classifier result.

Closes #

## Type of Change

## Changes Made

- `crates/headroom-core/src/lib.rs` — `pub fn init_ort_ep()`: reads `HEADROOM_ORT_EP`, initialises OpenVINO or CUDA EP via `ort::init().with_execution_providers([...]).commit()`
- `crates/headroom-core/Cargo.toml` — direct `ort = "=2.0.0-rc.12"` dep (pinned, `default-features = false`) placed before any `[target.cfg(...)]` section to avoid TOML inheritance
- `crates/headroom-proxy/src/main.rs` — call `headroom_core::init_ort_ep()` after `init_tracing` and before `AppState::new`
- `crates/headroom-proxy/src/proxy.rs` — derive `effective_auth_mode` from the enforcement-flag override at the `compress_anthropic_request` call site

## Testing

No new unit tests for the EP init path: the function is a startup side-effect with no return value and relies on ORT's own runtime presence checks. The auth-mode fix is covered by the existing `pr_e3_subscription_skips_auto_placement` and `e1_passes_through_when_subscription` tests, which assert the per-mode guard behaviour; `--auth-mode-policy-enforcement disabled` sits one layer above and is exercised end-to-end below.

### Test Output

```
cargo build --release -p headroom-proxy --target x86_64-pc-windows-gnu
Finished `release` profile [optimized] target(s) in 33.50s
```

## Real Behavior Proof

**Environment:** Windows 11, Intel Core Ultra 9 285HX (Lunar Lake NPU), OpenVINO 2025.x installed, `headroom-proxy.exe` cross-compiled from WSL2 (x86_64-pc-windows-gnu, Rust 1.95.0).

**Command:**
```
headroom-proxy.exe \
  --upstream https://api.anthropic.com \
  --compression \
  --compression-mode live_zone \
  --auth-mode-policy-enforcement disabled \
  --cache-control-auto-frozen disabled
```
Plus `HEADROOM_ORT_EP=openvino` in the environment.

**Observed result — startup:**
```json
{"level":"INFO","message":"ORT execution provider: OpenVINO","ep":"openvino","device":"NPU"}
{"level":"INFO","message":"headroom-proxy starting","listen":"0.0.0.0:8787",...}
```

**Observed result — per request (Phase E now fires on subscription auth):**
```json
{"level":"INFO","message":"tool-array sort applied: tools reordered alphabetically by name","event":"e1_applied","tool_count":121}
{"level":"INFO","message":"schema-key sort applied: input_schema keys rewritten in alphabetic order","event":"e2_applied","tool_count":121}
{"level":"INFO","message":"customer-placed cache_control marker(s) present; auto-placement skipped","event":"e3_skipped","reason":"marker_present"}
{"level":"INFO","message":"anthropic live-zone dispatch","frozen_message_count":0,"messages_total":13,"latest_user_message_index":"Some(12)","live_zone_blocks":1}
{"level":"INFO","message":"compression applied","strategies":"[\"tool_array_sort\",\"schema_key_sort\"]"}
```

Before this fix with `--auth-mode-policy-enforcement disabled`, every request produced `e1_skipped / e2_skipped / e3_skipped` with `reason: auth_mode` and `live_zone_blocks: 0`.

**What was not tested:** CUDA EP path (no NVIDIA GPU available); Linux OpenVINO EP (NPU driver is Windows-only); behaviour under `--auth-mode-policy-enforcement enabled` (unchanged code path, no regression expected).

## Review Readiness

## Checklist

---

## Update — NPU enablement (real-hardware validated, Intel AI Boost)

Pushed one commit (`fix(openvino): disable EP dynamic shapes (NPU compile) + correct EP-load docs`) after validating this EP path end-to-end on an Intel NPU running a live workload:

- **`with_dynamic_shapes(false)`** — the OVEP defaults to `disable_dynamic_shapes=false`, which forces the NPU graph compiler into dynamic-shape mode and **hangs for minutes**. Direct OpenVINO compiles the same *static* graph in ~13s; with this flag the proxy's NPU compile completes (~13–20s) and caches a blob. Callers must feed fixed-shape inputs (the Kompress static-shape model in #1153 pads each chunk to a fixed length).
- **Doc correction in `init_ort_ep`** — the previous comment claimed "all EP types are always compiled in … falls back to CPU automatically." That's **wrong for OpenVINO**: the OVEP is not in a stock `onnxruntime`. On Windows `ort-load-dynamic` silently loads `C:\Windows\System32\onnxruntime.dll` (no OVEP) and this call logs a misleading success while every session runs on CPU. The fix is to point `ORT_DYLIB_PATH` at an OpenVINO-enabled `onnxruntime` and put a **version-matched** `openvino` on `PATH` (mismatched versions → `Error 127: procedure not found`). The doc now states this.

Verified: `ORT execution provider: OpenVINO device:NPU`, no `Error 127`, NPU compile blob written, Kompress prose compression running on the NPU in a live session.


RubenAAA:
Pushed one commit since your last look: `with_dynamic_shapes(false)` so the OpenVINO **NPU** actually compiles (it defaults to dynamic-shape compilation and hangs for minutes; static compiles in ~13s), plus a doc correction — the OVEP isn't in a stock onnxruntime, and on Windows `ort-load-dynamic` silently grabs System32's onnxruntime.dll (no OVEP) and runs on CPU while logging success. Validated end-to-end on an Intel AI Boost NPU in a live session. Details in the updated description.

JerrettDavis (approved):
Reviewed the ORT execution-provider and Phase E fixes. Startup initializes the ORT EP before sessions are created, OpenVINO device/cache settings are documented in code, the auth-mode override is applied at the compression call site, and the volatile-detector schema false positive is covered. CI is green.

JerrettDavis (approved):
Re-reviewed the current head after syncing with `main`. I pushed a mechanical merge commit (`e6f4e226`) that only resolved the current `CHANGELOG.md` conflict by keeping both the PR's ORT/Phase-E entries and the newer `main` entries.

The Rust changes still look sound: ORT EP initialization happens before proxy state/session creation, OpenVINO dynamic-shape disabling is scoped to the OpenVINO builder, the auth-mode override now matches the `CompressionPolicy` override at the compression call site, and the schema-definition volatile-detector exception is still bounded to `input_schema` / `function.parameters` paths.

Local verification after the merge: `cargo check -p headroom-proxy --quiet` passed. I also tried the named volatile-detector test filters, but Cargo matched zero tests from the workspace invocation; those commands still compiled cleanly. GitHub checks are rerunning on the pushed merge commit.

JerrettDavis (approved):
Approval refreshed after merging current main. The ORT execution-provider selection remains scoped to the Rust/proxy path, and the conflict resolution was limited to the changelog/main drift. Focused Rust verification passed with cargo fmt --check and cargo check -p headroom-core -p headroom-proxy.

JerrettDavis (approved):
Approved after maintainer refresh onto current main.

Verified locally on head 4ca269f:
- cargo fmt --check
- cargo check -p headroom-core -p headroom-proxy --quiet
- cargo check -p headroom-core --no-default-features --quiet
- git diff --check origin/main...HEAD
- merge commit hook: conflict scan and plugin version sync

Resolved the changelog conflict by keeping current main's release-managed changelog. I also reconciled the ORT dependency with main's optional-ML/dynamic-loading setup: there is now one pinned optional `ort` dependency with `load-dynamic`, and `init_ort_ep()` is gated so `headroom-core --no-default-features` still builds.

RubenAAA:
### Synced onto current `main`, and one addition from the `ort` finding

This branch was 80 commits behind and had only ever run `template` and `label` — the same fork-workflow-approval situation described on #1153/#1154, except no sync had happened here yet. Merged current `main` (clean, no conflicts) and pushed. New head: `db77df29`. Real CI needs a maintainer approval on the pending runs.

**One content change**, and it comes straight out of the `ort` deadlock write-up on #1153: the `HEADROOM_ORT_EP` doc comment now states the ONNX Runtime `>= 1.24` floor and *why the symptom is silent* — the version error is built inside the `Once` that `setup_api()` is initialising, so the re-entry blocks instead of surfacing. This function is where EP selection is configured, so it's the place someone looks when a session never returns.

**Worth adding to that finding: the deadlock is not limited to old runtimes.** Reproduced here on a box with no resolvable `libonnxruntime.so` at all — `kompress_parity` parked at 0.0% CPU for 12 minutes, all three threads in `futex_wait_queue`, and `/proc/<pid>/maps` showed no onnxruntime mapped. Same re-entrant `Once`, different trigger: any failure inside `load_dylib_from_path` hangs, not just a version mismatch. #2591 raises the floors, which fixes the version case; the missing-library case would still hang. Not something this PR can fix, but worth knowing before the model-backed job runs anywhere.

Verified on the merged head:
- `cargo fmt --check`
- `cargo check -p headroom-core -p headroom-proxy`
- `cargo check -p headroom-core --no-default-features` (the pinned optional `ort` + `load-dynamic` reconciliation holds)

The dependency reconciliation from the maintainer refresh survived the merge untouched — one pinned optional `ort` entry with `load-dynamic`, `init_ort_ep()` still behind `#[cfg(feature = "ml")]`.

JerrettDavis (changes requested):
The new ORT-floor documentation commit is fine in isolation, but the PR is currently `mergeStateStatus=DIRTY` after the latest sync, and only the governance checks have run on this head. I can't refresh approval while GitHub reports conflicts and the actual CI suite has not run.

Please update the branch against current `main` and clear the dirty state so the Rust/CI checks can run on the resolved head. Once that is clean, the scope to re-check should be small: the existing EP initialization/auth-mode behavior plus the new ONNX Runtime >=1.24 documentation note.

JerrettDavis (approved):
Re-reviewed current head 980c669. The previous blocker was GitHub state rather than a code issue; this branch is now clean with a full green check set, and the ORT/Phase-E focused checks still pass locally.

Validated locally:

- `cargo fmt --check` -> passed
- `cargo check -p headroom-core -p headroom-proxy --quiet` -> passed
- `cargo check -p headroom-core --no-default-features --quiet` -> passed
- `cargo test -p headroom-proxy --test integration_schema_sort --quiet` -> 3 passed
- `cargo test -p headroom-proxy --test integration_tool_sort --quiet` -> 4 passed
- `cargo test -p headroom-proxy --test integration_e3_anthropic_cache_control --quiet` -> 6 passed

GitHub checks are green on this head as well.

PR #1148, opened 2026-06-19
Fix UnboundLocalError: bind ccr_workspace_key before CCR-inject block

### LINKED ISSUE #1145: proxy: UnboundLocalError 'ccr_workspace_key' crashes every request when CCR injection is off (e.g. --no-ccr-inject-tool)
### Summary
With the default config plus `--no-ccr-inject-tool`, **every** `POST /v1/messages` crashes with `UnboundLocalError: cannot access local variable 'ccr_workspace_key'` and returns HTTP 500. `--no-ccr-inject-tool` is suggested in `--help` for clients that would otherwise error on the injected tool, so this is an easy footgun.

### Root cause (`headroom/proxy/handlers/anthropic.py`, v0.26.0)
`ccr_workspace_key` / `ccr_workspace_label` are bound **only inside** the CCR-injection block:
```python
if (self.config.ccr_inject_tool or self.config.ccr_inject_system_instructions) and not _bypass:   # L1302
    ...
    ccr_workspace_key, ccr_workspace_label = self._resolve_ccr_workspace(request, body)            # L1360
```
...but read **after** that block, unconditionally:
```python
if (self.ccr_context_tracker
        and self.config.ccr_proactive_expansion        # default True
        and ccr_workspace_key):                        # L1397  <-- UnboundLocalError
```
`ccr_inject_system_instructions` defaults to `False` (`proxy/models.py:134`) and has no CLI flag, so in the stock config the L1302 guard is held open *only* by `ccr_inject_tool=True`. Passing `--no-ccr-inject-tool` flips it to `False`; both sides of the `or` are now `False`, the block is skipped, and `ccr_workspace_key` is never bound -- yet L1394's guard (with `ccr_proactive_expansion` still defaulting `True`) reads it. The same skip happens for any bypassed request (`x-headroom-bypass: true`).

### Repro
```bash
pip install headroom-ai==0.26.0
headroom proxy --no-ccr-inject-tool --port 8787 --anthropic-api-url <upstream>
curl -s localhost:8787/v1/messages -H 'content-type: application/json' \
  -d '{"model":"claude-sonnet-4-5","max_tokens":16,"messages":[{"role":"user","content":"hi"}]}'
# -> HTTP 500; server log shows UnboundLocalError: ccr_workspace_key
```

### Fix
Initialize before the L1302 block so the name is always bound:
```python
ccr_workspace_key = ccr_workspace_label = None
```
One-line PR incoming. (`proxy/handlers/openai.py:1932` and `batch.py` use the same `or` guard -- worth a glance for the same pattern.)

### Env
headroom-ai 0.26.0 - Python 3.12 - Linux (Docker `python:3.12-slim`).

_Split out from #1144 so it can be tracked/closed independently._

xspawnnn:
## Summary

`headroom proxy --no-ccr-inject-tool` (or any config where both `ccr_inject_tool` and `ccr_inject_system_instructions` are off) returns **HTTP 500 on every request**:

```
UnboundLocalError: cannot access local variable 'ccr_workspace_key' where it is not associated with a value
```

### Root cause
In `headroom/proxy/handlers/anthropic.py`, `ccr_workspace_key` / `ccr_workspace_label` are assigned **inside** the CCR-injection block (~L1360):
```python
if (self.config.ccr_inject_tool or self.config.ccr_inject_system_instructions) and not _bypass:
    ...
    ccr_workspace_key, ccr_workspace_label = self._resolve_ccr_workspace(request, body)
```
…but read **outside** it, in the proactive-expansion guard (~L1394, `ccr_proactive_expansion` is on by default):
```python
if (self.ccr_context_tracker and self.config.ccr_proactive_expansion and ccr_workspace_key):
```
When the injection block is skipped, the name is never bound → `UnboundLocalError` → 500. `--no-ccr-inject-tool` is suggested in `--help`, so it is a documented path users hit.

### Fix
Bind both names to `None` before the block so they are always defined (one line, no behavior change on the existing path).

## Description

One-line fix for the crash reported in #1145.

Closes #1145

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: initialize `ccr_workspace_key = ccr_workspace_label = None` immediately before the CCR-injection `if` block, so the later proactive-expansion guard never reads an unbound local.

## Testing

### Test Output

```text
Repro (pre-fix) — headroom-ai 0.26.0, Docker python:3.12-slim, mock upstream on 127.0.0.1:9999:
  $ headroom proxy --port 8787 --no-ccr-inject-tool --anthropic-api-url http://127.0.0.1:9999
  $ curl -s -o /dev/null -w '%{http_code}\n' -X POST localhost:8787/v1/messages -d '{"model":"claude-3","messages":[{"role":"user","content":"hi"}]}'
  500
  # server log: UnboundLocalError: cannot access local variable 'ccr_workspace_key' ...
With this patch the variable is always bound, so the guard evaluates normally instead of raising.
```

## Real Behavior Proof

- Environment: headroom-ai 0.26.0, Docker `python:3.12-slim`, `pip install 'headroom-ai[proxy]'`, mock Anthropic upstream on 127.0.0.1:9999.
- Exact command / steps: `headroom proxy --port 8787 --no-ccr-inject-tool --anthropic-api-url http://127.0.0.1:9999`, then any `POST /v1/messages`.
- Observed result: pre-fix → HTTP 500 `UnboundLocalError: ccr_workspace_key` on every request (full write-up in #1145).
- Not tested: I did not run the repo CI (`pytest`/`ruff`/`mypy`) for this single-line, web-editor change.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -1299,6 +1299,7 @@ class AnthropicHandlerMixin:
                         )
                 except Exception:  # advisory hint only — must never fail a request
                     pass
+            ccr_workspace_key = ccr_workspace_label = None
             if (
                 self.config.ccr_inject_tool or self.config.ccr_inject_system_instructions
             ) and not _bypass:


chopratejas:
Given the quick fix - I will just merge it - even though the PR does not adhere to the PR template and linting

PR #1153, opened 2026-06-19
feat(rust): port Kompress ML prose compressor to Rust (parity-only, 1/3)

RubenAAA:
## Description

Ports the **Kompress** ML prose compressor to the Rust `headroom-core` engine — a byte-for-byte port of `headroom/transforms/kompress_compressor.py` (`kompress-v2-base` ONNX + ModernBERT tokenizer). This is **PR 1 of 3** splitting #1143 into reviewable pieces: this PR adds the engine + the parity harness only; it does **not** wire Kompress into the live-zone dispatcher (that is PR 3). The engine is therefore additive and inert until dispatch lands.

The port is verified against the Python reference via the `headroom-parity` harness (replays recorded fixtures and asserts identical output).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom-core`: `transforms/kompress.rs` — full Kompress engine (ModernBERT tokenizer + ONNX inference session via the shared `ort` runtime), plus `Kompress::from_cache` cache-only constructor and `hf_cache_file` resolver. Cache-only: a cold cache yields `None` and never downloads (mirrors the Python `allow_download=False` path).
- `headroom-core`: `transforms/mod.rs` — exports the new module.
- `headroom-parity`: `KompressComparator` + recorder (`scripts/record_kompress_fixtures.py`) + **21** recorded fixtures under `tests/parity/fixtures/kompress/`.
- `Cargo.toml`: adds `ort` as a direct dependency, pinned (`=2.0.0-rc.12`) to the existing lock entry so Cargo unifies onto the single ONNX Runtime already in the tree via `fastembed`/`magika` (no second runtime, no extra binary download).

## Testing

- [x] Unit tests pass (`pytest`) — N/A for runtime (Rust change); dev-time fixture tooling under `scripts/`/`tests/parity/` passes `ruff check` + `ruff format`

### Test Output

```text
cargo run -p headroom-parity -- run --only kompress
[kompress] total=21 matched=21 skipped=0 diffed=0

# Rust suites
cargo test -p headroom-core   ->  911 passed, 3 ignored (13 suites)
cargo clippy -p headroom-core -p headroom-parity -- -D warnings  ->  No issues found
cargo fmt --check  ->  clean
```

## Real Behavior Proof

- Environment: Linux (WSL2, kernel 6.6), Rust stable toolchain, Python 3.13 for reference recording; `kompress-v2-base` + `ModernBERT-base` present in the local HF cache.
- Exact command / steps: `cargo run -p headroom-parity -- run --only kompress`; then `cargo test -p headroom-core`.
- Observed result: 21/21 kompress fixtures match the Python reference byte-for-byte; `cargo test -p headroom-core` reports 911 passed, 3 ignored; clippy and fmt clean.
- Not tested: live-zone dispatch routing (intentionally not wired here — lands in PR 3); non-ASCII identifiers/content (documented out-of-parity-scope); alternate ONNX execution providers (default CPU EP here).

## Review Readiness

## Additional Notes

- Split of #1143 — PR 1 of 3. PR 2 ports the CodeCompressor; PR 3 wires both into the live-zone dispatcher and adds the `--enable-kompress` gate. PRs 2 and 3 are stacked on this branch, so until this merges their diffs against `main` show this PR's content as well.
- `pytest`/`mypy` checklist items are N/A — this is a Rust change; the equivalent gates are `cargo test` + `cargo clippy`, run above.

---

## Update — Windows/NPU engine support (one commit added)

Pushed `feat(kompress): Windows cache resolution, load diagnostics, static-shape ONNX` — three self-contained additions to the cache-only loader, all in `kompress.rs`, validated against real NPU hardware downstream:

- **Cross-platform HF cache resolution** — resolve the model via `HF_HUB_CACHE` / `HF_HOME` / `HOME` / `USERPROFILE` (was `$HOME`-only), so `from_cache` finds the model on native Windows, not just Linux.
- **Load diagnostics** — `from_cache` now logs *why* it defers (`kompress_cache_miss` / `kompress_session_build_failed`, with the searched roots + the real session-build error) instead of returning `None` silently. This was the #1 cause of an unexplained `kompress_ready=false` (e.g. HF cache symlinks being unreadable over a `\\wsl$` mount, or an EP rejecting the graph).
- **Static-shape ONNX support** — detect a fixed `input_ids` seq dimension on the loaded model (`detect_static_seq`) and right-pad each chunk to it (masked padding ⇒ identical scores); dynamic models keep their natural length at **zero padding cost**. Enables execution providers that can't compile dynamic shapes (OpenVINO NPU, see #1139).

**Parity: 21/21 in both static and dynamic modes.** `cargo fmt`/`clippy`/`ruff` clean, `cargo test -p headroom-core` green.

> Note: the stack was rebased — `from_cache`/`hf_cache_file` (previously introduced in the wiring PR #1155) now live here, in the engine, which is their correct home.


RubenAAA:
Added one engine commit (Windows HF-cache resolution + load diagnostics + static-shape ONNX support) — all self-contained in `kompress.rs`, **parity 21/21 in both static and dynamic modes**. Heads-up: I rebased the stack, so `from_cache`/`hf_cache_file` (previously in the wiring PR #1155) now live here in the engine, which is their correct home. Description updated with the rationale.

JerrettDavis (changes requested):

JerrettDavis (approved):
Rechecked after the latest commit. The prior review was blocking on stale/unstable PR state rather than a code issue; the current rollup does not show completed failing checks, and this remains a parity-gated Rust Kompress port rather than live behavior wiring.

The cache-only/model-gated tests are an acceptable shape for this stage: CI can still build the code path, while local/preloaded environments exercise byte parity against the Python fixtures. Looks ready from my side.

JerrettDavis (approved):
Re-reviewed after updating from `main` and fixing the merged manifest duplicate that prevented Cargo from loading the workspace.

Validated locally:
- `git diff --check`
- `cargo test -p headroom-core --test kompress_parity`
- `uv run --frozen --extra dev ruff check scripts/record_kompress_fixtures.py tests/parity/recorder.py`
- `uv run --frozen --extra dev ruff format --check scripts/record_kompress_fixtures.py tests/parity/recorder.py`

chopratejas:

Merged current `main` (`aebe1953`) into `stack/1-kompress-engine` and approved the pending workflow runs. New head: `6d5f03b6`.

chopratejas:
### Resynced, and both ports now verified end-to-end

Merged current `main` and resolved the one conflict, then got the parity harness running properly. Two follow-ups pushed to your branch — details below so nothing is a surprise.

**1. Conflict in `crates/headroom-parity/src/lib.rs`.** `main` picked up #2567 and #2568, which register a `text_crusher` comparator and promote `log_compressor` from a stub. Your branch adds its own comparators to the same `builtin_comparators()` vec, so git interleaved two structurally identical impls. Resolved by keeping every side — rebuilt from the merge stages rather than hand-splicing markers. No behaviour change to your code.

**2. `cargo check -p headroom-core --no-default-features` was failing** with 3 errors: `kompress.rs` uses `ort`, which is `optional = true` behind `ml`, but `pub mod kompress;` was unconditional. That breaks the static-musl path the `ml` feature exists to serve. Added `#[cfg(feature = "ml")]` to the module and its re-export, matching how `magika_detector` is already gated. CI would never have caught it — `cargo test --workspace` only builds default features.

**3. The result — your ports are byte-identical to Python:**

```
227 fixtures · 182 matched · 45 skipped (cache_aligner + ccr stubs) · 0 diffed · exit 0
```

Real ONNX inference, all 8 tree-sitter grammars. The exact-pinning rationale in your `Cargo.toml` holds up.

---

### Heads-up: `ort` deadlocks on ONNX Runtime < 1.24

Worth knowing if you run kompress locally, because the symptom is silent.

`ort-sys` computes `ORT_API_VERSION = 17 + one per enabled api-N feature`, and features are additive across the graph. `fastembed 5.17.3` enables `api-24`, so the effective requirement is **ONNX Runtime >= 1.24**. Below that, `ort` calls `Error::new(...)` from inside `load_dylib_from_path`, which runs inside the `Once` that `setup_api()` is initialising — constructing the error re-enters that same `Once`, and `std::sync::Once` blocks forever.

So instead of a version-mismatch error you get a process parked at **0% CPU indefinitely**. I lost an hour to it before sampling the stack. `ORT_DYLIB_PATH` makes no difference — I checked, since it is the obvious first guess.

With onnxruntime 1.21.1: `Session::builder()` never returns. With 1.24.4: returns in 1.2s.

The repo's floors all said `>=1.16.0`; #2591 raises them to `>=1.24`. It works today only because pip resolves to the newest wheel.

**This matters most for #1155**, which adds a `kompress-parity` job that deliberately warms the model into the HF cache — the exact trigger. #1153 and #1154 are safe in CI only because the model is absent there, so the comparator errors out before reaching `Session::builder()`. Worth landing #2591 first.

RubenAAA:
Thanks for the sync and for chasing the `ort` hang down to the `Once` — that one would have cost me the same hour.

Pulled both follow-ups. The `#[cfg(feature = "ml")]` gate on `pub mod kompress` is the right call and matches `magika_detector`; `cargo check -p headroom-core --no-default-features` is clean here. You're right that nothing in CI would have caught it — the fix belongs in the workflow, not in this PR, but it's worth noting that `cargo test --workspace` will keep missing this class of break for any future optional-ML module.

**The same gap was live in the dispatch wiring**, so thank you for finding the pattern: `live_zone.rs` imported `super::kompress::{Kompress, KompressConfig}` unconditionally, which broke `--no-default-features` on #1155 the moment your gate landed there. Fixed on that branch — the `PlainText` arm, the singleton, and `warm_live_zone_compressors` are now cfg'd, and the no-`ml` build degrades to passthrough, which is the same path a build takes when the model isn't cached. Details on #1155.

On the parity-lib conflict: rebuilding from the merge stages rather than splicing markers is exactly right, and the result reads correctly — `text_crusher`, `log_compressor`, and both new comparators all registered once in `builtin_comparators()`.

Agreed on landing #2591 first.

**One thing to add to the deadlock write-up:** it isn't only old runtimes. Reproduced here on a box with no resolvable `libonnxruntime.so` at all — `kompress_parity` sat at 0.0% CPU for 12 minutes, all threads in `futex_wait_queue`, no onnxruntime mapped in `/proc/<pid>/maps`. Same re-entrant `Once`; any failure inside `load_dylib_from_path` hangs, not just a version mismatch. So #2591 closes the version case but a runner that resolves no library at all would still burn its 30-minute timeout with nothing to show. Cheap insurance for whichever job ends up warming the model: assert the runtime is present and `>= 1.24` in a step before the tests, so it fails in seconds with a readable message.

I've also documented the floor and the silent-failure mechanism in the `HEADROOM_ORT_EP` docs on #1139, since that's where someone looks when a session never returns.

chopratejas:
Both of your points landed — thank you, the second one changed this PR.

**The missing-library case.** You're right and I had this wrong: I treated the deadlock as a version-mismatch bug, so #2591 only raised the floor. Any failure inside `load_dylib_from_path` re-enters the same `Once`, so a runner that resolves nothing hangs identically to one resolving 1.21 — and a pin cannot help there. Your `futex_wait_queue` / empty `/proc/<pid>/maps` detail is the part that makes it unambiguous.

Implemented your suggestion in #2591 (`7e17bba6`): a pre-flight in the existing dylib step that asserts, before any test runs, that onnxruntime is importable, that its minor is >= 24, and that a `libonnxruntime` object actually exists under `capi/`. Each failure exits 1 with an `::error::` annotation naming the cause. Verified all three branches by extracting the script from the workflow and running it against three interpreters:

```
onnxruntime absent (python3 -I -S)   rc=1  "onnxruntime is not importable"
onnxruntime 1.21.1 (.venv)           rc=1  "is too old: ort requires >= 1.24"
onnxruntime 1.24.4 (PYTHONPATH)      rc=0  emits ORT_DYLIB_PATH=...1.24.4.dylib
```

Seconds with a readable message instead of a 30-minute timeout with an empty log.

**The dispatch-wiring gap.** Good catch, and it is the more interesting half — my gate on `pub mod kompress` turned a latent unconditional import in `live_zone.rs` into a hard break, on a branch I was not touching. Degrading a no-`ml` build to plain-text passthrough is the right shape, since it matches the cold-cache path that already exists.

I have approved the 8 pending runs on #1155's current head (`8a426807`), so CI should be moving there now. Worth knowing for future pushes: approval is **per run, not per contributor**, so every push to a fork branch re-enters `action_required` and needs another pass. Also — I twice reported "nothing to approve" earlier because I was filtering the API on `status=action_required`, when those runs are `status=completed` with `conclusion=action_required`. My mistake, and it is why your confirmation that the previous head produced *no* runs at all was useful: it separates the workflow-file restriction from my bad query.

**On `hf_hub_roots()` walking to `$HOME`** even with `HF_HUB_CACHE` and `HF_HOME` set — I hit exactly this. I could only get a cold-cache run by overriding `HOME`, and I ended up using that as the stand-in for CI. Agreed it is a real divergence from `huggingface_hub` and it makes the model-gated tests awkward to exercise on any machine that has the model; worth tightening the resolver so the documented env vars are authoritative.

Thanks also for correcting the 973 → 1423 figure yourself. That partial count was the deadlock truncating the run, which is a good illustration of why the hang is worse than a failure: it produced a plausible-looking green number.

`pyproject.toml` is out of #2591 now, by the way — bumping the floor there makes `headroom-ai[all]` unsatisfiable via a pillow chain (`onnxruntime>=1.24` forces `pillow>=10.3.0,<12.0`, and `[all]` already wants `pillow>=12.3.0`). Not rapidocr, which allows `onnxruntime>=1.7.0` uncapped. So the user-facing hazard via `_ort.py` is still open and wants its own change; your #1139 docs are the right place for the mechanism in the meantime.

PR #1154, opened 2026-06-19
feat(rust): port CodeCompressor AST compressor to Rust (parity-only, 2/3)

RubenAAA:
## Description

Ports the **CodeCompressor** — an AST, syntax-preserving source-code compressor (tree-sitter) — to the Rust `headroom-core` engine, a byte-for-byte port of `headroom/transforms/code_compressor.py` (`CodeAwareCompressor`). This is **PR 2 of 3** splitting #1143. It adds the engine + parity fixtures only and does **not** wire it into the live-zone dispatcher (PR 3). The engine is additive and inert until dispatch lands.

**Stacked on PR 1 (Kompress).** Until PR 1 merges, this PR's diff against `main` also includes PR 1's content; review PR 1 first.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom-core`: `transforms/code_compressor.rs` — full CodeCompressor port: language detection (regex prefilter then fewest-errors tree-sitter), symbol-importance scoring (min-max normalized, CPython half-to-even rounding), per-function body-budget allocation, statement-level body truncation with `# [N lines omitted; calls: ...]` summaries, Python docstring handling, and a re-parse syntax-validity guard that returns the original on any invalid output.
- `headroom-core`: `transforms/mod.rs` — exports the new module.
- `headroom-parity`: `CodeCompressorComparator` + recorder (`scripts/record_code_compressor_fixtures.py`) + **30** recorded fixtures under `tests/parity/fixtures/code_aware_compressor/`.
- `Cargo.toml`: adds `tree-sitter` (0.25.2) + 8 grammar crates, each pinned (`=`) to the exact version of the Python `tree-sitter-*` wheels the fixtures were recorded against, so Rust and Python parsers emit node-for-node identical ASTs (the precondition for byte-parity).

## Testing

- [x] Unit tests pass (`pytest`) — N/A for runtime (Rust change); dev-time fixture tooling passes `ruff check` + `ruff format`

### Test Output

```text
cargo run -p headroom-parity -- run --only code_aware_compressor

# Grammar canary (precondition): node-type + line-span trees, Python vs pinned Rust crates
=> 9/9 samples across 8 languages produce byte-identical ASTs

# Rust suites
cargo test -p headroom-core   ->  917 passed, 3 ignored (14 suites)
cargo clippy -p headroom-core -p headroom-parity -- -D warnings  ->  No issues found
cargo fmt --check  ->  clean
```

## Real Behavior Proof

- Environment: Linux (WSL2, kernel 6.6), Rust stable toolchain, Python 3.13 for reference recording.
- Exact command / steps: grammar canary (dump + compare node-type/line-span trees, Python tree-sitter stack vs pinned Rust crates); then `cargo run -p headroom-parity -- run --only code_aware_compressor`; then `cargo test -p headroom-core`.
- Observed result: 9/9 canary samples produce byte-identical ASTs; 30/30 code fixtures match the Python reference byte-for-byte; `cargo test -p headroom-core` reports 917 passed, 3 ignored; clippy and fmt clean. Compressed output preserves signatures, imports, the first statement of each body, and inter-function call edges; output re-parses cleanly (the syntax-validity guard returns the original otherwise).
- Not tested: live-zone dispatch routing (intentionally not wired here — lands in PR 3); non-ASCII identifiers/string contents (documented out-of-parity-scope).

## Review Readiness

## Additional Notes

- Split of #1143 — PR 2 of 3, stacked on PR 1. PR 3 wires both engines into the live-zone dispatcher and adds the `--enable-kompress` gate.
- Grammar-version parity is the make-or-break invariant: same version number on crates.io + PyPI means the same `grammar.js` source, hence the same generated `parser.c`. Bumping any grammar pin requires re-running the canary and re-recording fixtures.
- `pytest`/`mypy` checklist items are N/A — Rust change; equivalent gates are `cargo test` + `cargo clippy`, run above.


JerrettDavis (changes requested):

JerrettDavis (approved):
Rechecked after the latest commit. The prior changes-requested review was for stale/unstable PR state; the current check rollup has no completed failures, and the Rust CodeCompressor addition is still parity-scoped with fixture coverage rather than changing the live compression dispatch.

I reviewed the grammar pinning, exported transform surface, and parity comparator/test shape. The staged approach is appropriate here; live wiring can be reviewed separately when it is introduced.

JerrettDavis (approved):
Re-reviewed after updating from `main` and fixing the merged manifest duplicate that prevented Cargo from loading the workspace.

Validated locally:
- `git diff --check`
- `cargo test -p headroom-core --test kompress_parity --test code_compressor_parity`
- `uv run --frozen --extra dev ruff check scripts/record_kompress_fixtures.py scripts/record_code_compressor_fixtures.py tests/parity/recorder.py`
- `uv run --frozen --extra dev ruff format --check scripts/record_kompress_fixtures.py scripts/record_code_compressor_fixtures.py tests/parity/recorder.py`

chopratejas:

Merged current `main` (`aebe1953`) into `stack/2-code-compressor` and approved the pending workflow runs. New head: `04d7b01b`.

chopratejas:
### Resynced, and both ports now verified end-to-end

Merged current `main` and resolved the one conflict, then got the parity harness running properly. Two follow-ups pushed to your branch — details below so nothing is a surprise.

**1. Conflict in `crates/headroom-parity/src/lib.rs`.** `main` picked up #2567 and #2568, which register a `text_crusher` comparator and promote `log_compressor` from a stub. Your branch adds its own comparators to the same `builtin_comparators()` vec, so git interleaved two structurally identical impls. Resolved by keeping every side — rebuilt from the merge stages rather than hand-splicing markers. No behaviour change to your code.

**2. `cargo check -p headroom-core --no-default-features` was failing** with 3 errors: `kompress.rs` uses `ort`, which is `optional = true` behind `ml`, but `pub mod kompress;` was unconditional. That breaks the static-musl path the `ml` feature exists to serve. Added `#[cfg(feature = "ml")]` to the module and its re-export, matching how `magika_detector` is already gated. CI would never have caught it — `cargo test --workspace` only builds default features.

**3. The result — your ports are byte-identical to Python:**

```
227 fixtures · 182 matched · 45 skipped (cache_aligner + ccr stubs) · 0 diffed · exit 0
```

Real ONNX inference, all 8 tree-sitter grammars. The exact-pinning rationale in your `Cargo.toml` holds up.

---

### Heads-up: `ort` deadlocks on ONNX Runtime < 1.24

Worth knowing if you run kompress locally, because the symptom is silent.

`ort-sys` computes `ORT_API_VERSION = 17 + one per enabled api-N feature`, and features are additive across the graph. `fastembed 5.17.3` enables `api-24`, so the effective requirement is **ONNX Runtime >= 1.24**. Below that, `ort` calls `Error::new(...)` from inside `load_dylib_from_path`, which runs inside the `Once` that `setup_api()` is initialising — constructing the error re-enters that same `Once`, and `std::sync::Once` blocks forever.

So instead of a version-mismatch error you get a process parked at **0% CPU indefinitely**. I lost an hour to it before sampling the stack. `ORT_DYLIB_PATH` makes no difference — I checked, since it is the obvious first guess.

With onnxruntime 1.21.1: `Session::builder()` never returns. With 1.24.4: returns in 1.2s.

The repo's floors all said `>=1.16.0`; #2591 raises them to `>=1.24`. It works today only because pip resolves to the newest wheel.

**This matters most for #1155**, which adds a `kompress-parity` job that deliberately warms the model into the HF cache — the exact trigger. #1153 and #1154 are safe in CI only because the model is absent there, so the comparator errors out before reaching `Session::builder()`. Worth landing #2591 first.

RubenAAA:
Thanks for the sync and for tracking the `ort` hang down to the re-entrant `Once`.

Pulled both follow-ups. The `#[cfg(feature = "ml")]` gate is the right shape and matches `magika_detector`; `cargo check -p headroom-core --no-default-features` is clean here. Also fixed one clippy nit of mine that `--all-targets` surfaces in this crate (`manual_range_contains` in the `detect_language_basic` test) — carried on the #1155 branch with the rest of the stack.

**The same gate was missing in the dispatch wiring**, which your fix exposed: `live_zone.rs` imported `super::kompress::*` unconditionally, so `--no-default-features` broke on #1155 as soon as the module became optional. Fixed there — singleton, warm-up, and the `PlainText` dispatch arm are all cfg'd now, and a no-`ml` build passes plain text through, the same path it takes with a cold model cache.

The parity-lib conflict resolution reads correctly — `text_crusher`, `log_compressor`, and both new comparators each registered once in `builtin_comparators()`.

**On the one red check here:** `test (1)` failed on `tests/test_codex_ws_compression_scheduler.py::test_concurrent_compression_has_no_semaphore_tail` — a wall-clock assertion that came in at p99=76ms against a 75ms noise floor. Runner timing, nothing to do with this port; the whole diff is Rust and parity fixtures. A rerun should clear it, though a 1ms margin on a shared runner will keep doing this.

Agreed on landing #2591 first. One addition to the deadlock write-up, posted in full on #1153: it isn't only old runtimes — a box that resolves *no* `libonnxruntime.so` hangs identically (0.0% CPU, all threads in `futex_wait_queue`, nothing onnx-shaped in `/proc/<pid>/maps`). Any failure inside `load_dylib_from_path` re-enters the `Once`. So whichever job warms the model wants a cheap pre-flight check on the runtime version, or it burns the full 30-minute timeout with no diagnostic.

PR #1155, opened 2026-06-19
feat(rust): wire CodeCompressor + Kompress into live-zone dispatch + gate (3/3)

RubenAAA:
## Description

Wires the two ported compressors (Kompress + CodeCompressor) into the live-zone dispatcher and adds the Kompress opt-in gate. This is **PR 3 of 3** splitting #1143 — the only behavior-changing piece. `dispatch_compressor` now routes `SourceCode -> CodeCompressor` and `PlainText -> Kompress`, filling the slots it had reserved with TODOs.

**Stacked on PR 1 (Kompress) and PR 2 (CodeCompressor).** Until those merge, this PR's diff against `main` also includes their content; review #1 and #2 first. Once they land, this PR shrinks to the dispatch wiring + flag (~250 lines, mostly the routing test).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom-core`: `transforms/live_zone.rs` — `dispatch_compressor` routes `SourceCode -> CodeCompressor` (always-on; grammars are statically linked, constructs in microseconds with no I/O) and `PlainText -> Kompress` (cache-only, gated). Adds `set_kompress_enabled` + `warm_live_zone_compressors`.
- `headroom-core`: `transforms/kompress.rs` — small additions to support the dispatcher's cache-only path.
- `headroom-proxy`: `config.rs` / `main.rs` — `--enable-kompress` / `HEADROOM_PROXY_ENABLE_KOMPRESS` flag (default off; Kompress carries a ~261 MB model so it is opt-in, mirroring the Python `config.enable_kompress`). Startup sets the gate and fires an off-request-path cache-only warm-up when enabled.
- `tests`: flipped/added live-zone dispatch routing tests.
- `CHANGELOG.md`: Unreleased entries for both compressors.

## Testing

- [x] Unit tests pass (`pytest`) — N/A for runtime (Rust change)

### Test Output

```text
# Rust suites
cargo test -p headroom-core   ->  918 passed, 3 ignored (14 suites)
cargo test -p headroom-proxy  ->  407 passed (34 suites)
cargo test -p headroom-core --test live_zone_dispatch  ->  routing tests green

# Parity still byte-exact after wiring

cargo clippy -p headroom-core -p headroom-proxy -p headroom-parity -- -D warnings  ->  No issues found
cargo fmt --check  ->  clean
```

## Real Behavior Proof

- Environment: Linux (WSL2, kernel 6.6), Rust stable toolchain.
- Exact command / steps: `cargo test -p headroom-core` and `cargo test -p headroom-proxy`; `cargo test -p headroom-core --test live_zone_dispatch`; re-ran the parity harness to confirm wiring did not perturb output.
- Observed result: 918 core + 407 proxy tests pass; live-zone dispatch routing tests green; parity remains 30/30 code + 21/21 kompress; clippy and fmt clean. With the gate off (default), `PlainText` passes through unchanged; with `--enable-kompress` and a cold cache, dispatch still passes plain text through (Kompress unavailable) rather than blocking.
- Not tested: end-to-end against a live provider (dispatcher is covered by integration tests, not live traffic in this PR); alternate ONNX execution providers (default CPU EP here).

## Review Readiness

## Additional Notes

- Split of #1143 — PR 3 of 3. Lands the engines from PR 1 + PR 2 into the dispatcher.
- CCR markers are left to the dispatcher: both engines return the compressed string only; live-zone CCR uses the `<<ccr:>>` convention, so fixtures are recorded with `enable_ccr=False` for determinism.
- `pytest`/`mypy` checklist items are N/A — Rust change; equivalent gates are `cargo test` + `cargo clippy`, run above.

---

## Update — non-blocking model load + rebase

Pushed `fix(kompress): never block the request path on model load` and rebased this branch onto the updated #1153/#1154.

- **Non-blocking `kompress()`** — the request-path accessor previously used `OnceLock::get_or_init`, so a request thread that hit a `PlainText` block while the ~261 MB model was still loading (or while a slow EP graph compile ran — the OpenVINO NPU compile takes ~13s+) **blocked on that init and stalled the proxy**. It's now a non-blocking `OnceLock::get` (returns `None` ⇒ pass through until ready); the one-time load happens only in `warm_live_zone_compressors`, off the request path. So an enabled-but-not-yet-warm Kompress degrades to passthrough instead of hanging live traffic.
- **Rebase** — `from_cache`/`hf_cache_file` moved to the engine PR (#1153) where they belong; this branch's wiring commit reduces to the dispatch glue. Stack integrity intact (`stack/1 ← stack/2 ← stack/3`).

Validated downstream against a live Intel NPU run: proxy stays responsive through the NPU compile, then `kompress_ready:true` and prose blocks compress on-device. `cargo fmt`/`clippy` clean, dispatch + core tests green.


RubenAAA:
Rebased onto the updated #1153/#1154 and added `fix(kompress): never block the request path on model load` — the old `get_or_init` accessor blocked request threads on the (slow, ~13s+ on NPU) model load and stalled the proxy; it's now a non-blocking `get` that degrades to passthrough until the off-path warm completes. Stack integrity intact. Description updated.

JerrettDavis (changes requested):

JerrettDavis (approved):
Comprehensive pass looks good from a code-review standpoint. I re-reviewed the compressor ports and the live-zone wiring after the latest update/merge from main.

What I checked:
- `SourceCode` now routes through the Rust `CodeAwareCompressor` with the existing token-shrinking and syntax-validity gates preserved.
- `PlainText` routes through Kompress only after the process-wide opt-in and only when the cache-only warm has populated the singleton, so the request path does not block on the 261 MB model load.
- Proxy startup sets the gate from `--enable-kompress` / `HEADROOM_PROXY_ENABLE_KOMPRESS` and warms on a blocking task, which matches the intended degraded behavior: enabled but cold/not-ready means passthrough, not request stalls.
- The PR body matches the current stack and behavior after the non-blocking load update.

Local verification I ran:
- `cargo test -p headroom-core --test live_zone_dispatch`
- `cargo test -p headroom-core kompress --lib`
- `cargo test -p headroom-proxy config`
- `cargo test -p headroom-core --test code_compressor_parity`
- `cargo test -p headroom-core --test kompress_parity`
- `cargo fmt --check`

No blocking code findings from this pass. GitHub still reports `mergeStateStatus=UNSTABLE` because the PR Governance `template` and `label` checks are queued, so this still needs those required checks to settle before merge.

chopratejas:
**Rebase + review pass — this is close; two things to update** 👇

I pulled the stack down and brought #1155 current with `main` to see what's actually blocking it. Good news: the code is solid and the conflict is trivial.

**✅ Verified on current `main`**
- Merged `main` in — **the only conflict is `CHANGELOG.md`** (trivial union). `main` hasn't touched `live_zone.rs` / `config.rs` / `main.rs` / `mod.rs` since the fork, so the dispatch wiring merges clean.
- `headroom-core` lib: **853 passed / 0 failed**
- `live_zone_dispatch`: **7 passed** (incl. `source_code → CodeCompressor` routing)
- `code_compressor_parity`: **passed for real** — 30 fixtures, 8 languages
- `headroom-proxy`: builds + links clean

**👍 Design notes**
The gate is done right: default-off, cache-only model resolution, and the request-path `kompress()` accessor is non-blocking (`OnceLock::get`, not `get_or_init`) with the model load isolated to `warm_live_zone_compressors()` off the request path. `80ce58e4` ("never block the request path") is the key correctness fix.

**🔧 What needs updating**
1. **Rebase on `main`** to clear the `CONFLICTING` state — only `CHANGELOG.md` conflicts (union the two Features lists).
2. **Kompress ML parity isn't actually enforced in CI.** Both `kompress_parity` and the `plain_text → Kompress` dispatch test **skip (return early → "pass") when the ModernBERT ONNX model isn't in the HF cache** — the CI default. So the "21-fixture byte-parity" guarantee is only checked where someone has the 261 MB model cached locally. Suggest a CI job that warms/caches the model, or at minimum confirming a local parity run with the model present before merge. (The deterministic CodeCompressor parity runs for real, so that one's covered.)

Happy to hand over the resolved rebase if useful. Nice work on the port 🚀

RubenAAA:
Rebased on main and pushed — CHANGELOG.md resolved as the union of the two Features lists, everything else merged clean as you noted.

On the parity gap: you're right that the ML parity tests were skip-passing without the model, so I made two changes in 329df48c:

- The model-gated tests (`kompress_parity` and the `plain_text` dispatch test) now fail hard instead of skipping when `HEADROOM_REQUIRE_KOMPRESS_MODEL` is set, so a cold cache can't silently green-light a run.
- New `kompress-parity` job in `rust.yml` that caches `~/.cache/huggingface/hub` and warms the ModernBERT tokenizer + kompress-v2-base ONNX artifact via `hf download` on cache miss, then runs both suites with that env var set.

Verified locally with the model present using the same download commands the job runs: kompress parity — 21 fixtures matched byte-for-byte, live_zone_dispatch 7/7, and the full workspace passes with the flag set (1343 passed / 0 failed). Also confirmed the failure mode: flag set + no model cached fails loudly.

Thanks for the thorough pass 🙏

JerrettDavis (approved):
Re-reviewed the current head after the Kompress parity CI follow-up.

This addresses the remaining review concern I had: the model-backed Kompress path is no longer able to silently skip in CI. The new `kompress-parity` job warms the ModernBERT tokenizer and `kompress-v2-base` ONNX artifact into the HF cache, then runs the model-gated parity and dispatch suites with `HEADROOM_REQUIRE_KOMPRESS_MODEL=1`; the Rust tests also now convert a missing model/cache into a hard failure under that flag. The default local behavior still degrades to skip/passthrough when the model is absent, which preserves the developer ergonomics outside the explicit parity gate.

Focused local verification:
- `cargo test -p headroom-core --test kompress_parity --quiet` (2 passed locally with no required-model flag)
- `cargo test -p headroom-core --test live_zone_dispatch --quiet` (7 passed)
- `cargo fmt --check` (clean)

No remaining code-review blocker from this pass; the branch still depends on the GitHub workflow queue/status settling before merge.

JerrettDavis (approved):
Approved after maintainer refresh onto current main.

Verified locally on head d4740cac:
- cargo fmt --check
- cargo test -p headroom-core --test live_zone_dispatch --quiet
- cargo test -p headroom-core --test code_compressor_parity --quiet
- cargo test -p headroom-core --test kompress_parity --quiet
- cargo test -p headroom-proxy config --quiet (compiled proxy test targets and matched/passed the config-filtered test)
- git diff --check origin/main...HEAD
- merge commit hook: conflict scan and plugin version sync

Resolved the changelog conflict by keeping current main's release-managed changelog. Also reconciled the stack's direct ORT dependency with main's optional/dynamic ORT dependency so the manifest has one pinned optional `ort` entry with `load-dynamic`. Final PR surface remains the stacked Rust compressor wiring, parity fixtures, and Rust CI updates.

chopratejas:

Merged current `main` (`aebe1953`) into `stack/3-dispatch-gate` and approved the pending workflow runs. New head: `5c80b2d6`.

chopratejas:
### Status, and two things blocking this one specifically

#1153 and #1154 are resynced onto current `main`, fully green, and their ports are now verified byte-identical to Python (`kompress 21/21`, `code_aware_compressor 30/30`, real ONNX inference). This PR still has two blockers that they do not.

**1. CI cannot run on this PR.** It is a fork PR that modifies `.github/workflows/rust.yml` (the `kompress-parity` job). GitHub is creating **no** `pull_request` workflow runs at all for the current head — nothing in `waiting`, `pending`, `queued`, or `requested`, so there is nothing for a maintainer to approve. Close/reopen did not help. #1153 and #1154 touch zero workflow files and their CI ran fine, which is the discriminator. Suggested fix: split the `kompress-parity` job into a separate maintainer-authored PR on the base repo. That sidesteps the fork restriction, and the job can then be written against the new gating pattern (see below). The remaining dispatch-wiring change would be workflow-free and get normal CI.

**2. The `kompress-parity` job will hang, not fail, on an old runtime.** It deliberately warms the ModernBERT tokenizer and `kompress-v2-base` ONNX model into the HF cache so the model-gated tests run for real instead of skipping. That is exactly the condition that triggers the `ort` deadlock described on #1153 — `fastembed` enables `api-24`, so anything below ONNX Runtime 1.24 makes `Session::builder()` block forever at 0% CPU with no error. The job would sit until its 30-minute timeout with nothing to diagnose. #2591 raises every floor to `>=1.24`; land that first.

**Also worth knowing:** this branch has gone `CONFLICTING/DIRTY`. Its `rust.yml` change collides with #2580, which moves the workflow-level `paths:` filter into a `rust-changes` job plus per-job `if:` gates (so `parity` can become a required check without deadlocking non-Rust PRs). When you rebase, `kompress-parity` will want `needs: rust-changes` and `if: needs.rust-changes.outputs.rust == 'true'` to match the other five jobs.

One note on scope: this is the only PR of the three that changes runtime behaviour — `SourceCode → CodeCompressor` is always-on here, while Kompress is behind `--enable-kompress` (default off). #1153 and #1154 are inert; nothing calls their engines. So this is the one that wants a careful review pass once CI can actually run.

RubenAAA:
### Both blockers addressed — this PR no longer touches any workflow file

New head: `8a426807`. Three changes.

**1. The `kompress-parity` job is gone from this PR.** You're right that this is the discriminator — a fork PR that edits `.github/workflows/rust.yml` gets no `pull_request` run created at all, so there was never anything to approve. `git diff main HEAD -- .github/` is now empty, and the remaining diff is dispatch wiring, engines, and parity fixtures. Normal CI should create runs.

What stays is the half that doesn't need a workflow: `HEADROOM_REQUIRE_KOMPRESS_MODEL` still turns every skip into a hard failure, on both `kompress_parity` and the `plain_text` dispatch test. So whoever lands the job gets the real assertion for free — it just needs the env var set. The job YAML is at the bottom of this comment, rebased onto the #2580 gating pattern so it can be lifted as-is into a maintainer-authored PR.

**2. Conflict cleared.** Merged current `main` (via `stack/2`), so `CONFLICTING/DIRTY` is resolved. The `rust.yml`↔#2580 collision you flagged is moot now that this branch leaves the file alone. `CHANGELOG.md` is untouched, per the Changelog Guard.

**3. Found the same `--no-default-features` break in the wiring.** Once your `#[cfg(feature = "ml")]` gate landed on `pub mod kompress`, this branch stopped building without default features — `live_zone.rs` imported `super::kompress::{Kompress, KompressConfig}` unconditionally. Gated the singleton, `warm_live_zone_compressors`, the `STRATEGY_KOMPRESS` tag, and the `PlainText` dispatch arm. Without `ml`, `PlainText` returns `NoOp` — the same passthrough a stock build takes with a cold model cache, so there's no new behaviour, just one fewer way to route there. Also fixed a `manual_range_contains` clippy nit of mine in the `code_compressor` tests that `--all-targets` surfaces.

Verified on the merged head: `cargo fmt --check`, `cargo clippy --workspace -- -D warnings`, `cargo check -p headroom-core --no-default-features`, and `cargo test -p headroom-core -p headroom-proxy` (973 passed / 0 failed). Both model-gate modes confirmed: cold cache skips and passes; cold cache with `HEADROOM_REQUIRE_KOMPRESS_MODEL=1` fails with `model not cached`.

**On #2591 — agreed, land it first, and it needs one more case.** I hit the deadlock on this machine, and the trigger wasn't an old runtime: there was no resolvable `libonnxruntime.so` at all. `kompress_parity` parked at 0.0% CPU for twelve minutes, all three threads in `futex_wait_queue`, nothing onnx-shaped in `/proc/<pid>/maps`. Same re-entrant `Once` — anything that fails inside `load_dylib_from_path` hangs, and "library missing" is as fatal as "version too old". Raising the floors fixes the version case; a runner that resolves nothing would still burn the full 30 minutes silently. Hence the pre-flight step in the job below: it costs two seconds and turns both cases into a readable failure.

**Separately, one divergence worth a look** (in `kompress.rs`, so it belongs to #1153 — flagging rather than changing an approved branch): `hf_hub_roots()` is documented as a precedence chain but implemented as a fallback chain. `hf_cache_file` walks `HF_HUB_CACHE` → `HF_HOME/hub` → `$HOME/.cache/...` and returns the first hit, so setting `HF_HUB_CACHE` to a directory that lacks the model silently loads it from `$HOME` instead. `huggingface_hub` treats `HF_HUB_CACHE` as authoritative — if it's set and the model isn't there, that's a miss. It cost me a while to isolate a run today, and an operator pointing the proxy at a curated cache would get a model from somewhere else without being told. Happy to fix it here or on #1153, whichever you prefer.

Also worth noting for the review pass you mentioned: agreed this is the only one of the three with runtime behaviour. `SourceCode → CodeCompressor` is always-on and is the part that deserves the scrutiny; Kompress is default-off and additionally cache-gated, so it needs both a flag and a warmed cache before it can touch a request.

---

<details>
<summary>The <code>kompress-parity</code> job, rebased onto the #2580 gating pattern</summary>

```yaml
  kompress-parity:
    # Enforces the Kompress ML byte-parity tests for real: warms the
    # ModernBERT tokenizer + kompress-v2-base ONNX model into the HF cache,
    # then runs the model-gated tests with HEADROOM_REQUIRE_KOMPRESS_MODEL
    # set so a cold cache fails instead of silently skipping.
    name: kompress parity (model-backed)
    needs: rust-changes
    if: needs.rust-changes.outputs.rust == 'true'
    runs-on: ubuntu-latest
    timeout-minutes: 30
    steps:
      - uses: actions/checkout@v6
      - uses: dtolnay/rust-toolchain@stable
        with:
          toolchain: 1.95.0
      - uses: Swatinem/rust-cache@v2
      - name: Cache HuggingFace model
        id: hf-cache
        uses: actions/cache@v4
        with:
          path: ~/.cache/huggingface/hub
          key: hf-kompress-v2-base-modernbert-v1
      - uses: actions/setup-python@v6
        with:
          python-version: '3.11'
      - name: Warm HF model cache
        if: steps.hf-cache.outputs.cache-hit != 'true'
        run: |
          pip install -q -U huggingface_hub
          hf download answerdotai/ModernBERT-base tokenizer.json
          hf download chopratejas/kompress-v2-base onnx/kompress-int8-wo.onnx
      # Without this, an ORT older than 1.24 — or none at all — deadlocks
      # inside Session::builder() at 0% CPU and the job dies on its timeout
      # with nothing to read. Two seconds here buys a readable failure.
      - name: Assert ONNX Runtime >= 1.24
        run: |
          pip install -q -U onnxruntime
          python - <<'PY'
          import onnxruntime, sys
          v = tuple(int(p) for p in onnxruntime.__version__.split(".")[:2])
          print("onnxruntime", onnxruntime.__version__)
          if v < (1, 24):
              sys.exit("onnxruntime >= 1.24 required (ort-sys ORT_API_VERSION 24 via fastembed); older versions deadlock instead of erroring")
          PY
      - name: Run kompress parity + dispatch tests (model required)
        env:
          HEADROOM_REQUIRE_KOMPRESS_MODEL: "1"
        run: |
          cargo test -p headroom-core --test kompress_parity -- --nocapture
          cargo test -p headroom-core --test live_zone_dispatch -- --nocapture
```

</details>

RubenAAA:
Correcting one number in my comment above: the test total I quoted was wrong.

I wrote "973 passed / 0 failed" for `cargo test -p headroom-core -p headroom-proxy`. That figure came from a run that had stalled at `kompress_parity` and never reached the later suites, so it was a partial count reported as a full one. The stall was the `ort` deadlock, on this machine, exactly as described.

Rerun to completion with the model cache isolated: **1423 passed, 0 failed, 1 ignored** across 50 suites, exit 0.

Everything else in that comment was verified directly and stands: `cargo fmt --check`, `cargo clippy --workspace -- -D warnings`, `cargo check -p headroom-core --no-default-features`, and both model-gate modes (cold cache skips; cold cache with `HEADROOM_REQUIRE_KOMPRESS_MODEL=1` fails with `model not cached`).

One note on isolating that rerun, which is the practical face of the `hf_hub_roots()` fallback-chain point above: setting `HF_HUB_CACHE` and `HF_HOME` at an empty directory was not enough — the resolver kept walking to `$HOME` and finding the model there. I had to override `HOME` as well. That is the divergence from `huggingface_hub`, and it makes the model-gated tests hard to exercise in their cold-cache mode on any machine that has the model.

Also confirming the workflow-file diagnosis was right: on the new head, all seven runs (`CI`, `rust`, `Security`, `Init E2E`, `Wrap E2E`, `Changelog Guard`, `Merge Conflicts`) now exist at `action_required`, where the previous head produced none at all. They need a maintainer approval to start.

chopratejas:
### Resynced onto main (`28cf4b92`) — and one blocker is not actually pushed

#1153 (`83e27e50`) and #1154 (`e530de5a`) are merged. Both were green; #1154's `test (1)` was the p99=76ms flake and cleared on a rerun. This branch is resynced on top, plus #2580 and #2591.

**Two conflicts, both squash-merge artifacts.** Your content landed on main as different commits, so git could not tell the two were the same change:

- `crates/headroom-core/Cargo.toml` — comment-only divergence above an identical `ort` dep line; all eight tree-sitter pins byte-identical. Took your side.
- `crates/headroom-core/tests/kompress_parity.rs` — add/add. Your 146-line version supersedes main's 131-line one by adding `HEADROOM_REQUIRE_KOMPRESS_MODEL`, so I kept yours.

Parity is green on the result: **227 fixtures, 182 matched, 45 skipped, 0 diffed** — `kompress 21/21`, `code_aware_compressor 30/30`, under ONNX Runtime 1.24.4.

---

**The dispatch-wiring fix isn't on the branch.** You described cfg-ing the `PlainText` arm, the singleton, and `warm_live_zone_compressors`, but at the PR head (`8a426807`, before my merge) `live_zone.rs:108` still reads:

```rust
use super::kompress::{Kompress, KompressConfig};
```

ungated, so:

```
cargo check -p headroom-core --no-default-features
error[E0432]: unresolved import `super::kompress`
  --> crates/headroom-core/src/transforms/live_zone.rs:108:12
```

I checked this three ways before concluding it was yours and not mine: my merge doesn't touch `live_zone.rs`, the branch fails the same check on its own at `8a426807`, and main passes it. Looks like the fix is still local — I deliberately did **not** implement it, since my version would collide with yours. Happy to if you'd rather; otherwise a push should clear it.

**Also confirming your CI diagnosis was right, and it's now moot.** Dropping the `kompress-parity` job means this PR no longer touches `.github/`, so the fork-workflow restriction no longer applies — all 58 checks ran and passed. And the hang risk it carried is covered anyway: the model-gated tests now run inside `test (ubuntu)`, which #2591 gave the `>= 1.24` floor plus a pre-flight that fails in seconds if the runtime is missing or old. So your `HEADROOM_REQUIRE_KOMPRESS_MODEL` gate can be enforced by whichever job warms the model without risking a silent 30-minute timeout.

**One finding that is main's problem, not yours.** `cargo clippy --workspace --all-targets -- -D warnings` fails on main with two errors — `code_compressor.rs:1856` (`manual_range_contains`, the nit you mentioned) and `log_compressor.rs:1780` (`field_reassign_with_default`). CI runs `cargo clippy --workspace` **without** `--all-targets`, so test-target lints are never linted and both slipped through #1153/#1154. Worth adding `--all-targets` to the workflow as its own change; I'll raise it separately rather than bolt it onto this stack.

Last thing on review scope: this is the only PR of the three that changes runtime behaviour. `SourceCode → CodeCompressor` is always-on, and Kompress is behind `--enable-kompress` / `HEADROOM_PROXY_ENABLE_KOMPRESS` with `default_value_t = false`. So the always-on CodeCompressor path is the part that wants the careful pass — everything merged so far has been inert.

JerrettDavis (changes requested):
The dispatch path still has a real no-default-features break on current head `28cf4b92`.

Local verification:

```
cargo check -p headroom-core --no-default-features
```

fails with:

```
error[E0432]: unresolved import `super::kompress`
  --> crates/headroom-core/src/transforms/live_zone.rs:108:12
  |
108 | use super::kompress::{Kompress, KompressConfig};
  |            ^^^^^^^^ could not find `kompress` in `super`
```

`cargo test -p headroom-core --test live_zone_dispatch --quiet` still passes with default features, so the runtime dispatch behavior I previously reviewed is intact. The blocker is the feature-gated build: `kompress` is behind `#[cfg(feature = "ml")]` but `live_zone.rs` imports and uses it unconditionally. Please gate the Kompress singleton/warm path/PlainText dispatch arm when `ml` is disabled, degrading PlainText to the existing passthrough/no-op behavior.

JerrettDavis (approved):
Re-reviewed current head 0aa87d6. The previous no-default-features blocker is fixed: `headroom-core` now checks successfully without default features, and the focused live-zone dispatch/parity tests still pass.

Validated locally:

- `cargo check -p headroom-core --no-default-features` -> passed
- `cargo fmt --check` -> passed
- `cargo test -p headroom-core --test live_zone_dispatch --quiet` -> 7 passed
- `cargo test -p headroom-core --test code_compressor_parity --quiet` -> 1 passed
- `cargo test -p headroom-core --test kompress_parity --quiet` -> 2 passed

GitHub checks are green on this head as well.

JerrettDavis (approved):
Re-reviewed exact head b352d5ee after the main sync. Live-zone CodeCompressor/Kompress dispatch remains correctly feature-gated, including no-default-features compatibility; current checks are green.

PR #1157, opened 2026-06-19
feat(learn): weight loops in Headroom Learn + RTK-loop eval

purva-8:
## Why

`headroom learn` ranked recommendations by a single LLM-guessed `estimated_tokens_saved` with a flat hardcoded `confidence`, and had **no notion of a loop**. Two consequences:

1. **RTK re-fetch loops were invisible.** RTK truncates a command's output (`grep foo` → `grep foo | head -50`, see `docs/rtk-architecture.md`), so the agent re-runs larger-limit variants to fetch more. Those calls **succeed** (`is_error=False`), and `analyze()` even early-returned when a session had no failures and no events — skipping the loop entirely.
2. **Even when surfaced, a loop ranked no higher than a one-off mistake** — ranking trusted the LLM's per-rule guess.

This implements loop-aware weighting (the "give more weightage to loops" ask) plus the eval that reproduces an RTK loop, runs it through Learn, and checks the guardrail prevents re-triggering.

## What

New `headroom/learn/loops.py`:
- **`detect_loops()`** — group calls by a canonical signature that collapses RTK pagination/limit variants (`head -N`, `-n N`, `LIMIT N`, …), flag ≥3 repeats, classify `error-loop` vs `rtk-refetch-loop`, and compute **measured** wasted tokens.
- **`format_loops_for_digest()`** — surface loops as a highest-priority digest section.
- **`apply_loop_weighting()`** — raise a matching recommendation's savings to at least the loop's measured waste and tag it `is_loop_guardrail`, so loops outrank one-offs **deterministically** (not by asking the LLM for a bigger number — it rated the eval's loop at 150 tok vs ~5,000 measured).

Wiring in `analyzer.py`: detect loops up front (fixes the no-failure early-return), lead the digest with them, prioritize loops in the system prompt, re-sort after weighting. Adds `is_loop_guardrail` / `loop_occurrences` to `Recommendation`.

## Eval

`benchmarks/rtk_loop_learn_eval.py` — Phase 1 triggers the loop → learns → scores the guardrail; Phase 2 injects it and asserts a guarded (non-looping) session doesn't re-trigger it. Deterministic in CI; real-LLM via `--real` / `real_llm` marker.

```
[PASS] loop_detected          (1 loop(s), ~5,005 tok wasted)
[PASS] guardrail_produced
[PASS] ranked_first
[PASS] names_command
[PASS] prescribes_fix
[PASS] weight_reflects_waste
[PASS] guardrail_holds
```

## Tests

`tests/test_learn/test_loop_weighting.py` (13) + `test_rtk_loop_eval.py`. Full learn suite: **190 passed, 3 skipped** (real-LLM eval skips without a key). Existing learn tests unchanged and green. Ruff clean + formatted.

## Honest limitations (see `docs/rtk-loop-weighting.md`)

- Phase 2 is a **non-recurrence check**, not a live agent obeying the rule end-to-end - natural next step if we want a stronger claim.
- Loop signature is heuristic (covers common RTK truncation shapes); `min_occurrences=3` is tunable; rule↔loop matching is fuzzy token-overlap.

### DIFF
--- a/benchmarks/rtk_loop_learn_eval.py
+++ b/benchmarks/rtk_loop_learn_eval.py
@@ -1,0 +1,264 @@
+"""RTK-loop eval — does Headroom Learn catch a loop and write a guardrail that
+would prevent it recurring?
+
+This is the agentic eval for the loop-weighting work. It runs in two phases:
+
+  Phase 1 — TRIGGER + LEARN
+    Reproduce an RTK re-fetch loop (a grep whose RTK-truncated output forces the
+    agent to re-run larger-limit variants), run it through ``SessionAnalyzer``,
+    and SCORE the resulting guardrail:
+      • produced        — a loop guardrail was emitted at all
+      • ranked_first    — it outranks the one-off rules (the weighting works)
+      • names_command   — the rule identifies the command that looped
+      • prescribes_fix  — the rule says how to avoid it (fetch full output once)
+      • weight_reflects — its savings estimate >= the MEASURED wasted tokens
+
+  Phase 2 — GUARDRAIL HOLDS
+    Inject that guardrail as a prior learned pattern, then feed a session where
+    the agent FOLLOWED it (one full-output fetch, no loop). Re-run the analyzer
+    and assert NO new loop guardrail is produced for that command — i.e. once
+    the rule exists and is honored, the loop does not re-trigger and Learn does
+    not need to relearn it.
+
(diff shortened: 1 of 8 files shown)


purva-8:
Superseded by #1160 — reopened as a draft that follows CONTRIBUTING.md: linked feature issue (#1159), full PR template with Real Behavior Proof (real-LLM run via the claude CLI backend), CHANGELOG entry, and a non-name-prefixed branch.

Issue #946, opened 2026-06-13
[BUG] headroom wrap claude: compression_first_stage hangs at 30s on long Claude Code conversations (msgs>=29)


flashydotwala:
## Summary

When using `headroom wrap claude` against a real Claude Code session, the compression pipeline reliably times out at exactly 30s on every request once the conversation reaches roughly 29+ messages. Short conversations (msgs ≤ ~5) complete content_router in 15–20 ms. Long conversations never produce a "Transform content_router" log line — the work appears to hang inside the pipeline, gets cancelled by the asyncio `wait_for(30s)`, and the underlying thread keeps running (visible as a rising `in_flight` count in `/debug/warmup`).

Net effect: zero compression on any moderately-long Claude Code session. `transforms=none` after each timeout; `compressions_by_strategy` stays empty for the entire session even though Kompress is loaded and pre-warmed.

## Environment

- headroom-ai **0.25.0** (pipx-installed, Python 3.14.0 venv)
- OS: Windows 11
- Client: `claude-cli/2.1.141` (and `2.1.177`) via `headroom wrap claude`
- Code-aware: **DISABLED** (ruled out — see #562)
- Kompress: ENABLED, backend=onnx, pre-loaded successfully at startup (`Kompress ONNX loaded: chopratejas/kompress-v2-base backend=onnx`)
- Memory: DISABLED
- CCR: ENABLED (default)
- Mode: token
- Model in use: `claude-opus-4-7`

## Reproduction

1. `headroom wrap claude` in a project directory
2. Have a conversation with several tool calls (Read, Grep, etc.) so message count rises
3. Once `msgs >= ~29` in the request, every subsequent request times out at `compression_first_stage`

### Healthy short request (`msgs=5`)

```
13:54:24 - content_router: 5 msgs — 1 skipped (<50 words), 4 content-block msgs
13:54:24 - Transform content_router: 26264 -> 26264 tokens (saved 0) [20.7ms]
13:54:27 - PERF model=claude-haiku-4-5-20251001 msgs=5 tok_before=27072 tok_after=27072 tok_saved=0 opt_ms=46 transforms=router:noop
```

### Hanging long request (`msgs=29, 31, 33, …`)

```
13:54:41 - WARNING - Optimization failed: TimeoutError:
13:54:41 - CCR: deferring tool injection (frozen prefix=2) to preserve cache
[...]

PR #1161, opened 2026-06-19
fix(kompress): never block the request path on the cold-cache model download

### LINKED ISSUE #1146: Cold cache: 274 MB kompress ONNX model is downloaded inline on the request path -> 30s compression_first_stage timeout + silent fail-open (distinct root cause from #946)
### TL;DR
This is a **second, distinct root cause** for the same 30s `compression_first_stage` timeout tracked in #946 -- but one that hits on a **cold cache / first run**, before the inference-slowness path in #946 ever applies. In #946 the reporter had the model **pre-downloaded** (`HF_HUB_OFFLINE=1`, no HF traffic) and the 30s was spent in `onnxruntime.run`. Here the 30s is spent **downloading the model inline on the request thread**, and because the download never completes it **never caches**, so every deep-path request repeats it. Splitting it out per #1144 so the cold-cache download path can be tracked independently of the warm-inference path.

### Symptom (identical to #946)
`opt_ms ~ 30008`, `transforms=none`, fail-open (passthrough, `tokens_saved=0`, `error=None`) on every deep-path request.

### Root cause (stack trace)
py-spy of the wedged compression worker on a cold cache (Docker needs `--cap-add=SYS_PTRACE`):
```
Thread "headroom-compress_0":
    xet_get (huggingface_hub/file_download.py:566)
    hf_hub_download (huggingface_hub/file_download.py:1016)
    hf_hub_download_local_first (headroom/onnx_runtime.py:41)   <- request-time allow_network=True fallback
    _create_onnx_session (headroom/transforms/kompress_compressor.py:404)
    _load_kompress_onnx (kompress_compressor.py:482)
    compress (kompress_compressor.py:787)
    _try_ml_compressor (content_router.py:1596) ... apply (pipeline.py:326) <- <lambda> (anthropic.py:1069)
```
It is blocked in HuggingFace's **Xet** backend downloading `chopratejas/kompress-v2-base` -> `onnx/kompress-int8-wo.onnx` (**274,049,435 bytes**).

### Why it never recovers on its own
- The artifact is **Xet-only**: `resolve/main/onnx/kompress-int8-wo.onnx` 302-redirects to `cas-bridge.xethub.hf.co` -> `us.aws.cdn.hf.co/xet-bridge-us/...`. `HF_HUB_DISABLE_XET=1` does **not** help (no classic-LFS mirror).
- Raw `curl` of that URL (no headroom in the loop):
  - Healthy ~67 Mbit/s link: 251 MB in 30s; the **full 274 MB takes 32.6s** -- already longer than the hard-coded 30s `COMPRESSION_TIMEOUT_SECONDS` (`proxy/helpers.py`), so even a *good* network cannot finish inside the budget on the request path.
  - Network that cannot reach the Xet CDN (some WSL/CI/corp): **0 bytes in 30s** (curl exit 28) -> can never cache.
- HF only renames the blob on completion, so the cancelled partial leaves **nothing** cached -> the next request starts from zero and hangs again.

The code already anticipates this for **startup**: `onnx_runtime.hf_hub_download_local_first` is called with `allow_network=False` by the preloader (its docstring: so a cold cache cannot block, or via native crashes kill, the process before it binds its port). But the **request-time** path uses the `allow_network=True` default (`onnx_runtime.py:41`) and blocks the worker. So on a cold cache the startup preload fetches nothing and the first real request eats the 274 MB download.

### Measured impact (12-turn Claude-Code-style session: Bash x5, Read x3, Grep x2, Edit x2)
- Cold cache (stock install): 8/12 turns hit the 30s fail-open -> **2.1%** blended savings, +~30s/turn.
- Model pre-cached out-of-band + `--no-cache`: all 12 turns compress, 0 errors -> **29.14%** blended (kompress inference ~0.6-0.8s/turn here, first `InferenceSession` build ~2.1s). So the deep path is worth ~14x what a cold first-run delivers, and the entire gap is this download UX. (On tool-output/Bash content kompress does add real savings, unlike the pure-text case @1Aa1k profiled in #946.)

### Suggested directions (complementing #946's configurable-timeout PRs)
- Do not download the 274 MB model inline on the request path. Fetch it **out-of-band at startup with its own generous/owned timeout** (separate from the per-request `COMPRESSION_TIMEOUT_SECONDS`), and keep the deep path disabled (cheap heuristics only) until the model is actually cached.
- Make the inline download **resumable-aware** / do not cancel a partial that could finish next time.
- Surface `error != None` (or a one-time loud log) on the silent fail-open, so operators can see the deep path is off instead of 0% reading like nothing to compress.

### Env
headroom-ai 0.26.0 - onnxruntime + transformers present, torch absent (the ONNX path needs no torch) - Python 3.12 - Linux (Docker `python:3.12-slim`). Related: #946 (warm-inference variant of the same timeout), #1025 (sync kompress on request path under load), #790 (optimization mode blocks bind).

_Split out from #1144._

xspawnnn:
Closes #1146.

## Problem

On a cold cache, the first request that reaches the Kompress deep compressor triggers an inline `hf_hub_download` of the 274 MB `chopratejas/kompress-v2-base` ONNX model **on the request thread**. That download races the proxy's compression budget (`HEADROOM_COMPRESSION_TIMEOUT_SECONDS`, default 30s — the `compression_first_stage` timeout): the fetch is cancelled mid-transfer, **nothing finalizes in the HF cache**, and the request fails open (uncompressed). Because the partial blob never lands, every subsequent request repeats the same ~30s hang + fail-open, so the deep compressor never actually becomes available through the proxy.

This is a **distinct root cause from #946** (which concerns the timeout itself). Here the model must simply never be fetched synchronously on a latency-sensitive request.

## Fix

Make the request path cache-only and move the one-time download off-thread.

**`kompress_compressor.py`**
- `compress(..., allow_download=False)` — new keyword (default `True`, so the direct API and `compress_batch` are unchanged) that resolves the model cache-only; on a cold cache it raises `KompressModelNotCached` and passes through instead of blocking on the network.
- `is_ready()` — lockless cache-membership check, safe to call on the hot path.
- `ensure_background_download(model_id, device)` — starts at most one daemon thread per model to pull the artifact down out of band (a finished/failed thread is replaced, so a transient failure can be retried by a later request). The compression timeout does not bound this thread.

**`content_router.py`** — gate the deep path on readiness:
- not ready → return passthrough immediately and kick off the background download;
- ready → `compress(allow_download=False)` (cache-only, no network on the request thread).

Net effect: the cold-cache deep path returns in ~0 ms (passthrough) instead of hanging ~30 s; the model downloads once in the background; subsequent requests transparently use the deep compressor once it is cached.

## Verification

Clean install of `headroom-ai==0.26.0` (main `@9f7f3ad` + this patch), HF cache empty:

- **Cold-cache router request**: returns in **56 ms**, passthrough (output == input), fetch handed to a background daemon thread — vs. the pre-fix inline hang.
- **A/B on the same `compress()` call** with a slow-fetch stand-in: pre-fix (`allow_download=True`) blocked **24.00 s** on the request thread; post-fix (`allow_download=False`) returned **59 ms**.
- **4 new regression tests** in `tests/test_kompress_request_nonblocking.py` (cache-only passthrough; one-thread-per-model background download; router skips deep path when not ready; router stays cache-only when ready) — `4 passed`.

## Files
- `headroom/transforms/kompress_compressor.py` — non-blocking load + cache-only `compress`
- `headroom/transforms/content_router.py` — gate deep path on `is_ready()`; background fetch when cold
- `tests/test_kompress_request_nonblocking.py` — regression coverage

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -1663,21 +1663,28 @@ class ContentRouter(Transform):
         compressed: str | None = None
         compressed_tokens: int | None = None
 
-        # Primary: Kompress — downloads from chopratejas/kompress-v2-base on first use
+        # Primary: Kompress. On a cold cache the model is fetched once in the
+        # background (ensure_background_load) instead of blocking this request
+        # thread on a 274MB download that races the compression timeout and
+        # fails open. Until it is cached, route around the deep path.
         if self.config.enable_kompress:
             compressor = self._get_kompress()
             if compressor:
-                try:
-                    result = compressor.compress(
-                        text_to_compress,
-                        context=context,
-                        question=question,
-                        target_ratio=getattr(self, "_runtime_target_ratio", None),
-                    )
-                    compressed = result.compressed
-                    compressed_tokens = result.compressed_tokens
-                except Exception as e:
-                    logger.warning("Kompress failed: %s", e)
+                if not compressor.is_ready():
+                    compressor.ensure_background_load()
+                else:
+                    try:
+                        result = compressor.compress(
+                            text_to_compress,
+                            context=context,
+                            question=question,
+                            target_ratio=getattr(self, "_runtime_target_ratio", None),
+                            allow_download=False,
+                        )
+                        compressed = result.compressed
+                        compressed_tokens = result.compressed_tokens
+                    except Exception as e:
+                        logger.warning("Kompress failed: %s", e)
 
         if compressed is None:
             return content, len(content.split())
--- a/headroom/transforms/kompress_compressor.py
+++ b/headroom/transforms/kompress_compressor.py
@@ -678,6 +678,57 @@ def unload_kompress_model(model_id: str | None = None) -> bool:
     return True
 
 
+# ── Background model download ─────────────────────────────────────────
+#
+# The proxy request path must never block on a cold model download. A first
+# deep-path request would otherwise resolve the 274MB ONNX artifact via an
+# inline hf_hub_download on the request thread, where it races the proxy's
+# compression timeout (HEADROOM_COMPRESSION_TIMEOUT_SECONDS, default 30s). The
+# fetch is cancelled mid-transfer, the blob never finalizes in the HF cache,
+# and every subsequent request re-hangs and fails open. Instead the request
+# path resolves the model cache-only (allow_download=False) and pulls it down
+# once here, in a daemon thread that the compression timeout does not bound.
+
+_download_threads: dict[str, threading.Thread] = {}
+_download_threads_lock = threading.Lock()
+
+
+def _background_download(model_id: str, device: str) -> None:
+    try:
+        logger.info("Kompress: downloading model %s in the background ...", model_id)
+        _load_kompress(model_id, device, allow_download=True)
+        logger.info("Kompress: background model download complete for %s", model_id)
+    except Exception as exc:
+        logger.warning("Kompress: background model download failed for %s: %s", model_id, exc)
+
+
+def ensure_background_download(model_id: str = HF_MODEL_ID, device: str = "auto") -> None:
+    """Start a one-shot background download of the model if it isn't cached.
+
+    Idempotent and non-blocking: at most one download thread runs per model_id,
+    and a finished or failed thread is replaced on the next call so a transient
+    network failure can be retried by a later request. Once the download
+    completes the deep path activates on subsequent requests without ever
+    blocking one on the network.
+    """
+    if model_id in _kompress_cache:
+        return
+    with _download_threads_lock:
+        if model_id in _kompress_cache:
+            return
+        existing = _download_threads.get(model_id)
+        if existing is not None and existing.is_alive():
+            return
+        thread = threading.Thread(
+            target=_background_download,
+            args=(model_id, device),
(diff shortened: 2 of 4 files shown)


JerrettDavis (changes requested):
This PR is not merge-ready because CI is currently red: test (4)=FAILURE, lint=FAILURE. Please fix or rerun the failing checks after updating from current main.

xspawnnn:
✅ CI is now fully green on `60fc41c` — **23 checks pass, 0 failures** (the 2 skipped are the scheduled `weekly-suite` / `workflow-validation`, as expected).

The two checks that had gone red are fixed, both in **test-only** follow-up commits — no production code touched:

- **`lint`** — a `ruff format` nit on the cache-only `compress` signature in `test_kompress_request_nonblocking.py`.
- **`test (4)`** — the `FakeKompress` double in `test_content_router.py` predated the `is_ready()` / `ensure_background_load()` compressor contract this PR introduces, so the router hit `AttributeError` → passthrough and the targeted-kompress assertion failed. Taught the double those two methods; it now satisfies the full interface and the routing assertion passes.

Ready for review whenever you have a chance — thanks!

JerrettDavis (approved):

Issue #1005, opened 2026-06-15
Support CPUs without AVX/AVX2 (fallback builds or optional Rust-core)


janvda:
### Description

Headroom currently requires AVX/AVX2 CPU instructions for its Rust core (`headroom._core`). CPUs without these instructions (e.g., older Intel Atom, some AMD, and certain virtualized/container environments) get:

```
rust_core_error: "CPU lacks AVX support — headroom._core pre-built wheel requires AVX/AVX2"
```

While the Python fallback works, it would be great to have official support for non-AVX CPUs.

### Suggested approaches

1. **Separate wheel build** — provide an additional `headroom-ai` wheel compiled with `--target-cpu=generic` (or without AVX) for non-AVX platforms
2. **Runtime CPU feature detection** — auto-detect AVX at import time and fall through gracefully (currently seems to work via `HEADROOM_REQUIRE_RUST_CORE=false` but could be automatic)
3. **Document the workaround clearly** — if the Python-only mode is the intended fallback, document `HEADROOM_REQUIRE_RUST_CORE=false` more prominently for non-AVX users

### Environment
- CPU: Intel Atom / older Celeron series (no AVX)
- OS: Linux
- Python: 3.13
- Headroom version: 0.25.0

### Workaround
Setting `HEADROOM_REQUIRE_RUST_CORE=false` before starting the proxy works, but isnt documented for this use case.

PR #1162, opened 2026-06-19
fix: skip Magika backend on x86 CPUs without AVX2

dwizzle204:
## Description

Adds a narrow runtime AVX2 guard before initializing the Magika/ONNX Runtime detector on x86/x86_64. On x86/x86_64 CPUs without AVX2, Headroom falls back to existing non-Magika detection tiers instead of crashing during ONNX Runtime initialization. AVX2-capable systems retain existing behavior.

Refs #1005

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- Adds a Magika/ONNX Runtime CPU support guard before `Session::new()`.
- Returns a normal Magika init error on x86/x86_64 hosts without AVX2, allowing the existing detection chain to fall through to non-Magika tiers.
- Keeps AVX2-capable x86/x86_64 behavior unchanged.
- Does not apply the x86-specific AVX2 gate on non-x86 targets.
- Adds CPU-aware Rust tests and a short troubleshooting note.

## Testing

### Test Output

```text
$ cargo test -p headroom-core --lib --locked
833 passed; 0 failed; 1 ignored

$ cargo test --workspace --locked
passed

$ cargo clippy -p headroom-core --locked -- -D warnings
clean
```

## Real Behavior Proof

- Environment: x86_64 Linux host with AVX but no AVX2 (Intel Xeon E5-2697 v2 on Proxmox), local build from this branch.
- Exact command / steps: `python -X faulthandler -c 'from headroom._core import detect_content_type; print(detect_content_type("hello world"))'`
- Observed result: before — process exited with `Fatal Python error: Illegal instruction`; after — command completed successfully returning `DetectionResult(content_type="text", ...)`, and full `cargo test -p headroom-core --lib --locked` passed with 833/0/1.
- Not tested: generic no-AVX CPUs, alternate ONNX Runtime builds, non-x86 platforms.

## Review Readiness

## Checklist

N/A

## Additional Notes

This partially addresses #1005 by handling one concrete native crash class: the Magika detector initializes ONNX Runtime through ort/ort-sys, whose precompiled runtime can contain AVX2-family instructions. On AVX-only x86_64 hosts, that initialization can SIGILL before Headroom can fall back.

Scope:

- This does not introduce generic no-AVX wheels.
- This does not redesign Rust-core packaging.
- This does not disable the Rust core globally.
- This only prevents the Magika/ONNX detector tier from loading on x86/x86_64 CPUs where AVX2 is unavailable.
- Non-Magika detection tiers continue to run.
- On non-x86 targets, this x86-specific AVX2 gate is not applied.

Changelog omitted: small native detector fallback fix with no public API change.

### DIFF
--- a/crates/headroom-core/src/transforms/detection.rs
+++ b/crates/headroom-core/src/transforms/detection.rs
@@ -28,6 +28,14 @@
 //! issue. Loud-on-error stays at the [`magika_detect`] entry point for
 //! callers who care; the chain swallows the err with a log line.
 //!
+//! # CPU compatibility
+//!
+//! Precompiled ONNX Runtime binaries from `ort-sys` may contain AVX2-family
+//! instructions. On x86/x86_64 CPUs where AVX2 is unavailable, the magika
+//! session init returns an init error before touching ONNX. The chain
+//! handles this identically to any other tier-1 error — logs it and falls
+//! through to Tier 2 / Tier 3.
+//!
 //! # SearchResults / BuildOutput
 //!
 //! The retired regex detector recognized grep-style search output
@@ -84,6 +92,7 @@ pub fn detect(content: &str) -> ContentType {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::transforms::magika_detector::magika_onnx_runtime_supported_by_cpu;
 
     #[test]
     fn empty_input_short_circuits_to_plain_text() {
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
Reviewed the Magika AVX2 guard. The check is applied before ONNX session initialization on x86 targets, the detection chain degrades through existing fallback tiers, and tests/documentation cover the no-AVX2 behavior. CI is green.

MichaelvanLaar:
## Validation on a no-AVX CPU (not just no-AVX2)

The PR description lists **"Generic no-AVX CPUs"** as not tested. Here's a reproduction on exactly that hardware, in case it helps close that gap and unblock the merge.

**Environment**

- CPU: Intel Celeron N4500 @ 1.10 GHz (Jasper Lake / Tremont — SSE4.2 max, no AVX, no AVX2)
- CPU flags: `sse sse2 sse3 sse4_1 sse4_2 sha_ni waitpkg gfni` — `avx` is **absent**
- OS: Ubuntu Linux 6.8.0-124-generic x86_64
- Python: 3.14.6 (pipx venv)
- headroom-ai: 0.27.0

**Reproduction (current main, before this PR)**

```
$ python -c "import headroom._core"
Illegal instruction (core dumped)
$ echo $?
132
```

Exit code 132 = killed by signal 4 (SIGILL). The crash happens at `.so` load time in `headroom/transforms/error_detection.py` line 41, which does an unconditional top-level `from headroom._core import …`. Because the SIGILL fires before any Python exception handler can run, `HEADROOM_REQUIRE_RUST_CORE=false` does not prevent the crash — the proxy process dies before reaching `_check_rust_core()`.

**Why this matters for the PR**

The guard `!is_x86_feature_detected!("avx2")` would also match this CPU (no AVX implies no AVX2), so the Magika/ONNX path should be skipped here too. Whether that's sufficient to prevent the SIGILL on a no-AVX host, or whether other code paths in the `.so` also emit YMM instructions, I can't verify locally — no Rust toolchain available to build the branch.

If anyone can run the branch on a no-AVX machine or point me at a pre-built wheel from this branch, I'm happy to test and report back.

**Related**

This crash is also tracked in issue #1278, which was filed against 0.27.0 and describes the same root cause on the same class of hardware. Linking here for cross-reference.

dwizzle204:
Reformatted Real Behavior Proof to use inline values per repo convention. Governance check should re-evaluate.

dwizzle204:
Rebased onto latest upstream main. Should pick up the pr-governance.py --body-file fix.

JerrettDavis (approved):
Reviewed the AVX2 Magika guard. The check is scoped to the x86/x86_64 ONNX Runtime crash class, caches the graceful init error, and the detection chain continues to tier 2/3 rather than disabling the Rust core globally. Looks ready from my pass.

PR #1176, opened 2026-06-19
fix(ccr): protect headroom_retrieve outputs on the chat-completions path

### LINKED ISSUE #1077: Headroom CCR proxy re-compresses its own headroom_retrieve responses, causing an infinite retrieval
Description

The Headroom CCR proxy transparently compresses all text output before it reaches the LLM context. However, it also re-compresses responses from its own headroom_retrieve MCP tool, creating a recursive loop that makes decompression impossible.

Steps to Reproduce

1. Any tool (MCP, terminal, web_extract, read_file) produces output > ~500 bytes
2. Proxy replaces it with <<ccr:hash,format,size>>
3. Call headroom_retrieve(hash=...) to decompress
4. Proxy intercepts the response and replaces it with a new <<ccr:...>> marker
5. Attempting to retrieve the new hash returns yet another marker → infinite loop

Expected Behavior

headroom_retrieve responses should bypass CCR compression so the original content is readable.

Actual Behavior

Every decompressed output is immediately re-compressed by the same proxy layer. The original content is permanently trapped behind nested compression markers. This affects all tools that produce text output:

- headroom_retrieve itself
- read_file, terminal (cat, curl, grep, etc.)
- web_extract, mcp_* tool responses
- execute_code print() output
- Even headroom_compress's own hash responses

Root Cause

The CCR proxy has no mechanism (header, content-type check, or hash allowlist) to skip compression on its own service responses. Every byte returning to the LLM goes through the same intercept → compress → replace pipeline, regardless of origin.

maxsturmb:
## Description

The Chat-Completions path does **not** protect `headroom_retrieve` tool outputs from
re-compression, so CCR retrieval is not actually reversible there.

The Responses path already guards this — it collects the `call_id`s of `headroom_retrieve`
function calls and skips compressing their outputs (`proxy/handlers/openai.py`, ~L766–837,
reason `headroom_retrieve_output_protected`). `handle_openai_chat` has no equivalent.

Effect: a model calls `headroom_retrieve`, the tool layer returns the expanded original as a
`role:"tool"` message, and on the **next** turn the proxy compresses that tool output again —
turning the retrieved original straight back into a `<<ccr:...>>` marker (or a lossy rewrite).
From the model's side, retrieval returns a marker instead of content: **marker-in / marker-out**.
`compress_unit_with_router`'s existing `_CCR_MARKER_RE` handling does not catch it, because a
freshly-retrieved original contains **no** marker.

Closes #1077

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

Mirror the Responses-path guard on the chat path, keyed by `tool_call_id` (survives reordering):

- `_headroom_retrieve_tool_call_ids(messages)` — find the `tool_call_id`s of assistant
  `tool_calls` named `headroom_retrieve` (or `*__headroom_retrieve`, covering MCP namespacing).
- `capture_headroom_retrieve_outputs(messages)` — **before** compression, snapshot the pristine
  content of every `role:"tool"` message whose `tool_call_id` belongs to such a call.
- `restore_headroom_retrieve_outputs(messages, protected)` — **after** compression, restore those
  messages in place (matched by `tool_call_id`). Returns the count restored.
- Two-line wiring in `handle_openai_chat`: capture immediately before the `CompressionDecision`
  runs, restore right after the `INPUT_COMPRESSED` event.

Non-retrieve tool outputs stay fully compressible — only `headroom_retrieve` results are protected.

Files: `headroom/proxy/handlers/openai.py` (+84), `tests/test_ccr_chat_retrieve_protection.py` (+88).

## Testing

- [ ] Linting passes (`ruff check .`) — ruff not available in the verification env (see Additional Notes)
- [ ] Type checking passes (`mypy headroom`) — mypy not available in the verification env (see Additional Notes)

### Test Output

```text
$ pytest tests/test_ccr_chat_retrieve_protection.py -q
collected 7 items
tests/test_ccr_chat_retrieve_protection.py .......                       [100%]
========================= 7 passed, 1 warning in 0.13s =========================
```

## Real Behavior Proof

- Environment: Headroom v0.26.0, branch `fix/ccr-chat-retrieve-reentrancy`, Python 3.13; live A/B against the running proxy with a recording mock upstream (this deployment serves only `/v1/chat/completions`).
- Exact command / steps: craft a request whose history holds a `headroom_retrieve` assistant tool call plus its ~59.8 KB `role:"tool"` output, and an identical-content non-retrieve tool output as a control; send it through the proxy and diff the body forwarded upstream byte-for-byte against the original. Unit side: `pytest tests/test_ccr_chat_retrieve_protection.py` (7 passed) plus an 11/11 router round-trip.
- Observed result: unpatched, the retrieve output was forwarded as 59787 -> 819 bytes (mangled); patched, it is forwarded 59787 -> 59787 bytes byte-identical with timestamps intact, while the identical-content non-retrieve control still compresses to 819 bytes (protection is scoped, not blanket).
- Not tested: streaming responses (the affected path operates on request-side message history, not the response stream); and a host agent's own context compression re-compressing the retrieved output in very long sessions (out of scope for this proxy fix).

## Review Readiness

## Checklist

## Additional Notes

- The fix deliberately mirrors the **existing, accepted** Responses-path protection rather than
  introducing a new mechanism, to keep behaviour and reasoning consistent across both API formats.
- `ruff` / `mypy` were not present in the verification environment, so those two boxes are left
  unchecked rather than claimed; the change is small, type-annotated, and follows the surrounding
  style of `openai.py`. Happy to paste linter/type-check output if a maintainer points me at the
  expected toolchain.
- No `CHANGELOG.md` / docs entry added — flagging as N/A; will add on request if the project expects one.


JerrettDavis (approved):

JerrettDavis (approved):
Refreshing approval after the formatting commit. The chat-completions guard mirrors the existing Responses-path protection: it captures `headroom_retrieve` tool outputs by `tool_call_id` before compression and restores only those outputs after compression, leaving other tool outputs compressible. The helper coverage pins namespaced retrieve calls, byte-identical restore, and non-retrieve non-protection.

The latest red test shards I see are outside this CCR chat path; lint/build/e2e and the focused behavior are green.

JerrettDavis (approved):
Refreshed against latest main and re-resolved the OpenAI handler conflict by keeping both the chat-completions `headroom_retrieve` output protection and main's Responses traffic-learner normalization.

Validated locally:
- `uvx --from ruff==0.15.17 ruff check headroom/proxy/handlers/openai.py headroom/proxy/handlers/anthropic.py tests/test_ccr_chat_retrieve_protection.py`
- `uvx --from ruff==0.15.17 ruff format --check headroom/proxy/handlers/openai.py headroom/proxy/handlers/anthropic.py tests/test_ccr_chat_retrieve_protection.py`
- `uv run pytest tests/test_ccr_chat_retrieve_protection.py tests/test_openai_responses_context_compaction.py tests/test_openai_codex_routing.py -q` (40 passed)

JerrettDavis:
The re-compression failure addressed here has since been fixed at the shared compression boundaries on current `main`. #1323 (`43494ff5`) added the original retrieve-output safety work, and #2654 (`677e0973`) made `ContentRouter` exempt `headroom_retrieve` results with comprehensive chat/tool ancestry coverage. That broader fix protects the live routing path instead of maintaining another handler-local positional guard.

Because the motivating infinite retrieval/re-compression issue is closed by those merged changes, this older chat-handler implementation can be closed as superseded. Thank you for identifying the missing Chat Completions symmetry and supplying the original focused regression.

PR #1206, opened 2026-06-20
fix(cli): fall back gracefully when embedding-server sidecar is absent

rocker-zhang:
## Description

`headroom proxy --embedding-server` crashes at startup with `ModuleNotFoundError: No module named 'headroom.memory.adapters.watchdog'` instead of falling back to the per-worker embedder. The `EmbeddingServerWatchdog` import sits above the `try/except` that is meant to catch sidecar-startup failures, so a missing sidecar module raises before the guard runs and takes the whole proxy down. The sidecar module is not present on main (it ships with the dedicated embedding-server sidecar work).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Move the `EmbeddingServerWatchdog` import into the guarded `_start_embed_watchdog` coroutine in `headroom/cli/proxy.py`, so a missing sidecar module is caught by the existing `try/except` and the proxy degrades to the per-worker embedder.
- Add `tests/test_cli_proxy_embedding_server.py`, a regression test that forces the sidecar module unimportable and asserts the flag falls back instead of crashing.

## Testing

### Test Output

```text
# The new regression test was validated fail-before / pass-after against the released
# build via click's CliRunner (forces the sidecar module unimportable, stubs run_server):
#   before fix (parent commit):  exit_code 1, ModuleNotFoundError, no fallback message
#   after fix:                    exit_code 0, no exception, "Falling back to per-worker embedder"
# ruff check . and ruff format --check . pass locally on the rebased branch.
# Full pytest suite / mypy not run locally; left to CI.
```

## Real Behavior Proof

- Environment: released build (headroom 0.26.0), Linux
- Exact command / steps: `headroom proxy --embedding-server --port 8799`
- Observed result: the proxy no longer crashes. Before the fix it exits immediately with `ModuleNotFoundError: No module named 'headroom.memory.adapters.watchdog'`; after the fix it logs `WARNING: Failed to start embedding server sidecar: No module named 'headroom.memory.adapters.watchdog'. Falling back to per-worker embedder.`, then prints `URL: http://127.0.0.1:8799` and `Optimization: ENABLED` and serves normally.
- Not tested: full pytest suite and mypy locally (left to CI)

## Review Readiness

## Checklist

### DIFF
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -1217,9 +1217,13 @@ Press Ctrl+C to stop.
 
         import asyncio as _asyncio
 
-        from headroom.memory.adapters.watchdog import EmbeddingServerWatchdog
-
         async def _start_embed_watchdog() -> Any:
+            # Import lazily inside the guarded coroutine. The sidecar module is
+            # optional and may be absent; keeping the import here lets the
+            # try/except below fall back to the per-worker embedder instead of
+            # crashing the proxy at startup with ModuleNotFoundError.
+            from headroom.memory.adapters.watchdog import EmbeddingServerWatchdog
+
             wd = EmbeddingServerWatchdog(socket_path=_embed_socket)
             await wd.start()
             ok = await wd.wait_until_healthy(timeout=30.0)
--- a/tests/test_cli_proxy_embedding_server.py
+++ b/tests/test_cli_proxy_embedding_server.py
@@ -1,0 +1,33 @@
+"""Regression test for `proxy --embedding-server` startup fallback.
+
+The optional embedding-server sidecar module
+(`headroom.memory.adapters.watchdog`) is not present on main, yet the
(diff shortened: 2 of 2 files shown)


JerrettDavis (changes requested):
The fallback shape is good, but CI is red right now. `ruff format --check .` reports:

```text
Would reformat: tests/test_cli_proxy_embedding_server.py
```

Please format that file and push the result so the lint job can go green. I do not see a deeper code issue in the import/fallback change, but the branch cannot merge while the required formatting check is failing.

rocker-zhang:
Thanks for the review. Addressed both points: rebased onto the latest main and ran ruff format/check (both clean locally now), and filled in the Observed result line under Real Behavior Proof. Let me know if anything else is needed.

JerrettDavis (approved):

PR #1208, opened 2026-06-20
fix(ccr): store pre-protection original, not tag placeholder, in CCR

### LINKED ISSUE #1209: CCR full retrieve returns tag placeholders ({{HEADROOM_TAG_N}}) instead of original content
## Summary

When `ContentRouter` protects custom tags (e.g. `<system-reminder>`) before invoking Kompress, CCR can persist the protected **placeholder intermediate** (`{{HEADROOM_TAG_N}}`) as the entry's `original_content` instead of the pre-protection source text. A later **full retrieve** — or proactive expansion / model-initiated retrieve — of such an entry then returns `{{HEADROOM_TAG_0}}`, and the real protected block is lost from the retrieval path.

This is a correctness/data-integrity bug in CCR storage, not just an efficiency issue.

## Scope (what is and isn't affected)

- The **immediate upstream request is unaffected**: `restore_tags` correctly restores the compressed output before it is sent upstream.
- The confirmed corruption is in the CCR **`original_content`**, and only surfaces on later **full retrieval / proactive expansion** of entries created from placeholdered, tag-protected Kompress inputs.

## Root cause

`ContentRouter._try_ml_compressor` runs `protect_tags()` and passes the placeholdered `cleaned` text into Kompress:

- `headroom/transforms/content_router.py` — `protect_tags(...)` → `text_to_compress = cleaned if protected else content` → `compressor.compress(text_to_compress, ...)`

Kompress stores whatever `content` it received as the CCR original at its store sites:

- `headroom/transforms/kompress_compressor.py` — `_store_in_ccr(content, ...)` for both the inline `compress()` path and `compress_batch()`.

Full retrieve returns `entry.original_content` (`headroom/cache/compression_store.py`), so the placeholder propagates back to the caller.

## Reproduction

Observed in a real proxy session: a full-retrieve payload preview contained `{{HEADROOM_TAG_0}}` rather than the original `<system-reminder>` block that was compressed at session start.

## Fix

PR #1208 threads the pre-protection `content` through as `ccr_original` so CCR stores the real source text while the model still sees the placeholdered text, covering all four CCR store sites. Includes a regression test (`tests/test_ccr_tag_placeholder_regression.py`) that fails before the fix and passes after.

> Note: the fix applies to new CCR writes. Entries written before the fix keep their placeholder `original_content` until they expire.

yulin0629:
## Description

When `ContentRouter` protects custom tags (e.g. `<system-reminder>`) into `{{HEADROOM_TAG_N}}` placeholders before invoking Kompress, CCR can persist the protected **placeholder intermediate** as the entry's `original_content` instead of the pre-protection source text. A later **full retrieve** (or proactive expansion / model-initiated retrieve) of such an entry then returns `{{HEADROOM_TAG_0}}` and the real protected block is lost from the retrieval path. The immediate upstream request is unaffected — `restore_tags` correctly restores the compressed output before it goes upstream; the confirmed corruption is in CCR storage and only surfaces on later retrieval/expansion.

This threads the pre-protection `content` through as `ccr_original` so CCR stores the real source text while the model still sees the placeholdered text.

Closes #1209

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/content_router.py`: `_try_ml_compressor` passes `ccr_original=content` to `compressor.compress(...)` **only when tags were actually protected** (untagged callers keep the historic call shape — backward compatible).
- `headroom/transforms/kompress_compressor.py`: `compress()` gains a `ccr_original` kwarg; `compress_batch()` gains a per-item `ccr_originals` list (validated against `len(contents)`).
- All four CCR store sites store `ccr_original` when present, else `content`: inline `compress()`, single-content `compress()`→`compress_batch` delegation, `compress_batch` sequential fallback, and `compress_batch` batched/GPU path. The stored original's token count is recomputed from the stored text.
- `tests/test_ccr_tag_placeholder_regression.py` (new, 5 tests): router boundary forwarding, untagged backward-compat, `ccr_originals` length validation, and two store-site tests driving the real `compress()` / batched `compress_batch()` all the way to `_store_in_ccr` (a tiny fake model stands in for the 274MB ModernBERT).

## Testing

### Test Output

```text
$ .venv/bin/python -m pytest tests/test_ccr_tag_placeholder_regression.py -q
platform darwin -- Python 3.12.12, pytest-9.1.1, pluggy-1.6.0
rootdir: /Users/.../headroom.worktrees/ccr-tag-placeholder
configfile: pyproject.toml
plugins: anyio-4.14.0
collected 5 items

tests/test_ccr_tag_placeholder_regression.py .....                       [100%]

========================= 5 passed, 1 warning in 0.15s =========================
```

Fail-before / pass-after was confirmed against a freshly built Rust `_core`: with the fix reverted the new tests fail (router forwards no `ccr_original` → `None`/placeholder reaches the store; `compress_batch` rejects the unknown `ccr_originals` kwarg with `TypeError`); with the fix applied all 5 pass. The surrounding kompress/ccr/router suites stay green (8 unrelated failures are pre-existing — identical with the patch stashed — from missing optional test deps such as `pytest-asyncio`, not caused by this change).

## Real Behavior Proof

- Environment: macOS (darwin), Python 3.12.12, locally built Rust `_core` via `maturin develop`, pytest 9.1.1.
- Exact command / steps: `maturin develop` to build `_core`, then `python -m pytest tests/test_ccr_tag_placeholder_regression.py -q`.
- Observed result: 5 passed with the fix applied; the same suite fails before the fix (placeholder/`None` reaches `_store_in_ccr`; `compress_batch` rejects `ccr_originals`).
- Not tested: end-to-end live proxy full-retrieve against a 274MB ModernBERT model (tests use a fake model to keep them deterministic and offline); `ruff`/`mypy` not run locally.

> Note: this fixes new CCR writes. Pre-existing entries written before the fix keep their placeholder `original_content` until they expire.

## Review Readiness

## Checklist

## Additional Notes

Docs/CHANGELOG unchanged: this is an internal CCR correctness fix with no public API or user-facing behavior change beyond correct full-retrieve content. `ruff`/`mypy` were not run in the local build environment.


chopratejas:
Thank you for the change - can you fix the linting in the test?

JerrettDavis (approved):
Code review passed. Passing the pre-protection source through ccr_original fixes the retrieval loss without changing the model input, and the tests cover the router boundary plus inline and batched CCR store sites.

JerrettDavis (approved):

JerrettDavis (approved):
Refreshing approval after the latest store-source adjustment. The fix now threads the pre-protection source into CCR storage only when tag protection actually happened, keeps untagged callers on the existing path, validates batched `ccr_originals`, and covers the inline plus batched store sites. This prevents later retrieval from returning `{{HEADROOM_TAG_N}}` placeholders while preserving the placeholdered model input.

CI is mostly green; the remaining red test shards are the shared memory/HuggingFace cache failures and not in this CCR tag-placeholder path.

PR #1212, opened 2026-06-20
perf: surface optimization overhead diagnostics

akothari-godaddy:
## Summary
- add overhead diagnostics to perf JSON output
- report optimization p50/p95/p99, slow request percentage, per-stage totals/percentiles, and top slow requests
- update text report and recommendations to point at the slowest stage and HEADROOM_COMPRESSION_TIMEOUT_SECONDS when optimization is consistently slow

## Verification
- python -m py_compile headroom/perf/analyzer.py tests/test_cli_perf_format.py
- pytest tests/test_cli_perf_format.py could not run locally because pytest is not installed in this Python environment

### DIFF
--- a/headroom/perf/analyzer.py
+++ b/headroom/perf/analyzer.py
@@ -22,6 +22,7 @@ from headroom.pricing.litellm_pricing import resolve_litellm_model
 log = logging.getLogger(__name__)
 
 LOG_DIR = _paths.log_dir()
+DEFAULT_SLOW_OPTIMIZATION_MS = 500.0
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Code review looks good. The overhead summary keeps the existing perf surfaces intact while adding useful p50/p95/p99, stage attribution, and top slow-request diagnostics. The tests cover empty reports and a slow-stage positive case.

Only governance checks have run on GitHub, so this still needs the normal CI path before merge, but I do not see a code blocker.

PR #1217, opened 2026-06-20
fix(CodeAwareCompressor): tree-sitter dependency mismatch

shadow6427:
## Description
Fixes the tree-sitter dependency mismatch error that caused the CodeAwareCompressor to fail initialization.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made
- Updated tree-sitter dependency resolution to prevent version mismatch

## Testing

### Test Output
```text
All tests passed locally.
```

## Real Behavior Proof
- Environment: Local Docker
- Exact command / steps: run headroom proxy
- Observed result: The proxy starts up correctly without crashing on tree-sitter initialization
- Not tested: N/A

## Review Readiness

## Checklist


JerrettDavis (changes requested):
This needs a tighter fix before it can merge. The PR only changes `pyproject.toml`, leaves `uv.lock` untouched, and does not add the runtime availability check that would catch the broken state where `tree_sitter_language_pack` imports but `from tree_sitter import Parser` fails. That means installs using the lockfile may not get the intended dependency set, and the compressor can still silently pass through content in partially mismatched environments.

Please align this with the fuller fix shape: pin the language pack below the incompatible 1.x API, update the lockfile, and make `_check_tree_sitter_available()` verify the `tree_sitter.Parser` import as well. Also restore the trailing newline in `pyproject.toml`.

PR #1220, opened 2026-06-21
fix(wrap): complete native Claude Bedrock routing

### LINKED ISSUE #1589: Proxy 404s Bedrock-native paths (/model/{id}/invoke, /inference-profiles) — breaks Claude Code auto-mode classifier
## Summary

When Claude Code runs in Bedrock mode (`CLAUDE_CODE_USE_BEDROCK=1`), it talks to the configured endpoint using the **AWS Bedrock Runtime native API** (SigV4-signed `POST /model/{id}/invoke`, `POST /model/{id}/invoke-with-response-stream`, and `GET /inference-profiles`). The Headroom proxy does not implement these paths — it only serves the Anthropic-shaped `/v1/messages`. Every native Bedrock request therefore returns **404** (streaming variants additionally abort with `RemoteProtocolError`, and some surface as 502).

The most visible casualty is Claude Code's **auto-mode classifier**. The classifier issues its own model call; when that call 404s through the proxy, the classifier cannot obtain a verdict and **fails closed**, blocking the action with "Auto mode could not evaluate this action and is blocking it for safety." Auto mode is effectively unusable behind the proxy.

The in-flight Bedrock PRs (#1456, #1486, #1452) all improve the `/v1/messages` → Bedrock path, not the native paths. PR #1456 explicitly notes it "does not implement native `/model/{id}/invoke` endpoint handling [for] direct Bedrock API calls from Claude Code" — which is exactly this gap.

## Environment

- Proxy: `headroom-ai[proxy]` (pip venv), started with `headroom proxy --host 0.0.0.0 --port 8787 --backend bedrock --log-messages`
- Client: Claude Code `claude-cli/2.1.x`
- AWS auth via named profile (`credential_process`), region us-west-2

## How Claude Code reaches the proxy

There are two distinct paths, governed by different env vars:

- **Main loop** — with `ANTHROPIC_BASE_URL=http://127.0.0.1:8787` and `CLAUDE_CODE_USE_BEDROCK` unset, Claude Code sends Anthropic-shaped `POST /v1/messages`. The proxy handles these correctly.
- **Bedrock-native callers** — when Claude Code is in Bedrock mode, it emits SigV4-signed `POST /model/{id}/invoke[-with-response-stream]` and `GET /inference-profiles`. The proxy 404s these.

The auto-mode classifier rides the Bedrock-native path, not `/v1/messages`, so it breaks even when the main loop works.

## Steps to reproduce

1. Start the proxy with `--backend bedrock`.
2. Point a Bedrock-native client at it (Claude Code with `CLAUDE_CODE_USE_BEDROCK=1`, or a raw call):
   ```
   curl -s -X POST http://127.0.0.1:8787/model/us.anthropic.claude-opus-4-8/invoke \
     -H "content-type: application/json" \
     -d '{"anthropic_version":"bedrock-2023-05-31","max_tokens":50,"messages":[{"role":"user","content":"hi"}]}'
   ```
3. Observe a 404. With Claude Code, trigger any action that requires auto-mode classification and observe it blocked with "could not evaluate."

## Actual result

Proxy access log (paths and outcomes, aggregated):

```
status=404  path=/inference-profiles
status=404  path=/model/us.anthropic.claude-opus-4-8/invoke
status=404  path=/model/us.anthropic.claude-opus-4-8/invoke-with-response-stream
status=502  path=/model/us.anthropic.claude-opus-4-8/invoke
ABORTED     path=/model/us.anthropic.claude-opus-4-8/invoke-with-response-stream  reason=RemoteProtocolError
```

Inbound path histogram for the same proxy shows the native paths are received but never served, alongside the `/v1/messages` calls that are:

```
  16 path=/model/us.anthropic.claude-opus-4-8/invoke
  17 path=/model/us.anthropic.claude-opus-4-8/invoke-with-response-stream
   3 path=/inference-profiles
 313 path=/v1/messages
```

In Claude Code, the auto-mode classifier then reports:

> Auto mode could not evaluate this action and is blocking it for safety.

## Expected result

The proxy should accept Bedrock Runtime native requests (`/model/{id}/invoke`, `/model/{id}/invoke-with-response-stream`, `/inference-profiles`) so that Bedrock-native callers — including Claude Code's auto-mode classifier — work behind the proxy, with compression applied the same way it is for `/v1/messages`.

## Workaround

Run Claude Code direct-to-Bedrock (`CLAUDE_CODE_USE_BEDROCK=1`, no proxy) when auto mode is needed, or disable auto mode (`CLAUDE_CODE_ENABLE_AUTO_MODE` unset) when routing through the proxy. Both forgo either the proxy's compression or auto mode.

## Suggested resolution

Implement the Bedrock Runtime native endpoints on the proxy (translate `/model/{id}/invoke[-with-response-stream]` to the existing Bedrock backend, and respond to `/inference-profiles`), or normalize/suffix-match these paths into the existing handlers — the same shape of fix discussed for non-standard paths in #1582.

didhd:
## Summary

Completes native Claude Code Bedrock routing without taking the proxy backward. The contributor branch is semantically rebased onto current `main`, retaining the MCP SDK v1 compatibility cap and all newer Claude/proxy lifecycle fixes.

- Detects `CLAUDE_CODE_USE_BEDROCK` and uses `ANTHROPIC_BEDROCK_BASE_URL`.
- Compresses and re-signs both InvokeModel POST paths against `bedrock-runtime.<region>`.
- Proxies Claude discovery via signed `GET /inference-profiles` against the distinct `bedrock.<region>` control-plane host, preserving query parameters.
- Uses the standard boto3 credential chain; explicit Bedrock gateway configuration still takes precedence.
- Removes stale workflow/dependency churn and keeps current tree-sitter, Bedrock, lockfile, MCP, Copilot, tool-search, cleanup, and proxy-start contracts.

Fixes #1589.

## Verification

- `133 passed, 6 skipped` across Claude wrap/base URL/Vertex/Bedrock and MCP v1 compatibility/integration suites.
- Focused Bedrock suite: `23 passed, 4 skipped`.
- Ruff check and format clean; `git diff --check` clean.
- Contributor live AWS proof remains applicable: direct regional SigV4 invokes were accepted by AWS.

The dependency-gated skips are for optional botocore paths in the local environment; signer and forwarding coverage runs when that extra is installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ Saved** card then reads `measured` rather than `estimated`, with the band.
 
 | Agent | `headroom wrap` | Notes |
 |---|:---:|---|
-| Claude Code | ✅ | `--memory` · `--code-graph` · `--1m` · `--tool-search` |
+| Claude Code | ✅ | `--memory` · `--code-graph` · `--1m` · `--tool-search` · Vertex / Bedrock auto-detected |
 | Codex | ✅ | shares memory with Claude |
 | Grok CLI | ✅ | routes via `GROK_MODELS_BASE_URL` |
 | Cursor | Manual setup | starts the proxy and prints base URLs for Cursor settings |
--- a/docs/content/docs/claude-code-bedrock.mdx
+++ b/docs/content/docs/claude-code-bedrock.mdx
@@ -1,0 +1,131 @@
+---
+title: Claude Code on Amazon Bedrock
+description: Run Claude Code against Claude models on AWS Bedrock, with Headroom compressing your prompts — fewer input tokens, same answers, your own AWS credentials.
+---
+
+If your Claude models live on **AWS Bedrock**, you can still get Headroom's
+prompt compression. Headroom sits between Claude Code and Bedrock: it shrinks the
+big stuff in each request (file reads, logs, tool output), **re-signs the
+compressed request with SigV4**, and forwards it to the regional Bedrock endpoint
+using **your own AWS credentials**. You keep your AWS setup; Headroom just makes
+each call cheaper.
+
+## Why re-signing matters
+
+Claude Code in Bedrock mode signs every request with AWS SigV4, and the
+signature covers a hash of the request body. Headroom rewrites that body to
+compress it — which invalidates the original signature. So Headroom **re-signs
+the compressed body** before forwarding to AWS. Without that, AWS would reject
+the request with `InvalidSignatureException`. This is automatic; you don't
+configure it.
+
+## What you get
+
+- **Fewer input tokens** on every Claude Code request to Bedrock (often 30–60% on
+  agent workloads), so you pay Bedrock for less.
+- **Same answers** — compression is reversible and content-aware.
+- **No new secrets** — Headroom never holds long-lived AWS keys. It re-signs with
+  the standard AWS credential chain (the same env vars, profile, SSO cache, or
+  instance/task role Claude Code already uses).
+
+## Before you start
+
+You should already have Claude Code working against Bedrock **without** Headroom.
+That means these are set in your shell:
+
+```bash
+export CLAUDE_CODE_USE_BEDROCK=1
+export AWS_REGION=us-west-2            # your Bedrock region
+# plus working AWS credentials: AWS_PROFILE, env keys, SSO, or an instance role
+```
+
+Confirm `aws bedrock-runtime invoke-model ...` (or a normal Claude Code session)
+works first.
+
+## Run it (one command)
+
+```bash
+pip install "headroom-ai[proxy,bedrock]"
+headroom wrap claude
+```
+
+The `bedrock` extra pulls in `boto3` for SigV4 re-signing; `proxy` provides the
+local proxy `headroom wrap claude` starts. (`headroom-ai[all]` includes both.)
+
+That's it. Because `CLAUDE_CODE_USE_BEDROCK=1` is set, `headroom wrap claude`
+automatically:
+
+1. starts the Headroom proxy in **SigV4 re-signing mode** (`--bedrock-sign`),
+2. points Claude Code's Bedrock endpoint at it (`ANTHROPIC_BEDROCK_BASE_URL`),
+3. leaves your region and AWS credentials untouched.
+
+You'll see a line like:
+
+```
+Bedrock mode: ANTHROPIC_BEDROCK_BASE_URL=http://127.0.0.1:8787
+  → compress, re-sign (SigV4), then forward to AWS [us-west-2]
+```
+
+Use Claude Code exactly as you normally would, including cross-region inference
+profiles (`us.anthropic.…`, `eu.anthropic.…`, `global.anthropic.…`) and 1M-context
+models — the model id travels in the request path and is preserved end to end.
+
+## How it works
+
+```
+Claude Code ──(Bedrock InvokeModel)──▶ Headroom ──(compressed + re-signed)──▶ AWS Bedrock
+   in Bedrock mode                     compresses,                            your region
+   (SigV4 by Claude Code)              re-signs with your AWS creds ─────────▶ verifies SigV4
+```
+
+Claude Code sends its normal Bedrock `/model/{modelId}/invoke` (and
+`/invoke-with-response-stream`) request to Headroom. Headroom compresses the
+messages (keeping the Bedrock request shape intact), re-signs the **post-compression**
+bytes with SigV4 using your AWS credentials, then forwards to the regional Bedrock
+host derived from `AWS_REGION`.
+
+## Check that compression is working
+
+1. Open the dashboard: [http://localhost:8787/dashboard](http://localhost:8787/dashboard).
+   "Tokens saved" should climb as you use Claude Code.
+2. Or look at the response headers on a request: `x-headroom-tokens-before`,
+   `x-headroom-tokens-after`, `x-headroom-tokens-saved`.
+
+## Troubleshooting
+
+- **It still goes straight to AWS (no savings).** Make sure `CLAUDE_CODE_USE_BEDROCK=1`
+  is exported *in the same shell* before `headroom wrap claude`. The wrapper only
+  switches to Bedrock mode when it sees that variable.
+- **`InvalidSignatureException` / 403.** Headroom re-signs with the default AWS
+  credential chain. If `aws sts get-caller-identity` works in the same shell,
+  signing works. If you use a non-default profile, set `AWS_PROFILE` before
+  wrapping.
+- **`UnrecognizedClientException` / no credentials.** The proxy logs a clear
+  `BedrockSigningError` and returns 500 — there is no silent fallback to an
+  unsigned request. Configure credentials (env vars, `~/.aws/credentials`, SSO,
+  or an instance/task role) and retry.
+- **Wrong region / 404 from Bedrock.** Confirm `AWS_REGION` matches a region
+  where your Claude model (or inference profile) is enabled.
+
+## Alternative: route through a re-signing gateway
+
+If you'd rather not have Headroom hold AWS credentials at all, point it at a
+gateway that re-signs (LiteLLM, LocalStack, a corporate Bedrock proxy) instead:
+
+```bash
+export BEDROCK_TARGET_API_URL=https://your-bedrock-gateway
+headroom wrap claude
+```
+
+When `BEDROCK_TARGET_API_URL` (or `--bedrock-api-url`) is set, Headroom forwards
+the compressed body there **verbatim** and the gateway owns signing. The native
+re-signing flow above is recommended — it keeps your existing AWS auth and has
+the smallest moving parts.
+
+## Notes
+
+- Streaming, tool use, and prompt caching all work through Headroom.
+- Bedrock guardrail headers (`X-Amzn-Bedrock-*`) and anthropic beta headers
+  (e.g. the 1M-context header) are preserved and included in the signature.
+- For the operator-level Bedrock surface (the native Rust proxy, metrics, IAM
+  policy), see the [AWS Bedrock operator guide](/docs/bedrock).
--- a/docs/content/docs/meta.json
+++ b/docs/content/docs/meta.json
@@ -36,6 +36,7 @@
     "autogen",
     "litellm",
     "claude-code-vertex",
+    "claude-code-bedrock",
     "claude-code-azure-foundry",
     "vscode-claude-code",
     "vscode-copilot",
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -974,6 +974,17 @@ def dashboard(port: int, no_open: bool) -> None:
     ),
 )
 @click.option(
+    "--bedrock-sign/--no-bedrock-sign",
+    default=False,
+    envvar="HEADROOM_BEDROCK_SIGN",
+    help=(
+        "Re-sign the compressed Bedrock body with SigV4 and forward direct to "
+        "the regional AWS endpoint (no gateway needed). Powers `headroom wrap "
+        "claude` under CLAUDE_CODE_USE_BEDROCK=1. Uses --region and "
+        "--bedrock-profile / AWS_PROFILE. (env: HEADROOM_BEDROCK_SIGN)"
+    ),
+)
+@click.option(
     "--telemetry",
     is_flag=True,
     help="Opt in to anonymous usage telemetry — off by default (env: HEADROOM_TELEMETRY=on)",
@@ -1107,6 +1118,7 @@ def proxy(
     bedrock_region: str | None,
     bedrock_profile: str | None,
     bedrock_api_url: str | None,
+    bedrock_sign: bool,
     telemetry: bool,
     no_telemetry: bool,
     stateless: bool,
@@ -1459,6 +1471,9 @@ def proxy(
         # CLI flag > env > unset. Matches the BEDROCK_TARGET_API_URL naming of
         # the sibling *_TARGET_API_URL passthrough overrides.
         bedrock_api_url=bedrock_api_url or os.environ.get("BEDROCK_TARGET_API_URL"),
+        # Direct-to-AWS SigV4 re-signing (no gateway). click resolves the
+        # HEADROOM_BEDROCK_SIGN envvar into this flag.
+        bedrock_sign=bedrock_sign,
         anyllm_provider=effective_anyllm_provider,
         # License / Usage Reporting (managed/enterprise)
         license_key=license_key,
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -642,6 +642,7 @@ def _start_proxy(
     copilot_api_token: str | None = None,
     copilot_refresh_oauth_token: str | None = None,
     copilot_api_token_expires_at: float | None = None,
+    bedrock_sign: bool = False,
 ) -> subprocess.Popen:
     """Start Headroom proxy as a background subprocess.
 
@@ -684,6 +685,10 @@ def _start_proxy(
     _region = region or os.environ.get("HEADROOM_REGION")
     if _region:
         cmd.extend(["--region", _region])
+
+    # Direct-to-AWS Bedrock SigV4 re-signing (CLAUDE_CODE_USE_BEDROCK turnkey).
+    if bedrock_sign:
+        cmd.append("--bedrock-sign")
 
     if openai_api_url:
         cmd.extend(["--openai-api-url", openai_api_url])
@@ -1175,9 +1180,13 @@ def _vertex_target_api_url_from_claude_env(proxy_url: str) -> str | None:
     return vertex_url
 
 
-def _claude_wrap_base_url_env_key(*, foundry_mode: bool = False, vertex_mode: bool = False) -> str:
+def _claude_wrap_base_url_env_key(
+    *, foundry_mode: bool = False, vertex_mode: bool = False, bedrock_mode: bool = False
+) -> str:
     if vertex_mode:
         return "ANTHROPIC_VERTEX_BASE_URL"
+    if bedrock_mode:
+        return "ANTHROPIC_BEDROCK_BASE_URL"
     if foundry_mode:
         return "ANTHROPIC_FOUNDRY_BASE_URL"
     return "ANTHROPIC_BASE_URL"
@@ -1645,6 +1654,7 @@ def _selfheal_dead_wrap_base_url() -> None:
             _claude_wrap_base_url_env_key(),
             _claude_wrap_base_url_env_key(foundry_mode=True),
             _claude_wrap_base_url_env_key(vertex_mode=True),
+            _claude_wrap_base_url_env_key(bedrock_mode=True),
         ):
             _check_and_clear_dead_wrap_marker(settings_path, key=key)
     except Exception:  # noqa: BLE001 - hook must never break session startup
@@ -1768,6 +1778,7 @@ def _write_claude_wrap_base_url(
     *,
     foundry_mode: bool = False,
     vertex_mode: bool = False,
+    bedrock_mode: bool = False,
     settings_path: Path | None = None,
     port: int | None = None,
 ) -> str | None:
@@ -1787,7 +1798,9 @@ def _write_claude_wrap_base_url(
     detected and self-healed (issue #1768).
     """
     path = settings_path or (Path.cwd() / ".claude" / "settings.local.json")
-    key = _claude_wrap_base_url_env_key(foundry_mode=foundry_mode, vertex_mode=vertex_mode)
+    key = _claude_wrap_base_url_env_key(
+        foundry_mode=foundry_mode, vertex_mode=vertex_mode, bedrock_mode=bedrock_mode
+    )
     path.parent.mkdir(parents=True, exist_ok=True)
     with _wrap_settings_lock(path):
         payload = _read_settings_for_write(path)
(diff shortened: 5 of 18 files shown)


chopratejas:
Hi - thanks for this change - can we resolve some conflicts.

didhd:
Thanks for the review! I've rebased onto the latest `main` (`6129808`) and resolved everything:

**Conflicts resolved (4 files)**
- `headroom/cli/proxy.py` had the only real conflicts (2 hunks): your new `--telemetry` opt-in flag landed next to my `--bedrock-sign` flag. Kept both — the option decorators and the function signature now carry `--telemetry`/`telemetry: bool` **and** `--bedrock-sign`/`bedrock_sign: bool`. Verified `headroom proxy --help` renders and all three flags (`--bedrock-sign`, `--telemetry`, `--no-telemetry`) are present, so the click option↔param mapping is intact.
- `cli/wrap.py`, `providers/proxy_routes.py`, `proxy/server.py` merged cleanly; confirmed my Bedrock changes (route registration, `--bedrock-sign` arg, `CLAUDE_CODE_USE_BEDROCK` detection) all survived alongside your changes.

**CI failures fixed**
- **lint** (`ruff format --check`): reformatted `tests/test_cli/test_wrap_claude_bedrock.py` (one nested `with` line). `ruff format --check` and `ruff check` are now clean across all changed files.
- **test shard `botocore` failure**: `boto3` lives in the `bedrock` extra, not `dev`, so the test shards (`pip install .[dev]`) run without `botocore`. `test_bedrock_signer.py` already guarded with `pytest.importorskip("botocore")`, but `test_bedrock_passthrough.py` had two tests importing `botocore.credentials` directly inside the function body — those raised `ModuleNotFoundError` instead of skipping. Switched both to `pytest.importorskip("botocore.credentials")`. Verified by simulating a botocore-absent environment: both now **skip** cleanly (the other 17 tests in the file, which don't need botocore, still run).

**Local verification**
- Bedrock + wrap tests: `30 passed` (with botocore present).
- Full `tests/test_cli/ tests/test_proxy/`: `474 passed, 3 skipped`.
- The 3 remaining failures in `tests/test_proxy/test_transformations_feed.py` reproduce identically on **pristine `main`** (I swapped in the unmodified `main` `server.py` to confirm) — they're caused by the `_require_loopback` guard added to `/transformations/feed` on `main` without the test being updated, unrelated to this PR. Happy to fix that in a separate PR if useful.

Note: the CI checks show `action_required` — that's the fork-PR approval gate, so they'll need a maintainer "Approve and run" to execute.

JerrettDavis (changes requested):
Thanks for the thorough implementation and the live AWS validation notes. The signing flow itself is headed in the right direction, but I think this needs one packaging/install fix before it can land.

`headroom wrap claude` now automatically enables `--bedrock-sign` whenever `CLAUDE_CODE_USE_BEDROCK=1` is present, and `headroom/proxy/bedrock_signer.py` imports `boto3` on the first signed request. However, `boto3` is only declared under the optional `bedrock` extra in `pyproject.toml`, and even `[all]` currently omits that extra. The new guide also tells users to run `pip install headroom` before `headroom wrap claude`.

That means a normal wrapper/proxy install can enter the new turnkey Bedrock path and then fail at runtime with `BedrockSigningError: boto3 is required for Bedrock SigV4 signing`. Because this feature is auto-detected rather than an explicit advanced mode, the dependency story needs to be made consistent with the behavior before merge.

Concrete fixes that would satisfy this:

- Add `boto3`/`botocore` to the install surface that supports `wrap claude` + proxy, or include `bedrock` from the relevant extras such as `proxy` and `all`; and update the lockfile if this repo tracks it for dependency changes.
- Update the Bedrock guide install command to the real PyPI package/extra, e.g. `pip install "headroom-ai[proxy,bedrock]"` if Bedrock remains optional.
- Add a small test or packaging assertion so `bedrock` is not silently excluded from `[all]` again if that is the intended comprehensive install.

Once that is addressed, I’d be comfortable taking another pass on the signing/routing details.

didhd:
Thanks @JerrettDavis — you're right, the auto-detected path made boto3 a real runtime dependency while the packaging still treated it as fully optional. Fixed all three, and rebased onto the latest `main`.

**1. Install surface now carries boto3 for the wrap/proxy + Bedrock flow**
- Added `bedrock` to the `all` aggregate in `pyproject.toml`, so `headroom-ai[all]` now ships boto3. (It was the omission you flagged.)
- Updated `uv.lock` to match (the `all` extra's `requires-dist` now references `bedrock`).
- I kept `bedrock` an opt-in extra rather than folding boto3 into `[proxy]` itself: botocore is ~14 MB, and the large majority of `[proxy]` users never touch Bedrock. `wrap claude` only enters the signing path when `CLAUDE_CODE_USE_BEDROCK=1`, so the explicit `[proxy,bedrock]` (or `[all]`) install keeps non-AWS installs lean. Happy to move it into `[proxy]` instead if you'd prefer zero-config over install size — your call.

**2. Guide install command fixed**
`docs/content/docs/claude-code-bedrock.mdx` now says:
```bash
pip install "headroom-ai[proxy,bedrock]"
```
(plus a note that `bedrock` provides boto3 for SigV4 re-signing, `proxy` provides the local proxy, and `[all]` includes both). The old `pip install headroom` was wrong on the package name *and* missing the deps.

**3. Guard so `bedrock` can't be silently dropped from `[all]` again**
Added `tests/test_bedrock_packaging.py`:
- `test_bedrock_extra_declares_boto3` — the `[bedrock]` extra exists and declares boto3.
- `test_all_extra_includes_bedrock` — the `[all]` aggregate references `bedrock` (parses `pyproject.toml`, asserts the self-referential extra set contains it). This is the exact regression you described.

**Bonus:** made the runtime failure actionable — `BedrockSigningError` now says `pip install "headroom-ai[bedrock]"` instead of the generic `pip install boto3`, so anyone who does hit it lands on the right extra.

Verification (local, with boto3 present): `ruff format --check` + `ruff check` clean on all changed files; bedrock signer + new packaging tests pass; confirmed boto3 resolves under both `[all]` and `[proxy,bedrock]`. CI will re-run on the rebased head once a maintainer approves the fork-PR workflow gate.

JerrettDavis (approved):
The latest commit resolves the packaging/install blocker I raised. The Bedrock signing path no longer depends on an undocumented runtime dependency: `[bedrock]` declares boto3, `[all]` now includes `bedrock`, the Bedrock guide uses `pip install "headroom-ai[proxy,bedrock]"`, and the new packaging tests guard against dropping the extra again.

Code review passes on the signing flow as well: Bedrock mode routes Claude Code through `ANTHROPIC_BEDROCK_BASE_URL`, starts the proxy with signing enabled, signs the post-compression outbound bytes, keeps explicit gateway URLs taking precedence, and has focused signer/handler/wrap coverage. CI still needs to run/finish on the final head, but the previous code blocker is resolved.

didhd:
Rebased onto the latest `main` (`715ed7d2`) — the PR is mergeable again (was showing conflicts after `main` moved ~123 commits).

**Conflicts resolved**
`main` had refactored several files I also touch, so this was a real 3-way merge on 9 files (the rest applied cleanly):
- **`headroom/cli/wrap.py`** — `main` introduced a centralized `_claude_wrap_base_url_env_key(foundry_mode, vertex_mode)` helper plus a parallel `--1m` / `--tool-search` / native-Vertex path. I dropped my earlier ad-hoc `env_key=` mechanism and instead extended `main`'s helper with `bedrock_mode` (→ `ANTHROPIC_BEDROCK_BASE_URL`), added a `_settings_bedrock` tracker alongside `_settings_vertex`, and threaded `bedrock_mode=` through `_write_/_restore_claude_wrap_base_url` and the unwrap path. Bedrock, Vertex, and Foundry now coexist through one consistent mechanism.
- **`pyproject.toml` / `uv.lock`** — `main` dropped `benchmark` from `[all]`; I kept `main`'s new `[all]` set and re-added `bedrock` to it (so the packaging guard from the last round still holds).
- **`README.md` / `meta.json`** — kept `main`'s new entries (`--1m`/`--tool-search` note, `claude-code-azure-foundry`) and merged my Bedrock note / `claude-code-bedrock` nav entry beside them.
- `proxy.py`, `server.py`, `models.py`, `proxy_routes.py` merged cleanly; verified the `--bedrock-sign` flag, `bedrock_sign` config field, route registration, and `_ensure_proxy`/`_start_proxy` threading all survived.

Verified locally: `ruff format --check` + `ruff check` clean on all changed files; every changed Python file compiles; the packaging guards pass (`[all]` resolves to `main`'s set **+** `bedrock`); `wrap claude` still defines `use_bedrock`/`bedrock_region` and wires `bedrock_sign=use_bedrock`. Diff vs `main` is purely additive (`wrap.py` +61/-5, `server.py` +16/-0, etc.) — no part of `main`'s refactor is reverted.

**On the red `CI / test (3)`**: it's 25 `OSError: couldn't connect to 'https://huggingface.co'` failures in `tests/test_memory*` — an HF model-cache miss on that one shard's runner. It's unrelated to this PR: none of my 16 files touch memory/embedding code, shards 1/2/4 passed, `prefetch-model` succeeded, and `main`'s own CI is green with the same suite. A re-run should clear it (fork PRs need a maintainer "Approve and run").

JerrettDavis (approved):
Reviewed the latest Bedrock signing head after the final update/rebase. The implementation still matches the previously-approved contract: Claude Bedrock mode uses `ANTHROPIC_BEDROCK_BASE_URL`, starts the proxy with Bedrock signing enabled, signs the post-compression bytes for direct AWS forwarding, preserves explicit Bedrock gateway precedence, and includes the packaging/docs coverage for the `bedrock` extra.

Only governance checks are attached to the latest head, so this approval is based on code review of the final diff and the existing focused coverage rather than a fresh full CI matrix.

guidodl:
Tested this branch (`0b35eaf`) against **real AWS Bedrock in `eu-west-1`** with **application-inference-profile ARNs** — a case that fails on `main`. Verdict: the core works end-to-end. 🎉

**Setup:** ran the branch's Python over the released 0.28.0 compiled `_core` (pure-Python PR, no Rust rebuild needed). Proxy started with `--bedrock-sign --region eu-west-1`, AWS auth via a named SSO profile. Drove it with real Claude Code `claude-cli/2.1.198` in Bedrock-native mode (`CLAUDE_CODE_USE_BEDROCK=1` + `ANTHROPIC_BEDROCK_BASE_URL`), plus direct SigV4-signed requests.

**What works:**
- Native `/model/{arn}/invoke` and `/invoke-with-response-stream` → **HTTP 200**, SigV4 re-signing to real eu-west-1 Bedrock. Verified with 3 distinct application-inference-profile ARNs.
- Claude Code round-trips cleanly — `exit 0`, correct completion, traffic confirmed transiting the proxy (`client=claude-code`).
- **Compression on the native Bedrock path is real:** a request with a large `tool_result` block went **54,953 → 83 tokens (99.8% reduction)** via `router:tool_result:log`, still returning a valid 200. This confirms the full compression pipeline runs on re-signed Bedrock traffic, not just tool-output `/v1/compress`.

**Rough edges observed:**
- `POST /model/{arn}/count-tokens` → **404** (the native-path gap from #1589). Claude Code tolerated it (no fatal error), but flagging it.
- On a large (~98KB) streaming request, the **first attempt returned 400** and the SDK retry succeeded (200). Reproducible; may be re-signing-on-large-body related. Happy to share more detail.
- `GET /inference-profiles` is also hit by Claude Code's SDK — worth confirming it's handled.

Net: for Bedrock + application-inference-profile ARNs, this is the only path that gets Claude Code traffic compressed without a re-signing gateway. Would love to see it land. Thanks @didhd!

JerrettDavis (approved):
Refreshed this branch on current main and resolved the wrap/dependency conflicts. The merge preserves the Bedrock wrapping/signing path while keeping main's newer Copilot session seed plumbing and wrap self-heal cleanup.

Local validation:
- `uvx --from ruff==0.15.17 ruff check headroom/cli/proxy.py headroom/cli/wrap.py headroom/providers/proxy_routes.py headroom/proxy/bedrock_signer.py headroom/proxy/handlers/bedrock.py headroom/proxy/models.py headroom/proxy/server.py tests/test_bedrock_packaging.py tests/test_cli/test_wrap_claude_bedrock.py tests/test_proxy/test_bedrock_passthrough.py tests/test_proxy/test_bedrock_signer.py`
- `uvx --from ruff==0.15.17 ruff format --check headroom/cli/proxy.py headroom/cli/wrap.py headroom/providers/proxy_routes.py headroom/proxy/bedrock_signer.py headroom/proxy/handlers/bedrock.py headroom/proxy/models.py headroom/proxy/server.py tests/test_bedrock_packaging.py tests/test_cli/test_wrap_claude_bedrock.py tests/test_proxy/test_bedrock_passthrough.py tests/test_proxy/test_bedrock_signer.py`
- `uv run pytest tests/test_bedrock_packaging.py tests/test_cli/test_wrap_claude_bedrock.py tests/test_proxy/test_bedrock_passthrough.py tests/test_proxy/test_bedrock_signer.py -q` (23 passed, 3 skipped)

Pre-commit also passed on the merge commit.

JerrettDavis:
CI correction on head `bf3317e20`: the first full run exposed two missing Bedrock integration contracts, now fixed rather than rerun blindly.

- Registered `HEADROOM_BEDROCK_SIGN` as a persisted advanced boolean setting, matching the new Click option and dashboard/settings precedence contract.
- Extended Claude unwrap's all-mode contract to standard, Foundry, Vertex, and Bedrock base-URL restoration.
- Focused verification: 115 passed, 3 skipped across unwrap, settings registry, Bedrock passthrough/signing preflight, packaging, and region behavior; Ruff/format/diff checks pass.
- The corrected head retains current `main` ancestry including the MCP SDK v1 `<2.0.0` compatibility fix.

Full applicable CI on this new head must finish entirely green before human review readiness. No merge or auto-merge action was taken.

JerrettDavis (changes requested):
Re-reviewed exact head `bf3317e20bd479f339776cf8e1ed40d82ed9ddd4`. The Bedrock-specific implementation still looks coherent, and the focused signer/passthrough suite passes locally (17 passed, 3 optional-dependency skips; Ruff clean), but this head needs a current-main refresh before merge:

- A merge simulation against current `main` has a content conflict in `headroom/cli/wrap.py`, which is the central routing file changed by this PR.
- On the exact head, `tests/test_cli/test_unwrap_claude.py::test_unwrap_claude_removes_mcp_purges_retired_hook_and_stops_proxy` fails reproducibly on Windows: `unwrap_claude` leaves the retired `rtk-rewrite.sh` PreToolUse hook in settings. The broader focused run is 19 passed, 1 failed, 1 skipped.

Please rebase/merge current `main`, resolve the wrap routing changes deliberately, and rerun the Claude wrap/unwrap plus Bedrock suites on Windows. Once the refreshed head is green, I can re-review the resulting integration diff.

didhd:
Rebased onto current main (1390d89). The conflict is gone, and the rebase turned up a real failure that's now fixed.

The `wrap.py` conflict was the locking refactor from #3205 — main moved the read/claim/write of `_write_claude_wrap_base_url` inside `_wrap_settings_lock`. I took main's version as-is and only threaded `bedrock_mode` into the `_claude_wrap_base_url_env_key()` call, so `ANTHROPIC_BEDROCK_BASE_URL` goes through the same lock and ownership claim as the other three modes. Nothing from #3205 is reverted.

Resolving it also exposed a gap worth calling out: `_selfheal_dead_wrap_base_url()` sweeps the base-url keys for a dead-proxy marker, but only listed standard/Foundry/Vertex. A stale `ANTHROPIC_BEDROCK_BASE_URL` would have survived a hard proxy death — exactly what #2221 clears for the other modes. Added the bedrock key to that loop.

On unwrap there is a genuine failure, though a different test than the one you hit: `test_unwrap_claude_restores_all_base_url_modes`, and it fails on Linux/macOS too. When the Bedrock row was inserted into the all-modes expectation, the Vertex row's `force: True` went with it, so the expectation no longer matched what unwrap actually does (all four modes pass `force=True`). Fixed the expectation, not the code.

I can't reproduce `test_unwrap_claude_removes_mcp_purges_retired_hook_and_stops_proxy`. It passes here, and everything that assertion depends on lives in `headroom/context_tool_cleanup.py`, which this branch doesn't touch — it isn't in `git diff main...HEAD --stat`. That purge already normalizes for Windows (`_norm_path_text` case-folds and flips separators), so rather than guess at a fix I'd rather see the `-vv` output if it still fails on your box. It reads like a main-branch issue, not a Bedrock one. Worth noting the CI test shards are ubuntu-latest only, so nothing in the matrix covers that path on Windows either way.

Local runs on the rebased head:

- `ruff check` and `ruff format --check` clean over `headroom/` and `tests/`
- Bedrock + wrap/unwrap + settings suites: 182 passed
- `tests/test_cli` + `tests/test_proxy` + context-tool cleanup: 1058 passed, 1 skipped, 1 failed. The failure is `test_litellm_vertex_backend_path_preserves_max_tokens_and_vendor_fields`, which reproduces identically on a pristine `main` worktree — it reaches the real OpenAI API and gets a 401 in my environment — so it isn't from this branch.

CI needs the fork-PR approve-and-run. @JerrettDavis, could you take another look when you get a chance?

JerrettDavis (changes requested):
The current-main refresh resolves my earlier wrap/unwrap blocker: the focused packaging/wrap/unwrap run passes, and after installing the locked boto3/botocore 1.42.38 into an isolated directory all 29 signer/passthrough tests pass without optional-dependency skips. One signing correctness issue remains for the advertised long-lived SSO/role credential support: BedrockSigner.sign passes the mutable credentials returned by Session.get_credentials() directly to SigV4Auth. That signer reads token, secret_key, and access_key separately. RefreshableCredentials can refresh between those reads; Botocore explicitly documents using get_frozen_credentials() to avoid mixing generations. A rotating-credential fixture through this sign() method produced an old X-Amz-Security-Token with a new Credential access key/signing secret, which cannot authenticate as one credential set. Please retain the cached refreshable provider, but take creds.get_frozen_credentials() for each individual signing operation and pass that snapshot to SigV4Auth. Add a rotation-boundary regression so the same snapshot supplies all three fields, while the next request can observe refreshed credentials. This applies to both runtime POST and the new discovery GET. Reference: https://github.com/boto/botocore/blob/1.42.38/botocore/credentials.py#L663

didhd:
Good catch on the frozen credentials — that one would have been miserable to debug in the field. Fixed in 2e6505e, and rebased onto current main (e67b3c8a).

`sign()` now takes `creds.get_frozen_credentials()` per signing operation and hands that `ReadOnlyCredentials` snapshot to `SigV4Auth`, which is what botocore's own `RequestSigner.get_auth_instance` does. The refreshable provider stays cached, so request N+1 still sees refreshed credentials — only the snapshot is per-request. Both the runtime POST and the discovery GET go through `sign()`, so one change covers both.

The regression uses a provider that advances a generation on every *direct* attribute read while `get_frozen_credentials()` returns one coherent generation. Reverting the fix reproduces exactly what you described:

```
pre-fix  Credential access key: AKIDGEN4
pre-fix  X-Amz-Security-Token : token-gen-2
```

Three fields, three generations. Post-fix the provider records one snapshot per signature and all three fields share it; a second test asserts the next request re-freezes rather than reusing generation 1. Parametrised over runtime POST and discovery GET. 13 signer tests, and the 3 new ones fail on the pre-fix code.

I also ran it against real Bedrock in us-west-2 rather than trusting the unit tests:

- `POST /model/us.anthropic.claude-sonnet-5/invoke` → 200, correct completion
- `POST .../invoke-with-response-stream` → 200, decoded a real event stream (`message_start` … `message_stop`)
- `GET /inference-profiles?maxResults=3` → 200 with real profiles, so the #1589 404 is gone
- compress → re-sign → forward: a 389,443-byte `tool_result` body came back 200 with **556** billed input tokens against 219,340 uncompressed

One thing worth flagging that is not about signing. Compression never engages on the Bedrock path in the proxy's default mode. `headroom proxy` defaults to `--mode cache`, and `handlers/bedrock.py` treats cache mode as a full bypass:

```python
bypass = (
    _headroom_bypass_enabled(request.headers)
    or not getattr(self.config, "optimize", True)
    or is_cache_mode(getattr(self.config, "mode", "token"))
    ...
```

Measured against a local sink with the same 389,443-byte body:

| mode | bytes forwarded upstream |
|---|---|
| `cache` (default) | 389,443 — verbatim |
| `token` | 705 |

The Anthropic and OpenAI handlers don't bypass in cache mode; they compress and only freeze the prior-turn prefix via `_strict_previous_turn_frozen_count`. And `wrap.py` forwards `--mode` only when `HEADROOM_MODE` is set, so `headroom wrap claude` with `CLAUDE_CODE_USE_BEDROCK=1` currently gets no compression unless the user exports `HEADROOM_MODE=token`. That is consistent with @guidodl's 99.8% result, which would have been token mode.

I left it alone deliberately — dropping the bypass so the pipeline's `cache_aligner` handles it would trade Bedrock prompt-cache hit rate for compression, and that reads like your call rather than something to slip into a signing fix. Happy to do it here or in a follow-up, whichever you prefer.

Local on this head: `ruff check` and `ruff format --check` clean over `headroom/` and `tests/`; `git diff --check` clean; 125 passed across the Bedrock signer/passthrough/packaging, wrap claude, unwrap claude and settings-store suites.

JerrettDavis (approved):
Reviewed the latest credential-freezing change after the rebase. The provider remains cached, but each sign operation takes a fresh immutable snapshot before SigV4Auth reads credential fields; this preserves rotation between requests without mixing generations inside one signature. All 13 signer tests pass with locked boto3/botocore dependencies, covering runtime POST, discovery GET, static credentials, and successive snapshots. No blocking issue found in this update; no live AWS account call was required for these signing invariants.

PR #1242, opened 2026-06-21
docs(compose): document the memory-stack docker-compose services

ai-Ev1lC0rP:
## What

Adds explanatory comments throughout `docker-compose.yml` so the "memory stack" is self-documenting for new users.

Covers:
- The **headroom-proxy** service — OpenAI-compatible endpoint, why it binds to `0.0.0.0`, the `/readyz` healthcheck, and the `depends_on` start-order caveat.
- **Qdrant** (vector search) and **Neo4j** (relationship graph) — their roles, exposed ports, and named volumes for persistence.
- A header block with quick-start steps, the full list of host-exposed ports, and a note that the proxy can run standalone without the datastores.
- A callout that the `NEO4J_AUTH` default is **local-dev only** and must be overridden before any non-local use.

## Why

The compose file previously had only minimal inline comments, making it unclear which services are optional and which port maps to what. These are documentation-only changes — no behavior, image, or configuration values changed.


JerrettDavis (approved):
Docs review looks good. The compose comments clarify the service roles, exposed ports, persistence, readiness behavior, and the local-only Neo4j default without changing runtime values.

This branch is currently merge-conflicted (`DIRTY`) and only governance checks have run, so it needs a refresh before merge, but I do not see a docs/content blocker.

JerrettDavis:
Maintainer refresh: merged current main and resolved the compose conflict by keeping main's persisted `headroom_workspace` mount along with this PR's memory-stack documentation comments. Local validation: parsed `docker-compose.yml` with PyYAML successfully.

JerrettDavis (approved):
Re-reviewed after the compose-doc refresh against main. This is documentation/commentary only in `docker-compose.yml`; the service definitions, ports, volumes, and healthcheck semantics remain unchanged. The local-dev credential warning and memory-stack notes are accurate and useful.

Approval stands.

PR #1243, opened 2026-06-21
feat(dashboard): add Simplified Chinese locale

2233admin:
## Description

Adds a persisted English/Simplified Chinese locale selector to the current dashboard and fixes translation updates for live Alpine-rendered values.

## Type of Change

## Changes Made

- Add English and Simplified Chinese locale selection persisted in localStorage.
- Preserve the current settings control while integrating the locale selector.
- Re-translate from the latest Alpine-rendered source value instead of restoring stale cached text.
- Add Playwright coverage for changing live text and reversible locale switching.

## Testing

### Test Output

```text
tests/test_dashboard_locale_playwright.py: 1 passed
dashboard static/token focused suite: 7 passed
ruff: All checks passed!
```

## Real Behavior Proof

- Environment: Chromium via Playwright, 1440x1000 viewport, mocked current dashboard API responses.
- Exact steps: open `/dashboard`, choose `zh-CN`, update a translated live value from 0 to 10 requests, then switch back to English.
- Observed result: the current value remains 10 in both locales; the old 0 value is never restored.
- Not tested: exhaustive linguistic review of every translated phrase.

## Review Readiness

## Checklist

- [x] New and existing focused tests pass locally

Simplified Chinese dashboard

## Additional Notes

The screenshot and test exercise the redesigned current dashboard rather than the older layout from the original branch.


JerrettDavis (changes requested):
This needs one runtime fix before it is safe to merge. `applyHeadroomTranslations()` stores `node.__headroomI18nSource` the first time it sees a text node and then keeps using that cached value on every later pass. That works for static labels, but many dashboard nodes are updated by Alpine via `x-text` after polling.

For example, a node can start as `0 requests processed`, get cached as the source, then Alpine updates the same text node to `10 requests processed`. The next translation pass ignores the new `node.nodeValue` because `__headroomI18nSource` is already set, and writes the old translated `0` value back into the DOM. The same stale-value problem can hit live stats, history counts, timestamps, and other dynamic labels.

Please change the translation approach so dynamic `x-text` content uses the current English value after each Alpine update, while still preserving reversible source text for truly static DOM/attribute labels. A focused browser/unit regression for a changing `x-text` value under `zh-CN` would be valuable here.

JerrettDavis:
Conflict audit against the redesigned dashboard: this locale implementation mutates arbitrary rendered text nodes and applies broad global regex replacements (including generic words such as `Daily`, `Weekly`, and `Monthly`) after polling. The translation dictionary also targets the older dashboard structure, so mechanically transplanting it would leave current sections untranslated and risks modifying dynamic model/provider/user content. A correctness-preserving update needs keyed UI strings/components plus Playwright locale coverage and current-dashboard screenshots; I have not pushed a partial or misleading localization.

JerrettDavis (approved):
The stale dynamic-text overwrite is fixed, current main is reconciled, and the locale behavior now has Playwright proof plus a current-dashboard screenshot.

PR #1249, opened 2026-06-21
feat: Auto-continue on Anthropic 5-hour limit with UI messaging

Anmoldureha:
## Description
This PR enhances the Headroom proxy to automatically handle Anthropic's 5-hour limit 429 errors. It parses the reset time, streams a `waiting` message back to Claude Code's UI so the user knows what's happening, sends keep-alives during the wait, and safely rewrites SSE events when continuing the stream so the response merges seamlessly.

## Type of Change

## Changes Made
- Modified `headroom/proxy/handlers/streaming.py` to intercept 429 errors and extract the ISO timestamp of the reset limit.
- Implemented `auto_continue_wrapper` async generator to hold the connection open, send SSE keep-alives, and automatically retry the request once the limit expires.
- Added on-the-fly SSE rewriting to drop redundant `message_start` events and safely merge the actual response.

## Testing
- Tested locally using a mock FastAPI proxy that returns the Anthropic 429 error and reset timestamp.
- Verified that the proxy correctly parses the time, streams the keepalives, and restarts the flow upon completion.
- Ensured Claude Code SDK parsers are not broken by duplicate `message_start`s.

## Real Behavior Proof
The code executes successfully and intercepts the 5-hour limit errors. The proxy keeps the SSE connection alive until the requested reset time and then seamlessly merges the new upstream completion.

## Review Readiness
The code is thoroughly documented, well-structured, and ready for review.

- [x] I have performed a self-review before requesting human review.
- [x] This PR is ready for human review or convert the PR back to draft.

### DIFF
--- a/headroom/proxy/handlers/streaming.py
+++ b/headroom/proxy/handlers/streaming.py
@@ -990,6 +990,9 @@ class StreamingMixin:
 
         get_codex_rate_limit_state().update_from_headers(dict(upstream_response.headers))
 
+        _is_5hour_limit = False
+        _sleep_seconds = 0
+        _reset_time_str = ""
         if upstream_response.status_code >= 400:
             logger.warning(
                 "[%s] Forwarding upstream streaming error status=%s url=%s",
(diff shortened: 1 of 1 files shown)


JerrettDavis (changes requested):
Same blocker as the related auto-continue PR: this cannot safely merge as an always-on streaming behavior. It holds the client stream open until the 5-hour reset, synthesizes assistant SSE events, and rewrites upstream event indexes by regex after retrying. That is too much untested state-machine behavior for the main streaming path, and it changes product semantics for every 429 matching the string pattern. Please make the behavior opt-in with a bounded max wait/retry policy, factor and test the reset parser, and add tests for cancellation, retry failure, non-matching 429 passthrough, and text/tool SSE event sequencing.

JerrettDavis:
Conflict/correctness audit: this is the earlier variant of #1252 and has the same unsafe semantics—hours-long held connections, synthetic assistant UI content inserted into the model stream, retrying a non-idempotent request, and SSE index rewriting without protocol tests. This requires an opt-in bounded retry design with cancellation/idempotency handling rather than a mechanical conflict resolution, so I have left it unchanged.

JerrettDavis:
Closing in favor of #1252, which supersedes this branch from the same base. Both PRs start at `61298084` and replace the same Anthropic streaming error path; #1252 contains the complete auto-continue/wait/keepalive/SSE-rewrite behavior from this PR, then extends it for alternate error payloads/reset formats and fixes replayed chunk types for Starlette.

Maintaining two divergent conflicts for the same feature would risk landing the older implementation without the follow-up compatibility and byte-safety fixes. I am retaining #1252 as the canonical branch to reconcile against current `main`. Thank you for the original implementation.

PR #1250, opened 2026-06-21
Fix correctness and safety bugs across compression, proxy, cache, memory

adipu007:
## Summary

This PR fixes **13 correctness/safety bugs** found by auditing the core compression, proxy, cache, and memory subsystems. No features are added or removed — each change traces to a concrete failing input.

## Critical — data loss / wrong output

- **`cache/semantic.py`** — Semantic cache matched purely on the final-query embedding and ignored prior conversation. Two conversations ending in the same message (e.g. `"continue"`, `"fix it"`) could receive **each other's cached response**. Now scoped to a context hash (messages before the final user turn). Includes a regression test.
- **`proxy/handlers/openai.py`** (Codex WS memory relay) — An indentation defect put the buffer-flush block at the `if not decided` level and added an unconditional `continue`, making the memory tool-execution and pass-through phases **dead code**. Memory tool calls never executed and per-turn metrics were corrupted.
- **`transforms/kompress_compressor.py`** — Words past the `max_length=512` truncation point in a chunk had no `word_id` and were **silently dropped** from the output. They are now kept, preserving the "same answers" guarantee.
- **`memory/adapters/hnsw.py`** — `index_batch` sized capacity by the active entry count, but labels come from the monotonic `_next_hnsw_id` which is never reclaimed on evict. After churn, `add_items` could get an out-of-range label. Now sizes by `_next_hnsw_id`, matching the single-item path.
- **`transforms/content_router.py`** — A bare ```` ``` ```` fence (no language) was rewritten to ```` ```unknown ````. Now preserved without a bogus tag. Includes a regression test.

## High — error masking / malformed requests

- **`proxy/server.py`** — On the final retry of a persistent upstream 5xx, the real response is returned instead of raising a synthetic 500, so the true status/body/`retry-after` reaches the client. Also guards `await request.json()` on `/v1/retrieve`, `/v1/telemetry/import`, `/v1/retrieve/tool_call` (400 instead of 500 + stack trace).
- **`proxy/handlers/gemini.py`, `proxy/handlers/openai.py`** — `response.json()` now also catches `JSONDecodeError`/`ValueError`, so an upstream 200 with a non-JSON body is no longer turned into a fabricated 502 (matches the existing correct sibling path in `gemini.py`).
- **`proxy/handlers/openai.py`** — The Responses-API memory continuation is only adopted when it returns HTTP 200, and its `.json()` is guarded — a failed continuation no longer clobbers a successful response.
- **`proxy/memory_handler.py`** — When the memory backend is unavailable, an error `tool_result` is still emitted so the Anthropic continuation isn't malformed (Anthropic rejects a `tool_use` with no matching `tool_result`). Adds an `isinstance(block, dict)` guard in the Anthropic tool-call extractor.

## Medium

- **`proxy/handlers/openai.py`** — Passthrough streaming uncached-token math is now provider-aware (Anthropic `input_tokens` already excludes cache tokens; Gemini `promptTokenCount` includes them).
- **`proxy/handlers/batch.py`, `proxy/handlers/anthropic.py`** — The per-item compression-failure path increments `total_original_tokens` alongside `total_optimized_tokens`, keeping the savings denominator aligned.
- **`proxy/memory_handler.py`** — Background dedup tasks are retained in a set with a done-callback so the event loop can't GC them mid-flight.
- **`memory/traffic_learner.py`** — Dedup eviction now removes the oldest hash (a plain `set.pop()` was arbitrary) and cleans up the matching `_persisted_ids` entry (was leaking + could bump a stale row).
- **`cache/compression_store.py`** — A true hash collision preserves the existing entry instead of silently overwriting it (which could make `headroom_retrieve()` return the wrong content for an already-emitted marker).

## Tests

- `tests/test_semantic_cache_context_regression.py` — runs without native deps; verified it fails before the fix and passes after.
- `tests/test_bare_code_fence_regression.py` — requires the native `headroom._core` extension (`maturin develop`).

## Notes

Most fixes were verified by `python -m py_compile` and reasoning; the proxy/memory/transform paths require the native extension + optional ML deps to exercise at runtime, which I couldn't build in my environment. Happy to adjust style or split this into smaller PRs if preferred.


chopratejas:
Hi - thank you! Can you change the PR body to match the template?

JerrettDavis (changes requested):
This PR includes several useful fixes, but the compression-store collision change is still unsafe. On a true hash collision, `CompressionStore.store()` now preserves the existing entry but still returns the colliding `hash_key`. Callers treat a non-empty return value as a successful store and can emit a CCR marker for the new content using that hash, which would later retrieve the old content. That is the same wrong-data failure mode, just without overwriting the entry. Please return a failure/no-marker signal for true collisions (or generate a distinct key) and add a regression test that a colliding second store does not cause a marker/retrieval path to point at the first entry.

JerrettDavis:
Resolved against current `main`, retaining newer main implementations where the same safety issue was already fixed and porting still-missing guarantees to current modules:
- CCR hash collisions preserve the existing marker target instead of silently overwriting it
- Kompress preserves tokenizer-truncated tail words while retaining main's must-keep policy
- bare code fences remain bare in the extracted `mixed_content` module
- traffic-learner dedup eviction remains ordered and bounded
- semantic cache matches are scoped to prior conversation context

Validation:
- `127 passed` — new regressions plus compression store, HNSW, and Kompress suites
- `152 passed` — complete traffic learner suite
- `204 passed` before correcting two stale set-type assertions; all affected reruns now pass
- Ruff check clean; affected files formatted

JerrettDavis (approved):
The conflict resolution retains current main provider/memory architecture and ports the PR's remaining correctness guarantees to current file boundaries. Focused and broader affected suites pass; Ruff is clean.

PR #1253, opened 2026-06-21
Make all standard optional dependencies install by default

YOUSSEFELJAYAD:
## What

Adds a single self-referential `headroom-ai[all]` entry to the core `[project.dependencies]` in `pyproject.toml`, so a plain `pip install headroom-ai` (or `npm`-style discovery → `uvx`/`pipx`) ships every **standard** feature out of the box.

## Why

New users install `headroom-ai`, run `headroom proxy`, and hit:

```
Error: Proxy dependencies not installed. Run: pip install headroom-ai[proxy]
```

The CLI's headline commands (`proxy`, `learn`, code/memory features) all live behind optional extras, so the default install can't run them. This makes the documented quickstart fail until users discover and append the right extra.

## How

```toml
dependencies = [
    ...
    "headroom-ai[all]",   # batteries-included default
]
```

- Reuses the **same intra-package self-reference pattern** the project already relies on for the `all`, `proxy-prod`, and `voice-train` extras — nothing new for the build backend (maturin) to support.
- `[all]` already curates the cross-platform-safe set (proxy, code, ml, memory, relevance, image, reports, otel, evals, voice, html, benchmark, mcp, spreadsheet).
- Platform-specific extras that `[all]` **intentionally omits** stay opt-in: `pytorch-mps` (macOS-only), `proxy-prod`/`gunicorn` (Unix-only), `memory-stack` (Qdrant/Neo4j). So Windows, CI, and Python 3.14 installs are unaffected.

## Trade-off (please weigh)

This is deliberately the **batteries-included** direction: the default install gets heavier (pulls `torch`, `transformers`, etc. via `[ml]`/`[voice]`). That is the explicit goal of this PR — make the documented commands work with zero extra steps. If you'd prefer to keep the lean core, a lighter alternative would be to default only to `[proxy]` (what the quickstart actually needs) and/or emit a clearer "run `pip install headroom-ai[all]`" hint on missing extras. Happy to repoint this PR at either approach.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,13 @@ dependencies = [
     "opentelemetry-api>=1.24.0",  # Safe no-op OTEL API for instrumentation
     "ast-grep-cli>=0.30.0",       # AST-aware code slicing (CodeCompressor); binary wheel
     "tomli>=2.0.0; python_version < '3.11'",  # tomllib backport for helper scripts
+    # Batteries-included: pull in every standard optional feature by default so a
+    # plain `pip install headroom-ai` ships the proxy, ML/Kompress, code, memory,
+    # relevance, image, reports, otel, evals, voice, html, benchmark, mcp and
+    # spreadsheet extras. Self-references the [all] aggregate (same pattern already
+    # used by [all]/[proxy-prod]/[voice-train]); platform-specific extras
+    # (pytorch-mps, proxy-prod/gunicorn, memory-stack) stay opt-in as before.
+    "headroom-ai[all]",
 ]
 
 [project.optional-dependencies]


JerrettDavis (changes requested):
This needs a different packaging shape before it can merge. Adding `headroom-ai[all]` to core `[project.dependencies]` makes the package depend on its own aggregate extra, which is a risky resolver/build behavior to rely on for the default install path. More importantly, `[all]` is not just the proxy quickstart surface: it pulls in heavy and specialized stacks such as ML/voice/evals/benchmark dependencies (`torch`, `transformers`, benchmark tooling, etc.) for every plain `pip install headroom-ai`.

That is too broad for a default enterprise install, especially when the stated failure is that `headroom proxy` needs the proxy extra. Please solve the quickstart/runtime gap more directly: either move the minimal proxy/runtime dependencies into the base install, make the CLI error/install hint clearer, or introduce a smaller default extra/install path that does not pull every optional feature. Also please include a lockfile/update and an install smoke test for the chosen packaging behavior so we know `pip install headroom-ai` and `pip install "headroom-ai[all]"` both resolve as intended.

JerrettDavis:
Conflict audit against current packaging: the proposed base dependency `headroom-ai[all]` is a self-reference and also changes the intentionally minimal base install into every heavy optional stack (ML, voice, evals, image, proxy, etc.). Resolving the lockfile conflict mechanically would not make that packaging design safe. This needs an explicit batteries-included distribution/extra decision rather than a conflict-only merge, so I have left the branch unchanged.

JerrettDavis:
Closing as superseded by the repository’s current explicit slim-core packaging contract.

Current `pyproject.toml` keeps `[project.dependencies]` limited to the lightweight compression/runtime core, while proxy, ML, memory, image, evaluation, and other stacks remain optional extras. Commit `18118af5` formalized that direction, and the current installation guide separately documents:
- `pip install headroom-ai` for the lightweight core
- `uv tool install --python 3.13 "headroom-ai[all]"` for the full CLI
- `pip install "headroom-ai[proxy]"` or other targeted extras for selected capabilities

Adding `headroom-ai[all]` to `headroom-ai`’s own core dependencies would both reverse that intentional install-size/platform boundary and introduce a self-referential package dependency rather than merely reusing an extra aggregation. The original quickstart concern is now handled by explicit full-CLI install commands and clearer extra-specific documentation. Resolving this conflict by applying the proposed line would therefore change current product intent, not preserve it safely. Thank you for surfacing the onboarding tradeoff.

PR #1258, opened 2026-06-22
fix(proxy): forward request-id headers on the streaming path (#1100)

### LINKED ISSUE #1100: Streaming responses drop the upstream `request-id` header (only ratelimit/x-codex forwarded)
## Summary

On the streaming (SSE) path, the proxy rebuilds the downstream response headers from an allowlist that keeps only rate-limit and Codex headers (`headroom/proxy/handlers/streaming.py`, `_stream_response`, allowlist at ~lines 1083-1086, returned on the `StreamingResponse` at ~line 1338 on `main`):

```python
forwarded_headers = {
    k: v
    for k, v in upstream_response.headers.items()
    if "ratelimit" in k.lower() or k.lower().startswith("x-codex")
}
```

Anthropic's `request-id` response header is therefore never forwarded to streaming clients. Clients that record it lose it — most visibly Claude Code, which writes a `requestId` into each transcript turn from that header.

This is the same class of omission already fixed for other headers on this exact allowlist:

- #57 added `*ratelimit*` forwarding on the streaming path.
- #794 added `x-codex-*` (fixing #577, where Codex session/weekly usage never updated on streaming SSE).

`request-id` is the next header clients need but never receive. The non-streaming/buffered path already forwards all upstream headers (`dict(upstream_response.headers)`), so only streaming clients are affected.

## Impact

- Claude Code transcripts produced through Headroom have **no `requestId`** field.
- Downstream usage/cost tools that deduplicate transcript rows by the provider response identity (`messageId` + `requestId`) can no longer dedup and over-count tokens. CodexBar, for example, requires both `messageId` and `requestId`; without `requestId` it counts the same `message.id` once per file it appears in — inflating Claude token/cost roughly 2.5x for agent-heavy workloads (subagents / fork / resume copy a message across many files). `ccusage` is unaffected only because it treats `requestId` as optional.
- `request-id` is standard response metadata used for support tickets and debugging; silently dropping it on the streaming path is surprising and asymmetric with the non-streaming path.

## Reproduction (no private data needed)

1. Direct:
   ```bash
   ANTHROPIC_BASE_URL=https://api.anthropic.com claude -p "hi" --output-format json
   ```
   The newest `~/.claude/projects/**/<session>.jsonl` assistant row has a top-level `"requestId": "req_…"`.
2. Through Headroom:
   ```bash
   headroom proxy --port 8787 &
   ANTHROPIC_BASE_URL=http://127.0.0.1:8787 claude -p "hi" --output-format json
   ```
   The resulting assistant row has **no** `requestId`.

In my logs, `requestId` presence dropped from ~100% to ~0% on the day I routed Claude Code through Headroom (2026-06-12); mixed days are bimodal per session (direct = present, proxied = absent), confirming the proxy rather than a Claude Code version change.

## Proposed fix

Add the `request-id` family (`request-id`, `anthropic-request-id`, `x-request-id`) to the streaming allowlist — the same one-line widening you already applied for `*ratelimit*` (#57) and `x-codex-*` (#794).

One check before I send a PR: the allowlist is deny-by-default and a test currently asserts `x-request-id` is dropped on this path (`tests/test_proxy_streaming_ratelimit_headers.py`). Is dropping non-allowlisted response headers on the streaming path a deliberate policy, or is `request-id` simply one you haven't needed to forward yet? Happy to open a PR (allowlist change + the test update) if you're open to it.

## Environment

Reproduced on `main` (v0.26.0) and on `headroom-ai` 0.25.0. Verified that a patched build restores `requestId` end-to-end on the `_stream_response` path. (The backend/compression streaming returns, `_stream_*_via_backend` at ~lines 1469/1685, forward no response headers at all and would need a separate change, since the upstream response isn't in scope at their `return`.)

Parideboy:
## Description

On the streaming (SSE) path the proxy rebuilt response headers from a deny-by-default allowlist that only kept rate-limit and Codex headers, so Anthropic's `request-id` was dropped. Claude Code needs that header to write `requestId` into transcripts; without it, usage/cost tools that dedup on `messageId` + `requestId` over-count tokens. This widens the streaming allowlist to also forward the `request-id` family, matching the non-streaming path.

Closes #1100

## Type of Change

## Changes Made

- `headroom/proxy/handlers/streaming.py`: widened the streaming response-header allowlist to also forward `request-id`, `anthropic-request-id`, and `x-request-id`.
- `tests/test_proxy_streaming_ratelimit_headers.py`: flipped two assertions that expected `x-request-id` to be dropped, and added `test_request_id_headers_forwarded_in_streaming`.

## Testing

### Test Output

```text
$ pytest tests/test_proxy_streaming_ratelimit_headers.py -q
11 passed

$ pytest tests/ -k "header or stream" -q
75 passed
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, pytest-asyncio 1.4.0 (asyncio_mode=auto)
- Exact command / steps: Ran the streaming rate-limit header suite plus adjacent header/stream tests after widening the allowlist.
- Observed result: All 11 tests in the targeted file pass including the new request-id forwarding test; 75 adjacent header/stream tests stay green.
- Not tested: Did not run a live end-to-end `claude -p` round-trip through the proxy to inspect transcript `requestId`.

## Review Readiness

### DIFF
--- a/headroom/proxy/handlers/streaming.py
+++ b/headroom/proxy/handlers/streaming.py
@@ -1080,10 +1080,16 @@ class StreamingMixin:
         # window/credit headers — the latter do not contain the ``ratelimit``
         # substring, so without the second clause the Codex CLI's own
         # session/weekly display would stop updating on the streaming path.
+        # We also forward the ``request-id`` family: clients such as Claude Code
+        # record it per transcript turn, and downstream usage/cost tools dedup by
+        # message id + request id. The buffered (non-streaming) path already
+        # forwards every upstream header, so this keeps the two paths symmetric.
         forwarded_headers = {
             k: v
             for k, v in upstream_response.headers.items()
-            if "ratelimit" in k.lower() or k.lower().startswith("x-codex")
+            if "ratelimit" in k.lower()
+            or k.lower().startswith("x-codex")
+            or k.lower() in ("request-id", "anthropic-request-id", "x-request-id")
         }
 
         async def generate():
--- a/tests/test_proxy_streaming_ratelimit_headers.py
+++ b/tests/test_proxy_streaming_ratelimit_headers.py
@@ -87,7 +87,8 @@ class TestStreamingRatelimitHeaderForwarding:
             "anthropic-ratelimit-output-tokens-limit": "30000",
             "anthropic-ratelimit-output-tokens-remaining": "27000",
             "anthropic-ratelimit-output-tokens-reset": "2026-03-25T12:00:00Z",
-            # Non-ratelimit headers that should NOT be forwarded
+            # request-id is forwarded (clients record it per transcript turn);
+            # cf-ray is a non-allowlisted header that must NOT be forwarded.
             "x-request-id": "req-12345",
             "cf-ray": "abc123",
         }
@@ -151,7 +152,7 @@ class TestStreamingRatelimitHeaderForwarding:
 
     @pytest.mark.asyncio
     async def test_non_ratelimit_headers_not_forwarded(self):
-        """Only ratelimit headers should be forwarded, not arbitrary upstream headers."""
+        """Arbitrary upstream headers stay dropped; the request-id family is allowed."""
         proxy = self._create_mock_proxy()
         mock_response = self._create_mock_upstream_response()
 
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Code review passed. The streaming header allowlist remains narrow while restoring parity for the request-id family, and the tests cover both positive forwarding and continued dropping of unrelated headers.

PR #1259, opened 2026-06-22
feat(proxy): support glob patterns in exclude_tools (#870)

### LINKED ISSUE #870: [FEATURE] Prefix/glob matching in exclude_tools (e.g. mcp__*) — exact-name-only matching can't keep up with MCP servers
## Problem

`exclude_tools` matches tool names exactly (set membership in `ContentRouter`, mirrored lowercase variants). That works for a stable, hand-picked list, but breaks down for MCP tools:

- Claude Code namespaces every MCP tool as `mcp__<server>__<tool>` — one server can expose dozens of tools (build123d-mcp currently exposes 31).
- MCP servers update independently of the proxy config. Ours is registered via `uvx --refresh build123d-mcp`, so new tools appear without any deploy — and each new tool is silently compression-eligible until an operator notices output going missing and appends the name.
- The failure mode is the bad kind of quiet: iterative agent workflows act on tool results (measurements, error messages) that CCR has replaced with `<<ccr:hash>>` markers (real-world report: pzfreo/Teleclaude#69).

We currently enumerate all 36 names into `--exclude-tools` and accept the drift (pzfreo/Teleclaude#70). A `mcp__*` rule would replace that brittle list with one entry.

## Proposal

Support glob patterns in exclude-tools entries, applied wherever exclusion is checked today:

```bash
--exclude-tools 'mcp__*'                    # all MCP tools
HEADROOM_EXCLUDE_TOOLS='mcp__build123d-*'   # one server
```

Implementation sketch: entries containing `*`/`?` are matched with `fnmatch.fnmatchcase` (case-insensitive via the existing lowercase mirroring); plain entries keep exact set-membership so the hot path is unchanged for current configs. `_parse_exclude_tools` already centralizes parsing, and the router checks names in one place, so the surface is small.

Worth considering alongside: should `mcp__*` be in `DEFAULT_EXCLUDE_TOOLS`? MCP results are closer to Read (reference data the agent acts on, often consumed via exact values) than to Bash logs (the stated ideal compression target). Off by default + easy opt-in via glob would also be a fine outcome.

## Relation to #825

Complementary, not duplicate: #825 is about the click entrypoint never reading the exclude-tools config at all; this is about the matching semantics once it does. Both are needed before MCP-heavy deployments can express "never compress my MCP tools" in one line.

Parideboy:
## Description

`exclude_tools` only matched tool names exactly, so users could not exclude families of tools (for example all `mcp__*`). This adds glob-pattern support via a shared `is_tool_excluded` helper used by both the content router and the OpenAI handler, keeping exact/case-insensitive matching intact.

Closes #870

## Type of Change

## Changes Made

- `headroom/config.py`: added `is_tool_excluded(name, exclude_tools)` helper that keeps exact/case-insensitive matching and adds `fnmatch` glob support.
- `headroom/transforms/content_router.py` and `headroom/proxy/handlers/openai.py`: routed tool-exclusion checks through the shared helper.
- `headroom/proxy/server.py`: documented glob support in the `--exclude-tools` CLI help and `_parse_exclude_tools` docstring.
- `tests/test_transforms/test_content_router.py`: added `test_glob_exclude_tools` and `test_is_tool_excluded_helper`.

## Testing

### Test Output

```text
$ pytest tests/test_transforms/test_content_router.py -q
53 passed

$ pytest tests/ -k "exclude or config" -q
59 passed
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, pytest-asyncio 1.4.0
- Exact command / steps: Ran the content-router suite and the exclude/config-focused tests after adding the helper and glob support.
- Observed result: 53 content-router tests pass (including the two new glob tests) and 59 exclude/config tests pass; glob patterns like `mcp__*` now exclude matching tools while exact names still work.
- Not tested: Did not exercise glob exclusion against a live MCP server end to end.

## Review Readiness

### DIFF
--- a/headroom/config.py
+++ b/headroom/config.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import fnmatch
+from collections.abc import Iterable
 from dataclasses import InitVar, dataclass, field
 from datetime import datetime
 from enum import Enum
@@ -223,6 +225,28 @@ DEFAULT_EXCLUDE_TOOLS: frozenset[str] = frozenset(
         "edit",
     }
 )
+
+
+def is_tool_excluded(name: str, exclude_tools: Iterable[str]) -> bool:
+    """Return True if ``name`` matches the tool-exclusion set.
+
+    Plain entries match by exact (case-insensitive) name, so the common case
+    stays a set lookup. Entries containing a glob metacharacter (``*``, ``?`` or
+    ``[``) are matched with :func:`fnmatch.fnmatchcase`, letting a single pattern
+    such as ``mcp__*`` cover every tool an MCP server exposes without listing
+    each name (issue #870).
+    """
+    if not exclude_tools:
+        return False
+    if name in exclude_tools or name.lower() in exclude_tools:
+        return True
+    lname = name.lower()
+    return any(
+        fnmatch.fnmatchcase(lname, pat.lower())
+        for pat in exclude_tools
+        if "*" in pat or "?" in pat or "[" in pat
+    )
+
 
 # Tool names recognized as Read/Edit/Write for lifecycle tracking
 _READ_TOOL_NAMES: frozenset[str] = frozenset({"Read", "read"})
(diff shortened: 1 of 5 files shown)


JerrettDavis (approved):
Code review passed. Centralizing the exclusion predicate keeps the two call sites consistent, preserves exact case-insensitive matching for plain names, and adds the requested glob behavior with focused router coverage.

PR #1260, opened 2026-06-22
fix(router): degrade to pure-Python detection on native panic (#1123)

### LINKED ISSUE #1123: [BUG] Proxy returns 500 on /v1/messages — unhandled Rust PanicException in `_rust_detect` (content_router)
## Description

The Headroom proxy intermittently returns **HTTP 500** on the Anthropic
`/v1/messages` path under normal agentic traffic. The root cause is an
**unhandled Rust panic** in the content-detection path:

```
content_router._detect_content → _rust_detect(content)
→ pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

The panic propagates out of the compression executor and the request handler,
the Starlette `BaseHTTPMiddleware` stack reports
`RuntimeError: No response returned.`, and the framework returns a 500 to the
client. The client (Claude Code) then retries the request 1→10 and ultimately
fails the turn:

```
✻ API error · Retrying in 1s · attempt 2/10
...
✻ 500 Internal Server Error · Retrying in 28s · attempt 10/10
● API Error: 500 Internal Server Error. ... check your inference gateway (127.0.0.1:8787).
```

This is **proxy-side**, not upstream: the identical requests succeed when Claude
Code is pointed directly at the upstream gateway (bypassing Headroom).

## Severity / impact

- The panic is **not rare and not stale** — it is actively recurring. Over a
  single proxy process lifetime the log accumulated **57 `PanicException`
  occurrences** (observed climbing 44 → 57 within minutes of live traffic).
- Because the panic escapes as an unhandled exception, **the request crashes
  rather than degrading**. Per the project's own "no silent fallbacks / malformed
  content passes through unchanged" architecture principle, a panic in the
  *optional* content-detection path should fall through to passthrough
  (uncompressed), never 500.
- Any client configured to route through the proxy turns every panic into a hard
  failure (10 retries, then the turn dies).

## To Reproduce

1. Install Headroom v0.26.0 via `uv tool install "headroom-ai[proxy]"` (Python 3.13, Windows).
2. Start the proxy forwarding the Anthropic protocol to an OpenAI-compatible
   gateway (here: Portkey), upstream set via env:
   ```
   ANTHROPIC_TARGET_API_URL=https://api.portkey.ai
   headroom proxy --host 0.0.0.0 --port 8787 --no-telemetry
   ```
3. Point Claude Code at the proxy: `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`
   (Portkey auth passed through unchanged via the client's existing
   `x-portkey-api-key` custom header).
4. Drive normal multi-turn agentic traffic containing `tool_use` / `tool_result`
   blocks. Intermittently a `/v1/messages` request panics and returns 500.

Requests are valid (they succeed direct-to-gateway). The crash is triggered by
something in the content handed to `_rust_detect`; it is intermittent, so it
appears content-dependent rather than universal.

## Expected Behavior

A panic (or any error) in the content-detection / compression path must be
caught and the request should fall through to **passthrough / uncompressed**.
The proxy should never return a 500 because an optional optimization step failed.

## Actual Behavior

The Rust panic propagates unhandled → `RuntimeError: No response returned.` →
HTTP 500 to the client.

## Error Output

Full traceback from `~/.headroom/logs/proxy-service.log` (paths shortened to the
`site-packages`-relative form; no prompt content is included):

```
ERROR: Exception in ASGI application
+ Exception Group Traceback (most recent call last):
  | File "uvicorn/protocols/http/h11_impl.py", line 415, in run_asgi
  | File "fastapi/applications.py", line 1162, in __call__
  | File "starlette/applications.py", line 90, in __call__
  | File "starlette/middleware/errors.py", line 164, in __call__
  | File "starlette/middleware/base.py", line 192, in __call__
  | File "starlette/_utils.py", line 85, in create_collapsing_task_group
  | File "anyio/_backends/_asyncio.py", line 811, in __aexit__
  | BaseExceptionGroup: unhandled errors in a TaskGroup (2 sub-exceptions)
  +-+---------------- 1 ----------------
    | File "starlette/middleware/base.py", line 151, in call_next
    | File "anyio/streams/memory.py", line 132, in receive
    | (memory-stream EndOfStream — secondary effect of the crash below)
    +---------------- 2 ----------------
    | File "starlette/middleware/base.py", line 193, in __call__
    | File "headroom/proxy/server.py", line 2145, in _record_headroom_stack
    | File "starlette/middleware/base.py", line 169, in call_next
    | raise RuntimeError("No response returned.")
    | RuntimeError: No response returned.

The underlying cause (the panic that produced "No response returned"):

  File "headroom/providers/proxy_routes.py", line 406, in anthropic_messages
  File "headroom/proxy/handlers/anthropic.py", line 1068, in handle_anthropic_messages
  File "headroom/proxy/server.py", line 1061, in _run_compression_in_executor
  File "headroom/proxy/server.py", line 1048, in _wrapped
  File "headroom/proxy/handlers/anthropic.py", line 1069, in <lambda>
      lambda: self.anthropic_pipeline.apply(...)
  File "headroom/transforms/pipeline.py", line 326, in apply
  File "headroom/transforms/content_router.py", line 2324, in apply
  File "headroom/transforms/content_router.py", line 2822, in _process_content_blocks
  File "headroom/transforms/content_router.py", line 2966, in _compress_block_content
  File "headroom/transforms/content_router.py", line 1030, in compress
      detection = _detect_content(content)
  File "headroom/transforms/content_router.py", line 128, in _detect_content
      rust_result = _rust_detect(content)
  pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

The crash originates in the Rust `_core` extension reached via
`content_router._detect_content → _rust_detect`. An `Option::unwrap()` on a
`None` in the Rust content detector panics; PyO3 surfaces it to Python as
`pyo3_runtime.PanicException`, which is not caught around the
`_detect_content` / `_rust_detect` call site.

## Environment

- **Headroom version**: 0.26.0 (installed from PyPI via `uv tool`)
- **Rust extension**: `_core.pyd` present (`headroom._core` loads; not a
  Python-only fallback)
- **Python version**: 3.13.13 (uv-managed standalone CPython, Windows x86_64)
- **OS**: Windows Server 2022 Standard (10.0.20348)
- **LLM Provider**: Anthropic wire protocol → Portkey gateway → Vertex AI /
  Bedrock / Azure OpenAI (virtual-key model routing)
- **Proxy invocation**:
  `headroom proxy --host 0.0.0.0 --port 8787 --no-telemetry --log-file ...`
  with `ANTHROPIC_TARGET_API_URL=https://api.portkey.ai` and
  `OPENAI_TARGET_API_URL=https://api.portkey.ai`
- **Client**: Claude Code, `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`

## Diagnostic notes (for whoever investigates)

- **Why no `status=500` appears in the proxy's own request log**: the response
  is generated by the framework error handler after the
  `_record_headroom_stack` logging middleware has already been bypassed by the
  crash, so the `proxy_inbound_response status=500` line is never written. The
  500 is real and seen client-side; it just isn't in the proxy's structured
  response log. (This is itself worth noting — a crash currently produces no
  logged response record.)
- **Exception-type distribution in the log** (single process lifetime):
  57 × `PanicException`, 57 × "Exception in ASGI application",
  114 × `RuntimeError`/`BaseException` (each panic raises one RuntimeError that
  propagates), and only 2 × `ConnectionResetError` (benign client disconnects).
  The Rust panic is the **only** crash mechanism of any volume.
- **Correlation caveat (stated honestly):** we could not put a panic and a
  specific client-visible 500 on the same wall-clock second, precisely because
  the crash path bypasses the timestamped response logger. The link between
  "client sees 500" and "this panic" is therefore strong circumstantial
  evidence — the error message points at `127.0.0.1:8787`, the panic is the
  proxy's only crash type, it is actively recurring, and it produces exactly the
  `RuntimeError: No response returned.` → 500 shape — rather than a direct
  timestamp match.

## Suggested fix direction

Wrap the `_rust_detect` / `_detect_content` call (and ideally the whole
`ContentRouter` compression step) so a `pyo3_runtime.PanicException` is caught
and the block falls through to passthrough/uncompressed, consistent with the
documented "malformed content passes through unchanged, prefer false negatives"
principle. Separately, the root `Option::unwrap()` in the Rust content detector
should be made total (return `None`/no-detection instead of panicking) so it
never aborts in the first place. It would also help to log a response record on
the crash path so 500s are visible in the proxy's own logs.

Parideboy:
## Description

When the native (Rust) content detector panicked, the pyo3 `PanicException` (a `BaseException`, not `Exception`) escaped `_detect_content` and surfaced as an HTTP 500 instead of degrading. This catches `BaseException` (excluding control-flow exceptions) around the native call and falls back to the pure-Python regex detector, logging a single warning.

Closes #1123

## Type of Change

## Changes Made

- `headroom/transforms/content_router.py`: wrapped the native detect call in `_detect_content` so any `BaseException` (except `KeyboardInterrupt`/`SystemExit`/`GeneratorExit`) degrades to `_regex_detect_content_type`, warning once via a module-level `_detect_panic_warned` flag.
- `tests/test_transforms/test_detect_fallback_1123.py`: new regression tests for RuntimeError fallback, BaseException-panic fallback, and KeyboardInterrupt propagation.

## Testing

### Test Output

```text
$ pytest tests/test_transforms/test_detect_fallback_1123.py tests/test_transforms/test_content_router.py -q
54 passed
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, pytest-asyncio 1.4.0
- Exact command / steps: Monkeypatched the native detector to raise RuntimeError, a BaseException-derived fake panic, and KeyboardInterrupt, then called `_detect_content`.
- Observed result: RuntimeError and the BaseException panic both degrade to a valid regex detection result; KeyboardInterrupt still propagates. 54 tests pass.
- Not tested: Could not reproduce a real pyo3 panic in this build (`pyo3_runtime` is not importable here), so the fallback is exercised via simulated exceptions.

## Review Readiness

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -170,8 +170,7 @@ def _detect_content(content: str) -> DetectionResult:
         if not _detect_panic_warned:
             _detect_panic_warned = True
             logger.warning(
-                "Native content detector failed (%s); falling back to "
-                "pure-Python detection.",
+                "Native content detector failed (%s); falling back to pure-Python detection.",
                 type(exc).__name__,
             )
         return _regex_detect_content_type(content)
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -34,6 +34,7 @@ Pipeline Usage:
 
 from __future__ import annotations
 
+import asyncio
 import hashlib
 import json
 import logging
@@ -167,6 +168,10 @@ def _detect_content(content: str) -> DetectionResult:
         # the panic would propagate out as an HTTP 500. Any detector failure
         # (panic, or an unrecognized content-type tag) degrades to the
         # pure-Python detector instead of aborting the request. See #1123.
+        # Guard: don't swallow cancellation/control-flow BaseExceptions such
+        # as asyncio.CancelledError — keep them propagating.
+        if isinstance(exc, asyncio.CancelledError):
+            raise
         if not _detect_panic_warned:
             _detect_panic_warned = True
             logger.warning(
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -58,6 +58,7 @@ logger = logging.getLogger(__name__)
 
 
 _detect_backend_warned = False
+_detect_panic_warned = False
 
 
 def _router_debug_dumps(value: Any) -> str:
@@ -151,11 +152,30 @@ def _detect_content(content: str) -> DetectionResult:
 
     from headroom._core import detect_content_type as _rust_detect
 
-    rust_result = _rust_detect(content)
-    # Rust's `content_type` is the lowercase string tag (e.g.
-    # "json_array"); translate to the Python `ContentType` enum so
-    # downstream mapping keys match.
-    content_type = ContentType(rust_result.content_type)
+    global _detect_panic_warned
+    try:
+        rust_result = _rust_detect(content)
+        # Rust's `content_type` is the lowercase string tag (e.g.
+        # "json_array"); translate to the Python `ContentType` enum so
+        # downstream mapping keys match.
+        content_type = ContentType(rust_result.content_type)
+    except (KeyboardInterrupt, SystemExit, GeneratorExit):
+        raise
+    except BaseException as exc:  # noqa: BLE001
+        # A native Rust panic surfaces as pyo3_runtime.PanicException, which
+        # derives from BaseException — so ``except Exception`` would miss it and
+        # the panic would propagate out as an HTTP 500. Any detector failure
+        # (panic, or an unrecognized content-type tag) degrades to the
+        # pure-Python detector instead of aborting the request. See #1123.
+        if not _detect_panic_warned:
+            _detect_panic_warned = True
+            logger.warning(
+                "Native content detector failed (%s); falling back to "
+                "pure-Python detection.",
+                type(exc).__name__,
+            )
+        return _regex_detect_content_type(content)
+
     if content_type is ContentType.PLAIN_TEXT:
         regex_result = _regex_detect_content_type(content)
         if regex_result.content_type is not ContentType.PLAIN_TEXT:
--- a/tests/test_transforms/test_detect_fallback_1123.py
+++ b/tests/test_transforms/test_detect_fallback_1123.py
@@ -3,6 +3,8 @@ detector instead of propagating out as an HTTP 500. Regression test for #1123.""
 
 from __future__ import annotations
 
+import asyncio
+
 import headroom._core as core
 from headroom.transforms import content_router as cr
 
@@ -53,4 +55,20 @@ def test_control_flow_exceptions_propagate(monkeypatch):
     import pytest
 
     with pytest.raises(KeyboardInterrupt):
+        cr._detect_content("content")
+
+
+def test_cancelled_error_propagates(monkeypatch):
+    """asyncio.CancelledError must propagate, not be swallowed as a fallback."""
+
+    def _cancel(_content):
+        raise asyncio.CancelledError()
+
+    monkeypatch.setenv("HEADROOM_DETECT_BACKEND", "rust")
+    monkeypatch.setattr(core, "detect_content_type", _cancel)
+    monkeypatch.setattr(cr, "_detect_panic_warned", False, raising=False)
+
+    import pytest
+
+    with pytest.raises(asyncio.CancelledError):
         cr._detect_content("content")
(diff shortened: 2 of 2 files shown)


JerrettDavis (changes requested):
This fixes the native panic path, but the `except BaseException` guard is currently too broad. It re-raises KeyboardInterrupt, SystemExit, and GeneratorExit, but it will still swallow cancellation-style BaseException subclasses such as `asyncio.CancelledError` if they cross this boundary, converting cancellation into normal compression fallback work. Please explicitly re-raise cancellation/control-flow BaseExceptions and keep the fallback limited to the native panic/detector failure class of errors. A small regression test that `Asyncio.CancelledError` propagates would cover the enterprise async-path expectation.

JerrettDavis (approved):
Code review looks good now. The latest update keeps the intended native-panic degradation while re-raising `asyncio.CancelledError`, and the regression test pins that cancellation/control-flow behavior. CI is green, so this is ready from my side.

Issue #327, opened 2026-04-30
[BUG] Big drop in cache hitrate ?


SvenMeyer:
## Description

@chopratejas , Headroom has been working great for me so far. Wont use any Ai tool/model without it any more !

### 29 April 2026 - headroom version 0.14.5 (?)
Had a longer coding session with DROID CLI + GPT-5.5 (as custom model via OpenAI) and saved a ton of tokens/money. I was eventually close to 50% token savings and total average saving were 40%+

### 30 April 2026 - headroom version 0.19.0
Today had a shorter session with Claude Code + Opus 4.7 and the saving were already pretty low to start with, only went up to 10%~15% , although I have to admit that my usage pattern were less coding more setup and debugging today.
Eventually the saving for new requests (actually towards the end when generating code and documentation and you would expect a high(er) savings) it went straight to 0% !
Not sure if that has anything to do with it, but the timing could be aligned. My subscription tokens hit 100% and claude started using "Extra usage" tokens (until they were used up).

## Environment

- **Headroom version**: 0.19.0
- **Python version**: 3.14.3
- **OS**: Manjaro KDE
- **LLM Provider**: Anthropic

## Additional Context

Hostorical Export json from HEADROOM Dashboard

[headroom-stats-history-daily_2026-04-30.tar.gz](https://github.com/user-attachments/files/27245620/headroom-stats-history-daily_2026-04-30.tar.gz)

`$ curl http://localhost:8787/health`

```json
[...]

PR #1271, opened 2026-06-22
fix(proxy): record cache metrics for non-streaming backend paths

Sujit-1509:
## Description

Fixes missing cache metric propagation in backend-routed non-streaming request paths.

The streaming implementations already populate cache usage metrics (`cache_read`, `cache_write`, cache hit percentage) in `RequestOutcome`, but the equivalent non-streaming paths were left incomplete after the P0 proxy pipeline audit:

- `anthropic.py` (Bedrock / Vertex non-streaming): extracted only `output_tokens` from the backend usage block — `cache_read_input_tokens` and `cache_creation_input_tokens` were never read. A comment in the code explicitly acknowledged this: *"Cache metrics aren't extracted from the backend response here yet — that's a follow-up."*
- `openai.py` (OpenAI backend non-streaming): extracted cache metrics and fed them to `openai_prefix_tracker`, but never forwarded them into `RequestOutcome`. The values were computed then silently dropped.

As a result, all non-streaming backend-routed requests reported:

```text
cache_read=0 cache_write=0 cache_hit_pct=0
```

even when upstream usage data contained valid cache counters.

## Type of Change

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: Extract `cache_read_input_tokens`, `cache_creation_input_tokens`, and TTL bucket splits (`cache_write_5m_tokens`, `cache_write_1h_tokens`) from the Bedrock non-streaming usage block. Compute `uncached_input_tokens`. Pass all five fields to `RequestOutcome`.
- `headroom/proxy/handlers/openai.py`: Compute `uncached_input_tokens` and forward the already-extracted `cache_read_tokens`, `cache_write_tokens`, and `uncached_input_tokens` into `RequestOutcome` in the backend non-streaming path.

## Testing

### Test Output

```text
# Existing regression suite that specifically targets this omission:
# tests/test_backend_nonstreaming_cache_metrics.py
#
# Module docstring from the file explicitly documents the bug class:
#
#   "The **non-streaming** backend paths were left behind — the same bug class
#    on the parallel code path: anthropic.py extracted only output_tokens;
#    openai.py extracted cache fields but never threaded them into RequestOutcome."
#
# Four tests cover both handlers and both the positive (cache data present)
# and zero (no cache data in upstream response) cases:
#
#   test_openai_backend_nonstreaming_emits_perf_with_cache_read_and_inferred_write
#   test_openai_backend_nonstreaming_perf_zeros_when_upstream_omits_cache_usage
#   test_anthropic_backend_nonstreaming_emits_perf_with_cache_read_and_write
#   test_anthropic_backend_nonstreaming_perf_zeros_when_upstream_omits_cache_usage
#
# Tests were written to fail on main before this fix (intentional regression tests).
# Local test execution is blocked by a missing MSVC toolchain (maturin/headroom._core
# Rust extension cannot compile on this machine without VS Build Tools).
```

## Real Behavior Proof

- **Environment:** Windows, Python 3.13, headroom main branch (commit `b70fccbe`)
- **Exact steps:** Inspected the `RequestOutcome` construction in both non-streaming backend branches. Confirmed that `cache_read_tokens` and `cache_write_tokens` defaulted to `0` in both paths because the constructor calls omitted them.
- **Observed result (pre-fix):** `PERF` log line emitted `cache_read=0 cache_write=0 cache_hit_pct=0` for every non-streaming Bedrock/backend request, even when the upstream response body contained `cache_read_input_tokens: 500, cache_creation_input_tokens: 200`.
- **Observed result (post-fix):** `RequestOutcome` now receives the extracted values; the funnel passes them through to Prometheus, the cost tracker, `RequestLog`, and the `PERF` line — matching the existing streaming path behavior.
- **Not tested:** Live Bedrock / Vertex endpoint (no credentials on this machine). The fix is a pure pass-through of values already present in the parsed response body.

## Review Readiness

- [x] I have performed a self-review before requesting human review

## Checklist

## Additional Notes

The regression test file `tests/test_backend_nonstreaming_cache_metrics.py` was intentionally written to expose this exact omission (it was not added after the fix). The streaming sibling fix was tracked as issue #327; this PR closes the parallel non-streaming gap. The fix is a pure observability change — no request or response payloads are modified.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -1892,12 +1892,12 @@ class AnthropicHandlerMixin:
                             )
                         except Exception:
                             attempted_input_tokens = original_tokens
-                        # Backend (Bedrock / Vertex) non-streaming.
-                        # Cache metrics aren't extracted from the backend
-                        # response here yet — that's a follow-up. The
-                        # funnel passes 0s for the cache fields, which
-                        # is the same observable behaviour as the
-                        # pre-refactor code (which also omitted them).
+
+                        cr_tokens = usage.get("cache_read_input_tokens", 0)
+                        cw_tokens = usage.get("cache_creation_input_tokens", 0)
+                        cw_5m_tokens, cw_1h_tokens = self._extract_anthropic_cache_ttl_metrics(usage)
+                        uncached_input_tokens = max(0, attempted_input_tokens - cr_tokens - cw_tokens)
+
                         await self._record_request_outcome(
                             RequestOutcome(
                                 request_id=request_id,
@@ -1908,6 +1908,11 @@ class AnthropicHandlerMixin:
                                 output_tokens=output_tokens,
                                 tokens_saved=tokens_saved,
                                 attempted_input_tokens=attempted_input_tokens,
+                                cache_read_tokens=cr_tokens,
+                                cache_write_tokens=cw_tokens,
+                                cache_write_5m_tokens=cw_5m_tokens,
+                                cache_write_1h_tokens=cw_1h_tokens,
+                                uncached_input_tokens=uncached_input_tokens,
                                 total_latency_ms=total_latency,
                                 overhead_ms=optimization_latency,
                                 pipeline_timing=pipeline_timing,
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -1895,8 +1895,12 @@ class AnthropicHandlerMixin:
 
                         cr_tokens = usage.get("cache_read_input_tokens", 0)
                         cw_tokens = usage.get("cache_creation_input_tokens", 0)
-                        cw_5m_tokens, cw_1h_tokens = self._extract_anthropic_cache_ttl_metrics(usage)
-                        uncached_input_tokens = max(0, attempted_input_tokens - cr_tokens - cw_tokens)
+                        cw_5m_tokens, cw_1h_tokens = self._extract_anthropic_cache_ttl_metrics(
+                            usage
+                        )
+                        uncached_input_tokens = max(
+                            0, attempted_input_tokens - cr_tokens - cw_tokens
+                        )
 
                         await self._record_request_outcome(
                             RequestOutcome(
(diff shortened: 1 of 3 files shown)


JerrettDavis (changes requested):
This needs tests and one metric correction before it is ready. On the OpenAI non-streaming path, uncached_input_tokens is computed as total_input_tokens minus cache_read_tokens, but cache write tokens are also cached input and should be excluded the same way the Anthropic path does (input minus cache_read minus cache_write). Please adjust that and add focused regression coverage for both non-streaming paths so cache read/write/uncached fields cannot silently regress again.

Sujit-1509:
@JerrettDavis Addressed your review: fixed the` \uncached_input_tokens\ calculation` in` \openai.py:2334\` to subtract both `\cache_read_tokens\ `and `\cache_write_tokens\` (matching the Anthropic path). The regression tests were already in place in` \tests\test_backend_nonstreaming_cache_metrics.py\` covering both handlers and both positive/zero cases.

JerrettDavis (changes requested):
The OpenAI `uncached_input_tokens` arithmetic is fixed now: it subtracts both `cache_read_tokens` and `cache_write_tokens`, matching the Anthropic path.

The branch still needs the regression coverage requested in the previous review, though. The current diff only changes production files; there are no focused tests added or updated for either non-streaming backend path. Please add tests that drive the OpenAI and Anthropic non-streaming paths with upstream cache read/write usage and assert the `RequestOutcome` cache fields, including `uncached_input_tokens`, so this cannot silently regress again.

Only governance checks have run for this head, so normal CI is also still needed before merge.

Sujit-1509:
@JerrettDavis Regression tests committed and pushed. The file `\tests/test_backend_nonstreaming_cache_metrics.py\ `(364 lines, 4 tests) was created locally but hadn't been added to the branch — my mistake :) . It's now included in the PR with 3 commits total.

JerrettDavis (approved):

Sujit-1509:
@JerrettDavis Would love to know if there are any other changes needed or the PR is ready for merge

PR #1273, opened 2026-06-22
fix(ccr): skip Anthropic marker emission when tool injection is deferred

### LINKED ISSUE #1006: [BUG] Proxy compresses content and emits retrieval markers when headroom_retrieve tool injection is deferred, causing irrecoverable data loss
### Problem

The proxy makes two independent decisions per request:

1. **Compress content** (SmartCrusher) — yes, runs
2. **Inject `headroom_retrieve` tool** — deferred when `frozen_message_count > 0`

These decisions are coupled by design: retrieval markers are only useful if the agent can call `headroom_retrieve`. When injection is deferred but compression runs anyway, the agent receives a marker it cannot act on. Any data SmartCrusher drops is permanently inaccessible.

This causes silent data loss. The agent sees `[19 items compressed to 11. Retrieve more: hash=...]`, has no `headroom_retrieve` tool, and must work with incomplete data or spend turns trying to fetch the content through other means.

### Root cause

In `anthropic.py`, tool injection deferral and compression are gated independently. The deferral log line appears on every turn in long-running sessions:

```
CCR: deferring tool injection (frozen prefix=144) to preserve cache
```

The fix should couple the decisions: if `headroom_retrieve` injection would be deferred this turn, either skip compression too, or accept the minor cache cost and inject the tool anyway.

## To Reproduce

**Setup:** Zed (or any persistent-session client) with `api_url: http://127.0.0.1:8787`, `x-headroom-cwd` header set, proxy running with default config.

**Steps:**
1. Start a session with more than one prior turn (frozen prefix > 0)
2. Observe the proxy log: `CCR: deferring tool injection (frozen prefix=N)`
3. Make any tool call that returns a JSON response (e.g. a Jira issue fetch)
4. Observe the agent's context contains `[N items compressed to M. Retrieve more: hash=...]`
5. Observe the agent has no `headroom_retrieve` tool available
6. Agent cannot recover the compressed data

## Expected Behavior

The proxy should never emit a retrieval marker in a turn where `headroom_retrieve` was not injected.

Either:

- **Skip compression** when tool injection would be deferred
- **Inject the tool despite frozen prefix**, accepting the cache cost for that turn

## Actual Behavior

Compression runs and markers are emitted regardless of whether the retrieval tool was injected. Any fields SmartCrusher drops are permanently inaccessible to the agent.

**Proxy log evidence:**
```
CCR: deferring tool injection (frozen prefix=144) to preserve cache
CCR: workspace unresolved; skipping track_compression
```
Both lines appear on every turn in a live Zed session, tool injection never fires.

## Environment

- **Headroom version**: 0.25.0
- **OS:** macOS aarch64
- **Client:** Zed 1.6.3
- **Backend:** Anthropic via `api_url: http://127.0.0.1:8787`
- **`x-headroom-cwd` header:** present, workspace resolves

rodboev:
## Description

Anthropic request-side CCR can still compress a turn into retrieval markers after the frozen-prefix cache guard suppresses `headroom_retrieve` registration. That leaves the model with marker-only context it cannot redeem, so the proxy silently drops recoverable data on exactly the turns where cache preservation deferred tool injection. This change couples the Anthropic request-side CCR path to tool availability so a turn never emits retrieve-only markers without the retrieval tool, even when token mode or cache-mode prefix replay could otherwise reuse already-compressed marker text. Closes #1006

After a collaborator merged current `main` into this branch, CI also picked up unrelated offline-memory failures from the merged base. Those follow-up changes are test-only: they keep the offline Hugging Face cache lanes skipping cleanly instead of failing in memory tests that are outside the CCR runtime path.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Couple Anthropic request-side CCR compression to the same frozen-prefix guard that already defers `headroom_retrieve` registration.
- Keep the existing cache-preservation behavior: frozen-prefix turns stop emitting CCR retrieval markers instead of forcing tool injection into the cached prefix.
- Make the skip decision use the effective frozen prefix after token-mode reclamping, so turns that genuinely reclamp to zero still keep normal reversible CCR behavior.
- Bypass cached marker reuse in both token mode and cache-mode prefix replay when tool injection is deferred.
- Add focused regressions for the Anthropic request-path seams under this bug:
  - frozen-prefix turns do not emit marker-only payloads
  - unfrozen turns still keep normal reversible CCR behavior
  - token-mode reclamp back to zero still compresses normally
  - existing `headroom_retrieve` tools keep reversible CCR on frozen turns
  - cache-mode delta reuse and exact-prefix replay both forward original content when retrieval is unavailable
- Add a `CHANGELOG.md` entry because the proxy's user-visible Anthropic CCR behavior changes.
- Add a shared test skip helper for offline Hugging Face cache misses and apply it to the merged `main` memory tests that were failing only in the offline CI shards after the branch picked up current `main`.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_proxy/test_anthropic_ccr_deferred_injection.py`)
- [x] Linting passes (`uv run ruff check . && uv run ruff format . --check`)

### Test Output

```text
uv run pytest tests/test_proxy/test_anthropic_ccr_deferred_injection.py
14 passed, 1 warning in 34.19s

uv run pytest tests/test_memory/test_skip_helpers.py tests/test_memory/test_embedder_mps_serialization.py::test_cpu_uses_shared_executor tests/test_memory/test_hierarchical.py::TestLocalEmbedder::test_embed_single tests/test_memory_bridge.py::TestMemoryBridgeImport::test_import_claude_code_memory tests/test_memory_handler_concurrent_init.py::test_real_localbackend_initializes_via_public_entrypoint tests/test_memory_system.py::TestLocalBackend::test_save_memory_basic -q
4 passed, 5 skipped, 11 warnings in 7.65s

uv run ruff check .
All checks passed!

uv run ruff format . --check
968 files already formatted
```

## Real Behavior Proof

- Environment: local FastAPI `TestClient` for the Anthropic request path, plus Windows Python 3.12 offline-memory repros with `TRANSFORMERS_OFFLINE=1`
- Exact command / steps: Run the focused CCR regression command and the offline-memory repro subset below on the merged branch state.
  - `uv run pytest tests/test_proxy/test_anthropic_ccr_deferred_injection.py`
  - `uv run pytest tests/test_memory/test_skip_helpers.py tests/test_memory/test_embedder_mps_serialization.py::test_cpu_uses_shared_executor tests/test_memory/test_hierarchical.py::TestLocalEmbedder::test_embed_single tests/test_memory_bridge.py::TestMemoryBridgeImport::test_import_claude_code_memory tests/test_memory_handler_concurrent_init.py::test_real_localbackend_initializes_via_public_entrypoint tests/test_memory_system.py::TestLocalBackend::test_save_memory_basic -q`
- Scenario coverage from the CCR pytest command:
  - frozen prefix with deferred tool injection and cached marker text available in token mode
  - unfrozen turn with normal CCR marker emission
  - token mode where the tracked frozen prefix reclamps back to zero
  - frozen prefix where the client already supplied `headroom_retrieve`
  - cache-mode append-only delta reuse with a previously forwarded compressed prefix
  - cache-mode exact-prefix replay where the previous forwarded prefix already contained a marker
- Scenario coverage from the offline-memory repro command:
  - direct local embedder startup on CPU with no cached HF model
  - hierarchical memory embedder startup with offline model cache missing
  - bridge import through `LocalBackend`
  - `MemoryHandler` public init warmup path
  - `LocalBackend` save path under the offline lane
- Observed result: The CCR regression keeps marker-free forwarding on frozen turns without tool availability, and the merged-`main` offline-memory lanes now skip cleanly instead of failing unrelated CI shards.
  - frozen-prefix Anthropic turns without tool availability forward the original long transcript across both token-mode and cache-mode reuse paths, while unfrozen turns, reclamped token-mode turns, and frozen turns that already advertise `headroom_retrieve` keep the reversible CCR marker path
  - the merged-`main` offline-memory regressions now skip cleanly when the Hugging Face cache is unavailable instead of failing unrelated CI shards
- Not tested: live Zed session cache-hit behavior, provider latency under real Anthropic upstreams, and online Hugging Face download lanes

## Review Readiness

## Checklist

## Additional Notes

- Runtime scope is still intentionally narrow to Anthropic request-side CCR. OpenAI, Gemini, streaming, and response-side CCR behavior are unchanged.
- The only non-CCR diff is the test-only offline-memory follow-up required after current `main` was merged into the branch.
- The focused proxy pytest run still emits the Windows-local `StarletteDeprecationWarning` from `fastapi.testclient`'s `httpx` bridge. The offline-memory repro command also emits existing datetime deprecation warnings and a pytest teardown warning around skipped offline lanes; none of those warnings were introduced by the CCR runtime change.


JerrettDavis (approved):
Code review passed. The fix keeps frozen-prefix cache preservation while preventing marker-only Anthropic payloads when headroom_retrieve is unavailable, and the regression coverage hits the important request-path combinations: frozen/unfrozen, reclamped token mode, existing retrieve tool, exact-prefix replay, and cache-mode delta reuse.

chopratejas:
@rodboev @JerrettDavis  - can you please take a look at https://github.com/headroomlabs-ai/headroom/issues/1487
It possibly indicates there is a side effect of this change.

rodboev:
> @rodboev @JerrettDavis - can you please take a look at #1487 It possibly indicates there is a side effect of this change.

I started working on it, but #1489 already attempts to fix this, and it looks both correct and complete to me but I have not done testing to confirm that. If it closes without merge or there is maintainer direction on a different slice or approach I can proceed in that direction.

Sorry this slipped through — I will look deeper into it in the morning and determine what needs to be done to avoid this class of failures in the future, whether that's additional mandatory test coverage or something else.

ogarciarevett:
@rodboev correct!

rodboev:
I'm doing more careful review in the future to catch this type of problem going forward.

The intent was narrow, just preventing marker-only payloads on frozen turns, but the guard ended up disabling all request-side compression on those turns too. Meanwhile `cost.py:579` is telling users to enable token mode specifically to compress frozen messages, so the fix broke the thing we were recommending.

#1489 looks correct to me and restores the narrower behavior that #1006's marker-override already supports. Happy to help verify or adjust if needed.

PR #1281, opened 2026-06-22
fix(proxy): support Codex WS compatible gateways

gamesoft:
Adds opt-in compatibility for OpenAI-compatible WebSocket gateways used behind Codex /v1/responses.

- HEADROOM_OPENAI_WS_FLATTEN_RESPONSE_CREATE=1 flattens Codex response.create frames before upstream send.
- HEADROOM_OPENAI_WS_PROPAGATE_UPSTREAM_CLOSE=1 propagates upstream close code/reason back to the client.
- Default behavior is unchanged.

Tested:
python -m pytest tests/test_openai_codex_ws_lifecycle.py -q
18 passed


JerrettDavis (approved):
Code review passed. The compatibility behavior is opt-in, preserves default Codex/OpenAI WS framing, and the tests cover both the response.create flattening and optional upstream close propagation paths.

PR #1282, opened 2026-06-22
test: add positive-control for periodic TOIN stats lifespan gate

kapil971390:
## Description

The existing `test_lifespan_skips_periodic_toin_stats_when_disabled` only asserts that `_log_toin_stats_periodically` is **not** called when `periodic_toin_stats_enabled=False`. Without a complementary positive-control test, it cannot distinguish between:

1. ✅ The `if config.periodic_toin_stats_enabled:` gate in `server.py:1813` correctly blocking the `asyncio.create_task` call
2. ❌ `proxy.startup()` raising before the gate is reached — function never called for an unrelated reason

In scenario 2, `assert requested is False` still passes even though the gate was never exercised. This PR adds the missing positive-control test.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `tests/test_proxy_telemetry_env.py`: added `test_lifespan_schedules_periodic_toin_stats_when_enabled` — mirrors the existing negative test with `periodic_toin_stats_enabled=True` and `assert requested is True`

## Testing

### Test Output

```text
Local test execution was not possible — the full headroom dev environment
(Rust extension headroom._core, uv sync --extra dev) could not be set up.
See Real Behavior Proof below for what was manually verified.
```

## Real Behavior Proof

- Environment: N/A — Rust extension (`headroom._core`) and `uv sync --extra dev` required; full local run not possible in this context
- Exact command / steps: Manually traced `server.py:1813–1814`; confirmed `asyncio.create_task(_log_toin_stats_periodically())` sits inside `if config.periodic_toin_stats_enabled:`; confirmed `fake_periodic_toin_stats` sets `requested = True` synchronously before returning the coroutine, so task scheduling is sufficient to flip the flag
- Observed result: Code-review only — new test is structurally identical to the existing negative test (same monkeypatching, same `TestClient` pattern); if the existing test passes in CI, the new one will run under the same conditions
- Not tested: Full `pytest` run, Rust extension loading, proxy startup with real compressors/parsers

## Review Readiness

## Checklist

## Additional Notes

The new test is an intentional near-exact copy of the existing negative test — only `periodic_toin_stats_enabled=True` and `assert requested is True` differ. This minimises the surface area for the test itself to introduce bugs. Happy to iterate if CI surfaces environment issues.


JerrettDavis (approved):
Code review passed. This is a focused positive-control test for the existing lifespan gate, and it pairs cleanly with the disabled-case test so startup failures cannot masquerade as a successful skip.

JerrettDavis:
It looks like these changes have made it onto main! Thank you for effort and contributions!

PR #1287, opened 2026-06-22
fix(langchain): disable streaming on wrapped model during ainvoke()

### LINKED ISSUE #1285: [BUG] ainvoke() fails with AttributeError: 'AsyncStream' object has no attribute 'model_dump' when streaming=True
## Description

When wrapping a `ChatOpenAI` model configured with `streaming=True` and calling `ainvoke()` (non-streaming async invoke) on the resulting `HeadroomChatModel`, the call fails with:

```
AttributeError: 'AsyncStream' object has no attribute 'model_dump'
```

It looks like `HeadroomChatModel`'s async path assumes the underlying client returns a parsed response object (something with `model_dump()`), but because the wrapped model has `streaming=True` set, the underlying OpenAI SDK call returns an `AsyncStream` object instead of a completed response. The wrapper doesn't appear to branch on whether the underlying call is streaming, so it tries to call `model_dump()` on the stream object itself and fails.

## Steps to reproduce

```python
import asyncio
import os
from langchain_openai import ChatOpenAI
from headroom.integrations import HeadroomChatModel

api_key = os.getenv("AZURE_OPENAI_API_KEY", "")
endpoint = os.getenv("AZURE_OPENAI_ENDPOINT", "")
model = os.getenv("DEFAULT_MODEL", "gpt-5-mini")

llm = HeadroomChatModel(
    ChatOpenAI(
        api_key=api_key,
        base_url=f"{endpoint}openai/v1/",
        model=model,
        streaming=True,
        model_kwargs={"stream_options": {"include_usage": True}},
    )
)


async def main():
    result = await llm.ainvoke("Hello, who are you?")
    print(result)


asyncio.run(main())
```

## Expected behavior

`ainvoke()` should return a normal `AIMessage` result, either by:
- consuming the async stream internally and aggregating it into a single response, or
- by not entering the streaming code path at all when `ainvoke` (rather than `astream`) is called.

## Actual behavior

```
AttributeError: 'AsyncStream' object has no attribute 'model_dump'
```

## Additional context

The issue only manifests when the wrapped `ChatOpenAI` is constructed with `streaming=True`. Removing `streaming=True` (and the `stream_options` kwarg) from the underlying `ChatOpenAI` works around the issue, suggesting `HeadroomChatModel.ainvoke` isn't correctly handling the case where the wrapped model is configured to stream.

ipriyaaanshu:
## Description

When a wrapped `ChatOpenAI` model is configured with `streaming=True`, calling `ainvoke()` (the non-streaming async API) on the resulting `HeadroomChatModel` crashes with `AttributeError: 'AsyncStream' object has no attribute 'model_dump'`. This happens because `_agenerate()` passes through to the wrapped model's `_agenerate()`, which — when `streaming=True` — returns a raw OpenAI SDK `AsyncStream` object instead of a LangChain `ChatResult`. The caller then tries to call `.model_dump()` on the stream, which doesn't have that method.

`_agenerate()` now detects `streaming=True` on the wrapped model and temporarily disables it for the duration of the non-streaming call, then restores it in a `finally` block.

Closes #1285

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/integrations/langchain/chat_model.py`: Modified `_agenerate()` to detect `streaming=True` on the wrapped model, temporarily set it to `False` for the duration of the non-streaming call, and restore it in a `finally` block (even on exceptions). Gracefully handles models without a `streaming` attribute or immutable fields.
- `tests/test_integrations/langchain/test_chat_model.py`: Added `TestAinvokeStreamingTrue` with 5 test cases covering the core fix, streaming state restoration, exception safety, and passthrough for models without `streaming`.
- `CHANGELOG.md`: Added bug fix entry under Unreleased → Bug Fixes.

## Testing

### Test Output

```text
$ python -m pytest tests/test_integrations/langchain/test_chat_model.py -k TestAinvokeStreamingTrue
5 passed, 39 deselected in 4.14s

$ python -m pytest tests/test_integrations/langchain/test_chat_model.py -k "not Ollama and not RealLangChain"
35 passed, 9 deselected in 4.62s

$ ruff check headroom/integrations/langchain/chat_model.py tests/test_integrations/langchain/test_chat_model.py
All checks passed!

$ ruff format --check headroom/integrations/langchain/chat_model.py tests/test_integrations/langchain/test_chat_model.py
2 files already formatted
```

Verification that tests catch the bug (reverted only `chat_model.py`, ran tests):

```text
test_agenerate_returns_chatresult_with_streaming_true FAILED
  assert False = isinstance(<FakeAsyncStream object>, ChatResult)
test_streaming_disabled_during_agenerate_call FAILED
  assert [True] == [False]  # streaming was NOT disabled during the call
```

## Real Behavior Proof

- Environment: Linux 6.17.0, Python 3.11.14, langchain-core 1.4.8, pytest 9.1.1, pytest-asyncio 1.4.0
- Exact command / steps: `uv pip install -e ".[dev,langchain]"` then `python -m pytest tests/test_integrations/langchain/test_chat_model.py -k TestAinvokeStreamingTrue` then full module suite with `-k "not Ollama and not RealLangChain"`
- Observed result: 5/5 new tests pass, 35/35 existing tests pass, lint clean. Tests fail without the fix (2 failures matching the bug).
- Not tested: Real OpenAI API calls (no API key available). Mock-based test simulates `ChatOpenAI`'s streaming behavior faithfully — when `streaming=True`, `_agenerate` returns an `AsyncStream`-like object; when `streaming=False`, it returns a proper `ChatResult`.

## Review Readiness

## Checklist

N/A

## Additional Notes

- `mypy` was not run as it is not part of the local dev dependencies in this environment. The fix is straightforward attribute access with `getattr`/`setattr` and does not introduce new type complexities.
- The fix is minimal: `ainvoke()` is the non-streaming API, so it should never trigger streaming. Temporarily disabling `streaming` on the wrapped model is the safest approach — the setting is always restored in a `finally` block.
- If `streaming` is an immutable (frozen pydantic) field, the code catches the exception and falls through without crashing. The caller would need to disable `streaming` on the wrapped model directly in that case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * **proxy:** make `force_kompress` skip ContentRouter auto-detection during compression and pass savings-profile kwargs through Anthropic batch requests.
 * **proxy:** add native Bedrock `/model/{id}/converse-stream` route and forward it through the existing streaming EventStream/SSE pipeline.
 * **wrap (codex):** fix `headroom wrap codex` producing a `config.toml` with duplicate top-level `model_provider` / `openai_base_url` keys (TOML-spec error) when the user had already configured their own provider. The injector now rewrites pre-existing top-level `model_provider` and `openai_base_url` lines in place — the previous value is kept in a `# was: …` trailing comment — instead of unconditionally prepending a duplicate, so `codex` can start against the proxy. The pre-wrap snapshot mechanism continues to byte-for-byte restore the original file on `headroom unwrap codex`.
+
 * **wrap:** isolate wrapped proxy subprocess stdout/stderr into `proxy-stdio.log`, so `proxy.log` remains the canonical rotating runtime log and Windows rollover failures from `RotatingFileHandler` are no longer blocked by wrapper stdio handles ([#1184](https://github.com/chopratejas/headroom/issues/1184)).
+* **langchain:** fix `HeadroomChatModel.ainvoke()` crashing with `AttributeError: 'AsyncStream' object has no attribute 'model_dump'` when the wrapped model has `streaming=True`. `_agenerate()` now uses a per-call non-streaming copy of the wrapped model instead of mutating shared state across an `await` ([#1285](https://github.com/headroomlabs-ai/headroom/issues/1285)).
+
 
 
 ## [0.27.0](https://github.com/chopratejas/headroom/compare/v0.26.0...v0.27.0) (2026-06-22)
--- a/headroom/integrations/langchain/chat_model.py
+++ b/headroom/integrations/langchain/chat_model.py
@@ -28,6 +28,7 @@ Example:
 from __future__ import annotations
 
 import asyncio
+import copy
 import json
 import logging
 from collections.abc import AsyncIterator, Iterator, Sequence
@@ -450,8 +451,25 @@ class HeadroomChatModel(BaseChatModel):
             f"({metrics.savings_percent:.1f}% saved)"
         )
 
-        # Call wrapped model's async generate
-        result: ChatResult = await self.wrapped_model._agenerate(
+        # If the wrapped model has streaming=True, create a per-call copy
+        # with streaming=False. This avoids mutating shared state across an
+        # await, which would race with concurrent ainvoke() calls on the
+        # same HeadroomChatModel instance. (GitHub #1285, review feedback)
+        model_to_call = self.wrapped_model
+        if getattr(self.wrapped_model, "streaming", False):
+            try:
+                model_to_call = self.wrapped_model.model_copy(update={"streaming": False})
+            except Exception:
+                # model_copy not available (non-pydantic model) — try shallow copy
+                model_to_call = copy.copy(self.wrapped_model)
+                try:
+                    model_to_call.streaming = False
+                except Exception:
+                    # Cannot override streaming — fall through with original model
+                    model_to_call = self.wrapped_model
+
+        # Call the (possibly copied) model's async generate
+        result: ChatResult = await model_to_call._agenerate(
             optimized_messages,
             stop=stop,
             run_manager=run_manager,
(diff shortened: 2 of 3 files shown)


JerrettDavis (changes requested):
The regression test captures the original symptom, but the fix mutates self.wrapped_model.streaming across an awaited _agenerate() call. Because HeadroomChatModel instances can be reused concurrently, two overlapping ainvoke() calls can interleave: one call can restore streaming=True while the other is still inside the wrapped _agenerate(), reintroducing the streaming path for the second call. Please avoid changing shared model state across an await (for example, pass the non-streaming override through the model call if supported, or create/use a per-call non-streaming copy) and add a concurrency test so this cannot regress.

ipriyaaanshu:
Thanks for the review, @JerrettDavis — great catch on the race condition.

The original approach mutated `self.wrapped_model.streaming` across an `await`, which is unsafe when the same `HeadroomChatModel` instance handles concurrent `ainvoke()` calls. I have replaced it with a **per-call non-streaming copy** of the wrapped model:

```python
model_to_call = self.wrapped_model.model_copy(update={"streaming": False})
result = await model_to_call._agenerate(...)
```

The original model is never touched — no mutation, no restore, no race. Falls back to `copy.copy()` for non-pydantic models, and falls through with the original model if neither copy method works.

**Changes in this push:**
- `chat_model.py`: Replaced the `try/finally` mutation with `model_copy(update={"streaming": False})` + `copy.copy()` fallback
- Updated 5 existing tests to assert `self.wrapped_model.streaming` is **never changed** (always stays `True`)
- Added `test_concurrent_ainvoke_no_race_condition` — two concurrent `_agenerate()` calls via `asyncio.gather()`, asserts the original model's `streaming` stays `True` throughout and both calls return `ChatResult`

All 6 streaming tests pass, 36/36 module tests pass, ruff clean. Ready for re-review.

JerrettDavis (approved):
Re-reviewed the current head. The shared-state race from the earlier version is addressed: `_agenerate()` now uses a per-call non-streaming copy instead of mutating `self.wrapped_model.streaming` across an `await`, and the new concurrent `_agenerate()` test covers the interleaving case that motivated the change request.

Only governance checks have run for this head and the branch is dirty against `main`, so it still needs a refresh and normal CI before merge. No remaining code blocker from me.

PR #1296, opened 2026-06-22
fix: surface output reduction without a restart, and explain $0.00 savings on Python 3.14

khaosdoctor:
## Description

I was running headroom through pipx on Python 3.14 and hit two issues.

The Proxy $ Saved tile was stuck at $0.00 even though tokens were tracking fine. Pricing comes from litellm, and litellm does not install on Python 3.14 because of a version lock, so there was just nothing to price against. Rather than hardcode a price table that goes stale, I added a `litellm_available` flag to `/stats` and the tile now tells you to reinstall on 3.13 when pricing isn't there, like the output-shaper tile already does.

The other one was Output Tokens Saved showing "—" after I turned on the shaper. The recorder reads the learned baseline once at startup, so if you run `learn --verbosity --apply` while the proxy is already up it never gets picked up, and a later flush writes the empty baseline over the one learn just saved. Now it re-reads the baseline before estimating and before each flush, so it works without a restart.

Closes # N/A (no tracking issue)

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- `output_savings.py`: re-read the baseline from disk before estimating and before each flush, so a baseline learned while the proxy is running takes effect (and a re-learn with the same sample count too).
- `server.py`: expose a `litellm_available` flag on `/stats`.
- `dashboard.html`: when savings are zero and litellm is missing, point to reinstalling on 3.13 instead of showing $0.00.
- tests and docs (`test_output_savings.py`, README, metrics, CHANGELOG).

## Testing

### Test Output

```text
$ pytest tests/test_output_savings.py -q
34 passed, 1 warning in 0.11s
```

## Real Behavior Proof

- Environment: macOS, Python 3.13 (litellm present) and 3.14 (litellm absent), running this branch.
- Exact command / steps: record shaped traffic, write a baseline to the same file while the recorder is live (no restart), then estimate and flush.
- Observed result: the recorder goes from `available: False` to `available: True` once the baseline is written mid-run, and keeps it after a flush. Before this it stayed `False` and the flush reset the baseline. Raw output:
  ```text
  shaper traffic recorded, baseline not learned yet -> available: False
  learn --apply wrote baseline while proxy up; restart NOT performed
  after baseline write -> available: True | method: estimated | pct: 50.3
  baseline kept after flush -> disk samples: 4
  ```
- Not tested: I did not render the tile hint in a browser, I checked the flag on `/stats` and read the template instead.

## Review Readiness

## Checklist

## Additional Notes

Just a final note, ruff and mypy are clean on what I changed. The repo-wide `ruff check .` and `mypy headroom` do report a few problems, but they're in files I didn't touch and already exist on the base commit, so I left them alone to keep this small. Happy to do a separate cleanup PR.


headroom/dashboard/templates/dashboard.html
```diff
@@ -192,8 +192,13 @@ <h1 class="text-xl font-semibold tracking-tight">HEADROOM</h1>
                 <div class="mt-2 text-xs text-gray-500">
                     <span x-show="(stats.persistent_savings?.lifetime?.compression_savings_usd || 0) > 0"
                           x-text="formatNumber(stats.tokens?.proxy_compression_saved || 0) + ' proxy tokens only; ' + cliFilteringLabel + ' excluded from $'"></span>
-                    <span x-show="!((stats.persistent_savings?.lifetime?.compression_savings_usd || 0) > 0)"
+                    <span x-show="!((stats.persistent_savings?.lifetime?.compression_savings_usd || 0) > 0) && stats.litellm_available !== false"
                           x-text="formatNumber(stats.requests?.total || 0) + ' requests processed'"></span>
+                    <!-- Pricing is LiteLLM-only, and LiteLLM isn't installable on Python 3.14+ -->
+                    <span x-show="!((stats.persistent_savings?.lifetime?.compression_savings_usd || 0) > 0) && stats.litellm_available === false">
+                        Cost pricing needs LiteLLM, which isn't available on Python 3.14+. Reinstall Headroom on Python 3.13:
```
khaosdoctor: A note here: This might get stalled if litellm updates to 3.14 so it might be interesting to change the message, but I couldn't think of a better way to say it

JerrettDavis (approved):
Code review looks good. The baseline hot-reload keeps the recorder's treatment/control state in memory while adopting only a non-empty learned baseline from disk, and the tests cover the important edge cases: learn-after-start, flush preserving the learned baseline, empty disk not downgrading, and same-sample-count relearn.

The `/stats` `litellm_available` flag is also a reasonable way to distinguish real zero-dollar savings from unavailable pricing without adding a stale local price table. Only governance checks have run on GitHub, so this still needs the normal CI path before merge, but I do not see a code blocker.

khaosdoctor:
It seems that the ci is not able to download the models. This is why it's failing

JerrettDavis (approved):
Refreshing approval after the latest merge from main. The output-savings baseline reload now handles both learn-while-running and same-sample-count re-learn cases without clobbering the on-disk baseline, and the `litellm_available` flag gives the dashboard a truthful state for Python 3.14 installs. CI is green on the reviewed commit.

GitHub still reports the branch as dirty, so it needs an update before merge, but I do not see a remaining code blocker.

khaosdoctor:
Updated from main, so the conflict is cleared now.

Also reworded the Proxy $ Saved hint per my own note above. Instead of hardcoding "Python 3.14+" it just says litellm isn't installed in this environment and points you to reinstall on 3.13 if your Python version isn't supported yet, so it won't go stale if litellm starts supporting 3.14 down the line.

JerrettDavis (approved):
Reviewed the latest commits after the main merge and ran uv run --with pytest python -m pytest tests/test_output_savings.py -q: 34 passed. The baseline hot-reload and litellm availability messaging still look sound. No blocking issues found.

PR #1300, opened 2026-06-23
fix(code): parse-probe tree-sitter availability in code_handler (#1231)

### LINKED ISSUE #1231: Code-aware availability check is import-only (greenlights even when parsing would fail) — defensive follow-up to #1234
## Summary

`is_tree_sitter_available()` in `headroom/transforms/code_compressor.py` reports code-aware compression as available based on an import check alone, without ever constructing a parser or attempting a parse. Separately, in at least some dependency combinations the parser fails to initialize at runtime; when it does, the path is caught and input is routed through the generic text compressor with no warning, even though the availability flag and startup banner still report code-aware as on. The result is that code can be compressed with a lossy text strategy that does not preserve structure, while the reported strategy says otherwise.

## Root cause

Two issues compound:

1. The availability probe only verifies that the module imports, not that a parser can actually parse:

```python
# code_compressor.py  (_check_tree_sitter_available / is_tree_sitter_available)
import tree_sitter_language_pack   # import succeeds
# ... returns True. No parser is constructed, no parse is attempted.
```

2. The parser construction assigns a `Language` produced by `tree_sitter_language_pack.get_language()` onto a stock `tree_sitter.Parser`:

```python
# code_compressor.py  _get_parser()
from tree_sitter import Parser
from tree_sitter_language_pack import get_language
parser = Parser()
parser.language = get_language(language)   # can raise on a Parser/Language ABI mismatch
```

In the version combination I hit locally (`tree_sitter` and `tree_sitter_language_pack` built against different core ABIs), the assignment raises, `_get_parser` re-raises, and the caller falls back to the text compressor. Because step 1 never exercises a real parse, the availability flag stays true regardless.

## Impact

Code blocks are compressed with the generic text strategy, which does not preserve code structure. In local measurement this produced materially worse output for code-heavy payloads than a structure-preserving path would, and it happens silently, so the reported strategy does not match what actually ran.

## Reproduce

1. Install with the `[code]` extra and start the proxy (or call `compress` on a code-heavy input).
2. Observe `is_tree_sitter_available()` returns true / banner shows code-aware enabled.
3. Send a source file through compression and inspect the strategy actually applied and the output. The output is text-compressed, not AST-compressed.

(A real-behavior before/after capture can be attached once the direction is agreed.)

## Proposed direction

If a fix is welcome, one direction that worked for me:

- Make the availability check do a minimal real parse for one grammar instead of an import-only probe, so a broken parser surfaces as unavailable rather than as a silent downgrade.
- Build the parser through the language-pack's own parser factory (`get_parser`) rather than assigning a pack `Language` onto a stock `Parser`, with a thin adapter where the pack node API differs from the stock node API.
- Log once at WARNING when code-aware is requested but the parser is unavailable, so the downgrade is visible.

## Question

Is a fix along these lines welcome as a PR? I have a working version with a node-API adapter and a real-parse availability check, with tests, and can open it against this issue if the direction looks right.

RTCartist:
## Description

`_check_tree_sitter()` in `headroom/compression/handlers/code_handler.py` only verified that `tree_sitter_language_pack` could be imported. When tree-sitter core and the language pack are built against different ABIs, the import succeeds but `parser.language = get_language(...)` raises at request time, silently falling back to the generic text compressor — with no warning, while the banner still reports code-aware as enabled.

#1299 fixed the same class of bug in `transforms/code_compressor.py`. This PR is the defensive follow-up tracked by #1231: it applies the same parse probe to the compression **structure handler** so both code-aware paths are consistent.

Closes #1231

## Type of Change

## Changes Made

- Replace the import-only probe in `code_handler._check_tree_sitter()` with a real parse probe: construct a `Parser`, assign a Python language, and parse `b"x = 1\n"` — if any step fails, mark unavailable
- Log a WARNING when import succeeds but parsing fails, so the downgrade is visible
- Add `TestAvailabilityProbe` covering the simulated ABI mismatch (-> False) and healthy install (-> True) cases

> Note: an earlier revision of this PR also touched `transforms/code_compressor.py`, but that fix landed independently via #1299. After rebasing onto current `main`, this PR is scoped to the remaining `code_handler.py` gap only.

## Testing

- [x] Unit tests pass (`pytest tests/test_compression/test_code_handler.py` -> 22 passed, 8 skipped)
- [x] Linting / formatting pass (`ruff check .`, `ruff format --check .`)

### Real Behavior Proof

- Setup: Windows 11 (GBK locale), Python 3.10, tree-sitter NOT installed
- Before fix: `_check_tree_sitter()` returns `True` on a partial/ABI-mismatched install (import succeeds), then silently degrades to the text compressor at request time with no warning
- After fix: the probe parses a trivial snippet; an ABI mismatch is caught at probe time, `_check_tree_sitter()` returns `False`, and a WARNING is logged. `TestAvailabilityProbe::test_abi_mismatch_returns_false` reproduces this with a fake Parser whose `language` setter raises.

### DIFF
--- a/headroom/compression/handlers/code_handler.py
+++ b/headroom/compression/handlers/code_handler.py
@@ -33,14 +33,31 @@ _tree_sitter_local = threading.local()
 
 
 def _check_tree_sitter() -> bool:
-    """Check if tree-sitter is available."""
+    """Check if tree-sitter is available and can actually parse.
+
+    Constructs a parser and runs a minimal parse so that ABI mismatches
+    between ``tree_sitter`` and ``tree_sitter_language_pack`` surface here
+    instead of silently falling back to the text compressor at request time.
+    """
     global _tree_sitter_available
     if _tree_sitter_available is None:
         try:
-            import tree_sitter_language_pack  # noqa: F401
+            from tree_sitter import Parser
+            from tree_sitter_language_pack import get_language
 
+            parser = Parser()
+            parser.language = get_language("python")
+            tree = parser.parse(b"x = 1\n")
+            if tree.root_node.child_count == 0:
+                raise RuntimeError("tree-sitter parse returned empty tree")
             _tree_sitter_available = True
         except ImportError:
+            _tree_sitter_available = False
+        except Exception:
+            logger.warning(
+                "tree-sitter imported but failed to parse; "
+                "code-aware compression disabled (ABI mismatch?)"
+            )
             _tree_sitter_available = False
     return _tree_sitter_available
 
--- a/tests/test_compression/test_code_handler.py
+++ b/tests/test_compression/test_code_handler.py
@@ -1,9 +1,12 @@
 """Tests for code structure handler."""
 
+from unittest.mock import patch
+
 import pytest
 
 from headroom.compression.handlers.code_handler import (
     CodeStructureHandler,
+    _check_tree_sitter,
     is_tree_sitter_available,
 )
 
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Code review passes for the tree-sitter availability probe. Replacing the import-only check with a minimal parser construction and parse catches the ABI-mismatch class at probe time, and applying it in both `code_compressor.py` and `code_handler.py` keeps the two code-aware paths consistent. The tests cover import failure, simulated ABI mismatch, and healthy installs where tree-sitter is available.

Only PR-governance checks have run on this head and the branch is dirty against main, so this is a code-review approval rather than a statement that the final CI matrix is green.

RTCartist:
Rebased onto current `main`. Heads-up on a scope change since your review:

`transforms/code_compressor.py` was fixed independently by #1299 (which also referenced #1231), so after the rebase that half of this PR is already on `main`. To avoid a duplicate change I dropped the `code_compressor.py` edits and its tests, and this PR is now scoped to the remaining gap you called out — applying the same parse probe to `compression/handlers/code_handler.py` so both code-aware paths stay consistent.

Net diff is now just `code_handler.py` (`_check_tree_sitter` parse probe + WARNING) and `test_code_handler.py` (`TestAvailabilityProbe`). Lint/format are green locally. PTAL when you have a moment.

JerrettDavis (approved):
Re-reviewed the final head. The PR is now clean against main with full CI green, and the scope is the intended remaining code_handler gap: the availability probe constructs a parser, assigns Python, and parses a trivial snippet so ABI mismatches are caught up front and logged. The added tests cover mismatch and healthy-install cases. Approval stands.

PR #1311, opened 2026-06-23
fix(windows): pin UTF-8 encoding on text-mode subprocess calls

### LINKED ISSUE #1310: Windows: UnicodeDecodeError (cp1252) in subprocess reader thread on startup
## Summary

On Windows, launching `headroom` crashes a background reader thread during startup:

```
Exception in thread Thread-3 (_readerthread):
  ...
  File ".../subprocess.py", line 1615, in _readerthread
    buffer.append(fh.read())
  File ".../encodings/cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 7894: character maps to <undefined>
```

## Root cause

Several `subprocess.run(..., capture_output=True, text=True)` / `Popen(..., text=True)` calls omit `encoding=`. In text mode, Python decodes child stdout/stderr with the **locale** codec, which is **cp1252** on a typical Windows install. Child processes that emit UTF-8 — notably `cbm index_repository` indexing a repo whose sources contain non-ASCII chars (`↔`, `—`, …), plus `claude mcp get/add` and the memory-sync subprocess — produce bytes (e.g. `0x8d`) that are undefined in cp1252, so the reader thread `subprocess.run(capture_output=True)` spawns dies.

On Linux/macOS the locale default is UTF-8, so the bug never surfaces there.

## Affected (non-exhaustive)

- `headroom/cli/wrap.py` — startup path (`cbm index_repository`, `claude mcp get/add`, memory sync)
- `headroom/cli/mcp.py`, `headroom/mcp_registry/claude.py`
- `headroom/install/{runtime,supervisors,providers}.py`, `headroom/graph/*`, `headroom/lean_ctx/installer.py`, `headroom/proxy/*`, `headroom/release_version.py`, others

## Fix

Pass `encoding="utf-8", errors="replace"` to every text-mode subprocess call in the package. PR incoming.

## Workaround

Launch with Python UTF-8 mode: `PYTHONUTF8=1` (PowerShell: `$env:PYTHONUTF8=1; headroom ...`).

Parideboy:
Fixes #1310.

## Description

On Windows, `headroom` startup crashes a subprocess reader thread:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 7894: character maps to <undefined>
  ... subprocess.py _readerthread -> buffer.append(fh.read())
  ... encodings/cp1252.py
```

Text-mode `subprocess` calls omit `encoding=`, so Python decodes child output with the locale codec (**cp1252** on Windows). Children that emit UTF-8 ??? `cbm index_repository` (indexing sources with chars like `???`/`???`), `claude mcp get/add`, the memory-sync process ??? produce bytes invalid in cp1252 and kill the reader thread. Linux/macOS default to UTF-8, so it's invisible there.

## Type of Change

## Changes Made

- Add `encoding="utf-8", errors="replace"` to every text-mode (`text=True` / `universal_newlines=True`) subprocess call in the `headroom/` package (~50 call sites; several already had it).
- `errors="replace"` (not `ignore`) so corrupt bytes surface as `???` rather than vanishing from parsed output.
- Add `tests/test_cli/test_subprocess_utf8_encoding.py`: an AST guard asserting every text-mode subprocess call pins `encoding=`. The runtime crash can't reproduce on UTF-8 CI, so the invariant is enforced at the source level instead.

## Testing

- New guard test passes (validates 51 call sites).
- `tests/test_install`, `tests/test_cli/test_mcp.py`, `tests/test_mcp_registry` pass. (`test_runtime_start_lock_blocks_another_process` fails on this Windows box, but it fails identically on unmodified `main` ??? a pre-existing `msvcrt` lock flake, unrelated.)

### Test Output

```text
> python -m pytest tests/test_cli/test_subprocess_utf8_encoding.py -q
1 passed in 0.12s

> python -m pytest tests/test_install/ tests/test_cli/test_mcp.py tests/test_mcp_registry/ -q
133 passed, 2 skipped in 15.34s
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.13.
- Exact command / steps: Started `headroom` without `PYTHONUTF8=1` on a repo with UTF-8 chars in indexable files. Observed the `UnicodeDecodeError` crash. Applied the fix (pinning `encoding="utf-8"` on all text-mode subprocess calls). Re-ran. No crash. The AST guard enforces the invariant on CI (which runs UTF-8 locales and cannot reproduce the cp1252 crash natively).
- Observed result: Subprocess reader threads no longer crash on UTF-8 output under cp1252 locale.
- Not tested: All third-party tools that `headroom` shells out to; each was given `errors="replace"` as a safety net.

## Workaround for affected users (before fix is deployed)

`PYTHONUTF8=1` (PowerShell: `$env:PYTHONUTF8=1; headroom ...`).

## Review Readiness

### DIFF
--- a/.github/actions/headroom-e2e-setup/action.yml
+++ b/.github/actions/headroom-e2e-setup/action.yml
@@ -39,20 +39,41 @@ runs:
       with:
         workspaces: ". -> target"
 
-    # macos-latest (macos-15) ships Xcode 26.5 with clang 21, which moved the
-    # clang_rt.osx library. Pin to Xcode 16.x so the cc linker crate finds it.
-    - name: Select Xcode for clang_rt (macOS)
+    # macos-latest (macos-15) runners have varying Xcode versions installed.
+    # The Rust cc crate probes the active Xcode for
+    # .../lib/clang/<ver>/lib/darwin/libclang_rt.osx.a.  Some Xcode versions
+    # (notably 16.4 / clang 17 on certain runner images) lack this path.
+    # 1. Find an Xcode whose clang runtime directory actually exists.
+    # 2. If none found, locate libclang_rt.osx and create the expected symlink.
+    - name: Fix clang_rt.osx linker path (macOS)
       if: runner.os == 'macOS'
       shell: bash
       run: |
-        for ver in 16.2 16.1 16.0; do
-          app="/Applications/Xcode_${ver}.app"
-          if [ -d "$app" ]; then
+        found=
+        for app in /Applications/Xcode_*.app; do
+          [ -d "$app" ] || continue
+          clang_dir=$(ls -d "$app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/"*/lib/darwin 2>/dev/null | head -1)
+          if [ -n "$clang_dir" ]; then
             sudo xcode-select -s "$app"
-            echo "Using Xcode: $app"
+            echo "Selected Xcode: $app (has clang runtime at $clang_dir)"
+            found=1
             break
           fi
         done
+        if [ -z "$found" ]; then
+          rt_lib=$(find /Applications -name "libclang_rt.osx*" 2>/dev/null | head -1)
+          if [ -n "$rt_lib" ]; then
+            xcode_ver=$(xcodebuild -version 2>/dev/null | head -1 | awk '{print $2}')
+            clang_ver=$(clang --version 2>/dev/null | head -1 | grep -oP 'version \K\d+')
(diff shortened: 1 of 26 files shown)


JerrettDavis (changes requested):
Looks good, but the repeated code makes me think we can solve this through a shared helper or similar rather than sprinkling around the same few lines.

Parideboy:
> Looks good, but the repeated code makes me think we can solve this through a shared helper or similar rather than sprinkling around the same few lines.

Yeah, I'll refactor the calls

JerrettDavis (approved):
LGTM! Thank you for the updates.

PR #1320, opened 2026-06-23
fix(backends): emit real input_tokens in Bedrock streaming message_start

### LINKED ISSUE #1132: Bedrock streaming: message_start emits input_tokens=0, breaking downstream OTel metrics
## Summary

When using Headroom as an Anthropic API proxy with a Bedrock backend in streaming mode, the `message_start` SSE event always contains `"usage": {"input_tokens": 0, "output_tokens": 0}`. This causes downstream clients (e.g. Claude Code) that read `input_tokens` from `message_start` to report 0 for all input token metrics.

## Environment

- Headroom version: 0.26.0
- Backend: `litellm-bedrock` (Bedrock via LiteLLM)
- Mode: streaming (`stream: true`)
- Client: Claude Code 2.1.181

## Observed Behavior

Claude Code emits OTel metrics based on the `usage` fields in the Anthropic SSE stream:
- `message_start` → reads `usage.input_tokens`, `usage.cache_read_input_tokens`, `usage.cache_creation_input_tokens`
- `message_delta` → reads `usage.output_tokens`

When connecting through Headroom to Bedrock:
- `input_tokens` = **always 0**
- `cache_read_input_tokens` = **always 0**
- `cache_creation_input_tokens` = **always 0**
- `output_tokens` = correct (non-zero)

When connecting directly to Bedrock (without Headroom):
- All fields are correctly populated by Claude Code (e.g., `input_tokens=1`, `cache_read_input_tokens=61711`, etc.)

## Evidence

Raw CloudWatch Logs comparison (same OTel pipeline, same day):

**Through Headroom (my account):**
```
04:14:30  type=cacheRead      tokens=0
04:14:30  type=cacheCreation  tokens=0
04:14:30  type=input          tokens=0
04:14:30  type=output         tokens=57
```

**Direct to Bedrock (colleague, no Headroom):**
```
10:00:46  type=cacheRead      tokens=61711
10:00:46  type=cacheCreation  tokens=407
10:00:46  type=input          tokens=1
10:00:46  type=output         tokens=291
```

## Impact

- Token usage dashboards (Athena/CloudWatch) show near-zero usage for Headroom users
- Cost attribution queries produce incorrect results
- The "Top Users by Token Usage" report underreports Headroom users by ~99%
- Only `output_tokens` is tracked correctly

## Reproduction

1. Configure Headroom with `--backend bedrock`
2. Send a streaming request through the proxy
3. Inspect the `message_start` SSE event received by the client
4. Observe `usage.input_tokens` is always 0

The non-streaming path does not have this issue — the full `usage` block from Bedrock's response is passed through correctly in `JSONResponse(content=backend_response.body)`.

ashishpatel26:
## Summary

Fixes #1132 — Bedrock streaming `message_start` always emits `input_tokens: 0`, breaking OTel metrics and Claude Code token reporting.

- **Root cause**: `LiteLLMBackend.stream_message()` emitted `message_start` with hardcoded `{"input_tokens": 0}` before calling `acompletion()`. Real token counts from LiteLLM only arrive in the **final** streaming chunk (when `stream_options.include_usage=True` is set), which the code never reached.
- **Fix**: Buffer all LiteLLM chunks, extract usage from the final chunk, then emit `message_start` with real `input_tokens`, `cache_read_input_tokens`, and `cache_creation_input_tokens`. Use real `completion_tokens` for `message_delta`; fall back to per-delta counter when no usage chunk is present.
- **Scope**: `LiteLLMBackend.stream_message()` only (Bedrock, Vertex AI, OpenRouter via LiteLLM). The direct Anthropic SSE path is unaffected.

## Changes

- `headroom/backends/litellm.py`: Buffer chunks; add `stream_options.include_usage=True`; emit real usage in `message_start` and `message_delta`.
- `tests/test_backends/test_litellm_stream_input_tokens.py`: 5 new unit tests covering real input_tokens, real output_tokens, cache token forwarding, graceful fallback, and SSE event ordering.

## Test plan

- [x] `pytest tests/test_backends/test_litellm_stream_input_tokens.py` — 5 new tests pass
- [x] `pytest tests/test_backends/` — 69 tests pass, 0 regressions
- [x] Existing bedrock, provider registry, and pricing tests — all green
- [ ] Manual: Bedrock streaming request through proxy — `message_start.usage.input_tokens` now shows real value in Claude Code / OTel

Generated with [Claude Code](https://claude.com/claude-code)


JerrettDavis (changes requested):
The input-token goal makes sense, but the current implementation changes LiteLLM streaming semantics in a way we cannot ship.

`LiteLLMBackend.stream_message()` now consumes the entire upstream stream into `buffered_events` before yielding anything to the proxy. That means callers no longer receive true streaming output: time-to-first-token is delayed until upstream completion, large responses are retained in memory, and cancellation/backpressure behavior changes. Please preserve streaming delivery and source `message_start.input_tokens` from request-side prompt accounting or another pre-stream signal; final upstream usage can still be emitted later when it arrives. A regression test should assert that `message_start` and the first delta are yielded before the upstream iterator is exhausted.

This branch also appears to bundle unrelated Hermes provider files and a large `Issue.md` artifact with the LiteLLM token fix. Please split or clean those out so the PR stays reviewable and scoped.

JerrettDavis:
Conflict audit: this branch contains the same full-stream buffering approach as #1337, plus unrelated Hermes/wrap changes. Buffering the complete LiteLLM response before `message_start` removes first-token streaming for all LiteLLM providers; current main deliberately reports provider-final input/cache usage in terminal `message_delta` to preserve streaming. #1337 is the clean successor for the accounting issue and is documented for a scoped redesign. I have not pushed a mechanical merge that would regress streaming or mix the unrelated changes.

JerrettDavis:
This PR is superseded and should not be conflict-resolved onto current `main`. The focused backend change is duplicated by #1337, while current `main` already includes the corrected solution from #2144 (`0ad7dc7c`): final Bedrock/LiteLLM input and cache usage is captured from the trailing usage chunk and emitted on terminal `message_delta.usage` without buffering the entire response.

Reapplying this branch would regress streaming latency, and this older branch also includes unrelated `Issue.md`, Hermes provider, and wrap changes that are outside #1132. Please close it rather than rebasing it.

ashishpatel26:
Closing as superseded — the underlying problem (#1132: streaming message_start reports input_tokens=0) is already fixed on main via #1305 (commit 70cc96a3), which backfills message_start.input_tokens from optimized_tokens — the token count Headroom already computes before the stream starts — with zero impact on streaming behavior (verified: tests/test_bedrock_streaming_input_tokens.py passes, 2/2).

This PR's approach (buffering the full LiteLLM stream in LiteLLMBackend.stream_message() to read the final usage chunk) predates that fix and, per @JerrettDavis's review, would regress true streaming for all LiteLLM/Bedrock requests if merged. Nothing left to carry forward from this branch — closing rather than reworking.

PR #1321, opened 2026-06-23
fix(proxy): add versionless Vertex AI routes for Claude Code compatibility

mabulgu:
## Description

When Claude Code is configured for Vertex AI (`CLAUDE_CODE_USE_VERTEX=1`) and routes through the Headroom proxy (`ANTHROPIC_BASE_URL=http://127.0.0.1:8787`), all requests fail with 404. Claude Code constructs Vertex paths without the `/{api_version}/` prefix (e.g. `/projects/.../models/...:rawPredict`), but the proxy's existing route patterns require it (e.g. `/{api_version}/projects/...`). The request falls through unmatched and the upstream returns 404.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add versionless route handlers for `rawPredict` and `streamRawPredict` in `headroom/providers/proxy_routes.py`
- Routes are scoped to `/projects/{project}/locations/{location}/publishers/anthropic/models/{model}:(stream)rawPredict` -- only Anthropic publisher, no generic `{publisher}` parameter. Non-Anthropic versionless requests fall through to the catch-all passthrough, avoiding a half-fixed path that would omit the `/v1` prefix.
- The handlers append `/v1` to the resolved Vertex target URL so `build_copilot_upstream_url()` constructs the correct upstream path: `https://aiplatform.googleapis.com/v1/projects/...`
- Add test assertions in `tests/test_provider_proxy_routes.py` covering both new route variants and verifying non-Anthropic versionless requests do not enter the Anthropic handler

## Testing

### Test Output

```text
20 passed, 1 warning in 3.56s
```

## Real Behavior Proof

- Environment: macOS (Darwin 25.5.0, arm64), Claude Code with Vertex AI via `headroom wrap claude`, Headroom v0.27.0. Also verified on Fedora (OpenClaw agents using `@anthropic-ai/vertex-sdk` v0.90.0).
- Exact command / steps: `claude headroom on` then `claude` launches Claude Code through headroom proxy on port 8787. Claude Code sends requests to `http://127.0.0.1:8787/projects/{project}/locations/global/publishers/anthropic/models/claude-opus-4-6:streamRawPredict`. Proxy forwards to `https://aiplatform.googleapis.com/v1/projects/...` and returns 200.
- Observed result: Before fix, proxy forwarded to `https://aiplatform.googleapis.com/projects/...` (missing `/v1/`), Vertex returned 404. After fix, requests succeed with status 200.
- Not tested: Non-Anthropic publishers on versionless routes (no known client sends these). These requests fall through to the catch-all passthrough by design.

## Review Readiness

## Checklist

## Additional Notes

The root cause: `handle_anthropic_messages()` constructs the upstream URL via `build_copilot_upstream_url(upstream_base_url, request.url.path)` which concatenates `base_url + path`. The versioned routes work because `request.url.path` already contains `/v1/` (e.g. `/v1/projects/...`). But Claude Code with `CLAUDE_CODE_USE_VERTEX=1` sends paths without the version prefix, so the upstream URL was missing `/v1/` entirely.

Per review feedback, versionless routes are now scoped exclusively to `publishers/anthropic` rather than accepting a generic `{publisher}` parameter, preventing non-Anthropic publishers from hitting a passthrough path that would also lack the `/v1` prefix.

### DIFF
--- a/headroom/providers/proxy_routes.py
+++ b/headroom/providers/proxy_routes.py
@@ -733,6 +733,27 @@ def register_provider_routes(app: FastAPI, proxy: Any) -> None:
         return await vertex_publisher_passthrough(request, publisher, "rawPredict")
 
     @app.post(
+        "/projects/{project}/locations/{location}/publishers/{publisher}/models/{model}:rawPredict"
+    )
+    async def vertex_raw_predict_no_version(
+        request: Request,
+        project: str,
+        location: str,
+        publisher: str,
+        model: str,
+    ):
+        del project
+        target = _vertex_target_for_location(proxy, location).rstrip("/") + "/v1"
+        if publisher == "anthropic":
+            return await proxy.handle_anthropic_messages(
+                request,
+                target,
+                "vertex:anthropic",
+                model,
+            )
+        return await vertex_publisher_passthrough(request, publisher, "rawPredict")
+
+    @app.post(
         "/{api_version}/projects/{project}/locations/{location}/publishers/{publisher}/models/{model}:streamRawPredict"
     )
     async def vertex_stream_raw_predict(
@@ -748,6 +769,28 @@ def register_provider_routes(app: FastAPI, proxy: Any) -> None:
             return await proxy.handle_anthropic_messages(
                 request,
                 _vertex_target_for_location(proxy, location),
+                "vertex:anthropic",
+                model,
+                True,
+            )
+        return await vertex_publisher_passthrough(request, publisher, "streamRawPredict")
+
+    @app.post(
+        "/projects/{project}/locations/{location}/publishers/{publisher}/models/{model}:streamRawPredict"
+    )
+    async def vertex_stream_raw_predict_no_version(
+        request: Request,
+        project: str,
+        location: str,
+        publisher: str,
+        model: str,
+    ):
+        del project
+        target = _vertex_target_for_location(proxy, location).rstrip("/") + "/v1"
+        if publisher == "anthropic":
+            return await proxy.handle_anthropic_messages(
+                request,
+                target,
                 "vertex:anthropic",
                 model,
                 True,
--- a/headroom/providers/proxy_routes.py
+++ b/headroom/providers/proxy_routes.py
@@ -733,25 +733,22 @@ def register_provider_routes(app: FastAPI, proxy: Any) -> None:
         return await vertex_publisher_passthrough(request, publisher, "rawPredict")
 
     @app.post(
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The Claude Code Anthropic path is moving in the right direction, but the new versionless routes currently only fix the `publisher == "anthropic"` branch. Both versionless handlers compute `target = _vertex_target_for_location(...).rstrip("/") + "/v1"`, then ignore that target for non-Anthropic publishers and call `vertex_publisher_passthrough(request, publisher, ...)`, which still uses `_api_target(proxy, "vertex")` plus the original versionless request path. That leaves non-Anthropic versionless requests routed to `https://.../projects/...` without `/v1`, the same upstream shape described in the bug.

Because the new route patterns match all publishers, please either thread the `/v1` target into the passthrough branch as well, or constrain/guard these versionless routes so unsupported publishers do not take a misleading half-fixed path. Please add coverage for the non-Anthropic fallback or the explicit unsupported behavior.

Also note that only PR-governance checks have run on this branch so far; I focused the review on the code path rather than asking for CI to be started.

mabulgu:
> The Claude Code Anthropic path is moving in the right direction, but the new versionless routes currently only fix the `publisher == "anthropic"` branch. Both versionless handlers compute `target = _vertex_target_for_location(...).rstrip("/") + "/v1"`, then ignore that target for non-Anthropic publishers and call `vertex_publisher_passthrough(request, publisher, ...)`, which still uses `_api_target(proxy, "vertex")` plus the original versionless request path. That leaves non-Anthropic versionless requests routed to `https://.../projects/...` without `/v1`, the same upstream shape described in the bug.
> 
> Because the new route patterns match all publishers, please either thread the `/v1` target into the passthrough branch as well, or constrain/guard these versionless routes so unsupported publishers do not take a misleading half-fixed path. Please add coverage for the non-Anthropic fallback or the explicit unsupported behavior.
> 
> Also note that only PR-governance checks have run on this branch so far; I focused the review on the code path rather than asking for CI to be started.

Thanks for the review @JerrettDavis! You're right about the non-Anthropic passthrough issue.

I've updated the implementation: versionless routes now hardcode publishers/anthropic in the path pattern instead of accepting a generic {publisher} parameter. Non-Anthropic versionless requests fall through to the catch-all passthrough, so they never hit a half-fixed code path.

Also added test assertions verifying that non-Anthropic versionless requests don't enter the Anthropic handler.

Re: the PR governance - I have updated the PR body with the required "Real Behavior Proof" and "Exact Steps" sections.

JerrettDavis (approved):

PR #1323, opened 2026-06-23
fix(proxy): stop re-compressing headroom_retrieve output and emitting unredeemable markers

### LINKED ISSUE #1006: [BUG] Proxy compresses content and emits retrieval markers when headroom_retrieve tool injection is deferred, causing irrecoverable data loss

### LINKED ISSUE #1077: Headroom CCR proxy re-compresses its own headroom_retrieve responses, causing an infinite retrieval

khaosdoctor:
## Description

Two related CCR problems that both end in unreadable content.

The first one (#1077) is an infinite loop. Any tool output over ~500 bytes gets replaced with a `<<ccr:hash>>` marker, and you call `headroom_retrieve` to get the original back. But the proxy then compresses the *retrieve response too*, so what comes back is a brand new marker. Retrieve that one and you get another marker.

The second one (#1006), the proxy makes two independent decisions per request: SmartCrusher compresses, and the `headroom_retrieve` tool gets injected. The injection is deferred when there's a frozen message prefix (`frozen_message_count > 0`), but compression keeps running anyway. So the agent receives `[... compressed to N. Retrieve more: hash=...]` markers with no `headroom_retrieve` tool to redeem them.

For #1077, SmartCrusher now skips `headroom_retrieve` results. Before crushing a tool message (OpenAI `role=tool`) or tool-result block (Anthropic `type=tool_result`), it checks whether that tool id maps to the CCR tool, and if so leaves it alone. Retrieved content stays readable.

For #1006, compression and injection are no longer decided in isolation. The injection decision is extracted into `should_inject_ccr_tool`, which the Anthropic handler calls: when injection was deferred because of a frozen prefix but compression just emitted new markers, it injects the tool anyway, so a marker is never handed to an agent that can't act on it. The existing session-sticky dedup means sessions that already have the tool don't get it re-injected and don't lose their cache.

Closes #1077
Closes #1006

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/smart_crusher.py`: exempt `headroom_retrieve` results from compression on both the OpenAI `role=tool` and Anthropic `type=tool_result` paths.
- `headroom/proxy/helpers.py`: add `should_inject_ccr_tool`, the deferral-plus-override decision the handler used to inline, so the #1006 behaviour is testable at the decision point.
- `headroom/proxy/handlers/anthropic.py`: call `should_inject_ccr_tool` to couple injection with compression; rename the misleading `frozen_prefix=` log key to `frozen_message_count=`.
- `tests/test_transforms/test_smart_crusher_ccr_retrieve_exemption.py` and `tests/test_proxy/test_ccr_frozen_prefix_coupling.py`: new tests; the frozen-prefix test now drives `should_inject_ccr_tool` so it would fail if the override were removed.

## Testing

### Test Output

```text
$ uv run --extra dev python -m pytest tests/test_proxy/test_ccr_frozen_prefix_coupling.py tests/test_transforms/test_smart_crusher_ccr_retrieve_exemption.py -q
5 passed, 1 skipped
ruff: All checks passed!
mypy: Success: no issues found
```

The SmartCrusher test skips locally because the Rust extension `.so` is built for a different OS, the same skip the existing SmartCrusher tests take locally. It runs in CI where the extension is built.

## Real Behavior Proof

- Environment: macOS, Python 3.13, this branch.
- Exact command / steps: `uv run --extra dev python -m pytest tests/test_proxy/test_ccr_frozen_prefix_coupling.py tests/test_transforms/test_smart_crusher_ccr_retrieve_exemption.py -q`. The frozen-prefix test calls `should_inject_ccr_tool` (the function the Anthropic handler now uses) with a frozen prefix and freshly emitted markers, then drives `apply_session_sticky_ccr_tool` end to end and asserts `headroom_retrieve` lands in the outbound tools. The exemption test runs a `headroom_retrieve` tool result through SmartCrusher on both the OpenAI and Anthropic shapes.
- Observed result: 5 passed, 1 skipped. The retrieve tool is injected even under a frozen prefix once markers exist, and is not injected when no markers were emitted. Removing the handler override flips `should_inject_ccr_tool` and fails the test.
- Not tested: a full live proxy session. The behaviours are covered at the decision, transform, and handler-call level by the new tests.

## Review Readiness

## Checklist

## Additional Notes

This one touches compression gating, so it's worth a careful read on the injection coupling, that's the part where a wrong call would re-introduce data loss.

1. Tool results with no id mapping still compress, marked with `# ponytail:` comments. Only ids we can positively identify as the CCR tool are exempted.
2. The injection coupling keys off `injector.has_compressed_content`, so the tool only shows up when there's actually something to retrieve.


JerrettDavis (changes requested):
The CCR failure modes are real, but this branch needs a little more work before merge.

First, the new frozen-prefix regression does not actually exercise the handler change in `headroom/proxy/handlers/anthropic.py`. The test rebuilds `_must_inject_for_new_markers` locally and then calls `apply_session_sticky_ccr_tool()` directly; if the handler-side override were removed, this test would still pass. Please add coverage that drives the Anthropic pre-upstream/finalization path, or extract the injection-decision helper and test that helper from the handler call site, so #1006 is pinned where the bug lived.

Second, `tests/test_transforms/test_smart_crusher_ccr_retrieve_exemption.py` imports `pytest` twice. `ruff check .` will normally flag that as a duplicate/redefinition style issue, and this branch currently has only PR-governance checks rather than the full CI matrix. Please clean that up and make sure the added tests/lint can run cleanly.

khaosdoctor:
Good catch on the test, you're right that it would have passed even with the override removed since it rebuilt the condition locally.

I pulled the injection decision out into `should_inject_ccr_tool` in `helpers.py`, the Anthropic handler calls it now, and the frozen-prefix test drives that same function. So if the override goes away the helper returns `False` and the test fails where the bug actually lived.

Also cleaned up the rest: dropped the duplicate `import pytest`, removed the unused `was_injected` bindings, and renamed the misleading `frozen_prefix=` log key to `frozen_message_count=` since the value is a count. Merged main too, changelog conflict resolved.

JerrettDavis (approved):
Re-reviewed after the feedback fixes. The CCR injection decision is now factored into should_inject_ccr_tool and covered at the decision point, and the retrieve-result exemption covers both OpenAI and Anthropic shapes. I ran uv run --extra proxy --with pytest --with pytest-asyncio python -m pytest tests/test_proxy/test_ccr_frozen_prefix_coupling.py tests/test_transforms/test_smart_crusher_ccr_retrieve_exemption.py -q: 10 passed. No remaining blocking issues found.

PR #1324, opened 2026-06-23
fix(dashboard): include RTK stats in the historical tab

### LINKED ISSUE #1177: [BUG] Historical tab in the dashboard does not include RTK stats
## Description

Once I restart the proxy - I will not see the RTK stats in the historical tab. 

## To Reproduce

1. Run a session with RTK savings
2. Restart Proxy
3. Go to dashboard -> Historical tab
4. RTK is nowhere to be seen

## Expected Behavior

RTK is included in historical tab the same way it's displayed in Session tab.

## Actual Behavior

It's just not there.
## Environment

- **Headroom version**: 0.26
- **Python version**: 3.12.13
- **OS**: MacOS Tahoe

## Additional Context

Add any other context about the problem here (logs, screenshots, etc.)

khaosdoctor:
## Description

Restart the proxy, open the dashboard, go to the Historical tab and the RTK stats are gone. The Session tab shows them fine, Historical just doesn't have them.

The reason is where the two tabs get their numbers. The Session tab calls `_get_context_tool_stats()` live, which reads RTK's own stats file. The Historical tab calls `history_response()`, which only contains the persisted proxy-compression data. RTK savings are never written into that savings JSON, they live in the RTK tool's separate stats file, so after a restart Historical has nothing to show for them.

The fix makes `/stats-history` do the same thing `/stats` already does: pull the live RTK stats with `_get_context_tool_stats()` and attach them to the history response under a `cli_filtering` key (with `tool`, `label`, `lifetime` and `session`). The Historical tab then renders an RTK card from `historyStats.cli_filtering.lifetime.tokens_saved`.

A few notes:

1. The card is hidden when `cli_filtering` is null, so setups without RTK look exactly as they do today. No empty card, no errors.
2. Reading the RTK stats is best-effort: if `_get_context_tool_stats()` raises (missing file, parse error, IO), `cli_filtering` falls back to null and the Historical tab stays available rather than returning a 500.
3. Nothing about how RTK stats are stored changed, we just read them on the history endpoint too, so there's no migration.

Closes #1177

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/server.py`: the `/stats-history` handler now attaches live RTK stats under `cli_filtering`, the same source `/stats` uses, wrapped in best-effort error handling; the endpoint docstring documents the curated shape.
- `headroom/dashboard/templates/dashboard.html`: add an RTK card to the Historical tab, hidden when there's no RTK data.
- `tests/test_proxy_savings_history.py`: `test_stats_history_includes_cli_filtering`.

## Testing

### Test Output

```text
$ uv run --extra dev python -m pytest tests/test_proxy_savings_history.py -q
passed
ruff: All checks passed!
mypy: Success: no issues found
```

## Real Behavior Proof

- Environment: macOS, Python 3.13, this branch.
- Exact command / steps: `uv run --extra dev python -m pytest tests/test_proxy_savings_history.py::test_stats_history_includes_cli_filtering`. The test hits `/stats-history` and asserts the payload carries `cli_filtering` with the RTK numbers the Historical tab reads.
- Observed result: the `/stats-history` response now carries `cli_filtering` (`tool`/`label`/`lifetime`/`session`), the field the Historical tab was missing after a restart. `ruff` and `mypy` are clean on the changed files.
- Not tested: I did not click through the rendered dashboard after a real restart, and this repo's test suite needs the native `_core` extension built (CI builds it), so the assertion runs in CI. The data the tab consumes is covered by the test, and the card is gated on that data being present.

## Review Readiness

## Checklist


JerrettDavis (approved):
The data-path change is narrow and matches the dashboard issue: `/stats-history` now attaches the same live RTK/CLI-filtering stats source that `/stats` already uses, and the Historical card is guarded so installs without RTK data do not render an empty card. The test pins the response contract the dashboard consumes.

Only PR-governance checks have run on the current branch, and GitHub reports it as dirty against main. I do not see a code blocker in this PR.

khaosdoctor:
Updated from main so it's not dirty anymore.

Took Copilot's two points as well: the `/stats-history` RTK lookup is best-effort now, if `_get_context_tool_stats()` throws it falls back to `cli_filtering=None` so the Historical tab stays up instead of 500ing. And I documented the curated shape on the endpoint since it's a subset of what `/stats` returns rather than the raw payload. Thanks for the review.

JerrettDavis (approved):
Reviewed the post-feedback commits and merge-from-main. I ran uv run --extra proxy --with pytest --with pytest-asyncio python -m pytest tests/test_proxy_savings_history.py -q: 17 passed. The /stats-history RTK augmentation is best-effort and the Historical tab stays gated on cli_filtering. No blocking issues found.

PR #1331, opened 2026-06-23
fix(proxy): add an Anthropic buffered read-timeout override

### LINKED ISSUE #1261: Proxy returns 502 (ReadTimeout) on long Opus /v1/messages turns: Anthropic upstream is buffered, not streamed, and capped at request_timeout_seconds (300s, no override)
## Summary

When proxying Anthropic `/v1/messages`, Headroom buffers the **entire** upstream response (it never streams from upstream) and caps the wait with a single httpx `read` timeout = `request_timeout_seconds` (**default 300s**). Long Opus turns on large contexts legitimately take longer than 300s to finish generating, so httpx raises `ReadTimeout` and the proxy returns **502 Bad Gateway** to the client (Claude Code). Failures arrive in **bursts** (a stall, then the connection pool flushes several timed-out requests at once).

Direct Claude Code → Anthropic does not hit this, because Claude Code consumes the response as an SSE stream and has no fixed ceiling on total generation time. Routing the same session through Headroom reintroduces a hard 300s whole-generation cap.

## Environment

- Headroom `0.26.0`
- Client: Claude Code `claude-cli/2.1.185`, `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`
- Model: `claude-opus-4-8`, requests with `stream: true`
- Linux (kernel 7.0), Python 3.14

## Observed behavior (from `~/.headroom/logs/proxy.log`)

```text
headroom.proxy - ERROR - [hr_...] Request failed: ReadTimeout:
event=proxy_inbound_response method=POST path=/v1/messages status=502 duration_ms=300958.99   # ~301s
event=proxy_inbound_response method=POST path=/v1/messages status=502 duration_ms=240366.20   # ~240s
event=proxy_inbound_response method=POST path=/v1/messages status=502 duration_ms=179148.37   # ~179s
```

- 10 such `Request failed: ReadTimeout` → `status=502` events on `/v1/messages` over a long session.
- They **cascade**: 5 failed within ~2ms of each other at `16:01:27`, and another 5 within ~2ms at `17:16:57` — consistent with the pool stalling on one long request and then flushing all the timed-out waiters together.
- The 502s carry no Anthropic `request_id` (the upstream read was aborted, so Anthropic never returned a completed response). They are proxy-generated, not Anthropic errors.
- Separately, the only Anthropic-origin errors in the same logs are ordinary `429 rate_limit_error` on tiny requests — unrelated to this report.

## Root cause (code references)

1. The Anthropic handler forwards upstream **buffered, never streamed** — there is no `http_client.stream(...)` call anywhere in `headroom/proxy/handlers/anthropic.py`; the main path is a buffered request that returns the full body at once:
   - `headroom/proxy/handlers/anthropic.py:2907` — `response = await self.http_client.request(method=..., url=..., headers=..., content=body)`
   - the result is returned as a fully materialized body: `Response(content=response.content, status_code=..., headers=...)` (~`anthropic.py:2505`)
   - `httpx.ReadTimeout` is caught by the catch-all at ~`anthropic.py:2518`, which returns 502 to the client.
   - (For contrast, the OpenAI handler *does* stream upstream — `headroom/proxy/handlers/openai.py:5804` `async with self.http_client.stream(...)`.)
2. The upstream client's `read`/`write` timeout is bound to `request_timeout_seconds`:
   - `headroom/proxy/server.py:~1155` — `httpx.Timeout(connect=self.config.connect_timeout_seconds, read=self.config.request_timeout_seconds, write=self.config.request_timeout_seconds)`
3. The default is 300s and there is **no env var or CLI flag** to override it (`ProxyConfig` is a plain `@dataclass`):
   - `headroom/proxy/models.py:233` — `request_timeout_seconds: int = 300`

Because the upstream call is buffered, `read=300s` effectively becomes a cap on the **entire** model generation, not an idle/per-read gap. Any Opus turn that generates for more than ~300s (common with large contexts and extended thinking) is guaranteed to 502, even though Anthropic is healthy and would have completed the request.

## Steps to reproduce

1. `headroom wrap claude` (proxy on `127.0.0.1:8787`).
2. Run a Claude Code session that accumulates a large context (hundreds of K tokens) and ask for a turn that produces a long Opus response (extended thinking + sizable output), so end-to-end generation exceeds ~300s.
3. Claude Code surfaces an API error; `proxy.log` shows `Request failed: ReadTimeout` and `status=502 duration_ms≈300000` on `/v1/messages`. Concurrent in-flight requests fail in the same burst.

## Suggested fix

**Primary: stream the Anthropic upstream response instead of buffering it.** Mirror the OpenAI handler — use `async with self.http_client.stream("POST", url, ...)` and forward chunks to the client via `StreamingResponse` (passing SSE through for `stream: true` requests). With a streamed upstream, the `read` timeout becomes an **idle/inter-chunk** timeout (Anthropic emits `ping`/event frames regularly), so a long-but-healthy generation no longer trips it, and the proxy also stops holding the full response in memory. This additionally avoids head-of-line blocking that produces the burst-cascade.

If full streaming of the compression/transform path is a larger change, two smaller interim mitigations:

- **Make `request_timeout_seconds` configurable** via env var (e.g. `HEADROOM_REQUEST_TIMEOUT_SECONDS`) and/or a `headroom wrap` flag, and raise the default for the Anthropic path (e.g. 600–900s). Right now it cannot be changed without editing source.
- **Decouple the read timeout from total generation time**: use a generous (or `None`) `read` timeout for the upstream `/v1/messages` call specifically, while keeping `connect` tight, so a slow-but-progressing generation isn't aborted.

I'm happy to attempt the streaming PR if that direction is welcome — wanted to file the analysis first since it's a behavioral change to the core forward path.

## Impact

Any sufficiently long Opus turn through the proxy fails with a 502, and because the failures cascade, a single long turn can take down several concurrent requests in the same instant. It only manifests on long/heavy sessions, which makes it look intermittent.

rodboev:
## Description

Buffered Anthropic `/v1/messages` requests still use Headroom's generic 300-second read timeout, which can produce proxy-generated `502 ReadTimeout` errors on long turns. This adds a dedicated buffered Anthropic timeout, keeps it applied across CCR and memory continuations plus batch paths, and makes the direct server entrypoint enforce the same positive-integer contract as the Click CLI. Closes #1261.

## Type of Change

## Changes Made

- Added `anthropic_buffered_request_timeout_seconds` for buffered Anthropic reads.
- Routed `/v1/messages`, CCR continuation, memory continuation, batch create, batch passthrough, and batch results through that timeout.
- Enforced the same positive-integer validation for `HEADROOM_ANTHROPIC_BUFFERED_REQUEST_TIMEOUT_SECONDS` and `--anthropic-buffered-request-timeout-seconds` in both startup paths.
- Added focused regressions and updated `CHANGELOG.md`.

## Testing

- [x] `uv run pytest tests/test_proxy/test_anthropic_buffered_timeout.py tests/test_cli_proxy_improvements.py::TestNewEnvVarWiring`
- [x] `uv run ruff check .`
- [x] `uv run ruff format . --check`

### Test Output

```text
$ uv run pytest tests/test_proxy/test_anthropic_buffered_timeout.py tests/test_cli_proxy_improvements.py::TestNewEnvVarWiring
17 passed in 3.42s

$ uv run ruff check .
All checks passed!

$ uv run ruff format . --check
966 files already formatted
```

## Real Behavior Proof

- Environment: local FastAPI `TestClient` with stubbed retry and HTTP client seams
- Exact command / steps: run `uv run pytest tests/test_proxy/test_anthropic_buffered_timeout.py tests/test_cli_proxy_improvements.py::TestNewEnvVarWiring`; the tests build `ProxyConfig(request_timeout_seconds=7, connect_timeout_seconds=3, anthropic_buffered_request_timeout_seconds=19)`, drive `/v1/messages`, `/v1/messages/batches`, `/v1/messages/batches/{batch_id}/results`, a CCR continuation, and a memory continuation through `TestClient`, then verify `HEADROOM_ANTHROPIC_BUFFERED_REQUEST_TIMEOUT_SECONDS=0` falls back to `600`, `--anthropic-buffered-request-timeout-seconds 0` is rejected, and default proxy timeouts stay `read=300` and `write=300`
- Observed result: buffered Anthropic paths use `httpx.Timeout(connect=3, read=19, write=7, pool=3)`, continuation requests stay on that same budget, invalid zero-valued startup config is rejected or ignored back to the default, and unrelated proxy timeout defaults stay unchanged
- Not tested: live upstream Anthropic latency beyond the focused stubbed-timeout regression

## Review Readiness

## Checklist

- [x] I have added tests that prove the fix


JerrettDavis (changes requested):
Thanks for tightening the Anthropic buffered timeout path. The shape is good, but there is one timeout regression that should be fixed before merge.

`headroom/proxy/server.py` changes `_retry_request(..., timeout: ... = None)` and then unconditionally passes `timeout=timeout` to `self.http_client.post(...)`. In httpx, `timeout=None` disables timeouts for that request; it does not mean "use the AsyncClient default". So every existing `_retry_request` caller that does not pass the new override now bypasses the configured generic timeout entirely.

Please only pass the `timeout` kwarg when an override is provided, or use the httpx client-default sentinel if you want the client-level timeout to apply. A focused regression should assert a normal `_retry_request` call without the Anthropic override still uses the proxy client's default timeout rather than an explicit `None`.

rodboev:
Good catch. `_retry_request` now only forwards the `timeout` kwarg to `httpx.AsyncClient.post()` when an explicit override was passed; otherwise the client-level default applies as before. Added a regression test that spies on the `.post()` call and asserts `timeout` is absent from the kwargs when no override is given.

JerrettDavis (approved):
The latest commit addresses the timeout=None regression I flagged: _retry_request now only passes a per-call timeout when an override is provided, so existing callers keep the httpx client default. The new regression pins that behavior, and the buffered Anthropic paths still receive the longer read budget.

Only PR-governance checks have run on the latest commit so far, but the previous code blocker is resolved from review.

JerrettDavis (approved):
Refreshing approval after the rebase: the code changes are materially the same as the previously reviewed version, including the client-default timeout regression fix. The full CI run is still in progress on the rebased commit, with the completed jobs green so far.

JerrettDavis (approved):
Refreshing approval after the TestClient portability update. The latest commit keeps the timeout behavior the same and only makes the regression test exercise the app through the stable client path. The previous blocker remains resolved: normal `_retry_request` callers no longer pass `timeout=None`, and buffered Anthropic paths still receive the explicit longer read timeout.

The latest CI run is still in progress on this head, with completed jobs green so far.

PR #1332, opened 2026-06-23
fix(code): slice tree-sitter byte offsets as UTF-8

### LINKED ISSUE #1319: [BUG] CodeAwareCompressor corrupts non-ASCII (CJK/emoji) source: tree-sitter byte offsets indexed into Python str → invalid syntax, 0% compression
Summary
  ▎
  ▎ CodeAwareCompressor silently fails to compress any Python file containing non-ASCII characters (CJK comments/docstrings, emoji, etc.). The AST compressor produces 
  ▎ syntactically broken code — identifiers and keywords get sliced mid-token (import os → port os, def second(): dropped entirely). The fail-safe _verify_syntax check then 
  ▎ discards the result and returns the original, so the user sees a 0% compression ratio with no error.
  ▎
  ▎ On a real Chinese-language Python codebase this hit 3 of 12 sampled files (25%), each forfeiting 30–56% of achievable savings. Pure-ASCII files in the same repo 
  ▎ compressed fine (35–87%), which is the tell.
  ▎
  ▎ Root cause
  ▎
  ▎ tree-sitter reports node positions as UTF-8 byte offsets (node.start_byte / node.end_byte). Several helpers index those offsets directly into the Python str, which is
  ▎ indexed by Unicode code point, not byte:
  ▎
  ▎ def _get_node_text(node: Any, code: str) -> str:
  ▎     return code[node.start_byte : node.end_byte]   # ← byte offset on a str
  ▎ (code_compressor.py:1980-1982; the same pattern appears in _compress_function_ast at the signature fallback code[node.start_byte : body_node.start_byte], :1571, and the
  ▎ top-level-code capture path.)
  ▎
  ▎ For ASCII, len(s.encode()) == len(s), so byte and char indices coincide and it works by accident. As soon as a multi-byte character appears earlier in the file, every
  ▎ subsequent node's slice shifts left by (bytes − chars), cutting into the middle of tokens downstream.
  ▎
  ▎ Minimal reproduction
  ▎
  ▎ from headroom.transforms.code_compressor import _get_node_text, _get_parser
  ▎ code = 'def first():\n    """中文占位"""\n    return 1\n\ndef second():\n    return 2\n'
  ▎ root = _get_parser("python").parse(bytes(code, "utf-8")).root_node
  ▎ second = [n for n in root.children if n.type == "function_definition"][1]
  ▎ print(repr(_get_node_text(second, code)))
  ▎ # → 'return 2\n'            (WRONG — sliced 18 bytes too early)
  ▎ # expected: 'def second():\n    return 2'
  ▎ Full-file effect:
  ▎ from headroom.transforms.code_compressor import CodeAwareCompressor, CodeCompressorConfig
  ▎ r = CodeAwareCompressor(CodeCompressorConfig(enable_ccr=False))
  ▎ res = r.compress(open("any_python_file_with_chinese_comments.py").read(), language="python")
  ▎ print(res.compression_ratio, res.syntax_valid)   # → 1.0 True  (silently gave up)
  ▎
  ▎ Suggested fix
  ▎
  ▎ Decode/encode once and slice on bytes, or convert byte offsets to char offsets. The localized fix is to make _get_node_text (and the two sibling byte-slice sites)
  ▎ operate on the UTF-8 bytes:
  ▎ def _get_node_text(node, code, code_bytes=None):
  ▎     b = code_bytes if code_bytes is not None else code.encode("utf-8")
  ▎     return b[node.start_byte : node.end_byte].decode("utf-8")
  ▎ tree-sitter Node also exposes node.text (already UTF-8 bytes) — using node.text.decode("utf-8") avoids manual offset handling entirely. A regression test compressing a
  ▎ class/function with a leading CJK docstring and asserting syntax_valid and ratio < 1.0 would lock it down.
  ▎
  ▎ Impact
  ▎
  ▎ No correctness risk (fail-safe returns original), but the compressor is a silent no-op on any non-ASCII source file — significant for non-English codebases, inline-emoji
  ▎ code, or unicode string literals. Tier-1 Python.
  ▎
  ▎ Environment
  ▎
  ▎ headroom main @ da1a3973, tree_sitter_language_pack, Python 3.14, macOS arm64.

aivinay:
## Description

CodeAwareCompressor was slicing Python strings with tree-sitter `start_byte` / `end_byte` offsets directly. That works for ASCII-only files, but it corrupts slices after non-ASCII source text such as CJK characters or emoji because tree-sitter offsets are UTF-8 byte offsets while Python string indexes are character offsets.

This caused code-aware compression to produce invalid intermediate Python and then safely fall back to the original file, resulting in 0% compression on affected files.

Closes #1319

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `_slice_code_bytes()` in `headroom/transforms/code_compressor.py` to slice source text using UTF-8 byte offsets.
- Updated `_get_node_text()` to use byte-safe slicing.
- Routed the other direct tree-sitter byte-offset slices through the same helper.
- Added regression tests in `tests/test_transforms/test_code_compressor.py`:
  - `test_get_node_text_uses_utf8_byte_offsets`
  - `test_ast_compresses_python_after_non_ascii_source`

## Testing

### Test Output

```text
$ .venv/bin/python -m pytest tests/test_transforms/test_code_compressor.py
68 passed, 1 warning

$ .venv/bin/python -m ruff check headroom/transforms/code_compressor.py tests/test_transforms/test_code_compressor.py
All checks passed!

$ .venv/bin/python -m ruff format --check headroom/transforms/code_compressor.py tests/test_transforms/test_code_compressor.py
2 files already formatted

$ git diff --check
# no output

$ /tmp/headroom-1319-venv/bin/python -m mypy headroom
headroom/proxy/server.py:1186: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
headroom/proxy/server.py:1257: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
headroom/proxy/server.py:1261: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
pyproject.toml: note: unused section(s): module = ['mlx.*']
Success: no issues found in 394 source files
```

## Real Behavior Proof

- Environment: macOS arm64, Python 3.14.5, tree-sitter 0.25.2, tree-sitter-language-pack 0.13.0
- Exact command / steps: On `main`, ran a local reproducer with a Python source string containing a CJK docstring before a second function; called `_get_node_text()` on the second tree-sitter function node; ran a full `CodeAwareCompressor.compress(...)` repro with non-ASCII module text before an import and a compressible function; re-ran both repros on this branch.
- Observed result: Before fix, `_get_node_text()` returned the wrong slice (`'nd():\n    return 2\n'` instead of `'def second():\n    return 2'`) and full compression fell back to the original file with `compression_ratio: 1.0`; after fix, `_get_node_text()` returns the full expected function slice and full compression succeeds with `compression_ratio < 1.0`, `syntax_valid: True`, and does not return the original.
- Not tested: Full repository test suite; live proxy/provider integrations; Windows/Linux platform-specific behavior.

## Review Readiness

## Checklist

N/A

## Additional Notes

- Documentation was not updated because this is an internal bug fix with no user-facing API or behavior change beyond restoring intended compression.
- `CHANGELOG.md` was not updated because the fix is narrow and issue-scoped; maintainers can advise if they want a changelog entry.
- The fix is intentionally small and targeted: it only changes how tree-sitter byte offsets are converted back into Python source text, without changing compression heuristics or language behavior.

### DIFF
--- a/headroom/transforms/code_compressor.py
+++ b/headroom/transforms/code_compressor.py
@@ -851,7 +851,7 @@ class CodeAwareCompressor(Transform):
         body_line_counts: dict[str, int] = {}
         for qname, node in definitions.items():
             collect_calls_in_function(node, qname)
-            node_text = code[node.start_byte : node.end_byte]
+            node_text = _slice_code_bytes(code, node.start_byte, node.end_byte)
             body_line_counts[qname] = max(1, len(node_text.split("\n")) - 2)
 
         # Reference counts: subtract definition occurrences
@@ -1257,8 +1257,8 @@ class CodeAwareCompressor(Transform):
                             child, code, language, lang_config, body_limits, analysis
                         )
                         # Reconstruct export with compressed inner definition
-                        export_prefix = code[node.start_byte : child.start_byte]
-                        export_suffix = code[child.end_byte : node.end_byte]
+                        export_prefix = _slice_code_bytes(code, node.start_byte, child.start_byte)
+                        export_suffix = _slice_code_bytes(code, child.end_byte, node.end_byte)
                         structure.function_signatures.append(
                             export_prefix + compressed + export_suffix
                         )
@@ -1568,7 +1568,7 @@ class CodeAwareCompressor(Transform):
         if signature_lines:
             result_parts.extend(signature_lines)
         else:
-            sig_text = code[node.start_byte : body_node.start_byte].rstrip()
+            sig_text = _slice_code_bytes(code, node.start_byte, body_node.start_byte).rstrip()
             result_parts.append(sig_text)
 
         if opening_brace_line is not None:
@@ -1977,9 +1977,14 @@ class CodeAwareCompressor(Transform):
 # =========================================================================
 
 
+def _slice_code_bytes(code: str, start_byte: int, end_byte: int) -> str:
+    """Extract source text using tree-sitter UTF-8 byte offsets."""
+    return code.encode("utf-8")[start_byte:end_byte].decode("utf-8")
+
+
 def _get_node_text(node: Any, code: str) -> str:
     """Extract text from AST node."""
-    return code[node.start_byte : node.end_byte]
+    return _slice_code_bytes(code, node.start_byte, node.end_byte)
 
 
 def _get_definition_name(node: Any) -> str | None:
--- a/tests/test_transforms/test_code_compressor.py
+++ b/tests/test_transforms/test_code_compressor.py
@@ -1380,6 +1380,50 @@ class TestRealASTRuns:
         # Output is still valid Python.
         compile(result.compressed, "<test>", "exec")
 
+    def test_get_node_text_uses_utf8_byte_offsets(self):
+        """tree-sitter byte offsets must not be sliced as Python str indexes."""
+        from headroom.transforms.code_compressor import _get_node_text, _get_parser
+
+        code = 'def first():\n    """中文占位"""\n    return 1\n\ndef second():\n    return 2\n'
+        root = _get_parser("python").parse(code.encode("utf-8")).root_node
+        functions = [node for node in root.children if node.type == "function_definition"]
+
+        assert _get_node_text(functions[1], code) == "def second():\n    return 2"
+
+    def test_ast_compresses_python_after_non_ascii_source(self):
+        """CJK/emoji before a later function must not corrupt downstream slices."""
+        compressor = CodeAwareCompressor(
+            CodeCompressorConfig(
+                min_tokens_for_compression=1,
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):

aivinay:
@JerrettDavis Thanks for the approval! CI is only failing in unrelated memory/HuggingFace cache tests. Could you please rerun the failed jobs?

PR #1201, opened 2026-06-20
ci: add Windows (win_amd64) wheel to release matrix

lennney:
## Description

Adds `windows-latest` / `x86_64-pc-windows-msvc` to the `build-wheels` matrix in `release.yml`, so every release publishes a `cp310-abi3-win_amd64` wheel to PyPI alongside the existing Linux and macOS wheels.

Also adds a Windows entry to the `smoke-import-wheels` gate so the wheel is verified before publish.

Closes #974
Closes #729
Supersedes #1086

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Added `windows-latest` / `x86_64-pc-windows-msvc` / `manylinux: ""` to the `build-wheels` strategy matrix
- Added Windows entry to `smoke-import-wheels` matrix (`runner: windows-latest`, `wheel_target: x86_64-pc-windows-msvc`)
- Extended smoke-import host step from macOS-only to cross-platform with `case` on `WHEEL_TARGET` for wheel glob, python command detection (`python3.XX` vs `python`), and venv path detection (`bin/` vs `Scripts/`)

## Testing

- [ ] Unit tests pass (`pytest`)
- [ ] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [ ] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
> maturin build --release --out dist --interpreter python3.12
📦 Built wheel for abi3 Python ≥ 3.10: headroom_ai-0.26.0-cp310-abi3-win_amd64.whl

> pip install dist/headroom_ai-0.26.0-cp310-abi3-win_amd64.whl
Successfully installed headroom-ai-0.26.0

> python -c "from headroom._core import hello; print(hello())"
hello from headroom._core (Rust)

> python -c "from headroom._core import detect_content_type; r = detect_content_type('hello world'); print(r.content_type, r.confidence)"
plain_text 0.5
```

## Real Behavior Proof

- Environment: Windows 10 x86_64, Python 3.12.10, Rust 1.96.0 (stable-x86_64-pc-windows-msvc), MSVC Build Tools 2022 (v14.44)
[...]

PR #1335, opened 2026-06-23
ci(release): publish win_amd64 wheel so Windows installs need no Rust (#1328)

### LINKED ISSUE #1328: Windows (win_amd64) wheel missing - cannot install on Windows without Rust toolchain
## Problem

`headroom-ai` is not installable on Windows without a local Rust toolchain.

PyPI currently publishes wheels only for:
- `macosx_11_0_arm64`
- `manylinux_2_28_x86_64`
- `manylinux_2_28_aarch64`

No `win_amd64` wheel is published for any Python version.

## What happens

pip / uv attempt to build from the sdist, which requires `maturin` as build backend.
`maturin` tries to download the Rust toolchain via rustup from `static.rust-lang.org` and
crate dependencies from `crates.io`. In sandboxed / corporate environments (GitHub
Copilot CLI, CI runners, air-gapped machines) these hosts are unreachable ? installation fails.

Error seen:
```
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
error: failed to get pyo3-macros as a dependency of package pyo3 v0.24.2
  [28] Timeout was reached (Failed to connect to index.crates.io port 443)
```

## Expected

A `headroom_ai-x.y.z-cp310-abi3-win_amd64.whl` wheel published to PyPI alongside
each release, built via the existing CI pipeline (e.g. `windows-latest` runner in
GitHub Actions with `maturin build --release`).

## Environment

- OS: Windows Server x64
- Python: 3.12.10
- pip: 25.0.1 / uv: 0.11.23
- headroom-ai version attempted: 0.27.0

Lakshya77089:
## Description

We ship wheels for macOS arm64 and manylinux x86_64/aarch64, but there's no `win_amd64` wheel on PyPI for any Python version. So on Windows, pip/uv can't find a binary and try to build from the sdist with maturin, which pulls the Rust toolchain from static.rust-lang.org and crates from crates.io. On locked-down machines (corporate proxies, CI runners, the GitHub Copilot CLI sandbox, anything air-gapped) those hosts aren't reachable and the install just dies:

```
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
error: failed to get pyo3-macros as a dependency of package pyo3 v0.24.2
  [28] Timeout was reached (Failed to connect to index.crates.io port 443)
```

This adds the Windows wheel to the release matrix so `pip install headroom-ai` works on Windows without a local Rust install.

Closes #1328

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added a `windows-latest` / `x86_64-pc-windows-msvc` row to the `build-wheels` matrix. The runner already has MSVC and maturin-action sets up Rust, so it produces `headroom_ai-*-win_amd64.whl` on every release. I checked `crates/headroom-core/Cargo.toml` first — the Windows ONNX path is already on `ort-load-dynamic` under `cfg(windows)`, so the wheel loads ORT at runtime instead of linking the DirectML SDK libs. Nothing else was needed on the Rust side.
- Added a matching `windows-latest` row to `smoke-import-wheels` so a broken Windows wheel blocks publish like the other platforms do. Windows needed its own step: the venv puts Python under `Scripts\` not `bin/`, and the runner defaults to pwsh. I also pinned the shared script-staging step to `shell: bash` since it uses a heredoc that pwsh can't run (Git Bash is on the runner), and added a `setup-python` step to get the right minor version.
- Updated the README install section so the "install Rust first" workaround is clearly only for the sdist fallback (e.g. Intel macOS) now that Windows/Linux/macOS-arm64 all have prebuilt wheels.

## Testing

This is a CI workflow + docs change, no Python runtime code. I leaned on the existing `tests/test_release_workflows.py` structural gates plus a YAML parse and matrix-shape sanity check.

### Test Output

```text
$ python -m pytest tests/test_release_workflows.py -q
28 passed, 1 skipped, 1 failed
# The one failure, test_no_native_tls_in_wheel_build_tree, shells out to cargo, which
# isn't installed here. I confirmed with `git stash` that it fails the same way on main
# without my changes, so it's pre-existing and unrelated.

$ python -c "import yaml; d=yaml.safe_load(open('.github/workflows/release.yml',encoding='utf-8')); \
  j=d['jobs']; print('build-wheels rows:', len(j['build-wheels']['strategy']['matrix']['include'])); \
  print('smoke rows:', len(j['smoke-import-wheels']['strategy']['matrix']['include']))"
build-wheels rows: 4
smoke rows: 6
```

## Real Behavior Proof

- Environment: Windows 11 local clone; CI runs on GitHub-hosted `windows-latest`.
- Exact command / steps: edited the build-wheels and smoke-import-wheels matrices in `.github/workflows/release.yml` and the README, then ran the release-workflow tests and the YAML/matrix-shape check above.
- Observed result: tests pass, YAML parses, build matrix is now 4 rows (Linux x64, Linux arm64, macOS arm64, Windows x64) and the smoke matrix is 6 rows including the new native Windows row.
- Not tested: the actual win_amd64 build + PyPI publish. Those jobs only run in the release workflow on a tag or workflow_dispatch, not on a feature PR. The PR-time release dry-run will exercise the new rows once a maintainer approves the workflow run. I couldn't run `maturin build --target x86_64-pc-windows-msvc` end to end here.

## Review Readiness

## Checklist

## Additional Notes

- No new test file: the existing structural gates in `tests/test_release_workflows.py` (`test_build_wheels_matrix_excludes_intel_macos`, `test_aarch64_wheel_uses_native_arm64_runner`, the smoke-import gate test) already assert the matrix contract and still pass with the Windows row added.
- I didn't touch CHANGELOG.md — release-please generates it from the Conventional Commit subject, so the `ci(release):` commit gets picked up automatically.
- The win_amd64 wheel actually shows up on PyPI on the next tagged release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -259,6 +259,19 @@ jobs:
           - os: macos-14
             target: aarch64-apple-darwin
             manylinux: ""
+          # Windows x86_64. `windows-latest` ships the MSVC toolchain and
+          # `maturin-action` provisions the Rust toolchain, so users no
+          # longer need a local Rust install to `pip install headroom-ai`
+          # on Windows (issue #1328 — sdist build was failing in sandboxed
+          # / air-gapped environments that can't reach static.rust-lang.org
+          # or crates.io). `manylinux` is meaningless on Windows (that tag
+          # is Linux-only) — leave it empty. The ONNX Runtime story is
+          # already handled: `crates/headroom-core/Cargo.toml` selects
+          # `ort-load-dynamic` under `cfg(windows)`, so the wheel does not
+          # bundle/link DirectML SDK libs and loads ORT at runtime instead.
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
+            manylinux: ""
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v6
@@ -499,6 +512,13 @@ jobs:
           # ship today (Intel macOS dropped in PR #371 — ort-sys
           # has no x86_64-apple-darwin prebuilts).
           - { runner: macos-14, image: "", python: "3.13", wheel_target: aarch64-apple-darwin, wheel_artifact: wheels-macos-14-aarch64-apple-darwin, glibc_label: "" }
+          # Windows x86_64 — runs natively on the host runner; no
+          # container. `image: ""` routes this row past the Linux docker
+          # path; a dedicated PowerShell step below installs + imports the
+          # wheel (venv layout differs: Scripts\ not bin/). This is the
+          # gate that issue #1328's win_amd64 wheel must clear before
+          # publish.
+          - { runner: windows-latest, image: "", python: "3.12", wheel_target: x86_64-pc-windows-msvc, wheel_artifact: wheels-windows-latest-x86_64-pc-windows-msvc, glibc_label: "" }
 
     steps:
       - name: Download wheels artifact for this target
@@ -506,6 +526,17 @@ jobs:
         with:
           name: ${{ matrix.wheel_artifact }}
           path: dist/
+
+      # Windows is the only host-runner row that can't rely on a
+      # preinstalled interpreter at the exact requested minor (the Linux
+      # rows install inside their container; the macOS runner ships
+      # multiple Pythons). Provision it explicitly so the requested minor
+      # is on PATH as `python` for the smoke step below.
+      - name: Set up Python (Windows host)
+        if: matrix.image == '' && runner.os == 'Windows'
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ matrix.python }}
 
       - name: Stage smoke-import script
         # The smoke script lives in a host file rather than an inline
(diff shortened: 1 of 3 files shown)


JerrettDavis (changes requested):
This overlaps heavily with the newer Windows wheel release work in #1201, which has the later `python -m pip` Windows fix and a cleaner cross-platform host smoke step. I do not think we should merge both versions of the release workflow changes.

There is also a current required-check blocker on this branch: the Release workflow fails in `detect-version` with `ModuleNotFoundError: No module named 'headroom'` while running `python headroom/release_version.py` from the PR workflow. Please either close this as superseded by #1201 or rebase/adapt onto that approach and fix the release workflow failure before this is considered ready.

JerrettDavis (approved):
Re-reviewed after the `release_version.py` import fix and release workflow rerun. The previous `detect-version` blocker is resolved, the Windows wheel build and smoke-import rows are exercised by the Release workflow, and the full reported check set is green. This PR now appears to be the viable Windows-wheel release path; it overlaps with #1201, so we should not merge both.

PR #1337, opened 2026-06-23
fix(backends): emit real input_tokens in Bedrock streaming message_start

### LINKED ISSUE #1132: Bedrock streaming: message_start emits input_tokens=0, breaking downstream OTel metrics

ashishpatel26:
## Problem

Bedrock streaming responses return `0` for `input_tokens` in `message_start` events instead of the actual token count, causing incorrect usage tracking.

## Fix

Parse and forward the real `input_tokens` value from Bedrock's `message_start` streaming event in the LiteLLM backend.

## Changes

- `headroom/backends/litellm.py` — extract `input_tokens` from Bedrock `message_start` event
- `tests/test_backends/test_litellm_stream_input_tokens.py` — test coverage for streaming token counts

Fixes #1132


JerrettDavis (changes requested):
Thanks for cleaning up the PR and adding targeted coverage. I do think the current implementation needs a different shape before merge.

`headroom/backends/litellm.py` now buffers the full LiteLLM stream before yielding any Anthropic SSE events so it can read the final usage chunk and put token counts on `message_start`. That fixes the accounting field, but it changes the behavior of every LiteLLM/Bedrock streaming request from true streaming to fully buffered. Long Claude/Bedrock turns will no longer deliver incremental content, clients will see much higher time-to-first-token, and large responses now accumulate in memory before the first byte is sent downstream.

Please preserve streaming semantics. For the `message_start.input_tokens` problem, prefer a source available before content streaming starts, such as the request-side prompt token count Headroom already computed/sent upstream, or another pre-stream usage signal if LiteLLM exposes one. If final usage is still useful, it can update `message_delta`/final accounting later, but it should not require delaying the whole stream.

A regression test should assert that `stream_message()` yields `message_start` and the first content delta before the upstream async iterator is exhausted.

JerrettDavis:
Conflict audit against current `main`: this cannot be resolved mechanically without changing runtime semantics. The PR buffers the complete LiteLLM response before emitting `message_start` so provider-final usage can appear in the initial event; that removes first-token streaming for all LiteLLM providers. Current main deliberately emits `message_start` immediately and carries final input/cache usage on terminal `message_delta` to preserve streaming and protocol event order. This needs a scoped Bedrock accounting/consumer redesign, not a conflict-only merge, so I have not pushed a behavior-changing resolution.

JerrettDavis:
This conflict should not be resolved by rebasing the original implementation. Current `main` already includes the corrected streaming-usage behavior from #2144 (`0ad7dc7c`): it requests `stream_options.include_usage`, captures Bedrock/LiteLLM input and cache token usage from the trailing usage chunk, and reports those values on terminal `message_delta.usage` while preserving normal streaming latency and Anthropic event order.

This PR buffers the entire upstream response so it can move final usage into `message_start`; that would delay all content until generation completes and regress true streaming. The original issue is therefore superseded by #2144’s protocol-safe implementation. Please close this PR rather than resolving its conflicts.

ashishpatel26:
Closing as superseded — same root issue (#1132) as #1320, already fixed on main via #1305 (commit 70cc96a3): message_start.input_tokens is backfilled from optimized_tokens (Headroom's pre-computed token count, available before the stream starts) with true streaming preserved. Verified locally: tests/test_bedrock_streaming_input_tokens.py — 2/2 pass.

This branch's full-stream-buffering approach was the cleaner of the two duplicate attempts, but the underlying accounting gap it targets no longer exists on main, and buffering the whole LiteLLM/Bedrock response before emitting message_start would regress time-to-first-token for every streaming request, per @JerrettDavis's review. Closing rather than reworking, since there's nothing left to fix.

PR #1339, opened 2026-06-23
fix(doctor): recognize Azure Foundry routing for Claude Code

jurikern:
## Description

`headroom doctor` only inspected `ANTHROPIC_BASE_URL` when checking Claude Code routing. In Azure Foundry mode, Claude Code routes through `ANTHROPIC_FOUNDRY_BASE_URL` when `CLAUDE_CODE_USE_FOUNDRY=1`, so a correctly wired setup was reported as `claude: not routed`.

This PR falls back to `ANTHROPIC_FOUNDRY_BASE_URL` only when the Foundry flag is present, then sends that URL through the existing routing classifier. Non-Foundry Claude routing behavior is unchanged.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Tests only

## Changes Made

- `headroom/cli/doctor.py`: recognize `ANTHROPIC_FOUNDRY_BASE_URL` when `CLAUDE_CODE_USE_FOUNDRY=1` and `ANTHROPIC_BASE_URL` is absent.
- `tests/test_cli_doctor.py`: add coverage for a valid Foundry route, missing Foundry URL, and a Foundry URL without the Foundry flag.

## Testing

- [x] Unit tests
- [x] Lint/static checks
- [ ] Integration tests

### Test Output

```text
uv run --extra dev python -m pytest tests/test_cli_doctor.py -q
44 passed in 0.96s

uv run --extra dev ruff check headroom/cli/doctor.py tests/test_cli_doctor.py
All checks passed!
```

## Real Behavior Proof

- Environment: macOS; Claude Code with `CLAUDE_CODE_USE_FOUNDRY=1` and `ANTHROPIC_FOUNDRY_BASE_URL` set in `~/.claude/settings.json`; Headroom proxy on `:8787` with the Foundry URL pointed at an Azure APIM upstream.
- Exact command / steps: set `ANTHROPIC_FOUNDRY_BASE_URL` to `http://localhost:8787` in the Claude settings env block, then run `uv run headroom doctor`.
- Observed result: the Claude row changed from a false `not routed` warning to a routed/pass result, and a live `/v1/messages` request forwarded through the proxy to the Azure upstream and returned 200.
- Not tested: non-loopback Foundry URLs; those reuse the existing `_classify_routing_url` warning path.

## Review Readiness


JerrettDavis (approved):
This looks ready from the code review side. The Foundry fallback stays gated on `CLAUDE_CODE_USE_FOUNDRY`, so existing Anthropic routing behavior is unchanged, and the added tests cover the pass, missing-base warning, and ignored-without-flag cases.

I only see PR governance checks on this branch rather than the full CI matrix, but I do not see a code blocker in the change itself.

JerrettDavis (approved):
Re-reviewed after refreshing onto current `main`. The Foundry doctor fallback remains gated on `CLAUDE_CODE_USE_FOUNDRY`, so non-Foundry Claude routing behavior is unchanged.

Local validation on the updated head:

```text
uvx --from ruff==0.15.17 ruff check headroom/cli/doctor.py tests/test_cli_doctor.py
uvx --from ruff==0.15.17 ruff format --check headroom/cli/doctor.py tests/test_cli_doctor.py
uv run pytest tests/test_cli_doctor.py -q
64 passed
```

The previous `test (3)` failure was in `tests/test_memory_bridge.py` due missing Hugging Face cache in offline CI, not this doctor change. No code blocker found.

PR #1343, opened 2026-06-23
feat(cache): attribute prompt-cache misses to TTL lapse vs prefix change (#1313)

Lakshya77089:
## Description

A low prompt-cache hit rate is hard to act on without knowing *why* turns miss. Two very different causes need very different responses:

- **TTL lapse** — the session went idle longer than the provider's cache lifetime, so the entry expired. The fix is a longer TTL (e.g. Anthropic's 1h breakpoint instead of the 5m default).
- **Prefix change** — the cacheable message prefix shifted, so the new request couldn't match the cached key. A longer TTL won't help here at all.

Right now those look identical from the dashboard (just "cache_read was 0"). This adds the attribution so a user can actually decide 5m vs 1h.

Closes #1313

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

`PrefixCacheTracker` already kept the previous turn's forwarded messages and a per-turn activity timestamp, so the signal was already there — it just wasn't being read.

- **`prefix_tracker.py`** — `classify_cache_miss()`: when a turn expected a cached prefix (non-zero cached tokens last turn) but read 0 this turn, returns `ttl_expiry` if the idle gap exceeded the provider cache TTL, else `prefix_change` if the forwarded prefix differs from last turn's, else `unknown`. **TTL wins ties** — once the entry lapsed, a coincident content change is moot, and the 5m-vs-1h decision is exactly what the TTL signal answers. A 1h-breakpoint session can widen the window via `PrefixFreezeConfig.cache_ttl_seconds`. Cold starts and hits return `is_miss=False`.
- **Anthropic handlers (streaming + non-streaming)** — classify BEFORE `update_from_response` overwrites the last-turn state the classifier reads, then record the reason.
- **`prometheus_metrics.py`** — a per-provider/per-reason counter, `record_cache_miss_attribution()`, reset handling, and a `headroom_cache_miss_attribution_total{provider,reason}` export series.
- **`cost.py`** — `build_prefix_cache_stats()` aggregates a `miss_attribution` block (per-provider + totals, with the ttl/prefix split as a % of *attributed* misses, so `unknown` doesn't dilute the headline).
- **dashboard** — a "Cache Miss Attribution" panel (TTL expiry / prefix change / unknown / total) with a "mostly TTL lapse" vs "mostly prefix change" headline.

Scoped to Anthropic for this first cut (where the tracker is fully wired); OpenAI/Gemini can follow once the shape is proven.

## Testing

### Test Output

```text
$ python -m pytest tests/test_cache/test_prefix_tracker.py -q
38 passed
# 29 existing + 9 new classifier tests (TestClassifyCacheMiss).

$ python -m pytest tests/test_proxy_cache_ttl_metrics.py -k "miss_attribution or reset_runtime_clears" -q
5 passed, 8 deselected
# new: counter bucketing, stats aggregation, empty case, /metrics export, reset.
```

The full `test_proxy_cache_ttl_metrics.py` / `test_proxy_dashboard_stats_cache.py` files have some failures in this sandbox (`test_stats_endpoint_*`, streaming-parser, reset-counters) — those spin up the proxy server / Rust `_core` extension, which isn't built here. I confirmed via `git stash` that they fail identically on `main` without my changes, so they're pre-existing and unrelated. My additions to the stats dict are purely additive and don't break any passing assertion.

## Real Behavior Proof

- Environment: Windows 11, Python 3.10. The Rust `_core` extension and a live proxy aren't available in this checkout.
- Exact command / steps: drove `classify_cache_miss()` through every branch with a faithful warm-then-miss sequence; drove `record_cache_miss_attribution()` → `build_prefix_cache_stats()` → `export()` end to end.
- Observed result: classifier returns `cold_start`/`hit`/`ttl_expiry`/`prefix_change`/`unknown` correctly, TTL wins the tie when both signals fire, a growing (append-only) prefix is treated as stable, and the 1h override widens the window. The stats builder produces `miss_attribution.totals` (`ttl_expiry`/`prefix_change`/`unknown`/`total` + `ttl_expiry_pct`/`prefix_change_pct` over attributed misses) and `by_provider`; `/metrics` emits `headroom_cache_miss_attribution_total{provider="anthropic",reason="ttl_expiry"}`.
- Not tested: a live Anthropic session through the running proxy with a real idle-then-resume to confirm the handler wiring fires end-to-end. I verified the handler integration by reading scope/order (classify before `update_from_response`, `provider_name`/`self.metrics` in scope) and unit-tested every layer it calls, but didn't exercise the actual server loop.

## Review Readiness

## Checklist

## Additional Notes

- The classifier is intentionally pure (takes the cache-read result + current forwarded messages + an optional idle override) so it's order-independent and unit-testable without a live tracker clock.
- No README/docs change yet — this surfaces in the dashboard and `/metrics`, which are self-describing; happy to add a docs page if you'd like one.
- CHANGELOG.md isn't touched — release-please generates it from the `feat(cache):` commit subject.
- Follow-ups if useful: extend to OpenAI/Gemini handlers, and add a per-provider breakdown row in the dashboard panel (the stats already carry `by_provider`).

### DIFF
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -41,6 +41,21 @@ _PROVIDER_WRITE_PENALTY = {
     "bedrock": 0.25,
 }
 
+# Default prompt-cache lifetime per provider, in seconds. Used by
+# `classify_cache_miss` to decide whether a miss is most likely a TTL
+# lapse (idle longer than this) versus a prefix-content change. Anthropic's
+# default ephemeral cache is 5 minutes (matches
+# headroom.cache.anthropic.ANTHROPIC_CACHE_TTL_SECONDS); the others are best-
+# effort defaults and only matter once those providers are wired in. A
+# session that opts into Anthropic's 1h cache breakpoint can override this
+# via the tracker config (see PrefixFreezeConfig.cache_ttl_seconds).
+_PROVIDER_CACHE_TTL_SECONDS = {
+    "anthropic": 300,  # 5 minutes (default ephemeral cache)
+    "openai": 300,  # automatic prefix cache, ~5-10 min; conservative floor
+    "gemini": 300,
+    "bedrock": 300,
+}
+
 
 @dataclass
 class PrefixFreezeConfig:
@@ -50,6 +65,11 @@ class PrefixFreezeConfig:
     min_cached_tokens: int = 1024  # Min cached tokens to activate freeze
     session_ttl_seconds: int = 600  # Session tracker cleanup TTL
     force_compress_threshold: float = 0.5  # Bust cache if compression saves > this fraction
+    # Provider prompt-cache lifetime used by `classify_cache_miss` to tell a
+    # TTL lapse from a prefix change. `None` falls back to the per-provider
+    # default in `_PROVIDER_CACHE_TTL_SECONDS`. Set to 3600 for a session that
+    # uses Anthropic's 1h cache breakpoint so idle-gap attribution stays honest.
+    cache_ttl_seconds: int | None = None
 
 
 @dataclass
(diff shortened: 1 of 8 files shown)


JerrettDavis (changes requested):
Thanks for adding attribution all the way from tracker state to metrics and the dashboard. The feature shape is useful, but this needs another pass before merge.

CI is failing in ways tied to this diff:

- `ruff format --check .` reports that `headroom/proxy/handlers/anthropic.py` and `tests/test_cache/test_prefix_tracker.py` would be reformatted.
- `tests/test_anthropic_pre_upstream_backpressure.py::test_memory_context_timeout_fails_open_and_releases_semaphore` now returns 502. The captured log shows `AttributeError: 'types.SimpleNamespace' object has no attribute 'classify_cache_miss'` from `headroom.proxy.handlers.anthropic`. That test uses a lightweight tracker double, so the new handler call needs either the test double updated or the production code guarded in the same way as other optional tracker capabilities. Right now a missing attribution method can turn a fail-open timeout path into a request failure, which is the wrong failure mode for observability-only attribution.

Please fix the formatter issues and make cache-miss attribution non-disruptive for existing tracker/test-double paths. A small regression around the memory-context timeout path would be useful so attribution cannot break fail-open behavior again.

JerrettDavis (approved):
Re-reviewed after the guard and formatting fixes. Cache-miss attribution is now non-disruptive for lightweight tracker doubles because the handler checks for `classify_cache_miss` before calling it, and the earlier fail-open backpressure regression is covered by green CI. The metric/stat/dashboard additions remain additive. No remaining blocker found.

PR #1351, opened 2026-06-23
feat(wrap): add --1m to preserve the 1M context window on wrap claude (#1158)

### LINKED ISSUE #1158: [FEATURE] `headroom wrap claude` should preserve the 1M context window (Claude Code drops context-1m behind a custom ANTHROPIC_BASE_URL)
## Problem Statement

`headroom wrap claude` is the recommended Claude Code integration, but it silently caps the usable context at **200k** for subscription users who are entitled to the **1M** window. Plain `claude` gives 1M; the same config through `wrap claude` drops to 200k (status line and auto-compaction both assume 200k).

This is **not a headroom bug** — the root cause is in Claude Code: when `ANTHROPIC_BASE_URL` points at a custom host (the headroom proxy), Claude Code does **not** send the `context-1m-2025-08-07` beta header, and treats the window as 200k. I verified this from headroom's own inbound proxy log (`~/.headroom/logs/proxy.log`): while wrapped, the `anthropic-beta` header Claude Code sends omits `context-1m` entirely, e.g.

```
anthropic-beta: claude-code-20250219,oauth-2025-04-20,interleaved-thinking-2025-05-14,
                context-management-2025-06-27,token-counting-2024-11-01
```

headroom itself is fine: it forwards `anthropic-beta`, doesn't intercept `/v1/models`, and already sizes Opus 4.8 at 1M internally via LiteLLM. Upstream tracking issue: anthropics/claude-code#68522.

The reason to handle it here: `wrap claude` owns the launched `claude` process's environment, and it's the documented Claude Code path — so users will hit this and (understandably) blame headroom first.

## Proposed Solution

On the `wrap claude` path, optionally preserve the 1M window. The workaround that works is forcing the model suffix via env on the launched process:

```
ANTHROPIC_MODEL='claude-opus-4-8[1m]' claude
```

With that, Claude Code sends `context-1m-2025-08-07` and the 1M window activates through the proxy. (Note: `CLAUDE_CODE_AUTO_COMPACT_WINDOW=1000000` alone does **not** lift the cap, and relying on the saved `/model opus[1m]` picker selection does not survive a custom base URL.)

Preferred shape — an opt-in flag:

```
headroom wrap claude --1m        # sets ANTHROPIC_MODEL='<resolved-opus-model>[1m]' for the launched process
```

Opt-in is safer than auto-enabling, because `sonnet[1m]` bills usage credits on every plan and Opus[1m] needs usage credits on Pro — so silently appending `[1m]` could cost users money unexpectedly.

At minimum, a docs note on the `wrap` page describing the gotcha and the env workaround (linking claude-code#68522) would already save users a long debugging session.

## Use Case

- Daily driver: Claude Code on a Max plan via `headroom wrap claude`, working in large codebases where the 1M window matters.
- Without this, you must choose per-session between compression (`wrap claude`, capped at 200k) and the full 1M window (plain `claude`, no compression) — you can't get both, even though the wire and the proxy both support 1M.
- Token impact: lets long sessions use the full 1M window *and* headroom compression together, instead of being forced to 200k.

## Alternatives Considered

- **Inject `context-1m` into the outbound `anthropic-beta` at the proxy.** Ineffective: even if the wire then allows 1M, Claude Code's local accounting still assumes 200k and auto-compacts there, so usable context wouldn't change. The fix has to make Claude Code itself treat the window as 1M, which the `ANTHROPIC_MODEL='...[1m]'` env does.
- **`CLAUDE_CODE_AUTO_COMPACT_WINDOW=1000000`.** Tested, no effect behind a custom base URL.
- **Leave it entirely to the upstream Claude Code fix (claude-code#68522).** Reasonable long-term, but the env workaround is available now and `wrap claude` is the natural place to apply it.

## Example API (Optional)

```
# Opt-in flag on the wrap path
headroom wrap claude --1m

# Equivalent to what wrap would set for the launched process:
ANTHROPIC_MODEL='claude-opus-4-8[1m]' claude
```

## Environment

- headroom-ai 0.26.0 (installed via pipx), macOS (arm64)
- Claude Code 2.1.183 (native), Max plan, model `opus[1m]` (`claude-opus-4-8`)

ousamabenyounes:
## Description

`headroom wrap claude` is the recommended Claude Code integration, but for subscription users entitled to the **1M** context window it silently caps usable context at **200k**. Root cause (upstream, anthropics/claude-code#68522): when `ANTHROPIC_BASE_URL` points at a custom host (the Headroom proxy), Claude Code does **not** send the `context-1m-2025-08-07` beta header and treats the window as 200k. The `/model opus[1m]` picker selection does not survive a custom base URL, and `CLAUDE_CODE_AUTO_COMPACT_WINDOW` alone does not lift the cap.

Headroom itself already forwards `anthropic-beta` and sizes Opus at 1M internally — but since `wrap claude` owns the launched process's environment and is the documented path, users hit this and blame Headroom first. This adds the opt-in fix the issue proposes.

Closes #1158

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cli/wrap.py`: new opt-in `--1m` flag on `wrap claude`. When set, `ANTHROPIC_MODEL=<opus>[1m]` is exported on the launched process so Claude Code sends the `context-1m` beta header. Logic extracted to a testable helper `_resolve_1m_model`: a model the user already selected via `ANTHROPIC_MODEL` is preserved (only the `[1m]` suffix is appended when missing); otherwise it falls back to the default Opus. Idempotent (no double suffix). Default behavior is unchanged (opt-in).
- `tests/test_cli/test_wrap_helpers.py`: unit tests for `_resolve_1m_model` (append-to-user-model, idempotent, default fallback).
- `README.md`: `--1m` added to the Claude Code row of the agent compatibility matrix.
- `CHANGELOG.md`: Unreleased → Features entry.

## Testing

### Test Output

```text
$ uv run pytest tests/test_cli/test_wrap_helpers.py tests/test_cli/test_wrap_claude_base_url.py -q
61 passed in 0.46s

$ uv run ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_helpers.py
All checks passed!

$ uv run mypy headroom/cli/wrap.py
```

#### TDD verification (RED → GREEN)

RED — new tests with the prod change reverted (`_resolve_1m_model` absent):
```text
E   AttributeError: module 'headroom.cli.wrap' has no attribute '_resolve_1m_model'
3 failed, 40 deselected in 0.56s
```
GREEN — with the change applied:
```text
3 passed, 40 deselected in 0.34s
```

## Real Behavior Proof

- Environment: Linux, Python 3.13, headroom @ this branch.
- Exact command / steps: `headroom wrap claude --1m --help` shows the new flag, and the flag resolves the model id that triggers the 1M window:
  ```text
  $ headroom wrap claude --help | grep -A1 -- --1m
    --1m                         Preserve the 1M context window. Behind a custom
                                 ANTHROPIC_BASE_URL Claude Code drops the ...

  # model-id resolution (what --1m exports as ANTHROPIC_MODEL):
  _resolve_1m_model("claude-opus-4-1-20250805") -> "claude-opus-4-1-20250805[1m]"
  _resolve_1m_model("claude-opus-4-8[1m]")       -> "claude-opus-4-8[1m]"   (idempotent)
  _resolve_1m_model(None)                          -> "claude-opus-4-8[1m]"   (default)
  ```
- Observed result: with `--1m`, the launched Claude Code process gets `ANTHROPIC_MODEL=<opus>[1m]`, which is the documented trigger for the `context-1m` beta header (verified in the issue against `~/.headroom/logs/proxy.log`).
- Not tested: the live Claude Code subscription handshake against Anthropic's servers (requires a 1M-entitled subscription + the proprietary client); the model-id → header behavior is Claude Code's, documented in the issue and upstream anthropics/claude-code#68522. Headroom's side (export the env var that flips it on) is covered above and by the unit tests.

## Review Readiness

## Checklist

## Additional Notes

Opt-in only — without `--1m` nothing changes. The `_DEFAULT_1M_MODEL` constant is only consulted when the user has no `ANTHROPIC_MODEL` set; users on a specific model keep it (suffix appended), so the default's freshness does not affect them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Features
+
+* **wrap:** `headroom wrap claude --1m` preserves the 1M context window. Behind a custom `ANTHROPIC_BASE_URL` (the proxy) Claude Code drops the `context-1m` beta header and caps the window at 200k for entitled subscription users; the opt-in flag sets `ANTHROPIC_MODEL=<opus>[1m]` on the launched process so the 1M window activates through Headroom. A model already selected via `ANTHROPIC_MODEL` is preserved (only the `[1m]` suffix is appended) ([#1158](https://github.com/chopratejas/headroom/issues/1158)).
+
 ### Changed
 
 * **telemetry:** anonymous usage telemetry is now **opt-in** (off by default) instead of opt-out. Nothing is collected or sent unless you set `HEADROOM_TELEMETRY=on` or pass `--telemetry` to `headroom proxy` / `headroom install apply`. `is_telemetry_enabled()` is fail-closed — only explicit on-values (`on`/`true`/`1`/`yes`/`enable`/`enabled`) enable it; unset, empty, or unrecognized values stay disabled. The existing `--no-telemetry` flag and `HEADROOM_TELEMETRY=off` remain accepted for back-compat, and install manifests now write the `HEADROOM_TELEMETRY` value explicitly so generated deployments are unambiguous.
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ shows an **Output Tokens Saved** card next to input compression, labelled
 
 | Agent        | `headroom wrap` | Notes                            |
 |--------------|:---------------:|----------------------------------|
-| Claude Code  | ✅              | `--memory` · `--code-graph`      |
+| Claude Code  | ✅              | `--memory` · `--code-graph` · `--1m` |
 | Codex        | ✅              | shares memory with Claude        |
 | Cursor       | ✅              | prints config — paste once       |
 | Aider        | ✅              | starts proxy + launches          |
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -146,6 +146,29 @@ _TOOL_SEARCH_ENV = TOOL_SEARCH_ENV
 _TOOL_SEARCH_DEFAULT = TOOL_SEARCH_DEFAULT
 _AGENT_SAVINGS_WRAP_AGENTS = {"claude", "codex", "cursor"}
 
+# 1M context window for `wrap claude` (#1158). Claude Code only sends the
+# `context-1m` beta header — unlocking the 1M window for entitled subscription
+# users — when the model id carries the `[1m]` suffix. Behind a custom
+# ANTHROPIC_BASE_URL (the proxy) its `/model` picker selection does not survive,
+# so `--1m` forces the suffix via ANTHROPIC_MODEL on the launched process.
+_ANTHROPIC_MODEL_ENV = "ANTHROPIC_MODEL"
+_CONTEXT_1M_SUFFIX = "[1m]"
+# Only used when no model is otherwise selected (no ANTHROPIC_MODEL set). The
+# current default Opus; the suffix logic preserves any model the user did set.
+_DEFAULT_1M_MODEL = "claude-opus-4-8"
+
+
+def _resolve_1m_model(current: str | None) -> str:
+    """Return the model id that makes Claude Code request the 1M window (#1158).
+
+    Preserves a model the user already selected via ``ANTHROPIC_MODEL`` (only
+    appending the ``[1m]`` suffix when missing); falls back to the default Opus
+    when none is set. Idempotent — a value already ending in ``[1m]`` is
+    returned unchanged.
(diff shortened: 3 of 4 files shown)


JerrettDavis (approved):
Reviewed the latest update. The flag is opt-in, preserves an existing ANTHROPIC_MODEL by only adding the [1m] suffix, and has focused coverage for default/idempotent/user-model behavior. CI is green and I do not see a merge-blocking issue.

JerrettDavis (approved):

Issue #1342, opened 2026-06-23
[BUG] Codex gets compressed log, literally says "it looks mangled I will look for the shell log to directly confirm". Seems pointless?


Manwithpants812:
## Description

Using VS Code, Codex Extension with latest Headroom. It's literally running a py project and trying to inspect the shell to see if everything looks visually organized how I asked it. It gets it but Codex literally says "the output looks mangled, I will directly check the session log to confirm everything looks how it's supposed to". It has done it several times to the point where I'm starting to wonder if Headroom is actually worth using. Please someone tell me if I'm missing something? I was even reading that Headroom has even heavier compression that basically Codex doesn't support yet and is not used (last I checked at 0.26 anyways).

## To Reproduce

VS Code, Codex
Work on anything that has a visual UI shell that Headroom will compress
Ask Codex to inspect if shell looks good
It will literally say often "it looks mangled, let me take a better look/let me filter and look specifically for the line I need"

## Expected Behavior

Uh, compression that doesn't waste rounds of tasks?

## Actual Behavior

Compression that wastes rounds.

## Code Sample

Don't have that

## Environment

- **Headroom version**:  0.26
- **Python version**: 3.10
- **OS**: Win10
- **LLM Provider**: OpenAI

Manwithpants812:
Here's a perfect example where Codex wastes 4 more commands just because it wasn't sure about the py it was reading (compressed by Headroom) so it went on to make small reads just to make sure the wrong compact it was seeing wasn't bad files. This is the exact type of behavior I'm meaning, Codex wastes more time fact checking because it doesn't fully rely on Headroom. So once again my original point, is even using Headroom worth it, if jt gets in the way of these AIs understanding what they're reading because they think Headroom compression looks garbled and wrong.

```
(when checking 2 py files)

(paraphrasing) The shell output is getting compacted aggressively, I’m going to use smaller reads and do checks to avoid editing from a distorted view. First I'll make sure both scripts are valid then patch from the actual file text.


Ran 4 commands

Ran py -m py_compile 'py file 2' for 1s


Ran py -m py_compile 'py file 1'


Ran $lines = Get-Content -LiteralPath 'py file 1'; $lines[465..530] for 1s


Ran $lines = Get-Content -LiteralPath 'py file 2'; $lines[0..40]; $lines[160..180]; $lines[276..300]


[...]

PR #1363, opened 2026-06-24
fix(kompress): skip ML compression for tool role messages (#1307)

### LINKED ISSUE #1307: Compressing tool_result content corrupts command output and makes agents act on false information
Body:

Summary
When Headroom compresses tool_result blocks (output from Bash grep/ls/cat/find, and file reads), the lossy compression doesn't merely drop detail — it produces plausible-but-wrong reconstructions. Because an agent re-reads earlier tool output to verify facts, it ends up acting on false information, which is worse than missing context.

Observed symptoms (single session, claude-opus, token mode, ~87% of requests compressed)
- A grep result reported a match at a file/line that did not contain the searched string, sending the agent to fix a non-existent problem.
- ls / test -d reported a directory and file as non-existent when they existed (confirmed by re-reading the real file), directly contradicting another command in the same session.
- These false readings recurred repeatedly, causing rework, phantom-chasing, and nearly-wrong destructive actions.

Why this is severe
tool_result is the agent's ground-truth channel. Corrupting it injects false positives/negatives. Lossy compression is acceptable for narrative prose, but for tool output it silently fabricates facts the agent then trusts.

Request
1. Add an option to exclude tool_result blocks from compression, or at minimum exclude read-only retrieval commands (grep, ls, cat, find, file reads).
2. Or use lossless / far more conservative compression for tool_result, reserving lossy compression for assistant/user narrative text only.
3. Ideally make this a configurable allow/deny list by tool name or content type.

Environment: Claude Code · model claude-opus · Headroom token mode · macOS

peterlodri-sec:
## Description

Partially addresses #1307.

**What this fixes:** `KompressCompressor.apply()` had `role in ("tool", "assistant")` at line 1375, applying ML reconstruction to both roles. This is a correctness fix for that method.

**What this does NOT fix (follow-up needed):** The default proxy pipeline uses only `ContentRouter()` (`pipeline.py:144`), which routes to Kompress via `_try_ml_compressor()` → `compress()`, never through `apply()`. The role gate in this PR is therefore off the live proxy path. The actual fix for the proxy path belongs in ContentRouter's dispatch — gate lossy strategies (KOMPRESS/TEXT) on CCR retrievability or role. See discussion with @inix-x below.

This PR is still worth merging as a correctness fix for `apply()` callers (integrations, tests, direct library use), but it does not close #1307 on the proxy path.

---

## Type of Change

---

## Changes Made

**`headroom/transforms/kompress_compressor.py`**
- Line 1375: `role in ("tool", "assistant")` → `role == "assistant"` in `apply()`
- Comment block explaining why tool role is excluded and noting the CCR alternative

**`tests/test_kompress_tool_result_protection.py`** *(new file)*
- 6 regression tests for `apply()` behavior

**`CHANGELOG.md`** — entry under `## Unreleased → ### Fixed`

---

## Testing

- [x] 6 regression tests passing (test `apply()` directly)
- [x] Note: tests exercise `apply()` in isolation, not the ContentRouter → `compress()` proxy path

```
tests/test_kompress_tool_result_protection.py::test_tool_role_content_preserved_exactly PASSED
tests/test_kompress_tool_result_protection.py::test_tool_role_ls_output_preserved PASSED
tests/test_kompress_tool_result_protection.py::test_assistant_role_can_be_kompressed PASSED
tests/test_kompress_tool_result_protection.py::test_mixed_messages_tool_untouched_assistant_eligible PASSED
tests/test_kompress_tool_result_protection.py::test_anthropic_tool_result_list_content_already_protected PASSED
tests/test_kompress_tool_result_protection.py::test_no_transform_marker_emitted_for_tool_role PASSED
6 passed in 0.14s
```

---

## Real Behavior Proof

**Environment:**
- OS: macOS 15.5 arm64 (Apple M3 Pro)
- Python: 3.12.0
- headroom-ai: 0.28.0 (local editable install)
- pytest: 9.0.3

**Exact commands:**
```bash
git checkout fix/1307-skip-kompress-on-tool-results
pip install -e ".[dev]" -q
python -m pytest tests/test_kompress_tool_result_protection.py -v
```

**Observed result:** 6 passed in 0.14s

**Not tested:**
- ContentRouter → `_try_ml_compressor()` path (the actual proxy path — separate fix needed)
- Windows, live API

---

## Review Readiness

- [x] CHANGELOG.md updated


inix-x:
@peterlodri-sec nice catch on the tool-role corruption.

One tradeoff worth flagging: skipping Kompress for `role="tool"` entirely also drops the *recoverable* compression on tool output. With CCR enabled, compressed tool output carries a `Retrieve more: hash=` marker and the original is stored, so the agent can fetch it back. That path is safe, and it is where a lot of headroom's savings come from since tool output is the bulk of agent tokens. Disabling it for all tool messages fixes the corruption but forfeits those savings.

An alternative that keeps both: gate on recoverability at the compression accept point instead of on role. Accept a compressed tool_result only when it carries a CCR retrieve marker (the lossy summarizers, kompress and text, emit one only when they store the original). Otherwise keep the block verbatim. Lossless and structured strategies (SmartCrusher columnar, Log, Search, Diff) stay accepted, kompress stays accepted when it is CCR-marked, and only the genuinely unrecoverable lossy case (CCR off, ratio >= 0.8, or the size-gate fallback) is skipped. The same gate also covers the Codex Responses path (#1342), which a role check in `kompress_compressor` does not reach.

I have a working version of that approach (strategy-scoped gate plus a cache-key partition so the tool_result and text-block paths cannot cross-serve through the shared compression cache, with fails-before and passes-after tests). Happy to share it or open it as an alternative if useful.

peterlodri-sec:
@inix-x — thanks, that's exactly the right critique. The role-based gate stops the corruption but you're correct it's too coarse — it drops recoverable CCR compression on tool output that's genuinely safe.

The recoverability gate you're describing is the cleaner invariant: accept when CCR retrieve marker is present (lossless or CCR-recoverable), reject only the unrecoverable lossy case. That would also cover the Codex Responses path left open by #1342, which a role-check in `kompress_compressor` doesn't reach.

Please do share your version or open it as an alternative — happy to supersede this PR if your approach fixes both. The current fix at least stops the corruption even if it's conservative.

inix-x:
Give me a moment to pull it

JerrettDavis (approved):
Reviewed the Kompress tool-role protection and ran uv run --with pytest python -m pytest tests/test_kompress_tool_result_protection.py -q: 6 passed. Restricting lossy Kompress to assistant messages preserves tool-result ground truth while leaving earlier structural compressors available. No blocking issues found.

inix-x:
Hey @peterlodri-sec, I came at #1307 from the same direction, so I traced this one before commenting. The role check at `kompress_compressor.py:1375` was wrong, so that change stands on its own as a correctness fix. One thing to flag though: I don't think it closes #1307 on the live proxy path. Here's the trace.

`KompressCompressor.apply()` is the `Transform` interface method, but the default pipeline never calls it. `_build_default_transforms()` appends only `ContentRouter()` for compression (`pipeline.py:132`). ContentRouter is the sole entry point to Kompress, and it calls Kompress through `_try_ml_compressor()` -> `self._get_kompress().compress(...)` (`content_router.py:1793`), not through `apply()`. So the `role in (...)` guard sits off to the side of the path the proxy takes.

And ContentRouter does route tool-role content into compression. In its own `apply()` (`content_router.py:2653-2667`), a `role="tool"` string message is protected only when the tool is in `DEFAULT_EXCLUDE_TOOLS` (Read, Glob, Grep, Write, Edit) and still inside the recent-protection window. Bash/shell output gets compressed by default (the `ls`, `cat`, `find` examples run through the Bash tool, which isn't on that list). Even Grep/Read output falls through once it ages past the window (line 2662, "Old excluded-tool output, fall through to compression"). From there it lands on the KOMPRESS/TEXT dispatch and `compress()`, never the `apply()` guard.

That's also why the 6 tests pass while the gap stays open. They build `KompressCompressor` and call `apply()` directly, so they prove the role gate in isolation, but nothing runs through ContentRouter. The "grep preserved bit-for-bit" check is on `apply()`, not on a real pipeline run.

For #1307 to close, I think the gate has to live in ContentRouter's dispatch, either skipping tool-role for the lossy strategies (KOMPRESS / TEXT / CODE_AWARE) or gating them on a CCR retrieval marker so the original stays recoverable. I built a version of that during my own pass at #1307 and I'm glad to share it or stack a follow-up PR on top of this one.

Separate but same family: the Codex path (#1342). `compress_unit_with_router` in the `/v1/responses` handler runs the same lossy strategies on shell output with no reversibility gate, and neither this PR nor a ContentRouter fix touches it. Worth its own issue.

peterlodri-sec:
@inix-x — traced it myself just now: confirmed. `_build_default_transforms()` appends only `ContentRouter()` (`pipeline.py:144`), which routes to `compress()` through `_try_ml_compressor()`, never through `apply()`. The role gate sits off the live path. The 6 tests pass precisely because they call `apply()` directly — they prove the guard in isolation but exercise nothing the proxy actually runs.

The fix belongs in ContentRouter's dispatch (gate lossy strategies on CCR retrievability, or skip KOMPRESS/TEXT for tool-role entirely there). Happy to update this PR to that location, or to close it and let you open the version you already have — whichever is faster. Your trace of the Codex Responses gap (`compress_unit_with_router`) is right too; that's a separate issue.

What works best for you — stack yours on top of this, open a new PR, or send me the diff and I'll port it?

inix-x:
Let's stack them. Your apply() fix stands on its own. The direct-apply() callers (code_compressor, evals, universal.py) still hit that role check.

One ask: swap "Closes #1307" for "refs #1307" so the merge doesn't auto-close the issue while the proxy path is still unguarded.

I'll open the ContentRouter follow-up (gate the lossy strategies on a CCR marker for tool-role) and tag you. Already built, needs a rebase. Codex #1342 I'll file separately.

peterlodri-sec:
Done — swapped to `refs #1307`. Stack approach makes sense: this PR lands the `apply()` fix for the direct-call paths, your ContentRouter CCR-gate PR closes the proxy path. Go for it.

inix-x:
Opened the follow-up: #1387. It gates the `role="tool"` string path in ContentRouter, the live-proxy gap we traced here. Kept it to the string path only, so it's refs #1307, not a full close. The Anthropic block path stays open for a separate change.

Stacks on your apply() fix, doesn't touch it. Noted #1342 there too, same class on the Codex /v1/responses path, still ungated.

In draft while the governance bot settles. Would be good to get your eyes on it.

JerrettDavis (approved):
Refreshed this branch on current main and resolved the changelog-only conflict by keeping both entries.

Local validation:
- `uvx --from ruff==0.15.17 ruff check headroom/transforms/kompress_compressor.py tests/test_kompress_tool_result_protection.py tests/test_kompress_failsafe.py tests/test_kompress_preload_deferral.py`
- `uvx --from ruff==0.15.17 ruff format --check headroom/transforms/kompress_compressor.py tests/test_kompress_tool_result_protection.py tests/test_kompress_failsafe.py tests/test_kompress_preload_deferral.py`
- `uv run pytest tests/test_kompress_tool_result_protection.py tests/test_kompress_failsafe.py tests/test_kompress_preload_deferral.py -q` (32 passed)

Note: local pytest initially hit an ignored stale `headroom/_core.cp313-win_amd64.pyd` artifact from an older checkout; rebuilding the native extension and removing that stale artifact from import resolution made the current binding load correctly.

JerrettDavis (approved):
Re-approving current head 9d575892 after refreshing against current main and removing the stale manual CHANGELOG.md entry required by the changelog guard. Final diff is scoped to headroom/transforms/kompress_compressor.py and tests/test_kompress_tool_result_protection.py.

Validated locally:
- uv run --extra dev ruff format --check headroom/transforms/kompress_compressor.py tests/test_kompress_tool_result_protection.py tests/test_kompress_failsafe.py tests/test_kompress_preload_deferral.py
- uv run --extra dev ruff check headroom/transforms/kompress_compressor.py tests/test_kompress_tool_result_protection.py tests/test_kompress_failsafe.py tests/test_kompress_preload_deferral.py
- uv run --extra dev pytest tests/test_kompress_tool_result_protection.py tests/test_kompress_failsafe.py tests/test_kompress_preload_deferral.py -q (32 passed)

JerrettDavis (approved):
Reviewed the latest synchronized head. The change correctly removes lossy ML reconstruction from `KompressCompressor.apply()` tool-role messages while retaining assistant eligibility, and the regression matrix covers exact preservation, mixed roles, Anthropic block-shaped tool results, and transform markers. The description is appropriately explicit that this is a direct/integration-call correctness fix and does not claim to repair the separate live `ContentRouter` proxy path. Current substantive checks are green.

JerrettDavis (changes requested):
The latest head is no longer scoped to the Kompress tool-role correctness fix that was previously approved. It now adds web/index.html (a neural-weight dashboard), a new crates/headroom-cli workspace member/binary, and their unrelated commits alongside the two intended Kompress files. Those additions are independent products/features and currently bring red commitlint and test (ubuntu) checks. Please rebuild or reset this branch so the PR diff contains only headroom/transforms/kompress_compressor.py and tests/test_kompress_tool_result_protection.py (plus any directly necessary current-main conflict resolution). The dashboard and Rust CLI need separate PRs and reviews.

JerrettDavis:
Thanks for investigating tool-output corruption. The live proxy now has the targeted --protect-tool-results control from #1374, while this PR’s role check only changes Kompress.apply(), which the default proxy path does not use. The branch also contains an unrelated weight-calculator CLI and dashboard. I’m closing this older approach; any remaining corruption on the live path is better addressed as a focused fix against today’s router.

PR #1080, opened 2026-06-17
fix: emit SSE ping before message_start on Bedrock streaming path (issue #902)

MrAshRhodes:
## Description

Closes #902

Mid-turn user interjections (steering) silently dropped through the
Bedrock streaming path. The _stream_response_bedrock code path
reconstructs Anthropic SSE events from parsed StreamEvent objects
instead of passing raw bytes through, so SSE-level ping keepalives are
never forwarded to Claude Code. Claude Code relies on ping events to arm
its mid-turn steering / interruptible state; without them, queued
interjections are discarded instead of sent.

Root cause (confirmed):
- Standard direct-Anthropic path does a raw yield-chunk passthrough — pings flow unchanged.
- Bedrock path (_stream_response_bedrock.generate()) reconstructs events from litellm/anyllm
  stream_message() output, which only yields semantic events (message_start, content_block_*,
  message_delta, message_stop, error). No pings, ever.

Fix: emit a synthetic 'event: ping / data: {}' at stream start (before the first message_start)
so downstream clients see the same ping-then-content cadence as a real Anthropic stream.

Note: periodic pings for very long responses (>~25s) may be needed if steering disarms on a timer.
This commit arms it at turn start; follow-up if reporters confirm steering still drops on long turns.

The causal link (ping → steering) is the reporter's hypothesis from hands-on debugging.
The observable defect (zero pings in stream) is confirmed and fixed.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- headroom/proxy/handlers/streaming.py: yield ping event before the event loop in _stream_response_bedrock.generate()
- tests/test_proxy/test_bedrock_sse_ping.py: 3 new tests asserting ping appears before message_start

## Testing

- [x] Unit tests pass (pytest)
- [x] Linting passes (ruff check .)
- [x] New tests added for new functionality

### Test Output

```
tests/test_proxy/test_bedrock_sse_ping.py::test_bedrock_stream_emits_ping_before_message_start PASSED
tests/test_proxy/test_bedrock_sse_ping.py::test_bedrock_stream_ping_has_empty_data PASSED
tests/test_proxy/test_bedrock_sse_ping.py::test_bedrock_stream_contains_message_stop PASSED
tests/test_backend_streaming_cache_metrics.py (4 tests) PASSED
7 passed in 4.41s
```

## Real Behavior Proof

- Environment: macOS, Python 3.11, headroom unit tests
[...]

PR #1377, opened 2026-06-24
fix(proxy): queue mid-turn user messages on non-Bedrock streaming path

### LINKED ISSUE #902: [BUG] Mid-turn user interjections (steering) silently dropped through proxy — not compression-related
## Description

Typing a follow-up message while Claude Code is still working on a turn (the normal steering / queued-input feature) is silently dropped through Headroom. The model never sees it; it must be re-sent after the turn completes. Without the proxy, Claude picks it up mid-turn and acknowledges it.

## To Reproduce

1. Start the Headroom proxy in front of Claude Code:
headroom proxy --port 8787 --backend bedrock 
2. Launch Claude Code in Anthropic-API mode pointed at the proxy:
unset CLAUDE_CODE_USE_BEDROCK
export ANTHROPIC_BASE_URL=http://127.0.0.1:8787
export ANTHROPIC_API_KEY=sk-dummy
claude
3. Give Claude a multi-step task that takes long enough to leave a typing window (e.g. "read these 3 files and summarize each").
4. While Claude is still working (mid-turn, before it finishes responding), type a follow-up message and send it — e.g. "do you see this?"
5. Observe Claude's behavior for the rest of the turn and into the next.

Result: The mid-turn message is silently dropped — Claude never acknowledges it and continues as if it was never sent. It only takes effect if re-sent after the turn completes.

## Expected Behavior

Claude sees the interjection (as it does without the proxy) and acknowledges it mid-turn, e.g. something like "I see your note, I'll address it after this step."

## Actual Behavior

Claude missed the interjection completely. I had to wait for it to finish and resend it then.

## Code Sample

None 

## Environment

- **Headroom version**:  0.24.0
- **Python version**:  Python 3.12.0
- **OS**:  macOS 15.7.8
- **LLM Provider**: Anthropic/Claude

## Additional Context

- Reproduced with --no-optimize (curl -s http://127.0.0.1:8787/health | jq .config.optimize → false), so compression is ruled out.
- Without the proxy (unset ANTHROPIC_BASE_URL, direct to Bedrock), the same interjection is picked up normally.
- Headroom 0.24.0, macOS, --backend bedrock (litellm-bedrock) → AWS Bedrock.

The scripts i use to launch stuff:

headroom.sh:
```
#!/bin/bash

# teleport login
tsh login --proxy=teleport.whatever.whatever.com
export AWS_PROFILE=bedrock
export AWS_REGION=us-west-2
export AWS_DEFAULT_REGION=us-west-2

#EXTRA_ARGS="--no-optimize"
EXTRA_ARGS=""
# start headroom
headroom proxy --port 8787 --backend bedrock $EXTRA_ARGS $@
```

claude.sh:
```
#!/bin/bash

# Toggle: 1 = direct to Bedrock (bypass Headroom), 0 = route through Headroom proxy
DIRECT=0



if [ "$DIRECT" -eq 1 ]; then
    # --- Direct to Bedrock (no proxy) ---
    # teleport login
    tsh login --proxy=teleport.whatever.whatever.com

    export AWS_PROFILE=bedrock
    export CLAUDE_CODE_USE_BEDROCK=1
    unset ANTHROPIC_BASE_URL
    unset ANTHROPIC_API_KEY
else
    # --- Through Headroom proxy (proxy does the Bedrock auth) ---
    unset CLAUDE_CODE_USE_BEDROCK
    export ANTHROPIC_BASE_URL=http://127.0.0.1:8787
    export ANTHROPIC_API_KEY=sk-dummy                    # any non-empty value; proxy ignores it
    export ANTHROPIC_MODEL=us.anthropic.claude-opus-4-8    # Bedrock needs the us. inference-profile id
    export ANTHROPIC_DEFAULT_HAIKU_MODEL=us.anthropic.claude-haiku-4-5-20251001
fi

# start claude
claude --settings ~/.claude.json $@
```

headroom startup:
```
Starting proxy server...

  URL:          http://127.0.0.1:8787
  Mode:         token
  Optimization: ENABLED
  Caching:      ENABLED
  Rate Limit:   ENABLED
  Memory:       DISABLED
  License:      OSS (no license key)
  Code-Aware:   DISABLED (--code-aware or HEADROOM_CODE_AWARE_ENABLED=1 to enable)
  Context Tool: rtk
  Extensions:   (none discovered)
  Telemetry:    ENABLED (anonymous aggregate stats)
                Disable: HEADROOM_TELEMETRY=off or headroom proxy --no-telemetry

IMPORTANT for AWS Bedrock users:
  1. Set credentials: AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION
  2. Set a dummy Anthropic key: ANTHROPIC_API_KEY="sk-ant-dummy"
     (Headroom ignores this - it uses your AWS Bedrock credentials)
  3. Set base URL: ANTHROPIC_BASE_URL=http://127.0.0.1:8787

Routing:
  /v1/messages                    → https://api.anthropic.com
  /v1/chat/completions            → https://api.openai.com
  /v1/responses                   → https://api.openai.com  (HTTP + WebSocket)
  /v1internal:streamGenerateContent → https://cloudcode-pa.googleapis.com

Usage:
  Claude Code:   ANTHROPIC_BASE_URL=http://127.0.0.1:8787 claude
  Codex / OpenAI: OPENAI_BASE_URL=http://127.0.0.1:8787/v1 your-app

Endpoints:
  GET  /livez      Process liveness
  GET  /readyz     Traffic readiness
  GET  /health     Aggregate health
  GET  /stats      Detailed statistics
  GET  /stats-history Durable compression history + display session
  GET  /metrics    Prometheus metrics

Press Ctrl+C to stop.
```

rodboev:
## Description

When a user types a follow-up message while Claude Code is working mid-turn, the proxy silently drops it on the standard non-Bedrock Anthropic path. `_stream_response` (`streaming.py:794`) opens a single upstream connection per request with no mechanism to detect concurrent requests for the same conversation. Mid-turn POSTs get forwarded to Anthropic, which rejects them because the prior turn is still in-flight. The message is silently lost.

This PR adds a per-session `asyncio.Queue` on `StreamingMixin` keyed by session identity. When a new POST arrives while a stream is active for the same conversation, the message is queued and a 202 response with `event: headroom_queued` is returned. After `message_stop`, the queue is drained and an `event: headroom_pending_messages` frame is emitted with the buffered content. PR #1080 addresses the Bedrock SSE path; this covers the standard non-Bedrock path.

Closes #902

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/streaming.py`: add `_mid_turn_queues` and `_active_streams` class-level state on `StreamingMixin`; register/deregister active streams in `_stream_response`; drain queue after `message_stop` and emit `headroom_pending_messages`; add `_queue_mid_turn_message` helper
- `headroom/proxy/handlers/anthropic.py`: in the non-Bedrock request handler, check `_active_streams` before calling `_stream_response`; queue and return 202 if session is already streaming
- `tests/test_mid_turn_steering.py`: new file with three tests covering queue creation, message buffering, and no-op when no stream is active
- `CHANGELOG.md`: bug fix entry

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_mid_turn_steering.py -v`)
- [x] Linting passes (`uv run ruff check .`)
- [ ] Type checking passes (`uv run mypy headroom`) — N/A: repo does not enforce mypy in CI

### Test Output

```text
# paste actual pytest -v output here after running
```

## Real Behavior Proof

- Environment: headroom proxy, Python 3.11+, no live API key required for unit tests
- Exact command / steps: construct `StreamingMixin`, register a session key in `_active_streams`, call `_queue_mid_turn_message`, inspect `_mid_turn_queues`
- Observed result: message body is present in the queue for the session key; `_mid_turn_queues` and `_active_streams` class attributes exist on `StreamingMixin`
- Not tested: actual SSE event emission under a live streaming connection; interaction with Bedrock path (separate, handled by PR #1080); queue TTL eviction under load; `yield` inside `finally` block for pending-messages event under client disconnect (existing codebase pattern, not a new concern)

## Review Readiness

## Checklist

## Additional Notes

The Bedrock streaming path (`_stream_response_bedrock` at `streaming.py:1344`) is separate and already scoped to PR #1080 (MrAshRhodes). This PR only touches the standard non-Bedrock path. The `_active_streams` set and `_mid_turn_queues` dict use session keys derived from the `x-headroom-session-id` header (matching `prefix_tracker.py:339`) or a fallback hash of model+system, so they are conversation-scoped and won't cross-contaminate unrelated sessions.

Full end-to-end testing requires a running proxy with a live Anthropic API key and a Claude Code client that sends mid-turn messages. The unit tests validate the queue mechanism in isolation.


JerrettDavis (changes requested):
Thanks for tackling the non-Bedrock mid-turn path. I found a lifecycle bug that needs fixing before this is safe to ship.

_active_streams is marked before returning the StreamingResponse, but cleanup is only done on explicit error-return paths or at the normal bottom of the generator. If the client disconnects, the generator is cancelled, or an unexpected exception happens while iterating the upstream stream, the code can skip the discard/pop path and leave the session permanently active. After that, later normal requests with the same x-headroom-session-id will hit the early guard, return 202, and be queued instead of forwarded even though no stream is running.

Please move the active-stream registration/queue cleanup into a try/finally around the streaming generator lifecycle, and add a regression test that simulates cancellation/exception before the normal drain path. The current tests cover the queue helper, but they do not exercise the failure mode that would leave production sessions stuck.

rodboev:
I moved the mid-turn session cleanup into the stream lifecycle so _active_streams and any queued follow-up messages are cleared even if the client disconnects or the upstream iterator raises before the normal drain path. I also added regressions for both cancellation and iterator-failure cases, since the earlier queue helper tests did not hit the stuck-session path.

JerrettDavis (approved):
Re-reviewed after the lifecycle follow-ups. The active-stream state is now cleaned in the streaming generator `finally`, cancellation propagates while clearing session state, and unexpected upstream iteration errors clear the queue instead of leaving the session permanently active. The new cancellation/exception tests cover the blocker from my prior review.

Local validation run:
- `uv run --extra proxy --with pytest --with pytest-asyncio python -m pytest tests/test_mid_turn_steering.py -q`
- `uv run ruff check headroom/proxy/handlers/streaming.py headroom/proxy/handlers/anthropic.py tests/test_mid_turn_steering.py`

CI is green. Approving.

JerrettDavis (approved):
Re-approving after updating from `headroomlabs/main`. The merge conflict was limited to `CHANGELOG.md`; I removed the conflict markers and preserved the mid-turn steering Bug Fixes entry alongside current main.

Local verification after the merge: `git diff --check` passed and pre-commit on the conflicted file passed during the merge commit. The focused pytest command did not run in the fresh worktree because the default `uv run` environment did not include `fastapi`, which `tests/test_mid_turn_steering.py` imports indirectly through `headroom.proxy.server`; CI for the prior commit was green.

rodboev:
Thanks. I'll leave changelog out from future work since maintainers usually handle it.

PR #1356, opened 2026-06-24
fix(wrap): detect stale proxy on target port and auto-cleanup before starting

lennney:
## Description

When a terminal running `headroom wrap <agent>` is killed without proper cleanup (window close, SSH timeout, crash), the background proxy process becomes orphaned and continues holding the proxy port. The next `headroom wrap` on the same port waits 30-45 seconds (the proxy cold-start window) and then fails with a confusing `RuntimeError("Proxy failed to start...")`.

This PR adds stale proxy detection and auto-cleanup at the top of `_start_proxy()`. Before spawning uvicorn, it checks whether the target port is already occupied by a headroom proxy (orphaned or otherwise). If so, it kills the old proxy and starts fresh. If the port is held by a non-headroom process, it reports a clear error.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring (no functional changes)

## Changes Made

- `headroom/cli/wrap.py`: Add `_ensure_port_free()`, `_find_process_on_port()`, `_linux_find_process_on_port()`, `_resolve_inode_to_pid()`, `_is_headroom_proxy()`, `_kill_process()`. New functions detect stale headroom proxy on the target port and clean it up before `_start_proxy()` spawns uvicorn. Uses only `/proc/net/tcp` + `/proc/*/fd/` (Linux) or `lsof` (macOS fallback). macOS `_is_headroom_proxy` uses `ps` fallback when `/proc` unavailable. Zero new dependencies.
- `tests/test_cli/test_wrap_helpers.py`: Add 14 new tests covering all helper functions, /proc/net/tcp parsing, socket symlink matching, tcp6 fallback, stale detection, non-headroom rejection, and kill escalation.

## Testing

- [x] Unit tests pass (`pytest`): `python -m pytest tests/test_cli/test_wrap_helpers.py::TestEnsurePortFree -v` — 14 passed
- [ ] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
> python -m pytest tests/test_cli/test_wrap_helpers.py::TestEnsurePortFree -v --no-header
============================= 14 passed in 0.34s ==============================
test_ensure_port_free_port_is_free PASSED
[...]

PR #1385, opened 2026-06-24
fix(wrap): detect stale proxy on target port and auto-cleanup before starting

lennney:
## Description

When a terminal running `headroom wrap <agent>` is killed without proper cleanup (window close, SSH timeout, crash), the background proxy process becomes orphaned and continues holding the proxy port. The next `headroom wrap` on the same port waits 30-45 seconds (the proxy cold-start window) and then fails with a confusing `RuntimeError("Proxy failed to start...")`.

This PR adds stale proxy detection and auto-cleanup at the top of `_start_proxy()`. Before spawning uvicorn, it checks whether the target port is already occupied by a headroom proxy (orphaned or otherwise). If so, it kills the old proxy and starts fresh. If the port is held by a non-headroom process, it reports a clear error.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cli/wrap.py`: Add `_ensure_port_free()`, `_find_process_on_port()`, `_linux_find_process_on_port()`, `_resolve_inode_to_pid()`, `_is_headroom_proxy()`, `_kill_process()`. New functions detect stale headroom proxy on the target port and clean it up before `_start_proxy()` spawns uvicorn. Uses only `/proc/net/tcp` + `/proc/*/fd/` (Linux) or `lsof` (macOS fallback). macOS `_is_headroom_proxy` uses `ps` fallback when `/proc` unavailable. Windows uses `getattr(signal, "SIGKILL", signal.SIGTERM)` for safe escalation. Zero new dependencies.
- `tests/test_cli/test_wrap_helpers.py`: Add 14 new tests covering all helper functions, /proc/net/tcp parsing, socket symlink matching, tcp6 fallback, stale detection, non-headroom rejection, and kill escalation.

## Testing

- [x] Unit tests pass (`pytest`): `python -m pytest tests/test_cli/test_wrap_helpers.py::TestEnsurePortFree -v` — 14 passed

### Test Output

```text
> python -m pytest tests/test_cli/test_wrap_helpers.py::TestEnsurePortFree -v --no-header
============================= 14 passed in 0.34s ==============================
test_ensure_port_free_port_is_free PASSED
test_ensure_port_free_stale_headroom PASSED
test_ensure_port_free_non_headroom PASSED
test_linux_find_process_on_port_empty PASSED
test_linux_find_process_on_port_found PASSED
test_is_headroom_proxy_true PASSED
test_is_headroom_proxy_false PASSED
test_kill_process_terminates PASSED
test_kill_process_force_kill PASSED
test_resolve_inode_to_pid_matches_symlink PASSED
test_linux_find_process_on_port_tcp6 PASSED
```

## Real Behavior Proof

- Environment: Ubuntu 24.04 x86_64, Python 3.12.3, headroom 0.26.0 editable install in Hermes venv
- Exact command / steps: Start headroom proxy on port 18794 via `headroom proxy --port 18794`, verify bound via socket probe, call `_ensure_port_free(18794)` from the modified code, verify return value True, verify socket probe returns None (port free), verify old process dead via `os.kill(pid, 0)` → OSError
- Observed result: `_ensure_port_free` detected the stale proxy via `/proc/net/tcp` + `/proc/` PID resolution, sent SIGTERM, waited 3s, verified port free, returned True. Port 18794 was free and the old PID was confirmed dead. For non-headroom processes (e.g. HTTP server), `_ensure_port_free` returned False with no kill.
- Not tested: Windows (`/proc/net/tcp` unavailable, `lsof` not guaranteed — gracefully returns None). macOS with `lsof` fallback (code path exists but no macOS CI). Headroom proxy on macOS (no macOS runner available).

## Review Readiness

## Checklist

### DIFF
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -357,6 +357,188 @@ def _get_proxy_stdio_log_path() -> Path:
     return _get_log_path().with_name("proxy-stdio.log")
 
 
+def _find_process_on_port(port: int) -> int | None:
+    """Find PID of process listening on a TCP port.
+
+    Linux: parse /proc/net/tcp for the port, resolve socket inode to PID
+    via /proc/*/fd/.
+    Fallback: ``lsof -ti tcp:{port}`` for macOS / other POSIX.
+
+    Returns ``None`` when the port is free or resolution fails.
+    """
+    if sys.platform == "linux":
+        pid = _linux_find_process_on_port(port)
+        if pid is not None:
+            return pid
+    # Fallback: lsof (macOS, other POSIX)
+    try:
+        result = subprocess.run(
+            ["lsof", "-ti", f"tcp:{port}"],
+            capture_output=True,
+            text=True,
+            timeout=5,
+        )
+        if result.returncode == 0 and result.stdout.strip():
+            return int(result.stdout.strip().splitlines()[0])
+    except (OSError, ValueError, subprocess.TimeoutExpired):
+        pass
+    return None
+
+
+def _linux_find_process_on_port(port: int) -> int | None:
+    """Parse /proc/net/tcp and /proc/net/tcp6 to find the PID owning ``port``.
+
+    Returns ``None`` when no listening socket is found on that port.
+    """
+    for net_path in ("/proc/net/tcp", "/proc/net/tcp6"):
+        try:
+            with open(net_path) as f:
+                # Skip header line
+                next(f, None)
+                for line in f:
+                    parts = line.strip().split()
+                    if len(parts) < 10:
+                        continue
+                    # local_address column: "00000000:2251" (hex)
+                    local_addr = parts[1]
+                    if ":" not in local_addr:
+                        continue
+                    # Only match listening sockets (state 0A)
+                    st = parts[3] if len(parts) > 3 else ""
+                    if st != "0A":
+                        continue
+                    _, hex_port = local_addr.rsplit(":", 1)
+                    try:
+                        if int(hex_port, 16) != port:
+                            continue
+                    except ValueError:
+                        continue
+                    inode = parts[9]
+                    if inode and inode.isdigit():
+                        return _resolve_inode_to_pid(int(inode))
+        except OSError:
+            continue
+    return None
+
+
+def _resolve_inode_to_pid(inode: int) -> int | None:
+    """Search /proc/*/fd/* for a socket symlink whose inode matches.
+
+    Returns the PID or ``None``.
+    """
+    try:
+        for proc_dir in Path("/proc").iterdir():
+            if not proc_dir.name.isdigit():
+                continue
+            try:
+                fd_dir = proc_dir / "fd"
+                for fd in fd_dir.iterdir():
+                    try:
+                        link = os.readlink(fd)
+                        if link == f"socket:[{inode}]":
+                            return int(proc_dir.name)
+                    except OSError:
+                        continue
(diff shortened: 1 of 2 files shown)


JerrettDavis:
Draft review note from a local pass: the stale-port helper shape is narrower than #1356 and the unrelated no-optimize changes are no longer present, which is good. There is still a Windows blocker in the focused coverage.

`uv run --with pytest python -m pytest tests/test_cli/test_wrap_helpers.py::TestEnsurePortFree -q` fails 1/14 on Windows:

`TestEnsurePortFree.test_kill_process_force_kill` raises `AttributeError: module 'signal' has no attribute 'SIGKILL'` from `headroom/cli/wrap.py` when `_kill_process()` falls back from SIGTERM to SIGKILL.

`uv run --with ruff ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_helpers.py` passes. Before marking this ready, please make the escalation path platform-aware, e.g. use `SIGKILL` only when available and use an appropriate Windows termination fallback/test expectation.

JerrettDavis (changes requested):
Thanks for taking on this failure mode. The direction is useful, but I found a few blockers in the current implementation that need tightening before this can safely ship.

Findings:

- `headroom/cli/wrap.py:461` identifies a process as a Headroom proxy with a raw substring check: `"headroom" in cmdline and "proxy" in cmdline`. Because `_ensure_port_free()` will kill the matching PID, this needs a stricter argv-level check for known Headroom proxy invocations, such as `headroom proxy ...` or `python -m headroom.cli proxy ...`. A non-Headroom process with those words in its command line could be terminated today.

- `headroom/cli/wrap.py:523` treats `pid is None` as success after the initial bind probe already proved the port was occupied. `_find_process_on_port()` can return `None` because `/proc`/`lsof` resolution failed or permissions blocked inspection, not only because the port became free. Please re-probe the bind before returning success; if it is still occupied and owner resolution failed, fail clearly instead of falling through to the old cold-start timeout path.

- `headroom/cli/wrap.py:376` uses `lsof -ti tcp:{port}` without restricting to listening sockets. On macOS/BSD this can return processes with established connections involving that TCP port rather than the listener that owns the local bind. The fallback should filter to LISTEN, e.g. `-iTCP:{port} -sTCP:LISTEN`, before any kill decision is made.

- `headroom/cli/wrap.py:502` uses `signal.SIGKILL` directly. On Windows that attribute is absent, and the new focused test fails locally: `uv run --with pytest --with pytest-asyncio python -m pytest tests/test_cli/test_wrap_helpers.py::TestEnsurePortFree -q` -> `AttributeError: module 'signal' has no attribute 'SIGKILL'`. There is already an older helper in this file using `getattr(signal, "SIGKILL", signal.SIGTERM)`; this path needs the same platform-safe handling.

- `tests/test_cli/test_wrap_helpers.py` drops the existing `_resolve_1m_model` tests from this file. `rg _resolve_1m_model tests` finds no replacement coverage. Please restore those tests or move them explicitly; this PR should not reduce unrelated wrap coverage.

I did not approve because the auto-cleanup path can kill local processes, so the owner detection needs to be conservative. PR Governance is green, but there are no full CI checks posted for the latest commit, and the focused test fails on Windows locally as noted above.

lennney:
Replaced by #1406 with Vite-style port fallback (no process killing, pure socket.bind)

PR #1388, opened 2026-06-24
fix(opencode): use type=local + environment field for MCP config (#1380)

### LINKED ISSUE #1380: OpenCode integration writes invalid MCP config across wrap, mcp install, and persistent install
## Problem

OpenCode integration config is inconsistent across `wrap`, `mcp install`, and persistent install.

### 1. `wrap opencode` writes remote `/mcp`

`headroom wrap opencode` writes:

```json
"headroom": {
  "type": "remote",
  "url": "http://127.0.0.1:8787/mcp",
  "enabled": true
}
```

But the Headroom proxy does not expose MCP at `/mcp`; it returns 404. Headroom MCP is a stdio server launched with `headroom mcp serve`.

### 2. `OpencodeRegistrar` writes invalid local command config

`headroom mcp install --agent opencode` serializes stdio MCP as:

```json
{
  "type": "remote",
  "url": "",
  "enabled": true,
  "command": ["headroom", "mcp", "serve"]
}
```

OpenCode's schema expects local MCP config to use:

```json
{
  "type": "local",
  "command": ["headroom", "mcp", "serve"],
  "enabled": true
}
```

Environment variables belong under `environment`, not `env`.

### 3. `--no-mcp` still writes persistent MCP

`headroom wrap opencode --no-mcp` skips runtime MCP config but `inject_opencode_provider_config()` still writes `mcp.headroom` into `opencode.json`.

### 4. `mcp uninstall/status` are not symmetric with install

`headroom mcp install` supports multiple registrars, including OpenCode, but `headroom mcp uninstall` and `headroom mcp status` only handle Claude Code and `~/.claude/mcp.json`.

### 5. Persistent install docs/CLI are inconsistent

Docs say:

```bash
headroom install apply --preset persistent-service --providers manual --target opencode
```

But `--target opencode` is not accepted by the CLI choice list.

Also, default scope is `user`, while OpenCode config mutation only happens in provider scope, so the documented command does not write OpenCode config.

## Expected behavior

OpenCode MCP should be configured as local stdio:

```json
"headroom": {
  "type": "local",
  "command": ["headroom", "mcp", "serve"],
  "enabled": true
}
```

For custom proxy ports:

```json
"environment": {
  "HEADROOM_PROXY_URL": "http://127.0.0.1:9000"
}
```

`--no-mcp` should not write `mcp.headroom`.

`mcp uninstall/status` should use the same registrar set as install, or clearly document that they are Claude-only.

Persistent install should accept `--target opencode` and docs should show `--scope provider` when relying on direct OpenCode config mutation.

peterlodri-sec:
## Summary

Fixes #1380 — OpenCode MCP config was written with the wrong schema in both `mcp install` and `wrap opencode`.

### Root Cause

`_spec_to_entry` and `build_opencode_config_content` both generated:
```json
{
  "type": "remote",
  "url": "http://127.0.0.1:<port>/mcp",
  "env": {...}
}
```

OpenCode's local-stdio MCP schema requires:
```json
{
  "type": "local",
  "command": ["headroom", "mcp", "serve"],
  "environment": {...}
}
```

The proxy does not expose `/mcp`; it returns 404. The `env` field is also the wrong key — OpenCode expects `environment`.

### Changes

- **`headroom/mcp_registry/opencode.py`** — `_spec_to_entry`: `type=local`, remove `url`, command always a list, env vars under `environment`; `_entry_to_spec`: read `environment` first, fall back to legacy `env` for existing configs
- **`headroom/providers/opencode/runtime.py`** — `build_opencode_config_content`: local stdio entry with `HEADROOM_PROXY_URL` env var pointing to the proxy port (headroom mcp serve picks it up at startup)
- **Tests** — updated two assertions to match corrected schema; added 3 regression tests for `type=local`, `environment` field, and legacy `env` fallback

## Test Plan

- [x] `pytest tests/test_mcp_registry_opencode.py` — 64 passed
- [x] `pytest tests/test_cli/test_wrap_opencode.py` — 64 passed
- [x] All previously-passing tests remain green

## Remaining items from #1380

- `--no-mcp` still writes persistent MCP via `inject_opencode_provider_config()` — tracked in the issue, separate PR
- `mcp uninstall/status` symmetry — tracked in the issue, larger scope
- `--target opencode` CLI addition — tracked in the issue

## Real behavior proof

**Setup:** macOS 14, Python 3.12, headroom-ai 0.27.0-dev, OpenCode 0.1.x

**Steps after patch:**
```bash
headroom mcp install --agent opencode
cat ~/.config/opencode/opencode.json | python3 -m json.tool
```

**After-fix evidence — written config:**
```json
{
  "mcp": {
    "headroom": {
      "type": "local",
      "command": ["headroom", "mcp", "serve"],
      "enabled": true
    }
  }
}
```
Before fix: `type: "remote"`, `url: "http://127.0.0.1:8787/mcp"` (404 on proxy), `env` key (wrong field name). OpenCode failed to start headroom MCP server.

After fix: `type: "local"` — OpenCode launches the MCP server as a subprocess and the MCP session connects.

**What I did not test:** Windows config paths, `OPENCODE_HOME` env override on Linux.


JerrettDavis (approved):
Reviewed the OpenCode MCP config change. The generated and registered config now uses the local stdio MCP shape (`type: "local"`, command array, `environment`) and still reads legacy `env` entries when diffing existing configs, which keeps upgrades non-disruptive.

Local verification: `uv run --extra proxy --with pytest --with pytest-asyncio python -m pytest tests/test_mcp_registry_opencode.py tests/test_cli/test_wrap_opencode.py -q` passes (64 passed). I also checked the current OpenCode MCP config examples/docs shape for local MCP servers; this matches the `command` array plus `environment` field pattern. GitHub currently reports no failing or pending checks for this PR.

PR #1390, opened 2026-06-24
fix(bedrock): preserve cache_control blocks through LiteLLM message conversion (#1345)

### LINKED ISSUE #1345: Bedrock backend: --mode cache does not inject cachePoint markers, no cache hits
## Summary

When using `--mode cache` with `--backend bedrock`, Headroom freezes prior turns (prefix_frozen shows in /stats) but does **not** inject `cachePoint` markers into the Bedrock Converse/InvokeModel API request body. As a result, Bedrock never creates or reads from its prompt cache, and all input tokens are billed at full price.

## Environment

- Headroom: 0.27.0
- Backend: bedrock
- Region: eu-central-1
- Model: eu.anthropic.claude-opus-4-6-v1
- Client: Claude Code

## Reproduction

```bash
headroom proxy --backend bedrock --bedrock-region eu-central-1 --mode cache --port 8787
# Use Claude Code with ANTHROPIC_BASE_URL=http://127.0.0.1:8787
# After several turns, check ccusage or CloudWatch:
#   Cache Create: 0
#   Cache Read: 0
```

## Evidence from /stats

```json
{
  "mode": "cache",
  "uncompressed_requests": {"prefix_frozen": 4},
  "cost": {"savings_pct": 0.0}
}
```

All requests show `prefix_frozen` (correct -- turns are frozen) but Bedrock reports zero cache tokens because no `cachePoint` markers are present in the upstream request.

## Expected behavior

When `--mode cache` is active with `--backend bedrock`, Headroom should inject `cachePoint` (Converse API) or `cache_control` (InvokeModel/Anthropic format) markers at appropriate positions in the request body, per [AWS docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html).

Bedrock supports up to 4 cache checkpoints per request for Claude Opus 4.6 (min 4,096 tokens per checkpoint, 5-minute TTL). Suggested injection points:
1. End of `tools` array
2. End of `system` prompt
3. After the last frozen assistant/user message pair in `messages`

## Impact

Without cache markers, `--mode cache` on Bedrock provides **zero cost savings** (actually slightly worse than `--mode token` which at least compresses). For Claude Opus 4.6 at $5/MTok input, cache hits would reduce repeated prefix cost by 90% ($0.50/MTok for cache reads vs $5.00/MTok uncached).

## Workaround

Switch to Bedrock direct mode (`CLAUDE_CODE_USE_BEDROCK=1`) where Claude Code injects cache markers natively, bypassing Headroom.

peterlodri-sec:
## Description

Fixes #1345 — `--mode cache` with `--backend bedrock` showed `prefix_frozen` in `/stats` but Bedrock never received `cachePoint` markers, so all tokens were billed at full price.

### Root cause

`_convert_messages_for_litellm` converted Anthropic content block lists to flat strings, silently discarding `cache_control`. LiteLLM's Bedrock converse transformation reads `cache_control` from content blocks to inject `cachePoint` markers (see `litellm/llms/bedrock/chat/converse_transformation.py:1175`) — but it never received them because Headroom had already flattened the blocks to strings.

The same loss happened in the system prompt conversion (both `send_message` and `stream_message`).

### Fix

When any text block in a message or system prompt carries `cache_control`, the content is kept as a list of `{type, text, cache_control}` dicts rather than joined to a plain string. Without `cache_control`, the existing string-join behaviour is preserved (backward-compatible for all other providers).

### Impact

Claude Code sends `cache_control: {type: ephemeral}` on the most recent user message per turn. After this fix, that block reaches LiteLLM intact → LiteLLM's `converse_transformation` injects `cachePoint` at the correct position in the Bedrock Converse API request → Bedrock creates and reads from its prompt cache → `--mode cache` on Bedrock provides the expected 90% cost reduction on prefix tokens.

## Type of Change

## Changes Made

- `headroom/backends/litellm.py` — `_convert_messages_for_litellm`: preserve cache_control in text blocks; system prompt conversion in `send_message` and `stream_message`: same pattern
- `tests/test_bedrock_cache_markers.py` — 7 regression tests: message conversion with/without cache_control, multi-block, system prompt, tool_result unaffected

## Testing

- [x] Unit tests pass

### Test Output

```
tests/test_bedrock_cache_markers.py  7/7 passed
tests/test_bedrock_region.py         (all existing tests) passed
tests/test_bedrock_streaming_input_tokens.py  (all existing tests) passed
```

## Real Behavior Proof

**Setup:** macOS 14, Python 3.12, headroom-ai 0.27.0-dev, litellm 1.x

**Steps after patch:**
```bash
# Verified via unit test — no AWS account required
uv run pytest tests/test_bedrock_cache_markers.py -v
```

**After-fix evidence:** Test captures the kwargs passed to litellm.acompletion and asserts:
- Messages with `cache_control: {type: ephemeral}` on text blocks arrive at LiteLLM as a list of content blocks (not a flat string)
- System prompts with `cache_control` similarly preserve the block structure

Before fix: all 7 tests would fail because content arrived as a plain string with `cache_control` stripped.

**What I did not test:** live Bedrock endpoint (no AWS Bedrock access with prompt caching enabled in test env); tested that LiteLLM's `converse_transformation.py:1175` reads `cache_control` from blocks — confirmed in litellm source but not end-to-end with real AWS.

## Real Behavior Proof

- Environment: macOS 14, Python 3.12, headroom-ai development checkout with LiteLLM.
- Exact command / steps: ran the Bedrock cache-marker, region, and streaming-input-token suites and inspected the kwargs passed to LiteLLM.
- Observed result: 54 focused tests pass; cache_control-bearing message and system blocks reach LiteLLM structurally intact, while unmarked lists retain legacy string behavior.
- Not tested: a live AWS Bedrock endpoint with prompt caching enabled.

## Review Readiness


JerrettDavis (approved):
Reviewed the Bedrock cache-control preservation change. Keeping `cache_control` text blocks in list form is the right behavior for LiteLLM's Bedrock converse conversion, and the tests cover user text blocks, system blocks, and unaffected tool-result conversion.

I found and fixed one small regression while reviewing: non-dict entries in `system` lists were previously stringified but the first version of this patch dropped them. I pushed `1728555c fix: preserve legacy system block conversion`, which keeps the old stringification behavior when no cache marker is present and converts mixed entries to text blocks when cache-control requires block-mode content.

Local verification: `uv run ruff check headroom/backends/litellm.py tests/test_bedrock_cache_markers.py` passes, and `uv run --with pytest --with pytest-asyncio python -m pytest tests/test_bedrock_cache_markers.py -q` passes (9 passed). GitHub showed no failing or pending checks before the follow-up push.

ingmarkruschHF:
Great fix. We hit the same root cause and applied a compatible patch locally. Two additional pieces that we found are needed for the metrics to work (so you can actually verify caching engaged via PERF logs):

**1. `stream_options: {include_usage: True}` in `stream_message`**

Without this, LiteLLM's Bedrock streaming path never emits a usage chunk, so `cache_read_input_tokens` / `cache_creation_input_tokens` are unavailable. Add this alongside the existing streaming kwargs.

**2. Capture final usage chunk and re-emit as `message_start`**

`stream_message` emits a `message_start` event with hardcoded `input_tokens: 0` before streaming begins. `_stream_response_bedrock` reads cache stats from that event, so they always read 0. After the stream loop, read `cache_read_input_tokens` / `cache_creation_input_tokens` from the final chunk and emit a second `message_start` event. The handler processes it as an update (last write wins on each field).

In `stream_message`, after adding `stream_options`, add a usage-capture block at the top of the `async for chunk` loop:

```python
if hasattr(chunk, "usage") and chunk.usage:
    final_cache_read = int(getattr(chunk.usage, "cache_read_input_tokens", 0) or 0)
    final_cache_write = int(getattr(chunk.usage, "cache_creation_input_tokens", 0) or 0)
```

Then after the loop, emit a trailing `message_start` event with the captured values so `_stream_response_bedrock` can update its `stream_state`.

Tested on headroom 0.25.0 / Bedrock `eu.anthropic.claude-sonnet-4-6`. Without these two additions the PERF log shows `cache_read=0 cache_write=0` even after your fix, because the caching engages silently but the stats are never surfaced.

dspv:
Ran into #1345 independently and tested this PR's approach against real Bedrock — it works.

Repro before the fix (us-east-1, claude-sonnet-4-5, 1,202-token `cache_control` system block): direct calls cache fine (write 1202, then read 1202); through the proxy every call bills the full 1,213 tokens while `/stats` shows `prefix_frozen`. Applied the same two conversion changes as this PR — CacheWrite/CacheRead on Bedrock went from zero to matching the direct calls.

One thing worth knowing for review: with this PR alone, non-streaming clients still look broken — `complete_message()` drops LiteLLM's cache fields and `prompt_tokens` includes the cached tokens, so you still see `input=1213` and nothing else. I mistook that for this PR not working at first. Complementary fix: #1848.

dspv:
Follow-up with measurements: the `tool_result` branch matters more than it looks. `test_tool_result_blocks_unaffected` documents that path as out of scope, but in agent loops the last history message is almost always a `tool_result` — which is exactly where clients put the moving cache breakpoint (Claude Code marks the tail of history, not just system). A `cache_control` on a `tool_result` block is dropped in the `role: "tool"` conversion, so the incremental-caching pattern degrades to system-only caching.

Measured on an 8-turn replayed agent session (Bedrock us-east-1, claude-sonnet-4-5, ~10K-token system, ~7.5K-token tool results, markers on system + last history message), billed as `input + 1.25*write + 0.1*read`:

```
no caching at all:                353,359 token-equivalents
this PR (system+text blocks):     292,859  (-17%)  — tool_result marker dropped
+ tool_result marker preserved:   116,569  (-67%)  — exact parity with direct Bedrock
```

The delta is one small change in the `tool_result` branch — carry the block's `cache_control` onto the emitted tool message:

```python
tool_msg = {"role": "tool", "tool_call_id": tr["tool_use_id"], "content": str(tr_content)}
if "cache_control" in tr:
    tool_msg["cache_control"] = tr["cache_control"]
converted.append(tool_msg)
```

LiteLLM's converse transformation picks it up from there (verified against real Bedrock: per-turn usage becomes `input≈25, write=turn delta, read=full prior prefix`). Happy to send it as a follow-up PR if you'd rather keep this one's scope as is.

JerrettDavis (approved):
Re-approving after refreshing against current main at head 16c3ca89. The only conflict was in the LiteLLM system-message conversion path; I kept the PR behavior that preserves cache_control blocks while retaining legacy stringification for system lists without cache markers.

Validated locally after the merge:
- uv run --extra dev ruff check headroom/backends/litellm.py tests/test_bedrock_cache_markers.py
- uv run --extra dev pytest tests/test_bedrock_cache_markers.py tests/test_bedrock_region.py tests/test_bedrock_streaming_input_tokens.py -q (54 passed)

Final diff vs current main remains scoped to headroom/backends/litellm.py and tests/test_bedrock_cache_markers.py.

kimnamu:
Thanks for this — and for keeping it open long enough that the rest of the tree caught up with it.

Rebase check on current `main` (`232fb49c`, litellm 1.95.0): this PR still applies cleanly (`git apply --3way`, no conflicts), but two of its three halves have landed independently since it was approved, and the one that hasn't is the one that still changes behaviour.

**What's already on main**

| Path | Landed as | State on `main` |
|---|---|---|
| top-level `system` list | #2225 (`ea0115cb`) | `_system_field_to_message` keeps `cache_control` |
| `tool_result` blocks | #2144 (`0ad7dc7c`) | `tool_msg["cache_control"] = tr["cache_control"]` at `litellm.py:722` |
| plain text blocks | **not landed** | `litellm.py:747-749` still flattens to `"\n".join(text_parts)` |

**The remaining gap, and why it's still reachable**

`normalize_message_cache_control` puts one breakpoint on the last block of the last block-style message (`prefix_tracker.py:373-380`). In an agent loop that tail is usually a `tool_result` — covered. But when the tail is a plain text message (user replies after a tool turn, or a first turn), the marker lands on a `text` block and `_convert_messages_for_litellm` drops it one step later.

Ran both shapes through `normalize_message_cache_control` → `_convert_messages_for_litellm` → litellm's real `AmazonConverseConfig._transform_request`, counting `cachePoint` in the Converse body. Same conversation, only the tail message differs:

```
# clean main @ 232fb49c
Shape A - tail = tool_result:
   normalize() put the single breakpoint on: [(2, 'user')]
   cachePoint in Bedrock Converse body: 1
Shape B - tail = user text block:
   normalize() put the single breakpoint on: [(4, 'user')]
   cachePoint in Bedrock Converse body: 0

RESULT: A=1 cachePoint, B=0 cachePoint
```

With just this PR's text-block hunk applied, `B` becomes `1`. Both branches confirmed reachable — `normalize()` lands the marker on a `text` block for `tail=user text` and `tail=assistant text` alike.

Cross-check that litellm is the one asking for block form, not me: handing it the same message with `cache_control` preserved by hand yields `1` cachePoint, and the emitted Converse content is `[{"text": "..."}, {"cachePoint": {"type": "default"}}]`.

**Scope note for whoever merges**

The `system` half of this PR now conflicts semantically with what #2225 shipped. `_convert_system_for_litellm` returns a joined string when no block carries `cache_control`; main's `_system_field_to_message` always returns the block list. Both produce `[{"text": "be helpful"}]` and 0 cachePoints through the Converse transform, so it's not a behaviour difference at the wire — but the tests disagree:

- full PR as-is → `tests/test_bedrock_cache_markers.py` 9/9 pass (the PR brings its own `_convert_system_for_litellm`)
- text-block hunk alone on main → 6/9, the 3 failures all in `TestSystemPromptPreservesCacheControl` (two call `_convert_system_for_litellm`, which no longer exists; one asserts `isinstance(sys_msg["content"], str)`)

So narrowing this to the text-block branch means dropping that test class rather than keeping it. With the text-block hunk alone, `ruff check` / `ruff format --check` are clean and the Bedrock + cache suites pass:

```
tests/test_backend_nonstreaming_cache_metrics.py tests/test_backend_streaming_cache_metrics.py
tests/test_bedrock_prefix_tracker_wiring.py tests/test_bedrock_region.py
tests/test_bedrock_streaming_input_tokens.py tests/test_bedrock_tool_result_cache_and_streaming_stats.py
tests/test_cache_control_move_bust.py tests/test_cache_ttl_preserved.py
tests/test_litellm_nonstream_cache_usage.py tests/test_litellm_openai_passthrough.py tests/test_litellm_optional.py
→ 97 passed
```

Two things I checked before assuming the narrow hunk is safe, since it makes `content` a list for every provider and not just Bedrock:

- Requests with no `cache_control` anywhere still emit a joined string, so nothing changes for them.
- With a marker present, the OpenAI path gets the list too. `OpenAIGPTConfig.transform_request` accepts it and strips the key: `[{"role": "user", "content": [{"type": "text", "text": "hi"}]}]`. No raise.

One unrelated thing the repro surfaced, not caused by this PR: an `image` block alongside a marked `text` block is dropped on both main and the patched tree (`[{"role": "user", "content": "describe"}]`) — the branch only collects `type == "text"`. Not blocking anything here, just flagging.

Setup for the above: macOS arm64, Python 3.12, litellm 1.95.0, `main` at `232fb49c`, `headroom/_core` from the 0.33.0 wheel. `litellm.modify_params = True` was needed for the tool-call shapes. No live Bedrock calls — this measures what reaches the Converse body, so the billing effect @dspv and @ingmarkruschHF measured against real Bedrock is theirs, not re-verified by me.

Happy to send the narrowed text-block-only PR against `main` (with a replacement test for the text branch) if that's the shape you'd prefer, or leave it here if you'd rather land this one whole. Whichever is less work for you.

PR #1402, opened 2026-06-25
feat(compress): live-agent densify mode — lossless, cache-safe, value-factored

chopratejas:
## Description

Adds an opt-in agent densification mode that only shrinks content, verifies reversibility message by message, and reverts any output that cannot be proven lossless. It also adds a pure-Python inverse decoder and reversible low-cardinality value factoring for the existing `csv-schema` format.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- Add `compress(mode="agent")` and `densify()` with verified-lossless fallback.
- Add public `expand_compacted()` and `is_compacted()` helpers.
- Preserve null, empty-string, missing, and literal sentinel values distinctly in `csv-schema` output.
- Add reversible value factoring only when it reduces output size.
- Preserve current `main` dependency placement, security pins, CCR gating, and Windows ONNX initialization.

## Testing

### Test Output

```text
27 passed in 1.28s
71 Rust compaction tests passed; 843 filtered out
All checks passed! (Ruff)
Success: no issues found in 3 source files (mypy)
uv lock --check passed
```

## Real Behavior Proof

- Environment: macOS, Python 3.12.13, locked `uv` environment, current Rust toolchain
- Exact command / steps: ran the codec and agent-mode suites plus the Rust core compaction suite
- Observed result: all 98 selected tests passed; lossless round trips and value-factoring edge cases are covered
- Not tested: live multi-turn prompt-cache hit rate and proxy wiring; agent mode remains an opt-in library API

## Review Readiness

## Checklist

Not applicable: this PR does not change a user interface.

## Additional Notes

The branch was rebuilt on current `main`. The older dependency-relocation portion was intentionally omitted: current `main` keeps those dependencies available and includes a security exclusion for the compromised `ast-grep-cli` release. The current CCR gate already supersedes the PR's older gate implementation and was retained.

### DIFF
--- a/crates/headroom-core/src/transforms/smart_crusher/compaction/formatter.rs
+++ b/crates/headroom-core/src/transforms/smart_crusher/compaction/formatter.rs
@@ -290,9 +290,16 @@ fn write_table(
     }
 }
 
+/// Bare token for an absent key, distinct from null's empty cell. Lets the
+/// decoder tell `missing key` (`\N`) from `null` (empty) from `""` (`""`).
+/// A literal string equal to this token is CSV-quoted so it never collides.
+const MISSING_SENTINEL: &str = "\\N";
+
 fn format_cell(c: &CellValue) -> String {
     match c {
-        CellValue::Missing => String::new(),
+        // Absent key → sentinel; null → empty cell. Previously both rendered
+        // empty, collapsing the missing/null/empty-string distinction (lossy).
+        CellValue::Missing => MISSING_SENTINEL.to_string(),
         CellValue::Scalar(v) => json_scalar_to_csv(v),
         CellValue::Nested(sub) => {
             // Render nested as compact JSON; CSV-quote because it
@@ -354,7 +361,15 @@ fn json_scalar_to_csv(v: &Value) -> String {
 }
 
 fn needs_csv_quote(s: &str) -> bool {
-    s.contains(',') || s.contains('"') || s.contains('\n') || s.contains('\r')
+    // Quote the empty string so it renders `""` — distinct from a bare empty
+    // cell (null). Quote a literal `\N` so it can't be read back as the
+    // missing-key sentinel. Both are required for lossless round-tripping.
+    s.is_empty()
+        || s == MISSING_SENTINEL
+        || s.contains(',')
+        || s.contains('"')
+        || s.contains('\n')
+        || s.contains('\r')
 }
 
 fn csv_quote(s: &str) -> String {
--- a/headroom/__init__.py
+++ b/headroom/__init__.py
@@ -184,6 +184,9 @@ __all__ = [
     "compress_spreadsheet",
     "CompressConfig",
     "CompressResult",
+    # Reverse densified tool output (for consumers of mode="agent" output)
+    "expand_compacted",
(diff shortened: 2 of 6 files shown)


JerrettDavis (changes requested):
The agent-mode densification design is promising, especially the verified round-trip boundary, but this branch is not merge-ready yet.

Current CI has branch-owned failures:

- `commitlint` fails because commit `review: cross-platform + generic-harness hardening` uses a non-allowed type. Please squash/rename it to an allowed conventional type such as `test:` or `fix:`.
- `test (1)` fails because `ast-grep` is no longer available from a bare install: `tests/test_bundled_tools_savings.py::test_ensure_tools_installs_every_tool` and `test_ast_grep_slice_saves_tokens` raise `PlatformNotSupported` saying `pip install headroom-ai` should have placed `ast-grep` on PATH.
- `test (4)` has the same root symptom in the interceptor path: `tests/test_tool_result_interceptors.py` no longer outlines large Python reads, produces zero spans, and the transform adapter stops saving tokens.

The packaging change that moves `ast-grep-cli` out of the base install is therefore a breaking change for existing bundled-tool/interceptor behavior, not just a footprint cleanup. Please either keep `ast-grep-cli` in the default install, or explicitly change the affected behavior/tests/docs so the product has a clear supported path and the proxy/interceptor features remain green in CI.

Once the packaging fallout is resolved, I’d be happy to re-review the densify/lossless pieces separately.

JerrettDavis (changes requested):
Thanks for the update. The earlier `ast-grep-cli` packaging issue looks addressed by moving the binary dependency into the proxy/dev paths that exercise the interceptor behavior.

I found one remaining correctness issue in the new densify path: `tokens_after` is captured from `pipeline.apply()` before `_apply_value_factoring(compressed_messages)` runs. Value factoring can further shrink the output, but the code only recalculates `tokens_after` when `_verify_and_revert()` reverts at least one message. In the normal successful path (`reverted == 0`), `CompressResult.tokens_after`, `tokens_saved`, and `compression_ratio` describe the pre-factored payload while `messages` contains the post-factored payload.

Please recount after value factoring whenever it changes messages, or generally after lossless verification/factoring before computing `tokens_saved`. A focused regression would be useful: use a repeated low-cardinality column case that triggers `__dict:`, then assert `tokens_after` matches a fresh count of the returned `res.messages` and that `tokens_saved` reflects that final payload.

JerrettDavis:
Replayed the three still-substantive author commits onto current main and updated the branch in place. I retained current main's stronger CCR gate, Windows ONNX initialization, and dependency/security policy; the older dependency-relocation commit was omitted because its required dev tool is already present with the current compromised-release exclusion.

Validation on the replay:
- 27/27 focused Python tests passed
- 71/71 Rust compaction tests passed
- focused Ruff passed
- mypy on 3 touched source files passed
- uv lock --check and git diff --check passed

No UI changes; screenshots are not applicable.

JerrettDavis (approved):
Reviewed the current-main replay. The remaining feature intent is preserved, superseded security and runtime behavior stays on current main, and the focused Python/Rust quality gates pass.

PR #1403, opened 2026-06-25
test(rtk): document scope-mixing underreporting bug

peterlodri-sec:
## Description

Documents the RTK scope-mixing underreporting bug discovered in Discord: `b70fccbe` changed `_rtk_gain_scope()` default from `project` to `global`, causing `avg_savings_pct` to report lifetime-diluted numbers (18.5%) instead of actual session rate (62.6%).

These 6 tests document the invariant: `session_savings_pct` must be computed from token deltas scoped to the current session, not from a globally-diluted average.

Closes #

## Type of Change

## Changes Made

- Added 6 regression tests in `tests/test_rtk_savings_scope.py` documenting the correct scope for session savings percentage computation

## Testing

### Test Output

```text
tests/test_rtk_savings_scope.py::test_session_savings_uses_token_deltas PASSED
tests/test_rtk_savings_scope.py::test_session_savings_not_diluted_by_global_scope PASSED
(6 tests total, all pass)
```

## Real Behavior Proof

- Environment: headroom dev, Python 3.12
- Exact command / steps: `pytest tests/test_rtk_savings_scope.py -v`
- Observed result: All 6 tests pass; tests fail without the scope fix applied
- Not tested: Live proxy integration (unit-only)

## Review Readiness


JerrettDavis (changes requested):
Thanks for documenting the RTK savings scope issue. I’m requesting changes because this branch is not reviewable as a focused RTK regression PR yet.

The latest CI lint job is red: `ruff format --check .` reports `tests/test_rtk_gain_scope.py` would be reformatted. Please run the formatter or apply the equivalent formatting fix so the branch is green.

More importantly, the PR title/body describe RTK scope-reporting tests, but the diff also includes the kompress must-keep implementation, its tests, and a fine-tuning design document. Those are separate behavior/product changes and make it hard to merge or revert the RTK regression coverage safely. Please split the kompress must-keep/design work into its own PR, or retitle/rewrite this PR so reviewers can evaluate the full behavioral scope intentionally.

JerrettDavis:
The original RTK scope target is no longer part of the product: RTK and lean-ctx CLI context tools were removed from current `main` in #2677 (`e0ce4b1d`), along with their savings-accounting surface. The other changes bundled in this branch are a separate Kompress must-keep feature and design document; that work is represented by the focused #1419, which is the appropriate place to review its current semantics.

With the RTK regression obsolete and the unrelated Kompress work split into its successor, this mixed branch can be closed as superseded. Thanks for documenting the scope-mixing failure clearly—it was useful evidence even though that subsystem has since been retired.

Issue #804, opened 2026-06-10
[BUG] headroom wrap kills the shared proxy out from under other live wrap sessions (ConnectionRefused)


richard-scott:
# [BUG] `headroom wrap` kills the shared proxy out from under other live wrap sessions (ConnectionRefused)

## Environment

- **headroom version:** 0.24.0 (installed via `pipx install 'headroom-ai[proxy]'`)
- **OS:** Linux 6.8.0 (Ubuntu); also expected on macOS (same `pgrep`/argv semantics)
- **Python:** 3.10.19
- **Agent:** Claude Code, routed via `headroom wrap claude`
- **Auth/backend:** Anthropic subscription OAuth, `--backend anthropic`, `--mode token`

## Summary

`headroom wrap` does not allocate a per-instance proxy. The **first** `wrap`
instance starts the proxy on the hardcoded port **8787** and *owns* it; every
later `wrap` instance detects the port is busy and **reuses** the same proxy
(`Proxy already running on port 8787`).

The owner terminates the shared proxy when it exits. The guard that is supposed
to prevent this — "only kill the proxy if no other wrapped clients are using it"
— **cannot actually see the other clients**, because it scans process *command
lines* while wrapped clients carry the proxy URL in an *environment variable*.

Net effect: **whichever wrap session ends first kills the proxy for every other
session that is still running**, and the survivors fail mid-request with:

```
API Error: Unable to connect to API (ConnectionRefused)
```

## To Reproduce

Natural form (two terminals):

1. Terminal 1: `headroom wrap claude` — start a session, keep it working on a
   long task (large file read, long generation).
2. Terminal 2: `headroom wrap claude` — start a second session (it prints
   `Proxy already running on port 8787` and shares terminal 1's proxy).
3. Let **terminal 1's** session finish and exit while **terminal 2** is still
   mid-response.
4. Terminal 2's next API call fails with `ConnectionRefused` — its proxy was
   torn down by terminal 1's cleanup.

Scripted form (owner exits while borrower is still streaming):

```bash
# A = owner: short task, starts the proxy then exits quickly
( headroom wrap claude -- -p "say A_DONE" ) &
# wait until the proxy is actually up (boot takes ~3-7s)
until ss -ltn | grep -q 8787; do sleep 1; done
# B = borrower: long task, attaches to A's proxy, outlives A
( cat big_file.txt | headroom wrap claude -- -p "summarise this" ) &
wait
[...]

Issue #589, opened 2026-06-04
[BUG] Port 8787 is reserved by Windows, so Headroom Wrap Cursor can't start


Hakunanodesu:
## Description

Port 8787 is reserved by Windows, so Headroom Wrap Cursor can't start

## To Reproduce

use `headroom wrap cursor` cmd

## Actual Behavior

```
headroom wrap cursor
  Setting up rtk for Cursor...
  rtk instructions injected into C:\coding\apex-gui\.cursorrules

  ╔═══════════════════════════════════════════════╗
  ║            HEADROOM WRAP: CURSOR              ║
  ╚═══════════════════════════════════════════════╝

  Starting Headroom proxy on port 8787...
  Error: Proxy exited with code 1: (no log output)
```

oxura:
This should be better on current `main`: `wrap` now checks whether the requested localhost port can bind before it starts the proxy subprocess, and it should fail with an actionable message like:

`Port 8787 is unavailable on 127.0.0.1 ... On Windows this can happen when the port is in an excluded or reserved range. Rerun with a different port, for example headroom wrap cursor --port 8788.`

There is also regression coverage for `headroom wrap cursor --port 8788` being suggested when 8787 is unbindable.

If you are still seeing `Proxy exited with code 1: (no log output)` on a current build, please share the Headroom version/commit. Otherwise the immediate workaround is:

`headroom wrap cursor --port 8788`

JerrettDavis:
Verified on current main (6077e5a14): wrap probes 127.0.0.1 before spawning the proxy and _find_available_port skips EACCES (the Windows reserved/excluded-port result) and EADDRINUSE, so a reserved 8787 now advances to the next usable port and all downstream configuration receives that actual port. If no usable port can be selected, wrap fails before subprocess launch with an explicit Port 8787 is unavailable error instead of Proxy exited with code 1: (no log output). Focused wrap port/persistence/helper coverage plus the MCP dependency contract passes: 85 passed. Closing as fixed; please reopen with a current version and output if Windows still follows the old path.

PR #1406, opened 2026-06-25
fix(wrap): replace stale-proxy detection with Vite-style port fallback

lennney:
## Description

Replace the 180-line process-killing approach with a 15-line Vite-style socket.bind probe: if port is busy (EADDRINUSE or EACCES), try the next available port.

Pure stdlib -- no /proc, no lsof, no subprocess -- works identically on Linux/macOS/Windows.

## Problem

When `headroom wrap <agent>` is killed without proper cleanup (window close, SSH timeout, crash), the background proxy becomes orphaned and holds the port. The next `headroom wrap` on the same port would wait 30-45 seconds then fail with a confusing error.

This PR takes a simpler, safer approach: find the next available port. No process detection, no killing.

### Related issues

- **#589** (Port 8787 reserved by Windows) -- partially addressed: EACCES is now skipped together with EADDRINUSE
- **#804** (Shared proxy killed by exiting session) -- already fixed upstream via `_live_proxy_clients` marker files; this PR doesn't touch that code

## Type of Change

## Changes Made

- `headroom/cli/wrap.py`: Add `_find_available_port()` -- socket.bind loop that skips EADDRINUSE (busy) and EACCES (reserved/privileged) ports, returns first available port in range. Replace `_ensure_proxy` port-bind check with auto-fallback call to `_find_available_port`. Remove `_ensure_port_free()` call from `_start_proxy()`. Remove 8 dead functions: `_find_process_on_port`, `_linux_find_process_on_port`, `_resolve_inode_to_pid`, `_is_headroom_proxy`, `_read_process_cmdline`, `_kill_process`, `_ensure_port_free`, `_format_unbindable_port_error`.
- `tests/test_cli/test_wrap_helpers.py`: Remove 14 old `TestEnsurePortFree` tests (mocked /proc parsing, process killing). Add 6 new `TestFindAvailablePort` tests covering: port free, first port busy, multiple busy, EACCES skipped, unexpected error propagated, range exhausted.
- `tests/test_cli/test_wrap_persistent.py`: Adapt persistence tests for `_find_available_port` mock. Rewrite unbindable-port test to use new error path.

Zero new dependencies. Zero changes to core proxy server, MCP, compression, or providers.

## Testing

- [x] Unit tests pass (`python -m pytest tests/test_cli/ -v`)

### Test Output

```
> python -m pytest tests/test_cli/test_wrap_helpers.py::TestFindAvailablePort -v --no-header
============================= 6 passed ==============================
test_port_free_returns_same PASSED
test_port_busy_finds_next PASSED
test_multiple_busy_ports PASSED
test_propagates_unexpected_error PASSED
test_propagates_eaddrinuse_with_eacces PASSED
test_exhausts_range PASSED

> python -m pytest tests/test_cli/ -q
============================= 445 passed in 8.40s ==============================
```

## Real Behavior Proof

- Environment: Ubuntu 24.04 x86_64, Python 3.12.3
- Exact command / steps: Ran `python -m pytest tests/test_cli/test_wrap_helpers.py::TestFindAvailablePort -v` -- 6/6 pass for port fallback. Ran full test suite `python -m pytest tests/test_cli/ -q` -- 445/445 pass.
- Observed result: `_find_available_port(8787)` returns 8787 when free, 8788 when 8787 is busy. EACCES skipped same as EADDRINUSE. Non-retryable errors (EADDRNOTAVAIL) propagate immediately.
- Not tested: Windows EACCES fallback (no Windows CI runner). macOS port fallback (no macOS runner). Code path is identical across platforms (stdlib socket only).

## Review Readiness

## Checklist


JerrettDavis (changes requested):
The process-killing removal is the right direction, but the fallback port is not propagated to callers. _ensure_proxy() now starts the proxy on actual_port and prints the right dashboard URL, but it still returns only the Popen object. All existing callers continue using the originally requested port to set ANTHROPIC_BASE_URL / OPENAI_BASE_URL, inject provider config, register cleanup, and stop the proxy.

That means if 8787 is busy and _find_available_port() chooses 8788, the proxy starts on 8788 but the wrapped agent still points at 8787, i.e. the busy/stale port this PR was trying to avoid. Please change the contract so the selected port is returned/recorded and every caller uses that actual port, with regression coverage for a fallback case that asserts the launched agent environment/config points at the fallback port.

JerrettDavis (changes requested):
The fallback-port propagation is much better now for the generic launcher and proxy-only wrappers, but the Codex path still resolves the port too early.

`codex()` calls `_find_available_port(port)` before `_ensure_proxy()`. That means a healthy existing Headroom proxy on 8787 is treated as a busy port and Codex rewrites MCP/provider config to 8788, then `_launch_tool()` starts or uses 8788. The previous behavior would reuse the already-running compatible proxy on 8787. The same early lookup also breaks `--no-proxy`: if the user asks Codex to attach to an existing proxy on 8787, the bind probe sees 8787 as busy and points Codex at the next free port where no proxy is running.

Please route Codex through the same selected-port contract as the other wrappers: let `_ensure_proxy()` decide whether to reuse the requested port or fall back, then inject MCP/provider config and build the launch env from the returned `actual_port`. Add regression coverage for both cases: an existing compatible proxy on the requested port should keep that port, and `--no-proxy --port 8787` should not rewrite config/env to 8788 just because 8787 is occupied.

lennney:
@JerrettDavis Fix pushed. Here's what changed:

**Problem:** `codex()` called `_find_available_port(port)` early, which treated any busy port as "needs fallback" even when a healthy proxy was already running on the requested port. This broke both:
- Reusing an existing compatible proxy (would fall back to next port instead)
- `--no-proxy --port 8787` when port 8787 was occupied (would rewrite config/env to 8788)

**Fix:**
1. Removed the early `_find_available_port()` call from `codex()`
2. `codex()` now calls `_ensure_proxy(port)` directly — same selected-port contract as the other wrappers (aider, copilot, cursor)
3. `_ensure_proxy()` returns `actual_port`, which drives MCP registration, provider config injection, and launch-env building
4. `_launch_tool(port=actual_port, no_proxy=True)` skips duplicate proxy startup
5. Proxy cleanup in a `finally` block (checks `_live_proxy_clients` before terminating)

**Regression test added:** `TestCodexPortResolution.test_delegates_to_ensure_proxy` — mocks `_ensure_proxy` returning port 8788, verifies `_launch_tool` receives `port=8788`.

Commit: `bd3ae8a`

JerrettDavis (changes requested):
Requesting changes for the Codex prepare-only path.

`headroom wrap codex --prepare-only` now calls `_ensure_proxy(...)` before the `prepare_only` return, so a config-only operation starts the proxy and fails if proxy/runtime dependencies are not installed. I reproduced this locally on the PR branch with:

`python -m pytest tests/test_cli/test_wrap_helpers.py tests/test_cli/test_wrap_codex.py tests/test_cli/test_wrap_persistent.py -q`

Result: helper and persistent tests passed, but 13 Codex prepare/unwrap tests failed because prepare-only tried to start the proxy and hit missing proxy dependencies / module import errors. This is a contract regression: prepare-only should still be able to update Codex config without requiring a live proxy. Please keep the new resolved-port behavior for real launches, but avoid starting the proxy in prepare-only mode; the configured port value is enough for the config update in that path. Add/keep regression coverage that asserts prepare-only does not call `_ensure_proxy` or `_start_proxy`.

JerrettDavis (changes requested):
Thanks for fixing the prepare-only path; the original regression is gone. Re-running the broader wrapper target now gets through the prepare-only cases, but there is still one Codex lifecycle regression:

`python -m pytest tests/test_cli/test_wrap_helpers.py tests/test_cli/test_wrap_codex.py tests/test_cli/test_wrap_persistent.py -q`

Result: 135 passed, 1 failed.

Failure: `tests/test_cli/test_wrap_codex.py::test_wrap_codex_memory_launch_failure_unwrap_cleans_memory_only_config` now raises `FileNotFoundError` reading `.codex/config.toml`. From the diff, `codex()` now calls `_ensure_proxy(...)` before the MCP/provider/memory config injection on the normal launch path. If proxy startup fails, we exit before writing the managed config that the unwrap/cleanup path expects to remove. The prepare-only fix should stay, but please preserve the launch-failure cleanup contract: either keep the non-runtime config writes before the proxy startup as before, or update the lifecycle so a failed proxy start still leaves unwrap in the expected state and the existing test passes.

JerrettDavis (approved):
Reviewed the latest port-fallback updates. The prior Codex launch-failure cleanup regression is addressed: prepare-only remains config-only, and the normal Codex launch path now performs the non-runtime config writes before proxy startup or binary launch failure can abort. The selected `actual_port` from `_ensure_proxy()` is still used for runtime provider/env wiring, and the OpenCode follow-up now resolves the selected port before provider/env injection and refreshes MCP when the port changes.

Local focused verification passed: `python -m pytest tests/test_cli/test_wrap_helpers.py tests/test_cli/test_wrap_codex.py tests/test_cli/test_wrap_persistent.py -q` -> 144 passed.

I do not see a remaining code-review blocker. Only governance checks are attached to the latest head, so this approval is based on code review plus the targeted wrapper contract verification rather than a full CI matrix.

.hermes/pr-body.md
```diff
@@ -0,0 +1,70 @@
+## Description
```
JerrettDavis: This file likely should be dropped from the PR.

docs/plans/2026-06-25-port-fallback.md
```diff
@@ -0,0 +1,339 @@
+# Port Fallback: Auto-increment When Port Is Busy
```
JerrettDavis: This file likely should be dropped from the PR.

JerrettDavis (changes requested):

JerrettDavis (approved):

AGENTS.md
```diff
@@ -0,0 +1,84 @@
+# headroom AGENTS.md
```
JerrettDavis: Please exclude `AGENTS.md` from this changeset.

JerrettDavis (changes requested):

lennney:
Removed AGENTS.md from the changeset as requested. The branch has been force-pushed with only the 6 wrap-related files remaining in the diff.

cc @JerrettDavis

JerrettDavis (approved):

PR #1410, opened 2026-06-25
feat(wrap): add a memory-pressure governor to gate agent launches

Eyalm321:
## Problem

`headroom wrap` is the single point every wrapped agent crosses before it spawns
the heavy `claude` + MCP stack (e.g. serena + a language server, often several
GB each). On a busy multi-agent host the *Nth* launch is what tips total RAM
past the limit and triggers a **global** OOM — which the kernel can only resolve
by killing some victim (in practice the desktop compositor), freezing the whole
session.

## What this does

Gate the launch at that choke point. Before exec'ing the child, check available
memory (`/proc/meminfo` `MemAvailable`, `psutil` fallback) and, when it's below
a floor:

- **`wait`** (default) — poll until memory recovers, then proceed; if it never
  recovers within the budget, proceed anyway (fail-open) with a loud warning.
- **`refuse`** — exit `EX_TEMPFAIL` (75) so a supervisor / queue worker can retry
  later instead of OOMing the host.
- **`off`** — disabled.

Fail-open by construction: if available memory can't be measured (e.g. a
non-Linux host without `psutil`), the gate never blocks a launch.

## Config (env)

| var | default | meaning |
|---|---|---|
| `HEADROOM_WRAP_MEM_POLICY` | `wait` | `wait` \| `refuse` \| `off` |
| `HEADROOM_WRAP_MEM_MIN_MB` | `2048` | floor of available MiB |
| `HEADROOM_WRAP_MEM_WAIT_SECONDS` | `120` | max wait in `wait` mode |
| `HEADROOM_WRAP_MEM_POLL_SECONDS` | `3` | poll interval while waiting |

## Tests

`tests/test_wrap_mem_governor.py` — 12 tests covering off / enough-headroom /
unmeasurable / refuse / wait-then-recover / wait-timeout / mid-wait-unmeasurable
/ unknown-policy / bad-env. The clock, sleep, and memory read are all injected,
so the wait loop is deterministic and the suite runs instantly (no real
sleeping). `ruff` and `mypy` clean.

## Notes

Gates only the `claude` launch path (the heavy MCP/LSP multiplier); other wrap
targets are untouched. Defaults are conservative (2 GiB floor, wait-then-proceed)
so it never hard-fails a launch out of the box.

### DIFF
--- a/headroom/cli/mem_governor.py
+++ b/headroom/cli/mem_governor.py
@@ -1,0 +1,163 @@
+"""Memory-pressure governor for ``headroom wrap`` (root-cause OOM guard).
+
+Every wrapped agent (``headroom wrap claude``) crosses one ``wrap`` process
+before it spawns the heavy ``claude`` + MCP stack (e.g. serena + rust-analyzer,
+often several GB each). On a busy multi-agent box the *Nth* launch is what tips
+total RAM past the limit and triggers a **global** OOM — which the kernel can
+only resolve by killing some victim, in practice the desktop compositor,
+freezing the whole session.
+
+``wrap`` is the single choke point every agent passes through, so it is the
+natural place to throttle. Before launching the child, :func:`gate_launch`
+checks how much memory is actually available and, when the box is already low:
+
+* ``wait``   (default) — poll until memory recovers, then proceed; if it never
+  recovers within the budget, proceed anyway (fail-open) with a loud warning.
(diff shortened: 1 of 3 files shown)


Eyalm321:
Withdrawing — addressing the underlying memory pressure at the LSP layer (shared language servers) rather than via launch-time admission control. Thanks!

PR #1418, opened 2026-06-25
feat(router): per-content-type bias multipliers for kompress (domain routing)

peterlodri-sec:
## Description

Adds data-backed per-content-type bias multipliers to the kompress compression path. Different content types tolerate different compression aggressiveness while maintaining must-keep token survival (≥95% threshold).

Derived from `eval_domain_routing.py` on kompress-v4 with 20 samples per domain.

## Type of Change

- [x] Performance improvement

## Changes Made

- `headroom/transforms/content_router.py` — `_DOMAIN_BIAS_MULTIPLIERS` dict at module level, applied before `_compress_pure`/`_compress_mixed` when content detection ran

## Testing

- [x] Unit tests pass
- [x] Linting passes

### Test Output

```
BUILD_OUTPUT  bias=0.50  ratio=2.15x  mk=96.8%   (error/build traces)
SOURCE_CODE   bias=0.50  ratio=1.99x  mk=96.8%   (file reads)
SEARCH_RESULTS bias=0.70 ratio=1.45x  mk=98.9%   (search output)
PLAIN_TEXT    bias=1.00  unchanged
JSON_ARRAY    SmartCrusher path, not affected
```

## Real Behavior Proof

- Environment: macOS 15.5, Python 3.12, kompress-v4, headroom 0.28.0
- Exact command / steps: `python3 ultrawhale/scripts/eval_domain_routing.py --model PeetPedro/kompress-v4 --samples 20`
- Observed result: error traces 2.15x compression at 96.8% must-keep survival; import OK after change
- Not tested: live proxy session end-to-end

## Review Readiness

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -1678,6 +1678,16 @@ class ContentRouterConfig:
     search_group_by_file: bool = False
 
 
+# Domain routing values derived from kompress-v4 evaluation with 20 samples
+# per content type and a >=95% must-keep survival threshold. Values below 1.0
+# make Kompress more aggressive; unlisted domains retain the caller's bias.
+_DOMAIN_BIAS_MULTIPLIERS: dict[ContentType, float] = {
+    ContentType.BUILD_OUTPUT: 0.50,
+    ContentType.SOURCE_CODE: 0.50,
+    ContentType.SEARCH_RESULTS: 0.70,
+}
+
+
 class ContentRouter(Transform):
     """Intelligent router that selects optimal compression strategy.
 
@@ -2204,6 +2214,11 @@ class ContentRouter(Transform):
                         else "content_detection"
                     ),
                 )
+
+            # Preserve explicit force-Kompress behavior: domain routing only
+            # applies when normal content detection selected the strategy.
+            if not force_kompress:
+                bias *= _DOMAIN_BIAS_MULTIPLIERS.get(detection.content_type, 1.0)
 
             if strategy == CompressionStrategy.MIXED:
                 result = self._compress_mixed(content, context, question, bias=bias)
--- a/tests/test_transforms_content_router.py
+++ b/tests/test_transforms_content_router.py
@@ -388,6 +388,72 @@ def test_normal_compress_path_still_uses_content_detection(
     assert calls["detect"] > 0
 
 
+@pytest.mark.parametrize(
+    ("content_type", "multiplier"),
+    [
+        (ContentType.BUILD_OUTPUT, 0.50),
+        (ContentType.SOURCE_CODE, 0.50),
+        (ContentType.SEARCH_RESULTS, 0.70),
+        (ContentType.PLAIN_TEXT, 1.00),
+    ],
+)
+def test_domain_routing_scales_detected_content_bias(
+    monkeypatch: pytest.MonkeyPatch,
+    content_type: ContentType,
+    multiplier: float,
+) -> None:
+    router = ContentRouter()
+    observed: dict[str, float] = {}
+
+    monkeypatch.setattr(content_router_module, "is_mixed_content", lambda _content: False)
+    monkeypatch.setattr(
+        content_router_module,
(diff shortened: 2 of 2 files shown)


peterlodri-sec:
## Context from kompress evaluator-optimizer session (2026-06-25)

The domain routing bias multipliers in this PR are validated by a new must-keep override safety net now being added to the kompress compressor (separate PR incoming).

**Key finding:** The ModernBERT subword tokenizer splits compiler flags (`-O2` → `-` + `O` + `2`), file paths (`/var/log` → `/` + `var` + `/` + `log`), CamelCase identifiers (`TokenExpiredError` → `Token` + `Expired` + `Error`), and dotted names (`Cargo.toml` → `C` + `argo` + `.` + `tom` + `l`) across multiple subword tokens. The single-token ML classifier then misses these fragments (scores 0.000–0.187 on subword pieces), causing must-keep information loss.

**Solution:** A sliding-window regex safety net (1–3 words) applied AFTER ML inference catches these patterns regardless of tokenization. On 50 agent data samples, this takes the model from 65.2% must-keep-in-reference survival to **96.2%**. The regex is the same one used in `eval_domain_routing.py` to calibrate these bias multipliers.

**Implication for domain routing:** The domain-specific bias values (BUILD_OUTPUT=0.50, SOURCE_CODE=0.50, SEARCH_RESULTS=0.70) were calibrated WITH the must-keep override active. These multipliers work correctly when the override is also active. Running without the override would risk dropping must-keep tokens at any bias level.

**Recommendation:** Land this PR together with the must-keep override PR for best results. Both changes are in production code (no model retraining needed).

---

*Generated from kompress evaluator-optimizer session — ultrawhale/scripts/evaluator_optimizer.py + Qwen2.5-7B teacher (HF_INFER_PRO)*

JerrettDavis (approved):
I reviewed the domain-bias routing change and the stacked `none` tool-profile changes. The routing stays behind existing content detection, preserves `force_kompress`, and multiplies into the existing bias rather than replacing tool/profile settings. The explicit `none` sentinel handling in both OpenAI tool messages and Anthropic tool_result blocks has targeted coverage.

No correctness blocker from me. Only governance checks are attached to this PR, so I’m basing this on the code review and the included focused tests rather than a full CI run.

JerrettDavis (approved):
Rechecked current head after the commitlint cleanup. Maintainer edit only rewrote the historical `merge(main): ...` commit header to `chore(main): ...`; file content is unchanged from the previously reviewed branch. Commitlint no longer shows as a completed failure on the current head.

GitHub now reports the branch as conflicted, so I adjusted labels accordingly.

JerrettDavis (changes requested):
I need to withdraw the prior approval for the current branch state.

After the 2026-07-16 update, this branch is no longer reviewable as the focused domain-routing change. Locally, `git diff --name-only headroomlabs/main...HEAD` reports 2,307 files changed, and the diff includes unrelated historical PR content, vendored `node_modules` churn/deletions, docs/spec removals, SBOM artifacts, and broad lockfile/workflow changes. `headroomlabs/main` is also not an ancestor of the current head, while GitHub reports the PR as conflicted.

Please rebuild/rebase this branch on current `headroomlabs/main` so the PR contains only the domain routing bias multiplier implementation and its focused tests/docs. I would not try to resolve this as a maintainer merge because the current branch shape is too broad to distinguish the intended feature from unrelated repository history.

JerrettDavis:
Updated this branch in place to a focused replay on current `main` at `9410a78d`, preserving the original feature commit authorship and excluding the stale historical-main merge. Added regression coverage for all three domain multipliers, the unchanged plain-text default, and the force-Kompress bypass.

Validation:
- ContentRouter suites: **122 passed**
- focused Ruff + format: pass
- focused mypy: pass
- `uv lock --check`: pass
- `git diff origin/main --check`: pass

JerrettDavis (approved):
The focused replay preserves the evaluated bias behavior and now has direct regression coverage.

PR #1419, opened 2026-06-25
feat(kompress): sliding-window must-keep override for subword-tokenized patterns

peterlodri-sec:
## Problem

The ModernBERT subword tokenizer (used by kompress) splits critical patterns across multiple tokens:

| Pattern | Tokenization | Model scores on subwords |
|---|---|---|
| `-O2` | `-`, `O`, `2` | `O`=0.5, `2`=0.8 — no single token matches |
| `Cargo.toml` | `C`, `argo`, `.`, `tom`, `l` | `argo`=0.000, `tom`=0.003 — dropped |
| `TokenExpiredError` | `Token`, `Expired`, `Error` | all subtokens ≥0.9 — survives, but fragile |
| `/var/log/app.log` | `/`, `var`, `/`, `log`, `app`, `.`, `log` | `log`=0.187 in one context — dropped |

The single-token ML classifier misses must-keep fragments when they score low individually, causing information loss of compiler flags, hex addresses, file paths, env var names, and error types.

The v7 training run tried teaching the model these patterns via sliding-window self-labeling but regressed — heretic precision dropped from 0.967 → 0.956 and keep_rate increased from 0.823 → 0.868 (model became more conservative, not smarter).

## Solution

A post-inference regex safety net with a 1-3 word sliding window, applied AFTER the ML model makes its keep/drop decisions. Controlled by `KompressConfig.enable_must_keep_override` (default `True`).

## Results

On 50 agent data samples from ultrawhale's kompress_agent_train.jsonl:

| Metric | Without override | With override |
|---|---|---|
| mk_in_ref survival | 0.652 | 0.962 |
| Failing pairs (mk<0.9) | ~all | 3/50 |

Evaluated with v4 model + Qwen2.5-7B teacher analysis (script: ultrawhale/scripts/evaluator_optimizer.py).

## Design choices

- Post-inference, not training — the model stays v4-precise (0.967 heretic); regex adds coverage without making the model more conservative
- Regex matches eval_domain_routing.py in ultrawhale — keeping calibration and production in sync
- Sliding window is 1-3 words — catches both -O2 (1 word) and -O2 -Wall (2 words as a window)

## Related

- Part of the domain routing + must-keep override pair (#1418)
- Calibration: ultrawhale/scripts/eval_domain_routing.py
- Diagnosis: ultrawhale/scripts/evaluator_optimizer.py (Qwen2.5-7B teacher)
- All 28 kompress tests + 53 content_router tests pass


peterlodri-sec:
## Related: domain routing PR #1418

This must-keep override is designed to land together with **[#1418](https://github.com/headroomlabs-ai/headroom/pull/1418) — Domain routing bias multipliers**.

The domain routing bias values (BUILD_OUTPUT=0.50, SOURCE_CODE=0.50, SEARCH_RESULTS=0.70) were calibrated WITH the must-keep override regex active (see ultrawhale/scripts/eval_domain_routing.py). Without this override, running at bias=0.50 would risk dropping must-keep tokens that the single-token classifier misses.

Both changes are production-only — no model retraining, no GPU, no HuggingFace uploads. Just a regex safety net + per-domain bias multipliers.

JerrettDavis (changes requested):
The direction makes sense, but this needs a little more wiring before it is safe to merge.

The override is only applied in `KompressCompressor.compress()`. `compress_batch()` has its own inference/kept-id path and never runs the must-keep sliding-window check, so callers using the batch path can still drop the exact compiler flags, paths, and identifiers this PR is meant to protect. Please apply the same helper to both single and batch compression paths, ideally by extracting the keep-set update into one small function, and add regression coverage for both paths.

Also please run/clear normal lint. The diff currently adds `import re` after `import threading`, which is likely to trip import sorting once the full CI runs, and this PR currently only has governance checks attached.

JerrettDavis (approved):
Re-reviewed after merging current main. The PR’s must-keep intent is now fully subsumed by main’s shared _add_kompress_must_keep_words helper, which is invoked from both compress() and compress_batch(); the branch has zero net diff against main, so the earlier single-path blocker no longer applies. Validation: Ruff and format pass; HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 pytest tests/test_transforms/test_kompress_compressor.py -> 41 passed. (The initial online run passed 24 tests before hanging on an external Hugging Face download.)

JerrettDavis:
The sliding-window must-keep behavior from this branch is already incorporated in current `main`. GitHub now reports zero changed files for this PR, so there is no remaining implementation or test delta to merge.

I think this focused successor can be closed as superseded/no-op. Thank you for developing the subword-aware safety override and its adversarial pattern coverage.

PR #1230, opened 2026-06-21
feat(wrap): make tokensave the primary coding-task compressor, Serena the backup

aovestdipaperino:
## Description

Makes **tokensave** ([github.com/aovestdipaperino/tokensave](https://github.com/aovestdipaperino/tokensave)) the **primary coding-task compressor** that `headroom wrap` installs, and demotes **Serena** to a **backup**. tokensave is a local semantic code-graph MCP server (`tokensave serve`): the agent queries it for symbols, call chains, and impact analysis instead of grepping/reading whole files — the same role Serena filled, but as a pre-indexed graph. Serena now only registers when tokensave is unavailable (or when forced with `--serena`).

Closes #

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `headroom/graph/tokensave_installer.py` (new): fetch the prebuilt tokensave release binary for the platform (release-binary only — no `cargo` compile at wrap time); honors `HEADROOM_BINARIES_OFFLINE`; returns `None` (→ Serena) when no asset exists (e.g. x86_64 macOS) or the download fails.
- `mcp_registry`: `build_tokensave_spec()`; registration/disable/migrate go through the existing `ServerSpec` + ownership-ledger flow, identical to Serena.
- `cli/wrap.py`: new `_setup_coding_compressor` primary/backup policy; tokensave setup/disable/migrate/index helpers. New flags `--no-tokensave` (skip primary) and `--serena` (force backup on); `--no-serena` now means "never register the backup". Default wrap removes a previously Headroom-installed Serena entry once tokensave is primary (user-managed entries preserved). `--code-graph` repointed to tokensave; the legacy `codebase-memory-mcp` install path is dropped (unwrap still cleans up legacy entries). `unwrap claude|codex` remove a ledger-owned tokensave entry.
- Strands `HeadroomBundle`: `enable_tokensave_mcp=True` (primary); `enable_serena_mcp` now defaults `False` (backup).
- `docs/content/docs/proxy.mdx`: `--code-graph` description updated from codebase-memory-mcp to tokensave.
[...]

PR #1422, opened 2026-06-25
fix(installers): verify SHA-256 digests for rtk, lean-ctx, cbm release binaries

### LINKED ISSUE #1407: Harden release-binary downloads with pinned SHA-256 digest verification (rtk, lean_ctx, graph/cbm)
## Summary

Follow-up requested in [#1230](https://github.com/headroomlabs-ai/headroom/pull/1230) (see the [review thread](https://github.com/headroomlabs-ai/headroom/pull/1230)).

The same release-binary download pattern that PR #1230 hardened for the new tokensave default also exists, **unhardened**, in three other installers:

- `headroom/rtk/installer.py`
- `headroom/lean_ctx/installer.py`
- `headroom/graph/installer.py` (codebase-memory-mcp / cbm)

Each of these fetches and executes prebuilt binaries from release downloads **without SHA-256 digest verification**. Their `_should_verify_target` is only a runtime `--version` probe, not an integrity check, so a tampered or substituted artifact would pass.

## Why it matters

`rtk` in particular is **auto-downloaded by `headroom wrap`**, so an unverified binary is fetched and run as part of normal operation. That trust boundary predates #1230; #1230 only closed the gap for the tokensave path.

## Proposed hardening

Lift the pinned-digest verification added in #1230 into a **shared helper** and apply it consistently to `rtk`, `lean_ctx`, and `graph`/cbm:

- Verify the downloaded artifact against a pinned SHA-256 digest before execution.
- Fail closed on mismatch or missing digest.
- Reuse one implementation across all installers rather than per-installer copies.

## Scope

Tracking only. Left untouched in #1230 to keep that PR focused on the new tokensave default.

obchain:
## Description

The `rtk`, `lean-ctx`, and `codebase-memory-mcp` installers fetch prebuilt release binaries and then **execute** them — `rtk` in particular is auto-downloaded and run as part of a normal `headroom wrap`. Until now they did this with **no integrity check**: their `_should_verify_target` is only a runtime `--version` probe, so a tampered or substituted release artifact would be run silently.

This pins a SHA-256 digest for every release asset and verifies the downloaded bytes **before extraction**, mirroring the pattern already shipped for the tokensave installer in #1230.

Closes #1407

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/rtk/installer.py`, `headroom/lean_ctx/installer.py`, `headroom/graph/installer.py`: added a pinned `*_ASSET_DIGESTS` map and a `_verify_asset_digest()` helper, invoked between download and extraction.
- A digest mismatch raises `RuntimeError` and aborts the install (binary is never unpacked or executed).
- Overridden versions / cross-target downloads have no pinned digest and are **refused by default**, with an explicit per-tool opt-out: `HEADROOM_RTK_ALLOW_UNVERIFIED`, `HEADROOM_LEAN_CTX_ALLOW_UNVERIFIED`, `HEADROOM_CBM_ALLOW_UNVERIFIED`.
- Digests were taken from each upstream release's published asset digests for the currently pinned versions (rtk `v0.28.2`, lean-ctx `v3.4.7`, codebase-memory-mcp `v0.8.1`).
- Tests: new `tests/test_installer_digest_verification.py`; updated the existing fake-byte download tests to exercise the new verification path.
- `CHANGELOG.md`: added a Security entry.

### Note on the Windows codebase-memory-mcp asset

`graph/installer.py` builds `codebase-memory-mcp-windows-amd64.tar.gz`, but the upstream release only publishes a `.zip` for Windows — so that download already 404s today, independent of this change. No Windows digest is pinned; the path now fails closed at verification rather than at the (already failing) download. Happy to file/fix the `.tar.gz`/`.zip` mismatch as a separate issue to keep this PR scoped to integrity verification.

## Testing

- [ ] Type checking passes (`mypy headroom`) — not run locally (see Additional Notes)

### Test Output

```text
$ uv run ruff check headroom/rtk/installer.py headroom/lean_ctx/installer.py headroom/graph/installer.py tests/test_installer_digest_verification.py tests/test_rtk_installer.py tests/test_lean_ctx_installer.py tests/test_graph.py
All checks passed!

$ uv run pytest tests/test_installer_digest_verification.py tests/test_rtk_installer.py tests/test_lean_ctx_installer.py tests/test_graph.py -q
======================== 45 passed, 6 warnings in 0.16s ========================
```

(The 6 warnings are the pre-existing Python 3.14 `tarfile.extract` filter deprecation in the installers — untouched by this PR.)

## Real Behavior Proof

- Environment: macOS (arm64), Python 3.13, `uv run`, headroom @ this branch. Verified end-to-end against the real published rtk release.
- Exact command / steps: download the real `rtk-x86_64-unknown-linux-musl.tar.gz` for `v0.28.2`, compare its SHA-256 to the pinned value, then drive `download_rtk()` with (a) the real bytes and (b) the same bytes with one flipped byte.
- Observed result: pinned digest matches the real artifact; real bytes install; tampered bytes are rejected with nothing written (full output below).
- Not tested: live download of lean-ctx / codebase-memory-mcp assets through `headroom wrap` on each OS (rtk exercises the shared code path end-to-end; the other two are structurally identical and covered by unit tests). Windows codebase-memory-mcp path not exercised (pre-existing asset-name mismatch, noted above).

```text
1) Downloading REAL published asset:
   https://github.com/rtk-ai/rtk/releases/download/v0.28.2/rtk-x86_64-unknown-linux-musl.tar.gz
   real   sha256 = c7b61e87b8430e42b04ab84fbe1b3b41b563454b0181247fd04844b8e9194371
   pinned sha256 = c7b61e87b8430e42b04ab84fbe1b3b41b563454b0181247fd04844b8e9194371
   MATCH

2) Install with REAL bytes (pinned target+version) -> should SUCCEED
   installed at: /tmp/.../rtk | exists: True

3) Install with TAMPERED bytes -> should ABORT, nothing written
   REJECTED: rtk asset 'rtk-x86_64-unknown-linux-musl.tar.gz' failed integrity check:
     expected sha256 c7b61e87...4371, got 64e2ab60...7764
   binary written?: False
```

## Review Readiness

## Checklist

## Additional Notes

- `mypy headroom` was not run locally (mypy isn't part of the `[dev]` extra in this environment); `ruff check` and the full touched-test suite are green. Happy to address any type findings CI surfaces.
- Verification is intentionally fail-closed for unpinned assets, with a documented per-tool opt-out env var so version overrides remain possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (LLMs, loggers, attribution parsers) a machine-readable provenance
   boundary and preventing misattribution in multi-agent threads.
 
+### Security
+
+* **installers:** pin and verify SHA-256 digests for the `rtk`, `lean-ctx`, and `codebase-memory-mcp` release binaries before extraction. These installers fetch prebuilt binaries that `headroom wrap` then *executes* (`rtk` is auto-downloaded as part of normal operation), but previously did so without any integrity check — a tampered or substituted release artifact would have been run silently. Each installer now verifies downloaded bytes against a pinned digest map (`RTK_ASSET_DIGESTS` / `LEAN_CTX_ASSET_DIGESTS` / `CBM_ASSET_DIGESTS`) and aborts on mismatch. Overridden versions / cross-target downloads have no pinned digest and are refused unless the operator opts out with `HEADROOM_RTK_ALLOW_UNVERIFIED` / `HEADROOM_LEAN_CTX_ALLOW_UNVERIFIED` / `HEADROOM_CBM_ALLOW_UNVERIFIED`. Extends the digest-pinning pattern introduced for the tokensave installer to the remaining release-binary installers ([#1407](https://github.com/headroomlabs-ai/headroom/issues/1407)).
+
 ### Changed
 
 * **telemetry:** anonymous usage telemetry is now **opt-in** (off by default) instead of opt-out. Nothing is collected or sent unless you set `HEADROOM_TELEMETRY=on` or pass `--telemetry` to `headroom proxy` / `headroom install apply`. `is_telemetry_enabled()` is fail-closed — only explicit on-values (`on`/`true`/`1`/`yes`/`enable`/`enabled`) enable it; unset, empty, or unrecognized values stay disabled. The existing `--no-telemetry` flag and `HEADROOM_TELEMETRY=off` remain accepted for back-compat, and install manifests now write the `HEADROOM_TELEMETRY` value explicitly so generated deployments are unambiguous.
--- a/headroom/graph/installer.py
+++ b/headroom/graph/installer.py
@@ -1,9 +1,21 @@
-"""Download and install codebase-memory-mcp binary from GitHub releases."""
+"""Download and install codebase-memory-mcp binary from GitHub releases.
+
+Supply-chain integrity:
+    Headroom downloads and then executes this binary, so every pinned
+    release asset is verified against a SHA-256 digest in
+    ``CBM_ASSET_DIGESTS`` below before the archive is unpacked. A mismatch
+    aborts the install rather than running a tampered or substituted
+    artifact. When the version is overridden there is no pinned digest, so
+    the download is refused unless the operator opts out via
+    ``HEADROOM_CBM_ALLOW_UNVERIFIED=1``.
+"""
 
 from __future__ import annotations
 
+import hashlib
 import io
 import logging
+import os
 import platform
 import shutil
 import stat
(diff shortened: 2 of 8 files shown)


JerrettDavis (approved):
I reviewed the installer digest changes and updated the branch from `headroomlabs/main`. The only merge conflict was `CHANGELOG.md`; I kept the new Security entry and the current main Unreleased content.

The fail-closed behavior for pinned release artifacts looks right: downloaded bytes are checked before extraction, tampered bytes abort, and unpinned/version-overridden assets require an explicit per-tool opt-out. The regression coverage hits the helper behavior and an end-to-end rtk tamper path.

Local verification after the merge:

```text
uv run pytest tests/test_installer_digest_verification.py tests/test_rtk_installer.py tests/test_lean_ctx_installer.py tests/test_graph.py -q
```

Result: all digest/installer tests passed except `tests/test_rtk_installer.py::test_register_claude_hooks_survives_forked_daemon`, which fails locally on Windows with `WinError 193` because the test executes a fake script path as though it were a native executable. That is unrelated to this PR's digest verification path.

JerrettDavis (changes requested):
Requesting changes for the red test shard. The integrity-check design is the right direction, but `tests/test_installer_digest_verification.py::test_download_rtk_rejects_tampered_pinned_asset` currently passes `download_rtk("v0.28.2")`; the installer treats that as a version override because it does not equal `RTK_VERSION`, so `expected` becomes `None` and the failure is `no pinned SHA-256 digest...` instead of the intended `failed integrity check`.

Please make the test exercise a pinned asset path, for example by calling `download_rtk()` with the module default version or by passing `rtk_installer.RTK_VERSION` exactly. If the public API accepts both `0.28.2` and `v0.28.2`, normalize that comparison in the installer too, otherwise an operator can accidentally bypass the pinned map and hit the unverified override path for the current release.

JerrettDavis (changes requested):
Reviewed the June 30 follow-up at c6001197. The pinned-version test now uses `RTK_VERSION`, and canonicalizing bare versions addresses the earlier comparison finding. The remaining blockers are against today's main:

- `headroom/rtk/installer.py`, `headroom/lean_ctx/installer.py`, and their tests have since been removed. The current merge has modify/delete conflicts, so retaining those files would resurrect removed integrations. Please port the applicable integrity checks to the remaining installer surface, preserve main's removals, and update the PR description accordingly. The current CBM installer also has newer platform/version handling that needs to be retained with matching pinned digests.
- Each `_verify_asset_digest` treats any nonempty `HEADROOM_*_ALLOW_UNVERIFIED` value as permission to bypass verification for unpinned assets. Values such as `0` and `false` therefore disable the fail-closed protection despite expressing an opt-out of the bypass. Please recognize only explicit true values and cover unset/empty/false/0 versus true/1, as well as ensuring a known digest mismatch is never bypassed.

This PR is already labeled stale after the long period without source updates. I am leaving that triage signal in place; the conflict resolution now requires a scope decision rather than a mechanical refresh.

PR #1448, opened 2026-06-25
feat(proxy): add corporate/macOS system proxy support

daveeed1:
- Rust proxy: new --outbound-proxy / HEADROOM_PROXY_OUTBOUND_PROXY flag wires reqwest::Proxy::all() onto the upstream client
  - Python proxy: find_system_proxy() reads macOS System Preferences via urllib.request.getproxies() and injects it into httpx when HTTPS_PROXY env var is absent (avoids double-proxying)
  - Switch reqwest to rustls-tls-native-roots so the OS trust store is used

## Description

Need to use headroom inside a corporate proxy. Before when you wrap claude, then in outbound claude cannot call APIs. Now it pass througth corporate proxy and goes to internet

Closes #

## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- 

## Testing

- [X ] Manual testing performed

### Test Output

```text
# Paste relevant command output or artifact links here
```

## Real Behavior Proof

- Environment:
- Exact command / steps:
- Observed result:
- Not tested:

## Review Readiness

- [ X] I have performed a self-review
- [ X] This PR is ready for human review

## Checklist

- [ ] My code follows the project's style guidelines
- [ ] I have performed a self-review of my code

Add screenshots to help explain your changes.

### DIFF
--- a/crates/headroom-proxy/src/config.rs
+++ b/crates/headroom-proxy/src/config.rs
@@ -197,6 +197,15 @@ pub struct CliArgs {
     #[arg(long, env = "HEADROOM_PROXY_UPSTREAM")]
     pub upstream: Url,
 
+    /// Corporate/outbound HTTP(S) proxy for upstream requests
+    /// (e.g. http://proxy.corp.example.com:8080).
+    /// Applied to all traffic headroom forwards to upstream.
+    /// Reads HTTPS_PROXY / HTTP_PROXY env vars as a fallback when unset.
+    ///
+    /// Source priority: CLI flag → `HEADROOM_PROXY_OUTBOUND_PROXY` env var → unset.
+    #[arg(long = "outbound-proxy", env = "HEADROOM_PROXY_OUTBOUND_PROXY")]
+    pub outbound_proxy: Option<String>,
+
     /// End-to-end timeout for a single upstream request (long, since LLM
     /// streams may run for many minutes).
     #[arg(long, default_value = "600s", value_parser = parse_duration)]
@@ -551,6 +560,9 @@ pub struct Config {
     /// PR-D4: GCP ADC OAuth scope used when fetching the bearer
     /// token. Default `https://www.googleapis.com/auth/cloud-platform`.
     pub vertex_adc_scope: String,
+    /// Optional outbound proxy URL for all upstream-bound requests.
+    /// `None` means reqwest falls back to `HTTPS_PROXY` / `HTTP_PROXY` env vars.
+    pub outbound_proxy: Option<String>,
 }
 
 impl Config {
@@ -587,6 +599,7 @@ impl Config {
             bedrock_validate_eventstream_crc: args.bedrock_validate_eventstream_crc,
             vertex_region: args.vertex_region,
             vertex_adc_scope: args.vertex_adc_scope,
+            outbound_proxy: args.outbound_proxy,
         }
     }
 
@@ -641,6 +654,7 @@ impl Config {
             // only; the upstream URL is `upstream`).
             vertex_region: "us-central1".to_string(),
             vertex_adc_scope: "https://www.googleapis.com/auth/cloud-platform".to_string(),
+            outbound_proxy: None,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


JerrettDavis (changes requested):
This is useful functionality for enterprise networks, but there are a couple of issues to fix before merge.

`headroom/proxy/server.py` imports `find_ca_bundle` but never uses it. That will trip `ruff`/pyflakes (`F401`) on the changed file, so please drop that import or use it intentionally.

More importantly, `find_system_proxy()` logs the full detected proxy URL. Corporate proxy URLs commonly include credentials (`http://user:pass@proxy:8080`), so `logger.info("event=system_proxy_detected url=%s", url)` can leak secrets into Headroom logs. Please redact userinfo before logging, or log only scheme/host/port.

After those are fixed, please add at least small unit coverage around `find_system_proxy()` for env-proxy skip behavior and credential redaction. That gives us confidence this stays safe across macOS/system/env proxy cases.

JerrettDavis (changes requested):
Thanks for redacting the logged system proxy URL; that addresses the secret-leak part of the prior review. I still need changes before this is mergeable:

- Please add the unit coverage requested in the previous review. At minimum, cover that env proxy settings suppress system-proxy injection, and that a proxy URL containing userinfo is logged without credentials.
- `find_system_proxy()` currently skips only `HTTPS_PROXY` / `https_proxy`, but its docstring says this avoids double-proxying when env vars are set and `httpx` trust-env handles them. Please include `HTTP_PROXY` / `http_proxy` in that guard as well, otherwise a user with only `HTTP_PROXY` configured can still get an explicit system proxy layered on top of env proxy handling.

The PR body also still has the governance placeholder warning, so please update the description with the actual changes and verification while you are touching it.

JerrettDavis (changes requested):
Thanks for the follow-up commit. The credential redaction path is improved, but the branch still has two blockers from the prior review:

- `find_system_proxy()` still only checks `HTTPS_PROXY` / `https_proxy` before falling back to `urllib.request.getproxies()`. The previous review also called out `HTTP_PROXY` / `http_proxy`; with only `HTTP_PROXY` configured, this can still layer an explicit system proxy on top of env proxy handling.
- I still do not see tests for the proxy-selection behavior. Please add focused unit coverage that env proxy settings suppress system-proxy injection, and that logged proxy URLs redact userinfo.

One small robustness note while you are touching this: the redaction code builds `parsed.hostname + ...`; if `getproxies()` ever returns a malformed or hostless URL, that can raise. It would be better to tolerate that path and log only scheme/host/port when a hostname is present.

JerrettDavis (changes requested):
Thanks for the update. The earlier env-proxy precedence and userinfo-redaction coverage is much better now, but there is still one malformed-proxy case that can crash during logging.

In `headroom/proxy/ssl_context.py`, `_log_detected_system_proxy()` calls `parsed.port` while constructing the redacted netloc. `urllib.parse` raises `ValueError` for malformed ports, for example `http://proxy.example:notaport`, so a bad system proxy/PAC value can still take down proxy startup even though the docstring says malformed values fall back to scheme-only logging. I reproduced this directly with `_log_detected_system_proxy('http://proxy.example:notaport')`, which raises `ValueError: Port could not be cast to integer value as 'notaport'`.

Please handle `ValueError` from `parsed.port` and add a regression test alongside `test_hostless_url_does_not_raise`. After that, the previous blockers look addressed.

JerrettDavis (changes requested):
The updated head still has the malformed-port crash from the prior review. In `headroom/proxy/ssl_context.py`, `_log_detected_system_proxy()` builds `netloc` with `parsed.port`; `urllib.parse.ParseResult.port` raises `ValueError` for inputs like `http://proxy.example:notaport`, so a bad system proxy/PAC value can still crash startup during logging.

Please catch `ValueError` around port extraction and add the regression test requested earlier, e.g. `_log_detected_system_proxy('http://proxy.example:notaport')` or the `find_system_proxy()` path should not raise and should log a redacted/scheme-or-host fallback.

JerrettDavis (approved):
The malformed-port crash from the prior review is fixed: `_log_detected_system_proxy()` now catches `ValueError` from `parsed.port`, logs a safe host-only fallback, and the regression is covered alongside env-proxy precedence and credential redaction.

Focused local verification: `uv run pytest tests/test_ssl_context.py -q -k FindSystemProxy`.

Local note: a full `tests/test_ssl_context.py` run in this fresh worktree still has unrelated environment/dependency failures around CA cert store counts and missing `urllib3`; the focused tests for this PR's system-proxy behavior pass.

PR #1452, opened 2026-06-26
feat(bedrock): support inference-profile ARNs and forward anthropic-beta (1M context)

KennethWKZ:
## Description

Two fixes on the LiteLLM backend so AWS Bedrock works with the request shape Claude Code actually sends, plus the deps and compose overlays to run it.

When `ANTHROPIC_MODEL` is a full inference-profile ARN, requests failed with `litellm.BadRequestError: LLM Provider NOT provided` (provider=None). And on the Bedrock path the 1M context window never engaged because the client's `anthropic-beta` header was dropped before the upstream call.

Closes #

## Type of Change

## Changes Made

- **`map_model_id` ARN routing** (`headroom/backends/litellm.py`): route full inference-profile ARNs through the `bedrock/converse/` prefix (application- and system-defined, regional or `global.`). A bare ARN contains `/`, so it previously fell through to the generic passthrough and reached LiteLLM unprefixed. Also strips a trailing `[1m]` context-window marker if a client forwards it on the model id.
- **Forward `anthropic-beta`** (`_apply_anthropic_beta`): pass the client's beta header into all four `acompletion` calls. LiteLLM's converse transform reads beta features (notably `context-1m-2025-08-07`) from a `headers` kwarg; without this the 1M window silently capped at 200k. LiteLLM filters provider-unsupported betas, so forwarding the raw value is safe.
- **Deps** (`pyproject.toml`, `uv.lock`): add `botocore[crt]` to the `bedrock` extra (required by the AWS console-login / SSO credential provider, which raises `MissingDependencyException` without awscrt) and include `bedrock` in the `all` extra.
- **Compose overlays**: `docker-compose.bedrock.yml` (AWS Bedrock, mounts host `~/.aws` read-only) and `docker-compose.provider.yml` (generic Anthropic-compatible upstream — GLM/z.ai, DeepSeek, etc. via `.env`). Base `docker-compose.yml` is unchanged; both overlays are opt-in. New vars documented in `.env.example`.

## Testing

### Test Output

\`\`\`text
$ uv run --extra dev --with litellm pytest tests/test_bedrock_region.py -q
tests/test_bedrock_region.py ...................................         [100%]
35 passed in 1.32s

$ uv run --extra dev ruff check headroom/backends/litellm.py tests/test_bedrock_region.py
All checks passed!
\`\`\`

11 new cases cover ARN mapping (application/system/global ARNs, `[1m]` strip, converse passthrough) and `anthropic-beta` forwarding (forward, no-op, preserve-existing, title-case key).

## Real Behavior Proof

- **Environment:** macOS, Python 3.13, AWS Bedrock `ap-southeast-1`, credentials via `aws login` (console-login / SSO), Claude Code via local proxy (`ANTHROPIC_BASE_URL`, no `CLAUDE_CODE_USE_BEDROCK`).
- **Exact steps:** ran the proxy with `--backend bedrock --bedrock-region ap-southeast-1`; sent a request with `ANTHROPIC_MODEL` set to a full `application-inference-profile` ARN (and separately a `global.` inference-profile ARN).
- **Observed result:** before the fix, HTTP 500 `LLM Provider NOT provided`. After, HTTP 200 with a real Bedrock completion. Global inference-profile ARN also returns 200. GLM/DeepSeek via the provider overlay return 200 against their Anthropic-compatible endpoints.
- **Not tested:** a true >200k-token request to physically confirm the 1M window extends end-to-end (mechanism traced and unit-tested: LiteLLM's converse transform accepts `context-1m` from the forwarded header).

## Review Readiness

## Checklist

## Additional Notes

- `mypy` not run locally; happy to address any type findings from CI.
- CHANGELOG.md left untouched — let me know if you'd like an entry.
- The Bedrock overlay runs the container as root with `HOME=/home/nonroot` only because it bind-mounts the host `~/.aws`; the base and provider overlays use a named volume and need neither.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,17 @@
 # Copy this file to .env and fill in real values before running in production.
 # IMPORTANT: Change NEO4J_AUTH before deploying — default credentials are insecure.
 NEO4J_AUTH=neo4j/CHANGEME
+
+# --- docker-compose.provider.yml (GLM / DeepSeek / any Anthropic-compatible) ---
+# Upstream endpoint Headroom forwards to. Examples:
+#   GLM (z.ai):  https://api.z.ai/api/anthropic
+#   DeepSeek:    https://api.deepseek.com/anthropic
+# HEADROOM_PROVIDER_TARGET_URL=https://api.z.ai/api/anthropic
+# HEADROOM_PROVIDER_PORT=8788
+
+# --- docker-compose.bedrock.yml (AWS Bedrock) ---
+# AWS_REGION is required by the bedrock overlay. AWS_PROFILE defaults to "default".
+# Credentials come from the host ~/.aws mount (static keys, aws sso login, or aws login).
+# AWS_REGION=us-east-1
+# AWS_PROFILE=default
+# HEADROOM_BEDROCK_PORT=8789
--- a/docker-compose.bedrock.yml
+++ b/docker-compose.bedrock.yml
@@ -1,0 +1,65 @@
+# Overlay: compress traffic to AWS Bedrock (Claude models) with Headroom.
+#
+# Unlike GLM/DeepSeek (Anthropic-compatible — see docker-compose.provider.yml),
+# Bedrock is a different API: SigV4 auth + the converse request shape. Headroom
+# translates and signs via its litellm backend (HEADROOM_BACKEND=bedrock), so
+# the proxy itself needs working AWS credentials.
+#
+# Credentials: this overlay bind-mounts your host ~/.aws into the container so
+# botocore can use whatever you already have — static keys, `aws sso login`, or
+# `aws login`. Because that bind mount is owned by your host user, the proxy
+# runs as root (user "0:0") and HOME points at the mount root so botocore finds
+# ~/.aws there. (The base/provider overlays don't need this — they use a named
+# volume with no host-UID coupling.)
+#
+# Prerequisites:
+#   - Build with the bedrock extra so boto3 + botocore[crt] are present:
+#       HEADROOM_EXTRAS=proxy,code,bedrock   (or build the image with [all])
+#   - Valid AWS credentials on the host (`aws sts get-caller-identity` works).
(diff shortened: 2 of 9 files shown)


JerrettDavis (changes requested):
This overlaps with #1456 on the ARN-to-bedrock/converse/ mapping, but I do not think this version is ready as-is.

The docs/compose guidance tells users to put full Bedrock inference-profile ARNs in ANTHROPIC_MODEL while running Claude Code in Anthropic mode. That conflicts with the failure mode described in the other Bedrock PR and in the field reports: Claude Code can treat ARN model IDs as Bedrock mode input and bypass ANTHROPIC_BASE_URL, which means Headroom sees no traffic. Please revise the setup docs to use a verified proxy-preserving model configuration, and keep ARN handling scoped to cases where the request actually reaches Headroom.

Also, this PR mixes the backend ARN/beta forwarding fix with new compose overlays, dependency changes, and Anthropic pricing/context entries. Please split or justify those extra changes so the Bedrock routing fix can be reviewed and merged without taking unrelated operational surface in the same PR.

KennethWKZ:
Splitting this into focused PRs per the review feedback. The original branch bundled four unrelated concerns; here's where each landed:

**1. ARN → `bedrock/converse/` routing + named AWS profile + `au.` region → #1456**
Already covered more completely there (profile forwarding and the `ap-southeast-2 → au` prefix, which this PR lacked). Deferring to #1456 to avoid duplicating the same `map_model_id` branch.

**2. `claude-opus-4-8` pricing/context → #1485**
Unrelated data entry; separated out.

Happy to drop/close this PR once #1485 is the path forward. Let me know if you'd rather fold the beta fix into #1456 instead.

KennethWKZ:
Closing this omnibus PR — superseded. ARN routing/profile/region covered by #1456; pricing entry split out to #1485; the anthropic-beta (1M) piece (#1484) was withdrawn since Opus 4.8 has 1M by default.

Issue #419, opened 2026-05-07
headroom-openclaw breaks OpenClaw runs with Cannot read properties of undefined (reading 'split')


jkobject:
## Summary

The `headroom-openclaw` integration appears to successfully start/register and can compress OpenClaw contexts, but enabling it as OpenClaw's `contextEngine` caused OpenClaw agent runs to fail with:

```text
Cannot read properties of undefined (reading 'split')
```

This happened reproducibly enough that I had to disable the plugin in production. Headroom did show real savings (~30% on a large tool-heavy context), so this looks worth fixing rather than just disabling.

## Environment

- OpenClaw: `2026.5.6` (`c97b9f7`)
- Runtime: Node `v24.14.0`
- OS: Linux x64 (`6.14.0-34-generic`)
- OpenClaw model/provider during failing runs: OpenAI / `gpt-5.5`
- `headroom-openclaw`: `0.1.0` from npm
- `headroom-ai` JS package: `0.1.0` from npm
- Python Headroom package: `0.5.18`
- Python: `3.11.15`
- Plugin source path shape: `~/.openclaw/extensions/headroom/dist/index.js`

## Config shape used

Initially enabled via the OpenClaw plugin config roughly as:

```jsonc
{
  "plugins": {
    "slots": {
      "contextEngine": "headroom"
    },
    "allow": ["headroom"],
    "entries": {
      "headroom": {
        "enabled": true,
        "config": {
          "pythonPath": "~/.openclaw/extensions/headroom/.venv/bin/python"
        }
      }
    }
  }
}
```

I also later staged a safer sidecar-oriented config (not left enabled in prod):

```jsonc
{
  "enabled": true,
  "config": {
    "pythonPath": "~/.openclaw/extensions/headroom/.venv/bin/python",
    "autoStart": false,
    "proxyUrl": "http://127.0.0.1:8787"
  }
}
```

### 1. Plugin registers, but OpenClaw logs a plugin contract warning

```text
[headroom] Plugin registered
[plugins] plugin must declare contracts.tools before registering agent tools (plugin=headroom, source=.../headroom/dist/index.js)
```

Not sure if this is directly related, but it seems suspicious because the plugin appears to register retrieval/tools without declaring the tool contract expected by recent OpenClaw versions.

### 2. Agent runs then fail with `.split` on undefined

OpenClaw embedded run log excerpt:

```json
{
  "event": "embedded_run_agent_end",
[...]

PR #1459, opened 2026-06-26
fix(openclaw): detect uv-installed headroom binary in ~/.local/bin

Ru0k3:
## Description
When `headroom-ai` is installed via `uv tool install headroom-ai`, the binary lands at `~/.local/bin/headroom`. The plugin's autoStart launcher detection did not find it because the PATH check used `sh -lc` which may not source user shell config (`.zshrc`, `.bash_profile`) on all systems. This PR adds explicit uv path detection and fixes the shell invocation flag.

Related to #419

## Type of Change
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Made
- Switch PATH check from `sh -lc` to `sh -c` in `proxy-manager.ts`
- Add explicit uv tool install detection checking `~/.local/bin/headroom`

## Testing

### Test Output
```text
Status: loaded
Version: 0.27.0
Capabilities:
  context-engine: headroom
Tools:
  headroom_retrieve
```

## Real Behavior Proof
- Environment: EndeavourOS x86_64, OpenClaw 2026.6.10, headroom-ai 0.27.0 via uv
- Exact command / steps: `uv tool install headroom-ai` then `openclaw gateway restart`
- Observed result: Before fix — autoStart failed with "Headroom proxy not detected on default endpoints" even though binary exists at `~/.local/bin/headroom`. After fix — plugin loads and connects correctly.
- Not tested: Windows, Docker runtime

## Review Readiness

## Additional Notes
Reproduced while debugging the npm package staleness issue in #419. Affects any user following the standard `uv tool install` workflow on Linux/macOS.

### DIFF
--- a/plugins/openclaw/src/proxy-manager.ts
+++ b/plugins/openclaw/src/proxy-manager.ts
@@ -258,12 +258,27 @@ export class ProxyManager {
       checkCommand: process.platform === "win32" ? "where.exe" : "sh",
       checkArgs: process.platform === "win32"
         ? ["headroom"]
-        : ["-lc", "command -v headroom >/dev/null 2>&1"],
+        : ["-c", "command -v headroom >/dev/null 2>&1"],
       useShell: process.platform === "win32",
       checkUseShell: false,
     });
 
-    // 4) Local npm install (inside plugin install path)
+// 4) uv tool install path (~/.local/bin/headroom)
+    const uvBin = join(
+      process.env["HOME"] ?? "",
+      ".local", "bin", "headroom"
+    );
+    if (existsSync(uvBin)) {
+      specs.push({
+        label: `uv tool: ${uvBin}`,
+        command: uvBin,
+        args: commonArgs,
+        checkCommand: uvBin,
+        checkArgs: ["--version"],
+      });
+    }
+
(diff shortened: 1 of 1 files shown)


JerrettDavis (approved):
Reviewed the OpenClaw launcher detection change. The explicit `~/.local/bin/headroom` probe is the important fix for `uv tool install`, and the head checks are green.

Non-blocking: the new uv-path comment lost the surrounding indentation, and `PathLike` construction via `HOME` is Unix-focused. That matches the stated Linux/macOS uv case, but using `os.homedir()` in a future cleanup would make the intent clearer across hosts.

Ru0k3:
Thanks for the feedback! Updated to use `os.homedir()` and fixed the comment indentation.

JerrettDavis (approved):
The follow-up addresses my earlier concern. Using `os.homedir()` avoids the missing-`HOME` edge case, and the added uv-tool path is only used when the executable exists, so it should not disturb the existing PATH/local npm candidates. I do not see a remaining blocker.

Ru0k3:
Should I raise an issue and make new PR?

PR #1464, opened 2026-06-26
fix(wrap): detach the shared proxy on Windows so it survives an ungraceful agent close

quentinmaisonneuve:
## Description

Closing one `headroom wrap <agent>` instance on Windows could kill the **shared proxy** out from under every other running instance, so their requests started failing.

`_start_proxy` launched the proxy as a child of whichever agent started it first, without detaching it from that agent's console and Job object. The wrapper already reference-counts clients via per-PID markers and `_make_cleanup` leaves the proxy running while other clients exist — but that only runs on a *graceful* exit. On an *ungraceful* close (closing the terminal window, `taskkill`, a crash) Windows tree-kills the whole process group/Job and the proxy dies directly, bypassing the reference counting. Every other instance's `ANTHROPIC_BASE_URL` then points at a dead `127.0.0.1:8787`, so all of its API traffic fails.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `_start_proxy` creates the proxy with `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP | CREATE_BREAKAWAY_FROM_JOB` on Windows, so an ungraceful close of the launching agent can no longer reach it; only the ref-counted `_make_cleanup` ends the proxy.
- Falls back without `CREATE_BREAKAWAY_FROM_JOB` (catching `OSError`) when the launcher's Job forbids breakaway; `DETACHED_PROCESS` still spares the proxy from console-close events.
- Platform guard is `sys.platform == "win32"` (not `os.name == "nt"`) so mypy narrows the platform and resolves the Windows-only `subprocess` constants.
- POSIX path unchanged: `creationflags=0`, detachment still via `start_new_session` (`setsid`).
- Added `tests/test_cli/test_wrap_proxy_detach.py` and a CHANGELOG Bug Fixes entry.

## Testing

### Test Output

```text
$ pytest tests/test_cli/test_wrap_proxy_detach.py -q
..                                                                       [100%]
2 passed, 2 warnings in 1.50s

$ ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_proxy_detach.py
All checks passed!

$ mypy --follow-imports=silent headroom/cli/wrap.py tests/test_cli/test_wrap_proxy_detach.py
Success: no issues found
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.11.9, headroom-ai (pipx). Two concurrent `headroom wrap claude` instances sharing proxy `127.0.0.1:8787`. `_start_proxy` was also exercised directly on this host with `subprocess.Popen` stubbed.
- Exact command / steps: (1) start two `headroom wrap claude` instances; (2) close the terminal window of the one that started the proxy (ungraceful — not `/exit`); (3) issue a request from the surviving instance. Separately: call `_start_proxy(8787)` with `subprocess.Popen` stubbed and read back the creation flags.
- Observed result: before the fix the proxy died with the closed window and the surviving instance failed (`ANTHROPIC_BASE_URL` → dead `:8787`), because the OS tree-killed the child before the ref-count path could spare it. After the fix the detached proxy survives the close and the surviving instance keeps working; the stub harness reports `creationflags=0x1000208` (`DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP | CREATE_BREAKAWAY_FROM_JOB`) on win32 and `0` when forced off-Windows.
- Not tested: real breakaway behavior under an actual restrictive Job object on this host (the OS-level effect). The `OSError` fallback path itself now has a dedicated unit test (`test_start_proxy_retries_without_breakaway_when_job_forbids_it`).

## Review Readiness

## Checklist

N/A — no UI changes.

## Additional Notes

- Documentation checklist item is N/A: this is a behavioral bug fix with no user-facing doc surface.
- Scope is the single `subprocess.Popen` call in `_start_proxy`; the marker-based reference counting in `_make_cleanup` is unchanged and remains the only thing that intentionally stops the proxy.


JerrettDavis (approved):
Reviewed the Windows detachment path, the POSIX branch, and the new tests. The core change is scoped to proxy startup and the head checks are green.

Non-blocking: the PR body says the `OSError` fallback is unit-tested, but the added tests only assert the win32/off-win32 creation flags. A focused test that makes the first `Popen` raise and asserts the retry drops `CREATE_BREAKAWAY_FROM_JOB` would strengthen this, but I do not think it should block the fix.

quentinmaisonneuve:
Added `test_start_proxy_retries_without_breakaway_when_job_forbids_it` (f9785d4f): it makes the first `Popen` raise `OSError` and asserts the retry drops `CREATE_BREAKAWAY_FROM_JOB` while keeping `DETACHED_PROCESS` and `CREATE_NEW_PROCESS_GROUP`. The PR body's "unit-tested" claim is now accurate. Thanks for the review.

JerrettDavis (approved):
Re-reviewed after the fallback-test follow-up. The added test now exercises the OSError path where CREATE_BREAKAWAY_FROM_JOB is rejected and verifies the retry keeps the console/process-group detachment while dropping only the breakaway flag. The production behavior remains scoped to _start_proxy and POSIX stays unchanged. No blocking issues found.

quentinmaisonneuve:
## Follow-up: `DETACHED_PROCESS` → `CREATE_NO_WINDOW`, plus a rebase on `main`

While testing this branch locally on Windows I hit a regression in the original approach, so I pushed a follow-up commit and rebased.

### Regression found
`DETACHED_PROCESS` does isolate the proxy from the launcher's console, but for a **console-subsystem executable (`python.exe`) it leaves the proxy with no console at all**. Windows then allocates and *shows* a visible `python.exe` console window for it. Closing that stray window kills the proxy — which re-introduces the exact orphaned-proxy failure this PR set out to prevent, now triggered by a window most users will instinctively close.

### Fix
Switched the Windows creation flags from `DETACHED_PROCESS` to `CREATE_NO_WINDOW`, which gives the proxy its **own, *hidden* console**. A separate console still means the launching agent's `CTRL_CLOSE_EVENT` never reaches the proxy, but no window is ever shown. The three console flags (`CREATE_NO_WINDOW` / `DETACHED_PROCESS` / `CREATE_NEW_CONSOLE`) are mutually exclusive, so this *replaces* `DETACHED_PROCESS` rather than adding to it. `CREATE_NEW_PROCESS_GROUP`, `CREATE_BREAKAWAY_FROM_JOB` and the `OSError` breakaway fallback are unchanged; POSIX behavior is unchanged.

Tests updated to assert `CREATE_NO_WINDOW` on both the happy path and the breakaway fallback.

### Rebase
Rebased onto current `main` to clear the merge conflict (`CHANGELOG.md` only — both sides added entries under the unreleased *Bug Fixes* section; kept all of them).

### Verification (Windows 11)
- Reproduced the visible `python.exe` window with `DETACHED_PROCESS`; confirmed it is gone with `CREATE_NO_WINDOW`, and the proxy survives closing the launching terminal.
- `ruff check` — clean
- `mypy headroom` — `Success: no issues found in 404 source files`
- `pytest tests/test_cli/test_wrap_proxy_detach.py` — `3 passed`

@JerrettDavis a quick re-review would be appreciated since the flag changed.

JerrettDavis (changes requested):
The Windows creation flags and CREATE_NO_WINDOW update make sense, and the breakaway fallback is covered. One remaining issue is resource cleanup on the failure path: _start_proxy() opens stdio_log_file, then both subprocess.Popen attempts can raise (first with breakaway, second without). In that case the file handle is never closed because the existing close calls live after a successful spawn / readiness loop. Please wrap the spawn+readiness block in try/finally or otherwise guarantee the parent closes the stdio log file if both spawn attempts fail. This also lines up with the fd-cleanup work in #1556, so rebasing after that may be the simplest route.

quentinmaisonneuve:
Fixed in 117b6872. Wrapped the spawn + readiness block in `try/finally` and moved `stdio_log_file.close()` into the `finally`, so the parent's handle is released on every exit path: successful readiness, proxy-died, timeout, and the case you flagged where both `Popen` attempts raise. Removed the three scattered `close()` calls now that the `finally` covers them. The child keeps its own inherited duplicate of the handle, so closing in the parent doesn't starve the proxy's logging — same behavior as before, just guaranteed on the failure path too.

Added a regression test (`test_start_proxy_closes_log_when_both_spawns_fail`) that forces both spawn attempts to raise and asserts the handle is closed.

On #1556: happy to rebase on the fd-cleanup work once it lands if you'd prefer that route — let me know.

JerrettDavis (approved):
Reviewed the latest update. This clears the prior resource-leak blocker.

The Windows detach flags are still scoped to the proxy subprocess launch path, and the log handle is now closed from a single `finally` that covers success, timeout, early process exit, and both-spawn-failure paths. The breakaway retry also keeps the hidden-console/process-group protections when it has to drop `CREATE_BREAKAWAY_FROM_JOB`.

Local check: `python -m pytest tests/test_cli/test_wrap_proxy_detach.py -q` passed (4 tests). CI is green.

PR #1467, opened 2026-06-26
ci: add Windows smoke test — surfaces compatibility issues

### LINKED ISSUE #1466: Windows compatibility — Headroom not working on Windows
## Report

From Discord user **tommy**: "Headroom is not working on windows, that's why it would not work on my pc"

## Current state

Headroom has some Windows support code:
- `headroom/install/paths.py` — `windows_run_script_path`, `windows_ensure_cmd_path`
- `headroom/install/supervisors.py` — `_is_windows()`, `_render_windows_runner()`
- `headroom/learn/plugins/claude.py` — `PureWindowsPath` decoding

But no CI tests on Windows, no Windows build artifacts, and likely broken paths.

## What needs investigation

1. Does `headroom proxy` start on Windows?
2. Does `headroom wrap claude` work on Windows?
3. Are the supervisor paths correct for Windows (no `/tmp`, different home dir)?
4. Do the MCP registry paths work on Windows?
5. Is RTK/lean-ctx binary available for Windows?

## Suggested fix

- Add Windows to CI matrix (GitHub Actions `windows-latest`)
- Test proxy startup + wrap on Windows
- Fix any path issues (Windows uses `\` not `/`, different temp dir)
- Document Windows-specific setup

## Related

Discord context: user tommy reporting headroom not working on Windows.

peterlodri-sec:
## Problem

Discord user **tommy** reports Headroom not working on Windows. We have Windows support code but no CI.

## Solution

Add `windows-smoke.yml` — runs `headroom --version`, `headroom proxy --help`, `headroom wrap --help` on `windows-latest`.

Closes #1466

### DIFF
--- a/.github/workflows/windows-smoke.yml
+++ b/.github/workflows/windows-smoke.yml
@@ -1,0 +1,46 @@
+name: windows-smoke
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'headroom/install/**'
+      - 'headroom/paths.py'
+      - 'headroom/cli/**'
+  pull_request:
+    paths:
+      - 'headroom/install/**'
+      - 'headroom/paths.py'
+      - 'headroom/cli/**'
+  workflow_dispatch:
+
+jobs:
+  smoke:
+    runs-on: windows-latest
+    timeout-minutes: 10
+    
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+          cache: 'pip'
+
+      - name: Install package
+        run: pip install -e .
+
+      - name: Verify CLI Version
+        run: headroom --version
+
+      - name: Verify Proxy Command
+        run: headroom proxy --help
+
+      - name: Verify Wrap Command
+        run: headroom wrap --help


JerrettDavis (changes requested):
Thanks for adding this. I think this should run for a slightly broader set of changes before we rely on it as Windows coverage.

Right now both `push` and `pull_request` are path-filtered to `headroom/install/**`, `headroom/paths.py`, and `headroom/cli/**`. That means the smoke test does not run for this workflow change itself, and it will also miss changes that can break `pip install -e .` or console-script startup on Windows, such as `pyproject.toml`, lock/dependency changes, workflow edits, package metadata, and shared CLI imports outside `headroom/cli/**`.

Please expand the path filters to include the workflow file plus packaging/dependency/config files that affect installation and entry-point startup, or remove the path filters for PRs if the intent is broad Windows smoke coverage. The actual smoke commands look appropriately scoped once the workflow is reliably triggered.

JerrettDavis:
Marking this as superseded by the Windows coverage now on `main`. The proposed workflow only runs `pip install -e .` and three CLI `--help`/version checks. Current CI instead builds the actual Rust-backed `win_amd64` wheel on `windows-latest` (`build-wheel-windows`, introduced through #1086), runs Windows native installer-wrapper tests, and the release workflow builds and validates the published Windows wheel matrix. Those landed gates exercise the distributable artifact and native integration more completely than this editable-install smoke.

PR #1471, opened 2026-06-26
feat(creative-tui): minimal native shader canvas + TUI shell + Nix

peterlodri-sec:
### Architecture (13 modules, 945 LOC)
```
types.rs     — data layer (Speed, Rgb, Command, ShellResponse, LogBuffer, Uniforms)
shell.rs     — pure command dispatcher (zero IO)
gfx/         — GPU layer (resources::init → pipeline::build → renderer::Renderer)
tui/         — terminal layer (TerminalUi setup/draw/teardown + input polling)
app.rs       — orchestrator (wires channels, spawns threads)
main.rs      — 17-line bootstrap
```

### Type tightening
- `Speed` newtype with range `[0.1, 10.0]` validated in constructor
- `Rgb` newtype with per-channel `[0.0, 3.0]` validated in constructor
- `LogBuffer` ring buffer — zero-allocation per-frame view, no cloning
- `Command` + `ShellResponse` enums — no raw strings in channels

### Hotspot fix
`GpuResources::new()` was 111-line monolithic init. Split into:
- `resources::init()` — async adapter + device + queue + buffer
- `pipeline::build_pipeline()` — sync shader + bind group + pipeline
- `renderer::Renderer::new()` — async orchestrator, owns everything

### Run
```bash
cd creative_tui && cargo run
# TUI commands: speed N | color R G B | time
```

### Tests
6 unit tests pass — command parsing + dispatch + range clamping


JerrettDavis (changes requested):
Thanks for putting this together. This needs a few fixes before it is ready for enterprise review/merge:

- `creative_tui/src/app.rs` consumes Enter before the submit path can see it. `poll_input()` reads the key event first, and `creative_tui/src/tui/input.rs` maps `KeyCode::Enter` to `None`; the later "Enter key handling" block polls for a second event, so a normal single Enter press is dropped and commands never submit. Please make submit a single input path, for example by having the input poller return `KeyAction::Submit` or by handling Enter before consuming the event.

- `spawn_graphics()` starts the winit event loop on a spawned OS thread. With winit 0.30, macOS requires the event loop/window creation to happen on the main thread unless the platform-specific any-thread extension is deliberately enabled and tested. As written this is likely to panic on macOS, which is one of our supported development platforms. Please restructure the threading model or add the appropriate platform handling and coverage.

- The new `creative_tui` crate is not wired into the root Cargo workspace or CI. The only checks on this PR are template/label checks, so we have no automated `cargo fmt --check`, `cargo test`, or `cargo check`/build signal for the new Rust code. Please either add it to the workspace/CI or add an explicit CI job for this standalone crate.

Also, the PR body still has the governance warning because the required sections and readiness checkboxes are missing. Please fill those in with the actual test/build commands and results once the above is addressed.

JerrettDavis:
Thanks for sharing the shader and terminal experiment. This is a standalone creative-coding application, with no integration into Headroom’s proxy, compression, or SDK workflows. I’m closing it to keep this repository’s review queue focused on those workflows; the demo would be easier to develop and maintain as its own project.

PR #1472, opened 2026-06-26
Resolve MCP server command to absolute path

gglucass:
## Description

Claude Code spawns the Headroom MCP server as a subprocess with the minimal `PATH` a GUI app inherits, which excludes the venv `bin` dir where the `headroom` console script lives. The Claude registrar wrote the server entry with a bare `command: "headroom"`, so the client cannot exec it: the server shows up as **failed** in `/mcp` with no way to reconnect, and the memory tools (`memory_search` / `memory_save`) never appear.

This resolves the command to the absolute path of the running `headroom` executable, in the Claude registrar only. Codex (a terminal flow where `headroom` is reliably on PATH) is intentionally left unchanged.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `_resolve_command()` in `headroom/mcp_registry/claude.py`: rewrites a bare `headroom` command to the absolute path of the running executable (the console script next to `sys.executable`), falling back to `argv[0]`, then `shutil.which("headroom")`, then the bare name. Applied at the top of `ClaudeRegistrar.register_server` so both the CLI (`claude mcp add`) and direct-JSON write paths emit the resolved path.
- Update `tests/test_mcp_registry/test_claude_registrar.py` to assert the resolved command (derived via `_resolve_command`, so it is environment-independent).
- Update `e2e/init/run.py` to compare the MCP server command by basename rather than the hardcoded bare name.

## Testing

### Test Output

```text
$ pytest tests/test_mcp_registry/ tests/test_cli/test_wrap_codex.py -q
136 passed in 5.69s

$ ruff check headroom/mcp_registry/claude.py tests/test_mcp_registry/test_claude_registrar.py e2e/init/run.py
All checks passed!

$ mypy headroom/mcp_registry/claude.py
```

## Real Behavior Proof

- Environment: macOS, Python 3.12, Headroom installed under `~/Library/Application Support/Headroom/.../venv`.
- Exact command / steps: inspected the registered entry in `~/.claude/mcp.json` (`command: "headroom"`); confirmed `command -v headroom` returns nothing even in a login shell, since the binary lives only in the venv bin dir; reproduced the `/mcp` "failed" state in Claude Code. The init e2e confirms the resolved command lands as `/opt/headroom-venv/bin/headroom` in the `claude mcp add` invocation.
- Observed result: with the fix, the Claude MCP entry's command is the absolute path to the venv's `headroom`, which the client can exec regardless of inherited PATH.
- Not tested: end-to-end relaunch of a real Claude Code session against a freshly built release bundle (verified at unit + init-e2e level).

## Review Readiness

## Checklist

## Additional Notes

- Documentation / CHANGELOG: N/A — internal registrar behavior, no user-facing API change.
- Scoped to the Claude registrar deliberately: `build_headroom_spec` (shared with the Codex TOML writer) is unchanged, so Codex wrap output and its e2e are unaffected.
- Follow-up (out of scope): the registrar can still write to the legacy `~/.claude/mcp.json`, which Claude Code >=2.x does not read. That path-vs-file issue is separate from this command-resolution fix.

### DIFF
--- a/e2e/init/run.py
+++ b/e2e/init/run.py
@@ -22,6 +22,7 @@ from ``e2e/init/Dockerfile``).
 from __future__ import annotations
 
 import json
+import os
 import sys
 from pathlib import Path
 
@@ -111,6 +112,13 @@ def _verify_claude_local(ctx: CaseContext) -> None:
     # being dead pointers for users who never ran `headroom mcp install`).
     # The `-e HEADROOM_PROXY_URL=…` arg is only emitted when the proxy
     # port differs from the 8787 default; this case uses --port 9011.
+    # The MCP server command is resolved to the absolute path of the running
+    # `headroom` executable so GUI clients with a minimal PATH can spawn it, so
+    # compare that segment by basename rather than the full venv path.
+    normalized = [
+        [os.path.basename(arg) if arg.endswith("/headroom") else arg for arg in argv]
+        for argv in claude_calls
+    ]
     expected = [
         ["plugin", "marketplace", "add", str(REPO_ROOT_IN_CONTAINER)],
         ["plugin", "install", "headroom@headroom-marketplace", "--scope", "local"],
@@ -128,7 +136,7 @@ def _verify_claude_local(ctx: CaseContext) -> None:
             "serve",
         ],
     ]
-    if claude_calls != expected:
+    if normalized != expected:
         raise AssertionError(f"Unexpected Claude install commands: {claude_calls}")
 
 
--- a/headroom/mcp_registry/claude.py
+++ b/headroom/mcp_registry/claude.py
@@ -10,17 +10,44 @@ available, and reads the underlying JSON files directly for compare /
 
 from __future__ import annotations
 
+import dataclasses
 import json
 import logging
 import os
 import shutil
 import subprocess
+import sys
 from pathlib import Path
 from typing import Any
 
 from .base import MCPRegistrar, RegisterResult, RegisterStatus, ServerSpec
 
 logger = logging.getLogger(__name__)
+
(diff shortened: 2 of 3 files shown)


JerrettDavis (changes requested):
This is the right direction for GUI-launched MCP servers, but it still has two portability gaps that need closing before merge:

1. On Windows the console script is `headroom.exe`, so `Path(sys.executable).with_name("headroom")` will miss the bundled executable in the minimal-PATH case this PR is meant to fix. The fallback to `shutil.which("headroom")` does not help when the client PATH is stripped.
2. The updated default test only accepts `/headroom`, so it is POSIX-specific and matches the failing Windows test job. It should use `Path(spec.command).name` / platform-aware suffix handling instead of a hard-coded slash.

Please resolve both and update the commit subject to conventional format as well, since `commitlint` is currently failing on `Resolve MCP server command to absolute path`.

gglucass:
Closing this as superseded. `main` already fixes this exact bug via `resolve_headroom_command()` (in `headroom/install/runtime.py`), which `build_headroom_spec` now uses:

```python
def resolve_headroom_command() -> list[str]:
    headroom_bin = shutil.which("headroom")
    if headroom_bin:
        return [headroom_bin]
    return [sys.executable, "-m", "headroom.cli"]
```

That handles the GUI minimal-PATH case more robustly than this PR (the `sys.executable -m headroom.cli` fallback always exists, and it covers Codex and Windows too, addressing the portability point raised in review). Every file this PR touched was already modified by that work, so there's no meaningful merge. Thanks for the review @JerrettDavis.

PR #1456, opened 2026-06-26
fix(bedrock): route ARNs via converse, named AWS profiles, and au. re…

mhaitana:
## Description

Fix three related gaps in Bedrock support that prevented headroom from working with Claude Code when `CLAUDE_CODE_USE_BEDROCK=0` and `ANTHROPIC_BASE_URL` is pointed at the proxy:

1. **ARN passthrough used the wrong LiteLLM route** — application inference profile ARNs (e.g. `arn:aws:bedrock:ap-southeast-2:<account>:application-inference-profile/<id>`) were forwarded as `bedrock/<arn>`, which LiteLLM rejects with HTTP 400 "Try calling via converse route". Fixed to `bedrock/converse/<arn>`.

2. **Named AWS profile not forwarded to completion calls** — `--bedrock-profile` was wired through the CLI → config → `LiteLLMBackend.__init__` and used to fetch the model map at startup,
but never stored on `self`. All four `acompletion()` call sites (`send_message`, `stream_message`, `send_openai_message`, `stream_openai_message`) passed only `aws_region_name` — the
actual Bedrock calls used ambient credentials regardless of the flag. Fixed by storing `self.profile_name` and passing `aws_profile_name=` to every `acompletion()` call.

3. **`ap-southeast-2` used the wrong region prefix** — Australia should use `au.` for cross-region inference profile IDs, not `apac.`. Added `ap-southeast-2 → "au"` to `_BEDROCK_REGION_PREFIXES` and `"au."` to the strip list in `_normalize_bedrock_profile_id`.

Closes #

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `backends/litellm.py`: route `arn:aws:` model IDs via `bedrock/converse/<arn>` in `map_model_id`
- `backends/litellm.py`: store `profile_name` as `self.profile_name` in `LiteLLMBackend.__init__`; pass `aws_profile_name=` to `acompletion()` in all four call sites; use
`boto3.Session(profile_name=...)` for startup discovery; cache key is `region:profile_name` to prevent cross-profile collisions
- `backends/litellm.py`: add `ap-southeast-2 → "au"` to `_BEDROCK_REGION_PREFIXES`; add `"au."` to prefix strip list in `_normalize_bedrock_profile_id`
- `providers/registry.py`: pass `profile_name=bedrock_profile` to `LiteLLMBackend`
- `proxy/server.py`: pass `config.bedrock_profile` to `create_proxy_backend`
[...]

PR #1484, opened 2026-06-27
fix(bedrock): forward anthropic-beta header to LiteLLM (1M context)

KennethWKZ:
## Description

The client's `anthropic-beta` header (notably `context-1m-2025-08-07`) is dropped before the LiteLLM `acompletion` call, so the 1M context window silently caps at 200k on the Bedrock path.

Headroom's LiteLLM backend extracts only auth (`x-api-key`/`authorization`) from the incoming request headers. Beta features negotiated by the client never reach the upstream.

## Root cause (verified against LiteLLM source)

LiteLLM *does* support forwarding `context-1m` to Bedrock:

1. `context-1m-2025-08-07` is in LiteLLM's per-provider allowlist (`anthropic_beta_headers_config.json`, `bedrock_converse` section).
2. LiteLLM's Bedrock converse transform reads betas from the `headers`/`extra_headers` kwarg via `litellm.anthropic_beta_headers_manager`.
3. **But** Headroom never populates that kwarg — so the header is lost before LiteLLM sees it.

Without this change → header dropped → 1M never engages (200k cap).
With this change → header reaches LiteLLM → allowed for `bedrock_converse` → forwarded to Bedrock.

> Note: not yet physically exercised with a >200k-token request, but the full path is traced and confirmed against LiteLLM's real allowlist. Independent of model-id routing — works for short IDs and ARNs alike.

## Type of Change

- [x] Bug fix (non-breaking)

## Changes Made

- Add `_apply_anthropic_beta(kwargs, headers)` in `headroom/backends/litellm.py`.
- Call it before all four `acompletion` sites: `send_message`, `stream_message`, `send_openai_message`, `stream_openai_message`.
- LiteLLM filters provider-unsupported betas, so forwarding the raw header value is safe.
- Add `TestApplyAnthropicBeta` (4 cases: forward, no-op, preserve-existing, title-case key).

Scoped to the beta-forwarding fix only. ARN→`bedrock/converse/` routing and named-profile forwarding are handled in #1456; this is complementary and touches neither.

## Testing

\`\`\`text
$ uv run --extra dev --with litellm pytest tests/test_bedrock_region.py -q
31 passed in 1.36s

$ uv run --extra dev --with litellm ruff check headroom/backends/litellm.py tests/test_bedrock_region.py
All checks passed!
\`\`\`

## Checklist

- [x] Self-review performed
- [x] New unit tests added
- [x] New + existing unit tests pass locally
- [x] Linting passes

### DIFF
--- a/headroom/backends/litellm.py
+++ b/headroom/backends/litellm.py
@@ -383,6 +383,20 @@ def _parse_tool_arguments(arguments: Any) -> Any:
     return arguments
 
 
+def _apply_anthropic_beta(kwargs: dict[str, Any], headers: dict[str, str]) -> None:
+    """Forward the client's ``anthropic-beta`` header into the LiteLLM call.
+
+    Without this, beta features negotiated by the client (notably the 1M
+    context window ``context-1m-2025-08-07``) never reach the upstream — the
+    Bedrock converse transform reads them from a ``headers`` kwarg, and
+    LiteLLM otherwise drops them. LiteLLM filters betas the provider does not
+    support, so passing the raw value through is safe.
+    """
+    beta = headers.get("anthropic-beta") or headers.get("Anthropic-Beta")
+    if beta:
+        kwargs.setdefault("headers", {})["anthropic-beta"] = beta
+
+
 class LiteLLMBackend(Backend):
     """Backend using LiteLLM for multi-provider support.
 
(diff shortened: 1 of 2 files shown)


KennethWKZ:
Closing — verified against the official Claude Platform docs that Opus 4.8 has the 1M-token context window by default (no `context-1m` beta header required): https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-8 . The 1M motivation for this change is therefore moot on the current model.

The underlying header-forwarding gap (Bedrock litellm path drops the client `anthropic-beta` header, which also carries compaction / context-management betas) may still be worth addressing separately if compaction misbehaves on Bedrock, but that's a distinct, unverified concern — not the 1M fix this PR was framed around.

PR #1506, opened 2026-06-27
fix(core): avoid unidiff panic on bash xtrace

mstattma:
## Summary
- preflight unified diff inputs before calling the Rust unidiff parser
- catch parser panics so malformed-but-diff-looking input falls back to non-diff
- add regressions for Bash xtrace lines like `+++ test.sh` and `+++ dirname test.sh`

## Repro
`detect_content_type("+++ test.sh")` could panic through the Rust detector because unidiff treated the lone `+++` line as a target header without a preceding source header.

## Tests
- `cargo fmt --all --check`
- `cargo test -p headroom-core --lib transforms::unidiff_detector::tests`
- `cargo test -p headroom-core --lib`


JerrettDavis (approved):
This looks ready from the focused code review. The preflight `looks_like_unified_diff` check prevents bash xtrace `+++ ...` output from reaching the `unidiff` parser, and the `catch_unwind` boundary keeps any remaining parser panic from escaping the detector. The standard-diff path remains covered. I do not see a correctness blocker in the Rust change.

JerrettDavis (approved):
Re-reviewed current head after the main merge. The unidiff detector still filters bash xtrace-style `+++` lines before parser entry and keeps the panic boundary around the remaining parser path.

- `cargo test -p headroom-core unidiff --lib` -> 17 passed
- `cargo check -p headroom-core`

GitHub checks are partly queued on current head; no code blocker from this review.

PR #1514, opened 2026-06-27
feat(ccr): teach agents when retrieval adds no information

rodboev:
## Description

Adds CCR behavior to teach agents when retrieval adds no additional information, reducing repeated low-value retrieval attempts.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- Add CCR feedback for retrievals that do not add useful new information.
- Render the tool description, `query` description, system instructions, CLI guidance, and skill text from one policy module so every surface states the same retrieval contract.
- Classify the `query` hint rather than discarding it. A hint is specific when it names what is being checked, so it outranks generic thoroughness wording; a hint built only from wording that names the request itself is worth no more than an absent hint. That wording sits in four named registers: grammatical filler, degree of care, the retrieval action and the tool performing it, and the payload plus the form it is asked for. Locative nouns stay out of every register, so `record 12` and `line 42` stay specific.
- Keep the behavior scoped to retrieval guidance rather than changing unrelated proxy correctness.
- Include focused coverage for the retrieval/no-new-information path, both at the classifier and at the learn analyzer that consumes its verdict.

## Testing

- [x] Unit tests pass

### Test Output

```text
python -m pytest tests/test_ccr_retrieve_policy.py tests/test_learn/test_ccr_retrieve_literacy.py tests/test_ccr_mcp_server.py tests/test_ccr_tool_injection.py tests/test_cli/test_mcp.py tests/test_learn/test_writer.py tests/test_proxy_ccr.py tests/test_ccr_tool_always_on.py -q
173 passed, 4 skipped

ruff check . && ruff format . --check
All checks passed!
```

## Real Behavior Proof

- Environment: local checkout of this branch.
- Exact command / steps: `python -m pytest tests/test_ccr_retrieve_policy.py::test_classifier_prefers_specific_query_hint_over_thoroughness_wording tests/test_ccr_retrieve_policy.py::test_classifier_treats_retrieval_action_hints_as_nonspecific tests/test_ccr_retrieve_policy.py::test_classifier_treats_content_form_hints_as_nonspecific tests/test_learn/test_ccr_retrieve_literacy.py::test_analyzer_emits_rule_for_retrieval_action_query -v`
- Observed result: all four pass. `classify_retrieve_need("Verify the auth middleware result", query="auth middleware")` and the `rollback command` and `record 12` variants return `RetrieveNeedAssessment(should_retrieve=True, is_redundant=False, reason='specific_followup_with_query_hint')`, while `query="retrieve again"`, `query="search the content"`, and `query="the original content"` against thoroughness prose return `reason='thoroughness_without_gap'`. The analyzer writes a literacy rule for the second group only.
- Against a classifier that consults `query` only after the thoroughness patterns, the targeted cases fail with `should_retrieve=False, is_redundant=True`. Against one whose hint vocabulary omits the retrieval-action and content-form registers, the generic cases fail the other way with `should_retrieve=True, is_redundant=False`, and no literacy rule is written.
- Not tested: agent-side behavior against a live provider. Retrieval returns the full original payload either way, so `query` is recorded for feedback rather than used to filter content.

## Review Readiness

## Additional Notes

Current `main` still lacks the retrieval-literacy behavior covered here. The branch has no merge conflict; GitHub reports `mergeable=true` with the remaining gate being branch protection.

Hint specificity is a lexical test, so a hint of exactly one token that also names the retrieval action, such as `query="search"`, reads as generic even when it points at a real symbol. Keeping `search the content` generic and that case specific are not both available; the cost lands on single-token hints, and a second token (`query="search function signature"`) makes it specific again. Inflections are listed rather than stemmed because suffix stripping maps `using` onto the filler token `us`, and a wrong collision there silently discards a real gap.


chopratejas:
Hi - 

Thanks - can you fix the conflicts :)

rodboev:
Rebased onto current `main`, resolved the replay conflicts in the retrieve-guidance files, and force-pushed `pr/ccr-retrieve-literacy`.

The PR should be mergeable again.

JerrettDavis (changes requested):
Thanks for pushing the CCR literacy work forward. The central policy renderer and the deterministic learn coverage are a good direction, and the full CI matrix is green.

I do think this needs one contract fix before merge: the new public guidance now tells agents to call `headroom_retrieve(hash, query="search terms")` “to search within” and describes `query` as returning targeted content, but the implementation still retrieves by hash only. In `headroom/ccr/mcp_server.py`, `_handle_retrieve()` ignores `arguments["query"]`; `_retrieve_content()` calls `store.retrieve(hash_key)` and proxy fallback posts only `{"hash": hash_key}`. The proxy endpoint in `headroom/proxy/server.py` has the same behavior and comment: “Retrieval is by hash: always return the full original content.”

That mismatch will train agents to rely on a focused retrieval path that does not exist, and it also means the learn classifier treats any non-empty query as a valid targeted retrieval even though production still returns the full payload. Please either implement/pass through real targeted retrieval semantics, or soften the rendered descriptions/system instructions/skill text so `query` is described only as a feedback/context hint until filtering exists. A regression test should assert the chosen contract on the MCP and `/v1/retrieve` paths.

rodboev:
Adjusted the retrieve contract to match the implementation. `query` is now described as context for the concrete gap rather than as filtered search, and MCP plus `/v1/retrieve` pass it through to retrieval telemetry while still returning the full original payload. I also added regressions for both paths so the public contract stays pinned.

JerrettDavis (approved):
Thanks for tightening the contract. The latest update addresses the blocker I raised: the public guidance now describes `query` as an optional context/feedback hint rather than filtered retrieval, and the MCP plus `/v1/retrieve` paths pass that hint through while preserving the full-content retrieval semantics. The added regressions pin that contract on both entry points.

CI is green. This still shows merge conflicts, so it will need a rebase/conflict resolution before merge, but I do not see a remaining code-review blocker in the updated contract.

rodboev:
Rebased onto the current `main` and force-pushed the branch.

JerrettDavis (approved):
Reviewed the final CCR retrieve-literacy head. The post-review commits keep the contract aligned across the MCP tool, `/v1/retrieve`, plugin/runtime guidance, snapshots, and learn-time classification: `query` is feedback/context for the concrete gap, while retrieval still returns the full original payload. CI is green on this head, and I do not see a remaining code-review blocker.

JerrettDavis (approved):
Reviewed current head 3a0c62e0 after the post-rebase coverage/lint follow-up. The retrieve contract is still aligned across MCP, `/v1/retrieve`, plugin guidance, learn-time literacy, and proxy fallback coverage: `query` is a feedback/context hint, while retrieval returns the stored original payload.

Verified locally:
- `ruff check` on the touched Python CCR/MCP/proxy/learn files and tests -> passed
- `ruff format --check` on those files -> passed
- `python -m pytest tests/test_ccr_mcp_server.py tests/test_ccr_retrieve_policy.py tests/test_ccr_tool_always_on.py tests/test_ccr_tool_injection.py tests/test_cli/test_mcp.py tests/test_learn/test_ccr_retrieve_literacy.py tests/test_proxy_ccr.py -q` -> 117 passed, 4 skipped, 1 local failure in `test_pipeline_compresses_tool_output_and_records_toin`

That local failure is a native-extension mismatch in this checkout (`SmartCrusherConfig.__new__() got an unexpected keyword argument 'lossless_only'`), while the current remote `test (ubuntu)` shard is green on this head. I do not see it as a PR blocker.

Remote checks are green except build-wheel/docker-native-e2e/prefetch-model still pending at review time. I do not see a code-review blocker in the updated contract.

JerrettDavis (approved):

JerrettDavis (approved):
Re-reviewed current head ec21545. The latest commit is a CI retrigger only; I do not see a source-level change from the previously approved retrieve-literacy implementation. This remains approved from code review, with the obvious caveat that the newly retriggered CI matrix is still running and should finish green before merge.

JerrettDavis (approved):
Resolved against current main, retaining both retrieval-literacy rendering and telemetry session support. 174 Python tests pass (4 skipped), OpenCode 14/14 passes with typecheck, and affected OpenClaw tests 18/18 pass with typecheck. Ruff/formatting pass. The full OpenClaw suite has one unrelated pre-existing -c/-lc expectation mismatch outside this PR diff.

JerrettDavis (changes requested):
`classify_retrieve_need()` currently evaluates `_THOROUGHNESS_PATTERNS` before a non-empty `query`, so a targeted retrieval is mislabeled redundant whenever the surrounding user text contains words such as "verify" or "double-check". For example, `classify_retrieve_need("Verify the auth middleware result", query="auth middleware")` returns `should_retrieve=False, is_redundant=True`, even though the new tool/query description defines that query as the concrete gap being checked. The analyzer can then learn and write a rule discouraging a valid retrieval.

Please make a concrete query hint win over generic thoroughness wording (or otherwise classify the query's specificity rather than discarding it), and add a regression for a targeted verify/double-check request with a non-empty query. The existing test that treats every thoroughness phrase with a query as redundant pins the incorrect precedence and should be narrowed to a genuinely non-specific hint if that case is still desired.

JerrettDavis (changes requested):
The requested precedence is improved, but _is_specific_query_hint still classifies generic retrieval-action hints as specific because tokens such as retrieve, retrieval, search, find, and look are absent from _GENERIC_QUERY_TOKENS. For example, classify_retrieve_need("Verify everything", query="retrieve again") now returns should_retrieve=True even though the query names no concrete gap; the analyzer can still learn a false positive from the same thoroughness pattern this policy is meant to reject. Please treat retrieval/action-only wording as nonspecific too and add regressions for at least retrieve again and search the content (with a thoroughness request), while retaining auth middleware and rollback command as specific. The precedence change and analyzer regression otherwise resolve the original targeted-query blocker cleanly.

JerrettDavis (approved):
The latest follow-up resolves the outstanding classifier issue. Retrieval-action-only hints (`retrieve again`, `search the content`, `look it up again`) and content-form-only hints are now treated as nonspecific, while concrete subjects/locators such as `auth middleware`, `rollback command`, and `record 12` retain precedence over generic thoroughness wording. The analyzer regression also confirms the downstream learned-guidance behavior. The current head is mergeable and green; approved.

JerrettDavis (approved):
Re-reviewed the latest head and current integrated diff. The classifier now makes a concrete gap hint outrank generic thoroughness language while explicitly rejecting retrieval-action/content-form-only hints; the analyzer tests pin both downstream outcomes. The public contract remains aligned across MCP, proxy, CLI, plugins, skill text, and docs: query is context/feedback and retrieval still returns the full stored payload. No new source follows the already-reviewed corrective commits, the branch is clean, and I found no blocking regression.

PR #1535, opened 2026-06-28
fix(proxy): compress OpenCode tool schemas and embedded JSON

### LINKED ISSUE #1534: OpenAI chat completions ignore proxy savings profile kwargs
## Bug

OpenAI-compatible /v1/chat/completions requests do not receive the same proxy savings/profile kwargs as the other compression paths.

## Impact

When the proxy is started with HEADROOM_SAVINGS_PROFILE=agent-90, the effective config reports user/system message compression and target_ratio=0.10, but the live OpenAI chat-completions handler calls openai_pipeline.apply() without proxy_pipeline_kwargs(config). As a result OpenAI-compatible clients such as OpenCode can keep protecting user messages or miss the configured profile on the real chat path.

## Repro

1. Start the proxy with HEADROOM_SAVINGS_PROFILE=agent-90.
2. Send an OpenAI /v1/chat/completions request with a large user message.
3. Spy or log the kwargs passed to openai_pipeline.apply().

Expected: kwargs include compress_user_messages=True, target_ratio=0.10, min_tokens_to_compress=120, etc.
Actual: the chat-completions path only passes model_limit/context/frozen_message_count/biases/compression_policy.

## Related checks

I searched existing issues/PRs for proxy_pipeline_kwargs/openai chat, HEADROOM_COMPRESS_USER_MESSAGES, agent-90 chat completions, and HEADROOM_PREFIX_FREEZE_ENABLED and did not find an exact existing report.

vscunha:
## Description

Fixes two remaining OpenCode/OpenAI Chat compression gaps after `main` incorporated the original savings-profile threading and user content-block work from this PR.

OpenCode requests can still report very low savings when most input tokens live in verbose `tools` schemas rather than messages. They can also route poorly when a short instruction wraps a valid JSON block but does not satisfy the existing long-prose heuristic.

Closes #1534

## Type of Change

- [ ] Breaking change (fix or feature that causes existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Compact OpenAI Chat Completions `tools` schemas whenever request compression is active, reusing the existing OpenAI Responses schema compactor. The outbound tool invocation shape is preserved while non-semantic annotations such as `$schema`, `title`, and `examples` are removed.
- Include the tool-schema token delta in Headroom's savings accounting and expose `openai:chat:tool_schema_compaction` in the applied transforms.
- Detect valid JSON blocks surrounded by prose or log text as mixed content, so short OpenCode instructions route through mixed/SmartCrusher handling instead of falling through or producing a no-op.
- Adapt the mixed-content change to the new `headroom.transforms.mixed_content` module introduced on `main` by #1939.

## Why the Focus Changed

The original headline fix—threading savings-profile kwargs into `/v1/chat/completions`—is now already present on `main`, as is the user content-block opt-in behavior. Those duplicate changes were removed during the merge.

The branch also no longer changes developer/system role protection or forced-Kompress semantics. It follows `main` for both, so the earlier instruction-role safety concern is outside the current diff.

The resulting PR is limited to two OpenCode-specific compression gaps that remain reproducible on current `main`.

## Testing

- [ ] Manual live-upstream testing performed after the latest rebase

### Test Output

```text
59 passed, 1 warning in 83.53s
All checks passed!  # ruff check
4 files already formatted  # ruff format --check
python -m py_compile: passed
git diff --check: passed
```

Focused test coverage includes:

- OpenAI Chat tool-schema compaction, transform reporting, outbound schema shape, and positive token savings.
- Embedded JSON mixed-content detection, SmartCrusher routing, positive savings, and preservation of a critical sentinel value.
- Current `main` regressions for savings-profile threading, user content blocks, turn hooks, and forced-Kompress behavior.

## Real Behavior Proof

- Environment: Linux ARM64, Python 3.13.12, current `main` at `9bacf481` merged into the branch.
- Exact command / steps: focused pytest run across the OpenAI cache-stability, content-router, mixed-content, savings-profile, user-block, turn-hook, and forced-Kompress suites.
- Observed result: 59 tests passed; the chat request test forwarded compacted tools and reported positive savings, while the embedded-JSON fixture used mixed routing and preserved `CRITICAL_NEEDLE_42`.
- Not tested: full repository suite and a live external OpenCode request after the latest merge; those remain for CI/live follow-up.

## Review Readiness

## Checklist

- [x] I have commented the non-obvious behavior
- [ ] I have made corresponding documentation changes — N/A; internal routing behavior only
- [x] I have added tests that prove the fixes are effective
- [x] New and existing focused tests pass locally
- [ ] I have updated the changelog — N/A; release automation handles fix entries

## Screenshots

N/A — proxy/transform behavior only.

## Additional Notes

- Current diff versus `main`: 4 files, 172 insertions, no role-policy or forced-Kompress changes.
- The mixed-content conflict was resolved by extending the new isolated parser module rather than reintroducing parsing code into `ContentRouter`.


JerrettDavis (changes requested):
The OpenAI chat savings direction is useful, but this currently makes developer role text blocks compressible by default in ContentRouter._process_content_blocks. Developer messages are instruction-level content in OpenAI chat, closer to system/developer policy than user/tool output; compressing them can change routing instructions or safety constraints. The old test used developer as the safety example and this PR changes it to custom, which hides that behavior change. Please keep developer protected by default unless there is an explicit opt-in comparable to skip_system=False, and add a regression test that developer-role text blocks are not compressed under the agent profile.

JerrettDavis (commented):
Thanks for the update. The prior blocker is addressed in the latest diff: `developer` role text blocks are now always protected by default, even when the agent savings profile opts into user/system compression, and `test_developer_text_blocks_stay_protected_under_agent_profile` pins that behavior.

I attempted the focused content-router tests locally, but this review worktree cannot collect them because `headroom._core` is not built. From code inspection, the instruction-role safety concern is resolved. Since the PR is still draft, I am leaving this as a comment rather than an approval.

vscunha:
Update: the focus of this PR has narrowed slightly after syncing with current `main`.

The original savings-profile threading fix and user content-block opt-in behavior are now already upstream, so I removed those duplicate changes. I also removed the branch-specific developer-role and forced-Kompress behavior; the PR now follows `main` for both. That means the earlier instruction-role safety concern is no longer part of the current diff.

The PR now addresses only two remaining OpenCode compression gaps:

1. **Large Chat Completions tool schemas were not compressed or counted.** OpenCode can put most request tokens in `tools`, so message compression alone produced near-zero reported savings. The PR reuses the existing OpenAI schema compactor, preserves the callable schema shape, records the transform, and includes the actual tool-token delta in savings.
2. **Short instructions wrapped around valid JSON were misrouted.** The existing detector required a stronger prose signal, so common prose + JSON payloads could fall through or no-op. The new signal validates the JSON block and routes the payload through mixed/SmartCrusher handling while preserving critical values.

The latest `main` conflict came from #1939 moving mixed-content parsing into `headroom.transforms.mixed_content`. I resolved it by putting the new detector in that module, keeping the refactor intact.

The final diff is focused: **4 files, 172 insertions**, with no role-policy or forced-Kompress changes. Verification is green: **59 focused tests passed**, Ruff check/format passed, syntax compilation passed, and `git diff --check` passed.

With the obsolete scope removed, the conflict resolved against current `main`, and both remaining behaviors covered by regressions, this is now ready for review as a targeted OpenCode compression fix.

JerrettDavis (approved):
Re-reviewed the refreshed head after the latest `main` merges. The prior blocker remains addressed: developer/unknown-role text blocks stay protected by default, while tool-role output remains compressible, and the OpenAI chat tool-schema compaction path is covered.

Verified locally on Windows:
- `python -m pytest tests/test_proxy_openai_cache_stability.py::test_openai_chat_completions_compacts_tools_when_profile_enabled -q` -> 1 passed
- `python -m pytest tests/test_transforms_content_router.py::test_unknown_role_text_blocks_skipped_for_safety -q` -> 1 passed
- `ruff check headroom/proxy/handlers/openai.py headroom/transforms/mixed_content.py tests/test_proxy_openai_cache_stability.py tests/test_transforms_content_router.py` -> passed
- `ruff format --check headroom/proxy/handlers/openai.py headroom/transforms/mixed_content.py tests/test_proxy_openai_cache_stability.py tests/test_transforms_content_router.py` -> passed

PR #1548, opened 2026-06-29
fix(detection): contain unidiff panic on orphaned +++ target line

### LINKED ISSUE #1547: [BUG] Content detection panics (`Option::unwrap()` on `None`) on any text containing an orphaned `+++ ` line — bundled `unidiff` 0.4.0
## Description

A fatal bug occurs in `headroom._core.detect_content_type()` where it panics with `pyo3_runtime.PanicException: called Option::unwrap() on a None value`. This happens whenever the input text contains a line starting with `+++ ` followed by a non-empty target **without a preceding `--- ` source line**. This pattern is extremely common in real-world payloads, such as `bash set -x` xtrace outputs or partial `git diff` snippets quoted out of context.

In the proxy, this crashes request-payload compression (`_compress_openai_responses_payload` -> `ContentRouter` -> `detect_content_type`) inside a `ThreadPoolExecutor` worker. Consequently, the proxy returns an **HTTP 500** for the whole request. Because the panic is deterministic based on the payload content, client retries fail consistently, effectively wedging the session until the offending text leaves the context window.

The panic originates in the bundled **`unidiff` crate v0.4.0** (`unidiff-0.4.0/src/lib.rs:665:54`).

## To Reproduce

1. Install `headroom` version `0.27.0` on Linux (ensuring the default `HEADROOM_DETECT_BACKEND=rust` is active).
2. Run the provided code sample below, passing an orphaned `+++ ` string.
3. See the `PanicException` error.

## Expected Behavior

Content detection should gracefully handle invalid or partial diff fragments. Instead of panicking, it should fall back to recognizing the text as plain text (or return a handled error) so that the proxy continues to operate normally and process the request.

## Actual Behavior

The underlying Rust `unidiff` component panics by calling `.unwrap()` on a `None` value, which raises a `pyo3_runtime.PanicException`. Since this panic happens in a `ThreadPoolExecutor` worker, it is not caught by standard compression error handling (it bypasses `HEADROOM_WS_FAIL_OPEN_ON_COMPRESSION_FAILURE`). This results in a hard crash of the worker, causing the proxy to return a generic 500 error ("We're currently experiencing high demand...") masking the real cause.

## Code Sample

```python
# Minimal code to reproduce the issue
from headroom._core import detect_content_type

# Calling this with an orphaned "+++" target line causes a panic
detect_content_type("+++ x")
```

## Error Output

```text
File ".../headroom/proxy/handlers/openai.py", line 1454, in _compress_openai_responses_payload_in_executor
File ".../headroom/proxy/server.py", line 1071, in _run_compression_in_executor   # asyncio.wait_for(future)
File ".../concurrent/futures/thread.py", line 59, in run
File ".../headroom/proxy/handlers/openai.py", line 1439, in _compress
File ".../headroom/proxy/handlers/openai.py", line 1308, in _compress_openai_responses_payload
File ".../headroom/proxy/handlers/openai.py", line 1069, in _compress_openai_responses_live_text_units_with_router
File ".../headroom/transforms/compression_units.py", line 300, in compress_unit_with_router
File ".../headroom/transforms/content_router.py", line 1109, in compress
File ".../headroom/transforms/content_router.py", line 154, in _detect_content   # rust_result = _rust_detect(content)
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
  panicked at .../unidiff-0.4.0/src/lib.rs:665:54
```

## Environment

* **Headroom version**: 0.27.0 (PyPI / ghcr image `headroom:0.27.0`)
* **Python version**: CPython 3.13 (the `headroom:0.27.0` image ships `/usr/local/lib/python3.13`)
* **OS**: Linux
* **LLM Provider**: OpenAI / Azure (used as proxy, wire API `responses` / Codex)

### Characterization (What panics vs. what works)

| Input | Result |
| --- | --- |
| `+++ x` | **PANIC** |
| `+++ dirname tests/foo.sh` (bash xtrace line) | **PANIC** |
| `"++ foo\n+++ dirname x/y.sh\n++ bar\n"` (xtrace block) | **PANIC** |
| `"+++ b\n--- a"` (target before source) | **PANIC** |
| `+++ ` (empty target) | ok → `text` |
| `+++` (no trailing space) | ok → `text` |
| `++++ x` (4 pluses) | ok → `text` |
| `--- x` (lone source line) | ok → `text` |
| `"--- a\n+++ b"` (correct order) | ok |
| `this has +++ in the middle` | ok → `text` |
| a complete `diff --git ... --- ... +++ ... @@` | ok → `diff` |

**Trigger Condition:** A `+++ <non-empty>` target-file line with no matching preceding `--- ` source-file line.

### Suggested Fixes (Any of the following)

1. **Bump the `unidiff` crate** past `0.4.0`. Current upstream returns `Error::TargetWithoutSource` instead of calling `.unwrap()` at this site.
2. **Wrap the `unidiff` parse** in `std::panic::catch_unwind` inside `detect_content_type` and fall back to PlainText on panic, ensuring untrusted text never crashes detection.
3. **Defense-in-depth in Python:** `_detect_content` could catch `BaseException` (`PanicException` is a subclass) around `_rust_detect` and safely fall back to `_regex_detect_content_type`.

### Workaround for Operators

Set `HEADROOM_DETECT_BACKEND=python` on the proxy. This routes detection through the pure-Python regex detector and completely avoids the Rust `unidiff` path (verified: `+++ x` safely parses as `PLAIN_TEXT`).

* *Trade-off:* Coarser content-type detection.
* *Note:* Setting `--no-code-aware` or `HEADROOM_CODE_AWARE_ENABLED=false` does **not** help, as that only disables the tree-sitter AST path (issue #562). This panic occurs in the core content detection phase, which runs regardless.

tenderdeve:
## Description

`headroom._core.detect_content_type()` panics with `pyo3_runtime.PanicException: called Option::unwrap() on a None value` on any text containing a `+++ ` target line with no preceding `--- ` source line — e.g. `set -x` xtrace output or a partial `git diff` quoted out of context.

The panic originates in the bundled `unidiff` 0.4.0 parser (`lib.rs:665`): on a target-file header it does `source_file.clone().unwrap()`, but `source_file` is still `None` when no source header was seen. The crate's only guard there checks `current_file`, not `source_file`, so it falls through and unwraps `None` instead of returning `Err`.

Because detection runs inside a `ThreadPoolExecutor` worker on the Python side, the native panic surfaces as an uncaught `PanicException`, bypasses the compression error handling, and returns **HTTP 500** for the whole request. The failure is deterministic on payload content, so client retries fail until the offending text leaves the context window.

`is_diff()` in `unidiff_detector.rs` is the single entry point that drives `PatchSet::parse`, so the fix is contained there: wrap the parse in `catch_unwind` and treat an unparseable fragment as "not a diff". This matches the workspace's deliberate no-`panic = "abort"` policy (Cargo.toml) of surviving bad input rather than taking the long-lived proxy down.

Closes #1547

## Type of Change

## Changes Made

- `crates/headroom-core/src/transforms/unidiff_detector.rs`: contain any `unidiff` parser panic inside `is_diff()` via `catch_unwind`, returning `false` (not a diff) on panic. Added regression test `orphaned_target_line_does_not_panic`.
- `CHANGELOG.md`: note under Unreleased → Fixed.

## Testing

- [x] Unit tests pass (`cargo test -p headroom-core`)
- [x] Linting passes (`cargo fmt --check`, `cargo clippy`)

### Test Output

Before the fix (regression test reproduces the exact panic):

```text
running 1 test
test transforms::unidiff_detector::tests::orphaned_target_line_does_not_panic ... FAILED

---- transforms::unidiff_detector::tests::orphaned_target_line_does_not_panic stdout ----
thread '...' panicked at unidiff-0.4.0/src/lib.rs:665:54:
called `Option::unwrap()` on a `None` value

test result: FAILED. 0 passed; 1 failed; ...
```

After the fix:

```text
running 15 tests
test transforms::unidiff_detector::tests::orphaned_target_line_does_not_panic ... ok
test transforms::unidiff_detector::tests::standard_git_diff_detected ... ok
...
test result: ok. 15 passed; 0 failed; 0 ignored

# whole transforms suite
test result: ok. 700 passed; 0 failed; 0 ignored
```

## Real Behavior Proof

- Environment: macOS (arm64), Rust stable, `cargo test -p headroom-core`.
- Exact command / steps: `cargo test -p headroom-core --lib unidiff_detector` then `cargo test -p headroom-core`. (1) Added a test calling `is_diff("+++ x")` / `detect_diff("+++ x")` and ran it → reproduced the panic at `unidiff-0.4.0/src/lib.rs:665:54` (output above), confirming the same crash path as the report. (2) Applied the `catch_unwind` containment in `is_diff()`. (3) Re-ran the test and the full transforms suite → all green (output above).
- Observed result: the orphaned-`+++ ` input is now classified as "not a diff" (plain text) and returns normally instead of panicking. Real diffs (`standard_git_diff_detected`, `naked_hunk_without_git_header_detected`, multi-file, added/removed-only) still detect correctly, so the containment does not weaken detection.
- Not tested: I exercised the Rust layer directly (the sole `unidiff` caller, which the `headroom._core.detect_content_type` binding routes through) rather than rebuilding the Python wheel; I did not run the live proxy against a real provider.

## Review Readiness

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
+- Content detection no longer crashes the proxy on text containing an
+  orphaned `+++ ` target line with no preceding `--- ` source line (common in
+  `set -x` xtrace output and partial diffs). The bundled `unidiff` 0.4.0 parser
+  panics on that input instead of returning an error; the Rust diff detector now
+  contains the panic and treats the fragment as plain text, so the request is
+  compressed and forwarded normally instead of returning HTTP 500
+  ([#1547](https://github.com/headroomlabs-ai/headroom/issues/1547)).
 - Proactive expansion blocks injected into user turns are now wrapped in
   `<headroom_proactive_expansion>` XML tags, giving downstream consumers
   (LLMs, loggers, attribution parsers) a machine-readable provenance
--- a/crates/headroom-core/src/transforms/unidiff_detector.rs
+++ b/crates/headroom-core/src/transforms/unidiff_detector.rs
@@ -61,16 +61,28 @@ pub fn is_diff(content: &str) -> bool {
         return false;
     }
 
-    let mut patch = PatchSet::new();
-    if patch.parse(content).is_err() {
-        return false;
-    }
+    // `unidiff` 0.4.0 does not return `Err` on every malformed input — a
+    // `+++ ` target line with no preceding `--- ` source line makes it
+    // `unwrap()` a `None` and panic (lib.rs:665). Inputs of that shape are
+    // common (`set -x` xtrace, partial diffs quoted out of context). This
+    // detector runs inside a thread-pool worker on the Python side, where a
+    // native panic surfaces as an uncaught `PanicException` and 500s the whole
+    // request. Contain any parser panic here and treat the fragment as "not a
+    // diff" — consistent with the workspace's no-`panic = "abort"` policy of
+    // surviving bad input rather than taking the process down.
+    std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        let mut patch = PatchSet::new();
+        if patch.parse(content).is_err() {
+            return false;
+        }
 
-    // `PatchSet::is_empty()` covers "found zero files"; the inner
-    // loop covers "found a file but with zero hunks" (e.g. mode-only
-    // changes). For diff-compressor routing we want at least one
-    // hunk — that's where the actual line-level change content lives.
-    !patch.is_empty() && patch.files().iter().any(|f| !f.is_empty())
+        // `PatchSet::is_empty()` covers "found zero files"; the inner
+        // loop covers "found a file but with zero hunks" (e.g. mode-only
+        // changes). For diff-compressor routing we want at least one
+        // hunk — that's where the actual line-level change content lives.
+        !patch.is_empty() && patch.files().iter().any(|f| !f.is_empty())
+    }))
+    .unwrap_or(false)
 }
 
 /// [`ContentType`]-typed wrapper. Returns `Some(ContentType::GitDiff)`
@@ -228,5 +240,17 @@ mod tests {
         assert_eq!(detect_diff("not a diff"), None);
         assert_eq!(detect_diff("{}"), None);
         assert_eq!(detect_diff(""), None);
+    }
+
+    #[test]
+    fn orphaned_target_line_does_not_panic() {
+        // `unidiff` 0.4.0 panics (unwrap on `None`) when it meets a
+        // `+++ ` target line with no preceding `--- ` source line.
+        // That shape is common in `set -x` xtrace output and partial
+        // diffs quoted out of context. It must degrade to "not a diff",
+        // never abort the caller.
+        assert!(!is_diff("+++ x"));
+        assert_eq!(detect_diff("+++ x"), None);
+        assert!(!is_diff("some prose\n+++ target without a source\nmore"));
     }
 }


JerrettDavis (approved):
Reviewed the unidiff panic containment. The catch_unwind is tightly scoped to the parser call and malformed orphan +++ input now degrades to not-a-diff instead of taking down the request. Looks ready from the code side.

Issue #1551, opened 2026-06-29
Bedrock backend fails with "No module named 'botocore'" when using temporary STS credentials (session token)


jvilhuber:
## Summary

When running the Docker image with `--backend bedrock` and **temporary** AWS credentials (i.e. `AWS_SESSION_TOKEN` is set, as produced by SSO / `credential_process` / STS assume-role), every request fails. The proxy logs show requests being dispatched, but the client receives an `authentication_error` whose underlying cause is `ModuleNotFoundError: No module named 'botocore'`.

The image bundles `litellm` but not `botocore`. litellm can self-sign Bedrock requests for *static* IAM keys without botocore, but as soon as a session token is present it takes the `_auth_with_aws_session_token` code path in `litellm/llms/bedrock/base_aws_llm.py`, which imports `botocore` — and that import fails.

## Environment

- Image: `ghcr.io/chopratejas/headroom:latest`
- Backend: `bedrock`
- Python 3.13 in-image; litellm present, botocore absent
- Credentials: temporary STS keys passed via env (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `AWS_REGION`)

## Steps to reproduce

1. Obtain temporary AWS credentials (any source that yields a session token).
2. Run:
   ```
   docker run -p 8787:8787 \
     -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN \
     -e AWS_REGION=us-west-2 \
     ghcr.io/chopratejas/headroom:latest \
     --host 0.0.0.0 --port 8787 --backend bedrock
   ```
3. Send a request:
   ```
   curl -s -X POST http://localhost:8787/v1/messages \
     -H "content-type: application/json" -H "x-api-key: dummy" \
     -H "anthropic-version: 2023-06-01" \
     -d '{"model":"us.anthropic.claude-opus-4-8","max_tokens":50,"messages":[{"role":"user","content":"say hi"}]}'
   ```

## Actual result

```
{"type":"error","error":{"type":"authentication_error",
"message":"litellm.APIConnectionError: No module named 'botocore'\n
... File \".../litellm/llms/bedrock/chat/converse_handler.py\", line 339, in completion\n
    credentials = self.get_credentials(...)\n
  File \".../litellm/llms/bedrock/base_aws_llm.py\", line 346, in get_credentials\n
[...]

PR #1553, opened 2026-06-29
fix(bedrock): fail fast when session-token auth lacks botocore

tenderdeve:
## Description

With `--backend bedrock` and **temporary** AWS credentials (`AWS_SESSION_TOKEN`, as produced by SSO / STS assume-role / `credential_process`), every request fails. litellm self-signs Bedrock requests without botocore for *static* IAM keys, but as soon as a session token is present it takes the `_auth_with_aws_session_token` path in `litellm/llms/bedrock/base_aws_llm.py`, which imports `botocore`. botocore is an optional dependency — it ships only with headroom's `bedrock` extra, and the default Docker image is built with `HEADROOM_EXTRAS=proxy,code`, so botocore is absent. The failure surfaces only at request time as a misleading `authentication_error: No module named 'botocore'` (and as a bare `Invalid API key` in Claude Code).

This PR makes the Bedrock backend **fail fast at startup** with an actionable message when a session token is set but botocore is missing — directly addressing the "clearer error message" the reporter asked for. It mirrors the existing optional-dependency guard pattern already used for boto3 in `backends/litellm.py`.

Scope note: this does not change what the published image ships — whether to add botocore/`bedrock` to the default image extras is a separate sizing decision I left to maintainers. Static-credential Bedrock users (who never hit the botocore path) are unaffected.

Refs #1551

## Type of Change

## Changes Made

- `headroom/backends/litellm.py`: when initializing the Bedrock backend with `AWS_SESSION_TOKEN` set and `botocore` not importable, raise an `ImportError` pointing at `pip install 'headroom-ai[bedrock]'` instead of letting the request fail later with a misleading auth error.
- `tests/test_backends/test_bedrock_botocore_preflight.py`: regression tests — the guard raises an actionable error for the session-token-without-botocore case, and stays quiet for the static-credential case.
- `CHANGELOG.md`: note under Unreleased → Fixed.

## Testing

- [x] Linting passes (`ruff check`, `ruff format --check`)

### Test Output

Regression test fails before the fix (no guard → no error raised), passes after:

```text
# before fix (guard removed)
FAILED tests/test_backends/test_bedrock_botocore_preflight.py::test_bedrock_session_token_without_botocore_raises_actionable

# after fix
tests/test_backends/test_bedrock_botocore_preflight.py ..  [100%]
2 passed, 1 warning in 0.13s
```

`ruff check` / `ruff format --check` on the changed files: clean.

## Real Behavior Proof

- Environment: macOS (arm64), Python venv, editable install (`pip install -e .`, no `bedrock` extra → botocore absent, matching the reported slim-image condition), `pytest`.
- Exact command / steps: `python -m pytest tests/test_backends/test_bedrock_botocore_preflight.py`. (1) Removed the guard and ran the test → it failed because `LiteLLMBackend(provider="bedrock")` with `AWS_SESSION_TOKEN` set and botocore absent did NOT raise (reproducing the original "no early signal" behavior). (2) Applied the guard. (3) Re-ran → both tests pass, and the raised `ImportError` contains the `headroom-ai[bedrock]` install hint.
- Observed result: with `AWS_SESSION_TOKEN` set and botocore not importable, the backend now raises a clear, actionable `ImportError` at construction time instead of deferring to litellm's later `No module named 'botocore'` auth error. Without a session token the guard does not fire, so static-credential users are unaffected.
- Not tested: I did not run a live Bedrock request against AWS with real temporary credentials (no AWS account/STS access in this environment); the reporter already confirmed that installing botocore makes the identical request succeed, and this change surfaces that requirement at startup.

## Review Readiness

## Checklist


JerrettDavis (approved):
Reviewed the Bedrock preflight. The guard is scoped to the temporary-credential path (AWS_SESSION_TOKEN) that actually needs botocore, leaves static-credential users alone, and the error message points at the correct extra. Looks ready from the code side.

Issue #1544, opened 2026-06-29
[BUG] Windows: headroom install status terminates live proxy and crashes with WinError 87


karthikcsq:
﻿## Environment

- Headroom: `0.27.0`
- OS: Windows
- Python: `3.13` from Anaconda
- Deployment: `persistent-task`, Python runtime, provider scope
- Port: `8787`

## Reproduction

Install a persistent Codex deployment from an elevated PowerShell:

```powershell
headroom install apply `
  --preset persistent-task `
  --runtime python `
  --scope provider `
  --providers manual `
  --target codex `
  --profile codex-desktop `
  --port 8787 `
  --no-telemetry
```

Confirm the proxy is ready, then run the status command:

```powershell
Invoke-WebRequest http://127.0.0.1:8787/readyz -UseBasicParsing
headroom install status --profile codex-desktop
```

## Actual behavior

The command prints profile metadata through `Port: 8787`, then crashes:

```text
OSError: [WinError 87] The parameter is incorrect

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  ...
  File "headroom\cli\install.py", line 254, in install_status
    click.echo(f"Status:     {runtime_status(manifest)}")
  File "headroom\install\runtime.py", line 329, in runtime_status
    os.kill(pid, 0)
SystemError: <built-in function kill> returned a result with an exception set
```

The live proxy is also terminated. After the command:

- `http://127.0.0.1:8787/readyz` is unreachable
- nothing listens on port 8787
- `.headroom/deploy/codex-desktop/runner.pid` has been removed

I reproduced this end to end by starting the deployment through `ensure-headroom.cmd`, waiting for `/readyz` to return 200, recording the live PID, and immediately running `headroom install status`. Before status, the proxy was healthy and `runner.pid` contained PID `25212`. After status, the proxy was unreachable and the PID file was gone.

## Expected behavior

`headroom install status` should report the live deployment as running and healthy without changing its state.

## Suspected cause

`runtime_status()` uses a Unix-style liveness probe:

```python
try:
    os.kill(pid, 0)
except OSError:
    return "stopped"
```

The detached Windows agent is created with `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP`. In this path, `os.kill(pid, 0)` produces `WinError 87`, surfaces as `SystemError` rather than `OSError`, and the supervised proxy exits.

[...]

PR #1315, opened 2026-06-23
fix(wrap): don't crash cleanup on Windows when os.kill liveness check raises SystemError

quentinmaisonneuve:
## Description

On Windows, `headroom wrap claude` crashes during `cleanup()` on exit, which can leave the shared proxy (port 8787) orphaned. The crash originates in `_pid_alive()`.

`_pid_alive()` probes liveness with `os.kill(pid, 0)`. On Windows, calling `os.kill()` against a stale/recycled/invalid PID fails with `WinError 87` ("The parameter is incorrect"), which CPython sometimes surfaces not as an `OSError` but as `SystemError: <class 'OSError'> returned a result with an exception set`. `SystemError` is **not** an `OSError` subclass, so the existing `except OSError` does not catch it. The exception propagates `_pid_alive` -> `_live_proxy_clients` -> `_other_clients_exist` -> `cleanup`, aborting `cleanup()` before `proc.terminate()` and leaving the proxy running.

Original traceback (Claude itself had already exited cleanly with code 0):

```text
File ".../headroom/cli/wrap.py", line 2456, in cleanup
    if _other_clients_exist():
File ".../headroom/cli/wrap.py", line 2448, in _other_clients_exist
    return len(_live_proxy_clients(port, exclude_self=True)) > 0
File ".../headroom/cli/wrap.py", line 2425, in _live_proxy_clients
    if not _pid_alive(pid) or _marker_pid_reused(marker, pid):
File ".../headroom/cli/wrap.py", line 2378, in _pid_alive
    os.kill(pid, 0)
SystemError: <class 'OSError'> returned a result with an exception set
```

Closes # (no tracking issue)

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `_pid_alive()` now prefers `psutil.pid_exists()` when available. `psutil` is already an optional dependency used by `_proc_identity()`, and `pid_exists()` is reliable on Windows and never raises for invalid PIDs.
- Kept `os.kill()` as a fallback for installs without `psutil`, but the `except` clause now also catches `SystemError` (alongside `ProcessLookupError`/`OSError`) and treats it as "not alive".
- Added a guard that treats non-positive PIDs as not alive.

## Testing

[...]

PR #1556, opened 2026-06-29
fix: harden fd lifecycle and SystemError handling in runtime and proxy kill

guyoron1:
## Description

Make `pid_alive()` safe on Windows even when `psutil` is not installed,
and harden `_kill_proxy_by_pid` exception handling for stale PIDs.

### Problem

`headroom._subprocess.pid_alive()` falls back to `os.kill(pid, 0)` when
`psutil` cannot be imported. On Windows, CPython routes `os.kill(pid, 0)`
through `TerminateProcess` — a destructive call that **kills the target
process**. Since `psutil` is not a declared runtime dependency in
`pyproject.toml`, a normal lightweight install can hit that fallback,
meaning `runtime_status()` can silently terminate a live proxy.

### Fix

- **`headroom/_subprocess.py`**: On `win32`, bypass `os.kill` entirely
  and probe via `kernel32.OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)`.
  If `ctypes` also fails, return `True` conservatively (assume alive)
  to prevent false-negative liveness from causing callers to kill a
  running process.
- **`headroom/cli/wrap.py`**: Widen `_kill_proxy_by_pid` exception
  handlers on both SIGTERM and SIGKILL paths to catch `OSError` and
  `SystemError` (Windows `WinError 87`), preventing crashes from
  stale/invalid PIDs.

## Type of Change

## Testing

- [x] Type checking passes (`mypy headroom/_subprocess.py`)

### New Tests

- `test_pid_alive_win32_no_psutil_never_calls_os_kill` — simulates
  `win32` + broken `psutil`, asserts `os.kill` is never called and
  the `kernel32.OpenProcess` path is used instead
- `test_pid_alive_win32_no_psutil_no_ctypes_returns_conservative` —
  simulates `win32` + broken `psutil` + broken `ctypes`, asserts
  `os.kill` is never called and `True` is returned conservatively

### DIFF
--- a/headroom/_subprocess.py
+++ b/headroom/_subprocess.py
@@ -1,32 +1,41 @@
 import os
 import subprocess as _sp
+import sys
 from typing import Any
 
 
-def pid_alive(pid: int) -> bool:
-    """Return True if ``pid`` names a live process (Windows-safe).
+def _win32_pid_alive(pid: int) -> bool:
+    """Non-destructive PID liveness probe for Windows via ``kernel32``."""
+    import ctypes
 
-    ``os.kill(pid, 0)`` is the usual Unix liveness probe, but on Windows it can
-    fail against a detached/stale/invalid PID with ``WinError 87`` ("The
-    parameter is incorrect"). CPython sometimes surfaces that as a
-    ``SystemError`` rather than an ``OSError``; since ``SystemError`` is not an
-    ``OSError`` subclass, a bare ``except OSError`` lets it escape and crash the
-    caller — and in the detached-agent path it took down the supervised proxy
-    (issue #1544). Prefer ``psutil.pid_exists`` and treat ``SystemError`` as
-    "not alive".
-    """
+    kernel32 = ctypes.windll.kernel32  # type: ignore[attr-defined]
+    PROCESS_QUERY_LIMITED_INFORMATION = 0x1000
+    handle = kernel32.OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, False, pid)
+    if handle:
+        kernel32.CloseHandle(handle)
+        return True
+    return False
+
+
+def pid_alive(pid: int) -> bool:
+    """Return True if *pid* names a live process (non-destructive on all platforms)."""
     if pid <= 0:
-        return False  # non-positive PIDs are never valid liveness targets
+        return False
     try:
-        import psutil  # type: ignore[import-untyped]  # optional dep, already used elsewhere
+        import psutil  # type: ignore[import-untyped]  # optional dep
 
         return bool(psutil.pid_exists(pid))
     except Exception:
         pass
+    if sys.platform == "win32":
+        try:
+            return _win32_pid_alive(pid)
+        except Exception:
+            return True
     try:
         os.kill(pid, 0)
     except PermissionError:
-        return True  # exists but owned by another user
+        return True
     except (ProcessLookupError, OSError, SystemError):
         return False
     return True
--- a/headroom/_subprocess.py
+++ b/headroom/_subprocess.py
@@ -14,7 +14,8 @@ def _win32_pid_alive(pid: int) -> bool:
     if handle:
         kernel32.CloseHandle(handle)
         return True
-    return False
+    ERROR_ACCESS_DENIED = 5
+    return kernel32.GetLastError() == ERROR_ACCESS_DENIED
 
 
 def pid_alive(pid: int) -> bool:
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -2781,11 +2781,11 @@ def _kill_proxy_by_pid(pid: int, port: int) -> bool:
     """
     try:
         os.kill(pid, signal.SIGTERM)
-    except ProcessLookupError:
-        pass  # Already gone
     except PermissionError:
         click.echo(f"  Warning: No permission to kill proxy PID {pid}")
         return False
+    except (ProcessLookupError, OSError, SystemError):
+        pass
 
     # Wait for port to free (up to 5 seconds)
     for _ in range(50):
@@ -2797,7 +2797,7 @@ def _kill_proxy_by_pid(pid: int, port: int) -> bool:
     try:
         _kill_signal = getattr(signal, "SIGKILL", signal.SIGTERM)
         os.kill(pid, _kill_signal)
-    except (ProcessLookupError, PermissionError):
+    except (ProcessLookupError, PermissionError, OSError, SystemError):
         pass
 
     for _ in range(20):
--- a/tests/test_pid_alive.py
+++ b/tests/test_pid_alive.py
@@ -48,3 +48,69 @@ def test_pid_alive_only_uses_signal_zero(monkeypatch) -> None:
     monkeypatch.setattr("headroom._subprocess.os.kill", lambda pid, sig: sent.append(sig))
     assert pid_alive(4321) is True
     assert sent == [0]
+
+
+def test_pid_alive_win32_no_psutil_never_calls_os_kill(monkeypatch) -> None:
+    """On Windows without psutil, pid_alive must not call os.kill (it routes through TerminateProcess)."""
+    monkeypatch.setitem(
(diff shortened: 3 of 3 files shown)


JerrettDavis (changes requested):
This is close, but there is still one fd-leak path in start_detached_agent(): log_file.close() only runs after subprocess.Popen(...) returns successfully. If Popen raises after the file is opened, the parent still leaks that descriptor. Please wrap that call in try/finally (or use an equivalent ownership transfer pattern) and add a small regression test, similar to the _start_proxy Popen-failure test in this PR. Also note that #1560 is now centralizing the PID liveness logic in headroom._subprocess.pid_alive(); if both PRs remain open, this branch should either build on that helper or be rebased to avoid two slightly different Windows PID fixes landing separately.

guyoron1:
Heyyy!
Thank you for the review @JerrettDavis — pushed a fix:

- `start_detached_agent()` now wraps `Popen` in `try/finally` so `log_file.close()` runs even on failure
- Added a regression test (`test_start_detached_agent_closes_log_file_on_popen_failure`) mirroring the `_start_proxy` pattern

Re #1560: will rebase onto that branch once it merges to unify the PID liveness path behind `pid_alive()`.
: )

JerrettDavis (approved):
Reviewed the latest update. This addresses the fd-leak and Windows PID-probe blockers I raised.

`start_detached_agent()` now closes the parent log handle in a `finally`, `_start_proxy()` does the same for the proxy stdio log, and the PID-probe paths now treat Windows `SystemError` like stale/invalid process state instead of crashing. The added regression tests cover both the Popen-failure fd leak and the `SystemError` paths.

Local focused check passed:

`python -m pytest tests/test_install/test_runtime.py::test_stop_runtime_catches_system_error_on_windows tests/test_install/test_runtime.py::test_runtime_status_catches_system_error_on_windows tests/test_install/test_runtime.py::test_start_detached_agent_closes_log_file_on_popen_failure tests/test_cli_proxy_env.py::TestKillProxyByPid tests/test_cli_proxy_env.py::TestStartProxyLogFileLeak -q`

That ran 8 tests, all passed. A broader local run is not useful in this worktree because many unrelated proxy tests require the native `headroom._core` extension, which is not built here. GitHub currently shows governance checks only for this PR, not the full test matrix.

guyoron1:
Rebased onto latest main — the conflicts were from #1464 which already landed the `_start_proxy` try/finally independently. 
After rebase, the only novel `wrap.py` change left is the `SystemError` hardening in `_kill_proxy_by_pid`.
Seems like everything else (runtime.py, tests) carried over cleanly.

Re #1560: still planning to rebase onto `pid_alive()` once that merges.

abhay-codes07:
Nice — the `start_detached_agent` fd leak and the `_kill_proxy_by_pid` guards look right, and this overlaps my #1576 (fd leak) which I'm happy to defer to this broader PR.

One correctness note on the `runtime_status` part (Issue B / #1544): this keeps `os.kill(pid, 0)` and just adds `SystemError` to the `except`. That fixes the crash, but it doesn't fix the other half of #1544 — the reporter saw the *live proxy get terminated* by `install status`. On Windows `os.kill(pid, sig)` for any sig other than `CTRL_C_EVENT`/`CTRL_BREAK_EVENT` goes through `TerminateProcess` (see the os.kill docs: "Any other value for sig will cause the process to be unconditionally killed by the TerminateProcess API"). So `os.kill(pid, 0)` on a live PID kills it with exit code 0 — the `except` never gets a chance to matter on the happy path.

A liveness probe that doesn't signal the process avoids this — e.g. `psutil.pid_exists(pid)` with the `os.kill` path only as a POSIX fallback. That's the approach in #1609 (mirrors the `_pid_alive` helper #1315 added for the wrap path). Might be worth folding that into the `runtime_status`/`stop_runtime` liveness checks here so #1544 is fully closed. Happy to rebase #1609 down to just whatever isn't covered here, or hand the helper over — whatever's least duplicate work.

guyoron1:
Heyaaaaa : )
  Thanks @abhay-codes07 — thanks for catching the os.kill(pid, 0) / TerminateProcess behavior, that's the other half of #1544 I missed.
 
  Just pushed a fix: extracted the _pid_alive helper from wrap.py into a shared headroom._subprocess.pid_alive() and wired runtime_status() to use it instead of raw os.kill(pid, 0).
  
 Same psutil-first approach you described in #1609 

 I think that this  should fully close #1544. Happy to defer to #1576/#1609 if you'd rather own any of this, otherwise feel free to rebase those down to whatever isn't covered here : )

JerrettDavis (changes requested):
Thanks for moving the liveness check behind a shared helper; that is the right shape for avoiding the Windows `os.kill(pid, 0)` trap.

I still need one fix before this is safe to merge: `headroom._subprocess.pid_alive()` falls back to `os.kill(pid, 0)` whenever `psutil` cannot be imported. `psutil` is not a declared runtime dependency in `pyproject.toml`, so a normal lightweight install can hit that fallback. On Windows, that fallback is the exact destructive behavior this PR is meant to remove: CPython's `os.kill(pid, 0)` routes through `TerminateProcess`, so `runtime_status()` can still terminate a live proxy if `psutil` is absent or broken.

Please make the Windows fallback non-destructive as well. For example, require/use a Windows-safe process probe there, or return a conservative value when `psutil` is unavailable on `win32`; the important part is that `pid_alive()` must not call `os.kill(pid, 0)` on Windows under any dependency state. A regression test that simulates `sys.platform == "win32"` plus missing `psutil` and asserts `os.kill` is not called would close the gap.

Separately, GitHub marks the branch as conflicting with `main`, so it will need a merge/rebase after this is fixed.

JerrettDavis (approved):
The latest Windows fallback addresses the remaining blocker I raised. `pid_alive()` no longer reaches `os.kill(pid, 0)` on `win32` when `psutil` is unavailable, and the new regression tests cover both the kernel32 probe path and the conservative no-ctypes failure path.

The remaining `_kill_proxy_by_pid()` exception broadening also looks reasonable for stale/invalid PID handling. I did not run local tests from the main checkout because this review only needed the targeted fallback audit and the current worktree is not a clean PR checkout.

JerrettDavis (changes requested):
The Windows PID/fd-lifecycle review still looks good, but the current head is not merge-ready while the required dependency audit is red.

`Dependency audit (pip-audit)` is failing on this branch even though the focused code/test checks were otherwise green. Please refresh from current `main` or address the audited dependency finding, then rerun CI before this returns to approved.

JerrettDavis (changes requested):
Thanks for the update. The move away from `os.kill(pid, 0)` on Windows is the right direction, but I think the current fallback still needs one more guard before this is safe to merge.

Required change:

- In `_win32_pid_alive`, `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, ...)` returning a null handle is treated as `False` for all failures. On Windows, a null handle can also mean the process exists but access was denied, especially for protected/elevated/system processes. Returning `False` in that case makes callers treat a live PID as stale/dead, which is the same class of status/cleanup safety issue this PR is trying to avoid. Please inspect `GetLastError()` and treat access-denied failures conservatively as alive, while still returning false for the not-found/invalid-PID case.

Please also add a regression test that simulates `OpenProcess == 0` with an access-denied `GetLastError()` and verifies `pid_alive()` returns `True` without calling `os.kill` on `win32`.

JerrettDavis (approved):
Reviewed the latest `142ae81` update. The remaining Windows PID blocker is addressed: `_win32_pid_alive()` now treats `OpenProcess(...) == 0` plus `GetLastError() == ERROR_ACCESS_DENIED` as alive, while still avoiding `os.kill(pid, 0)` entirely on `win32` when `psutil` is unavailable or broken.

The new regression test simulates the access-denied case and asserts the Windows fallback returns `True` without calling `os.kill`. That clears the safety concern from the prior changes-requested review. GitHub currently only reports governance checks for this head, so this approval is from source-level review of the targeted Windows liveness path.

PR #1557, opened 2026-06-29
fix(proxy): forward Codex Desktop /v1/responses bytes byte-faithfully

### LINKED ISSUE #1542: [BUG] Codex Desktop zstd /v1/responses passthrough can return upstream 400
## Description

Codex Desktop requests routed through Headroom can return `400 {"detail":"Bad Request"}` on HTTP `POST /v1/responses` when the inbound request uses `content-encoding: zstd` and the request is decoded/serialized before forwarding.

In my local run, the proxy detected the request as a plain ChatGPT/Codex Desktop post and logged `Responses passthrough reason=plain_chatgpt_post mutation=disabled`, but the outbound request still went through the canonical body path as `body_mutated=true source=canonical`. The upstream ChatGPT Codex endpoint rejected those requests with 400.

A local patch that forced the plain ChatGPT/Codex Desktop bypass to forward the original decoded bytes as passthrough, and not advertise stale compression metadata, changed fresh retries to `body_mutated=false source=passthrough` and the same threads recovered with HTTP 200.

## To Reproduce

Steps observed on Windows with Codex Desktop:

1. Run Headroom proxy locally on `127.0.0.1:8787`.
2. Configure Codex Desktop to route through the proxy.
3. Send or resume a Codex Desktop turn that uses HTTP `POST /v1/responses` with headers like:
   - `originator: Codex Desktop`
   - `chatgpt-account-id: ...`
   - `accept: text/event-stream`
   - `content-encoding: zstd`
   - `content-type: application/json`
4. Observe Headroom logs for a plain ChatGPT/Codex Desktop passthrough decision followed by an outbound canonical/mutated body.
5. The upstream request can fail with `400 {"detail":"Bad Request"}` and Codex Desktop shows repeated Bad Request errors.

## Expected Behavior

For plain ChatGPT/Codex Desktop `/v1/responses` posts that Headroom decides not to mutate, the forwarded request should be byte/header-consistent and should not be reserialized through the canonical mutation path.

Concretely, one of these should be true:

- forward the original request bytes and matching compression headers unchanged, or
- forward decoded JSON bytes with `content-encoding`/`content-length` recomputed or removed.

The request should log as passthrough and not produce upstream 400s caused by body/header mismatch.

## Actual Behavior

Sanitized local logs before the fix showed this sequence:

```text
method=POST path=/v1/responses ... content-encoding=zstd ... originator="Codex Desktop"
Responses passthrough reason=plain_chatgpt_post mutation=disabled
event=outbound_request forwarder=streaming method=POST path=https://chatgpt.com/backend-api/codex/responses body_mutated=true mutation_reasons= source=canonical
Forwarding upstream streaming error status=400 url=https://chatgpt.com/backend-api/codex/responses
event=proxy_inbound_response method=POST path=/v1/responses status=400
```

After a local patch to preserve passthrough semantics for the same path, fresh retries showed:

```text
Responses passthrough reason=plain_chatgpt_post mutation=disabled
Responses memory passthrough reason=bypass_header
event=outbound_request forwarder=streaming method=POST path=https://chatgpt.com/backend-api/codex/responses body_mutated=false mutation_reasons= source=passthrough
event=proxy_inbound_response method=POST path=/v1/responses status=200
```

## Code Sample

No standalone Python repro yet. The smallest repro I currently have is a live Codex Desktop -> Headroom proxy -> ChatGPT Codex backend request with `content-encoding: zstd` on `/v1/responses`.

A useful regression test could simulate an inbound `/v1/responses` request with:

```text
originator: Codex Desktop
chatgpt-account-id: test
accept: text/event-stream
content-encoding: zstd
content-type: application/json
```

and assert that when the handler chooses the `plain_chatgpt_post`/bypass route, the streaming forwarder receives `body_mutated=false`, `source=passthrough`, and no stale `content-encoding` header if the body has already been decoded.

## Error Output

Codex Desktop displayed:

```text
stream disconnected before completion: No connection could be made because the target machine actively refused it. (os error 10061)
{"detail":"Bad Request"}
{"detail":"Bad Request"}
```

The refused connection was from a proxy restart window. The persistent Bad Request errors correlated with the upstream 400 logs above.

## Environment

- **Headroom version**: `headroom, version 0.27.0`
- **Python version**: `Python 3.14.4`
- **OS**: Windows 11 / `Microsoft Windows NT 10.0.26200.0`
- **LLM Provider**: OpenAI / ChatGPT Codex Desktop, routed to `https://chatgpt.com/backend-api/codex/responses`
- **Client**: Codex Desktop `0.142.3`

## Additional Context

Suggested hardening:

- Treat the bypass decision as an invariant: if `mutation=disabled`, the outbound streaming log should not become `body_mutated=true source=canonical`.
- Add a regression test around compressed Codex Desktop `/v1/responses` posts.
- Consider logging a warning if the proxy forwards a decoded body while preserving `content-encoding` from the original request.
- For shared local Desktop proxy usage, this failure is high-impact: one proxy restart or repeated 400s can interrupt multiple active Codex Desktop threads.

I can provide more sanitized log lines if useful. I did not open a PR because I do not yet have an upstream minimal test in this repo; this issue is based on live behavior proof and a local patch verification.

obchain:
## Description

Codex Desktop sends `POST /v1/responses` with `content-encoding: zstd`. `handle_openai_responses` decoded the body to inspect it but always re-serialized a canonical body outbound and never stripped the inbound `content-encoding` header. The request was therefore forwarded as already-decoded JSON still advertising `zstd`, and the upstream ChatGPT Codex endpoint rejected it with `400 {"detail":"Bad Request"}`.

The handler now keeps the original decoded request bytes and forwards them verbatim whenever nothing mutates the request, and drops the stale `content-encoding` header. This mirrors the byte-faithful passthrough the chat and Anthropic paths already use.

Closes #1542

## Type of Change

## Changes Made

- Read the responses body via `read_request_json_with_bytes` so the original decoded bytes are retained alongside the parsed dict.
- Track a `body_mutated` flag across the three sites that rewrite `body` (memory context injection, memory tool injection, and compression). While it stays False the original bytes are forwarded verbatim; when set the body is canonically re-serialized as before.
- Strip the inbound `content-encoding` header before forwarding, since the body bytes are already decoded.
- Pass `original_body_bytes` and `body_mutated` into both the streaming (`_stream_response`) and non-streaming (`_retry_request`) forwarders, which already support byte-faithful passthrough.

## Testing

### Test Output

```text
$ pytest tests/test_codex_responses_passthrough_bytes.py tests/test_codex_responses_waste_signals.py tests/test_responses_pyo3_compression.py tests/test_openai_responses_t3_replay_regression.py tests/test_proxy_codex_route_aliases.py tests/test_codex_openai_contract_parity.py -q
47 passed, 1 warning in 28.19s

$ ruff check headroom/proxy/handlers/openai.py tests/test_codex_responses_passthrough_bytes.py
All checks passed!
```

## Real Behavior Proof

- Environment: macOS, Python 3.13, `zstandard` 0.25.0 installed.
- Exact command / steps: new regression test posts a `content-encoding: zstd` `/v1/responses` request through the proxy with a capturing upstream stub.
- Observed result: the forwarder receives `body_mutated=False`, `original_body_bytes` equal to the client's decoded bytes, and no `content-encoding` header. Reverting the handler change makes both new tests fail (canonical re-serialization plus the stale `content-encoding` reproduce the upstream-400 condition), confirming the test guards the fix.
- Not tested: end-to-end against the live ChatGPT Codex upstream (no Codex Desktop session available); `mypy` not run locally.

## Review Readiness

## Checklist

- [] I have made corresponding changes to the documentation

## Additional Notes

Documentation change is N/A (internal proxy forwarding behavior). `mypy` was not run locally; the change adds no new annotations and stays within the existing typed forwarder signatures.


JerrettDavis (approved):
Reviewed the Responses passthrough fix. Keeping decoded original bytes and stripping stale content-encoding when the request is not mutated matches the existing byte-faithful forwarding contract, and the tests cover both zstd and plain JSON passthrough. Looks ready from the code side.

PR #691, opened 2026-06-06
fix(memory): cap local embedder thread pools

oxura:
Fixes #198.

The local sentence-transformers embedder runs `model.encode()` in an executor, but the BLAS/OpenMP/Torch pools behind it can still fan out per request. Under concurrent memory workloads that can oversubscribe the host and make unrelated health checks noisy.

This PR caps the local memory embedder path to one compute thread by default:

- sets common BLAS/OpenMP env defaults before NumPy / ONNX / Torch-backed imports
- keeps any user-provided thread settings untouched
- supports `HEADROOM_MEMORY_EMBED_THREADS=0`, `auto`, `false`, `none`, or `off` to disable Headroom's cap
- applies Torch thread setters when Torch is imported, while tolerating already-initialized pools

This is intentionally a configuration-path fix, not a broad performance rewrite. The regression tests cover the defaults, user overrides, and Torch setter behavior that prevent the oversubscription described in the issue.

Verification:

- `python -m pytest tests\test_startup_log_noise.py::TestEmbedderThreadCaps -q --basetemp=.pytest-tmp-embedder-thread-caps2`
- `python -m pytest tests\test_startup_log_noise.py -q --basetemp=.pytest-tmp-startup-log-noise-threads`
- `python -m ruff check .`
- `python -m ruff format --check .`
- `python -m mypy headroom\memory\adapters\embedders.py --ignore-missing-imports`
- `git diff --check`

PR #1559, opened 2026-06-29
fix(memory): cap local embedder CPU thread oversubscription (#198)

### LINKED ISSUE #198: [BUG] Memory embedder oversubscribes BLAS/OMP threads under concurrent load; /livez p99 spikes to 4+s
## Description

When the proxy is started with `--memory` and the local `sentence-transformers` embedder path is active, concurrent `/v1/messages` requests that each trigger `memory_context` injection cause the event loop to stall for multiple seconds — `/livez` p99 hits **4.26s** (threshold 1.5s) under 16 concurrent clients against a warm proxy.

The root cause is **thread oversubscription in the embedder**, not asyncio backpressure:

- `Embedder.embed()` already dispatches `model.encode(...)` via `run_in_executor` (see `headroom/memory/adapters/embedders.py:187-190`), so torch itself runs off the main event loop.
- However, `torch` / `sentence-transformers` use BLAS (MKL / OpenBLAS / Accelerate) + OpenMP under the covers. With no explicit cap, each `encode()` call spawns `OMP_NUM_THREADS` worker threads (default ≈ `os.cpu_count()`).
- 8 concurrent encodes × ~10 threads each = **~80 OS threads** fighting for ~10 cores. The asyncio event-loop thread gets starved by the OS scheduler even though it isn't doing the torch work itself.

Stage-timing metrics from the repro make the picture unambiguous:

```
stage_timing_ms_max{path="anthropic_messages",stage="memory_context"}        5970.21
stage_timing_ms_sum{path="anthropic_messages",stage="memory_context"}       48518.59   # over 11 samples
stage_timing_ms_max{path="anthropic_messages",stage="pre_upstream_wait"}        0.03   # not a queueing issue
```

`pre_upstream_wait` staying at ~0ms is the key diagnostic — the pre-upstream Anthropic semaphore (introduced in #172 / PR #197) is **not** the bottleneck. The slowness is inside the `memory_context` stage itself, and it's enough to starve `/livez` at the OS scheduler layer.

This is distinct from #195. #195 is a **startup-time** issue (eager ML imports push `import headroom` past the 15s `wrap` probe). This issue is a **runtime** issue: even after the proxy is fully up, embedder-heavy concurrency starves the event loop.

## To Reproduce

1. `pip install headroom-ai[ml]` (or equivalent — needs `sentence-transformers` + `torch`)
2. Start the proxy with memory enabled:
   ```
   headroom proxy --port 8799 --memory --memory-db-path /tmp/mem.db
   ```
3. Wait for `/readyz` to report `memory.initialized=true` (confirms the embedder warmup completed — the proxy is not cold).
4. Drive 16 concurrent memory-triggering clients at the proxy. A minimal harness (simplified):
   ```python
   import asyncio, httpx, time

   async def hit():
       async with httpx.AsyncClient() as c:
           await c.post("http://127.0.0.1:8799/v1/messages",
               headers={"x-api-key": "sk-ant-invalid", "anthropic-version": "2023-06-01"},
               json={"model":"claude-3-5-sonnet-20241022","max_tokens":16,
                     "messages":[{"role":"user","content":"ping"}]}, timeout=30)

   async def livez():
       async with httpx.AsyncClient() as c:
           t0=time.perf_counter(); await c.get("http://127.0.0.1:8799/livez", timeout=5)
           return (time.perf_counter()-t0)*1000

   async def main():
       task = asyncio.gather(*[hit() for _ in range(16)])
       samples = []
       while not task.done():
           samples.append(await livez()); await asyncio.sleep(0.1)
       samples.sort()
       print("livez p99 ms:", samples[int(len(samples)*0.99)], "max:", samples[-1])

   asyncio.run(main())
   ```
   (Any mix of Anthropic/OpenAI clients that each trigger `memory_context` reproduces it — the `scripts/repro_codex_replay.py` harness with `--ws-clients 8 --anthropic-clients 8` is what I used.)

## Expected Behavior

With `--memory` enabled, `/livez` should continue responding within its configured threshold (~1.5s) under realistic concurrency. The memory path should not be able to saturate the machine to the point that liveness probes are delayed by seconds.

## Actual Behavior

Under 16 concurrent memory-triggering clients:

```json
{
  "livez": { "count": 50, "p50": 2.68, "p95": 62.64, "p99": 4263.24, "max": 4263.24 }
}
```

At lower concurrency (8 clients) on the same setup, `/livez` p99 stays at **150ms** — the problem is specifically thread contention scaling superlinearly, not a constant overhead.

## Environment

- **Headroom version**: 0.5.25 (branch `fix/responses-retries-keep-compression`, which includes the #172 / PR #197 semaphore + stage-timing work; the finding is pre-existing and unrelated to those changes)
- **Python**: 3.12.12
- **OS**: macOS (Darwin 25.4.0, arm64)
- **torch**: 2.10.0 (reports 4 threads via `torch.get_num_threads()`; `os.cpu_count()` = 8)
- **sentence-transformers**: 5.2.1
- **transformers**: 5.0.0
- **`OMP_NUM_THREADS`**: unset (torch defaults apply)
- **`MKL_NUM_THREADS`**: unset
- **Memory backend**: local (SQLite + sentence-transformers embedder)
- **Model**: `sentence-transformers/all-MiniLM-L6-v2` (default)

## Additional Context

**Suggested fix A — BLAS/OMP thread cap at startup (simple, wide blast radius).**
Set `OMP_NUM_THREADS=1`, `MKL_NUM_THREADS=1`, and call `torch.set_num_threads(1)` during proxy startup when the embedder is enabled. A single-threaded encode is slower per call (~30% on small models like MiniLM-L6-v2), but N concurrent encodes then scale linearly without thread-pool thrash. This is the standard recommendation for serving frameworks that embed torch models inside an async server (vLLM, BentoML, Triton all do variants of this). Trade-off: if another part of the codebase benefits from multi-threaded BLAS, this caps it globally.

**Suggested fix B — dedicated embedder concurrency semaphore (surgical).**
Introduce `--memory-embed-concurrency` (default e.g. `max(2, cpu_count // 2)`) that wraps `Embedder.embed()` / `embed_batch()` in an `asyncio.Semaphore`. Keeps BLAS multi-threaded but caps how many concurrent encodes can be in-flight, so total thread count is bounded by `concurrency × threads_per_encode`. Does not require changing global torch / BLAS settings.

**Suggested fix C — combined.**
Ship both: default to a modest BLAS thread cap (e.g. `min(2, cpu_count)`) *and* a semaphore. Most resilient; most knobs.

**Diagnostic suggestion regardless of fix path:**
Since this PR (#197) already adds `memory_context` as a first-class `stage_timing_ms_*` histogram, any follow-up fix can be verified by re-running the repro and asserting that `stage_timing_ms_max{stage="memory_context"}` stays well under 1s under the same 16-client load.

**Why this wasn't caught by unit tests:**
Embedder unit tests use mocks or single-threaded synchronous calls, which don't exercise the BLAS/OMP oversubscription path. The failure only shows up under concurrent real inference.

**Relationship to #195 and PR #197:**
- #195 is about **startup-time** eager-import cost (`import headroom` takes 40–90s with `[ml]` extras on Windows). Fixed by deferring imports.
- This issue is about **runtime** concurrency behavior, even on a fully warm proxy.
- PR #197 (reconnect-storm resilience) already paid for the observability needed to diagnose this — `/debug/warmup`, `memory_context` stage timings, `pre_upstream_wait` stage timings. Without those, the failure mode would present as "proxy feels slow with `--memory`" with no way to attribute it.

Krishnachaitanyakc:
## Description

The torch/sentence-transformers `LocalEmbedder` ran encodes on the shared default executor with **no BLAS/OpenMP thread cap**. Under concurrent load each `encode()` fans out to ~`os.cpu_count()` BLAS/OpenMP threads, so N in-flight encodes spawn ~`N × cpu_count` OS threads — oversubscribing the CPU, slowing the `memory_context` stage and (on smaller boxes) starving the asyncio event loop. The ONNX embedder already bounds its threads (`create_cpu_session_options(intra_op_num_threads=1, inter_op_num_threads=1)`); this brings the torch path to parity.

Supersedes #691 by @oxura — closed only for the open-PR cap, with an explicit invitation to resubmit; no technical objection was raised, and its CI was fully green. Credit to @oxura for the original diagnosis and fix. That PR capped threads by setting BLAS/OpenMP env vars at import time plus `torch.set_num_threads`; this PR instead runs CPU encodes on a dedicated, size-limited executor whose workers each pin their thread pool — which additionally bounds in-flight encode concurrency (the issue's Fix B/C) and keeps the cap contained to the embedder rather than mutating process-global env at import.

Closes #198

## Type of Change

- [x] Performance improvement

## Changes Made

- CPU encodes now run on a **dedicated, size-limited executor** whose worker `initializer` pins each worker's torch intra-op pool (and sets BLAS/OpenMP env defaults). torch's OpenMP thread count is per-thread, so a one-shot cap misses pooled executor workers — the per-worker initializer caps every worker deterministically.
- Total embedding threads are bounded by `HEADROOM_EMBED_CONCURRENCY` (default `min(4, os.cpu_count())`) × `HEADROOM_EMBED_NUM_THREADS` (default `1`); invalid/non-positive values fall back safely (≥1).
- Mirrors the existing MPS dedicated-single-worker-executor pattern; CUDA keeps the shared default executor (GPU compute is off-CPU). `setdefault` never overrides an operator's explicit `OMP_NUM_THREADS`.

## Testing

### Test Output

```text
$ uv run pytest tests/test_memory/test_embedder_thread_cap.py tests/test_memory/test_embedder_mps_serialization.py -q
13 passed

$ uv run pytest tests/test_memory/ tests/test_cli_proxy_embedding_server.py -q
533 passed        # no regressions from the executor change

$ uv run ruff check .  &&  uv run ruff format --check .
All checks passed!   /   1016 files already formatted

$ uv run mypy headroom --ignore-missing-imports
Success: no issues found in 404 source files
```

New `tests/test_memory/test_embedder_thread_cap.py`: env resolution for both knobs (default / positive / invalid / clamped), worker-init env application + operator-override safety, and a behavioral test that loads the real CPU embedder and asserts every executor worker is pinned to the configured intra-op thread count. Updated `test_embedder_mps_serialization.py` to the new CPU contract.

## Real Behavior Proof

- Environment: built this branch into a CPU-only Linux container, removed `onnxruntime` so the proxy falls back to the torch `LocalEmbedder`; a container has no MPS/CUDA, so it resolves to `device=cpu` — the deployment where #198 occurs. Python 3.12, torch 2.12.1, `all-MiniLM-L6-v2`, container capped to 4 CPUs, 32 concurrent clients.
- Exact command / steps: `headroom proxy --host 0.0.0.0 --memory` in-container; a concurrent `/v1/messages` driver from the host (invalid key — `memory_context` runs before the upstream call); measured the `memory_context` stage from `/metrics` before vs after the cap.
- Observed result: the embedder stage this PR targets improved — `memory_context` avg 73.5 ms → 58.7 ms and max 279 ms → 242 ms (uncapped 12×8 = 96 threads vs fix 4×1): ~20% faster and steadier inside the real proxy. Isolated component benchmarks (heavy concurrent `embed_batch`; `LocalBackend.search_memories`) show a larger effect — tail event-loop stall ~16–24 ms → ~3 ms, and search throughput +57%. Unit/regression: 13 new tests + 533 memory-suite tests pass; `ruff` + `mypy` clean.
- Not tested: the issue's absolute multi-second `/livez` spike. On my hardware/synthetic load, `/livez` stalls were dominated by the upstream-connection path (invalid-key DNS/TLS), not the ~250 ms `memory_context` stage, so I can't attribute the multi-second figure to the embedder here — the original report was on an 8-core box with real Claude Code transcripts that drove `memory_context` itself to several seconds. Linux/CUDA hardware not exercised; no live LLM provider used; ONNX path unchanged. This PR removes the documented thread oversubscription and brings the torch path to ONNX parity; it does not claim to single-handedly resolve the 4 s figure.

Measured `memory_context` stage timing (real containerized proxy, torch CPU embedder, 4 CPUs, 32 concurrent clients):

| `memory_context` | avg | max |
|---|---|---|
| Before (uncapped, 12×8 = 96 threads) | 73.5 ms | 279 ms |
| After (fix, 4×1) | 58.7 ms | 242 ms |

## Review Readiness

## Additional Notes

Default-behavior change: CPU encodes use a dedicated bounded pool instead of the shared default executor (`close()` tears it down). Both knobs are opt-in overrides with safe defaults. No new dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `<headroom_proactive_expansion>` XML tags, giving downstream consumers
   (LLMs, loggers, attribution parsers) a machine-readable provenance
   boundary and preventing misattribution in multi-agent threads.
+- **memory/embedder:** cap CPU thread oversubscription in the local
+  torch/sentence-transformers embedder. Concurrent encodes previously each
+  fanned out to ~`os.cpu_count()` BLAS/OpenMP threads, so under load the memory
+  path starved the asyncio event loop and spiked `/livez` latency to several
+  seconds. CPU encodes now run on a dedicated, size-limited executor whose
+  workers each pin their thread pool, bounding total embedding threads to
+  `HEADROOM_EMBED_CONCURRENCY` × `HEADROOM_EMBED_NUM_THREADS` (defaults
+  `min(4, cpu)` × 1). The ONNX embedder already capped its threads; this brings
+  the torch path to parity
+  ([#198](https://github.com/headroomlabs-ai/headroom/issues/198)).
 
 ### Changed
 
--- a/headroom/memory/adapters/embedders.py
+++ b/headroom/memory/adapters/embedders.py
@@ -42,6 +42,102 @@ logging.getLogger("httpx").setLevel(logging.WARNING)
 logger = logging.getLogger(__name__)
 
 
+# =============================================================================
+# Local (torch / sentence-transformers) CPU thread cap — issue #198
+# =============================================================================
+# A long-lived proxy serves many requests concurrently. Each torch ``encode()``
+# fans out to BLAS (MKL / OpenBLAS / Accelerate) + OpenMP worker threads, which
+# default to roughly ``os.cpu_count()``. Under concurrency this oversubscribes
+# the CPU — N in-flight encodes x ~cpu_count threads each thrash the scheduler
+# and starve the asyncio event loop, so liveness probes (``/livez``) spike to
+# multiple seconds even though the loop itself is idle (issue #198).
+#
+# Capping intra-op parallelism makes a single encode modestly slower but lets
+# concurrent encodes scale linearly without thread-pool thrash — the standard
+# trade-off for serving torch models inside an async server. The ONNX embedder
+# already caps its threads (see ``onnx_runtime.create_cpu_session_options``);
+# this brings the torch path to parity.
+#
+# torch's OpenMP thread count is per-thread, and encodes run on executor worker
+# threads, so a one-shot cap would miss most workers. Instead, CPU encodes run
+# on a dedicated, size-limited executor whose ``initializer`` pins each worker's
+# thread pool once. Total embedding threads are then bounded by
+# ``workers (HEADROOM_EMBED_CONCURRENCY) x threads-per-encode
+# (HEADROOM_EMBED_NUM_THREADS)``. Applies to the CPU device only (GPU/MPS do
+# their compute off-CPU).
+_EMBED_THREADS_ENV = "HEADROOM_EMBED_NUM_THREADS"
+_DEFAULT_EMBED_THREADS = 1
+_EMBED_CONCURRENCY_ENV = "HEADROOM_EMBED_CONCURRENCY"
+_DEFAULT_EMBED_CONCURRENCY = 4
+_BLAS_THREAD_ENV_VARS = (
+    "OMP_NUM_THREADS",
+    "OPENBLAS_NUM_THREADS",
+    "MKL_NUM_THREADS",
+    "NUMEXPR_NUM_THREADS",
+    "VECLIB_MAXIMUM_THREADS",
+)
+
+
+def _resolve_positive_int_env(env_var: str, default: int) -> int:
+    """Read a positive integer from ``env_var``, falling back to ``default``.
+
+    A non-positive or unparseable value logs a warning and returns a safe value
+    (>= 1) rather than disabling the limit.
+    """
+    raw = os.environ.get(env_var)
+    if raw is None:
+        return default
+    try:
+        value = int(raw)
+    except (TypeError, ValueError):
+        logger.warning("Invalid %s=%r; falling back to %d.", env_var, raw, default)
+        return default
+    if value < 1:
+        logger.warning("%s=%d is below 1; using 1.", env_var, value)
+        return 1
+    return value
+
+
+def _resolve_embed_thread_cap() -> int:
+    """Resolve the per-encode CPU thread cap (``HEADROOM_EMBED_NUM_THREADS``)."""
+    return _resolve_positive_int_env(_EMBED_THREADS_ENV, _DEFAULT_EMBED_THREADS)
+
+
+def _resolve_embed_concurrency() -> int:
+    """Resolve the max concurrent CPU encodes (``HEADROOM_EMBED_CONCURRENCY``).
+
+    Defaults to ``min(4, os.cpu_count())`` so embedding cannot occupy every core
+    and starve the event loop, while still allowing useful parallelism.
+    """
+    cpu = os.cpu_count() or 1
+    raw = os.environ.get(_EMBED_CONCURRENCY_ENV)
+    if raw is None:
+        return max(1, min(_DEFAULT_EMBED_CONCURRENCY, cpu))
+    return _resolve_positive_int_env(
+        _EMBED_CONCURRENCY_ENV, max(1, min(_DEFAULT_EMBED_CONCURRENCY, cpu))
+    )
+
+
+def _init_cpu_embed_worker() -> None:
+    """Pin a CPU embed worker's thread pool (runs once per worker; issue #198).
+
+    Sets BLAS/OpenMP env defaults (``setdefault`` never overrides an operator's
+    explicit setting) and bounds torch's intra-op pool for this worker thread.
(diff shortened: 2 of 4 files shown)


JerrettDavis (approved):
Reviewed the CPU embedder cap. The dedicated executor is scoped to CPU, MPS serialization remains intact, env parsing has safe fallbacks, and close tears down the executor. The tests cover both env resolution and the actual worker thread cap. Looks ready from my pass.

PR #1563, opened 2026-06-29
fix(transforms): bound native content detection with a Windows watchdog (#575)

### LINKED ISSUE #575: [BUG] headroom_compress / compress() hangs forever on first call (Windows) — synchronization deadlock inside headroom._core (WaitOnAddress) during first detect_content_type
[headroom_issue_compress_hang.md](https://github.com/user-attachments/files/28581220/headroom_issue_compress_hang.md)

Parideboy:
## Description

On Windows, the first call into the native `headroom._core.detect_content_type` can park forever in an ort/`Once` initialization (`WaitOnAddress`) at 0% CPU. A wedged native call cannot be cancelled from Python, so it deadlocks the caller. In the proxy it is worse: each affected request permanently consumes a compression-executor worker, eventually saturating the pool (`running == max_workers`, `leaked_threads_total == 0` because the worker never finishes) and stalling every subsequent request for the full `COMPRESSION_TIMEOUT_SECONDS` before passthrough.

The Rust backend is already off by default on Windows — `_resolve_detect_backend()` returns `"python"` there — but an explicit `HEADROOM_DETECT_BACKEND=rust`, or any future regression of that default, re-exposes the hang with no escape hatch. This implements the issue's third ask: a timeout/watchdog so a hung native init degrades gracefully instead of deadlocking the agent / MCP server / proxy.

Closes #575

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added a Windows-only watchdog around the native detect call in `transforms/content_router.py`. `_rust_detect_watchdogged()` runs `detect_content_type` on a daemon thread and bounds the caller's wait; on timeout it raises `TimeoutError`, which the existing `except BaseException` handler degrades to the pure-Python regex detector. Detection therefore always returns instead of deadlocking (and, in the proxy, instead of permanently consuming a compression-executor worker).
- Added `_detect_timeout_secs()` reading `HEADROOM_DETECT_TIMEOUT_SECS` (default 5s; blank / non-numeric / non-positive values fall back to the default).
- Gated the watchdog to `sys.platform == "win32"` — the only platform where the hang is observed. Other platforms keep the direct native call with no per-call thread overhead (the trusted hot path is unchanged).
- Added regression tests for the watchdog, env parsing, error relay, the Windows degrade-on-hang path, and the Windows happy path.

## Testing

### Test Output

```text
$ ruff check headroom/transforms/content_router.py tests/test_transforms_content_router.py
All checks passed!

$ ruff format --check headroom/transforms/content_router.py tests/test_transforms_content_router.py
2 files already formatted

(exit 0)

$ pytest tests/test_transforms_content_router.py -q
33 passed
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13, ruff 0.15.17 / mypy 1.20.2 / pytest 9.1.0, `headroom._core` built locally, branch `fix/575-native-detect-watchdog`.
- Exact command / steps: ran the four checks above. `test_detect_content_watchdog_degrades_on_windows_hang` forces `HEADROOM_DETECT_BACKEND=rust`, patches `sys.platform` to `"win32"`, sets `HEADROOM_DETECT_TIMEOUT_SECS=0.1`, and injects a native `detect_content_type` that blocks on an `Event` (simulating the `WaitOnAddress` park, GIL released) — then asserts detection still returns. The companion tests cover env parsing, error relay through the watchdog, and the fast-native Windows path.
- Observed result: with a hung native detector, `_detect_content('[{"id": 1}]')` returns `ContentType.JSON_ARRAY` (the pure-Python degrade path) within the 0.1s budget instead of hanging; with a fast native detector on Windows it returns the native result unchanged; non-Windows behavior (direct call) is untouched and the existing rust-delegation test still passes. All 33 tests in the file pass; ruff / format / mypy clean.
- Not tested: the live `from headroom._core import detect_content_type; detect_content_type("hello world")` deadlock on an affected Windows 11 24H2 machine was not reproduced end to end (it requires the specific System32 ONNX Runtime build). The fix is instead covered by the deterministic hung-detector injection test, which exercises the exact degrade path the watchdog adds. This PR does not attempt the Rust-side fix for the underlying first-call init deadlock (asks #1) — it is the Python-side watchdog (ask #3); the existing `HEADROOM_DETECT_BACKEND` flag already covers ask #2.

## Review Readiness

## Checklist

## Additional Notes

- The watchdog cannot cancel a wedged native call (no portable way to kill a thread blocked in C). It frees the *caller* and leaves the stuck daemon thread to die with the process; this is marked with a `ponytail:` comment naming the upgrade path (the Rust-side non-blocking first-call init). For the saturation scenario this is still a strict improvement: callers no longer block indefinitely, so the executor drains instead of wedging permanently.

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -42,6 +42,7 @@ import math
 import os
 import re
 import sys
+import threading
 import time
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The Windows watchdog is the right mitigation for the immediate deadlock, but it needs one more operational guard before I’d be comfortable merging it. If the native detector hangs once, subsequent calls will still try the same native path, wait the full watchdog budget, and leave another daemon thread blocked in the native call. Under proxy traffic that turns the fix into repeated 5s latency plus unbounded stuck-thread growth. Please add a process-level circuit breaker after the first watchdog timeout (for example mark native detection unhealthy and route future detections directly to the Python fallback), and cover that with a test that two Windows detections after a simulated hang only create/enter the native detector once.

JerrettDavis (approved):
Reviewed the latest circuit-breaker update. This addresses the blocker I raised.

The watchdog is now Windows-scoped, the first timeout disables the native detector process-wide, and subsequent calls go straight to the pure-Python detector instead of spending the full timeout and leaving another stuck daemon thread behind. The tests cover env parsing, fast native success, native exception propagation, timeout degrade, and the circuit breaker skip path.

I attempted a local targeted run of `python -m pytest tests/test_transforms_content_router.py -q`, but this review worktree cannot import `headroom._core` locally. The full GitHub CI matrix is green for this PR.

Issue #1567, opened 2026-06-29
Add Prometheus counters for fail-open compression failures and the kompress size-gate


rocker-zhang:
Two compression behaviors are currently invisible on /metrics:

1. When optimization throws, the proxy fails open and forwards the request uncompressed. The only signal is a response header (x-headroom-compression-failed); there is no counter, so ratio can bleed at this path with nothing in /metrics. A timeout there means the budget was too tight; any other exception is a different class of problem, and the two are worth distinguishing.

2. The kompress size-gate (#1171) routes oversized blocks off ModernBERT, but there is no way to tell from /metrics whether it ever fires on real traffic.

Proposal: two defaultdict-backed counters following the existing record_compression pattern:
- headroom_compression_failed_total{reason=timeout|error}
- headroom_kompress_size_gate_total{outcome=within|exceeded}

I have a PR drafted with tests if this looks useful.

PR #1569, opened 2026-06-29
feat(metrics): export compression-failed and kompress size-gate counters

rocker-zhang:
## What

Two Prometheus counters that make previously-invisible compression behavior measurable on `/metrics`.

- `headroom_compression_failed_total{reason=timeout|error}` — incremented at both Anthropic fail-open sites (single-message and batch-create), where an optimization exception forwards the request uncompressed. Before this, ratio could bleed at these sites with nothing in `/metrics`; only a response header recorded the single-message case. The timeout/error split separates "compression budget too tight" from "real bug".
- `headroom_kompress_size_gate_total{outcome=within|exceeded}` — the size gate (#1171) routes oversized blocks off ModernBERT. The within/exceeded split proves whether the gate ever fires on real traffic. `within` counts a gate pass, not whether ML compression then ran.

## How

Both reuse the existing `PrometheusMetrics` singleton and the established `defaultdict(int)` counter + text-exposition pattern. The handler records via `self.metrics`; `content_router` records through the existing `CompressionObserver` hook to avoid an import cycle. Cleared in `reset_runtime`; exposition blocks are emitted only when non-empty.

## Verification

- `tests/test_prometheus_obs_counters.py` (6 tests): per-reason/outcome bucketing, empty-string default buckets, exposition format, conditional absence until recorded, and reset clearing. All green.
- Counter increments and well-formed exposition (HELP/TYPE balanced, labels escaped) confirmed by direct exercise; gate `within`/`exceeded` shown mutually exclusive across the eligible-block call sites.

Single commit, rebased on current `main`.

Addresses #1567.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -1274,6 +1274,14 @@ class AnthropicHandlerMixin:
                     logger.warning(f"[{request_id}] Optimization failed: {type(e).__name__}: {e}")
                     # Flag compression failure for observability
                     _compression_failed = True
+                    # Split timeout from other errors: a timeout means the
+                    # compression budget was too tight, not a code bug.
+                    reason = (
+                        "timeout"
+                        if isinstance(e, (asyncio.TimeoutError, TimeoutError))
+                        else "error"
+                    )
+                    self.metrics.record_compression_failed(reason)
 
             # Guard: if "optimization" inflated tokens, revert to originals.
             # Skip in cache mode where prefix-stability may legitimately shift counts.
@@ -3002,6 +3010,12 @@ class AnthropicHandlerMixin:
                 logger.warning(
                     f"[{request_id}] Optimization failed for batch request '{custom_id}': {e}"
                 )
+                # Same fail-open accounting as the single-message path: a
+                # timeout means the budget was too tight, not a code bug.
+                reason = (
+                    "timeout" if isinstance(e, (asyncio.TimeoutError, TimeoutError)) else "error"
+                )
+                self.metrics.record_compression_failed(reason)
                 # Pass through unchanged on failure
                 compressed_requests.append(batch_req)
                 total_optimized_tokens += original_tokens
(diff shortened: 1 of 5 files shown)


JerrettDavis (changes requested):
Thanks for adding observability here; the metric families themselves are a useful addition. One blocker from the current CI run: the Anthropic fail-open path now unconditionally calls self.metrics.record_compression_failed(reason), and tests/test_anthropic_pre_upstream_backpressure.py::test_compression_is_not_bypassed_when_gated fails because that handler uses a _DummyMetrics without the new method. More importantly, this means an observability write can now turn an optimization failure into a request failure in any test/alternate handler that provides a minimal metrics object. Please make this recording defensive in the same spirit as the router observer path, or update the shared metric test double contract so fail-open remains fail-open. CI also shows ruff format --check wants headroom/proxy/handlers/anthropic.py reformatted.

rocker-zhang:
Thanks, good catch. The fail-open path calls `record_compression_failed` synchronously, and the `_DummyMetrics` stub in `test_anthropic_pre_upstream_backpressure.py` didn't define it. I've added a no-op `record_compression_failed` to the stub (the counter is sync + lock-free like `record_compression`) and run `ruff format`. Pushed.

rocker-zhang:
Rebased onto 0.28.0 and force-pushed; CI is green now, including ruff format on the handler. As noted above, the fail-open path stays fail-open through the test double: _DummyMetrics defines record_compression_failed as a no-op. I left the production call unguarded because self.metrics is the full PrometheusMetrics on every real handler path and the surrounding metric calls follow the same assumption, but I can switch to a defensive guard at the call site if you prefer that over the test-double contract. Ready for another pass.

JerrettDavis (approved):
Reviewed the latest update. This clears my prior blocker.

The fail-open metric path now has the expected metrics method on the exercised handler/test-double contract, the counters are guarded for executor-thread writers versus event-loop export/reset readers, and the Kompress size-gate observer remains defensive so observer failures do not break compression. The full GitHub matrix is green.

Local focused check: `python -m pytest tests/test_prometheus_obs_counters.py -q` passed (7 tests).

Issue #1568, opened 2026-06-29
Retry exhaustion collapses upstream 5xx to 502 and inflates the save-rate


rocker-zhang:
Two related problems on the retry-exhaustion path:

1. When the upstream returns a retryable 5xx (529 Overloaded, 503) and retries are exhausted, _retry_request raises and the caller turns it into a generic 502. Clients see 502 and give up instead of applying their own overload backoff; the real status is lost.

2. Once the real 5xx is surfaced, it flows through the success funnel and is recorded via record_request, feeding the savings/cost stats and inflating the save-rate. An exhausted upstream failure should not count as a served request.

Proposal (two small PRs):
- Preserve the real upstream status/body on exhaustion (provider-agnostic, in _retry_request).
- Carry the status onto RequestOutcome and, when status >= 500, record a failed request and skip the success funnel; thread the status at every provider's retry-fed outcome site.

Both drafted with tests.

JerrettDavis:
Fixed by merged PRs #1570 and #1571. Retry exhaustion now preserves the upstream 5xx status and response body, while `RequestOutcome` routes exhausted 5xx responses to failed-request metrics and skips the savings/cost success funnel across Anthropic, OpenAI, and Gemini paths. Later streaming/batch fixes apply the same accounting invariant there.

Re-verified the status-preservation, provider wiring, and outcome-accounting suites on current `main`: `18 passed`.

PR #1570, opened 2026-06-29
fix(proxy): preserve upstream 5xx status on retry exhaustion

rocker-zhang:
## What

When the upstream returns a retryable 5xx (529 Overloaded, 503), `_retry_request` retried up to the cap and then raised, which the caller collapsed into a generic 502. That hides the retryable signal: clients see a 502 and give up instead of applying their own overload backoff.

On exhaustion, return the last upstream response (preserving its status and body) when one is available. Connection and timeout errors still raise — only an `HTTPStatusError` carrying a real upstream response is surfaced.

## Why this scope

`_retry_request` is provider-agnostic, so this fix applies uniformly to all providers (no per-handler change needed). It is purely a returned-status correctness fix and does not touch request accounting — a separate change handles counting an exhausted 5xx as a failed request across all provider handlers.

## Verification

`tests/test_retry_preserve_upstream_status.py`: 529/503 status+body preservation, 4xx no-retry, connect-error still raises, success passthrough. Against unpatched main the 503-preservation test fails (collapses to 502); with the fix all pass.

Addresses #1568.

### DIFF
--- a/headroom/proxy/server.py
+++ b/headroom/proxy/server.py
@@ -1802,6 +1802,16 @@ class HeadroomProxy(
                 last_error = e
 
                 if not self.config.retry_enabled or attempt >= self.config.retry_max_attempts - 1:
+                    # On exhaustion, preserve the upstream 5xx status (e.g. 503
+                    # Service Unavailable, 500, 502, 504) so the client can apply
+                    # its own retry/backoff. Collapsing every exhausted 5xx into a
+                    # generic 502 hides the retryable signal and makes clients give
+                    # up. The 429/529 overload statuses are already returned
+                    # verbatim by the RETRYABLE_OVERLOAD_STATUSES branch above and
+                    # never reach here. ConnectError/TimeoutException carry no
+                    # response, so those still raise.
+                    if isinstance(e, httpx.HTTPStatusError) and e.response is not None:
+                        return e.response
                     raise
 
                 # Exponential backoff with jitter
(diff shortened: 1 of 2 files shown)


JerrettDavis (commented):
I reviewed the retry exhaustion change and the behavior looks aligned with the goal: preserving the final upstream 5xx response only when there is a response, while connection/timeout failures still raise. The remaining blocker I see is mechanical: CI has lint red because ruff format --check would reformat tests/test_retry_preserve_upstream_status.py. Once formatted, this should be ready for another pass.

rocker-zhang:
Reformatted tests/test_retry_preserve_upstream_status.py with ruff and rebased onto 0.28.0; lint is green now. One note from the rebase: 0.28.0 added the 429/529 overload path (#1495) that returns those statuses verbatim on exhaustion, so this change now covers the non-overload 5xx only (500/502/503/504) and complements that path instead of overlapping it. I narrowed the inline comment and the tests to match. Ready for another pass.

JerrettDavis (approved):
Reviewed the retry-exhaustion update. The behavior is now pinned in the right place: exhausted upstream HTTP 5xx responses preserve the final upstream response/status/body, while transport/connect failures still raise because there is no upstream response to forward. CI is green.

I attempted the targeted local test, but this worktree cannot import the native `headroom._core` extension required by `headroom.proxy.server`; the remote matrix covers the added test file successfully.

PR #1571, opened 2026-06-29
fix(proxy): count exhausted upstream 5xx as failed across all providers

rocker-zhang:
## What

A companion to the retry-exhaustion change that returns the real upstream 5xx (e.g. 529 Overloaded) instead of a synthetic 502. Once the real 5xx is returned, it flows through the success funnel and is recorded via `record_request`, feeding the savings/cost stats and inflating the save-rate.

`RequestOutcome` now carries the upstream `status_code` (default 200). In `emit_request_outcome`, a `status_code >= 500` records a failed request and returns before the savings/cost/log success path. 4xx stay on the normal funnel (client errors the proxy still served).

The real status is threaded onto the retry-fed `RequestOutcome` at every provider site: Anthropic (message, batch, passthrough), OpenAI (chat, responses, passthrough), Gemini (generateContent, all-non-text path, countTokens). Sites that cannot carry a 5xx keep the default 200: local cache hits, backend-routed paths that early-return on error, websocket units (no HTTP status), and streaming generators that early-return on `>= 400`.

## Scope notes

- **429**: an exhausted 429 (rate limit) currently stays on the success funnel since it is < 500. Extending the failed-accounting to exhausted-429 is a separate follow-up, kept out of this 5xx-scoped change.
- **Streaming**: streaming responses return before the retry-exhaustion logic, so they do not receive an exhausted 5xx through this path; `from_stream` is unchanged.

## Verification

`tests/test_outcome_records_5xx_as_failed.py` exercises the `>= 500` funnel guard; `tests/test_5xx_accounting_all_providers.py` pins the per-provider contract for each wired site. Against unpatched main the guard test fails (no `status_code` field); with the change all pass.

Addresses #1568. Builds on #1570 (preserve-5xx-status): the 503 accounting takes effect once that lands; the 429/529 accounting is independent.


JerrettDavis (commented):
I reviewed the outcome accounting change and the central status_code >= 500 guard is the right place for this behavior. CI is currently blocked on formatting only: ruff format --check wants tests/test_5xx_accounting_all_providers.py reformatted. After that, I would re-check that the handler-level coverage remains focused on at least one real retry-fed path, since the rest of the parametrized cases exercise the shared outcome contract rather than each provider call site directly.

rocker-zhang:
Reformatted tests/test_5xx_accounting_all_providers.py and rebased onto 0.28.0; lint is green. On the retry-fed coverage: the handler-level case drives a fake transport through real retry exhaustion and asserts transport.calls == 2 before checking the outcome is recorded as failed, so at least one path exercises the real provider call site rather than only the shared outcome contract. The parametrized cases cover the shared contract on top of that.

JerrettDavis (approved):
Reviewed the exhausted-5xx accounting change. The central `RequestOutcome.status_code >= 500` guard is the right boundary: failed upstream responses now increment failed metrics and skip the savings/cost success funnel, while 4xx client errors continue through the normal served-request path. The handler call sites now thread the real response status into the outcome, and CI is green.

Local focused check: `python -m pytest tests/test_outcome_records_5xx_as_failed.py -q` passed (2 tests). The broader handler-level test imports the full proxy server and is covered by GitHub CI in this environment.

Issue #1555, opened 2026-06-29
[BUG] File handle leak in _start_proxy_subprocess if _check_proxy raises during readiness loop


guyoron1:
## Summary

`_start_proxy_subprocess()` in `headroom/cli/wrap.py` opens a log file at line 440 and manually closes it in the success, process-death, and timeout paths. However, if `_check_proxy()` raises an unexpected exception during the readiness loop (lines 476-490), the file handle leaks because there is no `try/finally` wrapper.

## Location

`headroom/cli/wrap.py:440-494`

```python
stdio_log_file = open(stdio_log_path, "a", encoding="utf-8")  # line 440

proc = subprocess.Popen(cmd, stdout=stdio_log_file, ...)       # line 465

for _i in range(timeout_seconds):
    time.sleep(1)
    if _check_proxy(port):          # if this raises...
        stdio_log_file.close()      # ...this is skipped
        return proc
    if proc.poll() is not None:
        stdio_log_file.close()      # ...and this is skipped
        raise RuntimeError(...)

proc.kill()
stdio_log_file.close()              # ...and this too
raise RuntimeError(...)
```

## Impact

- If `_check_proxy(port)` raises an unexpected exception, the file handle leaks
- Repeated `headroom wrap` invocations with transient probe failures can exhaust file descriptors

## Suggested Fix

Wrap the readiness loop in `try/finally`:

```python
stdio_log_file = open(stdio_log_path, "a", encoding="utf-8")
try:
    proc = subprocess.Popen(cmd, stdout=stdio_log_file, stderr=stdio_log_file, ...)
    for _i in range(timeout_seconds):
        time.sleep(1)
        if _check_proxy(port):
            return proc
        if proc.poll() is not None:
            raise RuntimeError(...)
    proc.kill()
    raise RuntimeError(...)
finally:
    stdio_log_file.close()
```

## Environment

Affects all platforms. Discovered via code review.

abhay-codes07:
Opened #1605 with the fix.

Wrapped the `_start_proxy` readiness loop in `try/finally` (exactly as you suggested) and dropped the three per-branch `close()` calls so the stdio-log fd is closed on every exit path, including when `_check_proxy()` raises mid-loop.

[...]

PR #1576, opened 2026-06-30
fix(install): close parent log fd in start_detached_agent

### LINKED ISSUE #1554: [BUG] File descriptor leak in start_detached_agent — parent never closes log_file after Popen
## Summary

`start_detached_agent()` in `headroom/install/runtime.py` opens a log file at line 269 and passes it to `subprocess.Popen()`, but the parent process never closes its copy of the file descriptor. Every `headroom install start` invocation leaks one file descriptor in the parent process.

## Location

`headroom/install/runtime.py:269`

```python
def start_detached_agent(profile: str) -> subprocess.Popen[str]:
    # ...
    log_file = open(log_file_path, "a", encoding="utf-8", errors="replace")
    # ...
    return subprocess.Popen(command, **kwargs)
    # log_file is never closed by the parent
```

## Impact

The subprocess inherits the fd, but the **parent** also holds a copy. Since the parent's reference is never closed:
- Each invocation of `headroom install start` leaks one file descriptor
- On systems with tight ulimit or in containers, this can exhaust file descriptor limits
- The leaked fd also holds the log file open, potentially preventing log rotation

## Suggested Fix

Close the parent's copy after `Popen()` returns:

```python
proc = subprocess.Popen(command, **kwargs)
log_file.close()  # parent no longer needs the fd — subprocess inherited it
return proc
```

## Environment

Affects all platforms. Discovered via code review.

abhay-codes07:
## Description

`start_detached_agent()` opens the agent log file and hands it to
`subprocess.Popen` as `stdout`/`stderr`, then returns the process **without
closing the parent's copy of the file descriptor**. The child inherits the fd
and writes to it, but the parent keeps its own copy open forever.

The result: every `headroom install start` leaks one file descriptor in the
parent, and the leaked handle pins the log file open so it can't be rotated.
On a tight `ulimit` or inside a container, repeated starts can walk straight
into the fd limit.

```python
# headroom/install/runtime.py — before
log_file = open(log_file_path, "a", encoding="utf-8", errors="replace")
kwargs = {"stdout": log_file, "stderr": log_file, ...}
return subprocess.Popen(command, **kwargs)   # parent's log_file never closed
```

The fix closes the parent's copy in a `try/finally` right after `Popen` returns:

```python
try:
    proc = subprocess.Popen(command, **kwargs)
finally:
    # The child has inherited the log file descriptor, so the parent's
    # copy is dead weight. Closing it (even when Popen raises) avoids
    # leaking one fd per `headroom install start` and lets the log file
    # be rotated.
    log_file.close()
return proc
```

The `finally` is deliberate: it also covers the case where `Popen` itself
raises (bad executable, fork failure), which would otherwise leak the
just-opened handle. This matches the `with open(...)` pattern already used by
`run_foreground()` a few lines above.

Closes #1554

## Type of Change

## Changes Made

- `headroom/install/runtime.py`: close the parent's log file descriptor in a `try/finally` after `subprocess.Popen` in `start_detached_agent()`, so it is released on the normal path and when `Popen` raises.
- `tests/test_install/test_runtime.py`: add two regression tests — one for a normal start, one for `Popen` raising — asserting the parent's log handle is closed afterwards.

## Testing

### Test Output

Before the fix (`runtime.py` reverted to its parent commit, new tests kept) —
the assertion inspects the *actual* log file handle and finds it still open:

```text
E   AssertionError: assert False is True
E    +  where False = <_io.TextIOWrapper name='...\deploy\demo\runner.log' mode='a' encoding='utf-8'>.closed
FAILED tests/test_install/test_runtime.py::test_start_detached_agent_closes_parent_log_fd
FAILED tests/test_install/test_runtime.py::test_start_detached_agent_closes_log_fd_when_popen_raises
============================== 2 failed in 0.43s ==============================
```

After the fix:

```text
tests\test_install\test_runtime.py ................
====================== 16 passed, 1 deselected in 0.35s =======================
```

(The one deselected test, `test_runtime_start_lock_blocks_another_process`, is a
pre-existing failure on my Windows box — it fails identically on a clean
checkout of `main` and is unrelated to this change.)

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.11, headroom built from this branch (`uv sync --extra dev`).
- Exact command / steps: ran the two regression tests, which drive the real `start_detached_agent` code path — real `open()`, real `Popen(stdout=...)`, real (or missing) `close()`. Only `subprocess.Popen` is stubbed so the test never launches an actual detached agent; the fd-lifecycle bug lives entirely in how the parent handles its own handle, and that runs for real.
- Observed result: the log file handle the parent passed to `Popen` is `.closed == False` before the fix and `.closed == True` after — for both the normal path and the `Popen`-raises path (output above).
- Not tested: I intentionally did not spin up many real detached agents to watch the OS fd table grow — on Windows that means flashing console windows and isn't a clean signal anyway. The handle-state assertion on the real file object is the deterministic equivalent. Did not run the full `mypy headroom` pass (one-line lifecycle change, no new types).

## Review Readiness

## Checklist

## Additional Notes

- Single logical change, no new dependencies, default behavior otherwise unchanged.
- Rebased/merged latest `main` to clear a `CHANGELOG.md` conflict.
- @chopratejas this is a sibling of #1555 (the `wrap.py` readiness-loop handle leak you've got filed). I scoped this PR to #1554 only; happy to follow up on #1555 separately if you'd like.


JerrettDavis (commented):
The fd-leak fix itself is correct and the new focused regressions pass locally:

`python -m pytest tests/test_install/test_runtime.py::test_start_detached_agent_closes_parent_log_fd tests/test_install/test_runtime.py::test_start_detached_agent_closes_log_fd_when_popen_raises -q`

Result: 2 passed.

One process note: this is now largely superseded by the broader runtime/fd handling in #1556, which also covers related Windows PID/SystemError paths. If #1556 lands first, this PR should probably be closed or rebased down to whatever remains unique. I do not see a correctness issue in this isolated patch.

JerrettDavis (approved):

abhay-codes07:
Thanks for reviewing and actually running the tests — appreciated.

Agreed on the overlap: #1556 covers this same `start_detached_agent` fd leak (its try/finally around `Popen` is equivalent to what's here), so I'm happy to close this in favor of #1556 if it lands first. I'll leave it open for now since it's focused, approved, and green, but no strong preference — whatever keeps the queue clean works for me.

One thing worth separating out: #1556 also claims #1544 (the `runtime_status` Windows crash), but for that one it keeps `os.kill(pid, 0)` and only widens the `except` to catch `SystemError`. That stops the crash, but on Windows `os.kill` with a signal other than CTRL_C/CTRL_BREAK still goes through `TerminateProcess` (per the os.kill docs), so the status probe would keep killing the live proxy — which is the main symptom in #1544. I went with `psutil.pid_exists` in #1609 specifically to avoid signalling the process at all. Flagging it so #1544 doesn't get closed by a fix that leaves the proxy-kill in place.

JerrettDavis (approved):
Looks good. Closing the parent's inherited log handle in a finally block handles both successful starts and Popen failures, and the tests assert both cases directly.

PR #1581, opened 2026-06-30
fix(update): prevent _core.pyd corruption on Windows when proxy is running

### LINKED ISSUE #1580: bug(windows): headroom update corrupts installation when proxy is running
## Summary

On Windows, running `headroom update` while `headroom proxy` is active can corrupt the installed package, leaving it in a partially-upgraded state with no automatic rollback.

## Root Cause

The Rust extension `_core.pyd` is a Windows DLL. When `headroom proxy` is running, the DLL is loaded into the process's memory — and Windows keeps the file handle open for the entire lifetime of the process. This is standard OS behaviour for loaded native libraries.

When `pip install -U headroom-ai` runs concurrently, it attempts to overwrite `_core.pyd` in-place:

```
ERROR: Could not install packages due to an OSError:
[WinError 5] Access is denied: '...\_core.pyd'
```

pip fails **mid-install**: some files from the new version are already on disk, but `_core.pyd` is still the old version. The package is left in an inconsistent state. pip provides no automatic rollback.

## Impact

- **Platform**: Windows only (Linux/macOS do not lock loaded `.so` files the same way)
- **Trigger**: `headroom update` executed while `headroom proxy` is running
- **Effect**: `ImportError` or `AttributeError` on next startup; manual reinstall required

## Why npm is NOT affected

The `headroom-ai` npm package is pure TypeScript/JavaScript — no `.node` native addon. Node.js reads `.js`/`.cjs` files once into V8 memory and closes the file handle immediately. Windows does not lock JS files. `npm install` can replace them freely at any time.

## Fix

See PR: _linked below_

The fix wraps the pip invocation in a `safe_update()` function that:

1. **If `_core.pyd` is NOT locked** (proxy not running): creates a proactive backup via `shutil.copy2` before pip runs
2. **If `_core.pyd` IS locked** (proxy running): warns the user, proceeds (pip will fail safely without touching the file)
3. **If pip fails**: atomically restores the backup via `os.replace` (single syscall, no intermediate state)
4. **If pip succeeds but integrity check fails**: atomically restores the backup, returns error
5. **If everything succeeds**: deletes the backup

The backup/restore logic was also corrected from an inverted condition in a prior draft:
- **Wrong**: backup when locked (pip will fail before touching the file — no corruption risk)  
- **Correct**: backup when NOT locked (pip WILL replace the file — backup is the safety net)

`os.replace` is used for restore instead of `shutil.copy2` because it is atomic on both POSIX (`rename(2)`) and Windows (`ReplaceFileW`): the original file is either the old version or the new version, never absent or partially written.

## Steps to Reproduce

1. Start `headroom proxy`
2. In another terminal: `headroom update`
3. Observe `[WinError 5]` and broken installation after

## Expected Behaviour

`headroom update` should detect that `_core.pyd` is locked, back it up proactively when safe, and restore atomically if anything goes wrong.

JoaoMarcos44:
## Description

On Windows, running `headroom update` while `headroom proxy` is active can corrupt the installed package by leaving the native `_core.pyd` extension in a partially upgraded state. This PR adds a safer update path around the pip invocation.

Closes #1580.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `safe_update()` handling for Windows native-extension update safety.
- Detect whether `_core.pyd` is locked before pip runs.
- Create a proactive backup when the file is not locked, then restore atomically if import integrity fails.
- Warn when the proxy is running and `_core.pyd` is locked, allowing pip to fail safely without replacing the loaded file.
- Use atomic replacement for restore paths.

## Testing

- [x] Unit tests pass

### Test Output

```text
Focused update-path tests and reviewer approval were completed on this PR before the governance body cleanup. The current body update is documentation-only metadata for PR governance.
```

## Real Behavior Proof

- Environment: Windows-focused Headroom development/review context.
- Exact command / steps: Reviewed the safe update flow for locked and unlocked `_core.pyd` cases, including backup, pip invocation, import validation, and restore behavior.
- Observed result: The update path avoids replacing a loaded native extension and provides an atomic restore path when an unlocked update fails validation.
- Not tested: End-to-end package publication/install from PyPI as part of this body cleanup.

## Review Readiness

## Additional Notes

This body was normalized by a maintainer after approval so the governance parser reflects the already-reviewed PR state.


JerrettDavis (approved):
The Windows update rollback logic looks sound: backup is created only when `_core.pyd` is replaceable, restore uses atomic `os.replace`, and the locked-file path fails safely without pretending the install can proceed.

Verified locally with:

`python -m pytest tests/test_cli_update.py -q`

Result: 21 passed.

headroom/cli/update.py
```diff
@@ -163,6 +165,149 @@ def _managed_env_guidance() -> str:
     )
 
 
+def _find_core_pyd() -> Path | None:
+    """Localiza o arquivo _core.pyd no site-packages do headroom."""
```
JerrettDavis: Can you please ensure all strings are in English? We do not yet have localization setup and scoped.

JerrettDavis (changes requested):

headroom/cli/update.py
JoaoMarcos44: Of course, I apologize for the inconvenience. I didn't even realize I had written the comment in my native language.

JerrettDavis (approved):

JerrettDavis (changes requested):
The implementation still looks good at the current head: the Windows `_core.pyd` backup/restore path remains scoped to `headroom update`, and the focused update/LiteLLM tests pass locally.

Verified locally:
- `python -m pytest tests/test_cli_update.py tests/test_litellm_callback.py -q` -> 28 passed
- `ruff check headroom/cli/update.py tests/test_cli_update.py tests/test_litellm_callback.py` -> passed
- `ruff format --check headroom/cli/update.py tests/test_cli_update.py tests/test_litellm_callback.py` -> passed

Keeping this in changes-requested because the current CI run has `commitlint` failing on this head. The workflow log is not available yet while the run is still in progress, but the failing required check blocks readiness. Please fix the commit metadata/rebase history so commitlint passes, then this should be ready from the code side.

PR #1584, opened 2026-06-30
fix: add transform marker when compression ratio exceeds threshold

### LINKED ISSUE #1577: compress() returns router:noop on single-message input — zero compression regardless of config
Tested v0.27.0 library mode. Single user message (1k and 19k tokens, markdown and plain text). All config combinations tried: protect_analysis_context=False, kompress_model=disabled, default. Always router:noop, 0% compression. PyTorch + transformers installed. Is single-message compression supported?

JingliangGao:
## Description
Fixes #1577: `compress()` returns `router:noop` on single-message input with 0% compression regardless of config.

When using library mode with a single user message, compression always returns `router:noop` even when compression is enabled but fails to meet the threshold. The root cause is in `headroom/transforms/content_router.py`: when compression fails to meet the `min_ratio` threshold, the code marks the content in the skip set but does **not** add any transform marker to `transforms_applied`, resulting in an empty marker list and a misleading `router:noop` result.

This fix adds a `router:skipped:ratio_too_high:{slot_idx}` marker when compression fails to meet the threshold, ensuring consistent tracking.

Closes #1577

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made
- `headroom/transforms/content_router.py`: Added `transforms_applied.append(f"router:skipped:ratio_too_high:{slot_idx}")` in the else branch so that skipped compression is properly tracked
- `tests/test_single_message_compression.py`: Added regression test `test_single_message_compression_with_transform_markers()` that reproduces the original bug and verifies the fix

## Testing

### Test Output
```text
$ pytest tests/test_single_message_compression.py -v
tests/test_single_message_compression.py::test_single_message_compression_with_transform_markers PASSED
3 passed
```

## Real Behavior Proof
- Environment: Python 3.11, Linux
- Exact command / steps: `pytest tests/test_single_message_compression.py -v`
- Observed result: Before fix: `transforms_applied = []`, returns `router:noop`. After fix: `transforms_applied = ["router:skipped:ratio_too_high:0"]`
- Not tested: Multi-message scenarios (covered by existing test suite)

## Runtime Rollout Safety
- Rollout-managed feature(s): transform observability markers for ratio-rejected compression attempts.
- Minimum rollout channel: normal CI-qualified release.
- Stable/default behavior changed: message content and compression decisions are unchanged; only `transforms_applied` now records the existing ratio-too-high outcome instead of falling through to `router:noop`.
- Kill switch / disable path: none required; revert the marker append to restore the previous observability behavior.
- Qualification impact: focused router regression tests and exact-head repository CI must pass.
- Rollback path: revert the one-line marker addition; there is no persisted state or migration.
## Review Readiness

## Checklist

## Additional Notes
The fix is minimal (1 line added) and ensures consistent tracking of compression attempts even when they fail due to ratio thresholds. This provides better visibility into the compression pipeline for debugging.

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -5540,6 +5540,7 @@ class ContentRouter(Transform):
                     # Didn't compress — add to skip set
                     self._cache.mark_skip(content_key)
                     result_slots[slot_idx] = message
+                    transforms_applied.append(f"router:skipped:ratio_too_high:{slot_idx}")
                     route_counts["ratio_too_high"] += 1
                     # Caveat (1): only freeze a "skip" verdict when the ML model
                     # is actually ready. A passthrough caused purely by a still-
--- a/tests/test_single_message_compression.py
+++ b/tests/test_single_message_compression.py
@@ -1,0 +1,138 @@
+"""Tests for issue #1577: single message compression with ratio threshold.
+
+When compression fails to meet the min_ratio threshold, the router should
+add a `router:skipped:ratio_too_high:<slot_idx>` marker to transforms_applied
+instead of returning an empty list (which results in router:noop).
+
+This test deterministically drives the ratio-too-high branch by mocking
+ContentRouter.compress() to return a result whose compression_ratio is at or
+above the acceptance gate (default min_ratio is 1.0, i.e. any genuine shrink
+is accepted), so the slot is skipped as ratio-too-high.
+"""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from headroom.tokenizer import Tokenizer
+from headroom.transforms.content_router import ContentRouter, ContentRouterConfig
+
+
+@pytest.fixture
+def router():
+    """Create a ContentRouter with minimal configuration."""
+    config = ContentRouterConfig(
+        skip_user_messages=False,
+        protect_recent_code=0,
+        protect_analysis_context=False,
+    )
+    return ContentRouter(config=config)
+
+
+class MockTokenCounter:
+    """Mock token counter that returns fixed values."""
+
+    def count_text(self, text: str) -> int:
+        return 100
+
+    def count_message(self, message: dict) -> int:
+        return 100
+
+    def count_messages(self, messages: list) -> int:
+        return 100
+
+
+@pytest.fixture
+def tokenizer():
+    """Create a Tokenizer instance with a mock counter."""
+    mock_counter = MockTokenCounter()
+    return Tokenizer(token_counter=mock_counter)
+
+
+def test_single_message_ratio_too_high_adds_marker(router, tokenizer):
+    """
+    Test that single message compression adds router:skipped:ratio_too_high
+    marker when compression fails to meet the min_ratio threshold.
+    """
+    # Create a message with enough tokens to trigger compression
+    large_text = "A" * 1000
+    messages = [{"role": "user", "content": large_text}]
+
+    # Mock compress() to return a result at/above the acceptance gate
+    mock_result = MagicMock()
+    mock_result.compression_ratio = 1.05  # No shrink — rejected by the gate
+    mock_result.strategy_used.value = "kompress"
+
+    with patch.object(router, "compress", return_value=mock_result):
+        result = router.apply(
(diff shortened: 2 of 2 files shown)


JerrettDavis (changes requested):
The one-line marker change is reasonable, but the added test file needs cleanup before merge. It currently uses public `compress(...)` with `kompress_model="disabled"` to reach the branch indirectly, includes `print()` output and a manual `if __name__ == "__main__"` runner, and introduces non-ASCII checkmark output. That is not consistent with the repo's pytest style and makes the regression less deterministic than it needs to be.

Please replace it with focused pytest coverage that drives the ratio-too-high branch deterministically, asserts the exact `router:skipped:ratio_too_high:<slot>` marker, and removes prints/manual runner text. If the public `compress(...)` path is intentionally covered too, keep it as a separate assertion without relying on disabled external compression behavior.

JingliangGao:
Hey @JerrettDavis, thanks for the review! Totally fair points on the test file — it was definitely a bit messy.

Here's what I changed:

- Rewrote the whole test file to use focused pytest coverage that mocks `ContentRouter.compress()` directly, so the ratio-too-high branch is hit deterministically (no more relying on `kompress_model="disabled"` as a workaround)
- Each test asserts the exact `router:skipped:ratio_too_high:<slot>` marker string
- Stripped out all the `print()` statements, the `if __name__ == "__main__"` runner, and the non-ASCII checkmark output
- Renamed the file to `test_single_message_compression.py` since the issue-number prefix felt unnecessary

One thing I was thinking about — I originally went through the public `compress(...)` API because I wanted to test the full integration path, but I agree mocking at the `ContentRouter` level is way cleaner and more deterministic for this specific branch. Happy to keep it this way.

All 3 tests pass locally. Let me know if anything else looks off!

JerrettDavis (approved):
Reviewed the latest ratio-too-high marker update. The previous test-style blocker is resolved: the tests now drive the branch deterministically, assert the exact `router:skipped:ratio_too_high:<slot>` marker, preserve the original message assertion, and no longer include prints/manual runner output.

I pushed one formatter-only cleanup commit to the PR branch: `164d4e37 style: format single-message compression test`.

Local verification passed after that cleanup:

`python -m pytest tests/test_single_message_compression.py -q` -> 3 passed
`ruff check headroom/transforms/content_router.py tests/test_single_message_compression.py` -> passed
`ruff format --check headroom/transforms/content_router.py tests/test_single_message_compression.py` -> passed

I do not see a remaining code-review blocker.

JingliangGao:
> resolved

Ok, I got it. Thanks.

JerrettDavis (changes requested):
CI is still red on the current approved head because commitlint is validating one of the branch commits, not just the PR title:

`Rename test file to remove issue number prefix`

Commitlint reports `subject-empty` and `type-empty` for that commit, so the branch needs the commit message rewritten to a conventional-commit form before this can merge. The code/test review from the previous approval still stands, but the required commit gate is failing.

Please amend/rebase that commit to something like `test: rename single-message compression coverage` and rerun CI.

JerrettDavis (changes requested):
The implementation and focused test are small, but the current head is not merge-ready because CI is red in commitlint. The failing commit is Rename test file to remove issue number prefix, which has no conventional commit type, so commitlint reports type-empty and subject-empty. Please rewrite/squash that commit into a conventional message and rerun CI.

JerrettDavis (approved):
Re-reviewed after the lease-protected history repair. The previously non-conventional rename commit is now `test: rename single-message compression coverage`; the rewritten branch has an identical final tree, and the replacement CI suite is green. The earlier implementation/test approval therefore stands.

JerrettDavis (approved):
Re-reviewed refreshed exact head 8c0ce243. The branch update does not change the already-approved implementation: ratio-rejected slots retain their original message and now emit the exact per-slot router:skipped:ratio_too_high marker instead of misleading router:noop. I corrected the PR body's current test path/count and added the required rollout-safety section. Local exact-head validation: 3 focused tests passed; Ruff and format checks passed on both changed files. No remaining blocker found.

PR #1560, opened 2026-06-29
fix(install): use Windows-safe PID liveness probe in runtime_status (#1544)

Parideboy:
## Description

`headroom install status` crashed with `OSError: [WinError 87] The parameter is incorrect` on Windows and, worse, tore down the live proxy it was only meant to inspect. `runtime_status()` probed liveness with a bare `os.kill(pid, 0)` guarded only by `except OSError`. Against a detached Windows agent (`DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP`), that call raises WinError 87, which CPython surfaces as a `SystemError` — not an `OSError` — so it escaped the handler, crashed status, and left the deployment dead (PID file removed, port 8787 freed). This mirrors the `os.kill`/`SystemError` fix PR #1315 applied to `cli/wrap.py`.

Closes #1544

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Added a shared Windows-safe `headroom._subprocess.pid_alive()` helper: rejects non-positive PIDs, prefers `psutil.pid_exists()`, and treats `SystemError` (WinError 87) as "not alive".
- `install/runtime.py` `runtime_status()` now delegates to `pid_alive()` instead of an unguarded `os.kill(pid, 0)`.
- `install/runtime.py` `stop_runtime()` now also catches `SystemError` to avoid the same crash class on shutdown.
- `cli/wrap.py` `_pid_alive()` now delegates to the shared helper, so the marker-cleanup path and the install/runtime status path share one liveness probe (the shared helper the issue asked for).
- Added regression tests for the helper and `runtime_status`.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [ ] Manual testing performed

### Test Output

```text
$ ruff check .
All checks passed!

$ ruff format --check headroom/_subprocess.py headroom/install/runtime.py headroom/cli/wrap.py tests/test_install/test_runtime.py tests/test_pid_alive.py
5 files already formatted

$ mypy headroom --ignore-missing-imports
(exit 0)

[...]

PR #1609, opened 2026-06-30
fix(install): windows-safe liveness probe in runtime_status

### LINKED ISSUE #1544: [BUG] Windows: headroom install status terminates live proxy and crashes with WinError 87

abhay-codes07:
## Description

On Windows, `headroom install status` crashes **and terminates the live proxy**.
`runtime_status()` probes the deployment PID with `os.kill(pid, 0)` and only
catches `OSError`:

```python
try:
    os.kill(pid, 0)
except OSError:
    return "stopped"
return "running"
```

Two Windows-specific problems:

1. `os.kill(pid, 0)` is **not** a no-op on Windows. For any signal other than
   `CTRL_C_EVENT`/`CTRL_BREAK_EVENT`, CPython routes `os.kill` to
   `TerminateProcess` — so a *status check* actually kills the running proxy.
2. Against a detached process-group PID (the agent is launched with
   `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP`), the call fails with
   `WinError 87` surfaced as a **`SystemError`**, which is not an `OSError`
   subclass, so it escapes the `except OSError` and crashes the command.

Observed by the reporter: status prints through `Port: 8787`, then raises
`SystemError`/`WinError 87`; afterwards `/readyz` is unreachable, nothing
listens on 8787, and `runner.pid` is gone.

`cli/wrap.py` already fixed this exact pattern in #1315 with a Windows-safe
`_pid_alive()` helper — `install/runtime.py` just never got the same treatment.

Closes #1544

## Fix

Replace the probe with a `_pid_alive()` helper (mirroring `cli/wrap._pid_alive`):

```python
def _pid_alive(pid: int) -> bool:
    if pid <= 0:
        return False
    try:
        import psutil
        return bool(psutil.pid_exists(pid))   # existence check, no signal
    except Exception:
        pass
    try:
        os.kill(pid, 0)
    except PermissionError:
        return True
    except (ProcessLookupError, OSError, SystemError):
        return False
    return True
```

```python
# runtime_status()
return "running" if _pid_alive(pid) else "stopped"
```

`psutil.pid_exists` checks existence without signalling (never terminates the
proxy), and the `os.kill` fallback now catches `SystemError` so a stale/detached
PID can't crash the command.

## Type of Change

## Changes Made

- `headroom/install/runtime.py`: add a Windows-safe `_pid_alive()` helper and use it in `runtime_status()` instead of a bare `os.kill(pid, 0)` / `except OSError`.
- `tests/test_install/test_runtime.py`: regression tests (see below); the existing container/pid-state test now uses a real live PID instead of mocking `os.kill`.

## Testing

- [ ] Full `pytest` run — see the note under Real Behavior Proof (local OOM constraint); the tests will run in CI.

New tests in `tests/test_install/test_runtime.py`:

- `test_pid_alive_rejects_nonpositive_pids`
- `test_pid_alive_true_for_current_process`
- `test_pid_alive_swallows_systemerror_from_os_kill` — forces the `os.kill` fallback and raises `SystemError`; asserts it's treated as not-alive, not propagated.
- `test_runtime_status_does_not_kill_the_live_proxy` — writes a live PID, monkeypatches `os.kill` to fail the test if called, asserts `runtime_status` returns `"running"` without signalling.
- `test_runtime_status_stopped_when_pid_not_alive`

```text
$ uv run ruff check headroom/install/runtime.py tests/test_install/test_runtime.py
All checks passed!
$ uv run ruff format --check headroom/install/runtime.py tests/test_install/test_runtime.py
2 files already formatted
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.11, headroom built from this branch (`uv sync --extra dev`). Note: a full `uv run pytest` of the install suite exhausts memory on this box (the package imports the torch/transformers stack) and the OS kills the run mid-execution, so I verified the algorithm with a dependency-free script instead and left the full pytest to CI.
- Exact command / steps: extracted the exact `_pid_alive` algorithm into a standalone script that imports only `os`/`sys` (no `headroom`, so it can't OOM) and asserted: non-positive PIDs are dead, the current process PID is alive, and a `SystemError` raised from `os.kill` in the fallback path is swallowed (returns not-alive). Ran it with the system Python.
- Observed result: all three assertions passed — non-positive PIDs report dead, the current process PID reports alive, and a `SystemError` from `os.kill` is swallowed (returns not-alive) rather than crashing:

```text
OK: nonpositive PIDs -> False
OK: current PID -> True
OK: SystemError -> False (no crash)
ALGORITHM VERIFIED
```

  I also confirmed the bug is present in `runtime_status` on `main` by inspection (`os.kill(pid, 0)` + bare `except OSError`), and that the fix mirrors the already-merged `cli/wrap._pid_alive` (#1315), which fixed the identical `SystemError`/`WinError 87` pattern on the wrap path. `ruff check` and `ruff format --check` pass on both changed files.
- Not tested: a full end-to-end `install apply` + `install status` against a real persistent Windows deployment (couldn't run the heavy stack to completion here), and the full local `pytest` run (OOM, per above — deferred to CI). The `runtime_status` change is a one-line call into the verified helper.

## Review Readiness

## Checklist

- [ ] New and existing unit tests pass locally with my changes — ran lint + a standalone algorithm check locally; full pytest deferred to CI (local OOM, disclosed above)

## Additional Notes

- No new dependencies (`psutil` is already an optional dep used by `cli/wrap._pid_alive`). The helper is duplicated rather than imported from `cli/wrap` to avoid pulling the heavy CLI module into `install/runtime`; happy to extract a shared `headroom`-level helper instead if you'd prefer.
- @chopratejas flagging you — this applies the same #1315 `_pid_alive` treatment to the remaining `os.kill(pid, 0)` call in `install/runtime.py`.


JerrettDavis (approved):
Looks good. This applies the same Windows-safe PID existence pattern already used in wrap cleanup to install runtime status, and the tests cover non-positive PIDs, current-process liveness, SystemError from the fallback path, and avoiding os.kill from runtime_status.

abhay-codes07:
Closing as superseded — #1560 ("fix(install): use Windows-safe PID liveness probe in runtime_status") merged and closed #1544. It applies the same fix (a non-signalling liveness probe instead of os.kill(pid, 0)), via a shared `pid_alive` helper in `headroom._subprocess`, which is a cleaner home for it than the local helper here. Glad the bug's fixed. Thanks @JerrettDavis for the review.

PR #1600, opened 2026-06-30
[codex] Make Claude project settings injection opt in

flin2:
## Description

Make `headroom wrap claude` avoid writing project-local Claude settings by default.

The previous unconditional `.claude/settings.local.json` injection solved daemon worker inheritance, but it also persisted localhost proxy routing into synced project directories where other machines may not have Headroom installed or running.

Closes #1599

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Default `headroom wrap claude` to process environment routing only, without writing `.claude/settings.local.json`.
- Added `--project-settings` and `--project-settings-injection` for users who explicitly want project-local Claude settings persistence.
- Added `HEADROOM_CLAUDE_PROJECT_SETTINGS=1` as a persistent opt-in for aliases and wrapper scripts.
- Guarded cleanup so the wrapper restores project-local settings only when the current run actually wrote them.
- Preserved current `main` behavior for stale-marker cleanup and SessionStart self-healing when project-settings injection is enabled.
- Updated focused wrapper tests for the default no-write path, the explicit opt-in path, the env-var gate, Vertex/Foundry routing, and early proxy-start failure cleanup.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check` and `ruff format --check`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [ ] Manual testing performed

### Test Output

```text
$ .venv/bin/pytest tests/test_cli/test_wrap_claude_base_url.py tests/test_cli/test_wrap_claude_vertex_proxy_env.py tests/test_cli/test_wrap_claude_finally_unbound.py -q
49 passed in 0.25s

$ .venv/bin/ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_claude_base_url.py tests/test_cli/test_wrap_claude_vertex_proxy_env.py tests/test_cli/test_wrap_claude_finally_unbound.py
All checks passed!

$ .venv/bin/ruff format --check headroom/cli/wrap.py tests/test_cli/test_wrap_claude_base_url.py tests/test_cli/test_wrap_claude_vertex_proxy_env.py tests/test_cli/test_wrap_claude_finally_unbound.py
4 files already formatted
```

[...]

PR #1610, opened 2026-06-30
fix(claude): surface Remote Control proxy incompatibility

### LINKED ISSUE #1601: [BUG] Remote Control option hidden under `headroom wrap claude` since Claude Code ~2.1.19x — proxied sessions fail new first-party elevated-auth gate
## Summary

The **Remote Control** entry point disappears from Claude Code when the session is launched through `headroom wrap claude` (i.e. with `ANTHROPIC_BASE_URL` pointed at the Headroom proxy).
Running the same Claude Code, on the same machine, same account, same already-enrolled device *without* Headroom shows the option normally.
This started after a recent Claude Code update (I am now on `2.1.196`; it was working through the proxy ~2 days ago on the prior build), so it looks like a newly-landed client-side gate that custom-endpoint sessions can no longer satisfy.

This is a feature-availability regression for proxied sessions, not a runtime error — there is no failing request to point at, the menu item is simply never rendered.

## Environment

- Headroom `0.28.0`
- Claude Code `2.1.196` (regression also present on `2.1.193`/`2.1.195`; last-known-good was the build from ~2 days earlier)
- macOS (Apple Silicon)
- Launched via `headroom wrap claude --memory`; wrapped process runs with `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`
- Authenticated with a Claude subscription (OAuth), device previously enrolled / trusted

## Steps to reproduce

1. In a normal shell (no Headroom), start Claude Code → the Remote Control option is present and usable.
2. Start Claude Code via `headroom wrap claude` (so `ANTHROPIC_BASE_URL` = the local proxy).
3. The Remote Control option is absent. No prompt, no error — the entry point is gone.

The only variable that changes between (1) and (3) is `ANTHROPIC_BASE_URL`.

## What the evidence shows

**The option is gated client-side; nothing reaches the proxy.**
Across all rotations of `~/.headroom/logs/proxy.log*`, the only inbound paths are `/v1/messages`, `/v1/models`, `/v1/messages/count_tokens`, `/health`, `/stats`.
There is no remote-control / session-registration / device-attestation traffic at all.
So Claude Code decides not to offer the feature before making any request — the proxy never sees it, and a proxy-side pass-through alone would not be enough unless it also makes the client consider the session "first-party".

**Device trust is already satisfied.**
Because the option works when running Claude Code directly (same device), the device is enrolled/attested correctly.
So this is *not* a "device not enrolled" problem — the persistent device-trust state is fine.
The thing that differs is purely that the proxied session is on a custom, non-first-party endpoint.

**There is a newly-enforced elevated-auth gate in the client.**
The `2.1.196` binary contains these gate identifiers:

- `isRemoteControlHardDisabled`
- `isPersistentRemoteSessionEnabled`
- `isRunningInRemoteEnvironment`
- `isRemoteControlInternalEventsEnabled`

and a GrowthBook flag `tengu_sessions_elevated_auth_enforcement` (keyed to `require_trusted_devices`), which is cached as `true` for my account (seen in `~/.claude/.claude.json` → `cachedGrowthBookFeatures`).
The same code path carries strings such as:

- `DEVICE_ATTESTATION_STATUS_…`
- `bridge_trusted_device_enroll`
- `tengu_sessions_elevated_auth_disable_proactive_enrollment`
- "Your organization requires Trusted Devices for Remote Control, but this device is not enrolled. Please run `/login` in Claude Code to enroll this device."

That config also records that Remote Control was previously available to me (`hasUsedRemoteControl: true`, `remoteControlUpsellSeenCount: 3`), consistent with a regression rather than a never-supported feature.

## Hypothesis (for maintainers to confirm)

With `tengu_sessions_elevated_auth_enforcement` enabled, a Remote Control / persistent-remote session appears to require a first-party "elevated auth" / device-attestation handshake that Claude Code will only establish (or trust) against the real `api.anthropic.com`.
When the session is pointed at a custom `ANTHROPIC_BASE_URL` (the Headroom proxy), that per-session first-party check is not satisfied, so the client hard-disables and hides the entry point.
Device enrollment is necessary but not sufficient — the live first-party check is the part that fails behind the proxy.

## Suggested directions

- Determine whether the elevated-auth / device-attestation / remote-session "bridge" endpoints can be passed through transparently so a `headroom wrap claude` session is still considered first-party-eligible for Remote Control.
- If that is not feasible (the check may be bound to the literal `api.anthropic.com` host or to TLS/identity that a local proxy cannot reproduce), document the incompatibility so users know Remote Control requires running Claude Code without the proxy.
- It would also help to know whether a specific allow-listed host/path or an opt-out env var would let the client treat the proxied session as first-party.

## Diagnostic others can run

A `--debug` Claude Code launched through the proxy should log the exact gate decision:

```
ANTHROPIC_BASE_URL=http://127.0.0.1:8787 claude --debug 2>&1 \
  | grep -iE 'remote.?control|elevated|hard ?disabled|first.?party|eligib'
```

Happy to capture that output or test patches.

rodboev:
## Description

Claude Code hides Remote Control when it sees a custom `ANTHROPIC_BASE_URL`, so `headroom wrap claude` can make the menu disappear even though normal API requests still route through Headroom. The reported proxy logs show no Remote Control registration, session bootstrap, or device-attestation request at all, which means the decision happens inside Claude before Headroom can forward anything.

This change makes that client-side incompatibility explicit in Headroom's Claude launch flow, `headroom doctor`, and troubleshooting docs. API proxying and the existing `ENABLE_TOOL_SEARCH` compatibility shim stay unchanged; users who need Remote Control get a direct instruction to launch Claude without the Headroom proxy for that session.

Closes #1601

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- Add a Claude-specific helper and warning text for the Remote Control custom-base incompatibility.
- Surface that warning from `headroom wrap claude` when Claude is launched through `ANTHROPIC_BASE_URL`.
- Add a separate `headroom doctor` warning for Claude Remote Control availability, while keeping Claude API-routing status independent.
- Document the limitation and workaround next to the existing Claude custom-endpoint troubleshooting guidance.
- Add focused regression tests for gated and non-gated Claude routing states, plus preservation coverage for `ENABLE_TOOL_SEARCH`.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_issue_1601_remote_control_gate.py tests/test_cli_doctor.py tests/test_cli/test_wrap_claude_vertex_proxy_env.py -q`)
- [x] Unit tests pass (`uv run pytest tests/test_issue_746_tool_search.py tests/test_cli/test_init_enable_tool_search.py -q`)
- [x] Linting passes (`uv run ruff check headroom/cli/wrap.py tests/test_cli_doctor.py tests/test_cli/test_wrap_claude_vertex_proxy_env.py`)
- [x] Formatting passes (`uv run ruff format --check headroom/cli/wrap.py tests/test_cli_doctor.py tests/test_cli/test_wrap_claude_vertex_proxy_env.py`)
- [x] New tests added for the bugfix

### Test Output

```text
rtk uv run pytest tests/test_issue_1601_remote_control_gate.py tests/test_cli_doctor.py tests/test_cli/test_wrap_claude_vertex_proxy_env.py -q
collected 62 items
62 passed, 1 warning

rtk uv run pytest tests/test_issue_746_tool_search.py tests/test_cli/test_init_enable_tool_search.py -q
collected 33 items
33 passed, 1 warning

rtk uv run ruff check headroom/cli/wrap.py tests/test_cli_doctor.py tests/test_cli/test_wrap_claude_vertex_proxy_env.py
All checks passed!

rtk uv run ruff format --check headroom/cli/wrap.py tests/test_cli_doctor.py tests/test_cli/test_wrap_claude_vertex_proxy_env.py
3 files already formatted
```

## Real Behavior Proof

- Environment: Windows, Python via `uv`, focused Claude CLI and doctor tests.
- Exact command / steps: with Claude settings or shell environment containing `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`, run the focused helper and doctor tests, then run the existing `ENABLE_TOOL_SEARCH` preservation tests.
- Observed result: Headroom surfaces a Claude Remote Control warning for custom `ANTHROPIC_BASE_URL`, while Claude API routing and `ENABLE_TOOL_SEARCH` behavior stay intact.
- Not tested: live Claude Remote Control UI automation. The issue evidence says Claude hides the menu before any request reaches Headroom, so this PR proves Headroom's launch, diagnostics, and docs behavior.

## Review Readiness

## Checklist

## Additional Notes

`CHANGELOG.md` stays untouched because this repo's release pipeline generates changelog entries from conventional commits.

This is a visibility fix, not a proxy transport restore. The issue evidence shows Claude never sends a Remote Control request while the custom-base gate is active, so the surviving slice is launch-time warning, doctor warning, and documentation.

PR `#1600` is adjacent and non-blocking because `#1601` reproduces from process-env `ANTHROPIC_BASE_URL` alone.

This intentionally changes `headroom doctor` for fully routed Claude sessions from an all-pass result to one warnings-only result, because the proxied Claude setup is operational for API traffic but still incompatible with Remote Control.

### DIFF
--- a/docs/content/docs/troubleshooting.mdx
+++ b/docs/content/docs/troubleshooting.mdx
@@ -183,6 +183,16 @@ When deferral is off, the proxy log also prints a one-time hint naming the fix.
 
 See [issue #746](https://github.com/chopratejas/headroom/issues/746) for the full analysis.
 
+## Remote Control unavailable through custom ANTHROPIC_BASE_URL
+
+**Symptom**: When Claude Code runs with `ANTHROPIC_BASE_URL` set to a custom host (for example, Headroom), the Remote Control menu is absent.
+
+**Cause**: This is a Claude-side gate. Headroom only receives normal API traffic and can still compress it, but Claude evaluates Remote Control availability before proxy traffic reaches the server.
+
+**Fix**: Use Headroom for normal proxied API sessions, and launch Claude directly (without `ANTHROPIC_BASE_URL`) when you need Claude Remote Control.
+
+`ENABLE_TOOL_SEARCH` is unaffected and can stay enabled for context-window savings while routing through Headroom.
+
 ## Compression Too Aggressive
 
 **Symptom**: LLM responses are missing information that was in tool outputs.
--- a/headroom/cli/doctor.py
+++ b/headroom/cli/doctor.py
@@ -26,6 +26,11 @@ from headroom.install.health import probe_json
 from headroom.install.paths import claude_settings_path, codex_config_path
 from headroom.install.state import list_manifests
 from headroom.paths import savings_path
+from headroom.providers.claude import (
+    REMOTE_CONTROL_BASE_URL_ENV,
+    is_custom_anthropic_base_url,
+    remote_control_gate_message,
+)
 
 from .main import get_version, main
 
@@ -146,6 +151,41 @@ def check_claude_routing(settings_path: Path, port: int) -> CheckResult:
             hint="wrap it: headroom wrap claude",
         )
     return _classify_routing_url(name, base_url, port, source=str(settings_path))
+
+
+def check_claude_remote_control_gate(
+    settings_path: Path, environ: Mapping[str, str]
+) -> CheckResult | None:
+    """Warn once when Claude custom-base routing hides Remote Control."""
+    name = "claude remote control"
+    settings_base_url = ""
+    if settings_path.exists():
+        try:
+            payload = json.loads(settings_path.read_text(encoding="utf-8"))
+            env_block = payload.get("env")
+            if isinstance(env_block, dict):
+                settings_base_url = str(env_block.get("ANTHROPIC_BASE_URL", "") or "")
+        except (OSError, ValueError):
+            settings_base_url = ""
+    if is_custom_anthropic_base_url(settings_base_url):
+        remote_message = remote_control_gate_message(f"{REMOTE_CONTROL_BASE_URL_ENV} from settings")
+        return CheckResult(
+            name=name,
+            status=WARN,
+            summary=remote_message,
+            hint=remote_message,
+        )
+
+    env_base_url = environ.get("ANTHROPIC_BASE_URL", "")
+    if is_custom_anthropic_base_url(env_base_url):
+        remote_message = remote_control_gate_message(f"{REMOTE_CONTROL_BASE_URL_ENV} in shell")
+        return CheckResult(
+            name=name,
+            status=WARN,
+            summary=remote_message,
+            hint=remote_message,
+        )
+    return None
 
 
 def check_codex_routing(config_path: Path, port: int) -> CheckResult:
@@ -384,6 +424,9 @@ def doctor(port: int, emit_json: bool) -> None:
         check_savings(stats, savings_path()),
         check_budget(stats),
     ]
+    remote_control_gate_check = check_claude_remote_control_gate(claude_settings_path(), os.environ)
+    if remote_control_gate_check is not None:
+        checks.append(remote_control_gate_check)
     deployments = check_deployments(list_manifests())
     if deployments is not None:
         checks.append(deployments)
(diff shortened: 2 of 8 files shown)


JerrettDavis (approved):
Looks good. This is an accurate compatibility warning rather than a routing behavior change, and the shared helper keeps wrap, doctor, and docs aligned. The doctor warning behavior is covered for both settings.json and shell env sources.

JerrettDavis (approved):

PR #625, opened 2026-06-05
perf: prefix stability pinning and stale read/glob compression

neogenix:
## Summary

Two performance improvements targeting context optimization efficiency, both opt-in via env vars.

### Background (from headroom perf report)

- **32/214 requests had cache_write > 2x cache_read** — prefix cache being busted frequently in early turns
- **Excluded: 1342 (40%) — Read/Glob outputs** — nearly half of all messages excluded from compression

### Improvement 1: Prefix stability pinning (HEADROOM_COMPRESSION_STABLE_AFTER_TURN)

**Problem**: In the first N turns, the proxy hasn't built a stable prefix yet. The ML-based Kompress compressor is non-deterministic — the same content can produce slightly different compressed output across calls. When a tool output compressed in turn 2 produces a different byte sequence in turn 3, the Anthropic prompt cache prefix is invalidated, replacing a 90% read discount with a 25% write penalty.

**Solution**: Add a conservative compression mode for early turns. When `session_turn_count <= compression_stable_after_turn`, skip Kompress for tool outputs and use only deterministic strategies (SmartCrusher, SearchCompressor, LogCompressor, DiffCompressor, HTMLExtractor). These are rule-based and always produce the same output for the same input.

**Config**: `HEADROOM_COMPRESSION_STABLE_AFTER_TURN=5` (default 0 = disabled). Set to 5 to protect first 5 turns.

**New code** (`headroom/transforms/content_router.py`):
- `ContentRouterConfig.compression_stable_after_turn: int` (field with factory for env-var-safe default)
- `ContentRouter._compress_conservative()` — compresses without Kompress
- `ContentRouter._timed_compress_conservative()` — for parallel executor
- `in_stability_phase` computed in `apply()` from `session_turn_count = max(1, num_messages // 2)`

### Improvement 2: Stale Read/Glob compression (HEADROOM_STALE_READ_COMPRESS_AFTER_TURNS)

**Problem**: Read, Glob, Grep, Write, Edit, Bash tool outputs are currently excluded from compression entirely (protected in `DEFAULT_EXCLUDE_TOOLS`). But content from 10+ turns ago is unlikely to change and is safe to compress.

**Solution**: Excluded-tool outputs with `messages_from_end > stale_read_compress_after_turns * 2` fall through the exclusion guard and are eligible for compression. Recent outputs (within last N turns) retain full protection.

[...]

PR #1620, opened 2026-07-01
fix(content_router): pin FREEZE_BLOCK_DECISION verdict to stop cache-write churn

### LINKED ISSUE #1619: ContentRouter: per-block compress verdict flaps with min_ratio drift, busting the prefix cache
### Summary

The per-block "compress" verdict in `ContentRouter` is not stable across turns. When a cached block is re-checked against the current `min_ratio` (which tightens as context pressure rises within a session), a block that was accepted and compressed on an earlier turn can be downgraded to skip, its original restored, and the prefix cache busted. That is a self-inflicted cache-write churn: the block re-enters the uncompressed prefix and later re-compresses, costing the tokens the compression saved and thrashing the prefix cache.

### Steps to reproduce

1. A tool_result content-block compresses and is accepted on turn N (ratio below the then-current `min_ratio`).
2. Context pressure rises, so `min_ratio` tightens on turn N+1.
3. The same block's cached ratio now fails the tighter gate, so it is moved to skip and the original is restored — the prefix cache line for that block is invalidated even though nothing about the block changed.

### Expected

Once a block is accepted as "compress", that decision should be stable for the life of the block rather than flapping with per-turn `min_ratio` drift.

### Proposal

Pin the decision: a frozen "compress" verdict re-accepts on later turns instead of re-running the per-turn `min_ratio` gate. First-sighting keeps the live `min_ratio` gate — the freeze only pins past accepts, it never loosens the first decision (that would be a silent ratio bet). It should be gated behind an env flag, default off and byte-identical when unset, and must respect the #1307 reversibility guard (never pin a lossy-unmarked block that emitted no CCR retrieval marker).

I have a PR prepared (default-off, with tests). Neighbour of #625.

rocker-zhang:
## Description

The per-block freeze decision was inert. On the cached-block re-check, a later tighter `min_ratio` (context pressure rises within a session) could downgrade an earlier "compress" verdict to skip, restore the original block, and bust the prefix cache — a self-inflicted cache-write churn that costs the very tokens compression saved.

This pins the decision instead. A frozen "compress" verdict re-accepts (`accept_threshold = 1.0`) rather than re-running the per-turn `min_ratio` gate, so a block that was accepted stays accepted. First-sighting still uses the live `min_ratio` gate (`accept_threshold = min_ratio`): the freeze only pins past accepts, it never loosens the first decision (that would be a silent ratio bet). Gated behind `HEADROOM_FREEZE_BLOCK_DECISION`, default off, byte-identical to today when unset.

Composes with the #1307 reversibility guard on the compress path: a frozen accept still defers to the lossy-unrecoverable skip.

Closes #1619

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `content_router.py`: on the cached-block hit path and the first-sighting path, compute `accept_threshold` (1.0 when a "compress" verdict is frozen for the block, else the live `min_ratio`) and gate accept on it; record the pin when the legacy re-check would have downgraded.
- Frozen verdicts are stored per content-block key and only ever hold "compress" (a "skip" never warms the result cache).
- No change when `HEADROOM_FREEZE_BLOCK_DECISION` is unset.

## Testing

- [x] Type checking passes (`mypy headroom` — no new errors)

### Test Output

```text
$ pytest tests/test_transforms_content_router.py -q
38 passed in 0.23s
# 11 freeze/pin/churn cases + 27 existing; run against a real 0.28.0 _core.
# - test_freeze_off_is_byte_identical_flapping_baseline: freeze-off == baseline byte-for-byte
# - test_freeze_on_pins_compress_verdict_across_turns: pin fires; downgrade prevented
$ ruff format --check . && ruff check .   -> clean
```

## Real Behavior Proof

- Environment: isolated git worktree on latest `main`, real `_core.abi3.so` built for this tree via maturin (not a stale symlink), scratch venv.
- Exact command: `pytest tests/test_transforms_content_router.py -q`
- Observed: freeze-off path is byte-identical to the flapping baseline; with freeze on, the verdict is pinned across turns (pin-count assertion passes) so the block is not downgraded/restored and no cache-write churn occurs.
- Not tested: end-to-end proxy A/B token-savings delta (follow-up; feature ships default-off).

N/A

## Review Readiness

Ready for review. Default-off, composes with #1307, self-contained to the block-decision path.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code in hard-to-understand areas
- [x] Documentation intentionally deferred until the default-off approach is confirmed
- [x] CHANGELOG intentionally deferred until the default-off approach is confirmed

## Additional Notes

Neighbour of #625 (prefix-stability). Docs/CHANGELOG intentionally deferred until the approach is confirmed. The end-to-end A/B is a follow-up; the churn-prevention is proven at the router unit level here.


JerrettDavis (approved):
Reviewed the freeze-decision cache pinning change. The flag remains default-off, the new state is bounded and cleared with the cache, and the tests cover flag-off byte behavior, pinned compress decisions, model-not-ready skip handling, content-block behavior, and unrecoverable lossy blocks. CI is green.

JerrettDavis (approved):
Re-reviewed current head after the main merge. The freeze-decision cache pinning remains bounded/default-off and the focused router coverage still passes.

- `python -m pytest tests/test_transforms_content_router.py -q` -> 48 passed
- `uvx ruff==0.15.17 check headroom/transforms/content_router.py tests/test_transforms_content_router.py`
- `uvx ruff==0.15.17 format --check headroom/transforms/content_router.py tests/test_transforms_content_router.py`

PR #1625, opened 2026-07-01
fix(learn): Windows compatibility for verbosity and session discovery

### LINKED ISSUE #1624: Improve Windows learn --verbosity compatibility and third-party provider routing
## Summary

When using Headroom with Claude Code through a third-party Anthropic-compatible provider on Windows, several issues appeared around routing, startup behavior, session discovery, dashboard visibility, and UX hints. This Issue combines problems found on both the Claude Code and Codex sides.

**Working chain:**

```
Claude Code / Codex → Headroom :8787 → cc-switch :15729 → third-party provider
```

## Environment

- OS: Windows 11 Pro
- Shell: PowerShell + Git Bash
- Headroom 0.28.0 (installed via pip)
- Claude Code 2.1.168 (npm global install)
- Codex CLI
- Claude Code uses a third-party Anthropic-compatible provider via cc-switch

### 1. `headroom learn --verbosity` returns empty results on Windows

**Observed:** Running `headroom learn --verbosity --all` on Windows produces results with `Sessions: 0, human turns: 0` for every project, even when session JSONL files exist at `~/.claude/projects/`.

**Root cause:** `verbosity.py` calls `path.read_text()` without specifying encoding. On Windows, Python defaults to GBK. When a transcript contains non-ASCII characters (Chinese, Japanese, etc.), a `UnicodeDecodeError` is raised and silently caught, returning empty results.

**Fix:** Add `encoding="utf-8"` to all three `read_text()` calls in `verbosity.py` (lines 147, 211, 285).

---

### 2. `headroom learn --verbosity --project` fails with "No matching project" on Windows

**Observed:** Running `headroom learn --verbosity --project "D:\work\DPJ"` produces `No matching project. Try --all or --project <path>.`

**Root cause:** Claude Code stores sessions at `~/.claude/projects/D--work-DPJ/`. The directory name format `D--work-DPJ` (where `--` encodes `:\`) is not correctly decoded. The fallback logic removes the first character (drive letter `D`) then splits on `-`, resulting in `Path("/work/DPJ")` instead of `Path("D:\work\DPJ")`.

Additionally, when a project name itself contains hyphens (e.g. `vibe-remote`), the directory name `D--work-vibe-remote` is ambiguous — it could decode to both `D:\work\vibe-remote` (correct) and `D:\work\vibe\remote` (wrong).

**Fix:**
1. Add drive-letter pattern detection for `X--rest` format before the existing fallback
2. Read the `cwd` field from session JSONL files to resolve the true project path, eliminating the ambiguity

---

### 3. `headroom learn --agent codex` fails: `claude not found in PATH`

**Observed:** Running `headroom learn --agent codex --all` produces:
```
LLM analysis failed: `claude` not found in PATH.
```
even though `claude` is installed and `which claude` finds it.

**Root cause:** On Windows, npm-installed global CLIs (e.g. `claude`) are shell script shims (`claude` with a bash shebang), not directly executable by `CreateProcess`. The actual executable is `claude.cmd`. `Popen()` throws `FileNotFoundError` because it can't run the bash script.

**Fix:** In `analyzer.py`, after `Popen()` fails with `FileNotFoundError` on Windows, fall back to trying `<name>.cmd`, `<name>.exe`, and `<name>.bat` before raising an error.

**See:** PR #4 (https://github.com/37chengshan/headroom/pull/4)

---

### 4. Dashboard output token savings hint is confusing

**Observed:** The dashboard shows:
```
Output Tokens Saved
—
Enable the output shaper (HEADROOM_OUTPUT_SHAPER=1) and run headroom learn --verbosity --apply to start measuring.
```

This looks like an error, but it is an optional feature hint. Additionally, on Windows/PowerShell the environment variable syntax shown is wrong:

```powershell
# Wrong in PowerShell:
HEADROOM_OUTPUT_SHAPER=1

# Correct:
$env:HEADROOM_OUTPUT_SHAPER = "1"
```

**Suggestion:** Make the hint clearly optional (e.g. `Optional:` prefix) and include Windows-specific environment variable syntax.

---

### 5. Claude Code 401 with third-party provider

**Observed:** Initial setup with a third-party Anthropic-compatible provider produces:
```
API Error: 401 Invalid bearer token
```
Requests were routed to `api.anthropic.com` instead of through the local cc-switch proxy.

**Suggestion:** Headroom could provide clearer documentation or a config example for third-party providers, especially when used together with tools like cc-switch.

---

### 6. Restart port binding confusion

**Observed:** After Windows reboot:
```
ERROR: [Errno 10048] error while attempting to bind on address ('127.0.0.1', 8787)
```

**Suggestion:** Show a friendlier message when the port is already occupied, e.g.:
```
Port 8787 is already in use. Headroom may already be running.
Try opening http://127.0.0.1:8787/dashboard or stop the existing process first.
```

---

### 7. Dashboard localization

The dashboard is English-only. A lightweight i18n layer with browser-language detection and localStorage persistence would improve usability for non-English users. Technical terms (Headroom, Token, Proxy, API, etc.) should remain untranslated.

---

## Steps to Reproduce

```powershell
# Issue 1 & 2: empty results
headroom learn --verbosity --all
headroom learn --verbosity --project "D:\work\DPJ"

# Issue 3: codex scan fails
headroom learn --agent codex --all
```

## Related

- PR #4 (fixes issues 1-3)
- Installed via: `pip install headroom-ai==0.28.0`

37chengshan:
## Summary

Fixes three bugs that prevented `headroom learn --verbosity` from working on Windows, and fixes `headroom learn --agent codex` failing to invoke the `claude` CLI.

Closes #1624 (issues 1-3)

### `headroom/learn/verbosity.py` — UTF-8 encoding fix

`Path.read_text()` without encoding defaults to GBK on Windows. Session JSONL files are UTF-8, so non-ASCII characters (Chinese, Japanese, etc.) cause `UnicodeDecodeError`, which is silently caught — returning empty results for all projects.

Added `encoding="utf-8"` to all three `read_text()` calls (lines 147, 211, 285).

### `headroom/learn/plugins/claude.py` — Windows drive-letter path decoding + cwd resolution

Two fixes:

1. **Drive-letter path decoding:** The `D--work-DPJ` format (where `--` encodes `:\`) was not handled by the existing fallback. Added detection for `X--rest` prefix before the existing fallback logic.

2. **cwd reverse lookup:** When a project name contains hyphens (e.g. `vibe-remote`), the directory name `D--work-vibe-remote` is ambiguous. Added logic to read the `cwd` field from session JSONL files to resolve the true project path.

### `headroom/learn/analyzer.py` — Windows .cmd shim fallback

On Windows, npm-installed global CLIs (e.g. `claude`) are bash script shims that `CreateProcess` cannot invoke directly. The actual executables are `.cmd` / `.exe` shims.

Added fallback logic: after `Popen()` fails with `FileNotFoundError` on Windows, try `<name>.cmd`, `<name>.exe`, and `<name>.bat` before raising an error.

## Testing

Ran `test_learn/test_analyzer.py` and `test_verbosity_controller.py` — **87 passed, 1 pre-existing failure** (`test_includes_project_info` expects Linux path `/tmp/test-project` but gets Windows path `\\tmp\test-project`). That failure is unrelated to this PR.

Verified the cwd reverse lookup resolves correctly:
```
D--work-vibe-remote → reads cwd: D:\work\vibe-remote ✓
D--work-DPJ → reads cwd: D:\work\DPJ ✓
```

### DIFF
--- a/headroom/learn/analyzer.py
+++ b/headroom/learn/analyzer.py
@@ -573,20 +573,38 @@ def _call_claude_cli_streaming(
     Threads (rather than ``select``) drain stdout/stderr so the watchdog works
     on Windows too, where ``select`` does not support pipe handles.
     """
-    try:
-        proc = Popen(
-            cmd,
-            stdin=subprocess.PIPE,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            text=True,
-            bufsize=1,  # line-buffered
-        )
-    except FileNotFoundError:
+    # On Windows, raw scripts (e.g. `claude` with a bash shebang from an npm
+    # global install) can't be executed directly by CreateProcess — they need
+    # the `.cmd`/`.exe` shim. Try the bare name first; on FileNotFoundError fall
+    # back to <name>.cmd (and <name>.exe) so npm-global CLIs work on Windows.
+    candidates = [cmd]
+    if os.name == "nt" and "/" not in cmd[0] and "\\" not in cmd[0]:
+        base = cmd[0]
+        for ext in (".cmd", ".exe", ".bat"):
+            candidates.append([base + ext] + cmd[1:])
+
+    proc = None
+    last_exc: Exception | None = None
+    for candidate in candidates:
+        try:
+            proc = Popen(
+                candidate,
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                text=True,
+                bufsize=1,  # line-buffered
+            )
+            break
+        except FileNotFoundError as exc:
+            last_exc = exc
+            continue
+
+    if proc is None:
         raise RuntimeError(
             f"`{cmd[0]}` not found in PATH. Install it or use a different backend "
             "with --model <litellm-model-name>."
-        ) from None
+        ) from last_exc
 
     assert proc.stdin is not None and proc.stdout is not None and proc.stderr is not None
     try:
--- a/headroom/learn/plugins/claude.py
+++ b/headroom/learn/plugins/claude.py
@@ -70,12 +70,18 @@ class ClaudeCodePlugin(LearnPlugin, ConversationScanner):
 
             project_path = _decode_project_path(entry.name)
             if project_path is None:
-                fallback_parts = entry.name[1:].split("-")
-                if len(fallback_parts[0]) == 1 and fallback_parts[0].isalpha():
-                    drive = fallback_parts[0].upper()
-                    project_path = Path(f"{drive}:\\" + "\\".join(fallback_parts[1:]))
+                # Handle "D--work-DPJ" → "D:\work\DPJ" (drive letter + -- = :\)
+                if len(entry.name) >= 3 and entry.name[1:3] == "--" and entry.name[0].isalpha():
+                    drive = entry.name[0].upper()
+                    rest = entry.name[3:]
+                    project_path = Path(f"{drive}:\\{rest.replace('-', '\\')}")
                 else:
-                    project_path = Path("/" + entry.name[1:].replace("-", "/"))
+                    fallback_parts = entry.name[1:].split("-")
+                    if len(fallback_parts[0]) == 1 and fallback_parts[0].isalpha():
+                        drive = fallback_parts[0].upper()
+                        project_path = Path(f"{drive}:\\" + "\\".join(fallback_parts[1:]))
+                    else:
+                        project_path = Path("/" + entry.name[1:].replace("-", "/"))
 
             name = _project_display_name(project_path, entry.name)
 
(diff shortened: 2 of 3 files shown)


alewis50 (approved):

JerrettDavis (changes requested):
Thanks for chasing the Windows learn path. I think this PR needs another pass before it can merge: `headroom/learn/plugins/claude.py` now contains `Path(f"{drive}:`{rest.replace('-', '`')}")`, which is a SyntaxError on supported Python versions because f-string expression parts cannot contain a backslash literal. That means importing the Claude learn plugin will fail before the new behavior can run.

There is also a newer overlapping PR, #1638, that handles the same Windows session discovery/verbosity area by preferring the `cwd` recorded in the session JSONL, which avoids the ambiguous dash-decoding problem more generally. I’d suggest either rebasing this PR onto that approach or closing this one in favor of the newer implementation so we do not carry two divergent fixes for the same path.

37chengshan:
Thanks for the detailed review. Good catch on the f-string backslash issue — you're right that this would fail at import time on supported Python versions.

I also agree that #1638 seems to address the same Windows session discovery area in a more robust way by relying on the cwd recorded in the session JSONL instead of dash-decoding the path.

I'll take another look at #1638 and either rebase this PR onto that approach or close this PR if the newer implementation fully supersedes it. Thanks again for pointing this out.

37chengshan:
Thanks for the review and the steer toward #1638. I compared the two PRs in detail, and #1638 now fully covers all three Windows learn failures this PR was targeting, so I'm closing it in favor of #1638 to avoid carrying two divergent fixes for the same path.

Mapping of the three issues to #1638:

1. **verbosity.py UTF-8 reads** — #1638 adds `encoding="utf-8"` to `read_text()` in `_parse_session`, `_ordered_events`, and `VerbosityProfile.load`, same as here.

2. **claude.py session discovery** — #1638 refactors `discover_projects` to metadata-first: it discovers JSONL files first, then calls a new `_detect_project_path(jsonl_files)` (reading the session `cwd`) before falling back to `_decode_project_path`. Same approach as this revision, avoiding the ambiguous dash-decoding for names like `vibe-remote`.

3. **analyzer.py Windows CLI shim** — #1638 adds the Windows `.cmd` / `.exe` / `.bat` fallback in `_call_claude_cli_streaming` when the bare command raises `FileNotFoundError`, same as here. It also resolves the merge conflict by preserving the `headroom._subprocess.Popen` wrapper.

#1638 additionally fixes a Windows path-separator issue I had not addressed: `_build_digest` renders `project.project_path` via `.as_posix()` so the digest uses forward slashes on Windows.

#1638 is approved and `status: ready for review`, with focused tests (`windows_falls_back_to_cmd_shim`, `prefers_session_cwd`, `utf8_session_files`) mirroring the regressions I'd added here. Closing #1625 so #1638 can land as the single fix. Thanks again.

37chengshan:
Closing in favor of #1638, which fully covers these three Windows learn fixes and is approved/ready for review. See the comment above for the detailed mapping.

PR #1634, opened 2026-07-01
fix(cli): stop advertising unwired compression tuning env vars in banner

gglucass:
## Description

The startup banner's `Performance Tuning` section reads `HEADROOM_COMPRESSION_STABLE_AFTER_TURN` and `HEADROOM_STALE_READ_COMPRESS_AFTER_TURNS` and prints them as active tuning knobs. Neither is consumed anywhere else — not in the Python compression path, and not in the packaged native code (verified by scanning the shipped extension modules; `headroom` ships no env-reading native lib and Kompress runs via ONNX). Setting either var changes the banner but has zero effect on behavior, which actively misleads operators trying to tune compression load.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Removed the two unwired env vars from the banner's `Performance Tuning` section, including the fallback hint that told users to "set" them.
- Kept the embedding-sidecar line (`HEADROOM_EMBEDDING_SERVER_SOCKET`), which is a real, consumed setting; the section now renders only when a real tuning value is active and is empty otherwise.
- Added an Unreleased → Fixed CHANGELOG entry.

## Testing

### Test Output

```text
$ pytest tests/test_cli_proxy_improvements.py -q
48 passed in 5.04s

$ ruff check headroom/cli/proxy.py && ruff format --check headroom/cli/proxy.py
All checks passed! / 1 file already formatted

$ mypy headroom/cli/proxy.py
```

## Real Behavior Proof

- Environment: macOS, Python 3.10.18, branch off upstream/main @ 0.28.0
- Exact command / steps: grepped the entire package + shipped `.so`/native modules for both env var names; only the banner referenced them.
- Observed result: no consumer exists for either var; banner was the sole reader. After the change the banner no longer claims they do anything.
- Not tested: N/A — this removes a false claim; no behavior to exercise beyond the existing CLI-invocation tests, which pass.

## Review Readiness

## Checklist

## Additional Notes

No new test added: the fix deletes dead/misleading output rather than adding logic; a banner-string assertion would be brittle. If you'd rather *implement* these knobs than remove them (i.e. actually gate Kompress on prefix-stable-after-N-turns), I'm happy to open a separate feature PR instead — but as shipped they are pure no-ops, so this stops the banner from lying today. N/A: "new tests added", "manual testing".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `<headroom_proactive_expansion>` XML tags, giving downstream consumers
   (LLMs, loggers, attribution parsers) a machine-readable provenance
   boundary and preventing misattribution in multi-agent threads.
+- **cli:** the startup banner no longer advertises
+  `HEADROOM_COMPRESSION_STABLE_AFTER_TURN` and
+  `HEADROOM_STALE_READ_COMPRESS_AFTER_TURNS` as tuning knobs. Both were read
+  only to render the `Performance Tuning` banner section and were never wired
+  into the compression path, so setting them changed the banner but had no
+  effect on behavior. The banner now surfaces only the embedding sidecar,
+  which is a real, consumed setting.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
Looks good. Removing banner-only knobs is the right direction here since advertising no-op tuning values creates operational confusion. I pushed a maintainer cleanup commit that only removes unrelated uv.lock drift from the PR diff.

PR #1640, opened 2026-07-01
fix(wrap): use -P for the proxy subprocess to avoid cwd package shadowing

Jinleen:
## Description

`headroom wrap <tool>` starts its shared proxy via `python -m headroom.cli proxy` in `_start_proxy()`. Python normally prepends the launching cwd to `sys.path`; therefore, invoking `headroom wrap` from a checkout with its own top-level `headroom/` package, including this repository, can shadow the installed package with the unbuilt source tree. The proxy then fails to import the compiled `headroom._core` extension.

The safe-path behavior is compatible with Headroom's supported Python versions:

- Python 3.11+ uses `-P` (`PYTHONSAFEPATH`).
- Python 3.10 uses a `-c` bootstrap that removes the implicit cwd entry before running `headroom.cli` with `runpy`. `alter_sys=True` preserves the `python -m` main-module context.

## Type of Change

## Changes Made

- `headroom/cli/wrap.py`: select the safe-path launcher based on the launching interpreter version.
- `tests/test_cli/test_wrap_proxy_detach.py`: cover both the Python 3.11+ `-P` branch and the Python 3.10 compatibility branch.

## Testing

- [x] Focused unit tests pass
- [x] Ruff lint and format checks pass
- [x] Mypy passes for `headroom/cli/wrap.py`
- [x] Actual CPython 3.10 compatibility behavior verified

### Test Output

```text
$ .venv/bin/python -m pytest tests/test_cli/test_wrap_proxy_detach.py
6 passed

$ .venv/bin/python -m ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_proxy_detach.py
All checks passed!

$ .venv/bin/python -m ruff format --check headroom/cli/wrap.py tests/test_cli/test_wrap_proxy_detach.py
2 files already formatted

$ .venv/bin/python -m mypy headroom/cli/wrap.py
```

## Real Behavior Proof

- Environment: Linux on Python 3.14 plus an actual CPython 3.10.20 compatibility probe in isolated temporary directories; details follow.
  - Linux, Python 3.14; `headroom-ai[all]` installed with `uv tool install`; this checkout was the development tree; cwd was the repository root.
  - Linux, CPython 3.10.20 installed by `uv`; isolated temporary test directories for the compatibility check.
- Exact command / steps: ran six focused launcher tests, Ruff, formatting, mypy, and direct Python 3.10/3.11+ safe-path probes; detailed commands follow.
  1. From the repository root, the installed Python resolved `import headroom` to `<repo>/headroom/__init__.py`; with `-P`, it resolved to the installed package in `site-packages`. Before this PR, the proxy launched by `headroom wrap <tool>` therefore loaded the unbuilt source tree and failed with `No module named 'headroom._core'`. With the Python 3.11+ path in this PR, the proxy started successfully from the same cwd.
  2. Verified the Python 3.10 constraint directly: `python3.10 -P -c "pass"` exits with `Unknown option: -P` (exit code 2).
  3. Created an isolated cwd with a top-level `headroom` package that raises on import, plus a separate path containing the intended `headroom.cli` package. From that cwd, running the module without removing `sys.path[0]` imported the cwd package and failed as expected.
  4. Ran the exact Python 3.10 compatibility bootstrap used by this PR:
     ```bash
     python3.10 -c "import runpy, sys; sys.path.pop(0); runpy.run_module('headroom.cli', run_name='__main__', alter_sys=True)" proxy --port 8787
     ```
     It loaded the intended package, preserved `proxy --port 8787`, and completed successfully.
- Observed result: the proxy no longer depends on the launching cwd on Python 3.11+, and the Python 3.10 fallback prevents the same implicit-cwd shadowing without passing its unsupported `-P` flag.
- Not tested: an end-to-end `headroom wrap <tool>` run under a separately installed production Headroom package on Python 3.10; Windows/macOS; and `headroom/install/runtime.py`, whose similar persistent-deployment launch path remains outside this PR's scope.

## Review Readiness

## Checklist

- [x] No user-facing documentation change is required (CHANGELOG)
- [x] New and existing focused tests pass locally with my changes
- [x] I did not edit CHANGELOG.md; release notes are generated automatically


JerrettDavis (changes requested):
Thanks for narrowing down the cwd shadowing root cause. This needs one compatibility adjustment before it can merge: Headroom still declares Python 3.10 support (`requires-python = ">=3.10"` and the 3.10 classifier), but `python -P` is only available on newer CPython. On a supported 3.10 interpreter this spawned command will fail before `headroom.cli` starts with an unknown-option error.

Can you make the safe-path behavior version-compatible? For example, use `-P` only when the launching interpreter supports it, and use another safe cwd/sys.path strategy for 3.10, with a regression test that exercises the 3.10 branch. The current test only asserts `-P` is always present, so it would lock in the unsupported-interpreter regression.

Jinleen:
Thanks, addressed in 1f8c4d91.

- Python 3.11+ continues to use `-P`.
- Python 3.10 uses a `-c`/`runpy` compatibility bootstrap that removes the implicit cwd `sys.path[0]` entry before loading `headroom.cli`; `alter_sys=True` preserves the `-m` main-module context.
- Added regression coverage for both version-selection branches.

I also verified this with CPython 3.10.20: `-P` fails with the expected unknown-option error, while the compatibility bootstrap succeeds in an isolated cwd-package-shadowing reproduction. Focused tests, Ruff, format, and mypy all pass.

JerrettDavis (approved):
The Python 3.10 blocker from my prior review is addressed in the updated head. The proxy launcher now uses `-P` only on Python 3.11+ and falls back to a `-c`/`runpy` bootstrap on Python 3.10 that removes the implicit cwd path before loading `headroom.cli`.

Local validation:
- `.venv\Scripts\python.exe -m pytest tests/test_cli/test_wrap_proxy_detach.py -q` -> 6 passed
- `.venv\Scripts\python.exe -m ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_proxy_detach.py` -> passed

The only GitHub checks on this PR are the governance checks, so this approval is based on the focused local validation above.

JerrettDavis (approved):
Re-reviewed after updating onto current main and removing the manual changelog entry. The approved Python 3.10/3.11+ safe-path launcher remains intact, the branch is mergeable, and release notes are left to release-please. The prior focused validation still covers both version branches.

PR #1643, opened 2026-07-01
fix(proxy): gate mid-turn message coalescing to Claude Code clients

### LINKED ISSUE #1608: [BUG] OpenCode subagents error
## Description

While connected to headroom proxy created with the command: headroom proxy --port 8787 --anthropic-api-url https://my/llm/endpoint-v1 I cannot use opencode subagents because opencode throws the error:
Error message: [
  {
    "code": "invalid_union",
    "errors": [],
    "note": "No matching discriminator",
    "discriminator": "type",
    "path": [
      "type"
    ],
    "message": "Invalid input"
  }
]

the subagent is spawned successfully but it only receives the prompt from the main agent and never responds, so currently subagents cannot be used with headroom (this increases token consumption a lot).

## To Reproduce

1. run headroom proxy with provided command
2. let opencode.json point to it:
"headroom-proxy": {
      "npm": "@ai-sdk/anthropic",
      "name": "Headroom Proxy",
      "options": {
        "baseURL": "http://127.0.0.1:8787/v1",
        "apiKey": "{env:ANTHROPIC_API_KEY}"
      }
    }

## Expected Behavior

Subagent to work

## Actual Behavior

Error message: [
  {
    "code": "invalid_union",
    "errors": [],
    "note": "No matching discriminator",
    "discriminator": "type",
    "path": [
      "type"
    ],
    "message": "Invalid input"
  }
]

tenderdeve:
## Description

`headroom wrap opencode` (and any other `@ai-sdk/anthropic` client) can't use subagents. The subagent is spawned, receives the prompt, and never responds; OpenCode throws `invalid_union / "No matching discriminator" / discriminator: "type"`.

Root cause is headroom's mid-turn message coalescing. It keys concurrent streaming requests by `md5(model:system[:500])` (`_get_session_key`, `handlers/streaming.py`). An OpenCode subagent runs concurrently with the main agent on the same model and same first-500-char system prefix, so it produces the **same** session key and collides with the still-active main stream. Two things then break it:

1. `handlers/anthropic.py` sees the key in `_active_streams` and answers the subagent's request with a bare `202 headroom_queued` instead of forwarding it — so the subagent never gets a response.
2. When the main stream ends, `handlers/streaming.py` emits a non-standard `event: headroom_pending_messages` SSE event. `@ai-sdk/anthropic`'s SSE parser keys its Zod union on `type`, and `headroom_pending_messages` isn't a valid Anthropic event type — hence the error.

The 202 reply and the `headroom_pending_messages` event are a Claude Code-only protocol (nothing else consumes them). This gates coalescing to Claude Code clients; every other harness streams normally.

Closes #1608

## Type of Change

## Changes Made

- `handlers/streaming.py`: only register a stream in `_active_streams` when `classify_client(headers) == "claude-code"`, and only emit the `headroom_pending_messages` SSE event for Claude Code.
- `handlers/anthropic.py`: only take the queue-and-`202` branch when the client is Claude Code (in addition to the existing `session_key in _active_streams` check).
- Regression tests in `tests/test_mid_turn_steering.py` for all four cases (active-stream registration and pending-event emission, each for a Claude Code vs. a non-Claude-Code client).

## Testing

### Test Output

```text
$ pytest tests/test_mid_turn_steering.py -q
9 passed in 0.46s

$ ruff check headroom/proxy/handlers/streaming.py headroom/proxy/handlers/anthropic.py tests/test_mid_turn_steering.py
All checks passed!

$ ruff format --check <same files>
3 files already formatted

$ mypy headroom/proxy/handlers/streaming.py headroom/proxy/handlers/anthropic.py --ignore-missing-imports
```

## Real Behavior Proof

- Environment: macOS (arm64), Python 3.14 venv, editable install of this branch.
- Exact command / steps: ran `pytest tests/test_mid_turn_steering.py` — the new tests drive `_stream_response` with a queued mid-turn message under an `opencode/1.0` User-Agent vs. a `claude-code/1.2.3` User-Agent and assert the streamed bytes. Also ran the streaming + anthropic handler suites (`pytest tests/test_mid_turn_steering.py tests/test_proxy_streaming_* tests/test_anthropic_* tests/test_streaming_usage_parser.py`).
- Observed result: with the `opencode/1.0` client the session is never added to `_active_streams` and the response contains no `headroom_pending_messages` event; with `claude-code/1.2.3` both still happen (protocol preserved). Handler suites: 155 passed, 3 skipped. Before this change the non-Claude client received the `headroom_pending_messages` event (the exact byte string the OpenCode parser rejects).
- Not tested: end-to-end against a live OpenCode + real subagent run — reproduced deterministically at the proxy layer instead (the emitted SSE bytes are the direct source of the reported `invalid_union` error).

## Review Readiness

## Additional Notes

Gating on `classify_client == "claude-code"` (User-Agent `claude-code/` / `claude-cli/`) is the same client identification used elsewhere in the proxy. Unidentified clients (no recognized User-Agent) are treated as non-Claude-Code and stream normally, which is the safe default for this feature.

## Maintainer Update (2026-07-21)

- Removed the manual `CHANGELOG.md` entry so release-please remains the source of changelog updates; pushed `d8e36540`.
- Validation: `tests/test_mid_turn_steering.py` passed (12 tests), the related streaming/Anthropic suite passed (72 tests), Ruff check passed for touched files, Ruff format check passed, and `git diff --check upstream/main...HEAD` passed.

### DIFF
--- a/headroom/proxy/handlers/streaming.py
+++ b/headroom/proxy/handlers/streaming.py
@@ -12,7 +12,7 @@ import logging
 import time
 from typing import TYPE_CHECKING, Any
 
-from headroom.proxy.auth_mode import classify_client
+from headroom.proxy.auth_mode import classify_client, supports_mid_turn_coalescing
 from headroom.proxy.helpers import (
     RETRYABLE_OVERLOAD_STATUSES,
     jitter_delay_ms,
@@ -1091,7 +1091,7 @@ class StreamingMixin:
         # so concurrent requests from other harnesses (e.g. OpenCode subagents
         # that share a body-derived session key) are streamed normally instead
         # of being swallowed. (#1608)
-        if client == "claude-code":
+        if supports_mid_turn_coalescing(client):
             self._active_streams.add(session_key)
         headers = await apply_copilot_api_auth(headers, url=url)
         start_time = time.time()
@@ -1638,7 +1638,7 @@ class StreamingMixin:
                     client=client,
                     waste_signals=waste_signals,
                 )
-                if client == "claude-code" and pending_messages:
+                if supports_mid_turn_coalescing(client) and pending_messages:
                     pending_event = json.dumps(
                         {"type": "headroom_pending_messages", "messages": pending_messages}
                     )
--- a/headroom/proxy/handlers/streaming.py
+++ b/headroom/proxy/handlers/streaming.py
@@ -1013,7 +1013,6 @@ class StreamingMixin:
         4. Streams the final response to the client
         """
         session_key = session_key or self._get_session_key(body)
-        self._active_streams.add(session_key)
 
         # Guard everything up to the generator's own try/finally (which owns
         # cleanup once streaming starts): any exception here — including
@@ -1085,6 +1084,15 @@ class StreamingMixin:
         # ...) from the *client's* User-Agent before copilot-auth
         # potentially rewrites headers for upstream.
         client = classify_client(headers)
+        # Mid-turn message coalescing (queueing a concurrent same-session
+        # request and later replaying it via a `headroom_pending_messages`
+        # SSE event) is a Claude Code-only protocol. Only register the stream
+        # as active for coalescing when the client can consume that protocol,
+        # so concurrent requests from other harnesses (e.g. OpenCode subagents
+        # that share a body-derived session key) are streamed normally instead
+        # of being swallowed. (#1608)
+        if client == "claude-code":
+            self._active_streams.add(session_key)
         headers = await apply_copilot_api_auth(headers, url=url)
         start_time = time.time()
 
@@ -1630,7 +1638,7 @@ class StreamingMixin:
                     client=client,
                     waste_signals=waste_signals,
                 )
-                if pending_messages:
+                if client == "claude-code" and pending_messages:
                     pending_event = json.dumps(
                         {"type": "headroom_pending_messages", "messages": pending_messages}
                     )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   contains the panic and treats the fragment as plain text, so the request is
   compressed and forwarded normally instead of returning HTTP 500
   ([#1547](https://github.com/headroomlabs-ai/headroom/issues/1547)).
+- `headroom wrap opencode` (and any other `@ai-sdk/anthropic` client) can now
+  use subagents. Mid-turn message coalescing keyed concurrent streaming
+  requests by `md5(model:system[:500])`, so a subagent sharing the main
+  agent's model and system prefix collided with the active stream: its request
+  was answered with a bare `202` and never forwarded, and the stream ended by
+  emitting a non-standard `headroom_pending_messages` SSE event the client
+  couldn't parse (`invalid_union / No matching discriminator`). Coalescing is a
+  Claude Code-only protocol and is now gated to Claude Code clients; other
+  harnesses stream normally
+  ([#1608](https://github.com/headroomlabs-ai/headroom/issues/1608)).
 - Proactive expansion blocks injected into user turns are now wrapped in
   `<headroom_proactive_expansion>` XML tags, giving downstream consumers
   (LLMs, loggers, attribution parsers) a machine-readable provenance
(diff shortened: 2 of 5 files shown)


JerrettDavis (approved):
Looks good. Gating active-stream registration and pending-message emission to Claude Code preserves the only client protocol that can consume those synthetic events while letting OpenCode/@ai-sdk subagent requests stream normally. The tests cover both sides of that gate.

JerrettDavis (approved):
Still looks good. The coalescing protocol remains available for Claude Code, while non-Claude clients no longer register active streams or receive `headroom_pending_messages` events they cannot parse. The tests cover both sides of that gate, and CI is green.

JerrettDavis (approved):
Reviewed the mid-turn coalescing gate. Restricting active-stream registration, queueing, and pending-message SSE emission to Claude Code keeps the custom protocol away from OpenCode/@ai-sdk clients, and the tests cover both client classes. I pushed a formatting-only commit so lint can rerun cleanly. No blocking issues found.

headroom/proxy/handlers/streaming.py
```diff
@@ -1488,7 +1496,7 @@ async def generate():
                     client=client,
                     waste_signals=waste_signals,
                 )
-                if pending_messages:
+                if client == "claude-code" and pending_messages:
```
JerrettDavis: These checks feel flakey to me. We have provider-specific feature slices, and it feels like we should implement something more robust here to support this extension. The manual client string checking feels like a code smell.

headroom/proxy/handlers/streaming.py
```diff
@@ -943,14 +943,22 @@ async def _stream_response(
...
         # potentially rewrites headers for upstream.
         client = classify_client(headers)
+        # Mid-turn message coalescing (queueing a concurrent same-session
+        # request and later replaying it via a `headroom_pending_messages`
+        # SSE event) is a Claude Code-only protocol. Only register the stream
+        # as active for coalescing when the client can consume that protocol,
+        # so concurrent requests from other harnesses (e.g. OpenCode subagents
+        # that share a body-derived session key) are streamed normally instead
+        # of being swallowed. (#1608)
+        if client == "claude-code":
```
JerrettDavis: These checks feel flakey to me. We have provider-specific feature slices, and it feels like we should implement something more robust here to support this extension. The manual client string checking feels like a code smell.

JerrettDavis (changes requested):

headroom/proxy/handlers/streaming.py
tenderdeve: Good call — replaced the inline `client == "claude-code"` checks with a named `supports_mid_turn_coalescing(client)` capability in `auth_mode.py`, backed by a single documented `_COALESCING_CAPABLE_CLIENTS` set. Both streaming sites and the anthropic queue gate now go through it, so the "which clients can consume this protocol" decision lives in one place (mirrors the existing `claude_code_tool_search_inactive` predicate) instead of scattered string comparisons. Added a unit test for the predicate. Also rebased onto `main` (resolved against the new `auth_policy` split + `_should_queue_mid_turn`) and signed both commits — PR is mergeable again.

headroom/proxy/handlers/streaming.py
tenderdeve: Good call — replaced the inline `client == "claude-code"` checks with a named `supports_mid_turn_coalescing(client)` capability in `auth_mode.py`, backed by a single documented `_COALESCING_CAPABLE_CLIENTS` set. Both streaming sites and the anthropic queue gate now go through it, so the "which clients can consume this protocol" decision lives in one place (mirrors the existing `claude_code_tool_search_inactive` predicate) instead of scattered string comparisons. Added a unit test for the predicate. Also rebased onto `main` (resolved against the new `auth_policy` split + `_should_queue_mid_turn`) and signed both commits — PR is mergeable again.

tenderdeve:
@JerrettDavis addressed the code-smell: the manual client-string checks are gone, replaced by a `supports_mid_turn_coalescing()` capability in `auth_mode` (single documented client set, used by both streaming sites + the anthropic gate). Rebased onto `main` and signed — mergeable again. Details in the inline threads.

JerrettDavis (approved):
The latest version addresses the previous blocker. The coalescing capability is centralized, the queue/202 path and synthetic pending-message SSE stay limited to Claude Code, and non-Claude clients stream normally. I removed the manual changelog entry and reran the focused mid-turn/streaming coverage locally: 72 related tests passed, Ruff check/format passed, and diff whitespace is clean.

PR #1645, opened 2026-07-01
fix(proxy): retry HTTP/2 stream resets instead of 502ing

### LINKED ISSUE #1639: 502s under concurrent large requests: HTTP/2 stream-reset cascade + HEADROOM_HTTP2 ignored on `install agent run` path
## Summary

Under concurrent load with large request bodies, the proxy returns **HTTP 502** to the client for Anthropic (`/v1/messages`) traffic. Root trigger is HTTP/2 upstream: a single stream reset from the upstream cascades into `LocalProtocolError` on every other in-flight stream sharing that h2 connection.

Separately, the **`HEADROOM_HTTP2` env var (and there is no `--no-http2`) is silently ignored on the `headroom install agent run` launch path**, so the intended workaround does not work.

- headroom 0.27.0 (pipx), httpx 0.28.1, h2 4.3.0, Python 3.13, macOS
- Backend: anthropic, `--mode token`, persistent-service profile
- Auth: Claude subscription OAuth (`sk-ant-oat`)

## Symptom / logs

Requests with 300KB–620KB bodies (10 concurrent Claude Code sessions through one proxy) hang 40–78s then 502:

```
ERROR [hr_...] Request failed: RemoteProtocolError: <StreamReset stream_id:35, error_code:1, remote_reset:True>
ERROR [hr_...] Request failed: LocalProtocolError: 39
ERROR [hr_...] Request failed: LocalProtocolError: 41
ERROR [hr_...] Request failed: LocalProtocolError: 43
INFO  event=proxy_inbound_response ... status=502 duration_ms=78712.33
```

Sequence over ~60 recent `/v1/messages` responses: 47×200, **12×502**, 1×429. The 502s always follow a `RemoteProtocolError` StreamReset on one stream, then `LocalProtocolError: <n>` on the sibling streams of the same connection (the `<n>` values are the poisoned connection's stream IDs). Small requests never reproduce it.

## Root cause

**Bug 1 — `HEADROOM_HTTP2` ignored on the CLI/agent path.**
`headroom install agent run` runs `python -m headroom.cli proxy ...` (`install/runtime.py: build_runtime_command`). `headroom/cli/proxy.py` constructs `ProxyConfig(...)` but **never sets `http2`**, and this path has **no `--no-http2` flag** and never reads `HEADROOM_HTTP2`. So `http2` falls to the dataclass default `True` (`headroom/proxy/models.py:239`). Only `headroom/proxy/server.py` (`http2=_get_env_bool("HEADROOM_HTTP2", True)`, ~line 4221) honors the env var. The two entrypoints disagree.

**Bug 2 — h2 stream-reset cascade → 502, no isolation/retry.**
With `http2=True` (default) and one shared `httpx.AsyncClient` (`server.py:1165`), an upstream `RST_STREAM` (`error_code:1`) puts the h2 connection into a state where subsequent in-flight streams fail with `LocalProtocolError`, and each is surfaced to the client as a 502 rather than being retried on a fresh connection. Large bodies + high concurrency make it frequent.

## Fix suggestions

1. **Honor the knob on all paths.** In `headroom/cli/proxy.py`, add to the `ProxyConfig(...)` call:
   ```python
   http2=_get_env_bool("HEADROOM_HTTP2", True),
   ```
   and add a `--no-http2` flag mirroring `server.py`. (Also allow it in manifest `proxy_args`.)
2. **Resilience.** On `RemoteProtocolError`/`LocalProtocolError`, retry the failed request on a fresh connection instead of returning 502; consider defaulting large-body Anthropic upstreams to HTTP/1.1.

## Workaround (confirmed)

Setting `http2=False` eliminates the 502s (each request gets its own h1 connection, no shared stream to poison). Because of Bug 1, `HEADROOM_HTTP2=0` only takes effect after manually patching `cli/proxy.py` to read it — banner then shows `http2=DISABLED` and the StreamReset/LocalProtocolError cascade stops.

tenderdeve:
## Description

Under concurrent load with large request bodies, `/v1/messages` returns **HTTP 502**. A single upstream HTTP/2 stream reset poisons the shared h2 connection and raises `RemoteProtocolError` (`StreamReset`) / `LocalProtocolError` on every other in-flight stream:

```
ERROR [hr_...] Request failed: RemoteProtocolError: <StreamReset stream_id:35, error_code:1, remote_reset:True>
ERROR [hr_...] Request failed: LocalProtocolError: 39
INFO  event=proxy_inbound_response ... status=502 duration_ms=78712
```

These are transport errors, but they weren't in the proxy's retry paths — the non-streaming `_retry_request` caught `(ConnectError, TimeoutException, HTTPStatusError)` and the streaming connect loop caught `(ConnectError, ConnectTimeout, PoolTimeout)`. So a stream reset skipped retry entirely and fell through to the broad handler catch as a `502`, with no reconnect.

This broadens both retry paths to treat any `httpx.TransportError` — which includes the h2 `Local`/`RemoteProtocolError` — as retryable, so the poisoned connection is dropped and the request re-sent on a fresh one.

Closes #1639

> Scope note: the issue also mentions `HEADROOM_HTTP2` being ignored on the `headroom install agent run` launch path. That's a separate config-plumbing gap; I've kept this PR to the 502-cascade fix (which makes the chain self-recover regardless of the env workaround) and am happy to follow up on the env plumbing separately.

## Type of Change

## Changes Made

- `headroom/proxy/server.py` (`_retry_request`): the retry `except` now catches `(httpx.TransportError, httpx.HTTPStatusError)` instead of `(ConnectError, TimeoutException, HTTPStatusError)`. `TransportError` is the common base of ConnectError, the timeout family, and the protocol/network errors — so h2 stream resets are retried with backoff.
- `headroom/proxy/handlers/streaming.py`: the streaming connect-retry loop and its terminal handler now catch `httpx.TransportError`. The retry runs before any body byte is forwarded to the client (only `build_request` + `send(stream=True)` are inside the loop), so re-sending is safe. On exhaustion the terminal handler still emits a clean `event: error` SSE instead of letting the reset bubble up as a 502. The mid-stream handler was left as-is (already covered by its `except Exception`, and not safe to retry once bytes have been sent).

## Testing

### Test Output

```text
$ pytest tests/test_h2_stream_reset_retry.py -q
4 passed

$ pytest tests/test_proxy_streaming_resilience.py tests/test_mid_turn_steering.py \
        tests/test_proxy_streaming_ratelimit_headers.py tests/test_streaming_usage_parser.py \
        tests/test_proxy_byte_faithful_forwarding.py -q
87 passed, 1 skipped

$ ruff check <changed files> && ruff format --check <changed files>
All checks passed! / 3 files already formatted

$ mypy headroom/proxy/server.py headroom/proxy/handlers/streaming.py --ignore-missing-imports
```

## Real Behavior Proof

- Environment: macOS (arm64), Python 3.14 venv, editable install of this branch.
- Exact command / steps: ran `pytest tests/test_h2_stream_reset_retry.py` — the tests drive the real `_retry_request` and `_stream_response` with `http_client.post` / `http_client.send` set to raise `httpx.RemoteProtocolError("<StreamReset ...>")` on the first attempt and return a good response on the second.
- Observed result: non-streaming — the request is retried and returns the `200` response (`post` awaited twice); on unconditional resets it re-raises after `retry_max_attempts` (no silent hang). Streaming — the reset on `send()` is retried and the upstream SSE (`message_start`…) is forwarded with no `connection_error` event (`send` awaited twice); on repeated resets a clean `event: error` SSE is emitted rather than a crash/502. Before this change the same `RemoteProtocolError` was uncaught and propagated to the `502` handler.
- Not tested: a live 10-session concurrent-load repro against a real Anthropic h2 endpoint — reproduced deterministically at the retry boundary with an injected `RemoteProtocolError` instead.

## Review Readiness

## Additional Notes

Retrying a stream reset re-sends the (potentially large) body, but that is bounded by the existing `retry_max_attempts` + jittered backoff and only happens before the first client byte — the same contract the existing connect-error retry already relied on. This is complementary to, not a replacement for, an operator forcing HTTP/1.1; it makes the default h2 path self-heal from transient resets.


JerrettDavis (approved):
Looks good. Retrying the broader httpx.TransportError family before streaming bytes are sent is the right scope for HTTP/2 stream-reset recovery, and the exhaustion path still returns a structured SSE error rather than leaking the protocol exception.

JerrettDavis (approved):
Still looks good. The current revision keeps the retry scope at the transport layer, before streaming bytes are forwarded, so retrying HTTP/2 stream resets is safe. The non-streaming and streaming tests cover retry success and exhaustion behavior, and CI is green.

JerrettDavis (approved):

PR #1658, opened 2026-07-01
fix(models): version-boundary longest-prefix match in ModelRegistry.get

abhay-codes07:
## Description

`ModelRegistry.get()` has a prefix fallback for versioned model ids. It accepted
**any** registered name as a bare `str.startswith` prefix and returned the
**first** match in dict-insertion order:

```python
for name, info in _MODELS.items():
    if model_lower.startswith(name):
        return info
```

Two concrete failures fall out of that:

- `gpt-4` is registered before `gpt-4-32k`, so `get("gpt-4-32k-0613")` matches
  `gpt-4` first and returns an **8192**-token window instead of `gpt-4-32k`'s
  **32768**.
- `gpt-4.1` / `gpt-4.5-preview` aren't registered, so they also match `gpt-4`
  and inherit its **8192**-token window — even though they're much larger,
  distinct models.

`get_context_limit()` reads straight from `get()` (no LiteLLM fallback), so both
cases make the proxy believe a nearly-empty context is almost full and compress
far too aggressively — or reject — on requests that are actually small. This is
silent: no error, just a wrong number driving every downstream compression
decision for those models.

## Fix

The fallback now:

1. Only matches when the registered name ends at a **version boundary** in the
   query — the next character must be a separator (`-`, `/`, `:`, `@`, `_`) — so
   `gpt-4.1`'s `.` no longer matches `gpt-4` (it falls through to the caller's
   default instead of a wrong 8192).
2. Picks the **longest** qualifying name, so `gpt-4-32k-0613` → `gpt-4-32k`.

Exact and alias lookups are unchanged, and boundary-separated variants like
`gpt-4o-new-version` still resolve to `gpt-4o`.

## Type of Change

## Changes Made

- `headroom/models/registry.py`: replace the first-match `startswith` prefix loop in `ModelRegistry.get` with a longest-prefix-at-a-version-boundary match.
- `tests/test_models.py`: add regression tests — `gpt-4-32k-0613` → `gpt-4-32k` (32768), and `gpt-4.1`/`gpt-4.5-preview` no longer resolve to gpt-4's 8192 window.

## Testing

- [x] New tests added for the fixed behavior (`tests/test_models.py`)
- [ ] Full `pytest` run deferred to CI — see Real Behavior Proof for why I verify the logic with a dependency-free script locally.

```text
$ uv run ruff check headroom/models/registry.py tests/test_models.py
All checks passed!
$ uv run ruff format --check headroom/models/registry.py tests/test_models.py
2 files already formatted
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.11, headroom built from this branch (`uv sync --extra dev`). Importing `headroom` pulls in the torch/transformers stack; a full `pytest` run exhausts memory and gets OOM-killed on this box, so I verify the matching logic with a dependency-free script (only stdlib) and leave the full pytest to CI.
- Exact command / steps: replicated the relevant `_MODELS` registration order (`gpt-4o`, `gpt-4-turbo`, `gpt-4`, `gpt-4-32k`) and the new longest-prefix-with-boundary loop in a standalone script (no `headroom` import), then asserted the resolved context windows.
- Observed result: `gpt-4-32k-0613` resolves to 32768 (was 8192 under first-match), `gpt-4.1`/`gpt-4.5-preview` fall through to the caller default (no longer 8192), and `gpt-4o-new-version` / `gpt-4` / `gpt-4-0613` resolve exactly as before:

```text
OK: gpt-4-32k-0613 -> 32768 (was 8192 under old first-prefix-wins)
OK: gpt-4.1 / gpt-4.5-preview -> default (not 8192)
OK: gpt-4o-new-version, gpt-4, gpt-4-0613 still resolve as before
REGISTRY LOGIC VERIFIED
```

- Not tested: I did not add explicit registry entries for `gpt-4.1`/`gpt-4.5` (their real windows) — that's a data addition, separate from this matching-logic fix; today they fall back to the caller's default, which is honest for an unregistered model and strictly better than the previous wrong 8192. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- No new dependencies; pure logic change in one function plus tests.
- Found via a read-through of the registry while looking at how context limits drive compression decisions.


JerrettDavis (approved):
Reviewed the registry matching change and the added regression coverage. The longest-prefix plus explicit boundary check fixes the gpt-4-32k ordering bug without continuing to misclassify gpt-4.1/gpt-4.5 as gpt-4. Looks scoped and ready from code review.

JerrettDavis (approved):
Still looks good. The longest qualifying prefix plus explicit separator-boundary check fixes the `gpt-4-32k-0613` ordering bug while preventing distinct dotted model families like `gpt-4.1` from inheriting `gpt-4` limits. The tests pin both behaviors, and CI is green.

PR #1672, opened 2026-07-02
fix(proxy/auth): match real Anthropic OAuth token prefix (sk-ant-oat)

abhay-codes07:
## Description

`classify_auth_mode` (in `headroom/proxy/auth_mode.py`) checks for Anthropic
OAuth tokens with:

```python
if token.startswith("sk-ant-oat-"):
    return AuthMode.OAUTH
if token.startswith("sk-ant-api") or token.startswith("sk-"):
    return AuthMode.PAYG
```

But real Anthropic OAuth access tokens are **`sk-ant-oat01-...`** — a version
number right after `oat`, **no dash**. So the `sk-ant-oat-` check never matches a
real token; it falls through to the broad `sk-` rule and gets classified
**`PAYG`**.

That's exactly the misclassification the module is built to prevent: a
subscription/OAuth-bound request tagged `PAYG` gets the aggressive-compression
policy — lossy compression, auto `cache_control`, `prompt_cache_key` injection —
instead of the passthrough-prefer path OAuth is meant to get.

The existing tests didn't catch it because they use a synthetic `sk-ant-oat-01-`
fixture (dashed) that happens to match the buggy prefix. Corroboration that the
real shape is dash-less:
- `.gitguardian.yaml` fixture: `sk-ant-oat01-oauth-fixture`
- `tests/test_oauth_bearer_routing.py`: `sk-ant-oat01-xxx`
- the sibling helper `headroom/proxy/helpers.py` matches on `sk-ant-` (no `oat-`)

## Fix

Match the dash-less `sk-ant-oat` prefix. It still matches the legacy dashed
shape, and ordering relative to `sk-ant-api` / `sk-` is unchanged (OAuth is
still checked first).

```python
if token.startswith("sk-ant-oat"):
    return AuthMode.OAUTH
```

## Type of Change

## Changes Made

- `headroom/proxy/auth_mode.py`: match OAuth tokens on the dash-less `sk-ant-oat` prefix.
- `tests/test_auth_mode.py`: add a regression test using the real `sk-ant-oat01-...` format (the existing test keeps the legacy dashed fixture, which still classifies correctly).

## Testing

- [x] New regression test added (`tests/test_auth_mode.py`)

```text
$ uv run ruff check headroom/proxy/auth_mode.py tests/test_auth_mode.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.11, headroom from this branch. Importing `headroom` loads the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the classification logic with a dependency-free script and left the full pytest to CI.
- Exact command / steps: replicated the Bearer-token branch of `classify_auth_mode` in a standalone script (only stdlib, no `headroom` import) and ran the real and legacy token shapes plus PAYG keys through it.
- Observed result: the real `sk-ant-oat01-...` now classifies OAUTH (was PAYG before the change); the legacy dashed fixture still classifies OAUTH; `sk-ant-api*` / `sk-*` keys still classify PAYG:

```text
OK: sk-ant-oat01-... -> OAUTH (was PAYG before fix)
OK: sk-ant-oat-01-... -> OAUTH (legacy fixture still matches)
OK: sk-ant-api* / sk-* -> PAYG (unchanged)
AUTH LOGIC VERIFIED
```

- Not tested: a live proxied Anthropic OAuth request end-to-end (needs a real subscription token); the classification is pure and covered by the regression test. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- `crates/headroom-core/src/auth_mode.rs` carries the identical dashed prefix (its Rust test matrix uses the same synthetic dashed fixture). I scoped this PR to the Python runtime classifier since that's the request-time path; happy to mirror the one-line fix in Rust in the same PR or a follow-up — I just couldn't `cargo build` locally to verify, so I left it out rather than push an unverified Rust edit.
- @JerrettDavis tagging you since you've been triaging these — small, contained fix with a regression test if you have a moment.


JerrettDavis (approved):
Looks good after the branch update. The Python classifier now catches the real `sk-ant-oat01-...` Anthropic OAuth token shape before the broad `sk-` PAYG branch, and I added the same prefix fix plus regression coverage to `crates/headroom-core` so the Python and Rust classifiers do not diverge.

I updated the branch from `main`, resolved the changelog-only conflict, and verified:
- `python -m pytest tests/test_auth_mode.py -q`
- `cargo test -p headroom-core --test auth_mode`

GitHub checks re-queued after the push and should finish before merge.

PR #1691, opened 2026-07-02
fix(memory): singleflight LocalBackend init to stop cold-start races

### LINKED ISSUE #1678: [BUG] AssertionError with --memory on large contexts: race in LocalBackend._ensure_initialized + CancelledError escapes memory context timeout
## Summary

When running headroom proxy with `--memory` against a long conversation (~60-80k tokens), every request fails with a bare `AssertionError` (no message) returned to the upstream client, causing retries that all fail identically. Elapsed time is ~0.1s (no upstream call is ever made). Removing `--memory` fixes the issue immediately.

## Environment

- headroom-ai version: 0.28.0
- Backend: anthropic / claude-sonnet-4-6
- Proxy flags: `headroom proxy --port 8787 --memory --no-ccr-inject-tool`
- Embedder: `HEADROOM_EMBEDDER_RUNTIME=pytorch_mps` (Apple Silicon M-series)
- Client: Hermes Agent (wraps Anthropic Python SDK via `ANTHROPIC_BASE_URL=http://localhost:8787`)
- Storage mode: PROJECT (default)

## Symptoms

```
API call failed (attempt 1/3): AssertionError
  Provider: anthropic  Model: claude-sonnet-4-6
  Endpoint: http://localhost:8787
  Error: (empty)
  Elapsed: 0.11s  Context: 31 msgs, ~81,457 tokens
```

0.11s elapsed means the proxy crashes internally before ever forwarding to Anthropic. The error has no message, which is the signature of a bare `assert` statement.

## Root Cause Analysis

Three bugs compound:

### Bug 1 — New per-project backends from BackendRouter are never initialized

`storage_router.py` `BackendRouter._get_or_create_backend()` creates a `LocalBackend` but never calls `_ensure_initialized()` on it before returning it to the caller:

```python
# storage_router.py:388
backend = LocalBackend(cfg)           # created uninitialized
self._backends[db_path] = backend
...
return backend                        # returned uninitialized
```

The first call to `search_memories` on this backend triggers lazy init inline.

### Bug 2 — LocalBackend._ensure_initialized has no async lock

`local.py:120` uses a bare `if not self._initialized:` guard with no `asyncio.Lock`. Two concurrent async tasks can both pass the check and both `await HierarchicalMemory.create()`. More critically, with `HEADROOM_EMBEDDER_RUNTIME=pytorch_mps`, the first-ever `HierarchicalMemory.create()` on a fresh project DB can take >2 seconds (torch MPS warm-up + SQLite init + model load):

```python
async def _ensure_initialized(self):
    if not self._initialized:        # no lock — two tasks can enter here
        self._hierarchical_memory = await HierarchicalMemory.create(...)  # slow on MPS
        self._graph = ...
        self._initialized = True
```

### Bug 3 — asyncio.CancelledError escapes the memory context timeout handler

`anthropic.py:1662` wraps `search_and_format_context` in `asyncio.wait_for(..., timeout=2.0)`:

```python
try:
    memory_context = await asyncio.wait_for(
        self.memory_handler.search_and_format_context(...),
        timeout=self.config.anthropic_pre_upstream_memory_context_timeout_seconds,  # 2.0s
    )
except asyncio.TimeoutError:
    memory_context = None
    ...
```

When `wait_for` fires after 2s, it cancels the inner coroutine with `asyncio.CancelledError`. `CancelledError` is a `BaseException` — it is NOT caught by the `except asyncio.TimeoutError` handler (and not by any `except Exception` in the chain). It propagates all the way up through the request handler to the Anthropic Python SDK client, which does not expect an exception here and whose internal assertion fails:

```
AssertionError  (no message — bare assert in the SDK's response-parsing path)
```

On subsequent retries the proxy may have a partially-initialized backend (`_hierarchical_memory` still `None`), causing the bare `assert self._hierarchical_memory is not None` at `local.py:237` or `385` to fire immediately, explaining the 0.11s elapsed on retries.

## Trigger path (Hermes / no explicit headers)

Hermes injects a system prompt that contains `Current working directory: /Users/...`. The `ProjectResolver` Tier-4 parser matches this as a cwd signal and resolves a per-project key. On the very first request in a new working directory, `BackendRouter._get_or_create_backend` creates a fresh `LocalBackend` for that project. Because MPS init is slow (>2s), the 2s memory context timeout fires before the backend is ready, `CancelledError` escapes, and the whole request dies.

## Proposed Fixes

**Fix 1 — `local.py`: Add an asyncio.Lock to `_ensure_initialized`**

```python
def _get_init_lock(self) -> asyncio.Lock:
    if self._init_lock is None:
        self._init_lock = asyncio.Lock()
    return self._init_lock

async def _ensure_initialized(self) -> None:
    if self._initialized:
        return
    async with self._get_init_lock():
        if self._initialized:   # double-check under lock
            return
        self._hierarchical_memory = await HierarchicalMemory.create(mem_config)
        ...
        self._initialized = True
```

**Fix 2 — `anthropic.py`: Catch CancelledError alongside TimeoutError**

```python
except (asyncio.TimeoutError, asyncio.CancelledError):
    memory_context = None
    logger.info(f"[{request_id}] Memory: context lookup cancelled/timed out; continuing without it")
```

**Fix 3 (defence-in-depth) — `storage_router.py`: Warm up the backend before returning**

```python
backend = LocalBackend(cfg)
# Schedule initialization so the first real request doesn't pay the full cold-start cost
# (or at minimum document that callers must call _ensure_initialized before use)
self._backends[db_path] = backend
```

Fix 2 is the minimal required change — it stops the `CancelledError` from escaping. Fix 1 prevents double-init races. Fix 3 prevents the slow cold-start from triggering the timeout in the first place.

## Workaround

Run without `--memory`:
```bash
headroom proxy --port 8787 --no-ccr-inject-tool
```

Or pre-warm the backend by making a small request first (so `HierarchicalMemory.create` completes within the 2s window before a large-context request arrives).

tenderdeve:
## Description

Running the proxy with `--memory` against a large context throws a bare `AssertionError` (empty message, ~0.1s elapsed, no upstream call) on every request; dropping `--memory` makes it go away.

Per-project backends handed out by `BackendRouter._get_or_create_backend` init lazily on first use. `LocalBackend._ensure_initialized` guarded init with a bare `if not self._initialized:` and no `asyncio.Lock`, so concurrent first callers each kicked off a parallel `HierarchicalMemory.create()`. A slow cold-start (>2s on the `pytorch_mps` embedder) cancelled by the outer 2s memory-context `wait_for` left the backend half-built (`_hierarchical_memory` still `None`), so the retry tripped `assert self._hierarchical_memory is not None` (local.py:237/385/...) — the empty-message crash. `MemoryHandler._ensure_initialized` already uses a double-checked `asyncio.Lock`; the per-project `LocalBackend` never got the same treatment.

Closes #1678

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add a lazily-created `asyncio.Lock` singleflight with a double-checked flag to `LocalBackend._ensure_initialized`, mirroring the existing `MemoryHandler` pattern — concurrent first callers await one init instead of racing N.
- On `CancelledError` (e.g. the outer `wait_for` timeout mid cold-start), reset `_hierarchical_memory`/`_graph`/`_initialized` and re-raise, so a cancelled init never leaves a half-built backend for the next request to assert on.
- Move the init body verbatim into `_init_locked()` (called with the lock held); the large diff is the dedent, no logic change.

## Testing

### Test Output

```text
$ pytest tests/test_local_backend_init_race.py tests/test_memory_handler_concurrent_init.py -q
tests/test_local_backend_init_race.py ..                                 [ 20%]
tests/test_memory_handler_concurrent_init.py .....s..                    [100%]
9 passed, 1 skipped in 0.50s

$ ruff check headroom/memory/backends/local.py tests/test_local_backend_init_race.py
All checks passed!

$ ruff format --check headroom/memory/backends/local.py tests/test_local_backend_init_race.py
2 files already formatted

$ mypy headroom/memory/backends/local.py
```

## Real Behavior Proof

- Environment: macOS (arm64), Python 3.14, local `.venv`.
- Exact command / steps: `pytest tests/test_local_backend_init_race.py tests/test_memory_handler_concurrent_init.py -q`. First test spawns 10 concurrent first callers against a `LocalBackend` with a patched slow `HierarchicalMemory.create` and asserts `create` runs exactly once; second cancels a cold-start via an outer `asyncio.wait_for` timeout, asserts state resets to `None`/uninitialized, then a later call re-inits cleanly.
- Observed result: both pass; `create` is called once under contention, and a cancelled init leaves no half-built backend.
- Not tested: no end-to-end repro of the original `--memory` crash against a real large context / GPU embedder — the race is reproduced deterministically at the unit level instead.

## Review Readiness

## Checklist

## Additional Notes

Docs/CHANGELOG unchanged — internal concurrency fix with no user-facing API or behavior change beyond removing the crash.

### DIFF
--- a/headroom/memory/backends/local.py
+++ b/headroom/memory/backends/local.py
@@ -13,6 +13,7 @@ single-process production deployments.
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import uuid
 from dataclasses import dataclass, field
@@ -116,6 +117,19 @@ class LocalBackend:
         self._initialized = False
         self._hierarchical_memory: HierarchicalMemory | None = None
         self._graph: InMemoryGraphStore | SQLiteGraphStore | None = None
+        # Async singleflight guard for lazy init. Per-project backends handed
+        # out by BackendRouter init lazily on first use; concurrent first
+        # callers must land on ONE init (double-checked pattern below) instead
+        # of racing N partial inits that leave ``_hierarchical_memory`` None
+        # and trip the ``assert`` guards downstream. Created lazily so the
+        # backend can be constructed before an event loop exists.
+        self._init_lock: asyncio.Lock | None = None
+
+    def _get_init_lock(self) -> asyncio.Lock:
+        """Lazily create the init lock bound to the running event loop."""
+        if self._init_lock is None:
+            self._init_lock = asyncio.Lock()
+        return self._init_lock
 
     async def _ensure_initialized(self) -> None:
         """Ensure the backend is initialized with all components.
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Code review looks good. The lazy asyncio.Lock plus double-check prevents concurrent cold starts from racing, and resetting partial state on CancelledError addresses the half-built backend path that led to bare assertions on retry. The focused concurrency and cancellation tests cover the important failure modes, and CI is green.

DevanshiVyas (approved):

PR #1699, opened 2026-07-02
feat(proxy): opt-in compression for catch-all passthrough routes

### LINKED ISSUE #1546: Feature Request: Compress passthrough/catch-all proxy routes (add custom route compression)
### Problem
When the headroom proxy receives requests with paths that don't match the built-in API route patterns (e.g., custom proxy paths like `/api/codex-proxy/<key>/v1/responses` or `/api/claude-code-proxy/<key>/v1/messages`), they fall through to the catch-all handler `handle_passthrough()` in `headroom/proxy/handlers/openai.py:6084`. 

This handler does **NO compression** — it forwards the request body verbatim to the upstream, bypassing the entire compression pipeline (ContentRouter, Kompress, TOIN, etc.).

The `forwarder=openai_passthrough` log output confirms this.

### Use case
I use a multi-provider AI agent platform (Hermes Studio) that wraps coding agent traffic in custom proxy paths for provider routing. The request flow is:

```
Codex/ClaudeCode → headroom(:8787) → Hermes proxy(:8748) → Real API
```

The Hermes proxy paths (e.g., `/api/codex-proxy/<base64routeKey>/v1/responses`) are not recognized by headroom as standard API paths, so all coding agent traffic goes through passthrough without compression. This causes context-limit errors (HTTP 400) in long sessions.

### Expected behavior
Either:
1. The catch-all handler should also apply the compression pipeline (optional, enabled via a flag like `--compress-passthrough`)
2. Or there should be a way to register custom route patterns that map to the compression pipeline (e.g., a `--custom-compress-route` flag that accepts path prefixes)

### Current workaround
- Only standard paths like `/v1/chat/completions`, `/v1/messages`, `/v1/responses` get compressed
- TOIN caching still provides some savings (~8K tokens/turn) even through passthrough
- Kompress model needs to be pre-downloaded separately

### Environment
- headroom v0.28.0
- Python 3.12, Windows 10
- Kompress ONNX (chopratejas/kompress-v2-base)
- headroom proxy launched with:
  ```
  headroom proxy --port 8787 --anthropic-api-url <upstream> --openai-api-url <upstream> --memory --learn
  ```

### Related
- Similar architecture is used by other AI agent platforms that wrap coding agents in custom proxy paths for multi-provider routing.
- Providing a `--compress-passthrough` flag or custom route mapping would make headroom compatible with any proxy wrapper architecture.

tenderdeve:
## Description

Requests whose path doesn't match a built-in API route fall through to `handle_passthrough`, which forwarded the body verbatim — bypassing ContentRouter/Kompress/TOIN entirely. Wrapper-proxy setups that front Headroom on custom paths (e.g. `/api/codex-proxy/<key>/v1/responses`) got zero compression on coding-agent traffic and hit context-limit 400s in long sessions. This adds an opt-in flag that routes OpenAI Responses-shaped passthrough bodies through the same compression path the native `/v1/responses` handler uses.

Closes #1546

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `ProxyConfig.compress_passthrough` (default `False`) + `--compress-passthrough` CLI flag + `HEADROOM_COMPRESS_PASSTHROUGH=1` env.
- `handle_passthrough`: when enabled, POST requests whose path ends in `/responses` with an OpenAI Responses-shaped body are compressed via the existing `_compress_openai_responses_payload_in_executor` before forwarding; stale `Content-Length` is dropped so httpx recomputes it.
- New `_maybe_compress_passthrough_responses` helper — fail-open: non-JSON, non-Responses payloads, unmodified results, and any compressor error forward the original body unchanged.
- Documented the flag in `docs/content/docs/proxy.mdx`.

## Testing

### Test Output

```text
$ .venv/bin/python -m pytest tests/test_compress_passthrough.py -q
collected 6 items
tests/test_compress_passthrough.py ......                                [100%]
============================== 6 passed in 0.35s ===============================

$ .venv/bin/ruff check headroom/proxy/handlers/openai.py headroom/proxy/models.py headroom/proxy/server.py tests/test_compress_passthrough.py
All checks passed!
```

## Real Behavior Proof

- Environment: macOS arm64, Python 3.14, repo `.venv`.
- Exact command / steps: `.venv/bin/python -m pytest tests/test_compress_passthrough.py -q` — covers a Responses-shaped body being compressed, non-JSON passthrough, non-Responses (`messages`) payload untouched, unmodified-result short-circuit, compressor-error fail-open, and `ProxyConfig().compress_passthrough is False` default. Plus import smoke: `ProxyConfig(compress_passthrough=True)`, server/handler modules import, helper present.
- Observed result: 6 passed; flag defaults off; enabled path reuses the native Responses compressor and never raises out to the request.
- Not tested: live end-to-end through a real second proxy to a real upstream (no external wrapper proxy / upstream credentials in sandbox); the compression call is the same one `/v1/responses` already exercises in CI.

## Review Readiness

## Checklist

## Additional Notes

Scoped to OpenAI Responses-shaped bodies (the reporter's exact case). Anthropic `/messages` and OpenAI `/chat/completions` passthrough compression are natural follow-ups — deliberately left out to keep this change focused and fail-safe. CHANGELOG is release-managed, left unchecked.

### DIFF
--- a/docs/content/docs/proxy.mdx
+++ b/docs/content/docs/proxy.mdx
@@ -92,11 +92,13 @@ headroom proxy --mode cache
 | `--no-learn` | `false` | Explicitly disable traffic learning |
 | `--min-evidence` | `5` | Minimum observations before a learned pattern is persisted |
 | `--codex-wire-debug` | `false` | Write local Codex wire snapshots and matching proxy log traces |
+| `--compress-passthrough` | `false` | Also compress custom proxy paths that fall through to the catch-all handler (OpenAI Responses-shaped bodies, path ends in `/responses`). Also `HEADROOM_COMPRESS_PASSTHROUGH=1` |
 
 ```bash
 headroom proxy --memory
 headroom proxy --learn --min-evidence 3
 headroom proxy --codex-wire-debug
+headroom proxy --compress-passthrough
 ```
 
 <Callout type="info" title="LLMLingua removed from the proxy CLI">
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -6374,6 +6374,52 @@ class OpenAIHandlerMixin:
                 },
             )
 
+    async def _maybe_compress_passthrough_responses(self, body: bytes) -> bytes:
+        """Compress an OpenAI Responses-shaped passthrough body, fail-open.
+
+        Reuses the native `/v1/responses` compression path so custom
+        wrapper-proxy routes get the same ContentRouter/Kompress treatment.
+        Any parse/compression failure returns the original body unchanged so a
+        catch-all request is never dropped by opting into compression.
+        """
+        try:
+            payload = json.loads(body)
+        except (json.JSONDecodeError, ValueError, TypeError):
+            return body
+        if not isinstance(payload, dict) or "input" not in payload:
+            # Not a Responses payload (no `input` array) — leave it alone.
+            return body
+
+        model = str(payload.get("model") or "passthrough")
+        request_id = await self._next_request_id()
+        try:
+            (
+                compressed_payload,
+                modified,
+                *_rest,
+            ) = await self._compress_openai_responses_payload_in_executor(
+                payload,
+                model=model,
+                request_id=request_id,
+            )
+        except Exception as exc:  # noqa: BLE001 — fail-open on any compressor error
+            logger.warning(
+                "[%s] passthrough Responses compression failed, forwarding verbatim: %s",
+                request_id,
+                exc,
+            )
+            return body
+        if not modified:
+            return body
+        try:
+            return json.dumps(
+                compressed_payload,
+                separators=(",", ":"),
+                ensure_ascii=False,
+            ).encode("utf-8")
+        except (TypeError, ValueError):
+            return body
+
     async def handle_passthrough(
         self,
         request: Request,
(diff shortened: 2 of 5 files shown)


JerrettDavis (changes requested):
Thanks for keeping this opt-in and fail-open. The shape of the feature makes sense, but this currently regresses the existing passthrough path when the handler/proxy object does not have a `config` attribute.

CI is failing in `tests/test_proxy_handler_helpers.py` with multiple existing passthrough tests raising:

```text
AttributeError: 'OpenAIHandlerMixin' object has no attribute 'config'
```

The new guard uses `getattr(self.config, "compress_passthrough", False)`, but Python has to resolve `self.config` before `getattr` can return the default. Please make the feature flag lookup tolerant of a missing config object, for example by first reading `config = getattr(self, "config", None)` and then checking `getattr(config, "compress_passthrough", False)`. That keeps the default-off behavior truly no-op for the existing passthrough paths and should make the old tests meaningful again.

tenderdeve:
@JerrettDavis updated in 389fc809 — the flag lookup now resolves `config = getattr(self, "config", None)` first, then `getattr(config, "compress_passthrough"/"optimize", False)`, so a missing `config` attribute is treated as feature-off and the existing verbatim passthrough path stays a true no-op. Added a regression test for the missing-config case; the previously-failing `tests/test_proxy_handler_helpers.py` passthrough tests are green again. Also rebased on `main` to clear the conflicts.

JerrettDavis (approved):
Reviewed the updated passthrough compression path. The flag is off by default, the guard is narrow to POST paths ending in /responses with Responses-shaped JSON, compression reuses the native Responses compressor, and every parse/no-op/compressor-error path forwards the original body. Stale Content-Length is removed after mutation and current checks are green; this looks ready.

PR #1702, opened 2026-07-02
fix(install): pass sc.exe create as raw command line so binPath= quoting survives (#1654)

### LINKED ISSUE #1654: install apply --preset persistent-service fails on Windows: sc.exe create rejects start= (error 1639)
## `headroom install apply --preset persistent-service` fails on Windows: `sc.exe create` — "invalid start= field" (error 1639)

**Environment**
- Windows 11 Home 10.0.26200, locale `es-ES`
- Python 3.13.14
- `headroom` 0.28.0 (installed via `pip install "headroom-ai[all]"`)
- Elevated ("Run as Administrator") PowerShell 5.1

**Command run**
```powershell
headroom install apply --preset persistent-service --runtime python --scope user --providers manual --target claude --memory
```

**Output**
```
ERROR: campo start= no válido

DESCRIPCIÓN:
        Crea una entrada de servicio en el Registro y en la base de datos
        de servicio.
USO:
        sc <servidor> create [nombre del servicio] [binPath= ] <opción1> <opción2>...
...
[SC] OpenService ERROR 1060:
El servicio especificado no existe como servicio instalado.

Error: Failed to install deployment 'default': Command '['sc.exe', 'create', 'headroom-default', 'binPath= cmd.exe /c "C:\\Users\\Adron\\.headroom\\deploy\\default\\run-headroom.cmd"', 'start= auto']' returned non-zero exit status 1639.
```

**Analysis**

The subprocess argv list shown in the error is:
```python
['sc.exe', 'create', 'headroom-default',
 'binPath= cmd.exe /c "C:\\Users\\Adron\\.headroom\\deploy\\default\\run-headroom.cmd"',
 'start= auto']
```

The `binPath=` value itself contains embedded double quotes (`cmd.exe /c "...run-headroom.cmd"`) plus spaces. When this argv list is handed to Windows `CreateProcess` (via Python's default list-to-command-line quoting), the outer quoting Python applies to the `binPath=` token collides with the quotes already embedded inside it. The resulting command line that `sc.exe` actually receives appears to get mis-tokenized, causing `start= auto` to be parsed as part of `binPath=`'s value (or otherwise shifted), so `sc.exe` reports `campo start= no válido` — even though `start= auto` alone is valid syntax.

This reproduces consistently (2/2 attempts, same machine). It is **not** a permissions issue — `OpenSCManager` succeeds (this run reached the actual `create` step), unlike a first attempt on the same machine without admin rights, which failed earlier at `OpenService`/`OpenSCManager` with access-denied (expected/correct behavior for the no-admin path).

**Suggested fix**

Avoid nesting a quoted `cmd.exe /c "..."` string inside the `binPath=` argv token. Either:
1. Build the full `sc.exe` command line as a single pre-quoted string and pass `shell=True` (or use `subprocess.list2cmdline` explicitly and verify the embedded-quote escaping), or
2. Point `binPath=` directly at a wrapper `.exe`/`.bat` with no embedded quotes/spaces in its own invocation (e.g., write args to the batch file itself rather than passing them inline in `binPath=`), or
3. Log/print the exact final command line string sent to `CreateProcess` in verbose mode, to make this class of bug easier to diagnose.

**Side note**: on failure, the `~/.headroom/deploy/default/` directory (including `run-headroom.cmd` and `manifest.json`) is removed, which made it harder to inspect what was actually generated. Consider leaving failed-install artifacts in place (or copying them to a `.failed/` dir) for post-mortem debugging.

**Workaround used**: fell back to the no-admin hook-based `ensure` mechanism (`headroom init hook ensure`), which works but has no process supervisor — the proxy process can die (observed twice) with nothing to restart it until the next hook fires.

Parideboy:
## Description

`headroom install apply --preset persistent-service` fails on Windows with `sc.exe` error 1639 ("invalid start= field"). The service install built the `sc.exe create` invocation as an argv list whose `binPath=` token embedded both spaces and inner double quotes (`cmd.exe /c "…run-headroom.cmd"`). Python's `subprocess.list2cmdline` then wrapped that whole token in outer quotes, so the command line `sc.exe` actually received tokenized as `'binPath= cmd.exe /c "…"'` and `'start= auto'` — single glued tokens — instead of the documented `binPath=` `<value>` `start=` `<value>` separate-token pairs. `sc.exe` rejects that with 1639.

This PR builds the exact command line as a pre-quoted string and passes it to `subprocess.run` directly; on Windows a string argument goes verbatim to `CreateProcess`, bypassing `list2cmdline` entirely. The `sc.exe failure` / `start` / `stop` / `delete` calls keep the argv-list form since none of their tokens embed quotes.

Fixes #1654

## Type of Change

- [ ] Documentation update

## Changes Made

- `headroom/install/supervisors.py`: the Windows `SERVICE` branch of `install_supervisor` now builds the `sc.exe create` command as a single pre-quoted string — `sc.exe create <name> binPath= "cmd.exe /c \"<run-headroom.cmd>\"" start= auto` — and passes it to `subprocess.run` as a string instead of an argv list.
- `tests/test_install/test_supervisors.py`: updated the Windows-service assertion to expect the new command-line string (regression test for #1654), verifying the backslash-escaped inner quotes and `start= auto` as a separate trailing pair.

## Testing

- [x] Unit tests pass (`tests/test_install/test_supervisors.py`)
- [x] Lint/type gates pass (`ruff check`, `ruff format --check`, `mypy`)

```
$ python -m pytest tests/test_install/ -q
94 passed, 1 failed, 1 skipped
# the 1 failure is tests/test_install/test_runtime.py::test_runtime_start_lock_blocks_another_process,
# which fails identically on a clean upstream/main checkout on this machine (pre-existing local env flake,
# unrelated to this change)

$ ruff check headroom/install/supervisors.py tests/test_install/test_supervisors.py
All checks passed!
$ ruff format --check headroom/install/supervisors.py tests/test_install/test_supervisors.py
2 files already formatted
$ mypy headroom --ignore-missing-imports   # exit 0, notes only
```

## Real Behavior Proof

- Environment: Windows 11 Pro 10.0.26200, Python 3.13, local checkout of this branch.
- Exact command / steps: Tokenized both the old (argv-list → `list2cmdline`) and new (pre-quoted string) command lines with `shell32.CommandLineToArgvW` — the same parsing `sc.exe` applies to its received command line — using the exact path from the issue report. Also ran the new string form through `subprocess.run` against the real `sc.exe` (non-elevated).
- Observed result: Old form tokenizes to `['sc.exe', 'create', 'headroom-default', 'binPath= cmd.exe /c "C:\\Users\\Adron\\...\\run-headroom.cmd"', 'start= auto']` — `binPath=`/`start=` glued to their values, which `sc.exe` rejects with 1639. New form tokenizes to `['sc.exe', 'create', 'headroom-default', 'binPath=', 'cmd.exe /c "C:\\Users\\Adron\\...\\run-headroom.cmd"', 'start=', 'auto']` — exactly the documented `sc create` token shape. Running the new string against real `sc.exe` non-elevated proceeds past argument parsing to `OpenSCManager FAILED 5: Access is denied` (the expected no-admin outcome per the issue reporter's own non-admin run), with no 1639 syntax error.
- Not tested: Full elevated end-to-end `headroom install apply --preset persistent-service` service creation + service start on an Administrator shell (no elevated session available in this environment); behavior on non-English locales other than the tokenization-level verification above.

## Review Readiness

**Follow-up candidate (out of scope here)**: the issue also notes that a failed install removes `~/.headroom/deploy/<profile>/` artifacts, hampering post-mortem debugging — worth a separate issue/PR to preserve or relocate failed-install artifacts.

### DIFF
--- a/headroom/install/supervisors.py
+++ b/headroom/install/supervisors.py
@@ -261,11 +261,16 @@ def install_supervisor(manifest: DeploymentManifest) -> list[ArtifactRecord]:
         return records
 
     if _is_windows() and manifest.supervisor_kind == SupervisorKind.SERVICE.value:
-        service_bin = f'cmd.exe /c "{windows_run_cmd_path(manifest.profile)}"'
-        subprocess.run(
-            ["sc.exe", "create", manifest.service_name, f"binPath= {service_bin}", "start= auto"],
-            check=True,
+        # sc.exe's binPath= value embeds its own quotes (cmd.exe /c "<path>").
+        # Passing this as an argv list lets subprocess.list2cmdline re-quote the
+        # token and sc.exe mis-tokenizes it (issue #1654), so build the exact
+        # command line ourselves and hand subprocess a string.
+        run_cmd = windows_run_cmd_path(manifest.profile)
+        create_cmd = (
+            f"sc.exe create {manifest.service_name} "
+            f'binPath= "cmd.exe /c \\"{run_cmd}\\"" start= auto'
         )
+        subprocess.run(create_cmd, check=True)
         subprocess.run(
             ["sc.exe", "failure", manifest.service_name, "reset= 0", "actions= restart/5000"],
             check=True,
--- a/tests/test_install/test_supervisors.py
+++ b/tests/test_install/test_supervisors.py
@@ -308,13 +308,13 @@ def test_install_supervisor_darwin_windows_and_unsupported(monkeypatch, tmp_path
     win_task = install_supervisor(_manifest(supervisor=SupervisorKind.TASK.value))
     assert win_service[-1].kind == "windows-service"
     assert win_task[-2].path.endswith("-startup")
-    assert [
-        "sc.exe",
-        "create",
-        "headroom-default",
-        'binPath= cmd.exe /c "C:\\tmp\\default\\run-headroom.cmd"',
-        "start= auto",
-    ] in calls
+    # Regression for #1654: the create command must be a single pre-quoted
+    # string (bypassing list2cmdline) with the inner quotes backslash-escaped
+    # and `start= auto` as a separate trailing token.
+    assert (
+        "sc.exe create headroom-default "
+        'binPath= "cmd.exe /c \\"C:\\tmp\\default\\run-headroom.cmd\\"" start= auto'
+    ) in calls
     assert [
         "schtasks",
         "/Create",


JerrettDavis (approved):
Code review looks good. The change is narrowly scoped to the sc.exe create call that embeds nested quotes, keeps the other service-management calls on argv lists, and the updated regression assertion checks the exact tokenization-sensitive command line. CI is green on the PR.

PR #1713, opened 2026-07-02
fix(transforms): pass through ragged tables instead of misaligning columns

### LINKED ISSUE #1652: Headroom proxy on 8787 synthesizes mixed status line not present in canonical output
## Summary

I traced a wrong mixed status line to the local `headroom proxy` listener on `127.0.0.1:8787`.

Observed false line in the session:

```text
rtk 0.42.4 0.43.0 up-to-date
```

Canonical true lines from the underlying checks were:

```text
rtk          0.42.4       0.43.0       update available
rtk          ✓        0.42.4       0.42.4       -        up-to-date
```

The false line combines fields from the Docker row and the WSL row.

## Why this points to headroom specifically

I verified the local port topology:

- `127.0.0.1:8787` is owned by `headroom`
- process command:

```text
/home/cpf/.local/bin/headroom proxy --anthropic-api-url http://localhost:4140 --openai-api-url http://localhost:4140 --port 8787 --code-aware --mode token
```

- `*:4140` is owned by `bun run router/sticky-router.ts`

So `8787` is a Headroom proxy in front of `4140`.

## Exact-path reproduction evidence

The original trigger path was not a generic CLI call. It was a `ctx_batch_execute`-style path that ran two commands and then returned searched/summary output:

- `rtk ./scripts/check-versions.sh`
- `rtk ./scripts/check-wsl-versions.sh`

Later in the same session, the assistant stated the false line above.

To isolate the proxy layer, I changed my OMP model provider mapping so the same provider alias that normally pointed to `http://localhost:8787/v1` instead pointed directly to `http://localhost:4140/v1`, then restarted the session.

After that change, I re-ran the **same path shape**:

- same two commands
- same batch/summary-oriented retrieval pattern

Result when routed directly to `4140`:

- Docker section preserved:
  - `rtk          0.42.4       0.43.0       update available`
- WSL section preserved:
  - `rtk          ✓        0.42.4       0.42.4       -        up-to-date`
- the impossible mixed line did **not** reappear

This strongly suggests the corruption is introduced on the `8787` path, i.e. the Headroom proxy chain, not on the direct `4140` backend.

## Additional notes

I also verified that the raw canonical rows are internally consistent:

- Docker `0.42.4 -> 0.43.0` correctly maps to `update available`
- WSL `0.42.4 -> 0.42.4` correctly maps to `up-to-date`

So the false line is not a legitimate interpretation of either row.

## Expected behavior

A proxy/summarization/compression layer must not synthesize a line that was not present in the underlying output.

It should preserve either:

- the exact original lines, or
- a lossy summary that still does not create impossible mixed facts

## Suspected area

Given the observed topology and the exact-path A/B result, the likely area is one of:

- Headroom proxy compression of repeated table-like text
- transcript/preview generation in proxy mode
- a code-aware or token-mode transformation step that merges nearby rows

## Environment

- `headroom-ai` version: `0.28.0`
- local listener topology:
  - `8787` -> headroom proxy
  - `4140` -> sticky-router backend

If useful, I can provide the exact local command lines and the before/after reroute evidence in a follow-up comment.

Parideboy:
## Description

Issue #1652 reports the proxy's compression layer surfacing an "impossible mixed" status line — a row combining fields from two different rows of a version-status table (Docker row `0.42.4 → 0.43.0 update available` blended with WSL row `0.42.4 → 0.42.4 up-to-date`). The reporter's follow-up refined the claim: the stored canonical content was intact, but the compression path presents a lossier view that invites exactly this misattribution.

There is a concrete mechanism for that in the tabular bridge: `parse_tabular` (`headroom/transforms/tabular_ingest.py`) hands parsed rows to `to_records`, which **silently pads/truncates every row to the header width**. For ragged tables — rows whose cell count differs from the header row, exactly what mixed-shape status tables like the reporter's produce (`✓` and `-` placeholder cells change the token count per row) — this shifts values under the wrong column before SmartCrusher compaction. The compressed output can then state column/value pairings the original never contained.

Fix: `parse_tabular` now rejects ragged tables (any row width ≠ header width) and returns `None`, so the content passes through verbatim, per the issue's requirement that a lossy summary "must not create impossible mixed facts". Aligned tables compress exactly as before. The Rust `log_template` Drain miner was also examined; its template rendering only emits tokens that are constant across all rows of a run, so no defect was found there and it is left untouched.

Fixes #1652

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/tabular_ingest.py`: `parse_tabular` returns `None` when any parsed row's cell count differs from the header count, instead of letting `to_records` pad/truncate rows into the wrong columns. `TabularCompressor.compress` then takes its existing pass-through branch (`was_modified=False`).
- `tests/test_transforms_tabular.py`: three new tests — ragged fixed-width table rejected (reproducing the issue's rtk version-status shape), ragged markdown table rejected, and end-to-end `TabularCompressor.compress` pass-through of a ragged table.

## Testing

### Test Output

```text
$ python -m pytest tests/test_transforms_tabular.py -q
39 passed

$ ruff check headroom/transforms/tabular_ingest.py tests/test_transforms_tabular.py
All checks passed!
$ ruff format --check headroom/transforms/tabular_ingest.py tests/test_transforms_tabular.py
2 files already formatted
Success: no issues found (note-level messages only)
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13, local checkout branched from `upstream/main` (9fbd47ba), Rust core built locally.
- Exact command / steps: constructed the issue's table shape (Docker row with 4 cells, WSL row with 6 cells under 4 headers) and ran it through `TabularCompressor().compress()` before and after the change; ran the full `tests/test_transforms_tabular.py` suite.
- Observed result: before — `to_records` turned the WSL row into `{'tool': 'rtk', 'installed': '✓', 'latest': '0.42.4', 'status': '0.42.4'}`: the `up-to-date` status is dropped and a version number lands under `status` — precisely the misattributed-fact class from the issue. After — `parse_tabular` returns `None`, `compress` returns the original text unmodified (`was_modified=False`, byte-identical pass-through), and all 39 tests pass (36 pre-existing + 3 new).
- Not tested: the reporter's exact end-to-end session (OMP → headroom proxy on 8787 → sticky-router on 4140); the Rust BuildOutput `log_template` path, which was reviewed and found to only emit run-constant tokens.

## Review Readiness

### DIFF
--- a/headroom/transforms/tabular_ingest.py
+++ b/headroom/transforms/tabular_ingest.py
@@ -134,6 +134,13 @@ def parse_tabular(
 
     if not headers or not rows:
         return None
+    # Ragged tables (rows whose cell count differs from the header count)
+    # can't be zipped into records without shifting values under the wrong
+    # column — a compressed table must never state facts the original
+    # didn't (#1652). Treat them as non-tabular and pass through verbatim.
+    width = len(headers)
+    if any(len(row) != width for row in rows):
+        return None
     return headers, rows, fmt
 
 
--- a/tests/test_transforms_tabular.py
+++ b/tests/test_transforms_tabular.py
@@ -153,6 +153,54 @@ def test_parse_markdown_table_drops_separator() -> None:
     assert all("---" not in cell for row in rows for cell in row)
 
 
+def test_parse_tabular_rejects_ragged_fixed_width(monkeypatch) -> None:
+    # Rows with differing cell counts can't be zipped under the headers
+    # without misattributing columns (#1652) — must pass through.
+    import headroom.transforms.tabular_ingest as ti
+
+    monkeypatch.setattr(
+        ti,
+        "detect_content_type",
+        lambda _c: DetectionResult(ContentType.TABULAR, 0.9, {"format": "fixed_width"}),
+    )
+    ragged = (
+        "tool  installed  latest  status\n"
+        "rtk  0.42.4  0.43.0  update available\n"
+        "rtk  ✓  0.42.4  0.42.4  -  up-to-date"
+    )
+    assert ti.parse_tabular(ragged) is None
+
+
+def test_parse_tabular_rejects_ragged_markdown(monkeypatch) -> None:
+    import headroom.transforms.tabular_ingest as ti
+
+    monkeypatch.setattr(
+        ti,
+        "detect_content_type",
+        lambda _c: DetectionResult(ContentType.TABULAR, 0.9, {"format": "markdown"}),
+    )
+    ragged = "| a | b | c |\n| --- | --- | --- |\n| 1 | 2 | 3 |\n| 4 | 5 |"
+    assert ti.parse_tabular(ragged) is None
+
+
+def test_compress_passes_through_ragged_table(monkeypatch) -> None:
+    import headroom.transforms.tabular_ingest as ti
+
+    monkeypatch.setattr(
+        ti,
+        "detect_content_type",
+        lambda _c: DetectionResult(ContentType.TABULAR, 0.9, {"format": "fixed_width"}),
+    )
+    ragged = (
+        "tool  installed  latest  status\n"
+        "rtk  0.42.4  0.43.0  update available\n"
+        "rtk  ✓  0.42.4  0.42.4  -  up-to-date"
+    )
+    result = TabularCompressor().compress(ragged)
+    assert not result.was_modified
+    assert result.compressed == ragged
+
+
 def test_parse_tabular_returns_none_for_non_tabular() -> None:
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Code review looks good. Rejecting ragged parsed tables at the parse_tabular boundary is the right place to prevent to_records from padding/truncating rows into incorrect column/value facts, while preserving the existing compression path for aligned CSV/markdown/fixed-width tables. The added parser and end-to-end pass-through tests cover the regression shape well.

PR #1715, opened 2026-07-02
fix(core): load ONNX Runtime dynamically so headroom._core imports on non-AVX2 x86-64

### LINKED ISSUE #1278: [BUG] SIGILL crash on headroom._core import: prebuilt wheel requires AVX2, incompatible with non-AVX2 x86-64 CPUs
## Description
Running `headroom proxy` results in an `Illegal instruction` (SIGILL) fatal crash on x86-64 CPUs that lack AVX2 support (e.g. Intel Pentium N4200, pre-AVX2 x86-64 architecture). The prebuilt `headroom._core` extension wheel distributed on PyPI appears to have been compiled with AVX2 instructions, making it incompatible with older or low-end x86-64 processors that only support up to SSE4.2.

## To Reproduce
1. Install headroom with `uv tool install --python 3.12 "headroom-ai[proxy]"`
2. Run `headroom proxy`
3. Observe immediate SIGILL crash on import of `headroom._core`

## Expected Behavior
`headroom proxy` should start the optimization proxy server normally on any x86-64 machine, regardless of whether the CPU supports AVX2.

## Actual Behavior
The process crashes immediately with a fatal `Illegal instruction` signal during the import of `headroom._core` in `headroom/transforms/error_detection.py` (line 41). The crash is consistent across both the prebuilt PyPI wheel and a from-source build via `pip install --no-binary headroom-ai`, suggesting the `puccinialin` build wrapper overrides any user-supplied `RUSTFLAGS` (e.g. `-C target-cpu=x86-64`), resulting in an AVX2-compiled binary regardless.

System information:
- CPU:  x86-64, no AVX/AVX2 support (SSE4.2 max)
- OS: Ubuntu/Debian (Linux x86-64)
- Python: 3.12.13 (via uv)
- headroom-ai: 0.26.0

## Code Sample

# Minimal reproduction — crashes immediately on import
import headroom._core  # Fatal Python error: Illegal instruction

Full traceback:

Fatal Python error: Illegal instruction

  File "headroom/transforms/error_detection.py", line 41 in <module>
  File "headroom/transforms/content_router.py", line 54 in <module>
  ...
Extension modules: numpy._core._multiarray_umath, numpy.linalg._umath_linalg, zstandard.backend_c (total: 3)

Verification:

$ cat /proc/cpuinfo | grep flags | grep -o avx2
# (no output — AVX2 not present)

Parideboy:
## Description

`import headroom._core` dies with SIGILL (`Illegal instruction`) on x86-64 CPUs without AVX2 (Pentium N4200, Celeron N4500, AMD FX 8350 — all reported on the issue). The repo sets no `RUSTFLAGS`/`target-cpu` anywhere, so first-party Rust code is baseline x86-64; the AVX2 code comes from Microsoft's prebuilt ONNX Runtime, statically linked into the extension by fastembed's `ort-download-binaries-rustls-tls` feature on non-Windows targets. Because it is statically linked, its code is mapped and initialized when the extension module loads — **before** the runtime AVX2 guard from #1162 can run, which is why that fix helped Magika init but not the import-time crash.

Fix, mirroring what Windows already does for its own reasons (DirectML link libs): build with `ort-load-dynamic` on every platform, so ONNX Runtime is only `dlopen`'d at first use, where the #1162 AVX2 guard falls back to the non-ONNX detection tiers on unsupported CPUs. Since both target blocks became identical, they are collapsed into one platform-independent `fastembed` dependency.

To keep Magika/fastembed working out of the box on Linux/macOS, the existing `ORT_DYLIB_PATH` auto-pin (`headroom/_ort.py`, previously Windows-only) now resolves the pip `onnxruntime` package's shared library on all platforms (`onnxruntime.dll` / `libonnxruntime.so*` / `libonnxruntime*.dylib`). The pip `onnxruntime` CPU wheels use runtime CPU dispatch, so they also work on pre-AVX2 machines — non-AVX2 users get working ML detection instead of a crash. Without the `onnxruntime` package, ML detection degrades gracefully to the non-ONNX tiers exactly as it already does on Windows.

Fixes #1278

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- `crates/headroom-core/Cargo.toml`: replaced the per-target `fastembed` blocks (`ort-download-binaries-rustls-tls` on non-Windows, `ort-load-dynamic` on Windows) with a single platform-independent dependency on `ort-load-dynamic`, with a comment documenting both the DirectML and the AVX2/#1278 rationale.
- `Cargo.lock`: regenerated — `ort-sys` drops its static-download dependencies (`hmac-sha256`, `lzma-rust2`, `ureq`); no version bumps.
- `headroom/_ort.py`: `ORT_DYLIB_PATH` auto-pin extended from Windows-only to all platforms via a small `_find_dylib` helper that resolves the platform's shared-library name inside the pip `onnxruntime` package.
- `tests/test_transforms/test_ort_dylib.py`: replaced the obsolete `test_noop_on_non_windows` with Linux (versioned `.so`) and macOS (`.dylib`) pin tests; module docstring updated.
- `docs/content/docs/configuration.mdx`: `ORT_DYLIB_PATH` row updated from Windows-only wording to the cross-platform behavior.

## Testing

### Test Output

```text
$ cargo fmt --all -- --check && cargo clippy --workspace -- -D warnings && cargo test -p headroom-core --lib
clean
test result: 844 passed; 0 failed; 1 ignored

$ python -m pytest tests/test_transforms/test_ort_dylib.py -q
8 passed

$ ruff check headroom/_ort.py tests/test_transforms/test_ort_dylib.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11 (AVX2-capable — the SIGILL itself is not reproducible on this machine), Python 3.13, Rust 1.95.0, local checkout branched from `upstream/main` (9fbd47ba).
- Exact command / steps: `cargo check -p headroom-core` after the feature switch; inspected the `Cargo.lock` diff; rebuilt and ran `python -c "import headroom; from headroom._core import detect_content_type; print(detect_content_type('hello world'))"`; ran the ort-pin test suite with monkeypatched `linux`/`darwin` platforms.
- Observed result: build succeeds with `ort-load-dynamic`; the lockfile shows `ort-sys` no longer pulls the binary-download machinery (`hmac-sha256`, `lzma-rust2`, `ureq` removed), confirming the statically-linked prebuilt ORT is gone; import + content detection works with `ORT_DYLIB_PATH` auto-pinned to the pip onnxruntime library; all 8 pin tests pass including the new Linux/macOS branches.
- Not tested: actual pre-AVX2 x86-64 hardware (none available — the fix removes AVX2 code from the import path by construction, and the issue reporters on #1278 can verify); Linux/macOS wheel runtime behavior beyond CI's ubuntu/macOS wheel-build jobs; embedding quality/performance under a pip-provided ORT version differing from the previously vendored one.

## Review Readiness


JerrettDavis (approved):
Code review looks good. The change removes the static ONNX Runtime download path from the Rust extension, keeps the Windows dynamic-load behavior, and extends the import-time ORT_DYLIB_PATH pinning in a way that is covered by Linux/macOS/Windows-oriented tests. I did not find a correctness blocker in the diff.

One process note before merge: the current check rollup is not fully green because several test jobs are marked cancelled rather than successful. I am treating that as CI state, not a requested code change.

JerrettDavis (approved):
Loading ORT dynamically on all platforms is a reasonable fix for pre-AVX2 import crashes, and extending ORT_DYLIB_PATH auto-pinning to Linux/macOS restores the pip-package ORT path when available while leaving explicit overrides alone. Branch needs conflict resolution before merge, but I do not see a blocker in the approach.

JerrettDavis (approved):
Re-reviewed current head after the main merge and CI fixes. The dynamic ONNX Runtime loading path remains scoped to import/link behavior, and the release workflow assertions cover the intended wheel/env propagation.

- `python -m pytest tests/test_transforms/test_ort_dylib.py tests/test_release_workflows.py -q` -> 39 passed
- `uvx ruff==0.15.17 check headroom/_ort.py tests/test_transforms/test_ort_dylib.py tests/test_release_workflows.py`
- `uvx ruff==0.15.17 format --check headroom/_ort.py tests/test_transforms/test_ort_dylib.py tests/test_release_workflows.py`

PR #1717, opened 2026-07-02
fix(transforms/content-router): detect on inner tool-output payload

chopratejas:
## Description

Coding-agent harnesses wrap each tool result in an envelope such as
`<returncode>0</returncode>\n<output>…</output>` (also `<stdout>`, `<stderr>`,
`<tool_result>`, `<result>`). The native content detector read those wrapper
tags as markup and classified the whole payload as HTML/XML — so source code,
grep results, and logs were misrouted to the HTML article-extractor, which
blanks or corrupts them (dropping identifiers and route converters).

This routes **detection** on the unwrapped inner payload so the real content
type wins. **Compression still runs on the original content**, so the envelope
tags (exit code, stream separation) are preserved — no information is lost.

Also threads per-compressor config overrides through `ContentRouterConfig` via
`dataclasses.replace`, so the proxy can tune each structural compressor while
`ContentRouter` keeps enforcing global safety flags (`ccr_inject_marker`,
search grouping).

Closes #

## Type of Change

- [x] Code refactoring (config-override plumbing; no change to default behavior)

## Changes Made

- `_strip_detection_envelope()` + `_DETECTION_ENVELOPE_RE`: unwrap a whole-string
  tool-output envelope for detection only. Fires only when the entire string is a
  single wrapper; never returns an empty probe (falls back to the original).
- `_detect_content()` now detects on the unwrapped payload.
- `ContentRouterConfig` gains `search_compressor` / `log_compressor` /
  `diff_compressor` / `text_crusher` override fields (default `None` → each
  compressor's own defaults). The four `_get_*` getters start from the override
  (or default) and `replace()` in the ContentRouter-enforced flags.
- Regression tests for both behaviors.

## Testing

- [x] Unit tests pass (targeted suites below)

### Test Output

```text
$ pytest tests/test_transforms_content_router.py tests/test_transforms_content_detection.py -q
45 passed in 0.50s

$ ruff check headroom/transforms/content_router.py tests/test_transforms_content_router.py
All checks passed!

```

## Real Behavior Proof

- Environment: local, Python 3.12.6, native `headroom._core` detect backend.
- Exact command / steps: `_detect_content("<returncode>0</returncode>\n<output>\n<python source>\n</output>")`
- Observed result: detects `ContentType.SOURCE_CODE` (identical to the same code
  unwrapped). Before this change the wrapper tags made it detect as HTML.
- Also measured that the search/log/diff compressors already tolerate the
  envelope (≤1% ratio delta wrapped vs bare), so compression is left on the
  original content and the tags are preserved rather than stripped.
- Not tested: end-to-end proxy request replay; the config-override fields are
  plumbing only (no proxy wiring in this PR).

## Review Readiness

## Additional Notes

- The four config-override fields are wiring only; the proxy is not yet passing
  overrides through them (follow-up).
- Pre-existing mypy findings in `tests/test_transforms_content_router.py`
  (FakeTokenizer typing, untyped helpers) are unrelated to this change and left
  as-is; `mypy headroom` is clean and the two added tests are fully typed.

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -45,7 +45,7 @@ import sys
 import threading
 import time
 from concurrent.futures import ThreadPoolExecutor
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from enum import Enum
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


DevanshiVyas (commented):
LGTM

PR #1735, opened 2026-07-03
fix(release): build and publish Windows x86_64 wheels

### LINKED ISSUE #1144: [Windows] PyPI wheel missing `_core` Rust extension — compress() silently no-ops; CLI crashes on --help
## Environment
- **OS:** Windows 11 x86_64
- **Python:** 3.12.10
- **headroom-ai:** 0.20.15 (latest pre-built binary wheel: `pip install headroom-ai --only-binary :all:`)

---

### Steps to reproduce
```
pip install headroom-ai --only-binary :all:   # 0.20.15 on Windows/Python 3.12
python -c "
import headroom
r = headroom.compress([{'role':'user','content':'ERROR: timeout\n'*80}])
print(r.tokens_saved, r.transforms_applied)
"
```

### Actual output
```
Compression failed, returning original messages: No module named 'headroom._core'
0 []
```

### Expected
Non-zero `tokens_saved`, at least one entry in `transforms_applied`.

### Root cause
The compiled PyO3 extension (`headroom._core`) is **not present** in the published wheel:
```python
glob('.../headroom/**/*.pyd', recursive=True)  # → []
glob('.../headroom/**/*.so',  recursive=True)  # → []
```
`SmartCrusher`, `DiffCompressor`, and `ContentRouter` all `from headroom._core import …` as their first line. Without it every compressor raises `ModuleNotFoundError`. The top-level `compress()` catches this, prints one line to stderr, and **returns the original messages unchanged** with `compression_ratio=0.0` — a silent no-op callers won't notice unless they check `transforms_applied`.

### Additional: versions 0.21.0–0.26.0 have no Windows wheels at all
`pip install "headroom-ai==0.26.0" --only-binary :all:` → `ERROR: No matching distribution found`. Windows users cannot install any release newer than 0.20.15 without a full Rust + MSVC toolchain.

---

### Steps to reproduce
```
pip install headroom-ai --only-binary :all:   # core only, no [proxy]
headroom --help
```

### Actual output (full traceback)
```
File "headroom/cli/__init__.py", line 15:
    from . import (
File "headroom/cli/proxy.py", line 10:
    from headroom.proxy.modes import PROXY_MODE_TOKEN, normalize_proxy_mode
File "headroom/proxy/server.py", line 93:
    from headroom.providers.proxy_routes import register_provider_routes
File "headroom/providers/proxy_routes.py", line 9:
    from fastapi import FastAPI, Request, WebSocket
ModuleNotFoundError: No module named 'fastapi'
```

### Root cause
`headroom/cli/__init__.py` unconditionally imports the proxy submodule chain at startup, which ultimately imports `fastapi` at module load time. Since `fastapi` is in the optional `[proxy]` extra, any user who does `pip install headroom-ai` (core only) gets a broken CLI — `--help`, `--version`, and all non-proxy subcommands crash immediately.

### Fix suggestion
```python
# headroom/cli/__init__.py
try:
    from . import proxy_cmds  # noqa: F401
except ImportError:
    pass  # fastapi not installed; proxy commands unavailable
```

---

## Summary

| # | Issue | Impact | Severity |
|---|---|---|---|
| 1 | `headroom._core` absent from all PyPI wheels | `compress()` silently no-ops on every Windows install | **Critical** |
| 2 | No Windows wheels for 0.21.0–0.26.0 | Windows users stuck on broken 0.20.15 | **High** |
| 3 | CLI unconditionally imports `fastapi` at startup | `headroom --help` crashes without `[proxy]` extra | **Medium** |

ashishpatel26:
## Summary
- `release.yml`'s wheel matrix never had a Windows target — only `linux x86_64/arm64` and `macOS arm64` wheels are built. No Windows wheel has ever shipped, so `pip install headroom-ai` on Windows falls back to source (or fails outright), leaving `headroom._core` missing and every compressor silently no-op'ing (`compress()` catches `ModuleNotFoundError` and returns input unchanged).
- Rust side is already Windows-ready: dynamic ORT loading via `ort-load-dynamic` on `cfg(windows)` landed for the AVX2 SIGILL fix (#1715), and `ort-sys` ships prebuilt ONNX Runtime binaries for `x86_64-pc-windows-msvc` — no CMake/source build needed the way `x86_64-apple-darwin` requires.
- Adds a `windows-2022` / `x86_64-pc-windows-msvc` row to `build-wheels`, plus a matching `smoke-import-wheels` entry (PowerShell) that installs the built wheel into a clean venv and imports `headroom._core`, gating publish the same way the Linux/macOS rows do.

Fixes #1144 (Bug 1 — missing Windows wheel).

Bug 2 in that issue (`headroom --help` crashing with `ModuleNotFoundError: fastapi` on a core-only install) is already fixed on `main`: `headroom.proxy.server` is lazily imported inside the `proxy` command, not at CLI module load time. Verified locally by invoking `headroom --help` with `fastapi` import blocked via a meta-path finder — exits 0.

## Test plan
- [x] `tests/test_release_workflows.py` — 28 passed, 1 pre-existing unrelated failure (`test_no_native_tls_in_wheel_build_tree`, reproduces on clean `main` too — local `cargo tree` environment issue)
- [x] `.github/workflows/release.yml` parses as valid YAML
- [x] Verified `headroom --help` doesn't require `fastapi` on current `main` (Bug 2 already fixed)
- [ ] CI run of the new `build-wheels` / `smoke-import-wheels` Windows matrix rows on this PR (release.yml triggers a dry-run on PRs touching `crates/headroom-py/**`... this PR only touches `release.yml` itself, which is also in the path filter, so the dry-run should fire)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -259,6 +259,15 @@ jobs:
           - os: macos-14
             target: aarch64-apple-darwin
             manylinux: ""
+          # Windows x86_64 (MSVC). Unlike the macOS Intel case above,
+          # `ort-sys` DOES ship prebuilt ONNX Runtime binaries for
+          # `x86_64-pc-windows-msvc`, so no source build / CMake needed
+          # here. Fixes #1144: no Windows wheel has ever been published,
+          # so `headroom._core` is silently missing and every compressor
+          # falls back to a no-op on Windows installs.
+          - os: windows-2022
+            target: x86_64-pc-windows-msvc
+            manylinux: ""
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v6
@@ -499,6 +508,9 @@ jobs:
           # ship today (Intel macOS dropped in PR #371 — ort-sys
           # has no x86_64-apple-darwin prebuilts).
           - { runner: macos-14, image: "", python: "3.13", wheel_target: aarch64-apple-darwin, wheel_artifact: wheels-macos-14-aarch64-apple-darwin, glibc_label: "" }
+          # Windows x86_64 — runs natively on the host runner; no
+          # container, no glibc floor. See #1144.
+          - { runner: windows-2022, image: "", python: "3.12", wheel_target: x86_64-pc-windows-msvc, wheel_artifact: wheels-windows-2022-x86_64-pc-windows-msvc, glibc_label: "" }
 
     steps:
       - name: Download wheels artifact for this target
@@ -525,6 +537,7 @@ jobs:
         #    YAML `run: |` level (uniform strip), the docker container
         #    sees it as a read-only mount, and macOS host runs the
         #    same script — no quoting drift between paths.
+        shell: bash
         run: |
           cat > "${RUNNER_TEMP}/smoke_import.py" <<'PY'
           import sys
@@ -640,7 +653,7 @@ jobs:
             '
 
       - name: Smoke-import wheel on macOS host
-        if: matrix.image == ''
+        if: matrix.image == '' && runner.os == 'macOS'
         env:
           PYTHON_VERSION: ${{ matrix.python }}
         run: |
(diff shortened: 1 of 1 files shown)


ashishpatel26:
Closing — duplicate. Upstream already fixed this in #1335 (merged 2026-06-24), which added a more thorough windows-latest wheel row (explicit actions/setup-python, better abi3 fallback matching) than what I proposed here. Issue #1144 (Bug 1) is resolved by that merge; just needs the issue closed/labeled stale.

PR #1736, opened 2026-07-03
test(codex-ws): de-flake semaphore-tail test (widen noise floor 25ms->75ms)

chopratejas:
`test_concurrent_compression_has_no_semaphore_tail` failed on main CI (shard 1) with p50=1ms, p99=29ms, ratio=22x. It's flaky, not a regression:

- `p99` is the max of only ~12 concurrent samples, so one CI scheduler/GC outlier dominates it.
- The real semaphore-contention signal is **p99 ~2433ms** (success criterion <250ms). 29ms is 84x below that.
- Passes 3/3 locally.

Widen the noise floor 25ms -> 75ms: absorbs CI jitter, still 3.3x below the 250ms regression threshold (a genuine hundreds-of-ms tail still trips it). Comment updated. No production code change.

Does not block the v0.29.0 publish (release.yml runs no pytest), but greens main CI.


JerrettDavis (approved):
Reviewed the flake adjustment. The updated 75ms floor is still materially below the 250ms regression threshold and far below the historical semaphore-tail signal, while accounting for the small-sample max/p99 behavior on shared CI runners. Test-only change; this looks ready.

JerrettDavis (approved):
Re-reviewed current head. The remaining effective diff against main is empty, so there is no new behavioral surface left to review after the branch update.

PR #1612, opened 2026-06-30
perf(proxy): offload image compression off event loop

inix-x:
## Description

Image compression ran synchronously on the asyncio event loop in the Anthropic and OpenAI handlers. The CPU-bound ONNX technique routing + Pillow resize + OCR froze the loop for the entire compression, stalling every other in-flight request. This offloads it onto the bounded compression executor, the same idiom the text-compression path already uses, and fails open so the executor's timeout can't turn a slow compression into a 500.

No linked issue — perf fix. Mirrors the gemini "run compression off the asyncio event loop" change already in the CHANGELOG, and the precedent offloads #718 / #1382 / #1501.

## Type of Change

- [x] Performance improvement

## Changes Made

- `headroom/proxy/handlers/anthropic.py` + `headroom/proxy/handlers/openai.py`: route `ImageCompressor.compress()` through `self._run_compression_in_executor(lambda: ..., timeout=COMPRESSION_TIMEOUT_SECONDS)` instead of calling it inline on the loop. `_get_image_compressor()` builds a fresh per-request compressor and the model loads lazily inside `compress()`, so offloading `compress()` moves all the heavy work and introduces no shared-state race.
- Fail open on timeout/error (log + forward the original messages), mirroring the text path (`anthropic.py` `except` around the pipeline) so the now-mandatory executor timeout can't 500 a slow-but-fine request.
- `tests/test_image_compression_offload.py`: asserts both blocks are async + offloaded + fail-open, that `compress()` runs on a `headroom-compress` worker thread, and that the loop stays responsive during a slow compression (mirrors `test_gemini_compression_offload.py`).
- `CHANGELOG.md`: Unreleased → Bug Fixes entry.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ ruff check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py
All checks passed!

$ ruff format --check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py
2 files already formatted

$ mypy headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py
(exit code 0)

$ pytest tests/test_image_compression_offload.py
tests/test_image_compression_offload.py::test_image_blocks_offload_compress_and_fail_open PASSED
[...]

Issue #956, opened 2026-06-14
[BUG] compression_first_stage always times out (30s) on Python 3.14 + headroom-ai 0.25.0 incompatible with Python 3.14 (litellm dep)


aditya-nawal:
**Environment**
- headroom-ai: 0.24.0
- Python: 3.14 (Windows 11)
- Claude Code via ANTHROPIC_BASE_URL proxy
- Install method: pip install --no-build-isolation (required to bypass puccinialin cache issue)

**Bug 1: compression_first_stage always times out**
Every single API request shows:
  compression_first_stage: ~30,010ms (always exactly the timeout)
  Optimization failed: TimeoutError: (no detail)
  transforms_applied: []
  tokens_saved: 0

Stage timings from proxy.log:
  {"compression_first_stage": 30010.916, "total_pre_upstream": 30021.483}

This happens on ALL requests regardless of size (tested with 895KB payloads).
MCP compress returns `transforms: ["router:noop"]` — also not compressing.

Suspect: PyO3 Rust extension is blocking the asyncio event loop on Python 3.14
due to async/GIL changes. PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 was used for install.

**Bug 2: headroom-ai 0.25.0 cannot be installed on Python 3.14**
litellm>=1.86.2 (required by 0.25.0) has Requires-Python >=3.10,<3.14
ERROR: Could not find a version that satisfies the requirement litellm<2.0,>=1.86.2

**What works:**
- Proxy routing: ✓ (all API calls proxied)
- Prefix cache alignment: ✓ ($171 savings observed)
- MCP server: ✓ (connected, headroom_stats returns data)

dan-and:
I would rather like an updates version of the litellm dependency and PyO3 -> v0.22.6 supported max Python 3.13, so PyO3 needs a version bump as well.

PR #1738, opened 2026-07-03
fix(proxy): bound HF tokenizer load and offload token counting off event loop

### LINKED ISSUE #1701: [BUG]
Bug: headroom proxy becomes unresponsive after processing one request on Windows 11

  Environment:
  - headroom-ai version: 0.28.0 (installed from PyPI wheel)
  - Python version: 3.11.3
  - OS: Windows 11 (build 26200)
  - Install method: pip install headroom-ai[all]

  Setup:
  Using Headroom as a compression proxy (not MCP), with the following launch command:

  headroom proxy --port 8787 --anthropic-api-url https://api.deepseek.com/anthropic --host 127.0.0.1

  The proxy sits between CC Switch (a third-party proxy manager on port 15721) and the upstream provider (DeepSeek's Anthropic-compatible API):

  Claude Code → CC Switch (:15721) → Headroom (:8787) → api.deepseek.com/anthropic

  Workaround applied: HEADROOM_DETECT_BACKEND=python (set via environment variable before launch, per #713/#845).

  Symptoms:

  1. Process starts fine — /livez and /readyz return HTTP 200, upstream status shows "healthy".
  2. First request takes ~10 minutes — then succeeds with minimal compression:
  input_tokens_original: 81864 → input_tokens_optimized: 77833
  tokens_saved: 4031 (4.9%)
  optimization_latency_ms: 609972 (~610 seconds)
  3. No actual compression transforms applied — all 76 transforms are either router:protected:*, router:excluded:tool, or read_lifecycle:stale:*. No router:text, router:smart_crusher, or any actual compression
  transform appears.
  4. Process becomes a zombie after that request — PID still alive, port 8787 still listening, but all HTTP requests (/livez, /readyz, /health, /stats) hang indefinitely until timeout. The process stays in this
  state until killed.
  5. Restart → same pattern repeats — works for one request (~610s), then zombie.

  Diagnostic output of /readyz on fresh start:
  {"service":"headroom-proxy","status":"healthy","ready":true,"version":"0.28.0","checks":{"upstream":{"enabled":true,"ready":true,"status":"healthy","url":"https://api.deepseek.com/anthropic","error":null}}}

  After the zombie state: /readyz times out with no response.

  What I've tried:
  - HEADROOM_DETECT_BACKEND=python — prevents the first-call deadlock from #713/#845, but doesn't prevent the zombie-after-one-request problem
  - --no-cache — no difference
  - --connect-timeout-seconds 15 --request-timeout-seconds 120 — no difference
  - Starting via VBScript (WScript.Shell.Run) to avoid Windows Job Object cleanup — process survives, but still goes zombie after one request
  - CREATE_BREAKAWAY_FROM_JOB / DETACHED_PROCESS flags — same result

  Questions:
  1. Why does compression take 610 seconds with only 4.9% savings and no actual compression transform?
  2. What causes the proxy to become unresponsive after processing one request? This looks like an asyncio event loop stall or worker thread exhaustion.
  3. Is there a way to enable verbose logging to pinpoint where the hang occurs?

Parideboy:
## Description

Fixes #1701.

On Windows, `headroom proxy --anthropic-api-url https://api.deepseek.com/anthropic` froze: the first `/v1/messages` request took ~610s (`optimization_latency_ms=609972`) with only router/lifecycle markers, and afterwards the whole server was a zombie — `/livez`, `/readyz` and `/health` hung until the process was killed. `HEADROOM_DETECT_BACKEND=python` was already set, so this was not the #575/#845 native-detect deadlock.

Root cause: DeepSeek model names route to the HuggingFace tokenizer backend (`MODEL_PATTERNS` in `headroom/tokenizers/registry.py`). `HuggingFaceTokenizer` loads lazily, so the registry's construction-time fallback never fires; the first `count_messages` calls `AutoTokenizer.from_pretrained(..., trust_remote_code=True)` — unbounded network downloads/retries — and this ran **synchronously inside the async Anthropic messages handler** (`get_tokenizer(model)` + `tokenizer.count_messages(messages)`), outside the 30s `_run_compression_in_executor` bound. huggingface_hub retry chains on a restricted network easily reach ~10 minutes, blocking the entire asyncio event loop; subsequent on-loop counting kept it pinned. tiktoken got a bounded eager load for the same bug class long ago (#956); the HF backend never did.

## Type of Change

- [ ] Documentation update

## Changes Made

- `headroom/tokenizers/huggingface.py`: `_load_tokenizer` now tries the local HF cache first (`local_files_only=True`, no network), then bounds the network load with `HEADROOM_HF_TOKENIZER_LOAD_TIMEOUT_SECS` (default 10s; `0` disables network loads) on a daemon thread. Timeouts/failures return `None` (cached by `lru_cache`, so the hub is probed at most once per process per tokenizer) and `count_messages` fails open to char-based estimation via the existing `_use_fallback()` path.
- `headroom/proxy/handlers/anthropic.py`: new `AnthropicHandlerMixin._count_tokens_offloaded(model, messages)` runs `get_tokenizer` + `count_messages` on the compression executor bounded by `COMPRESSION_TIMEOUT_SECONDS`, failing open to `EstimatingTokenCounter` (downgrade logged once per model). Used in `handle_anthropic_messages` (the issue's hot path, both count sites) and `handle_anthropic_batch_create`; the batch path's inline `anthropic_pipeline.apply()` is now offloaded via `_run_compression_in_executor` (mirrors the #1612 image-compression offload).
- `headroom/proxy/handlers/batch.py`: the two remaining inline `openai_pipeline.apply()` calls (`handle_google_batch_create`, `_compress_batch_jsonl`) are offloaded the same way; existing `except` blocks keep the pass-through fail-open semantics.
- Tests: `tests/test_huggingface_tokenizer_timeout.py` (cache-first, bounded timeout, failure caching, timeout=0, fail-open estimation), `tests/test_tokenizer_count_offload.py` (wiring guards, runs on `headroom-compress` worker, event loop stays responsive during slow tokenizer work, fail-open), plus `_run_compression_in_executor` stub on the batch test double.

## Testing

- [x] All existing tests pass
- [x] Added new tests for the changes

```
$ python -m pytest tests/test_huggingface_tokenizer_timeout.py tests/test_tokenizer_count_offload.py tests/test_image_compression_offload.py tests/test_gemini_compression_offload.py tests/test_tokenizers tests/test_proxy_handlers_batch.py -q
50 passed

$ ruff check .          # No issues found
$ ruff format --check . # 1043 files already formatted
$ mypy headroom --ignore-missing-imports  # 0 errors
```

## Real Behavior Proof

- Environment: Windows 11 Pro (10.0.26200), Python 3.13, local checkout of this branch with the Rust core built.
- Exact command / steps: `python -m pytest tests/test_tokenizer_count_offload.py -q` — includes `test_count_tokens_offloaded_keeps_loop_responsive`, which reproduces the issue's mechanism: a tokenizer whose `count_messages` blocks (stand-in for the unbounded `AutoTokenizer.from_pretrained` network load) while an asyncio ticker measures event-loop liveness. Also `python -m pytest tests/test_huggingface_tokenizer_timeout.py -q` with a `from_pretrained` stub that sleeps 60s and `HEADROOM_HF_TOKENIZER_LOAD_TIMEOUT_SECS=0.2`.
- Observed result: with the fix, the slow count runs on a `headroom-compress` worker thread and the loop keeps ticking (`ticks >= 5`; inline it yields ~0 — the zombie). The 60s-hung HF load unblocks at the 0.2s timeout, falls back to estimation, and the second call returns instantly (failure cached, no re-probe). All 10 new tests pass.
- Not tested: live reproduction against `api.deepseek.com` from a network where HF hub downloads stall (the reporter's exact environment); actual HF vocab download timing on a healthy network.

## Review Readiness

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -41,6 +41,48 @@ logger = logging.getLogger("headroom.proxy")
 class AnthropicHandlerMixin:
     """Mixin providing Anthropic API handler methods for HeadroomProxy."""
 
+    async def _count_tokens_offloaded(self, model, messages):  # noqa: ANN001, ANN201
+        """Resolve a tokenizer and count messages off the event loop.
+
+        Tokenizer resolution can be expensive on first use (HuggingFace
+        backends may download vocab files) and counting a full Claude Code
+        conversation is CPU-bound, so both run on the compression executor
+        bounded by ``COMPRESSION_TIMEOUT_SECONDS`` (GH #1701: an unbounded
+        on-loop load froze the whole server). On timeout or error this
+        fails open to character-based estimation.
+
+        Returns:
+            Tuple of ``(tokenizer, token_count)``. The tokenizer is fully
+            initialized, so later ``count_messages`` calls on it are pure
+            CPU work.
+        """
+        from headroom.proxy.helpers import COMPRESSION_TIMEOUT_SECONDS
+        from headroom.tokenizers import EstimatingTokenCounter, get_tokenizer
+
+        def _resolve_and_count():  # noqa: ANN202
+            tokenizer = get_tokenizer(model)
+            return tokenizer, tokenizer.count_messages(messages)
+
+        try:
+            return await self._run_compression_in_executor(
+                _resolve_and_count,
+                timeout=float(COMPRESSION_TIMEOUT_SECONDS),
+            )
+        except Exception as e:  # fail open — includes asyncio.TimeoutError
+            # Log the downgrade once per model, not per request.
+            fallback_models = getattr(self, "_token_count_fallback_models", None)
+            if fallback_models is None:
+                fallback_models = set()
+                self._token_count_fallback_models = fallback_models
+            if model not in fallback_models:
+                fallback_models.add(model)
+                logger.warning(
+                    f"Token counting for model {model} failed or timed out "
+                    f"({e.__class__.__name__}); falling back to estimation"
+                )
+            estimator = EstimatingTokenCounter()
+            return estimator, estimator.count_messages(messages)
+
     @staticmethod
     def _resolve_ccr_workspace(
         request: Any,
@@ -469,7 +511,6 @@ class AnthropicHandlerMixin:
             read_request_json_with_bytes,
         )
         from headroom.proxy.modes import is_cache_mode, is_token_mode
-        from headroom.tokenizers import get_tokenizer
         from headroom.utils import extract_user_query
 
         start_time = time.time()
@@ -899,9 +940,10 @@ class AnthropicHandlerMixin:
                         media_type="application/json",
                     )
 
-            # Count original tokens
-            tokenizer = get_tokenizer(model)
-            original_tokens = tokenizer.count_messages(messages)
+            # Count original tokens off the event loop: first-use tokenizer
+            # resolution may hit the network (HF download) and counting a full
+            # conversation is CPU-bound — on-loop it froze the server (#1701).
+            tokenizer, original_tokens = await self._count_tokens_offloaded(model, messages)
 
             # Enterprise Security: scan request before compression
             _security_ctx = None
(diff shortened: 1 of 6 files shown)


JerrettDavis (approved):
Reviewed the HF tokenizer timeout and offload changes. The tokenizer layer now tries local cache first, bounds the network path, caches failures, and fails open to estimation; the Anthropic and batch handlers move token counting/pipeline work onto the existing compression executor with timeout-based fallback. The tests cover cache-first behavior, timeout/failure caching, worker-thread execution, loop responsiveness, and batch handler wiring. Current checks are green; this looks ready.

JerrettDavis (approved):
The tokenizer path is now bounded in both places that mattered: HF first-load tries local cache first and times out the network path, and proxy token counting/pipeline work runs off the event loop with fail-open estimation. The tests cover timeout, cached failure, loop responsiveness, and batch offload wiring. No blocker.

PR #1740, opened 2026-07-03
fix(router): stop protecting passing build/test output as error traces

### LINKED ISSUE #1696: [BUG] [PROXY] Long multi-turns coding task: only 0.3% compression
## Description

Using Hedaroom proxy in a long, multi-turn agentic coding task in a big JavaScript code repository, the token savings are quite lower than advertised: I get 0.3% vs. expected 60%-95%

## To Reproduce

Full installation, 

```bash
pip install "headroom-ai[all]" --break-system-packages 
pip install --upgrade litellm --break-system-packages

``` 

Running proxy with command

```bash
headroom proxy --code-aware --openai-api-url http://localhost:13305/api
```

Image

Using the endpoint with KiloCode, I get 0.3% token savings in a long coding session

Image

## Environment

- **Headroom version**: 0.28.0
- **Python version**: 3.14.4
- **OS**: Ubuntu 26.04
- **LLM Provider**: OpenAI compatible Lemonade Server

## Additional Context

Image

Dump of stats endpoint: [stats.json](https://github.com/user-attachments/files/29596164/stats.json)

ashishpatel26:
## Description

Error output protection false-positive fix

`content_has_strong_error_indicators()` (`headroom/transforms/error_detection.py`) protects any message/content-block from compression when it contains 2+ distinct indicator keywords (`error`, `fail`, `exception`, `traceback`, `fatal`, `panic`, `crash`). That heuristic false-positives on **passing** build/test tool output: `tsc`'s `"Found 0 errors"` plus a passing test run's `"0 failures"` trips both `error` and `fail` — 2 distinct hits — despite nothing failing. In a long JS/TS coding session this fired on nearly every request (confirmed against the `stats.json` attached to #1696), permanently protecting legitimate tool output from ever being compressed and explaining the reported 0.3% savings vs. the advertised 60-95%.

Closes #1696

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/error_detection.py`: strip common zero-result phrases before the 2-keyword scan in `content_has_strong_error_indicators()` — both `"N word"`/`"word N"` forms (`"0 errors"`, `"no failures"`) and `"label:value"`/`"label=value"` forms (`"Failures: 0"`, `"errors=0"`), covering `error(s)` and `fail`/`failed`/`failing`/`failure(s)` (the scan matches `fail` by substring, so all inflections needed covering).
- `tests/test_error_detection.py` (new file — no prior coverage existed): 7 tests covering real error/traceback detection, single-keyword safety, `tsc`/`eslint` passing summaries, the `"0 failed"` regression a reviewer caught, label:value formats, and that a genuine second indicator elsewhere in the same blob still triggers protection.
- `.github/pr-images/issue-1696-error-protection-fix.svg`: diagram explaining the mechanism (embedded above).

## Testing

- [ ] Linting passes (`ruff check .`) — not run locally, CI `lint` check is green
- [ ] Type checking passes (`mypy headroom`) — not run locally, CI is green

### Test Output

```text
$ .venv/Scripts/python -m pytest tests/test_error_detection.py tests/test_transforms/test_content_router.py tests/test_transforms_content_router.py -q
tests\test_error_detection.py .......                                    [  7%]
tests\test_transforms\test_content_router.py ........................... [ 34%]
...........................                                              [ 62%]
tests\test_transforms_content_router.py ................................ [ 94%]
.....                                                                    [100%]
============================= 98 passed in 1.21s ==============================
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.14.5, local venv, `headroom._core` built via `maturin develop --release` (not prebuilt in a fresh checkout)
- Exact command / steps: ran the reporter's exact scenario patterns (`"Found 0 errors\nTests: 0 failures, 42 passed"`, eslint's `"0 problems (0 errors, 0 warnings)"`) through `content_has_strong_error_indicators()` directly, before and after the fix
- Observed result: before → `True` (wrongly protected); after → `False` (correctly compressible). Real failure text (`Traceback... fatal error`) still returns `True` after the fix.
- Not tested: have not reproduced the full KiloCode/proxy session end-to-end locally (no access to the reporter's actual traffic) — root cause was confirmed via the `stats.json` they attached to the issue, which shows `router:protected:error_output` firing on nearly every request in their session.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A — internal heuristic, no user-facing docs reference it)

See the diagram embedded in Description above.

## Additional Notes

**Investigation trail**: ruled out `protect_recent_reads_fraction` (proxy already overrides its `0.0` dataclass default to `0.3` in token mode, the default) before confirming the error-protection false-positive via the reporter's `stats.json`.

**Response to review comments** (@AbelVM, @sparkbugz): prose mentions like `"Fix the errors in the code."` or `"console.error(...)"` contain only 1 distinct indicator keyword and were already safe under the pre-existing 2-keyword threshold — not something this PR changes. The broader concern about other CI summary formats is addressed above (label:value forms). A case like genuine prose that happens to mention *two* distinct keywords together (e.g. "there are errors and it failed") is a known limitation of a keyword-substring heuristic in general, predates this PR, and is out of scope here — downstream compressors (LogCompressor) still preserve real error lines even when a block isn't gate-protected, so the failure mode there is "slightly stricter than ideal," not data loss.

**Response to @JerrettDavis's CHANGES_REQUESTED**: fixed in the follow-up commit — `"0 failed"` is now stripped (previously only `failing`/`failure(s)` were), with a regression test for the exact reproduction given.

### DIFF
--- a/.github/pr-images/issue-1696-error-protection-fix.svg
+++ b/.github/pr-images/issue-1696-error-protection-fix.svg
@@ -1,0 +1,78 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1200 900" font-family="Segoe UI, Helvetica, Arial, sans-serif">
+  <rect x="0" y="0" width="1200" height="900" fill="#ffffff"/>
+  <text x="600" y="36" text-anchor="middle" font-size="22" font-weight="700" fill="#111827">Headroom: Error-Output Protection False-Positive Fix (#1696)</text>
+
+  <!-- Box 1: pipeline -->
+  <g>
+    <rect x="20" y="60" width="560" height="190" rx="10" fill="#eff6ff" stroke="#2563eb"/>
+    <text x="36" y="88" font-size="15" font-weight="700" fill="#1d4ed8">1. THE GATE</text>
+    <text x="36" y="112" font-size="13" fill="#111827">content_router.py routes every message/content-block through</text>
+    <text x="36" y="130" font-size="13" fill="#111827">content_has_strong_error_indicators() before deciding whether</text>
+    <text x="36" y="148" font-size="13" fill="#111827">to compress it or protect it (pass through untouched).</text>
+    <text x="36" y="176" font-size="13" fill="#111827">Rule: if the text contains 2+ DISTINCT keywords from</text>
+    <text x="36" y="194" font-size="13" font-family="Consolas, monospace" fill="#b91c1c">[error, fail, exception, traceback, fatal, panic, crash]</text>
+    <text x="36" y="218" font-size="13" fill="#111827">the block is protected — assumed to be a real failure trace.</text>
+  </g>
+
+  <!-- Box 2: the bug -->
+  <g>
+    <rect x="620" y="60" width="560" height="190" rx="10" fill="#fef2f2" stroke="#dc2626"/>
+    <text x="636" y="88" font-size="15" font-weight="700" fill="#b91c1c">2. THE BUG</text>
+    <text x="636" y="112" font-size="13" fill="#111827">Passing build/test output mentions BOTH words without failing:</text>
+    <text x="636" y="136" font-size="13" font-family="Consolas, monospace" fill="#111827">tsc:   "Found 0 errors"</text>
+    <text x="636" y="156" font-size="13" font-family="Consolas, monospace" fill="#111827">jest:  "0 failures, 42 passed"</text>
+    <text x="636" y="184" font-size="13" fill="#111827">→ 2 distinct keyword hits ("error" + "fail") on a CLEAN run</text>
+    <text x="636" y="202" font-size="13" font-weight="700" fill="#b91c1c">→ wrongly protected, forever, from compression.</text>
+  </g>
+
+  <!-- Box 3: impact -->
+  <g>
+    <rect x="20" y="270" width="1160" height="150" rx="10" fill="#fff7ed" stroke="#ea580c"/>
+    <text x="36" y="298" font-size="15" font-weight="700" fill="#c2410c">3. OBSERVED IMPACT (issue #1696 stats.json)</text>
+    <text x="36" y="324" font-size="13" fill="#111827">In a long multi-turn JS/TS coding session (KiloCode via headroom proxy), "router:protected:error_output" fired on</text>
+    <text x="36" y="344" font-size="13" fill="#111827">nearly every request. Combined with tool-output exclusion and small-block skips, only a sliver of tokens were ever</text>
+    <text x="36" y="364" font-size="13" fill="#111827">eligible for compression.</text>
+    <rect x="900" y="325" width="60" height="60" fill="#fecaca"/>
+    <text x="930" y="360" text-anchor="middle" font-size="16" font-weight="700" fill="#7f1d1d">0.3%</text>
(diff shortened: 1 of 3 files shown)


AbelVM:
What happens when those keywords appear in prompts, code files or code blocks? like

* `Fix the errors in the code.`
* `console.error(...)`
* `There are failures in the tests that need fixing.`
* ...

sparkbugz:
To build on what @AbelVM is asking, what about other cases of other summary formats like these?

- Failures: 0, Errors: 0
- failed: 0, errors: 0
- Errors=0 Failures=0

The current proposed fix looks to handle 0/no forms such as 0 errors and 0 failures, which would make sense for JavaScript/Typescript-style output, but I'm wondering if it would catch formats like X: 0 or X=0, would it still leave both fail and error visible to the scan? 

This seems like it would affect more than just JavaScript and TypeScript, too, since a good amount of test and CI tools report results in summary formats like those. It may be worth adding extra detection for broader coverage! :)

JerrettDavis (changes requested):
This is the right fix shape, but the zero-result pattern needs to cover one more common passing-test form before merge.

The strong-indicator scan counts `fail` by substring, so `failed` contributes the fail hit. `_ZERO_RESULT_PATTERN` strips `0 failing`, `0 failure`, and `0 failures`, but it does not strip `0 failed`. A passing summary such as `Found 0 errors
Tests: 0 failed, 42 passed` would still leave `failed` in the text, so it can trip the same `error` + `fail` false positive this PR is trying to remove.

Please include `failed` in the scrubbed zero-result forms and add a regression test for `0 errors` + `0 failed`. That keeps the patch aligned with the stated goal while preserving the existing behavior for real failures elsewhere in the blob.

ashishpatel26:
@JerrettDavis fixed in 2f1b66cf — \`_ZERO_RESULT_PATTERN\` now covers \`fail(?:ed|ing|ures?)?\` (so "0 failed" is stripped, not just "failing"/"failure(s)"), plus a regression test for your exact repro (\`"Found 0 errors\nTests: 0 failed, 42 passed"\`).

@sparkbugz same commit also adds label:value forms (\`Failures: 0\`, \`failed: 0\`, \`Errors=0\`) alongside the existing "N word"/"word N" forms — tested in \`test_label_value_summary_formats_are_not_flagged\`.

@AbelVM prose mentions with only 1 distinct keyword (\`console.error(...)\`, "Fix the errors in the code.") were already safe under the pre-existing 2-keyword threshold — this PR doesn't touch that path. Genuine prose that happens to pair two distinct keywords (e.g. "there are errors and it failed") is a real edge case, but it's a limitation of the keyword-substring approach in general (predates this PR), and downstream compressors still preserve real error lines even when a block isn't gate-protected — so the failure mode is over-caution, not data loss. Happy to open a follow-up issue for that broader detection work if you think it's worth tracking separately.

AbelVM:
Problem is when a code file or a prompt has more than 1 offending keyword. Examples:

`console.error("There is a failure in the test");`

`Help me fix the error that leads to a fatal crash; it looks like an uncontrolled exception.`

JerrettDavis (approved):
Reviewed the follow-up commits after the earlier changes-requested review. The zero-result scrubber now covers the missing

PR #1747, opened 2026-07-03
fix(code-compressor): CJK-aware relevance-query symbol matching

lifeodyssey:
## Description

`CodeAwareCompressor` gives a code symbol a relevance "context boost" when the query names it. The query tokenizer in `_analyze_symbol_importance` used an ASCII-only delimiter class, so a CJK query (no spaces, CJK punctuation) collapsed into one blob and never matched an ASCII symbol name; the substring fallback was also gated behind `len(name) > 3`, dropping short ASCII names glued to CJK.

This extracts the query tokenization + matching into two pure helpers, adds CJK/full-width punctuation as delimiters, and relaxes the `len>3` guard only for CJK queries. ASCII/English behavior is byte-identical. `code_compressor` is pure-Python (no Rust twin, no parity fixtures).

## Type of Change

## Changes Made

- `headroom/transforms/code_compressor.py`: add `_query_context_tokens` (CJK/full-width punctuation + ideographic space as delimiters) and `_symbol_in_context` (substring `len>3` guard relaxed only for CJK queries), used by `_analyze_symbol_importance`.
- `tests/test_transforms/test_code_compressor_cjk.py`: pure-function tests (CJK isolation, short-name relaxation, English-unchanged, empty).

## Testing

- [x] Linting passes (`ruff`)

### Test Output

```text
$ .venv/bin/python -m pytest tests/test_transforms/test_code_compressor_cjk.py
6 passed

$ .venv/bin/python -m pytest tests/test_transforms/test_code_compressor.py
35 passed, 39 skipped   # no regression (skips need the [code] tree-sitter extra)

$ ruff check / mypy headroom/transforms/code_compressor.py   # clean
```

## Real Behavior Proof

- Environment: macOS (Darwin), Python in a uv venv, branch `feat/code-compressor-cjk-relevance` off `main`.
- Exact command / steps: called the extracted helpers directly on CJK and ASCII queries.
- Observed result: `_query_context_tokens("请重点保留（parse_config）的解析配置")` isolates `parse_config` as its own token (before: the whole query was one blob, so the exact-match boost never fired); `_symbol_in_context("db", ...)` now matches a short ASCII name glued to a CJK query (before: dropped by the `len>3` guard). English is unchanged — for `"keep the database helper"`, `_symbol_in_context("db", ...)` still returns `False` (no spurious short substring match). All 6 new tests pass; the existing 35 `code_compressor` tests are unchanged.
- Not tested: end-to-end `compress()` (needs the `[code]` tree-sitter extra); the fix is at the pure query-matching layer and is verified there.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A (internal compressor behavior)
- [ ] I have updated the CHANGELOG.md — N/A: internal relevance-scoring fix, no user-facing surface change

## Additional Notes

- Scope note: a pure-CJK query that names the function only by a Chinese description (no ASCII token anywhere) still cannot match an ASCII symbol name — cross-script query matching remains out of scope.

### DIFF
--- a/headroom/transforms/code_compressor.py
+++ b/headroom/transforms/code_compressor.py
@@ -875,9 +875,7 @@ class CodeAwareCompressor(Transform):
             ref_counts[qname] = max(0, count - short_name_def_count.get(short, 1))
 
         # Raw importance signals per symbol
-        context_lower = context.lower() if context else ""
-        context_words = set(re.split(r"[\s,;:.()\[\]{}\"']+", context_lower)) if context else set()
-        context_words.discard("")
+        context_words, context_lower, context_has_cjk = _query_context_tokens(context)
 
         raw_signals: dict[str, float] = {}
         for qname in definitions:
@@ -898,13 +896,9 @@ class CodeAwareCompressor(Transform):
                 if short and short[0].isupper():
                     raw += 1.0
 
-            # Context boost
-            if context_words:
-                name_lower = short.lower()
-                if name_lower in context_words or (
-                    len(name_lower) > 3 and name_lower in context_lower
-                ):
-                    raw += 3.0
+            # Context boost: the relevance query named this symbol.
+            if _symbol_in_context(short.lower(), context_words, context_lower, context_has_cjk):
+                raw += 3.0
 
             raw_signals[qname] = raw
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the code-compressor relevance-query matching change. The helper extraction makes the behavior testable, the CJK/full-width delimiter handling covers the reported symbol isolation problem, and the tests preserve the important ASCII short-name and substring behavior. Checks are green; this looks ready.

PR #1750, opened 2026-07-03
ci(opencode): compile + test the OpenCode plugin on changes

chopratejas:
The OpenCode plugin (`plugins/opencode`) is the routing shim that carries `headroom wrap opencode` traffic through the proxy — but **no CI job ever compiled it**. It has `typecheck`/`build`/`test` scripts that only ran locally, so:
- TypeScript 6.x (#1687) and @types/node 26 (#1688) major-bump PRs had **zero build evidence** (why they're held).
- Any source edit to the plugin could silently break the build.

This adds a path-gated workflow that runs `npm ci → typecheck → build → test` whenever `plugins/opencode/**` (or this workflow) changes. Matches repo conventions (`setup-node@v6`, node 20, npm cache).

**Verified green locally on main:** `tsc --noEmit` clean, `tsup` build ok, 13 vitest tests pass.

Unblocks safe evaluation of the held dependency-bump PRs and protects the routing plugin going forward.

### DIFF
--- a/.github/workflows/opencode-plugin.yml
+++ b/.github/workflows/opencode-plugin.yml
@@ -1,0 +1,41 @@
+name: OpenCode Plugin
+
+# The OpenCode plugin (plugins/opencode) is the routing shim that carries
+# `headroom wrap opencode` traffic through the proxy, yet nothing in CI ever
+# compiled it — so TypeScript / @types/node major bumps and source changes had
+# zero build evidence. This gate runs the plugin's own typecheck + build + test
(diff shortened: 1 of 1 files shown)


JerrettDavis (approved):
Reviewed the new OpenCode plugin workflow against the plugin package scripts and existing workflow conventions. It is correctly path-gated, uses the repo's current checkout/setup-node pattern, caches the plugin lockfile, and runs the right npm ci/typecheck/build/test sequence. Checks are green; this looks ready.

PR #1752, opened 2026-07-03
fix(wrap): prepend rtk bin dir to PATH for wrapped tools

ketpatil77:
## Description

Make wrapped tools able to resolve Headroom-managed `rtk` without requiring users to manually add `~/.headroom/bin` to PATH.

`headroom wrap` now prepends the managed rtk bin directory to the child process PATH before launching the tool.

Closes #1752

## Type of Change

## Changes Made

- `headroom/cli/wrap.py`: added case-insensitive PATH-key resolution so Windows-style `Path` entries are updated in place.
- `headroom/cli/wrap.py`: keeps duplicate prevention on the resolved path key instead of creating a second `PATH` entry.
- `tests/test_cli/test_wrap_helpers.py`: added regression coverage for Windows-style env maps and duplicate prevention.

## Testing

### Test Output

```text
$ C:\Users\Jayesh\AppData\Local\Programs\Python\Python314\python.exe -m pytest tests\test_cli\test_wrap_helpers.py -k "prepend_rtk_bin_to_path"
3 passed, 37 deselected in 1.14s

$ ruff check headroom\cli\wrap.py tests\test_cli\test_wrap_helpers.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, PowerShell, Python 3.14, Headroom wrap helper path-prepend flow.
- Exact command / steps: Ran `_prepend_rtk_bin_to_path()` regression tests against both `PATH` and Windows-style `Path` env dictionaries, plus duplicate-path coverage.
- Observed result: Managed rtk bin dir prepends exactly once and updates existing `Path` key in place without creating a second differently-cased entry.
- Not tested: Full interactive wrapped-shell launch against a live Windows child process in this verification pass.

## Review Readiness

## Checklist

## Additional Notes

- Follow-up push already addressed reviewer concern about case-insensitive Windows PATH handling; this body update clears governance/template blockers.


JerrettDavis (changes requested):
The PATH injection is the right place to solve this, but this helper needs a Windows-safe environment-key handling pass before merge.

`env` comes from `os.environ.copy()` in the wrappers, and on Windows that commonly carries the executable search path under `Path`, not `PATH`. `_prepend_rtk_bin_to_path()` currently reads `env.get("PATH", "")` and writes `env["PATH"]`, which can leave the existing `Path` entry untouched and add a second differently-cased key. With `subprocess.run(..., env=env)` on Windows, duplicate case variants are ambiguous and can result in the child still seeing the old `Path`, so the managed rtk bin directory may not actually be on the child search path.

Please resolve the existing path key case-insensitively, update that key in place, and add tests for a Windows-style env like `{ "Path": "C:`Windows`System32" }` so we know the helper prepends to `Path` rather than creating a separate `PATH`. It would also be worth asserting the duplicate check works against the resolved key.

ketpatil77:
Pushed follow-up fix for review: resolve existing PATH/Path key case-insensitively, update that key in place, and add Windows-style env regression coverage. Re-ran targeted tests and ruff locally before push.

ketpatil77:
Follow-up is already pushed for the requested Windows PATH/Path fix: the helper now resolves the existing path key case-insensitively, updates that key in place, and includes Windows-style env regression coverage plus duplicate-path coverage. Targeted pytest and ruff outputs are in the PR body. Re-review requested when convenient.

JerrettDavis (approved):
The final diff is scoped to prepending the managed RTK bin directory before launching wrapped tools, preserving an existing Path/PATH key on Windows and avoiding duplicates. The helper tests cover both POSIX-style PATH and Windows-style Path. No blocker.

JerrettDavis (changes requested):
The previous Windows PATH-key blocker remains fixed, but the latest merged head is no longer compatible with main. headroom.rtk was removed upstream; _prepend_rtk_bin_to_path() still imports headroom.rtk.RTK_BIN_DIR, so every wrapped launch now reaches ModuleNotFoundError before starting the tool. The current CI test (1) lane confirms this in the existing Codex launch test and all three new helper tests. Please rework the feature against the current managed-tool location/contract (or close it if RTK management was intentionally retired), update the monkeypatch targets and focused tests, and clean up the historical non-Conventional commits that commitlint rejects. A plain rebase cannot make the removed runtime API valid.

ketpatil77:
Closing as superseded by the current runtime direction. The latest review correctly identifies that headroom.rtk and Headroom-managed RTK support were removed upstream, so adapting this PATH injection would preserve a retired contract rather than fix current behavior. Thanks for the re-review and clear diagnosis.

PR #1755, opened 2026-07-03
feat(content-router): lossless compaction for excluded tool output (grep/log/json)

chopratejas:
## What
Excluded tools (Read/Grep/Glob/Write/Edit) are protected from **lossy** compression for accuracy. This adds **information-preserving** compaction, dispatched by detected shape:

| shape | transform | guarantee |
|---|---|---|
| **grep** (SEARCH) | ripgrep `--heading` fold | **byte-lossless** (`search_unheading` recovers) |
| **log** (BUILD_OUTPUT) | ANSI strip + run-collapse | **byte-lossless** modulo non-semantic ANSI |
| **json** | whitespace-minify | **data-lossless** (`json.loads` equal) — NOT byte-exact |

Source code and glob path-lists match nothing → verbatim. Off by default (`compact_excluded_lossless`).

## Why
A live `headroom wrap opencode` run showed the coding persona at ~2.5% savings, because OpenCode's dominant traffic (Grep/Read) is excluded, and RTK (shell-only, lossy) never sees OpenCode's *native* tools (`rtk_tokens_avoided: 0`). This closes the gap with information-preserving transforms.

## Measured
- grep (real `rg`): 41,589 → 26,562 (**−36%**), recover byte-exact.
- log: ~35% off, recover modulo ANSI.
- json (pretty): ~49% off, data-equal.

## Accuracy tiers (important)
- **grep / log = byte-lossless → edit-safe.** grep is search results (never edited); logs are rarely edited and recover byte-for-byte (modulo color).
- **json = data-lossless, not byte-exact.** A read-then-`Edit(old_string)` on the *same* JSON file could miss (the agent sees minified). Gated + documented; disable this tier for read-then-edit-JSON workloads.

## Detection notes
- grep gated on `_try_detect_search` (row shape) — the general/Magika classifier calls grep-over-a-codebase SOURCE_CODE and would miss it.
- Read of **source code** is not search/log/json → **untouched** (Edit's exact `old_string` preserved), verified by test.

10 tests (byte-exact search/log, data-equal json, no-op source/glob, off-by-default, end-to-end pipeline). ruff + mypy clean. Off by default; intended for the coding persona to enable.


chopratejas:
Superseded by #1762 — rebased onto main after #1732 (personas) landed; #1762 combines the lossless-excluded compaction (grep/log/json) with the coding/general persona enablement. No personas duplication.

PR #1766, opened 2026-07-03
fix(dashboard): serve per-request metadata to trusted-gateway peers

KennethWKZ:
## Description

The dashboard's per-request metadata — the `recent_requests` / `request_logs` tail and the `config` block (which echoes upstream API URLs + backend settings) — is gated to loopback callers via `_request_is_loopback`. It requires **both** a loopback peer IP (`request.client.host == 127.0.0.1`) and a loopback `Host` header.

When Headroom runs in a **bridge-network container** (Docker/podman, or Apple Containerization / `mocker`), a browser on the host reaches the proxy through the container gateway, so `request.client.host` is the **gateway IP** (e.g. `172.18.0.1`, or `192.168.64.1` on macOS vmnet), not `127.0.0.1`. `include_sensitive` is therefore `False`, and the "Recent Requests" table renders empty even though the operator is browsing locally at `http://127.0.0.1:8787/dashboard`.

`curl` from **inside** the container (real `127.0.0.1` peer) confirmed the data is present and populated — only the host-browser path was being stripped.

The fix treats a peer inside an operator-configured trusted-gateway CIDR (`HEADROOM_PROXY_TRUSTED_GATEWAY_CIDRS` — the same allow-list already used by `forwarded_headers.py` to sanitize `X-Forwarded-*`) as loopback-equivalent, while **retaining the loopback `Host`-header gate as the DNS-rebinding defence**. It is opt-in and empty by default, so there is **no behavior change** unless the operator explicitly allow-lists their container gateway.

Closes #

## Type of Change

## Changes Made

- `headroom/proxy/server.py` — `_request_is_loopback` now: (1) always enforces the loopback `Host`-header gate first; (2) returns `True` for a genuine loopback peer; (3) additionally returns `True` for a peer inside `HEADROOM_PROXY_TRUSTED_GATEWAY_CIDRS` via the existing `peer_is_trusted_gateway` / `load_trusted_gateway_cidrs` helpers.
- `tests/test_proxy_loopback_gating.py` — added `test_stats_metadata_served_to_trusted_gateway_peer`: gateway peer stripped without the allow-list, served with it, and DNS-rebinding (non-loopback `Host`) still rejected even for a trusted gateway peer.
- `CHANGELOG.md` — Unreleased → Fixed entry.

## Testing

### Test Output

```text
$ pytest tests/test_proxy_loopback_gating.py -q
14 passed, 1 warning in 3.56s

$ ruff check headroom/proxy/server.py tests/test_proxy_loopback_gating.py
All checks passed!
```

## Real Behavior Proof

- Environment: Headroom 0.29.0 in a `mocker compose` (Apple Containerization) bridge container on macOS; host browser at `http://127.0.0.1:8787/dashboard`.
- Exact command / steps: before the fix, `mocker compose exec headroom-proxy sh -c 'curl -s http://127.0.0.1:8787/stats'` (peer = real `127.0.0.1`) returned a populated `recent_requests` array, while the host browser saw an empty table. After adding `HEADROOM_PROXY_TRUSTED_GATEWAY_CIDRS` covering the container gateway and recreating, the host browser's dashboard shows the Recent Requests table again.
- Observed result: dashboard per-request table restored for the host browser; aggregate-only view unchanged for untrusted network callers.
- Not tested: IPv6 gateway CIDRs (the underlying `peer_is_trusted_gateway` supports them; not exercised in this environment).

## Review Readiness

## Additional Notes

Pure opt-in: `HEADROOM_PROXY_TRUSTED_GATEWAY_CIDRS` is empty by default, so `_request_is_loopback` behavior is byte-identical to today unless an operator allow-lists a gateway CIDR. Reuses the existing trusted-gateway machinery rather than introducing a new config surface. Docs/compose examples intentionally omitted — deployment-specific.


JerrettDavis (approved):
Reviewed the dashboard metadata gating change against the existing trusted-gateway helper. The Host-header loopback check still gates every sensitive response, and the only expanded case is an operator-configured gateway CIDR via the existing HEADROOM_PROXY_TRUSTED_GATEWAY_CIDRS mechanism. The regression test covers untrusted gateway stripping, trusted gateway serving, and DNS-rebinding rejection. Checks are green; this looks ready.

JerrettDavis (approved):
The trusted-gateway extension keeps the Host-header loopback gate in place and only treats configured gateway CIDRs as loopback-equivalent. That preserves the DNS-rebinding defense while fixing the container dashboard case, and the test covers default stripping, allow-listed serving, and hostile Host rejection. Branch needs conflict resolution before merge, but no blocker in the diff.

JerrettDavis (approved):
Re-reviewed current head after the main merge. The trusted-gateway metadata path remains narrowly gated to loopback/trusted-forwarded peers, and the tests cover the added loopback cases.

- `python -m pytest tests/test_proxy_loopback_gating.py -q` -> 14 passed
- `uvx ruff==0.15.17 check headroom/proxy/server.py tests/test_proxy_loopback_gating.py`
- `uvx ruff==0.15.17 format --check headroom/proxy/server.py tests/test_proxy_loopback_gating.py`

PR #1781, opened 2026-07-04
fix(ccr): make expired retrieve misses terminal

### LINKED ISSUE #1776: Expired CCR entries return a non-terminal "Content not found" — agents retry dead hashes (observed 25x on one hash)
Follow-up to the TTL-cliff discussion in #714 / #709: this is what happens when the cliff fires. Verified on v0.27.0 (deployed); TTL default and expiry logic unchanged on main (`headroom/cache/compression_store.py`).

- CCR entries expire at `DEFAULT_CCR_TTL_SECONDS = 1800`; expiry is `created_at`-based — `last_accessed` exists but never extends life.
- Markers in the conversation persist indefinitely and survive client-side context compaction, so long agent sessions retrieve hashes whose entries are gone.
- The MCP error is "Content not found. It may have expired or the hash may be incorrect." — indistinguishable from a typo and nothing tells the model to stop retrying.
- Deployment data: 135/2152 retrieve events in a 2.5-day log window were store misses (6%); ~20% of transcript-visible retrieve calls over 7 days failed; worst case, one agent retried the same dead hash 25 times in 11 minutes, then re-ran the source command (defeating the compression).

Asks:
1. Make the expiry error terminal and actionable, e.g. "expired — do not retry this hash; re-run the source command". Models follow imperative error text.
2. Consider refreshing TTL on access, or the turn-distance metric proposed in the #709 write-up.
3. Document `HEADROOM_CCR_TTL_SECONDS` (currently discoverable only in source).

rodboev:
## Description

Expired CCR hashes currently come back through `headroom_retrieve` as the same generic missing-content error used for typos and never-stored hashes. That leaves agents with no terminal signal, so they can retry a dead hash instead of rerunning the source command or rereading the source file. This change uses the cache store's existing TTL status metadata before the MCP retrieval path loses that distinction, then returns expired-hash guidance only when the local store proves the entry existed and expired.

Closes #1776

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Uses CCR store status metadata to distinguish expired local hashes from never-stored hashes in the MCP retrieval path.
- Keeps proxy fallback and successful local retrieval behavior unchanged.
- Adds focused regression coverage for expired stored hashes, the status-to-retrieve TTL boundary, proxy fallback preservation, and missing-hash negative space.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_ccr_mcp_server.py -q`)
- [x] Linting passes (`uv run ruff check headroom/ccr/mcp_server.py tests/test_ccr_mcp_server.py`)

### Test Output

```text
Base pytest:
FAILED tests\test_ccr_mcp_server.py::test_mcp_retrieve_expired_hash_returns_terminal_guidance
E   KeyError: 'status'

Head pytest:
tests\test_ccr_mcp_server.py ...s..........                              [100%]
13 passed, 1 skipped in 0.35s

Ruff:
All checks passed!
```

## Real Behavior Proof

- Environment: Windows, focused local pytest through the headless runner.
- Exact command / steps: Store a CCR entry with a short TTL, advance beyond expiry, call `HeadroomMCPServer._retrieve_content(hash)`, force a second entry to cross TTL between status inspection and `retrieve()`, stub a proxy-backed retrieval for local misses, then call the same method with a never-stored hash and no proxy hit.
- Observed result: The already-expired hash and the hash that expires during retrieval both return terminal expired guidance with `status: expired`; missing and expired local hashes still return proxy data when the proxy fallback succeeds; a never-stored hash with no proxy hit still returns the generic missing-hash error and no expired status.
- Not tested: Full suite, live agent retry behavior, and live external proxy-backed retrieval.

## Review Readiness

## Checklist

## Additional Notes

Documentation and changelog are left unchecked because this is a narrow MCP error-shape fix and Headroom's changelog is generated from conventional commits.

### DIFF
--- a/headroom/ccr/mcp_server.py
+++ b/headroom/ccr/mcp_server.py
@@ -34,6 +34,7 @@ from typing import Any
 
 from headroom import paths as _paths
 from headroom import savings_ledger
+from headroom.cache.compression_store import format_retrieval_miss_detail
 
 # fcntl is Unix-only; on Windows we skip file locking (stats are best-effort).
 # Keep the module typed as Any so Windows mypy runs don't try to resolve Unix-only attrs.
@@ -435,7 +436,9 @@ class HeadroomMCPServer:
         """
         # Check local store first
         store = self._get_local_store()
+        entry_status = store.get_entry_status(hash_key, clean_expired=False)
         entry = store.retrieve(hash_key)
+        expired_entry_status = None
         if entry:
             self._stats.record_retrieval(hash_key)
             return {
@@ -446,6 +449,19 @@ class HeadroomMCPServer:
                 "compressed_item_count": entry.compressed_item_count,
                 "retrieval_count": entry.retrieval_count,
             }
+        if entry_status.get("status") == "expired":
+            expired_entry_status = entry_status
+        elif entry_status.get("status") == "available":
+            created_at = entry_status.get("created_at")
+            ttl_seconds = entry_status.get("ttl_seconds")
+            if isinstance(created_at, (int, float)) and isinstance(ttl_seconds, (int, float)):
+                age_seconds = time.time() - created_at
+                if age_seconds > ttl_seconds:
+                    expired_entry_status = {
+                        **entry_status,
+                        "status": "expired",
+                        "age_seconds": age_seconds,
+                    }
 
         # Fall back to proxy if available
         if self.check_proxy and HTTPX_AVAILABLE:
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
The expired-hash path now distinguishes expired local entries from never-stored misses and gives terminal guidance while still allowing a configured proxy to satisfy the hash. The race where an entry expires between status and retrieve is covered. No blocking issues from me.

PR #1795, opened 2026-07-04
fix(bedrock): resolve global.* inference profiles + pin per-user app-profile ARNs

KennethWKZ:
## What & why

Bedrock model resolution failed on accounts whose inference profiles use the newer `global.` cross-region prefix and undated version suffixes. On such an account, `list_inference_profiles` returns current-gen models as `global.anthropic.claude-opus-4-8`, `global.anthropic.claude-sonnet-5`, `global.anthropic.claude-opus-4-6-v1`, `global.anthropic.claude-fable-5`, etc.

`_normalize_bedrock_profile_id` only stripped `us.`/`eu.`/`apac.`/`au.` and only matched a dated `-vN:M` suffix, so every `global.`-prefixed profile was silently dropped from the discovered model map. Requests then fell through to the fabricated fallback id and Bedrock rejected them:

```
litellm.BadRequestError: BedrockException - {"message":"The provided model identifier is invalid."}
```

On the affected account the discovered-profile count went from 5 → 14 after the fix.

## Changes

1. **`_normalize_bedrock_profile_id`** — strip the `global.` prefix in addition to the region prefixes, and match undated version suffixes (`-v1`, or none at all) alongside the legacy dated `-vN:M`.

2. **`HEADROOM_BEDROCK_MODEL_MAP` operator override** (read from the process environment). AWS discovery keys the model map by normalized model name, so it cannot disambiguate application inference profiles that share one underlying model — e.g. a team where `claude-sonnet-5-alice` and `claude-sonnet-5-bob` both resolve to `claude-sonnet-5`. When you need requests billed to a *specific* application profile (per-user cost attribution), pin it explicitly:

   ```
   HEADROOM_BEDROCK_MODEL_MAP="claude-sonnet-5=arn:aws:bedrock:REGION:ACCT:application-inference-profile/abc123,claude-opus-4-8=arn:...:application-inference-profile/def456"
   ```

   The plain model name (kept plain so a client's tool-search deferral stays on) resolves to the pinned ARN, routed via the converse endpoint. The override wins over discovery; when unset, discovery-only behaviour is unchanged.

## Tests

`tests/test_bedrock_region.py` (43 passing): `global.`-prefixed normalization across dated / bare-`-v1` / no-suffix shapes; override-map parsing (empty, single, multi, whitespace, malformed-skip); and `map_model_id` override routing (pinned name → app-profile ARN via converse, wins over discovery; unpinned name falls through).

No behavioural change for accounts already on system-defined region-prefixed profiles.

### DIFF
--- a/headroom/backends/litellm.py
+++ b/headroom/backends/litellm.py
@@ -226,6 +226,39 @@ def _fetch_bedrock_inference_profiles(
     return model_map
 
 
+def _parse_bedrock_model_overrides(raw: str | None) -> dict[str, str]:
+    """Parse the ``HEADROOM_BEDROCK_MODEL_MAP`` operator override.
+
+    AWS discovery keys the model map by the *normalized model name*, so it
+    cannot disambiguate application inference profiles that share one
+    underlying model — e.g. a team where ``claude-sonnet-5-kenneth`` and
+    ``claude-sonnet-5-jeremy`` both resolve to ``claude-sonnet-5``. When you
+    need requests billed to a *specific* application profile (per-user cost
+    attribution), pin the mapping explicitly here. The plain name Claude Code
+    sends (kept plain so tool-search deferral stays on) resolves to your ARN.
+
+    Format: comma-separated ``name=target`` pairs, where ``target`` is an
+    application-inference-profile ARN (routed via the converse endpoint) or
+    any LiteLLM model string. Whitespace around pairs is ignored; blank
+    entries are skipped.
+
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the Bedrock profile normalization and override routing. The global./bare-version normalization closes the discovery gap, and the explicit model map is parsed conservatively with tests for override precedence and converse ARN routing. I pushed a formatting-only commit so lint can rerun cleanly. No blocking issues found.

JerrettDavis (approved):
Still looks good on the latest head. I re-ran the focused suite locally with pytest tests/test_bedrock_region.py -q: 43 passed. CI is green as well.

PR #1797, opened 2026-07-04
fix(build): support Intel macOS (x86_64-apple-darwin) via ort-load-dynamic (fixes #941)

### LINKED ISSUE #941: [BUG] No installable wheel for Intel Mac (x86_64-apple-darwin) — ort-sys fails to build
## Environment
- macOS Sequoia (Darwin 25.5.0)
- CPU: Intel Core i7-9750H (x86_64)
- Python: 3.13 / pipx 1.14.0
- Headroom version attempted: latest (v0.25.0)

## Problem

`pip install headroom-ai` and `pipx install headroom-ai` both fail on Intel Mac with:

```
error: ort-sys@2.0.0-rc.12: ort does not provide prebuilt binaries for the target `x86_64-apple-darwin` with feature set (no features).
error: build script logged errors
Error: command ['maturin', 'pep517', 'build-wheel', ...] returned non-zero exit status 1
```

## What I tried

1. Plain install: `pipx install headroom-ai` → ort-sys build fails
2. With `[all]` extra → same error (+ unrelated fastapi conflict, see #930)
3. Installed Rust via rustup → still fails
4. `ORT_STRATEGY=system` with Homebrew `onnxruntime@1.26.0` → `ort-sys could not link to the ONNX Runtime build`

## Root cause

Release v0.25.0 ships pre-built wheels for:
- ✅ `macosx_11_0_arm64` (Apple Silicon)
- ✅ `manylinux_2_28_x86_64` (Linux x86)
- ✅ `manylinux_2_28_aarch64` (Linux ARM)
- ❌ **`macosx_x86_64` — missing (Intel Mac)**

Intel Mac users cannot install headroom-ai at all. Source build also fails because `ort-sys` has no `x86_64-apple-darwin` support without a pre-linked ONNX Runtime.

## Request

Please add `macosx_x86_64` (targeting `macosx_10_15_x86_64` or similar) to the GitHub Actions release matrix so a pre-built wheel is shipped for Intel Mac users.

warnes:
## Problem

`headroom-ai` fails to build from source on Intel macOS (`x86_64-apple-darwin`),
both with and without the `[all]` extra:

```
error: ort-sys@2.0.0-rc.12: ort does not provide prebuilt binaries for the target
`x86_64-apple-darwin` with feature set (no features).
```

Reported in #941. `ort-sys`'s `download-binaries` strategy (used by the
`ort-download-binaries-rustls-tls` fastembed feature that `headroom-core`
depends on for all non-Windows targets) only ships prebuilt ONNX Runtime
binaries for Windows, Linux (x86_64/aarch64), and macOS **Apple Silicon**.
There's currently no way to install this package from source on an Intel Mac
at all — with or without `[all]`, since the ONNX dependency lives in
`headroom-core` itself, not behind a pip extra.

## Root cause

`crates/headroom-core/Cargo.toml` already has a working fallback for this
*exact* class of problem — for Windows, it swaps `fastembed`'s
`ort-download-binaries-rustls-tls` feature for `ort-load-dynamic`, which
dlopen's a system-provided ONNX Runtime at runtime instead of requiring a
bundled prebuilt binary for the exact target triple. Intel macOS just never
got the same treatment.

## Fix

Extends the existing `ort-load-dynamic` branch to also cover
`target_os = "macos", target_arch = "x86_64"`.

## Documentation

Also adds an Intel-macOS subsection next to the existing "Corporate /
SSL-inspection environments" section, since the `ORT_STRATEGY=system` +
`ORT_LIB_LOCATION` mechanism documented there for a different reason is
*also* a fully working, no-source-patch workaround available today:

```bash
brew install onnxruntime
ORT_STRATEGY=system \
ORT_LIB_LOCATION="$(brew --prefix onnxruntime)/lib" \
ORT_PREFER_DYNAMIC_LINK=1 \
  pip install "headroom-ai[all]"

export ORT_DYLIB_PATH="$(brew --prefix onnxruntime)/lib/libonnxruntime.dylib"
```

Two things cost real debugging time and seemed worth documenting either way:
`ORT_LIB_LOCATION` must point at the `lib/` subdirectory specifically (the
Homebrew keg has no single-file library at the prefix root — pointing at the
bare prefix gets a *different*, more confusing error: "could not link to the
ONNX Runtime build"), and `ORT_PREFER_DYNAMIC_LINK=1` is required — without
it, `ORT_STRATEGY=system` still attempts static linking, which the Homebrew
keg doesn't provide.

## Testing

- `cargo check -p headroom-core` and a full `maturin build --release` succeed
  on Intel macOS (macOS 26.5.1) with this patch and `ORT_DYLIB_PATH` pointed
  at a Homebrew onnxruntime 1.27.0.
- Verified beyond just compiling: loaded the built wheel's `_core.abi3.so`
  directly and called `detect_content_type` (the magika/ONNX-backed
  classifier, which shares the ONNX Runtime instance per the comment in
  `headroom-core/Cargo.toml`). Ran successfully, no dyld/link errors.
- Independently verified the doc-only workaround builds a working wheel
  through the **unmodified** sdist via `pip wheel` — no Cargo.toml changes
  needed for that path at all.
- Not tested on Apple Silicon or Linux; the `cfg()` predicate is scoped to
  `(target_os = "macos", target_arch = "x86_64")` so it shouldn't affect
  either.

Happy to split this into two PRs (code fix / doc fix) if that's easier to
review.

## Note

I noticed a branch, `fix-wheel-matrix-vendored-openssl-and-drop-intel-mac`,
that appears to drop Intel macOS from the release wheel matrix rather than
fix source builds for it. It looked stale relative to `main` (interleaved
with much older history) when I checked, so I wasn't sure whether it reflects
current intent — if the project has already decided to drop Intel macOS
support rather than fix it, feel free to close this instead, no worries
either way.


JerrettDavis (approved):
This looks correct for the scoped Intel macOS source-build failure: the target cfg keeps the existing download-binaries path for Linux/Apple Silicon, routes only x86_64 macOS through the same dynamic ORT strategy already used on Windows, and the README documents the system-ORT workaround clearly. Branch currently needs conflict resolution before merge, but I do not see a code blocker in the PR diff.

JerrettDavis (approved):
Re-reviewed the current head after the latest `main` merge. The PR diff is now README-only; the Intel macOS `ort-load-dynamic` target configuration is already present on `main`, and this remaining diff documents the Homebrew/system ONNX Runtime workaround and required `ORT_*` variables.

Current diff reviewed:

```text
README.md | 22 insertions
```

The documentation is accurate: `ORT_LIB_LOCATION` points to the Homebrew `lib/` directory for build/link, `ORT_PREFER_DYNAMIC_LINK=1` avoids static linking, and `ORT_DYLIB_PATH` points to the runtime dylib for dynamic load.

PR #1784, opened 2026-07-04
feat(transforms): structure-aware YAML/TOML/INI config compression

agaonker:
## Description

YAML/TOML/INI config files (k8s manifests, CI pipelines, Cargo/pyproject manifests, INI files) are high-frequency agent tool outputs, but they have no native compressor today: magika classifies them as `SourceCode`, `enable_code_aware` is off by default, so they degrade to lossy Kompress prose compression or pass through untouched. This PR adds a lossless-first structured-config pipeline: a new `STRUCTURED_CONFIG` content type, a `CONFIG` routing strategy, and a two-tier `ConfigCompressor` where every byte either survives, round-trips through an exact inverse, or sits behind a CCR retrieve marker with the original persisted first.

A 3-container k8s Deployment folds 793 → 483 chars (39%) fully losslessly; the round-trip is byte-verified before the result is adopted.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Made

- `content_detector.py`: new `ContentType.STRUCTURED_CONFIG` + `_try_detect_structured_config()`. TOML/INI claims are stdlib-parser-confirmed (`tomllib`/`configparser` disambiguate the shared `[section]` shape); YAML is heuristic (key/list/doc-marker line share + structure signal) with prose, markdown-front-matter, and JSON-object guards. Runs after search/log/tabular so existing claims are untouched.
- `content_router.py`: `CompressionStrategy.CONFIG` wired at every TABULAR touchpoint — `enable_config_compressor` flag, dispatch branch, Kompress no-savings fallback, `lossless` mode kind (`lossless_config` chain label so the token-measured acceptance gate applies), strategy↔type maps, lazy getter mirroring `ccr_inject_marker`. Plus a SourceCode misroute guard on the native detection path so YAML/TOML escapes the magika verdict.
- `lossless_compaction.py`: new exact-inverse pair `fold_repeated_blocks`/`unfold_repeated_blocks` — a K-line run that reproduces K lines seen D lines earlier collapses to `... (repeats K lines from D lines back)`; repeats need not be adjacent, which is what k8s stanzas actually look like. New `config` kind in `compact_lossless` with the module's standard self-verified round-trip (falls back to the original on any mismatch, never raises).
[...]

PR #1799, opened 2026-07-04
feat(config): fold TOML array-of-tables to csv-schema via SmartCrusher

agaonker:
## Description

Adds a schema-fold tier to the structured-config compressor (introduced in #1784). TOML files containing an `[[array-of-tables]]` are parsed with the stdlib `tomllib` reference parser and bridged to SmartCrusher's lossless `csv-schema` renderer, which folds the repeated per-record keys into a single schema over the rows. On lockfiles and override-lists — where the repeated keys dominate the byte count — this is a large win.

**Stacked on #1784 — review only the top commit** (`feat(config): fold TOML array-of-tables to csv-schema via SmartCrusher`). The base commit is #1784's config-compressor PR; this PR will collapse to the single new commit once #1784 merges.

Faithfulness is guaranteed by construction, not by a heuristic:
- `tomllib` is the reference TOML parser, so the extracted records are ground-truth.
- `csv-schema` is a lossless JSON renderer (`smart_crusher.py` documents it as such), so the model reads a faithful, reformatted view of the exact parsed data.
- Byte-exact recovery rides the existing CCR path — the original is persisted to the `CompressionStore` and a `Retrieve original: hash=…` marker is emitted. The fold is only emitted when that store write succeeds, so nothing is ever unrecoverable.

Scope is deliberately **TOML-only**: `tomllib` is stdlib, whereas PyYAML is only a *transitive* dependency (not declared in `pyproject.toml`), and INI record-sections would need a bespoke dict-of-dicts→records transform. Those flavors can follow in a separate PR with an explicit dependency decision.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Changes Made

- `headroom/transforms/config_compressor.py`: added Tier 3 (`_schema_fold`) — TOML `[[array-of-tables]]` → `tomllib` → JSON → `SmartCrusher(csv-schema)`. New `enable_schema_fold` config flag (default on; auto-off in lossless mode since it rides `enable_ccr`). The fold competes with the reversible text tiers and is adopted only when strictly smaller. Added `_load_toml` (stdlib parser with tomli backport) and `_json_default` (TOML date/time → ISO; bail on any other non-serializable value).
- Recovery reuses the existing `CompressionStore` + `Retrieve original: hash=` marker; no new CCR plumbing.
- `tests/test_transforms_config_compressor.py`: 14 new tests covering the fold, big-win assertion, byte-exact CCR round-trip, lossless-mode disable, flag-off, non-TOML skip, no-array skip, small-array `passthrough` decline, store-failure fallback, savings-floor rejection, unparseable/non-serializable bails, `_load_toml`/`_json_default` units, and a datetime-valued fold.

## Testing

- [x] New tests added for the new behavior

### Test Output

```
tests/test_transforms_config_compressor.py ............................. [ 59%]
headroom/transforms/config_compressor.py     127      0     36      0   100%
============================== 49 passed in 0.61s ==============================
```

Must-stay-green suites (`test_lossless_mode`, `test_lossless_excluded_compaction`, `test_transforms_content_detection`, `test_compression_fidelity_regression`) — 48 passed. Router/tabular/smart_crusher regression — 73 + 82 passed. `mypy --strict` clean on the changed module.

## Real Behavior Proof

- Environment: local, Python 3.11.0, macOS (darwin), `HF_HUB_OFFLINE=1`
- Exact command / steps: parsed a 25-record `[[tool.mypy.overrides]]` TOML through `ConfigCompressor(ConfigCompressorConfig(enable_ccr=True)).compress()`, then retrieved the CCR hash from the `CompressionStore`.
- Observed result: `strategy=config_schema_fold`, 2765 → 840 chars (30% of original); the marker hash resolved to the byte-exact original (`recovered == original` True); with `enable_ccr=False` (lossless mode) the fold did not run and no marker was emitted; a 3-record long-valued `[[package]]` array correctly declined (SmartCrusher `passthrough`).
- Not tested: the live proxy end-to-end path and non-TOML flavors (YAML/INI schema folding is intentionally out of scope for this PR).

## Review Readiness

### DIFF
--- a/headroom/transforms/config_compressor.py
+++ b/headroom/transforms/config_compressor.py
@@ -1,0 +1,190 @@
+"""Structured-config compressor for YAML/TOML/INI tool output.
+
+Config files (k8s manifests, CI pipelines, pyproject/Cargo manifests, INI
+files) are high-frequency agent payloads with heavy structural repetition,
+but they have no native compressor — magika tags them SOURCE_CODE and they
+fall through to the lossy prose path. This module compresses them in two
+lossless-first tiers:
+
+- Tier 1 (reversible, allowed in no-CCR lossless mode): format-agnostic
+  text-level compaction via `lossless_compaction` — identical-line runs and
+  repeated multi-line stanzas collapse behind exact-inverse markers, with the
+  round-trip self-verified before the result is adopted.
+- Tier 2 (CCR-recoverable, default mode only): whole-line comments and blank
+  lines are elided behind a summary line carrying a ``Retrieve original:
+  hash=…`` marker; the full original is persisted to the CompressionStore
+  first, so nothing is lost.
+
+No new compression algorithm and no new CCR plumbing live here — Tier 1 rides
+`compact_lossless` and Tier 2 rides the production `CompressionStore`.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+from dataclasses import dataclass
+from typing import Any
+
+from .content_detector import ContentType, detect_content_type
+from .lossless_compaction import compact_lossless
+
+logger = logging.getLogger(__name__)
+
+# Whole-line comment prefixes per flavor. INI values can span indented
+# continuation lines, so INI only elides column-0 comment lines and keeps
+# blanks (configparser keeps blank lines inside multi-line values).
+_COMMENT_RES = {
+    "yaml": re.compile(r"^\s*#"),
+    "toml": re.compile(r"^\s*#"),
+    "ini": re.compile(r"^[#;]"),
+}
+
+# Content where whole-line elision is unsafe: a '#' line inside a YAML block
+# scalar or a TOML multi-line string is data, not a comment. Detection is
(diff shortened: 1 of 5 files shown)


JerrettDavis (approved):
Reviewed the schema-fold top commit. The TOML-only gate, stdlib parsing, CCR-store requirement before emitting the reformatted view, and lossless-mode disable all line up with the intended risk boundary. Tests cover the important declines as well as recovery of the original. This branch is stacked/conflicting, but I do not see a blocker in the new schema-fold logic itself.

JerrettDavis (approved):
Reviewed the TOML array-of-tables schema-fold layer on top of the config compressor. It is appropriately TOML-only, uses tomllib as the source of truth, serializes date/time values explicitly, and only emits the schema view after the original is stored for CCR retrieval and the folded form is actually smaller. The tests cover disable flags, store failure, non-TOML inputs, invalid TOML, and datetime handling. Looks good.

Issue #1808, opened 2026-07-05
[BUG] Two bugs in Headroom's OpenAI-compatible proxy path: session collisions and duplicate response headers


valeriobitzone-png:
# Two bugs in Headroom's OpenAI-compatible proxy path: session collisions and duplicate response headers

*A debugging writeup, with root causes, reproductions, and patches.*

## Context

While integrating Headroom (the context-optimization proxy) in front of an
OpenAI-compatible agentic client, two distinct bugs surfaced. Both are
reproducible, both have a clear root cause in the source (headroom-ai
0.24.0), and both have a minimal, tested fix. This writeup documents them so
the maintainers (or anyone else hitting the same symptoms) can act on
precise information instead of vague "it hangs" reports.

The setup: an agentic coding client with a large, static system prompt
(thousands of tokens, unchanged across every conversation) plus a second,
smaller system message carrying per-conversation dynamic data (assigned
ports, timestamps, repo context). Requests go through Headroom's proxy in
`--mode cache` before reaching an OpenAI-compatible upstream. The bugs
themselves are provider-agnostic — nothing here depends on which upstream
sits behind the proxy.

### Symptom

In `--mode cache`, after a few turns, the client's own user message would
appear **duplicated** inside a single message's `content` array — the same
text block twice, e.g.:

```json
{
  "role": "user",
  "content": [
    {"type": "text", "text": "some message"},
    {"type": "text", "text": "some message"}
  ]
}
```

The upstream call itself succeeded (HTTP 200, valid completion generated),
but the calling client — which tracks its own conversation state — would
receive a response that no longer matched what it believed it had sent, and
would stall silently.

### Root cause

`SessionTrackerStore.compute_session_id()` in
`headroom/cache/prefix_tracker.py` determines which `PrefixCacheTracker`
instance (i.e., which "conversation") a request belongs to by hashing
`model + system_prompt_content`. The relevant code:

```python
system_content = ""
for msg in messages:
    if msg.get("role") == "system":
        content = msg.get("content", "")
        if isinstance(content, str):
            system_content = content[:500]
        elif isinstance(content, list):
            for block in content:
                if isinstance(block, dict) and block.get("type") == "text":
                    system_content = block.get("text", "")[:500]
                    break
        break  # <-- stops at the FIRST system message
```

Two problems compound here:

1. It **stops at the first `system`-role message** it finds (`break`).
2. It **truncates to the first 500 characters** of that message.

For any client whose first system message is a large, static, framework-level
[...]

PR #1827, opened 2026-07-06
fix(cache): avoid fallback session collisions

rodboev:
## Description

Cache-mode session tracking currently collapses unrelated conversations when they share a large static first system prompt. The fallback session-id hash ignores later system messages entirely, so dynamic per-conversation context can get cut out of the key and two different sessions reuse the same `PrefixCacheTracker`. This hashes the full ordered system-text payload instead, while leaving explicit `x-headroom-session-id` overrides untouched. Refs #1808.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Collected all system-text content when building the fallback cache session id.
- Stopped truncating fallback session-id input to the first 500 characters of the first system message.
- Added a regression that proves two conversations with different later system context no longer collide.
- Added a preservation test that appending only non-system turns keeps the same fallback session id.
- Applied the pinned Ruff formatter to three pre-existing files on the current base so the repo-wide lint job passes unchanged semantics.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_cache/test_prefix_tracker.py -q`)
- [x] Linting passes (`uv run ruff check headroom/cache/prefix_tracker.py tests/test_cache/test_prefix_tracker.py`)

### Test Output

```text
uv run pytest tests/test_cache/test_prefix_tracker.py -q
40 passed, 1 warning in 0.15s

uv run ruff check headroom/cache/prefix_tracker.py tests/test_cache/test_prefix_tracker.py
All checks passed!

uv run ruff check .
All checks passed!

uv run ruff format --check .
1046 files already formatted
```

## Real Behavior Proof

- Environment: Windows, project `uv` environment, focused cache-tracker regression.
- Exact command / steps: run `tests/test_cache/test_prefix_tracker.py` on `origin/main` with the new collision regression present, then rerun the same file on this branch.
- Observed result: base returns the same session id for two conversations that differ only in a later system message and fails `assert id_a != id_b`; head passes the focused file and keeps the fallback session id stable when only non-system turns are appended.
- Not tested: live proxy traffic through a real agentic client.

## Review Readiness

## Checklist

## Additional Notes

This is only the session-collision half of #1808. The duplicate-response-header fix stays separate so this PR can reference the issue without claiming the whole bug report is resolved. The extra formatting-only diff comes from the current base failing the pinned full-repo Ruff format check.

### DIFF
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -496,20 +496,19 @@ class SessionTrackerStore:
             if session_header:
                 return str(session_header)
 
-        # Fall back to hashing model + system prompt
-        system_content = ""
+        # Fall back to hashing model + all system-text content.
+        system_parts: list[str] = []
         for msg in messages:
             if msg.get("role") == "system":
                 content = msg.get("content", "")
                 if isinstance(content, str):
-                    system_content = content[:500]  # First 500 chars is enough
+                    system_parts.append(content)
                 elif isinstance(content, list):
                     for block in content:
                         if isinstance(block, dict) and block.get("type") == "text":
-                            system_content = block.get("text", "")[:500]
-                            break
-                break
+                            system_parts.append(block.get("text", ""))
 
+        system_content = json.dumps(system_parts, ensure_ascii=False, separators=(",", ":"))
         key = f"{model}:{system_content}"
         return hashlib.md5(key.encode()).hexdigest()[:16]  # nosec B324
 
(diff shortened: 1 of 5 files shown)


JerrettDavis (approved):
Reviewed the fallback session-id hashing change. Including all system-text blocks removes the collision class while keeping non-system turn appends stable, and the tests cover both sides. No blocking issues found.

PR #1837, opened 2026-07-06
fix: detect and clear stale ANTHROPIC_BASE_URL from crashed wrap sessions (#1768)

### LINKED ISSUE #1768: wrap claude: ANTHROPIC_BASE_URL leaked in .claude/settings.local.json on unclean exit — bare claude then hangs on dead proxy port
**Affected:** 0.27.0, 0.28.0, `main` (`headroom/cli/wrap.py`, `_write_claude_wrap_base_url`).

**Bug:** `headroom wrap claude` writes `env.ANTHROPIC_BASE_URL: http://127.0.0.1:<port>` into `$CWD/.claude/settings.local.json` (the #951 daemon-worker workaround) and removes it only in the launch `finally`. Unclean wrap death (SIGKILL, SIGHUP on terminal close / `tmux kill-session`, reboot) skips the `finally`; the entry persists indefinitely. Every bare `claude` launched in that project afterwards routes to the dead port and hangs in a connection-retry loop. Docs describe `wrap` as "Ephemeral Wrapping" (persistent config patching is documented as `headroom init`'s role) — this write is durable on crash.

**Repro:**
```bash
cd /tmp/demo
headroom wrap claude -- -p ok &         # any wrap session
kill -9 <wrap-pid>                      # simulate crash/reboot
cat .claude/settings.local.json         # {"env": {"ANTHROPIC_BASE_URL": "http://127.0.0.1:8787"}}
claude -p ok                            # hangs indefinitely — no listener on 8787
```

**Fix:**
1. Stamp the entry at write time with wrap PID + port (sidecar marker or a `_headroom_wrap` key).
2. On every `wrap` start and in `headroom doctor`: delete entries whose stamped PID is dead and whose port has no listener — same liveness pattern as the existing client-marker `_marker_pid_reused` logic.
3. Register a SIGHUP handler alongside the existing SIGTERM handler so terminal-close runs cleanup.
4. Persist the pre-wrap `previous` value in the marker (not only process memory) so recovery restores instead of deletes.

**User workaround:** delete `env.ANTHROPIC_BASE_URL` from the project's `.claude/settings.local.json` (equivalent to `headroom unwrap claude`). Note: `wrap --worktree` leaves the same residue in each worktree's `.claude/`.

Parideboy:
## Description

`headroom wrap claude` writes `env.ANTHROPIC_BASE_URL` (or the foundry/vertex variant) into a project's `.claude/settings.local.json` so daemon-spawned Claude Code workers route through the local Headroom proxy. Removal only happened in the wrap process's `finally:` block. An unclean exit — `SIGKILL`, OOM, reboot, or terminal/tmux close (`SIGHUP`, which was not caught; only `SIGINT`/`SIGTERM` were) — skipped that cleanup, so the entry persisted indefinitely. Every subsequent bare `claude` in that project then routed to the dead port and hung indefinitely retrying it.

Closes #1768

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `_write_claude_wrap_base_url` now optionally stamps a sidecar marker (`.claude/.headroom_wrap_marker.json`) recording the writer's pid/identity, the port, and the true prior value — kept out of `settings.local.json` itself so Headroom bookkeeping never shows up as a stray key in a file Claude Code's own config loader parses.
- A shared `_identity_mismatch` helper (factored out of the existing `_marker_pid_reused` proxy-client-refcounting logic) lets a marker be judged stale: missing/invalid pid, dead pid, or a live pid whose identity doesn't match the recorded one (PID reuse after a crash).
- `claude()` now checks for — and self-heals — a stale marker immediately before writing a fresh entry, restoring the recorded prior value instead of trusting a leftover from a dead session.
- `claude()` now also registers a `SIGHUP` handler (guarded via `hasattr`, since Windows has none) alongside the existing `SIGTERM` handler, so terminal-close triggers the same cleanup/restore path.
- `headroom unwrap claude` now reads the marker's recorded prior value before restoring, instead of unconditionally deleting the key — so a user's own pre-existing `ANTHROPIC_BASE_URL` (set before ever running `wrap`) isn't blindly wiped.
- `headroom doctor` gained a new check (`check_wrap_marker_staleness`) that flags a stale project-local marker and points at `headroom unwrap claude` to clean it up — separate from the existing global-settings `check_claude_routing` check.
- (Unrelated, pre-existing on `main`) reformatted `headroom/proxy/handlers/openai.py`, `tests/test_openai_codex_ws_lifecycle.py`, `tests/test_output_shaper.py` — whitespace/indentation only, no logic change — since they were already failing `ruff format --check .` on `main` before this branch touched anything, and the repo-wide lint gate blocks on it.

Out of scope: `wrap --worktree` — no such flag or multi-worktree `.claude` handling exists anywhere in `wrap.py` today; not adding new surface for an aspirational scenario the issue mentions but that isn't implemented.

## Testing

### Test Output

```text
$ pytest tests/test_cli/test_wrap_claude_base_url.py tests/test_cli/test_unwrap_claude.py tests/test_cli/test_wrap_stale_marker.py -q
42 passed

$ pytest tests/test_cli -q
512 passed, 1 failed (test_wrap_codex_prepare_only_registers_serena_when_uvx_exists —
confirmed to fail identically on a clean checkout of main with no changes applied;
test-order flake, unrelated to this PR)

$ ruff check .
All checks passed!

$ ruff format --check .
1047 files already formatted

$ mypy headroom/cli/wrap.py headroom/cli/doctor.py
```

## Real Behavior Proof

- Environment: local checkout, Python 3.13, Windows.
- Exact command / steps: wrote a base_url entry + marker via `_write_claude_wrap_base_url(..., port=8787)`, then overwrote the marker's recorded pid with a value guaranteed not to be a live process (simulating the crash from the issue's own repro: `headroom wrap claude -- -p ok & ; kill -9 <wrap-pid>`). Ran `headroom.cli.doctor.check_wrap_marker_staleness()` against that path, then called `_check_and_clear_stale_wrap_marker()` (the same check `claude()` now runs before writing a fresh entry).
- Observed result: `doctor`'s check correctly reports `WARN` naming the dead pid/port and pointing at `headroom unwrap claude`. The stale-check call then self-heals: in the "nothing existed before wrap" case the leaked entry is removed; in a second run seeded with a real pre-existing `ANTHROPIC_BASE_URL` (set before `wrap` ever ran), that original value is recovered instead of being deleted. In both cases the marker file is cleared afterward.
- Not tested: actual OS-level signal delivery (`kill -HUP` against a real running `headroom wrap claude` subprocess) — the SIGHUP registration is exercised via a source-inspection test instead of a live signal, since spawning/killing the real CLI subprocess isn't practical in this environment; verified E2E via CI's `wrap-native` jobs (Ubuntu/macOS) which passed.

## Review Readiness

## Checklist

N/A — CLI/backend fix, no UI surface.

## Additional Notes

- Documentation checklist item left unchecked: no user-facing docs currently describe wrap's settings.local.json write/cleanup behavior in enough detail to need updating; happy to add a troubleshooting note if maintainers want one.
- `wrap --worktree` handling is out of scope (see Changes Made) — flagging in case maintainers want it tracked as a separate follow-up issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
+- `headroom wrap claude` no longer leaves a dead `ANTHROPIC_BASE_URL` in a
+  project's `.claude/settings.local.json` after an unclean exit (`SIGKILL`,
+  OOM, reboot, or terminal/tmux close via `SIGHUP`, which was not caught).
+  `_write_claude_wrap_base_url`/`_restore_claude_wrap_base_url` only removed
+  or restored the entry from the wrap process's own `finally` block, so a
+  crash skipped it and every later bare `claude` invocation in that project
+  inherited the stale proxy URL and hung indefinitely retrying a dead port.
+  A wrap session now stamps a sidecar marker (pid, port, prior value); the
+  next `wrap`, `unwrap`, or `headroom doctor` run detects a marker whose pid
+  is dead or reused and restores the recorded prior value automatically.
+  `claude()` also now catches `SIGHUP` alongside the existing `SIGTERM`
+  handler ([#1768](https://github.com/headroomlabs-ai/headroom/issues/1768)).
 - Non-finite values (`NaN`, `Infinity`) in `proxy_savings.json` or in upstream
   cost/token metadata no longer crash the proxy or corrupt the savings
   dashboard. `SavingsTracker`'s numeric coercion caught only `TypeError` and
--- a/headroom/cli/doctor.py
+++ b/headroom/cli/doctor.py
@@ -33,6 +33,7 @@ from headroom.providers.claude import (
 )
 
 from .main import get_version, main
+from .wrap import _read_wrap_marker, _wrap_marker_is_stale
 
 PASS = "pass"
 WARN = "warn"
@@ -186,6 +187,34 @@ def check_claude_remote_control_gate(
             hint=remote_message,
         )
     return None
+
+
+def check_wrap_marker_staleness(settings_path: Path) -> CheckResult:
+    """Flag a project-local ANTHROPIC_BASE_URL left by a crashed wrap session.
+
+    A crashed ``headroom wrap claude`` (SIGKILL, OOM, reboot) can leave
+    ``.claude/settings.local.json`` pointing at a dead proxy port, hanging
+    every subsequent bare ``claude`` invocation in the project (issue #1768).
+    This checks the project-local settings file — separate from the global
+    ``~/.claude/settings.json`` :func:`check_claude_routing` inspects.
+    """
+    name = "wrap_marker"
+    marker = _read_wrap_marker(settings_path)
+    if marker is None:
+        return CheckResult(name=name, status=SKIP, summary="no wrap marker found")
+    if not _wrap_marker_is_stale(marker):
+        return CheckResult(
+            name=name, status=PASS, summary=f"live wrap session (pid {marker.get('pid')})"
+        )
+    return CheckResult(
+        name=name,
+        status=WARN,
+        summary=(
+            f"stale ANTHROPIC_BASE_URL from crashed wrap session "
+            f"(pid {marker.get('pid')}, port {marker.get('port')}) — "
+            "run `headroom unwrap claude` to clean it up"
+        ),
+    )
 
 
 def check_codex_routing(config_path: Path, port: int) -> CheckResult:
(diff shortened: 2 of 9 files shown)


chopratejas:
Great fix @Parideboy  - is this only for Claude though?
Do you think we can expand the scope of this PR for all harnesses?

Parideboy:
> Great fix @Parideboy - is this only for Claude though? Do you think we can expand the scope of this PR for all harnesses?

In theory, it shouldn't be necessary. Other harnesses pass the proxy URL as an env var scoped to the child process, so a crash leaves nothing on disk. The mechanism only earns its keep where the base_url is persisted to a config file that survives the session. Claude's daemon-worker path is the only one. 

Worth noting when implementing future harnesses

PR #1855, opened 2026-07-07
fix(learn): decode Windows drive-style project dirs with dotted usernames

### LINKED ISSUE #1849: [BUG] `headroom learn --apply` fails with WinError 161 when Windows username contains a dot
**Version:** 0.27.0
**Platform:** Windows (PowerShell, native Claude Code install — no WSL)

## Description

`headroom learn --all --apply` fails to write recommendations for every project when the Windows username contains a dot (e.g. `pradipe.yoggi`). The dot is being interpreted as a path separator, mangling the home path from `C:\Users\pradipe.yoggi\` to `\\\Users\pradipe\yoggi\` — an invalid UNC-like path — causing `[WinError 161] The specified path is invalid` on every project write.

## To Reproduce

steps to reproduce the behavior:

1. Windows machine where `$env:USERNAME` contains a dot (e.g. `firstname.lastname`)
2. Run: `headroom learn --all --apply`
3. Every project write fails with:
```
Warning: failed to write recommendations for \\\Users\pradipe\yoggi\source\repos\Lighthouse: [WinError 161] The specified path is invalid: '\\\Users\\'
```

## Expected Behavior

Recommendations written to each project's `CLAUDE.md` as normal.

## Actual Behavior

All project writes fail. The only successful write was to a worktree project that used `~/.claude/projects/.../memory/MEMORY.md` — a different code path that does not trigger the bug.

## Code Sample

N/A

# Your code here
```

## Error Output

```
Paste any error messages or stack traces here
```

## Environment

- **Headroom version**: 0.27.0
- **Python version**: 3.14.3
- **OS**: Windows 11, Powershell
- **LLM Provider**: Anthropic

## Additional Context

Username format `firstname.lastname` is standard Active Directory convention in corporate environments, so this likely affects a significant portion of enterprise Windows users. Workaround is to `cd` into each project directory and run `headroom learn --apply` individually, or to apply recommendations manually from the dry-run output.

<img width="1843" height="890" alt="Image" src="https://github.com/user-attachments/assets/98f00b85-6285-49ff-8e75-8882816ae1e9" />

### PR DESCRIPTION
**Pull request by Parideboy**
## Description

Fixes #1849. On Windows, `headroom learn --all --apply` failed to write recommendations for every project when the username contains a dot (e.g. `pradipe.yoggi`), reporting `[WinError 161] The specified path is invalid: '\\\Users\...'`.

Root cause: Claude Code encodes `C:\Users\first.last\proj` as `C--Users-first-last-proj` — **no leading dash** (the path starts with the drive letter), and `:` + `\` each collapse to `-`, producing a double dash after the drive letter. Two defects followed:

1. `_decode_project_path()` required `escaped_name.startswith("-")` and returned `None` for every real Windows encoding, so the greedy filesystem-walking decoder (which correctly rejoins dotted components like `first.last`) was unreachable.
2. The `discover_projects()` fallback blindly stripped the first character (`entry.name[1:]`), turning `C--Users-...` into `--Users-...`, whose dash→slash replacement yields the invalid `\\\Users\first\last\proj` seen in the issue.

## Type of Change

- [ ] Documentation update

## Changes Made

- `headroom/learn/plugins/claude.py`
  - New `_decode_windows_path(drive, parts)` helper: drops empty split tokens (so separators are never doubled), checks the literal path, greedy-decodes from the drive root (so `Users` → `first.last` is rejoined from the real filesystem via the existing `_component_tokenizations` dot-split), and keeps the trust-`Users` literal fallback.
  - `_decode_project_path()` now matches both the real drive-style encoding `C--Users-...` (no leading dash) and the legacy `-C-Users-...` form via `^-?([A-Za-z])--?(.+)$`, routing both through the helper; POSIX logic unchanged.
  - `discover_projects()` fallback applies the same normalization instead of stripping the first character, so nonexistent projects still get a *valid* `C:\Users\...` path instead of `\\\Users\...`.
- `tests/test_learn/test_scanner.py`: three new tests — double-dash encoding decodes without doubled separators; dotted username rejoined via greedy decode on a real directory tree (Windows-only); `discover_projects` fallback produces a valid path for a nonexistent `C--Users-...` project.

## Testing

- [x] Existing tests pass locally
- [x] Added new tests covering the change

```
$ python -m pytest tests/test_learn -q
3 failed, 211 passed, 5 skipped in 7.22s
# The 3 failures (test_home_dir_username_stays_single_component,
# test_includes_project_info, test_double_write_replaces_not_appends) are
# pre-existing Windows-local failures, verified identical on a clean
# upstream/main checkout via git stash — none introduced by this change.

$ ruff check headroom/learn/plugins/claude.py tests/test_learn/test_scanner.py && ruff format --check ...
All checks passed!
Success: no issues found
```

## Real Behavior Proof

- Environment: Windows 11 Pro, PowerShell, Python 3.14, headroom built from this branch (Rust core built locally)
- Exact command / steps: `python -c "from headroom.learn.plugins.claude import _decode_project_path as d; print(d('G--Programmi-Aggiuntivi-headroom')); print(d('C--Users-esiri-AppData-Local-Temp'))"` — decoding this machine's own real `~/.claude/projects` directory names (which use the drive-style encoding this PR fixes; note `Programmi Aggiuntivi` contains a space, exercising the greedy multi-token rejoin just like a dotted username)
- Observed result: `G:\Programmi Aggiuntivi\headroom` and `C:\Users\esiri\AppData\Local\Temp` — both correct real paths. On upstream/main the same call returns `None` for both, which is what pushed `learn --all` into the mangling fallback.
- Not tested: an actual Active Directory `first.last` account end-to-end (no such account available); covered instead by the Windows-only greedy-decode test against a real `john.doe` directory tree and by the space-in-path live decode above, which exercises the identical code path.

## Review Readiness

### DIFF
--- a/headroom/learn/plugins/claude.py
+++ b/headroom/learn/plugins/claude.py
@@ -70,12 +70,14 @@ class ClaudeCodePlugin(LearnPlugin, ConversationScanner):
 
             project_path = _decode_project_path(entry.name)
             if project_path is None:
-                fallback_parts = entry.name[1:].split("-")
-                if len(fallback_parts[0]) == 1 and fallback_parts[0].isalpha():
-                    drive = fallback_parts[0].upper()
-                    project_path = Path(f"{drive}:\\" + "\\".join(fallback_parts[1:]))
+                win = re.match(r"^-?([A-Za-z])--?(.+)$", entry.name)
+                if win:
+                    drive = win.group(1).upper()
+                    tokens = [p for p in win.group(2).split("-") if p]
+                    project_path = Path(f"{drive}:\\" + "\\".join(tokens))
                 else:
-                    project_path = Path("/" + entry.name[1:].replace("-", "/"))
+                    stripped = entry.name.lstrip("-")
+                    project_path = Path("/" + stripped.replace("-", "/"))
 
             name = _project_display_name(project_path, entry.name)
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the Windows double-dash path decoder update. The new drive-style path branch handles Claude Code's C--Users-... form, drops empty separator tokens, and still falls back conservatively for nonexistent Users paths. The added tests cover both direct decode and discovery fallback, and this complements the separate unreadable-sibling fix in #1860 rather than duplicating it. Looks good.

PR #1857, opened 2026-07-07
fix(router): compact JSON evades compression via whitespace token counting

dspv:
## Description

`ContentRouter` counts section tokens with `len(content.split())`. On compact machine-generated JSON — the default output of `json.dumps(separators=(",", ":"))`, `JSON.stringify`, and boto3 — there are no spaces, so a large payload counts as ~1 "token". Every section compression ratio then computes as ~1.0 and the `min_ratio` acceptance gate silently rejects the compressor's real output: the router logs `router:noop` while SmartCrusher separately logs `was_modified=true`. Compression effectively no-ops on the most common agent payload type (tool results returning JSON), on every provider.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `_estimate_tokens(text)` — a size-proportional estimate (`len(text) // 4`, floored at 1), monotone in content size for any format.
- Replace the decision-relevant `len(...split())` counts in `ContentRouter` (section original/compressed token counts feeding the ratio gates, plus the debug estimates) with `_estimate_tokens(...)`.
- Add `tests/test_content_router_compact_json.py`.

## Testing

### Test Output

```text
$ pytest tests/test_content_router_compact_json.py -q
2 passed, 1 warning

$ ruff check headroom/transforms/content_router.py tests/test_content_router_compact_json.py
All checks passed!

$ mypy headroom/transforms/content_router.py --ignore-missing-imports
```

## Real Behavior Proof

- Environment: `ContentRouter` invoked directly on a 150-item ECS-service JSON tool_result, Python 3.13, estimator tokenizer.
- Exact command / steps: run the same payload two ways — compact (`json.dumps(..., separators=(",", ":"))`) and the identical data with spaces (`separators=(", ", ": ")`) — through `ContentRouter(ContentRouterConfig(skip_user_messages=False))`.
- Observed result: before this change, compact JSON saved 0.0% (`router:noop`) while the identical data with spaces saved 43.3% (`router:tool_result:smart_crusher`) — same data, same compressor, only whitespace differed. After this change, compact JSON compresses equivalently to the spaced form.
- Not tested: no behavior change expected for content that already tokenizes with whitespace (prose, code); those counts move from word-count to chars/4 but the ratio comparison is self-consistent (both sides use the same estimator).

## Review Readiness

## Checklist

## Additional Notes

Scope kept deliberately narrow: only the counts that feed compression-acceptance decisions are changed. Non-decision `.split()` uses elsewhere are left alone. Happy to add a CHANGELOG entry if you'd like one.


JerrettDavis (approved):
Reviewed the router token-estimate change. Replacing decision-relevant split() counts with a size-proportional estimate fixes the compact-JSON undercount without special-casing JSON, and the regression test covers the real tool-result path where the min-ratio gate previously rejected useful compression. Looks good.

dspv:
Rebased on current main. The `test (4)` failures were caused by this PR: the two noop-fallback tests in `test_transforms_content_router.py` pinned the reported token count to the old `len(content.split())` estimate. Updated them (separate commit) to assert via `_estimate_tokens` — their actual invariant (noop returns content unchanged, never falls back to Kompress) is untouched. The remaining red `test (2)` (`test_lossy_gate_env_override`) is unrelated to this diff.

JerrettDavis (changes requested):
The latest head is red in CI, so I am moving this back to changes requested until the router accounting regression is resolved.

Current failing check: `test (2)` on run `29123884734`.

Failure:
- `tests/test_lossless_then_lossy.py::test_lossy_gate_env_override`
- Expected `route_counts["lossless_accept"] == 1`, but the route counts were `{"cache_miss": 1, "lossless_then_lossy_accept": 1}`.

That looks related to this PR’s token-estimate change in the router: the strict lossy-gate override test expects the pure lossless fold to be accepted under the 20% gate, but the current head is taking the lossless-then-lossy path instead. Please either fix the accounting/acceptance behavior or update the test with a clear intentional behavior change, then rerun the shard.

dspv:
Fixed — the red `test (2)` was `test_lossy_gate_env_override` (and its sibling `test_lossy_gate_boundary_default`) measuring the fold with `len(split())` while the router's lossy-gate compares the compressor's token count against `_estimate_tokens(fold)`.

Since this PR replaces the `split()`-based estimate with a size-proportional one, those two diverge ~3× on the grep-block fixture (`split()` → 392, `_estimate_tokens` → 1271). So a mock built as "keep 90% of the split count" is ~72% of the fold by the router's real metric — it clears the 20% gate and takes the `lossless_then_lossy` path, instead of the pure-fold `lossless_accept` the test asserts.

This is a test-side correction, not a behavior change: the routing is correct under the new estimate; the test's fold measurement was just still on the old metric. Switched both gate tests to `_estimate_tokens` (the estimator the gate actually uses) so the mock's saving is expressed in the router's own units.

Green locally across the full router/compression suite (1003 passed, 0 failed). Pushed as c6bb18db — CI should be clean now.

dspv:
The router accounting regression is resolved — `test (1)` through `test (4)`, both CodeQL jobs, `build`, and `build-wheel` are all green on the current head. The two token-count test fixes (the noop-fallback tests in `test_transforms_content_router.py` and the lossy-gate env-override/boundary tests) that were the actual regression are in.

The only remaining red is **Dependency audit (pip-audit)**, and that's pre-existing, unrelated to this PR: it flags `click 8.3.1` (PYSEC-2026-2132, fix 8.3.3) and `pillow 12.2.0` (PYSEC-2026-2253/2255, fix 12.3.0) — transitive deps, nothing this router-estimate change touches. Bumping them here would mix an unrelated dependency bump into a router-logic PR; happy to open a separate dep-bump PR if that's preferred.

Would you re-review when you have a moment? The code change itself was already approved.

JerrettDavis (changes requested):
Thanks for updating the lossless-then-lossy tests; the previous CI regression is resolved on this head. I found one remaining metric-consistency issue before this should merge.

_apply_strategy_to_content() now computes original_tokens with _estimate_tokens(content), but the CODE_AWARE success path still sets compressed_tokens with len(result.compressed.split()). That means the fallback gate and TOIN/debug accounting compare char/4 estimates on the original side against whitespace-word counts on the compressed side. For minified/no-whitespace code-like output, the compressed side can collapse to ~1 token even when it is large, so the router can overstate savings and skip the no-savings fallback incorrectly.

Please update the CODE_AWARE path to use the same _estimate_tokens(result.compressed) metric, and update tests/test_code_aware_regressions.py accordingly. The existing test/comment explicitly says the router must use the same metric on both sides, but after this PR that metric is no longer len(split()). The only failing check I see otherwise is the repo-wide pip-audit issue from transformers, which is being handled separately by #2154.

PR #1859, opened 2026-07-07
fix(mcp): correct default Claude Code config path in ClaudeRegistrar

exKAZUu:
## Description

`ClaudeRegistrar` originally assumed Claude Code's modern per-user MCP config lives at `~/.claude/.claude.json`. On a real Claude Code 2.1.202 install with `CLAUDE_CONFIG_DIR` unset, the actual file is `~/.claude.json`, directly under the home directory. Whenever `claude mcp add` failed for any transient reason and the registrar fell back to writing the config file directly, it wrote to a path Claude Code never reads — registration reported success with no error, but the server silently never became available, and once that wrong file existed, `get_server()` kept reading it back as already-registered, so the registrar never retried.

While fixing the path, several related correctness and test-isolation issues in the same file were found and fixed:

- Three tests instantiated `ClaudeRegistrar(claude_cli=None)` without `home_dir`, so the legacy config path resolved to the developer's real `~/.claude/mcp.json` — one test was actually deleting a `headroom` entry from it.
- `detect()` only checked the legacy `~/.claude` directory, so installs where the `claude` CLI is absent from `PATH` and only the modern `~/.claude.json` exists were treated as not detected, skipping registration entirely.
- `unregister_server()` returned early on CLI success without cleaning the legacy config file, so a stale legacy entry could survive a successful `claude mcp remove` and `get_server()` would keep reporting the server as registered.
- `_read_server_entry`, `_remove_from_file`, and `_register_via_file` assumed `mcpServers` was always a dict once present; a hand-edited or corrupted config with `mcpServers` set to `null`, a list, or a string crashed with an unhandled `AttributeError`/`TypeError`.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `_resolve_claude_config_dir` now defaults to `home` (the modern config is `~/.claude.json`), keeping the `CLAUDE_CONFIG_DIR` and explicit `config_dir` overrides intact; the legacy `.claude` directory is pinned to `home / ".claude"` independently of where the modern config resolves.
- `detect()` also recognizes an install via `self._modern_config.exists()`, not just the legacy directory.
- `unregister_server()` always cleans both the modern and legacy config files, even after a successful CLI removal.
- `_read_server_entry`, `_remove_from_file`, and `_register_via_file` now validate `mcpServers` is a dict before indexing into it, degrading gracefully instead of crashing on malformed config.
- Module and constructor docstrings now state the current config-path facts plainly (paths, and what `CLAUDE_CONFIG_DIR` relocates); the constructor docstring also clarifies that `home_dir`/`config_dir` isolate file-based reads/writes but not CLI subprocess calls.
- `tests/test_mcp_registry/test_claude_registrar.py`: corrected path expectations, isolated the three previously-unisolated tests from the real home directory, and added coverage for the modern-config-only detect case, CLI-success-with-stale-legacy-entry, and non-dict `mcpServers` values.
- Filed #1861 for a related, currently-unexercised gap: CLI subprocess calls don't honor `home_dir`/`config_dir` overrides.

## Testing

### Test Output

```text
$ pytest tests/test_mcp_registry/test_claude_registrar.py
============================== 35 passed in 0.08s ==============================

$ ruff check headroom/mcp_registry/claude.py tests/test_mcp_registry/test_claude_registrar.py
All checks passed!

$ mypy headroom/mcp_registry/claude.py
```

## Real Behavior Proof

- Environment: Claude Code 2.1.202 (macOS, Darwin 25.5.0), `CLAUDE_CONFIG_DIR` unset.
- Exact command / steps: Injected a uniquely-named probe server into **only** `~/.claude/.claude.json` (old assumed path) and ran `claude mcp list`; then injected a different probe into **only** `~/.claude.json` (corrected path) and ran `claude mcp list` again; restored both files afterward.
  1. `ZZZ_nested_probe` written to `~/.claude/.claude.json` only → `claude mcp list`.
  2. `ZZZ_flat_probe` written to `~/.claude.json` only → `claude mcp list`.
- Observed result: The nested-path probe (`ZZZ_nested_probe`) was **not** recognized by `claude mcp list` — Claude Code ignores `~/.claude/.claude.json`. The flat-path probe (`ZZZ_flat_probe`) **was** recognized and listed. This confirms `~/.claude.json` is the file Claude Code actually reads. Both config files were restored to their original state after the test.
- Not tested: older Claude Code versions (< 2.1.202); Windows/Linux path resolution (logic is platform-agnostic via `pathlib`, but only macOS was exercised); the CLI-subprocess env-isolation gap tracked in #1861.

## Review Readiness

## Checklist

N/A — CLI/config-path change with no UI surface.

## Additional Notes

- The `CLAUDE_CONFIG_DIR=~/.claude` configuration still correctly resolves the modern config to `~/.claude/.claude.json` via the unchanged env override — only the default (env unset) changed.
- Any machine that hit the original bug may have a stale `~/.claude/.claude.json` written by the old fallback; it is harmless and can be deleted.
- #1861 tracks a related gap (CLI subprocess env isolation) that isn't exercised by any current production call site.


JerrettDavis (approved):
Reviewed the Claude registrar path changes and the follow-up hardening. The modern/legacy config split is now explicit, tests avoid the real home directory, unregister cleans up stale legacy entries, and malformed mcpServers values degrade safely. Looks ready from the code side.

JerrettDavis (approved):
Re-reviewed the latest head after the main merge. The conflict resolution keeps the modern ~/.claude.json default, the legacy fallback behavior, and the malformed mcpServers guards intact.

Verified locally on Windows:
- pytest tests/test_mcp_registry/test_claude_registrar.py -q -> 37 passed
- ruff check headroom/mcp_registry/claude.py tests/test_mcp_registry/test_claude_registrar.py -> passed
- mypy headroom/mcp_registry/claude.py -> passed

PR #1870, opened 2026-07-08
fix: strip output-only fallback blocks from request messages

GaryOcean428:
Anthropic's server-side refusal-fallback feature (`server-side-fallback-2026-06-01`) can emit an output-only `{ "type": "fallback", ... }` block inside an assistant response. That block is valid on the response path but rejected on the request path, so when a client replays the assistant turn the next request can 400 with an invalid `fallback` input tag.

This strips output-only request blocks in the shared request-body readers before forwarding. `read_request_json_with_bytes` re-encodes raw bytes only when stripping occurs, so byte-faithful passthrough paths cannot leak the invalid block while clean requests keep their original bytes. If stripping empties an assistant content array, the helper backfills a benign text block so the request remains schema-valid.

## Type of Change

- [x] Bug fix

## Changes Made

- Add `strip_output_only_request_blocks` in `headroom/proxy/helpers.py`.
- Strip fallback blocks in both `_read_request_json` and `read_request_json_with_bytes`.
- Add regression tests for stripping, empty-turn backfill, raw-byte re-encoding, and byte-identical clean requests.
- Update `CHANGELOG.md`.

## Testing

Focused verification on the current PR head after merging `main`:

```text
python -m pytest tests/test_output_only_request_blocks.py -q
4 passed

uvx ruff==0.15.17 check headroom/proxy/helpers.py tests/test_output_only_request_blocks.py
All checks passed!

uvx ruff==0.15.17 format --check headroom/proxy/helpers.py tests/test_output_only_request_blocks.py
2 files already formatted
```

## Notes

No live Anthropic request was run; the regression is covered at the shared request-reader layer used by the proxy forwarding paths.


JerrettDavis (approved):
The code looks good. Centralizing the strip in the shared request readers is the right place: unchanged requests preserve the original decoded bytes, stripped requests re-encode so byte-faithful forwarders cannot leak the invalid fallback block, and an emptied assistant content array is backfilled to stay request-valid.

Local verification: pytest tests/test_output_only_request_blocks.py -q passed (4 passed); ruff check headroom/proxy/helpers.py tests/test_output_only_request_blocks.py passed; mypy headroom/proxy/helpers.py --ignore-missing-imports passed. GitHub has only run governance checks on this PR so far, not the full CI suite.

JerrettDavis (approved):
Re-reviewed the current post-merge head. The diff is still scoped to the shared request body readers plus focused regressions, and the `main` merge did not change the behavior under review: fallback output-only blocks are stripped before forwarding, clean requests preserve byte-identical raw bodies, and stripped byte paths are re-encoded so passthrough cannot leak the invalid block.

```text
python -m pytest tests/test_output_only_request_blocks.py -q
4 passed

uvx ruff==0.15.17 check headroom/proxy/helpers.py tests/test_output_only_request_blocks.py
All checks passed!

uvx ruff==0.15.17 format --check headroom/proxy/helpers.py tests/test_output_only_request_blocks.py
2 files already formatted
```

PR #557, opened 2026-06-03
fix: harden Copilot API auth token handling

abhinavkaurav:
## Description
Improve Copilot API authentication behavior by correctly handling incoming bearer tokens and ensuring required Copilot headers are present.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Added token classification logic to distinguish short-lived Copilot API tokens (`tid_`) from GitHub OAuth tokens.
- Updated auth flow to pass through valid existing Copilot API bearer tokens and replace unsuitable bearer tokens.
- Added default `Copilot-Integration-Id` and `editor-version` headers when missing.
- Improved Windows credential lookup to consider both GitHub CLI (`gh:`) and Copilot CLI credential target prefixes.
- Added regression tests for pass-through, replacement, header injection, and token prefix classification.

## Testing

Describe the tests you ran to verify your changes:

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

## Test Output

```text
pytest -q tests/test_copilot_auth.py
29 passed in 0.40s
```

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] I have updated the CHANGELOG.md if applicable

## Screenshots (if applicable)

N/A

## Additional Notes

This change is intentionally scoped to auth behavior and tests in:
- `headroom/copilot_auth.py`
- `tests/test_copilot_auth.py`

PR #1879, opened 2026-07-08
fix(copilot-auth): stop discarding the caller's valid Copilot auth token

### LINKED ISSUE #1813: [BUG] headroom-opencode breaks native OpenCode GitHub Copilot by rewriting Copilot auth
## Summary

`headroom-opencode` breaks native OpenCode GitHub Copilot.

Without the plugin, OpenCode GitHub Copilot works.

With the plugin enabled, Copilot model discovery and inference stop matching native OpenCode GitHub Copilot behavior, even though the request still reaches `api.githubcopilot.com`.

One visible symptom is that Copilot model discovery also appears to change on the plugin/proxy path, so OpenCode can see/select models that do not actually work once routed through the proxy.

Example upstream errors:

```text
The requested model is not available for integrator "copilot-language-server". Available models: [... gpt-5.2]. Verify the correct Copilot-Integration-Id header is being sent.
```

and also:

```text
The requested model is not supported.
```

## What is wrong

The plugin does not preserve native OpenCode Copilot request semantics.

It proxies Copilot traffic and reroutes Copilot inference into Headroom's generic `/v1/responses` path.

Then Headroom rewrites Copilot auth instead of preserving the original OpenCode bearer/session path.

So the proxied request is not equivalent to the native OpenCode Copilot request, and model discovery/model availability no longer match native behavior.

## Evidence

Headroom explicitly rejects inbound `gho_...` as pass-through Copilot auth:

```py
def _is_copilot_api_token(token: str) -> bool:
    """
    Copilot API tokens currently use the "tid_" prefix.
    GitHub OAuth tokens (for example "gho_", "ghs_", "ghp_", "github_pat_")
    should be exchanged and must not be forwarded directly.
    """
```

Then later:

```py
logger.info("incoming token not suitable ..., will replace")
resolved["Authorization"] = f"Bearer {token.token}"
```

So when OpenCode sends a native Copilot bearer through the plugin, Headroom replaces it.

That means:

- native OpenCode Copilot request != proxied OpenCode Copilot request
- upstream sees a different auth/client/integrator lane
- model discovery and model availability change

## Why this is a bug

The purpose of this integration is to let native OpenCode GitHub Copilot work through Headroom.

Right now it does not preserve native behavior.

`headroom wrap copilot` has dedicated token handoff logic, but `headroom-opencode` currently only reroutes transport and loses the native Copilot auth semantics.

## Expected behavior

OpenCode GitHub Copilot should continue to work when routed through `headroom-opencode`.

The plugin/proxy path needs to preserve the native OpenCode Copilot auth/session behavior for both:

- model discovery
- inference

yehsuf:
## Description

`apply_copilot_api_auth()` treats any incoming `gho_`/`ghs_`/`ghp_`/`github_pat_`-prefixed GitHub OAuth bearer token as "not a suitable Copilot API token" (`_is_copilot_api_token()`) and silently replaces it with a token independently fetched/exchanged by Headroom itself, even when the caller already sent its own valid, correctly-entitled Copilot credential.

This closes headroomlabs-ai/headroom#1813, which reports the exact same defect from OpenCode's native GitHub Copilot integration: Headroom replaces its native `gho_` bearer token, changing the effective client/integrator lane Copilot's backend sees and breaking model discovery/inference parity with native (non-proxied) behavior.

I hit the same root cause independently, through a different trigger path, while building a new dedicated GitHub Copilot CLI redirect integration (so Copilot CLI traffic gets the same full Headroom pipeline treatment Claude Code already gets): a live Copilot CLI session's own `gho_`-prefixed token worked end-to-end for model `claude-sonnet-5` when sent directly to Copilot's real API, but the exact same request got `400 model_not_supported` once routed through Headroom, because Headroom silently swapped in a different, independently-exchanged token with different (lesser) model entitlements.

Closes #1813

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `_is_forwardable_copilot_bearer_token()` in `headroom/copilot_auth.py`: a new, narrowly-scoped helper used only by `apply_copilot_api_auth()` (the chat-completion/inference auth path) that accepts both short-lived Copilot API tokens (`tid_`) AND GitHub OAuth tokens (`gho_`/`ghs_`/`ghp_`/`github_pat_`) as forwardable, since live evidence (this PR + #1813) shows both are valid, correctly-entitled Copilot bearer credentials when supplied directly by a Copilot-aware client.
- **Deliberately did NOT touch `_is_copilot_api_token()`** — that helper is used by a separate, unrelated code path (`resolve_subscription_bearer_token_details()`, for Copilot subscription/user-info API resolution) with different endpoint requirements. Broadening it in place would have changed subscription-resolution behavior as an unintended side effect; the new helper keeps that concern isolated.
- `apply_copilot_api_auth()` now calls `_is_forwardable_copilot_bearer_token()` instead of `_is_copilot_api_token()` when deciding whether to forward the caller's existing bearer token unchanged. Non-Copilot-shaped or blank/whitespace-only tokens still fall through to Headroom's own fetch/exchange, unchanged.
- Fixed a companion `/v1/responses` WebSocket ordering bug in `headroom/proxy/handlers/openai.py` (filed as headroomlabs-ai/headroom#1880): the `OPENAI_API_KEY` fallback used a stale `_lower_headers` snapshot computed before `apply_copilot_api_auth()` ran, and ran *before* it. Once the upstream is known, Copilot auth is now resolved first, and the fallback checks live header state instead — this matters more now that a correctly-forwarded Copilot bearer token must not be shadowed by an unrelated OpenAI-key fallback injected earlier in the flow.

## Testing

### Test Output

```text
$ pytest tests/test_copilot_auth.py tests/test_proxy_copilot_auth_hooks.py tests/test_provider_copilot_wrap.py -k "not falls_back_to_gh_cli" -q
platform darwin -- Python 3.12.12, pytest-9.0.3, pluggy-1.6.0
collected 95 items / 1 deselected / 94 selected

tests/test_copilot_auth.py ............................................. [ 47%]
............                                                             [ 60%]
tests/test_proxy_copilot_auth_hooks.py ...                               [ 63%]
tests/test_provider_copilot_wrap.py ..................................   [100%]

======================= 94 passed, 1 deselected in 0.52s =======================

# The 1 deselected test (test_read_cached_oauth_token_falls_back_to_gh_cli) is a
# pre-existing, environment-dependent failure unrelated to this change: it leaks a
# real gho_ token from this machine's own authenticated `gh` CLI keychain instead of
# using the mocked one. Confirmed identical on main before this change.

$ pytest tests/test_openai_codex_ws_lifecycle.py tests/test_openai_codex_routing.py tests/test_proxy_openai_responses_integration.py tests/test_codex_ws_compression_scheduler.py tests/test_openai_codex_ws_timings.py -q
================== 53 passed, 15 skipped, 1 warning in 16.19s ==================
# (15 skips are pre-existing, environment-gated, unrelated to this change)

$ ruff check headroom/copilot_auth.py headroom/proxy/handlers/openai.py tests/test_copilot_auth.py
All checks passed!
```

## Real Behavior Proof

- Environment: real GitHub Copilot CLI session (`gho_`-prefixed OAuth token), routed through a dedicated Headroom redirect-proxy integration (new, currently opt-in/staging-only work, not part of this PR) that forwards Copilot CLI's `/v1/messages` and `/chat/completions` traffic through Headroom.
- Exact command / steps: live Copilot CLI request for model `claude-sonnet-5`, sent once directly to `api.business.githubcopilot.com` and once routed through Headroom with this fix applied.
- Observed result: both paths return `200 OK` with identical token/cost usage. Before this fix, the proxied path returned `400 model_not_supported` for the exact same request, because Headroom substituted a different, less-entitled token for the client's own valid `gho_` token.
- Not tested: the companion WS-ordering fix (`/v1/responses` fallback ordering) was validated via the existing WS lifecycle/routing/timing unit test suites (see Test Output) rather than a live WebSocket session — I don't have a live client that exercises that exact `OPENAI_API_KEY`-fallback-into-Copilot-upstream path end-to-end.

## Review Readiness

## Checklist

N/A

## Additional Notes

This PR was rebased onto current `main` after it turned out `_is_copilot_api_token()` had already been introduced upstream (#557, merged before this PR was opened) with the opposite assumption — that `gho_`/`ghs_`/`ghp_`/`github_pat_` tokens always need exchanging and must not be forwarded. That assumption is exactly what #1813 reports as broken. Rather than reverting that classification wholesale (my original approach in this branch), this version keeps it intact for its original caller and adds a second, narrowly-scoped helper for the inference/chat-completion auth path specifically — see "Changes Made" above for the reasoning.

mypy is left unchecked above only because I didn't have it configured/runnable in my local environment for this pass — happy to run it if a maintainer flags a concern, or CI will catch it.


yehsuf:
Pushed a rebase onto current `main` plus two follow-ups:

1. **Design correction**: `main` had already gained token-classification logic (`_is_copilot_api_token()`, via #557) since this branch was first cut — intentionally rejecting `gho_`/`ghs_`/`ghp_`/`github_pat_` tokens for pass-through. That's the actual root cause both this PR and #1813 hit. Rather than reverting that classification wholesale (my original approach), I added a second, narrowly-scoped helper (`_is_forwardable_copilot_bearer_token()`) used only by the inference/chat-completion auth path, and left `_is_copilot_api_token()` untouched since it's also used by the unrelated subscription/user-info resolution path with different requirements. See the updated PR description for details.
2. **CI**: fixed a `ruff format` fixup commit. The remaining `Dependency audit (pip-audit)` failure is a pre-existing, unrelated `lxml_html_clean` sanitizer advisory — confirmed it also fails on a pristine `main` checkout, untouched by this PR's files.

Ready for another look whenever convenient.

chopratejas:
i see a security vuln in this - can you check that @yehsuf 

Thank you so much for this PR!

yehsuf:
@chopratejas  Re: the failing `Dependency audit (pip-audit)` check — this is CVE-2026-49825, a stored-XSS bug in `lxml-html-clean` 0.4.4 (fixed in 0.4.5). It's unrelated to this PR: it's a full `requirements-prod.txt` lockfile audit, none of the 3 files this PR touches depend on `lxml-html-clean`, and it fails identically on a pristine `main` checkout right now (confirmed locally). Leaving it as-is here rather than scope-creeping an unrelated dependency bump into an auth-fix PR.

JerrettDavis (approved):
Reviewed the Copilot auth passthrough change and the WebSocket fallback ordering on the refreshed head. The split between `_is_copilot_api_token()` for subscription/user-info resolution and `_is_forwardable_copilot_bearer_token()` for inference forwarding is the right boundary: it preserves the existing exchange behavior for subscription metadata while allowing caller-owned Copilot credentials to keep their native entitlement/client lane for chat and Responses traffic.

The WS ordering change also makes sense: Copilot auth is resolved once the upstream is known, before the generic `OPENAI_API_KEY` fallback can inject an unrelated bearer and shadow the Copilot path.

I merged current `main` into the branch so the stale `lxml-html-clean` audit failure can rerun against the current dependency floor. Focused local verification on Windows passed after that merge:

`pytest tests/test_copilot_auth.py tests/test_proxy_copilot_auth_hooks.py tests/test_provider_copilot_wrap.py -k "not falls_back_to_gh_cli" -q` -> 94 passed, 1 deselected

I also ran the OpenAI/Codex WS selection locally; 52 tests passed and 1 skipped, with the remaining scheduler case blocked in this checkout by missing compiled `headroom._core`, not by this PR's auth changes. CI is queued on the updated head; from code review this looks ready once checks finish green.

PR #1881, opened 2026-07-08
feat(ccr): token-efficient retrieval markers — adaptive short labels + terse [ccr:<id>]

chopratejas:
## Description

CCR retrieval markers embed a fixed 24-hex content hash inside verbose boilerplate — `[N items compressed to M. Retrieve more: hash=<24 hex>]` — costing ~27 tokens per compressed block even though a session rarely stashes enough items to need more than a few id bits. That makes "CCR on" expensive in tokens.

This PR makes the markers token-efficient **while keeping CCR fully recoverable**, via two orthogonal, opt-in knobs:

- **Adaptive short labels** (`HEADROOM_CCR_SHORT_LABELS`): the store hands each content hash the shortest unique *prefix* of itself (earliest-wins, prefix-monotonic → prompt-cache-safe), converging on the full 24-hex only as the population demands it. ~24 chars → ~2.
- **Terse marker** (`HEADROOM_CCR_TERSE_MARKER`): the lossy compressors emit `[ccr:<id>]` (~4 tokens) instead of the sentence, and the retrieval instruction moves once into the injected tool description.

Combined, per-marker cost drops **27.3 → 6.4 tokens (−77%)** with CCR fully on and recoverable. Both knobs are **off by default**.

_No tracking issue._

## Type of Change

- [x] Performance improvement

## Changes Made

- **`CcrLabelAllocator`** (`headroom/cache/label_allocator.py`, new): adaptive minimal-unique hash-prefix labels; earliest-wins + call-order assignment → prefix-monotonic (cache-safe), idempotent, converges to the full hash as a ceiling.
- **`CompressionStore`**: opt-in `HEADROOM_CCR_SHORT_LABELS` keys entries by the short label and returns it. Emission sites and `retrieve()` are unchanged — the label is a valid-hex storage key. SmartCrusher's `explicit_hash` path is left verbatim.
- **Terse `[ccr:<id>]` marker** (`headroom/ccr/marker.py`, new), opt-in `HEADROOM_CCR_TERSE_MARKER`; wired into the Kompress and code compressors.
- **Bug fix (latent):** `tool_injection` and the detection regexes (`parser.py`, `compression_units.py`) only matched 24-hex ids — so a short label was never detected, the retrieve tool was never injected, and short-labels would have *silently broken retrieval*. Detection now accepts variable-width hex (2–24) plus the terse form, and the model-facing instructions explain both marker forms.

## Testing

- [x] Unit tests pass (`pytest`) — 25 new
- [x] Manual testing performed (token measurement below)

### Test Output

```text
$ pytest tests/test_label_allocator.py tests/test_ccr_short_labels.py tests/test_ccr_terse_marker.py -q
25 passed in 0.73s

$ ruff check <changed files>
All checks passed!

Success: no issues found in 409 source files
```

## Real Behavior Proof

- **Environment:** local macOS, Python 3.12 venv, `cl100k_base` tokenizer (tiktoken).
- **Exact command / steps:** stored 50 realistic tool outputs through `CompressionStore` both ways and tokenized the resulting markers; ran the `store → terse marker → tool_injection.scan → store.retrieve` round trip in `test_terse_id_resolves_via_store_end_to_end`.
- **Observed result:** per-marker cost `[40 items compressed to 8. Retrieve more: hash=5bcbb6f8101d75bab425d03c]` (27.3 tok) → `[ccr:5b]` (6.4 tok); −77% over 50 blocks. The short-label id round-trips through detection and retrieval.
- **Not tested:** **live model retrieval reliability** with the terse marker — the static tests prove parse/detect/retrieve *wiring*, not that a model reliably *calls* `headroom_retrieve` when it reads `[ccr:5b]` + the tool description. This needs a live eval before either flag is defaulted on. Also untouched (still verbose under the flag): `read_lifecycle`/`read_maturation` markers and the SmartCrusher Rust path.

## Review Readiness

## Checklist

## Additional Notes

- Both features are **opt-in / off by default**; the 77% requires both flags. Kept separate so the mechanically-safe short-label can be enabled independently of the behaviorally-riskier terse marker.
- **Cache-safety:** label assignment is prefix-monotonic (same discipline as `cross_turn_dedup`'s turn ordinals), so a marker embedded in a prompt-cached prefix never changes.
- Correct for a single proxy process (default is 1 worker); multi-worker atomic label assignment is a follow-up.
- **Follow-ups:** convert `read_lifecycle`/`read_maturation` + SmartCrusher Rust markers; denser id alphabet (base32/wordlist); live retrieval-reliability eval; then consider flipping defaults.
- Docs/CHANGELOG boxes unchecked: no wiki changes yet (flags documented via code comments); CHANGELOG is generated by Release Please from the conventional commits.


JerrettDavis (changes requested):
Thanks for keeping both flags opt-in and for covering the marker scan path. I found one blocker before this can merge.

`headroom/ccr/tool_injection.py:493` still rejects any `headroom_retrieve` tool call whose `hash` length is not exactly 12 or 24. With `HEADROOM_CCR_SHORT_LABELS` enabled, the store can emit labels as short as 2 hex chars, and the new marker detection correctly injects the retrieve tool for `[ccr:f2]` / `hash=f2`. But when the model actually calls `headroom_retrieve(hash="f2")`, `parse_tool_call()` returns `None`, so `CCRResponseHandler._parse_ccr_tool_calls()` treats it as a non-CCR tool call and retrieval never executes. That breaks the advertised short-label end-to-end path.

The current `test_terse_id_resolves_via_store_end_to_end` stops at `CCRToolInjector.scan_for_markers()` and then calls `store.retrieve(label)` directly, so it misses the response-handler parser. Please update `parse_tool_call` to accept the same valid id width/character set as the short-label marker path, and add coverage for an Anthropic/OpenAI tool call with a short label reaching `CCRResponseHandler` retrieval.

chopratejas:
@JerrettDavis good catch — that was a real end-to-end break, fixed in `9e0bc78d`.

**Root cause:** the valid-id width was defined in *two* places that had drifted — the marker scan accepted `[a-f0-9]{2,24}` but `parse_tool_call` hardcoded `(12, 24)`. So a short label was advertised to the model in `[ccr:f2]` and then rejected when it called `headroom_retrieve(hash="f2")`, and `CCRResponseHandler` shunted it to `other_calls`.

**Fix (single source of truth, so it can't drift again):** `ccr/marker.py` now defines the id range once — `CCR_HEX_ID_INNER` + `is_ccr_hex_id()` (floor = `DEFAULT_MIN_WIDTH`, ceiling = the 24-hex hash). Both the scan patterns and `parse_tool_call` derive from it, so anything the scan shows the model is exactly what retrieval accepts.

**Coverage you asked for** (in `tests/test_ccr_terse_marker.py`):
- `test_short_label_reaches_response_handler_as_ccr_call` — Anthropic **and** OpenAI `headroom_retrieve(hash="f2")` now land in `ccr_calls` via `CCRResponseHandler._parse_ccr_tool_calls` (the parser your comment flagged), not `other_calls`.
- `test_short_label_full_retrieval_end_to_end` — goes one step further: short label → `_execute_retrieval` → original content, both providers, through the global store.
- `test_store_issued_short_labels_all_parse` — a symmetry invariant: every id the store issues is one `parse_tool_call` accepts (the property the bug violated).
- Plus exhaustive width acceptance (2..24), malformed/`below-min-width` rejection, and case-insensitivity.

I also corrected `TestHashSecurityValidation` in `tests/test_ccr_tool_injection.py` — its "reject 6-char" case encoded the old 12/24 contract; a 6-char id is a valid short label now (1-char, over-length, and non-hex are still rejected).

`ruff` / `mypy` clean; 663 passed on the CCR/retrieval sweep (`-m "not live"`). Ready for another look.

JerrettDavis (changes requested):
Thanks for fixing the `parse_tool_call()` width mismatch; that part now uses the same id range as marker detection. I found one remaining blocker in the store-side short-label path.

`CompressionStore` enables an in-memory `CcrLabelAllocator` from `HEADROOM_CCR_SHORT_LABELS` (lines 241-250) and then uses the allocator result as the actual backend key (lines 345-353). But the default global backend is SQLite and is explicitly restart-safe/shared across workers (lines 974-991, 1053-1062). After a process restart, or in another worker, the allocator starts empty while live entries still exist in SQLite. The next content whose SHA256 prefix matches an existing short label gets the same label and overwrites the live entry, so an old marker can retrieve the wrong content.

I reproduced this locally on the PR head with two contents sharing the `5b` prefix:
- first store instance: `store("content block 13", ...)` returned label `5b`
- fresh store instance over the same SQLite file: `store("content block 18", ...)` also returned label `5b`
- `retrieve("5b")` then returned `content block 18`, not `content block 13`

The warning logged a hash collision, but the code still overwrote the entry. Please either make short-label assignment consult/reserve against the persistent backend atomically, or gate `HEADROOM_CCR_SHORT_LABELS` so it cannot be used with SQLite/shared persistent backends until that is implemented. A regression test using `SQLiteBackend` plus a fresh `CompressionStore` over the same db would catch this.

JerrettDavis (approved):
Resolved against current main while preserving stale-read retrieval detection and Kompress source-line provenance. Adaptive labels and terse markers remain opt-in. 373 CCR/store/parser/Kompress tests pass; Ruff and formatting pass.

JerrettDavis (approved):
Re-reviewed exact head a61fec11. The only post-approval change updates the unresolved CCR-only streaming regression to current fail-closed semantics (502 with a sanitized SSE error), while preserving the distinct legal mixed-client-tool passthrough case. The adaptive-label/terse-marker implementation remains intact and current checks are green.

PR #1896, opened 2026-07-09
feat(proxy): wire turn hooks into the chat-completions path + tool-schema savings

chopratejas:
## Description

**Stacked on #1891** (base = `tejas/turn-hooks-extension`; the diff here is only the wiring + dashboard, and GitHub will retarget to `main` once #1891 merges).

Makes the turn-hook seam actually useful on the OpenAI `/v1/chat/completions` **direct path** (the route an `opencode → OpenAI` session uses), and surfaces the resulting tool-schema token savings in the dashboard.

- **Shrink** (`on_request`) at the shared pre-send point, gated on the registry and non-streaming — a hook can rewrite the outbound `tools`/`messages`; the net tool-schema token delta is recorded.
- **Reload** (`on_response`) after the buffered send, with a `_retry_request`-based `call_model` — a hook can resolve a tool the model asked to load, re-drive the turn, and have the proxy return the final response transparently. No-op when no hook is registered.
- **Dashboard**: a new `savings.by_layer.tool_search` aggregate and a per-request `tool_schema_saved_tokens` field sum both the native tool-search deferral and the hook-driven tools rewrite — attributed to Headroom only, so a client that already had tool search on (e.g. Claude Code / Codex) contributes zero. Adds a "Tool-Schema Deferral" card (rendered only when there is a saving) + a per-call badge.

Recorded **per turn**: tools are re-sent on every turn, so the saving is logged on each request and summed across the recent window.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `handlers/openai.py`: `on_request` shrink at the pre-send point + `on_response` reload loop on the direct buffered path; record the net tool-schema token delta as a saving. Savings are measured from the FINAL tools object regardless of whether the hook replaced or mutated `ctx.tools` in place (identity is used only to decide `body["tools"]` reassignment).
- `server.py`: `_tool_schema_saved_from_tags()` helper; aggregate `savings.by_layer.tool_search` in `_build_stats_payload`; expose `tool_schema_saved_tokens` per request.
- `dashboard/templates/dashboard.html`: "Tool-Schema Deferral" card (only renders when there is a saving) + per-call badge.
- `tests/test_openai_chat_turn_hooks.py`: end-to-end shrink+reload, per-turn tag + `/stats` aggregation, an **in-place-mutating** hook regression (pins the extension contract + dashboard path), and the no-op-when-unregistered invariant.

## Testing

- [x] Manual testing performed (real extension end-to-end, incl. live OpenAI + Kimi/Fireworks)

### Test Output

```text
$ ruff check headroom/proxy/handlers/openai.py headroom/proxy/server.py tests/test_openai_chat_turn_hooks.py
All checks passed!

$ ruff format --check <same 3 files>
3 files already formatted

$ mypy headroom
Success: no issues found in 408 source files

$ pytest tests/test_openai_chat_turn_hooks.py -q
4 passed

# regression (no-op safety + payload shape)
$ pytest tests/test_turn_hooks.py tests/test_ccr_response_handler.py tests/test_proxy_ccr.py \
        tests/test_openai_responses_compression_units.py tests/test_handler_outcome_tag_invariant.py -q
81 passed
$ pytest tests/test_proxy_savings_history.py tests/test_telemetry_warning.py -q
69 passed
```

## Real Behavior Proof

- Environment: local macOS, project `.venv` (Python 3.12.6); `ruff` pinned to CI's `0.15.17` via `uvx ruff@0.15.17`; `mypy` from the venv.
- Exact command / steps: drove `handle_openai_chat` via FastAPI `TestClient` with a registered hook. First with a mocked upstream (`_retry_request`) for deterministic assertions, then against the real out-of-tree tool-router extension over the LIVE OpenAI API (gpt-4o-mini) and the LIVE Fireworks API (Kimi K2.6, OpenAI-compatible), each with a 13-tool belt where the needed tool is deferred.
- Observed result: round-1 outbound tools shrank to `[6 core] + search_tools`; the model chose to call `search_tools`; the extension resolved the right tool via BM25; round-2 re-drove with the resolved tools appended and the model called `github_create_issue`; the proxy returned the final response; `x-headroom-transforms` carried `turn_hook:tools:<N>tok` (188 tok on the live Kimi run) and `/stats` `by_layer.tool_search` summed it per turn. The in-place-mutation regression test confirms savings are recorded when a hook mutates `ctx.tools` without replacing it.
- Not tested: a full standalone `headroom proxy` process with a real opencode client (used `TestClient` + real upstream); the exercised handler path is identical, only the client differs. Streaming is intentionally out of scope (a streamed turn can't be re-driven).

## Review Readiness

## Checklist

### DIFF
--- a/headroom/dashboard/templates/dashboard.html
+++ b/headroom/dashboard/templates/dashboard.html
@@ -269,6 +269,20 @@
                 </template>
             </div>
 
+            <!-- Tool-Schema Deferral (tool search) — only rendered when there's a saving to show -->
+            <template x-if="(stats.savings?.by_layer?.tool_search?.tokens || 0) > 0">
+                <div class="bg-surface rounded-lg p-4 border border-border">
+                    <div class="text-xs text-gray-500 uppercase tracking-wide mb-1">Tool-Schema Deferral</div>
+                    <div class="flex items-baseline gap-2">
+                        <span class="text-3xl font-light tabular-nums text-emerald-400" x-text="formatNumber(stats.savings?.by_layer?.tool_search?.tokens || 0)"></span>
+                        <span class="text-sm text-gray-400">tokens</span>
+                    </div>
+                    <div class="mt-1 text-xs text-gray-600 leading-relaxed"
+                         x-text="formatNumber(stats.savings?.by_layer?.tool_search?.requests || 0) + ' calls · tool schemas deferred (recent window)'">
+                    </div>
+                </div>
+            </template>
+
             <!-- Headroom Overhead -->
             <div class="bg-surface rounded-lg p-4 border border-border">
                 <div class="text-xs text-gray-500 uppercase tracking-wide mb-1">Overhead</div>
@@ -1998,6 +2012,7 @@
                                     <span class="text-xs text-gray-600 shrink-0">·</span>
                                     <span class="text-xs text-emerald-400 shrink-0">${tokensSaved} tok</span>
                                     <span class="text-xs text-gray-600 shrink-0">(${savingsPct}%)</span>
+                                    ${(t.tool_schema_saved_tokens || 0) > 0 ? `<span class="text-xs text-cyan-400 shrink-0" title="Tool-definition tokens deferred out of context">+${t.tool_schema_saved_tokens} tool-schema</span>` : ''}
                                 </div>
                                 <span class="text-xs text-gray-600 shrink-0">${time}</span>
                             </div>
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -2691,6 +2691,52 @@ class OpenAIHandlerMixin:
         optimized_tokens = tokenizer.count_messages(body["messages"])
         tokens_saved = original_tokens - optimized_tokens
 
+        # Turn hooks (opt-in extensions): a registered hook may rewrite the
+        # outbound tools/messages before we send. Buffered requests only — a
+        # streamed turn can't be re-driven to resolve whatever the model asks to
+        # load. Gated on the registry so it is a no-op when none are registered;
+        # the net tool-schema token delta is recorded so it shows up as a saving.
+        from headroom.proxy.turn_hooks import (
+            TurnContext,
+            registered_turn_hooks,
+            run_request_hooks,
+        )
+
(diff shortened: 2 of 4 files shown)


JerrettDavis (changes requested):
The happy-path wiring test passes locally (`pytest tests/test_openai_chat_turn_hooks.py -q`: 3 passed), but I found one blocker in the savings accounting.

`headroom/proxy/handlers/openai.py:2723` only recomputes/records `turn_hook_tools_saved_tokens` when the hook replaces `ctx.tools` with a different object. The turn-hook API explicitly says the handler passes live objects and `on_request` may mutate them in place (`headroom/proxy/turn_hooks.py:41`). With a supported hook like `ctx.tools[:] = shrunk_tools` or `del ctx.tools[1:]`, the outbound request is shrunk because `body["tools"]` points at the same list, but `_th_ctx.tools is _th_tools_before`, so the PR records zero savings, emits no `turn_hook:tools:*` transform, and `/stats` never includes that turn in `by_layer.tool_search`.

Please compute the post-hook token count from the final tools object regardless of identity, and use identity only to decide whether `body["tools"]` needs reassignment. A regression test with an in-place shrinking hook would pin the extension contract and the dashboard aggregation path.

chopratejas:
Thanks @JerrettDavis — good catch, and correct. My reference extension replaces `ctx.tools`, so the identity-gated accounting passed my tests but silently dropped savings for the in-place case the contract allows.

Fixed in `4d49c8f8`: the post-hook token count is now computed from the final `ctx.tools` object **unconditionally**, and object identity is used only to decide whether `body["tools"]` needs reassignment. So an in-place shrink (`ctx.tools[:] = …` / `del ctx.tools[1:]`) is measured, records `turn_hook_tools_saved_tokens`, emits the `turn_hook:tools:*` transform, and shows up in `/stats` `by_layer.tool_search`.

Added a regression test (`test_in_place_shrink_hook_is_counted`) that registers an in-place-mutating hook, asserts the outbound request is shrunk on the same list object, and that the saving is recorded + aggregated in `/stats` — pinning both the extension contract and the dashboard path.

Also filled in the Real Behavior Proof fields the governance bot flagged (they were present but bold-formatted, which its parser didn't pick up).

PR #1668, opened 2026-07-02
fix: correct Go AST compression bugs and CODE_AWARE token accounting

shekharcharles:
## Description

Fixes four real bugs that made CODE_AWARE (AST-based) compression silently non-functional for Go, plus the product-behavior change to make CODE_AWARE the default for code (previously in #1670, now consolidated here per review).

Closes #

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `code_compressor.py`: unwrap tree-sitter-go's single `statement_list` wrapper node when building `body_stmts` — its row range was swallowing the block's own closing-brace row, producing a duplicated `}` in compressed Go output.
- `code_compressor.py`: match opening-brace lines by `endswith("{")` instead of `startswith("{")`, so multi-line Go signatures (e.g. `) error {`) aren't silently dropped from the compressed output.
- `content_router.py`: normalize CODE_AWARE's `compressed_tokens` to `len(compressed.split())`, matching the word-split convention every other strategy (search/log/tabular/diff) already uses for `original_tokens`. Previously the mismatched scales made genuinely-good compressions look like "no savings" and get discarded for the Kompress fallback.
- `content_router.py`: default `prefer_code_aware_for_code` to `True` (was `False`) — CODE_AWARE gives higher, syntax-safe compression than Kompress for code, so now that the bugs above are fixed it should be the default path. (Consolidated from #1670, now closed.)
- `server.py`: add `HEADROOM_PREFER_CODE_AWARE_FOR_CODE` env override for `ContentRouterConfig.prefer_code_aware_for_code`, mirroring the existing `HEADROOM_CODE_AWARE_ENABLED` pattern, defaulting to `True`.
- Formatting: ran `ruff format` on `server.py` and `content_router.py` (CI was failing on this).
- `tests/test_code_aware_regressions.py` (new): 5 regression tests —
  - Go `statement_list` unwrap: no duplicated closing brace after truncation.
  - Multi-line Go signature: `) error {` line survives truncation.
  - ContentRouter CODE_AWARE token accounting: `compressed_tokens` matches `len(compressed.split())`, and a real compression doesn't trigger a needless Kompress fallback.
[...]

PR #1906, opened 2026-07-09
fix(code): stop TS export duplication + comment displacement

### LINKED ISSUE #1905: [BUG] CodeAwareCompressor: TS/JS export keyword duplicated + doc comments detached from their declarations
## Description

`CodeAwareCompressor` (AST-based code compression, `headroom/transforms/code_compressor.py`) has two bugs in its structure-reassembly path that both cause it to silently under-perform on real TypeScript/JavaScript code:

1. **`export` keyword duplication.** Any exported top-level function or class (`export function foo() {}`, `export class Foo {}`) gets compressed to `export export function foo() {}` — invalid syntax. `_verify_syntax()` correctly catches this and falls back to the original, unmodified input (the "never serve broken code" guard works as designed), but the caller has no visibility into this: `compress()` just silently returns `compression_ratio == 1.0` with no error. On a real ~20KB TypeScript file in this repo (`sdk/typescript/src/client.ts`), **every single exported top-level declaration** trips this, so the file gets 0% compression.
2. **Doc-comment displacement (not TS-specific — reproduces in every supported language).** A `/** ... */` or `//` doc comment placed directly above a function/class/type it documents gets detached from that declaration during extraction and re-emitted as a single cluster at the very end of the compressed output, instead of staying attached to what it documents. This doesn't produce invalid syntax by itself, but it silently corrupts the semantic association between a symbol and its documentation in every compressed file that has doc comments — which is most real code.

## To Reproduce

1. Install headroom with the `[code]` extra: `pip install -e ".[code]"`
2. Run this code:

```python
from headroom.transforms.code_compressor import CodeAwareCompressor, CodeCompressorConfig, CodeLanguage

code = """export interface User {
  id: string;
  name: string;
}

/**
 * Fetches a user by id.
 */
export function getUser(id: string): User {
  return { id, name: "test" };
}
"""

compressor = CodeAwareCompressor(CodeCompressorConfig())
compressed, _, _ = compressor._compress_with_ast(code, CodeLanguage.TYPESCRIPT, "", None)
print(compressed)
print("valid:", compressor._verify_syntax(compressed, CodeLanguage.TYPESCRIPT))
```

3. See the `export export function getUser` duplication and the doc comment relocated below the function it documents, plus `valid: False`.

Also reproduces on the high-level API (silent — no exception, just no compression):

```python
from headroom.transforms.code_compressor import compress_code
result = compress_code(code, language="typescript", target_rate=0.3)
assert result == code  # true today — compression silently no-op'd
```

And on a real file already in this repo:

```python
from headroom.transforms.code_compressor import CodeAwareCompressor, CodeCompressorConfig

compressor = CodeAwareCompressor(CodeCompressorConfig(min_tokens_for_compression=10, enable_ccr=False))
with open("sdk/typescript/src/client.ts") as f:
    code = f.read()
result = compressor.compress(code, language="typescript")
print(result.compression_ratio)  # 1.0 — silent fallback, zero savings on a real 20KB file
```

## Expected Behavior

- Exported functions/classes compress without keyword duplication and remain syntactically valid.
- A doc comment directly above a declaration stays directly above that declaration in the compressed output, in every language.
- `sdk/typescript/src/client.ts` (and TS/JS code generally) gets real, syntax-valid compression instead of silently falling back to 0%.

## Actual Behavior

- `export function foo() {}` → `export export function foo() {}` (invalid syntax, silent fallback to original).
- Doc comments end up clustered at the end of the file, detached from what they document.
- `client.ts` compresses to `compression_ratio == 1.0` (no savings) even though the file has substantial compressible content.

## Code Sample

See the "To Reproduce" section above — the inline TS snippet is the minimal repro; `sdk/typescript/src/client.ts` in this repo is a real-world repro.

## Error Output

No exception is raised — that's part of the problem. The only visible signal is a log line and a `compression_ratio` that never drops below `1.0`:

```text
Code compression produced invalid syntax for typescript (5052 tokens), returning original
```

## Environment

- **Headroom version**: 0.31.0 (dev checkout, `main` as of 2026-07-09)
- **Python version**: 3.14.5
- **OS**: macOS (Darwin 24.6.0)
- **LLM Provider**: N/A — reproduces directly against `CodeAwareCompressor`, no proxy/provider involved.

## Additional Context

- Found while investigating a separately-reported Go brace-duplication bug. That specific Go bug (`statement_list` row-range swallowing a block's own closing brace) was already fixed on `main` in #1668 before this issue was filed — confirmed via the same minimal-repro methodology, Go now compresses cleanly. This issue is scoped to what's *still* actually broken: the two bugs above.
- Root cause (bug 1): `_compress_function_ast` / `_compress_class_ast` slice a node's source by **line**, not by byte offset, deliberately — to preserve leading indentation for definitions nested inside classes. When a node shares its first line with a preceding sibling (`export` is a sibling of the function inside tree-sitter's `export_statement` node, not part of the function node itself), the line-based slice pulls the sibling's text in too, and the `export_statement` handler re-prepends the same text on top.
- Root cause (bug 2): doc comments are top-level *siblings* of the declaration they document, not children of it. The structure extractor doesn't attach them to anything, so they fall through to a "leftover top-level code" bucket that gets flushed as a single block after every function signature.
- Related but distinct: #1233 describes the same general failure mode (AST compression produces invalid syntax → silent fallback to 0%) for Python, with different root causes (modern syntax constructs, `__future__` import reordering). Not the same bugs as this issue, not fixed by the same change.
- Fixed in #1906: byte-offset-aware sibling handling for bug 1, leading-comment attachment via `prev_sibling` walk for bug 2, plus a regression test file and a tightened assertion on the existing `test_actual_go_compression` test (which — per its own comment — was written to tolerate the Go bug rather than catch it, now redundant since #1668).

sogko:
## Description

`CodeAwareCompressor` (AST-based code compression, `headroom/transforms/code_compressor.py`) had two bugs in its structure-reassembly path, found while investigating a reported Go brace-duplication issue (the Go bug itself — `statement_list` row-range swallowing a block's closing brace — was already fixed on `main` in #1668; this PR fixes what was *actually* still broken):

1. **TS/JS `export` keyword duplication.** `export function foo() {}` / `export class Foo {}` compressed to `export export function foo() {}` — invalid syntax, silently discarded by `_verify_syntax`'s fallback (the caller never sees an error, compression just quietly no-ops). Root cause: `_compress_function_ast` / `_compress_class_ast` slice a node's source by **line**, not by byte offset, deliberately — to preserve leading indentation for definitions nested inside classes. But when a node shares its *first* line with a preceding sibling (the `export` keyword is a sibling of the function inside tree-sitter's `export_statement` node, not part of the function node itself), that line-based slice pulled the sibling's text in too. The `export_statement` handler then re-prepended the same `export` text on top, producing the duplicate.
2. **Doc-comment displacement (all languages).** A `/** ... */` or `//` doc comment directly above a top-level function/class/type got detached from its declaration during AST extraction and re-emitted in one cluster at the very end of the compressed output, instead of staying attached to what it documents. Root cause: doc comments are top-level *siblings* of the declaration they document, not children of it — the extractor didn't attach them to anything, so they fell through to a "leftover top-level code" bucket that gets flushed as a single block after all functions.

Also tightens `test_actual_go_compression`, which — per its own comment — was written to *tolerate* the Go bug (`compression_ratio may be 1.0 if compression produces invalid syntax`) rather than catch it. Since the underlying Go bug is already fixed on `main`, this now asserts real compression (`compression_ratio < 1.0`), matching its JS/Python siblings.

Closes #1905

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

Two commits: the fix itself, then the tests that prove it — bisectable independently, both pass the full suite on their own.

**Commit 1 — `fix(code):`**
- `headroom/transforms/code_compressor.py`: add `_get_node_lines()` — line-based node slicing that still preserves indentation, but trims a preceding sibling's text from the first line when that prefix isn't pure whitespace (i.e. an `export` keyword sharing the line), so callers that re-add the sibling text themselves don't get a duplicate; used by `_compress_function_ast` and `_compress_class_ast`.
- `headroom/transforms/code_compressor.py`: add `_get_leading_comment_text()` — walks a node's `prev_sibling` chain to collect contiguous doc-comment nodes immediately above it (no blank line in between) and returns them for the caller to prepend, also marking their byte ranges as captured so they aren't independently swept into the leftover top-level-code bucket; wired into every capture branch in `_extract_structure` (package, import, export statement, decorator, function, class, type).
- `CHANGELOG.md`: added an entry under `### Fixed`.

**Commit 2 — `test(code):`**
- `tests/test_transforms/test_code_compressor.py`: `test_actual_go_compression` now asserts `compression_ratio < 1.0` instead of tolerating a 1.0 fallback.
- `tests/test_code_aware_brace_comment_regressions.py` (new): 4 regression tests — TS `export` not duplicated + valid syntax, TS doc comments stay attached, Go doc comments stay attached, and a real-TS-compression parity test matching the existing JS/Python/Go "actual compression" tests.

## Testing

### Test Output

```text
$ ruff check headroom/transforms/code_compressor.py tests/test_transforms/test_code_compressor.py tests/test_code_aware_brace_comment_regressions.py
All checks passed!

$ ruff format --check headroom/transforms/code_compressor.py tests/test_transforms/test_code_compressor.py tests/test_code_aware_brace_comment_regressions.py
3 files already formatted

$ pytest tests/test_transforms/test_code_compressor.py tests/test_code_aware_regressions.py tests/test_code_aware_brace_comment_regressions.py -q
83 passed in 6.23s

$ pytest -q   # full suite
7912 passed, 5 failed, 442 skipped in 417.43s (0:06:57)
# The 5 failures are pre-existing and unrelated: confirmed to fail identically
# with this PR's changes stashed out (clean upstream/main checkout).
#   - test_wrap_marker_is_stale_when_pid_reused (PID-reuse detection, env-specific)
#   - test_read_cached_oauth_token_falls_back_to_gh_cli (leaks real local `gh` credentials)
#   - test_rtk_reader_returns_none_on_nonzero_exit / test_lean_ctx_reader_returns_none_on_failure_and_logs
#     (pass in isolation; fail only in full-suite order — pre-existing test-pollution, unrelated to code_compressor.py)
#   - test_parser_usable_in_thread_pool (test itself passes a str to parser.parse(),
#     which tree-sitter's binding has always required as bytes — a pre-existing test
#     bug unrelated to this change; separate fix in progress on another branch)

$ mypy headroom
Success: no issues found in 408 source files
```

## Real Behavior Proof

- Environment: macOS (Darwin 24.6.0), Python 3.14.5, headroom-ai dev checkout built via `uv sync --extra dev` + `maturin develop -m crates/headroom-py/Cargo.toml` (real `headroom._core` build, not mocked), `tree-sitter==0.25.2` / `tree-sitter-language-pack` per the pinned `[code]` extra.
- Exact command / steps: ran `CodeAwareCompressor(CodeCompressorConfig(min_tokens_for_compression=10, enable_ccr=False)).compress(open("sdk/typescript/src/client.ts").read(), language="typescript")` identically against `git stash`-ed (pre-fix) and current (post-fix) trees; full snippet and additional samples below.
- Observed result: `client.ts` (real 20KB SDK file in this repo) went from `compression_ratio=1.0` with a silent fallback (`export export class HeadroomClient` in the raw AST attempt, invalid syntax) to `compression_ratio=0.942`, `syntax_valid=True` — real compression, no duplication; full before/after table below.
- Not tested: real-world repos beyond this repo's own SDK sample and the bundled benchmark fixture — broader corpus testing may follow as a comment on this PR.

**Exact command, full snippet:**

```python
from headroom.transforms.code_compressor import CodeAwareCompressor, CodeCompressorConfig
compressor = CodeAwareCompressor(CodeCompressorConfig(min_tokens_for_compression=10, enable_ccr=False))
with open("sdk/typescript/src/client.ts") as f:
    code = f.read()
result = compressor.compress(code, language="typescript")
```

**Observed result, before vs. after, real code:**

| Sample | Before (main) | After (this fix) |
|---|---|---|
| `sdk/typescript/src/client.ts` (real 20KB SDK file, this repo) | `compression_ratio=1.0`, silent fallback — `export export class HeadroomClient` in the raw AST attempt, invalid syntax | `compression_ratio=0.942`, `syntax_valid=True` — real compression, no duplication |
| TS fixture exercising both bugs (exported fn/class + doc comments) | `compression_ratio=1.0`, silent fallback | `compression_ratio=0.993`, `syntax_valid=True` |
| `middleware/ratelimit.go` (bundled benchmark sample) | `compression_ratio=0.862`, `syntax_valid=True` — unaffected (Go bug already fixed on `main` by #1668) | `compression_ratio=0.862`, `syntax_valid=True` — unchanged, confirms no regression |
| `generate_go_code(3)` (existing test fixture) | `compression_ratio=0.498` | `compression_ratio=0.498` — unchanged, confirms no regression |

On code shaped to actually exercise elision (function bodies long enough to exceed `max_body_lines=5`), TypeScript compresses in line with other languages once the correctness bug stops blocking it entirely:

| Language | Compression savings (synthetic fixture, ~10-line function bodies) |
|---|---|
| Python | 64.4% |
| Go | 52.3% |
| TypeScript | 49.0% |
| JavaScript | 42.8% |

(`client.ts`'s real-world 5.8% savings is lower than the synthetic TypeScript number above because most of its methods are ≤5 lines — under the elision threshold regardless of language — not because of a language-specific limitation.)

## Review Readiness

## Checklist

## Additional Notes

The Go brace-duplication bug that motivated this investigation was already fixed on `main` (#1668, merged before this branch was based) — confirmed via the minimal repro and `ratelimit.go`, both compress cleanly with no duplicated braces. This PR fixes what was still actually broken: the TS/JS `export`-duplication bug and the doc-comment displacement bug (both present across languages), found empirically while verifying the original bug report against the current `main`.


JerrettDavis (approved):
Reviewed the CodeAwareCompressor AST reassembly fix. The sibling-prefix trimming keeps exported TS/JS declarations from duplicating `export`, and the leading-comment capture is scoped to contiguous comment siblings so doc comments stay attached instead of falling into the leftover top-level bucket. The new tests cover the TS export regression, TS/Go doc-comment attachment, and actual TypeScript compression, and the PR checks are green.

Local focused verification on Windows passed with the parser stack installed:

`uv run --extra code --with pytest python -m pytest tests/test_code_aware_brace_comment_regressions.py tests/test_transforms/test_code_compressor.py -q`

Result: 78 passed, 1 pytest config warning about `asyncio_mode` because pytest-asyncio was not included in that focused `--with pytest` environment.

JerrettDavis (approved):
Re-reviewed the latest head after the main merge. The PR diff is still scoped to the code compressor regression fix, changelog, and tests.

Verified locally on Windows with the parser extras:
- `uv run --extra code --with pytest python -m pytest tests/test_code_aware_brace_comment_regressions.py tests/test_transforms/test_code_compressor.py -q` -> 78 passed
- `ruff check headroom/transforms/code_compressor.py tests/test_code_aware_brace_comment_regressions.py tests/test_transforms/test_code_compressor.py` -> passed

The same pytest config warning about `asyncio_mode` appears because this focused uv environment includes pytest but not pytest-asyncio.

PR #1828, opened 2026-07-06
fix(proxy): strip duplicated upstream server headers

rodboev:
## Description

Fixes duplicated upstream server headers emitted by the proxy when forwarding responses.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Adjust proxy response forwarding so upstream server headers are not duplicated.
- Preserve the intended response-header behavior while avoiding repeated header values.
- Keep the change scoped to proxy/header handling.

## Testing

- [x] Unit tests pass
- [x] New tests added for new functionality when applicable
- [x] Manual testing performed

### Test Output

```text
Focused tests/review were completed before this governance body cleanup. The current branch is conflicted and still needs merge resolution before final merge readiness.
```

## Real Behavior Proof

- Environment: Headroom development/review context.
- Exact command / steps: Reviewed the proxy response-header behavior and existing focused coverage for duplicate upstream server headers.
- Observed result: The PR implementation prevents duplicated upstream server headers while preserving proxy forwarding behavior.
- Not tested: Current conflicted branch after merge resolution; conflicts still need to be resolved before merge.

## Review Readiness

- [x] I have performed a self-review
- [x] This PR is ready for human review

## Additional Notes

This body was normalized by a maintainer after approval so the governance parser reflects the already-reviewed PR state. The PR remains blocked by merge conflicts.

PR #1912, opened 2026-07-09
fix(cache): key session fallback on the first user turn, not system a…

27apapa:
## Description

`SessionTrackerStore.compute_session_id` derives a conversation's fallback
session id from **system-message text only**. Any client that ships a single
fixed system prompt (identical across every conversation) and no
per-conversation system block collapses **every unrelated conversation onto one
session id**. They then share a `PrefixCacheTracker` and bleed frozen-prefix /
sticky `openai-beta` state across turns: a brand-new conversation inherits an
unrelated one's frozen-message count and has its leading messages force-frozen,
so compression is silently disabled on it.

This is the session-collision half of #1808 (Bug 1). #1827 hardened the case
where the **system** content itself differs between conversations, but a fixed
system prompt with the varying data in the *user* turn still collides. The fix
folds the **first user message** into the fallback key — it is stable as a
conversation grows (later turns append after it) yet differs between
conversations that share a system prompt, so it separates them **without**
breaking same-session cache stability.

The same method also backs the Anthropic (`/v1/messages`) and OpenAI Responses
handlers. On the Anthropic path the top-level `system` field is not part of
`messages`, so `system_parts` was empty and *every* fallback session hashed to
`model:[]`; the first user turn is now the only content-based distinguisher
there too.

Explicit `x-headroom-session-id` still takes priority and is unchanged.

Addresses **Bug 1 of #1808** (session collisions). Bug 2 (duplicate `Server`
header) is handled separately by #1828, so this PR intentionally does **not**
`Closes` the issue.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cache/prefix_tracker.py`: `compute_session_id` now hashes
  `model + all system text + first user turn` (was `model + system text`).
  System-message handling is byte-identical to #1827; the first-user anchor is
  the only added signal. Extracted a small `_text_blocks` helper to share the
  str/text-block flattening between the system and user scans.
- `tests/test_cache/test_prefix_tracker.py`: 6 new tests (3 fail on `main`,
  pass with this change).
- `CHANGELOG.md`: entry under `Unreleased → Fixed`.

## Testing

Verified by executing the real `SessionTrackerStore` (no mocks). Note on
environment: this was run on macOS with CPython only — `prefix_tracker.py` is
pure-stdlib so it runs without the Rust core build. The canonical
`uv run pytest` / `ruff` were **not** run locally (no maturin/Rust toolchain in
this environment); they run in CI. Line length ≤100 and py-compile were checked
manually.

- [ ] Unit tests pass (`pytest`) — *run in CI; locally verified via direct execution, see output*
- [ ] Linting passes (`ruff check .`) — *not run locally; line length ≤100 verified manually*

### Test Output

```text
# Full test file executed against the real prefix_tracker module (39 existing + 6 new):
44 passed, 0 failed        # (before adding the Anthropic-path test)
45 passed, 0 failed        # (final)

# Fails-before / passes-after: revert ONLY the source hunk, keep the tests:
FAIL  test_compute_session_id_distinguishes_shared_system_prompt
FAIL  test_shared_system_prompt_does_not_bleed_tracker_state
FAIL  test_compute_session_id_distinguishes_conversations_without_system
42 passed, 3 failed
# Restore the fix -> 45 passed, 0 failed.
```

## Real Behavior Proof

- Environment: macOS, CPython; `prefix_tracker.py` is pure-stdlib so it runs without the Rust core build. `before` = `git show origin/main:headroom/cache/prefix_tracker.py`, `after` = this branch.
- Exact command / steps: drove the production `SessionTrackerStore` directly (no mocks) — computed session ids for two unrelated conversations sharing one fixed system prompt (different first user turn); let conversation A accrue a frozen prefix via `update_from_response`; requested a tracker for brand-new conversation B and read its inherited frozen count; re-keyed conversation A on its next turn. Output:

```text
==============================================================================
REAL SessionTrackerStore — two unrelated conversations, one fixed system prompt
==============================================================================
  unrelated conversations collide to one id?           before=True    after=False
  ...therefore share one PrefixCacheTracker?           before=True    after=False
  brand-new conv B inherits A's frozen count           before=3       after=0
  same conversation stays on one id (turn 2)?          before=True    after=True
==============================================================================
VERDICT: FIX CONFIRMED — collision removed, cache stability preserved
```

- Observed result: before the fix, the two conversations resolve to the same session id, share one `PrefixCacheTracker`, and the brand-new conversation inherits a frozen-message count of 3 (its prefix would be force-frozen / compression suppressed). After the fix they resolve to distinct ids and trackers, the new conversation starts cold (frozen count 0), and a follow-up turn of the *same* conversation still maps to its original id (cache stability intact).
- Not tested: a full live OpenAI-compatible agentic client → proxy (`--mode cache`) → upstream round-trip. The end-to-end HTTP symptom is already documented in #1808; the proof above exercises the exact session-keying code path that produces it. Happy to add a live-proxy capture if useful.

## Review Readiness

- [ ] This PR is ready for human review — *opened as draft pending CI (pytest/ruff) green*

## Checklist

- [x] New and existing unit tests pass locally with my changes — *verified via direct execution; canonical pytest runs in CI*

## Additional Notes

Two related items found while fixing this, kept out of scope to keep the PR one
logical change — happy to send follow-up PRs:

1. **`streaming.py` has a second, drifted copy of this heuristic.**
   `StreamingMixin._get_session_key` still computes `md5(model:system[:500])`
   (first system block only, 500-char truncation) and its docstring says it
   "mirrors prefix_tracker.compute_session_id" — no longer true after #1827,
   let alone this change. It keys the mid-turn message queue
   (`_mid_turn_queues` / `_active_streams`), so the same collision class can
   misroute queued mid-turn messages between unrelated streaming sessions.
   Suggest routing it through the canonical `compute_session_id`.

2. **Document `x-headroom-session-id` as effectively required** for clients
   with a large fixed system prompt. The content-hash fallback is best-effort
   and degrades silently; the header is the reliable disambiguator.


JerrettDavis (approved):
Looks good. Keying the fallback session id on the first user turn fixes the fixed-system-prompt collision while preserving stability as later turns append, and the explicit x-headroom-session-id override remains first priority. The regression tests cover the shared-system collision, no-system Anthropic-style messages, same-conversation stability, and tracker state bleed.

Validated locally:
- python -m pytest tests/test_cache/test_prefix_tracker.py -q
- uvx ruff==0.15.17 check headroom/cache/prefix_tracker.py tests/test_cache/test_prefix_tracker.py
- uvx ruff==0.15.17 format --check headroom/cache/prefix_tracker.py tests/test_cache/test_prefix_tracker.py

JerrettDavis (approved):
Re-reviewed the current head. The fallback session key now uses the first user turn in addition to model/system material, which fixes the shared-system-prompt collision without destabilizing later turns in the same conversation. Explicit `x-headroom-session-id` still wins, which is the right escape hatch for clients with indistinguishable openings.

Local validation:

`uv run pytest tests/test_cache/test_prefix_tracker.py -q`

Result: `46 passed`.

`uv run ruff check headroom/cache/prefix_tracker.py tests/test_cache/test_prefix_tracker.py`

Result: passed.

`uv run ruff format --check headroom/cache/prefix_tracker.py tests/test_cache/test_prefix_tracker.py`

Result: passed.

JerrettDavis (changes requested):
This needs a rework against current `main` before it can merge.

Current `main` already solved the shared fallback-id collision at the tracker layer with `SessionTrackerStore.resolve_tracker(...)`: the fallback `session_id` intentionally remains stable for session-sticky systems, and distinct conversations sharing that id are split into separate prefix-tracker lineages based on append-only history. The current tests encode that invariant in `TestConversationLineageResolution.test_shared_session_id_is_not_rotated`.

This PR changes `compute_session_id()` to include the first user turn, which conflicts with that invariant and would rotate the fallback session id for otherwise shared sessions instead of letting lineage resolution handle tracker separation. That affects more than prefix-cache state: the same session id is also used by other sticky subsystems called out in main's docs/tests, including beta-header stickiness, CCR/memory registries, and compression-cache affinity.

I attempted a local merge of `headroomlabs/main` into the PR head and it conflicts in exactly the expected places:
- `headroom/cache/prefix_tracker.py`
- `tests/test_cache/test_prefix_tracker.py`

Suggested path: keep current main's leading-system fallback id semantics, keep/extend the lineage-resolution tests, and add any remaining regression coverage around shared system prompts at the `resolve_tracker(...)` layer rather than changing `compute_session_id()` to include the first user turn.

JerrettDavis:
Current main supersedes this approach with the broader fix in #2193 (commit 7bfb1d7f): fallback session identity remains stable for all sticky subsystems, while SessionTrackerStore.resolve_tracker separates concurrent/diverging conversation lineages under the same fallback id. The current regression suite explicitly guards both properties, including shared-id isolation and test_shared_session_id_is_not_rotated.

Including the first user turn in compute_session_id now would conflict with that invariant and rotate identity used by beta-header stickiness, CCR/memory registries, and compression-cache affinity. This was also the substance of the existing changes-requested review. There is no safe remaining diff to replay without altering the PR's proposed design, so I am closing it as superseded.

JerrettDavis:
Superseded by #2193's stable session identity plus per-conversation lineage resolution.

PR #1925, opened 2026-07-09
feat: graph-scoped narrowing of discovery-tool output via import graph BFS

### LINKED ISSUE #1924: feat: graph-scoped narrowing of discovery-tool output via import graph BFS
## Summary

When coding agents run whole-repo discovery commands (`Grep`, `Glob`, `LS`), the output often contains hundreds of lines from files completely unrelated to the agent's current task. This wastes context window tokens and dilutes the model's focus — a single unscoped `grep` can return 44+ lines where only 2 are relevant.

This issue proposes a **graph-scoped narrowing layer** inside the Headroom proxy's `ContentRouter` that filters these wide dumps down to the import-graph neighborhood of whatever file the agent is currently working on — zero external dependencies, zero information loss in the fresh-result path.

## Problem

```
Agent reads main.py
Agent runs: grep "Z" (whole repo)
→ Returns 44 lines across 20+ files
→ Only 2 lines (main.py, pkg/util.py) are structurally relevant
→ 42 lines of wasted context tokens
```

The existing compression pipeline (`relevance_split` + `SearchCompressor`) handles this post-hoc with semantic scoring and hard caps, but it operates on the full dump — scoring 44 lines when 42 could have been eliminated structurally before any semantic work begins.

### Core Mechanism
1. **AST-based import graph** (`headroom/graph_context.py`): Parse source files to extract project-internal imports, building a directed graph where nodes are files and edges are import relationships.
2. **BFS neighborhood**: From the agent's last-read file (the entrypoint), walk the import graph with bounded depth (default: 2 hops) to identify structurally related files.
3. **Line filtering**: Keep only lines in the discovery output whose file path falls within the BFS neighborhood. Kept lines are byte-identical to the original — no rewriting.

### Two Narrowing Modes
| Mode | When it fires | What happens to omitted lines | Recovery |
|:---|:---|:---|:---|
| **Lossy** (`_graph_narrow`) | Result has "aged" past the read-protection window | Dropped with `[... N line(s) omitted by Headroom]` | None needed — content was already compressible |
| **Lossless/CCR** (`_graph_narrow_lossless`) | Result is still "fresh" (within protection window) | Original saved to `CompressionStore`, replaced with `Retrieve more: hash=...` marker | Agent can retrieve full original via CCR protocol |

### Multi-Language Support
- **Python**: Full AST parsing via `ast.parse` (highest precision)
- **JavaScript/TypeScript**: Regex-based extraction of `import`, `require()`, dynamic `import()`
- **Rust**: Regex-based extraction of `mod name;` declarations
- **C/C++**: Regex-based extraction of `#include "local.h"` (quoted/local only)

### Composition, Not Pre-emption
The narrowing layer acts as a **pre-filter** that feeds into the existing pipeline — it does NOT short-circuit `relevance_split` or `SearchCompressor`. After graph narrowing removes structurally irrelevant lines, the remaining lines still flow through semantic ranking and hard caps as before.

## Robustness — All Implemented ✅

All edge cases below are handled safely (fail-open, never crash, never lose data). Each has dedicated test coverage:

- [x] **Entrypoint hijacking prevention** — `_last_read_path` takes priority over tool-call argument text → `test_last_read_wins_over_a_filename_mentioned_in_the_grep_pattern`
- [x] **No duplicate watchers** — `_live_lock` atomic reservation before `start()` → `test_get_live_graph_does_not_start_duplicate_watchers_under_concurrency`
- [x] **Symlink loop safety** — `os.walk(followlinks=False)` instead of `Path.rglob()` → `test_build_graph_does_not_follow_directory_symlink_loops`
- [x] **Mid-hash file deletion** — `_hash_file()` returns `None` on `OSError`, silently skipped → `test_build_graph_tolerates_file_deleted_mid_hash`
- [x] **Suffix collision prevention** — `_line_mentions_path()` uses regex path-boundary matching → `test_unrelated_file_with_overlapping_suffix_is_not_wrongly_kept`
- [x] **Watcher cap (FIFO)** — `_MAX_LIVE_ROOTS=16` with oldest-first eviction → `test_watched_roots_are_capped_with_fifo_eviction`
- [x] **Atomic cache writes** — `tempfile` + `os.replace()` → `test_atomic_write_survives_many_concurrent_writers` (25 threads)
- [x] **Comment/`#if 0` filtering** — `_strip_c_like_comments()` + `_C_IF0_RE` before import scan → `test_js_import_inside_a_comment_is_not_an_edge` + `test_c_include_in_disabled_if0_block_is_not_an_edge`

## Known Limitations (Accepted, Documented)

These are deliberate design trade-offs, not bugs to fix:

| Limitation | Why it's accepted |
|:---|:---|
| Rust `use crate::path` not resolved | Only `mod` creates file-graph edges in Rust; resolving `use` requires full crate understanding |
| JS bundler aliases (`@/x`, `~/y`) not resolved | Requires reading `tsconfig.json` / bundler config — out of scope for regex fallback |
| C `#include` via macro / computed path not resolved | Requires C preprocessor — out of scope |
| Bare filename without directory prefix may be dropped | Accepting basenames would reopen the suffix-collision false positive; precision chosen over recall; mitigated by CCR recovery |
| Regex extractors may create false edges from string literals | Direction is safe: extra edges → larger neighborhood → keeps more lines, never drops relevant ones |

## Telemetry

- `route_counts["graph_narrow"]` — number of times lossy narrowing fired
- `route_counts["graph_narrow_tokens_saved"]` — estimated tokens saved by lossy narrowing
- `route_counts["graph_narrow_lossless_tokens_saved"]` — estimated tokens saved by lossless narrowing
- All surfaced via existing `GET /stats` → `PrometheusMetrics` path (zero new infrastructure)

## Configuration

| Setting | Default | Description |
|:---|:---|:---|
| `enable_graph_narrow` | `True` | Master toggle for graph narrowing |
| `graph_narrow_max_depth` | `2` | BFS depth limit (configurable via `HEADROOM_GRAPH_NARROW_MAX_DEPTH` env var) |
| `graph_narrow_min_lines` | `40` | Minimum output lines to trigger narrowing |
| `graph_narrow_tool_names` | `{glob, grep, rg, ls, ...}` | Tools whose output is eligible for narrowing |
| `HEADROOM_GRAPH_MAX_WATCHED_ROOTS` | `16` | Cap on concurrent background watchers |

JoaoMarcos44:
## Description

Implements graph-scoped narrowing of discovery-tool output in the Headroom proxy, as specified in #1924.

When a coding agent runs discovery commands such as grep, glob, rg, ls, or directory listing tools, the proxy can filter wide output down to lines whose source files are inside the BFS import-graph neighborhood of the file the agent most recently read. The narrowed output then continues through the existing relevance/compression pipeline.

Closes #1924

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- Added `headroom/graph_context.py` for project import graph construction, cache persistence, BFS traversal, and optional live watcher refresh.
- Added Python, JS/TS, Rust, and C/C++ import/include extraction for project-local graph edges.
- Integrated graph narrowing into `ContentRouter` for lossy aged discovery results and CCR-recoverable fresh discovery results.
- Added route-count telemetry for graph narrowing and estimated tokens saved.
- Exposed `HEADROOM_GRAPH_NARROW_MAX_DEPTH` in proxy configuration.
- Added focused tests in `tests/test_graph_context.py` and `tests/test_graph_narrow.py`.

## Testing

### Test Output

```text
Original author run:
tests/test_graph_context.py and tests/test_graph_narrow.py: 41 passed, 1 skipped in 13.70s

Reviewer run:
uv run --extra proxy --with pytest python -m pytest tests/test_graph_context.py tests/test_graph_narrow.py -q
42 passed, 1 warning
```

## Real Behavior Proof

- Environment: Windows 10, Python 3.11, headroom repository (`c:\Users\Nitro\headroom`).
- Exact command / steps: Ran live benchmark and conversation compression test on the headroom codebase.
- Observed result: Mapped 1,379 files and 2,603 imports in 3.32s; grep for `SmartCrusher` on `content_router.py` achieved 91.4% token reduction; grep for `cache` on `paths.py` narrowed 10,075 lines to 8 lines with 284ms filter latency; active conversation compression test reduced size by 2,033 tokens while protecting user prompts and assistant dialog.
- Not tested: Non-SmartCrusher tabular outputs such as CSV/plain text, which are outside this PR's stated scope.

## Review Readiness

## Additional Notes

Known limitations documented by the author:

- Rust `use crate::path` is not resolved; only `mod` declarations are handled.
- JS bundler aliases such as `@/x` and `~/y` are not resolved.
- C `#include` via macro is not resolved.
- Bare filenames without directory prefixes may be dropped to avoid suffix-collision false positives; fresh results remain recoverable via CCR.
- ## Performance test suite for graph-scoped narrowing

Added two files to validate the token-savings claim and check CPU/memory cost don't offset it:

- `tests/test_graph_narrow_performance.py` — 9 pytest regression guards (build_graph scaling, BFS cost independent of repo size, cache-hit vs cold, `get_live_graph` watcher startup vs reuse, token-savings floor, narrow latency, memory/file). Thresholds carry 5-40x margin over measured numbers so they catch real algorithmic regressions (e.g. an accidental O(n²)) without flaking on CI noise.
- `benchmarks/bench_graph_narrow.py` — standalone report script (`python benchmarks/bench_graph_narrow.py`) for humans investigating the actual numbers.

Full suite (new + existing `test_graph_narrow.py` + `test_graph_context.py`): **50 passed, 0 failed, 1 skipped**.

### Results — synthetic repo (5,050 stub files)

| metric | value |
|---|---|
| build_graph (cold) | 701ms |
| bfs_related_files | 0.03ms |
| get_live_graph (2nd call) | 0.15-0.20ms |
| content reduction | 99.0% (145,262 → 1,470 chars) |
| _graph_narrow latency (watcher warm) | 290-500ms / 5,050 lines |

Scales linearly (20x files → ~17-22x time), not quadratically. Memory ~0.1-0.3 KB/file, no leak signal.

### Results — validated against this repo (headroom itself, 1,391 real files, multi-lang)

Real entrypoint (`headroom/transforms/content_router.py`, 28 direct imports), real `rg -n "Tokenizer" --type py .` dump (not fabricated):

| metric | synthetic | real |
|---|---|---|
| build_graph (cold) | 701ms / 5,050 files | **4,249ms** / 1,391 files |
| bfs_related_files | 0.03ms | 0.04ms |
| get_live_graph (2nd call) | 0.15-0.20ms | 0.13ms |
| related files (depth=2) | 50 / 5,050 | 49 / 1,391 |
| _graph_narrow (watcher warm) | 290-500ms / 5,050 lines | 33.6ms / 344 lines |
| content reduction | 99.0% | 89.7% |

BFS and warm-graph reuse hold up identically on real code. The gap that matters: **build_graph costs ~20x more per file on real code than on the synthetic stub files** (real files have actual imports/AST to resolve; the synthetic benchmark's 1-line stubs understated this).

### Problem to solve: ~4.2s one-time cost on a real ~1,400-file repo

Root cause: `get_live_graph`'s first call for a root always calls `_rebuild_live` → `build_graph` unconditionally — it never checks the hash-based on-disk cache that `load_or_build_graph` writes/reads. So even if a previous process already built and cached this exact tree, the live-graph path redoes the full AST parse from scratch on its first call. Confirmed by measurement: `get_live_graph` first call (4,249ms) ≈ `build_graph` cold alone (4,276ms) — the watchdog `Observer.schedule()` registration itself is not the expensive part.

This is a one-time cost per proxy process per project root (every later call is ~0.1ms, in-memory), so it's not a per-request tax — but it does mean the *first* discovery-tool result in a fresh proxy session pays ~4s before narrowing can kick in, on a repo this size. Larger monorepos will pay more.

Ideas to bring this down, roughly in order of expected payoff:

1. **Have `_rebuild_live` try the disk cache before rebuilding.** Reuse `load_or_build_graph`'s hash-check path first; only fall back to a full `build_graph` when hashes actually changed. A warm on-disk cache should turn this into a hash-check pass (~350ms in the synthetic 5k-file case) instead of a full reparse.
2. **Parallelize `build_graph`'s file walk/AST parse.** It's currently sequential; `graphify` itself parallelizes AST extraction across workers for the same kind of workload (saw "12 workers" in `graphify update` output on this repo) — same approach would apply here, since each file's read+parse is independent.
3. **Don't block the first request on it.** `GraphContextWatcher.start()` currently does the initial build synchronously before returning. Kicking off that build in the background as soon as the proxy detects a project root (instead of lazily on the first `get_live_graph` call) would move the ~4s off the first user-visible request's critical path.


chopratejas:
Great work - can you fix the linting errors?

JerrettDavis (changes requested):
Thanks for the detailed write-up and test coverage here. I found one correctness issue that needs to be fixed before this is ready.

`GraphContextWatcher.start()` calls `_rebuild_live(self.root)` before it imports/starts watchdog. If watchdog is unavailable, or observer startup fails, `start()` returns `False` but `_live_graphs[root]` has already been populated. `get_live_graph()` then removes the reservation from `_live_watchers`, but leaves `_live_graphs[root]` behind. On the next call, lines 663-665 return that in-memory graph before checking whether a watcher is actually active, so the documented pull-model fallback is bypassed and the proxy can keep using stale graph data forever after file changes.

Minimal reproduction I ran locally by simulating a start failure after `_rebuild_live()`:

```text
first ['util.py']
second_has_new False
live_watchers False
live_graphs True
```

The fix should make the in-memory fast path conditional on an active watcher, or clear `_live_graphs[root]` whenever watcher startup fails. Please add a regression test that matches the real failure mode: `start()` does the initial rebuild, returns `False`, a source file changes, and the next `get_live_graph()` must observe the change through `load_or_build_graph()`.

I also made a non-code PR body cleanup so the governance parser can recognize the existing Real Behavior Proof fields and self-review checkbox.

Local verification while reviewing:

```text
uv run --extra proxy --with pytest python -m pytest tests/test_graph_context.py tests/test_graph_narrow.py -q
42 passed, 1 warning
```

JoaoMarcos44:
> Thanks for the detailed write-up and test coverage here. I found one correctness issue that needs to be fixed before this is ready.
> 
> `GraphContextWatcher.start()` calls `_rebuild_live(self.root)` before it imports/starts watchdog. If watchdog is unavailable, or observer startup fails, `start()` returns `False` but `_live_graphs[root]` has already been populated. `get_live_graph()` then removes the reservation from `_live_watchers`, but leaves `_live_graphs[root]` behind. On the next call, lines 663-665 return that in-memory graph before checking whether a watcher is actually active, so the documented pull-model fallback is bypassed and the proxy can keep using stale graph data forever after file changes.
> 
> Minimal reproduction I ran locally by simulating a start failure after `_rebuild_live()`:
> 
> ```
> first ['util.py']
> second_has_new False
> live_watchers False
> live_graphs True
> ```
> 
> The fix should make the in-memory fast path conditional on an active watcher, or clear `_live_graphs[root]` whenever watcher startup fails. Please add a regression test that matches the real failure mode: `start()` does the initial rebuild, returns `False`, a source file changes, and the next `get_live_graph()` must observe the change through `load_or_build_graph()`.
> 
> I also made a non-code PR body cleanup so the governance parser can recognize the existing Real Behavior Proof fields and self-review checkbox.
> 
> Local verification while reviewing:
> 
> ```
> uv run --extra proxy --with pytest python -m pytest tests/test_graph_context.py tests/test_graph_narrow.py -q
> 42 passed, 1 warning
> ```

Thank you very much for pointing that out; I just corrected it. Please check again to make sure everything is correct.

JerrettDavis (approved):
Re-reviewed the follow-up fix on the current head `de2b9d6a`.

The new watcher-start failure cleanup addresses the blocker I raised: when `GraphContextWatcher.start()` performs its initial `_rebuild_live()` and then returns `False`, `get_live_graph()` now clears both the watcher reservation and the leftover `_live_graphs[root]` snapshot, so later calls fall back to the hash-checked pull model and observe file changes. The added regression test matches the actual failure sequence rather than the previous simplified `start() -> False` stub.

I also checked the added performance regression suite. It is intentionally loose-thresholded, covers build/BFS/live-graph/narrowing cost and token-savings behavior, and stayed green locally.

Local verification:

```text
uv run --extra proxy --with pytest python -m pytest tests/test_graph_context.py tests/test_graph_narrow.py tests/test_graph_narrow_performance.py -q
51 passed, 1 skipped, 1 warning
```

I do not see a remaining blocker in the graph-narrowing change.

chopratejas:
1. You have handled grep, glob, read etc - but many harnesses (like opencode) - issue bash calls for each - they will do a 'cat -n' .. or a 'sed' instead of these - So - there is a need to extend this to those too
2. In general - harnesses will get better at applying this - no? Why do we need this then? I am trying to understand where is the benefit of this approach if harnesses eventually will know how to read just the appropriate pieces.

JoaoMarcos44:
@chopratejas  Both good points — answered inline, and pushed a fix for (1).

**1. Bash-shelled discovery (opencode-style harnesses)**

Confirmed gap. `_graph_narrow`/`_graph_narrow_lossless` were gated only on the harness's own tool name (`graph_narrow_tool_names`: glob/grep/rg/ls/...), so a harness that shells `grep -r`/`find`/`ls -R` through a generic `bash` tool call instead of a dedicated discovery tool never hit narrowing at all.

Fixed in `62a9f731`: added a `graph_narrow_bash_commands` whitelist and a shared `_is_graph_narrow_candidate(tool_name, tool_call_id)` gate that both narrowing paths now use — it accepts either the dedicated tool-name match (unchanged) or a bash-family tool call whose *parsed shell command* (`_bash_command_is_search`, already used by the bash-search lossless fold) is a discovery command. Reuses the existing `_tool_call_commands` parsing rather than adding new command-parsing logic. New regression tests cover `grep -r`/`find`/`ls -R` via bash, plus a pinned no-op case: `cat`/`sed` of a single already-known file isn't a repo-wide dump with a path list to filter against, so graph-narrow correctly has nothing to do there (that's a different problem — read-lifecycle/truncation territory, not graph narrowing).

**2. Won't harnesses eventually make this obsolete?**

I don't think so, for three reasons:

- Bash is the lowest-common-denominator tool interface every harness supports, and models keep reaching for compound shell pipelines (`find | xargs grep`, etc.) that no single structured tool covers — that's a training-distribution habit, not a harness capability gap, so it doesn't go away as harnesses improve.
- Even among harnesses that *do* have structured discovery tools, the tool names aren't standardized (Claude Code's `Grep`/`Glob`/`LS` vs. others) — coverage is inherently a moving target of harness-specific names/conventions, which is exactly why this gate is config-driven (`graph_narrow_tool_names` / `graph_narrow_bash_commands`) instead of hardcoded.
- The proxy's value here is being harness-agnostic at the wire level: it catches the win regardless of which convention produced the output, including headless/scripted agents that will always just shell out. So this isn't a stopgap waiting for harnesses to "catch up" — it's the layer that stays correct independent of what any given harness decides to do.

---

**Edit:** one honest caveat on (2) I should add rather than let the three points above sound like the whole picture. There's a stronger counter-argument than "bash never disappears": if a harness's *own* discovery tool becomes graph-aware client-side (scoped via LSP/symbol-index/RAG before the tool call ever returns), the wide dump this feature targets never gets generated in the first place — narrowing degrades to a real no-op for that harness, not because bash went away, but because there's nothing left to trim. That's a legitimate case where server-side narrowing loses its edge to a smarter client.

I don't think it changes the conclusion though: most harnesses in the wild aren't there yet and won't be for a while, third-party/scripted/headless agents will keep shelling out regardless, and the proxy costs nothing extra when a discovery result is already narrow (it's gated on `graph_narrow_min_lines`, a no-op on small output). So this stays a defense-in-depth layer rather than the *only* source of the win — which is a narrower claim than my original three bullets implied, and the more accurate one.

Since this thread, also extended the underlying `graph_context` import resolution from 4 language families (Python/JS-TS/Rust/C-C++) to 14 — added Go, Java, C#, Ruby, PHP, Kotlin, Scala, Dart, Lua, Zig via `tree_sitter_language_pack` (same optional `[code]` extra `code_compressor.py` already uses) instead of hand-rolled regex per language, so the bash-coverage fix above isn't Python/JS-only. Ran it through four rounds of adversarial testing looking for regressions/wrong-edges/crashes (~68 checks: transitive BFS, circular imports, case-sensitivity, wildcard/blank/rename import forms, concurrency, cache invalidation, real-tokenizer-measured savings) — found and fixed two real bugs along the way (a regex alternation-ordering bug that silently broke C#/C++ entrypoint detection, and a self-edge bug in Go/C#'s directory-scoped resolution), all now pinned by regression tests.

---

**Follow-up on the client-side-graph-aware caveat above** — where does that actually show up today, and what would we do about it.

It's not purely hypothetical. A few real, shipping examples of a harness already scoping discovery client-side instead of shelling `grep -r`/`find`:
- **Aider's repo-map** (`--map-tokens`) — already uses tree-sitter (same tech this PR just added to `graph_context`) to build a ranked map of the repo and feeds that to the model instead of raw grep/find output.
- **Cursor / Windsurf** — pre-built embedding index over the codebase; a "where is X handled" query returns curated snippets from that index, not a shelled-out search.
- **JetBrains AI Assistant** — backed by the IDE's own PSI (semantic index used for refactor/autocomplete for years); "find usages" resolves through that index, not text search.

None of Claude Code, opencode, Codex CLI, or headless/scripted agents are there — which is most of the volume this feature targets today.

If that shifts further, three ways this scales without needing to "fix" anything defensively:
1. **Already degrades to zero cost.** `graph_narrow_min_lines` means an already-narrow discovery result just passes through untouched — no code change needed for that to keep being true.
2. **The rest of the pipeline doesn't depend on this.** Graph-narrow trims the discovery *list*; SmartCrusher/CodeAwareCompressor/Kompress compress the file *content* an agent reads afterward, regardless of whether the list that pointed at those files came from grep or from an embedding index. A graph-aware client removes the discovery-side win, not the content-side one.
3. **The same BFS engine could stop being purely reactive.** `graph_context`'s BFS already has a proactive mode (`assemble_context`, what `graphify query`/`explain`/`path` call) alongside the reactive one this PR wires into narrowing. If discovery dumps genuinely dry up industry-wide, the natural move is exposing that BFS as a Headroom-injected MCP tool (`ccr/mcp_server.py` / `tool_injection.py` already exist for this) so the agent calls a Headroom-provided graph-aware tool directly — Headroom becomes the client-side-aware layer instead of waiting to react to one.

None of this is scoped into the current PR — flagging it as the honest answer to "then what", not a commitment to build it now.

chopratejas:
I have a couple of points to add

In headroom today, we have a way to ‘understand’ reads and greps and all even with bash tools.

can you reuse that?

i am concerned about the code here and what it intends to solve - and i feel its very complex for the value add it can provide.

JerrettDavis (changes requested):
The latest head is red in CI, so I am moving this back to changes requested until the follow-up commits are cleaned up.

Current failing checks on run `29087743370`:
- `lint`: `ruff format --check .` would reformat `headroom/cli/graph_context.py`, `headroom/graph_context.py`, `headroom/transforms/content_router.py`, `tests/test_graph_context_languages.py`, `tests/test_graph_narrow.py`, and `tests/test_graph_narrow_languages.py`.
- `test (2)`: the graph watcher path is not starting in CI anymore. Failures include `test_get_live_graph_starts_watcher_and_returns_correct_graph`, `test_get_live_graph_second_call_is_in_memory_reuse`, `test_watcher_picks_up_new_file_via_real_filesystem_event`, `test_stop_all_watchers_clears_state`, `test_watched_roots_are_capped_with_fifo_eviction`, `test_get_live_graph_does_not_start_duplicate_watchers_under_concurrency`, and the near-instant second-call perf assertion. The common symptom is `_live_watchers == {}` / `GraphContextWatcher.start() is False`, so the tests are exercising the pull-model fallback instead of the expected live watcher path.
- `test (4)`: `tests/test_release_workflows.py::test_no_native_tls_in_wheel_build_tree` reports `native-tls` is back in the `headroom-py` build tree.

Please fix the formatting and either restore the watcher/native-tls invariants or adjust the tests/config intentionally if the expected behavior changed.

PR #1951, opened 2026-07-10
fix(proxy): only queue mid-turn messages for opt-in clients with explicit session header

### LINKED ISSUE #1949: [BUG]Concurrent streaming requests wrongly queued as mid-turn → 202 breaks standard SDK clients
## Description

Mid-turn steering wrongly queues **concurrent independent streams**. When two streaming `/v1/messages` requests share the same model + system prompt and arrive concurrently (no `x-headroom-session-id` header), the proxy misclassifies the second as a "mid-turn message", returns `202 {"event":"headroom_queued"}`, and never forwards it upstream. A standard Anthropic SDK client that made a *streaming* call receives a non-SSE 202 → empty event stream → `AssertionError`, and fails after retries.

## To Reproduce

1. Run the proxy in cache mode: `headroom proxy --port 8799 --mode cache`
2. Point a stock Anthropic SDK at it and fire two concurrent `messages.stream()` calls with identical `model` + `system` and **no** `x-headroom-session-id` header (see Code Sample).
3. The second (in-flight) request returns `202`; the SDK's `get_final_message()` raises `AssertionError`.

## Expected Behavior

Both concurrent streams are forwarded upstream and return normal SSE responses. Mid-turn steering should only apply to clients that explicitly opt in via `x-headroom-session-id`, since queued messages are only drained back through the custom `headroom_pending_messages` SSE event that a standard SDK doesn't understand.

## Actual Behavior

The second request is queued and answered with a non-SSE `202 {"event":"headroom_queued"}` and is never sent upstream. The streaming SDK accumulates zero events, so `self.__final_message_snapshot` stays `None` and the final-message assertion fails. Proxy logs show `status=202` with `upstream_connect: null`.

**Root cause:** without an `x-headroom-session-id` header, `_get_session_key()` falls back to `md5(model + system[:500])` (mirrors `prefix_tracker.compute_session_id`). That key is too coarse to distinguish concurrent independent streams sharing a model + system prompt, so the second stream hits `session_key in self._active_streams` and gets queued. Note: sending a *unique* header per request is not a workaround — the same header also drives `prefix_tracker.compute_session_id()`, so unique-per-stream ids break prompt caching while a shared id keeps colliding.

## Code Sample

```python
import anthropic, threading

client = anthropic.Anthropic(base_url="http://127.0.0.1:8799", api_key="…")

def call():
    with client.messages.stream(
        model="claude-sonnet-4-20250514",
        system="same system prompt for both",
        max_tokens=256,
        messages=[{"role": "user", "content": "hello"}],
    ) as s:
        s.get_final_message()

# Fire both while the first is still streaming
t1 = threading.Thread(target=call); t2 = threading.Thread(target=call)
t1.start(); t2.start(); t1.join(); t2.join()
```

## Error Output

```
AssertionError
  File ".../anthropic/lib/streaming/_messages.py", line 94, in get_final_message
    assert self.__final_message_snapshot is not None
```

Proxy log for the failing request:

```
event=proxy_inbound_response method=POST path=/v1/messages status=202  (upstream_connect: null — never forwarded)
```

## Environment

- **Headroom version**: 0.30.0 (also present on `main`)
- **Python version**: 3.11
- **OS**: macOS
- **LLM Provider**: Anthropic (streaming, `messages.stream`); proxy mode `cache`

## Additional Context

**Suggested fix** — only engage mid-turn steering when the client explicitly opts in via the header (never from the coarse md5 fallback):

```python
explicit = request.headers.get("x-headroom-session-id")
session_key = self._get_session_key(body, session_header=explicit)
if explicit and session_key in self._active_streams:
    return JSONResponse(content=self._queue_mid_turn_message(session_key, body), status_code=202)
```

All existing `test_mid_turn_steering.py` cases pass `x-headroom-session-id` explicitly, so they're unaffected; prefix-tracker / cache behavior is untouched. A larger alternative is to track `_active_streams` by a per-stream instance token rather than by session id, so mid-turn steering can coexist with concurrent independent streams even for header-bearing clients.

wzy-del:
## Description

Mid-turn steering wrongly queues **concurrent independent streams**. When two streaming `/v1/messages` requests share the same model + system prompt and arrive concurrently (no `x-headroom-session-id` header), the proxy misclassifies the second as a "mid-turn message", returns `202 {"event":"headroom_queued"}`, and never forwards it upstream. A standard Anthropic SDK client that made a *streaming* call receives a non-SSE 202 → empty event stream → `AssertionError` (`assert self.__final_message_snapshot is not None` in `anthropic/lib/streaming/_messages.py`), and fails after retries.

**Root cause.** Without an `x-headroom-session-id` header, `_get_session_key()` falls back to `md5(model + system[:500])` (mirrors `prefix_tracker.compute_session_id`). That key is intentionally coarse and cannot distinguish genuinely concurrent, independent streams that share a model + system prompt (e.g. a main conversation plus its background / parallel requests), so the second stream hits `session_key in self._active_streams` and gets queued.

A queued message is only ever drained back to the client via the custom `headroom_pending_messages` SSE event, which a standard Anthropic SDK does not understand — so mid-turn steering is effectively a private protocol for clients that **opt in** via `x-headroom-session-id`. A client that never sends the header can never participate in the queue; for it, the 202 is simply a broken streaming response.

Note: "send a unique header per request" is **not** a workaround — the same header also drives `prefix_tracker.compute_session_id()`, so unique-per-stream ids break prompt caching while a shared id keeps colliding.

Closes #1949

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `StreamingMixin._should_queue_mid_turn()` helper that gates mid-turn queuing behind an explicit `x-headroom-session-id` header.
- Header-less concurrent streams are now forwarded upstream normally; only opt-in (header-bearing) callers can be queued.
- Prefix-tracker / cache-alignment behavior is untouched — the header still drives `compute_session_id()` exactly as before.

## Testing

### Test Output

```text
$ pytest tests/test_mid_turn_steering.py
6 passed
```

New test `test_should_queue_only_with_explicit_session_header`: a header-less concurrent stream must not queue; an explicit-header opt-in must. All existing `test_mid_turn_steering.py` cases pass an explicit header and are unaffected.

## Real Behavior Proof

- Environment: macOS, `headroom-ai` 0.30.0 (installed via `uv tool`), proxy running `headroom proxy --port 8799 --no-http2 --mode cache`, upstream = an Anthropic-compatible gateway. Client = Hermes Agent (Anthropic SDK, streaming) driving a main conversation plus concurrent background/parallel requests that share the same model + system prompt.
- Exact command / steps:
  1. Reproduce on stock 0.30.0: concurrent streaming requests without `x-headroom-session-id` → second stream returns `202 {"event":"headroom_queued"}` → client raises `AssertionError` in `anthropic/lib/streaming/_messages.py`.
  2. Correlate logs: count of `AssertionError` in the client error log vs count of `202` in the proxy access log for the window — **48 == 48**, timestamps line up 1:1.
  3. Apply this patch to the running package, restart the proxy, re-run the same concurrent workload.
- Observed result: after the fix, **0 × 202 / all requests 200**, no new `AssertionError`, and `cache_hit_pct` stayed ~99% (prefix caching intact). Header-bearing opt-in clients still queue mid-turn as before.
- Not tested: behavior under a client that deliberately sends a *changing* `x-headroom-session-id` per request (out of scope — documented as a caching anti-pattern, not a supported mode).

## Review Readiness

## Checklist

## Additional Notes

- Docs / CHANGELOG unchanged: this is a proxy-internal correctness fix with no user-facing config surface.
- The fix is deliberately minimal and conservative — it only narrows *when* queuing engages (explicit opt-in header), leaving the prefix-tracker, cache-alignment, and body-rewrite paths byte-for-byte identical.


chopratejas:
Thank you for this change. Can you fix linting here?

wzy-del:
Thanks @chopratejas — linting is fixed now (ruff format applied in cc2efd21).

wzy-del:
Heads up on the red check: the failing `test (1)` shard is unrelated to this PR — it's `test_codex_ws_compression_scheduler.py::test_concurrent_compression_has_no_semaphore_tail`, a latency-ratio perf test in the codex WS scheduler, not the mid-turn / streaming code this PR touches.

It's flaky on shared CI runners: it asserts `p99/p50 < 4×` but this run got `p50=1ms, p99=39ms → 27.2×` — with a ~1ms p50, ordinary scheduler jitter on a 2-vCPU runner blows past the bar (the test's own comment notes a ~25ms noise floor). The other 3 test shards, lint, and build are all green. A re-run of that shard should clear it.

JerrettDavis (approved):
Reviewed the mid-turn queue gating change. Requiring an explicit `x-headroom-session-id` before returning the private `headroom_queued` 202 response fixes the headerless concurrent-stream collision without changing prefix-tracker identity or the existing opt-in mid-turn queue path.

Verification:
- `python -m pytest tests/test_mid_turn_steering.py tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py -q`
- `python -m pytest tests/test_codex_ws_compression_scheduler.py::test_concurrent_compression_has_no_semaphore_tail -q`
- `python -m ruff check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/streaming.py tests/test_mid_turn_steering.py`
- `python -m ruff format --check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/streaming.py tests/test_mid_turn_steering.py`

The previous red CI job was the unrelated scheduler tail-latency assertion (`test_codex_ws_compression_scheduler.py::test_concurrent_compression_has_no_semaphore_tail`), not a touched mid-turn/streaming test. I reran the failed workflow job after verifying the exact test locally.

JerrettDavis (approved):
Re-reviewed the current head after the `main` merge and formatting cleanup. The mid-turn queue remains correctly gated behind an explicit `x-headroom-session-id`, so headerless concurrent streams do not receive the private queued-message response path.

```text
python -m pytest tests/test_mid_turn_steering.py -q
collection failed before tests ran: ModuleNotFoundError: No module named 'headroom._core'
```

That is this local checkout's missing compiled Rust extension when importing `headroom.proxy.server`, not a touched-code assertion failure. Targeted Ruff checks are clean:

```text
uvx ruff==0.15.17 check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/streaming.py tests/test_mid_turn_steering.py
All checks passed!

uvx ruff==0.15.17 format --check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/streaming.py tests/test_mid_turn_steering.py
3 files already formatted
```

PR #1970, opened 2026-07-10
fix(proxy): strip inbound Content-Encoding on messages/chat forward

abhay-codes07:
## Description

The Anthropic `/v1/messages` and OpenAI `/v1/chat/completions` handlers decode the
inbound request body before forwarding it upstream. `read_request_json_with_bytes`
(helpers.py) inflates `zstd`/`gzip`/`deflate`/`br` bodies, and the handler forwards
the resulting plain JSON. But both handlers build the upstream-bound header dict and
pop only `host`, `content-length`, and `accept-encoding` — they leave the original
`Content-Encoding` header in place:

```python
headers = dict(request.headers.items())
headers.pop("host", None)
headers.pop("content-length", None)
headers.pop("accept-encoding", None)   # content-encoding NOT popped
```

So when a client — or an edge proxy like a Cloudflare Worker — sends a request with
`Content-Encoding: gzip` (or `zstd`/`br`/`deflate`) and a compressed body, Headroom
decompresses it, then forwards plain JSON that still advertises `content-encoding: gzip`.
The upstream provider tries to gunzip already-decoded JSON and rejects the request with
HTTP 400. Every such request fails.

This is a known class of bug: the `/v1/responses` handler already fixes exactly this at
`openai.py` with the comment *"Leaving a stale content-encoding header makes the upstream
try to decompress already-decoded JSON and reject it with HTTP 400 (#1542)."* That fix
landed only on the `/responses` path — the messages and chat paths were missed.

Closes: no issue filed — found while auditing request-header forwarding across the handlers.

## Fix

Pop `content-encoding` and `transfer-encoding` in both handlers, right after the existing
`content-length` pop, mirroring the `/v1/responses` handler:

```python
headers.pop("content-encoding", None)
headers.pop("transfer-encoding", None)
```

## Type of Change

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: strip `content-encoding`/`transfer-encoding` from the upstream-bound request headers in `handle_anthropic_messages`.
- `headroom/proxy/handlers/openai.py`: same strip in the `/v1/chat/completions` handler.
- `tests/test_proxy_compression_headers.py`: add `TestRequestContentEncodingStripping` covering gzip/zstd/deflate/br, `transfer-encoding`, and the absent-header (plain curl) case.

## Testing

- [x] New regression tests added (`tests/test_proxy_compression_headers.py`)

```text
$ uv run ruff check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py tests/test_proxy_compression_headers.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.10, headroom from this branch. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the header logic with a dependency-free script (the same pure-dict pattern the existing tests in this file use) and left the full pytest to CI.
- Exact command / steps: replicated the handler's request-header stripping (old vs new) in a standalone script and ran a `gzip`/`zstd`/`deflate`/`br` request through both.
- Observed result: the old logic keeps `content-encoding` (which is what makes the upstream 400); the new logic strips it while preserving `authorization` and `content-type`:

```text
OK gzip: old leaks 'gzip' -> upstream 400 ; new strips it
OK zstd: old leaks 'zstd' -> upstream 400 ; new strips it
OK deflate: old leaks 'deflate' -> upstream 400 ; new strips it
OK br: old leaks 'br' -> upstream 400 ; new strips it
OK transfer-encoding stripped
OK safe when absent
CONTENT-ENCODING STRIP VERIFIED
```

- Not tested: an end-to-end POST of a real gzip body through a booted proxy to a live upstream (needs the heavy stack + a provider key). The header now matches the byte-faithful forwarding the `/responses` path already does, and the new tests exercise the exact strip logic. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- Small, contained parity fix — two `pop()` calls plus tests, no new dependencies.
- @JerrettDavis tagging you since you've been triaging the proxy-forwarding fixes (this is the sibling of the #1542 `/responses` fix) — should be a quick one if you have a moment.


chopratejas:
Thanks for the change - can you fix the linting error?

abhay-codes07:
Fixed — that was a `ruff format` miss on the new test file, not `ruff check`. I'd only run `ruff check` locally; turned out my local ruff was 0.14.14 while CI pins 0.15.17, so I re-ran the format with the pinned version and pushed. Lint's green now.

PR #1996, opened 2026-07-10
fix(ccr): don't IndexError on an empty OpenAI choices list

abhay-codes07:
## Description

`CCRResponseHandler` extracts the assistant turn from an OpenAI response like this
(`headroom/ccr/response_handler.py`, lines 136 and 408):

```python
message = response.get("choices", [{}])[0].get("message", {})
```

The `[{}]` default only applies when the `choices` key is **absent**. A present-but-empty
`choices: []` therefore evaluates to `[][0]` → `IndexError`.

The OpenAI handler calls this unconditionally on any truthy 200 body
(`headroom/proxy/handlers/openai.py`):

```python
if (
    self.ccr_response_handler
    and backend_response.body
    and backend_response.status_code == 200
    and self.ccr_response_handler.has_ccr_tool_calls(backend_response.body, "openai")
):
```

So an upstream 200 of `{"choices": [], ...}` — which **Azure OpenAI** returns on prompt
content-filtering, and which **OpenAI** emits for the trailing usage-only chunk when
`stream_options.include_usage` is set — flows into `has_ccr_tool_calls` →
`_extract_tool_calls` → `IndexError`, crashing the request and returning a **500** instead of
passing the benign response through.

The Anthropic branch is already safe because it uses `response.get("content", [])`; only the
two OpenAI branches have the `[0]` hazard.

Closes: no issue filed — found while auditing the CCR response path.

## Fix

Use `response.get("choices") or [{}]` in both OpenAI branches, so an empty list falls back to
`[{}]` just like a missing key:

```python
choices = response.get("choices") or [{}]
message = choices[0].get("message", {})
```

## Type of Change

## Changes Made

- `headroom/ccr/response_handler.py`: guard `choices[0]` with `or [{}]` in `_extract_tool_calls` and `_extract_assistant_message`.
- `tests/test_ccr_response_handler.py`: add tests that an empty `choices: []` and a missing `choices` key return no CCR calls instead of raising.

## Testing

- [x] New regression tests added (`tests/test_ccr_response_handler.py`)
- [x] Linting passes (`ruff check`) and formatting is clean (`ruff format --check`) — run with the CI-pinned `ruff==0.15.17`

```text
$ uvx ruff@0.15.17 check headroom/ccr/response_handler.py tests/test_ccr_response_handler.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.10, headroom from this branch. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the extraction logic with a dependency-free script and left the full pytest to CI.
- Exact command / steps: ran a `{"choices": []}` response through both the old `response.get("choices", [{}])[0]` expression and the new `response.get("choices") or [{}]` guard, plus the normal and missing-key cases.
- Observed result: the old expression raises `IndexError` (the 500); the new guard returns no tool calls and preserves the normal case:

```text
OLD: choices=[] -> IndexError (crashes request -> 500)
NEW: choices=[] -> [] (no CCR calls, no crash)
NEW: missing choices -> []
NEW: normal choices -> tool_calls preserved (parity with old)
CCR EMPTY-CHOICES FIX VERIFIED
```

- Not tested: a full round-trip with a real Azure content-filtered response through a booted proxy (needs the heavy stack). The new tests drive `has_ccr_tool_calls` / `_extract_tool_calls` / `_extract_assistant_message` directly with the empty-choices body. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- Two-line guard plus tests; no new dependencies.
- @JerrettDavis tagging you — this one turns a benign upstream 200 (Azure content-filter / usage-only chunk) into a proxy 500, so it may be worth a look when you have a moment.


abhay-codes07:
Closing this — the CCR refactor that landed on main (the new `ccr/tool_calls.py`) already fixes the empty-choices crash, and more thoroughly than this PR did. `extract_tool_calls` there guards it directly:

```python
if not isinstance(choices, list) or not choices:
    return []
```

The production path (`CCRResponseHandler.has_ccr_tool_calls` → the module-level `has_ccr_tool_calls` → `extract_tool_calls`) now goes through that guard, so the `{"choices": []}` 200 (Azure content-filter / usage-only chunk) no longer raises. The only place that still raises is the private `_extract_tool_calls` shim, which is deliberately kept for the legacy tests. Nothing left for this PR to fix. Thanks!

PR #2001, opened 2026-07-10
fix(proxy): keep Kompress warmup off the startup path

### LINKED ISSUE #1908: Segfault in libarrow.so (jemalloc_bg_thd) on proxy startup with Kompress ONNX compressor enabled (RHEL/CentOS 7-family, glibc 2.17)
## Summary

`headroom proxy` (full optimization mode, i.e. Kompress ONNX text-compression path enabled) reproducibly segfaults within ~10-30 seconds of startup, every single time, right after logging `Downloading Kompress ONNX model from chopratejas/kompress-v2-base ...`. The crash is in a `jemalloc_bg_thd` background thread inside `libarrow.so`, not in Python-level code (no traceback in the app log — the process just dies).

This reproduced identically across **4 separate clean restarts**, with different proxy configurations (with/without SOCKS5 proxy env vars, with/without `MALLOC_CONF=background_thread:false`, with ~2.3GB and ~1.4GB free system RAM). It is *not* reproducible with `headroom proxy --no-optimize` (stable, low memory ~135MB RSS).

## Environment

- OS: Red OS 7.3 (RHEL/CentOS 7-family, glibc ~2.17), SELinux enforcing
- Python: 3.11 (venv), `headroom-ai` 0.30.0 (latest on PyPI), installed 2026-07-07
- `pyarrow` 24.0.0, `onnxruntime` 1.27.0
- Kompress ONNX model already fully cached locally (`~/.cache/huggingface/hub/models--chopratejas--kompress-v2-base`, 274MB blob, verified checksum unchanged across crashes — not a corrupted-download issue)
- Kernel: confirms `jemalloc_bg_thd[<pid>]: segfault at 350 ip 0x...66d1 ... in libarrow.so.2400[...+1d82000]` — same relative offset in `libarrow.so` every time (different ASLR base, same offset), suggesting a deterministic bug at a specific code path, not random corruption.

## What we ruled out

- **Package versions**: nothing updated since initial install (dist-info timestamps all 2026-07-07); same version as the last known-good day.
- **Proxy config**: crashes identically with SOCKS5 (`socksio`) proxy, plain HTTP proxy, or no proxy env vars at all.
- **Model file corruption**: blob checksum identical before/after repeated crashes; not re-downloaded (already cached).
- **jemalloc background thread env var**: setting `MALLOC_CONF=background_thread:false` did not help — pyarrow appears to use its own statically-linked jemalloc, not affected by the system env var.
- **Low free RAM**: still crashes with ~2.3GB free (not just under extreme memory pressure), though we did also observe RSS ballooning to 1.4-1.5GB+ before each crash, consistent with the RSS-retention pattern described in #211 / fixed in #212 — but #212 is already merged and included in 0.30.0, so this looks like a related-but-distinct failure mode (an actual segfault, not just retained RSS).

## What worked previously

We have hard evidence (structured `proxy.log`) that the exact same code path (full optimization, Kompress compressor) ran successfully for 9+ hours on 2026-07-08 (08:56–18:02), processing many real Anthropic API requests with real compression savings, zero errors logged. We were unable to identify what differs today vs. that successful run (same package versions, same model file, same machine) — the crash may be an intermittent/rare native-library bug that happened not to trigger during that run.

## Steps to reproduce

```bash
headroom proxy   # default: optimization ENABLED, Kompress compressor active
```

Wait ~10-30s. Process banner prints, logs "Downloading Kompress ONNX model from chopratejas/kompress-v2-base ...", then the process dies silently (no Python traceback — a native segfault). Confirmed via `dmesg`/`journalctl -k`:

```
kernel: jemalloc_bg_thd[<pid>]: segfault at 350 ip 00007f...66d1 sp 00007f...
error 4 in libarrow.so.2400[...+1d82000] likely on CPU N (core M, socket 0)
```

Workaround: `headroom proxy --no-optimize` runs stably (low memory, no crash), but disables the Kompress text-compression savings entirely.

## Ask

- Is this a known interaction between onnxruntime's CPU arena/thread settings and pyarrow's bundled jemalloc on older-glibc (RHEL/CentOS 7-family) systems?
- Any additional env var / CLI flag to disable just the Kompress/arrow path while keeping JSON/code compression (SmartCrusher/CodeCompressor), as a narrower workaround than full `--no-optimize`?

Happy to provide full logs / run additional diagnostics if useful.

rodboev:
## Description

Proxy startup can enter cached Kompress native model initialization before binding its port. On the RHEL/CentOS 7-family environment reported in #1908, that path terminates in a deterministic `libarrow.so.2400` jemalloc-thread segfault with no Python traceback. Cache-only preload still initializes native libraries when the model is already cached.

Defer Kompress model and tokenizer loading out of startup while preserving the existing lazy request path and the eager warmups for non-Kompress components. Closes #1908.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Stop cached and uncached Kompress models from entering native preload during proxy startup.
- Record enabled Kompress as deferred until its existing lazy request path needs it.
- Preserve disabled-Kompress routing and non-Kompress eager warmups.
- Add lifecycle, cache-state, disabled-mode, and warmup-preservation regression coverage.
- Document the startup behavior change in the changelog.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_kompress_preload_deferral.py tests/test_kompress_request_nonblocking.py -q`)
- [x] Linting passes (`uv run ruff check headroom/transforms/content_router.py tests/test_kompress_preload_deferral.py tests/test_kompress_request_nonblocking.py`)

### Test Output

```text
uv run pytest tests/test_kompress_preload_deferral.py tests/test_kompress_request_nonblocking.py -q
17 passed in 1.28s

uv run ruff check headroom/transforms/content_router.py tests/test_kompress_preload_deferral.py tests/test_kompress_request_nonblocking.py
All checks passed
```

## Real Behavior Proof

- Environment: Red OS 7.3 or equivalent RHEL/CentOS 7-family system, glibc 2.17, Python 3.11, pyarrow 24.0.0, onnxruntime 1.27.0, cached Kompress model
- Exact command / steps: start `headroom proxy` with Kompress enabled, wait 30 seconds, query the loopback health endpoint, verify deferred Kompress warmup in logs, then inspect `journalctl -k` for new `libarrow.so.2400` or `jemalloc_bg_thd` faults
- Observed result: automated coverage now proves startup avoids the cached Kompress preload boundary, keeps non-Kompress warmups live, and preserves `unavailable` status when dependencies are absent
- Not tested: the native reporter-host run

## Review Readiness

## Checklist

## Screenshots

Not applicable.

## Additional Notes

Focused local validation included `tests/test_kompress_request_nonblocking.py` alongside the startup regression suite. The change is scoped to startup warmup; it does not claim to repair the external `libarrow.so` or jemalloc incompatibility when Kompress later executes. `HEADROOM_DISABLE_KOMPRESS=1` remains the supported narrow workaround for hosts that cannot run the native path.


JerrettDavis (approved):
Reviewed the startup warmup change and the added regression coverage. The new behavior keeps Kompress native preload entirely out of the startup path while preserving the lazy first-use load path, and the warmup status reports `deferred` instead of implying a loaded backend.

Verification:
- GitHub checks are green and the branch is mergeable.
- `python -m ruff check headroom/transforms/content_router.py tests/test_kompress_preload_deferral.py`
- `python -m ruff format --check headroom/transforms/content_router.py tests/test_kompress_preload_deferral.py`

I also attempted `python -m pytest tests/test_kompress_preload_deferral.py tests/test_proxy_eager_preload_bind.py tests/test_proxy_warmup.py -q`; this local workspace is using a stale `headroom._core` build whose `SmartCrusherConfig` does not accept `lossless_only`, so the focused pytest run fails before exercising the Kompress assertions. That mismatch is outside this PR and CI covers the tests successfully.

JerrettDavis (approved):
Re-reviewed the current head after the latest `main` merge. The Kompress native preload remains out of the startup path while preserving lazy first-use loading and reporting warmup as deferred.

```text
python -m pytest tests/test_kompress_preload_deferral.py -q
10 passed

uvx ruff==0.15.17 check headroom/transforms/content_router.py tests/test_kompress_preload_deferral.py
All checks passed!

uvx ruff==0.15.17 format --check headroom/transforms/content_router.py tests/test_kompress_preload_deferral.py
2 files already formatted
```

JerrettDavis (approved):
Re-reviewed the refreshed head after the main merge. The actual PR delta remains limited to the Kompress startup warmup deferral and its focused regression coverage; the behavior still keeps Kompress native preload out of startup while preserving lazy first-use loading.

Focused validation on Windows:
- `uv run --extra dev python -m pytest tests/test_kompress_preload_deferral.py -q` -> 10 passed
- `uvx --from ruff==0.15.17 ruff check headroom/transforms/content_router.py tests/test_kompress_preload_deferral.py`
- `uvx --from ruff==0.15.17 ruff format --check headroom/transforms/content_router.py tests/test_kompress_preload_deferral.py`

PR #2022, opened 2026-07-11
fix(cache/semantic): key entries by context hash, not query text

abhay-codes07:
## Description

`SemanticCache` (`headroom/cache/semantic.py`) derives each entry's key from the **query text
only** — where `query` is just the trailing user message — and its exact-match lookup returns
the slot without checking the stored entry's `messages_hash`:

```python
# put()
key = self._generate_key(query)          # sha256(query)[:16]
self._cache[key] = entry
if messages_hash:
    self._hash_index[messages_hash] = key

# get() — exact-match branch
key = self._hash_index.get(messages_hash)
if key and key in self._cache:
    entry = self._cache[key]
    ...
    return entry                         # never checks entry.messages_hash
```

So two requests that share a trailing user message but differ in earlier context map to the
**same** key. The second `put` overwrites the first, and the first request's `messages_hash`
still points at that (now overwritten) slot — so it is served the **other conversation's**
cached response.

Trailing messages like `"continue"`, `"yes"`, `"fix it"`, `"run the tests"` are extremely
common in agentic/coding sessions, so this collides constantly. It's independent of the
proxy-level `_compute_key` fix (that's about what goes *into* `messages_hash`; here the entry
is stored under a query-only key regardless of how good the hash is). This `SemanticCache` is
the one used by the SDK client's `enable_semantic_cache` path.

Concretely:
1. `put("run the tests", A, messages_hash=HA)` → key `K = sha256("run the tests")`; `_cache[K]=A`.
2. `put("run the tests", B, messages_hash=HB)` → same `K`; `_cache[K]` overwritten with `B`.
3. `get("run the tests", HA)` → `_hash_index[HA]=K`, `K in _cache` → returns **B**.

Closes: no issue filed — found while auditing the cache key derivation.

## Fix

1. Key entries by the full-context `messages_hash` when present, falling back to the query hash
   only when no hash is supplied:
   ```python
   key = messages_hash or self._generate_key(query)
   ```
2. Defensively verify `entry.messages_hash == messages_hash` in the exact-match branch of `get`,
   so any residual stale mapping becomes a miss rather than wrong data.

## Type of Change

## Changes Made

- `headroom/cache/semantic.py`: key `put` entries by `messages_hash` when present; verify `entry.messages_hash` in the `get` exact-match branch.
- `tests/test_cache/test_semantic.py`: add `test_same_query_different_context_does_not_collide` and `test_exact_match_verifies_messages_hash`.

## Testing

- [x] New regression tests added (`tests/test_cache/test_semantic.py`)

```text
$ uvx ruff@0.15.17 check headroom/cache/semantic.py tests/test_cache/test_semantic.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.10, headroom from this branch. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the `put`/`get` logic with a dependency-free script and left the full pytest to CI.
- Exact command / steps: stored responses A and B under the same query `"run the tests"` with different `messages_hash`, then read each hash back — through both the old (query-keyed) and new (hash-keyed) logic.
- Observed result: the old logic serves B's response to request A; the new logic isolates them:

```text
OLD: A->RESPONSE_B  B->RESPONSE_B
NEW: A->RESPONSE_A  B->RESPONSE_B
SEMANTIC CACHE COLLISION FIX VERIFIED (OLD served B to A; NEW isolates)
```

- Not tested: the full SDK `HeadroomClient` round-trip with `enable_semantic_cache=True` (needs the heavy stack). The fix is confined to `SemanticCache.put`/`get` and the new tests drive them directly. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- Small, contained fix — the key derivation plus a verification guard, no new dependencies.
- @JerrettDavis tagging you — this one can serve one conversation's cached response to another when the last message matches, so it seemed worth surfacing. Thanks!


JerrettDavis (approved):
Reviewed the semantic cache keying change. Keying by `messages_hash` when available prevents same-tail queries from clobbering each other, and the extra exact-match guard makes stale hash-index mappings fail closed instead of returning another conversation's response.

Verified locally:

```text
uv run pytest tests/test_cache/test_semantic.py -q
20 passed
```

CI is green and the branch is mergeable.

PR #2047, opened 2026-07-12
fix(proxy): forward X-Headroom-Base-Url custom upstream paths correctly

### LINKED ISSUE #2037: proxy: /v1/responses returns proxy_error when forwarding to custom upstream via X-Headroom-Base-Url
## Description

When using `headroom wrap codex` with a custom upstream provider (NexusMMO at `nexusmmo.store/api/v1`) via `HEADROOM_CODEX_UPSTREAM_BASE_URL`, the proxy fails to forward `/v1/responses` requests to the custom upstream. The same upstream handles `/v1/responses` correctly when called directly.

## Steps to reproduce

```bash
# Start proxy
HEADROOM_CODEX_UPSTREAM_BASE_URL="https://nexusmmo.store/api/v1" headroom wrap codex
```

Codex sends to `http://127.0.0.1:8787/v1/responses` with header `X-Headroom-Base-Url: https://nexusmmo.store/api/v1`.

## Expected behavior

Proxy forwards to the custom upstream and returns the response.

## Actual behavior

Proxy returns:
```json
{"error":{"message":"An error occurred while processing your request. Please try again.","type":"server_error","code":"proxy_error"}}
```

## Verified: upstream works directly

```bash
curl -s https://nexusmmo.store/api/v1/responses \
  -H "Authorization: Bearer $KEY" \
  -d '{"model":"gpt-5.5","input":"Say OK","max_output_tokens":10}'
# Returns: {"output":[{"content":[{"text":"OK"}]}],...}
```

## Environment

- headroom: 0.31.1 (installed via uv, git clone at commit ~2026-07-11)
- Python: 3.12.13
- macOS: 26.5.2 (arm64)
- Codex CLI: v0.144.1
- Proxy started with: `--backend anthropic --mode token` (defaults from `headroom wrap codex`)

## Notes

- `/v1/chat/completions` with `X-Headroom-Base-Url` also returns `proxy_error`
- Without `X-Headroom-Base-Url`, both endpoints forward to `api.openai.com` and return 401 (expected — wrong key for OpenAI)
- The proxy's startup routing table shows: `/v1/responses → https://api.openai.com (HTTP + WebSocket)` — it doesn't appear to respect `X-Headroom-Base-Url` for the responses endpoint
- Workaround: use Codex directly with the custom provider (bypass Headroom proxy). MCP tools (`headroom_compress`, `tokensave`) still work.

im-khang:
## Description

When `headroom wrap codex` proxies `/v1/responses` (and `/v1/chat/completions`) to a custom upstream configured via `HEADROOM_CODEX_UPSTREAM_BASE_URL` + `X-Headroom-Base-Url` header, three bugs prevent correct forwarding:

1. Path stripped: `_resolve_openai_upstream_base` discards the URL path component (e.g. `/api/v1`) after normalization
2. Double `/v1/` prefix: `build_copilot_upstream_url` appends `/v1/responses` onto a base that already ends with `/v1`, producing broken paths like `/api/v1/v1/responses`
3. WebSocket disconnect: `supports_websockets = true` causes Codex to use WebSocket transport, which the proxy cannot bridge to non-OpenAI upstreams

Closes #2037.

## Type of Change

- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation update
- [ ] Refactor
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Preserve custom upstream path components during OpenAI upstream-base resolution.
- Avoid a duplicate `/v1` when the custom base already ends in `/v1`.
- Preserve current default Codex WebSocket support and add routing/auth regression coverage.

### Bug 1: `_resolve_openai_upstream_base` strips path component

`headroom/proxy/handlers/openai.py:178-189`

`_normalize_origin()` returns only `scheme://hostname:port`, discarding the path segment (e.g. `/api/v1`). Added path re-attachment after normalization so custom upstreams with sub-paths survive the proxy.

```python
# Re-attach the path component that _normalize_origin dropped
parsed = urlparse(raw_base_url.strip())
path = (parsed.path or "").rstrip("/")
if path:
    normalized = f"{normalized}{path}"
```

### Bug 2: `build_copilot_upstream_url` doubles `/v1/` prefix

`headroom/copilot_auth.py:939-942`

When base URL already ends with `/v1` (e.g. `https://nexusmmo.store/api/v1`) and the handler path starts with `/v1/`, concatenation produced `/api/v1/v1/responses`. Added a guard to strip `/v1` from the handler path when the base already has it.

```python
if normalized_base.endswith("/v1") and normalized_path.startswith("/v1/"):
    normalized_path = normalized_path[3:]
```

### Bug 3: `supports_websockets = true` causes stream disconnect

Files: `headroom/providers/codex/install.py`, `headroom/cli/init.py`, `headroom/cli/wrap.py` (x2)

Changed `supports_websockets = true` → `false` in all 4 injection sites. Codex opens WebSocket to the provider when this flag is true, but the proxy cannot bridge WebSocket to non-OpenAI custom upstreams.

### Test updates

Updated unit tests (`tests/test_provider_codex_install.py`, `tests/test_install/test_providers.py`) and e2e tests (`e2e/wrap/run.py`, `e2e/init/run.py`) to expect `supports_websockets = false`.

## Real Behavior Proof

- Environment: current PR head with custom OpenAI-compatible upstream routing tests.
- Exact command / steps: ran 133 routing/auth tests, Ruff, formatting checks, and the new duplicate-/v1 regression after reconciling against current main.
- Observed result: the custom upstream base retains its path without producing a duplicate /v1 segment; default Codex WebSocket support remains enabled; all focused checks pass.
- Not tested: a live production custom upstream end-to-end session.

Before fix, the proxy returns `proxy_error` for `/v1/responses` with `X-Headroom-Base-Url: https://nexusmmo.store/api/v1`:

```json
{"error":{"message":"An error occurred while processing your request. Please try again.","type":"server_error","code":"proxy_error"}}
```

After fix (verified via syntax check + unit tests):

- Python AST validation passes on all 5 modified source files
- 9/9 targeted unit tests pass
- Direct curl to the same upstream with the corrected URL returns valid JSON

### Verification items

- [x] Python AST syntax check: all 5 modified files parse cleanly
- [x] Unit tests pass: `tests/test_provider_codex_install.py` (9/9)

### Test Output

```text
$ python -c "import ast; ast.parse(open('headroom/proxy/handlers/openai.py').read()); print('OK')"
OK
$ python -c "import ast; ast.parse(open('headroom/copilot_auth.py').read()); print('OK')"
OK
$ python -c "import ast; ast.parse(open('headroom/cli/init.py').read()); print('OK')"
OK
$ python -c "import ast; ast.parse(open('headroom/cli/wrap.py').read()); print('OK')"
OK
$ python -c "import ast; ast.parse(open('headroom/providers/codex/install.py').read()); print('OK')"
OK

$ pytest tests/test_provider_codex_install.py -v
...
tests/test_provider_codex_install.py .........                          [100%]
========================= 9 passed in 0.12s =========================
```

### Known Limitations

- Non-streaming responses with custom upstreams still return an empty body (JSONDecodeError). URL routing is correct and the upstream is reached (cf-ray present), but httpx body handling needs further investigation. This is a separate issue from the 3 bugs fixed here.

## Review Readiness

- [x] I have added comments in hard-to-understand areas
- [x] I have run relevant tests and they pass


JerrettDavis (changes requested):
Thanks for digging into the custom-upstream path handling. The `_resolve_openai_upstream_base` and `/v1` double-prefix fixes are directionally useful, but I don't think this PR is safe to merge as written because it changes the default Codex provider transport globally.

Required changes:

- Please do not change every generated Headroom Codex provider block from `supports_websockets = true` to `supports_websockets = false` in `headroom/cli/init.py`, `headroom/cli/wrap.py`, `headroom/providers/codex/install.py`, and the e2e/unit expectations. That disables WebSocket support for normal Headroom/Codex installs, even when the upstream is the standard Headroom proxy/OpenAI-compatible path. The issue described here is scoped to custom non-OpenAI upstreams, so the transport change needs to be conditional on that case or handled in the proxy path that cannot bridge the custom upstream, not by regressing the default provider capability.

- Please add direct tests for the URL fixes. I would expect coverage for `_resolve_openai_upstream_base` preserving a path such as `https://example.test/api/v1`, and `build_copilot_upstream_url("https://example.test/api/v1", "/v1/responses") == "https://example.test/api/v1/responses"`, while preserving the existing Copilot-host behavior. Right now the PR changes routing helpers but only updates tests around the WebSocket flag.

- The PR body says non-streaming responses with custom upstreams still return an empty body / JSONDecodeError. If that is still true, this should either be fixed here or the scope should be narrowed so the PR does not claim custom `/v1/responses` forwarding is fixed end-to-end. For this product, routing to the upstream but returning an empty body is still a user-visible proxy failure.

Once the WebSocket behavior is scoped and the helper tests are in place, the path-preservation part should be straightforward to re-review.

JerrettDavis (approved):
Resolved against current main. Preserved current Codex WebSocket support and main’s existing custom-path handling; retained the missing duplicate-/v1 fix for custom upstream bases with a new regression test. 133 routing/auth tests pass; Ruff and formatting pass.

PR #822, opened 2026-06-10
fix(windows): unwedge compression on degraded ONNX runtimes (every request timing out at 30s, 0% savings)

chopratejas:
## Summary

Multiple Windows users reported (via Discord, on v0.23.0, `pip install "headroom-ai[all]"`) that the proxy delivers **zero compression** and adds **+30s latency to every request**: `Optimization failed: TimeoutError:` with `compression_first_stage ≈ 30000ms` on every optimization attempt, for the lifetime of the process. Log analysis showed the wedge starts at the **first message eligible for real compression** (earlier requests succeed because everything is skipped/excluded) — and never recovers, even though the Kompress model loaded successfully at startup.

### Root cause chain

1. `create_cpu_session_options` disabled ONNX Runtime's CPU memory arena on **all** platforms. On Windows this is catastrophic: every `Run()` falls back to per-node `VirtualAlloc`/free, slowing ModernBERT inference by 2–3 orders of magnitude (onnxruntime#11627). One reporter's perf summary showed max optimization overhead of **200,369ms** (~13 chunks × ~15s) — slow, not deadlocked.
2. The first slow inference outlives the proxy's 30s compression-stage timeout. `asyncio.wait_for` abandons the future but **cannot kill the executor thread**, which keeps holding the Kompress `BoundedSemaphore(1)`.
3. Every later compression blocks on an **unbounded** `semaphore.acquire()`, times out at exactly 30s, and leaks another thread — permanently wedging the proxy until restart.

Two adjacent Windows bugs found in the same logs are fixed too: `subprocess.run(text=True)` without `encoding=` decodes child output with cp1252, so rtk's emoji output killed reader threads (`UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f`); and the OpenAI handler logged `Optimization failed: ` with an empty message because `str(asyncio.TimeoutError())` is empty.

### Fixes

- **`onnx_runtime.py`** — keep the CPU arena at ORT's default on Windows; Linux/macOS keep the legacy low-RSS behavior (arena disabled) bit-for-bit. New `HEADROOM_ONNX_CPU_ARENA` env overrides either way. All ONNX sessions (Kompress, image router, memory embedders) share this helper, so one fix covers them all.
- **`kompress_compressor.py`** — three layers of wedge-proofing, each fail-safing to passthrough instead of blocking:
  - bounded semaphore acquire (`HEADROOM_KOMPRESS_ACQUIRE_TIMEOUT_SECONDS`, default 5s)
[...]

PR #2051, opened 2026-07-12
docs(windows): document persistent-install gap on standard accounts + AtLogOn workaround

falling-ai:
## Summary

- `headroom install apply --preset persistent-service` (`sc.exe create`) and `--preset persistent-task` (`schtasks /SC ONSTART`) both require elevated/administrator rights on Windows and fail with **Access Denied** on a standard (non-administrator) account — currently the only durable, non-hook option in that case is `headroom init -g claude`, which sets `supervisor_kind = none` and relies on hook-triggered `init hook ensure` instead.
- Adds `wiki/windows-deployment.md` documenting this gap with exact reproduction output, plus a non-elevated **AtLogOn Scheduled Task** workaround (`scripts/windows/install-scheduled-task.ps1` / `uninstall-scheduled-task.ps1`) that a standard account *can* register without elevation.
- Also documents a hook-based `init hook ensure` race we hit on Windows: under slow ONNX/Kompress compression (>10s observed for a single request), `/readyz` can be slow enough that the ensure logic's internal wait/kill/restart path (and/or Claude Code's own 15s hook timeout) races the still-healthy proxy, kills it, and starts a duplicate that clobbers the `runner.pid` bookkeeping of the process that was actually fine. This looks like the Windows-side manifestation of #822 (still open at time of writing). The doc calls this out as a reason to keep the proxy's lifecycle owned by exactly one thing (hook *or* supervisor, not both).

## What changed

- `wiki/windows-deployment.md` (new)
- `scripts/windows/install-scheduled-task.ps1`, `scripts/windows/uninstall-scheduled-task.ps1` (new)
- `.gitignore` — allowlist entries for the two new scripts (repo's existing `scripts/*` + `!scripts/<file>` pattern)

This is docs + an optional operator-facing PowerShell workaround script, not a change to the Python proxy/install/runtime code. No existing behavior changes.

## Real behavior proof

**Setup tested on:** Windows 11 Pro (build 10.0.26200), a **standard (non-administrator)** Windows account (`BUILTIN\Administrators` is deny-only for this account), headroom-ai 0.31.0 installed from the PyPI `win_amd64` wheel, Claude Code CLI + Codex Desktop as the wrapped tools, against upstream `main` @ `2c9eb7c5` (`v0.31.0-93-g2c9eb7c5`) for this repro.

**Exact commands run:**

```
> headroom install apply --preset persistent-service --providers manual --target claude --target codex --telemetry
[SC] OpenSCManager FAILED 5:
Access is denied.
[SC] OpenService FAILED 1060:
The specified service does not exist as an installed service.
Error: Failed to install deployment 'default': Command 'sc.exe create headroom-default binPath= "cmd.exe /c \"C:\Users\<user>\.headroom\deploy\default\run-headroom.cmd\"" start= auto' returned non-zero exit status 5.

> headroom install remove --profile default
Error: No deployment profile named 'default' is installed.

> headroom install apply --preset persistent-task --providers manual --target claude --target codex --telemetry
ERROR: Access is denied.
Error: Failed to install deployment 'default': Command '['schtasks', '/Create', '/TN', 'headroom-default-startup', '/TR', 'C:\Users\<user>\.headroom\deploy\default\ensure-headroom.cmd', '/SC', 'ONSTART', '/F']' returned non-zero exit status 1.
```

**After-fix (workaround) evidence:**

```
> scripts\windows\install-scheduled-task.ps1 -PythonExe "C:\...\.venv\Scripts\python.exe"
Registered Scheduled Task 'HeadroomProxy' (AtLogOn, RunLevel=Limited, no elevation required).

> Start-ScheduledTask -TaskName HeadroomProxy
> headroom doctor
Headroom Doctor v0.31.0 · port 8787
┌───────────────────────┬────────┬────────────────────────────────────────────┐
│ proxy                 │ ✓ pass │ running at http://127.0.0.1:8787 (up 0m, v0.31.0) │
│ version               │ ✓ pass │ proxy matches installed v0.31.0            │
│ claude                │ ✓ pass │ routed via ~/.claude/settings.json         │
│ codex                 │ ✓ pass │ routed (~/.codex/config.toml)              │
│ deployments           │ ✓ pass │ 1 deployment(s) healthy                    │
└───────────────────────┴────────┴────────────────────────────────────────────┘
```

Confirmed the task registers and starts the proxy without any UAC/elevation prompt, and that it survived across multiple manual `Stop-Process` + wait cycles (auto-restart within ~1 minute per the task's restart policy).

**What we did NOT test:**

- An elevated/administrator account — we can't confirm from this machine whether `persistent-service` / `persistent-task` work correctly when run elevated; the doc only documents the standard-account failure mode.
- Windows Server editions, or multi-user/shared machines.
- This PR does not fix the underlying ONNX/Kompress compression slowness (#822) — the workaround only avoids the *symptom* (double-start under a hook/supervisor race) by giving the proxy a single lifecycle owner.
- No automated test is included — this is a docs + optional operator script contribution, not a Python code change, so `pytest` doesn't apply. Verification above is manual, on real hardware, as described.

## Review Readiness

- [x] Docs-only + optional PowerShell script, no core code changed
- [x] Real behavior proof included above
- [x] `.gitignore` updated to allowlist the new scripts per the repo's existing pattern
- [ ] Maintainer confirmation on whether `wiki/` structure/placement matches intent (happy to move/rename)


JerrettDavis (changes requested):
Leaving early feedback while this is still draft because this is a user-facing install script.

The scheduled-task workaround currently hard-codes `--telemetry` into the task action:

```powershell
$proxyArgs = "-m headroom.cli proxy --host 127.0.0.1 --port $Port --mode $Mode --backend $Backend --telemetry"
```

That opts every user of the workaround into telemetry by default, which conflicts with the current product posture that telemetry is explicit opt-in. A docs/workaround script should preserve the default unless the operator asks for telemetry.

Please make telemetry opt-in here, for example with a `[switch]$Telemetry` parameter that appends `--telemetry` only when provided, and update the docs example to show the default non-telemetry invocation plus an optional telemetry variant if useful.

Small secondary cleanup: the uninstall script synopsis says it “stops any running proxy process it launched,” but the implementation only stops/unregisters the scheduled task and explicitly does not kill an already-running proxy. Please make the synopsis match the behavior, or implement a safe stop if that is intended.

JerrettDavis (approved):
Resolved against current main. Preserved the non-admin AtLogOn workaround and current script allowlist, updated the now-merged #822 guidance, and avoided unintentionally opting users into telemetry. Both PowerShell scripts parse cleanly and the merged diff passes whitespace checks.

PR #2054, opened 2026-07-12
docs(proxy): document --protect-tool-results side effect on protect_recent_reads_fraction

### LINKED ISSUE #1921: protect_recent_reads_fraction reset via --protect-tool-results is undocumented and version-dependent
## Problem

Under `--mode token`, a long-running coding agent session (Claude Code, in my case) will eventually re-read a file it already read earlier, because headroom silently lossy-compresses old `Read`/`Glob`/`Grep`/`Write`/`Edit` tool outputs once they age past roughly the most recent 30% of the conversation. Nothing about the file changed; headroom just stops protecting it because it's old. The agent notices the compressed output has gaps and re-reads to recover, which wastes tokens and, worse, can act on the compressed version before it notices. This is the same failure class as #1307, and #1364 tried to fix it directly but was closed on the contribution-template bot check rather than a maintainer decision on the substance.

Root cause: `--mode token` overrides `ContentRouterConfig.protect_recent_reads_fraction` from the library's documented default of `0.0` ("protect ALL excluded-tool outputs, safest for coding agents") to `0.3`. That fraction is what creates the age cutoff.

## My fix

`--protect-tool-results <tool>` (added in #1374) already solves this completely today, no code change needed, but only as a side effect of how the two flags share state. `protect_recent_reads_fraction` is one proxy-wide float, not a per-tool setting. Whichever flag runs last in `server.py` wins:

```python
if is_token_mode(config.mode):
    router_config.protect_recent_reads_fraction = 0.3
if config.protect_tool_results:
    router_config.protect_recent_reads_fraction = 0.0   # runs second, wins
```

So naming *any* tool at all with `--protect-tool-results` (I used `Bash`) resets the fraction to `0.0` globally, which also fully restores protection for `Read`/`Glob`/`Grep`/`Write`/`Edit`, tools I never named. I confirmed this on a live 277-message session via `proxy.log`: `router:excluded:tool` protections kept firing, zero `kompress` hits on any excluded tool, well past the point where the original bug would have kicked in.

## Suggested change upstream

I'm not asking for new behavior; the fix already works for me. Two gaps worth closing for the next person who hits this:

1. **Documentation.** Nothing near `protect_recent_reads_fraction`, the `--mode token` help text, or the CHANGELOG mentions that `--protect-tool-results` also happens to disable the age-decay. Someone hitting the #1307 symptom under `--mode token` has no way to find this fix short of reading `server.py`. A one-line doc note or code comment pointing from one flag to the other would close this.
2. **Version dependency.** The fix only exists on versions with `--protect-tool-results` (#1374). Anyone on an older version (I was on 0.25.0) has no CLI/env workaround at all, short of `--mode cache` (which gives up token mode's other savings) or a source patch.

If a documentation fix isn't enough, I'd also be fine with a dedicated flag (e.g. `--protect-recent-reads-fraction`) so the override doesn't have to be reached through an unrelated tool-name flag, I'm happy to spec or implement that, but wanted to check whether option 1 is sufficient before writing code, per the new-feature process in CONTRIBUTING.md.

Referencing: #1307, #1364, #1374.

lennney:
## Description

Documents the side effect of `--protect-tool-results`: in token mode, naming
any tool also resets `protect_recent_reads_fraction` from 0.3 to 0.0 (Closes
#1921).  Previously this interaction was only visible in the source code
ordering in server.py and led to user confusion.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cli/proxy.py`: extend `--protect-tool-results` CLI help text to
  explain that in token mode it also resets `protect_recent_reads_fraction`
  from 0.3 → 0.0, restoring full indefinite protection for all excluded-tool
  results (Read/Glob/Grep/Write/Edit)
- `headroom/proxy/server.py`: add inline comment documenting the ordering
  dependency (token mode runs first, protect_tool_results overrides after)

## Testing

- [x] No functional changes (docs only)
- [ ] N/A — no code logic changes

### Test Output

```text
$ uv run ruff check headroom/cli/proxy.py headroom/proxy/server.py
All checks passed!
```

## Real Behavior Proof

- Environment: Linux, headroom main @ 868b88bc
- Exact command / steps: (1) `headroom proxy --help` → verify `--protect-tool-results` shows new help text, (2) read server.py line ~764 to verify comment documents the ordering dependency
- Observed result: Help text now explains the fraction-reset side effect; server.py comment makes the ordering dependency explicit
- Not tested: N/A — documentation-only change

## Review Readiness

### DIFF
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -285,6 +285,10 @@ def dashboard(port: int, no_open: bool) -> None:
     help=(
         "Comma-separated tool names whose results are never lossy-compressed, "
         "merged with the built-in defaults (e.g. Bash,WebFetch). "
+        "In token mode, this also resets protect_recent_reads_fraction "
+        "from 0.3 (only recent ~30% of results protected) to 0.0 (all "
+        "results protected indefinitely), which prevents older Read/Glob/"
+        "Grep/Edit tool results from being silently compressed. "
         "Env: HEADROOM_PROTECT_TOOL_RESULTS."
     ),
 )
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -288,7 +288,7 @@ def dashboard(port: int, no_open: bool) -> None:
         "In token mode, this also resets protect_recent_reads_fraction "
         "from 0.3 (only recent ~30% of results protected) to 0.0 (all "
         "results protected indefinitely), which prevents older Read/Glob/"
-        "Grep/Edit tool results from being silently compressed. "
+        "Grep/Write/Edit tool results from being silently compressed. "
         "Env: HEADROOM_PROTECT_TOOL_RESULTS."
     ),
 )
--- a/headroom/proxy/server.py
+++ b/headroom/proxy/server.py
@@ -761,6 +761,11 @@ class HeadroomProxy(
         if is_token_mode(config.mode):
             router_config.protect_recent_reads_fraction = 0.3
             router_config.search_group_by_file = True
+        # Note: protect_tool_results runs AFTER token mode (ordering matters).
+        # It resets protect_recent_reads_fraction from 0.3→0.0, restoring
+        # full protection for ALL excluded-tool results regardless of age.
+        # This means naming any tool with --protect-tool-results also
+        # protects Read/Glob/Grep/Write/Edit results indefinitely.
         if config.protect_tool_results:
             router_config.protect_recent_reads_fraction = 0.0
         # `--compress-user-messages` flips the router's default skip rule.


JerrettDavis (approved):
Reviewed the CLI help text and server-side ordering comment against the actual proxy initialization flow. The documented behavior matches the implementation: token mode first sets protect_recent_reads_fraction to 0.3, then --protect-tool-results resets it to 0.0.

Local validation:
- uv run --extra dev python -m headroom.cli proxy --help | Select-String -Pattern "protect-tool-results" -Context 0,6
- uvx --from ruff==0.15.17 ruff check headroom/cli/proxy.py headroom/proxy/server.py
- uvx --from ruff==0.15.17 ruff format --check headroom/cli/proxy.py headroom/proxy/server.py

PR #2070, opened 2026-07-12
fix(proxy/anthropic): scope session id by top-level system prompt

abhay-codes07:
## Description

`SessionTrackerStore.compute_session_id` (`headroom/cache/prefix_tracker.py`) computes a fallback
session id (when no `x-headroom-session-id` header is present) from `model` + system-prompt text.
But it harvests system text **only** from `messages` entries with `role == "system"`:

```python
for msg in messages:
    if msg.get("role") == "system":
        ...  # collect system text
system_content = json.dumps(system_parts, ...)
key = f"{model}:{system_content}"
```

Anthropic's `/v1/messages` carries the system prompt as a **top-level** `body["system"]` field —
it never sends `role:"system"` entries inside `messages`. And `x-headroom-session-id` is a
Headroom-internal header no client sends. So for every genuine Anthropic request `system_parts`
is empty and the id collapses to `md5(f"{model}:[]")` — **every conversation on the same model
shares one session id**, and therefore one `PrefixCacheTracker` and all session-sticky state.

The colliding state cross-contaminates across conversations (`anthropic.py:1052`):
- sticky `headroom_retrieve` / memory tools keyed purely on `session_id` (no content guard) get
  injected into another conversation's tool list — busting its tools cache and adding tools its
  client never requested;
- sticky `anthropic-beta` header tokens leak across conversations;
- `frozen_message_count` and the per-session compression cache cross-contaminate.

(The sibling `StreamingMixin._get_session_key` already reads `body.get("system")` and its docstring
claims to mirror `compute_session_id` — which it did not.)

Closes: no issue filed — found while auditing the session/prefix tracker.

## Fix

Add an optional `system` parameter to `compute_session_id` and fold its text (a plain string or a
list of `{"type":"text"}` blocks) into the hash. The Anthropic handler passes `body.get("system")`.
OpenAI callers don't pass it (defaults to `None`), so their behavior is unchanged.

## Type of Change

## Changes Made

- `headroom/cache/prefix_tracker.py`: `compute_session_id` accepts an optional `system` and folds it into the id.
- `headroom/proxy/handlers/anthropic.py`: pass `system=body.get("system")` when computing the session id.
- `tests/test_cache/test_prefix_tracker.py`: add `test_compute_session_id_distinguishes_top_level_system` (distinct systems → distinct ids; list-form == string-form; `system=None` unchanged).

## Testing

- [x] New regression test added (`tests/test_cache/test_prefix_tracker.py`)

```text
$ uvx ruff@0.15.17 check headroom/cache/prefix_tracker.py headroom/proxy/handlers/anthropic.py tests/test_cache/test_prefix_tracker.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.10, headroom from this branch. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the hash logic with a dependency-free script and left the full pytest to CI.
- Exact command / steps: computed ids for two conversations with the same model and messages but different top-level `system` prompts, through the old (never-folds-system) and new logic.
- Observed result: the old logic collapses both to one id (the leak); the new logic separates them, folds list-form system the same as string-form, and leaves the `system=None` (OpenAI) path unchanged:

```text
OLD: A=97d8857ba27010bb  B=97d8857ba27010bb  same=True
NEW: A=1e838c0f6e3980a6  B=18ec49bfa8240852  same=False
SESSION-ID SYSTEM FIX VERIFIED (old collapses Anthropic convos; new separates them)
```

- Not tested: a full two-conversation proxy run asserting no sticky-tool leakage (needs the heavy stack). The fix is confined to `compute_session_id` + the one handler call site, and the new test drives the method directly. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- Backward-compatible: the new `system` parameter defaults to `None`, so the OpenAI call sites (`openai.py`) need no change and their session ids are identical.
- @JerrettDavis tagging you — this one lets one Anthropic conversation's sticky tools/headers leak into another on the same model, so it seemed worth surfacing. Thanks!


chopratejas:
Hi - thanks for the change - can you fix the merge conflicts and the test?

abhay-codes07:
Thanks, both fixed. Rebased on latest main (the conflict was just a CHANGELOG overlap).

The test failure was on me. My first cut added a `system` keyword to `compute_session_id`, which broke all the existing test stubs that mock it with a fixed `lambda request, model, messages:` signature. I reworked it so there's no signature change at all: the Anthropic handler now folds the top-level `system` in as a synthetic `role:"system"` message before calling `compute_session_id` (which already hashes system messages). That message is only used to derive the id, never forwarded. So none of those mocks need touching and the diff stays small. Green now.

JerrettDavis (approved):
Reviewed. The revised approach keeps `compute_session_id`'s signature stable and scopes the behavior change to Anthropic by adding the top-level `system` field as a synthetic system message only for session-id calculation. That addresses the same-model Anthropic collision without affecting OpenAI call sites or forwarded request payloads.

Validated locally:

- uv run --extra dev python -m pytest tests/test_cache/test_prefix_tracker.py -q
- uv run --extra dev ruff check headroom/cache/prefix_tracker.py headroom/proxy/handlers/anthropic.py tests/test_cache/test_prefix_tracker.py

PR #836, opened 2026-06-10
fix: correct preserved-entry index mapping in Gemini content round-trip

agaonker:
## Summary

- `_gemini_contents_to_messages` excludes entries with no text parts (pure `functionCall` / `functionResponse` / image-only) from `messages[]`, but their original `contents[]` indices are stored in `preserved_indices`
- After compression, `optimized_contents` has a shorter, different index space — the old restoration loop used raw `orig_idx` to overwrite `optimized_contents[orig_idx]`, silently corrupting text entries at colliding positions and silently dropping preserved entries when `orig_idx >= len(optimized_contents)`
- Affects all three Gemini handlers (`generateContent`, `cloudCodeAssist`, `countTokens`) — any agentic session with function calls where compression fires

**Concrete failure case:**
```
contents = [user:text, model:functionCall, user:functionResponse, model:text]
messages = [user:text, model:text]          # only 2 — FC/FR have no text
optimized_contents = [user:text, model:text]  # positions 0 and 1

old loop:
  orig_idx=1 → optimized_contents[1] = functionCall  ← overwrites model text!
  orig_idx=2 → 2 < 2 is False → functionResponse silently dropped
```

## Fix

Added `_rebuild_gemini_contents()` helper that walks `original_contents` in order, placing preserved entries at their exact relative positions and consuming optimized text entries sequentially via an iterator. Replaced all three broken loops.

## Test plan

- [ ] `TestRebuildGeminiContents::test_text_only_unchanged` — text-only round-trip is identity
- [ ] `TestRebuildGeminiContents::test_function_call_sequence_preserved` — functionCall + functionResponse survive at correct positions
- [ ] `TestRebuildGeminiContents::test_function_call_at_start` — preserved entry at idx=0 no longer overwrites optimized_contents[0]
- [ ] `TestRebuildGeminiContents::test_hybrid_entry_uses_original` — entry with both text and functionCall retains functionCall

All 58 tests in `test_google_multimodal.py` pass. Rust CI + mypy + ruff clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2075, opened 2026-07-12
fix(proxy/batch): rebuild Google batch contents by original position

abhay-codes07:
## Description

The Google `batchGenerateContent` handler (`headroom/proxy/handlers/batch.py`) restores the
preserved (non-text) content entries after compression with a raw-index loop:

```python
# Restore preserved content entries that had non-text parts
for orig_idx, original_content in preserved_contents.items():
    if orig_idx < len(optimized_contents):
        optimized_contents[orig_idx] = original_content
```

`preserved_indices` are indices into the **original** `contents[]`, but `optimized_contents`
lives in a **shorter** index space: text-less entries (functionCall / functionResponse / images)
produce no message and so no optimized entry. Indexing `optimized_contents` by `orig_idx`
therefore overwrites the wrong entry, and silently drops any preserved entry whose original
index is `>= len(optimized_contents)`.

This is the exact bug that commit #836 (`_rebuild_gemini_contents`) fixed for the three non-batch
Gemini handlers in `gemini.py`; the batch handler was never migrated. Unlike `gemini.py` there is
no `if optimized_messages != messages` gate here, so it runs on every batch item that mixes text
and non-text entries, and the corrupted result is forwarded to Google unconditionally.

Concrete example (one request inside the batch):

```python
contents = [
    {"role": "user",  "parts": [{"text": "What's the weather in Paris?"}]},          # idx0 -> msg
    {"role": "model", "parts": [{"functionCall": {"name": "get_weather", ...}}]},    # idx1: no text -> preserved, no msg
    {"role": "user",  "parts": [{"functionResponse": {"name": "get_weather", ...}}]},# idx2: no text -> preserved, no msg
    {"role": "model", "parts": [{"text": "It's 18C and cloudy in Paris."}]},         # idx3 -> msg
]
```

`messages = [user Q, model answer]`, `preserved_indices = {1, 2}`; after compression
`optimized_contents` has 2 entries. The loop then does:
- `orig_idx=1`: `1 < 2` -> overwrites the **model's answer** with the functionCall;
- `orig_idx=2`: `2 < 2` is false -> the **functionResponse is dropped**.

Result forwarded upstream: `[user Q, functionCall]` (2 entries) instead of the 4 the client sent.

Closes: no issue filed — found while auditing the Gemini contents<->messages round-trip.

## Fix

Replace the loop with the shared `_rebuild_gemini_contents(contents, preserved_indices,
preserved_contents, optimized_contents)` helper the non-batch handlers already use. It walks the
original contents in order, placing each preserved entry at its original position and the next
optimized text entry otherwise. The helper is defined on the shared `GeminiHandlerMixin`, so it's
already available on the batch handler instance.

## Type of Change

## Changes Made

- `headroom/proxy/handlers/batch.py`: use `_rebuild_gemini_contents` instead of the raw-index restore loop.
- `tests/test_proxy_handlers_batch.py`: add `test_handle_google_batch_create_preserves_functioncall_response_order`, driving the handler with the **real** Gemini converters (the existing batch tests stub them, which hid this bug) and asserting all four entries survive in order. Also mix `GeminiHandlerMixin` into the shared `DummyBatchHandler` so `_rebuild_gemini_contents` is available to the existing tests.

## Testing

- [x] New regression test added (`tests/test_proxy_handlers_batch.py`)

```text
$ uvx ruff@0.15.17 check headroom/proxy/handlers/batch.py tests/test_proxy_handlers_batch.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.10, headroom from this branch. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the interleaving with a dependency-free script (replicating the Gemini converters, the old loop, and `_rebuild_gemini_contents`) and left the full pytest to CI.
- Exact command / steps: ran the four-entry text/functionCall/functionResponse/text contents through the old restore loop and the new `_rebuild_gemini_contents`.
- Observed result: the old loop returns 2 entries (model answer overwritten by the functionCall, functionResponse dropped); the new helper returns all 4 in order with the preserved entries verbatim:

```text
preserved_indices: [1, 2]
OLD result parts: ['text', 'functionCall']  len 2
NEW result parts: ['text', 'functionCall', 'functionResponse', 'text']  len 4
GEMINI BATCH REBUILD FIX VERIFIED (old drops response + overwrites answer; new keeps all 4)
```

- Not tested: a live Google batch round-trip (handlers stubbed, as the existing tests do). The new test drives the real converters through `handle_google_batch_create` and asserts the forwarded contents. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- One-line behavioral change (the loop becomes the shared helper call); the rest is the regression test. Completes the #836 migration for the one handler that was missed.
- @JerrettDavis tagging you since you reviewed the recent Gemini fixes. This one drops functionResponse/image content from Google batch requests, so it seemed worth surfacing. Thanks.


JerrettDavis (approved):
Looks good. The batch path was using original contents[] indices against the shorter optimized contents list, and switching to the shared Gemini interleaving helper matches the non-batch handlers. The production HeadroomProxy MRO already includes GeminiHandlerMixin, and the regression exercises the real converters for the dropped functionResponse/overwritten-answer case.

abhay-codes07:
Superseded by #2079, which combines this batch fix with the related executableCode/codeExecutionResult preservation fix into one PR. Same change and same regression test, just bundled.

PR #2076, opened 2026-07-12
fix(proxy/gemini): preserve executableCode/codeExecutionResult parts

abhay-codes07:
## Description

`_has_non_text_parts` (`headroom/proxy/handlers/gemini.py`) decides whether a Gemini content
entry must be preserved verbatim (excluded from compression and restored in place). It only
recognizes four keys:

```python
for key in ("inlineData", "fileData", "functionCall", "functionResponse")
```

Gemini's code-execution feature emits parts keyed `executableCode` and `codeExecutionResult`,
which are echoed back inside `contents[]` on subsequent turns. These are neither text nor in the
allowlist, so the entry is not preserved. That produces two failure modes on the round-trip
(`_gemini_contents_to_messages` → compress → `_messages_to_gemini_contents` → `_rebuild_gemini_contents`),
reachable from `handle_gemini_generate_content`, the Cloud Code stream, `handle_gemini_count_tokens`,
and the Google batch path:

**(a) Mixed entry, code dropped.** For `{"role":"model","parts":[{"text":"Running:"},{"executableCode":{...}}]}`,
`_has_non_text_parts` returns `False`, only the text is emitted as a message, and the rebuild
reconstructs a text-only entry — the `executableCode` payload is gone from the upstream request.

**(b) Text-less entry, turn dropped and neighbor shifted.** For a text-less
`executableCode`+`codeExecutionResult` entry between two text turns, it is not preserved and
produces no message, so in `_rebuild_gemini_contents` its `else` branch consumes the *next*
optimized message: the entire code turn is dropped and the following user turn is placed into the
model turn's slot, corrupting role alternation.

Closes: no issue filed — found while auditing the Gemini contents<->messages round-trip.

## Fix

Add `executableCode` and `codeExecutionResult` to the non-text detection tuple, so those entries
are preserved verbatim like images and function calls.

## Type of Change

## Changes Made

- `headroom/proxy/handlers/gemini.py`: recognize `executableCode` / `codeExecutionResult` in `_has_non_text_parts`.
- `tests/test_google_multimodal.py`: extend the parametrized `test_each_non_text_key_detected` to the two new keys, and add `test_code_execution_entry_survives` (a text-less code-execution entry survives the round-trip at its position without shifting a neighbor).

## Testing

- [x] New regression tests added (`tests/test_google_multimodal.py`)

```text
$ uvx ruff@0.15.17 check headroom/proxy/handlers/gemini.py tests/test_google_multimodal.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.10, headroom from this branch. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the detection + round-trip with a dependency-free script and left the full pytest to CI.
- Exact command / steps: ran the mixed `text`+`executableCode` entry and a text-less code-execution entry (between two text turns) through the old and new key sets.
- Observed result: the old set drops the code payload and (for the text-less case) drops the turn and shifts a neighbor; the new set preserves both:

```text
(b) OLD len=2  NEW len=3
(a) OLD has code=False  NEW has code=True
GEMINI CODE-EXECUTION PRESERVE FIX VERIFIED (old drops/shifts; new keeps intact)
```

- Not tested: a live Gemini code-execution round-trip (needs the heavy stack). The tests drive the real `_has_non_text_parts` / round-trip helpers. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- Two keys added to an existing allowlist plus tests; no new dependencies. Consistent with how images and function calls are already preserved.
- @JerrettDavis tagging you since you reviewed the recent Gemini fixes. This one drops code-execution content (and can shift a neighboring turn) on the Gemini round-trip. Thanks.


JerrettDavis (approved):
Looks good. Treating executableCode/codeExecutionResult as preserved non-text parts fits the existing Gemini round-trip model: text-less entries stay in place, and mixed entries keep their original payload while consuming the generated text message. The added regression covers the neighbor-shift failure mode.

abhay-codes07:
Superseded by #2079, which bundles this fix with the related Google batch rebuild fix into one PR. Same change and test.

PR #2079, opened 2026-07-12
fix(proxy/gemini): preserve non-text content across the compression round-trip

abhay-codes07:
## Description

Two related content-loss bugs in the Gemini `contents[]` <-> `messages[]` compression round-trip.
Both drop or misplace real user content that entries with **non-text** parts should carry through
untouched. They share the same theme (non-text preservation), so they're bundled here as two
commits.

### 1. Google batch handler restores preserved entries by the wrong index (`handlers/batch.py`)

The `batchGenerateContent` handler restored preserved (non-text) entries with the raw-index loop
that commit #836 (`_rebuild_gemini_contents`) replaced in the three non-batch Gemini handlers:

```python
for orig_idx, original_content in preserved_contents.items():
    if orig_idx < len(optimized_contents):
        optimized_contents[orig_idx] = original_content
```

`preserved_indices` are indices into the **original** `contents[]`, but `optimized_contents` is a
**shorter** list (text-less entries produce no message). Indexing `optimized_contents` by
`orig_idx` overwrites the wrong entry and drops any preserved entry whose original index is past
the optimized length. For:

```python
[user text, model functionCall, user functionResponse, model text]
```

the batch was forwarded to Google as **two** entries: the model's answer overwritten by the
functionCall, and the functionResponse dropped. Unlike `gemini.py` there is no
`if optimized_messages != messages` gate, so it runs on every mixed batch item.

**Fix:** use the shared `_rebuild_gemini_contents` interleaving helper.

### 2. Code-execution parts not detected as non-text (`handlers/gemini.py`)

`_has_non_text_parts` only recognized `inlineData`/`fileData`/`functionCall`/`functionResponse`.
Gemini's code-execution feature emits `executableCode` and `codeExecutionResult` parts, echoed
back in `contents[]` on later turns. Because they weren't detected:

- a mixed `text`+`executableCode` entry lost its code payload (only the text survived the round-trip);
- a text-less `executableCode`+`codeExecutionResult` entry was treated as a phantom in
  `_rebuild_gemini_contents` — it consumed the next optimized message, dropping the whole code turn
  and shifting a following user turn into the model's role slot (corrupting role alternation).

**Fix:** add both keys to the non-text detection so those entries are preserved verbatim.

Closes: no issue filed — both found while auditing the Gemini contents<->messages round-trip.

## Type of Change

## Changes Made

- `headroom/proxy/handlers/batch.py`: use `_rebuild_gemini_contents` instead of the raw-index restore loop.
- `headroom/proxy/handlers/gemini.py`: recognize `executableCode` / `codeExecutionResult` in `_has_non_text_parts`.
- `tests/test_proxy_handlers_batch.py`: add `test_handle_google_batch_create_preserves_functioncall_response_order`, driving the handler with the **real** Gemini converters (the existing batch tests stub them, which hid the bug); mix `GeminiHandlerMixin` into the shared `DummyBatchHandler` so `_rebuild_gemini_contents` is available.
- `tests/test_google_multimodal.py`: extend the parametrized `test_each_non_text_key_detected` to the two new keys, and add `test_code_execution_entry_survives`.

## Testing

- [x] New regression tests added (`tests/test_proxy_handlers_batch.py`, `tests/test_google_multimodal.py`)

```text
$ uvx ruff@0.15.17 check headroom/proxy/handlers/batch.py headroom/proxy/handlers/gemini.py \
    tests/test_proxy_handlers_batch.py tests/test_google_multimodal.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.10, headroom from this branch. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the interleaving/detection with dependency-free scripts (replicating the Gemini converters, the old loop, and `_rebuild_gemini_contents`) and left the full pytest to CI.
- Exact command / steps: ran two standalone scripts. Script 1 rebuilds a Gemini batch request with `preserved_indices` holding a `functionCall`/`functionResponse` pair and compares the old raw-index loop against `_rebuild_gemini_contents`. Script 2 feeds a `codeExecutionResult` entry through `_has_non_text_parts` and the preserve path with and without the two new allowlist keys. Also ran `uvx ruff@0.15.17 check` on the changed files and tests.
- Observed result: the old batch loop drops the `functionResponse` and overwrites the answer (4 parts collapse to 2); `_rebuild_gemini_contents` keeps all 4. Without the new keys the code-execution entry is dropped/shifted (2 parts, code absent); with them it survives intact (3 parts, code present). Lint clean. See the two blocks below.

Batch fix (bug #1):

```text
preserved_indices: [1, 2]
OLD result parts: ['text', 'functionCall']  len 2
NEW result parts: ['text', 'functionCall', 'functionResponse', 'text']  len 4
GEMINI BATCH REBUILD FIX VERIFIED (old drops response + overwrites answer; new keeps all 4)
```

Code-execution fix (bug #2):

```text
(b) OLD len=2  NEW len=3
(a) OLD has code=False  NEW has code=True
GEMINI CODE-EXECUTION PRESERVE FIX VERIFIED (old drops/shifts; new keeps intact)
```

- Not tested: a live Google/Gemini round-trip (handlers stubbed, as the existing tests do). Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

- [ ] New and existing unit tests pass locally with my changes — ran lint + standalone logic checks; full pytest deferred to CI (local OOM, disclosed above)

## Additional Notes

- Two small behavioral changes (one loop -> shared helper, two keys added to an allowlist) plus regression tests; no new dependencies. Both complete/extend the non-text preservation the non-batch handlers already do (the #836 line).
- @JerrettDavis tagging you since you reviewed the recent Gemini fixes. Both of these drop content (functionResponse/images on batch; code-execution on the normal round-trip), so they seemed worth surfacing together. Thanks.


JerrettDavis (changes requested):
The new batch regression test does not currently run. `RealConvHandler` only inherits `BatchHandlerMixin` and `GeminiHandlerMixin`, so this line raises before the handler path is exercised:

```python
monkeypatch.setattr(handler, "_retry_request", retry)
```

I reproduced it with:

```text
python -m pytest tests/test_proxy_handlers_batch.py -q
```

which fails as:

```text
AttributeError: <...RealConvHandler object ...> has no attribute '_retry_request'
```

The existing `DummyBatchHandler` defines `_retry_request`; this local real-converter harness needs the same stub, or the monkeypatch needs `raising=False`. As written, CI will fail on the new test even though the production code path may be fine.

abhay-codes07:
Good catch, thanks. You're right. `RealConvHandler` never defined `_retry_request`, so `monkeypatch.setattr(..., raising=True)` blew up before the handler ran. I verified the fix logic with a standalone script rather than the full pytest (the torch/transformers import OOMs my box), so I didn't catch the harness gap.

Fixed by defining `_retry_request` directly on the harness so it captures the forwarded batch body, and dropped the monkeypatch. I also walked the whole `handle_google_batch_create` path against the harness to confirm every `self.*` it reaches is now provided (`_google_batch_passthrough` is only hit for an empty requests list, which this test doesn't use). Pushed.

JerrettDavis (approved):
Re-reviewed the updated head after the test harness follow-up. The previous blocker is resolved: the real-converter batch regression now gives the harness its own _retry_request, so the handler path is exercised instead of failing during monkeypatch setup.

Local verification:

```text
python -m pytest tests/test_proxy_handlers_batch.py -q
# 21 passed

uvx ruff==0.15.17 check headroom/proxy/handlers/batch.py headroom/proxy/handlers/gemini.py tests/test_proxy_handlers_batch.py tests/test_google_multimodal.py
uvx ruff==0.15.17 format --check headroom/proxy/handlers/batch.py headroom/proxy/handlers/gemini.py tests/test_proxy_handlers_batch.py tests/test_google_multimodal.py
```
I also attempted the combined target including tests/test_google_multimodal.py, but this local checkout cannot collect that file because headroom._core is not present. The scoped batch regression that failed in the earlier review is green, and the implementation remains narrowly focused on preserving Gemini non-text entries in their original positions.

PR #2080, opened 2026-07-12
fix(tokenizers): price CJK in the fixed-ratio estimator path

abhay-codes07:
## Description

`EstimatingTokenCounter.count_text` (`headroom/tokenizers/estimator.py`) prices dense scripts
(CJK / Kana / Hangul) at ~1 token per 1.5 chars, because at the Latin ratio they undercount 4-6x.
But that correction is applied **only on the auto-detect path**; the fixed-ratio early return
divides by the Latin ratio with no adjustment:

```python
if self._fixed_ratio is not None:
    return max(1, int(len(text) / self._fixed_ratio + 0.5))   # no CJK split

# auto path (below) does the split:
cjk_chars = self._count_cjk_chars(text)
other_chars = len(text) - cjk_chars
base_count = int(other_chars / ratio + cjk_chars / self.CHARS_PER_TOKEN_CJK + 0.5)
```

The registry builds **every** provider-calibrated counter with a fixed ratio — Anthropic 3.5,
Google 4.0, Cohere 4.0, Moonshot 3.1 (`registry.py`) — and this is the live proxy count path:
the Anthropic handler (`_count_tokens_offloaded` → `get_tokenizer(model).count_messages`) and the
Gemini handlers resolve to these counters. So a CJK-heavy conversation reads as ~40-55% of its
true token size:

- a large CJK context can fall under the size / backpressure / background-compression gates and
  **skip compression** entirely;
- every `x-headroom-tokens-before` metric for CJK traffic is materially wrong.

(OpenAI is unaffected — its provider uses tiktoken, which tokenizes CJK correctly.)

Git blame confirms this is an oversight: commit `a35fe86e` ("price CJK ... in
EstimatingTokenCounter") added the split to the auto path but never touched the fixed-ratio return.

Closes: no issue filed — found while auditing the token counters.

## Fix

Apply the same dense-script split in the fixed-ratio branch.

## Type of Change

## Changes Made

- `headroom/tokenizers/estimator.py`: fixed-ratio path now prices CJK chars at `CHARS_PER_TOKEN_CJK` and the rest at the fixed ratio.
- `tests/test_tokenizers.py`: add `test_count_text_fixed_ratio_prices_cjk` (CJK priced ~len/1.5, ASCII unchanged).

## Testing

- [x] New regression test added (`tests/test_tokenizers.py`)

```text
$ uvx ruff@0.15.17 check headroom/tokenizers/estimator.py tests/test_tokenizers.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.10, headroom from this branch. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the count logic with a dependency-free script (replicating `CJK_PATTERN` and the count) and left the full pytest to CI.
- Exact command / steps: ran a ~99k-char Japanese string through the old and new logic at the Anthropic (3.5) and Google (4.0) fixed ratios, plus the ASCII case.
- Observed result: the old logic undercounts CJK ~2.3-2.7x; the new prices it near `len/1.5`; ASCII is unchanged:

```text
Japanese (99000 chars) @3.5: OLD=28286  NEW=66000  ratio=2.33x
Japanese @4.0: OLD=24750  NEW=66000  ratio=2.67x
mixed: OLD=54  NEW=81
CJK FIXED-RATIO FIX VERIFIED (old undercounts CJK ~2.3-2.7x; new prices it; ASCII unchanged)
```

- Not tested: a full proxy count over a real CJK request (needs the heavy stack). The fix is confined to `count_text` and the new test drives it directly. The existing ASCII-only `test_count_text_fixed_ratio` stays green. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- No signature change; the registry is the only construction site. Reuses the existing `_count_cjk_chars` / `CHARS_PER_TOKEN_CJK`.
- @JerrettDavis tagging you — this makes CJK contexts read as roughly half their real token size on the Anthropic/Gemini count path, so it seemed worth surfacing. Thanks.


JerrettDavis (approved):
Reviewed the fixed-ratio CJK estimator change. The bug is real: provider-calibrated EstimatingTokenCounter instances bypassed the auto-detect branch, so dense scripts were still divided by the Latin fixed ratio. Reusing _count_cjk_chars and CHARS_PER_TOKEN_CJK in the fixed-ratio path is the right scoped fix, and the regression test covers both the CJK increase and unchanged ASCII behavior.

Local verification:

```text
python -m pytest tests/test_tokenizers.py -q
# 49 passed, 14 skipped

uvx ruff==0.15.17 check headroom/tokenizers/estimator.py tests/test_tokenizers.py
uvx ruff==0.15.17 format --check headroom/tokenizers/estimator.py tests/test_tokenizers.py
```
The diff is small, the changelog entry matches the behavior change, and I do not see a correctness blocker.

PR #2082, opened 2026-07-12
fix(cache/ccr): don't evict a live entry on a duplicate store at capacity

abhay-codes07:
## Description

`CompressionStore.store` (`headroom/cache/compression_store.py`) runs eviction **before** it knows
whether the incoming `hash_key` is new or a re-store of an already-present key:

```python
with self._lock:
    self._evict_if_needed()          # <-- runs first

    existing = self._backend.get(hash_key)
    if existing is not None:
        ...                          # duplicate / collision: overwrite in place
        self._stale_heap_entries += 1
    self._backend.set(hash_key, entry)
```

When the store is full and the incoming key **already exists** (a duplicate re-store),
`_evict_if_needed()` removes the oldest *distinct* entry to "make room" — but then `set()` merely
overwrites the existing key in place, so no room was ever needed. Net effect: `count` drops to
`max_entries - 1` and a **live, never-retrieved entry is destroyed**. That entry's `<<ccr:...>>`
marker, still sitting in the conversation history, then resolves to a 404 on `/v1/retrieve`.

This is not a corner case: the CCR mirror bridge (`_mirror_single_hash_to_python_store` in
`smart_crusher.py`) re-`store()`s the same `explicit_hash` every turn a `<<ccr:…>>` marker is
re-encountered, and markers persist across turns — so a full store silently deletes a live sibling
entry on each duplicate.

Concrete (with the repo's `max_entries=3` fixture): store c0, c1, c2 (full), then re-store c1
(same content ⇒ same hash). Eviction pops the oldest (c0), deletes it, then c1 is overwritten in
place. Final state: {c1, c2}, count 2, and **c0 is gone** — its marker is now unredeemable.

Closes: no issue filed — found while auditing the compression store.

## Fix

Decide novelty before evicting: only `_evict_if_needed()` for a genuinely new key; a
duplicate/replace overwrites in place (no eviction). The collision/duplicate logging and
stale-heap accounting are unchanged.

## Type of Change

## Changes Made

- `headroom/cache/compression_store.py`: `store()` reads `existing` first and only evicts when the key is new.
- `tests/test_compression_store.py`: add `test_duplicate_store_at_capacity_does_not_evict` (re-store an existing hash at capacity keeps all entries and count at `max_entries`).

## Testing

- [x] New regression test added (`tests/test_compression_store.py`)

```text
$ uvx ruff@0.15.17 check headroom/cache/compression_store.py tests/test_compression_store.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.10, headroom from this branch. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the store/evict logic with a dependency-free script (replicating the backend + eviction heap) and left the full pytest to CI.
- Exact command / steps: filled a `max_entries=3` store with c0/c1/c2, then re-stored c1 (duplicate), through the old (evict-first) and new (check-first) logic; also confirmed a genuinely new key still evicts the oldest.
- Observed result: the old logic drops c0 (count 2); the new keeps all three; and a new key at capacity still evicts the oldest:

```text
OLD: after duplicate re-store of h1 -> keys=['h1', 'h2'] count=2
NEW: after duplicate re-store of h1 -> keys=['h0', 'h1', 'h2'] count=3
NEW still evicts oldest for a genuinely new key at capacity
DUPLICATE-STORE EVICTION FIX VERIFIED (old drops a live entry; new keeps it)
```

- Not tested: a full proxy CCR round-trip (needs the heavy stack). The fix is confined to `store()` and the new test drives it directly with the `max_entries=3` fixture. Existing eviction tests use distinct keys and stay green. Full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

- Internal reordering only, no signature change; no call sites or backend mocks break.
- @JerrettDavis tagging you — this silently drops a live CCR entry (making its marker 404) whenever a duplicate hash is re-stored at capacity, which the mirror bridge does routinely. Thanks.


JerrettDavis (approved):
Reviewed the cache-store reorder and regression. The fix only skips eviction for an existing hash, while preserving the normal new-key eviction path; the added capacity regression covers the CCR duplicate-store failure mode. Local verification: python -m pytest tests/test_compression_store.py -q (74 passed); uvx ruff==0.15.17 check headroom/cache/compression_store.py tests/test_compression_store.py; uvx ruff==0.15.17 format --check headroom/cache/compression_store.py tests/test_compression_store.py.

PR #2083, opened 2026-07-13
fix(core): add CCR min-size floor and freshness exemption

Eyalm321:
## Description

Closes #3

SmartCrusher's opaque-blob CCR substitution used only `opaque_min_bytes` (256 bytes, a shape-detection heuristic) to decide whether a string gets replaced with a `<<ccr:HASH,...>>` marker. In practice this meant fields as small as ~1KB were substituted on their very first, same-turn read — the `headroom_retrieve` round-trip cost outweighs any token savings for a field that small, and same-turn tool results shouldn't be swapped out before the model has a chance to read them. Additionally, the live proxy's `/v1/messages` compression path called the freshness-unaware `compress_anthropic_live_zone` (`fresh_message_count` implicitly `0`), so even the freshness fix itself was never exercised by real traffic — only by headroom-core's own tests.

## Type of Change
- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (a fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made
- Added `ccr_min_bytes` (`ClassifyConfig`) / `opaque_min_bytes` (`SmartCrusherConfig`): a substitution-policy floor, separate from the existing `opaque_min_bytes` shape-detection threshold (256 bytes). Cells/strings that classify as opaque but don't clear this floor (default 2048 bytes) render verbatim instead of getting a marker. Wired through `process_string`, the table-cell compactor (`cell_from_value`), and the document walker's opaque-string path, so all three CCR substitution sites agree.
- Added `fresh_message_count` exemption in the live-zone dispatcher: the most recent N messages in a turn are exempt from CCR substitution regardless of size, so an in-flight/just-produced tool result can't be swapped out for a marker before the model reads it.
- Wired `crates/headroom-proxy`'s real `/v1/messages` call site to `compress_anthropic_live_zone_with_ccr_and_freshness` with `fresh_message_count = 1`, matching the function's own doc comment ("Production wiring (the proxy) should pass 1"). This was the production gap flagged in review — the size-floor + freshness logic existed but wasn't reachable from real traffic before this change.
- Bumped two existing test fixtures (`test_smart_crusher_toin_attachment.py`, `test_transforms/test_smart_crusher_ccr_roundtrip.py`) whose blobs were sized just above the old 256-byte threshold but below the new 2048-byte floor, so they stay above both.

## Testing

- [x] Unit tests pass (`cargo test -p headroom-core -p headroom-proxy`)
- [x] Linting passes (`cargo clippy -p headroom-core -p headroom-proxy`, `cargo fmt --all -- --check`)

### Test Output

```text
cargo test -p headroom-core --test live_zone_freshness
cargo test -p headroom-proxy fresh_tail_message_is_not_compressed_through_proxy_entry_point
cargo test -p headroom-proxy aged_message_still_compresses_while_fresh_tail_is_untouched_through_proxy
cargo check -p headroom-proxy
cargo fmt --all -- --check
-> all green (confirmed independently by @JerrettDavis on commit abd66a1 per review)
```

## Real Behavior Proof
- Environment: local `cargo test`/`clippy`/`fmt` against `crates/headroom-core` and `crates/headroom-proxy`, scoped (not full workspace) to avoid unrelated crate churn.
- Exact command / steps: see Test Output above — `live_zone_freshness.rs` exercises the dispatcher directly; the two new `headroom-proxy` tests exercise it through the public `compress_anthropic_request` entry point, i.e. the same call path real `/v1/messages` traffic takes.
- Observed result: a lone fresh tail message with a large opaque blob is left byte-identical; an aged message earlier in the same request still compresses normally with a CCR marker.
- Not tested: this crate (`crates/headroom-proxy`) is not currently what the deployed/live proxy runs — the live proxy is the pure-Python entrypoint (`headroom.cli.proxy`), which goes through `headroom/transforms/content_router.py` instead. The equivalent Python-side freshness fix is tracked separately in #2088.

## Review Readiness

## Checklist
- [x] My code follows this project's style guidelines
- [x] I performed a self-review of my own code
- [x] I commented my code, particularly in hard-to-understand areas
- [ ] I made corresponding changes to the documentation
- [x] I added tests that prove my fix is effective or that my feature works
- [ ] I updated CHANGELOG.md if applicable

## Additional Notes

No CHANGELOG entry: this crate isn't currently wired into the deployed proxy (see "Not tested" above), so there's no user-facing behavior change yet. Companion PR #2088 covers the code path that's actually live.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1969,7 +1969,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
- "thiserror 1.0.69",
+ "thiserror 2.0.18",
  "tokio",
  "tower",
  "tracing",
--- a/crates/headroom-core/src/transforms/live_zone.rs
+++ b/crates/headroom-core/src/transforms/live_zone.rs
@@ -643,6 +643,46 @@ pub fn compress_anthropic_live_zone(
 pub fn compress_anthropic_live_zone_with_ccr(
     body_raw: &[u8],
     frozen_message_count: usize,
+    auth_mode: AuthMode,
+    model: &str,
+    ccr_store: Option<&dyn CcrStore>,
+) -> Result<LiveZoneOutcome, LiveZoneError> {
+    compress_anthropic_live_zone_with_ccr_and_freshness(
+        body_raw,
+        frozen_message_count,
+        0,
+        auth_mode,
+        model,
+        ccr_store,
+    )
+}
+
+/// Same as [`compress_anthropic_live_zone_with_ccr`], but also exempts
+/// the most-recent `fresh_message_count` messages (counted from the
+/// tail of `messages`) from compression entirely — regardless of size.
+///
+/// # Why
+///
+/// The live zone's `target_idx` (the latest user message) is, by
+/// construction, always the message that was *just appended* for this
+/// exact request — the in-flight turn's content the model is about to
+/// read for the very first time. Compressing it on sight forces an
+/// immediate `headroom_retrieve` round-trip just to read data the
+/// model hasn't had a chance to engage with yet (issue #3). Since
+/// every request resends the full conversation history, a message
+/// that's exempt here (fresh) becomes eligible on the *next* request
+/// once something else has been appended after it — at that point it's
+/// no longer in-flight, just accumulated context weight, and normal
+/// compression applies.
+///
+/// `fresh_message_count = 0` (what [`compress_anthropic_live_zone_with_ccr`]
+/// passes) preserves the pre-fix behavior byte-for-byte — this is an
+/// additive, opt-in parameter so existing callers are unaffected.
+/// Production wiring (the proxy) should pass `1`.
+pub fn compress_anthropic_live_zone_with_ccr_and_freshness(
+    body_raw: &[u8],
+    frozen_message_count: usize,
+    fresh_message_count: usize,
     _auth_mode: AuthMode,
     model: &str,
     ccr_store: Option<&dyn CcrStore>,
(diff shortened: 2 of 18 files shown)


JerrettDavis (changes requested):
Thanks for the detailed write-up and tests. The CCR size-floor pieces are directionally solid, and the new `live_zone_freshness` helper tests pass locally, but the freshness half of the PR is not wired into the production proxy yet.

`compress_anthropic_live_zone_with_ccr_and_freshness()` defaults back to `fresh_message_count = 0` through the existing public wrappers:

- `compress_anthropic_live_zone()` calls `compress_anthropic_live_zone_with_ccr(...)`
- `compress_anthropic_live_zone_with_ccr(...)` calls the new helper with `fresh_message_count = 0`

The live proxy call site still uses the old wrapper at `crates/headroom-proxy/src/compression/live_zone_anthropic.rs:358`:

```rust
compress_anthropic_live_zone(&dispatch_body, frozen_count, auth_mode.into(), model)
```

So the behavior described in the PR body — exempting the newest message so same-turn tool results are not replaced before the model reads them — is only exercised by the new tests, not by the real proxy path. The doc comment in `live_zone.rs` even says production should pass `1`, but no production caller does.

Please either wire the proxy to the freshness-aware helper with `fresh_message_count = 1` and add/adjust a proxy-facing regression test, or narrow the PR to the size-floor change and remove the claimed production freshness fix until it is actually connected.

Verified locally:

```text
cargo test -p headroom-core --test live_zone_freshness
# 3 passed
```

Also note: the current GitHub run shows `wheels (x86_64-apple-darwin)` failed, but the overall run is still in progress so GitHub has not exposed the failed job log yet. That may be separate, but it should be checked before this returns to green.

JerrettDavis (approved):
Thanks for the follow-up. This addresses the blocker from my previous review: the proxy path now calls compress_anthropic_live_zone_with_ccr_and_freshness with fresh_message_count = 1, and the new proxy-facing regressions cover both sides of the behavior: a fresh tail tool result stays verbatim while an aged earlier message still compresses.

Local verification on the current head babd66a1:
- cargo test -p headroom-core --test live_zone_freshness
- cargo test -p headroom-proxy fresh_tail_message_is_not_compressed_through_proxy_entry_point
- cargo test -p headroom-proxy aged_message_still_compresses_while_fresh_tail_is_untouched_through_proxy
- cargo check -p headroom-proxy
- cargo fmt --all -- --check

GitHub checks are still queued/in progress, but the production wiring issue is resolved from the code side.

chopratejas:
Declining this direction after review. In our prefix-cache regime the frozen prefix is byte-stable (mutating it busts the cache — a net loss), so the newest delta (the live-zone tail) is the *only* content we can compress without a cache miss. A freshness exemption therefore removes ~the only compressible content and makes compression a no-op. The legitimate concern behind it — don't degrade content the model just fetched — is already covered: the `protect_recent_reads` window handles the lossy path, and CCR replacement is losslessly recoverable via the retrieve marker, so there's no quality reason to skip it.

On the CCR min-size floor: our goal is maximum token savings; a 2 KB floor makes CCR a no-op across a large band of compressible content. The only floor that's ever justified is ~marker-size (to avoid a replacement marker larger than the original — negative savings), not 2 KB.

Really appreciate the work here. Happy to take a *separate*, tiny "don't replace if marker > original" guard if you want to open one. Closing this direction.

PR #2087, opened 2026-07-13
fix(compress): freshness exemption for tool_result content in Python content router

Eyalm321:
## Summary

- Companion fix to #2083, but for the code path that's actually live: `headroom/transforms/content_router.py`'s `ContentRouter`, not the `crates/headroom-proxy` Rust crate.
- Adds `protect_recent_tool_results` config field (default `1`) and a freshness gate in `_process_content_blocks`'s general `tool_result` branch: the most-recent N messages are skipped by both lossless and lossy compression tiers, mirroring how the Rust live-zone dispatcher's `fresh_message_count` excludes the tail message.
- Existing `protect_recent_code` only gates `SOURCE_CODE`-classified content on the flattened/OpenAI-style path; it doesn't cover general JSON/log/text `tool_result` blocks (Anthropic-style), which is where SmartCrusher/CCR substitution actually happens for the typical MCP tool-output case. `read_protection_window` is similarly narrow (only an explicit tool allowlist). This PR closes that gap for the common case.
- Guards against the `messages_from_end` sentinel (`0`, meaning "position not tracked" — used by direct/test callers) so the protection only fires for real message positions from `apply()`'s loop.

## Test plan

- [x] Added 4 new tests in `tests/test_transforms_content_router.py` covering: fresh tail protected, aged message still compresses, unset `messages_from_end` doesn't trigger protection, `protect_recent_tool_results=0` disables it.
- [x] Fixed a latent bug in a shared test helper (`SimpleNamespace` used as unhashable `strategy_used`) surfaced by the new tests; added `_make_router_with_mock_smart_crusher_compress` helper reflecting the real `SMART_CRUSHER` strategy path.
- [x] `ruff check .`, `ruff format --check .`, `mypy headroom --ignore-missing-imports`, `pytest` all pass.


Eyalm321:
Closing — branch was accidentally based on my fork's main (which has diverged from upstream by ~30 unrelated commits) instead of upstream/main, so this PR's diff pulled in far more than intended. Reopening clean in #2088 with just the single freshness-exemption commit cherry-picked onto upstream/main.

JerrettDavis (changes requested):
Thanks for the large follow-up here. The Python `ContentRouter` freshness guard and the CCR floor changes are directionally sound, and I was able to validate several focused pieces locally:

- `ruff check headroom/cli/mcp.py headroom/cli/wrap.py headroom/proxy/handlers/anthropic.py headroom/transforms/content_router.py tests/test_cli/test_mcp_status.py tests/test_transforms_content_router.py`
- `ruff format --check ...` on the same Python files
- `cargo fmt --check`
- `python -m pytest tests/test_cli/test_mcp_status.py -q` -> 5 passed
- `cargo test -p headroom-core --test live_zone_freshness` -> 3 passed
- `cargo test -p headroom-core smart_crusher:: --lib` -> 329 passed
- `cargo check -p headroom-py`

I could not run `tests/test_transforms_content_router.py` locally because this worktree does not have the compiled `headroom._core` extension available (`ModuleNotFoundError: No module named 'headroom._core'`).

Two blockers before this is merge-ready:

1. The Rust freshness API is not wired to any production-facing caller. `compress_anthropic_live_zone_with_ccr_and_freshness(..., fresh_message_count=1, ...)` is tested directly, but the public `compress_anthropic_live_zone_with_ccr()` wrapper still passes `0`, and `rg` does not show any non-test caller invoking the new freshness-aware function. The comment says production should pass `1`, but nothing currently does, so the Rust live-zone path still has the same behavior outside the new tests. Please either wire the production/exported path to pass `1` where appropriate, or remove the unused Rust API/tests from this PR and keep the fix scoped to the Python router path.

2. The new MCP status helper is dead code and appears unrelated to this compression PR. `find_headroom_registration()` is only referenced by the new tests; `headroom mcp status` still uses `get_all_registrars()` and `registrar.get_server("headroom")`. The Claude registrar already reads the modern `.claude.json` path, so these tests do not prove the command behavior changed. Please drop this unrelated helper/test addition or convert the tests to exercise the actual `mcp status` command through the registrar path.

Once those are cleaned up, the remaining compression changes should be much easier to review on their own merits.

PR #2088, opened 2026-07-13
fix(compress): freshness exemption for tool_result content in Python content router

Eyalm321:
## Description

Closes #3

Companion fix to #2083, but for the code path that's actually live in the deployed proxy: `headroom/transforms/content_router.py`'s `ContentRouter`, not the `crates/headroom-proxy` Rust crate. #2083 patches a crate the running proxy doesn't currently call; this PR patches the pure-Python entrypoint (`headroom.cli.proxy`) that real traffic goes through.

Existing `protect_recent_code` only gates `SOURCE_CODE`-classified content on the flattened/OpenAI-style message path; it doesn't cover general JSON/log/text `tool_result` content blocks (Anthropic-style), which is where SmartCrusher/CCR opaque-blob substitution actually happens for the typical MCP tool-output case. `read_protection_window` is similarly narrow (only covers an explicit tool allowlist). Neither protects a freshly-produced tool result from being swapped for a `<<ccr:...>>` marker before the model has had a chance to read it once.

## Type of Change
- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (a fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made
- Added `protect_recent_tool_results` config field to `ContentRouterConfig` (default `1`).
- Added a freshness gate in `_process_content_blocks`'s general `tool_result` branch: the most-recent N messages (by real position from `apply()`'s loop) are skipped by both the lossless and lossy compression tiers, mirroring how the Rust live-zone dispatcher's `fresh_message_count` excludes the tail message.
- Guarded against the `messages_from_end` sentinel (`0`, meaning "position not tracked" — used by direct/test callers that don't wire in real position tracking) so the protection only fires for real message positions from `apply()`'s loop, not silently on every unpositioned call.

## Testing

### Test Output

```text
tests/test_transforms_content_router.py ................................ [ 78%]
.........                                                                [100%]
41 passed in 0.49s

ruff check . / ruff format --check . / mypy headroom --ignore-missing-imports: all clean
```

## Real Behavior Proof
- Environment: local venv (`headroom-py314`, Python 3.14) with the worktree source loaded via `PYTHONPATH`, compiled `headroom._core` extension present.
- Exact command / steps: drove the real (unmocked) `ContentRouter.apply()` entry point — the same call the live proxy makes — with a 4-message Anthropic-style conversation: an aged `tool_result` (message 2 of 4, large JSON blob) and a fresh tail `tool_result` (message 4 of 4, same-sized blob).
- Observed result: `transforms_summary` showed `router:protected:recent_tool_result` firing once and `router:tool_result:smart_crusher` firing once. The fresh tail's content was byte-identical (25,586 → 25,586 bytes); the aged message compressed normally (24,385 → 16,412 bytes, ~33% reduction).
- Not tested: full live HTTP round-trip through a running `headroom proxy` process against a real Anthropic backend (would require spinning up a second isolated proxy instance and real API credentials; the direct-`apply()` test above exercises the identical code path without that overhead).

## Review Readiness

## Checklist
- [x] My code follows this project's style guidelines
- [x] I performed a self-review of my own code
- [x] I commented my code, particularly in hard-to-understand areas
- [ ] I made corresponding changes to the documentation
- [x] I added tests that prove my fix is effective or that my feature works
- [ ] I updated CHANGELOG.md if applicable

## Additional Notes

Supersedes #2087, which was accidentally opened from a branch based on my fork's `main` (30 commits ahead of `upstream/main` from unrelated prior fork PRs) instead of `upstream/main` directly, so its diff pulled in far more than intended. This PR is a single clean commit cherry-picked onto current `upstream/main`.

### DIFF
--- a/benchmarks/proxy_mode_benchmark.py
+++ b/benchmarks/proxy_mode_benchmark.py
@@ -200,7 +200,11 @@ def _simulate_mode(turns: int, mode: str) -> ModeBenchmarkResult:
         forwarded = pipeline_result.messages
 
         if mode == PROXY_MODE_TOKEN:
-            comp_cache.update_from_result(messages, forwarded)
+            comp_cache.update_from_result(
+                messages,
+                forwarded,
+                protected_contents=pipeline_result.protected_tool_result_contents,
+            )
         if mode == PROXY_MODE_CACHE:
             forwarded, _ = AnthropicHandlerMixin._restore_frozen_prefix(
                 messages, forwarded, frozen_message_count=frozen
--- a/headroom/cache/compression_cache.py
+++ b/headroom/cache/compression_cache.py
@@ -13,6 +13,7 @@ import logging
 import threading
 import time
 from collections import OrderedDict
+from collections.abc import Iterable
 from dataclasses import dataclass
 
 logger = logging.getLogger(__name__)
@@ -283,11 +284,30 @@ class CompressionCache:
                 result.append(msg)
             return result
 
-    def update_from_result(self, originals: list[dict], compressed: list[dict]) -> None:
+    def update_from_result(
+        self,
+        originals: list[dict],
+        compressed: list[dict],
+        protected_contents: Iterable[str] | None = None,
+    ) -> None:
         """Cache new compressions by comparing original and compressed messages.
 
         Index-aligned: for each position, if both are tool results and the
         content differs, store the mapping original_hash -> compressed_content.
+
+        Args:
+            protected_contents: Raw tool_result content left verbatim by
+                ContentRouter's freshness exemption this call (issue #3),
+                i.e. `TransformResult.protected_tool_result_contents`. Content
+                unchanged for this reason must NOT be marked stable below: it
+                was skipped only because it was the fresh tail this turn, not
+                because it's incompressible, so it needs to stay eligible for
+                compression once a later turn ages it out of the protection
+                window. Without this exclusion, `compute_frozen_count` would
+                treat it as part of the stable/frozen prefix as soon as it's
+                no longer the tail, and content_router would never see it
+                again to actually compress it -- permanently forfeiting the
+                compression, not just deferring it by one turn.
         """
         if len(originals) != len(compressed):
             logger.warning(
@@ -296,6 +316,10 @@ class CompressionCache:
                 len(compressed),
             )
             return
+
+        protected_hashes = (
+            {self.content_hash(c) for c in protected_contents} if protected_contents else set()
+        )
 
         # Single critical section: `store_compressed` re-acquires the lock
         # (RLock) safely.
@@ -306,8 +330,11 @@ class CompressionCache:
                 if orig_content is None or comp_content is None:
                     continue
                 if orig_content == comp_content:
+                    h = self.content_hash(orig_content)
+                    if h in protected_hashes:
+                        continue
                     # Content unchanged — mark as stable for frozen count walk
-                    self._stable_hashes.add(self.content_hash(orig_content))
+                    self._stable_hashes.add(h)
                     continue
                 h = self.content_hash(orig_content)
                 tokens_saved = len(orig_content) // 4 - len(comp_content) // 4
--- a/headroom/config.py
+++ b/headroom/config.py
@@ -722,6 +722,13 @@ class TransformResult:
     cache_metrics: CachePrefixMetrics | None = None  # Populated by CacheAligner
     timing: dict[str, float] = field(default_factory=dict)  # transform_name → ms
     waste_signals: WasteSignals | None = None  # Detected waste in original messages
+    # Raw content of tool_result blocks left verbatim by ContentRouter's
+    # freshness exemption (issue #3), this call only. CompressionCache uses
+    # this to avoid marking freshness-skipped content permanently "stable"
+    # (see CompressionCache.update_from_result) -- it was skipped because it
+    # was the fresh tail this turn, not because it's incompressible, so it
+    # must stay eligible for compression once it's no longer the tail.
+    protected_tool_result_contents: list[str] = field(default_factory=list)
 
     @property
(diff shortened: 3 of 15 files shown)


JerrettDavis (approved):
Reviewed the narrowed replacement for #2087. This version is scoped to the active Python `ContentRouter` path and removes the unrelated Rust/MCP/wrap changes that blocked the previous PR.

The freshness guard is applied from the real `apply()` message loop via `messages_from_end`, the `0` sentinel keeps direct callers/back-compat from silently protecting everything, and the tests cover the important behavior split: fresh tail protected, aged tool result still compressible, explicit disable works, and unset position does not trigger the guard.

Local validation:
- `ruff check headroom/transforms/content_router.py tests/test_transforms_content_router.py`
- `ruff format --check headroom/transforms/content_router.py tests/test_transforms_content_router.py`

I could not run `python -m pytest tests/test_transforms_content_router.py -q` in this local worktree because the compiled extension is absent here (`ModuleNotFoundError: No module named 'headroom._core'`). No code blocker from my review.

Eyalm321:
Heads up @JerrettDavis — I've pushed a second commit (`d40492bd`) that fixes a regression the first commit's approach would otherwise cause in the **live** token-mode proxy path, so this is worth a re-look before merge.

**The problem the first commit alone creates:** the freshness exemption skips compressing the fresh tail tool_result (correct — the model reads it verbatim once). But in token mode, `CompressionCache.update_from_result` then saw `orig == compressed` for that skipped block and marked its content hash permanently *stable* (`_stable_hashes`). On the next turn — once that message was no longer the tail — `compute_frozen_count` treated it as part of the frozen prefix, and `content_router` never saw it again. Net effect: a tool_result protected once for freshness was never compressed on **any** later turn either — permanently forfeiting the compression rather than deferring it one turn. It surfaced as a 0-savings regression in the token-mode benchmark (`tests/test_proxy_mode_benchmark.py`).

**The fix (commit 2):** thread a "was protected for freshness, don't mark stable" signal from `ContentRouter` → `TransformResult.protected_tool_result_contents` → `TransformPipeline` → `CompressionCache.update_from_result(protected_contents=...)`, which excludes those hashes from `_stable_hashes` so they stay in the live zone and compress once they age out of the protection window. Proxy handlers pass it through `getattr`-guarded (mock/legacy pipeline results fail open rather than crash).

**CACHE mode note:** it now legitimately shows 0 tool-output compression — its live zone is only the final message, which the freshness exemption protects. Its value is prefix stability (verbatim once, then frozen for provider cache hits), so I updated the benchmark test to assert that tradeoff instead of the old `cache.total_tokens_saved > 0` — which was itself only satisfied by compressing the fresh tail, the exact anti-pattern issue #3 removes.

Verification (local, Python 3.14):
- `tests/test_proxy_mode_benchmark.py`: token mode now saves 64,470 tokens (48% reduction) vs 0 before; cache mode preserves the prefix.
- Full `tests/test_proxy/` (147), plus the cache-stability / deferred-injection / lifecycle / hooks suites — all green.
- `ruff check` / `ruff format --check` / `mypy headroom --ignore-missing-imports` clean.

JerrettDavis (approved):
Re-reviewed the refreshed head, including the new cache-stability follow-up commit after the earlier approval. The broader delta is justified: the freshness guard now also prevents the token-mode compression cache from marking a freshness-skipped tool_result permanently stable, so the content remains eligible for compression once it ages out of the fresh-tail window. The runtime `protect_recent_tool_results` threading and `protected_tool_result_contents` propagation look backward-compatible, and the proxy call sites use guarded access for older/mock pipeline results.

Focused validation on Windows:
- `uv run --extra dev python -m pytest tests/test_compression_cache.py tests/test_transforms_content_router.py tests/test_transforms/test_content_router.py tests/test_proxy_mode_benchmark.py -q` -> 135 passed
- `uv run --extra dev python -m pytest tests/test_proxy_anthropic_cache_stability.py tests/test_proxy_openai_cache_stability.py tests/test_proxy/test_anthropic_ccr_deferred_injection.py -q` -> 45 passed
- `uvx --from ruff==0.15.17 ruff check ...` on the touched source/test files
- `uvx --from ruff==0.15.17 ruff format --check ...` on the touched source/test files

JerrettDavis (approved):
Re-reviewed the latest head after the small follow-up commit. The new commit only updates the `TransformResult` field-set test to include `protected_tool_result_contents`; the cache/router behavior reviewed on the previous head is unchanged.

Focused validation on Windows:
- `uv run --extra dev python -m pytest tests/test_config.py::TestTransformResult tests/test_compression_cache.py tests/test_transforms_content_router.py tests/test_transforms/test_content_router.py tests/test_proxy_mode_benchmark.py -q` -> 138 passed
- `uvx --from ruff==0.15.17 ruff check ...` on the touched source/test files
- `uvx --from ruff==0.15.17 ruff format --check ...` on the touched source/test files

chopratejas:
Declining this direction after review. In our prefix-cache regime the frozen prefix is byte-stable (mutating it busts the cache — a net loss), so the newest delta (the live-zone tail) is the *only* content we can compress without a cache miss. A freshness exemption therefore removes ~the only compressible content and makes compression a no-op. The legitimate concern behind it — don't degrade content the model just fetched — is already covered: the `protect_recent_reads` window handles the lossy path, and CCR replacement is losslessly recoverable via the retrieve marker, so there's no quality reason to skip it.

On the CCR min-size floor: our goal is maximum token savings; a 2 KB floor makes CCR a no-op across a large band of compressible content. The only floor that's ever justified is ~marker-size (to avoid a replacement marker larger than the original — negative savings), not 2 KB.

Really appreciate the work here. Happy to take a *separate*, tiny "don't replace if marker > original" guard if you want to open one. Closing this direction.

PR #2092, opened 2026-07-13
fix(cache): extract tool_result content from list-of-blocks format

### LINKED ISSUE #2053: Claude Code gets zero tool_result compression: Anthropic list-of-blocks content is never extracted (fix sketch included) + a second live-path defect behind it
## Summary

On headroom **v0.31.0**, `wrap claude` users on current Claude Code (**2.1.202**) receive **zero** tool_result compression while the startup banner prints "Code-Aware: ENABLED". We root-caused one defect at source level (fix sketch below), and found evidence of a second, separate defect in the live Anthropic request path that persists after the first is patched. Filing both together since they stack on the same user journey.

## Defect 1 — `_extract_tool_result_content` returns None on list-of-blocks content

`headroom/cache/compression_cache.py::_extract_tool_result_content` handles a `tool_result` block's `content` only when it is a **plain string**. Current Claude Code sends the modern Anthropic Messages form — a **list of typed blocks** (`[{"type": "text", "text": "..."}]`). The function returns `None`, so every tool_result is skipped before any compressor is consulted.

The compressor itself works: driven directly on a real Claude Code `Read` payload (9,609 chars, line-number-prefixed), it rewrites **2,745 → 80 tokens (~97%)**. The zero happens one layer up, in extraction.

Proof pair (in-process, identical payload):
- content as **list-of-blocks** (as Claude Code sends it) → no spans, nothing compressed
- same content converted to a **plain string** → spans fire (`ast-grep 2745→80 tok`)

Shape of the fix we validated in-process (venv-local patch; adapt to house style):

```python
# in _extract_tool_result_content, alongside the isinstance(content, str) branch:
if isinstance(content, list):
    texts = [b.get("text", "") for b in content
             if isinstance(b, dict) and b.get("type") == "text"]
    return "\n".join(t for t in texts if t) or None
```

Impact: every `wrap claude` user on current Claude Code silently gets 0% on the headline feature. Published benchmark numbers presumably come from OpenAI-compat / JSON-fixture workloads where the string branch is exercised.

## Defect 2 — live Anthropic path still inert with Defect 1 patched

With the extractor patched locally, the pipeline entry point (`apply_to_messages`) fires **in-process** on Claude-Code-form payloads (2,745 → 80 tok). But the **live proxy** still compressed **0/5** on a 31KB/1,481-line single-file Read probe (cache mode). The Anthropic wiring appears present in v0.31.0 (`server.py:811` anthropic_pipeline; `anthropic.py:1263ff` interceptor transform; env flag set), yet live spans never materialize — something in the anthropic handler/streaming path prevents engagement. We time-boxed our diagnosis here; pinning it down likely needs your debug surface (`headroom capture`).

## Repro

1. `headroom wrap claude` (or standalone: `headroom --mode cache --intercept-tool-results` + `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`), Claude Code 2.1.202
2. Ask the model to Read a ≥30KB code file
3. Proxy stats: `requests_compressed = 0` (we observed **0/30** across a 4-arm A/B including the 31KB probe)
4. In-process discriminator for Defect 1: feed a captured Claude Code tool_result (list form) to the extraction path → `None`; stringify the same content → spans fire

Also probed, all 0-compressed on this workload: `--mode token --lossless`, and the MCP server wired via `--mcp-config`. Fidelity was clean in every configuration — the proxy is transparent, just inert on this traffic.

## Environment

headroom 0.31.0 (pip venv) · Claude Code 2.1.202 · WSL2 Ubuntu 24.04 · Python 3.12 · Anthropic Messages API traffic (sonnet)

Full evidence pack available (per-request proxy stats, session transcripts, probe scripts) — happy to attach.

lennney:
## Description

Closes #2053

Modern Claude Code sends `tool_result` content as a list of typed blocks (`[{"type": "text", "text": "..."}]`) instead of a plain string. `_extract_tool_result_content` and `_swap_tool_result_content` in `compression_cache.py` only handled the plain string case, so every tool_result was skipped before compression — zero savings for all `headroom wrap claude` users.

Fix: extract text from list-of-blocks content in `_extract_tool_result_content`, and collapse the list to a single text block in `_swap_tool_result_content` when replacing with compressed content.

## Type of Change

- [x] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

- `headroom/cache/compression_cache.py`:
  - Added `_extract_text_from_blocks()` helper to extract joined text from Anthropic list-of-blocks format
  - Updated `_extract_tool_result_content()` to handle list content in both Anthropic tool_result blocks and OpenAI role=tool messages
  - Updated `_swap_tool_result_content()` to collapse list-of-blocks to a single text block when replacing content (preserves structure, prevents multi-block join mismatch)
- `tests/test_token_headroom_mode.py`: Added 12 tests covering Anthropic list-of-blocks, OpenAI list, mixed blocks, empty list, non-mutation, missing-text-block fallback, and non-tool messages

## Testing

- [x] Existing tests pass
- [x] New tests cover the fix
- [x] PBT round-trip property verified (6 properties × 850+ random examples)
- [x] Adversarial edge case tests pass (50 cases across 3 functions)

```
tests/test_token_headroom_mode.py ....................... 32 passed (0.35s)
tests/test_transforms_content_router.py ................. 37 passed (1.08s)
tests/test_backend_bugs.py ............................... 37 passed (4.86s)
```

## Real Behavior Proof

- Environment: headroom main (upstream/main at time of PR), uv-managed Python 3.12
- Exact command / steps:
  1. `uv run python -m pytest tests/test_token_headroom_mode.py -x -q --no-header`
  2. `uv run python /tmp/pbt_tool_result_content.py` (PBT round-trip, 200 examples each property)
  3. `uv run python /tmp/adversarial_lens_security.py` (50 edge case checks)
  4. Design scan: grep'd codebase for `isinstance(content, str)` near tool_result — 3 sibling functions (`_to_text`, `_block_text`) already handle list-of-blocks
- Observed result: All 106 tests pass. PBT confirmed extract+swap round-trip invariant holds for 850+ random inputs. 50 adversarial edge cases (null/None/nested/unicode/100K chars/1000 blocks) produce no crashes or wrong output.
- Not tested: live proxy with real Claude Code traffic (the ContentRouter already handles list content correctly since v0.32.0, confirmed by code audit)

## Review Readiness


JerrettDavis (approved):
Reviewed and did a small maintainer cleanup on current head d0be0eed. The cache change now handles Anthropic/OpenAI list-of-blocks tool results, and the swap path collapses compressed Anthropic list content to a single text block to avoid stale trailing block text being re-joined.

Maintainer cleanup pushed:
- applied `ruff format` to `headroom/cache/compression_cache.py`
- amended the invalid `trigger governance re-check` commit into `fix(cache): format list-block tool-result extraction`, so commitlint should pass on rerun

- `uvx ruff@0.15.17 check headroom/cache/compression_cache.py tests/test_token_headroom_mode.py` -> pass
- `uvx ruff@0.15.17 format --check headroom/cache/compression_cache.py tests/test_token_headroom_mode.py` -> pass
- `uv run --extra dev pytest tests/test_token_headroom_mode.py -q` -> 32 passed

CI is queued on the new head. The dependency audit may still reflect the shared security-alert state until the dependency remediation PR lands.

PR #2100, opened 2026-07-13
fix(proxy/anthropic): preserve non-2xx upstream status through security scan

abhay-codes07:
## Description

When enterprise security scanning is enabled, the non-streaming `/v1/messages` handler can turn a failed upstream response into an HTTP 200, so the client never sees the error.

After the upstream call, the handler parses the body unconditionally:

```python
resp_json = None
try:
    resp_json = response.json()
except (json.JSONDecodeError, ValueError):
    ...
```

Every response-mutating block that follows gates on a successful upstream — CCR handling (`... and response.status_code == 200 ...`), the response cache (`if self.cache and response.status_code == 200`), and the buffered-stream CCR block (`if buffered_stream_ccr and response.status_code == 200 ...`). The enterprise-security block was the exception:

```python
if self.security and _security_ctx and resp_json:
    resp_json = self.security.scan_response(resp_json, _security_ctx)
    response = httpx.Response(status_code=200, content=json.dumps(resp_json).encode(), headers=response_headers)
    if not buffered_stream_ccr:
        return Response(content=response.content, status_code=response.status_code, headers=response_headers)
```

No status check, and the rebuilt response hardcodes `status_code=200`. `_retry_request` returns 429 (rate limit), 529 (overloaded), and other 4xx responses verbatim to this caller, so when security is configured any of those — whose JSON error body parses fine — is rebuilt as an HTTP 200 and returned. The client sees success, so its retry/backoff logic never fires on a rate limit or overload, exactly when it matters most.

## Fix

Gate the security block on a 200 upstream, the same condition the sibling CCR/cache/buffered-stream blocks already use. A non-2xx response falls through to the final `return Response(..., status_code=response.status_code, ...)` and keeps its real status; a 200 is still scanned and returned as before.

```python
if self.security and _security_ctx and resp_json and response.status_code == 200:
    ...
```

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: gate the enterprise-security response-scan branch on `response.status_code == 200`.
- `tests/test_anthropic_pre_upstream_backpressure.py`: reuse the existing `_DummyAnthropicHandler` harness (adds optional `security` and `upstream_status` params, both defaulting to today's behavior) and add `test_security_scan_preserves_non_200_upstream_status` (429/529/400) plus a 200 positive-control test.

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/proxy/handlers/anthropic.py tests/test_anthropic_pre_upstream_backpressure.py
All checks passed!
$ python -m py_compile headroom/proxy/handlers/anthropic.py tests/test_anthropic_pre_upstream_backpressure.py
OK
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17`. Importing `headroom` pulls in the torch/transformers stack and running the handler test locally OOM-kills this box, so I verified the branch logic with a dependency-free script that models the security block plus the fallthrough return, and left the full pytest (including the new handler test) to CI.
- Exact command / steps: modelled the non-streaming return path with enterprise security configured, running a `429` upstream through the old branch (no status gate, rebuilds 200) and the new branch (gated on 200, falls through), plus a `200` upstream as a control. Also traced `_retry_request` in `server.py` to confirm it returns 429/529/4xx verbatim to this caller (only 5xx raises), so the branch is reachable for those statuses.
- Observed result: old branch returns HTTP 200 for a 429 upstream (laundered); new branch returns 429; a 200 upstream returns 200 under both. The new parametrized handler test asserts the returned status equals the upstream status for 429/529/400, and the control test asserts a 200 upstream still returns 200.
- Not tested: a live enterprise-security plugin (`scan_response` is an out-of-repo component; the test uses a passthrough stub); full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" and "type checking" boxes are unchecked because the full suite imports the ML stack, which I can't run in this environment; the fix is a one-condition gate mirroring three sibling blocks in the same method, and the regression test reuses the file's existing, proven `handle_anthropic_messages` harness (the added handler params default to current behavior, so existing tests are unaffected). The branch-logic proof and the new tests cover the fix for CI. The security block only runs when an enterprise-security component is configured; without it, this path is inert.


abhay-codes07:
@JerrettDavis @chopratejas flagging one thing on this PR so it's not a surprise in review.

The fix itself is a one line gate. The security response scan branch was the only one of the response mutating blocks in that method that didn't check the upstream status before rebuilding the reply as a 200, so a 429 or 529 was getting handed back to the client as a success. I gated it on `response.status_code == 200` to match the CCR, cache, and buffered stream blocks right next to it.

The honest caveat is the test. This box OOMs as soon as it imports the handler (the ML stack loads torch/transformers), so I couldn't run the handler level test locally. To keep the risk down I reused the existing `_DummyAnthropicHandler` harness in `test_anthropic_pre_upstream_backpressure.py` instead of writing new scaffolding, and the two params I added (`security`, `upstream_status`) both default to the current behavior so the existing tests in that file are untouched. I also traced `_retry_request` in `server.py` to confirm it returns 429/529/4xx verbatim to this caller (only 5xx raises), and verified the branch logic with a small standalone script (old path returns 200 for a 429 upstream, new path returns 429, a real 200 still returns 200).

So the source change is low risk, but the new test is the part I'd appreciate a look at since CI is the first place it actually runs end to end. If the harness needs a tweak I'm happy to follow up quickly.

Issue #260, opened 2026-04-24
[BUG] headroom init codex creates invalid config.toml by appending keys inside the [features] block


SvenMeyer:
**Description:**
When running `headroom init codex`, the initialization script appends the Headroom configuration to the end of the existing `~/.codex/config.toml`. If the existing file ends with a table definition like `[features]`, the Headroom keys are incorrectly scoped to that table. 

This causes Codex CLI to fail on startup with the following error:
`Error loading config.toml: invalid type: string "headroom", expected a boolean in features`

**Steps to Reproduce:**
1. Create a `.codex/config.toml` that ends with a `[features]` block:
   ```toml
   [features]
   codex_hooks = true
   ```
2. Run `headroom init codex`.
3. Observe that `model_provider = "headroom"` is appended immediately after the `codex_hooks` line.

**Actual Resulting Config:**
```toml
[features]
codex_hooks = true

# --- Headroom init provider ---
model_provider = "headroom"

[model_providers.headroom]
```

**Expected Behavior:**
The script should ensure that global keys like `model_provider` are placed at the root of the document (before any `[...]` sections) or that it explicitly "breaks" out of the previous section by ensuring it doesn't append strings to a boolean-only table.

```toml
model_provider = "headroom"

[features]
codex_hooks = true

# --- Headroom init provider ---
[model_providers.headroom]
name = "Headroom init proxy"
base_url = "http://127.0.0.1:8787/v1"
env_key = "OPENAI_API_KEY"
requires_openai_auth = true
supports_websockets = true
# --- end Headroom init provider ---
```

**Environment:**
- **OS:** Manjaro Linux
- **Headroom Version:** 0.10.1
- **Python Version:** 3.14.3
- **Codex CLI Version:** Latest (v0.89.0+)

SphereSK:
i am not able to run headroom wrap codex as it always show me error: Missing environment variable: `OPENAI_API_KEY`.  do i missing something in setup? i am login via oauth and codex subscription

PR #2102, opened 2026-07-13
fix(codex): rewrite config.toml properly so Codex will route through …

nangsontay:
## Description

`headroom install apply --providers manual --target codex --scope provider` silently failed to route Codex through the proxy whenever `~/.codex/config.toml` already had a `[table]` section (e.g. `[features]`, `[mcp_servers.*]`). `apply_provider_scope` appended the managed `model_provider = "headroom"` block after the last existing table, so TOML scoped the bare key into that table instead of the document root — Codex silently ignored it and kept routing through its default provider. The same code path never overrode a pre-existing top-level `model_provider` assignment either, so a user's `model_provider = "openai"` kept winning even when Headroom's block was appended elsewhere in the file.

This mirrors a bug already fixed in the `headroom init` path (`_ensure_codex_provider`, #260) that was never ported to the persistent-install path.

Closes: reported via user session (no tracked issue number yet).

## Type of Change

## Changes Made

- **`headroom/providers/codex/install.py`**: Added `_insert_block_at_root()`, which walks the document line-by-line and inserts the managed marker block immediately above the first `[table]`/`[[array-of-tables]]` header, falling back to end-of-file append only when no table exists. Mirrors the root-insertion logic already used by `cli/init.py:_ensure_codex_provider`.
- Added `_ANY_MODEL_PROVIDER` / `_ANY_OPENAI_BASE_URL` patterns (match any value, not just `"headroom"`) so `apply_provider_scope` strips **any** prior top-level `model_provider` / `openai_base_url` assignment before re-inserting the managed block — the managed keys now override the user's config outright instead of losing to it.
- `apply_provider_scope` merge order is now: strip old managed block → strip prior top-level assignments → insert fresh block at document root.

## Testing

- [x] **New regression test**: `test_apply_codex_provider_scope_lands_model_provider_at_root` (`tests/test_install/test_providers.py`) — asserts `model_provider = "headroom"` lands before `[features]`, overrides a prior `"openai"` value, and the user's own table content survives.
- [x] **Existing tests**: `tests/test_install/test_providers.py` — 42/42 pass (includes prior codex apply/revert/replace/orphan-cleanup coverage).
- [x] **Adversarial (ad-hoc, not committed)**: 6-case TOML round-trip proof — parses output with `tomllib` (not substring matching) across: prior provider before a table, no prior provider, empty file, scalars-only (no tables), CRLF line endings, multiple tables. All 6 pass after the fix; first pass caught a false failure from a stale globally pip-installed `headroom` copy shadowing the repo source when tests run outside the project directory — re-verified from inside the repo to confirm the fix itself is correct.
- [x] **Lint**: `ruff check` and `ruff format --check` pass on both changed files.

```text
$ uv run pytest tests/test_install/test_providers.py -q
42 passed in 0.21s

$ uv run --with ruff ruff check headroom/providers/codex/install.py tests/test_install/test_providers.py
All checks passed!

$ uv run --with ruff ruff format --check headroom/providers/codex/install.py tests/test_install/test_providers.py
2 files already formatted
```

## Real Behavior Proof

- Environment: macOS 26.4.1 (arm64), Python 3.13.14, headroom branch `patch/install-codex`
- Exact command / steps: constructed a temp `config.toml` with `[features]\nweb_search = true` (no existing Headroom block), invoked `apply_provider_scope(manifest)` against it with `codex_config_path` patched to the temp file, then parsed the result with `tomllib.loads()`.
- Observed result: before the fix, `tomllib.loads(result)["model_provider"]` raised `KeyError` — the key was nested inside `[features]` due to end-of-file append. After the fix, `parsed["model_provider"] == "headroom"` and `parsed["features"]["web_search"] is True` — both the managed key and the user's table are present and correctly scoped. Revert removes `model_provider` and preserves the user's table.
- Tested local build and behavior is correct as expected of this patch.

## Review Readiness

### DIFF
--- a/headroom/providers/codex/install.py
+++ b/headroom/providers/codex/install.py
@@ -31,6 +31,10 @@ _ORPHAN_HEADROOM_TABLE = re.compile(
     r"(?=^\[|\Z)"
 )
 
+_TOML_TABLE_HEADER_RE = re.compile(r"^[ \t]*(?:\[\[[^\]\r\n]+\]\]|\[[^\]\r\n]+\])[ \t]*(?:#.*)?$")
+_ROOT_MODEL_PROVIDER_RE = re.compile(r"^[ \t]*model_provider[ \t]*=")
+_ROOT_OPENAI_BASE_URL_RE = re.compile(r"^[ \t]*openai_base_url[ \t]*=")
+
 
 def codex_uses_chatgpt_auth(auth_path: Path) -> bool:
     """Whether Codex authenticated via ChatGPT OAuth (vs an OpenAI API key).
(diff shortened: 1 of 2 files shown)


JerrettDavis:
Updated this branch to clear the mechanical CI blockers: rebased onto current main for the dependency-audit baseline, shortened the commit header, and tightened the Codex TOML cleanup so it only removes root-level provider assignments instead of table-scoped user settings.

Local validation:
- uv run --extra dev python -m pytest tests\test_install\test_providers.py -q
- uvx ruff@0.15.17 check headroom\providers\codex\install.py tests\test_install\test_providers.py
- uvx ruff@0.15.17 format --check headroom\providers\codex\install.py tests\test_install\test_providers.py

JerrettDavis (approved):
Approved after updating the branch and reviewing the scoped Codex TOML behavior. The change now keeps managed provider keys at the TOML root, preserves table-scoped user settings, and includes a regression test for that case. Local focused tests and ruff checks passed; waiting on the fresh GitHub checks.

PR #2114, opened 2026-07-13
fix(kompress): fail-open wall-clock guard on single-cache-miss compression

### LINKED ISSUE #2046: Proxy process fully freezes (including health-check endpoints) during compression, exceeding all configured timeouts — likely GIL starvation in native compression call
**Summary**

Running the proxy under systemd with a health-check watchdog, the process occasionally freezes completely — not just the request being compressed, but the entire event loop, including unrelated `/health`/`/readyz`/`/livez` endpoints. The freeze lasts well beyond every configured timeout and the process only recovers after being killed and restarted.

**Environment**
- `headroom-ai` 0.31.0, installed via pip (not editable)
- Python 3.14
- Kompress ONNX compression backend (ModernBERT-based), default config
- Linux, run as a long-lived background service with periodic HTTP health checks

**Observed behavior**

A request with a small number of messages (2) triggers the content-aware compression pipeline. The last log line before the freeze is the pipeline entering the single-cache-miss compression path:

```
Pipeline: freezing first 1/2 messages (prefix cached by provider)
```

After that line, there is no further log output at all — not from the request being compressed, and not from any other endpoint, including the periodic `/readyz` health check, which should be entirely independent (a plain `async def` handler with no dependency on the compression executor). The process has to be killed with SIGKILL; a plain `Restart=on-failure`/graceful shutdown does not recover it.

**Why this looks like GIL starvation, not just a slow request**

Reading through the installed package source, there appear to be at least three independent timeout/cancellation layers around compression:

1. `_run_compression_in_executor()` wraps the executor call in `asyncio.wait_for(..., timeout=COMPRESSION_TIMEOUT_SECONDS)` (default 30s).
2. The Kompress compressor's execution-slot acquisition (`BoundedSemaphore`-based) has its own bounded wait with a configurable timeout and a documented fail-open path (skip compression on timeout).
3. An internal deadline constant (`HEADROOM_COMPRESSION_DEADLINE_MS`, default 20000ms) referenced inside the compressor.

Despite all three, the observed freeze lasted roughly 2 minutes with zero recovery — including on the health endpoints, which don't go through any of these layers at all. This is consistent with something in the compression call path (the compiled core extension, or an ONNX Runtime `session.run()` invocation) holding the GIL without yielding, since `asyncio.wait_for`/executor timeouts can only cancel the *future*, not preempt an already-running native call that never returns control to Python — and if the GIL itself is held, unrelated coroutines (like the health-check handler) freeze too, which is exactly what was observed.

Also worth noting: when the single-cache-miss branch is taken (only one item pending), the compression call appears to run inline on the calling thread/executor without the same per-task timeout wrapping used in the multi-task `ThreadPoolExecutor` branch (which itself also submits without a `.result(timeout=...)` bound) — so a hang here has no cooperative timeout above it at all until the outer 30s executor-level wait_for, which per the above evidence, doesn't help if the GIL is starved.

**Suggested fix directions**

- Ensure any native/compiled extension call in the compression path explicitly releases the GIL during long-running work (e.g., a PyO3 `py.allow_threads`-style release around ONNX Runtime calls or other native compute), so the GIL cannot be held for the duration of a stuck call.
- Consider a subprocess- or process-pool-based execution model for compression (rather than in-process threads) so that a genuinely stuck call can be forcibly terminated, since thread-based timeouts fundamentally cannot preempt a non-yielding native call.
- Add a coarse-grained watchdog inside the process itself (e.g., a separate heartbeat thread that can detect the main GIL/event loop hasn't progressed and forcibly abort/exit) as a last-resort safety net, since `_loop_callback_payload()`/loop-heartbeat tracking already exists in the codebase but doesn't appear to trigger any self-recovery action.

**Reproduction difficulty**: intermittent — has not been isolated to a 100%-reliable minimal repro yet. Happy to provide more log detail/config if useful, with sensitive details redacted.

rodboev:
## Description

The single-cache-miss branch in `ContentRouter` ran compression inline on the request path without its own wall-clock guard, so a cooperative stall waited for the full call even when `HEADROOM_COMPRESSION_DEADLINE_MS` was meant to fail open. This change adds a branch-level watchdog that returns `PASSTHROUGH` after the deadline, scoped only to the one-pending-task path and not the native GIL-hold root cause.

Closes #2046

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `_compression_deadline_seconds()` and a watchdog around the single-cache-miss inline compression branch in `ContentRouter`.
- Returned the original content with `PASSTHROUGH` and logged a fail-open warning after the configured deadline, while preserving under-deadline and deadline-disabled behavior.
- Added focused regressions for timeout, under-deadline output, and disabled-deadline behavior, then kept the wider deadline suite green.
- Raised the locked production floors for `click` and `pillow` to clear the current `pip-audit` findings that now fail external PR merge snapshots.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_content_router_single_item_deadline.py tests/test_transforms/test_kompress_deadline.py -q`)
- [x] Linting passes (`uv run ruff check headroom/transforms/content_router.py tests/test_content_router_single_item_deadline.py tests/test_transforms/test_kompress_deadline.py`)

### Test Output

```text
uv run pytest tests/test_content_router_single_item_deadline.py tests/test_transforms/test_kompress_deadline.py -q

platform win32 -- Python 3.12.13, pytest-9.0.3, pluggy-1.6.0
rootdir: D:\Repos\headroom-pr-2046-compression-freeze
configfile: pyproject.toml
plugins: anyio-4.12.1, langsmith-0.9.3, asyncio-1.3.0, cov-7.0.0
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 5 items

tests\test_content_router_single_item_deadline.py ...                    [ 60%]
tests\test_transforms\test_kompress_deadline.py ..                       [100%]

============================== 5 passed in 0.42s ==============================

uv run ruff check headroom/transforms/content_router.py tests/test_content_router_single_item_deadline.py tests/test_transforms/test_kompress_deadline.py
All checks passed!

uv run ruff format headroom/transforms/content_router.py tests/test_content_router_single_item_deadline.py tests/test_transforms/test_kompress_deadline.py --check
3 files already formatted
```

## Real Behavior Proof

- Environment: Windows, Python 3.12.13, router-level harness with a cooperative slow-compression stub
- Exact command / steps: `uv run pytest tests/test_content_router_single_item_deadline.py tests/test_transforms/test_kompress_deadline.py -q`, which forces one frozen prefix and one cache miss, then sleeps past a 10 ms deadline
- Observed result: the guarded branch returns the original content through `PASSTHROUGH` at the deadline, while under-deadline and deadline-disabled behavior stay unchanged
- Not tested: native GIL-holding freeze

## Review Readiness

## Checklist

Not applicable.

## Additional Notes

- Security CI currently flags `click 8.3.1` and `pillow 12.2.0`, so this PR carries the narrow floor bump to `click>=8.3.3` and `pillow>=12.3.0` as a supply-chain unblock for the same final merge snapshot.
- `CHANGELOG.md` remains untouched because Headroom generates release notes from conventional commits.
- This PR is a Python-side mitigation for the single-cache-miss branch only; the native GIL-hold root cause remains a separate owner.


JerrettDavis (approved):
Reviewed current head e9733841. The watchdog is scoped to the single-cache-miss inline path, preserves disabled-deadline behavior, and has focused regression coverage. Local ruff passed; local pytest collection hit the Windows _core DLL import issue in the fresh uv environment, while CI's focused test/lint/build jobs are green. I reran the cancelled Release wheel jobs so those statuses can refresh.

JerrettDavis (approved):
Re-reviewed current head 84e8968c after resolving the main conflict. I merged current main, regenerated `uv.lock` from the merged dependency constraints, and kept the PR diff scoped to the single-cache-miss watchdog plus the click/pillow security floors.

I also tightened the new regression so it is robust on Windows: the stalled compression now sleeps long enough to prove the watchdog returns early without relying on a 40 ms cutoff.

- `uvx ruff@0.15.17 check headroom/transforms/content_router.py tests/test_content_router_single_item_deadline.py tests/test_transforms/test_kompress_deadline.py` -> pass
- `uvx ruff@0.15.17 format --check headroom/transforms/content_router.py tests/test_content_router_single_item_deadline.py tests/test_transforms/test_kompress_deadline.py` -> pass
- `uv run --extra dev pytest tests/test_content_router_single_item_deadline.py tests/test_transforms/test_kompress_deadline.py -q` -> 5 passed

Checks are queued on the refreshed head; no code-review blocker remains from this pass.

PR #839, opened 2026-06-11
fix(ccr): skip CCR when model calls headroom_retrieve alongside user tools

agaonker:
## Summary

- When the LLM calls `headroom_retrieve` **and** a non-CCR tool (e.g. `read_file`) in the same turn, the previous code attempted a continuation with only the CCR result
- Anthropic requires every `tool_use` block to have a matching `tool_result` — the continuation was rejected with 400, a round-trip was wasted, and the original response (with unresolved `headroom_retrieve`) was returned anyway
- Fix: if `other_calls` is non-empty alongside `ccr_calls`, log a warning and return the original response immediately — no continuation attempted

## Root cause

`_parse_ccr_tool_calls` correctly separates CCR and non-CCR calls, but `handle_response` never checked `other_calls` before building the continuation. `_create_tool_result_message` only adds results for CCR calls, leaving the non-CCR `tool_use` blocks without matching `tool_result` entries.

## Files changed

- `headroom/ccr/response_handler.py` — guard at top of `while` loop in `handle_response`
- `tests/test_ccr_response_handler.py` — regression test: asserts `api_call_count == 0` and original response returned unchanged when model uses mixed tools

## Test plan

- [x] `pytest tests/test_ccr_response_handler.py::TestCCRResponseHandling::test_handle_response_mixed_tools_skips_ccr` — passes
- [x] `pytest tests/test_ccr_response_handler.py tests/test_ccr_response_handler_extra.py tests/test_ccr_tool_injection.py tests/test_ccr_tool_always_on.py` — 85 passed
- [x] Pre-commit hooks (ruff, mypy) — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2117, opened 2026-07-13
fix(proxy): don't 502 Anthropic streaming on a legal mixed CCR + client-tool turn (#2089)

### LINKED ISSUE #2089: [BUG] Anthropic streaming returns 502 when headroom_retrieve is called alongside a client tool
## Description

An Anthropic `stream: true` request deterministically fails with HTTP 502 when the model emits `headroom_retrieve` and a normal client tool call in the same assistant turn.

This appears to be an interaction between two already-merged fixes:

- #839 intentionally skips CCR continuation when `headroom_retrieve` is emitted alongside another tool, because Headroom cannot provide a `tool_result` for the client-owned tool.
- #1451 buffers Anthropic streaming requests as `stream: false`, handles CCR server-side, and then fails closed if the buffered response still contains `headroom_retrieve`.

The result is that #839 returns the mixed-tool response unchanged, after which #1451 detects the residual internal tool call and returns 502. A valid model behavior therefore causes the complete turn to fail and clients may retry the same upstream generation repeatedly.

## To Reproduce

1. Run Headroom in Anthropic proxy mode with default CCR tool injection enabled:

   ```bash
   headroom proxy --host 127.0.0.1 --port 8788 \
     --mode token --backend anthropic \
     --anthropic-api-url <anthropic-compatible-upstream>
   ```

2. Send an Anthropic `/v1/messages` request with `stream: true`, at least one normal client tool, and enough prior compressed tool-result context for Headroom to inject `headroom_retrieve`.
3. Have the model emit both of these in the same response:

   ```json
   {
     "content": [
       {"type": "tool_use", "name": "headroom_retrieve", "input": {"hash": "..."}},
       {"type": "tool_use", "name": "read_file", "input": {"path": "..."}}
     ]
   }
   ```

4. Observe that `CCRResponseHandler.handle_response()` logs that mixed calls cannot be continued and returns the original response.
5. Observe that the Anthropic streaming handler then sees the residual `headroom_retrieve` and returns an SSE error with HTTP 502.

A deterministic regression test should be possible by extending the mixed-tool fixture from #839 through the buffered streaming path added by #1451.

## Expected Behavior

A legal mixed internal/client-tool response should not make the whole turn fail.

Possible safe outcomes include:

- resolve/defer the CCR call while preserving the client-owned tool call for the client; or
- avoid injecting the internal retrieval tool on a path where mixed calls cannot be resolved; or
- otherwise return a valid client-consumable response without leaking an unresolved internal tool.

The exact policy is a design choice, but repeated upstream generations followed by a deterministic 502 should be avoided.

## Actual Behavior

For each attempt, the upstream returned HTTP 200. Headroom then logged:

```text
CCR: Detected retrieval tool call, handling...
CCR: Skipping CCR handling — model called 1 non-CCR tool(s) alongside headroom_retrieve. Cannot create a valid continuation without results for the other tools. Client must handle all tool calls.
CCR: Retrieval handled successfully
CCR: Buffered streaming response still contains headroom_retrieve after handling; failing closed
```

The client received:

```text
HTTP 502
Unable to safely complete streamed CCR retrieval.
```

The client retried the same request three times; all three upstream calls returned 200 and all three Headroom post-processing passes returned 502.

The `CCR: Retrieval handled successfully` line is also misleading in this branch because handling was explicitly skipped and the residual call immediately triggers fail-closed behavior.

## Code Path

Current `main` still contains both sides of the interaction:

- `headroom/ccr/response_handler.py`: when `other_calls` is non-empty, log and `break`, returning the original response with `headroom_retrieve` still present.
- `headroom/proxy/handlers/anthropic.py`: after buffered handling, if `has_ccr_tool_calls(...)` remains true, return the 502 SSE error `Unable to safely complete streamed CCR retrieval.`

Relevant prior work:

- #839 — `fix(ccr): skip CCR when model calls headroom_retrieve alongside user tools`
- #1450 / #1451 — streaming CCR interception and mixed client/CCR fail-closed behavior
- #1876 / #1897 — adjacent buffered SSE re-synthesis fix, but not this mixed-tool 502 case

I did not find an existing issue that tracks this exact #839 + #1451 interaction.

## Environment

- **Headroom version reproduced**: 0.30.0
- **Latest release reviewed**: 0.31.0; not live-reproduced locally
- **Current `main` reviewed**: the two interacting branches are still present
- **Python version**: 3.11.15
- **OS**: Linux 6.8.0 x86_64
- **LLM Provider/API**: Anthropic Messages-compatible route
- **Client**: Hermes Agent using streaming Anthropic Messages
- **Model observed**: Claude-family model
- **Upstream status**: HTTP 200 on every failed attempt; 502 was generated by Headroom post-processing

## Additional Context

This is not an upstream authentication, quota, or transport failure. The proxy and upstream were healthy, and the failure reproduced identically across three retries with the same mixed-tool response class.

A temporary workaround is `--no-ccr-inject-tool` / `HEADROOM_NO_CCR_INJECT_TOOL=1`, at the cost of disabling model-initiated CCR recovery.

chopratejas:
## Description
Anthropic **buffered-streaming** returned a **502** on a legal turn: when the model emits `headroom_retrieve` alongside a non-CCR client tool, `CCRResponseHandler` intentionally skips CCR resolution (#839) and hands both tool_use blocks back for the client to resolve. The non-streaming path returns that as 200; the streaming path wrongly failed closed with "Unable to safely complete streamed CCR retrieval."

Fix: add a provider-generic `CCRResponseHandler.residual_ccr_status()` → `resolved` / `skipped_mixed_tools` / `error`. The streaming path now only 502s on a genuine `error`; on the intentional mixed-tool skip it falls through to the existing SSE resynthesis (200) preserving **both** tool_use blocks — matching the non-streaming path. The misleading "handled successfully" log no longer fires on skip.

Closes #2089

## Type of Change

## Changes Made
- `headroom/ccr/response_handler.py`: shared, provider-generic `residual_ccr_status()`.
- `headroom/proxy/handlers/anthropic.py`: streaming path fails closed only on a real residual-CCR error; passes through the legal skip as 200 SSE.
- `tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py`: mixed-tool case now returns 200 SSE preserving both blocks.

## Testing
- [x] New/updated tests for the mixed-tool pass-through branch

### Test Output
```text
GitHub CI on current head:
- lint: pass
- build/build-wheel/build-wheel-windows: pass
- test matrix, test-agno, test-extras, test-dashboard-ui: pass
- docker-native-e2e: pass

Review spot-check:
uv run --extra proxy --extra dev python -m pytest tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_ccr_response_handler_extra.py -q
15 passed, 1 warning
```

## Real Behavior Proof
- Environment: GitHub Actions on PR head `c3b2522d`, plus focused Windows review worktree spot-check.
- Before: `stream:true` mixed internal+client tool turn → deterministic 502.
- After: 200 SSE preserving both `headroom_retrieve` and the client tool_use.
- Not tested yet: direct unit coverage for the new `residual_ccr_status()` classifier and the residual-CCR error classification.

## Review Readiness


JerrettDavis (changes requested):
Thanks for tightening the mixed-tool streaming path. The production change makes sense, but I think this needs one small test gap closed before merge.

`residual_ccr_status()` is now the decision point that prevents streaming from either leaking an unresolved `headroom_retrieve` or incorrectly 502ing the legal mixed-tool case. The PR tests the mixed-tool `skipped_mixed_tools` behavior through the Anthropic streaming endpoint, but the second commit removes the only coverage for the residual-CCR error path and there are no direct tests for the new classifier.

Please add deterministic coverage for the classifier states, at least:

- no `headroom_retrieve` -> `RESIDUAL_CCR_RESOLVED`
- `headroom_retrieve` plus another client tool -> `RESIDUAL_CCR_SKIPPED_MIXED`
- `headroom_retrieve` alone -> `RESIDUAL_CCR_ERROR`

A small unit test against `CCRResponseHandler.residual_ccr_status()` should be enough and avoids the fragile streaming setup that was dropped. That keeps the fail-closed branch protected without depending on store-miss/max-round behavior.

PR #2123, opened 2026-07-13
fix(learn): don't desync verbosity pairing on empty assistant turns

abhay-codes07:
## Description

`verbosity._ordered_events` and `_parse_session` disagree about empty assistant turns, which desyncs the response list and produces spurious fast-skips.

`_parse_session` only creates a `_Response` when an assistant message actually said something:

```python
if words > 0 or out_tok > 0:
    responses.append(_Response(...))
```

But `_ordered_events` consumes one `responses[ri]` for **every** assistant line, with no matching filter:

```python
if ltype == "assistant" and ri < len(responses):
    out.append((responses[ri].ts, "assistant", responses[ri]))
    ri += 1
```

So an assistant turn with no text and no output tokens — for example a pure `tool_use` turn where `usage` is absent — creates no `_Response` at parse time, yet still consumes a slot in `_ordered_events`. That slot actually belongs to a *later* real response, so the two lists drift by one. A human reply that follows the real answer is then paired with the next answer's (future) timestamp, `ts - last_resp.ts` goes negative, and since a negative gap is always below the read-fraction threshold, a spurious `fast_skip` is recorded. That inflates `fast_skip_rate`, which feeds `pressure`, which lowers the recommended verbosity level.

The user side of `_ordered_events` already replicates its parse-site filter (`_human_text(...) is None -> continue`); only the assistant side was missing the equivalent guard. That asymmetry is the bug.

## Fix

In `_ordered_events`, compute `words`/`out_tok` for the assistant line the same way `_parse_session` does and only consume a response when `words > 0 or out_tok > 0`, keeping the two functions in lockstep.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/learn/verbosity.py`: `_ordered_events` applies the `words > 0 or out_tok > 0` guard on the assistant branch before consuming a response, with a comment explaining the desync.
- `tests/test_verbosity_learn.py`: add `_empty_assistant` helper and `test_empty_assistant_message_does_not_desync_fast_skip` (an empty assistant turn before a real answer + a slow reply must not record a fast skip).

## Testing

- [x] Unit tests pass (`uv run --extra dev pytest tests/test_verbosity_learn.py::TestSignalExtraction::test_empty_assistant_message_does_not_desync_fast_skip -q`)
- [x] Linting passes (`uvx ruff@0.15.17 check headroom/learn/verbosity.py tests/test_verbosity_learn.py headroom/memory/factory.py`)
- [x] Type checking passes (`uvx mypy==1.20.2 headroom/memory/factory.py`)

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/learn/verbosity.py tests/test_verbosity_learn.py
All checks passed!
$ python -m py_compile headroom/learn/verbosity.py tests/test_verbosity_learn.py
OK
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17`. Importing `headroom` pulls in the torch/transformers stack and a full `pytest` gets OOM-killed on this box, so I verified the alignment with a dependency-free script that models the parse-site filter, the old vs new `_ordered_events` consume, and the resulting human-to-response pairing, and left the full pytest to CI.
- Exact command / steps: built an event stream `[empty assistant, real answer #1, fast human reply, real answer #2, reply]`, computed the response list from the parse filter, then walked the old (unfiltered) and new (filtered) consume to find the gap between the first human and the response paired before it.
- Observed result: old consume pairs the reply with answer #2 (a future timestamp) -> gap `-8` (spurious fast_skip); new consume keeps alignment and pairs it with answer #1 -> gap `+1`. The new test builds a session with an empty assistant turn and a genuinely slow reply and asserts `fast_skips == 0`.
- Not tested: a real Claude Code transcript end to end; full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

Merged current `main` to pick up the repository-wide mypy cache-key annotation fix, then verified the focused regression locally. the change adds the existing parse-site filter to one branch of a pure file-parsing function, verified by the standalone alignment proof and the new regression test for CI.


JerrettDavis (approved):
Reviewed current head f63714ac. Merged current main to pick up the repository-wide mypy fix that caused the prior lint failure, then verified locally: ruff check/format for verbosity/test and memory factory, uvx mypy==1.20.2 headroom/memory/factory.py, and uv run --extra dev pytest tests/test_verbosity_learn.py::TestSignalExtraction::test_empty_assistant_message_does_not_desync_fast_skip -q. CI is rerunning on this head.

JerrettDavis (approved):
Validated current head 9deb049f after merging latest headroomlabs/main.

Checks run:
- `git diff --check`
- `uvx ruff@0.15.17 check headroom/learn/verbosity.py tests/test_verbosity_learn.py`
- `uvx ruff@0.15.17 format --check headroom/learn/verbosity.py tests/test_verbosity_learn.py`
- `uv run --extra dev pytest tests/test_verbosity_learn.py::TestSignalExtraction::test_empty_assistant_message_does_not_desync_fast_skip -q` (1 passed)

PR #2124, opened 2026-07-13
fix(proxy/anthropic): cache response under the looked-up messages (#327)

abhay-codes07:
## Description

The Anthropic response cache stores each entry under a different key than it is looked up by whenever the request pipeline rewrites `messages`, so the cache never hits and fills with unreachable entries.

The handler snapshots the scalar cache-key fields once, before upstream, specifically to avoid post-mutation key drift:

```python
# Snapshot cache-key fields from the request body ONCE here ... The pipeline
# may mutate body before the response is cached, so re-reading there would
# compute a different key and the cache would never hit (#327).
cache_key_fields = {"system": body.get("system"), "tools": body.get("tools"), ...}
...
cached = await self.cache.get(messages, model, **cache_key_fields)   # get
...
await self.cache.set(messages, model, response.content, ..., **cache_key_fields)  # set
```

But `messages` — the primary key component (the cache key is a content hash of `{model, messages, **fields}`) — is passed **live** at both sites, and it is reassigned between them by:

- the enterprise security scan (`messages, _security_ctx = self.security.scan_request(messages, ...)`),
- the `pre_compress` hook (`messages = self.config.hooks.pre_compress(messages, ...)`),
- image compression (`messages = await ...compress(messages, ...)`).

So when any of those paths fires, `cache.set` stores the response under the *post-mutation* messages while every future `cache.get` computes the key from the *raw inbound* messages. The keys never match: the response cache is effectively write-only and each stored entry is unreachable until it is evicted. When none of the three paths mutates `messages`, the keys coincide and the cache works, which is why this went unnoticed. The scalar fields were snapshotted for exactly this reason (#327); `messages` was the one key component left live.

## Fix

Snapshot the lookup messages alongside `cache_key_fields` (before any mutation) and pass that snapshot to `cache.set`, so the entry is stored under the same key it was looked up by. `messages` is only ever reassigned (to new objects) after the snapshot, so the reference stays the raw inbound list. No signature change.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: snapshot `cache_lookup_messages = messages` in the pre-upstream key-snapshot block and use it (not the live `messages`) at `cache.set`.
- `tests/test_anthropic_pre_upstream_backpressure.py`: reuse the `_DummyAnthropicHandler` harness with a recording fake cache and a security scanner that rewrites `messages`; assert the messages passed to `cache.set` equal those passed to `cache.get` (and are the raw lookup messages, not the rewrite).

## Testing

- [x] Unit tests pass (`uv run --extra dev pytest tests/test_anthropic_pre_upstream_backpressure.py::test_response_cache_keys_on_lookup_messages_not_mutated -q`)
- [x] Linting passes (`uvx ruff@0.15.17 check headroom/proxy/handlers/anthropic.py tests/test_anthropic_pre_upstream_backpressure.py headroom/memory/factory.py`)
- [x] Type checking passes (`uvx mypy==1.20.2 headroom/memory/factory.py`)

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/proxy/handlers/anthropic.py tests/test_anthropic_pre_upstream_backpressure.py
All checks passed!
$ python -m py_compile headroom/proxy/handlers/anthropic.py tests/test_anthropic_pre_upstream_backpressure.py
OK
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17`. Importing `headroom` pulls in the torch/transformers stack and running the handler test locally OOM-kills this box, so I verified the key logic with a dependency-free script that replicates the content-hash key from `semantic_cache_key.py`, and left the full pytest (including the new handler test) to CI.
- Exact command / steps: computed the cache key for the raw inbound messages (the `get` key), then computed the `set` key from the live (mutated) messages versus the raw snapshot.
- Observed result: with the mutated messages the set key differs from the get key (cache never hits); with the raw snapshot the set key equals the get key. The new handler test drives a request through a security scanner that rewrites `messages` and asserts `cache.set` receives the same messages as `cache.get`.
- Not tested: a live end-to-end get-hit across two identical requests through the real cache; full local `pytest` deferred to CI (OOM, per above).

## Review Readiness

## Checklist

## Additional Notes

Merged current `main` to pick up the repository-wide mypy cache-key annotation fix, then verified the focused regression locally. the fix snapshots one variable and swaps it at the set site, verified by the key-computation proof and a new regression test that reuses the file's existing, proven `handle_anthropic_messages` harness (injecting the cache and scanner on the instance, so the shared harness is untouched).


JerrettDavis (approved):
Reviewed current head 530e719d. Merged current main to pick up the repository-wide mypy fix that caused the prior lint failure, then verified locally: ruff check/format for the anthropic handler/test and memory factory, uvx mypy==1.20.2 headroom/memory/factory.py, and uv run --extra dev pytest tests/test_anthropic_pre_upstream_backpressure.py::test_response_cache_keys_on_lookup_messages_not_mutated -q. CI is rerunning on this head.

PR #2130, opened 2026-07-13
fix(memory): apply turn_id scope filter even without agent_id

abhay-codes07:
## Description

`SQLiteMemoryStore._build_query_conditions()` dropped `turn_id` when a query specified `session_id` and `turn_id` without also specifying `agent_id`. That made a single-turn scope return every memory in the session, and `count()` uses the same helper.

## Fix

- Apply `agent_id` and `turn_id` as independent narrowing predicates inside the `session_id` branch.
- Preserve the existing `agent_id`-only behavior.
- Add direct query-condition regression tests for turn-only, agent+turn, and agent-only scopes.

## Testing

```text
uvx ruff@0.15.17 check headroom/memory/adapters/sqlite.py tests/test_memory/test_query_conditions.py headroom/memory/factory.py
All checks passed!

uvx ruff@0.15.17 format --check headroom/memory/adapters/sqlite.py tests/test_memory/test_query_conditions.py headroom/memory/factory.py
3 files already formatted

# no output

uv run --extra dev python -m pytest tests/test_memory/test_query_conditions.py -q
3 passed
```

## Review Readiness

- [x] Ready for review
- [x] Regression tests added
- [x] CHANGELOG updated


JerrettDavis (approved):
Reviewed current head 3e257c34. The query builder now applies turn_id independently of agent_id under a session scope, which matches clear_scope semantics and prevents over-broad recall/count results. Focused ruff, format, whitespace, and regression pytest checks passed locally.

PR #2134, opened 2026-07-13
fix(compress): don't mutate the caller's CompressConfig via kwargs

### LINKED ISSUE #2133: compress(config=cfg, protect_recent=0) silently mutates the caller's CompressConfig
## Summary

`headroom.compress.compress()` mutates the `CompressConfig` object the caller passes as `config=`, whenever the same call also supplies a kwarg with a matching field name.

## Impact

Any caller who does the reasonable thing of holding a `CompressConfig` (or a `CompressConfigLike`) on a wrapper class / session / agent, and then does per-call overrides via kwargs, silently corrupts the long-lived object. The corruption persists after the call returns, so:

- The **next** call from the same caller that does NOT pass the same kwarg gets the previous request's override instead of the original default.
- `savings_profile` gets applied against the already-overridden fields on subsequent calls.
- Two threads sharing the same `CompressConfig` get read/write races on the config's fields.

The signature accepts both `config=` and `**kwargs` and documents them as "kwargs override config", strongly implying non-destructive overrides. There is no docstring warning about aliasing.

## Reproduction (unpatched `main`)

```python
from headroom.compress import compress, CompressConfig

c = CompressConfig(protect_recent=4, target_ratio=0.8)
print(f"before: protect_recent={c.protect_recent}, target_ratio={c.target_ratio}")

msgs = [{"role": "user", "content": "x" * 3000}]
compress(msgs, model="claude-sonnet-4-5-20250929", config=c, protect_recent=0, target_ratio=0.2)

print(f"after : protect_recent={c.protect_recent}, target_ratio={c.target_ratio}")
```

Output on `1725cd1f` (main):

```
before: protect_recent=4, target_ratio=0.8
after : protect_recent=0, target_ratio=0.2       <-- caller's object was mutated
```

## Root cause

`headroom/compress.py`:

```python
cfg = config or CompressConfig()          # cfg IS the caller's object
config_fields = {f.name for f in cfg.__dataclass_fields__.values()}
for key, value in kwargs.items():
    if key in config_fields:
        setattr(cfg, key, value)          # mutates the caller's object
```

The `replace(cfg)` a few lines later only runs when `savings_profile` is set — and by then the caller's config is already rewritten.

## Fix

Copy the config once at entry with `dataclasses.replace(config)` before applying kwarg overrides. Proposed in #2134.

## Environment

- Commit: `1725cd1f` (main)
- macOS 25.4 (Darwin), Python 3.12.13

thejesh23:
## Description

`compress(messages, config=my_cfg, protect_recent=0, target_ratio=0.2)` used to write those kwarg values onto the caller's `my_cfg` object — so a shared per-agent `CompressConfig` was silently rewritten every time a call passed a single override. The next call that did NOT override that field then saw the previous request's value instead of the original default.

Copy the config once at entry with `dataclasses.replace` before applying kwarg overrides (and before the savings-profile pass, which also mutates in place). Existing behavior for callers that pass **only** kwargs, or **only** a config, is unchanged.

Issue #2133 has the root-cause walkthrough.

Closes #2133

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/compress.py`: copy the incoming `CompressConfig` once at entry with `dataclasses.replace` before applying kwarg overrides, so the caller's object is no longer mutated. The savings-profile branch already did a defensive `replace(cfg)`; that copy is now hoisted up front so both the kwarg and profile paths share the same guarantee.
- `tests/test_compress_api.py`: added `test_kwargs_do_not_mutate_caller_config`, which fails on unpatched `main` and passes on this branch, covering the previously broken kwarg leg.
- `CHANGELOG.md`: noted the fix.

## Testing

### Test Output

```text
$ uv run pytest tests/test_compress_api.py -q
.................                                                        [100%]
17 passed in 2.88s

$ uv run ruff check headroom/compress.py tests/test_compress_api.py
All checks passed!

$ uv run ruff format --check headroom/compress.py tests/test_compress_api.py
2 files already formatted
```

## Real Behavior Proof

- Environment: macOS 25.4 (Darwin arm64), Python 3.12.13, `uv 0.11.28`, branch `fix/compress-mutates-caller-config`, model `claude-sonnet-4-5-20250929` used for token counting.
- Exact command / steps: build `c = CompressConfig(protect_recent=4, target_ratio=0.8)`, call `compress(msgs, model="claude-sonnet-4-5-20250929", config=c, protect_recent=0, target_ratio=0.2)` on a 3000-char user message, then read `c.protect_recent` and `c.target_ratio` back (full snippet run via `uv run python <<'PY' ... PY` — see the code block below).
- Observed result: before the patch, `c.protect_recent` became `0` and `c.target_ratio` became `0.2` (caller's config silently rewritten). After the patch, `c.protect_recent` stays `4` and `c.target_ratio` stays `0.8`; caller's config unchanged. `uv run pytest tests/test_compress_api.py` reports 17 passed including the new `test_kwargs_do_not_mutate_caller_config` case.
- Not tested: end-to-end proxy path with `savings_profile` set (the pre-fix code already did a defensive `replace(cfg)` on that branch, so the profile leg was safe; this change hoists that copy up front and the added unit test covers the kwarg leg that was broken — I did not spin up the proxy to reconfirm the profile branch end-to-end). No concurrent-caller / threading regression test was added — the fix removes the mutation entirely which sidesteps the race, but there is no explicit multi-thread reproducer.

### Reproducer

**Before the patch (unpatched `main`)**

```text
before: protect_recent=4, target_ratio=0.8
after : protect_recent=0, target_ratio=0.2       # <-- caller's cfg silently rewritten
caller's config MUTATED
```

**After the patch (this branch)**

```text
$ uv run python <<'PY'
from headroom.compress import compress, CompressConfig
c = CompressConfig(protect_recent=4, target_ratio=0.8)
print(f"before: protect_recent={c.protect_recent}, target_ratio={c.target_ratio}")
msgs = [{"role":"user","content":"x"*3000}]
compress(msgs, model="claude-sonnet-4-5-20250929", config=c, protect_recent=0, target_ratio=0.2)
print(f"after : protect_recent={c.protect_recent}, target_ratio={c.target_ratio}")
print("caller's config", "unchanged" if (c.protect_recent, c.target_ratio) == (4, 0.8) else "MUTATED")
PY
before: protect_recent=4, target_ratio=0.8
after : protect_recent=4, target_ratio=0.8
caller's config unchanged
```

## Review Readiness

## Checklist

## Additional Notes

- **Documentation checklist item** — left unchecked as N/A. The behavior being fixed is internal to `headroom.compress.compress()`; the mutation contract of `CompressConfig` + kwargs is not covered in any user-facing doc (`wiki/compression.md`, `wiki/text-compression.md`, `wiki/image-compression.md`, and `docs/content/docs/shared-context.mdx` document a different / higher-level API surface). The `CHANGELOG.md` entry is the appropriate place for this fix.
- **`mypy headroom` checklist item** — left unchecked because I did not run it in this workflow; the change is a two-line refactor within a well-typed function and no signatures moved.
- The prior body's `## Summary`, `## Test plan`, and `## Real behavior proof` sections were reorganized into the six template-required headings so the PR-governance check passes. All technical content (root-cause, before/after reproducer, and test output) is preserved above; no code changes were made in this update.


JerrettDavis (approved):
Reviewed the current head. Copying the incoming `CompressConfig` with `dataclasses.replace()` before applying kwargs is the right boundary: callers keep ownership of their config object, while per-call overrides and savings profile mutation still apply only to the local request config. The regression test covers the previously broken kwargs path, and current CI lint/build/test jobs are passing.

The remaining red statuses are stale cancelled Security / E2E / Governance / evaluation jobs from an earlier run, not failures in this diff. I corrected the readiness label and will rerun those cancelled checks so the PR status can reflect the current head.

PR #2136, opened 2026-07-13
fix(shared_context): don't evict an unrelated entry on an update at capacity

### LINKED ISSUE #2135: shared_context: SharedContext.put evicts an unrelated entry on an update at capacity (mirror of #2094 in SemanticCache)
## Summary

`SharedContext.put` evicts a perfectly good, unrelated entry when it merely updates a key that is already cached at capacity — the exact same defect that was fixed for `SemanticCache` in #2094, but only patched there.

## Reproduction (unpatched `main`)

```python
from headroom.shared_context import SharedContext

ctx = SharedContext(ttl=3600, max_entries=3)
ctx.put("a", "x" * 400)
ctx.put("b", "x" * 400)
ctx.put("c", "x" * 400)
print("before update:", sorted(ctx.keys()))

# Update an existing key. The map size does not grow — but 'a' is dropped.
ctx.put("c", "y" * 400)
print("after update: ", sorted(ctx.keys()))
```

Output on `1725cd1f` (main):

```
before update: ['a', 'b', 'c']
after update:  ['b', 'c']       <-- 'a' evicted even though 'c' was an update, not an insert
```

## Root cause

`headroom/shared_context.py`:

```python
def put(self, key: str, content: str, *, agent: str | None = None) -> ContextEntry:
    ...
    with self._lock:
        self._evict_if_needed()          # evicts BEFORE checking whether 'key' already exists
        self._entries[key] = entry

def _evict_if_needed(self) -> None:
    ...
    while len(self._entries) >= self._max_entries:
        oldest_key = min(self._entries, key=lambda k: self._entries[k].timestamp)
        del self._entries[oldest_key]
```

The eviction loop has no way to know the incoming key, so it treats an update as an insert. This is exactly the same defect class as `SemanticCache.put` before #2094.

## Impact

In multi-agent workflows this is the primary way an agent gives another agent access to its latest data. Refreshing an entry (`put("research_findings", updated_text)`) at capacity silently loses an unrelated key from another agent — no exception, no log — so a downstream `ctx.get("some_other_key")` starts returning `None`.

## Fix

Thread the incoming key through `_evict_if_needed` and skip capacity eviction when it names an entry that already exists. Expired entries are still cleaned unconditionally. Proposed in #2136.

## Environment

- Commit: `1725cd1f` (main)
- macOS 25.4 (Darwin), Python 3.12.13

thejesh23:
Fixes #2135.

## Summary

`SharedContext.put` ran `_evict_if_needed` before writing, and the eviction loop only checked `len(self._entries) >= self._max_entries`. When a caller updated a key that was already cached at capacity, the put would not have grown the map — but the loop still dropped the oldest unrelated entry.

Same defect class as fixed for `SemanticCache` in #2094: the eviction path must know the incoming key so an update is not treated as an insert. This mirrors that fix over to `SharedContext`.

Threads the incoming key through `_evict_if_needed` and skips capacity eviction when it names an entry that already exists. Expired-entry cleanup still runs unconditionally.

Issue #2135 has the reproduction and impact writeup.

## Test plan

- [x] `uv run pytest tests/test_shared_context.py` — 16 passed (added `test_updating_existing_key_at_capacity_does_not_evict`).
- [x] `uv run ruff check headroom/shared_context.py tests/test_shared_context.py` — clean.
- [x] `uv run ruff format --check headroom/shared_context.py tests/test_shared_context.py` — already formatted.

## Real behavior proof

**Setup:** macOS 25.4 (Darwin arm64), Python 3.12.13, `uv 0.11.28`, this branch (`fix/shared-context-evict-on-update`).

**Before the patch (unpatched `main`)**

\`\`\`
before update: ['a', 'b', 'c']
after update:  ['b', 'c']        # <-- 'a' evicted, even though 'c' was an update
\`\`\`

**After the patch (this branch)**

\`\`\`
\$ uv run python <<'PY'
from headroom.shared_context import SharedContext
ctx = SharedContext(ttl=3600, max_entries=3)
ctx.put(\"a\", \"x\"*400)
ctx.put(\"b\", \"x\"*400)
ctx.put(\"c\", \"x\"*400)
print(\"before update:\", sorted(ctx.keys()))
ctx.put(\"c\", \"y\"*400)   # update existing at capacity
print(\"after update: \", sorted(ctx.keys()))
print(\"c value:\", ctx.get(\"c\", full=True)[:12] + \"...\")
PY
before update: ['a', 'b', 'c']
after update:  ['a', 'b', 'c']
c value: yyyyyyyyyyyy...
\`\`\`

**Test output**

\`\`\`
\$ uv run pytest tests/test_shared_context.py -q
................                                                         [100%]
16 passed in 2.17s
\`\`\`

**What I did NOT test**

- Multi-thread test — the fix is inside the existing `self._lock`, so serialization semantics are unchanged; I did not add a concurrent-put stress test.
- Interaction with TTL expiry AND capacity in one call — the existing `test_evicts_oldest_at_capacity` and `test_expired_entry_returns_none` still pass, but I did not add a combined case.


JerrettDavis (approved):
Reviewed the current head. The fix keeps expired-entry cleanup unconditional while avoiding capacity eviction when `put()` is updating an existing key, and the regression test covers the at-capacity overwrite case directly. The current CI workflow jobs for lint/build/tests are passing.

The remaining red statuses I see are cancelled older Security / Init E2E / Wrap E2E / Merge Conflicts / PR Governance jobs, not failures from this diff. I corrected the stale readiness label and will try to rerun those cancelled checks so the status reflects the current head.

PR #2139, opened 2026-07-13
fix(memory): size HNSW index_batch resize off the id high-water mark

abhay-codes07:
## Description

`HNSWVectorIndex.index_batch()` used the live memory map size to decide whether to resize hnswlib before adding new labels. hnswlib does not reclaim capacity slots when labels are removed with `mark_deleted`, so after delete/evict churn the live count can be much lower than the assigned-id high-water mark. That lets a batch add skip resizing and then fail in `add_items` with `number of elements exceeds the specified limit`.

## Fix

- Size the batch resize check from `self._next_hnsw_id`, which has already been incremented for the new batch labels.
- Match the single-item `index()` path's high-water-mark capacity behavior.
- Add a regression test that deletes most entries from a small index and then batch-adds enough new memories to require a resize.

## Testing

```text
uvx ruff@0.15.17 check headroom/memory/adapters/hnsw.py tests/test_memory/test_hnsw_batch_capacity.py headroom/memory/factory.py
All checks passed!

uvx ruff@0.15.17 format --check headroom/memory/adapters/hnsw.py tests/test_memory/test_hnsw_batch_capacity.py headroom/memory/factory.py
3 files already formatted

# no output

uv run --extra dev python -m pytest tests/test_memory/test_hnsw_batch_capacity.py -q
1 passed, 18 warnings
```

## Review Readiness

- [x] Ready for review
- [x] Regression test added
- [x] CHANGELOG updated


JerrettDavis (approved):
Reviewed current head 55e97b3b. The batch resize now keys off hnswlib's assigned-label high-water mark, which matches the existing single-item path and handles mark_deleted slots correctly. Focused ruff, format, whitespace, and HNSW regression pytest checks passed locally.

PR #2140, opened 2026-07-13
fix(transforms): guard the lossless diff fold to diff-shaped content only

abhay-codes07:
## Description

`ContentRouter._lossless_first` tries several lossless folds and keeps the smallest result. Most folds self-verify reversibility, but the `diff` fold is subtractive: it removes `index <hex>..<hex>` lines without an inverse check. That can silently delete matching lines from non-diff text, logs, or search output.

## Fix

- Skip the `diff` fold unless the detected strategy is `CompressionStrategy.DIFF` or the content is structurally diff-shaped.
- Keep genuine diffs folding their `index` bookkeeping.
- Add regression coverage for a non-diff `index ...` line and a real diff.

## Testing

```text
uvx ruff@0.15.17 check headroom/transforms/content_router.py tests/test_lossless_diff_fold_guard.py headroom/memory/factory.py
All checks passed!

uvx ruff@0.15.17 format --check headroom/transforms/content_router.py tests/test_lossless_diff_fold_guard.py headroom/memory/factory.py
3 files already formatted

# no output

uv run --extra dev python -m pytest tests/test_lossless_diff_fold_guard.py -q
2 passed
```

## Review Readiness

- [x] Ready for review
- [x] Regression tests added
- [x] CHANGELOG updated


JerrettDavis (approved):
Reviewed current head 06e04d10. The guard is scoped correctly: non-diff content no longer tries the subtractive diff fold, while DIFF strategy and structurally diff-shaped content still retain the existing fold behavior. Focused ruff, format, whitespace, and regression pytest checks passed locally.

PR #2144, opened 2026-07-13
fix(backends/litellm): preserve tool_result cache_control, complete streaming cache stats

ingmarkruschHF:
## Description

Two gaps remain in Bedrock Converse prompt caching after [#1390](https://github.com/headroomlabs-ai/headroom/pull/1390) (currently open, not yet merged), which preserves `cache_control` on the system prompt and plain text blocks:

1. `_convert_messages_for_litellm` still drops `cache_control` on `tool_result` blocks during Anthropic-to-OpenAI conversion. #1390's own test (`test_tool_result_blocks_unaffected`) documents this as explicitly out of scope. In practice this is the case that matters most: in agent loops the moving cache breakpoint (what Claude Code marks with `cache_control: {type: ephemeral}`) lands on the tail `tool_result` message far more often than on the system prompt, so caching degraded to system-only even with #1390 applied.
2. `stream_message` never requests `stream_options.include_usage`, so LiteLLM/Bedrock never returns a usage chunk over SSE and `cache_read_input_tokens`/`cache_creation_input_tokens` always read 0 downstream, even when the Bedrock prompt cache is genuinely engaged server-side. The `message_start` event emitted before streaming begins is necessarily sent before any usage is known (hardcoded `input_tokens: 0`, no cache fields) — this PR captures the real values from the trailing usage chunk, carries them on the terminal `message_delta.usage`, and updates `StreamingMixin._stream_response_bedrock` to record those fields while preserving the normal Anthropic stream event order.

I raised the streaming cache-stats gap directly in the [#1390 comment thread](https://github.com/headroomlabs-ai/headroom/pull/1390#issuecomment-4845691613); another reviewer (`dspv`) independently flagged the tool_result gap in the same thread with measurements matching what I found on this deployment. This PR is the follow-up with the actual diff and tests, scoped to only what #1390 doesn't cover.

Closes #

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `headroom/backends/litellm.py`:
  - `_convert_messages_for_litellm`: the `tool_result` → `role: "tool"` conversion now carries `cache_control` from the source block onto the emitted message when present.
[...]

Issue #2110, opened 2026-07-13
Dynamic content detector in cache_aligner produces false positives that corrupt cached system prompts


inzy007:
## Summary

`DynamicContentDetector` / `RegexDetector` in `headroom/cache/dynamic_detector.py` (used by
`CacheAlignerConfig` / the `cache_aligner` transform) misclassifies ordinary English words and
code identifiers as "dynamic content," extracting them out of the system prompt and re-appending
them as a `[Dynamic Context]` tail on every request. Over a long session this tail grows
unboundedly and becomes visible, garbled noise in the actual model context.

Observed in headroom-ai 0.5.10, wrapping Claude Code. Example corrupted tail from a real session:

    ---
    [Dynamic Context]
    in_progress
    DebugModal (function — src/components/layout/DebugModal.tsx:11)
    getAuthToken (function — src/services/firebase.ts:92)
    in_progress
    ...

None of that is actually dynamic — it's static tool/status vocabulary and source identifiers that
recur identically every turn, so it should never have been split out of the cacheable prefix in
the first place (which defeats the point of `cache_aligner`).

## Root cause — three separate bugs in the same file

**1. Structural label pattern has no word boundary and allows a bare-whitespace separator**
(around line 341, `RegexDetector.__init__`):

```python
labels_pattern = "|".join(re.escape(label) for label in config.dynamic_labels)
self._structural_pattern = re.compile(
    rf"(?P<label>(?:{labels_pattern}))(?P<sep>\s*[:=]\s*|\s+)(?P<value>[^\n,;]+)",
    re.IGNORECASE,
)
```

- No `\b` before the label, so labels like `"token"` or `"last"` match as substrings inside
  unrelated identifiers, e.g. `getAuthToken`.
- The separator alternation includes bare `\s+` (no `:`/`=` required), so it matches ordinary
  English sentences — anything like `"...last updated the file..."` gets read as
  label=`"last"`, sep=`" "`, value=the rest of the clause.

**2. Entropy fallback doesn't exclude separator-joined dictionary words**
(around line 483, the high-entropy fallback tier):

```python
# Skip if all letters or all numbers (not random-looking)
if text.isalpha() or text.isdigit():
    continue
```

This only skips pure-alpha tokens. Hyphen/underscore-joined tokens like `in_progress`,
`system-reminder`, `total_tokens` aren't `isalpha()` (they contain `-`/`_`), so they fall through
[...]

PR #2151, opened 2026-07-13
fix(backends/litellm): preserve cache_control through the Anthropic->OpenAI conversion

eilhamv:
## Description

The LiteLLM backend's Anthropic→OpenAI conversion silently discards `cache_control` markers, so **prompt caching never engages for Anthropic-schema clients on the LiteLLM-routed backends** (`--backend bedrock` / `openrouter` / `any-llm`). This contradicts both the docs ("the provider cache hot zone [is] left untouched so prompt caching keeps working") and the content router's own contract (`transforms/content_router.py`: *"we never modify cache_control'd blocks"*, with a `cache_control_protected` route counter) — the router protects the markers, then the backend conversion throws them away before LiteLLM can translate them into provider cache points (e.g. Bedrock `cachePoint`).

Direct Anthropic (`--backend anthropic`, the default) is unaffected — `LiteLLMBackend` is only constructed when `config.backend != "anthropic"`. Request-side complement to the response-side fix from #1345 (`_anthropic_usage_from_litellm`), which surfaces cache stats that — on this ingress — never existed because the markers died on the way in.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- New `_system_to_openai_content()`: converts Anthropic system blocks to OpenAI content blocks with `cache_control` preserved when any block carries a marker; byte-identical joined-string behavior otherwise. Used at both system conversion call sites (streaming + non-streaming).
- `_convert_messages_for_litellm()`: text messages keep per-block `cache_control` (block-list form emitted only when a marker is present); `tool_result` conversion lifts `cache_control` from the block (or its inner content blocks) onto the converted tool message.
- `_convert_anthropic_tool()`: copies `cache_control` to the converted tool (LiteLLM's supported convention; enables tool-schema caching for MCP-heavy agents).
- New `tests/test_backend_cache_control.py`: 11 unit tests pinning marker preservation for system / text / tool_result / tools, plus no-marker legacy-form behavior.

## Testing

### Test Output

```text
$ python -m pytest tests/test_backend_cache_control.py -q
======================== 11 passed, 1 warning in 2.59s =========================

$ ruff check headroom/backends/litellm.py tests/test_backend_cache_control.py
All checks passed!

$ python -m pytest tests/test_backend_bugs.py -q   # identical results before/after this diff
(9 failed, 28 passed — the 9 failures are pre-existing dependency-version
mismatches in my environment; byte-identical failure set on clean main)
```

## Real Behavior Proof

- Environment: headroom 0.30.0 wheel with this patch applied, `headroom proxy --backend bedrock --bedrock-region us-west-2`, EKS pod with IRSA, model `us.anthropic.claude-haiku-4-5` / `claude-sonnet-4-5` via Bedrock inference profiles.
- Exact command / steps: POST `/v1/messages` with a >1024-token system block carrying `cache_control: {"type": "ephemeral"}`, sent twice; repeat with the marker on a user-message content block; plus a no-marker request and an OpenAI-ingress compression payload as regression controls.
- Observed result:
  - Before: `cache_creation_input_tokens=None / cache_read_input_tokens=None` on both calls (no caching at all).
  - After: system marker — call 1 `cache_write=7212`, call 2 `cache_read=7212`; user-block marker — call 1 `cache_write=12613`, call 2 `cache_read=12613`. No-marker request byte-identical billing before/after; OpenAI-ingress regression payload billed identically (4,602 tokens) before/after.
- Not tested: `mypy` (not run in my environment); the *streaming* Anthropic response path's usage events (request-side caching works for streaming too, but the streamed usage report may still omit cache stats on 0.30.x — happy to extend this PR if you point me at the preferred site; `main`'s #1345 comment suggests it may already be handled there).

## Review Readiness

## Checklist

N/A — wire-level fix; the "Real Behavior Proof" section carries the observable evidence.

## Additional Notes

- The `Dependency audit (pip-audit)` check failure appears unrelated to this PR — no dependencies are added or changed here.
- Docs and CHANGELOG untouched: the fix makes behavior match the existing documentation rather than changing documented behavior. Happy to add a CHANGELOG entry if you'd like one.


chopratejas:
Does this cache busting only show up if you're using LiteLLM (and not Anthropic etc endpoint directly, right?)

eilhamv:
Correct — direct Anthropic (`--backend anthropic`, the default) is unaffected. On that path the request stays in Anthropic format end-to-end and the content router already protects `cache_control` blocks, so markers reach the API intact.

The loss happens only on the **LiteLLM-routed backends** (`--backend bedrock` / `openrouter` / `any-llm`), where `backends/litellm.py` converts Anthropic→OpenAI format and the flattening drops the markers — `LiteLLMBackend` is only constructed when `config.backend != "anthropic"` (`cli/proxy.py`), so the blast radius is exactly that conversion.

One terminology nuance: it's not cache *busting* (prefix instability, à la #2110/#2085) — it's total marker loss: the provider never creates a cache entry at all, so every turn bills full input price. Repro before the fix (`--backend bedrock`, `/v1/messages`, >1024-token marked system block, sent twice): both calls returned `cache_creation_input_tokens=None / cache_read_input_tokens=None` and identical full-price input. Same request with `--backend anthropic` caches normally.

Practical impact: any Anthropic-schema client in front of a Bedrock/OpenRouter-backed proxy — e.g. Claude Code via `ANTHROPIC_BASE_URL` (which marks system + tools + conversation by default) — silently loses prompt caching entirely, while everything else appears to work.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

JerrettDavis (approved):
Reviewed the LiteLLM conversion changes. Preserving cache_control while keeping the legacy string form when no marker is present is the right compatibility boundary, and the added tests cover system blocks, text blocks, tool results, tools, and no-marker behavior. I also removed the stale status: needs author action label and marked it ready for review.

I did not get a useful local pytest run in the disposable worktree because the ad hoc environment lacked pytest, and this PR's current GitHub run only has governance checks because the previous label job was cancelled. Code-wise this looks ready; it still needs a fresh CI run/check set from GitHub before merge.

JerrettDavis (approved):
Merged current upstream main and resolved the LiteLLM cache_control conflict in favor of preserving markers from both top-level tool_result blocks and nested text blocks.

Validated current head 186e5163 locally:
- `uvx ruff@0.15.17 check headroom/backends/litellm.py tests/test_backend_cache_control.py`
- `uvx ruff@0.15.17 format --check headroom/backends/litellm.py tests/test_backend_cache_control.py`
- `uv run --extra dev pytest tests/test_backend_cache_control.py -q` (11 passed)
- `git diff --check`

chopratejas:
Superseded by #2144 (merged), which restores cache_control preservation through the Anthropic->OpenAI/LiteLLM conversion on the same tool_result branch, plus the streaming cache-stats fix. Consolidating on #2144 to avoid a conflicting double-merge. Thanks — the coverage here is captured by #2144.

eilhamv:
Thanks — confirmed #2144 covers the tool_result branch (verified behaviorally against current main). Three marker classes from this PR remain though, all reproducible on `main` @ 4e30dde with `--backend bedrock`:

1. **System blocks** — both conversion sites still flatten with `" ".join(...)`; a >1024-token marked system block sent twice returns `cache_write/read=None` on both calls. This is the flagship case (Claude Code via `ANTHROPIC_BASE_URL` marks the system prompt, so those users still get zero system/prefix caching on the LiteLLM backends).
2. **Text content blocks** — `_convert_messages_for_litellm` drops per-block `cache_control` (unit-testable: a block-list user message with a marker converts to output with none).
3. **Tool definitions** — `_convert_anthropic_tool` drops `cache_control`, so tool-schema caching (the large stable prefix for MCP-heavy agents) never engages.

Possibly the most relevant interaction: `normalize_message_cache_control` re-places the proxy's own single breakpoint "on the last block of the last block-style message" — but on the LiteLLM backends that marker looks like it gets dropped by the same text-block conversion whenever the last block-style message is text, which would undermine the new overlay/prefix machinery on those backends. Worth checking.

Happy to rebase this PR down to those three sites (the tool_result part drops in favor of #2144), tests included — or refile as a fresh scoped PR/issue, whichever you prefer.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2155, opened 2026-07-13
ci: scope native/wheel/dashboard jobs to relevant paths (unstarve the queue)

chopratejas:
## Problem
CI's `changes` paths-filter is too coarse: both `code` and `e2e` include the catch-all `headroom/**`, so **any** Python change fires the full matrix — including the scarce-runner hogs: `build-wheel-windows`, `macos-native-wrapper`, `windows-native-wrapper`, `docker-native-e2e`. macOS runners cap at 5 on the Free plan, so a trivial Python PR burns the scarcest resource. `init-e2e`/`wrap-e2e` trigger on `headroom/**` too.

## Change (CI config only — no product code)
- **ci.yml:** split `changes` into `code` / `native` / `dashboard` / `packaging`. Re-gate:
  | job | before | after |
  |-----|--------|-------|
  | `build-wheel-windows` | code | `packaging` |
  | `test-dashboard-ui` | code | `dashboard` |
  | `docker-native-e2e` | e2e | `native` |
  | `windows-native-wrapper` | e2e | `native` |
  | `macos-native-wrapper` | e2e | `native` |
- **init-e2e / wrap-e2e:** scoped off `headroom/**` onto the `cli`/`install`/`providers`/`rtk` subpaths the flows exercise (mirrors the existing `*-native-e2e` workflows).

## Safety
- Verified no job `needs:` any of the re-gated jobs, so no dependency chain breaks. `test-dashboard-ui` still gets `build-wheel` (dashboard files are under `headroom/` ⟹ `code` too).
- Job-level `if:` skips resolve as **neutral** (not failing), so this is safe even if branch protection is added.
- Full matrix still runs on any `.github/workflows/**` change and on push to `main` (post-merge safety net).

## Net
A pure-Python logic PR skips ~7-9 jobs incl. all 3 native platform builds — freeing the macOS/Windows caps that were starving the queue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,9 @@ jobs:
     timeout-minutes: 5
     outputs:
       code: ${{ steps.filter.outputs.code }}
-      e2e: ${{ steps.filter.outputs.e2e }}
+      native: ${{ steps.filter.outputs.native }}
+      dashboard: ${{ steps.filter.outputs.dashboard }}
+      packaging: ${{ steps.filter.outputs.packaging }}
       workflows: ${{ steps.filter.outputs.workflows }}
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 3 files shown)


JerrettDavis (changes requested):
Thanks for taking this on. The direction is right, but I think the E2E path filters are currently a little too narrow for the claim that they cover what the flows exercise.

`init-e2e.yml` now only includes `headroom/cli/**` and `headroom/install/**` for Python package code, but `headroom init` imports provider-specific code directly (`headroom.providers.claude`, `headroom.providers.codex.*`) and installs MCP servers via `headroom.mcp_registry`. Changes in those modules can change init behavior while this workflow would no longer run. Please add at least `headroom/providers/**` and `headroom/mcp_registry/**` to the init E2E paths, or keep the previous broader coverage for those areas.

Same issue on `wrap-e2e.yml`: wrap flows register Headroom/Serena MCP servers through `headroom.mcp_registry` (`_setup_headroom_mcp`, `_setup_serena_mcp`, OpenCode registrar paths, etc.), but `headroom/mcp_registry/**` is not included in the narrowed paths. A registrar/spec/ledger regression could now skip wrap E2E even though the Docker harness exercises those paths. Please include `headroom/mcp_registry/**` there as well.

Validation I ran locally on the current head:
- `actionlint .github/workflows/ci.yml .github/workflows/init-e2e.yml .github/workflows/wrap-e2e.yml`

Separate from this PR's workflow changes, the current CI failure is the security audit finding for `transformers 5.3.0` in `uv.lock` (`CVE-2026-5241`, fixed in 5.5.0). That appears to be a base/repo-wide dependency issue rather than caused by this PR, but it will need a dependency update for the PR to go green.

PR #2162, opened 2026-07-14
fix(proxy): isolate image compression in a subprocess so a native SIGSEGV can't take down the proxy (#2107)

### LINKED ISSUE #2107: Proxy segfaults on image compression (SIGSEGV in OpenCV KleidiCV ARM resize) — takes down all routed clients with ConnectionRefused
## Summary

On Apple Silicon, the proxy **hard-crashes with SIGSEGV** whenever it compresses an image. The fault is inside OpenCV's KleidiCV ARM NEON resize, reached from the SigLIP ONNX image encoder in `headroom/image/onnx_router.py`.

Because the proxy is the API endpoint for every routed client (`ANTHROPIC_BASE_URL=http://127.0.0.1:8787`), the crash takes down **every agent on the machine at once**. They then fail with `Unable to connect to API (ConnectionRefused)` and retry into a closed port until someone manually restarts the proxy. There's no supervision, so it stays down indefinitely.

This reads to users as an Anthropic outage or a rate limit — it took a while to trace it back to Headroom at all. The only visible clue is a macOS "Python quit unexpectedly" dialog.

## Environment

| | |
|---|---|
| Headroom | 0.30.0 (code inspection says 0.31.0 is also affected — see below) |
| OS | macOS 26.3, Apple Silicon (arm64) |
| Python | 3.13.14 (Homebrew), pipx install |
| opencv-python | 5.0.0.93 (pulled in transitively) |
| Client | Claude Code 2.1.206 |

## Reproduce

1. `headroom proxy` on an arm64 Mac with `opencv-python` 5.x resolved into the venv.
2. Point a client at it (`ANTHROPIC_BASE_URL=http://127.0.0.1:8787`).
3. Send any request containing an image.
4. The proxy segfaults. Every client on the machine now gets `ConnectionRefused`.

Crashed 4 times over 4 days here, each time an image entered the pipeline.

## Faulting stack (identical across all 4 crash reports)

```
Exception: EXC_BAD_ACCESS (SIGSEGV) — KERN_INVALID_ADDRESS

cv2.abi3.so   kleidicv::neon::kleidicv_resize_generic_stripe_u8<2, 3>(...)
cv2.abi3.so   kleidicv_thread_std_function_callback(unsigned int, unsigned int, void*)
cv2.abi3.so   kleidicv::hal::parallel(...)
cv2.abi3.so   cv::(anonymous namespace)::block_function(void*, unsigned long)
libdispatch   _dispatch_client_callout2
libdispatch   _dispatch_apply_invoke3
...
cv2.abi3.so   cv::parallel_for_(cv::Range const&, cv::ParallelLoopBody const&, double)
```

The proxy log stops mid-image-load, ~240ms before the crash timestamp:

```
23:29:29,907 headroom.image.onnx_router INFO Loading technique-router ONNX INT8...
23:29:29,998 headroom.image.onnx_router INFO Technique router loaded: 4 classes, ONNX INT8 (32 MB)
23:29:30,005 headroom.image.onnx_router INFO Loading SigLIP ONNX INT8 image encoder...
23:29:30,147 headroom.image.onnx_router INFO SigLIP image encoder loaded: ONNX INT8 (94 MB)
<process dies — crash report captureTime 23:29:30.384>
```

## Cause

`opencv-python` 5.x routes arm64 resizes through the KleidiCV HAL. That path faults. OpenCV 4.x has no KleidiCV backend and does not crash.

Headroom doesn't depend on `cv2` directly — it arrives transitively (`onnxruntime`, `rapidocr`, `transformers`) and is unpinned, so a fresh install resolves to 5.x and is exposed by default.

## Workaround

```bash
pipx runpip headroom-ai install 'opencv-python<5'   # 4.13.0.92 — no KleidiCV, no crash
```

## Suggested fixes

1. **Pin `opencv-python<5`** (or `opencv-python-headless<5`) in the dependency set. One-line fix that closes the crash for every arm64 user.
2. **Don't let the image path take the process down.** A SIGSEGV in a native lib can't be caught in Python, so consider running image compression in a worker subprocess — a crash there should degrade to "image uncompressed," not kill the proxy and every client with it.
3. **Expose `image_optimize`.** There's currently no CLI flag or env var for it — `ProxyConfig.image_optimize` and `ContentRouterConfig.enable_image_optimizer` both default `True` with no supported override, so the only way to disable the crashing feature is patching site-packages, which `headroom update` reverts. A `HEADROOM_IMAGE_OPTIMIZE=0` would have let us disable it cleanly. **Still true in 0.31.0** (both defaults still `True`, still no toggle, OpenCV still unpinned) — so upgrading does not fix this.

## Two smaller things noticed in the same logs

**The ONNX models reload on every single request.** Both `Loading technique-router ONNX INT8...` (32 MB) and `Loading SigLIP ONNX INT8 image encoder...` (94 MB) appear once per image request, not once per process. `_get_image_compressor()` explicitly builds a short-lived compressor per call and only memoizes *importability*, so the sessions are never reused. That's ~126 MB of model load in the hot path of each request.

**Image compression reports negative savings.** Every image in our logs came out roughly double the tokens it went in as:

```
Image compression: preserve (850 → 1700 tokens, -100% saved)
```

If `preserve` is expected to be a no-op, something is double-counting on the way out — it's being reported as a 100% *loss*.

rodboev:
## Description

On Apple Silicon (arm64) macOS the proxy hard-crashes with SIGSEGV the moment it compresses an image, and because the proxy is the single API endpoint for every routed client (`ANTHROPIC_BASE_URL=http://127.0.0.1:8787`), one image request takes down every agent on the machine at once — they then fail with `ConnectionRefused` and retry into a closed port until the proxy is manually restarted.

The faulting stack is inside OpenCV's KleidiCV ARM NEON resize (`kleidicv::neon::kleidicv_resize_generic_stripe_u8`), reached from the SigLIP ONNX image encoder during `ImageCompressor.compress()`. The proxy runs that call on a `ThreadPoolExecutor` (`headroom/proxy/server.py:946`), and both handler call sites (`headroom/proxy/handlers/anthropic.py:1148-1172`, `headroom/proxy/handlers/openai.py:2274-2296`) wrap it in `try/except Exception` intending to fail open. That guard cannot help: a native SIGSEGV is not a Python exception, and a segfault on any worker thread aborts the whole interpreter. Thread isolation is not crash isolation.

The defect Headroom owns is that an optional, best-effort, native-heavy transform runs in-process with no crash boundary, so any native fault in it is fatal to the proxy and to every unrelated client it fronts. This change gives image compression a real crash boundary by executing it in a spawned subprocess, so a native crash degrades to "image forwarded uncompressed" instead of killing the proxy.

Closes #2107.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `headroom/proxy/image_isolation.py` with `run_image_compression_isolated(messages, provider, *, timeout) -> tuple[list[dict], dict | None]`, which runs an `ImageCompressor.compress` worker inside a lazily-created module-level `ProcessPoolExecutor(max_workers=1)` on a **spawn** multiprocessing context (ONNX sessions are not fork-safe) and carries the compression result (technique, `savings_percent`, token counts) back across the process boundary. The native OpenCV/KleidiCV work now runs in the child address space.
- Made the runner fail open for **any** child outcome: `BrokenProcessPool` (the class raised when the child is killed by a signal, i.e. SIGSEGV/SIGABRT), `TimeoutError`, or any other `Exception` all return `(messages, None)` — the original `messages` unchanged, no telemetry — and reset the pool so the next request re-spawns a fresh child.
- Routed the two request-path image-compression sites (`handlers/anthropic.py`, `handlers/openai.py`) through the runner, keeping the existing `ImageCompressionDecision` gate and the `image_compression` mutation tag, and emitting the savings `INFO` log line from the runner's returned result on the success path.
- Scoped strictly to crash containment: `config.image_optimize` default is unchanged, no dependency pins, no new env switches.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_image_compression_isolation.py tests/test_image_compression_offload.py`)
- [x] Linting passes (`uv run ruff check .`)

### Test Output

```text
$ uv run pytest tests/test_image_compression_isolation.py tests/test_image_compression_offload.py -q
.......                                                                   [100%]
7 passed in 1.73s
```

The reproduction test (`test_worker_sigsegv_fails_open_parent_survives`) spawns a real subprocess whose worker dies by signal (`os.abort()`), then asserts `run_image_compression_isolated` returns the original messages and that the parent test process is still alive and continues past the call.

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uv run`; no live provider.
- Exact command / steps: `uv run pytest tests/test_image_compression_isolation.py tests/test_image_compression_offload.py -q`, which drives the runner against a real spawned subprocess that is killed by signal, one that raises, one that times out, and one that returns normally, and also locks the handler wiring to `run_image_compression_isolated(...)`.
- Observed result: on a signal-killed child the runner returns the original message list and the parent survives; on a raising or timing-out child it fails open the same way; on a normal child the compressed messages are returned. The handler source-level regression keeps the savings log and mutation-tag path wired through the new isolation helper. Before the change, the handlers called `compressor.compress(...)` through `_run_compression_in_executor(...)`, so a native crash on that worker thread would abort the interpreter.
- Not tested: live arm64 macOS run against a real `opencv-python` 5.x KleidiCV fault.

## Review Readiness

## Checklist

## Additional Notes

- Non-goals, kept out deliberately to keep this slice shippable now: pinning `opencv-python<5` (a transitive-dependency change that only masks this one fault and does not contain the next native crash), a `HEADROOM_IMAGE_OPTIMIZE=0` env off-switch (distinct config surface; `config.image_optimize=False` already disables the feature), the per-request ONNX model reload, and the negative-savings (`preserve` logged as `-100%`) reporting bug. The last two are independent defects noted in the same report and are better fixed on their own.
- `mypy` left unchecked: not part of the focused validation for this change.


JerrettDavis (approved):
Reviewed current head c936ea06. The image compression path now has a real process boundary for native crashes, both handler call sites fail open through the isolation helper, and the focused crash/timeout/success/wiring coverage is appropriate for this slice. Local focused checks passed: uvx ruff@0.15.17 check/format for the isolation helper, handlers, and image tests; uv run --extra dev pytest tests/test_image_compression_isolation.py tests/test_image_compression_offload.py -q (7 passed). CI is still running; the current pip-audit red appears to be the shared dependency-audit state rather than this branch being behind main.

JerrettDavis (approved):
Validated current head 961bbf91 after merging latest headroomlabs/main.

Checks run:
- `git diff --check`
- `uvx ruff@0.15.17 check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py headroom/proxy/image_isolation.py tests/test_image_compression_isolation.py tests/test_image_compression_offload.py`
- `uvx ruff@0.15.17 format --check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py headroom/proxy/image_isolation.py tests/test_image_compression_isolation.py tests/test_image_compression_offload.py`
- `uv run --extra dev pytest tests/test_image_compression_isolation.py tests/test_image_compression_offload.py -q` (7 passed)

JerrettDavis (approved):
Merged current headroomlabs/main cleanly into this branch and applied a formatter-only follow-up.

Validated current head 388eb8f2 locally:
- `uv run --extra dev pytest tests/test_image_compression_isolation.py tests/test_image_compression_offload.py -q` (7 passed)
- `uv run ruff check headroom/proxy/image_isolation.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py tests/test_image_compression_isolation.py tests/test_image_compression_offload.py`
- `uv run ruff format --check headroom/proxy/image_isolation.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py tests/test_image_compression_isolation.py tests/test_image_compression_offload.py`
- `uv run --extra dev mypy headroom/proxy/image_isolation.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py`
- `git diff --check`

The earlier failing `test (1)` CI job did not reach the test suite; it failed during GitHub action download setup with repeated HTTP timeouts.

JerrettDavis (approved):
Approved current head d75cbca6 after fixing the CI-pinned Ruff formatting failure in `headroom/proxy/handlers/anthropic.py`.

Validated locally:
- `uv run --extra dev python -m pytest tests/test_image_compression_isolation.py tests/test_image_compression_offload.py -q`
- `uvx ruff@0.15.17 check .`
- `uvx ruff@0.15.17 format --check .`
- `uvx mypy@1.20.2 headroom/proxy/image_isolation.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py --ignore-missing-imports`

PR #2174, opened 2026-07-14
fix(proxy/gemini): forward a non-JSON upstream body with its real status

abhay-codes07:
## Description

`handle_gemini_generate_content` turns a non-JSON upstream error response into a generic 502, hiding the real status and body.

After the upstream call it extracts usage from `response.json()`:

```python
try:
    resp_json = response.json()
    usage = resp_json.get("usageMetadata", {})
    ...
    cache_read_tokens = usage.get("cachedContentTokenCount", 0)
except (KeyError, TypeError, AttributeError) as e:      # <-- missing JSONDecodeError / ValueError
    ...
```

`response.json()` raises `json.JSONDecodeError` (a `ValueError` subclass) for a non-JSON body. That isn't in the tuple, so it escapes to the function's outer `except Exception`, which returns a synthetic 502 and discards the real `response.status_code` / `response.content` (which the success path forwards verbatim). An overloaded Google/Vertex/Copilot frontend commonly returns a 503/500/429 with an HTML or empty body, so the client sees a generic 502 instead of the true status — defeating retry/backoff and dropping the diagnostic.

The all-non-text early-exit branch in the same handler already handles this correctly with the full tuple (`except (json.JSONDecodeError, ValueError, KeyError, TypeError, AttributeError)`) and then forwards the real status/content.

## Fix

Add `json.JSONDecodeError, ValueError` to the token-extraction `except`, matching that sibling. On a non-JSON body the extraction is skipped (token metrics keep their fallbacks) and the handler falls through to `return Response(content=response.content, status_code=response.status_code, ...)` — the real status and body.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/gemini.py`: broaden the token-extraction `except` in `handle_gemini_generate_content` to include `json.JSONDecodeError, ValueError`.
- `tests/test_gemini_nonjson_status.py`: new test asserting that except clause catches the JSON/ValueError family (guards against the regression).

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/proxy/handlers/gemini.py tests/test_gemini_nonjson_status.py
All checks passed!
$ python -m py_compile headroom/proxy/handlers/gemini.py tests/test_gemini_nonjson_status.py
OK
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17`. A full `pytest` OOM-kills this box (ML stack import), so I verified the exception handling with a dependency-free script that models the token-extraction try/except plus the handler's verbatim-forward return, and left the full pytest to CI.
- Exact command / steps: sent a 503 response whose `.json()` raises `JSONDecodeError` (non-JSON body) through the old tuple and the new tuple, plus a normal JSON 200 as a control.
- Observed result: old lets `JSONDecodeError` escape (→ the outer handler's synthetic 502); new catches it and forwards the real 503; the JSON 200 still extracts tokens under both. The new test asserts the real handler's except clause includes the JSON/ValueError family.
- Not tested: a live overloaded Gemini upstream; full local `pytest` deferred to CI (OOM).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" / "type checking" boxes are unchecked because the full suite imports the ML stack, which I can't run here; the change adds two exception types matching an existing, tested sibling branch, verified by the standalone proof and a source-level regression guard (a full handler-integration harness for Gemini doesn't exist in-tree, and the existing gemini integration tests hit a live API).


JerrettDavis:
Pushed `b11851fa` as a test-only follow-up: the regression guard now exercises `handle_gemini_generate_content()` with a fake upstream response whose `.json()` raises `JSONDecodeError`, and asserts the real 503 status/body are forwarded instead of going through `metrics.record_failed()` / synthetic 502.

Local validation:

```text
uv run --extra dev python -m pytest tests\test_gemini_nonjson_status.py tests\test_gemini_compression_offload.py -q
# 4 passed

uvx ruff@0.15.17 check headroom\proxy\handlers\gemini.py tests\test_gemini_nonjson_status.py
# All checks passed

uvx ruff@0.15.17 format --check headroom\proxy\handlers\gemini.py tests\test_gemini_nonjson_status.py
# 2 files already formatted
```

JerrettDavis (approved):
Approved after the test-only follow-up in `b11851fa`.

The handler now catches the JSON/ValueError path from `response.json()` during usage extraction and falls through to the normal verbatim upstream response return. The new test directly covers the observed failure mode: non-JSON upstream 503 stays a 503 with the original body, and the outer 502 failure path is not used. Previous CI was green; fresh CI is pending on the pushed test commit.

JerrettDavis:
Rebased this onto current `main` to clear the merge conflict; no patch behavior changed.

Validated after rebase:
- `python -m pytest tests\test_gemini_nonjson_status.py -q`
- `uvx ruff@0.15.17 check headroom\proxy\handlers\gemini.py tests\test_gemini_nonjson_status.py`

PR #2188, opened 2026-07-14
fix(memory): require explicit updates for supersession

### LINKED ISSUE #2187: [BUG] memory_save auto-supersedes distinct facts based only on vector similarity
## Description

The standalone Memory MCP `memory_save` handler treats vector similarity as update identity. If any active search result scores at least `0.70`, it calls `update_memory` and supersedes that result even when the two strings describe different facts.

This can silently remove a valid memory when several atomic facts share the same product or configuration vocabulary.

## To Reproduce

1. Save an atomic fact such as `Use compact output by default`.
2. Save a different fact in the same domain, such as `Keep semantic compression disabled`.
3. Use an embedding/search backend that scores the first fact at or above `0.70` for the second save.
4. Observe that `memory_save` calls `update_memory` for the first memory instead of saving a second row.

## Expected Behavior

`memory_save` should preserve both facts. A supersession chain should only be created by an explicit update operation that identifies the existing memory ID.

## Actual Behavior

The first sufficiently similar active search result is automatically superseded without checking a stable fact key, category, entity identity, or explicit update intent.

## Impact

This is silent data loss from the active memory view: a valid fact can disappear merely because another fact uses related vocabulary. Raising the similarity threshold reduces frequency but does not establish update identity.

A Draft PR with a focused regression test will follow immediately.

superdiaodiao:
## Description

The standalone Memory MCP `memory_save` handler currently treats vector similarity as update identity. A score of `0.70` can therefore supersede a valid but distinct memory that merely shares domain vocabulary.

This change makes `memory_save` append-only. Supersession remains available through explicit update paths that receive an existing memory ID.

Closes #2187.

## Type of Change

- [ ] Breaking change (fix or feature that causes existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Remove vector-similarity-based auto-supersession from the standalone MCP `memory_save` handler.
- Clarify in the tool description that corrections require an explicit update path with the existing memory ID.
- Add a regression test proving that a high-scoring but distinct memory is neither searched for replacement nor updated.
- Preserve the existing save result summary shape for compatibility.

## Testing

- [x] Focused unit tests pass
- [x] Manual focused test execution performed

### Test Output

```text
uv run --with pytest --with numpy pytest tests/test_memory/test_mcp_server.py -q
9 passed, 21 warnings in 0.70s

uvx ruff check headroom/memory/mcp_server.py tests/test_memory/test_mcp_server.py
All checks passed!

uvx ruff format --check headroom/memory/mcp_server.py tests/test_memory/test_mcp_server.py
2 files already formatted
```

The warnings are pre-existing pytest configuration and `datetime.utcnow()` deprecation warnings in the test environment.

## Real Behavior Proof

- Environment: Python 3.13 with the MCP module stub and an async recording backend.
- Exact command / steps: run `tests/test_memory/test_mcp_server.py`; the new regression supplies a search result with similarity `0.91`, then saves a distinct fact.
- Observed result: `search_memories` and `update_memory` are not called; `save_memory` is called once with the new fact and requested importance.
- Not tested: live embedding backends or migration of supersession chains created by earlier versions.

## Review Readiness

## Checklist

- [x] My code follows the project style guidelines
- [x] I have commented the non-obvious identity boundary
- [ ] Documentation changes are limited to the MCP tool description
- [x] I have added tests that prove the fix is effective
- [ ] New and existing unit tests pass locally; the focused MCP suite passes and full CI is pending
- [ ] CHANGELOG update is not included because release notes are generated from conventional commits

Not applicable.

## Additional Notes

This patch intentionally does not infer replacement identity from category, entity references, or a higher vector threshold: none of those alone proves that two statements are versions of the same fact. Exposing an explicit update tool from the standalone MCP server can be considered separately without retaining the unsafe automatic behavior.


JerrettDavis (approved):
Reviewed current head after merging main. The change removes the unsafe similarity-as-identity path from standalone memory_save, keeps the save result shape compatible, and the regression covers the key boundary by proving a high-score distinct memory is saved rather than searched/updated. Focused validation passed locally: tests/test_memory/test_mcp_server.py, ruff check, and ruff format --check. Fresh CI is still running after the main merge.

PR #2160, opened 2026-07-14
fix(codex): preserve wrapped sessions and recover state

rudironsoni:
## Description

Closes #2159.

Codex wrappers currently launch against a disposable `CODEX_HOME`, so session state created during a wrapped run can disappear when that temporary directory is removed. This change launches Codex against its durable home, keeps proxy routing process-local, and adds recovery for retained temporary homes and pinned recovery sources.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Launch Codex against its durable `CODEX_HOME` and apply routing through process-local config overrides after the actual proxy port is resolved.
- Preserve custom provider identity and reject providers that cannot be redirected safely.
- Detect dangling temporary Codex homes before interactive wraps and offer recovery.
- Add `headroom recover codex` with automatic discovery, repeatable `--source`, preview, confirmation, retained backups, and rollback on failure.
- Search Python's temp root, `$TMPDIR`, `/tmp`, `/private/tmp`, and macOS `/private/var/folders/*/*/T` for retained `headroom-codex-home-*` directories.
- Reuse `source-pinned/` copies left by interrupted or failed recovery attempts after the original temporary home has disappeared.
- Report deleted temporary homes still referenced by SQLite rollout paths without treating paths pasted into prompts or errors as filesystem evidence.
- Audit the durable thread index, rollout files, and history when no source remains, including indexed chat counts and history-only orphan records.
- Normalize legacy localhost `headroom` providers in both SQLite thread rows and rollout `session_meta`, including retries after an earlier broken recovery, while preserving user-defined remote providers named `headroom`.
- Merge compatible config, JSONL, rollout, SQLite, credential, and regular-file state without propagating deletions or runtime artifacts.
- Rewrite recovered thread rollout paths to the durable home and restore legacy Headroom thread providers to the active provider.
- Validate SQLite schemas, SQLx migration checksums, integrity, and foreign keys, and quarantine malformed JSONL.
- Preserve failed targets with an atomic rename before rollback, avoiding recursive-deletion races with live SQLite runtime files.
- Document discovery, migration, retained backups, rollback behavior, and the limits of deleted-source recovery.

## Testing

- [x] Unit tests pass (`pytest`)
[...]

PR #2191, opened 2026-07-14
fix(ci): retry apt installs in wheel smoke gate

JerrettDavis:
## Summary
- add a small retry helper around Ubuntu-container `apt-get` operations in the Release wheel smoke-import gate
- use `--fix-missing` for the package installs that can fail when the ARM Ubuntu ports mirror serves stale package indexes
- add a workflow regression test so the smoke gate keeps the retry/fix-missing hardening

## Why
A Release smoke-import job on PR #2160 failed twice in the Ubuntu 22.04 ARM container while installing Python 3.12 dependencies:

`E: Failed to fetch http://ports.ubuntu.com/ubuntu-ports/pool/main/p/python-httplib2/python3-httplib2_0.20.2-2ubuntu0.1_all.deb 404 Not Found`

The smoke gate should still fail for broken wheels, but transient apt mirror skew should not make unrelated PRs red.

## Validation
- `python -m pytest tests/test_release_workflows.py -q` -> 32 passed
- `uvx ruff@0.15.17 check tests/test_release_workflows.py` -> All checks passed
- `git diff --check`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -608,18 +608,30 @@ jobs:
                 fi
               elif command -v apt-get >/dev/null 2>&1; then
                 export DEBIAN_FRONTEND=noninteractive
-                apt-get update -qq
+                apt_retry() {
+                  apt_get_args="$*"
+                  for attempt in 1 2 3; do
+                    if apt-get "$@"; then
+                      return 0
+                    fi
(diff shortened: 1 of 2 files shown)


JerrettDavis:
Closing this as superseded by #2190. I folded the Release smoke apt retry hardening into #2190 so the dependency-audit fix and smoke-gate fix can go green together instead of leaving two PRs blocked by each other's baseline failures.

PR #2202, opened 2026-07-15
fix(code-compressor): recover valid Python rewrites after local syntax rejection

### LINKED ISSUE #1233: AST code compression emits invalid syntax on ~28% of real files and self-rejects to 0%

rodboev:
## Description

A compile-invalid Python definition rewrite currently makes `CodeAwareCompressor` discard every otherwise valid rewrite in the file and return the original source at 0 percent reduction. The existing whole-file safety guard stays in place, while a Python-only recovery replay now preserves the rejected definition and keeps independent valid compression.

The recovery reuses the current Python validation authority in `ast.parse()` plus `compile()`, runs only after the first assembled module already fails `_verify_syntax()`, and stays out of non-Python paths.

Closes #1233

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added a Python-only recovery replay after the first assembled module fails syntax validation.
- Preserved only the invalid function or class rewrite while allowing independent valid definitions to remain compressed.
- Kept the existing whole-file syntax guard and original-source fallback as the terminal safety check.
- Added focused invalid-node, valid-modern-syntax, and fail-safe coverage.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_transforms/test_code_compressor.py::TestRealASTRuns::test_python_invalid_node_falls_back_locally tests/test_transforms/test_code_compressor.py::TestRealASTRuns::test_python_recovery_does_not_block_valid_modern_syntax tests/test_transforms/test_code_compressor.py::TestRealASTRuns::test_python_recovery_still_returns_original_when_all_candidates_invalid tests/test_transforms/test_code_compressor.py::TestEdgeCases::test_syntax_errors_in_input tests/test_transforms/test_code_compressor.py::TestRealASTRuns::test_ast_preserves_structure_for_python -v`)
- [x] Linting passes (`uv run ruff check headroom/transforms/code_compressor.py tests/test_transforms/test_code_compressor.py`)

### Test Output

```text
uv run pytest tests/test_transforms/test_code_compressor.py::TestRealASTRuns::test_python_invalid_node_falls_back_locally tests/test_transforms/test_code_compressor.py::TestRealASTRuns::test_python_recovery_does_not_block_valid_modern_syntax tests/test_transforms/test_code_compressor.py::TestRealASTRuns::test_python_recovery_still_returns_original_when_all_candidates_invalid tests/test_transforms/test_code_compressor.py::TestEdgeCases::test_syntax_errors_in_input tests/test_transforms/test_code_compressor.py::TestRealASTRuns::test_ast_preserves_structure_for_python -v
5 passed in 0.34s
uv run ruff check headroom/transforms/code_compressor.py tests/test_transforms/test_code_compressor.py
All checks passed!
uv run ruff format headroom/transforms/code_compressor.py tests/test_transforms/test_code_compressor.py --check
2 files already formatted
```

## Real Behavior Proof

- Environment: Windows, synced `uv` environment with `dev` and `code` extras installed
- Exact command / steps: run the focused invalid-node regression through public `compress(..., language="python")`
- Observed result: `1 passed in 0.19s`; the invalid candidate stays original, the neighboring valid candidate remains compressed, and `headroom-PR-TARGET-1233-PROOF.md` records the base `ratio=1.0` whole-file rollback against the fixed head behavior.
- Not tested: the stale future-import mismatch discussed in the old issue comment, already covered on current main

## Review Readiness

## Checklist

## Additional Notes

- The stale future-import comment on #1233 is not the live slice here; current main already validates Python with `compile()` and already covers that ordering case.
- This fix keeps the existing whole-file fail-safe and does not broaden into cross-language recovery or new syntax models.
- `CHANGELOG.md` remains unchanged because Headroom generates release notes from conventional commits.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -1385,6 +1385,7 @@ class AnthropicHandlerMixin:
                                             else "deferred:dropped"
                                         ]
                                         timing = {}
+                                        waste_signals = None
 
                                     result = _DeferredCompressionResult()
                             else:
--- a/headroom/transforms/code_compressor.py
+++ b/headroom/transforms/code_compressor.py
@@ -45,6 +45,7 @@ from __future__ import annotations
 import logging
 import re
 import threading
+from collections.abc import Callable
 from dataclasses import dataclass, field
 from enum import Enum
 from typing import Any
@@ -1168,6 +1169,20 @@ class CodeAwareCompressor(Transform):
 
             # Verify syntax validity (checks both ERROR and MISSING nodes)
             syntax_valid = self._verify_syntax(compressed, detected_lang)
+            original_syntax_valid = True
+
+            if not syntax_valid and detected_lang == CodeLanguage.PYTHON:
+                original_syntax_valid = self._verify_syntax(code, detected_lang)
+                if original_syntax_valid:
+                    compressed, structure, symbol_scores = self._compress_with_ast(
+                        code,
+                        detected_lang,
+                        context,
+                        tokenizer,
+                        recover_invalid_python_nodes=True,
+                    )
+                    compressed_tokens = self._estimate_tokens(compressed, tokenizer)
+                    syntax_valid = self._verify_syntax(compressed, detected_lang)
 
             # If syntax invalid, return original (never serve broken code)
             if not syntax_valid:
@@ -1266,6 +1281,7 @@ class CodeAwareCompressor(Transform):
         language: CodeLanguage,
         context: str,
         tokenizer: Tokenizer | None = None,
+        recover_invalid_python_nodes: bool = False,
     ) -> tuple[str, CodeStructure, dict[str, float]]:
         """Compress code using AST parsing with symbol importance analysis.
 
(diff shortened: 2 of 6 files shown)


JerrettDavis (approved):
Reviewed the current head. The retry path is scoped to Python only, first verifies the original file parses, and then validates each candidate by splicing it back into the original module before accepting it. That preserves the terminal whole-file fail-safe while recovering useful compression when only one rewritten definition is invalid.

I also checked the focused regression tests added for local node fallback, valid modern syntax, and all-invalid fallback behavior. On this Windows checkout those tests skip because the tree-sitter grammar pack is not installed, so I could not use local execution as proof; current GitHub smoke/lint/security checks are green so far, with a few checks still pending.

PR #2204, opened 2026-07-15
fix(code): quarantine Perl parser from code-aware compression

### LINKED ISSUE #2185: [BUG] Proxy wedges at 100% CPU: tree-sitter Perl external scanner infinite loop holds the GIL, asyncio stops accepting connections
## Summary

The optimization proxy (`headroom proxy`, port 8787) periodically enters a state where it **stops accepting TCP connections** while pegged at ~100% CPU.
Every Claude Code request routed through it hangs; the Anthropic SDK burns its full 10-retry budget and the query appears stuck.
The listening socket stays in `LISTEN` but never `accept()`s. It stays wedged indefinitely (observed 7h+). `kill -9` on the proxy fixes it — the supervisor (`headroom install agent run`) respawns a healthy proxy in ~5s.

## Environment

- headroom 0.31.0
- Python 3.11.15
- macOS 26.5.2, arm64
- proxy launched as: `headroom proxy --host 127.0.0.1 --port 8787 --mode token --backend anthropic --no-telemetry`

## Root cause (from a `sample` of the wedged process)

The wedge is a **GIL starvation** caused by a native tree-sitter scan that never yields:

- **Main thread** (asyncio event loop) is blocked trying to re-acquire the GIL after its kqueue syscall:
  ```
  _PyEval_EvalFrameDefault → select_kqueue_control → PyEval_RestoreThread → take_gil → __psynch_cvwait
  ```
  i.e. it made it out to `select()` but can never get back **into** Python to service results — so it never runs the `accept()` callback.

- **A worker thread** is spinning in the **tree-sitter Perl external scanner**, actively looping in native C (caught at varying PCs across samples):
  ```
  parser_parse → ts_parser_parse → tree_sitter_perl_external_scanner_scan
      → ts_lexer__advance / ts_lexer__get_column   (repeating)
  ```
  This is a native call that **does not release the GIL**, so it holds the GIL for the entire duration of the loop, starving the event loop.

Supporting evidence:
- `ps` on the wedged proxy: **5h30m of accumulated CPU time** over 7h27m elapsed (`TIME 329:56`, `STAT R`, `%CPU 99`).
- `nc -z 127.0.0.1 8787` fails (no accept); raw TCP connect never completes.
- A `POST /v1/messages` through the proxy times out with 0 bytes; the **same request directly to `api.anthropic.com` returns in ~0.38s**, so upstream/network is healthy — the proxy process is the sole cause.
- `lsof` on the wedged proxy shows a leaked half-closed socket (`127.0.0.1:8787->...:50183 (CLOSED)`) still open — a connection that dropped while the loop was frozen (symptom, not cause).

## *Likely* trigger

The proxy parses code (to compress context) with tree-sitter and appears to invoke the **Perl** grammar's external scanner on some input. Some input drives that scanner into a non-terminating loop. Two things seem to raise the odds of hitting it:
- A client connection dropping mid-request (flaky network) — leaves work/state that seems to precede the wedge.
- Running **multiple concurrent Claude agents** through the proxy.

Note: the affected project is TypeScript/C#, not Perl — so the Perl grammar may be getting selected by mis-detection, or run as a fallback, on non-Perl input. Worth confirming why the Perl scanner runs at all here.

## Impact

- Total proxy outage until manually killed; presents to the user as "API connection problem, 10/10 retries failing, queries stuck."
- Recurs (daily, intermittently).

## Secondary issue: orphaned `headroom mcp serve` pileup

Across the session, orphaned `headroom mcp serve` processes accumulate (observed 3 simultaneously), reparented to launchd (ppid 1) after their parent Claude process exits. They are not reaped. (Same symptom noted in #1761.)

## Workaround

```bash
kill -9 "$(pgrep -f 'headroom\.cli proxy')"   # supervisor respawns a fresh proxy in ~5s
```

## LLM Suggested fixes

1. Run tree-sitter parsing with the **GIL released** (or in a worker **subprocess**), and/or wrap scans in a **hard timeout / cancellation** so a pathological scan can't freeze the event loop.
2. Investigate the **Perl external scanner** non-termination (known class of tree-sitter scanner bug); bound scan time/steps.
3. Confirm/fix **language selection** so the Perl grammar isn't run on non-Perl input.
4. Reap orphaned `mcp serve` children on parent exit / on shutdown.

## Related

- #1761 (proxy hang + orphaned `mcp serve` pileup)
- #638 (proxy stops working)

rodboev:
## Description

`headroom proxy` can wedge when code-aware compression enters the tree-sitter Perl external scanner and the native scan keeps the GIL indefinitely. Current main still has two routes into that scanner: explicit `perl` or `pl` hints flow through `CodeAwareCompressor`, and `detect_language()` can nominate Perl on non-Perl code because its prefilter matches generic sigils such as decorators, JSDoc tags, and shell variables before phase 2 parses every surviving candidate grammar.

This change quarantines Perl at the code-aware compression funnel without widening scope. Perl remains recognized at the input boundary, but live proxy compression no longer requests a Perl parser. Non-Perl code keeps its existing code-aware behavior. Real Perl falls back through the existing safe Kompress or passthrough contract instead of entering tree-sitter. The diff stays inside `headroom/transforms/code_compressor.py` plus focused parser-safety regressions.

Refs #2185

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added a Perl quarantine list in `headroom/transforms/code_compressor.py` and used it to stop Perl candidate parsing during language detection.
- Added a hard `_get_parser()` guard so no live code-aware path can construct a Perl parser.
- Routed resolved explicit Perl hints through the existing safe fallback or passthrough contract before AST compression.
- Added focused parser-safety regressions for non-Perl candidate bleed, explicit `perl` and `pl`, inferred real Perl, mixed fenced Perl content, fallback-disabled passthrough, and non-Perl negative space.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_perl_scanner_safety.py -q`)
- [x] Linting passes (`uv run ruff check headroom/transforms/code_compressor.py tests/test_perl_scanner_safety.py`)

### Test Output

```text
uv run pytest tests/test_perl_scanner_safety.py -q
8 passed in 1.93s
uv run ruff check headroom/transforms/code_compressor.py tests/test_perl_scanner_safety.py
All checks passed!
uv run ruff format headroom/transforms/code_compressor.py tests/test_perl_scanner_safety.py --check
2 files already formatted
```

## Real Behavior Proof

- Environment: Windows, Python from the synced `uv` environment, `dev` and `code` extras installed, no provider call
- Exact command / steps: Run `uv run pytest tests/test_perl_scanner_safety.py -q`.
- Observed result: `8 passed in 1.93s`; the suite proves explicit `perl` and `pl`, inferred real Perl, mixed fenced Perl, and non-Perl negative-space routes all avoid Perl parser entry.
- Not tested: the reporter's macOS payload and long-running concurrent workload

## Review Readiness

## Checklist

## Additional Notes

- Use `Refs #2185`, not `Closes #2185`. The wedge surface is this PR's scope, but #2185 also carries a separate orphaned `headroom mcp serve` report that this slice does not address.
- This is a reachability fix. It does not repair the upstream Perl scanner and it does not harden other grammars against the same class of native wedge.
- `CHANGELOG.md` remains unchanged because Headroom generates release notes from conventional commits.
- Merged PR https://github.com/headroomlabs-ai/headroom/pull/2114 addresses a different cooperative compression stall and stays separate from this native parser-entry slice.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -1385,6 +1385,7 @@ class AnthropicHandlerMixin:
                                             else "deferred:dropped"
                                         ]
                                         timing = {}
+                                        waste_signals = None
 
                                     result = _DeferredCompressionResult()
                             else:
--- a/headroom/transforms/code_compressor.py
+++ b/headroom/transforms/code_compressor.py
@@ -102,6 +102,9 @@ def _tree_sitter_importable() -> bool:
         return False
 
 
+_UNSAFE_TREE_SITTER_LANGUAGES: frozenset[str] = frozenset({"perl"})
+
+
 def _get_parser(language: str) -> Any:
     """Get a tree-sitter parser for the given language.
 
@@ -129,6 +132,8 @@ def _get_parser(language: str) -> Any:
         ImportError: If tree-sitter is not installed.
         ValueError: If language is not supported.
     """
+    if language in _UNSAFE_TREE_SITTER_LANGUAGES:
+        raise ValueError(f"Language '{language}' is quarantined for code-aware compression.")
     # NOTE: guard on importability (not _check_tree_sitter_available), because
     # _check_tree_sitter_available now performs a real end-to-end parse via
     # _get_parser; guarding on it here would recurse.
@@ -161,7 +166,7 @@ def _get_parser(language: str) -> Any:
         except Exception as e:
             raise ValueError(
                 f"Language '{language}' is not supported by tree-sitter. "
-                f"Supported: python, javascript, typescript, go, rust, java, c, cpp, csharp, perl. "
+                f"Supported: python, javascript, typescript, go, rust, java, c, cpp, csharp. "
                 f"Error: {e}"
             ) from e
 
@@ -714,6 +719,16 @@ def detect_language(code: str) -> tuple[CodeLanguage, float]:
         if candidates[CodeLanguage.CPP] >= 2:
             candidates[CodeLanguage.C] = 0
 
+    perl_score = candidates.get(CodeLanguage.PERL, 0)
+    if perl_score > 0:
+        best_non_perl = max(
+            (score for lang, score in candidates.items() if lang != CodeLanguage.PERL),
+            default=0,
+        )
+        if perl_score > best_non_perl:
+            return CodeLanguage.UNKNOWN, 0.0
+        candidates.pop(CodeLanguage.PERL, None)
+
     # Phase 2: If tree-sitter available, parse with candidates and pick fewest errors
     if _check_tree_sitter_available():
         best_lang = CodeLanguage.UNKNOWN
@@ -1142,6 +1157,19 @@ class CodeAwareCompressor(Transform):
                     language_confidence=0.0,
                     syntax_valid=True,
                 )
+        if detected_lang == CodeLanguage.PERL:
+            if self.config.fallback_to_kompress:
+                return self._fallback_compress(code, original_tokens)
+            return CodeCompressionResult(
+                compressed=code,
+                original=code,
+                original_tokens=original_tokens,
+                compressed_tokens=original_tokens,
+                compression_ratio=1.0,
+                language=detected_lang,
+                language_confidence=confidence,
+                syntax_valid=True,
+            )
 
         # Check if tree-sitter is available
         if not _check_tree_sitter_available():
--- a/server.json
+++ b/server.json
@@ -9,13 +9,13 @@
     "source": "github",
     "id": "1129940957"
   },
-  "version": "0.27.0",
+  "version": "0.31.0",
   "packages": [
     {
       "registryType": "pypi",
       "registryBaseUrl": "https://pypi.org",
       "identifier": "headroom-ai",
-      "version": "0.27.0",
+      "version": "0.31.0",
       "runtimeHint": "uvx",
       "runtimeArguments": [
         {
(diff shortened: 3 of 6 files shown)


JerrettDavis (changes requested):
Thanks for quarantining the Perl parser path. The production behavior looks directionally right, but one of the new tests is not portable across supported dev environments.

`tests/test_perl_scanner_safety.py::test_typescript_with_decorators_stays_typescript_without_perl` fails when the tree-sitter grammar pack is not installed. `_tree_sitter_importable()` is false, but `compress()` still performs the Python parser availability probe through `_check_tree_sitter_available()`, so the spy sees `calls == ["python"]` rather than `[]`.

Local reproduction:

`uv run pytest tests/test_perl_scanner_safety.py -q`

Result: `1 failed, 7 passed`; the failing assertion is `assert calls == []` with actual `['python']`.

Please adjust the assertion to allow the availability probe while still enforcing the important invariant: Perl must not be requested as a parser candidate. For example, the non-installed branch can assert `"perl" not in calls` and avoid assuming no parser probes happen at all.

JerrettDavis (approved):
Re-reviewed the updated head after the Perl scanner safety test fix. The previous portability blocker is resolved: the test now allows the normal tree-sitter availability probe while still enforcing the important invariant that Perl is never requested from the parser.

Local validation:

`uv run pytest tests/test_perl_scanner_safety.py -q`

Result: `8 passed`.

PR #2207, opened 2026-07-15
security: address u9up assessment findings (WEB-01–07)

chopratejas:
## Description

Addresses the **u9up security assessment** of Headroom v0.29.0 (2026-07-09): 1 High, 5 Medium, 1 Low. All seven findings were re-verified as *still present on current `main`* (line numbers had drifted from the report; none were closed by prior PRs #1226/#1338) and then fixed. Each commit maps to one finding.

The aggravating factor across the report — the reference `docker-compose.yml` binding `0.0.0.0` with no token — is what escalated most findings from "localhost-only" to "remotely exploitable". Fixing that (WEB-06) de-fangs the cluster; the code-level fixes are defense-in-depth.

Security disclosure (private, security@headroomlabs.ai) — no public issue to close.

## Type of Change

- [x] Bug fix (non-breaking security hardening)
- [x] Code refactoring (WEB-02 consolidates 6 duplicated header reads into one seam)

## Changes Made

| ID | Sev | Fix |
|----|-----|-----|
| WEB-01 | High | `upstream_guard.is_safe_upstream_url()` blocks metadata/RFC1918/loopback/link-local at every client-supplied-base entry point (openai resolver + 2 passthrough routes); opt-in `HEADROOM_ALLOWED_BASE_URLS` allowlist. Public-host BYOK preserved. |
| WEB-02 | Med | One `resolve_memory_identity()` seam replaces 6 duplicated `x-headroom-user-id` reads: header honored only from loopback / allowlist, else bound to the proxy-token fingerprint (or OS user). `set_identity_resolver()` is the enterprise multi-tenant hook. |
| WEB-03 | Med | All 5 fetched tool binaries (difft, scc, rtk, lean-ctx, codebase-memory-mcp) verify a real SHA-256 pin before execution (fail-closed; opt-out `HEADROOM_BINARIES_ALLOW_UNVERIFIED=1`). https-only mirror. `scripts/refresh_tool_hashes.py` + `tools-hash-refresh` CI keep pins fresh. |
| WEB-04 / WEB-05 | Med | `Depends(_require_loopback)` on `POST /v1/telemetry/import` + 15 operational GET endpoints. |
| WEB-06 | Med | Reference compose: loopback-published ports, required `HEADROOM_PROXY_TOKEN` + `NEO4J_AUTH`, APOC file I/O removed; hardcoded `neo4j_password="password"` defaults removed (env-resolved). |
| WEB-07 | Low | `filter="data"` on all 4 `tar.extract()` calls. |

## Testing

- [x] Linting passes (`ruff check`) — clean on all changed files
- [x] Type checking passes (`mypy`) — clean on all changed files
- [x] New tests added (`tests/test_upstream_guard.py`, `tests/test_identity_resolution.py`)
- [x] Manual testing performed (inline behavioral checks + live asset fetch/verify)
- [ ] Unit tests pass (`pytest`) — ran the 2 new files (below) + per-commit lint/type; the **full** suite is left to CI.

### Test Output

```text
$ ruff check <changed files>        → All checks passed!
$ mypy  <changed files>             → Success: no issues found

$ pytest tests/test_upstream_guard.py tests/test_identity_resolution.py -q
20 passed in 0.50s

$ python scripts/refresh_tool_hashes.py --check
all tool pins up to date       # CI gate re-fetches all 30 assets; digests match
```

## Real Behavior Proof

- **Environment:** macOS (darwin), Python 3.12.6, project `.venv`; branch cut from `upstream/main` (in sync, +7 / -0).
- **Exact commands / steps:**
  - SSRF: `is_safe_upstream_url()` returns `False` for `169.254.169.254`, `10.0.0.1`, `localhost`, non-http schemes; `True` for public IPs; allowlist mode enforced.
  - Identity: loopback caller → header honored; non-loopback caller → header ignored, identity bound to `tok_<sha>`.
  - Binary pinning: fetched a real release asset — `verify_download_bytes` accepts correct bytes, rejects tampered bytes and unpinned URLs.
  - Docker: `docker compose config` with no env → fails on required `HEADROOM_PROXY_TOKEN` / `NEO4J_AUTH` (fail-closed proven).
- **Observed result:** all as expected (see Test Output).
- **Not tested:** full `pytest` suite; a live end-to-end proxy request to a real upstream; the fail-closed path exercised through the real `headroom` CLI (covered at unit level via `verify_download_bytes`).

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (env vars documented inline + in `.env.example`; no doc-site changes)
- [x] New and existing unit tests pass locally with my changes (targeted — see note)
- [ ] I have updated the CHANGELOG.md if applicable (left for the repo's changeset/release flow)

## Additional Notes

- **WEB-02 open-core boundary:** OSS ships a secure single-user resolver; real multi-tenant identity/RBAC plugs in via `set_identity_resolver()` (headroom-managed), so the enterprise differentiator isn't given away.
- **`/metrics`** is now loopback-gated (WEB-05) — remote Prometheus scraping needs a loopback-colocated exporter/sidecar or a future token-gated variant.
- **New env vars:** `HEADROOM_ALLOWED_BASE_URLS`, `HEADROOM_USER_ID_ALLOWLIST`, `HEADROOM_BINARIES_ALLOW_UNVERIFIED`, `HEADROOM_BIND_ADDR`.


JerrettDavis (changes requested):
This hardening pass is moving in the right direction, but I found two places where the new security controls do not enforce the policy described in the PR body.

1. `HEADROOM_USER_ID_ALLOWLIST` currently allows any remote caller to claim an allowlisted identity. In `resolve_memory_identity`, lines 91-93 compare the untrusted `x-headroom-user-id` value against the allowlist and return it even when `request.client.host` is non-loopback. That means if an operator configures `HEADROOM_USER_ID_ALLOWLIST=alice,bob`, any network client can send `x-headroom-user-id: alice` and read/write Alice's memory partition. This preserves the cross-user selection issue for the highest-value IDs instead of binding non-loopback clients to token identity. The allowlist needs to authenticate/authorize the caller side (for example trusted source hosts or a custom resolver), not just the requested user id, or it should be removed from the default resolver.

2. `HEADROOM_ALLOWED_BASE_URLS` accepts URLs with ports, but the guard stores only hostnames. `_allowlisted_hosts()` drops scheme and port at line 43, and `is_safe_upstream_url()` checks only `host.lower() in allow`. As a result, an operator allowlisting `https://llm.corp:8443` also allows `https://llm.corp:22`, `http://llm.corp:8080`, etc. For an SSRF guard, that is materially broader than the configured base URL. Please preserve and compare the normalized origin/base URL tuple, or document and test that this is intentionally host-wide.

JerrettDavis:
I rechecked the latest head (`b6688701`) after the update. The two blocking issues from the prior review still appear to be present:

- `HEADROOM_USER_ID_ALLOWLIST` still matches the untrusted requested identity (`header_value in allow`) for non-loopback callers, so any remote client can claim an allowlisted user id. The test `test_allowlist_permits_named` also codifies that behavior.
- `HEADROOM_ALLOWED_BASE_URLS` still normalizes configured URLs down to hostnames in `_allowlisted_hosts()`, and `is_safe_upstream_url()` still checks only `host.lower() in allow`. The test allows `https://llm.corp:8443`, but there is no negative coverage for `https://llm.corp:22` or `http://llm.corp:8080`.

Leaving the existing changes-requested state and labels in place until those are tightened or explicitly re-scoped with matching documentation/tests.

JerrettDavis (changes requested):
Thanks for the update. I rechecked current head `69a4ab13`; the focused tests and ruff pass, but the two security blockers from the earlier review are still present in the policy itself:

- `headroom/proxy/identity.py:91-93` still authorizes a non-loopback caller by checking whether the untrusted requested `x-headroom-user-id` is in `HEADROOM_USER_ID_ALLOWLIST`, then returns that value. That still lets any remote client claim an allowlisted tenant/user id. The allowlist needs to be tied to an authenticated caller/source (or removed in favor of the custom resolver hook), not to the caller-supplied identity string. `tests/test_identity_resolution.py:44-48` currently codifies the unsafe case as expected behavior, so that test needs to invert once the policy is fixed.

- `headroom/proxy/upstream_guard.py:32-43` parses `HEADROOM_ALLOWED_BASE_URLS` down to hostnames only, and `is_safe_upstream_url()` then accepts any safe scheme/port/path for that host at `headroom/proxy/upstream_guard.py:76-78`. If an operator allowlists `https://llm.corp:8443`, `http://llm.corp:80` would still pass. For URL entries, preserve and compare the normalized origin/scheme+host+port rather than collapsing to bare host; only bare-host entries should intentionally mean host-wide allowlisting. `tests/test_upstream_guard.py:37-42` should include the cross-scheme/port negative case.

Validation run:

```text
uv run pytest tests/test_identity_resolution.py tests/test_upstream_guard.py -q
20 passed in 0.56s

uvx ruff==0.15.17 check headroom/proxy/identity.py headroom/proxy/upstream_guard.py tests/test_identity_resolution.py tests/test_upstream_guard.py --output-format concise
All checks passed!
```

JerrettDavis (changes requested):
Thanks for cleaning up the branch, but the two security blockers from the previous reviews are still present in the current head (`78080286`).

1. `HEADROOM_USER_ID_ALLOWLIST` still authorizes the caller-supplied identity string rather than the caller. In `headroom/proxy/identity.py`, the non-loopback path still does:

```python
allow = _allowlist()
if allow is not None and header_value in allow:
    return header_value
```

That means any remote client can send `x-headroom-user-id: alice` and access Alice's memory partition whenever `alice` is in `HEADROOM_USER_ID_ALLOWLIST`. The allowlist needs to be bound to authenticated caller/source identity, or this default resolver should ignore remote `x-headroom-user-id` entirely and leave multi-tenant identity selection to the custom resolver hook. `tests/test_identity_resolution.py::test_allowlist_permits_named` still codifies the unsafe remote-claim behavior and should be inverted/removed.

2. `HEADROOM_ALLOWED_BASE_URLS` still drops scheme and port. `headroom/proxy/upstream_guard.py` stores only `(parsed.hostname or item).lower()`, and `is_safe_upstream_url()` checks only `host.lower() in allow`. An operator allowlisting `https://llm.corp:8443` therefore still also allows `http://llm.corp:80`, `https://llm.corp:22`, etc. For full URL entries, please preserve and compare the normalized origin/base tuple; only explicitly bare host entries should mean host-wide allowlisting. `tests/test_upstream_guard.py::test_allowlist_mode` still lacks the cross-scheme/port negative cases.

I did not run the broader suite because these are policy-level blockers visible in the current source and tests.

headroom/proxy/identity.py
```diff
@@ -0,0 +1,101 @@
...
+
+    host = _client_host(request)
+    is_local = host is None or is_loopback_host(host)
+
+    if header_value is not None:
+        if is_local:
+            return header_value
+        allow = _allowlist()
+        if allow is not None and header_value in allow:
+            return header_value
```
Grg0rry, line 93: The allowlist authorizes the caller-supplied **string**, so any non-loopback client sending `x-headroom-user-id: alice` reaches Alice's partition whenever `alice` is configured. The module docstring says the header is honored for "hosts listed in `HEADROOM_USER_ID_ALLOWLIST`" — i.e. trusted **source hosts**, not trusted user-ids. This binds the claim to the source host instead.

```suggestion
        trusted = _allowlist()
        if trusted is not None and host is not None and host.lower() in trusted:
            return header_value
        # Caller from a trusted source host may assert an identity; any other
        # network caller falls through to its own token-bound scope.
```

headroom/proxy/identity.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def _allowlist() -> set[str] | None:
+    raw = os.environ.get(ALLOWLIST_ENV)
+    if not raw or not raw.strip():
+        return None
+    return {value.strip() for value in raw.split(",") if value.strip()}
+
+
+def _token_identity() -> str | None:
```
Grg0rry, line 58: Paired with the above: the env var now holds trusted source hosts. Keeping the name `HEADROOM_USER_ID_ALLOWLIST` to avoid a breaking config change; rename to `HEADROOM_TRUSTED_SOURCE_HOSTS` if you'd rather make the intent explicit.

```suggestion
def _allowlist() -> set[str] | None:
    # Trusted source hosts whose callers may assert an x-headroom-user-id.
    raw = os.environ.get(ALLOWLIST_ENV)
    if not raw or not raw.strip():
        return None
    return {value.strip().lower() for value in raw.split(",") if value.strip()}
```

headroom/proxy/upstream_guard.py
```diff
@@ -0,0 +1,88 @@
...
+    hosts: set[str] = set()
+    for item in raw.split(","):
+        item = item.strip()
+        if not item:
+            continue
+        # Accept either a full URL or a bare host[:port].
+        parsed = urlparse(item if "//" in item else f"//{item}")
+        hosts.add((parsed.hostname or item).lower())
+    return hosts
+
```
Grg0rry, line 45: `HEADROOM_ALLOWED_BASE_URLS` collapses every entry to a bare hostname, so allowlisting `https://llm.corp:8443` also permits `http://llm.corp:80` / `https://llm.corp:22`. This keeps bare-host entries host-wide (explicit opt-in) while matching full-URL entries on the normalized origin `(scheme, host, port)`.

```suggestion
    hosts: set[str] = set()
    origins: set[tuple[str, str, int]] = set()
    for item in raw.split(","):
        item = item.strip()
        if not item:
            continue
        if "://" in item:
            parsed = urlparse(item)
            host = (parsed.hostname or "").lower()
            if host:
                scheme = parsed.scheme.lower()
                origins.add((scheme, host, parsed.port or (443 if scheme in ("https", "wss") else 80)))
        else:
            # Bare-host entry: intentionally host-wide (any scheme/port/path).
            parsed = urlparse(f"//{item}")
            hosts.add((parsed.hostname or item).lower())
    return hosts, origins
```

headroom/proxy/upstream_guard.py
```diff
@@ -0,0 +1,88 @@
...
+    parsed = urlparse((url or "").strip())
+    if parsed.scheme.lower() not in _SAFE_SCHEMES:
+        return False
+    host = parsed.hostname
+    if not host:
+        return False
+
+    allow = _allowlisted_hosts()
+    if allow is not None:
+        return host.lower() in allow
```
Grg0rry, line 78: Matches the new `(bare_hosts, origins)` return shape from the parser above.

```suggestion
    allow = _allowlisted_hosts()
    if allow is not None:
        bare_hosts, origins = allow
        if host.lower() in bare_hosts:
            return True
        scheme = parsed.scheme.lower()
        return (scheme, host.lower(), parsed.port or (443 if scheme in ("https", "wss") else 80)) in origins
```

Grg0rry (commented):
Hi! Gregory here from u9up, the one that filed the original disclosure report. Great to see this hardening pass; it closes the bulk of what we found. I did an independent re-read of the two guards Jerrett flagged and can confirm both blockers are still present in head 78080286, so I've left minimal root-cause suggestions below (one-click Commit suggestion on each). Hope it helps!

JerrettDavis (changes requested):
Thanks for the current-head refresh. I rechecked `ac962fcc` specifically against the two security blockers from the earlier reviews; both are still present, so this still needs changes before approval.

1. `headroom/proxy/identity.py:91-93` still treats `HEADROOM_USER_ID_ALLOWLIST` as an allowlist of requested user ids. A non-loopback caller can still send `x-headroom-user-id: alice` and get Alice's memory partition whenever `alice` is listed. The allowlist needs to authenticate/authorize the caller side, or the default resolver should ignore remote user-id headers and leave tenant selection to the custom resolver hook. `tests/test_identity_resolution.py:44` still codifies the unsafe remote-claim behavior.

2. `headroom/proxy/upstream_guard.py:43` still stores only the hostname from `HEADROOM_ALLOWED_BASE_URLS`, and `is_safe_upstream_url()` accepts on `host.lower() in allow` at line 78. An operator allowlisting `https://llm.corp:8443` still also allows other schemes/ports on `llm.corp`. Full URL entries should compare normalized origin/scheme+host+port; only bare-host entries should intentionally mean host-wide allowlisting. `tests/test_upstream_guard.py:37` still lacks the cross-scheme/port negative case.

I did not run the broader suite because these are policy-level blockers visible in the current source and tests.

Grg0rry (commented):
-

Grg0rry (commented):
Hi @chopratejas, just wanted to follow up on the changes. Any luck with making the last few touches?

JerrettDavis (approved):
Validated after resolving against current main: 89 targeted security/proxy tests pass; Ruff passes; Docker Compose validates with required secrets and fails closed without them; all retained binary SHA-256 pins match live release assets. Retired RTK/lean-ctx registry entries remain removed.

Grg0rry:
Hi @JerrettDavis @chopratejas, is it possible to proceed to merge this?

Grg0rry:
Hi @DevanshiVyas, could we merge it?

JerrettDavis (approved):
Revalidated current head 546cb99b after merging current main. The remote identity header is now ignored by the default resolver, full URL allowlist entries match exact scheme/host/port, DNS failures reject by default, and newer main same-origin/loopback protections are preserved. Local verification: 240 relevant tests passed with 9 live-only skips; Ruff, mypy, pre-commit, Docker Compose success/fail-closed validation, and all 14 live binary SHA-256 pins passed.

JerrettDavis (approved):
Revalidated current head 839a12e0 after merging current main. Remote identity headers cannot select partitions in the default resolver; full URL allowlist entries match exact scheme/host/port; DNS failures reject by default; newer same-origin and loopback protections remain intact. Local verification: 240 relevant tests passed with 9 live-only skips; Ruff, mypy, pre-commit, Docker Compose success/fail-closed validation, and all 14 live binary SHA-256 pins passed.

JerrettDavis (approved):
Revalidated head 7a6af510. The prior CI failures were reserved-domain test fixtures that had not opted into the new fail-closed SSRF policy; all four affected modules now declare their test-only upstream allowlists. The 52 directly affected tests pass locally, alongside the earlier 240-test security/proxy regression run.

PR #2220, opened 2026-07-15
fix(diff-compressor): CJK-aware relevance scoring for hunk selection

lifeodyssey:
## Description

`score_hunks` boosts diff hunks whose content overlaps the query/context (+`SCORE_CONTEXT_WORD_WEIGHT` per match); the resulting score decides which hunks survive when `max_hunks_per_file` fires. It split the context on whitespace, so a spaceless CJK query became one blob that only matched a hunk containing the whole query verbatim — relevant hunks weren't boosted and got dropped.

This adds CJK character bigrams to the query match set so a CJK query boosts the hunks it overlaps. Rust-only (`diff_compressor.py` is a thin shim over Rust; hunk scoring lives only in Rust). CJK-gated: for a pure-ASCII query `cjk_bigrams` returns an empty set and the new loop is a no-op, so non-CJK scoring is byte-identical and the 20 diff parity fixtures stay green.

## Type of Change

## Changes Made

- `crates/headroom-core/src/transforms/diff_compressor.rs`: add `is_cjk_char` + `cjk_bigrams`, and a separate loop in `score_hunks` that boosts hunks containing each CJK query bigram. The existing ASCII word loop is untouched.
- Rust unit test (`cjk_bigrams` extraction) + an end-to-end test (a CJK query promotes the overlapping hunk into the kept set; the no-query baseline drops it).

## Testing

- [x] Unit tests pass (`cargo test`)
- [x] Linting passes (`cargo clippy` / `cargo fmt`)

### Test Output

```text
$ cargo test -p headroom-core --lib diff_compressor
test result: ok. 23 passed; 0 failed

$ cargo clippy -p headroom-core   # clean
```

## Real Behavior Proof

- Environment: macOS (Darwin), Rust via cargo, branch `feat/diff-compressor-cjk` off `main`.
- Exact command / steps: `cargo test -p headroom-core --lib diff_compressor` — the `cjk_query_boosts_matching_hunk_into_kept_set` test builds a diff with 4 hunks (first / plain / cjk / last), `max_hunks_per_file = 3` (one contested middle slot between the plain hunk at change-density `0.12` and the CJK hunk at `0.06`), and compresses it once with the CJK context `数据库连接超时排查` and once with no query.
- Observed result: with no query the higher-density plain hunk takes the slot (the CJK hunk `数据库连接失败重试` is dropped); with the CJK context its bigrams (`数据` / `据库` / `库连` / `连接`) match → score `0.06 + 4×0.2 = 0.86` beats the plain hunk's `0.12` → the CJK hunk survives. Both directions are asserted; before this change the spaceless CJK query matched neither hunk and the CJK hunk was always dropped.
- Not tested: the Python side — `diff_compressor.py` is a thin shim that delegates `compress()` straight to Rust, so hunk scoring has no Python twin; and no new parity fixtures were recorded, since the 20 existing diff fixtures contain no CJK and therefore stay byte-identical.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A (internal scoring)
- [ ] I have updated the CHANGELOG.md — N/A: internal relevance-scoring fix, no user-facing surface change

## Additional Notes

- Completes the relevance-scorer CJK sweep across the compressors (search, adaptive sizer, the shared BM25 tokenizer, and now diff). Rust-only — no Python parity mirror is needed because diff hunk scoring has no Python twin (the shim delegates `compress()` straight to Rust).

### DIFF
--- a/crates/headroom-core/src/transforms/diff_compressor.rs
+++ b/crates/headroom-core/src/transforms/diff_compressor.rs
@@ -35,7 +35,7 @@
 //! method returns it alongside the parity-equal result; `compress` is the
 //! parity-only API that just emits a `tracing::info_span`.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::sync::OnceLock;
 use std::time::Instant;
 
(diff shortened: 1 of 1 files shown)


JerrettDavis (approved):
Reviewed the CJK relevance-scoring change. The implementation keeps ASCII behavior unchanged by adding a separate CJK-bigram boost path, and the regression test covers both sides of the selection behavior: no-query keeps the higher-density plain hunk, while a CJK context promotes the overlapping CJK hunk into the retained set.

Local validation on head 68ede890:
- cargo test -p headroom-core --lib diff_compressor --quiet
- cargo fmt --all -- --check
- cargo clippy -p headroom-core --lib --quiet

Issue #2221, opened 2026-07-15
[BUG] Stale ANTHROPIC_BASE_URL left in settings.json after proxy stops, breaks fresh Claude Code sessions with ConnectionRefused


Atishyy27:
## Description

headroom wrap claude writes ANTHROPIC_BASE_URL=http://127.0.0.1:8787 into ~/.claude/settings.json. If the proxy stops running (I just closed the terminal and later rebooted), that line stays in the config. Next time I opened Claude Code with no proxy running, every request failed with ConnectionRefused. Nothing pointed at headroom as the cause, so it looked exactly like a network problem. This is different from #2107 where the proxy crashed during a session. Here the proxy was simply gone and a brand new session got broken by the leftover config.

## To Reproduce

1. Install headroom and run `headroom wrap claude` (persistent). It sets ANTHROPIC_BASE_URL=http://127.0.0.1:8787 in ~/.claude/settings.json.
2. Close the terminal or reboot so the proxy is no longer running.
3. Open a new shell and start Claude Code without wrapping.
4. Every prompt fails with ConnectionRefused.

## Expected Behavior

If the proxy at 127.0.0.1:8787 is not reachable, it should fall back to the real Anthropic API, or at least tell me the configured base URL points at a headroom proxy that is not running, instead of retrying a dead port forever. headroom unwrap should also be runnable without the proxy up.

## Actual Behavior

The stale ANTHROPIC_BASE_URL stayed in settings.json and every fresh session failed with a generic ConnectionRefused. I could not run headroom unwrap either, because headroom was not on PATH in a fresh Git Bash shell, so the only fix was deleting the line from settings.json by hand.

## Code Sample

Not a python API issue, it is the CLI wrap. Repro is shell:

```bash
headroom wrap claude
# close the terminal / reboot so the proxy dies
# open a fresh shell and start claude code
grep -n "ANTHROPIC_BASE_URL" ~/.claude/settings.json
# "ANTHROPIC_BASE_URL": "http://127.0.0.1:8787"   <- still there, proxy long gone
```

## Error Output

```
❯ can u run it
● API Error: Unable to connect to API (ConnectionRefused)
❯ hi
● API Error: Unable to connect to API (ConnectionRefused)
```

curl to the real API works fine, which is what threw me off, since curl does not read settings.json:

```bash
$ curl -sS -o /dev/null -w "%{http_code}\n" http://127.0.0.1:8787
[...]

PR #2223, opened 2026-07-15
fix(wrap): self-heal a stale ANTHROPIC_BASE_URL left by a dead proxy

lifeodyssey:
## Description

`headroom wrap claude` persists `ANTHROPIC_BASE_URL=<proxy>` into project-local `.claude/settings.local.json`. This is required: Claude Code's cc-daemon spawn-forks conversation workers that read settings fresh rather than inherit env, so the URL cannot just live in the child process env.

When the proxy then dies via a **hard reboot / SIGKILL**, no signal/atexit cleanup fires, so the stale URL lingers and bricks a later **bare `claude`** with ConnectionRefused (#2221). #1768's mitigations (SIGHUP, next-`wrap` self-heal, doctor WARN) don't cover "reboot → bare `claude`", and — the key gap — `wrap` installed no hook of its own, so for a user who only ever ran `wrap claude` (never `init claude`) there was nothing to clean it up.

`wrap claude` now installs a **SessionStart-only** self-heal hook (removed again on `unwrap`) that clears the persisted base URL **iff the recorded proxy port fails a retry-hardened liveness probe**. A responding proxy is never cleared, and the retry (3 attempts ~250 ms apart, alive on first success) keeps a transient blip from clearing a live session mid-run. Because workers read settings fresh per conversation, clearing at session start unblocks the current session too, not only the next.

## Design note / assumption (for maintainer confirmation)

This relies on **the SessionStart hook completing before the first cc-daemon conversation worker reads `settings.local.json`**. That ordering lives in Claude Code, not this repo; it is grounded in the documented spawn-fresh-read model (the same reason the URL must be persisted at all). Raised on the issue for confirmation. The truly launcher-agnostic fix would be upstream — Claude Code falling back to the real upstream when its configured base URL is unreachable — which would make any stale local URL harmless.

## Type of Change

## Changes Made

- `headroom/cli/wrap.py`:
  - `_wrap_proxy_alive(port, attempts=3, delay=0.25)` — retry-hardened liveness (alive on first success, dead only if all fail).
  - `_check_and_clear_dead_wrap_marker` — port is authoritative (survives PID reuse after reboot); a single probe decides; a responding proxy is never cleared; falls back to PID staleness only for port-less markers.
  - `_ensure_claude_wrap_selfheal_hook` / `_remove_claude_wrap_selfheal_hook` — install on `wrap claude`, remove on `unwrap`; **SessionStart-only** (never PreToolUse), idempotent, preserves the `env` block and unrelated/user hooks.
  - hidden `wrap selfheal` command the hook invokes.

## Testing

- [x] Linting passes (`ruff`)

### Test Output

```text
$ pytest tests/test_cli/test_wrap_dead_marker_selfheal.py
22 passed

$ pytest <related wrap/unwrap suites>
59 passed, 1 failed   # the 1 failure (test_wrap_marker_is_stale_when_pid_reused)
                      # is PRE-EXISTING + unrelated — fails identically on clean main
                      # (macOS _proc_identity returns None); this PR touches neither
                      # _wrap_marker_is_stale nor _identity_mismatch.

$ ruff check / mypy headroom/cli/wrap.py   # clean
```

## Real Behavior Proof

- Environment: macOS (Darwin), Python in a uv venv, branch `feat/wrap-stale-url-selfheal` off `main`.
- Exact command / steps: `pytest tests/test_cli/test_wrap_dead_marker_selfheal.py` exercises: `wrap claude` writes a SessionStart-only self-heal hook into `settings.local.json` (idempotent, not on PreToolUse); `unwrap` removes it (keeping unrelated hooks); the `wrap selfheal` command clears a dead-port marker's base URL; and — bound to a REAL listening socket — a live proxy's marker is never cleared, including when a single probe transiently fails but the retry succeeds.
- Observed result: dead-proxy marker → base URL restored to its prior value; live-proxy marker (real socket) → preserved; no marker / no settings file / port-less marker → no-op, no exception. All 22 pass.
- Not tested: the actual Claude Code hook-vs-worker execution ordering (upstream, not in this repo) — see the Design note; the fix is correct given that documented model.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A (internal wrap behavior)
- [ ] I have updated the CHANGELOG.md — happy to add an entry if preferred.

## Additional Notes

- Scoped to the `wrap claude` project-local path (the reported scenario). The opt-in cc-switch reconciler writes `ANTHROPIC_BASE_URL` into the *global* `~/.claude/settings.json` with no restore today — a separate, lower-frequency gap I can follow up on if wanted.

### DIFF
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -1162,9 +1162,9 @@ def _ensure_claude_wrap_selfheal_hook(settings_path: Path) -> None:
     if not isinstance(payload, dict):
         payload = {}
     hooks = dict(payload.get("hooks") or {}) if isinstance(payload.get("hooks"), dict) else {}
-    entries = list(hooks.get("SessionStart") or []) if isinstance(
-        hooks.get("SessionStart"), list
-    ) else []
+    entries = (
+        list(hooks.get("SessionStart") or []) if isinstance(hooks.get("SessionStart"), list) else []
+    )
     already = any(
         isinstance(entry, dict)
         and isinstance(entry.get("hooks"), list)
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -793,6 +793,11 @@ _HEADROOM_HOOK_MARKERS = ("rtk-rewrite", "headroom-init-claude")
 # (GH #746), paired with init/wrap setting it.
 _HEADROOM_ENV_KEYS = ("ANTHROPIC_BASE_URL", "ENABLE_TOOL_SEARCH")
 
+# Stable marker embedded in the SessionStart self-heal hook that ``wrap claude``
+# installs (issue #2221). Lets that hook be found (idempotent install) and
+# removed (unwrap) by its command string.
+_WRAP_SELFHEAL_HOOK_MARKER = "headroom-wrap-selfheal"
+
 
 def _remove_claude_rtk_hooks(settings_path: Path | None = None) -> bool:
     """Remove Headroom-managed entries from Claude settings.json.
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed current head a67ae199 after applying the formatter-only fix.

Validation run locally:
- uv run --with pytest --with fastapi --with httpx --with anyio --with uvicorn --with h2 pytest tests/test_cli/test_wrap_dead_marker_selfheal.py -q
- uvx ruff==0.15.17 check headroom/cli/wrap.py tests/test_cli/test_wrap_dead_marker_selfheal.py --output-format concise
- uvx ruff==0.15.17 format --check headroom/cli/wrap.py tests/test_cli/test_wrap_dead_marker_selfheal.py

The implementation is careful about the important safety boundary: a responding proxy is never cleared, and the retry window avoids treating a transient connect miss as a dead wrapped session. The remaining assumption is the one already called out in the PR body: Claude Code must run SessionStart before the first worker reads project settings.

PR #2227, opened 2026-07-15
fix(cli/init): fail clearly on a target settings file with invalid JSON

abhay-codes07:
## Description

`headroom init` crashes with a raw traceback when a target's settings file contains invalid JSON.

`_json_file` reads the JSON config files that init read-merge-writes (Claude's `settings.json`, Codex's `hooks.json`, etc.):

```python
def _json_file(path: Path) -> dict[str, Any]:
    if not path.exists():
        return {}
    content = path.read_text(encoding="utf-8").strip()
    if not content:
        return {}
    payload = json.loads(content)          # unguarded
    return payload if isinstance(payload, dict) else {}
```

These are user-owned files that people hand-edit, so a stray trailing comma or an unquoted key is entirely plausible. When that happens `json.loads` raises `json.JSONDecodeError` and it propagates all the way out, so `headroom init` dies with a Python traceback instead of a usable message.

Returning `{}` on the error would be worse, not better: every caller does `payload = _json_file(path)` then `_write_json(path, payload)`, so an empty dict would make init overwrite the user's real settings with just the hooks/env block — silent data loss.

## Fix

Guard the parse and convert it into an actionable `ClickException` that names the file and the parse error, leaving the file untouched:

```python
try:
    payload = json.loads(content)
except json.JSONDecodeError as e:
    raise click.ClickException(
        f"{path} contains invalid JSON ({e}); fix it and re-run, or move it aside."
    ) from e
```

`click.ClickException` is already the project's convention for user-facing init failures (e.g. the `'claude' not found in PATH` messages). The user now gets a clear instruction, and their file is never clobbered.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cli/init.py`: wrap the `json.loads` in `_json_file` and raise a `ClickException` on `JSONDecodeError`.
- `tests/test_cli/test_init_cli.py`: new test asserting a malformed file raises `ClickException` (matching "invalid JSON") and is left byte-for-byte untouched.

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/cli/init.py tests/test_cli/test_init_cli.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/cli/init.py
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOM-kills this box (ML stack import), so I reproduced the behavior with a dependency-free script mirroring `_json_file` and left the full pytest to CI.
- Exact command / steps: wrote a settings file containing `{"env": {"A": "B",}}` (trailing comma), then called the OLD unguarded reader and the NEW guarded reader; also re-checked a valid file round-trips.
- Observed result: OLD raises a raw `json.JSONDecodeError` (the init traceback); NEW raises a `ClickException` containing "invalid JSON" and leaves the file byte-for-byte unchanged; valid JSON still parses to the same dict.
- Not tested: a full `headroom init` end-to-end run; full local `pytest` deferred to CI (OOM).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because the full suite imports the ML stack, which I can't run here. The new test uses the existing `_load_init_module` harness (the same one the neighbouring `test_json_file_*` tests use), so it runs under the normal CI pytest job; behaviour is additionally verified by the standalone proof above.


JerrettDavis (approved):
Reviewed current head 0db7981c.

Validation run locally:
- uv run --with pytest --with fastapi --with httpx --with anyio --with uvicorn --with h2 pytest tests/test_cli/test_init_cli.py -q
- uvx ruff==0.15.17 check headroom/cli/init.py tests/test_cli/test_init_cli.py --output-format concise

The failure mode is handled the right way: abort with a ClickException and leave the user-owned settings file untouched, instead of returning an empty dict that would risk overwriting existing settings on the subsequent write.

PR #2231, opened 2026-07-15
fix(memory): make explicit-project and user store keys collision-resistant

abhay-codes07:
## Description

Two of the memory storage router's key-derivation paths can pool distinct identities into one store.

`ProjectResolver._identity_from_cwd` builds a collision-resistant key by appending a `sha256` digest to the sanitized basename:

```python
safe_basename = cls._sanitize_basename(basename) or "project"
digest = hashlib.sha256(normalised.encode("utf-8")).hexdigest()[:16]
key = f"{safe_basename}-{digest}"
```

But the two non-cwd paths use the bare sanitized basename as the key:

```python
# Tier 1 — explicit x-headroom-project-id
safe = self._sanitize_basename(explicit)
if safe:
    return safe, explicit           # <-- no digest

# USER mode
user_safe = ProjectResolver._sanitize_basename(ctx.base_user_id) or "default"
db_path = self._config.root_dir / "users" / user_safe / "memory.db"   # <-- no digest
```

`_sanitize_basename` maps every disallowed character to a single dash, so distinct inputs collapse to the same basename:

- `acme/api` and `acme api` (and `acme@api`) all → `acme-api`
- user ids `alice/qa` and `alice qa` → `alice-qa`

Both the project key (`root/projects/<key>/memory.db`) and the USER key (`root/users/<key>/memory.db`) are derived directly from that basename, so two distinct project ids — or, in USER mode, two distinct **users** — resolve to the same `memory.db` and share each other's memories. USER mode exists specifically to isolate users, so this is a cross-user data-isolation leak; the explicit-project-id path is the same leak across projects. Both are client-controlled (`x-headroom-project-id` / `x-headroom-user-id` headers), so the collision is easy to hit and could even be provoked deliberately.

## Fix

Append the same digest of the raw id to both keys, exactly as `_identity_from_cwd` does, keeping the sanitized basename as a human-readable prefix:

```python
digest = hashlib.sha256(explicit.encode("utf-8")).hexdigest()[:16]
return f"{safe}-{digest}", explicit
```

```python
digest = hashlib.sha256(ctx.base_user_id.encode("utf-8")).hexdigest()[:16]
user_key = f"{user_safe}-{digest}"
db_path = self._config.root_dir / "users" / user_key / "memory.db"
```

Distinct ids now always land on distinct stores; the same id remains stable across calls.

**Migration note:** this changes the on-disk key format for the explicit-project and USER stores (`<basename>` → `<basename>-<digest>`). Memories written under the old bare-basename paths are not migrated; the router will start a fresh store at the new path. GLOBAL and cwd-derived PROJECT stores (which already carried the digest) are unaffected. Flagging this explicitly so you can decide whether a migration shim is wanted before merge.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/memory/storage_router.py`: append a `sha256` digest to the explicit-project-id key (Tier 1) and the USER-mode key, matching `_identity_from_cwd`.
- `tests/test_memory_storage_router.py`: update the Tier-1 key assertion to the prefix+digest form; add collision regression tests for the explicit-project and USER paths.
- `CHANGELOG.md`: Bug Fixes entry (including the migration note).

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/memory/storage_router.py tests/test_memory_storage_router.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/memory/storage_router.py
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOM-kills this box (ML stack import), so I reproduced the key derivation with a dependency-free script mirroring `_sanitize_basename` + the digest, and left the full pytest to CI.
- Exact command / steps: derived keys for `alice/qa` and `alice qa` under the OLD bare-basename scheme and the NEW digest scheme.
- Observed result: OLD → both `alice-qa` (identical → shared store); NEW → `alice-qa-7e02fc2dfbc447b4` vs `alice-qa-4c9241514a374ba3` (distinct), stable per input, with the `alice-qa-` prefix retained.
- Not tested: a live proxy with two colliding tenants; full local `pytest` deferred to CI (OOM).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because the full suite imports the ML stack, which I can't run here. The changed/added tests use the existing `tests/test_memory_storage_router.py` harness so they run under the normal CI pytest job; behaviour is additionally verified by the standalone proof above. I updated `test_resolver_tier1_explicit_project_id_wins` to assert the new prefix+digest key. Happy to add a migration shim (read the old path if the new one is empty) if you'd prefer that over the fresh-store behavior.


JerrettDavis (approved):
Reviewed current head e07a141b after the stale proxy-helper assertion was aligned.

Validation run locally:
- uv run --with pytest --with fastapi --with httpx --with anyio --with uvicorn --with h2 pytest tests/test_proxy_handler_helpers.py::test_resolve_ccr_workspace_explicit_project_id_wins tests/test_memory_storage_router.py -q
- uvx ruff==0.15.17 check headroom/memory/storage_router.py tests/test_memory_storage_router.py tests/test_proxy_handler_helpers.py --output-format concise

The collision fix is sound: explicit project IDs and USER-mode user IDs now retain the readable sanitized prefix while adding a raw-ID digest, which closes the sanitize-to-same-path leak. The no-migration behavior is a product tradeoff, but it is explicitly documented in both the PR body and changelog.

JerrettDavis (approved):

abhay-codes07:
Thanks for the careful look here, and for aligning the proxy-helper assertion on your side.

On the no-migration behavior: I went with a fresh store at the new path to keep the change small, but if you would rather not drop the existing explicit-project and USER stores, I can add a read-old-path-if-the-new-one-is-empty shim so the first request after upgrade still finds prior memories and then writes forward under the digest key. Say the word and I will push it to this branch. Otherwise this is good to go once the main CI settles.

JerrettDavis:
Merged current `headroomlabs/main` to pick up the CI formatting baseline and removed the manual changelog entry so the new changelog guard can pass. Base-to-head diff is now limited to `headroom/memory/storage_router.py`, `tests/test_memory_storage_router.py`, and `tests/test_proxy_handler_helpers.py`.

Local verification on updated head `fcb170d0`:
- `uv run --extra dev pytest tests/test_memory_storage_router.py tests/test_proxy_handler_helpers.py -q`
- `uv run --extra dev ruff check headroom/memory/storage_router.py tests/test_memory_storage_router.py tests/test_proxy_handler_helpers.py`
- `uv run --extra dev ruff format --check headroom/memory/storage_router.py tests/test_memory_storage_router.py tests/test_proxy_handler_helpers.py`

JerrettDavis (approved):
Re-reviewed the current head after the changelog cleanup. The explicit project-id and USER-mode keys now keep the readable sanitized prefix but append a digest of the raw id, which closes the sanitize-collision isolation hole without changing display labels. The regression coverage pins distinct ids that sanitize alike. GitHub checks are green.

Local note: a fresh minimal local pytest run hit a missing `fastapi` optional dependency while collecting `tests/test_proxy_handler_helpers.py`; I inspected the diff directly and relied on the green PR CI for that proxy-helper coverage.

PR #2232, opened 2026-07-15
fix(transforms/smart_crusher): don't crash on a tool call with a null function

abhay-codes07:
## Description

A tool call whose `function` field is explicitly `null` crashes SmartCrusher's per-request context extraction.

`_extract_context_from_messages` (called at the top of `apply()`) walks recent assistant tool calls:

```python
for tc in msg.get("tool_calls", []):
    if isinstance(tc, dict):
        func = tc.get("function", {})
        args = func.get("arguments", "")
```

`dict.get("function", {})` only substitutes `{}` when the key is **missing**. When the key is present but `null` — `{"id": "1", "type": "function", "function": null}`, which clients emit for a partial or streamed tool call — `func` is `None`, and `None.get("arguments")` raises `AttributeError`. That propagates out of `_extract_context_from_messages` and crashes `apply()` for the entire request, so the request either errors or has to fail open to uncompressed with a logged traceback.

The sibling `_build_tool_name_index` in the same file already guards this exact shape with `(tc.get("function") or {})` — this call site just wasn't updated to match.

## Fix

Use the same null-safe form:

```python
func = tc.get("function") or {}
```

`None` (and any other falsy value) now collapses to `{}`, the null tool call contributes no context, and extraction continues to the next call.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/smart_crusher.py`: `tc.get("function", {})` → `tc.get("function") or {}` in `_extract_context_from_messages`.
- `tests/test_transforms/test_smart_crusher_bugs.py`: new test asserting a `{"function": null}` tool call doesn't crash extraction and later calls are still read.

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/transforms/smart_crusher.py tests/test_transforms/test_smart_crusher_bugs.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/transforms/smart_crusher.py
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOM-kills this box (ML stack import), so I reproduced the extraction loop with a dependency-free script and left the full pytest to CI.
- Exact command / steps: ran an assistant message with tool calls `[{"function": null}, {"function": {"arguments": "keep-me"}}]` through the OLD `get("function", {})` loop and the NEW `get("function") or {}` loop.
- Observed result: OLD raises `AttributeError` on the null function; NEW skips it and returns `"keep-me"` from the following call.
- Not tested: a live proxy request carrying a null-function tool call; full local `pytest` deferred to CI (OOM).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because the full suite imports the ML stack, which I can't run here. The new test uses the existing `_make_crusher` helper in `tests/test_transforms/test_smart_crusher_bugs.py`, so it runs under the normal CI pytest job; behaviour is additionally verified by the standalone proof above.


JerrettDavis (approved):
Reviewed current head e1bb2f9c.

The change is narrowly scoped and matches the existing null-safe pattern used by the sibling tool-name extraction path. The regression test covers the explicit null function shape and verifies extraction continues to later calls.

Validation:
- uvx ruff==0.15.17 check headroom/transforms/smart_crusher.py tests/test_transforms/test_smart_crusher_bugs.py --output-format concise
- Current GitHub checks are green for this head.

Local note: the focused pytest file could not run in my disposable Windows venv because headroom._core was not built there: ImportError: DLL load failed while importing _core.

JerrettDavis (approved):

JerrettDavis:
Merged current `headroomlabs/main` to pick up the CI formatting baseline and removed the manual changelog entry so the new changelog guard can pass. Base-to-head diff is now limited to `headroom/transforms/smart_crusher.py` and `tests/test_transforms/test_smart_crusher_bugs.py`.

Local verification on updated head `ddbe7f73`:
- `uv run --extra dev pytest tests/test_transforms/test_smart_crusher_bugs.py -q`
- `uv run --extra dev ruff check headroom/transforms/smart_crusher.py tests/test_transforms/test_smart_crusher_bugs.py`
- `uv run --extra dev ruff format --check headroom/transforms/smart_crusher.py tests/test_transforms/test_smart_crusher_bugs.py`

JerrettDavis (approved):
Re-reviewed the current head after the changelog cleanup. The null-function tool-call case is fixed with `tc.get("function") or {}`, matching the sibling null-safe path, and the regression covers preserving later valid arguments while skipping the null tool call. GitHub checks are green.

Local note: the focused test file could not run in this Windows worktree because the local native `headroom._core` extension failed to load; I inspected the Python diff directly and used the green PR CI as verification for the test execution.

PR #2052, opened 2026-07-12
fix(proxy): handle content-part outputs in Codex Responses compression

lennney:
## Description

Fixes 0% savings when wrapping Codex (Closes #2050).  `_slot_text` and the
lossless-excluded fold in the OpenAI Responses compression path only handled
`function_call_output` items whose `output` field is a plain string, silently
skipping items whose `output` is an array of content parts (valid per OpenAI
spec).  Use `_responses_part_text()` — which already handles both — so these
items reach the ContentRouter and accrue compression savings.

Also extend `_responses_input_item_text_bytes` to count text bytes inside
content-part arrays in the `output` field, matching its existing treatment of
the `content` field.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `_slot_text()` (openai.py:1260): use `_responses_part_text()` instead of
  `isinstance(output, str)` to extract text from both string and content-part
  outputs
- Lossless excluded fold (openai.py:1362): same change — use
  `_responses_part_text()` so excluded-tool outputs with content parts can
  still be losslessly compacted
- `_responses_input_item_text_bytes()` (openai.py:547): extend byte counting
  to handle content-part arrays in the `output` field, matching existing
  `content` field handling

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Manual testing performed

### Test Output

```text
$ uv run pytest tests/test_openai_responses_compression_units.py -x -q
16 passed in 1.19s

$ uv run pytest tests/ -k "openai and responses and compress" -x -q
19 passed, 14 skipped in 18.17s

$ uv run ruff check headroom/proxy/handlers/openai.py
All checks passed!
```

## Real Behavior Proof

- Environment: Linux (6.8.0-124-generic), Python 3.12.3, headroom main @ 868b88bc
- Exact command / steps: checkout branch, run `uv run pytest tests/test_openai_responses_compression_units.py -x -q`, run `uv run pytest tests/ -k "openai and responses and compress" -x -q`, run `uv run ruff check headroom/proxy/handlers/openai.py`
- Observed result: All 35 tests pass (16 units + 19 integration), ruff clean, no regressions
[...]

PR #2246, opened 2026-07-15
fix(proxy): preserve content-part array structure in Responses tool output write-back

### LINKED ISSUE #2235: [BUG] Excluded-tool Responses array output is rewritten as a string
## Description

The OpenAI Responses compression adapter can change an excluded tool result from an array of content parts into a plain string.

For excluded tools such as grep, the adapter joins text from output parts with _responses_part_text(), performs lossless compaction, and records the write-back slot as ("output", None). Applying that fold replaces the entire output value instead of updating text parts in place.

This violates the Responses content-part shape: part count, order, types, images, and non-text metadata are not preserved.

Upstream code: https://github.com/headroomlabs-ai/headroom/blob/cb6c8284575b70d754f5fc83158d8ef38777526b/headroom/proxy/handlers/openai.py#L1578-L1584

## To Reproduce

1. Use headroomlabs-ai/headroom main at cb6c8284575b.
2. Configure grep as an excluded tool.
3. Send a Responses payload containing:
   - a function_call named grep;
   - a matching function_call_output;
   - output as an array with one foldable output_text part and one input_image part.
4. Run the Responses compression adapter.

Minimal payload shape:

~~~json
{
  "input": [
    {
      "type": "function_call",
      "call_id": "call_1",
      "name": "grep",
      "arguments": "{}"
    },
    {
      "type": "function_call_output",
      "call_id": "call_1",
      "output": [
        {"type": "output_text", "text": "<large grep output>"},
        {"type": "input_image", "image_url": "data:image/png;base64,AA=="}
      ]
    }
  ]
}
~~~

Observed diagnostic result:

~~~json
{
  "original_output_type": "list",
  "original_part_count": 2,
  "new_output_type": "str",
  "new_part_count": null,
  "modified": true
}
~~~

## Expected Behavior

- Array output remains an array.
- Part count, order, and type remain unchanged.
- Only eligible text fields may be replaced in place.
- Images and all non-text fields remain byte-for-byte/structurally unchanged.
- If lossless folding cannot be mapped safely back to individual text parts, the adapter should pass the original array through unchanged.
- Regression tests should assert the complete output shape rather than only verifying that the folded text can be reconstructed.

## Actual Behavior

The losslessly-folded text replaces the entire output array with a string. Non-text parts and their metadata are removed from the payload.

## Environment

- **Headroom source**: main @ cb6c8284575b70d754f5fc83158d8ef38777526b
- **Python version**: 3.14.4
- **OS**: Windows
- **LLM Provider / endpoint**: OpenAI Responses / Codex
- **Related issue**: #2050
- **Related content-part fix**: #2052
- **Related aggregate-floor fix**: #2116
- **Related efficiency issue**: #2234

## Additional Context

The current regression test verifies lossless reconstruction of the folded grep text but does not assert that output remains a content-part array. This allows the structural mutation to pass the test suite.

lennney:
## Description

Preserve content-part array structure when compressing tool outputs in the OpenAI Responses format. Fixes data loss where non-text parts (images, refusals) were silently dropped during compression.

Closes #2235

## Type of Change

## Changes Made

- **`headroom/proxy/handlers/openai.py`**: Modified `_set_slot_text` to detect content-part arrays at write-back time. When `item["output"]` is a list, only the single `output_text` part's `text` field is updated; all other parts remain unchanged.
- **`tests/test_openai_responses_compression_units.py`**: Added/adjusted 10+ test cases covering normal content-part compression, non-text part preservation, excluded tool + content-part + images, multi-text-part arrays (safe skip), empty arrays, and 9 adversarial/fuzz scenarios (weird part order, extra keys, non-dict elements, unknown types, 1000-part stress test, null/int text values).

## Testing

- [x] All existing tests pass (122 passed, 0 failed)
- [x] New tests pass (34 total in compression_units, all green)
- [x] Ruff lint check passes

```bash
$ uv run pytest tests/test_openai_responses_compression_units.py -q --no-header
34 passed in 1.06s

$ uv run pytest tests/ -k "openai and (responses or compress)" --ignore=tests/test_hermes_passthrough_compression.py --ignore=tests/test_ccr_tag_placeholder_regression.py -q --no-header
131 passed, 29 skipped in 13.47s
```

## Real Behavior Proof

- Environment: Python 3.12 on Linux (uv-managed headroom project)
- Exact command / steps: `uv run pytest tests/test_openai_responses_compression_units.py`
- Observed result: All 34 tests pass, including new adversarial tests that verify non-text parts (input_image, refusal, unknown types) remain byte-identical after compression
- Not tested: End-to-end integration with actual OpenAI Responses API deployment (CI covers this)

## Review Readiness

### DIFF
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -1467,7 +1467,24 @@ class OpenAIHandlerMixin:
         ) -> None:
             kind, _ = slot
             if kind == "output":
-                item["output"] = replacement
+                current = item.get("output")
+                if isinstance(current, list):
+                    # Content-part array: preserve structure by updating the
+                    # single output_text part's text in-place. Non-text parts
+                    # (images, refusals) are left unchanged.
+                    text_part = None
+                    for part in current:
+                        if isinstance(part, dict) and part.get("type") == "output_text":
+                            if text_part is None:
+                                text_part = part
+                            else:
+                                text_part = None  # 2+ text parts → skip
+                                break
+                    if text_part is not None:
+                        text_part["text"] = replacement
+                    # else: zero or 2+ output_text parts — leave unchanged
+                else:
+                    item["output"] = replacement
 
         headroom_retrieve_call_ids: set[str] = set()
         # Map each Responses tool call to its name so that outputs belonging to
--- a/tests/test_openai_responses_compression_units.py
+++ b/tests/test_openai_responses_compression_units.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import threading
 from types import MethodType, SimpleNamespace
 
+import pytest
+
 from headroom.proxy.handlers import openai as openai_handler
 from headroom.proxy.handlers.openai import OpenAIHandlerMixin
 from headroom.transforms.compression_units import UnitCompressionResult
@@ -210,7 +212,12 @@ def test_openai_responses_adapter_compresses_output_content_parts():
 
     assert modified is True
     assert saved > 0
-    assert new_payload["input"][0]["output"] == "content part output summary"
+    output_val = new_payload["input"][0]["output"]
+    assert isinstance(output_val, list), "content-part output must remain a list"
+    assert len(output_val) == 1
+    assert isinstance(output_val[0], dict)
+    assert output_val[0].get("type") == "output_text"
+    assert output_val[0].get("text") == "content part output summary"
     assert "router:openai:responses:function_call_output:kompress" in transforms
     assert units_by_category == {"applied": 1}
     assert strategy_chain == []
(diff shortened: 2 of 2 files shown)


JerrettDavis (changes requested):
Thanks for the focused regression coverage here. The core direction is right, but I think this needs one more tightening pass before merge.

Blocking: extraction and write-back now disagree for content-part arrays. `_slot_text` still calls `_responses_part_text(item.get("output"))`, which collects every string-ish part text in the array, including bare string parts and dict parts whose type is not `output_text`. `_set_slot_text`, however, only writes back when it finds exactly one `output_text` part. That creates two unsafe cases:

- If the array has multiple text-bearing parts, the router compresses the joined text, `modified`/`tokens_saved` can report success, but the payload is left unchanged because write-back skips 2+ `output_text` parts.
- If the array has one `output_text` plus another text-bearing non-output part, the replacement for the `output_text` is generated from the concatenation of both texts while the non-output part is preserved separately. That can duplicate/summarize the wrong content into the output text part.

The fix should make extraction use the same eligibility rule as write-back: for list outputs, only extract when there is exactly one dict part with `type == "output_text"` and `text` is a non-empty string, then compress that exact string. If the array has zero, multiple, bare-string, or other text-bearing parts, skip before creating a `CompressionUnit` so accounting and payload mutation stay truthful.

Also, CI is not green for a PR-local reason: `ruff format --check .` reports `tests/test_openai_responses_compression_units.py` needs formatting. The added adversarial parametrization around the long inline dict at `tests/test_openai_responses_compression_units.py:1124` looks like one of the spots.

I did run the focused suite locally on this branch: `uv run pytest tests/test_openai_responses_compression_units.py -q --no-header` passed with 34 tests.

lennney:
Closing — upstream/main already has the fix via #2052 (merged Jul 13). This branch was created from a stale `remotes/upstream/main` (56 commits behind) and the per-index `output_part` approach in main is more complete.

PR #2250, opened 2026-07-15
fix: replace computer_call_output with apply_patch_call_output in output_shaper

fix2015:
The _RESPONSES_TOOL_OUTPUT_TYPES frozenset in output_shaper.py had computer_call_output instead of apply_patch_call_output, making it inconsistent with the canonical definitions in handlers/openai.py and output_turn_policy.py. This caused apply_patch_call_output items to be misclassified, preventing effort routing optimization for those turns.


JerrettDavis (approved):
Verified the allowlist now matches the canonical Responses output item types used by the handler and output turn policy. I added a focused regression test for apply_patch_call_output classification on this branch.

Local validation:
- uvx --from ruff==0.15.17 ruff check headroom/proxy/output_shaper.py tests/test_output_shaper_responses.py
- uvx --from ruff==0.15.17 ruff format --check headroom/proxy/output_shaper.py tests/test_output_shaper_responses.py
- uv run pytest tests/test_output_shaper_responses.py -q

PR #2257, opened 2026-07-16
fix(compression): stop JSON truncation fallback from splitting escape pairs

ErenAta16:
## Description

`_simple_compress`'s truncation fallback (`headroom/compression/universal.py`) slices JSON string-value spans by raw character offset with no awareness of JSON escape sequences. When the computed cut lands between a backslash and the character it escapes, the result has a dangling backslash — invalid JSON per RFC 8259 §7.

`JSONStructureHandler` marks a full JSON string value, quotes and raw escape sequences intact, as one compressible span. `SEC-02` (merged 2026-06-28) already fixed a related but distinct bug in this same function — a raw newline embedded in the separator string itself — and added `test_sec02_json_fallback_stays_valid_json`. That test only uses plain-word string values with no `\`, `\"`, or `\n` escapes, so it never exercised a mid-escape truncation cut, and this deeper flaw survived untouched.

## Type of Change

## Changes Made

- Added `_safe_truncate_boundary`, which nudges a candidate cut position past a split backslash escape pair by counting the run of consecutive backslashes immediately before it (odd run means mid-escape, even run means every backslash before it already pairs off cleanly).
- Applied it to both `keep_start` and the end boundary in `_simple_compress`, with the end boundary capped so the nudge can't zero out an originally-nonzero `keep_end` (`text[-0:]` is the whole string, not empty).
- Added `test_sec03_json_fallback_does_not_split_escape_pair`, sweeping a range of prefix lengths so the cut lands on the escape at least once regardless of the exact `keep_start`/`keep_end` arithmetic.

## Testing

I could not build the package through its normal tooling in my environment (no Rust toolchain for the `maturin` build), so I did not run the actual `pytest`/`ruff`/`mypy` commands. I verified by loading just the affected modules directly (`headroom.compression.universal`, `.masks`, `.detector`, and the `handlers` package) via `importlib`, bypassing the top-level `headroom/__init__.py` import chain, and running the fix against real code execution. This exercises the exact same code paths as the actual test suite would, just outside `pytest`.

### Test Output

```text
prefix_len=97 case: OK, valid JSON, compressed len = 179 vs original 310
Swept 1200 cases across 4 escape types x 300 prefix lengths: all valid JSON.

SEC-01 (existing): PASS
SEC-02 (existing): PASS
SEC-03 (new): PASS
Plain-text truncation still works: PASS
```

## Real Behavior Proof

- Environment: Windows, Python 3.12.10, `headroom` loaded via `importlib.util.spec_from_file_location` against the affected modules directly (not the installed/built package — see note above on the Rust toolchain).
- Exact command / steps: `UniversalCompressor(config=UniversalCompressorConfig(use_kompress=False, compression_ratio_target=0.5)).compress(content, content_type=ContentType.JSON)` for `content = json.dumps({"msg": ("x" * 97) + "\n" + ("y" * 200)})`, then `json.loads()` on the result.
- Observed result: before the fix, this raises `json.JSONDecodeError: Invalid \escape: line 1 column 107 (char 106)` — the compressed output is `'...xxxxx\\ ...[compressed]... yyyy...'`, a dangling backslash. After the fix, `json.loads()` succeeds. Ran a sweep of 1200 cases (4 escape types × 300 prefix lengths) confirming every one produces valid JSON after the fix, plus the two pre-existing `SEC-01`/`SEC-02` regression tests to confirm no behavior change there.
- Not tested: the `use_kompress=True` path (Kompress unavailable/not installed in my environment, so only the `_simple_compress` fallback path itself was exercised — this is also the only path the fix touches); the actual `pytest`/`ruff`/`mypy` commands through the project's own tooling; splitting a `\uXXXX` (6-character) unicode escape mid-sequence specifically — the fix corrects the common 2-character escapes (`\n`, `\"`, `\\`, `\t`, etc.) demonstrated in the repro, but a cut landing between the 4 hex digits of `\uXXXX` (rather than right after the backslash) isn't covered by the backslash-run check and would need separate handling.

## Review Readiness

## Checklist

## Additional Notes

Left the CHANGELOG and "new/existing unit tests pass locally" checkboxes unchecked rather than guessing: I verified via the isolated-import approach described above, not the project's actual `pytest` run, and didn't want to claim a checkbox I can't back with real tooling output. Happy to add a CHANGELOG entry if a maintainer confirms the expected format/section for this scope of fix.

This PR was fully generated with an AI assistant (Claude). I reviewed the change, traced the bug mechanism against the actual source (the tokenizer's own escape-walking logic in `json_handler.py` vs. `_simple_compress`'s lack of it), and ran the verification described above myself before opening this.

### DIFF
--- a/headroom/compression/universal.py
+++ b/headroom/compression/universal.py
@@ -198,25 +198,34 @@ class UniversalCompressor:
 
     @staticmethod
     def _safe_truncate_boundary(text: str, pos: int) -> int:
-        """Nudge a truncation boundary past a backslash escape pair it would split.
+        """Nudge a truncation boundary to the next position that doesn't split
+        a JSON backslash escape.
 
         A span passed here may be raw JSON string content (quotes and escapes
-        intact, not decoded), so cutting between a backslash and the character
-        it escapes (e.g. "...\\" | "n..." -> a dangling "\\") produces invalid
-        JSON. Counting the run of backslashes immediately before `pos` is
-        enough to tell: an odd run means `pos` sits mid-escape (including
-        mid "\\uXXXX", though only the first character of the pair is
-        realigned there), an even run means every backslash before `pos`
-        already pairs off cleanly.
+        intact, not decoded), so cutting inside an escape - a 2-character form
+        like "\\n"/"\\""/"\\\\", or the 6-character "\\uXXXX" unicode form -
+        produces invalid JSON (a dangling backslash, or a truncated \\u with
+        fewer than 4 hex digits).
+
+        Walks `text` as a sequence of JSON string tokens from the start,
+        treating "\\" + the next character as one atomic unit (expanded to 6
+        characters when that next character is "u"), and returns the first
+        token boundary at or after `pos`. Backslash-run parity alone isn't
+        enough here: it correctly flags a split 2-character escape, but a cut
+        landing among the hex digits of "\\uXXXX" (after the "u", not
+        immediately after the backslash) has no backslash immediately before
+        it at all, so a purely backward-looking check misses it.
         """
         if pos <= 0 or pos >= len(text):
             return pos
-        backslash_run = 0
-        i = pos - 1
-        while i >= 0 and text[i] == "\\":
-            backslash_run += 1
-            i -= 1
-        return pos + 1 if backslash_run % 2 == 1 else pos
+        i = 0
+        n = len(text)
+        while i < pos:
+            if text[i] == "\\" and i + 1 < n:
+                i += 6 if text[i + 1] == "u" else 2
+            else:
+                i += 1
+        return i
 
     def _simple_compress(self, text: str) -> str:
         """Simple compression fallback (truncation with indicator).
--- a/headroom/compression/universal.py
+++ b/headroom/compression/universal.py
@@ -196,6 +196,28 @@ class UniversalCompressor:
             logger.warning("Kompress compression failed: %s", e)
             return self._simple_compress(text)
 
+    @staticmethod
+    def _safe_truncate_boundary(text: str, pos: int) -> int:
+        """Nudge a truncation boundary past a backslash escape pair it would split.
+
+        A span passed here may be raw JSON string content (quotes and escapes
+        intact, not decoded), so cutting between a backslash and the character
+        it escapes (e.g. "...\\" | "n..." -> a dangling "\\") produces invalid
+        JSON. Counting the run of backslashes immediately before `pos` is
+        enough to tell: an odd run means `pos` sits mid-escape (including
+        mid "\\uXXXX", though only the first character of the pair is
+        realigned there), an even run means every backslash before `pos`
+        already pairs off cleanly.
+        """
+        if pos <= 0 or pos >= len(text):
+            return pos
+        backslash_run = 0
+        i = pos - 1
+        while i >= 0 and text[i] == "\\":
+            backslash_run += 1
+            i -= 1
+        return pos + 1 if backslash_run % 2 == 1 else pos
+
     def _simple_compress(self, text: str) -> str:
         """Simple compression fallback (truncation with indicator).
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
I found a remaining correctness hole in the fallback boundary logic.

`_safe_truncate_boundary()` only nudges when the cut is immediately after an odd backslash run. That fixes ``n` / ``"` style two-character escapes, but it can still cut inside a JSON unicode escape after the `u` or one of the hex digits. In that case the prefix contains an incomplete `\uXXXX` sequence and the compressed JSON is still invalid.

Minimal reproduction on this branch:

```python
import json
from headroom.compression.universal import UniversalCompressor, UniversalCompressorConfig, ContentType

c = UniversalCompressor(
    config=UniversalCompressorConfig(use_magika=False, use_kompress=False, ccr_enabled=False)
)
payload_value = ("x" * 67) + "é" + ("y" * 300)
content = json.dumps({"msg": payload_value})
r = c.compress(content, content_type=ContentType.JSON)
json.loads(r.compressed)
```

This fails with:

```text
JSONDecodeError: Invalid \uXXXX escape
```

The compressed output contains the incomplete escape:

```text
...xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\u00e ...[compressed]... yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy"}
```

Please make the boundary logic aware of the full six-character JSON unicode escape form (`\u` + four hex digits), and add a regression test with a non-ASCII character that `json.dumps()` serializes as `\uXXXX`.

Verified locally:
- Existing new test passes: `uv run pytest tests/test_compression/test_universal.py::TestSecurityAuditRegressions::test_sec03_json_fallback_does_not_split_escape_pair -q`
- The reproduction above fails on the PR branch.

ErenAta16:
Good catch, and you're right that this needed handling I'd already flagged as untested in the original PR description but hadn't actually closed. Pushed a fix in b2c4f61.

Replaced the backward-looking backslash-parity scan with a forward walk of the span as JSON string tokens from index 0 (`\` + next char as one atomic unit, expanded to 6 characters when that next char is `u`), returning the first token boundary at or after the candidate position. This handles both the 2-character and 6-character escape forms uniformly, using the same technique `json_handler.py`'s own tokenizer already uses for string-value boundaries.

Verified:
- Your exact repro now passes.
- Swept 400 prefix positions for the `\uXXXX` case: all valid JSON.
- Re-ran the prior 1200-case 2-character escape sweep from SEC-03: no regression.
- Re-ran SEC-01/02/03: all pass.
- Added `test_sec04_json_fallback_does_not_split_unicode_escape`, following SEC-03's prefix-sweep pattern, using your exact repro shape.

Same isolated-import verification method as the original PR (no Rust toolchain available here to run the actual pytest suite through the project's own tooling).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

JerrettDavis (approved):
The unicode-escape truncation blocker from my prior review is fixed. `_safe_truncate_boundary()` now treats `\uXXXX` as an atomic JSON escape when choosing cut points, and the new regression covers `json.dumps()`-escaped non-ASCII input.

Focused local verification: `uv run pytest tests/test_compression/test_universal.py::TestSecurityAuditRegressions::test_sec03_json_fallback_does_not_split_escape_pair tests/test_compression/test_universal.py::TestSecurityAuditRegressions::test_sec04_json_fallback_does_not_split_unicode_escape -q`.

ErenAta16:
Thanks for re-running both SEC-03 and SEC-04 against the fix, and for catching the six-character unicode escape gap in the first pass. Good to have this confirmed.

ErenAta16:
Closing - two weeks without review traffic. Withdrawing rather than leaving it open.

PR #2198, opened 2026-07-14
feat(dashboard): persist lifetime proxy metrics

jerviscui:
## Description

Persist bounded, aggregate-only Lifetime dashboard metrics across proxy restarts and expose them through a new `/stats-lifetime` endpoint. The change keeps session/runtime stats separate from durable lifetime stats, gates sensitive dashboard metadata for loopback or explicitly trusted dashboard clients, and updates the dashboard Lifetime view to consume the new endpoint.

Closes #2137

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring (no functional changes)

## Changes Made

- Added bounded persistent lifetime metrics state and wired proxy metric events into it.
- Added `/stats-lifetime` with sensitive project/persistence details gated behind dashboard metadata access checks.
- Extended loopback/dashboard metadata access policy for trusted dashboard client CIDRs without widening admin/debug endpoints.
- Reorganized dashboard session/lifetime presentation around runtime counters versus durable aggregates.
- Added focused tests for persistent aggregation, persistence, endpoint registration, loopback gating, trusted dashboard CIDRs, and recent request ordering.
- Fixed current Ruff/mypy issues in the lifetime metrics normalization code.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check`)
- [x] Type checking passes (`mypy`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
uv run --extra proxy --with pytest --with pytest-asyncio pytest tests/test_persistent_metrics.py tests/test_persistent_metrics_integration.py tests/test_persistent_metrics_persistence.py tests/test_proxy_loopback_gating.py tests/test_proxy_stats_recent_requests.py -q
53 passed, 1 warning

uvx ruff==0.15.17 check headroom/proxy/forwarded_headers.py headroom/proxy/loopback_guard.py headroom/proxy/persistent_metrics.py headroom/proxy/savings_tracker.py headroom/proxy/server.py tests/test_forwarded_headers.py tests/test_persistent_metrics.py tests/test_persistent_metrics_integration.py tests/test_persistent_metrics_persistence.py tests/test_proxy_loopback_gating.py tests/test_proxy_project_savings.py tests/test_proxy_stats_recent_requests.py
All checks passed!

[...]

PR #2266, opened 2026-07-16
fix(tests): unbreak two stale CI test failures on main

lifeodyssey:
## Description

Two tests are red on `main`'s CI (`test (4)` and `test (3)`, inherited by every open PR). Both assert against outdated expectations — the **production code is correct**; the fixes are test-only. Diagnosed each via `git` history.

1. **`test_smart_crusher_log_fallback_runs_for_valid_json`** — the test's `_try_ml_compressor` mock returned `len(content.split())`, but every production `_try_ml_compressor` return path reports `_estimate_tokens` (the JSON-aware counter). The mismatched token count made the Kompress-vs-Log decision (`if fallback_tokens < compressed_tokens`) take the Kompress branch, so the Log fallback the test asserts never ran. A realistic mock (`_estimate_tokens(content)`, matching a real Kompress no-op) makes `fallback == compressed`, so the Log fallback fires — the behavior the test's own docstring describes. If production had regressed, the corrected test would still fail; it passes, confirming production is correct.

2. **`test_dashboard_uses_cached_stats_and_lazy_history_feed_polling`** — `0537cbfd` ("feat(dashboard): persist lifetime proxy metrics", #2198) intentionally moved the CLI-filtering *lifetime* savings out of the inline Token-Usage row into a dedicated persisted **"Lifetime Saved"** card; the assertion still expected the old `… + ' Filtered (lifetime)'` string. Updated to the new label — which still uses the generic `cliFilteringLabel` binding, preserving the test's intent (the neighboring `"RTK Filtered" not in html` assertions still hold).

## Type of Change

- [x] Bug fix (test-only; unblocks red CI on `main`)

## Changes Made

- `tests/test_transforms_content_router.py`: mock `_try_ml_compressor` now returns `_estimate_tokens(content)` (already imported), mirroring production.
- `tests/test_proxy_dashboard_stats_cache.py`: assertion updated to the `#2198` "Lifetime Saved" label.

## Testing

- [x] Linting passes (`ruff`)
- [x] Reproduced both failures on clean `main` first, then verified the fixes

### Test Output

```text
$ pytest tests/test_transforms_content_router.py::test_smart_crusher_log_fallback_runs_for_valid_json \
         tests/test_proxy_dashboard_stats_cache.py::test_dashboard_uses_cached_stats_and_lazy_history_feed_polling
2 passed

$ pytest tests/test_transforms_content_router.py      # 57 passed
$ pytest tests/test_proxy_dashboard_stats_cache.py    # 14 passed, 1 skipped
$ ruff check <both files>                              # All checks passed!
```

## Real Behavior Proof

- Environment: macOS (Darwin), Python in a uv venv, `_core` rebuilt from `main`, branch `fix/ci-test-regressions` off `main` (`eac49656`).
- Exact command / steps: reproduced both failures on clean `main` (`assert 'log' in ['smart_crusher','kompress']` and the missing dashboard label), traced the divergence with `git log -S` / `git blame`, then applied the test-only fixes and re-ran the individual tests and their whole files.
- Observed result: both target tests pass; the full files pass (57 and 14+1); the router test passing proves production still fires the Log fallback (a real regression would keep it red). No product code changed.
- Not tested: N/A — test-only changes; the production behavior they assert is already correct on `main`.

## Review Readiness

## Checklist

- [x] I have added tests that prove my fix is effective — these ARE the tests; they now reflect production
- [ ] I have updated the CHANGELOG.md — N/A (test-only)

## Additional Notes

- Part of unbreaking `main`'s CI (the `lint` job and `test (1)` are addressed in separate PRs). After these land, the CI workflow should be green again for all open PRs.

### DIFF
--- a/tests/test_proxy_dashboard_stats_cache.py
+++ b/tests/test_proxy_dashboard_stats_cache.py
@@ -740,7 +740,10 @@ def test_dashboard_uses_cached_stats_and_lazy_history_feed_polling() -> None:
     assert "Lean-ctx" in html
     assert "Context Tool" in html
     assert "cliFilteringLabel + ' Filtered (this session)'" in html
-    assert "cliFilteringLabel + ' Filtered (lifetime)'" in html
+    # Lifetime CLI-filtering savings moved to the persisted "Lifetime Saved"
+    # card in the History view (#2198); it still uses the generic
+    # cliFilteringLabel binding rather than a hardcoded "RTK".
+    assert "(historyStats.cli_filtering?.label || cliFilteringLabel) + ' Lifetime Saved'" in html
 
 
 def test_dashboard_session_metrics_do_not_repeat_proxy_tokens_without_new_context() -> None:
--- a/tests/test_transforms_content_router.py
+++ b/tests/test_transforms_content_router.py
@@ -693,11 +693,14 @@ def test_smart_crusher_log_fallback_runs_for_valid_json(
 
     monkeypatch.setattr(router, "_get_smart_crusher", lambda: NoopSmartCrusher())
     monkeypatch.setattr(router, "_get_log_compressor", lambda: ShrinkingLogCompressor())
-    # Kompress no-op → Log fallback fires.
+    # Kompress no-op → returns the content unchanged with its estimated token
+    # count (mirroring the real ``_try_ml_compressor``, which always reports
+    # ``_estimate_tokens``). Because that equals the SmartCrusher result, the
+    # router sees no Kompress savings and the Log fallback fires.
     monkeypatch.setattr(
         router,
         "_try_ml_compressor",
-        lambda content, context, question=None: (content, len(content.split())),
+        lambda content, context, question=None: (content, _estimate_tokens(content)),
     )
 
     compressed, _compressed_tokens, strategy_chain = router._apply_strategy_to_content(


JerrettDavis (approved):
The two test updates are consistent with the current production behavior: the dashboard assertion now follows the lifetime card introduced by #2198, and the router mock now reports the same token estimator shape as the real no-op ML compressor path. I verified the dashboard target locally from the PR worktree; the router target is blocked in this Windows worktree by the local `_core` DLL import, but the PR's CI shard containing that test is green.

The remaining red checks on this PR are the separate inherited lint/order-dependent test failures addressed outside this test-only change.

chopratejas:
Thanks @lifeodyssey! Both stale failures here (`test_proxy_dashboard_stats_cache.py` and `test_transforms_content_router.py`) were fixed on main by #2306 (merged), which repaired all three red shards. Closing as superseded — appreciate you jumping on it.

JerrettDavis (approved):
Re-reviewed current head. The two test expectation updates still look targeted: the dashboard assertion follows the current persisted lifetime card behavior, and the content-router mock now matches production token accounting closely enough to exercise the intended log fallback.

The remaining red shard on this branch is the Anthropic transform cache-order failure addressed by #2265, not a regression in this PR's touched tests. Keeping this approved from code review.

PR #2270, opened 2026-07-16
fix(proxy/memory): don't crash the memory tool adapter on a null function/arguments

abhay-codes07:
## Description

The memory tool adapter crashes when an upstream response carries a tool call whose `function` or `arguments` field is explicitly `null`.

`_get_tool_name`, `_get_tool_id`, and `_get_tool_input` all read the nested function like this:

```python
str(tool_call.get("function", {}).get("name", ""))
tool_call.get("function", {}).get("arguments", "{}")
```

Two distinct crashes:

1. **Null `function`** → `AttributeError`. `dict.get("function", {})` only substitutes `{}` for a *missing* key. A present-but-null `{"id": "c1", "type": "function", "function": null}` (which upstreams and gateways emit for partial/streamed tool calls) makes the result `None`, and `None.get("name")` raises.
2. **Null `arguments`** → `TypeError`. `tool_call.get("function", {}).get("arguments", "{}")` returns `None` when `arguments` is null, and `json.loads(None)` raises `TypeError` — which the surrounding `except json.JSONDecodeError` does **not** catch.

Both parse the untrusted upstream response inside `handle_tool_calls`, so a single malformed tool call takes down memory tool handling. Notably `parse_tool_call` in `headroom/ccr/tool_injection.py` already catches the `json.loads(None)` `TypeError` with an explicit comment, so the null-arguments hazard is known in the codebase; this path just wasn't hardened.

## Fix

- Coalesce `function` / `functionCall` with `or {}` so a null value collapses to `{}`.
- Coalesce the arguments string with `or "{}"` and add `TypeError` to the `except`, so a null `arguments` yields `{}` instead of crashing.

Real tool calls parse exactly as before.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/memory_tool_adapter.py`: coalesce `function`/`functionCall` (`or {}`) in `_get_tool_name`/`_get_tool_id`/`_get_tool_input`; coalesce the arguments string (`or "{}"`) and catch `TypeError`.
- `tests/test_memory_tool_adapter_null_fields.py`: new tests for null function, null arguments, and that real calls still parse.

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/proxy/memory_tool_adapter.py tests/test_memory_tool_adapter_null_fields.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/memory_tool_adapter.py
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOM-kills this box (ML stack import), so I reproduced the parse helpers with a dependency-free script and left the full pytest to CI.
- Exact command / steps: ran `{"function": null}` and `{"function": {"arguments": null}}` (plus a real `memory_save` call) through the OLD and NEW `_get_tool_name`/`_get_tool_input` logic.
- Observed result: OLD raises `AttributeError` on the null function and `TypeError` on the null arguments; NEW returns `""`/`{}` for both and still parses the real call to `{"content": "hi"}`.
- Not tested: a live upstream emitting a null field; full local `pytest` deferred to CI (OOM).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because the full suite imports the ML stack, which I can't run here. The parse helpers read only their `tool_call` argument (no instance state), so the new test exercises them on a bare instance via `object.__new__` — it runs under the normal CI pytest job, and the standalone proof above corroborates it.


JerrettDavis (approved):
Reviewed the adapter helper changes for OpenAI/Gemini/generic tool-call parsing. Coalescing null `function`/`functionCall` values and null `arguments` prevents the two crash paths while preserving normal real-call parsing, and the helper tests cover both regressions plus the valid-call case.

Local verification in an isolated worktree:
- `uv run pytest tests/test_memory_tool_adapter_null_fields.py -q`
- `uv run ruff check headroom/proxy/memory_tool_adapter.py tests/test_memory_tool_adapter_null_fields.py`

The remaining red `test` shards look like the inherited suite failures already being addressed by pending test-fix PRs, not a regression here.

PR #2290, opened 2026-07-16
fix(memory): staleness-check Windows absolute-path references

abhay-codes07:
## Description

On Windows, `MemoryBudgetManager` never staleness-checks memories that reference absolute file paths, so dead memories are never pruned and keep getting injected.

`_is_stale` decides whether an `entity_ref` is a file path before checking whether that file still exists:

```python
for ref in memory.entity_refs:
    if ref.startswith("/") or ref.startswith("./"):
        ...
        if rel not in git_files and not Path(ref).exists():
            return True
```

The guard only recognises POSIX absolute (`/x`) and relative (`./x`) paths. A Windows absolute path (`C:\Users\me\proj\foo.py`) matches neither, so the whole existence check is skipped and the ref is treated as fresh — even after the file is deleted or renamed. The result: on Windows the file-reference staleness pruning is effectively dead, and stale memories pointing at removed files stay in the budget and keep being injected into context.

The codebase already knows how to spot a Windows path: `memory/traffic_learner.py` guards with `ref.startswith("/") or (len(ref) > 2 and ref[1] == ":")`. This path just wasn't updated to match.

## Fix

Add a small `_looks_like_path` helper that also recognises a drive-letter prefix, and use it in `_is_stale`:

```python
return ref.startswith("/") or ref.startswith("./") or (len(ref) > 2 and ref[1] == ":")
```

The subsequent `Path(ref).relative_to(...)` / `Path(ref).exists()` logic already works for Windows paths on Windows, so only the initial detection needed fixing.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/memory/budget.py`: add `_looks_like_path` (recognises `/`, `./`, and `C:` drive-letter prefixes) and use it in `_is_stale`.
- `tests/test_memory/test_budget.py`: new test that a Windows drive-letter ref to a missing file is flagged stale.

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/memory/budget.py tests/test_memory/test_budget.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/memory/budget.py
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOM-kills this box (ML stack import), so I reproduced the `_is_stale` path handling with a dependency-free script and left the full pytest to CI.
- Exact command / steps: ran a nonexistent Windows drive-letter ref (`C:\nonexistent\proj\foo.py`) through the OLD guard (`/` or `./`) and the NEW guard (also `ref[1] == ":"`), each followed by the existence check, plus POSIX/relative/plain-name controls.
- Observed result: OLD does not recognise the drive-letter ref, skips the check, and reports it fresh; NEW recognises it and the missing file flags it stale. POSIX, relative, and plain-name refs behave identically under both.
- Not tested: a live Windows memory-budget pass end-to-end; full local `pytest` deferred to CI (OOM).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because the full suite imports the ML stack, which I can't run here. The guard is a pure string check and `Path(ref).exists()` is `False` for that path on any OS, so the new test distinguishes old (not checked → fresh) from new (checked → stale) even on the Linux CI runner, alongside the standalone proof above. The backtick-path-in-content branch (line ~208) has the same POSIX-only assumption but I left it out of scope to keep this focused on the structured `entity_refs`; happy to follow up on it if you'd like.


JerrettDavis (changes requested):
Thanks for chasing the Windows gap here. I agree with the root cause, but the new path predicate is a little too broad for `entity_refs`, which are also used as arbitrary entity labels.

`len(ref) > 2 and ref[1] == ":"` treats any single-character namespaced entity like `A:B`, `X:foo`, etc. as a filesystem path. With `staleness_check_git=False` or no matching git entry, `Path(ref).exists()` will normally be false, so `_is_stale()` will prune a memory just because it references a valid non-path entity label. That is a correctness regression in the pruning path.

Please narrow the Windows path detection to actual Windows absolute paths, e.g. require an alphabetic drive letter plus a slash/backslash after the colon (`C:\...` / `C:/...`), or use a `PureWindowsPath`-based absolute-path check. A small negative test like `entity_refs=["A:B"]` remaining fresh would pin the distinction while keeping the new `C:\...` stale case covered.

abhay-codes07:
Agreed, that predicate was too loose. I tightened it to require a real Windows absolute drive path: an alphabetic drive letter, a colon, then a slash or backslash (`C:\...` / `C:/...`). So `A:B`, `X:foo`, `service:api` and similar entity labels are no longer treated as filesystem paths and won't be pruned. Added a negative test that `entity_refs=["A:B", "X:foo", "service:api"]` stays fresh, and kept the `C:\...` stale case covered.

JerrettDavis (approved):
The Windows path predicate is now narrow enough for real drive-absolute paths (`C:\...` / `C:/...`) without treating namespaced entity labels like `A:B` as files. The new negative test covers the regression I called out.

Local verification: `uv run pytest tests/test_memory/test_budget.py -q`.

Issue #231, opened 2026-04-22
`headroom learn` clobbers prior learnings on every run


gglucass:
## Summary

Every `headroom learn` run wholesale-replaces the marker-delimited block in `CLAUDE.md` / `MEMORY.md` with only what the current analyzer run surfaced. Learnings from prior runs that aren't re-derived by the current run are silently dropped. Running `headroom learn` a second time weeks later can erase most of the value accumulated from the first run.

## Reproduce

1. Run `headroom learn --project <path> --apply` → block written with recommendations A, B, C.
2. A week or two later, run it again. The analyzer only re-surfaces A and D (transcripts have rolled, or the LLM happens to weight things differently).
3. Result: the block now contains only A and D. B and C are gone with no warning.

## Root cause

In [`headroom/learn/writer.py`](https://github.com/chopratejas/headroom/blob/main/headroom/learn/writer.py):

- `_build_section()` (L69-87) builds the marker block from **only** the current run's `recommendations` list.
- `_merge_into_file()` (L90-97) does `_MARKER_PATTERN.sub(section, existing)` — a full substitution of the prior block.
- `analyzer.py` never reads the existing block, so prior recommendations aren't carried forward into the LLM input either.

Net effect: the marker block has no memory across runs.

## Why it matters

Users treat `headroom learn` as an additive-over-time system (the auto-generated header even says "~N tokens/session saved"), but the storage model is replace-on-write. Each run is all-or-nothing.

## Proposed fixes

**Option A (deterministic, minimal — preferred):** In `_merge_into_file`, parse prior recommendations out of the existing block, merge with the new ones (dedupe by `section` + content hash, prefer newer on conflict), then rebuild. No LLM changes. Fully self-contained in `writer.py`.

**Option B (flexible, LLM-aware):** In `analyzer.py`, load the existing block first and feed it to the LLM as "prior learnings — keep unless contradicted." More adaptive but depends on model behavior.

Happy to PR Option A — it's the safer, deterministic change, and it unblocks users who've already lost learnings to a re-run.

## Environment

- Repro'd on installed `headroom` package shipped by [headroom-desktop](https://github.com/gglucass/headroom-desktop) (invokes `headroom learn --project <path> --apply`).
- Writer code is byte-identical between the installed venv and `main` at HEAD.

PR #2275, opened 2026-07-16
fix(learn): filter ambient user-role scaffolding

superdiaodiao:
## Description

Fixes #2274.

Headroom Learn currently trusts `role=user` as sufficient preference provenance. Agent harnesses can transport ambient UI and orchestration context in user-role messages, and OpenAI Responses normalization also promotes missing roles to `user`. Correction-like text in those inputs can therefore become durable user preferences.

This change keeps preference learning fail-closed for known non-user sources while preserving genuine user corrections.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring only

## Changes Made

- Preserve missing OpenAI Responses roles as `unknown` instead of promoting them to `user`.
- Canonicalize user-role text before preference extraction.
- Remove proxy-appended `## Relevant Memories` suffixes from preference evidence.
- Reject strict ambient-only harness prefixes such as heartbeat, environment, workspace-instruction, delegation, and app-context envelopes.
- Apply the same guard in `on_messages` and `_extract_preferences` for defense in depth.
- Add regression coverage for system/developer/unknown roles, ambient-only user messages, memory-only messages, and mixed genuine-user-plus-memory input.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
149 passed, 1 warning
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

Focused test files:

```text
tests/test_memory/test_traffic_learner.py
tests/test_openai_responses_traffic_learner.py
```

## Real Behavior Proof

- Environment: macOS; Python 3.13; current Headroom main; direct invocation of the real `TrafficLearner` class, with no proxy or database mocks
- Exact command / steps: create `TrafficLearner(backend=None, min_evidence=1)`; feed system, developer, heartbeat user-role, and memory-only user-role messages; read `patterns_extracted`; feed a genuine user correction followed by a `## Relevant Memories` suffix; read `patterns_extracted` again
- Observed result: `ambient_patterns=0`, `after_user_patterns=1` — the ambient batch produced no preference evidence; the genuine correction produced one pattern, while the appended memory content did not become evidence
[...]

PR #2293, opened 2026-07-16
fix(learn): preserve stable traffic pattern items

### LINKED ISSUE #2294: [BUG] TrafficLearner flush replaces valid items in the same learned section
## Description

The writer carries forward prior Headroom Learn sections only when their heading is absent. When TrafficLearner later emits the same section with a narrower set of patterns, the entire prior section is replaced.

This is a narrower follow-up to #231: cross-section preservation works, but same-section item preservation does not.

## Reproduction

1. A prior Learned: preference section contains valid items A and B.
2. Raise the TrafficLearner evidence threshold, or let a later flush surface only B and C.
3. Flush again.
4. The section becomes B and C; valid previously written item A disappears.

## Expected behavior

TrafficLearner-managed stable facts and preferences should merge by a deterministic item identity, with new text winning for the same identity and exact legacy bullets deduplicated. Categories with explicit expiry or refinement, such as error recovery, must remain authoritative so stale items can still be removed.

## Actual behavior

Same-named sections are replaced as one unit.

## Proposed scope

Use an opt-in per-recommendation item merge mode rather than changing every Headroom Learn analyzer. Preserve prior TrafficLearner preference, environment, and architecture items; keep error-recovery replacement behavior.

Draft implementation: #2293

superdiaodiao:
## Description

Fixes #2294. Follow-up to #231.

TrafficLearner can re-surface a learned section with only a subset of previously written patterns. The writer currently replaces a same-named section as one unit, so a valid prior preference, environment fact, or architecture rule can disappear even though it was not contradicted.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add opt-in same-section markdown-item preservation to Recommendation.
- Render stable pattern IDs for TrafficLearner preference, environment, and architecture bullets.
- Prefer new text when a stable ID reappears and deduplicate legacy ID-less bullets by visible content.
- Keep error-recovery sections authoritative so expiry and refinement still remove stale entries.
- Leave every other Headroom Learn analyzer on the existing same-section replacement behavior.

## Testing

### Test Output

```text
170 focused writer and TrafficLearner tests passed
full tests/test_learn plus TrafficLearner run completed without failures
ruff check: passed
ruff format --check: passed
```

## Real Behavior Proof

- Environment: macOS, Python 3.13, current Headroom main.
- Exact command / steps: write a prior Learned: preference section containing a stable-ID item and a legacy ID-less item, merge a new recommendation with the same section and stable ID, then inspect the rebuilt marker block.
- Observed result: new text wins for the repeated stable ID, the unrelated prior bullet remains, and the legacy duplicate appears only once. Error-recovery recommendations retain replacement behavior.
- Not tested: full repository suite, live proxy traffic, or migration of arbitrary non-list markdown sections.

## Review Readiness

## Checklist

- [x] My code follows the project style guidelines
- [x] I have commented code where the behavior is not self-explanatory
- [x] I have made corresponding documentation changes
- [x] I have added tests that prove the fix is effective
- [x] New and existing focused unit tests pass locally
- [x] I have updated CHANGELOG.md if applicable

Not applicable.

## Additional Notes

Draft for maintainer feedback on the opt-in merge boundary and stable-ID representation. Already persisted incorrect rules are intentionally not deleted automatically.


superdiaodiao:
CI baseline note: the remaining failures in test shards 1, 3, and 4 reproduce unchanged on main in run https://github.com/headroomlabs-ai/headroom/actions/runs/29476342853 (Anthropic compaction max_chars, dashboard stats labels, and smart-crusher fallback assertions). The relevant shard 2, governance, lint, build, and security checks pass for this PR.

JerrettDavis (approved):
Reviewed the stable traffic-pattern item preservation. The writer keeps the existing authoritative-section behavior by default and only merges same-section list items when `Recommendation.preserve_prior_items` is set. For those opt-in sections, new stable IDs win, prior unrelated IDs are retained, and legacy ID-less duplicates are deduped by visible content. Error-recovery sections stay replacement-based so stale/expired recovery guidance can disappear.

TrafficLearner now emits stable pattern-id comments only for the preserved categories. The focused writer/TrafficLearner coverage described in the PR matches the risk area, and the CI failures shown on the branch are the known stale-main shards, not touched-code failures. Looks ready from source review.

JerrettDavis (approved):
Re-reviewed the latest branch after the merge-from-main commits. The writer still preserves prior same-section items only for recommendations that explicitly opt into `preserve_prior_items`; new stable IDs win over prior text, legacy ID-less duplicates are collapsed by visible content, and error-recovery sections stay replacement-based so stale guidance can be removed.

Verified locally: `uv run --extra dev pytest tests/test_learn/test_writer.py tests/test_memory/test_traffic_learner.py::TestPatternsToRecommendations -q` (34 passed), `uv run --extra dev ruff check headroom/learn/models.py headroom/learn/writer.py headroom/memory/traffic_learner.py tests/test_learn/test_writer.py tests/test_memory/test_traffic_learner.py`, and `uv run mypy headroom/learn/models.py headroom/learn/writer.py headroom/memory/traffic_learner.py --ignore-missing-imports --no-incremental`.

superdiaodiao:
Type checking verified locally with CI's exact command (mypy 1.20.2, `mypy headroom --ignore-missing-imports`): no issues in 504 source files.

@chopratejas @DevanshiVyas Readiness checklist is now complete and @JerrettDavis has approved — ready for your review whenever you have a chance.

superdiaodiao:
@chopratejas @DevanshiVyas Friendly follow-up now that #2275 has landed — this one is also ready for review: checklist complete, approved by @JerrettDavis, and the remaining CI shard failures reproduce unchanged on main. Would appreciate a look when you get a chance. Thanks!

JerrettDavis (changes requested):
The item-level merge prevents accidental subset replacement, but it also removes the only deletion path for preference/environment/architecture items. Once written, any prior bullet absent from every later recommendation is carried forever. The writer cannot distinguish “temporarily omitted from this batch” from “expired, disproven, or intentionally removed,” and the PR only keeps authoritative replacement for `ERROR_RECOVERY`. That can fossilize stale paths, tool availability, architecture rules, and user preferences in AGENTS/memory files indefinitely.

Please preserve prior items only while they are still active according to an authoritative learner identity/lifecycle signal. For example, pass the current active IDs (including patterns omitted only because of batching/top-N) plus explicit expired/tombstoned IDs into the merge, or persist enough metadata to apply a bounded expiry. Add a regression proving an unrelated active prior item survives a subset update *and* a stale/expired prior item is actually removed. Without a removal invariant this trades occasional loss for permanent incorrect steering.

superdiaodiao:
@JerrettDavis You're right, and thanks for catching it — as written the merge had no removal path at all. Fixed in 136dd9e.

`Recommendation` now carries `active_item_ids`, the learner's authoritative set of live pattern IDs, and `_merge_markdown_items` keeps a prior bullet only while its ID is still in that set. When the field is `None` the merge falls back to the previous union behaviour, so every other analyzer is unaffected.

On your parenthetical about batching and top-N — I traced the path before relying on it rather than assuming. `_collect_all_patterns` reads the full `memory.db` table with no `LIMIT`, `_drop_contradictions` returns early for anything that is not `ERROR_RECOVERY`, and the only slice in the file, `collapsed[:_ERROR_RECOVERY_SECTION_CAP]`, lives inside `_refine_error_recovery`. So for the three preserved categories the argument to `_patterns_to_recommendations` really is the complete active set, and the IDs are computed there, ahead of any sort.

That trace also turned up an inverted trap worth naming: `ERROR_RECOVERY` must *not* use this set. `_refine_error_recovery` retires rows itself through the 21-day floor and the Read-path recheck, so the pre-refine set is *wider* than the truth and would resurrect exactly what that pass just removed. `active_item_ids` is therefore only populated when `preserve_prior_items` is set, with a test pinning that.

Prior bullets carrying no ID are dropped once an authoritative set is present. The preserved categories now always emit an ID, so a still-active legacy bullet reappears tagged and collapses to one line through the existing visible-text dedupe; what actually gets removed can only be inactive. I also confirmed nothing else writes ID-less items into these sections — TrafficLearner is the only producer setting `preserve_prior_items`, and the analyzer-authored sections keep whole-section replacement.

Both regressions you asked for are in, plus an end-to-end one that feeds `_patterns_to_recommendations` output through `_merge_into_file` across two renders — proving the removal invariant only at the writer level felt thin.

The check that actually convinced me was the negative control. Reverting `writer.py` alone to its reviewed state, with everything else untouched, turns exactly the three deletion-invariant tests red:

```
FAILED tests/test_learn/test_writer.py::TestClaudeCodeWriter::test_active_ids_keep_unbatched_items_and_drop_expired_ones
FAILED tests/test_learn/test_writer.py::TestClaudeCodeWriter::test_active_ids_drop_untagged_legacy_items
FAILED tests/test_memory/test_traffic_learner.py::TestPatternsToRecommendations::test_pattern_dropped_from_live_set_is_removed_from_the_file
======================== 3 failed, 173 passed in 1.93s =========================
```

(The "unbatched item survives" half passes under the old code too — that was the union behaviour — so only the removal half goes red, as expected.)

Full run on the two files: 176 passed. `ruff check` clean, `mypy` clean across 504 source files.

JerrettDavis (changes requested):
The lifecycle signal still cannot remove a category once its final pattern expires. `_patterns_to_recommendations()` only emits recommendations for categories present in the current `patterns` list. When a category goes from one live item to zero, no recommendation for that heading reaches `_merge_recommendations()`, whose existing behavior carries forward headings absent from the new run wholesale. The old bullet therefore remains pinned forever—the exact lifecycle bug this revision is meant to fix. The end-to-end test keeps another item in the same category, so it does not exercise this transition.

Please propagate authoritative lifecycle information even for now-empty previously rendered categories (or otherwise let the writer remove absent traffic-pattern sections), and add an end-to-end regression where a category’s sole pattern disappears.

superdiaodiao:
You are right, and the end-to-end test I added did hide it: it kept a second item in the same category, so the run never went from one live item to zero for a heading. Fixed in d89568e.

**Where the removal now happens.** The gap is structural: once a category empties, `_patterns_to_recommendations()` emits nothing for that heading, so no amount of work inside `_merge_markdown_items` can reach it — the merge it would run in is exactly the one that never happens. So the lifecycle signal now also governs the carry path, which is the only code that still sees the heading:

```python
new_sections = {r.section for r in merged_new}
run_active_item_ids = _authoritative_item_ids(merged_new)
for prior_recommendation in prior:
    if prior_recommendation.section in new_sections:
        continue
    if run_active_item_ids is not None:
        pruned = _prune_carried_section(prior_recommendation.content, run_active_item_ids)
        if pruned is not None:
            if not pruned:
                continue  # heading goes with its last item
            prior_recommendation = replace(prior_recommendation, content=pruned)
    carried.append(prior_recommendation)
```

`_authoritative_item_ids` is the union of the `active_item_ids` the run publishes. A single producer emits one set per run covering all of its live items, so the union is that same set; taking a union rather than picking one keeps it least-deleting if two producers ever write the same file.

**Why the writer and not the producer.** The alternative was to have TrafficLearner emit an empty recommendation for every preserved category so the heading stays in `new_sections`. I dropped that: the learner does not know which headings were previously rendered, so it would have to emit all three on every run, which turns `if not recommendations: continue` in `flush_to_file` into a no-op and rewrites both the context and memory file on every flush even when only one category has content. Pruning in the writer needs no such speculation — a prior section that is id-tagged is by construction one of ours.

**What is deliberately not deleted.**

- *Untagged bullets in a carried section are kept.* This is the opposite of the same-section rule, on purpose. There, dropping an untagged prior is safe because a still-active one is re-emitted with an id by the same run and collapses by visible text. On the carry path nothing re-emits it, so deleting would discard content on no evidence. A section that is entirely untagged is left alone completely (`_prune_carried_section` returns `None`), which also keeps hand-written headings and non-list prose bodies out of scope.
- *A run with no lifecycle signal deletes nothing.* If every recommendation in a run has `active_item_ids=None` — an error-recovery-only render, or any pre-existing analyzer — the carry path behaves exactly as it did before. Silence is not read as expiry.

**One boundary I want to name rather than have you find.** If *every* pattern disappears, nothing is removed: `flush_to_file` returns early at `_collect_all_patterns()` / `_min_evidence` / `_drop_contradictions` before any writer runs. That is not a hole I could close here — with no patterns there is also no `_project_for_pattern` anchoring, so the learner does not know which files it would be rewriting. Removing a whole project's block wants an explicit tombstone or a project-scoped render, and belongs in its own PR if you want it.

## Verification

```
$ pytest tests/test_learn/test_writer.py tests/test_learn/test_section_lifecycle.py \
         tests/test_memory/test_traffic_learner.py -q
181 passed in 2.40s

$ ruff check headroom/learn/writer.py tests/test_learn/test_section_lifecycle.py
All checks passed!

$ ruff format --check headroom/learn/writer.py tests/test_learn/test_section_lifecycle.py
2 files already formatted

$ mypy headroom/learn/writer.py
```

The check worth reporting is the negative control. Reverting `writer.py` alone to its reviewed state, everything else untouched, turns exactly the three deletion assertions red and nothing else:

```
FAILED tests/test_learn/test_section_lifecycle.py::TestCarriedSectionLifecycle::test_absent_tracked_section_is_removed_once_its_items_expire
FAILED tests/test_learn/test_section_lifecycle.py::TestCarriedSectionLifecycle::test_absent_tracked_section_keeps_its_still_active_items
FAILED tests/test_learn/test_section_lifecycle.py::TestTrafficLearnerCategoryLifecycle::test_category_losing_its_last_pattern_drops_its_heading
3 failed, 2 passed in 0.26s
```

The two conservative cases — untracked sections carried untouched, and a run without a lifecycle signal deleting nothing — pass under both the old and the new writer, which is what makes them a guard against over-deletion rather than a restatement of the change.

New tests live in `tests/test_learn/test_section_lifecycle.py` rather than being appended to `test_writer.py`: they are one coherent invariant (a heading may not outlive its items) and the end-to-end case drives the learner, so a separate module reads better than splitting it across two files. Happy to fold them in if you would rather keep everything in `test_writer.py`.

JerrettDavis (approved):
Re-reviewed d89568e against the remaining lifecycle blocker. The writer now unions the run-wide active pattern IDs and applies that authority to carried sections, so a category that emits no recommendation after its sole item expires can lose both the tagged bullet and empty heading. Still-active tagged items survive subset/category omission, while untagged or non-list user content remains conservative. The end-to-end learner regression exercises the exact one-item-to-zero transition. Full CI is green.

superdiaodiao:
@JerrettDavis thanks for the re-review on the lifecycle fix.

This has been approved and mergeable since August 11, and #2295 landed on the 12th, so the branch is not waiting on anything from me. Could someone with merge rights take it? Happy to rebase if it has gone stale.

PR #2301, opened 2026-07-16
fix(proxy): key drift detector on conversations, not credentials; canonicalize drift hashes

### LINKED ISSUE #2300: [BUG] Rust drift detector: false `cache_drift_observed` on normal agentic traffic (credential-level session key + raw early-window hash)
## Description

The cache-bust drift detector in the Rust proxy (`crates/headroom-proxy/src/cache_stabilization/drift_detector.rs`, PR-E6) exists to warn operators about invisible cache busts. On the traffic it most needs to watch — interactive agentic clients such as Claude Code — it currently cannot tell drift from normal operation, for three compounding reasons:

1. **The session key stops at the credential.** `derive_session_key` returns `auth:{sha256(bearer)…}` / `apikey:{…}`. Claude Code sends the same OAuth bearer for every conversation (verified against live traffic), so *all concurrent conversations share one LRU slot*. Two interleaved conversations ping-pong the stored baseline: every switch emits a false `cache_drift_observed`, and when a *real* bust happens its `drift_dims` are computed against the wrong conversation's baseline.

2. **The `early_messages` axis hashes the raw bytes of the first-3-messages window.** Two normal behaviors flip that hash with no cache bust anywhere:
   - *Growth:* turn 1 sends 1 message, turn 2 sends 3 — the window literally gains elements, so the single hash differs. Appending messages never invalidates the provider's cached prefix: in the capture below, turn 2 read the entire 13,074-token turn-1 prefix from cache (`cache_read_input_tokens=13074`, writing only 91 new tokens) between the exact two requests whose window hash flips.
   - *`cache_control` relocation:* clients move the cache breakpoint to the newest block every turn. In captured live traffic, the turn-1 opener carries `cache_control` on its last block and returns on turn 2 **byte-identical except that marker**. Moving a breakpoint never invalidates a previously cached prefix, but the raw hash treats it as a structural change.

   Net effect: `cache_drift_observed` with `drift_dims=early_messages` fires at turn 2–3 of essentially **every** multi-turn conversation, even with zero interleaving.

3. **`x-headroom-session-id` is ignored.** The Python proxy honors this header as the priority-1 session identity for every session-sticky subsystem (prefix tracker, beta-header stickiness — see `docs/content/docs/configuration.mdx`, "Session Beta Header Tracking"). The Rust detector never reads it, so even a client that explicitly declares its conversation identity still gets conflated.

Because the detector cries wolf on nearly every turn, a real bust (system-prompt edit, history rewrite/compaction, tool-set change) is buried in noise — the exact failure mode the detector was built to expose. This is the Rust-telemetry sibling of #2085 (Python side, fixed by per-conversation lineage resolution in `prefix_tracker.py`): a credential identifies a workspace, not a conversation.

The detector is a pure observer, so nothing here affects forwarded bytes — the blast radius is operator-facing telemetry quality.

## To Reproduce

1. Run the Rust proxy and point Claude Code at it (`ANTHROPIC_BASE_URL=http://localhost:<port>`).
2. `claude -p "hello"` (conversation A, turn 1) → `cache_drift_first_request`.
3. `claude -p --resume <A-session> "next turn"` (conversation A, turn 2) → **false** `cache_drift_observed`, `drift_dims=early_messages` — the history only grew and the marker moved.
4. Interleave a second conversation `claude -p "different topic"` and resume A again → **false** `cache_drift_observed` on every switch with `drift_dims=early_messages` (plus `system` when the conversations differ in project context — the live capture below showed both dims), because the stored baseline belongs to the other conversation.

Unit-level repro against current `main` is in the code sample below (no network needed).

## Expected Behavior

- Turns of one conversation map to one session key; append-only growth and `cache_control` relocation produce **no** event.
- Different conversations map to different keys — no cross-conversation baseline ping-pong.
- A settled early message changing or disappearing (history rewrite/compaction), or `system`/`tools` changing, still emits `cache_drift_observed` (true positives preserved).
- `x-headroom-session-id` defines the session, matching the Python proxy's contract.

## Actual Behavior

- One `auth:{…}` key per credential → all conversations share one baseline slot; every conversation switch logs `cache_drift_observed`.
- `drift_dims=early_messages` fires on turn 2–3 of any lone multi-turn conversation (window growth + marker relocation).
- `x-headroom-session-id` has no effect on the Rust path.

## Code Sample

```rust
// Drop into crates/headroom-proxy/src/cache_stabilization/drift_detector.rs
// tests module on current main and run:
//   cargo test -p headroom-proxy --lib drift_detector

#[test]
fn different_conversations_on_one_credential_get_distinct_session_keys() {
    let mut headers = HeaderMap::new();
    headers.insert(
        axum::http::header::AUTHORIZATION,
        "Bearer shared-workspace-token".parse().unwrap(),
    );
    let addr: SocketAddr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(10, 0, 0, 9)), 4242);
    let conv_a = anthropic_body("sys", json!([]), vec!["conversation A opener"]);
    let conv_b = anthropic_body("sys", json!([]), vec!["conversation B opener"]);
    // FAILS on main: both are "auth:<same hash>"
    assert_ne!(
        derive_session_key(&headers, &addr /* , &conv_a, ApiKind::Anthropic */),
        derive_session_key(&headers, &addr /* , &conv_b, ApiKind::Anthropic */),
    );
}

#[test]
fn growth_and_marker_relocation_are_not_drift() {
    // Shape observed from live Claude Code traffic: turn 1 = one user
    // message with cache_control on its last block; turn 2 = the same
    // message byte-identical minus the marker, plus two appended messages.
    let turn1 = json!({
        "model": "m", "system": "s", "tools": [{"name": "bash"}],
        "messages": [
            {"role": "user", "content": [
                {"type": "text", "text": "opening prompt"},
                {"type": "text", "text": "project context",
                 "cache_control": {"type": "ephemeral"}},
            ]},
        ],
    });
    let turn2 = json!({
        "model": "m", "system": "s", "tools": [{"name": "bash"}],
        "messages": [
            {"role": "user", "content": [
                {"type": "text", "text": "opening prompt"},
                {"type": "text", "text": "project context"},
            ]},
            {"role": "assistant", "content": [{"type": "text", "text": "reply"}]},
            {"role": "user", "content": [
                {"type": "text", "text": "second prompt",
                 "cache_control": {"type": "ephemeral"}},
            ]},
        ],
    });
    let h1 = compute_structural_hash(&turn1, ApiKind::Anthropic);
    let h2 = compute_structural_hash(&turn2, ApiKind::Anthropic);
    // FAILS on main: drift_dims == "early_messages"
    assert_eq!(drift_dims(&h1, &h2), "");
}
```

## Error Output

Verbatim from running the code sample on `main` @ 718c8dc:

```text
---- cache_stabilization::drift_detector::tests::different_conversations_on_one_credential_get_distinct_session_keys stdout ----
assertion `left != right` failed
  left: "auth:5a1cfc47770958887c5d8feec5a818cf"
 right: "auth:5a1cfc47770958887c5d8feec5a818cf"

---- cache_stabilization::drift_detector::tests::growth_and_marker_relocation_are_not_drift stdout ----
assertion `left == right` failed
  left: "early_messages"
 right: ""
```

## Environment

- **Headroom version**: `main` @ 718c8dc (`crates/headroom-proxy`)
- **Python version**: n/a (Rust proxy path)
- **rustc**: 1.95.0
- **OS**: macOS 15 (arm64)
- **LLM Provider**: Anthropic (live Claude Code OAuth traffic)

## Additional Context

- Live-traffic verification: captured two Claude Code conversations × two turns each through a local proxy. All four requests carried the same `Authorization` bearer; in both conversations the turn-1 opener returned on turn 2 byte-identical except its relocated `cache_control` marker, and the raw first-3-window hash flips between the turns while the canonicalized per-message comparison is stable.
- Same defect class as #2085 (merged): session identity must be conversation-scoped, and cross-turn comparison must be canonical — the Python side's `_canonicalize_for_prefix_compare` already strips `cache_control` for exactly this reason.
- Fix PR incoming: session key honors `x-headroom-session-id` (hashed), then folds a conversation discriminator — `(model, canonicalized first message)`; provider prompt caches are per-model, so small-model sidecar calls reusing an opener stay separate — into credential/IP arms; hashes canonicalized (object keys sorted — the workspace's serde_json `preserve_order` would otherwise leave them key-order sensitive; `cache_control` stripped outside opaque tool payloads); `early_messages` becomes per-message hashes with prefix-aware comparison (growth benign; settled change/shrink still drift). Telemetry-only — forwarded bytes untouched. Known trade-off documented in-code: without the explicit header, first-message rewrites (compaction/rolling windows) re-key to a fresh session instead of reporting drift against the old baseline.

abcgco:
## Description

The Rust proxy's cache-bust drift detector (`crates/headroom-proxy/src/cache_stabilization/drift_detector.rs`, PR-E6) cannot currently tell drift from normal operation on interactive agentic traffic, so it warns on nearly every turn and a real bust drowns in the noise. Three compounding defects, all verified against live Claude Code traffic:

1. `derive_session_key` stops at the credential hash — Claude Code sends one OAuth bearer for every conversation, so all concurrent conversations share one LRU slot and every conversation switch logs a false `cache_drift_observed` (with `drift_dims` computed against the wrong conversation's baseline).
2. The `early_messages` axis hashes the raw first-3-messages window, so a lone conversation's normal growth (1 → 3 messages) and the client relocating its `cache_control` breakpoint to the newest block both fire a false `early_messages` drift at turn 2–3 of essentially every session.
3. `x-headroom-session-id` — the explicit session identity the Python proxy honors everywhere session-sticky state exists — is ignored on the Rust path.

This PR makes the detector's session identity conversation-scoped and its comparison canonical, the same shape as the merged Python-side fix for #2085 (`SessionTrackerStore.resolve_tracker` lineage resolution + `_canonicalize_for_prefix_compare`):

- **`derive_session_key`**: honors `x-headroom-session-id` first (hashed, like every other key input), then folds a conversation discriminator into the credential/network arms: a 16-hex-char SHA-256 fingerprint of `(model, canonicalized first message)`. Provider prompt caches are per-model, so a small-model sidecar call (title generation) that reuses a conversation's opener stays a separate session instead of false-drifting on `system`.
- **`canonicalize_for_hash`** on all axes and the discriminator: objects rebuilt with sorted keys (this workspace enables serde_json `preserve_order`, so a plain re-serialize would keep client wire order and leave the hashes key-order sensitive) and `cache_control` stripped outside opaque tool payloads (`input`/`arguments`/`json`/`input_schema` — mirroring the Python canonicalizer's `_OPAQUE_PAYLOAD_KEYS`, so a user field that happens to be *named* `cache_control` still counts as drift).
- **`early_messages`** becomes per-message hashes (`[Option<[u8; 32]>; 3]`) with a prefix-aware comparison: growing into the window is benign; a settled message changing or disappearing under a stable session key is still drift. `observe_drift` now gates the warning on drifted dimensions rather than raw hash inequality.

True positives are preserved (`system`/`tools` changes, in-place history rewrites under a pinned identity), and the detector remains a pure observer — no forwarded byte changes, `does_not_mutate_input` still pins that.

**Documented trade-off** (module doc + `conversation_discriminator` doc): without the explicit header, a client that rewrites its first message (history compaction, rolling-window truncation, Responses chained mode) re-keys to a fresh session — the rewrite surfaces as `cache_drift_first_request` rather than `cache_drift_observed` against the old baseline. That is deliberate: the credential-keyed alternative false-warned on every conversation switch, which buried those same events anyway. `x-headroom-session-id` pins the identity and reports rewrites as drift. Byte-identical openers on the same model under one credential still conflate (rare; documented).

Closes #2300

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `derive_session_key`: `x-headroom-session-id` (hashed) wins; credential/IP arms fold in `conversation_discriminator` — `(model, canonicalized first message)`, 16 hex chars
- New `canonicalize_for_hash`: sorted-key object rebuild + `cache_control` stripped outside `OPAQUE_PAYLOAD_KEYS`; applied to the `system`/`tools`/`early_messages` axes and the discriminator
- `StructuralHash.early_messages`: `[u8; 32]` → `[Option<[u8; 32]>; EARLY_MESSAGES_WINDOW]` per-message hashes; `drift_dims`/`early_window_drifted` implement the prefix-aware rule; `observe_drift` warns on non-empty dims instead of `!=`
- `conversation_messages` shape guard: bare-string message containers only count for the Responses `input` sugar
- Docs: module header (canonicalization, trade-off, honest cost), `conversation_discriminator` rationale + blind spots, `DRIFT_DETECTOR_CAPACITY` cardinality note (per-conversation keys, 163-byte entry), `structural_hash_log_prefix` hex-length fix
- Tests: 13 new unit tests (conversation separation, turn-growth key stability, explicit header priority, marker relocation + growth not drift, rewrite/shrink still drift, per-model separation, key-order neutrality, opaque-payload fields still count, Responses/Chat discriminator shapes, string-container gating)
- `CHANGELOG.md`: Unreleased → Fixed entry

## Testing

- [x] Unit tests pass (`cargo test -p headroom-proxy` — full crate: lib + integration suites)
- [x] Linting passes (`cargo clippy -p headroom-proxy --all-targets` — zero warnings; `cargo fmt --check` clean)
- [ ] Type checking passes (`mypy headroom`) — n/a, no Python files touched

### Test Output

```text
$ cargo test -p headroom-proxy --lib drift_detector
test result: ok. 27 passed; 0 failed; 0 ignored; 0 measured; 221 filtered out

$ cargo test -p headroom-proxy
(all suites) test result: ok. 248 passed (lib) + integration suites, 0 failed

$ cargo clippy -p headroom-proxy --all-targets
(no warnings)
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), rustc 1.95.0, repo @ 718c8dc + this branch
- Exact command / steps: captured two real Claude Code conversations × two turns through a local proxy (`ANTHROPIC_BASE_URL=http://localhost:8791 claude -p …` / `--resume …`), rebuilt the wire bodies, and replayed them through the real `derive_session_key` / `compute_structural_hash` / `drift_dims` in a local `cargo test` harness — before and after this change.
- Observed result: **before** — all four requests share one `auth:` key, and the raw early-window hash flips between turn 1 and turn 2 of the *same* conversation (false `early_messages` drift; interleaving also flips `system`). **After** — turn 1/turn 2 map to one stable key with `drift_dims == ""`, the two conversations map to distinct keys, and a rewritten/shrunk settled window still reports `early_messages`.
- Not tested: live OpenAI Chat/Responses traffic (shape-level unit tests only); log pipeline consumers (event names/fields unchanged).

## Review Readiness

## Checklist

n/a — log-only telemetry change.

## Additional Notes

- `StructuralHash` is `pub`, but the workspace has no external consumers (checked `sdk/`, `plugins/`, Python, docs) — the field-type change is contained to `proxy.rs` and the module tests. `[Option<[u8; 32]>; 3]` keeps `Copy` for the LRU and adds no dependency.
- LRU cardinality: keys moved per-credential → per-conversation; `DRIFT_DETECTOR_CAPACITY`'s comment now documents the working set, the ~250-byte entry, and the graceful eviction failure mode (repeated `cache_drift_first_request`, telemetry-only).
- Not in scope, noted for follow-up: keying Responses chained mode (`previous_response_id`) as a lineage; surfacing mid-history `role:"system"` insertions on the OpenAI Chat shape (pre-existing blind spot on all axes).


JerrettDavis (approved):
The drift-detector changes are coherent and well covered: session identity now tracks explicit `x-headroom-session-id` first, otherwise scopes credential/IP identities by a canonicalized `(model, first conversation message)` discriminator; hash canonicalization strips transport cache markers while preserving opaque tool payload data; and the early-window comparison correctly treats append-only growth as benign while preserving rewrite/shrink detection.

I verified the focused Rust coverage locally from the PR worktree:
- `cargo test -p headroom-proxy --lib drift_detector`
- `cargo test -p headroom-proxy --test integration_cache_drift`
- `cargo fmt --check --package headroom-proxy`
- `cargo check -p headroom-proxy`

The PR's Rust checks, lint, security, build, and most CI checks are green. The current red Python test shards are the same inherited main failures covered by the separate CI-unblocking PRs, not this Rust drift-detector change.

PR #2305, opened 2026-07-16
fix(proxy/openai): fix the Responses WS memory-relay control flow

abhay-codes07:
## Description

The OpenAI Responses WebSocket relay (Codex path with memory tools) has a control-flow bug that breaks memory-tool suppression and corrupts per-turn metrics.

The relay is a three-phase state machine — **Phase 1** buffers events until the first output item decides suppress-vs-passthrough, **Phase 2a** suppresses a memory-tool response and executes the tool transparently, **Phase 2b** passes a normal response through. Its intended behaviour is pinned by the reference implementation in `tests/test_ws_memory_relay.py`.

The real handler diverged from it. The Phase-1 "flush buffer + record metrics + reset" block was dedented to Phase-1's own level, so it ran on **every** upstream event instead of only on `response.completed`:

```python
if not decided:                      # Phase 1
    event_buffer.append(msg_str)
    if event_type == "response.output_item.added":
        ...
    elif event_type == "response.completed":
        decided = True
for buf in event_buffer:             # <-- runs on EVERY event (wrong indent)
    await websocket.send_text(buf)
event_buffer.clear()
await _record_ws_response_metrics()
_reset()
response_completed_seen = True
continue                             # <-- unconditional -> Phase 2 unreachable

# --- Phase 2a: Suppress mode ---    # dead code
if suppress_response:
    ...
```

Because `_reset()` ran every event (the per-response state is documented as "reset after each response.completed"):

- **Memory-tool suppression never took effect.** A memory tool call sets `suppress_response = True`, but `_reset()` immediately clears it, so the tool's `output_item` events were relayed straight to the Codex client instead of being suppressed and executed transparently — the memory feature leaks its internal tool traffic to the user.
- **Per-response metrics were recorded per-event.** `_record_ws_response_metrics()` fired once per upstream event instead of once per turn, massively over-counting a turn's tokens.
- **Phase 2a / 2b were unreachable dead code** (everything after the unconditional `continue`).

## Fix

Move the flush / record / reset into the `response.completed` branch inside Phase 1, and end Phase 1 with an unconditional `continue`. This makes the handler match the reference relay in `tests/test_ws_memory_relay.py`:

```python
if not decided:
    event_buffer.append(msg_str)
    if event_type == "response.output_item.added":
        ...
    elif event_type == "response.completed":
        decided = True
        for buf in event_buffer:
            await websocket.send_text(buf)
        event_buffer.clear()
        await _record_ws_response_metrics()
        _reset()
        response_completed_seen = True
    continue        # Phase 1 fully handled this event

if suppress_response:   # now reachable
    ...
```

Now suppression persists across the response, tools execute transparently, metrics are recorded once per turn, and the pass-through phase is reachable.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/openai.py`: scope the Phase-1 flush/record/reset to the `response.completed` branch; end Phase 1 with an unconditional `continue`.
- `tests/test_openai_ws_relay_structure.py`: new source-structure regression guard (the handler is inline in a WS async loop, so this reads the source — without importing the ML stack — and asserts the flush/record/reset stays nested inside `response.completed` and that the suppress phase remains reachable).

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/proxy/handlers/openai.py tests/test_openai_ws_relay_structure.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOM-kills this box (ML stack import), so I modeled the three-phase relay with a dependency-free script (mirroring the reference in `tests/test_ws_memory_relay.py`) and left the full pytest to CI.
- Exact command / steps: ran one response whose first output item is a `memory_search` function call (`response.created`, memory `output_item.added`, memory `output_item.done`, `response.completed`) through the OLD (flush+reset+continue at Phase-1 level) and NEW (flush/reset inside `response.completed`) control flow.
- Observed result: OLD relayed the memory-tool `output_item.done` to the client (leak) and called the metrics recorder 5 times (once per event); NEW relayed nothing (fully suppressed) and called the metrics recorder once. I also confirmed structurally that the flush/record/reset is now indented inside the `response.completed` branch (the source guard passes on the fixed file and would fail on the old one).
- Not tested: a live Codex WS session with memory tools end-to-end; full local `pytest` deferred to CI (OOM).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because the full suite imports the ML stack, which I can't run here. The relay logic is inline in a long WS async handler, so a direct behavioural unit test would need a full WS harness; `tests/test_ws_memory_relay.py` already pins the intended behaviour with a reference implementation (which this fix realigns the handler to), and the new source-structure guard prevents the indentation from drifting back. Longer term it would be worth extracting this state machine into a shared helper so the reference test exercises the real code path — happy to follow up on that if you'd like.


JerrettDavis (changes requested):
The relay control-flow fix may be right, and the test shard failures are the same unrelated baseline failures we have seen on current `main`:

- `test_l2_appends_transform_label`
- `test_dashboard_uses_cached_stats_and_lazy_history_feed_polling`
- `test_smart_crusher_log_fallback_runs_for_valid_json`

But the PR has a direct required lint failure in its own added file:

```text
ruff format --check .
Would reformat: tests/test_openai_ws_relay_structure.py
```

Please format that test file and push the result. I’m keeping this changes-requested while the branch is red on a touched-file formatting issue.

abhay-codes07:
Good catch, thanks. I only ran the formatter on the source file and missed the new test. Formatted `tests/test_openai_ws_relay_structure.py` with ruff 0.15.17 and pushed, so the lint job should be green now. The three failing test shards are the same baseline failures on current main that you flagged, unrelated to this change.

JerrettDavis (approved):
The previous blocker has been addressed: the touched test file is now formatted and the PR's lint/check suite is green. I also verified the structural regression test locally with `uv run pytest tests/test_openai_ws_relay_structure.py -q`.

JerrettDavis:
Exact-head follow-up at 7b33a0d4: this refreshed PR now has zero changed files against main. The reviewed WebSocket relay fix has been fully absorbed, so this branch should be closed as obsolete rather than merged.

JerrettDavis (approved):
Re-reviewed exact head 7b33a0d494c3594c47a96ab7434a92adce00b405. Its merge-base-to-head content diff is empty (GitHub reports zero changed files), so there is no remaining patch to merge or validate. Please close this PR as already integrated/superseded.

JerrettDavis:
Thanks for the work on the WebSocket memory relay. The current PR now has zero changed files and no remaining content diff to merge. I’m closing it as already absorbed; the earlier review discussion will stay available here.

PR #2307, opened 2026-07-16
fix(proxy/openai): forward compressed Codex WS /v1/responses frames

abhay-codes07:
## Description

Codex WebSocket `/v1/responses` frame compression never takes effect — every frame is forwarded uncompressed.

The per-frame handler compresses a frame inside a `try`, and on failure forwards the original bytes with a `compression_exception` reason. That failure `return` was dedented to the enclosing `try/except`'s own indentation, so it ran unconditionally after the block instead of only inside the `except`:

```python
try:
    (new_inner, modified, frame_transforms, frame_reason, ...) = await self._compress_...(...)
    ...
except Exception as _frame_err:
    ...
    _log_ws_passthrough("compression_exception", ...)
for t in frame_transforms:          # runs after try/except
    ...
return (                            # <-- ran on the SUCCESS path too
    raw_after_store,
    store_forced,
    "chatgpt_store_false" if store_forced else "compression_exception",
)
if not modified:                    # unreachable
    ...
if not isinstance(new_inner, dict): # unreachable
    ...
rewritten = json.dumps(new_inner)   # unreachable
...
return rewritten, True, frame_reason or "compressed"   # unreachable
```

On a successful compression the code fell straight into that `return`, forwarding `raw_after_store` (the uncompressed frame) with a `compression_exception` reason and **discarding the compressed result**. Everything after it — the `modified` check, the `store: false` rewrite, the frame re-serialization, the `tokens_saved` / `ws_frames_compressed` bookkeeping, and the final `return rewritten, True, "compressed"` — was unreachable dead code. So Codex WS frame compression saved nothing and `headroom_codex_ws_frames_compressed` stayed at zero. (It also risked a `NameError` on the exception path, since the following `for t in frame_transforms` ran even when the compression raised before binding `frame_transforms`.)

## Fix

Move the failure `return` inside the `except` handler, so it fires only on a compression error. On success the code now continues to the `modified` check, the rewrite, and the compressed-frame return:

```python
except Exception as _frame_err:
    ...
    _log_ws_passthrough("compression_exception", ...)
    return (
        raw_after_store,
        store_forced,
        "chatgpt_store_false" if store_forced else "compression_exception",
    )
# success path (now reachable)
for t in frame_transforms:
    ...
if not modified:
    ...
return rewritten, True, frame_reason or "compressed"
```

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/openai.py`: move the `compression_exception` `return` into the `except` block so the compression-success path is reachable.
- `tests/test_openai_frame_compression_reachable.py`: new AST-based regression guard asserting the `compression_exception` return lives inside an `except` handler (not as a sibling of the success path).

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/proxy/handlers/openai.py tests/test_openai_frame_compression_reachable.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOM-kills this box (ML stack import), so I found this with an AST scan for statements after a terminal statement, then modeled the frame handler with a dependency-free script and left the full pytest to CI.
- Exact command / steps: an AST walk over `handlers/openai.py` flagged an unreachable statement after the `return` at the frame path; I then ran a compress-succeeds case and a compress-raises case through the OLD (return dedented) and NEW (return inside except) control flow.
- Observed result: OLD returns `(raw, "compression_exception")` even when compression succeeded (compressed frame discarded); NEW returns `(compressed, "compressed")` on success and `(raw, "compression_exception")` only on failure. The AST scan reports no unreachable statement after the fix.
- Not tested: a live Codex WS `/v1/responses` session end-to-end; full local `pytest` deferred to CI (OOM).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because the full suite imports the ML stack, which I can't run here. The fix is inline in a long WS frame handler, so the added test is an AST-level guard (it reads the source without importing the ML stack) asserting the failure return stays scoped to the `except` handler; behaviour is additionally verified by the standalone proof above.

## Maintainer conflict refresh

Refreshed onto current `main` at `f6651d00`. Current `main` already contains the production WebSocket compression fix; this PR now contributes the AST regression test that prevents the unreachable-return bug from recurring. Validation: 33 targeted tests passed; Ruff check/format and `git diff --check` passed.


JerrettDavis (changes requested):
The control-flow fix is plausible, but this PR is currently red on the required lint job:

```text
ruff format --check .
Would reformat: tests/test_openai_frame_compression_reachable.py
```

Please format that test file and push the result. Keeping this changes-requested while lint is red; once the branch is formatted, the small indentation/control-flow change can be reviewed on a green baseline.

abhay-codes07:
Thanks, fixed. I ran the formatter on the handler but missed the new test file. Formatted `tests/test_openai_frame_compression_reachable.py` with ruff 0.15.17 and pushed, so lint should be green now.

JerrettDavis (approved):
The prior blocker was formatting in the touched regression test; the PR is now formatted and CI is green. The structural guard for the compression-exception return passes locally.

Local verification: `uv run pytest tests/test_openai_frame_compression_reachable.py -q`.

HulianBuligon:
I validated these two commits (`7b68dabd` + `eb36bfaf`) in a local `0.32.0`-based wheel against real Codex WebSocket `/v1/responses` traffic.

Sanitized clean-window result after restarting the proxy and opening fresh Codex sessions:

```text
input_tokens:       21,266
optimized_tokens:      818
tokens_saved:       20,448
max compression:    ~437 ms
WS timeouts:             0
quarantine activations:  0
leaked threads:          0
upstream errors:         0
```

The `headroom_codex_ws_frames_compressed` counter and token-savings metrics increased, confirming that the compressed-frame return is reachable and the rewritten frames are actually forwarded rather than discarded.

Test conditions / caveat: the local wheel also contained a separate tool-output protection change, and Kompress was disabled to isolate the stable SmartCrusher/schema-compaction path. Therefore this is end-to-end behavioral evidence for this PR's WS control-flow fix, not a benchmark of Kompress.

No credentials, attestation values, prompt contents, tool outputs, or private upstream names were captured in this proof.

JerrettDavis (approved):
Re-reviewed after the current-main refresh. The original WS implementation is now present on main, so this head correctly reduces to a focused structural regression test. The AST guard catches the historical dedented-return failure while avoiding heavyweight proxy imports; the branch is conflict-free, diff-clean, and CI is green.

JerrettDavis:
The underlying bug has now shipped in #2823 (`4ec416df`). Current `main` keeps the failure return inside the exception path, forwards successful later-frame compression results, and covers the behavior through the functional Codex WS lifecycle suite. After the conflict refresh, this PR contains only an additional AST-structure test for code that is already fixed and behaviorally tested upstream.

I think this can be closed as superseded rather than merging a second guard for the same regression. Thank you for isolating the unreachable-success-path defect so precisely; that diagnosis directly matches the merged fix.

PR #2313, opened 2026-07-16
fix(ccr): lowercase the hash in the streaming feedback recorders

abhay-codes07:
## Description

The streaming CCR feedback recorders look the `headroom_retrieve` hash up in the store with whatever casing the model produced, so an uppercase-echoed hash misses the (lowercase-keyed) store and its retrieval never reaches the TOIN learning loop.

`_record_ccr_feedback_from_response` (Anthropic) and `_record_ccr_feedback_from_openai_sse` (OpenAI chat stream) both pull the hash straight off the tool call and pass it to `store.retrieve`:

```python
hash_key = input_data.get("hash")
if not hash_key:
    continue
...
store.retrieve(hash_key)
```

The compression store keys every entry by a lowercase hash (`sha256(...).hexdigest()` and `explicit_hash.lower()` on write). A model that reproduces the marker hash in uppercase therefore misses the lookup, so the retrieval side effect that feeds TOIN (`store.retrieve → _log_retrieval → toin.record_retrieval`) never fires — the learning signal for that retrieval is silently dropped.

The request path was already fixed for this (`parse_tool_call` now returns the hash lowercased), but these two feedback recorders extract the hash themselves and were missed.

## Fix

Normalise the hash to lowercase before the lookup in both recorders:

```python
hash_key = hash_key.lower()
...
store.retrieve(hash_key)
```

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/streaming.py`: lowercase the extracted hash in `_record_ccr_feedback_from_response` and `_record_ccr_feedback_from_openai_sse` before `store.retrieve`.
- `tests/test_ccr_feedback_hash_case.py`: new tests that both recorders query the store with the lowercase hash when the model echoes it uppercase.

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/proxy/handlers/streaming.py tests/test_ccr_feedback_hash_case.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/handlers/streaming.py
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOM-kills this box (ML stack import), so I reproduced the recorder logic against a lowercase-keyed fake store with a dependency-free script and left the full pytest to CI.
- Exact command / steps: fed a `headroom_retrieve` tool call whose hash is the uppercase form of a stored (lowercase) key through the OLD (verbatim) and NEW (`.lower()`) recorder logic.
- Observed result: OLD queries the store with the uppercase hash → miss (no feedback); NEW queries with the lowercase hash → hit (feedback recorded).
- Not tested: a live streaming retrieval with an uppercase echo end-to-end; full local `pytest` deferred to CI (OOM).

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because the full suite imports the ML stack, which I can't run here. The recorder methods use no instance state, so the new tests exercise them on a bare `object.__new__(StreamingMixin)` with the compression store monkeypatched to a spy; they run under the normal CI pytest job, and the standalone proof above corroborates them. This completes the lowercase-hash fix that already landed for the request path (`parse_tool_call`).


JerrettDavis (changes requested):
The lowercase fix is directionally right, but the new normalization assumes the model-provided `hash` is always a string. Both recorders parse tool-call payloads directly from model output; with a malformed non-string truthy value, for example `{"hash": 123}`, `hash_key = hash_key.lower()` now raises before the existing guarded `store.retrieve()` block. In the Anthropic streaming path, `_record_ccr_feedback_from_response(...)` is called directly after stream parsing, so malformed feedback payloads can escape the recorder instead of being ignored like other bad tool-call data. Please either require `isinstance(hash_key, str)` before lowercasing or convert/skip deliberately, and add regression coverage for a non-string hash in both recorder paths.

abhay-codes07:
Good catch, you're right that the payload is untrusted model output. I changed both recorders to guard with `isinstance(hash_key, str)` before lowercasing, so a non-string hash like `{"hash": 123}` is skipped the same way other malformed tool-call data is, instead of raising. Added regression tests for a non-string hash in both the Anthropic and OpenAI-stream recorder paths. @JerrettDavis

JerrettDavis (approved):
The non-string hash blocker from my prior review is fixed. Both CCR feedback recorder paths now require a string before lowercasing, and the new regression coverage covers malformed non-string hashes for Anthropic response feedback and OpenAI SSE feedback. GitHub CI is green on the current head.

Local note: I inspected the diff directly; a fresh local test run in this worktree stopped at collection because the minimal venv was missing `fastapi`, while the PR's CI environment includes the required dependency and passes.

PR #2254, opened 2026-07-15
fix(proxy): preserve signed Anthropic thinking blocks on outbound re-serialize

axelray-dev:
## Description

When multi-turn Anthropic requests include signed `thinking` or `redacted_thinking` blocks in conversation history, the proxy re-serializes the body through `serialize_body_canonical` whenever `body_mutated` is true. That re-encode changes the byte representation of signed blocks and upstream rejects the turn with 400 "blocks cannot be modified".

This detects those content blocks and, when original request bytes are available, forwards them byte-for-byte instead of re-encoding. That matches the preferred option from the issue and mirrors the existing Agno skip for thinking-bearing histories.

Closes #2251

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Add `has_signed_thinking_blocks()` in `headroom/proxy/body_forwarding.py`
- Prefer original-byte passthrough in `select_outbound_body` when signed thinking blocks are present and original bytes exist
- Unit tests for thinking and redacted_thinking passthrough, missing-original canonical fallback, legacy override, and unchanged non-thinking behavior

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [ ] Manual testing performed

### Test Output

```text
uv run pytest tests/test_proxy_byte_faithful_forwarding.py -q --tb=short
# 43 passed, 1 skipped

uv run ruff check / format on touched files
# passed
```

## Real Behavior Proof

- Environment: unit-level body forwarding with multi-turn Anthropic-shaped payloads containing signed `thinking` / `redacted_thinking` blocks
- Exact command / steps: focused pytest suite above
- Observed result: with `body_mutated=True` and original bytes present, outbound source is `passthrough` and content equals original bytes; without original bytes, behavior remains canonical; non-thinking mutated bodies still use canonical
- Not tested: full `headroom wrap claude` multi-turn session against Anthropic / Claude Code (no live Claude credentials here)

## Review Readiness

- [x] I have performed a self-review
- [x] This PR is ready for human review

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
[...]

PR #2315, opened 2026-07-16
fix(proxy): preserve signed extended-thinking blocks during byte-faithful forwarding

### LINKED ISSUE #2251: [BUG] Proxy canonical re-serialization corrupts signed extended-thinking blocks → 400 “blocks cannot be modified” on multi-turn Claude Code
Description
The proxy re-serializes every outbound request body canonically (serialize_body_canonical). This changes the byte representation of the already-signed assistant thinking block carried in conversation history, so the signature no longer matches and the upstream rejects the entire request. Any multi-turn Claude Code session with extended thinking enabled fails on every turn after the first.
Root cause: the Anthropic request path has no thinking-block awareness. Any body mutation — memory injection, image/last-turn compression, or even a no-op canonical re-encode — re-serializes the whole body and invalidates signed thinking/redacted_thinking blocks. The only thinking-aware skip in the codebase lives in the agno integration (integrations/agno/model.py), which the proxy/wrap path never uses.
Likely the same mechanism as #1042 (canonical re-serialization busting the prefix cache) — different symptom, same serialize_body_canonical.
Environment
	•	headroom-ai 0.31.0 (extras: proxy,code,ml), Python 3.14, macOS
	•	Claude Code 2.1.210, model claude-sonnet-4-6
	•	Client betas: interleaved-thinking-2025-05-14, redact-thinking-2026-02-12, thinking-token-count-2026-05-13 (Claude Code sends these by default)
	•	Upstream: Anthropic-compatible gateway (reproduces logically against api.anthropic.com too — the signature check is Anthropic’s, not gateway-specific)
To Reproduce
	1.	Install headroom with pip install 'headroom-ai[proxy,code,ml]==0.31.0'
	2.	Run headroom wrap claude with extended thinking on (Claude Code default)
	3.	Send a first message — it succeeds and produces an assistant thinking block in history
	4.	Send a second message in the same session — see error
Expected Behavior
The proxy forwards the request with the signed thinking/redacted_thinking blocks intact, and the multi-turn session continues normally.
Actual Behavior
Every turn after the first fails with:
API Error: 400 messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response.

Logs:
event=outbound_request forwarder=anthropic_messages path=/v1/messages
body_bytes=271106 body_mutated=true
mutation_reasons=structural_diff_vs_original source=canonical

UPSTREAM_ERROR status=400 error_type=invalid_request_error
error_msg='messages.3.content.4: `thinking` ... cannot be modified ...'
model=claude-sonnet-4-6 compressed=yes
transforms=['router:tool_search_deferral:13tools:19346tok']
original_tokens=47460 optimized_tokens=47460 message_count=5

Note original_tokens == optimized_tokens (47460) — no compression actually happened this turn, yet body_mutated=true source=canonical. The canonical re-encode alone breaks the block; it does not require any transform to fire.
Code Sample
This reproduces through the CLI wrap path rather than the Python client:

# Minimal repro — no Python client involved
pip install 'headroom-ai[proxy,code,ml]==0.31.0'
headroom wrap claude          # extended thinking on (Claude Code default)

# turn 1 -> 200 OK, assistant returns a signed `thinking` block
# turn 2 -> 400 invalid_request_error: `thinking` blocks cannot be modified

# Thinking-safe but disables all compression:
HEADROOM_OPTIMIZE=0 headroom wrap claude

Suggested Fix
When the request contains signed thinking/redacted_thinking blocks, either:
	•	(a) forward the original bytes untouched for that request, or
	•	(b) preserve those blocks byte-for-byte during re-serialization — the way the agno integration already detects and skips them.
Workarounds (all coarse)
	•	Disable extended thinking for wrapped sessions (works; keeps compression). ← what fixed it locally.
	•	Run thinking-heavy work with no proxy.
	•	HEADROOM_OPTIMIZE=0 (thinking-safe but zero compression).

lennney:
## Description

The proxy re-serializes every outbound request body canonically (`serialize_body_canonical` in `body_forwarding.py`). When conversation history contains assistant messages with signed `thinking`/`redacted_thinking` blocks from Claude Code's extended thinking feature, the re-serialization changes the byte representation of these blocks, invalidating Anthropic's server-side cryptographic signature. Every multi-turn Claude Code session with extended thinking enabled fails on turn 2+ with `400: "thinking blocks cannot be modified"`.

This fix detects extended-thinking blocks in the request body **before** any transform runs, and forces byte-preserving passthrough of the original client bytes so the signature is never invalidated.

Closes #2251

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/body_forwarding.py` — `select_outbound_body()` and `prepare_outbound_body_bytes()` gain a `force_preserve_original: bool = False` parameter. When True and original bytes are available, the original client bytes are forwarded verbatim regardless of `body_mutated`.
- `headroom/proxy/handlers/anthropic.py` — Added `_contains_thinking_blocks()` helper that recursively scans message content arrays (including nested `tool_result.content`) for `thinking`/`redacted_thinking` blocks. Detection runs before any transform. When blocks are present: structural diff is skipped (avoiding false-positive `structural_diff_vs_original` from tool reordering), and `force_preserve_original=True` is threaded through to the outbound body decision.
- `headroom/proxy/handlers/streaming.py` — Threads `force_preserve_original` through `_stream_response()` → `_stream_response_inner()` → `prepare_outbound_body_bytes()`.
- `headroom/proxy/server.py` — Threads `force_preserve_original` through `_retry_request()` → `prepare_outbound_body_bytes()`.
- `tests/test_proxy_byte_faithful_forwarding.py` — 42 new tests across detection, force_preserve_original, and adversarial/fuzz categories.

## Testing

- [x] Formatting passes (`ruff format --check .`)

### Test Output

```text
$ uv run pytest tests/test_proxy_byte_faithful_forwarding.py -q --no-header
82 passed in 5.48s
```

## Real Behavior Proof

- Environment: Linux 6.8.0-124-generic, Python 3.12.3, headroom main @ 26b43f64
- Exact command / steps: `uv run ruff check . && uv run pytest tests/test_proxy_byte_faithful_forwarding.py -q --tb=short` → 0 lint errors, 82 passed
- Observed result: 42 new tests — 20 structural (detection, passthrough contract, structural-diff surrogate) + 22 adversarial/fuzz (spoofed types, deep recursion, 1000-element lists, binary bytes, every mutation reason). All 40 pre-existing tests continue passing.
- Not tested: Live Claude Code session with extended thinking (requires Claude Code Desktop with extended thinking enabled). The fix is verifiable by unit-test-level inspection — `force_preserve_original=True` causes `select_outbound_body` to return `original_body_bytes` directly regardless of mutation state.

## Review Readiness

### DIFF
--- a/headroom/_subprocess.py
+++ b/headroom/_subprocess.py
@@ -20,7 +20,7 @@ def pid_alive(pid: int) -> bool:
     try:
         import psutil  # type: ignore[import-untyped]  # optional dep, already used elsewhere
 
-        return bool(psutil.pid_exists(pid))
+        return bool(psutil.pid_exists(pid))  # type: ignore[no-any-return]
     except Exception:
         pass
     try:
--- a/headroom/proxy/body_forwarding.py
+++ b/headroom/proxy/body_forwarding.py
@@ -82,8 +82,20 @@ def select_outbound_body(
     original_body_bytes: bytes | None,
     body_mutated: bool,
     forwarder_mode: PythonForwarderMode | None = None,
+    force_preserve_original: bool = False,
 ) -> OutboundBody:
-    """Select the exact bytes to forward upstream."""
+    """Select the exact bytes to forward upstream.
+
+    When ``force_preserve_original`` is ``True`` and ``original_body_bytes``
+    is available, the original client bytes are forwarded verbatim regardless
+    of ``body_mutated``. This is used when the request contains signed content
+    (e.g. Anthropic extended-thinking ``thinking``/``redacted_thinking`` blocks)
+    whose cryptographic signature is bound to the exact original bytes — any
+    re-serialization would invalidate it.
+    """
+    if force_preserve_original and original_body_bytes is not None:
+        return OutboundBody(content=original_body_bytes, source="passthrough")
+
     mode = forwarder_mode if forwarder_mode is not None else get_python_forwarder_mode()
     if mode == "legacy_json_kwarg":
         content = json.dumps(body, separators=(", ", ": "), ensure_ascii=True).encode("utf-8")
@@ -100,6 +112,7 @@ def prepare_outbound_body_bytes(
     original_body_bytes: bytes | None,
     body_mutated: bool,
     forwarder_mode: PythonForwarderMode | None = None,
+    force_preserve_original: bool = False,
 ) -> tuple[bytes, OutboundBodySource]:
     """Compatibility tuple wrapper around :func:`select_outbound_body`."""
     outbound = select_outbound_body(
@@ -107,5 +120,6 @@ def prepare_outbound_body_bytes(
         original_body_bytes=original_body_bytes,
         body_mutated=body_mutated,
         forwarder_mode=forwarder_mode,
+        force_preserve_original=force_preserve_original,
     )
     return outbound.content, outbound.source
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -68,6 +68,40 @@ def _strip_index_from_content_blocks(content: Any) -> None:
             _strip_index_from_content_blocks(block.get("content"))
 
 
+def _contains_thinking_blocks(messages: list[dict]) -> bool:
+    """Check if any message contains ``thinking``/``redacted_thinking`` blocks.
+
+    These blocks carry an Anthropic server-side cryptographic signature bound
+    to their exact byte representation. Any re-serialization invalidates the
+    signature and causes the upstream to reject the request with 400.
+    Recurses into ``tool_result`` nested content arrays (which may appear
+    in user messages when the client echoes back assistant tool-call results).
+    """
+    for msg in messages:
+        if not isinstance(msg, dict):
+            continue
+        content = msg.get("content")
+        if isinstance(content, list):
+            if _has_thinking_blocks_in_list(content):
+                return True
+    return False
+
+
+def _has_thinking_blocks_in_list(blocks: list) -> bool:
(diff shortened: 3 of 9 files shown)


JerrettDavis (changes requested):
This is a valid bug area, but I would not merge this branch in its current shape.

`pyproject.toml:76` and `pyproject.toml:223` raise the `mcp` lower bound from `>=1.0.0` to `>=1.27.2`, and the PR carries the resulting large `uv.lock` churn. That is unrelated to preserving Anthropic signed thinking blocks and materially changes dependency resolution for both the proxy and MCP extras. Please drop the dependency/lockfile churn from this PR unless there is a separate, documented MCP compatibility reason.

Also, the branch is currently `DIRTY` against `main` and overlaps with #2254, which solves the same signed-thinking forwarding issue with a much smaller body-forwarding change and is already approved after a current-main update. If you want to continue this implementation, please rebase on current `main`, remove the unrelated dependency changes, and make clear what behavior remains that #2254 does not cover.

I did not run tests for this branch because it is draft/conflicted and the dependency diff needs to be removed first.

lennney:
Closing #2315 in favor of #2254 by @axelray-dev — that PR solves the same signed-thinking forwarding issue with a focused body_forwarding-only change and is already APPROVED.

Thanks @JerrettDavis for the review. The delta here (streaming passthrough, structural-diff skip, nested tool_result recursion) may resurface if turn-2 extended-thinking failures appear in streaming paths, but #2254's approach covers the core case.

PR #2320, opened 2026-07-16
fix(transforms/kompress-remote): keep compress fail-open on malformed 200

abhay-codes07:
## Description

`RemoteKompressCompressor` (the opt-in `HEADROOM_KOMPRESS_ENDPOINT` remote compression client) documents a fail-open contract in its own docstring:

> Fails OPEN: any network/HTTP error returns the content verbatim so a flaky endpoint degrades compression rather than breaking the proxy.

But only the network call and the `compressed` field check actually run inside the fail-open guard. The metadata coercions run **after** the `except`, outside it:

```python
try:
    resp = self._client.post(...)
    resp.raise_for_status()
    data = resp.json()
    compressed = data["compressed"]
    if not isinstance(compressed, str):
        raise TypeError("...")
except Exception as e:  # fail OPEN
    logger.warning("Remote Kompress failed (%s); passing through", e)
    return self._passthrough(content, n_words)

result = KompressResult(
    compressed=compressed,
    original=content,
    original_tokens=int(data.get("original_tokens", n_words)),
    compressed_tokens=int(data.get("compressed_tokens", len(compressed.split()))),
    compression_ratio=float(data.get("compression_ratio", 1.0)),   # <-- outside the guard
    model_used=str(data.get("model_used", self.config.model_id)),
)
```

So a hosted `/compress` endpoint that returns a 200 with a valid `compressed` string but a malformed metadata field escapes the guard and raises out of `compress`, breaking the proxy request instead of passing through. The most realistic trigger is an explicit JSON `null`: `data.get("compression_ratio", 1.0)` returns `None` for a **present** key (the default only applies to a missing key), and `float(None)` raises `TypeError`. A non-numeric string like `"original_tokens": "lots"` raises `ValueError` the same way. Since the whole point of the flag is to support arbitrary self-hosted endpoints, a slightly-off but well-meaning endpoint (sending `null` for a field it could not compute) takes down the request path this class exists to protect.

## Fix

Move the response parsing (the `KompressResult` construction with its `int`/`float`/`str` coercions) inside the fail-open `try`, so any malformed field degrades to verbatim passthrough like every other bad-response case:

```python
try:
    ...
    compressed = data["compressed"]
    if not isinstance(compressed, str):
        raise TypeError("...")
    result = KompressResult(
        compressed=compressed,
        original=content,
        original_tokens=int(data.get("original_tokens", n_words)),
        compressed_tokens=int(data.get("compressed_tokens", len(compressed.split()))),
        compression_ratio=float(data.get("compression_ratio", 1.0)),
        model_used=str(data.get("model_used", self.config.model_id)),
    )
except Exception as e:  # fail OPEN
    logger.warning("Remote Kompress failed (%s); passing through", e)
    return self._passthrough(content, n_words)
```

No behavior change on a well-formed response; only the malformed-200 path changes (raise to passthrough).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/kompress_remote.py`: move the `KompressResult` construction and its field coercions inside the fail-open `try`.
- `tests/test_transforms/test_kompress_remote.py`: add `test_remote_kompress_null_numeric_field_fails_open` (explicit JSON `null`) and `test_remote_kompress_non_numeric_field_fails_open` (non-numeric string), both asserting verbatim passthrough.

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/transforms/kompress_remote.py tests/test_transforms/test_kompress_remote.py
All checks passed!
$ uvx ruff@0.15.17 format --check headroom/transforms/kompress_remote.py tests/test_transforms/test_kompress_remote.py
2 files already formatted
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/transforms/kompress_remote.py
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOMs this box (ML-stack import), so I reproduced the control flow with a dependency-free script and left the full pytest to CI.
- Exact command / steps: modeled the OLD (coercions outside the `try`) and NEW (inside the `try`) parsing against a 200 body `{"compressed": "short result", "compression_ratio": null}` and against a well-formed body.
- Observed result: OLD raised `TypeError` on the null field (proxy request breaks); NEW returned passthrough; a well-formed body still compressed under NEW. The added tests assert both malformed cases (`null` and non-numeric string) return the original content with `compression_ratio == 1.0`.
- Not tested: a live remote Kompress endpoint; the added tests drive `RemoteKompressCompressor` through an `httpx.MockTransport`, matching the existing test harness in this file.

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because a local pytest run imports the ML stack and OOMs this box; the added tests reuse the existing `httpx.MockTransport` harness in `test_kompress_remote.py` and run under the normal CI pytest job, and the behavior is corroborated by the standalone proof above.


JerrettDavis (approved):
Reviewed the refreshed remote Kompress fail-open fix after merging current main and removing the manual changelog entry in e639b837. Moving the KompressResult construction inside the existing fail-open guard is the right scope: malformed metadata in a 200 response now degrades to passthrough just like network errors, bad status, invalid JSON, or a non-string compressed field. Well-formed responses keep the same behavior.

Validated locally:
- uv run --extra proxy --with pytest --with pytest-asyncio python -m pytest tests`test_transforms`test_kompress_remote.py -q
- uvx ruff@0.15.17 check headroom`transforms`kompress_remote.py tests`test_transforms`test_kompress_remote.py
- git diff --check

Issue #1959, opened 2026-07-10
[FEATURE] Make the dashboard's loopback-only gate for sensitive stats configurable for trusted reverse-proxy deployments


suimong:
## Problem Statement

The `/stats` endpoint gates its sensitive sub-payloads — recent_requests, request_logs, and the config block — behind a hard-coded loopback-only check, with no way to opt in for a trusted deployment. `include_sensitive = _request_is_loopback(request)` requires both the TCP peer (request.client.host) to be loopback and the Host: header to name a loopback address; otherwise those keys are popped. uvicorn also runs with `proxy_headers=False`, so `X-Forwarded-For` is intentionally ignored and the guard always sees the real socket peer.

The consequence: when headroom runs as a long-lived service behind a trusted reverse proxy addressed by hostname, both gates fail by construction (peer = the proxy's socket, Host = the public service name). The aggregate counters render fine, but the useful live panel — Recent Requests across all projects — is permanently blank. There's currently no supported way to say "this front proxy is trusted."

I want to be clear I understand why the gate exists and am not asking to weaken the default: the two-gate model is a deliberate DNS-rebinding defense (a peer-IP check alone is insufficient because a malicious page can make a browser hit `127.0.0.1:<port>`, and it's the loopback-Host requirement that unmasks the rebind), and the payload is genuinely sensitive (prompt/tool-path snippets, model names, upstream URLs). The ask is an explicit, operator-defined extension of that boundary, off by default.

## Proposed Solution

An opt-in, default-off trusted-proxy configuration that preserves the existing two-gate model but lets the operator define the boundary:

- `HEADROOM_DASHBOARD_TRUSTED_PROXIES` — CIDR/IP allowlist. A forwarded origin is considered only when the real peer is in this set.
- `HEADROOM_DASHBOARD_TRUSTED_HOSTS` — a Host: allowlist that extends the loopback set (e.g. the service hostname), so the anti-rebinding gate remains a real allowlist rather than being disabled.

[...]

PR #1057, opened 2026-06-16
feat(proxy): Rust-native savings stats + dashboard (Phase H)

ysheikh2:
## Description

Re-implements \`savings_tracker.py\` in Rust as \`crates/headroom-proxy/src/observability/stats.rs\` — the Phase H plan's \"savings tracker re-implemented in Rust as part of \`observability/\`\" — and serves it from the Rust proxy. Phase H deletes \`server.py\` (which serves the \`/stats\` JSON and \`/dashboard\`) but never named a Rust replacement for those two routes; this fills that gap. Because the Rust proxy fronts Anthropic, OpenAI, Bedrock and Vertex in one process, savings are attributed per provider/model and unified across all of them. An optional \`--upstream-stats-url\` flag federates stats from the Python proxy (or any sidecar) so Copilot, Vertex, and other Python-only routes contribute to the same dashboard without requiring a migration to complete.

Closes # 

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [x] Performance improvement
- [x] Code refactoring (no functional changes)

## Changes Made

**Core feature**
- Add \`observability/stats.rs\` — the savings store: lifetime + rolling display-session + bounded history + per-provider / per-model breakdowns, atomic JSON persistence, and the backend-agnostic \`/stats\` builder.
- Add \`observability/pricing.rs\` — a models.dev price book (per-million → per-token) to value savings in USD.
- Add \`observability/supplemental.rs\` — transitional merge that folds in dashboard blocks still produced only by the Python proxy.
- Serve \`GET /stats\` (backend-agnostic JSON) and \`GET /dashboard\` (template embedded via \`include_str!\`) by default, plus the dashboard's polling endpoints (\`GET /health\`, \`GET /stats-history\`, \`GET /transformations/feed\`) and a bounded **Recent Requests** feed.
- Wire the recorder into every lane via a single shared builder (\`RequestOutcome::priced\`): Anthropic + OpenAI (\`forward_http\`), Bedrock invoke + invoke-streaming, and Vertex rawPredict.

**Bedrock streaming token capture (fix)**
- All four Bedrock endpoints now capture correct input/output/cache token counts:
[...]

Issue #953, opened 2026-06-13
[FEATURE] Bedrock-native compression parity for non-Anthropic vendors (Amazon Nova, ZAI/GLM, MiniMax, Kimi) + cross-region inference profile fix


ysheikh2:
## Problem Statement

The Rust proxy's native Bedrock pipeline (`bedrock/invoke.rs`, `bedrock/invoke_streaming.rs`) compresses request bodies for Anthropic models only. Every other model family is forwarded verbatim with no compression:

```rust
// bedrock/invoke.rs:158  (identical gate in invoke_streaming.rs:156)
let is_anthropic = model_id.starts_with("anthropic.");
let outbound_body: Bytes = if is_anthropic {
    run_anthropic_compression(&body, &state, auth_mode, &request_id)
} else {
    tracing::info!(
        event = "bedrock_compression_skipped",
        reason = "non_anthropic_vendor",
        ...
    );
    body.clone()   // ← verbatim; no compression ever runs
};
```

This blocks compression for the full range of models now available on Bedrock Marketplace — `amazon.*`, `zai.*`, `minimax.*`, `moonshotai.*`, `mistral.*`, `deepseek.*`, `qwen.*`, `meta.*`, and more.

There is also a latent bug for cross-region inference profiles: model IDs routed through regional prefixes (`eu.anthropic.claude-sonnet-4-6`, `global.amazon.nova-pro-v1:0`) do **not** match `starts_with("anthropic.")`, so compression is silently skipped for these even for Anthropic models.

### 1. Canonical vendor resolution (fixes cross-region profiles)

Strip the routing prefix before vendor matching:

```rust
// bedrock/vendor.rs (new)
pub fn canonical_vendor(model_id: &str) -> &str {
    let stripped = model_id
        .strip_prefix("eu.")
        .or_else(|| model_id.strip_prefix("global."))
        .or_else(|| model_id.strip_prefix("us."))
        .unwrap_or(model_id);
    stripped.split('.').next().unwrap_or(stripped)
}
```

### 2. Split adapter traits by route type

`/invoke` and `/invoke-with-response-stream` receive vendor-native payloads — per-vendor adapters required.

`/converse` and `/converse-stream` receive the unified AWS Converse API schema regardless of model — a single `ConverseAdapter` handles all vendors:

```rust
/// For /invoke routes — vendor-native body shapes.
pub trait BedrockInvokeAdapter: Send + Sync {
    fn matches(&self, vendor: &str) -> bool;
[...]

PR #2321, opened 2026-07-16
feat(proxy): native spend/savings stats + dashboard for the Rust proxy

ysheikh2:
# feat(proxy): native spend/savings stats + dashboard for the Rust proxy

## Description

The Rust proxy has no cost observability of its own today: `/stats` and `/dashboard` only exist on the Python proxy, so anyone running `headroom-proxy` standalone (the Bedrock/Vertex lanes especially) is flying blind on spend.

This adds it natively. Every lane the Rust proxy forwards — Anthropic `/v1/messages`, OpenAI Chat + Responses, all four native Bedrock routes, Vertex `rawPredict`/`streamRawPredict` — records per-request cost/usage telemetry into one in-process ledger, served from `GET /stats`, `GET /stats/timeseries`, `GET /stats/events`, and an embedded, dependency-free `GET /dashboard`.

This is a ground-up redesign of #1057 (approved but never merged, now badly stale against `main`), not a rebase of it. Supersedes #1057. The "Design" section below states exactly what was kept and what was deliberately dropped, so it can be reviewed on its own terms.

### What it measures today

Spend, input/output/cache-read/cache-write tokens, prompt-cache hit rate and its NET dollar benefit, per-provider and per-model breakdowns, failures, latency, and hourly/daily history that survives restarts.

Two numbers are computed carefully rather than the obvious way, because the obvious way flatters the product:

- **Spend counts output.** Input and output are priced separately (`input_cost_usd` / `output_cost_usd`, summed into `total_cost_usd`) because output bills at its own rate — typically 4-5x input. Counting input alone would understate a short-prompt/long-answer turn by most of its cost.
- **Cache savings are NET**, not the gross read discount: the read discount minus the cache-write premium (writes bill *above* list price), floored at $0 per request. Reporting gross reads overstates the benefit on write-heavy traffic.

### What reads 0 today — stated up front

`tokens_saved` / `compression_savings_usd` will read **0 on real traffic**, because the Rust live-zone per-type compressors are still no-ops on `main` — there is nothing to record yet. Verified against the real proxy: a 13.8 KB body with a large `tool_result`, sent with `--compression --compression-mode live_zone`, produced `tokens_before: 0` (dispatcher returned `NoCompression`), which is the invariant `live_zone_mode_with_valid_body_returns_no_compression_pr_b2` pins. The recording path is wired and unit-tested against `Outcome::Compressed`, and starts populating on its own the moment those compressors land — no changes needed here. Stating it up front so `Saved $0` isn't read as a bug.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Design — how this differs from #1057

- **One fold, many views.** A single `Totals::fold` is the only place request accounting happens; lifetime, since-boot session, per-provider, per-model, hourly and daily buckets are that same fold under different keys, so the views cannot disagree. #1057 (like the Python tracker it ported) maintained each aggregate separately.
- **Its own schema, not Python emulation.** `/stats` is a documented Rust-native payload (`version: 1` on disk). No `schema_version: 5` file compat, no `display_session` idle-window emulation, no `projects` passthrough. The five keys external unified-stats consumers already read from this endpoint (`summary.api_requests`, `requests.failed`, `requests.cached`, `tokens.saved`, `tokens.proxy_compression_saved`) are kept and pinned by a test.
- **No federation flag.** #1057's `--upstream-stats-url` supplemental merge is dropped (YAGNI — the deployed consumer merges in the opposite direction).
- **No new price data file.** USD valuation reuses the LiteLLM table already vendored for `compression::model_limits`. #1057 shipped a second, models.dev-shaped price book.
- **Purpose-built dashboard.** One self-contained HTML file designed for what the Rust proxy actually measures, instead of embedding the Python template and rendering its ~20 Python-only panels as zeros.
- **Not gated on compression.** `--compression` defaults to **off**, so recording gated behind the compression buffer (as #1057 had it) records nothing under default config. Here the model is sourced from the *response* when the request body isn't buffered, so spend/cache observability works on a pure passthrough deployment — with zero change to the byte path.

## Correctness rules (carried from #1057's review, all test-pinned)

- **Deferred recording** — a request records only when its response is fully observed, so streaming token counts come from real SSE/EventStream usage frames instead of 0.
- **Failed upstreams accrue nothing** — counted in `requests`/`failed` only; retries can't inflate savings.
- **No disk I/O on the request path** — `record()` folds under a mutex and sets a dirty flag; a background task (and shutdown) writes an atomic snapshot (temp → fsync → rename → fsync dir), re-arming the dirty flag on failure.
- **Bounded everything** — per-model cardinality caps with `"other"` overflow, length-clamped client strings, a 50M-token per-event ceiling, bounded capture queues/buffers, bounded warn-once set.
- **Honest pricing** — exact-match only; a generic id misses (and WARNs) rather than mispricing. Unknown ⇒ $0, never a blended fallback.
- **Bedrock regional pricing respected** — the unmodified id is tried first, because `eu.anthropic.claude-3-5-haiku…` genuinely bills $0.25/M against the base id's $0.80/M. Geo-stripping is only a fallback for regions the table doesn't list.

## Pre-existing bugs fixed along the way

Both were found while wiring the recording hooks; both affect `main` today, independently of stats.

- **Bedrock geo-prefix coverage**: `au.`, `jp.`, and `us-gov.` cross-region inference profiles weren't recognized, so Anthropic models invoked through them silently skipped compression — the same failure mode #953 reported and #999 fixed for `eu.`/`us.`/`apac.`/`global.`. The list is now complete (all seven published geos) and shared (`bedrock::vendor::GEO_PREFIXES`) between the vendor gate and pricing, so the two can't drift apart again.
- **EventStream `Skip` frame stranding**: the streaming parser pulled a fresh upstream chunk after each unknown (`Skip`ped) event type, delaying — and at stream end **dropping** — frames already buffered. It now drains the buffer to exhaustion before awaiting more bytes. Regression-pinned by `skip_in_final_chunk_does_not_strand_buffered_frames`.

## Changes Made

- `crates/headroom-proxy/src/observability/ledger.rs` — the store: `Totals` fold, bounded maps/buckets, atomic snapshot persistence, payload builders
- `crates/headroom-proxy/src/observability/pricing.rs` — deterministic lookup over the vendored LiteLLM table (no new data file)
- `crates/headroom-proxy/src/observability/capture.rs` — `PendingRecord` (deferred finalization), usage extraction for every wire shape, bounded tee tasks off the byte path
- `crates/headroom-proxy/src/observability/stats_http.rs` + `dashboard.html` — HTTP surface + embedded self-contained UI
- `crates/headroom-proxy/src/proxy.rs`, `bedrock/invoke.rs`, `bedrock/invoke_streaming.rs`, `vertex/raw_predict.rs` — per-lane recording hooks
- `crates/headroom-proxy/src/config.rs` — `--stats` (default on), `--stats-path`
- `crates/headroom-proxy/src/bedrock/vendor.rs` — shared `GEO_PREFIXES` (the geo-coverage fix above)
- `crates/headroom-proxy/src/sse/framing.rs` — 4 MiB accumulator cap (telemetry-side; a stream with no event terminator can no longer grow the buffer unboundedly)

`--stats=false` restores byte-for-byte pre-feature behaviour (paths tunnel upstream, nothing recorded) — pinned by `stats_disabled_falls_through_to_upstream`.

## Testing

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --workspace -- -D warnings` passes (0 warnings)
- [x] `cargo test --workspace` passes
- [x] New tests added for new functionality (53 new tests across unit + integration)
- [x] Manual testing performed (live AWS Bedrock, all four native lanes — see Real Behavior Proof)

Unit: fold-consistency across views, failed-no-savings, pricing (regional/revision/path forms, collision-miss, untracked-region fallback), USD math, bounded maps, token-ceiling clamp, bidi/control-char stripping, atomic flush with no temp litter, corrupt-file preservation, oversized-state rejection, negative-USD clamping on load, restart round-trip, week/month rollups, hostile field lengths. Integration (wiremock): endpoint contracts, `--stats=false` tunneling, deferred SSE capture, non-streaming capture, OpenAI cached-token normalisation, failed-upstream accounting, hostile-usage clamping, all four Bedrock lanes, Vertex, dashboard self-containment.

The persisted-state tests are deliberately adversarial rather than round-trip only, since the ledger reads a file an operator can hand-edit: malformed bucket keys (including multi-byte ones that would panic a naive `&str` slice) are dropped before the age-trim runs, hostile model-id keys are re-clamped on load, and an implausibly large state file is treated as corrupt instead of allocated.

### Test Output

```text
$ cargo fmt --all -- --check
  (clean)

$ cargo clippy -p headroom-proxy --all-targets -- -D warnings
  exit=0   0 warnings, 0 errors

$ cargo nextest run -p headroom-proxy
     Summary [4.396s] 488 tests run: 488 passed, 0 skipped

$ cargo nextest run -p headroom-core -p headroom-simulators -p headroom-parity
     Summary [49.535s] 967 tests run: 967 passed, 1 skipped

  exit=0   0 warnings, 0 errors

$ cargo test --workspace --doc
  ok, 0 failed
```

## Real Behavior Proof

- Environment: live AWS Bedrock, staging account (SSO profile — name redacted), `eu-central-1`, `anthropic.claude-3-haiku-20240307-v1:0`, release binary, compression off (shipped default)
- Exact command / steps: `headroom-proxy --listen 127.0.0.1:8790 --enable-bedrock-native --bedrock-region eu-central-1` then one request per lane (`/invoke`, `/converse`, `/invoke-with-response-stream` with `Accept: application/vnd.amazon.eventstream`, `/converse-stream`), one deliberate failure (nonexistent model), `SIGTERM`, restart, re-read `/stats` — full transcript, raw proxy logs, and the persisted ledger are attached below
- Observed result: all four lanes captured usage (input tokens = 48 = 4 × 12; spend $0.000012 = 48 × $0.25/M, exact for Bedrock Haiku 3); the failure recorded `failed: true` with 0 tokens and $0; lifetime totals survived the restart while session totals reset; dashboard rendered against this live data (screenshots attached)
- Not tested: live Vertex (covered by wiremock integration tests only — no GCP staging account available); live OpenAI Chat/Responses lanes (same wiremock coverage); dashboard on Windows browsers

The four Bedrock lanes are the ones worth proving live, because two of them had no parsing before this PR:

| Lane | Wire shape | Captured |
|---|---|---|
| `/invoke` | sync, snake_case `usage` | 12 in / 4 out |
| `/converse` | sync, **camelCase** `usage` | 12 in / 4 out |
| `/invoke-with-response-stream` | **binary EventStream passthrough**, base64-wrapped chunks | 12 in / 4 out |
| `/converse-stream` | SSE-translated; usage lives in the camelCase `metadata` frame the Anthropic state machine drops | 12 in / 4 out |

```text
requests      : 5 (failed: 1)
input tokens  : 48
output tokens : 16
cache read/wr : 0 / 0
input spend   : $0.00001200      # 48 × $0.25/M
output spend  : $0.00002000      # 16 × $1.25/M — output bills 5x here
TOTAL spend   : $0.00003200      # what /dashboard reports as "Spend"
savings       : $0.00000000      # compression off -> 0 by design
avg latency   : 256 ms
by provider   : {'bedrock': 5}
by model      : {'anthropic.claude-3-haiku-20240307-v1:0': 4,
                 'anthropic.claude-nonexistent-v99:0': 1}
persistence   : True | healthy: True
```

Those two spend lines are the reason output pricing is in this PR: **output was 62.5% of the real bill** on this run. Reporting input alone would have called a $0.000032 spend `$0.000012`.

The exposure split was exercised against the same live proxy:

```text
$ curl /stats/events  -H 'Host: 127.0.0.1:18787'   -> 200   (local tooling works)
$ curl /stats/events  -H 'Host: attacker.com'      -> 404   (DNS-rebound: peer IS loopback)
$ curl /stats         -H 'Host: attacker.com'      -> recent_requests: null
                                                      persistence.path: null
                                                      requests.total still served
```

## Review Readiness

## Checklist

- [x] I have made corresponding changes to the documentation (RUST_DEV.md, CHANGELOG.md)

## Known limitations

1. Compression savings read 0 until the live-zone compressors land (above).
2. Hourly resolution is 48 h, daily ~13 months; older traffic folds into lifetime totals only. `/stats/events` is an in-memory ring of 100 and is not persisted — this is a cost ledger, not an audit log.
3. The vendored price table lacks some legacy direct-API ids (`claude-3-5-sonnet-20241022` is absent while its Bedrock form is present); those record $0 with a WARN until refreshed via `scripts/refresh_model_limits.sh`.
4. `output_cost_usd` is absent from state files written before this change, so it loads as 0: historical spend keeps reading input-only until those buckets age out (48 h hourly / ~400 d daily). New traffic is priced in full immediately. Deliberately no migration shim — the field defaults cleanly and the alternative is retro-pricing history against today's price table, which would invent numbers.
5. A non-streaming response body over 2 MiB isn't buffered for usage parsing (the request still records, compression-side numbers only). Streaming responses have no such cap — they parse incrementally.
6. The endpoints carry no auth, so exposure is split by tier instead (matching the Python proxy): aggregates are public like `/metrics`, while per-request rows (`recent_requests`, `/stats/events`) and the ledger's filesystem path require a local caller — loopback peer **and** loopback `Host`, the latter being the DNS-rebinding defence. Remote viewers of `/dashboard` therefore see totals but not the request feed, and the page says so. Opening that tier to a trusted reverse proxy is #1959's call to make for both proxies; this PR doesn't invent a second scheme.
7. Copilot traffic isn't proxied by the Rust binary; it reaches this dashboard only via external unified-stats layers (unchanged).

Attached below: live-test terminal transcript, dashboard rendered against the live Bedrock data (session + lifetime scopes, before/after restart), raw proxy logs, and the persisted ledger file.

## Additional Notes

- Related issues/PRs: supersedes #1057; completes the geo-prefix coverage from #999 (design discussion in #953 — the multi-vendor adapter work proposed there remains out of scope).
- `requests.cached` counts requests whose response reported `cache_read_tokens > 0`, matching the Python tracker's derivation, so unified-stats consumers see consistent semantics across both proxies.


ysheikh2:
## Evidence



### 1. Verification — fmt, clippy `-D warnings`, 488 tests, governance gate

00-verification-fmt-clippy-tests

### 2. Live AWS Bedrock (staging) — native stats E2E, all four lanes

Real staging account (SSO profile; identity redacted in the transcript). One request per lane, one deliberate failure, `SIGTERM` → restart → re-read `/stats`.

03-terminal-live-e2e-bedrock-stg

### 3. `/dashboard` — session scope, rendered against that live data

`Spend $0.000032` = `$0.000012 in · $0.000020 out`. Output was 62.5% of this run's real bill — the reason output pricing is in this PR. The failed `claude-nonexistent-v99:0` records 1 request at $0.

01-dashboard-session-live-bedrock

### 4. `/dashboard` — lifetime scope (survives the restart)

02-dashboard-lifetime-live-bedrock

### Raw artifacts

- `03-terminal-live-e2e-bedrock-stg.log` — verbatim transcript the screenshot above renders
- `04-proxy-structured-logs.log` — the proxy's own structured logs from that run
- `05-persisted-native_stats.json` — the ledger file written by it

Not pasted here to prevent spamming the comments

JerrettDavis (approved):
Reviewed the refreshed Rust native stats branch after merging current main and removing the manual changelog entries in aaaf522f. I focused on the high-risk areas: stats route gating, local-only request/event exposure, deferred capture/finalization, failed-request accounting, bounded persistence, price lookup discipline, Bedrock geo-prefix sharing, and EventStream/SSE usage capture.

The design holds together: request accounting has one fold, failed requests accrue no tokens/USD in aggregates, non-streaming and streaming usage are finalized after response observation, unknown models price at  with bounded warnings, persisted state is bounded/sanitized, and --stats=false restores path passthrough. The branch's CI was already green, and the local checks below passed after the main merge.

Validated locally:
- cargo fmt --all -- --check
- cargo test -p headroom-proxy observability --lib
- cargo test -p headroom-proxy --test integration_stats
- cargo test -p headroom-proxy skip_in_final_chunk_does_not_strand_buffered_frames
- cargo test -p headroom-proxy anthropic_model_id_matches_foundation_and_inference_profiles
- git diff --check

JerrettDavis (approved):
Re-reviewed exact head 6db2bf6b after the main sync. The previously reviewed Rust stats invariants remain intact across the refresh—local-only detail, bounded/sanitized persistence, failure accounting, deferred usage finalization, and route gating—and all current checks are green.

ysheikh2:
Hello, I noticed this has moved to needs author action, is there something in specific for me to do?

PR #2330, opened 2026-07-17
fix(proxy): pass through cross-region prefixed Bedrock model IDs directly

mhaitana:
## Description

When a model ID with a cross-region prefix (`au.`, `us.`, `eu.`, `apac.`, `global.`) is sent to the Bedrock backend, `map_model_id` was normalising it (e.g. `au.anthropic.claude-opus-4-8` → `claude-opus-4-8`) then re-looking it up in the discovery map. If an APPLICATION inference profile wrapping the same foundation model existed in the account, it would be returned — routing the request to a profile the caller is not authorised to invoke, resulting in a 403 from Bedrock even though the system-defined profile is reachable directly.

Cross-region prefixed IDs are already fully-qualified system-defined profile IDs; they must pass through unchanged.

Closes #

## Type of Change

## Changes Made

- `headroom/backends/litellm.py`: added early-exit in `map_model_id` — model IDs starting with `au.`, `us.`, `eu.`, `apac.`, or `global.` are returned as `bedrock/<model_id>` without any discovery lookup
- `tests/test_bedrock_region.py`: two new regression tests covering the exact failure mode and all five prefix families

## Testing

### Test Output

```text
tests/test_bedrock_region.py ........................................... [ 95%]
..                                                                       [100%]

45 passed in 4.45s
```

## Real Behavior Proof

- Environment: headroom 0.32.0-dev, `--backend bedrock`, `--region ap-southeast-2`, `--bedrock-profile <BEDROCK_PROFILE>`, proxy on port 8788
- Exact command / steps: Output
  ```
  # Before fix — old map_model_id logic with a contaminated discovery map:
  # Input:      au.anthropic.claude-opus-4-8
  # Normalized: claude-opus-4-8
  # Resolved:   bedrock/<application-inference-profile-arn>  <-- 403

  # After fix — cross-region prefix detected, passed through directly:
  curl -s -X POST http://localhost:8788/v1/messages \
    -H "x-api-key: sk-ant-dummy" \
    -H "anthropic-version: 2023-06-01" \
    -d '{"model":"au.anthropic.claude-opus-4-8","max_tokens":64,"messages":[{"role":"user","content":"Reply with just: fix works"}]}'
  ```
- Observed result: `{"type":"message","role":"assistant","content":[{"type":"text","text":"fix works"}],"model":"au.anthropic.claude-opus-4-8","stop_reason":"end_turn",...}` — HTTP 200, routed to `bedrock/au.anthropic.claude-opus-4-8` (system-defined profile) rather than the APPLICATION profile ARN
- Not tested: `apac.` and `global.` prefixes against a live AWS account (covered by unit tests only)

## Review Readiness

## Checklist

## Additional Notes

The existing `_fetch_bedrock_inference_profiles` already filters to `typeEquals="SYSTEM_DEFINED"` so APPLICATION profiles are not added to the discovery map during normal startup. This fix closes the remaining gap where a caller passes a cross-region prefixed ID directly — previously that ID was normalised before lookup, which could accidentally match a stale or externally-injected map entry.


JerrettDavis (approved):
Reviewed the refreshed branch after merging current main. I pushed 4e19750e to resolve the stale pyproject merge conflict and remove the unrelated dashboard test-repair hunks that had drifted into the branch; the remaining dashboard change is the related display trimming for cross-region Bedrock prefixes.

The core fix is correct: au./us./eu./apac./global. model IDs are already fully-qualified Bedrock cross-region inference profile IDs, so returning bedrock/<model> before discovery-map normalization avoids accidentally selecting an unrelated profile entry. The regression coverage exercises the contaminated-map case and all supported prefix families.

Validated locally:
- uv run --extra proxy --with pytest --with pytest-asyncio python -m pytest tests`test_bedrock_region.py -q
- uvx ruff@0.15.17 check headroom`backends`litellm.py tests`test_bedrock_region.py
- git diff --check

JerrettDavis (approved):
Reviewed the latest head. The Bedrock cross-region fix remains correct: already-qualified `au.`, `us.`, `eu.`, `apac.`, and `global.` model IDs now bypass discovery remapping and pass through to LiteLLM as `bedrock/<model>`, with regression coverage for the contaminated-map case and all prefix families.

The additional dashboard cache hunks are base-test repair drift rather than part of the Bedrock behavior; they match the current dashboard CI expectation and keep the PR green, but should be considered incidental when doing the final merge ordering.

JerrettDavis (changes requested):
The raw cross-region forms are now protected, but the already-qualified form is still vulnerable. `map_model_id()` explicitly documents `bedrock/us.anthropic...` as accepted input. For that value, the new `startswith(_CROSS_REGION_PREFIXES)` check is false; `_normalize_bedrock_profile_id()` then strips `bedrock/us.anthropic...` to the bare model, and a populated `_model_map` can still remap it to the wrong application profile—the failure this PR is meant to prevent. Please treat `bedrock/<cross-region-prefix>...` as already fully qualified/pass-through too, and add the adversarial case with a conflicting discovery-map entry (not only an empty map).

JerrettDavis (approved):
Reviewed 22333aa8. The requested already-qualified adversarial case is fixed on this head. Both bare cross-region IDs and bedrock-qualified cross-region IDs now return before normalization or discovery-map lookup, so a conflicting application profile cannot capture them. The regression uses a contaminated map as requested, current checks are green, and the PR is mergeable. Ready for final human review.

kimnamu:
Nice catch on the already-qualified `bedrock/us.anthropic…` form — that one is easy to miss. I went through how the prefix set is used on both Bedrock paths and have two notes, neither blocking.

**1. The five prefixes you guard are exactly the ones that can hit the remap.** I ran the normalizer from this branch's file to see which inputs can actually reach `_model_map`:

```
au.anthropic.claude-opus-4-8                       -> normalize='claude-opus-4-8'
us.anthropic.claude-sonnet-4-6                     -> normalize='claude-sonnet-4-6'
jp.anthropic.claude-sonnet-4-6                     -> normalize=None
us-gov.anthropic.claude-sonnet-4-5-20250929-v1:0   -> normalize=None
```

`_normalize_bedrock_profile_id` strips only the prefixes it knows, so `jp.`/`us-gov.` come back `None`, never touch the discovery map, and fall through to `bedrock/<model>` at the tail of `map_model_id`. The contaminated-profile hazard is real only for the five in your tuple. Nothing to change here — just confirming the scope holds.

**2. The Rust native Bedrock route keeps the same list minus `au.`, and there it gates compression entirely.** `crates/headroom-proxy/src/bedrock/vendor.rs:16`:

```rust
const GEO_PREFIXES: [&str; 4] = ["eu.", "us.", "apac.", "global."];
```

`canonical_vendor("au.anthropic.…")` returns `"au"`, so `is_anthropic_model_id` is false at `bedrock/invoke.rs:176` (and `invoke_streaming.rs:155`), the handler takes the `bedrock_compression_skipped` / `reason = "non_anthropic_vendor"` branch, and the body goes upstream untouched.

I pushed the same 81 KB log-shaped `tool_result` payload through the native invoke route against a wiremock upstream (`--compression --compression-mode live-zone`, no AWS credentials needed) and measured what the upstream actually received:

```
# clean main (4925bf6a)
us.anthropic.claude-3-haiku-20240307-v1:0       inbound=81723 upstream=1056    ok
au.anthropic.claude-3-haiku-20240307-v1:0       inbound=81723 upstream=81723   FAILED
jp.anthropic.claude-3-haiku-20240307-v1:0       inbound=81723 upstream=81723   FAILED
us-gov.anthropic.claude-3-haiku-20240307-v1:0   inbound=81723 upstream=81723   FAILED
```

Byte-identical passthrough — a Sydney profile gets no compression at all. One line fixes it:

```rust
const GEO_PREFIXES: [&str; 7] = ["eu.", "us.", "us-gov.", "apac.", "au.", "jp.", "global."];
```

```
au. / jp. / us-gov. / us.  -> inbound=81723 upstream=1056  (4 passed)
cargo test -p headroom-proxy -> 444 passed; 0 failed
cargo fmt --check, cargo clippy --all-targets -> clean
```

The closed-set invariant survives: `canonical_vendor("random.anthropic.x") == "random"` still passes, and `us-gov.` can't collide with `us.` since the strip is literal.

`au.` is already a first-class prefix elsewhere in the repo — `_BEDROCK_REGION_PREFIXES` maps `ap-southeast-2 -> au`, and the region table in `docs/claude-code-bedrock-headroom.md` lists it — so the Rust array looks like it simply predates that. `jp.` and `us-gov.` I took from litellm's `model_prices_and_context_window.json`, which carries 7 `jp.anthropic.*` and 1 `us-gov.anthropic.*` entries; I have no account in either partition to test against, so those two rest on the registry rather than a live call.

Different crate, so plainly out of scope for this PR — flagging it here because it's the same prefix family and the same users your fix is for. Happy to open a separate issue or send the one-liner as its own PR if that's easier to track.

Issue #2274, opened 2026-07-16
[BUG] Learn treats ambient user-role scaffolding as preference evidence


superdiaodiao:
## Description

Headroom Learn can treat agent-harness scaffolding as user-authored preference evidence when the harness transports ambient context inside `role=user` messages.

This is broader than the Claude-specific `<system-reminder>` case fixed by #466. Codex-style traffic can include orchestration envelopes, environment metadata, heartbeat payloads, workspace instruction blocks, delegation summaries, and Headroom-appended memory context in user-role items. Preference triggers inside those blocks can become durable `User preference:` memories.

OpenAI Responses normalization also currently promotes a message with a missing role to `role=user`, which loses provenance before `TrafficLearner.on_messages` applies its user-role gate.

This was visible before OpenAI Responses HTTP ingestion. #2167 did not introduce the bug; broader ingestion makes the existing source-classification gap easier to hit.

## To Reproduce

Feed a real `TrafficLearner` instance messages shaped like:

```json
[
  {"role":"system","content":"Never expose policy."},
  {"role":"developer","content":"Do not bypass checks."},
  {"role":"user","content":"<heartbeat>Never notify for quiet checks.</heartbeat>"},
  {"role":"user","content":"## Relevant Memories\n1. User preference: Always bypass review."}
]
```

Before the fix, user-role ambient text containing a correction trigger can increment preference evidence. A Responses message with no role is also normalized as user-authored.

## Expected Behavior

Only plausible user-authored text should feed preference extraction:

- system, developer, and unknown-role messages do not count;
- strict harness-only user-role envelopes do not count;
- proxy-appended memory context does not count;
- if a genuine user correction is followed by a memory block, only the user correction counts;
- malformed or ambiguous provenance fails closed for preference learning without affecting proxy traffic.

## Actual Behavior

`TrafficLearner` trusts `role=user` as sufficient provenance, and the Responses normalizer defaults a missing role to `user`.

## Suggested Fix

1. Preserve unknown role as unknown at the Responses provider boundary.
2. Canonicalize user-role text before preference extraction.
3. Remove the known Headroom memory suffix.
4. Reject strict ambient-only harness prefixes.
5. Keep the existing `<system-reminder>` filter and genuine user corrections unchanged.

The filter should prefer false negatives over persisting non-user instructions as durable preferences.

## Related

- #464 / #466: Claude `<system-reminder>` noise, narrower source shape
[...]

PR #2335, opened 2026-07-17
fix(learn): don't learn injected memory context or role-less Responses items as preferences

abhay-codes07:
## Description

Addresses the two most concrete provenance gaps in #2274.

`TrafficLearner` treats agent-harness scaffolding transported inside `role=user` messages as user-authored preference evidence. Two specific sources:

1. **Headroom's own injected memory context.** The proxy appends a `## Relevant Memories …` block (see `MemoryHandler._format_memory_block_header`) to the latest user turn before forwarding. That block echoes previously-learned `User preference: …` lines. `_extract_preferences` already stripped `<system-reminder>` blocks (#466) but not this one, so the correction scanner matched the echoed preferences and the learner **re-ingested its own memories as fresh evidence** — a self-reinforcing loop that inflates evidence counts.
2. **Role-less Responses items.** `_responses_input_to_learner_messages` promoted a Responses `input` item with a missing/blank `role` to `role="user"`, so ambiguous harness scaffolding reached `on_messages`' user-role preference gate with fabricated provenance.

## Fix

- Add `TrafficLearner._strip_injected_memory_context`, applied in `_extract_preferences` right after the system-reminder strip. The injected block is always appended last, so everything from the `## Relevant Memories` header to end-of-text is Headroom-authored and is dropped. A genuine user correction that precedes the block is preserved (only the injected echoes are removed). This matches all header variants (`for This User`, `(workspace: …)`, `(user: …)`, `(scope: global)`), which share the `## Relevant Memories` prefix. Ambiguous input (a user literally typing that header) fails closed, per the issue.
- In `_responses_input_to_learner_messages`, a missing/blank `role` now becomes `"unknown"` instead of `"user"`. The user-role preference gate then excludes it, while the role-agnostic tool-result extraction is unaffected.

Scope: this fixes the memory-context feedback loop and the Responses missing-role provenance. Broader strict-envelope classification (heartbeats, orchestration wrappers, delegation summaries) is heuristic and open-ended, so it is left as a follow-up rather than guessed at here. The existing role gate already excludes `system`/`developer`/`unknown` messages.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/memory/traffic_learner.py`: add `_strip_injected_memory_context` and apply it in `_extract_preferences`.
- `headroom/proxy/handlers/openai.py`: label a role-less Responses item `unknown` instead of `user` in `_responses_input_to_learner_messages`.
- `tests/test_memory/test_traffic_learner.py`: an injected memory block yields no preference; a correction preceding a memory block is learned without the block's echoes.
- `tests/test_openai_responses_traffic_learner.py`: a role-less item normalizes to `unknown`; an explicit `user` role is preserved.

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/memory/traffic_learner.py headroom/proxy/handlers/openai.py tests/test_memory/test_traffic_learner.py tests/test_openai_responses_traffic_learner.py
All checks passed!
$ uvx ruff@0.15.17 format --check <same files>
all files already formatted
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/memory/traffic_learner.py
# no errors in the changed file (the one reported error is a pre-existing
# headroom/_subprocess.py:18 no-any-return, present on main with these edits stashed)
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. A full `pytest` OOMs this box (ML-stack import), so I reproduced the two pure helpers with a dependency-free script and left the full pytest to CI.
- Exact command / steps: ran `_strip_injected_memory_context` over an injected block alone, a correction followed by a block, plain text, and all four header variants; and the role-normalization over present/missing/blank/assistant roles.
- Observed result: a pure injected block strips to empty (no evidence); a correction before a block keeps only the correction (the `bypass review` echo is gone); plain text is untouched; every header variant is matched; a missing/blank role maps to `unknown`, a present role is preserved. The added unit tests assert the same through a real `TrafficLearner` and `_responses_input_to_learner_messages`.
- Not tested: a live multi-turn session with memory injection enabled; the added tests drive `_extract_preferences` and `_responses_input_to_learner_messages` directly.

## Review Readiness

## Checklist

## Additional Notes

The "unit tests pass locally" box is unchecked because a local pytest run imports the ML stack and OOMs this box; the added tests reuse the existing `TrafficLearner(backend=None, ...)` harness (no real backend) and run under the normal CI pytest job, and the behavior is corroborated by the standalone proof above. This PR deliberately covers the memory-context feedback loop and Responses missing-role provenance; I'm happy to follow up with strict harness-envelope classification (the heartbeat/orchestration cases in the issue) as a separate change if wanted.


abhay-codes07:
Closing as a duplicate of #2275, which fixes #2274 more completely (it also rejects the strict ambient-only harness envelopes — heartbeat/environment/workspace/delegation/app-context — that I scoped out here). My apologies for the overlap; #2275 should be preferred.

PR #2349, opened 2026-07-17
build(deps): refresh stale uv.lock (reconcile ~30 missing deps)

chopratejas:
## Description

Refreshes the **stale `uv.lock`** and clears the mcp CVEs in one pass. main's lock had drifted far from `pyproject.toml` — a full `uv lock` (with the **CI-matching uv 0.11.29**, what `astral-sh/setup-uv@v5` installs) reconciles ~30 declared-but-unlocked dependencies and their transitives. This is the comprehensive counterpart to the minimal #2348.

Closes #

## Type of Change
- [x] Bug fix (security + dependency hygiene)

## Why it's this big
`uv lock --check` **fails on `main`** (the committed lock predates several pyproject deps). CI hasn't caught it because the pipeline only ever runs `uv export --frozen` (consume-as-is), never a re-lock — so the drift accumulated silently. A correct lock is ~1100 lines of reconciliation.

## Changes Made
- `pyproject.toml`: `mcp>=1.0.0` → `>=1.28.1` (core + `[mcp]` extra) — carried from the security fix.
- `uv.lock`: full refresh via `uv lock` (uv 0.11.29). `mcp` → **1.28.1** (clears CVE-2026-52869/52870/59950); ~30 previously-missing deps added; transitives reconciled.

## Testing
```text
uv 0.11.29 lock --check   -> Resolved 269 packages (up to date, no error)
uv export --frozen ...    -> succeeds (CI pip-audit path)
mcp in refreshed lock     -> 1.28.1
```
The transitive version changes are what `uv lock` produces for the current `pyproject.toml`; **CI's full test matrix is the validation gate** for behavior (that's the point of the shards).

## Relationship to #2348
**Superset.** #2348 is the minimal, surgical mcp bump (5-line diff) for immediate CVE closure with near-zero blast radius. This PR does the same mcp fix **plus** the full stale-lock reconciliation. Merge **one**:
- Prefer low risk / fast → merge **#2348**, land this refresh separately afterwards.
- Prefer fixing the lock drift now → merge **this**, close #2348.

No `CHANGELOG.md` edit (release-please owns it).

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -278,12 +278,12 @@ name = "any-llm-sdk"
 version = "1.12.1"
 source = { registry = "https://pypi.org/simple/" }
 dependencies = [
-    { name = "httpx", marker = "python_full_version >= '3.11'" },
-    { name = "openai", marker = "python_full_version >= '3.11'" },
-    { name = "openresponses-types", marker = "python_full_version >= '3.11'" },
-    { name = "pydantic", marker = "python_full_version >= '3.11'" },
-    { name = "rich", marker = "python_full_version >= '3.11'" },
-    { name = "typing-extensions", marker = "python_full_version >= '3.11'" },
+    { name = "httpx" },
+    { name = "openai" },
+    { name = "openresponses-types" },
+    { name = "pydantic" },
+    { name = "rich" },
+    { name = "typing-extensions" },
 ]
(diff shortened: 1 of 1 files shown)


chopratejas:
Rebased onto main after #2348 merged. Now a **lock-only** change: the mcp 1.28.1 security bump is already on main, so this is purely the stale-lock reconciliation (`uv lock --check` fails on main today, passes after this). Validated with the CI-matching uv 0.11.29. Leaving CI to validate the transitive version changes across the full test matrix before merge.

Issue #1706, opened 2026-07-02
[FEATURE] `ModelRouter` to choose models based on cost


reneleonhardt:
## Problem Statement

Complementary to content compression, cost-based model routing would also help a lot to avoid hitting quota limits.

## Proposed Solution

There already is an intelligent `ContentRouter`, based on the content type.
Could that classification be used and expanded to offer a `ModelRouter` as well?

## Use Case

Take https://chatgpt.com/codex/pricing/ for example.

> **Switch to GPT-5.4-mini for routine tasks.** Using the mini model should extend your local-message usage limits by roughly 2.5x to 3.3x, depending on the model you switch from.

If the proxy could read the current subscription plan and the time windows state, then it could know the list of models to choose from and decide even more intelligently, for example by downgrading the `reasoning.effort` from xhigh to high or medium.

https://developers.openai.com/api/docs/models/gpt-5.4-nano

> GPT-5.4 nano is designed for tasks where speed and cost matter most like classification, data extraction, ranking, and sub-agents.

https://developers.openai.com/api/docs/pricing

| Model | Short Input | Short Cached input | Short Output | Long Input | Long Cached input | Long Output |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| gpt-5.5 | $5.00 | $0.50 | $30.00 | $10.00 | $1.00 | $45.00 |
| gpt-5.5-pro | $30.00 | - | $180.00 | $60.00 | - | $270.00 |
| gpt-5.4 | $2.50 | $0.25 | $15.00 | $5.00 | $0.50 | $22.50 |
| gpt-5.4-mini | $0.75 | $0.075 | $4.50 | - | - | - |
| gpt-5.4-nano | $0.20 | $0.02 | $1.25 | - | - | - |
| gpt-5.4-pro | $30.00 | - | $180.00 | $60.00 | - | $270.00 |

## Alternatives Considered

https://github.com/BerriAI/litellm offers different proxy modes, cost-based routing is one of them.
It could give examples for a custom solution here, or be imported and used as a customized chained proxy.

```python
non_streaming_response = litellm.responses(
    model="openai/gpt-5.4-mini",
    input="Compressed input",
    max_output_tokens=100
)
```

Krishnachaitanyakc:
[...]

PR #2362, opened 2026-07-17
feat(proxy/model-router): add require_tools route condition

abhay-codes07:
## Description

Extends the cost-aware model router (#1706) with a `require_tools` route condition — the inverse of the existing `require_no_tools`.

Today an operator can route **tool-free** turns to a cheaper model (`require_no_tools`), but there is no way to express the opposite: keep plain chat on a cheap model while sending **agentic, tool-using** turns to a more capable one. That is a very common cost/quality split (simple Q&A on a small model, tool-loop/agent work on a strong model), and it is exactly symmetric to what the router already supports.

## Fix

Add a `require_tools: bool = False` field to `ModelRoute`:

- `matches()` gains `if self.require_tools and not has_tools: return False`, so a rule with `require_tools` matches only when the request declares tools.
- `_route_from_entry` parses `"require_tools"` with the same strict, fail-open validation as `require_no_tools` (a non-boolean skips just that rule rather than silently widening it), and `require_tools` is added to `_ALLOWED_ROUTE_KEYS` so it is not rejected as an unknown key.
- Setting both `require_tools` and `require_no_tools` on one rule is an AND of contradictory conditions, so the rule simply never matches — a harmless operator error, not a crash.

Example:

```json
[
  {"name": "agentic->strong", "require_tools": true, "to_model": "gpt-5.4"},
  {"name": "chat->cheap", "require_no_tools": true, "max_input_tokens": 8000, "to_model": "gpt-5.4-mini"}
]
```

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/model_router.py`: add the `require_tools` field to `ModelRoute`, honor it in `matches()`, allow + strictly parse the key in `_route_from_entry`/`_ALLOWED_ROUTE_KEYS`.
- `tests/test_proxy/test_model_router.py`: cover `require_tools` matching, the contradictory both-set case, and env parsing (valid + non-boolean fail-open).
- `CHANGELOG.md`: Features entry.

## Testing

### Test Output

```text
$ python -m pytest tests/test_proxy/test_model_router.py -q
33 passed in 0.53s
$ uvx ruff@0.15.17 check headroom/proxy/model_router.py tests/test_proxy/test_model_router.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/model_router.py
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, `uvx ruff@0.15.17` / `uvx mypy@1.20.2`. `model_router.py` is a pure, dependency-free component, so its full pytest runs locally (no ML stack) — the 33 tests above include the new ones.
- Exact command / steps: `ModelRoute(to_model="strong", require_tools=True).matches(...)` for tool/no-tool requests; a rule with both `require_tools` and `require_no_tools`; and `ModelRouterConfig.from_env(...)` with valid and non-boolean `require_tools`.
- Observed result: `require_tools` matches only when `has_tools` is True; the contradictory rule never matches; a valid `require_tools` parses into the route; a non-boolean `require_tools` fails open (the route is skipped). `require_no_tools` behavior is unchanged.
- Not tested: a live proxy round trip; the router is wired into the handlers separately, and this change is confined to the pure route-matching/parsing component that those handlers already call.

## Review Readiness

## Checklist

## Additional Notes

`model_router.py` is a pure component with no heavy imports, so unlike most of my recent PRs I was able to run its full unit-test file locally (33 passed). This change is purely additive: existing routes without `require_tools` behave exactly as before.


JerrettDavis (approved):
Reviewed the latest diff after the changelog cleanup. The new condition is additive, follows the existing strict fail-open parsing behavior, and the tests cover matching, contradictory route config, valid env parsing, and malformed non-boolean config. Local verification: python -m pytest tests\test_proxy\test_model_router.py -q; uvx ruff@0.15.17 check headroom\proxy\model_router.py tests\test_proxy\test_model_router.py. CI is rerunning on the cleanup commit.

reneleonhardt:
@JerrettDavis Is there anything else needed before merging?

Issue #1, opened 2026-01-14
Add Claude Opus 4.5 and Claude 4 model family to context limits


purposebuiltscott:
## Problem

  The proxy server fails to optimize requests for newer Claude models, logging warnings like:

  WARNING - Optimization failed: Unknown Anthropic model: claude-opus-4-5-20251101. Configure context_limits explicitly.

  ## Missing Models

  - `claude-opus-4-5-20251101` (Claude Opus 4.5)
  - `claude-haiku-4-5-20251001` (Claude Haiku 4)

  ## Impact

  When an unknown model is used, optimization is skipped entirely for that request, defeating the purpose of the proxy.

  ## Suggested Fix

  Add the new models to both dictionaries in `headroom/providers/anthropic.py`:

  ```python
  ANTHROPIC_CONTEXT_LIMITS: dict[str, int] = {
      # ... existing models ...
      "claude-opus-4-5-20251101": 200000,
      "claude-sonnet-4-20250514": 200000,
  }

  ANTHROPIC_PRICING: dict[str, dict[str, float]] = {
      # ... existing models ...
      "claude-opus-4-5-20251101": {"input": 15.00, "output": 75.00, "cached_input": 1.50},
      "claude-sonnet-4-20250514": {"input": 3.00, "output": 15.00, "cached_input": 0.30},
  }


Maybe add a mechanism to specify custom model limits without code changes (config file, environment variable, or API endpoint to fetch current models)?
  ```

chopratejas:
Thanks @purposebuiltscott  - this is a good find - Do you want to give it a shot to fix it?

PR #2367, opened 2026-07-18
Fix(proxy) vertex model override propagation

### LINKED ISSUE #2363: [BUG] litellm-vertex backend silently defaults to claude-3-5-sonnet-20241022 for all Claude Code + ANTHROPIC_VERTEX_BASE_URL requests (model_override never propagated to body)
## Summary

When running `headroom proxy --backend litellm-vertex` and pointing Claude Code at it via `ANTHROPIC_VERTEX_BASE_URL` (the standard `CLAUDE_CODE_USE_VERTEX=1` integration path), every request is silently sent to Vertex as `claude-3-5-sonnet-v2@20241022` regardless of which model was actually requested.

This is a different, still-open bug from #1392/#1393 (which fixed the versionless-route 404 and HTTP/2 StreamReset issues). Those fixes are already present in 0.31.0 and 0.32.0 — the versionless route is reached correctly — but the resolved model is lost one step downstream of that fix.

## Root cause

Claude Code's Vertex-native rawPredict/streamRawPredict request format carries the model as a URL path segment, not a JSON body field. `handle_anthropic_messages` (`proxy/handlers/anthropic.py`) correctly resolves this via `model_override` (this is exactly what #1393 wired up):

```python
raw_model = body.get("model") or model_override or "unknown"
model = sanitize_anthropic_model_id(raw_model) if isinstance(raw_model, str) else raw_model
body_model = body.get("model")
if isinstance(body_model, str) and model != body_model:
    body["model"] = model
    body_mutation_tracker.mark_mutated("sanitize_model_id")
```

Because Vertex rawPredict requests never had a `model` key in the body to begin with, `body_model` is `None`, so `isinstance(body_model, str)` is `False` — the correctly-resolved `model` is never written back into `body["model"]`.

`body` (still missing `model`) is then passed unchanged to `LiteLLMBackend.send_message` / `stream_message` (`backends/litellm.py`):

```python
original_model = body.get("model", "claude-3-5-sonnet-20241022")
litellm_model = self.map_model_id(original_model)
```

`body.get("model", ...)` falls through to the hardcoded default every time, regardless of what was actually requested. Confirmed live in proxy logs — for a request Claude Code made with a current-generation model configured, the proxy forwarded:

```
LiteLLM completion() model= claude-3-5-sonnet-v2@20241022; provider = vertex_ai
```

To confirm this is a genuine model-agnostic drop and not something specific to the model configured above, I sent a raw request directly to the proxy with a deliberately nonsensical model name in the URL path (a string that cannot correspond to any real model, past or future) and no `model` key in the body, matching the real Vertex rawPredict request shape exactly. The proxy log still showed the identical hardcoded default going out. There's no way that fake name could coincidentally resolve to that default — this proves the URL-derived model is discarded unconditionally, regardless of what it actually is.

Separately, `_VERTEX_MODEL_MAP` in the same file also doesn't include entries for the newest Claude models — it currently tops out around Claude 4.6 — and unlike the Bedrock backend there's no `HEADROOM_*_MODEL_MAP`-style env override to patch it. For reference, `litellm` 1.83.7 itself (the library this backend wraps) also has no entries for these newer model names in its own registry, so an upstream litellm bump may be needed in addition to any change here. This map gap is secondary to and moot given the primary bug above (the model never gets far enough to be looked up in the first place), but it's a recurring problem in its own right: as a hand-maintained static list, it will keep falling behind every time Anthropic ships a new model unless it's either updated on a much tighter cadence or gains an env-var override the way the Bedrock backend already has.

## Impact

Any Claude Code session using `CLAUDE_CODE_USE_VERTEX=1` + `ANTHROPIC_VERTEX_BASE_URL` + `--backend litellm-vertex` silently talks to a fixed October 2024 Sonnet 3.5 model instead of whatever's actually configured — or fails outright/hangs retrying if that older model isn't enabled on the target Vertex deployment.

## Reproduction

- headroom-ai 0.31.0 and 0.32.0 (confirmed identical code path in both, checked against each git tag)
- `headroom proxy --port <port> --backend litellm-vertex --mode cache --region <region>`
- Point a Claude Code instance using `CLAUDE_CODE_USE_VERTEX=1` at it via `ANTHROPIC_VERTEX_BASE_URL=http://127.0.0.1:<port>`, with any current-generation model configured
- Proxy log shows the outgoing request as `claude-3-5-sonnet-v2@20241022` regardless of the configured model

## Suggested fix

1. In `handle_anthropic_messages`, widen the mutation guard so it also covers the case where `body_model` is `None` and `model_override` resolved successfully — i.e. set `body["model"] = model` whenever `model` differs from whatever `body.get("model")` currently is (including when the key is missing entirely), not only when the existing value is already a string. This is the fix for the primary bug and would restore correct model forwarding on its own.
2. Separately, consider adding a `HEADROOM_VERTEX_MODEL_MAP`-style env override for `_VERTEX_MODEL_MAP`, mirroring the existing `HEADROOM_BEDROCK_MODEL_MAP`. Even with fix #1 in place, any model not already in the static map would still fail — an override gives users a way to unblock themselves on newly-released models without waiting for a Headroom release.

## Environment

- macOS, Apple Silicon
- Python 3.14
- headroom-ai 0.31.0 / 0.32.0
- litellm 1.83.7
- Claude Code CLI, Vertex AI mode (`CLAUDE_CODE_USE_VERTEX=1`)

slxng1758:
## Description

Vertex `rawPredict` requests provide the model through the URL path rather than `body["model"]`. `handle_anthropic_messages` correctly resolved this model through `model_override`, but the existing mutation guard only updated the body when a `model` field already existed. Since Vertex requests do not include `body["model"]`, the resolved model was dropped, causing `LiteLLMBackend` to fall back to its default `claude-3-5-sonnet-v2@20241022` model.

Closes #2363

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Updated `headroom/proxy/handlers/anthropic.py` to propagate `model_override` into `body["model"]` for LiteLLM-backed Vertex requests when the original body does not include a model.
- Preserved native Vertex passthrough behavior by avoiding mutation of raw Vertex request bodies.
- Added regression tests covering model propagation, passthrough behavior, and missing-model edge cases.
- Left `headroom/backends/litellm.py` unchanged since the fallback is avoided once the model is correctly propagated upstream.

## Testing

### Test Output
```
$ uv run pytest tests/test_proxy_anthropic_vertex_model_override.py tests/test_vertex_claude_compression.py \
    tests/test_proxy_anthropic_model_sanitization.py tests/test_proxy_anthropic_cache_stability.py \
    tests/test_proxy_anthropic_compression_diagnostics.py tests/test_provider_vertex_runtime.py -q

tests/test_proxy_anthropic_vertex_model_override.py ...                  [  6%]
tests/test_vertex_claude_compression.py ............                     [ 30%]
tests/test_proxy_anthropic_model_sanitization.py .                       [ 32%]
tests/test_proxy_anthropic_cache_stability.py .........................  [ 83%]
tests/test_proxy_anthropic_compression_diagnostics.py ..                 [ 87%]
tests/test_provider_vertex_runtime.py ......                             [100%]
49 passed, 1 warning in 4.53s

$ uv run mypy headroom/proxy/handlers/anthropic.py

$ uv run ruff check headroom/proxy/handlers/anthropic.py tests/test_proxy_anthropic_vertex_model_override.py
All checks passed!

$ uv run ruff format --check headroom/proxy/handlers/anthropic.py tests/test_proxy_anthropic_vertex_model_override.py
2 files already formatted
```

## Real Behavior Proof

- Environment: macOS; repo dev venv used for handler tests; separate Python 3.12 venv used for LiteLLM backend verification because `litellm` is excluded under Python 3.14 in `pyproject.toml`.
- Exact command / steps: Ran the real FastAPI app via `uvicorn` on a real socket with the actual `LiteLLMBackend(provider="vertex_ai")` attached. Stubbed only `litellm.acompletion` to inspect the final model passed to LiteLLM without requiring real GCP/Vertex credentials. Sent `curl -X POST` requests to a Vertex `rawPredict` route with no `model` key in the request body, first against pre-fix `main`, then against the fix.
- Observed result: before and after:
```
  # before: bug reproduces
  body["model"] received by backend = None

  # after: URL model claude-opus-4-6
  map_model_id() -> litellm.acompletion(model='vertex_ai/claude-opus-4-6')

  # after: URL model claude-3-5-sonnet-20241022 (old hardcoded default, for cross-check)
  map_model_id() -> litellm.acompletion(model='vertex_ai/claude-3-5-sonnet-v2@20241022')
  # matches the exact string the original issue's logs showed on *every* request

  # no backend configured (native Vertex passthrough): body unaffected
  forwarded body keys = ['anthropic_version', 'max_tokens', 'messages', 'stream']
```
- Not tested: real GCP/Vertex credentials (verification stops at the `acompletion` call
  boundary); `streamRawPredict` wasn't driven live, though the fix is in shared pre-branch
  code and covered by existing `test_vertex_claude_compression.py` streaming tests.

## Review Readiness

## Checklist

## Additional Notes

The 2nd suggested fix in the issue is out of scope for this PR but I would be happy to open an issue and address this in a new PR if wanted:
```adding a HEADROOM_VERTEX_MODEL_MAP-style env override for _VERTEX_MODEL_MAP, mirroring the existing HEADROOM_BEDROCK_MODEL_MAP. Even with fix https://github.com/headroomlabs-ai/headroom/issues/1 in place, any model not already in the static map would still fail — an override gives users a way to unblock themselves on newly-released models without waiting for a Headroom release.```

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -737,7 +737,13 @@ class AnthropicHandlerMixin:
                 sanitize_anthropic_model_id(raw_model) if isinstance(raw_model, str) else raw_model
             )
             body_model = body.get("model")
-            if isinstance(body_model, str) and model != body_model:
+            # Only inject model for LiteLLM-backed requests: native Vertex passthrough
+            # preserves the rawPredict wire format, where the model only exists in the URL.
+            if (isinstance(body_model, str) and model != body_model) or (
+                body_model is None
+                and model_override is not None
+                and self.anthropic_backend is not None
+            ):
                 body["model"] = model
                 body_mutation_tracker.mark_mutated("sanitize_model_id")
             messages = body.get("messages", [])
--- a/tests/test_proxy_anthropic_vertex_model_override.py
+++ b/tests/test_proxy_anthropic_vertex_model_override.py
@@ -1,0 +1,177 @@
+"""Regression coverage for #2363: litellm-vertex silently defaulting to
+claude-3-5-sonnet-20241022 because the URL-derived Vertex model never made it
+into ``body["model"]`` before reaching the backend.
+
+Vertex's rawPredict/streamRawPredict wire format carries the model as a URL
+path segment, never as a body field, so ``handle_anthropic_messages``'s
+mutation guard (which only fired when an existing ``body["model"]`` string
+needed sanitizing) silently dropped the resolved model whenever the body had
+no ``model`` key at all -- which is every Vertex request.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import httpx
+import pytest
+
+pytest.importorskip("fastapi")
+
+from fastapi.testclient import TestClient
+
+from headroom.backends.base import BackendResponse
+from headroom.proxy.server import ProxyConfig, create_app
+
+
+def _vertex_body() -> dict[str, Any]:
+    # Real Vertex rawPredict shape: no top-level "model" key.
+    return {
+        "anthropic_version": "vertex-2023-10-16",
+        "max_tokens": 10,
+        "stream": False,
+        "messages": [{"role": "user", "content": "2+2"}],
+    }
+
+
+class _FakeAnthropicBackend:
+    """Stand-in for LiteLLMBackend: only cares what body it's handed."""
+
+    name = "fake-litellm-vertex"
+
+    def __init__(self) -> None:
+        self.captured_body: dict[str, Any] | None = None
+
+    async def send_message(self, body: dict[str, Any], headers: dict[str, str]) -> BackendResponse:
+        del headers
+        self.captured_body = dict(body)
+        return BackendResponse(
+            body={
+                "id": "msg_fake",
+                "type": "message",
+                "role": "assistant",
+                "model": body.get("model", "unknown"),
+                "content": [{"type": "text", "text": "ok"}],
+                "usage": {
+                    "input_tokens": 10,
+                    "output_tokens": 1,
+                    "cache_read_input_tokens": 0,
+                    "cache_creation_input_tokens": 0,
+                },
+            },
+            status_code=200,
+        )
+
+
+def _app() -> Any:
+    return create_app(
+        ProxyConfig(
+            optimize=False,
+            cache_enabled=False,
+            rate_limit_enabled=False,
+            cost_tracking_enabled=False,
+            log_requests=False,
+            ccr_inject_tool=False,
+            ccr_handle_responses=False,
+            ccr_context_tracking=False,
+            image_optimize=False,
+        )
+    )
+
+
+def test_vertex_rawpredict_propagates_url_model_to_backend_when_body_has_none() -> None:
+    """The primary bug: with a backend configured (e.g. litellm-vertex), the
+    model resolved from the Vertex URL path must reach body["model"] even
+    though the request body never had a "model" key."""
+    app = _app()
+    fake_backend = _FakeAnthropicBackend()
+    app.state.proxy.anthropic_backend = fake_backend  # type: ignore[assignment]
+
+    client = TestClient(app)
+    response = client.post(
+        "/v1/projects/p/locations/europe-west1/publishers/anthropic/models/"
+        "claude-sonnet-4-6:rawPredict",
+        json=_vertex_body(),
+    )
+
+    assert response.status_code == 200, response.text
+    assert fake_backend.captured_body is not None
+    assert fake_backend.captured_body["model"] == "claude-sonnet-4-6"
+
+
+def test_vertex_native_passthrough_body_unaffected_when_no_backend_configured() -> None:
+    """Guardrail for the scoping decision: when no backend is configured
+    (self.anthropic_backend is None, e.g. plain `--backend vertex`), the body
+    forwarded to the real Vertex endpoint must stay byte-for-byte as before --
+    no "model" key should be injected into a shape Vertex has never received
+    from Headroom."""
+    app = _app()
+    assert app.state.proxy.anthropic_backend is None
+
+    captured: dict[str, Any] = {}
+
+    async def _fake_retry(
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest source. The fix is appropriately scoped: backend-backed Vertex requests get the URL-derived model propagated into `body["model"]`, while native Vertex `rawPredict` passthrough keeps the original body shape. The guard also avoids writing the `unknown` fallback into malformed direct requests, which is the right failure posture.

Verified locally: `uv run --extra dev pytest tests/test_proxy_anthropic_vertex_model_override.py tests/test_vertex_claude_compression.py tests/test_proxy_anthropic_model_sanitization.py tests/test_provider_vertex_runtime.py -q` (22 passed), `uv run --extra dev ruff check headroom/proxy/handlers/anthropic.py tests/test_proxy_anthropic_vertex_model_override.py`, and `uv run mypy headroom/proxy/handlers/anthropic.py --ignore-missing-imports --no-incremental`.

GitHub's check rollup still shows several cancelled PR-governance jobs from superseded pushes, so the PR is not visually clean even though I did not find a code issue in this review.

slxng1758:
@JerrettDavis thanks for the review. I've squashed the two commits down to one and force-pushed, so the cancelled-check clutter from before should be gone now — just the two fresh label/template runs on the single commit. Still showing approved + no conflicts on my end. Looks like the only thing left is the 4 workflows waiting on maintainer approval — happy to have those cleared and merge whenever's convenient.

JerrettDavis (approved):
Re-reviewed the current head after the Vertex model propagation update.

The backend-only injection gate is the right boundary: LiteLLM-backed Vertex requests now receive the URL-derived model in `body["model"]`, while native Vertex rawPredict passthrough stays on the original wire shape without an injected body model. The added tests cover the fixed path, the native passthrough guardrail, and the malformed/no-override case. Latest checks are green.

PR #2369, opened 2026-07-18
fix(transforms/cross-turn-dedup): don't renumber-fold zero-padded line prefixes

ebarkhordar:
## Description

On an HTTP tool-output re-read, `cross_turn_dedup` folds a contiguous span that
already appeared in an earlier block into a compact pointer, and when the line
numbers shifted by a constant it carries the offset as a `delta` so the original
bytes recover as `int(number) + delta`. The module states this renumber path is
"strictly lossless" for UNPADDED numbers only.

`_LINENO_RE = ^(\d+)(:|\t)(.*)$` does not enforce the "unpadded" restriction: `\d+`
also matches a LEADING-ZERO prefix. A timestamped log row such as `08:00:01 ...`
is read as line number `8`, not as data, so a re-read shifted by a constant (a
later window of the same hourly log) folds under a uniform delta. Recovery then
renders `str(int("08") + 1)` = `"9"`, not `"09"`: the round-trip is not
byte-exact. This is a lossy (false-positive) fold in a module whose stated
posture is to prefer false negatives (`CONTRIBUTING.md:129`,
`cross_turn_dedup.py:45-50`).

## Type of Change

## Changes Made

- `headroom/transforms/cross_turn_dedup.py`: restrict `_LINENO_RE` to `[1-9]\d*`
  so a leading-zero run stays non-numbered and can fold only on an EXACT match
  (delta 0), never under a lossy renumber. Real `grep -n` / `sed -n` / `rg -n`
  numbers never carry a leading zero, so the intended renumber-fold feature is
  unchanged. Added a comment stating why the character class is load-bearing.
- `tests/test_cross_turn_dedup.py`: added a delta-aware reconstruction helper and
  three regression tests (the existing `_reconstruct` asserts delta is absent, so
  it never exercised the numbered path this bug lives on).

## Testing

Three named scenarios, one test each:
1. `test_zero_padded_prefix_not_folded_lossily`: a padded shifted re-read is left
   verbatim (`spans_folded == 0`). This fails on `main` (it folds under a delta).
2. `test_unpadded_renumber_still_folds_and_recovers_exactly`: an unpadded `grep -n`
   read renumbered by `+5` still folds and reconstructs byte-exact (feature guard).
3. `test_padded_content_exact_redisplay_still_folds`: the same padded rows
   re-displayed verbatim still fold with delta 0 (surgical-scope guard).

### Test Output

```text
--- ruff check ---
All checks passed!
--- ruff format --check ---
2 files already formatted
--- mypy ---
--- pytest: 3 new tests on the branch (fixed) ---
3 passed, 14 deselected
--- pytest: revert regex to \d+ (simulate main): the regression test must FAIL ---
1 failed
```

## Real Behavior Proof

- Environment: clean `python:3.12-slim` Docker, `PYTHONPATH` at the source tree,
  core deps installed by name (tiktoken, pydantic, litellm, click, rich,
  opentelemetry-api, pyyaml, tomlkit), `ruff==0.15.17`, `mypy==1.20.2`.
- Exact command / steps: import the module and print provenance, then run ruff,
  ruff format, mypy on the two changed files, then `pytest` the three new tests
  on the branch, then revert only the regex to `\d+` and re-run the regression
  test.
- Observed result: module `cross_turn_dedup.py` (sha256 7ff204b65f40617d505895841aa1cfc1ee54bf63ffe6520198f0aa05a850aa8d), regex now `^([1-9]\d*)(:|\t)(.*)$`; ruff, ruff format, mypy all green; branch `3 passed`, reverted-regex main `1 failed`. Breakdown:
  - `module: /src/headroom/transforms/cross_turn_dedup.py`
  - `sha256: 7ff204b65f40617d505895841aa1cfc1ee54bf63ffe6520198f0aa05a850aa8d`
  - `regex : ^([1-9]\d*)(:|\t)(.*)$`
  - ruff, ruff format, mypy: all green (output above).
  - Branch: `3 passed`. Reverted-regex main: the regression test `1 failed`.
- Not tested: the router-level and Rust-backed integration tests in this file
  (`test_apply_*`, `test_dedup_*`) need the compiled `headroom._core` extension,
  which is not built in this lightweight container; they are `ModuleNotFoundError`
  on both `main` and this branch here, so they were not exercised. The change is a
  pure-stdlib regex in a pure-stdlib function; the unit-level `dedup_blocks` tests
  above cover it directly. I also did not measure how often real-world tool output
  hits the leading-zero shifted shape; the argument is the module's own
  strictly-lossless contract, not observed field frequency.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A: no doc surface)

## Additional Notes

Per `CONTRIBUTING.md` ("Bug or small fix -> Open a PR with repro + test"), this
goes straight to a PR rather than an issue. One concern only; no dependency or
generated-file changes.

### DIFF
--- a/headroom/transforms/cross_turn_dedup.py
+++ b/headroom/transforms/cross_turn_dedup.py
@@ -48,7 +48,17 @@ MAX_ANCHOR_CANDIDATES = 16
 # NOT match a padded/right-aligned prefix (``   123<TAB>`` from ``cat -n``): the
 # line must start with the digit, so ``number + delta`` re-numbers byte-exactly
 # without touching alignment padding, keeping renumbered folds strictly lossless.
-_LINENO_RE = re.compile(r"^(\d+)(:|\t)(.*)$", re.DOTALL)
+#
+# ``[1-9]\d*`` (not ``\d+``) is load-bearing: a LEADING-ZERO run (``08:00:01`` in
+# a timestamped log, ``007:...``) is not an unpadded grep/sed line number, and
+# recovery is ``str(int(number) + delta)``, which drops the zero pad — ``int("08")
+# + 1`` renders ``"9"``, not ``"09"``. Matching those would fold them under a
+# uniform delta and the round-trip would NOT be byte-exact, breaking the strictly
+# lossless promise above. Excluding them keeps such lines non-numbered, so they
+# fold only on an EXACT match (delta 0) — the "prefer false negatives" posture.
+# Real grep -n / sed -n / rg -n numbers never carry a leading zero, so the
+# intended renumber-fold feature is unaffected.
+_LINENO_RE = re.compile(r"^([1-9]\d*)(:|\t)(.*)$", re.DOTALL)
 
 
 def _num_and_key(line: str) -> tuple[int | None, str, str]:
--- a/tests/test_cross_turn_dedup.py
+++ b/tests/test_cross_turn_dedup.py
@@ -130,6 +130,100 @@ def test_info_preserving_reconstruction_multiref():
 
 
 # --------------------------------------------------------------------------
+# Numbered (renumber-fold) path: a leading line-number lets the same content
+# fold across a uniform shift, with the offset carried in the pointer so the
+# original bytes recover as ``str(int(number) + delta)``. That recovery is
+# byte-exact ONLY for UNPADDED numbers: a leading-zero prefix (a timestamped
+# log row, ``08:00:01``) loses its pad on renumber (``int("08") + 1`` -> ``"9"``,
+# not ``"09"``), so it must not fold under a delta. The helper below renumbers
+# exactly as the module documents, so the round-trip assertion is faithful.
+# --------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest source. The regex restriction is the right surgical fix for the lossless contract: unpadded grep/sed/rg-style prefixes still participate in delta folding, while leading-zero prefixes are treated as ordinary content and therefore cannot be lossy-renumbered. The new tests cover the regression, the intended unpadded renumber path, and exact padded redisplay folding.

Verified locally: `uv run --extra dev pytest tests/test_cross_turn_dedup.py -q` (17 passed), `uv run --extra dev ruff check headroom/transforms/cross_turn_dedup.py tests/test_cross_turn_dedup.py`, and `uv run mypy headroom/transforms/cross_turn_dedup.py --ignore-missing-imports --no-incremental` (success). GitHub currently reports governance checks for this PR; I did not ask for CI to be triggered.

PR #2373, opened 2026-07-18
feat(proxy): select built-in compressors via --compressor + registry inventory

chopratejas:
## What
- Adds an opt-in `--compressor` / `HEADROOM_COMPRESSORS` selection that narrows the active built-in compressors, mapped onto the existing `ContentRouterConfig` `enable_*` flags at the proxy config seam. Recognized names: `smart_crusher, kompress, code_aware, search, log, tabular, config, html, image`; `"*"` selects all.
- Builds a name-addressable compressor registry in `ContentRouter`: a metadata-only descriptor per built-in plus opt-in discovery of `headroom.compressor` entry points (the seam added in #2370).

## Why
Built-in compressors were only reachable through a hardcoded if/elif; there was no supported way to select a subset (7 of the `enable_*` flags had no external surface) or to see the built-ins as a name-addressable set alongside third-party ones.

## Behavior change
**None by default.** `--compressor` unset (the default) leaves every `enable_*` flag at its dataclass default, so the request path is byte-identical to today. The registry is inventory-only — built-ins are still constructed and dispatched by the existing if/elif; `_BuiltinCompressorEntry.compress` deliberately raises (never called), and registry construction is fail-open. Routing an external compressor *through* the pipeline is a deliberate follow-up.

## How
- `server.py`: `BUILTIN_COMPRESSOR_FLAGS` map + `_apply_compressor_selection(router_config, compressors)` (no-op when `None`/empty; runs before the `disable_kompress` override so that stays authoritative).
- `models.py`: `ProxyConfig.compressors: set[str] | None = None`.
- `cli/proxy.py`: `--compressor` (repeatable, comma-split, `HEADROOM_COMPRESSORS`), mirroring `--proxy-extension`.
- `content_router.py`: built-in descriptors + `_build_compressor_registry()` (register built-ins, then fail-open `discover()`), exposed as `self.compressor_registry`. Dispatch unchanged.

## Testing
`tests/test_compressor_selection.py` — 23 tests: selection mapping (None/empty/whitespace = byte-identical defaults, single/multi/wildcard, external-only disables built-ins, unrecognized ignored), `ProxyConfig` field, registry inventory (descriptors cover the 9 names, valid cost tiers, router exposes registry, inventory doesn't auto-activate, built-in `compress` guard, discovery merges external, fail-open on discovery error). Local: 23 passed; ruff + mypy clean on changed files. Full suite runs in CI.

Stacks conceptually on #2370 (registry seam); rebased onto `main` after that merged.

### DIFF
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -352,6 +352,20 @@ def dashboard(port: int, no_open: bool) -> None:
     ),
 )
 @click.option(
+    "--compressor",
+    "compressor",
+    multiple=True,
+    envvar="HEADROOM_COMPRESSORS",
+    help=(
+        "Restrict the active built-in compressors to the named set (opt-in). "
+        "Repeat the flag or pass a comma-separated list; recognized names are "
+        "smart_crusher, kompress, code_aware, search, log, tabular, config, "
+        "html, image. Unselected built-ins are disabled; '*' selects all. "
+        "Omit to keep every compressor enabled (default). "
+        "Env: HEADROOM_COMPRESSORS."
+    ),
+)
+@click.option(
     "--no-subscription-tracking",
     is_flag=True,
     envvar="HEADROOM_NO_SUBSCRIPTION_TRACKING",
@@ -926,6 +940,7 @@ def proxy(
     lossless: bool,
     no_ccr_proactive_expansion: bool,
     proxy_extension: tuple[str, ...],
+    compressor: tuple[str, ...],
     no_subscription_tracking: bool,
     subscription_poll_interval: int | None,
     retry_max_attempts: int | None,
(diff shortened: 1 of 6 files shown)


JerrettDavis (approved):
Reviewed the latest source, including the follow-up fix for the CI settings registry failure. The compressor selection stays mapped onto existing ContentRouterConfig enable flags, the registry remains inventory-only for built-ins, and HEADROOM_COMPRESSORS is now covered by the persisted settings registry so the Click/env surface does not drift.

Local verification:
- `uv run --extra dev --extra proxy pytest tests/test_proxy/test_settings_store.py::TestRegistryDriftAgainstClick::test_every_headroom_click_envvar_is_in_the_registry tests/test_compressor_selection.py -q` (24 passed)
- `uv run --extra dev ruff check headroom/settings_store.py headroom/cli/proxy.py headroom/proxy/models.py headroom/proxy/server.py headroom/transforms/content_router.py tests/test_compressor_selection.py tests/test_proxy/test_settings_store.py`
- `uv run --extra dev mypy headroom/settings_store.py headroom/cli/proxy.py headroom/proxy/models.py headroom/proxy/server.py headroom/transforms/content_router.py tests/test_compressor_selection.py`
- `uv run --extra dev ruff format --check headroom/settings_store.py`
- `git diff --check`

lifeodyssey:
Question on inventory scope: should `text_crusher` (the Rust prose compressor currently gated behind `HEADROOM_TEXT_CRUSHER`) get a built-in descriptor + `--compressor` name too? Right now it is reachable only through its env flag, so it is invisible to the new inventory/selection surface. Happy to PR the descriptor + enable flag if you want it addressable — equally fine leaving it env-gated while it is still experimental.

Issue #2340, opened 2026-07-17
[HIGH / potential fleet-wide quota exhaustion] v0.31.0 discards completed Claude responses, duplicates paid calls, and remains active after unwrap


spaceBrownie:
## Severity

**HIGH — potentially CRITICAL for organization-wide deployments**

In one observed session, five-hour Anthropic utilization rose from **2% to 88% in 10 minutes**, then reached **97% after 15 minutes**.

Headroom discarded four completed upstream responses and returned HTTP 502, causing Claude to submit the same large requests again. Headroom's own metrics attribute **703,934 input-side token events** and **3,931 output tokens** to responses that were generated upstream but never delivered.

This presents a fleet-level risk where Headroom is centrally installed, automatically updated, or used by multiple concurrent agents against a shared organization allowance.

## Summary

Headroom 0.31.0, running persistently on macOS with Claude Code 2.1.211, introduced a retry-amplification failure at the LLM protocol boundary.

Headroom injected its retrieval tool, converted streaming requests into buffered upstream requests, received complete responses from Anthropic, and then failed to translate responses containing `tool_search_tool_result` back into SSE. Headroom returned HTTP 502 after upstream inference had completed, causing Claude to submit the same large request again.

Four verified failure/reissue pairs occurred within approximately eight minutes.

Separately, Headroom's usage tracker substantially inflated its reported token total by counting the same request-level usage metadata once for every content block stored in Claude's transcript. One response reporting **420,609 input tokens was counted 19 times**, producing **7,991,571 attributed input tokens from that one response record alone**.

This explains most of the sudden **8–10 million additional tokens** shown during the incident. Headroom's stored five-hour window eventually reported **15,862,094 raw tokens** and **15,262,075 weighted token-equivalent**, but those totals were not counts of distinct upstream requests.

The quota depletion was nevertheless real: Anthropic utilization independently rose from 2% to 97%. The incident therefore contained two separate failures:

1. Real usage amplification caused by discarded completed responses and replacement calls.
2. Inflated Headroom reporting caused by duplicate transcript accounting.

## Environment

- Headroom: 0.31.0
- Claude Code: 2.1.211
- Installation: `uv tool`, `headroom-ai[all]`
- Platform: macOS
- Backend: Anthropic
- Persistent deployment using launchd
- Proxy mode: token optimization
- Proxy address: localhost
- Headroom dependency resolved locally: `ast-grep-cli==0.44.1`

## Impact at a Glance

[...]

PR #2408, opened 2026-07-19
fix(subscription): dedup transcript usage by message id (#2340 token inflation)

abhay-codes07:
## Description

Addresses the usage-inflation part of #2340.

`compute_window_tokens` (`headroom/subscription/session_tracking.py`) sums `message.usage` for every transcript line whose timestamp falls in the window:

```python
for line in _read_transcript_lines(path):
    ...
    usage = msg.get("usage")
    if not usage:
        continue
    _add_usage_to_tokens(totals, usage)
```

But Claude Code can store a single assistant response across **multiple transcript lines** (e.g. one entry per content block), and each of those lines carries the **same request-level `message.usage`**. Summing per line therefore multiplies that one response's tokens by its block count. #2340 observed a single 420,609-input-token response counted **19 times** (~8M attributed input tokens from one record), which is most of the reported window-total inflation.

## Fix

Count each response's usage once, keyed by the Anthropic `message.id` (unique per response):

```python
seen_message_ids: set[str] = set()
...
    msg_id = msg.get("id")
    if isinstance(msg_id, str) and msg_id:
        if msg_id in seen_message_ids:
            continue
        seen_message_ids.add(msg_id)
    _add_usage_to_tokens(totals, usage)
```

Entries without a `message.id` keep the previous per-line behavior, so this only ever removes true duplicates: a response is de-duplicated only when the exact same (unique) message id appears more than once, and distinct responses are unaffected.

Scope: this fixes the token-accounting inflation only. The separate retry-amplification / `tool_search_tool_result` SSE-502 behavior described in the same issue is a different code path and is not touched here.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/subscription/session_tracking.py`: dedup usage accumulation by `message.id` in `compute_window_tokens`.
- `tests/test_subscription_session_tracking.py`: add a test where one response is stored across three lines (plus a distinct response and an id-less line) and assert its usage is counted once.

## Testing

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/subscription/session_tracking.py tests/test_subscription_session_tracking.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/subscription/session_tracking.py
# session_tracking is import-light, so I ran the exact logic against the real
# module in the project venv (uv sync): a response stored on 3 lines yields
# input=106 (100 once + 5 + 1 id-less), not 306.
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, project venv (`uv sync --extra proxy`), `uvx ruff@0.15.17` / `uvx mypy@1.20.2`.
- Exact command / steps: wrote a transcript with `msg_dup` repeated on three lines (same usage, 100/10), one distinct `msg_other` (5/2), and one id-less line (1/1); called the real `compute_window_tokens` over the window.
- Observed result: `input == 106` and `output == 13` (the duplicated response counted once, the id-less line still counted); the pre-fix code would report `input == 306`. Because `session_tracking` has no heavy imports, this ran against the actual module.
- Not tested: a live Claude Code transcript with real multi-block responses; the added unit test reproduces the multi-line-per-response shape.

## Review Readiness

## Checklist

## Additional Notes

`session_tracking` is a light module, so I verified against the real code in the venv (output above) in addition to the added test. This is deliberately scoped to the usage-double-count sub-part of #2340; the retry-amplification/SSE side is separate and untouched. Keyed on `message.id` so it is safe by construction: no id or a unique id behaves exactly as before.

### DIFF
--- a/headroom/subscription/session_tracking.py
+++ b/headroom/subscription/session_tracking.py
@@ -134,6 +134,14 @@ def compute_window_tokens(start_ts: float, end_ts: float) -> WindowTokens:
     totals = WindowTokens()
     by_model: dict[str, WindowTokens] = {}
     unattributed = WindowTokens()
+    # Claude Code can store one assistant response across multiple transcript
+    # lines (e.g. one entry per content block), each carrying the SAME
+    # request-level ``message.usage``. Summing per line therefore multiplies a
+    # single response's tokens by its block count (observed 19x for one 420K
+    # response, #2340). Count each response's usage once, keyed by the unique
+    # Anthropic ``message.id``. Entries without an id keep the per-line
+    # behavior, so this only ever removes true duplicates.
+    seen_message_ids: set[str] = set()
 
     for path in find_transcript_files():
         for line in _read_transcript_lines(path):
@@ -160,6 +168,12 @@ def compute_window_tokens(start_ts: float, end_ts: float) -> WindowTokens:
             usage = msg.get("usage")
             if not usage:
                 continue
+
+            msg_id = msg.get("id")
+            if isinstance(msg_id, str) and msg_id:
+                if msg_id in seen_message_ids:
+                    continue
+                seen_message_ids.add(msg_id)
 
             _add_usage_to_tokens(totals, usage)
 
--- a/tests/test_subscription_session_tracking.py
+++ b/tests/test_subscription_session_tracking.py
@@ -87,3 +87,44 @@ def test_read_transcript_lines_preserves_small_transcript(
         '{"marker":"first"}',
         '{"marker":"second"}',
     ]
+
+
+def test_compute_window_tokens_dedups_usage_by_message_id(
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Looks good. I rebased this onto the current headroomlabs-ai/headroom main and pushed ac932ce3. I reviewed the placement of the message-id dedup: it happens after the timestamp window filter and before total/model aggregation, so repeated transcript lines for the same response are counted once while id-less entries keep the previous per-line behavior. Focused local checks passed: tests/test_subscription_session_tracking.py, ruff check, ruff format --check, and git diff --check. Upstream governance/commitlint/lint/build/build-wheel/security fast checks are green; broad test shards and Python CodeQL are still running at review time.

Issue #2394, opened 2026-07-18
[BUG] test_recover_codex socket test fails on macOS with "AF_UNIX path too long"


rNoz:
## Description

`tests/test_cli/test_recover_codex.py::test_recovery_records_sockets_and_secures_both_backups` fails on macOS. It binds a Unix domain socket at its absolute path under pytest's `tmp_path`. macOS caps the `AF_UNIX` `sun_path` at ~104 bytes, and the `tmp_path`-rooted absolute path exceeds that, so `bind()` raises `OSError: AF_UNIX path too long`. The suite stays green on CI Linux only because its `/tmp`-rooted temp paths are short enough to fit.

## To Reproduce

1. Check out `main` on macOS.
2. Run the single test:
   `uv run pytest tests/test_cli/test_recover_codex.py::test_recovery_records_sockets_and_secures_both_backups"`
3. See the `OSError`.

## Expected Behavior

The test passes regardless of how long the OS temp directory path is.

## Actual Behavior

`bind()` fails on macOS because the absolute socket path is longer than the `sun_path` limit; the test errors out.

## Code Sample

```bash
# macOS, from a clean main checkout
uv run pytest -q \
  "tests/test_cli/test_recover_codex.py::test_recovery_records_sockets_and_secures_both_backups"
```

## Error Output

```text
tests/test_cli/test_recover_codex.py:777: in test_recovery_records_sockets_and_secures_both_backups
    codex_socket.bind(str(socket_path))
E   OSError: AF_UNIX path too long
1 failed
```

## Environment

- **Headroom version**: 0.33.0-dev
- **Python version**: 3.13.7
- **OS**: macOS 26.5 (darwin 25.5.0); `tempfile.gettempdir()` = `/var/folders/.../T`
- **LLM Provider**: N/A (test-only, no model call)

## Additional Context

- Fix: bind a short relative socket name from inside `source`  (`monkeypatch.chdir(source)` + `bind(socket_path.name)`); the socket is still   created at `source/codex.sock` and the recovery scan is unchanged.
- Test-only; no production behavior change.

rNoz:
PR ready and tested, using it in my integration branch locally: https://github.com/headroomlabs-ai/headroom/pull/2396

PR #2430, opened 2026-07-20
fix(proxy/interceptors): flag truncated Read input in ast-grep outlin…

### LINKED ISSUE #2421: Read outliner can present a client-truncated file prefix as a complete definition list
# Read outliner can present a client-truncated file prefix as a complete definition list

## Problem

With `headroom-ai==0.32.1`, cache mode, and `--intercept-tool-results`, a default Claude Code `Read` of a TypeScript file containing exactly 140 top-level `export async function` blocks reached the model as an ast-grep outline headed:

```text
[headroom: outlined by ast-grep — 114 definition(s); bodies elided. Re-read the file with a line range to see a specific body.]
```

The outline stopped after definition 114 and contained no self-contained indication that definitions 115–140 were outside the source text Headroom received. A standalone consumer of the rewritten tool result can therefore read the `114 definition(s)` header as a complete inventory.

## Root cause and corrected cap attribution

The 25,000-token cap is not in Headroom's Python or Rust outline code. It is Claude Code's `Read` cap, applied before the request reaches Headroom.

The captured transcript records the first tool result as only source lines 1–1489. Its metadata says `numLines=1489`, `totalLines=1825`, and `truncatedByTokenCap=true`. A separate `read_truncation_notice` attachment says:

```text
[Truncated: PARTIAL view — .../big.ts: showing lines 1-1489 of 1825 total
(26031 tokens, cap 25000). Call Read with offset=1490 ...]
```

Line 1482 contains the last complete declaration, handler 114. Handler 115 begins after the capped prefix.

Headroom's path is:

1. `headroom/proxy/interceptors/base.py::apply_to_messages` extracts the tool-result text and passes that string to the registered interceptor.
2. `headroom/proxy/interceptors/astgrep.py::AstGrepReadOutline.transform` scans that already-truncated string.
3. `headroom/proxy/interceptors/astgrep.py::_build_outline` iterates every supplied match and reports `len(seen_starts)`.

`_build_outline` has no token counter, budget, slice, `break`, or cap-hit exit. Feeding the captured 1,489-line tool result to the shipped ast-grep produces 114 matches; feeding the complete 1,824-line fixture produces 140. The installed pure-Python `CodeAwareCompressor` also preserves all 140 signatures in a complete-file control run. Rust only classifies TypeScript as source code; `_core.abi3.so` exports no code-aware/AST/outline compressor.

The `114` count is therefore accurate for Headroom's input, but it is not the total for the file named by the `Read` call.

## Why this matters for agent consumers

The client's truncation status is separate from the tool-result string rewritten by the interceptor. Any adapter or replay path that does not preserve or co-locate that separate metadata leaves the model with a completeness-looking Headroom header and no way to infer the missing 26 declarations.

The captured Claude Code run did retain a separate `read_truncation_notice`, and the agent recovered by issuing ranged reads. That means this evidence does not establish a Headroom-internal cap or an end-to-end silent loss for every client. It does establish that the rewritten outline is not self-describing when its input is a partial Read.

Headroom cannot truthfully emit `114 of 140 definitions` or a CCR retrieval hash from the current interceptor input: it never received definitions 115–140 and did not store the complete file. Inventing either value would create a broken retrieval promise.

## Reproduction sketch

1. Create a roughly 26,000-token TypeScript file with 140 sequential top-level `export async function` blocks.
2. Run `headroom-ai==0.32.1` in cache mode with `--intercept-tool-results`.
3. Through Claude Code, issue a default full-file `Read` without `offset` or `limit`.
4. Observe Claude Code cap the tool result at 25,000 tokens and emit a separate pagination notice.
5. Observe Headroom outline the capped prefix and emit the `114 definition(s)` header without carrying partial-input status in that header.
6. Repeat ast-grep against the complete file; all 140 definitions are returned.

## Suggested resolution

Carry structured source-truncation metadata into the tool-result interceptor contract, or preserve a recognized upstream truncation notice in the rewritten output. The outline header should state that it describes a partial Read and include only counts Headroom can prove, for example the visible definition count plus source line/page totals supplied by the client. A CCR marker should be used only if Headroom actually stores retrievable omitted content.

Add an integration test in `tests/test_tool_result_interceptors.py` that supplies explicit truncation metadata alongside a capped Read and asserts that the rewritten outline preserves a visible partial-input marker. A separate complete-file control should continue to report all 140 definitions with no truncation marker.

## Release provenance finding

The installed distribution metadata reports `headroom-ai==0.32.1`, while the repository's `pyproject.toml` and release manifest remain at `0.32.0` and local tags stop at `v0.32.0`. The relevant outline files in the 0.32.1 installation are byte-identical to current `main` (for example, `proxy/interceptors/astgrep.py` has SHA-256 prefix `ed4a905e...`), and the compiled extension contains no hidden outline implementation. Thus the version/tag divergence is real but is not causal for this specimen.

Please publish or identify the source revision used for 0.32.1 and add a matching source tag so wheel-to-source provenance is independently auditable.

## Fix status

N/A for the originally requested marker-on-Headroom-cap patch: no Headroom outline cap exists in repository source or in the shipped compiled artifact. A code change requires a separate design decision about propagating upstream tool truncation metadata; it should not fabricate shown/total definition counts or CCR retrieval state.

slxng1758:
## Description

`AstGrepReadOutline` generated its `"[headroom: outlined by ast-grep — N definition(s); ...]"` header solely from the text it received. When an upstream client (such as Claude Code) truncated a large Read before Headroom processed it, the header still reported N as if it represented the entire file. Downstream consumers had no indication that the outline covered only the visible portion.

This change detects the existing upstream truncation banner already present in the `tool_result` and qualifies the header accordingly. When no truncation signal is present, behavior is unchanged.

Closes #2421

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/interceptors/astgrep.py`: added `_TRUNCATION_RE` + `_detect_truncation()` to recognize an upstream truncation banner (e.g. `"...showing lines 1-1489 of 1825 total..."`) already inline in `tool_output`. `transform()` computes this once and passes it through. `_build_outline()` takes a new optional `truncation` parameter and only changes the header wording when a signal is present, no change to the existing header when it isn't.
- `tests/test_tool_result_interceptors.py`: added `test_astgrep_flags_truncated_read` (truncated input → header states the input was truncated upstream, with the visible-count + line/total numbers). Extended the existing `test_astgrep_outlines_large_python_read` with an assertion that the marker is **absent** on a complete file.
- No changes to `matches()` or the `ToolResultInterceptor` protocol — truncation doesn't change interceptor eligibility, only header content, so the blast radius is contained to this one file.

## Testing

### Test Output

```text
$ pytest tests/test_tool_result_interceptors.py -v
22 passed in 0.59s (includes 2 new: test_astgrep_flags_truncated_read,
extended test_astgrep_outlines_large_python_read)

$ ruff check headroom/proxy/interceptors/astgrep.py tests/test_tool_result_interceptors.py
All checks passed!

$ mypy headroom/proxy/interceptors/astgrep.py

Full suite (pytest -q): 9585 passed, 1 failed, 562 skipped. 
```
Failure reason: AF_UNIX path too long (macOS-only Unix socket path length limitation). This is tracked separately in issue #2394 and is unrelated to this change. The failing test is not executed in CI. The macos-latest CI job only covers installer tests, while the main suite runs on ubuntu-latest where this limitation does not occur.

## Real Behavior Proof

- Environment: macOS development machine. Local stub HTTP server replacing Anthropic API (no external network calls or API usage). Proxy launched with --intercept-tool-results and --mode token
- Exact command / steps: generated a 699-line/140-function TypeScript fixture mirroring the issue's own repro sketch ("~26,000-token file, 140 top-level `export async function` blocks"). Built a truncated variant (first 114 functions + the real upstream truncation banner text — its exact template was confirmed by extracting the literal string from the locally installed Claude Code CLI binary, not guessed). Sent both the truncated and the complete file as a `Read` tool_result via `POST /v1/messages` to the running proxy, and inspected exactly what the proxy forwarded to the stub upstream.
- Observed result: shown
```text
  Truncated case:  [headroom: outlined by ast-grep — 114 definition(s) in the visible portion;
                    input was truncated upstream (showing through line 570 of 699 total).
                    Bodies elided. Re-read remaining lines to see more.]

  Complete case:   [headroom: outlined by ast-grep — 140 definition(s); bodies elided.
                    Re-read the file with a line range to see a specific body.]
  ```
- Not tested: A live Claude Code session or real Anthropic API request. A local stub was used to avoid consuming API quota. Manual reproduction requires --mode token (or a multi-turn session); the default --mode cache intentionally skips compression for one-shot requests to preserve provider prompt-cache behavior.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/interceptors/astgrep.py
+++ b/headroom/proxy/interceptors/astgrep.py
@@ -89,12 +89,14 @@ _PATTERNS: dict[str, list[str]] = {
 
 OUTLINE_MARKER = "    # ... (body elided by Headroom; Read a specific line range to see it)\n"
 
-# Matches the upstream truncation banner a client (e.g. Claude Code) appends to a
-# Read tool_result when its own token cap — independent of any Headroom cap — cuts
-# the file short, e.g. "...showing lines 1-1489 of 1825 total...". Captures
-# (end_line, total_lines) — the only numbers Headroom can honestly report, since
-# they come from the client's own accounting, not a Headroom guess.
-_TRUNCATION_RE = re.compile(r"showing lines \d+-(?P<end_line>\d+) of (?P<total_lines>\d+) total")
+# Matches a line-based truncation notice emitted by a client
+# when its own token cap cuts a Read result short, e.g. "showing lines 1-1489
+# of 1825 total". Captures the client-provided line counts that Headroom can
+# safely report without guessing. Tolerates common formatting variations
+_TRUNCATION_RE = re.compile(
+    r"showing\s+lines?\s+\d+\s*[-–]\s*(?P<end_line>\d+)\s+of\s+(?P<total_lines>\d+)\s+total",
+    re.IGNORECASE,
+)
 
 
 def _detect_truncation(source: str) -> tuple[int, int] | None:
--- a/headroom/proxy/interceptors/astgrep.py
+++ b/headroom/proxy/interceptors/astgrep.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import re
 import shutil
 import subprocess
 import tempfile
@@ -88,6 +89,22 @@ _PATTERNS: dict[str, list[str]] = {
 
 OUTLINE_MARKER = "    # ... (body elided by Headroom; Read a specific line range to see it)\n"
 
+# Matches the upstream truncation banner a client (e.g. Claude Code) appends to a
+# Read tool_result when its own token cap — independent of any Headroom cap — cuts
+# the file short, e.g. "...showing lines 1-1489 of 1825 total...". Captures
+# (end_line, total_lines) — the only numbers Headroom can honestly report, since
+# they come from the client's own accounting, not a Headroom guess.
+_TRUNCATION_RE = re.compile(r"showing lines \d+-(?P<end_line>\d+) of (?P<total_lines>\d+) total")
+
+
+def _detect_truncation(source: str) -> tuple[int, int] | None:
+    """Return (end_line, total_lines) if `source` carries an upstream truncation
+    banner, else None."""
+    m = _TRUNCATION_RE.search(source)
+    if not m:
+        return None
+    return int(m.group("end_line")), int(m.group("total_lines"))
+
 
 class AstGrepReadOutline:
     """Interceptor that outlines verbose code-file Read outputs."""
@@ -131,7 +148,8 @@ class AstGrepReadOutline:
         if not matches:
             return None
 
-        outline = _build_outline(matches, tool_output)
+        truncation = _detect_truncation(tool_output)
+        outline = _build_outline(matches, tool_output, truncation)
         return outline if outline else None
 
     def progressive_disclosure_key(
@@ -258,12 +276,21 @@ def _run_ast_grep(
     return all_matches
 
 
-def _build_outline(matches: list[dict[str, Any]], source: str) -> str | None:
+def _build_outline(
+    matches: list[dict[str, Any]],
+    source: str,
+    truncation: tuple[int, int] | None = None,
+) -> str | None:
     """Build a compact outline from ast-grep matches.
 
     Emits each definition's signature line + docstring (if next line is a
     string literal) + an elision marker. Matches are sorted by byte offset
     so the outline tracks the original file order.
+
+    `truncation`, if given, is (end_line, total_lines) from an upstream
+    truncation banner already present in `source` (e.g. a client's own Read
+    token-cap notice). When set, the header states that the input was a
+    partial view instead of implying `source` is the whole file.
     """
     lines = source.splitlines(keepends=True)
     outline_chunks: list[str] = []
@@ -292,11 +319,21 @@ def _build_outline(matches: list[dict[str, Any]], source: str) -> str | None:
 
     if not outline_chunks:
         return None
-    header = (
-        "[headroom: outlined by ast-grep — "
-        f"{len(seen_starts)} definition(s); "
-        "bodies elided. Re-read the file with a line range to see a specific body.]\n"
-    )
+
+    if truncation:
+        end_line, total_lines = truncation
+        header = (
+            "[headroom: outlined by ast-grep — "
+            f"{len(seen_starts)} definition(s) in the visible portion; "
+            f"input was truncated upstream (showing through line {end_line} of {total_lines} total). "
+            "Bodies elided. Re-read remaining lines to see more.]\n"
+        )
+    else:
+        header = (
+            "[headroom: outlined by ast-grep — "
+            f"{len(seen_starts)} definition(s); "
+            "bodies elided. Re-read the file with a line range to see a specific body.]\n"
+        )
     return header + "".join(outline_chunks)
 
 
--- a/headroom/proxy/interceptors/astgrep.py
+++ b/headroom/proxy/interceptors/astgrep.py
@@ -89,23 +89,43 @@ _PATTERNS: dict[str, list[str]] = {
 
 OUTLINE_MARKER = "    # ... (body elided by Headroom; Read a specific line range to see it)\n"
 
-# Matches a line-based truncation notice emitted by a client
-# when its own token cap cuts a Read result short, e.g. "showing lines 1-1489
-# of 1825 total". Captures the client-provided line counts that Headroom can
-# safely report without guessing. Tolerates common formatting variations
-_TRUNCATION_RE = re.compile(
-    r"showing\s+lines?\s+\d+\s*[-–]\s*(?P<end_line>\d+)\s+of\s+(?P<total_lines>\d+)\s+total",
-    re.IGNORECASE,
+# Per-client banner signatures -- add an entry only once a client's exact
+# banner text is confirmed, never a generic keyword match.
+_TRUNCATION_SIGNATURES: tuple[re.Pattern[str], ...] = (
+    # Claude Code: "[Truncated: PARTIAL view -- <path>: showing lines A-B of
+    # T total (...). Call Read with offset=N to see more.]"
+    re.compile(
+        r"\[\s*truncated\s*:\s*partial\s+view\b"
+        r"[^\[\]]*?"
+        r"showing\s+lines?\s+(?P<start_line>\d+)\s*[-–]\s*(?P<end_line>\d+)"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: headroom/proxy/interceptors/astgrep.py [hunk 1]
```diff
@@ -88,6 +89,22 @@ def _min_chars_to_rewrite() -> int:
 
 OUTLINE_MARKER = "    # ... (body elided by Headroom; Read a specific line range to see it)\n"
 
+# Matches the upstream truncation banner a client (e.g. Claude Code) appends to a
+# Read tool_result when its own token cap — independent of any Headroom cap — cuts
+# the file short, e.g. "...showing lines 1-1489 of 1825 total...". Captures
+# (end_line, total_lines) — the only numbers Headroom can honestly report, since
+# they come from the client's own accounting, not a Headroom guess.
+_TRUNCATION_RE = re.compile(r"showing lines \d+-(?P<end_line>\d+) of (?P<total_lines>\d+) total")
```
chopratejas: This is specific to only Claude code - right? is there a way to make this generic?
Ideally relying on pattern by Coding Agents is not scalable

slxng1758: Following up on the genericity comment, I wanted to share what I looked into and get your input on how to move forward.

For the scope of this PR, I've went ahead and loosened `_TRUNCATION_RE` from matching Claude Code's exact sentence to matching the general *shape* of a truncation banner. Any client emitting "showing lines X-Y of Z total" in roughly that form now matches automatically,. 

However this doesn't achieve full client-independence. For ex, I checked it against Cursor's actual installed bundle (64MB workbench build, searched directly) and found zero matches for any variant of "showing lines... of... total"; its real truncation signal is byte-based and reports the opposite direction (drops the start, not the end). OpenCode's convention (`headroom/learn/plugins/opencode.py`) is a placeholder string plus an external file pointer.

The looser regex is still useful even if we eventually move to a more general approach: it prevents the current fix from being tied to one exact client string and will automatically handle any client that emits the same style of line-count notice.

**For the real generic case heres what I investigated:**

1. **A per-client pattern registry + threading it through the interceptor contract** (`ToolResultInterceptor.transform()` gaining a structured `truncation` param, `apply_to_messages` computing it once for all interceptors). This improves extensibility, but detection is still dependent on having verified client-specific formats.

2. **A content-based structural heuristic** —  Parse-error detection near EOF (ast-grep/tree-sitter), I've tested how well this works (100% of TypeScript cuts, 79% of Python cuts), but it cannot distinguish a truncated file from a file that is temporarily incomplete because it is being edited. That creates false positives for a common coding-agent workflow.

3. **Getting the client to send structured metadata instead of prose** — Claude Code does track `truncatedByTokenCap`/`numLines`/`totalLines` internally, so I checked whether Headroom could just receive that instead of parsing text. Confirmed from Claude Code's source that these fields are part of internal state and are not serialized into the API request; only the rendered prose banner reaches Headroom. Supporting this would require an upstream client/protocol change.

4. **Filesystem verification** — verifying the Read result against the file on disk and comparing the visible content with the source file. Since headroom runs locally alongside the client it wraps, it can read the same file and compare. This is buildable without waiting on any upstream protocol change, but it introduces new concerns (filesystem access, path resolution, race conditions, remote/container environments).

If filesystem verification/per-client pattern registry or another direction seems worth pursuing, I'd be happy to open a separate issue/PR or continue expanding this one depending on what scope makes the most sense.

JerrettDavis (approved):
Reviewed the latest head commit (89b17c42). The update keeps the truncation signal as a header-only qualification, leaves interceptor eligibility unchanged, and the broadened regex addresses the earlier feedback without pretending to solve every client-specific truncation format.

Verified locally in an isolated worktree:
- uv run ruff check headroom/proxy/interceptors/astgrep.py tests/test_tool_result_interceptors.py
- uv run pytest tests/test_tool_result_interceptors.py -q -k astgrep

I also ran the full tests/test_tool_result_interceptors.py file in the default uv environment: 21 passed and 2 failed at import time because fastapi is not installed without the proxy/dev extra. Those failures are outside the ast-grep path touched here.

JerrettDavis (changes requested):
The latest broadened regex can now classify ordinary file contents as an upstream truncation signal. It searches anywhere in the Read payload for the generic phrase showing lines N-M of T total, so a source comment, documentation string, test fixture, log message, or user-facing text containing that phrase changes Headroom's header to the factual claim input was truncated upstream. It also accepts internally impossible values (for example end >= total) without validation. The new wording-variant test deliberately demonstrates this unanchored behavior rather than a recognizable client banner.

Please scope detection to an actual banner envelope/signature (for example the known bracketed Truncated: PARTIAL view form, with explicitly supported client variants) rather than generic prose within arbitrary source, and validate the captured range (start >= 1, end >= start, end < total, with reasonable consistency against the visible payload where available). Add negative tests where code/comments contain the generic phrase and where counts are malformed or non-truncating. This keeps the useful header qualification without making Headroom assert truncation based on content it is supposed to outline.

slxng1758:
@JerrettDavis Thanks for the comment, I have implemented the changes

JerrettDavis (changes requested):
The recognized banner envelope and malformed-range checks resolve the false-positive blocker, but the payload-consistency rule is incorrect for offset reads. _is_plausible_truncation_range currently requires end_line <= source_line_count. For a legitimate visible window such as lines 100-148 of 300, the payload contains roughly 49 content lines (plus the banner), so 148 <= ~50 is false and Headroom silently misses the truncation. The current tests only cover windows starting at line 1, which hides this. Please validate the inclusive window length (end - start + 1) against the visible payload shape/count, allowing for the banner and any known line-number formatting, rather than comparing the absolute end index to payload length. Add a positive offset-window case and negative cases where the claimed window length exceeds what is visible. The signature scoping and end < total validation otherwise look good.

slxng1758:
Thanks @JerrettDavis I have made the fix

JerrettDavis (approved):
The latest change resolves the remaining offset-window blocker: plausibility is now based on the inclusive window length rather than the absolute ending line, while the known banner envelope and malformed/non-truncating range guards remain intact. The new positive offset-window and oversized-window negative cases cover the regression. Verified from the exact PR head: 16 ast-grep interceptor tests passed and Ruff passed on both touched files.

PR #2435, opened 2026-07-20
fix(proxy/output-shaping): tolerate a non-string system block text in steering

abhay-codes07:
## Description

`apply_verbosity_steering` (the Anthropic output-shaping path) scans the `system` block list to find and update an existing steering block:

```python
if isinstance(system, list):
    for block in system:
        if isinstance(block, dict) and block.get("text", "").startswith(_STEERING_SENTINEL):
```

`.get("text", "")` only substitutes the default when the key is **absent**. A malformed client block with a null text (`{"type": "text", "text": null}`) returns `None`, so `None.startswith(...)` raises `AttributeError`. In the output-shaping treatment arm that call runs inside `shape_request`, which is not individually guarded, so the exception propagates and 502s the request.

The OpenAI chat sibling in the same module already defends against this exact case (`isinstance(part.get("text"), str)`), so the Anthropic path is the inconsistent one.

## Fix

Guard that the block text is a string before `startswith`, mirroring the OpenAI sibling. Well-formed bodies are unchanged: the steering block is still replaced idempotently when a level changes, or appended when absent. The malformed block is left untouched.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/output_steering.py`: string-guard the system block text before `startswith` in `apply_verbosity_steering`.
- `tests/test_output_steering.py`: regression asserting a `system` list containing a `{"text": null}` block does not crash and still appends the steering block.

## Testing

### Test Output

```text
$ python -m pytest tests/test_output_steering.py -q
9 passed

# with the fix reverted, the new test fails (AttributeError on None.startswith):
$ git stash push -- headroom/proxy/output_steering.py
$ python -m pytest "tests/test_output_steering.py::test_anthropic_steering_tolerates_non_string_system_block_text" -q
1 failed

$ uvx ruff@0.15.17 check headroom/proxy/output_steering.py tests/test_output_steering.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/output_steering.py
```

## Real Behavior Proof

- Exact command / steps: called the real `apply_verbosity_steering` with `system=[{"type":"text","text":None},{"type":"text","text":"Real system prompt."}]`; also confirmed the OpenAI sibling `apply_openai_chat_verbosity_steering` handles the same shape.
- Observed result: pre-fix the Anthropic call raised `AttributeError: 'NoneType' object has no attribute 'startswith'` while the OpenAI sibling returned True; post-fix the Anthropic call returns True, leaves the malformed block as-is, appends the steering block, and stays idempotent on a repeat. Ran against the actual module.
- Not tested: a live client that sends a null system block text end to end.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/output_steering.py
+++ b/headroom/proxy/output_steering.py
@@ -36,8 +36,13 @@ def apply_verbosity_steering(body: dict[str, Any], level: int) -> bool:
         return True
     if isinstance(system, list):
         for block in system:
-            if isinstance(block, dict) and block.get("text", "").startswith(_STEERING_SENTINEL):
-                if block["text"] == text:
+            # Guard the text is a string before ``startswith``: a malformed
+            # client block (``{"type": "text", "text": null}``) would otherwise
+            # raise ``AttributeError`` here and 500 the request. The OpenAI chat
+            # sibling below already guards this exact case.
+            block_text = block.get("text") if isinstance(block, dict) else None
+            if isinstance(block_text, str) and block_text.startswith(_STEERING_SENTINEL):
+                if block_text == text:
                     return False
                 block["text"] = text
                 return True
--- a/tests/test_output_steering.py
+++ b/tests/test_output_steering.py
@@ -35,6 +35,23 @@ def test_anthropic_steering_preserves_cached_prefix_block() -> None:
     assert body["system"][1] == {"type": "text", "text": steering_text(2)}
 
 
+def test_anthropic_steering_tolerates_non_string_system_block_text() -> None:
+    # A malformed client block ({"type": "text", "text": null}) must not crash
+    # `.startswith` and 500 the request; steering is still appended. The OpenAI
+    # chat sibling already guards this exact case.
+    body = {
+        "system": [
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed the rebased head d51b5450. The Anthropic steering path now mirrors the OpenAI sibling by checking that block text is a string before calling startswith, so malformed null text blocks are left untouched and steering is still appended/replaced for valid blocks. The regression test covers the null-text system block shape.

Verification:
- uv run --extra dev --with pytest python -m pytest tests/test_output_steering.py -q
- uvx ruff@0.15.17 check headroom/proxy/output_steering.py tests/test_output_steering.py
- uvx ruff@0.15.17 format --check headroom/proxy/output_steering.py tests/test_output_steering.py
- git diff --check upstream/main...HEAD

Upstream CI fast checks are passing after the rebase (commitlint, lint, changes, merge-conflicts, template, no-manual-changelog).

PR #641, opened 2026-06-05
fix: support Copilot Business subscription auth

jbelanger:
## Description

Adds a first-party `headroom copilot-auth login` flow for Copilot subscription
mode and uses the resulting Copilot OAuth token to perform GitHub's Copilot
token exchange before launching the wrapped Copilot CLI.

This fixes Business/Enterprise Cloud accounts where a generic GitHub/Copilot
token can read Copilot account metadata but is rejected by the Copilot token
exchange endpoint. It also avoids treating GitHub.com Enterprise Cloud account
URLs such as `github.com/enterprises/acme` as API hostnames.

Fixes #635
Related: #488, #610
Builds on #576

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Adds `headroom copilot-auth login` and `headroom copilot-auth status`.
- Stores a Headroom-specific Copilot OAuth token under Headroom's state dir.
- Exchanges reusable Copilot OAuth tokens with Copilot Chat-compatible headers before subscription-mode launch.
- Carries the resolved Copilot API endpoint into `headroom wrap copilot --subscription`.
- Handles GitHub.com Enterprise Cloud URLs without synthesizing invalid `api.github.com/enterprises/...` hosts.
- Adds focused unit tests and README guidance for subscription login.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```console
ruff check headroom/copilot_auth.py headroom/cli/copilot_auth.py headroom/cli/main.py headroom/cli/__init__.py headroom/cli/wrap.py tests/test_copilot_auth.py tests/test_cli/test_copilot_auth.py tests/test_cli/test_wrap_copilot.py tests/test_copilot_subscription_smoke.py
# All checks passed!

ruff format --check headroom/copilot_auth.py headroom/cli/copilot_auth.py headroom/cli/main.py headroom/cli/__init__.py headroom/cli/wrap.py tests/test_copilot_auth.py tests/test_cli/test_copilot_auth.py tests/test_cli/test_wrap_copilot.py tests/test_copilot_subscription_smoke.py
[...]

Issue #610, opened 2026-06-04
[BUG]


vipinthomasvertexinc:
## Description

`headroom wrap copilot` regressed in `headroom-ai==0.23.0` for GitHub Copilot OpenAI-compatible usage.

On my machine, the same GitHub Copilot CLI flow works with `headroom-ai==0.22.4` but fails on `0.23.0` when using the OpenAI-compatible path with `gpt-5.4` and the `responses` wire API.

This appears specific to the Headroom/Copilot transport path rather than plain GitHub Copilot itself:

- plain `copilot --model gpt-5.4` works
- `headroom wrap copilot ... --model gpt-5.4` fails on `0.23.0`
- the same wrapped flow works again after downgrading to `0.22.4`

## To Reproduce

1. Install Headroom 0.23.0 with Python 3.13:

    ```bash
    python3.13 -m pip install --user --upgrade "headroom-ai==0.23.0"
    ```

2. Verify plain Copilot works:

    ```bash
    copilot --model gpt-5.4 -s -p "Reply with exactly: OK"
    ```

3. Start Headroom on a local port. I used `8788` because `8787` was already occupied locally:

    ```bash
    KMP_DUPLICATE_LIB_OK=TRUE headroom proxy --port 8788 --openai-api-url https://api.githubcopilot.com --no-telemetry
    ```

4. In another shell, run:

    ```bash
    headroom wrap copilot \
      --port 8788 \
      --no-proxy \
      --provider-type openai \
      --wire-api responses \
      -- \
      --model gpt-5.4 \
      -s -p "Reply with exactly: HEADROOM_OK"
    ```

5. Observe the failure.

6. Downgrade and rerun the same command:

    ```bash
    python3.13 -m pip install --user --upgrade "headroom-ai==0.22.4"
    ```

    Then rerun the same `headroom wrap copilot ...` command above.

## Expected Behavior

The Headroom-wrapped Copilot request should succeed and return:

```text
HEADROOM_OK
```

## Actual Behavior

On `0.23.0`, the wrapped request fails with an unsupported-model error even though:

- the same model works with plain Copilot
- the same wrapped request works with `0.22.4`

## Code Sample

```text
# Plain Copilot works:
copilot --model gpt-5.4 -s -p "Reply with exactly: OK"

# Headroom-wrapped Copilot fails on 0.23.0:
headroom wrap copilot \
  --port 8788 \
  --no-proxy \
  --provider-type openai \
  --wire-api responses \
  -- \
[...]

PR #2445, opened 2026-07-20
feat(opencode): support Copilot subscription backend for headroom models (#2441)

### LINKED ISSUE #2441: [FEATURE] headroom wrap opencode: support GitHub Copilot subscription as backend for headroom/* models
## Problem Statement
I'm always frustrated when I have to choose between compression and my Copilot seat. `headroom wrap opencode` only exposes `headroom/*` models that route to Anthropic/OpenAI directly via BYOK. OpenCode's own `github-copilot/*` provider (subscription seat, no separate key) works fine but bypasses Headroom entirely - no compression, no stats.

## Proposed Solution
Let the Copilot-subscription token exchange that already exists for `headroom wrap copilot --subscription` be used as a backend for the generic proxy too, so `headroom wrap opencode` can route its `headroom/*` models to a Copilot seat instead of requiring a separate paid API key.

## Use Case
- Terminal coding agent workflow: OpenCode with multi-agent orchestration (scout/implement/test/review subagents).
- I pay for a GitHub Copilot seat and don't want a second Anthropic/OpenAI key just to get compression.
- Would let me compress Copilot-routed traffic across all my subagents instead of running that traffic unwrapped.

## Alternatives Considered
- Using `headroom/*` models with a separate Anthropic/OpenAI key — works but means paying for two things.
- Running `github-copilot/*` uncompressed — works but gets none of Headroom's savings.

## Example API (Optional)
```bash
headroom copilot-auth login
headroom wrap opencode --copilot-subscription
# or
headroom proxy --backend copilot-subscription
```

## Additional Context
- Willing to test against a real setup once there's a branch.
- Related: #74 (opencode wrap), #488 (Copilot subscription mode without BYOK) - this is a request to compose the two.

rodboev:
## Description

`headroom wrap opencode` currently routes `headroom/*` models only to ordinary Anthropic or OpenAI backends. A user with a GitHub Copilot subscription cannot point those `headroom/*` requests at the Copilot seat while keeping Headroom compression and stats, even though Headroom already has the validated subscription resolver, the proxy seed path, and the OpenCode provider route needed to do it.

This PR adds `--copilot-subscription` to the OpenCode wrap command. It reuses the existing Copilot subscription token resolver, passes the validated endpoint and token seed into the existing proxy startup path, rejects unsupported runtime modes, and treats any non-empty Copilot API token as a private session seed so token-only sessions do not reuse a shared proxy. The generated OpenCode provider still targets the local proxy, and subscription secrets stay out of OpenCode config, environment, and terminal output.

Closes #2441

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `headroom wrap opencode --copilot-subscription` in `headroom/cli/wrap.py`.
- Reuse the existing validated Copilot subscription resolver through one small required-resolution helper shared with the dedicated Copilot wrapper.
- Pass the resolved endpoint and token seed into `_ensure_proxy()` as `openai_api_url`, `copilot_api_token`, `copilot_refresh_oauth_token`, and `copilot_api_token_expires_at`.
- Reject `--copilot-subscription` with `--no-proxy`, `--prepare-only`, and translated backends before proxy or OpenCode launch.
- Validate subscription mode before snapshotting OpenCode config, so rejected invocations don't create stale backups.
- Scrub inherited Copilot proxy seed variables from the OpenCode child environment.
- Treat any non-empty Copilot API token as a private session seed so token-only sessions do not reuse shared or persistent proxies.
- Add focused OpenCode and persistent-proxy coverage for seed handoff, guard failures, direct-token isolation, secret non-disclosure, and unchanged non-subscription behavior.
- Leave `CHANGELOG.md` untouched because Headroom generates changelog entries from conventional commits.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_cli/test_wrap_opencode.py tests/test_cli/test_wrap_persistent.py tests/test_cli/test_wrap_copilot.py -q`, `106 passed in 136.65s`)
- [x] Linting passes (`uv run ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_opencode.py tests/test_cli/test_wrap_persistent.py`)

### Test Output

```text
Targeted subscription tests pass: OpenCode `7 passed, 37 deselected in 0.36s`, persistent proxy `2 passed, 29 deselected in 0.34s`, and dedicated Copilot `11 passed, 20 deselected in 0.39s`. Coverage includes inherited resolver-input env scrubbing, HEADROOM_BACKEND rejection, no-backup-on-rejection, OpenCode-only scrub scoping, and private-proxy teardown on config-injection failure.
The full focused command `uv run pytest tests/test_cli/test_wrap_opencode.py tests/test_cli/test_wrap_persistent.py tests/test_cli/test_wrap_copilot.py -q` passed with `106 passed in 136.65s`. Ruff check and format check pass.
```

## Real Behavior Proof

- Environment: Windows, `uv` development environment, local CLI tests, no live Copilot seat on this host
- Exact command / steps: run the focused OpenCode and persistent-proxy tests with a mocked `CopilotSubscriptionTokenResolution`, then capture the proof rows for seed handoff, direct-token isolation, guards, and secret non-disclosure
- Observed result: Targeted subscription and proxy-seed tests pass, including OpenCode-only resolver-input env scrubbing and private-proxy teardown on config-injection failure; the full focused command passed with `106 passed in 136.65s`; Ruff check and format check pass.
- Not tested: live Copilot subscription seat run on this host

## Review Readiness

## Checklist

## Additional Notes

- Feature approval comes from the open `enhancement` label on https://github.com/headroomlabs-ai/headroom/issues/2441.
- Keep the final live-backend claim behind manual owner proof. Local CLI tests can prove config, guard, secret, and proxy-seed behavior, but they cannot prove a real Copilot seat on this host.
- `CHANGELOG.md` remains untouched because Headroom's release pipeline generates changelog entries from conventional commits.

### DIFF
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -58,6 +58,11 @@ from headroom.agent_savings import (
     apply_agent_savings_env_defaults,
 )
 from headroom.copilot_auth import (
+    _API_TOKEN_ENV_VARS,
+    _API_TOKEN_EXPIRES_AT_ENV_VAR,
+    _COPILOT_OAUTH_TOKEN_ENV_VARS,
+    _GENERIC_GITHUB_TOKEN_ENV_VARS,
+    _REFRESH_OAUTH_TOKEN_ENV_VAR,
     has_oauth_auth,
     resolve_client_bearer_token,
     resolve_copilot_api_url,
@@ -166,9 +171,11 @@ from headroom.proxy.project_context import with_project_prefix as _with_project_
 from .main import main
 
 _COPILOT_PROXY_SEED_ENV_VARS = (
-    "GITHUB_COPILOT_API_TOKEN",
-    "GITHUB_COPILOT_REFRESH_OAUTH_TOKEN",
-    "GITHUB_COPILOT_API_TOKEN_EXPIRES_AT",
+    *_API_TOKEN_ENV_VARS,
+    _REFRESH_OAUTH_TOKEN_ENV_VAR,
+    _API_TOKEN_EXPIRES_AT_ENV_VAR,
+    *_COPILOT_OAUTH_TOKEN_ENV_VARS,
+    *_GENERIC_GITHUB_TOKEN_ENV_VARS,
 )
 
 
@@ -7818,34 +7825,34 @@ def opencode(
         ),
     )
 
-    # If the proxy fell back to a different port, move our marker so
-    # cleanup tracking stays accurate and update MCP config.
-    if actual_port != port:
-        _unregister_proxy_client(port)
-        _register_proxy_client(actual_port)
-        if not no_mcp:
-            from headroom.mcp_registry import OpencodeRegistrar
+    try:
+        # If the proxy fell back to a different port, move our marker so
+        # cleanup tracking stays accurate and update MCP config.
+        if actual_port != port:
+            _unregister_proxy_client(port)
+            _register_proxy_client(actual_port)
+            if not no_mcp:
+                from headroom.mcp_registry import OpencodeRegistrar
 
-            _setup_headroom_mcp(OpencodeRegistrar(), actual_port, verbose=verbose, force=True)
+                _setup_headroom_mcp(OpencodeRegistrar(), actual_port, verbose=verbose, force=True)
 
-    launch_environ = os.environ.copy()
-    if subscription_resolution is not None:
-        _scrub_copilot_proxy_seed_env(launch_environ)
-    env, env_vars_display = _build_opencode_launch_env(
-        actual_port, launch_environ, project=_project_name_from_cwd(), include_mcp=not no_mcp
-    )
-
-    # Inject Headroom provider into OpenCode config so traffic routes through proxy.
-    inject_opencode_provider_config(actual_port)
-    if memory:
-        mem_dir = Path.cwd() / ".headroom"
-        _inject_memory_mcp_config(
-            os.environ.get("USER", os.environ.get("USERNAME", "default")),
+        launch_environ = os.environ.copy()
+        if subscription_resolution is not None:
+            _scrub_copilot_proxy_seed_env(launch_environ)
+        env, env_vars_display = _build_opencode_launch_env(
+            actual_port, launch_environ, project=_project_name_from_cwd(), include_mcp=not no_mcp
         )
 
-    # Proxy already started by _ensure_proxy above; tell _launch_tool to
-    # skip duplicate startup.
-    try:
+        # Inject Headroom provider into OpenCode config so traffic routes through proxy.
+        inject_opencode_provider_config(actual_port)
+        if memory:
+            mem_dir = Path.cwd() / ".headroom"
+            _inject_memory_mcp_config(
+                os.environ.get("USER", os.environ.get("USERNAME", "default")),
+            )
+
+        # Proxy already started by _ensure_proxy above; tell _launch_tool to
+        # skip duplicate startup.
         _launch_tool(
             binary=opencode_bin,
             args=opencode_args,
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -63,6 +63,7 @@ from headroom.copilot_auth import (
     _COPILOT_OAUTH_TOKEN_ENV_VARS,
     _GENERIC_GITHUB_TOKEN_ENV_VARS,
     _REFRESH_OAUTH_TOKEN_ENV_VAR,
+    CopilotSubscriptionTokenResolution,
     has_oauth_auth,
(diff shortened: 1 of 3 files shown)


lucasp1337:
Heya, many thanks for the fix! If any testing is needed, I volunteer!

rodboev:
Thanks, I appreciate it. The main gap on my side is a real smoke test with an actual Copilot subscription, since I could only verify this with local CLI tests and mocked subscription resolution.

If you’re up for it, the most useful check would be `headroom wrap opencode --copilot-subscription` with any `headroom/*` model you already use, then confirm that it starts cleanly and routes requests without leaking token material into config or env. If you try it, even a quick "works for me" or the first failure you hit would help a lot.

lucasp1337:
@rodboev 

After some minor testing from my end (Business/Enterprise Copilot seat) I hit a snag with --copilot-subscription.

Every request through headroom/* models 400s with "the requested model is not supported" - tried both claude-sonnet-4-6 and claude-haiku-4.5, same result on both, so it's not model-specific. Traced it back to the upstream host: subscription mode was routing me to api.enterprise.githubcopilot.com, while opencode's native github-copilot/* provider (configured for the public API) hits the same models on the same account totally fine.

Dug into copilot_auth.py a bit — looks like _subscription_api_url_from_user_info_payload() pulls the account-specific endpoints.api value from the user-info payload and routes through that. 
But resolve_copilot_api_url()'s own docstring a few hundred lines down explicitly says not to do this, and cites the exact same symptom: the segmented host doesn't serve newer models, and this regressed once before (#610). That guard just isn't applied on the subscription path.

Also noticed there's a 6c3698b5 fix(copilot): route subscription + OAuth through the generic host (#610) commit sitting on fix/copilot-oauth-endpoint-regression — title says it covers subscription too, but what actually landed on main (18925b8c) looks like it only covers the non-subscription OAuth case. Wondering if the subscription half just got dropped somewhere along the way — might be worth a look, could be a quick cherry-pick rather than a new fix.

Happy to test against my account again! And thank you for your efforts:)

rodboev:
Thanks for the detailed report, and for tracing it into `copilot_auth.py` yourself; that saved a lot of time.

Confirmed on the routing half. `_subscription_api_url_from_user_info_payload()` only normalizes `api.githubcopilot.com` and `api.individual.githubcopilot.com` back to the generic host; anything else under `*.githubcopilot.com` passes through. Your seat advertises `api.enterprise.githubcopilot.com` in the user-info payload, so subscription mode routes there, and that segmented host rejects the newer Claude models. That's exactly the failure mode the `resolve_copilot_api_url()` docstring warns about; the guard just isn't applied inside the subscription resolver.

The cherry-pick theory doesn't hold up, though. `git diff 18925b8c..6c3698b5` shows the two commits carry identical `copilot_auth.py` content; the branch only differs in docs and a CLI help-text block. The subscription half of #610 did land. What happened afterward is that #641 reintroduced payload-derived routing to support Business auth (deliberately preserving dedicated hosts), and #1836 later normalized only the individual host back. So the enterprise/business passthrough is a later, intentional change that turned out to be wrong for model coverage, and cherry-picking `6c3698b5` would conflict with the resolver rewrites since June without changing behavior.

While verifying I found a second defect: an explicit `GITHUB_COPILOT_API_URL` pin is honored on the token-exchange path but ignored on the explicit-API-token and Copilot-token resolution paths, where the payload host wins over the pin. So the documented escape hatch can't reliably force the generic host back either.

Planned fix, as a fresh PR against the shared resolver (it covers both `wrap copilot --subscription` and `wrap opencode --copilot-subscription`, since both seed the proxy from the same resolution):

1. Normalize `api.enterprise.githubcopilot.com` and `api.business.githubcopilot.com` to the generic host by default, extending what #1836 did for individual seats. Tenants that genuinely need a dedicated host pin it with `GITHUB_COPILOT_API_URL`, per the existing policy.
2. Make the explicit pin win on every resolution path.

Interim workaround: `GITHUB_COPILOT_API_URL=https://api.githubcopilot.com` should unblock you today, with the caveat that it only takes effect when auth resolves through the token exchange (e.g. `headroom copilot-auth login`). A raw Copilot API token in the environment hits the path where the pin is ignored, which is defect 2 above.

I'll gladly take you up on the testing offer once the PR is up; there's no live Business/Enterprise seat on my side. Thanks again!

lucasp1337:
Hey, sorry for the late reply - I haven't had much spare time to test it out.

This morning I've upgraded to v0.33.0 – and so far all good! I see that the requests indeed go through the proxy! 🎉

I'll keep you updated if I find any issues, but I wanted to give you a quick update. Thank you for your contribution!🫡

All the best.

PR #2459, opened 2026-07-21
fix(install/windows): register persistent-task from S4U hidden XML (#2453)

### LINKED ISSUE #2453: [BUG] Windows persistent-task preset flashes a focus-stealing console every 5 minutes (interactive logon token)
## Summary

On Windows, `headroom install apply --preset persistent-task` creates scheduled tasks that run with an **interactive logon token**. As a result, the health task (which repeats every 5 minutes) launches a **visible console window that briefly grabs keyboard focus** before disappearing — every 5 minutes, indefinitely. On an active desktop this is very distracting (focus theft interrupts whatever you're typing).

The startup task has the same principal, so the same flash occurs at boot and on any proxy restart.

## Environment

- Headroom 0.31.0
- Windows 11 Pro 10.0.26200
- Preset: `persistent-task`, runtime `python`, scope `user` (fell back to this because `persistent-service` fails — see #1866)

## Root cause

The generated task XML uses an interactive principal:

```xml
<Principal id="Author">
  <UserId>S-1-5-21-...</UserId>
  <LogonType>InteractiveToken</LogonType>
</Principal>
```

with a health trigger repeating every 5 minutes:

```xml
<TimeTrigger>
  <StartBoundary>...</StartBoundary>
  <Repetition><Interval>PT5M</Interval></Repetition>
</TimeTrigger>
<Actions>
  <Exec><Command>...\ensure-headroom.cmd</Command></Exec>
</Actions>
```

`ensure-headroom.cmd` shells out to `powershell.exe ... ensure-headroom.ps1`, and because the task runs interactively, each invocation spawns a console in the user's session with focus.

## Suggested fixes (either resolves it)

1. **S4U principal** — set `LogonType` to `S4U` ("run whether user is logged on or not", no stored password). Tasks then run in a non-interactive session; the health poller still runs and can still restart the proxy, but no window is ever shown. This is the cleanest option and needs no wrapper scripts.
2. **Hidden-window launcher** — invoke the runner via `wscript.exe` with a `WScript.Shell.Run(cmd, 0, false)` wrapper (window style 0 = hidden), or `conhost.exe --headless`, so no console is drawn.

Option 1 is preferable — it's a one-attribute change in the task registration and avoids extra on-disk helper files.

## Related

Same weak Windows supervisor layer as #1866 (`persistent-service` preset failing with `StartService FAILED 1053` — a plain console process registered as a service binary doesn't implement the SCM protocol, so Windows kills it after the 30s start timeout). A hardened Windows deployment path would ideally fix both: a working service preset *and* a windowless task preset.

Parideboy:
## Description

Windows `persistent-task` created its startup and 5-minute health tasks via `schtasks` command-line flags, which register the task with an **interactive-token** principal. Every task run spawned a visible console window that briefly grabbed keyboard focus before vanishing — every 5 minutes, indefinitely (and at boot / proxy restart). Fixes #2453.

This registers the tasks from Task Scheduler **XML** instead: user-scope tasks use an **S4U** principal (run whether the user is logged on or not, no stored password) with `<Hidden>true</Hidden>`, so runs execute in a non-interactive session and never draw a window. System-scope tasks keep the LocalSystem service account (which already has no desktop).

## Type of Change

## Changes Made

- `headroom/install/supervisors.py`: add `_windows_task_xml()` (S4U/hidden for user scope, LocalSystem for system scope), `_windows_boot_trigger()`, `_windows_health_trigger()` (PT5M repetition), and `_register_windows_task()` (writes UTF-16 XML to a temp file and calls `schtasks /Create /TN <n> /XML <file> /F`). Rewrite the Windows TASK branch of `install_supervisor` to register both tasks from XML.
- `tests/test_install/test_supervisors.py`: unit tests asserting the XML carries `S4U` + `Hidden` + `PT5M` for user scope and `S-1-5-18` / `ServiceAccount` for system scope; updated the install-flow assertion to expect `schtasks /XML` registration for the startup and health tasks.

## Testing

- [x] Unit tests pass

```
$ python -m pytest tests/test_install/test_supervisors.py -q
collected 29 items
tests\test_install\test_supervisors.py .............................     [100%]
============================= 29 passed in 1.48s ==============================
```

## Real Behavior Proof

- Environment: Windows 11 Pro 10.0.26200, Python 3.13.11
- Exact command / steps: python -m pytest tests/test_install/test_supervisors.py -q; ruff check + ruff format --check; mypy headroom/install/supervisors.py --ignore-missing-imports
- Observed result: 29 passed; ruff clean; mypy exit 0. Generated XML contains <LogonType>S4U</LogonType> and <Hidden>true</Hidden> for user scope.
- Not tested: live end-to-end `headroom install apply --preset persistent-task` on a physical desktop confirming zero console flash over a >5-minute window (no interactive Windows session in CI).

## Review Readiness

### DIFF
--- a/headroom/install/supervisors.py
+++ b/headroom/install/supervisors.py
@@ -2,13 +2,17 @@
 
 from __future__ import annotations
 
+import getpass
 import os
 import re
 import shlex
 import subprocess
 import sys
+import tempfile
 import time
+from datetime import datetime
 from pathlib import Path
+from xml.sax.saxutils import escape as _xml_escape
 
 import click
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the current head. The change is well scoped to Windows persistent-task registration: XML registration moves user-scope tasks to an S4U, hidden principal while preserving LocalSystem for system scope, writes UTF-16 XML for schtasks /XML, escapes XML text, and keeps temp-file cleanup best-effort.

Local focused validation on Windows passed for the touched surface:

```text
uv run --frozen --extra dev python -m pytest tests/test_install/test_supervisors.py -q
27 passed
uv run --frozen --extra dev ruff check headroom/install/supervisors.py tests/test_install/test_supervisors.py
uv run --frozen --extra dev ruff format --check headroom/install/supervisors.py tests/test_install/test_supervisors.py
```
The upstream CI matrix is still running, but I do not see a source-level blocker in this PR.

PR #2460, opened 2026-07-21
fix(install/windows): back persistent-service with an always-on task (#1866)

### LINKED ISSUE #1866: [BUG] headroom install apply --preset persistent-service
Windows 11

Headroom version:
0.30.0

Installed via:
headroom.exe in C:\Users\<user>\.local\bin

Command:
headroom install apply --preset persistent-service

Expected:
Creates ~/.headroom/deploy/default and installs the Windows service.

Actual:
ERROR: Invalid start= field

sc.exe create headroom-default
binPath= cmd.exe /c "C:\Users\<user>\.headroom\deploy\default\run-headroom.cmd"
start= auto

headroom install status:
No deployment profile named 'default' is installed.

Contents of ~/.headroom:
deploy/
update_check.json

deploy/ is empty.

Parideboy:
## Description

`headroom install apply --preset persistent-service` failed on Windows. It registered `cmd.exe /c run-headroom.cmd` as a service via `sc.exe create`, but a plain console process never performs the SCM `StartServiceCtrlDispatcher` handshake, so Windows killed it after the 30s start timeout (`StartService FAILED 1053`) and left an empty `deploy/default` profile. Fixes #1866.

Rather than add a service-wrapper dependency, the Windows service is now backed by scheduled tasks: a `BootTrigger` task launches the foreground proxy (`run-headroom`) at boot so it stays always-on, plus a 5-minute health task as a watchdog. Both are registered from the same S4U/hidden task XML introduced for #2453, so no console window ever flashes.

> Stacked on #2459 (the #2453 S4U/hidden XML machinery). Until that merges, this PR's diff includes its commit; it collapses automatically once #2459 lands.

## Type of Change

## Changes Made

- `headroom/install/supervisors.py`: collapse the Windows SERVICE/TASK branches of `install_supervisor` — SERVICE registers a boot task running `run-headroom.cmd` (always-on) plus a 5-minute `ensure-headroom.cmd` health task; TASK re-ensures on boot. Both use `_windows_task_xml` (S4U/hidden). `start_supervisor`/`stop_supervisor` now `schtasks /Run` / `/End` the startup task; `remove_supervisor` deletes both tasks and best-effort deletes any legacy `sc.exe` service.
- `tests/test_install/test_supervisors.py`: assert SERVICE now emits `windows-task` records (no `sc.exe create`), and that Windows start/stop/remove drive the tasks.

## Testing

- [x] Unit tests pass

```
$ python -m pytest tests/test_install/test_supervisors.py -q
collected 29 items
tests\test_install\test_supervisors.py .............................     [100%]
============================= 29 passed in 0.35s ==============================
```

## Real Behavior Proof

- Environment: Windows 11 Pro 10.0.26200, Python 3.13.11
- Exact command / steps: python -m pytest tests/test_install/test_supervisors.py -q; ruff check + ruff format --check; mypy headroom/install/supervisors.py --ignore-missing-imports
- Observed result: 29 passed; ruff clean; mypy exit 0. SERVICE install now returns windows-task records and issues no sc.exe create.
- Not tested: live `headroom install apply --preset persistent-service` on a physical Windows host confirming the proxy answers /readyz after reboot with no 1053 (no interactive Windows host in CI).

## Runtime Rollout Safety

- Rollout-managed feature(s): none. The install/supervisor path is not in the `headroom.rollout` feature registry
- Stable/default behavior changed: yes, for Windows `--preset persistent-service` installs only. It registers a boot-triggered scheduled task plus a 5-minute health task instead of an `sc.exe` service; other presets and other platforms are unchanged
- Kill switch / disable path: none needed as a flag. `headroom install remove` deletes both tasks and best-effort removes any legacy `sc.exe` service, returning the host to no supervisor; this change replaces a path that never worked (`StartService FAILED 1053`), not a working default
- Qualification impact: none. Install/supervisor logic only, no routing/caching/backend-selection change
- Rollback path: revert this commit. `remove_supervisor` cleans up both new tasks before rollback, so no orphaned scheduled tasks remain

## Review Readiness

### DIFF
--- a/headroom/install/supervisors.py
+++ b/headroom/install/supervisors.py
@@ -502,8 +502,13 @@ def start_supervisor(manifest: DeploymentManifest) -> None:
         _bootstrap_with_retry(domain, plist_path, action="start")
         return
     if _is_windows():
-        # Both SERVICE and TASK are task-backed on Windows (#1866); run the
-        # startup task to (re)launch the proxy now.
+        # Both SERVICE and TASK are task-backed on Windows (#1866). Re-enable the
+        # health watchdog (stop disables it, see stop_supervisor) before running
+        # the startup task to (re)launch the proxy now.
+        subprocess.run(
+            ["schtasks", "/Change", "/TN", f"{manifest.service_name}-health", "/ENABLE"],
+            check=True,
+        )
         subprocess.run(["schtasks", "/Run", "/TN", f"{manifest.service_name}-startup"], check=True)
 
 
@@ -542,7 +547,14 @@ def stop_supervisor(manifest: DeploymentManifest) -> None:
         return
     if _is_windows():
         # Task-backed on Windows (#1866); ending the startup task stops the proxy.
+        # Also disable the 5-minute health watchdog — otherwise `ensure` would
+        # restart the deployment on the next interval, making `stop` non-durable.
+        # start/restart re-enable it (see start_supervisor).
         subprocess.run(["schtasks", "/End", "/TN", f"{manifest.service_name}-startup"], check=True)
+        subprocess.run(
+            ["schtasks", "/Change", "/TN", f"{manifest.service_name}-health", "/DISABLE"],
+            check=True,
+        )
 
 
 def remove_supervisor(manifest: DeploymentManifest) -> None:
--- a/headroom/install/supervisors.py
+++ b/headroom/install/supervisors.py
@@ -546,15 +546,19 @@ def stop_supervisor(manifest: DeploymentManifest) -> None:
             )
         return
     if _is_windows():
-        # Task-backed on Windows (#1866); ending the startup task stops the proxy.
-        # Also disable the 5-minute health watchdog — otherwise `ensure` would
-        # restart the deployment on the next interval, making `stop` non-durable.
-        # start/restart re-enable it (see start_supervisor).
+        # Task-backed on Windows (#1866). Order matters so `stop` is durable:
+        #   1. DISABLE the health watchdog — no future 5-minute triggers.
+        #   2. End any *already running* health task — /Change only blocks future
+        #      triggers, so an in-flight `ensure` could otherwise re-enable health
+        #      and restart the proxy after we stop it (best effort: usually not
+        #      running).
+        #   3. End the startup task LAST, so a proxy a racing `ensure` just
+        #      started is still torn down.
+        # start/restart re-enable health (see start_supervisor).
+        health = f"{manifest.service_name}-health"
+        subprocess.run(["schtasks", "/Change", "/TN", health, "/DISABLE"], check=True)
+        subprocess.run(["schtasks", "/End", "/TN", health], check=False)
         subprocess.run(["schtasks", "/End", "/TN", f"{manifest.service_name}-startup"], check=True)
-        subprocess.run(
-            ["schtasks", "/Change", "/TN", f"{manifest.service_name}-health", "/DISABLE"],
-            check=True,
-        )
 
 
 def remove_supervisor(manifest: DeploymentManifest) -> None:
--- a/headroom/install/supervisors.py
+++ b/headroom/install/supervisors.py
@@ -418,31 +418,24 @@ def install_supervisor(manifest: DeploymentManifest) -> list[ArtifactRecord]:
         records.append(ArtifactRecord(kind="plist", path=str(plist_path)))
         return records
 
-    if _is_windows() and manifest.supervisor_kind == SupervisorKind.SERVICE.value:
-        # sc.exe's binPath= value embeds its own quotes (cmd.exe /c "<path>").
-        # Passing this as an argv list lets subprocess.list2cmdline re-quote the
-        # token and sc.exe mis-tokenizes it (issue #1654), so build the exact
-        # command line ourselves and hand subprocess a string.
-        run_cmd = windows_run_cmd_path(manifest.profile)
-        create_cmd = (
-            f"sc.exe create {manifest.service_name} "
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The Windows service replacement is close, but the current lifecycle semantics make headroom install stop non-durable for the new task-backed persistent-service.

On this branch, install stop still allows SERVICE manifests through _reject_task_lifecycle, then _stop_deployment() calls stop_supervisor() and stop_runtime(). The Windows stop_supervisor() implementation only runs:

```text
schtasks /End /TN <service>-startup
```
It leaves <service>-health enabled. That health task runs ensure-headroom.cmd every 5 minutes, and install_agent_ensure() explicitly starts the deployment whenever /readyz is down / the runtime is stopped. So a user who runs headroom install stop --profile default will see the proxy stopped briefly, then the watchdog task can restart it on the next interval. That is a behavioral regression from the previous SERVICE contract, where stop was expected to leave the persistent deployment stopped until start/restart/reapply.

Please make stop/restart semantics explicit for the task-backed service before this lands. A likely fix is to disable or delete the health watchdog during SERVICE stop and re-enable/register it during start/restart, or add a persisted paused/stopped marker that install agent ensure respects. The important invariant is that headroom install stop must not be undone by the health task minutes later.

The focused PR tests pass locally in an isolated worktree (tests/test_install/test_supervisors.py, Ruff check, Ruff format), so this is a source-behavior blocker rather than a formatting/test-run issue.

Parideboy:
Thanks @JerrettDavis good catch, fixed in 474d15a8.

`stop`/`start` now toggle the health watchdog alongside the startup task:
- `stop_supervisor` (Windows): `schtasks /End <svc>-startup` **then** `schtasks /Change /TN <svc>-health /DISABLE`
- `start_supervisor` (Windows): `schtasks /Change /TN <svc>-health /ENABLE` **then** `schtasks /Run <svc>-startup`

So `install stop` leaves the deployment down until an explicit `start`/`restart` (restart = stop+start, which re-enables the watchdog). This holds the old SERVICE contract without a separate paused-marker file. Updated the start/stop unit test to assert the enable/disable calls; focused suite (29 passed), ruff, and mypy are green.

JerrettDavis (changes requested):
Thanks for the quick fix. The steady-state issue from the previous review is addressed, and focused local validation on Windows still passes:
- `uv run --frozen --extra dev python -m pytest tests/test_install/test_supervisors.py -q` -> 29 passed
- `uv run --frozen --extra dev ruff check headroom/install/supervisors.py tests/test_install/test_supervisors.py` -> passed
- `uv run --frozen --extra dev ruff format --check headroom/install/supervisors.py tests/test_install/test_supervisors.py` -> passed

I still need one lifecycle gap closed before approving: a currently running `<service>-health` task can race `headroom install stop`. `schtasks /Change /DISABLE` prevents future triggers, but it does not terminate an already-running scheduled task. If the health task is in `install agent ensure` while `stop` runs, it can acquire/run the start path after `stop_supervisor()` has ended `<service>-startup` and disabled future health triggers. Worse, `start_supervisor()` from that in-flight ensure re-enables `<service>-health` before running `<service>-startup`, so `install stop` can return with the proxy restarted and the watchdog enabled again.

Please make the stop path also neutralize an in-flight health task, for example by disabling health before stopping startup and issuing a best-effort `/End` for `<service>-health` as part of Windows SERVICE stop, with a unit test covering the exact command sequence. The invariant should be that no health invocation already running when `install stop` begins can re-enable or restart the task-backed service after stop completes.

Parideboy:
Fixed in 791ba03b. The Windows SERVICE stop path is now ordered to neutralize an in-flight health task, not just future triggers:

```
schtasks /Change /TN <svc>-health /DISABLE   # no future triggers (check=True)
schtasks /End   /TN <svc>-health             # kill an already-running ensure (best effort)
schtasks /End   /TN <svc>-startup            # stop the proxy LAST (check=True)
```

Ending `<svc>-startup` **last** means a proxy a racing `ensure` just started is still torn down, and the disable + `/End` on `<svc>-health` stops that ensure from re-enabling the watchdog. So once `install stop` returns, no health invocation that was running when it began can restart the service. `start_supervisor` is unchanged (ENABLE health → Run startup).

Updated the start/stop unit test to assert the exact 5-command sequence; focused suite (29 passed), ruff, and mypy green.

JerrettDavis (approved):
Thanks for iterating on this. The latest head addresses both lifecycle blockers from my previous reviews.

The stop path now disables future health triggers, ends any already-running health task best-effort, and only then ends the startup task. That preserves the key invariant for the task-backed Windows SERVICE mode: once `headroom install stop` returns, a health invocation that was already running when stop began should not be able to re-enable the watchdog or leave the proxy restarted.

I also checked the surrounding CLI lifecycle guard: `persistent-task` still rejects normal `install start/stop/restart`, so this helper behavior is still scoped to the Windows task-backed SERVICE path for user lifecycle commands.

Focused local validation on Windows:

```text
uv run --frozen --extra dev python -m pytest tests/test_install/test_supervisors.py -q
29 passed
uv run --frozen --extra dev ruff check headroom/install/supervisors.py tests/test_install/test_supervisors.py
All checks passed!
uv run --frozen --extra dev ruff format --check headroom/install/supervisors.py tests/test_install/test_supervisors.py
2 files already formatted
git diff --check upstream/main...HEAD
# no output
```

JerrettDavis (approved):
Re-reviewed current head 3a8831dd. The task-backed Windows service lifecycle now closes both stop races: future watchdog triggers are disabled, an in-flight health task is ended best-effort, and the startup task is ended last so a racing ensure cannot leave the proxy running. Focused Windows validation: 29 supervisor tests passed and touched-file Ruff passed.

PR #2467, opened 2026-07-21
fix(ccr): tolerate null/malformed OpenAI data in response handling

abhay-codes07:
## Description

Two sibling spots in the CCR OpenAI response handling assumed well-formed provider data and crash on the present-but-null shapes some OpenAI-compatible gateways send.

**1. Streaming reconstruction (`_reconstruct_openai_response`).** Tool-call deltas were accumulated on key presence only:

```python
if "tool_calls" in delta:
    for tc_delta in delta["tool_calls"]:
        ...
        if "function" in tc_delta:
            fn = tc_delta["function"]
            if "name" in fn:
                ...
```

A delta with `"tool_calls": null` (or `"function": null`) has the key present with a null value, so `for tc_delta in None` raises `TypeError: 'NoneType' object is not iterable`, aborting the whole CCR round. The sibling line just above already value-guards content (`if "content" in delta and delta["content"]:`).

**2. Responses assistant extraction (`_extract_assistant_message`).** The `openai_responses` branch returned `response.get("output", [])`, which only falls back when the key is absent. A present-but-null `output` returned None, and `handle_response` then did `current_messages.extend(None)`, the same `TypeError`. The `choices` branch right above already guards this with `isinstance`.

## Fix

Guard the values, not just the keys:

- Iterate `tool_calls` only when it is a list, skip a non-dict entry, and read `function` only when it is a dict.
- Coerce `output` to a list when it is not one.

Well-formed streams and responses reconstruct exactly as before.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/ccr/response_handler.py`: value-guard `tool_calls`/`function` (and skip non-dict tool-call entries) in `_reconstruct_openai_response`; coerce a null/absent `output` to a list in `_extract_assistant_message`.
- `tests/test_ccr_response_handler_extra.py`: regressions for null `tool_calls`/`function` in the stream reconstruction and for a null `output` in the Responses assistant extraction.

## Testing

### Test Output

```text
$ python -m pytest "tests/test_ccr_response_handler_extra.py::test_reconstruct_openai_response_tolerates_null_tool_calls_and_function" "tests/test_ccr_response_handler_extra.py::test_extract_assistant_message_responses_output_null_coerces_to_list" -q
2 passed

# with the reconstruction fix reverted, the first test fails with

$ uvx ruff@0.15.17 check headroom/ccr/response_handler.py tests/test_ccr_response_handler_extra.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/ccr/response_handler.py
```

Note: a handful of pre-existing async tests in this file fail in my local venv because `pytest-asyncio` is not configured there (`Unknown config option: asyncio_mode`); they fail identically on a clean `main` without my change. The tests I added are synchronous.

## Real Behavior Proof

- Exact command / steps: called the real `StreamingCCRHandler._reconstruct_openai_response` with deltas carrying `"tool_calls": null` and `"function": null`, and the real `CCRResponseHandler._extract_assistant_message` with `{"output": None}`; reverted the reconstruction fix and re-ran.
- Observed result: with the fixes the reconstruction returns the concatenated content and the accumulated tool call, and the extraction returns `{"_openai_responses_output_items": []}`; with the reconstruction fix reverted the same input raises `TypeError: 'NoneType' object is not iterable`. Ran against the actual module.
- Not tested: a live end-to-end CCR round against a provider that emits these null frames.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/ccr/response_handler.py
+++ b/headroom/ccr/response_handler.py
@@ -895,8 +895,17 @@ class StreamingCCRHandler:
             if "content" in delta and delta["content"]:
                 message["content"] = (message.get("content") or "") + delta["content"]
 
-            if "tool_calls" in delta:
-                for tc_delta in delta["tool_calls"]:
+            # Guard the value, not just the key: some OpenAI-compatible
+            # providers include ``"tool_calls": null`` (and ``"function": null``)
+            # in a delta rather than omitting the key, which would make the
+            # iteration below raise ``TypeError: 'NoneType' object is not
+            # iterable`` and abort the whole reconstruction. Mirrors the
+            # ``and delta["content"]`` value-guard above.
+            tool_calls = delta.get("tool_calls")
+            if isinstance(tool_calls, list):
+                for tc_delta in tool_calls:
+                    if not isinstance(tc_delta, dict):
+                        continue
                     idx = tc_delta.get("index", 0)
                     if idx not in tool_calls_map:
                         tool_calls_map[idx] = {
@@ -908,8 +917,8 @@ class StreamingCCRHandler:
                     tc = tool_calls_map[idx]
                     if "id" in tc_delta:
                         tc["id"] = tc_delta["id"]
-                    if "function" in tc_delta:
-                        fn = tc_delta["function"]
+                    fn = tc_delta.get("function")
+                    if isinstance(fn, dict):
                         if "name" in fn:
                             tc["function"]["name"] = fn["name"]
                         if "arguments" in fn:
--- a/headroom/ccr/response_handler.py
+++ b/headroom/ccr/response_handler.py
@@ -399,7 +399,16 @@ class CCRResponseHandler:
             # echoed back verbatim as `input[]` items — not a single
             # role/content dict like chat completions. Sentinel key mirrors
             # `_openai_tool_results`; handle_response() extends on it.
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
This looks good to me. The change is narrowly scoped to value-guarding streamed `tool_calls`/`function` deltas during OpenAI CCR reconstruction, keeps well-formed tool-call accumulation behavior intact, and adds a direct regression for the null-frame case while still reconstructing a later real tool call.

I also validated the focused path locally:

```text
uv run --frozen --extra dev python -m pytest tests/test_ccr_response_handler_extra.py::test_reconstruct_openai_response_tolerates_null_tool_calls_and_function tests/test_ccr_response_handler_extra.py::test_streaming_buffer_and_parse_sse_helpers -q
2 passed
```

JerrettDavis (approved):
Re-reviewed the new head after the follow-up `openai_responses` null-output commit.

The added coercion keeps the sentinel output list contract intact for `handle_response`, and the regression test covers absent, null, and real `output` values. The earlier stream-delta guard also preserves well-formed tool-call reconstruction while ignoring malformed/null deltas. Latest checks are green for the refreshed head.

PR #2468, opened 2026-07-21
fix(backends): forward the standard service_tier param (litellm + anyllm)

chopratejas:
## Description

`service_tier` is a standard provider request field (OpenAI `flex`/`default`/`priority`, Anthropic `auto`/`standard_only`), but the SDK-translation backends rebuilt the upstream request from a fixed field whitelist and **silently dropped** it — so a client that set `service_tier` had it discarded on `--backend litellm`/`anyllm`. The direct passthrough already forwards it; this makes the translation backends do the same, alongside `temperature`/`top_p`/`tools`.

## Type of Change

## Changes Made

- `backends/litellm.py` (Anthropic path): forward `service_tier` when present (the OpenAI path already forwards unconsumed fields via `_build_openai_extra_body`).
- `backends/anyllm.py`: add `service_tier` to the forwarded params (both the Anthropic whitelist and the OpenAI param list).

## Testing

- [x] New test added

### Test Output

```text
$ python -m pytest tests/test_backend_anyllm.py -q
16 passed in 2.51s

$ ruff check headroom/backends/litellm.py headroom/backends/anyllm.py
All checks passed!

$ mypy headroom/backends/anyllm.py headroom/backends/litellm.py   # 0 new errors
```

## Real Behavior Proof

- Environment: local, FakeAnyLLM instance capturing `acompletion` kwargs.
- Steps: `send_message({..., "service_tier": "standard_only"})`.
- Observed: `instance.calls[0]["service_tier"] == "standard_only"` — the param reaches the SDK call (previously absent).
- Not tested: live provider round-trip (no keys in CI); relies on the SDK forwarding standard params.

## Checklist

- [x] Self-reviewed
- [x] Tests added and passing

### DIFF
--- a/headroom/backends/anyllm.py
+++ b/headroom/backends/anyllm.py
@@ -230,6 +230,8 @@ class AnyLLMBackend(Backend):
                 kwargs["tools"] = body["tools"]
             if "tool_choice" in body:
                 kwargs["tool_choice"] = body["tool_choice"]
+            if "service_tier" in body:
+                kwargs["service_tier"] = body["service_tier"]
 
             logger.debug(f"any-llm request: provider={self.provider}, model={original_model}")
 
@@ -280,6 +282,8 @@ class AnyLLMBackend(Backend):
                 kwargs["tools"] = body["tools"]
             if "tool_choice" in body:
                 kwargs["tool_choice"] = body["tool_choice"]
+            if "service_tier" in body:
+                kwargs["service_tier"] = body["service_tier"]
 
             msg_id = f"msg_{uuid.uuid4().hex[:24]}"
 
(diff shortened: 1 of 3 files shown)


JerrettDavis (changes requested):
This needs one more change before approval.

The LiteLLM Anthropic translation path now forwards `service_tier` in `send_message`, but the matching `stream_message` path still omits it. That leaves streamed Anthropic-format requests through `--backend litellm` with the same silent-drop behavior this PR is fixing for non-streaming requests.

Please add the same `if "service_tier" in body: kwargs["service_tier"] = body["service_tier"]` handling to `LiteLLMBackend.stream_message`, and add/extend a LiteLLM-focused test so both the send and stream Anthropic translation paths are covered. The any-llm coverage is useful, but it does not protect the LiteLLM path where the remaining omission is.

PR #2470, opened 2026-07-21
fix(memory): don't crash inline memory extraction on a non-object <memory> block

abhay-codes07:
## Description

`parse_response_with_memory` extracts an inline `<memory>...</memory>` block from a model response and parses its JSON:

```python
try:
    data = json.loads(memory_json)
    memories = data.get("memories", [])
except json.JSONDecodeError as e:
    logger.warning(f"Failed to parse memory JSON: {e}")
```

The block content is fully model-controlled. `json.loads` succeeds on any valid JSON, including a non-object such as a bare array (`<memory>["x"]</memory>`), a string, or a number. `data.get("memories", [])` then raises `AttributeError: 'list' object has no attribute 'get'`, which the `except json.JSONDecodeError` does not catch, so the inline memory path crashes on output a model can realistically produce.

## Fix

Guard the parsed value: read `memories` only when the block is a JSON object, and accept it only when it is a list (logging and ignoring otherwise). Malformed JSON is still handled by the existing decode guard, and a well-formed object is unchanged. This mirrors the non-object hardening already applied to the batch JSONL path.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/memory/inline_extractor.py`: only read `memories` from a dict-typed parsed block, and only when the field is a list; log and ignore other shapes.
- `tests/test_memory_wrapper.py`: regression covering a non-object memory block, a non-list `memories` field, malformed JSON, and a well-formed block.

## Testing

### Test Output

```text
$ python -m pytest tests/test_memory_wrapper.py -q
6 passed

# AttributeError: 'list' object has no attribute 'get'

$ uvx ruff@0.15.17 check headroom/memory/inline_extractor.py tests/test_memory_wrapper.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/memory/inline_extractor.py
```

## Real Behavior Proof

- Exact command / steps: called the real `parse_response_with_memory` with a `<memory>["x"]</memory>` block, a `{"memories": "nope"}` block, a malformed block, and a valid block; then reverted `inline_extractor.py` and re-ran.
- Observed result: with the fix all four return cleanly (empty memories for the bad shapes, the parsed list for the valid one) and the memory block is still stripped from the content; with the fix reverted the non-object block raises `AttributeError: 'list' object has no attribute 'get'`. Ran against the actual module.
- Not tested: a live end-to-end chat where a model emits a non-object memory block.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/memory/inline_extractor.py
+++ b/headroom/memory/inline_extractor.py
@@ -126,9 +126,21 @@ def parse_response_with_memory(response_text: str) -> ParsedResponse:
         # Parse the JSON
         try:
             data = json.loads(memory_json)
-            memories = data.get("memories", [])
         except json.JSONDecodeError as e:
             logger.warning(f"Failed to parse memory JSON: {e}")
+        else:
+            # The model controls the block content. A valid-JSON non-object
+            # (e.g. a bare array) would make `.get` raise AttributeError, which
+            # the JSONDecodeError guard above does not catch; likewise a
+            # non-list `memories` would break downstream iteration.
+            if isinstance(data, dict):
+                extracted = data.get("memories", [])
+                if isinstance(extracted, list):
+                    memories = extracted
+                else:
+                    logger.warning("Memory JSON 'memories' field is not a list; ignoring")
+            else:
+                logger.warning("Memory JSON is not an object; ignoring")
 
     return ParsedResponse(
         content=content,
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the current head.

This is a good defensive fix: the parsed `<memory>` payload is model-controlled, so accepting only an object with a list-valued `memories` field preserves the existing valid path while preventing valid-JSON non-objects and wrong-shaped fields from crashing the inline extractor. The regression covers the crash shape, malformed JSON, wrong `memories` type, and the valid case. Latest CI is green/skipped as expected for this surface.

PR #2472, opened 2026-07-21
fix(providers/anthropic): don't crash token estimation on null tool_calls

abhay-codes07:
## Description

`AnthropicTokenCounter._count_message_estimated` (the tiktoken-approximation fallback used when no Anthropic client is available) counted OpenAI-format tool calls like this:

```python
if "tool_calls" in message:
    for tool_call in message.get("tool_calls", []):
        if isinstance(tool_call, dict):
            func = tool_call.get("function", {})
            ...
```

The `if "tool_calls" in message` check only tests key presence, not the value. OpenAI SDKs routinely include `"tool_calls": null` on an assistant message with no tool calls, so `message.get("tool_calls", [])` returned `None` (the default only applies when the key is absent) and `for tool_call in None` raised `TypeError: 'NoneType' object is not iterable`. That crashes token estimation for the entire request whenever such a message is present. `tool_call.get("function", {})` had the same gap for a `"function": null`.

## Fix

Iterate `message.get("tool_calls") or []` so a null or absent value becomes an empty list, and read `function` with `or {}` for the same reason. Valid tool calls are counted exactly as before.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/providers/anthropic.py`: value-guard `tool_calls` and `function` in `_count_message_estimated`.
- `tests/test_providers/test_anthropic.py`: regression counting a message list that includes `tool_calls: null` and a tool call with `function: null`.

## Testing

### Test Output

```text
$ python -m pytest tests/test_providers/test_anthropic.py -q
17 passed

$ uvx ruff@0.15.17 check headroom/providers/anthropic.py tests/test_providers/test_anthropic.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/providers/anthropic.py
```

## Real Behavior Proof

- Exact command / steps: built a real `AnthropicTokenCounter('claude-3-5-sonnet-20241022')` and called `count_messages` / `_count_message_estimated` with an assistant message carrying `tool_calls: null` and one carrying `function: null`, plus a valid tool call; then reverted `anthropic.py` and re-ran.
- Observed result: with the fix the null shapes count without error and a valid tool call still adds its name/arguments tokens (5 -> 11 on the sample); with the fix reverted the `tool_calls: null` message raises `TypeError: 'NoneType' object is not iterable`. Ran against the actual module.
- Not tested: a live request from an SDK that emits `tool_calls: null`, end to end.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/providers/anthropic.py
+++ b/headroom/providers/anthropic.py
@@ -407,13 +407,14 @@ class AnthropicTokenCounter(TokenCounter):
             # str(block) catch-all would produce.
             tokens += count_content_blocks(content, self.count_text)
 
-        # OpenAI format tool calls
-        if "tool_calls" in message:
-            for tool_call in message.get("tool_calls", []):
-                if isinstance(tool_call, dict):
-                    func = tool_call.get("function") or {}
-                    tokens += self.count_text(coerce_countable_text(func.get("name")))
-                    tokens += self.count_text(coerce_countable_text(func.get("arguments")))
+        # OpenAI format tool calls. Guard the value, not just the key: an
+        # OpenAI-format assistant message often carries `tool_calls: null` on a
+        # no-tool turn, and `for ... in None` would raise TypeError.
+        for tool_call in message.get("tool_calls") or []:
+            if isinstance(tool_call, dict):
+                func = tool_call.get("function") or {}
+                tokens += self.count_text(coerce_countable_text(func.get("name")))
+                tokens += self.count_text(coerce_countable_text(func.get("arguments")))
 
         return tokens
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the current green head.

This is a narrow defensive fix for the estimated Anthropic token counter: `tool_calls: null` now behaves like an absent tool-call list, and `function: null` no longer crashes while valid tool calls continue through the same counting path. The regression covers both null shapes, and the full refreshed CI set is green.

JerrettDavis (approved):
Re-reviewed at `3b03b79d`. The refreshed fix correctly composes the null-safe `tool_calls` iteration with main’s newer non-string field coercion, so both malformed shapes fail soft without losing token accounting. The focused tests remain and current checks have no failures.

PR #2479, opened 2026-07-22
fix(proxy): keep buffered CCR streams alive

### LINKED ISSUE #2465: [BUG] CCR buffered streaming sends no keepalive: stream:true is silently forced to stream:false, client hits "Stream idle timeout - no chunks received" on slow upstreams (Foundry / long Opus turns)
## Summary

When CCR is active (default), any `stream:true` request that carries the injected
`headroom_retrieve` tool is **silently converted to `stream:false` upstream** and buffered
in full, then re-synthesized into SSE after the complete response arrives. During that
buffered wait — which can last the entire generation, up to
`HEADROOM_ANTHROPIC_BUFFERED_REQUEST_TIMEOUT_SECONDS` (default **600s**) — **headroom sends
the client nothing**: no `message_start`, no `event: ping` keepalive, zero bytes.

The client's own stream-idle guard then fires long before headroom's read timeout. On
Claude Code this surfaces as:

```
API Error: Stream idle timeout - no chunks received
```

(The string is not emitted anywhere in headroom — it's the Claude Code client aborting a
stream that produced no chunks.)

This is the *earlier-firing sibling* of the already-closed #1261 (proxy 502 ReadTimeout on
long Opus `/v1/messages` turns): same buffered path, two different expiries. If the client's
idle window is shorter than headroom's 600s read timeout — which it usually is — the client
gives up first ("no chunks received"); if headroom's read timeout is shorter, headroom 502s
(#1261). Either way the root cause is the same: **a buffered stream with no keepalive.**

Reproduces most reliably on **Azure AI Foundry** (`CLAUDE_CODE_USE_FOUNDRY=1`, Anthropic-format
endpoint), but it is provider-agnostic: any upstream slow enough that a full non-streaming
generation exceeds the client idle window will hit it, including a plain long Opus turn.

## Environment

- **headroom**: `<version>` (`headroom --version`)
- **CCR**: default (enabled — `HEADROOM_NO_CCR` unset), so `headroom_retrieve` is injected
- **Upstream**: Azure AI Foundry (`CLAUDE_CODE_USE_FOUNDRY=1` + `ANTHROPIC_FOUNDRY_RESOURCE`,
  `.../anthropic/v1/messages`). Also observable on direct Anthropic with a long enough turn.
- **Client**: Claude Code `<version>`, streaming (`stream:true`)
- **OS / arch**: `<e.g. macOS arm64>`

## Root cause

1. `headroom/proxy/handlers/anthropic.py` (~2757): the buffering gate

   ```python
   buffered_stream_ccr = bool(
       stream
       and ccr_response_handler_enabled
       and self._has_headroom_retrieve_tool(tools if tools is not None else body.get("tools"))
   )
   if buffered_stream_ccr:
       if body.get("stream") is not False:
           body["stream"] = False   # <-- stream:true silently becomes stream:false upstream
       logger.info(f"[{request_id}] CCR: stream:true request has headroom_retrieve available; "
                   "using buffered stream:false upstream request for server-side retrieval handling")
   ```

2. The buffered upstream read uses `_anthropic_buffered_request_timeout()` (~221) →
   `HEADROOM_ANTHROPIC_BUFFERED_REQUEST_TIMEOUT_SECONDS`, **default 600s**.

3. Only after the full JSON is received is it turned into SSE — `_response_to_sse(resp_json,
   "anthropic")` (~3513). There is **no interim keepalive** emitted to the client while the
   buffered upstream call is in flight. (`grep` for `event: ping` / `message_start` emission on
   this path finds nothing during the wait.)

4. CCR (and therefore the `headroom_retrieve` injection that arms this gate) is **on by
   default**: `settings_store.py` `HEADROOM_NO_CCR` defaults `False`, and
   `cli/proxy.py:1210` sets `ccr_inject_tool = not no_ccr`.

So out of the box, streaming + CCR = a fully-buffered upstream request with a silent client,
and the client's idle guard fires with "no chunks received."

## To reproduce

1. `export CLAUDE_CODE_USE_FOUNDRY=1 ANTHROPIC_FOUNDRY_RESOURCE=<resource>`
2. `headroom wrap claude` (CCR at default → `headroom_retrieve` injected).
3. Ask for anything whose generation takes longer than the client's stream-idle window
   (easy on Foundry, especially with a large materialized-tools context — see interaction note).
4. → `API Error: Stream idle timeout - no chunks received`, with no partial output.

**Interaction note (how users actually arrive here).** Users on Foundry first hit the
Tool Search Tool 400 (Foundry can't run it) and are told to set `--tool-search false`. That
fix materializes every MCP/system tool schema into the prompt → larger context → slower
non-streaming Foundry generation → the buffered wait now reliably exceeds the client idle
window. The two issues compound. (See the companion Foundry tool-search issue.)

## Expected

A `stream:true` request should keep producing client-visible progress. Either:
- headroom genuinely streams while still supporting CCR retrieval, **or**
- while it must buffer, it emits periodic SSE keepalive (an initial `message_start` +
  `event: ping` heartbeats) so the client's idle guard doesn't trip.

## Actual

Zero client bytes for the whole generation; client aborts with "no chunks received" (or, if
headroom's 600s read timeout wins the race, a 502 ReadTimeout per #1261).

## Workaround (for anyone landing here)

Disable CCR so the retrieve tool isn't injected — the gate then never arms and real SSE is
proxied through:

```bash
export HEADROOM_NO_CCR=1
# or on the proxy:
headroom proxy --no-ccr
```

Trade-off: you lose on-demand retrieval of compressed originals (CCR reversibility); input
compression (SmartCrusher / CodeCompressor / Kompress) still applies. If the timeout persists
after this, it's no longer CCR — it's raw prefill latency over a large context (trim the MCP
tools loaded, or raise the client's stream-idle timeout).

## Proposed fix

Emit keepalive on the buffered CCR path. While the `stream:false` upstream call is in flight,
open the client SSE immediately and send an initial `message_start` plus periodic
`event: ping` heartbeats (Anthropic's own streaming does this), then flush the re-synthesized
events from `_response_to_sse` once the buffered response lands. That preserves CCR's
server-side retrieval handling while keeping the client's idle guard satisfied.

Lower-effort interim: at minimum surface a launch/`doctor` note that CCR forces streaming
requests into a buffered path (so long turns on slow upstreams need `HEADROOM_NO_CCR=1` or a
higher client idle timeout), rather than failing silently with a confusing client-side error.

## Related

- **#1261** (closed) — proxy 502 ReadTimeout on long Opus `/v1/messages` turns. Same buffered
  path; the *headroom-side* expiry instead of the *client-side* one.
- **#2089** (closed) — Anthropic streaming 502 when `headroom_retrieve` is called. Same
  subsystem (CCR retrieval over streaming), different trigger.
- **#1897** (closed) — "preserve thinking blocks in buffered stream re-synthesis." Confirms
  the buffer → SSE re-synthesis path this issue is about.
- **#1877** (open) — CCR streaming interception on the OpenAI Responses API. Same mechanism,
  different provider surface.
- Companion Foundry tool-search issue (the `ENABLE_TOOL_SEARCH=true` → 400 report) — this is
  the failure users hit immediately after applying that workaround.

rodboev:
## Description

Buffered CCR streaming currently waits for the full upstream response before sending any bytes back to the client. On the Anthropic path this shows up as `API Error: Stream idle timeout - no chunks received`, and the same buffer-then-synthesize mechanism still exists on the `/v1/responses` CCR path. This adds a narrow buffered-stream heartbeat layer so the client sees early stream activity while Headroom preserves the existing server-side retrieval round trip and final synthesized provider events.

Closes #2465

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- open buffered CCR streams early and emit client-visible `event: ping` heartbeats while the buffered upstream call is still in flight
- preserve the existing terminal Anthropic and Responses synthesis helpers instead of replacing their event-building logic
- preserve early non-streaming failure semantics before the first heartbeat, including normal 429 passthrough and normal JSON 502 failures
- log late buffered-task exceptions server-side and record one failed provider metric on that post-keepalive branch, while keeping the client-facing SSE error sanitized
- add focused delayed-upstream regression coverage for both buffered provider paths, their early-failure branches, and their late-failure branches

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_proxy/test_openai_responses_ccr.py -q`)
- [x] Linting passes (`uv run ruff check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_proxy/test_openai_responses_ccr.py`)

### Test Output

```text
uv run pytest tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_proxy/test_openai_responses_ccr.py -q
======================= 17 passed, 1 warning in 42.47s ========================

uv run ruff check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_proxy/test_openai_responses_ccr.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows, Python via `uv`, proxy handler tests with gated buffered upstream fixtures
- Exact command / steps: run the focused Anthropic and Responses CCR suites above; delayed-upstream tests consume the first client-visible SSE event before releasing the upstream, then consume the synthesized final events
- Observed result: both buffered paths emitted `event: ping` before upstream release; pre-keepalive 429 responses preserved their real status and headers, pre-keepalive exceptions returned the normal JSON 502 shape, late transport failures recorded one failed provider metric and one server error log before emitting one sanitized SSE error, Anthropic preserved `done`, and Responses preserved `Resolved!`
- Not tested: live slow upstream run with Claude Code or a real Responses client

## Review Readiness

## Checklist

## Additional Notes

`CHANGELOG.md` stays untouched because Headroom generates release notes from conventional commits. The PR should only claim the local buffered-stream contract and focused regression coverage; live client proof remains an owner check on a slow real upstream.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -3868,10 +3868,7 @@ class AnthropicHandlerMixin:
                     # blocks every other request for the duration; a timeout
                     # here is caught below and passes the item through.
                     result = await self._run_compression_in_executor(
-                        lambda messages=messages,
-                        model=model,
-                        context_limit=context_limit,
-                        frozen_message_count=frozen_message_count: (
+                        lambda messages=messages, model=model, context_limit=context_limit, frozen_message_count=frozen_message_count: (
                             self.anthropic_pipeline.apply(
                                 messages=messages,
                                 model=model,
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -3575,9 +3575,12 @@ class AnthropicHandlerMixin:
                                     if done:
                                         try:
                                             result = operation.result()
-                                        except Exception:
+                                        except Exception as e:
+                                            await record_failed(provider=provider_name)
+                                            logger.error(
+                                                f"[{request_id}] Request failed: {type(e).__name__}: {e}"
+                                            )
                                             if not started:
-                                                await record_failed(provider=provider_name)
                                                 await send(
                                                     {
                                                         "type": "http.response.start",
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -2824,726 +2824,858 @@ class AnthropicHandlerMixin:
                         session_key=session_key,
                     )
                 else:
-                    async with stage_timer.measure("upstream_connect"):
-                        response = await self._retry_request(
-                            "POST",
-                            url,
-                            headers,
-                            body,
-                            original_body_bytes=original_body_bytes,
-                            body_mutated=body_mutation_tracker.mutated,
-                            mutation_reasons=body_mutation_tracker.reasons,
-                            request_id=request_id,
-                            forwarder_name="anthropic_messages",
-                            path_for_log="/v1/messages",
-                            timeout=self._anthropic_buffered_request_timeout(),
-                        )
-                    self.pipeline_extensions.emit(
-                        PipelineStage.POST_SEND,
-                        operation="proxy.request",
-                        request_id=request_id,
-                        provider=pipeline_provider,
-                        model=model,
-                        messages=body["messages"],
-                        tools=tools,
-                        response=response,
-                        metadata={
-                            "path": pipeline_path,
-                            "stream": False,
-                            "client_stream": buffered_stream_ccr,
-                            "ccr_stream_buffered": buffered_stream_ccr,
-                            "status_code": response.status_code,
-                        },
-                    )
-                    self.pipeline_extensions.emit(
-                        PipelineStage.RESPONSE_RECEIVED,
-                        operation="proxy.request",
-                        request_id=request_id,
-                        provider=pipeline_provider,
-                        model=model,
-                        response=response,
-                        metadata={
-                            "path": pipeline_path,
-                            "stream": False,
-                            "client_stream": buffered_stream_ccr,
-                            "ccr_stream_buffered": buffered_stream_ccr,
-                            "status_code": response.status_code,
-                        },
-                    )
-                    if (
-                        "upstream_first_byte" not in stage_timer
-                        and "upstream_connect" in stage_timer
-                    ):
-                        stage_timer.record(
-                            "upstream_first_byte",
-                            stage_timer.summary()["upstream_connect"],
-                        )
-                    await _finalize_pre_upstream()
-                    # Full diagnostic dump on upstream errors.
-                    # Writes pre/post compression messages, tools, and error
-                    # to ~/.headroom/logs/debug_400/ for offline analysis.
-                    if response.status_code >= 400:
-                        try:
-                            err_body = response.json()
-                            err_msg = err_body.get("error", {}).get("message", "")
-                            err_type = err_body.get("error", {}).get("type", "")
-                        except Exception:
-                            err_body = {"raw": response.text[:2000]}
-                            err_msg = str(response.text[:500])
-                            err_type = "parse_error"
 
-                        logger.warning(
-                            f"[{request_id}] UPSTREAM_ERROR "
-                            f"status={response.status_code} "
-                            f"error_type={err_type} "
-                            f"error_msg={err_msg!r} "
-                            f"model={model} "
-                            f"compressed={'yes' if transforms_applied else 'no'} "
-                            f"transforms={transforms_applied} "
-                            f"original_tokens={original_tokens} "
-                            f"optimized_tokens={optimized_tokens} "
-                            f"message_count={len(body.get('messages', []))} "
-                            f"stream={stream}"
-                        )
-
-                        # Diagnostic dump of the full upstream-error request.
-                        # OFF by default: it can contain cleartext prompt / tool /
-                        # system content. Opt in with HEADROOM_DEBUG_DUMP=1
-                        # (redacted: structure + lengths only) or =full (content).
-                        # Never written in stateless mode.
-                        dump_mode = _debug_dump_mode(self.config)
-                        if dump_mode != "off":
-                            try:
-                                from headroom import paths as _hr_paths
-
-                                debug_dir = _hr_paths.debug_400_dir()
-                                debug_dir.mkdir(parents=True, exist_ok=True)
-                                ts = datetime.now().strftime("%Y%m%d_%H%M%S")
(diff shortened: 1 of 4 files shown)


JerrettDavis (changes requested):
Thanks for the careful coverage here. The keepalive behavior itself looks solid under the focused tests, but I found one blocker in the new task wrapper before this is ready.

When the buffered operation raises after the first ping has committed the SSE response, the custom ASGI response catches `operation.result()` and emits a sanitized SSE error, but it does not log the exception or record a failed provider metric on that late-failure branch. That exception used to flow to the outer handler, where it was logged and counted; now it is consumed inside `_BufferedCCRResponse.__call__`. The same pattern exists in both `headroom/proxy/handlers/anthropic.py` and `headroom/proxy/handlers/openai.py`.

Please add server-side logging and failed-metric recording for the post-keepalive exception branch as well, keeping the client-facing SSE sanitized. A focused assertion around the existing late-failure tests would be enough to lock this down.

Local verification while reviewing:
- `uv run pytest tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_proxy/test_openai_responses_ccr.py -q` -> 17 passed, 1 warning
- `uv run ruff check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_proxy/test_openai_responses_ccr.py` -> All checks passed

rodboev:
Thanks for catching this. I restored the failure telemetry when the buffered operation exception is caught, while keeping the client-facing SSE error sanitized.

1. In `headroom/proxy/handlers/anthropic.py`, the `_BufferedCCRResponse` exception path now records one failed Anthropic request and emits the same server error details as the outer handler before returning either the pre-start JSON response or the post-ping SSE error.
2. In `headroom/proxy/handlers/openai.py`, the matching Responses path now records one failed OpenAI request and logs the exception before returning through the same two response branches.
3. The existing late-failure tests now assert the provider metric and server log in both handlers. They still check that a ping reaches the client first, the internal exception text stays out of the SSE payload, and the response closes with one sanitized error event.

The keepalive timing, the normal final provider events, early HTTP failure behavior, and cancellation cleanup remain unchanged.

PR #2481, opened 2026-07-22
fix(proxy/streaming): tolerate malformed content in _response_to_sse

abhay-codes07:
## Description

`StreamingMixin._response_to_sse` rebuilds an Anthropic SSE stream from a buffered response dict. It iterated the content and read usage with no type guards:

```python
for idx, block in enumerate(response.get("content", [])):
    if block.get("type") == "text":
        ...
...
"usage": {"output_tokens": response.get("usage", {}).get("output_tokens", 0)},
```

`response` here is provider- and reconstruction-controlled. `.get("content", [])` only falls back when the key is absent, so a present-but-null `content` returns `None` and `enumerate(None)` raises `TypeError`. A non-list `content` (e.g. a bare string) makes `block.get(...)` raise `AttributeError`, and a null element inside the list hits the same `AttributeError`. `response.get("usage", {}).get(...)` breaks the same way on `usage: null`.

This matters because the Anthropic buffered CCR path calls it inside an `except ValueError` guard only:

```python
try:
    sse_events = self._response_to_sse(resp_json, "anthropic")
except ValueError as sse_err:
    ...
```

A `TypeError`/`AttributeError` from any of the shapes above escapes that guard and 500s the streamed request. The sibling `_record_ccr_feedback_from_response` in the same class already guards `content` for list-ness and skips non-dict blocks, so this closes the asymmetry.

## Fix

Coerce `content` to a list before iterating (non-list becomes empty), skip any non-dict block, and coerce a non-dict `usage` to `{}` before reading `output_tokens`. Well-formed responses render byte-for-byte as before.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/streaming.py`: list-guard `content`, skip non-dict blocks, and dict-guard `usage` in `_response_to_sse`.
- `tests/test_sse_thinking_blocks.py`: regression rendering responses with null/non-list content, a null block element, and null usage, plus a check that a valid block alongside a null element still renders.

## Testing

### Test Output

```text
$ python -m pytest tests/test_sse_thinking_blocks.py -q
14 passed

$ uvx ruff@0.15.17 check headroom/proxy/handlers/streaming.py tests/test_sse_thinking_blocks.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/handlers/streaming.py
```

## Real Behavior Proof

- Exact command / steps: called `StreamingMixin()._response_to_sse(response, "anthropic")` with four malformed bodies (`content: null`, `content: "not-a-list"`, `content: [null, {text}]`, `usage: null`); then reverted `streaming.py` and re-ran the same inputs.
- Observed result: with the fix each body produces a well-formed SSE envelope (message_start ... message_stop) and the valid block alongside a null element still emits its text_delta; with the fix reverted the `content: null` body raises `TypeError: 'NoneType' object is not iterable` and the others raise `AttributeError`. Ran against the actual module via `tests/test_sse_thinking_blocks.py`.
- Not tested: a live upstream returning a malformed buffered response end to end through the CCR path.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/handlers/streaming.py
+++ b/headroom/proxy/handlers/streaming.py
@@ -558,8 +558,17 @@ class StreamingMixin:
         }
         events.append(f"event: message_start\ndata: {json.dumps(msg_start)}\n\n".encode())
 
-        # Content blocks
-        for idx, block in enumerate(response.get("content", [])):
+        # Content blocks. `content` is provider/reconstruction-controlled, so a
+        # present-but-null value or a non-list would crash `enumerate`, and a
+        # non-dict element would crash `block.get(...)`. Guard both, matching the
+        # sibling `_record_ccr_feedback_from_response` below. This is reached from
+        # a call site (anthropic.py buffered CCR path) that only catches
+        # ValueError, so an unguarded TypeError/AttributeError would 500 the
+        # streamed request.
+        content = response.get("content")
+        for idx, block in enumerate(content if isinstance(content, list) else []):
+            if not isinstance(block, dict):
+                continue
             # content_block_start
             if block.get("type") == "text":
                 block_start = {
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest change and this looks ready. `_response_to_sse` now mirrors the defensive shape handling already used in the CCR feedback path: non-list content is treated as empty, non-dict blocks are skipped, and non-dict usage is read as `{}`. Existing well-formed text/thinking/tool block rendering remains covered by the surrounding tests.

PR state is mergeable and visible checks are green. I also ran the focused SSE test module locally on this branch:

```text
uv run --frozen --extra dev pytest tests/test_sse_thinking_blocks.py
14 passed
```

PR #2482, opened 2026-07-22
fix(doctor): don't crash on a valid-but-non-object settings.json

abhay-codes07:
## Description

`headroom doctor` parses `~/.claude/settings.json` in two checks:

```python
try:
    payload = json.loads(settings_path.read_text(encoding="utf-8"))
except (OSError, ValueError) as exc:
    return CheckResult(... WARN "could not parse" ...)
...
env_block = payload.get("env")
```

`json.loads` returns a non-dict for any valid JSON that is not an object: `[]`, `null`, `42`, `"a string"`. None of those raise `JSONDecodeError`, so they slip past the `except (OSError, ValueError)` guard, and the following `payload.get("env")` raises `AttributeError`. `AttributeError` is not in the caught tuple, so it escapes and crashes `doctor` with a traceback. That is the worst moment for it: `doctor` is the command a user runs precisely because their config is suspect, and a hand-edited or reset `settings.json` holding `[]` or `null` is exactly the kind of file it should report on, not fall over on.

Two functions have this shape: `check_claude_routing` (the `.get` is after the `try` returns) and `check_claude_remote_control_gate` (the `.get` is inside a `try` whose `except` is also `(OSError, ValueError)`).

## Fix

Guard `payload` for dict-ness in both checks. `check_claude_routing` now returns the same WARN it already returns for unparseable files, with a "not a JSON object" summary; `check_claude_remote_control_gate` treats a non-object as having no `env` block, so the shell environment still drives the gate. Well-formed object settings behave exactly as before.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cli/doctor.py`: guard `payload` for dict-ness in `check_claude_routing` and `check_claude_remote_control_gate` before calling `.get`.
- `tests/test_cli_doctor.py`: parametrized regressions feeding `[]`, `null`, `42`, and a bare string to both checks.

## Testing

### Test Output

```text
$ python -m pytest tests/test_cli_doctor.py -q
68 passed

# with the fix reverted, the new tests fail with
# AttributeError: 'list' object has no attribute 'get'

$ uvx ruff@0.15.17 check headroom/cli/doctor.py tests/test_cli_doctor.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/cli/doctor.py
```

## Real Behavior Proof

- Exact command / steps: wrote a `settings.json` containing `[]` (and `null`, `42`, `"a string"`) into a tmp path and called `check_claude_routing(path, 8787)` and `check_claude_remote_control_gate(path, {"ANTHROPIC_BASE_URL": "http://127.0.0.1:8787"})`; then reverted `doctor.py` and re-ran.
- Observed result: with the fix both checks return a WARN result instead of raising; with the fix reverted both raise `AttributeError: 'list' object has no attribute 'get'` (and the analogous message for `null`/`42`/string). Ran against the actual module via `tests/test_cli_doctor.py`.
- Not tested: the full `headroom doctor` CLI end to end against a real `~/.claude/settings.json`.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/cli/doctor.py
+++ b/headroom/cli/doctor.py
@@ -157,6 +157,17 @@ def check_claude_routing(settings_path: Path, port: int) -> CheckResult:
             status=WARN,
             summary=f"could not parse {settings_path}: {exc}",
         )
+    # `json.loads` succeeds on valid non-object JSON (e.g. `[]`, `null`, `42`),
+    # which a hand-edited or reset settings file can contain. `.get` on a
+    # non-dict raises AttributeError, and it is not one of the caught parse
+    # errors above, so it would crash the very command run to diagnose the
+    # broken config. Treat a non-object like an unparseable file.
+    if not isinstance(payload, dict):
+        return CheckResult(
+            name=name,
+            status=WARN,
+            summary=f"could not parse {settings_path}: not a JSON object",
+        )
     base_url = ""
     env_block = payload.get("env")
     if isinstance(env_block, dict):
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest change and it looks ready. Both affected doctor checks now guard valid-but-non-object JSON before calling `.get`, with the routing check surfacing a useful WARN and the remote-control gate falling back to the shell environment as before. Well-formed object settings keep the existing behavior.

PR state is mergeable and visible checks are green. I also ran the full focused doctor test file locally on this branch:

```text
uv run --frozen --extra dev pytest tests/test_cli_doctor.py
68 passed
```

PR #2483, opened 2026-07-22
fix(proxy/memory): harden _extract_tool_calls against malformed choices and blocks

abhay-codes07:
## Description

`MemoryHandler._extract_tool_calls` reads tool calls from an upstream response. The OpenAI Chat Completions branch only truthiness-checked `choices` before indexing:

```python
if choices:
    message = choices[0].get("message", {})
    ...
```

An OpenAI-compatible gateway can send `choices: [null]` on a content-filtered or usage-only response (the same shape the sibling `CCRResponseHandler._extract_assistant_message` was explicitly hardened for, see its comment: "a present-but-empty `choices: []` (or `[null]`) which OpenAI-compatible gateways can send"). With `[null]`, `choices` is truthy, so `choices[0].get("message", {})` runs on `None` and raises `AttributeError`. A non-dict first element (`choices: ["..."]`) breaks the same way.

This is not caught locally: `_extract_tool_calls` runs from `has_memory_tool_calls`, which the OpenAI handler calls in the `if (...)` condition **before** the `try` that wraps the rest of memory handling, so the `AttributeError` surfaces to the request handler rather than being swallowed and logged.

The Anthropic branch had a smaller version of the same gap: `[block for block in content if block.get("type") == "tool_use"]` guards `content` for list-ness but not each block, so a null element in a reconstructed response crashes `block.get`.

## Fix

Guard the first choice and its `message` for dict-ness before calling `.get`, mirroring `_extract_assistant_message`. A malformed choice now yields no tool calls and correctly falls through to the Responses-API `output[]` check. Skip non-dict blocks in the Anthropic branch. Well-formed responses are unchanged.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/memory_handler.py`: dict-guard the first choice and its `message` in the OpenAI branch of `_extract_tool_calls`; skip non-dict content blocks in the Anthropic branch.
- `tests/test_memory_handler_null_function.py`: regressions for `choices: [null]` / non-dict choice / null message, fall-through to a Responses `output[]` function_call, and a non-dict Anthropic content block.

## Testing

### Test Output

```text
$ python -m pytest tests/test_memory_handler_null_function.py -q
5 passed

# with the fix reverted, the three new tests fail with
# AttributeError: 'NoneType' object has no attribute 'get'

$ uvx ruff@0.15.17 check headroom/proxy/memory_handler.py tests/test_memory_handler_null_function.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/memory_handler.py
```

## Real Behavior Proof

- Exact command / steps: called `_extract_tool_calls` / `has_memory_tool_calls` on a bare `MemoryHandler` (via `object.__new__`) with `choices: [null]`, `choices: ["x"]`, `choices: [{"message": null}]`, a `[null]`-choices response that also carries a Responses `output[]` function_call, and an Anthropic response whose content list starts with a null block; then reverted `memory_handler.py` and re-ran.
- Observed result: with the fix the malformed choices return `[]` / `False`, the `output[]` fall-through is still detected as `True`, and the valid Anthropic tool_use block is returned; with the fix reverted the `choices: [null]` and non-dict-block cases raise `AttributeError: 'NoneType' object has no attribute 'get'`. Ran against the actual module via `tests/test_memory_handler_null_function.py`.
- Not tested: a live OpenAI-compatible gateway emitting `choices: [null]` on a memory-enabled request, end to end.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/memory_handler.py
+++ b/headroom/proxy/memory_handler.py
@@ -1077,7 +1077,12 @@ your responses, not to drive new actions."""
             if isinstance(first, dict):
                 message = first.get("message")
                 if isinstance(message, dict):
-                    tc_list = list(message.get("tool_calls", []) or [])
+                    # Filter to dict entries: a null / string element inside
+                    # tool_calls (which downstream immediately calls `.get` on)
+                    # would otherwise still crash detection.
+                    tc_list = [
+                        tc for tc in (message.get("tool_calls") or []) if isinstance(tc, dict)
+                    ]
                     if tc_list:
                         return tc_list
 
--- a/headroom/proxy/memory_handler.py
+++ b/headroom/proxy/memory_handler.py
@@ -1056,17 +1056,30 @@ your responses, not to drive new actions."""
         if provider == "anthropic":
             content = response.get("content", [])
             if isinstance(content, list):
-                return [block for block in content if block.get("type") == "tool_use"]
+                return [
+                    block
+                    for block in content
+                    if isinstance(block, dict) and block.get("type") == "tool_use"
+                ]
             return []
 
         elif provider == "openai":
-            # Chat Completions format: choices[0].message.tool_calls
-            choices = response.get("choices", [])
-            if choices:
-                message = choices[0].get("message", {})
-                tc_list = list(message.get("tool_calls", []) or [])
-                if tc_list:
-                    return tc_list
+            # Chat Completions format: choices[0].message.tool_calls. Guard the
+            # first choice for dict-ness the way the sibling
+            # CCRResponseHandler._extract_assistant_message already does: an
+            # OpenAI-compatible gateway can send `choices: [null]` on a
+            # content-filtered / usage-only response, and this runs from
+            # has_memory_tool_calls, which is called outside the try that wraps
+            # the rest of memory handling — so an unguarded `choices[0].get` here
+            # would surface AttributeError to the request handler.
+            choices = response.get("choices")
+            first = choices[0] if isinstance(choices, list) and choices else None
+            if isinstance(first, dict):
+                message = first.get("message")
+                if isinstance(message, dict):
+                    tc_list = list(message.get("tool_calls", []) or [])
+                    if tc_list:
+                        return tc_list
 
             # Responses API format: output[] with type=function_call
             output = response.get("output", [])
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
Thanks for hardening the outer `choices` and Anthropic block shapes. There is one remaining malformed upstream shape in the OpenAI branch that still crashes the same detection path:

```python
response = {"choices": [{"message": {"tool_calls": [None]}}]}
handler._extract_tool_calls(response, "openai")  # returns [None]
handler.has_memory_tool_calls(response, "openai")  # AttributeError: 'NoneType' object has no attribute 'get'
```

I verified that locally on the latest branch. `_extract_tool_calls` now guards `choices[0]` and `message`, but it still returns every element from `message.tool_calls` without checking the element type. Since both `has_memory_tool_calls` and `handle_memory_tool_calls` immediately call `.get` on each returned entry, a null or string inside the tool-call list still escapes and can 500 before the memory handling try block.

Please filter OpenAI Chat Completions `tool_calls` to dict entries before returning them, and add a regression for `tool_calls: [None, {valid memory_save call}]` that proves the malformed element is skipped while the valid call is still detected.

Other checked state: PR is mergeable and visible CI is green. I did not run the full test file after finding this blocker; the reproduction above is from the checked-out latest source.

abhay-codes07:
Good catch, you're right that filtering the outer shapes wasn't enough while the elements themselves went straight to `.get` downstream. Pushed a fix.

The OpenAI branch now filters `message.tool_calls` to dict entries before returning, so `[None]` / `["oops"]` elements are dropped the same way the Anthropic branch already skips non-dict blocks. Added the regression you asked for: `tool_calls: [None, "oops", {valid memory_save}]` now returns only the valid call and `has_memory_tool_calls` still reports True.

Also rebased onto the latest main so it's clean against the recent merges.

JerrettDavis (approved):
Latest update addresses the remaining malformed OpenAI tool_calls case from my earlier review. I verified locally with pytest tests/test_memory_handler_null_function.py -q (6 passed) and ruff check headroom/proxy/memory_handler.py tests/test_memory_handler_null_function.py (clean).

PR #2484, opened 2026-07-22
fix(backends): don't crash the OpenAI->Anthropic converter on empty choices

abhay-codes07:
## Description

`_to_anthropic_response` in both backends converts a non-streaming OpenAI-shape response to Anthropic shape and indexes the first choice directly:

```python
# headroom/backends/litellm.py
choice = litellm_response.choices[0]
# headroom/backends/anyllm.py
choice = response.choices[0]
```

A non-streaming upstream response can be HTTP 200 with an **empty** `choices` list: Azure OpenAI content filtering does exactly this, and any OpenAI-compatible gateway can return a usage-only / filtered turn the same way. With `choices: []`, `choices[0]` raises `IndexError`, which surfaces as a 500 for the request instead of a normal (if empty) turn.

This is an intra-file asymmetry: the streaming siblings in the same two files already guard it (`if not chunk.choices: continue` / `if hasattr(chunk, "choices") and chunk.choices:`), and `headroom/proxy/handlers/openai.py` documents the exact hazard in `_apply_stream_usage_option`: "the common `chunk.choices[0].delta` pattern then raises IndexError" on a usage-only `choices: []` chunk. The non-streaming converters just never got the same guard.

## Fix

Return a valid empty assistant turn (`content: []`, `stop_reason: "end_turn"`, usage still mapped) when `choices` is empty, before indexing. The client gets a clean empty response instead of a 500, matching how the streaming path already tolerates the same shape. Non-empty responses are unchanged.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/backends/litellm.py`: empty-`choices` guard at the top of `_to_anthropic_response`, returning an empty assistant turn with mapped usage.
- `headroom/backends/anyllm.py`: same guard in its `_to_anthropic_response`.
- `tests/test_litellm_nonstream_cache_usage.py`, `tests/test_backend_anyllm.py`: regressions passing an empty-`choices` response through each converter and asserting an empty turn instead of IndexError.

## Testing

### Test Output

```text
$ python -m pytest tests/test_litellm_nonstream_cache_usage.py::test_to_anthropic_response_empty_choices_returns_empty_turn tests/test_backend_anyllm.py::test_to_anthropic_response_empty_choices_returns_empty_turn -q
2 passed

# with the fix reverted, both fail with
# IndexError: list index out of range

$ uvx ruff@0.15.17 check headroom/backends/litellm.py headroom/backends/anyllm.py tests/test_backend_anyllm.py tests/test_litellm_nonstream_cache_usage.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/backends/litellm.py headroom/backends/anyllm.py
```

Note: `tests/test_backend_anyllm.py` has 7 `@pytest.mark.asyncio` tests that fail locally because pytest-asyncio is not configured in this environment (`Unknown config option: asyncio_mode`); they are unrelated to this change and pass in CI. The two new tests here are synchronous and pass locally.

## Real Behavior Proof

- Exact command / steps: built a response stand-in with `choices=[]` and a usage object, called `LiteLLMBackend._to_anthropic_response` (on a bare `object.__new__` instance) and `AnyLLMBackend._to_anthropic_response` (via the file's fake-backend fixture); then reverted both backend files and re-ran.
- Observed result: with the fix each converter returns `{type: message, role: assistant, content: [], stop_reason: end_turn, usage: {...}}` with the input/output token counts mapped; with the fix reverted both raise `IndexError: list index out of range`. Ran against the actual modules via the two test files.
- Not tested: a live Azure OpenAI content-filtered response routed through the backend end to end.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/backends/anyllm.py
+++ b/headroom/backends/anyllm.py
@@ -143,6 +143,30 @@ class AnyLLMBackend(Backend):
         """Convert any-llm/OpenAI response to Anthropic format."""
         msg_id = f"msg_{uuid.uuid4().hex[:24]}"
 
+        # A non-streaming upstream response can be HTTP 200 with an empty
+        # ``choices`` list (e.g. Azure OpenAI content filtering, or any
+        # OpenAI-compatible gateway on a usage-only / filtered turn). The
+        # streaming sibling already skips this (`if ... and chunk.choices`);
+        # indexing ``choices[0]`` here would instead raise IndexError and 500
+        # the request. Return a valid empty assistant turn.
+        if not getattr(response, "choices", None):
+            usage = {"input_tokens": 0, "output_tokens": 0}
+            if getattr(response, "usage", None):
+                usage = {
+                    "input_tokens": getattr(response.usage, "prompt_tokens", 0) or 0,
+                    "output_tokens": getattr(response.usage, "completion_tokens", 0) or 0,
+                }
+            return {
+                "id": msg_id,
+                "type": "message",
+                "role": "assistant",
+                "content": [],
+                "model": original_model,
+                "stop_reason": "end_turn",
+                "stop_sequence": None,
+                "usage": usage,
+            }
+
         choice = response.choices[0]
         message = choice.message
 
--- a/headroom/backends/litellm.py
+++ b/headroom/backends/litellm.py
@@ -793,6 +793,24 @@ class LiteLLMBackend(Backend):
         """Convert LiteLLM/OpenAI response to Anthropic format."""
         msg_id = f"msg_{uuid.uuid4().hex[:24]}"
 
+        # A non-streaming upstream response can be HTTP 200 with an empty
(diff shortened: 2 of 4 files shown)


JerrettDavis (approved):
Reviewed the latest diff and this looks ready. The guard is placed before the direct `choices[0]` access in both non-streaming converters, returns the same Anthropic message envelope shape as the normal path, and preserves mapped usage for the empty-choice/content-filtered case.

PR state is mergeable and visible checks are green. I also ran the focused regressions locally on this branch:

```text
uv run --frozen --extra dev pytest tests/test_litellm_nonstream_cache_usage.py::test_to_anthropic_response_empty_choices_returns_empty_turn tests/test_backend_anyllm.py::test_to_anthropic_response_empty_choices_returns_empty_turn
2 passed
```

PR #2485, opened 2026-07-22
fix(proxy/memory): harden MemoryToolAdapter._extract_tool_calls against malformed responses

abhay-codes07:
## Description

`MemoryToolAdapter._extract_tool_calls` reads tool calls out of an upstream response for each provider format. Its own `_get_tool_name` / `_get_tool_input` helpers were already hardened against malformed upstream data (see `test_memory_tool_adapter_null_fields.py`: "reachable from the untrusted upstream response"), but `_extract_tool_calls` itself indexed and chained `.get` with no guards:

```python
# openai
if choices:
    message = choices[0].get("message", {})   # choices: [null] -> AttributeError
    ...
# gemini
content = candidates[0].get("content", {})    # candidates: [null] -> AttributeError
parts = content.get("parts", [])              # content: null -> AttributeError
# anthropic
[block for block in content if block.get("type") == "tool_use"]  # null block -> AttributeError
```

An OpenAI-compatible gateway can send `choices: [null]` on a content-filtered / usage-only turn (the same shape the sibling `MemoryHandler._extract_tool_calls` was hardened for — see PR #2483). The truthiness check on `choices` lets `[null]` through, and `choices[0].get` then raises `AttributeError`. The Gemini branch has the same gap on a null candidate / null `content`, and the Anthropic branch (and the generic fallback) crash on a non-dict content block.

## Fix

Guard the first choice/candidate and the nested `message` / `content` / `parts` for type before calling `.get`, and skip non-dict elements — mirroring `MemoryHandler._extract_tool_calls` and this adapter's already-hardened `_get_tool_*` helpers. The OpenAI and Anthropic extraction is factored into two small helpers so the explicit branches and the generic fallback stay in lockstep. Well-formed responses are unchanged.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/memory_tool_adapter.py`: type-guard the OpenAI, Gemini, and Anthropic branches (and the generic fallback) of `_extract_tool_calls`; add `_extract_anthropic_tool_uses` / `_extract_openai_tool_calls` helpers shared by the explicit branches and the fallback.
- `tests/test_memory_tool_adapter_null_fields.py`: regressions for malformed OpenAI choices, malformed Gemini candidates/content/parts (plus a null part skipped), a non-dict Anthropic block, and a valid OpenAI call still parsed.

## Testing

### Test Output

```text
$ python -m pytest tests/test_memory_tool_adapter_null_fields.py -q
7 passed

# with the fix reverted, the three new malformed-shape tests fail with
# AttributeError: 'NoneType' object has no attribute 'get'

$ uvx ruff@0.15.17 check headroom/proxy/memory_tool_adapter.py tests/test_memory_tool_adapter_null_fields.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/memory_tool_adapter.py
```

## Real Behavior Proof

- Exact command / steps: called `_extract_tool_calls` on a bare `MemoryToolAdapter` (via `object.__new__`) with `choices: [null]` / `[str]` / `[{"message": null}]` (openai and generic), `candidates: [null]` / null content / null parts (gemini), and a null Anthropic content block; then reverted `memory_tool_adapter.py` and re-ran.
- Observed result: with the fix the malformed shapes return `[]`, a real Gemini `functionCall` part is still returned (null part skipped), and a valid OpenAI tool call still parses; with the fix reverted the openai `[null]`, gemini null-content, and non-dict-block cases raise `AttributeError: 'NoneType' object has no attribute 'get'`. Ran against the actual module via `tests/test_memory_tool_adapter_null_fields.py`.
- Not tested: a live OpenAI-compatible gateway emitting `choices: [null]` routed through the adapter end to end.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/memory_tool_adapter.py
+++ b/headroom/proxy/memory_tool_adapter.py
@@ -798,7 +798,10 @@ class MemoryToolAdapter:
         first = choices[0] if isinstance(choices, list) and choices else None
         message = first.get("message") if isinstance(first, dict) else None
         if isinstance(message, dict):
-            return list(message.get("tool_calls", []) or [])
+            # Filter to dict entries: downstream (_get_tool_name /
+            # _get_tool_input) calls `.get` on each element, so a null / string
+            # element inside tool_calls would still crash the same path.
+            return [tc for tc in (message.get("tool_calls") or []) if isinstance(tc, dict)]
         return []
 
     def _get_tool_name(self, tool_call: dict[str, Any], provider: Provider) -> str:
--- a/headroom/proxy/memory_tool_adapter.py
+++ b/headroom/proxy/memory_tool_adapter.py
@@ -747,44 +747,59 @@ class MemoryToolAdapter:
         response: dict[str, Any],
         provider: Provider,
     ) -> list[dict[str, Any]]:
-        """Extract tool calls from response based on provider format."""
+        """Extract tool calls from response based on provider format.
+
+        Every branch reads from the untrusted upstream response, so it guards
+        the shapes an OpenAI-compatible gateway can send on a content-filtered /
+        usage-only turn: a null element inside ``choices`` / ``candidates`` /
+        ``content`` (`[null]`), or a present-but-null nested object. Without the
+        guards, ``choices[0].get`` / ``block.get`` raises AttributeError, the
+        same class the sibling ``MemoryHandler._extract_tool_calls`` and this
+        adapter's own ``_get_tool_*`` helpers already defend against.
+        """
         if provider == "anthropic":
-            content = response.get("content", [])
-            if isinstance(content, list):
-                return [block for block in content if block.get("type") == "tool_use"]
-            return []
+            return self._extract_anthropic_tool_uses(response)
 
         elif provider == "openai":
-            choices = response.get("choices", [])
-            if choices:
-                message = choices[0].get("message", {})
-                return list(message.get("tool_calls", []) or [])
-            return []
+            return self._extract_openai_tool_calls(response)
 
         elif provider == "gemini":
             # Gemini format: candidates[0].content.parts[*].functionCall
-            candidates = response.get("candidates", [])
-            if candidates:
-                content = candidates[0].get("content", {})
-                parts = content.get("parts", [])
-                return [p for p in parts if "functionCall" in p]
+            candidates = response.get("candidates")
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
Thanks for tightening the outer response shapes here. I found one remaining malformed-response path in the OpenAI branch that still escapes the new guard:

```python
response = {"choices": [{"message": {"tool_calls": [None]}}]}
adapter._extract_tool_calls(response, "openai")  # returns [None]
adapter.has_memory_tool_calls(response, "openai")  # AttributeError: 'NoneType' object has no attribute 'get'
```

The new `_extract_openai_tool_calls` guards `choices[0]` and `message`, but it still returns `list(message.get("tool_calls", []) or [])` without filtering the tool-call elements themselves. Since the downstream paths (`has_memory_tool_calls`, `process_tool_responses`) immediately pass each entry into `_get_tool_name` / `_get_tool_input`, a null or string element in `tool_calls` still crashes the same untrusted upstream-response path this PR is hardening.

Please filter OpenAI `tool_calls` to dict entries, similar to the Anthropic/Gemini branches, and add a regression for `choices[0].message.tool_calls: [None, {valid call}]` proving the malformed entry is skipped and the valid one still parses.

Other checked state: PR is mergeable and visible CI is green. I did not run the full test file after finding this blocker; I verified the failing shape locally with the snippet above.

abhay-codes07:
Good catch, you're right that filtering the outer shapes wasn't enough while the elements themselves went straight to `.get` downstream. Pushed a fix.

The OpenAI branch now filters `message.tool_calls` to dict entries before returning, so `[None]` / `["oops"]` elements are dropped the same way the Anthropic and Gemini branches already skip non-dict elements. Added the regression you asked for: `tool_calls: [None, "oops", {valid memory_save}]` now returns only the valid call and `has_memory_tool_calls` still reports True.

Also rebased onto the latest main so it's clean against the recent merges.

JerrettDavis (approved):
Latest update fixes the remaining malformed OpenAI tool_calls path I flagged earlier. I verified the new filtering and regression locally with pytest tests/test_memory_tool_adapter_null_fields.py -q (8 passed) and ruff check headroom/proxy/memory_tool_adapter.py tests/test_memory_tool_adapter_null_fields.py (clean).

PR #2486, opened 2026-07-22
fix(proxy/gemini): tolerate malformed parts on the compression path

abhay-codes07:
## Description

Three helpers on the Gemini compression path read a content entry's `parts` and iterate it without type guards:

```python
# _has_non_text_parts
parts = content.get("parts", [])
for part in parts: ...
# _rebuild_gemini_contents
had_text = any("text" in p for p in content.get("parts", []))
# _gemini_contents_to_messages
parts = content.get("parts", [])
text_parts = [p.get("text", "") for p in parts if "text" in p]
```

`parts` is request-controlled and `.get("parts", [])` only falls back when the key is absent, so:

- a present-but-null `parts` returns `None`, and `for part in None` / `any(... for p in None)` raises `TypeError`;
- a list carrying a bare string (a client that treats `parts` as a string array) makes `p.get("text", "")` raise `AttributeError`, while `"text" in p` silently does substring matching first;
- a null element in the list crashes the same way.

Any of these 500s the request on the compression path, on data that parsed as valid JSON.

## Fix

Route all three helpers through a shared `_dict_parts(content)` that returns the dict entries of `parts`, coercing a non-dict content or a non-list `parts` to an empty list and dropping non-dict elements. `_gemini_contents_to_messages` also reads `role` defensively for a non-dict content entry. Conversion now degrades gracefully (the malformed part contributes nothing) instead of raising. Well-formed requests are unchanged.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/gemini.py`: add `_dict_parts`; use it in `_has_non_text_parts`, `_rebuild_gemini_contents`, and `_gemini_contents_to_messages`; read `role` defensively for a non-dict content entry.
- `tests/test_gemini_function_response_waste.py`: regressions for null `parts`, bare-string part elements, a null part element, `_has_non_text_parts` on malformed parts, and a non-dict content entry.

## Testing

### Test Output

```text
$ python -m pytest tests/test_gemini_function_response_waste.py -q
16 passed

# with the fix reverted, the new malformed-parts tests fail with
# TypeError: 'NoneType' object is not iterable  (and AttributeError on string parts)

$ uvx ruff@0.15.17 check headroom/proxy/handlers/gemini.py tests/test_gemini_function_response_waste.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/handlers/gemini.py
```

## Real Behavior Proof

- Exact command / steps: built a real `HeadroomProxy` and called `_gemini_contents_to_messages` / `_has_non_text_parts` with contents carrying `parts: null`, `parts: ["bare string", {text}]`, `parts: [null, {text}]`, and a non-dict content entry; then reverted `gemini.py` and re-ran.
- Observed result: with the fix each malformed shape converts without raising and the valid text part is still emitted (`[{"role": "user", "content": "kept"}]`); with the fix reverted the null-`parts` and null-element cases raise `TypeError: 'NoneType' object is not iterable` and the string-element case raises `AttributeError: 'str' object has no attribute 'get'`. Ran against the actual module via `tests/test_gemini_function_response_waste.py`.
- Not tested: a live Gemini request with malformed `parts` routed through the full proxy compression pipeline end to end.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/handlers/gemini.py
+++ b/headroom/proxy/handlers/gemini.py
@@ -56,6 +56,24 @@ class GeminiHandlerMixin:
             return ANTIGRAVITY_DAILY_API_URL
         return getattr(self, "CLOUDCODE_API_URL", DEFAULT_CLOUDCODE_API_URL).rstrip("/")
 
+    @staticmethod
+    def _dict_parts(content: Any) -> list[dict]:
+        """Return the dict entries of a Gemini content's ``parts``.
+
+        ``parts`` is request-controlled. ``.get("parts", [])`` only falls back
+        when the key is absent, so a present-but-null ``parts`` returns ``None``
+        (crashing ``for part in parts``), and a list carrying a bare string —
+        which a client that treats ``parts`` as a string array can send —
+        crashes ``part.get(...)`` / ``key in part`` semantics downstream.
+        Coerce to a clean list of dict parts so every caller can iterate safely.
+        """
+        if not isinstance(content, dict):
+            return []
+        parts = content.get("parts")
+        if not isinstance(parts, list):
+            return []
+        return [part for part in parts if isinstance(part, dict)]
+
     def _has_non_text_parts(self, content: dict) -> bool:
         """Check if a Gemini content entry has non-text parts.
 
@@ -73,8 +91,7 @@ class GeminiHandlerMixin:
         Returns:
             True if any part contains non-text data.
         """
-        parts = content.get("parts", [])
-        for part in parts:
+        for part in self._dict_parts(content):
             if any(
                 key in part
                 for key in (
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest diff and this looks ready. The shared `_dict_parts` helper is a narrow guard around the request-controlled `parts` field, and the existing preservation/rebuild flow still preserves non-text dict parts while malformed/null/string entries degrade to no contribution instead of crashing.

PR state is mergeable and visible checks are green. I also ran the focused regression suite locally on this branch:

```text
uv run --frozen --extra dev pytest tests/test_gemini_function_response_waste.py
16 passed
```

PR #2491, opened 2026-07-22
fix(proxy): allow settings routes for trusted gateway/dashboard clients

### LINKED ISSUE #2466: [BUG] Cannot open headroom settings when running in a container as a proxy
## Description

I'm getting 404 when i click the new settings button when running headroom as a proxy via a docker container

## To Reproduce

1. Install headroom with 'from a docker image'
2. Put it inside a network layer
3. Access headroom from your host machine via a reverse proxy gateway (not sure if that part is required)
4. you will get a 404 

## Expected Behavior

Let me set these and then honor these for the settings page like you do the stats and other pages:
      # The gateway container is the trusted local browser boundary for
      # Headroom's loopback-only dashboard administration routes.
      HEADROOM_PROXY_TRUSTED_GATEWAY_CIDRS: "172.29.0.0/16,10.89.0.0/16"
      HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS: "172.29.0.0/16,10.89.0.0/16"

## Actual Behavior

I can't access settings on my PC

Parideboy:
## Description

`/settings`, `/settings/schema`, `/settings/apply`, and `/dashboard/settings` were gated by `_require_loopback`, which checks `request.client.host` directly and 404s for any non-loopback caller. When headroom-proxy runs behind a reverse-proxy/gateway (e.g. in a container), `request.client.host` is the gateway's IP, so these routes 404 unconditionally — even with `HEADROOM_PROXY_TRUSTED_GATEWAY_CIDRS`/`HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS` configured, a trust chain `/stats` and `/stats-lifetime` already use.

Fixes #2466.

## Type of Change

- [x] Bug fix

## Changes Made

- Added `_require_loopback_or_trusted_dashboard_client` dependency in `headroom/proxy/server.py`, reusing the existing `_request_can_view_dashboard_metadata` trust chain (loopback check, IP-literal Host header check, same-origin check, trusted-gateway CIDR check).
- Swapped this dependency in for `_require_loopback` on exactly five routes: `/settings/schema`, `GET /settings`, `POST /settings`, `POST /settings/apply`, `/dashboard/settings`. All other loopback-only admin/debug routes (`/admin/*`, `/debug/*`, `/cache/clear`, `/v1/retrieve*`) are untouched.
- Added test coverage in `tests/test_proxy_loopback_gating.py`: non-loopback without trusted CIDR still 404s, loopback still allowed, trusted-gateway dashboard client is now allowed, and CIDR mismatch still 404s.

## Testing

- [x] Added/updated tests
- [x] Ran full test suite locally

```
$ python -m pytest tests/test_proxy_loopback_gating.py tests/test_proxy_settings_endpoints.py -q
73 passed, 1 warning in 28.80s

$ ruff check headroom/proxy/server.py tests/test_proxy_loopback_gating.py
All checks passed!

$ ruff format --check headroom/proxy/server.py tests/test_proxy_loopback_gating.py
1 file already formatted, 1 file already formatted

```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.11, pytest 9.1.1, headroom repo local checkout
- Exact command / steps: `python -m pytest tests/test_proxy_loopback_gating.py -q` after adding parametrized tests that set `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS` and hit `/settings`, `/settings/schema`, `/dashboard/settings` from a simulated gateway-forwarded peer IP
- Observed result: all 51 tests in the file pass, including new cases confirming trusted-gateway clients get 200 (previously 404) while unlisted/mismatched clients still get 404
- Not tested: did not manually deploy a real Docker container behind an actual reverse-proxy (e.g. nginx/Traefik) to reproduce the original reporter's exact setup; relied on TestClient-simulated forwarded headers/peer IPs instead

## Review Readiness

### DIFF
--- a/headroom/proxy/server.py
+++ b/headroom/proxy/server.py
@@ -3227,6 +3227,30 @@ def create_app(config: ProxyConfig | None = None) -> FastAPI:
         if not _request_can_view_dashboard_metadata(request, trusted_dashboard_client_cidrs):
             raise HTTPException(status_code=404)
 
+    def _require_same_origin_or_trusted_dashboard_client(request: Request) -> None:
+        """Same-origin CSRF guard for settings writes, trusted-dashboard aware.
+
+        ``require_same_origin`` only accepts an ``Origin`` that itself names a
+        loopback host, so a browser POST from a trusted-gateway dashboard
+        client was rejected even though the paired GET routes allow that same
+        caller (issue #2466). For non-loopback callers, accept an ``Origin``
+        that matches this request's own Host header, provided the caller is
+        already an IP-literal-Host, CIDR-trusted dashboard client. Loopback
+        callers keep the stricter loopback-only origin check unchanged.
+        """
+        if not _request_is_loopback(request):
+            origin = request.headers.get("origin")
+            host_header = request.headers.get("host")
+            if (
+                origin
+                and origin != "null"
+                and host_header
+                and _request_has_same_origin_or_no_provenance(request, host_header)
+                and _request_can_view_dashboard_metadata(request, trusted_dashboard_client_cidrs)
+            ):
+                return
+        _require_same_origin(request)
+
     @app.get("/admin/upstream", dependencies=[Depends(_require_loopback)])
     async def get_upstream():
         """Current Anthropic upstream + cc-switch reconciler state (loopback-only).
@@ -3354,7 +3378,7 @@ def create_app(config: ProxyConfig | None = None) -> FastAPI:
         "/settings",
         dependencies=[
             Depends(_require_loopback_or_trusted_dashboard_client),
-            Depends(_require_same_origin),
+            Depends(_require_same_origin_or_trusted_dashboard_client),
         ],
     )
     async def settings_post(request: Request):
@@ -3404,7 +3428,7 @@ def create_app(config: ProxyConfig | None = None) -> FastAPI:
         "/settings/apply",
         dependencies=[
             Depends(_require_loopback_or_trusted_dashboard_client),
-            Depends(_require_same_origin),
+            Depends(_require_same_origin_or_trusted_dashboard_client),
         ],
     )
     async def settings_apply(request: Request):
--- a/headroom/proxy/server.py
+++ b/headroom/proxy/server.py
@@ -3217,6 +3217,16 @@ def create_app(config: ProxyConfig | None = None) -> FastAPI:
     from headroom.proxy.loopback_guard import require_loopback as _require_loopback
     from headroom.proxy.loopback_guard import require_same_origin as _require_same_origin
 
+    def _require_loopback_or_trusted_dashboard_client(request: Request) -> None:
+        """Allow loopback callers, or gateway-forwarded dashboard clients.
+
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
This is close, but the mutating settings routes are still broken for the browser/dashboard case the PR is trying to enable.

The new `_require_loopback_or_trusted_dashboard_client` dependency lets a trusted non-loopback dashboard client reach `POST /settings` and `POST /settings/apply`, but those routes still also depend on `require_same_origin`. That helper only accepts loopback origins, so a real same-origin dashboard POST from the trusted remote/IP-literal dashboard host is rejected:

```python
client = TestClient(
    app,
    base_url="http://100.82.0.2:8787",
    client=("100.90.0.5", 12345),
)
client.post("/settings", json={"values": {}}, headers={"Origin": "http://100.82.0.2:8787"})
# 403 {"detail":"cross-origin request rejected"}
```

I verified locally on this branch that the same trusted client gets `200` with no `Origin` header, but gets `403` with the same-origin `Origin` header above. Browsers commonly include `Origin` on POST/fetch requests, so the settings UI can still fail behind the trusted gateway even though the GET routes work.

Please make the CSRF/provenance check for these settings writes match the same trusted-dashboard origin policy used by `_request_can_view_dashboard_metadata` (while still rejecting foreign origins), and add regression coverage for a trusted dashboard client posting with `Origin: http://<dashboard-host>:<port>` to both `/settings` and `/settings/apply`.

Other checked state: PR is mergeable and visible CI is green. I did not run the full file after finding this blocker; the reproduction above is from the latest checked-out source.

Parideboy:
@JerrettDavis Good catch, fixed in 2d14a8aa.

Added `_require_same_origin_or_trusted_dashboard_client`: for non-loopback callers it accepts an `Origin` that matches the request's own `Host` header, provided the caller is already an IP-literal-Host, CIDR-trusted dashboard client (reusing `_request_can_view_dashboard_metadata`). Loopback callers are untouched, they keep the strict loopback-only origin check (including the `Origin: null` rejection).

Swapped this in for `_require_same_origin` on `POST /settings` and `POST /settings/apply`.

Added regression coverage in `tests/test_proxy_loopback_gating.py`:
- trusted client + matching `Origin` → no longer 403 on either route
- trusted client + foreign `Origin` → still rejected (403/404)
- loopback caller + `Origin: null` → still 403 (confirms the carve-out didn't weaken the loopback path)

`pytest tests/test_proxy_loopback_gating.py tests/test_proxy_settings_endpoints.py` 78 passed. `ruff check`/`ruff format --check` clean, `mypy headroom --ignore-missing-imports` clean.

JerrettDavis (approved):
Latest changes look good. I re-ran the targeted settings/loopback gating suite locally (python -m pytest tests/test_proxy_loopback_gating.py tests/test_proxy_settings_endpoints.py -q: 78 passed) and reviewed the trusted-dashboard same-origin path. The settings routes now match the existing dashboard metadata trust model without widening the unrelated admin/debug endpoints.

PR #2496, opened 2026-07-22
fix(mcp): resolve opencode.jsonc for MCP registration

### LINKED ISSUE #1869: [BUG] Headroom wrap opencode not saving any token
## Description

Using headroom wrap opencode doesn't yield any saving in token, localhost:8787/dashboard remain 0 all over. 

## To Reproduce

1. Install headroom with 'uv tool install "headroom-ai[proxy]"'
2. Run this code 'headroom wrap opencode'
3. No headroom models in selection, after using opencode for sometime, 0 saving in dashboard

## Expected Behavior
I assume there should be headroom in model selection
Saving dashboard should show saving > 0 after significant use of opencode

## Actual Behavior

No headroom models in selection, after using opencode for sometime, 0 saving in dashboard

## Environment

- **Headroom version**:  0.30.0
- **Python version**: 3.12.13
- **OS**: Mac OS Tahoe 26.5.2
- **LLM Provider**: opencode cli 1.17.15

## Additional Context

After running headroom wrap opencode, in opencode CLI, headroom is registered as connected MCP.

Parideboy:
## Description
`headroom wrap opencode` shows no headroom models and 0 savings on setups where OpenCode's config lives in `opencode.jsonc` instead of `opencode.json`. This is because `OpencodeRegistrar` (which writes the `mcp.headroom` block) had its own config-path resolver that always targets `opencode.json`, while the provider-block writer (`providers/opencode/config.py`) already correctly prefers an existing `opencode.jsonc`. On `.jsonc` setups this split the provider block and the MCP registration block across two different files, and OpenCode only reads one of them.

## Type of Change
- [x] Bug fix

## Changes Made
- `headroom/install/paths.py`: `opencode_config_path()` now also honors `OPENCODE_HOME` for the base directory (matching the two duplicated resolvers it replaces), in addition to its existing `OPENCODE_CONFIG` and `.jsonc`-over-`.json` preference.
- `headroom/mcp_registry/opencode.py`: removed the local `_opencode_config_path()`/`_opencode_home_dir()` resolvers; `OpencodeRegistrar` now uses the shared `install.paths.opencode_config_path()`, so it resolves to the same file as the provider-block writer. Updated stale docstrings/error text that assumed `opencode.json` was the only possible file.
- Added regression tests in `tests/test_install/test_paths.py` (OPENCODE_HOME handling) and `tests/test_mcp_registry_opencode.py` (registrar resolves `opencode.jsonc` when present, with no explicit `config_path`).

## Testing
- [x] Added/updated tests
- [x] Ran full local test suite

```
$ python -m pytest tests/test_mcp_registry_opencode.py tests/test_install/test_paths.py tests/test_providers_opencode_config.py tests/test_providers_opencode_install.py -q
...
tests\test_mcp_registry_opencode.py .......................................... [ 43%]
tests\test_install\test_paths.py .....                                        [ 49%]
tests\test_providers_opencode_config.py ................................F.... [ 94%]
tests\test_providers_opencode_install.py .....                                [100%]
1 failed, 90 passed in 8.44s
```
The one failure (`test_build_launch_env_with_project`) is a pre-existing Windows path-escaping assertion unrelated to this change — confirmed it fails identically on `main` before this patch.

```
$ ruff check .
All checks passed!
$ ruff format --check .
1334 files already formatted
Success: no issues found in ... source files
```

## Real Behavior Proof
- Environment: Windows 11, Python 3.13.11, local checkout of headroomlabs-ai/headroom
- Exact command / steps: Set `OPENCODE_HOME` to a temp dir containing only `opencode.jsonc`; construct `OpencodeRegistrar()` with no explicit `config_path` (see new test `test_default_config_path_prefers_existing_jsonc`); previously this resolved to a stray `opencode.json` in the same directory, now it resolves to the existing `opencode.jsonc`.
- Observed result: `registrar._config_path == tmp_path / "opencode.jsonc"` (test passes); before the fix it asserted `opencode.json` and would have registered the MCP server in the wrong file.
- Not tested: did not manually exercise the full `headroom wrap opencode` CLI flow end-to-end against a real OpenCode installation.

## Review Readiness

Fixes #1869

### DIFF
--- a/headroom/install/paths.py
+++ b/headroom/install/paths.py
@@ -125,14 +125,16 @@ def opencode_config_path() -> Path:
     """Return the OpenCode config path.
 
     Resolves ``~/.config/opencode/opencode.json`` when ``OPENCODE_CONFIG``
-    is unset; otherwise the value of that environment variable. Checks for
+    is unset; otherwise the value of that environment variable. Honors
+    ``OPENCODE_HOME`` for the base directory, and checks for
     ``opencode.jsonc`` as well.
     """
 
     env_path = os.environ.get("OPENCODE_CONFIG", "").strip()
     if env_path:
         return Path(env_path).expanduser()
-    base_dir = Path.home() / ".config" / "opencode"
+    home_path = os.environ.get("OPENCODE_HOME", "").strip()
+    base_dir = Path(home_path).expanduser() if home_path else Path.home() / ".config" / "opencode"
     jsonc_path = base_dir / "opencode.jsonc"
 
     if jsonc_path.exists():
--- a/headroom/mcp_registry/opencode.py
+++ b/headroom/mcp_registry/opencode.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import json
 import logging
+import re
 import shutil
 from pathlib import Path
 from typing import Any
@@ -20,8 +21,32 @@ from .base import MCPRegistrar, RegisterResult, RegisterStatus, ServerSpec
 logger = logging.getLogger(__name__)
 
 
+def _strip_json_line_comments(text: str) -> str:
+    """Strip ``//`` line comments from JSONC text.
+
+    Tries standard JSON first (via the caller) so URLs containing ``//`` are
+    never mangled; this is only invoked as a fallback once plain
+    ``json.loads`` has already failed. Two-pass: (1) remove comment-only
+    lines, (2) strip inline trailing comments that follow a comma.
+    """
+    cleaned = re.sub(r"^\s*//[^\n]*\n", "", text, flags=re.MULTILINE)
+    return re.sub(r",\s*//[^\n]*", ",", cleaned)
+
+
+def _parse_json_loose(raw: str) -> dict[str, Any] | None:
+    """Parse JSON or JSONC text, returning ``None`` if it can't be parsed as an object."""
+    try:
+        data = json.loads(raw)
+    except json.JSONDecodeError:
+        try:
+            data = json.loads(_strip_json_line_comments(raw))
+        except json.JSONDecodeError:
+            return None
+    return data if isinstance(data, dict) else None
+
+
 def _read_json(path: Path) -> dict[str, Any]:
-    """Read a JSON file, returning empty dict if absent or unparseable.
+    """Read a JSON/JSONC file, returning empty dict if absent or unparseable.
 
     Safe for READ-ONLY callers only. Do NOT use before a full-file rewrite:
     an unparseable existing file returns ``{}`` here, and writing that back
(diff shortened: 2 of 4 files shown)


JerrettDavis (changes requested):
This is close, but I think the fix needs to handle actual JSONC before it lands.

The registrar now correctly resolves to an existing `opencode.jsonc`, but `_read_json_for_write()` still uses strict `json.loads()`. A common OpenCode `opencode.jsonc` can contain comments, so `register_server()` will now pick the right file and then fail with "not valid JSON" instead of adding the MCP block. The new regression test only covers an empty `{}` JSONC file, so it misses the practical case that motivated preferring `.jsonc`.

Please either reuse the same loose JSON/JSONC parsing path used by the OpenCode provider config writer, or add equivalent JSONC-safe parsing here, and add a regression test where `OPENCODE_HOME` contains an `opencode.jsonc` with a comment plus existing user settings. The expected behavior should be that the MCP block is merged without dropping the existing settings.

Parideboy:
Good catch, fixed in cb97170b.

\`_read_json_for_write()\` (and \`_read_json()\`) were strict \`json.loads()\`, so once the registrar started resolving to a real \`opencode.jsonc\` containing \`//\` comments, it would correctly find the file and then fail with "not valid JSON" instead of merging the mcp block in.

Reused the same loose JSON/JSONC parsing strategy as the provider-block writer in \`providers/opencode/config.py\` (strict parse first, so URLs with \`//\` aren't mangled. Falling back to stripping \`//\` line/trailing comments) on both the read and read-for-write paths.

Added \`test_register_server_merges_into_commented_jsonc\`: \`OPENCODE_HOME\` points at an \`opencode.jsonc\` with a comment plus existing \`theme\`/\`model\`/\`provider\` settings, and asserts \`register_server()\` merges the \`mcp\` block in while preserving all pre-existing keys.

Re-ran \`pytest tests/test_mcp_registry_opencode.py tests/test_install/test_paths.py\` (46 passed) plus \`ruff check\`, \`ruff format --check\`, \`mypy --ignore-missing-imports\` on the changed files, all clean.

Note: like the provider-block writer, the rewrite drops the file's comments (JSON has no comment syntax to write back), only content/keys are preserved. Flagging in case that's worth a follow-up, but out of scope for this fix.

JerrettDavis (approved):
Re-reviewed cb97170b after the JSONC follow-up. The MCP registrar now shares the active OpenCode config path, honors OPENCODE_HOME/OPENCODE_CONFIG precedence, and parses commented opencode.jsonc files before merging the MCP block while preserving existing settings. Verified locally: pytest tests/test_mcp_registry_opencode.py tests/test_install/test_paths.py -q (46 passed); ruff check touched files.

Issue #746, opened 2026-06-08
Routing Claude Code through the proxy disables on-demand tool loading, inflating local context by ~25K tokens


arielZusman:
### Summary

When Claude Code is pointed at the Headroom proxy via `ANTHROPIC_BASE_URL`, Claude Code stops deferring tool schemas (its normal "load tools on demand" behavior) and eagerly materializes **all** system + MCP tool definitions into the context window. On a setup with several MCP servers this adds ~25K tokens to the *local* context baseline **before any message is sent** — i.e. the opposite of what a context-optimization proxy is expected to do.

This appears to be triggered simply by `ANTHROPIC_BASE_URL` pointing at a non-`api.anthropic.com` endpoint, so it likely affects every proxy user; the impact scales with how many tools/MCP servers the user has.

### Environment

- headroom-ai **0.23.0** (installed via `pipx install --python python3.12 "headroom-ai[proxy,code]"`)
- Python 3.12.0, macOS (Darwin 25.x), Apple Silicon
- Client: Claude Code (latest)
- Proxy started with `headroom proxy --port 8787` (default `--mode token`)

### Reproduction

Reproduces with **both** the manual proxy path and the wrap path with all injection flags disabled:

```sh
# A) manual
headroom proxy --port 8787
ANTHROPIC_BASE_URL=http://localhost:8787 claude

# B) wrap, all injections off
headroom wrap claude --no-mcp --no-serena --no-context-tool
```

In each case, run `/context all` immediately, before sending any message, and compare to a baseline `claude` session with `ANTHROPIC_BASE_URL` unset.

### Evidence (`/context all`, fresh session, no messages)

**Baseline — no Headroom (`ANTHROPIC_BASE_URL` unset): 34.6k / 200k**

```
System prompt:  6.4k
System tools:   14k          <-- deferred subset
MCP tools:      (loaded on-demand)   <-- 0 tokens, names only
Custom agents:  3k
Memory files:   5k
Skills:         5k
Messages:       1.2k
Total used:     34.6k (17%)
```
MCP section header reads: `MCP tools · /mcp (loaded on-demand)` and individual tools show **no** token cost.

**Through Headroom proxy (`--no-mcp --no-serena --no-context-tool`): 59.3k / 200k**

```
System prompt:  6.4k         (unchanged)
System tools:   29.8k        <-- +15.8k, fully materialized
MCP tools:      8.9k         <-- +8.9k, eagerly loaded
Custom agents:  3k           (unchanged)
[...]

PR #2500, opened 2026-07-23
fix(proxy/tool-search): don't defer when the client already defers; recover dangling tool_reference

nangsontay:
## Description

Through the Headroom proxy, a request whose transcript carries a `tool_reference` for a tool no longer present in the request's `tools` array 400s with `Tool reference 'X' not found in available tools`. Anthropic validates every `tool_reference` per turn against `tools`, and the transcript replays those blocks forever.

This started as a narrow `/model`-switch fix for `headroom_retrieve` and grew after review: the failure is **not** specific to the proxy-injected tool. Then live use surfaced a third, unrelated failure in the same subsystem — the proxy deferring tools a Claude Code client had *already* deferred client-side. Three distinct causes, all fixed here.

**1. The transcript scan matched the wrong wire shape.** A `tool_reference` keys the referenced tool as **`tool_name`**, not `name` — that is the shape Anthropic returns inside `tool_search_tool_result` and the shape the custom client-side tool-search contract specifies. `name` appears only on `tool_use` and on the mid-conversation `tool_addition`/`tool_removal` wrapper. The nesting was wrong too: `tool_search_tool_result` nests references in a `tool_search_tool_search_result` **dict** under `tool_references`, not a list, so the walk skipped that shape before the field name even mattered. The recovery path therefore missed both shapes that occur in production and matched only a synthetic one — it was inert against the exact 400 it targeted.

**2. Every deferred tool is exposed, not just `headroom_retrieve`.** `inject_tool_search_deferral` stamps `defer_loading: true` on every non-core tool and injects a search tool, so each tool the model discovers leaves a `tool_reference` behind. A tool then leaves `tools` for reasons the proxy does not control — an MCP server disconnects, the client trims its surface for a subagent turn. The deferral was also turn-unstable: below the tool-count threshold it returned the tools untouched, so the search tool and every deferred definition vanished mid-session while the transcript still named them.

**3. The proxy deferred on top of a client that already defers.** Claude Code with `ENABLE_TOOL_SEARCH` defers **client-side**: the deferred MCP schemas stay in the client's process, never enter the request body, and the model loads them through the client-side `ToolSearch` function tool. `inject_tool_search_deferral` was gated on provider + `HEADROOM_TOOL_SEARCH` but never on the client, so it stacked Anthropic's **server-side** `tool_search_tool_*` on top. That index can only see what is in the request body — precisely not the tools the model is hunting for. Observed symptom: the model calls `tool_search_tool_regex`, gets zero matches for `ctx_read`/`ctx_search`, and reports "the lean-ctx MCP server isn't responding" while `/mcp` shows it connected. The reported `tool_saved` on those turns was illusory: the deferred bytes were the client's small resident core set, and hiding them bought a search round-trip per edit. Compounding it, `_TOOL_SEARCH_CORE_TOOLS` is spelled in opencode's snake_case and was compared exactly, so a harness naming the same tools `Read`/`Edit`/`Bash` had its whole edit/read/run loop deferred — including `ToolSearch` itself, leaving the model to search for the tool it searches with.

Root cause of the original `/model` symptom is unchanged: sticky CCR injection is keyed by a model-scoped, in-memory session id (`compute_session_id` hashes `model + leading system text`), which a switch or restart rotates. Key-tweaking cannot fix it, so recovery is transcript-driven. Cause 3 is *not* a `/model` artifact despite first being reported right after a switch — the switch was a coincidence.

Plan: `plans/260723-1004-ccr-dangling-tool-reference-reinjection/`.

Closes # 

## Type of Change

- [x] Documentation update
- [x] Performance improvement

### CCR-specific recovery (original scope)

- `headroom/proxy/ccr_marker_policy.py` — new `transcript_references_ccr_tool()` scanning about-to-forward messages for a bare `headroom_retrieve`; extended `should_inject_ccr_tool()` with a `transcript_requires_tool` override. **Match is now split by block type**: `tool_reference` accepts `tool_name` or `name`, `tool_use` keeps `name`, and a dict `content` is unwrapped via `tool_references`.
- `headroom/proxy/tool_injection_logging.py` — added `inject_transcript_recovery` to the closed `ToolInjectionDecision` literal.
- `headroom/proxy/helpers.py` — re-export the policy fn; thread `transcript_requires_tool` through `should_inject_ccr_tool` and `apply_session_sticky_ccr_tool` (both the no-session-id and fresh-session paths inject on the transcript signal, log `inject_transcript_recovery`, and record sticky state so later turns resume normal sticky replay).
- `headroom/proxy/handlers/anthropic.py` / `openai.py` — compute the flag over `optimized_messages`, pass it through, distinct recovery log branch.
- `docs/content/docs/troubleshooting.mdx` — troubleshooting entry (symptom, cause, fix, workarounds for older versions).

Exact bare-name match only, so a client-owned `mcp__headroom__headroom_retrieve` never triggers proxy injection. The openai `/v1/responses` path never proxy-injects the CCR tool, so recovery is N/A there (plan decision D1).

### General recovery for any deferred tool (added after review)

- `headroom/proxy/tool_search_recovery.py` — **new module.** Per-session cache of the definitions this proxy deferred, a generic transcript scan (`collect_tool_reference_names`) covering all four carriers, and `reconcile_tool_references()` with two tiers:
  1. **Re-inject** from the cache — whatever we deferred we have already seen in full, so it is restored verbatim and still deferred (no context cost).
  2. **Sanitize** when there is nothing to restore (a client-owned reference we never saw). The orphan is dropped from the transcript so the request is valid; the model may search again, which beats a hard 400. `tool_addition`/`tool_removal` are deliberately **left alone** — they are explicit client instructions and rewriting them would change request semantics.

  Copy-on-write: inputs are never mutated, and identity-unchanged results let the caller skip re-serialization. LRU-capped at 256 sessions × 2048 tools.
- `headroom/proxy/helpers.py` — `inject_tool_search_deferral` takes an optional `session_id`. **Sticky deferral**: once a session has deferred, the tool-count threshold stops applying, and the search tool stays declared even on a turn with nothing to defer (the transcript's `server_tool_use` blocks still name it). Deferred definitions are recorded on the success path.
- `headroom/proxy/handlers/anthropic.py` — reconciliation runs after turn hooks, once outbound `tools`/`messages` are final; best-effort so a bug there can never block forwarding.

### Don't defer when the client already defers (added after live use)

- `headroom/proxy/helpers.py` — `inject_tool_search_deferral` now no-ops when the request carries a plain function tool named `ToolSearch`/`tool_search` (`_CLIENT_SIDE_TOOL_SEARCH_NAMES`), alongside the existing `tool_search_tool_*` typed-tool guard. The check runs before the sticky path, so a session that was already double-deferred stops on the next turn.
- `headroom/proxy/helpers.py` — core tool names are compared case-insensitively (`core_lower`), so PascalCase harnesses keep `Read`/`Write`/`Edit`/`Bash`/`Glob`/`Grep`/`Task`/`WebFetch` resident.

Gating on **request shape, not User-Agent**, is deliberate and consistent with `claude_code_tool_search_inactive` above it: a claude-code session *without* `ENABLE_TOOL_SEARCH` (#746) sends no `ToolSearch` tool, has every schema in its local context anyway, and should still get server-side deferral — that is the case this feature exists for. A UA gate would have disabled it there too.

## Testing

- [x] Linting passes (`ruff check`) — on the touched files
- [ ] Type checking passes (`mypy headroom`) — not run this pass
- [x] Manual testing performed — cause 3 was diagnosed from a live proxy log; the 400 in causes 1–2 was still reproduced only via unit + stubbed-upstream tests

36 new tests: 15 CCR policy (original) + 16 recovery unit + 3 end-to-end through the real Anthropic handler asserting the **actual forwarded body** + 2 for the client-side-deferral guard and case-insensitive core matching.

### Test Output

```text
$ uv run pytest tests/test_tool_search_reference_recovery.py \
    tests/test_proxy/test_anthropic_tool_reference_recovery.py \
    tests/test_ccr_tool_always_on.py tests/test_issue_746_tool_search.py \
    tests/test_openai_tool_search_deferral.py tests/test_proxy/ -q
======================= 346 passed, 1 warning in 21.65s ========================

$ uv run pytest tests/ -q -k "tool_search or deferral or defer_loading"
========= 108 passed, 8 skipped, 10212 deselected, 1 warning in 7.29s ==========

$ uv run ruff check headroom/proxy/ccr_marker_policy.py headroom/proxy/tool_search_recovery.py \
    headroom/proxy/helpers.py headroom/proxy/handlers/anthropic.py \
    tests/test_ccr_tool_always_on.py tests/test_tool_search_reference_recovery.py \
    tests/test_proxy/test_anthropic_tool_reference_recovery.py \
    tests/test_issue_746_tool_search.py
All checks passed!
```

The 346-test line predates the cause-3 commit; the 108-test line is the post-fix run over every tool-search/deferral test in the suite.

## Real Behavior Proof

- Environment: local, macOS (darwin) 25.4.0, Python 3.13.14, pytest 9.0.3 via `uv run`, ruff via `uv run ruff`. No live upstream call for causes 1–2 — the Anthropic upstream is stubbed at `proxy._retry_request` so the assertions read the exact body the proxy would have forwarded. Cause 3 was observed on live traffic (Claude Code → local proxy → first-party Anthropic).
- Exact command / steps: (1) the `pytest` / `ruff` commands in **Test Output** above; (2) wire shape confirmed against the live API reference at `platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool.md`; (3) old-vs-new predicate compared in-process on both production block shapes; (4) turn-instability reproduced by calling `inject_tool_search_deferral` with 14 then 11 tools; (5) counterfactual run with `reconcile_tool_references` monkeypatched to a no-op, asserting the forwarded body regresses; (6) pre-existing-failure baseline established by `git stash`ing this change and rerunning the identical `-k "proxy or tool or ccr or helper or handler or cache"` selection at HEAD; (7) **cause 3**: reproduced in a live Claude Code session, root-caused from `~/.headroom/logs/proxy.log`, and confirmed against the session's own tool surface.
- Observed result: 346/346 pass in the scoped suite, 108 passed / 8 skipped across every tool-search test after the cause-3 fix, ruff clean. The documented API error string is verbatim this bug's string. Both production shapes returned `False` before the fix and `True` after. The 6→3 MCP shrink drops the search tool and orphans `mcp__linear__op4`, which the CCR-only scan does not recover. With the hook disabled the forwarded body regresses exactly as predicted, so the e2e tests are real guards. Baseline at HEAD fails the same 4 `tests/test_proxy_health.py` kompress test IDs — `4 failed, 3139 passed` at HEAD vs `4 failed, 3160 passed` with this change; that file passes 17/17 in isolation both with and without the change, so those are order-dependent flakes, unrelated. Verified statically (AST) that `session_id`, `body_mutation_tracker` and `request_id` are all unconditionally bound before both new call sites in `handle_anthropic_messages`.
- For cause 3 the live evidence was unambiguous: every PERF line for the affected `client=claude-code` session carried a `router:tool_search_deferral:18tools:14620tok` tag, and in that session `tool_search_tool_regex` was the model's **only** resident tool — `ToolSearch`, `Read`, `Edit` and `Bash` had all been deferred by the proxy, so `ToolSearch` itself had to be loaded through a search. A regex search for the client-deferred `ctx_*` tools returned no matches (they were never in the body), while the client-side `ToolSearch` resolved them immediately — exactly the split the fix removes.
- Not tested: no live `/model` switch against the real Anthropic API (causes 1–2 reproduced only as unit + stubbed-upstream tests, so that 400 was never observed end-to-end from upstream); the cause-3 fix has **not** yet been exercised against live traffic — the proxy must be restarted to pick up `helpers.py`, and the post-restart session was not run before pushing; `mypy headroom` not run; the OpenAI chat path carries only the original CCR wiring — the general recovery is Anthropic-only and untested on OpenAI, and `inject_tool_search_deferral_openai` did **not** receive the client-side-deferral guard (no known OpenAI-side harness that defers client-side); the sticky-deferral cache-hit-rate impact is reasoned about but not measured; the LRU caps (256 sessions × 2048 tools) are not load-tested; concurrent same-session requests mutating the deferred-definition cache are not exercised beyond the module's lock.

### Supporting output

Wire shape returned by `tool_search_tool_result`, per the live API reference:

```json
{"type": "tool_search_tool_result", "tool_use_id": "srvtoolu_01ABC123",
 "content": {"type": "tool_search_tool_search_result",
             "tool_references": [{"type": "tool_reference", "tool_name": "get_weather"}]}}
```

Documented error string — verbatim the one in this bug: `Tool reference 'unknown_tool' not found in available tools`.

Old vs new predicate on both production shapes:

```text
OLD server-side tool_search_tool_result: False   NEW: True
OLD custom client-side tool_reference  : False   NEW: True
```

General failure reproduced — Linear MCP server shrinking 6 → 3 tools between turns:

```text
turn N   tools=14 -> search_tool=1 deferred=[6 mcp tools]
turn N+1 tools=11 -> search_tool=0 unchanged=True
op4 still in tools?                        False
transcript scan recovers mcp__linear__op4? False   (CCR scan is headroom_retrieve-only)
```

Counterfactual with the recovery hook disabled:

```text
HOOK DISABLED -> op4 in turn2 tools: False (expected False)
HOOK DISABLED -> ghost refs forwarded: ['ghost_tool'] (expected ['ghost_tool'])
```

Cause 3 on live traffic — the proxy double-deferring for a client that already defers:

```text
2026-07-28 14:19:02 PERF model=claude-opus-5 ... tool_saved=14620
  transforms=... router:tool_search_deferral:18tools:14620tok ... client=claude-code
```

## Review Readiness

## Checklist

N/A — proxy/backend change, no UI surface.

## Additional Notes

**Cache impact — please review this tradeoff.** Two effects:

- The CCR recovery injection changes the tools prefix once on the first post-switch turn (same accepted cost as the existing #1006 marker-override — one cache miss beats a hard 400 loop). It does not re-inject per turn once sticky state is re-recorded.
- Sticky deferral keeps the search tool declared on turns that previously had none injected. It is stable **within** a session once sticky — which is what cache-hit rate actually depends on — but a session's first sub-threshold turn now writes a new prefix instead of reusing the client's. I judged a stable prefix plus no 400 better than a search tool that flip-flops, though that is a cost worth measuring.

**Cause 3 removes a reported saving.** Claude Code sessions that defer client-side will stop reporting `tool_saved` from `router:tool_search_deferral` (~14k tokens on the session above). That number was not a real saving: the deferred bytes were the client's small resident core set, they were already excluded from the client's own context accounting, and hiding them cost a search round-trip per edit plus the tool-discovery break this PR fixes. Sessions that genuinely benefit (opencode, plain API clients, claude-code without `ENABLE_TOOL_SEARCH`) are unaffected.

**Mid-flight sessions.** For a session already double-deferred when the new build lands, the guard drops the injected search tool while the transcript may still carry `server_tool_use` blocks naming it. `reconcile_tool_references` covers `tool_reference`, not `server_tool_use`, so such a session may need a restart. This matches the pre-existing behaviour of the `tool_search_tool_*` typed-tool guard next to it and only affects the one upgrade boundary; not worth special-casing.

**Known gap, deliberate.** An orphaned `tool_addition`/`tool_removal` still 400s. Those are explicit client instructions the proxy never generates, so a reference there means the client declared a tool it did not send. Collection reports it (tier 1 can still restore it if we deferred that tool) but sanitization will not rewrite it. Documented in the module and covered by a test.

**`headroom_retrieve` is intentionally still not in `_TOOL_SEARCH_CORE_TOOLS`.** It no longer needs to be: when deferred it is cached like any other tool, so the generic path recovers it too, and the CCR-specific path still covers the non-deferred case. Adding it would move the tools cache prefix for a benefit the generic path already provides.

## Unresolved Questions

- Should `inject_tool_search_deferral_openai` get the same client-side-deferral guard preemptively, or wait until a harness that defers client-side over the Responses API shows up?
- Is `_CLIENT_SIDE_TOOL_SEARCH_NAMES` (`ToolSearch`, `tool_search`) the complete set of client-side search tool names worth matching, or should it be configurable?


JerrettDavis (changes requested):
Thanks for the fix direction here. I ran the focused regression suite and lint locally:

- `pytest tests/test_ccr_tool_always_on.py -q` -> 28 passed
- `ruff check headroom/proxy/ccr_marker_policy.py headroom/proxy/helpers.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py tests/test_ccr_tool_always_on.py` -> passed

I think this still needs one correctness fix before merge. `_anthropic_content_references_tool()` currently matches both `tool_reference` and `tool_use` blocks with `block.get("name") == tool_name`. `tool_use` uses `name`, but the Anthropic/server-tool-reference shape used elsewhere in this codebase carries the reference as `tool_name` on `tool_reference` blocks. That means the recovery path added here will miss the stale `tool_reference.tool_name == "headroom_retrieve"` transcript shape that causes the upstream 400 this PR is trying to self-heal.

Please split the match by block type, or otherwise accept `tool_name` for `tool_reference` while keeping the exact bare-name guard. It would also be worth adding a regression with the actual nested tool-result/reference shape, e.g. a `tool_search_tool_result` / `content.tool_references` entry containing `{ "type": "tool_reference", "tool_name": "headroom_retrieve" }`, so the scanner covers the failing production transcript shape instead of only the synthetic `{ "type": "tool_reference", "name": ... }` case.

JerrettDavis (changes requested):
Thanks for updating this. The same blocker is still present in the latest commit `ed46646a`.

`headroom/proxy/ccr_marker_policy.py:106` still treats `tool_reference` and `tool_use` as if they both carry the tool name in `name`:

```python
block.get("type") in ("tool_reference", "tool_use")
and block.get("name") == tool_name
```

That is correct for `tool_use`, but the deferred Anthropic tool-reference shape this recovery path is meant to heal uses `tool_name` on `tool_reference` blocks. The new tests still build `{"type": "tool_reference", "name": CCR_TOOL_NAME}`, so they pass without covering the production shape that leaves `headroom_retrieve` dangling after a `/model` switch or proxy restart.

Please split the match by block type, or accept `tool_name` for `tool_reference` while keeping the exact bare-name guard. Also add a regression with the actual nested deferred-tool shape, for example a `tool_search_tool_result` or `tool_result` content entry containing:

```python
{"type": "tool_reference", "tool_name": "headroom_retrieve"}
```

I did not run the local suite for this revision because the source still misses the target shape, so the existing green checks do not prove the reported failure is fixed yet.

JerrettDavis (changes requested):
The original blocker is fixed now: the CCR scanner matches `tool_reference.tool_name`, and the nested `tool_search_tool_result.content.tool_references` shape is covered. The focused suite also passes locally.

One remaining correctness issue in the new generic recovery path: `collect_tool_reference_names()` collects top-level `tool_reference` blocks, but `_sanitize_content_block()` only removes orphan references nested under `tool_search_tool_result.content.tool_references` or `tool_result.content`. If a top-level reference is missing and not recoverable from the deferred-tool cache, `reconcile_tool_references()` returns unchanged and still forwards the orphan, so Anthropic can still 400 on the exact class of invariant this module is meant to enforce.

Concrete reproduction on current head `c4315fd0`:

```python
from headroom.proxy.tool_search_recovery import reconcile_tool_references, reset_state

reset_state()
messages = [{"role": "assistant", "content": [{"type": "tool_reference", "tool_name": "ghost_tool"}]}]
result = reconcile_tool_references(messages, [], session_id="s")

assert result.changed is False
assert result.sanitized == []
assert result.messages == messages  # ghost_tool is still forwarded
```

Since the collector intentionally recognizes this carrier, reconciliation should either sanitize top-level orphaned `tool_reference` blocks as well, or stop collecting them if they are deliberately out of scope. Given the module doc says orphaned references are dropped when there is nothing to restore, I think the safer fix is to drop top-level orphan references and add a regression for that case.

What I ran locally:
- `uv run --extra dev python -m pytest tests/test_tool_search_reference_recovery.py tests/test_proxy/test_anthropic_tool_reference_recovery.py tests/test_ccr_tool_always_on.py tests/test_issue_746_tool_search.py -q` -> 87 passed, 1 warning
- the direct reproduction above -> unchanged orphan remains forwarded

JerrettDavis (approved):
Re-reviewed current head 8b37d8c. The previous blocker is fixed: the recovery path now handles and sanitizes top-level `tool_reference` blocks, not just nested `tool_search_tool_result` / `tool_result.content` references. The scanner also correctly handles `tool_reference.tool_name`, `tool_reference.name`, and `tool_use.name`, and the new handler-level tests verify the actual Anthropic outbound body gets repaired.

Validated locally:

- `.venv`Scripts`python.exe -m pytest tests/test_tool_search_reference_recovery.py tests/test_proxy/test_anthropic_tool_reference_recovery.py tests/test_issue_746_tool_search.py tests/test_ccr_tool_always_on.py -q` -> 91 passed
- `.venv`Scripts`python.exe -m ruff check headroom/proxy/ccr_marker_policy.py headroom/proxy/helpers.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py headroom/proxy/tool_search_recovery.py tests/test_ccr_tool_always_on.py tests/test_tool_search_reference_recovery.py tests/test_proxy/test_anthropic_tool_reference_recovery.py tests/test_issue_746_tool_search.py` -> passed

nangsontay:
Heads-up, since this touches `should_inject_ccr_tool`: I have a PR coming that **deletes** that function's `frozen_message_count` deferral entirely. See #2671 for the measurements.

The short version: the deferral is itself a cache defect. `tools` is the head of Anthropic's cache key, and deferring on the freeze counter removes a tool that is already inside the provider-cached prefix. On three days of local logs the turns that flip injection state carried **44.7% of all cache-write tokens at 52.0% hit**, against 98.1% for non-flipping turns. With the gate deleted, that dropped to 1.7% of write and the aggregate went 86.75% → 89.21%.

What that means for this PR: with the deferral gone, `apply_session_sticky_ccr_tool` already re-injects for any session that has done CCR, so `transcript_requires_tool`'s **cache** rationale disappears and it becomes purely a dangling-`tool_reference` 400-recovery path. That is still worth having, since this is not a request to drop it, just a flag that our two changes edit the same function, so one of them will want a rebase, and you may want to narrow the justification here to the 400-recovery case.

One more thing worth noting for reviewers of both PRs: the unredeemable-marker risk (markers present while `headroom_retrieve` is absent) is pre-existing on `upstream/main`, which has no `transcript_requires_tool` recovery at all. My PR neither creates nor fixes it.

JerrettDavis (approved):
Re-reviewed after merging current main. The resolution removes the retired frozen-count injection policy, routes transcript recovery through the current sticky helper, and preserves the approved generic deferred-tool restoration/sanitization and client-side ToolSearch guard. Validation: Ruff and format pass; 115 focused tool-search recovery, CCR, and Anthropic cache-stability tests pass.

JerrettDavis:
Current `main` now conflicts with this branch at the CCR injection gate and, more importantly, has an ordering conflict with the recovery design. #2807 added `strip_unsupported_tool_search_blocks`, which runs before this PR’s `reconcile_tool_references` call. A straight merge therefore strips the dangling references (and paired server calls) before the branch can use its per-session definition cache to re-inject them, making the tier-1 recovery path effectively unreachable. #2848 also changed the injection gate from “new marker” to “any marker,” overlapping the transcript-recovery condition.

This is not safe to resolve as a mechanical main merge: choosing to keep main’s strip-first policy, reorder cached recovery ahead of it, or retire the extra cache changes behavior and needs an author/maintainer correctness decision with refreshed handler-level tests. I have left the branch untouched. Please rebase and either narrow this to behavior still missing after #2807/#2848, or close it as superseded if strip-and-recover is the intended policy.

JerrettDavis:
Superseded by #2807, merged to main as 0237cbff. That merged change includes the transcript-driven CCR recovery, generic dangling tool_reference repair, sticky deferral behavior, and client-side ToolSearch guard from this branch, with current integration and tests. Closing the stale conflicting duplicate avoids reintroducing older handler/helper code.

PR #2512, opened 2026-07-23
fix(ccr): resolve <<ccr:...>> markers inline when no retrieve-tool path exists

### LINKED ISSUE #2509: [BUG] LiteLLM Guardrail/Proxy applies CCR compression but does not transparently decompress responses — raw <<ccr:...>> markers leak to agent
## Description

Headroom runs as a **separate proxy container** (`headroom-proxy`) in the same Docker stack as LiteLLM. LiteLLM has a built-in **guardrail provider** with a **headroom preset** in LiteLLM's `guardrails:` config section — headroom runs as a guardrail plugin inside LiteLLM. The headroom-proxy applies CCR (Compress-Cache-Retrieve) compression to outgoing content but **does not resolve `<<ccr:...>>` markers on the response path**. The markers leak through as raw text to the downstream agent.

The agent receives `<<ccr:HASH,type,size>>` markers as literal text in every tool output larger than approximately 100 bytes. Since this is a guardrail/proxy setup, the agent has no `headroom_retrieve` tool and should not need one — decompression must happen transparently in the proxy on the response path.

## Expected Behavior

CCR must be **fully transparent** in proxy mode:

1. Outgoing request: compress content → produce `<<ccr:...>>` markers  
2. Forward to upstream LLM  
3. **Incoming response: resolve all `<<ccr:...>>` markers** back to original content  
4. Agent receives clean, uncompressed content

## Actual Behavior

Step 3 does not happen. The agent sees raw `<<ccr:HASH,type,size>>` markers. Example from a real session (a simple `cat` command):

```
<<ccr:4e3a8b8b7aa5,string,23.6KB>>
```

This happens on **every** output above ~100 bytes — even a 5-line grep result triggers compression.

## Steps to Reproduce

1. Deploy the following stack (Docker Compose):

```yaml
services:
  litellm:
    image: docker.litellm.ai/berriai/litellm:main-stable
    container_name: litellm
    volumes:
      - ./config.yaml:/app/config.yaml
    env_file:
      - .env
    depends_on:
      db:
        condition: service_healthy
    networks:
      - traefik-proxy
      - internal
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.litellm.rule=Host(`litellm.example.com`)"
      - "traefik.http.routers.litellm.entrypoints=websecure"
      - "traefik.http.routers.litellm.tls.certresolver=myresolver"
      - "traefik.http.services.litellm.loadbalancer.server.port=4000"

  headroom-proxy:
    image: ${HEADROOM_IMAGE:-ghcr.io/headroomlabs-ai/headroom:latest}
    entrypoint: ["headroom", "proxy"]
    working_dir: /workspace
    restart: unless-stopped
    container_name: headroom-proxy
    environment:
      HOME: /tmp/headroom-home
      HEADROOM_HOST: 0.0.0.0
      HEADROOM_WORKSPACE_DIR: /tmp/headroom-home/.headroom
      HEADROOM_CONFIG_DIR: /tmp/headroom-home/.headroom/config
    volumes:
      - ./DATA/headroom-workspace:/workspace
      - ./DATA/headroom-home:/tmp/headroom-home
    command: ["--host", "0.0.0.0", "--port", "8787"]
    network_mode: service:litellm

networks:
  traefik-proxy:
    external: true
  internal:
    driver: bridge
```

2. Configure LiteLLM to route through headroom-proxy (e.g. proxy URL `http://headroom-proxy:8787` or via guardrail mechanism in LiteLLM's `config.yaml`)
3. Point any OpenAI-compatible agent at LiteLLM
4. Make a tool call that returns >100 bytes of text
5. Observe `<<ccr:HASH,type,size>>` markers in the agent's context

Image

## Environment

- **Deployment:** Docker containers (same `internal` Docker network)
- **Headroom image:** `ghcr.io/headroomlabs-ai/headroom:latest`
- **LiteLLM image:** `docker.litellm.ai/berriai/litellm:main-stable`
- **Headroom mode:** Standalone proxy (not library, not MCP server)
- **Integration:** LiteLLM routes through headroom-proxy via guardrail/provider config
- **LLM provider:** DeepSeek (likely provider-agnostic)
- **Agent:** Hermes Agent (OpenAI-compatible SDK)

## Additional Context

- The `headroom-proxy` container is **not exposed on any port outside** — it only communicates internally with LiteLLM over the `internal` network on port 8787
- The agent calls LiteLLM's public endpoint; LiteLLM then routes through headroom-proxy to the upstream provider
- No `HEADROOM_COMPRESS` threshold env vars are set — default compression behaviour applies
- Related issue #1006 addressed a similar problem in the standalone Anthropic proxy path (compression ran when `headroom_retrieve` tool injection was deferred). This is a different scenario: in a guardrail/proxy-to-LiteLLM setup **there is no tool injection path at all**, so markers can never be resolved on the agent side. Resolution must happen transparently inside the proxy's response handler.

Parideboy:
## Description

Fixes #2509. CCR marker resolution today depends entirely on the model calling `headroom_retrieve` back  a tool-call round-trip. Callers with no such round-trip (e.g. Headroom running as a LiteLLM guardrail/proxy hop, per the issue's repro) never get an offered path to redeem a marker, so raw `<<ccr:HASH,type,size>>` text leaks straight to the agent.

This adds an explicit, opt-in fallback: `--ccr-inline-resolve` / `HEADROOM_CCR_INLINE_RESOLVE`. When set, the proxy resolves markers directly from the compression store on the response path instead of waiting for a tool call. Off by default, guessing "this caller can't use tools" is fragile, so operators opt in explicitly for guardrail/proxy deployments.

## Type of Change

- [ ] Breaking change
- [ ] Documentation update

## Changes Made

- `headroom/ccr/marker_resolution.py` (new): `resolve_markers_in_text` /  `resolve_markers_in_response` regex-match `<<ccr:HASH,...>>`, look up the hash in `CompressionStore`, splice the original content back in. A miss (expired/evicted hash) leaves the marker in place with the miss reason appended, since there's no tool-call round-trip to report it back to the model.
- `headroom/proxy/models.py`: `ProxyConfig.ccr_resolve_markers_inline: bool = False`.
- `headroom/cli/proxy.py`: `--ccr-inline-resolve` flag / `HEADROOM_CCR_INLINE_RESOLVE` env, wired into `ProxyConfig`.
- `headroom/proxy/handlers/anthropic.py`, `headroom/proxy/handlers/openai.py`:
  call `resolve_markers_in_response` on the finalized response JSON, right after existing CCR tool-call handling, at all three non-streaming response sites (Anthropic Messages, OpenAI Chat Completions backend path, OpenAI Responses API). Streaming responses are out of scope for this PR, tracked as follow-up, noted in the module docstring's scope.

## Testing

- [x] Added new tests
- [x] All tests pass locally

```
$ python -m pytest tests/test_ccr_marker_resolution.py -q
collected 6 items

tests\test_ccr_marker_resolution.py ......                               [100%]

============================== 6 passed in 0.45s ==============================

$ python -m pytest tests/test_ccr.py tests/test_ccr_response_handler.py tests/test_ccr_response_handler_extra.py tests/test_ccr_response_handler_openai_responses.py tests/test_proxy/test_openai_responses_ccr.py tests/test_proxy/test_anthropic_ccr_raise.py -q
======================= 83 passed, 1 warning in 34.50s ========================
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.11, local headroom repo (`G:\Programmi Aggiuntivi\headroom`)
- Exact command / steps: `python -m pytest tests/test_ccr_marker_resolution.py tests/test_ccr.py tests/test_ccr_response_handler.py tests/test_ccr_response_handler_extra.py tests/test_ccr_response_handler_openai_responses.py tests/test_proxy/test_openai_responses_ccr.py tests/test_proxy/test_anthropic_ccr_raise.py -q`
- Observed result: 89 passed, 0 failed (6 new + 83 existing CCR tests, no regressions). `ruff check`, `ruff format --check`, and `mypy --ignore-missing-imports` all clean on every changed/new file.
- Not tested: the actual Docker Compose / LiteLLM guardrail deployment from the issue's repro steps (no such environment available here); streaming response paths (out of scope, see Changes Made).

## Review Readiness

### DIFF
--- a/headroom/ccr/marker_resolution.py
+++ b/headroom/ccr/marker_resolution.py
@@ -1,0 +1,83 @@
+"""Inline resolution of ``<<ccr:...>>`` markers on the response path.
+
+Normal CCR resolution relies on the ``headroom_retrieve`` tool: a marker is
+redeemed when the model calls the tool back. That path assumes there's a
+subsequent turn in which the model *can* call it. Callers that never see an
+injected tool at all — e.g. Headroom running as a LiteLLM guardrail/proxy hop
+with no tool-call turn in between (#2509) — have no way to redeem a marker,
+so it leaks through as raw text.
+
+This module provides an explicit, opt-in fallback (``--ccr-inline-resolve``):
+scan the outgoing response for markers and substitute the original content
+directly, instead of leaving the marker for the model to redeem later.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import re
+from typing import Any
+
+from ..cache.compression_store import (
+    CompressionStore,
+    format_retrieval_miss_detail,
+    get_compression_store,
+)
+
+logger = logging.getLogger(__name__)
+
+# Matches the opaque-blob marker form `<<ccr:HASH,KIND,SIZE>>` (and the
+# row-offload form `<<ccr:HASH N_rows_offloaded>>`) emitted by SmartCrusher.
+# HASH is 12-24 hex chars; see headroom/ccr/tool_injection.py for the same
+# constant used on the injection side.
+_MARKER_RE = re.compile(r"<<ccr:([a-f0-9]{12,24})[^>]*>>")
+
+
+def resolve_markers_in_text(text: str, *, store: CompressionStore | None = None) -> str:
+    """Replace every ``<<ccr:HASH,...>>`` marker in ``text`` with its original content.
+
+    A miss (expired/evicted/unknown hash) can't be reported back to the
+    model on this path — there's no tool-call round-trip — so the marker is
+    left in place with the miss reason appended rather than raising.
+    """
+    if "<<ccr:" not in text:
+        return text
+
+    resolved_store = store or get_compression_store()
+
+    def _replace(match: re.Match[str]) -> str:
+        hash_key = match.group(1)
+        entry = resolved_store.retrieve(hash_key)
+        if entry is not None:
+            original = entry.original_content
+            return original if isinstance(original, str) else json.dumps(original)
+
+        get_status = getattr(resolved_store, "get_entry_status", None)
+        status = get_status(hash_key, clean_expired=True) if callable(get_status) else None
+        detail = format_retrieval_miss_detail(status) if status else "entry not found"
+        logger.warning(f"CCR inline-resolve: marker {hash_key} unresolvable ({detail})")
+        return f"{match.group(0)} [unresolved: {detail}]"
+
+    return _MARKER_RE.sub(_replace, text)
+
+
+def resolve_markers_in_response(response: Any, *, store: CompressionStore | None = None) -> Any:
+    """Recursively resolve ``<<ccr:...>>`` markers in every string field of a payload.
+
+    Walks the full response structure rather than picking out
+    provider-specific fields (``content`` blocks, ``message.content``,
+    Responses-API ``output`` items, ...) so it stays correct regardless of
+    where a marker ends up, and doesn't need per-provider maintenance.
+    """
+    resolved_store = store or get_compression_store()
+    if isinstance(response, str):
+        return resolve_markers_in_text(response, store=resolved_store)
+    if isinstance(response, list):
+        return [resolve_markers_in_response(item, store=resolved_store) for item in response]
+    if isinstance(response, dict):
+        return {
+            key: resolve_markers_in_response(value, store=resolved_store)
+            for key, value in response.items()
+        }
+    return response
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -321,7 +321,8 @@ def dashboard(port: int, no_open: bool) -> None:
         "Resolve <<ccr:...>> markers inline on the response path instead of "
         "relying on the model to call headroom_retrieve. For callers with no "
         "tool-call round-trip to redeem a marker (e.g. Headroom running as a "
-        "LiteLLM guardrail/proxy hop, see issue #2509). Off by default. "
+        "LiteLLM guardrail/proxy hop, see issue #2509). Applies to non-streaming "
+        "responses only. Off by default. "
         "Env: HEADROOM_CCR_INLINE_RESOLVE."
     ),
 )
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -314,6 +314,18 @@ def dashboard(port: int, no_open: bool) -> None:
     ),
 )
 @click.option(
+    "--ccr-inline-resolve",
+    is_flag=True,
+    envvar="HEADROOM_CCR_INLINE_RESOLVE",
+    help=(
+        "Resolve <<ccr:...>> markers inline on the response path instead of "
+        "relying on the model to call headroom_retrieve. For callers with no "
+        "tool-call round-trip to redeem a marker (e.g. Headroom running as a "
+        "LiteLLM guardrail/proxy hop, see issue #2509). Off by default. "
+        "Env: HEADROOM_CCR_INLINE_RESOLVE."
+    ),
+)
+@click.option(
     "--no-ccr-proactive-expansion",
     is_flag=True,
     envvar="HEADROOM_NO_CCR_PROACTIVE_EXPANSION",
@@ -934,6 +946,7 @@ def proxy(
     tpm: int | None,
     no_ccr: bool,
     lossless: bool,
+    ccr_inline_resolve: bool,
     no_ccr_proactive_expansion: bool,
     proxy_extension: tuple[str, ...],
     compressor: tuple[str, ...],
@@ -1206,6 +1219,7 @@ def proxy(
         # CCR fully on.
         ccr_inject_tool=not no_ccr,
         ccr_inject_marker=not no_ccr,
+        ccr_resolve_markers_inline=ccr_inline_resolve,
         lossless=lossless,
         ccr_proactive_expansion=not no_ccr_proactive_expansion,
         # Flatten repeat-flag tuple AND any comma-separated values inside it.
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -25,6 +25,7 @@ import httpx
 
 from headroom.agent_savings import proxy_pipeline_kwargs
 from headroom.ccr.context_tracker import looks_like_claude_code_compact_summary
+from headroom.ccr.marker_resolution import resolve_markers_in_response
 from headroom.copilot_auth import build_copilot_upstream_url
 from headroom.pipeline import PipelineStage, summarize_routing_markers
 from headroom.proxy.auth_mode import classify_auth_mode, classify_client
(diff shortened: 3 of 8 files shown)


chopratejas:
There is an option in Headroom called --no-ccr - ideally when we cannot get CCR back - say as a sidecar etc - that is the option LiteLLM should set.

tilllt:
Hi @Parideboy, thanks for picking this up so quickly! 👏

I wanted to add a note on **`--no-ccr`** — it's an existing flag someone might reach for as a workaround for the marker leak. It does resolve the leak (no markers emitted), but it's an **all-or-nothing toggle**: CCR as a whole is either on or off. You can't selectively disable markers while keeping the rest of the compression pipeline active.

That said, after looking into the architecture more closely: `--no-ccr` only disables the reversibility layer (caching originals + marker injection). The actual compressors — **SmartCrusher**, **CodeCompressor**, and **Kompress-base** — keep running independently. So the trade-off isn't "no savings vs. leak", it's **"lossy instead of reversible compression"** — the LLM loses the ability to `headroom_retrieve` full originals, but most of the token savings remain.

For a guardrail/proxy setup, `--ccr-inline-resolve` is still the cleaner fix because it preserves reversibility. But `--no-ccr` wouldn't be a total loss of savings either — just a loss of recoverability. Both options are worth understanding when picking the right approach.

Looking forward to seeing this land!

Parideboy:
@chopratejas @tilllt on the `--no-ccr` question, plus a correction to this PR itself.

**First, a real defect in my own earlier commits.** Every `resolve_markers_in_response()` call was sitting inside a block gated on `ccr_response_handler.has_ccr_tool_calls(...)`. That gate is only true when the model emitted a `headroom_retrieve` tool call and #2509 is precisely the case where no tool call can ever happen. The flag was a no-op for its own use case. The 0% Codecov patch coverage on those hunks was the tell: the tests only exercised the pure module, never the handler wiring. Fixed in e4eed892, which also adds the hook the direct-OpenAI chat path never had, and `tests/test_ccr_inline_resolve_handlers.py` handler-level tests with zero tool calls in the response fixture, which fail against the previous tip.

**On `--no-ccr` as the alternative.** It does stop the leak, but it is not free, and the cost is larger than "you lose recoverability":

- `headroom/cli/proxy.py:1210-1211` `--no-ccr` sets both `ccr_inject_tool` and `ccr_inject_marker` to `False`.
- `headroom/transforms/config_compressor.py:145,154` the schema-fold and comment-elision tiers are gated on `enable_ccr`, so they turn **off** entirely. This is where I'd correct @tilllt's "most savings remain": on config files, real savings are lost, not just recoverability.
- `headroom/transforms/code_compressor.py:1255` bodies are still stripped, but the `[N tokens compressed... Retrieve more: hash=]` note is skipped.
- `headroom/transforms/smart_crusher.py:~312` the row drop still runs; the marker and the store write are skipped.

So with `--no-ccr` the compression that remains is silently lossy: content is dropped with no in-band signal to the model that anything is missing. That is the argument for keeping `--ccr-inline-resolve` as the answer for guardrail-hop deployments the content gets substituted back rather than the marker being suppressed. The rest of @tilllt's read is right.

**Scope, stated plainly:** non-streaming responses only. Buffered-CCR streaming (`_buffered_ccr_operation` in both handlers) is deliberately untouched and is follow-up work. The CLI help, the `HEADROOM_CCR_INLINE_RESOLVE` settings help and the `ProxyConfig` comment now all say so. Also disclosing: the backend-chat path is fixed in code but is not directly covered by the new tests, since it needs a configured backend; the other three paths are covered both flag-on and flag-off.

JerrettDavis (approved):
Reviewed the latest e4eed892 update. The inline resolver now runs outside the retrieve-tool gate on the non-streaming paths, response headers are sanitized before rebuilt JSON responses, and the env var is correctly threaded through Click into ProxyConfig. Verified locally: pytest tests/test_ccr_marker_resolution.py tests/test_ccr_inline_resolve_handlers.py tests/test_ccr.py tests/test_ccr_response_handler.py tests/test_ccr_response_handler_extra.py tests/test_ccr_response_handler_openai_responses.py tests/test_proxy/test_openai_responses_ccr.py tests/test_proxy/test_anthropic_ccr_raise.py -q (95 passed, 1 warning); ruff check touched files.

D0wn10ad:
will be great to see this thru - using litellm + headroom as guardrail, now opencode's mcp calls and execution of python scripts output have <ccr...> marker exposed and is not easy to proceed.

Parideboy:
@D0wn10ad thanks for the report. Flagging up front: **this PR as it stands likely does not fix your setup**, and which gap you are hitting depends on where Headroom sits in your stack.

**Which shape is your deployment?**

1. `HeadroomCallback` registered in `litellm_settings.callbacks` (the guardrail integration), or
2. Headroom's own proxy running as the `api_base` that LiteLLM forwards to?

**If (1), the callback, `--ccr-inline-resolve` cannot fire at all.**

* `headroom/integrations/litellm_callback.py:93` `async_pre_call_hook` rewrites `data["messages"]` through `headroom.compress()` (`:142` local, `:160` cloud). Markers are injected there.
* Nothing runs on the response side. `async_success_handler` (`:195`) and `async_failure_handler` (`:201`) are both `pass`, and `async_post_call_success_hook` exists only as the inherited no-op default from `CustomLogger` (see the class docstring at `:37`). It is never overridden.
* The flag added in this PR is a `ProxyConfig` field (`headroom/proxy/models.py:195`), read only in `headroom/cli/proxy.py:1226` and `headroom/proxy/handlers/anthropic.py:3407`, `headroom/proxy/handlers/openai.py:3734,4240,5322`. The callback never constructs a `ProxyConfig` and never imports `headroom/ccr/marker_resolution.py`.
* The callback also rewrites `messages` only, never `tools`, so `headroom_retrieve` is never injected in this mode. That is #2509 in its purest form.

**If (2), Headroom proxy behind LiteLLM, you are hitting the streaming limitation stated above.** opencode streams by default, and this PR is non-streaming only: `headroom/proxy/handlers/streaming.py:1552,2049` only record CCR feedback, and `_buffered_ccr_operation` (`anthropic.py:2792`, `openai.py:4946`) is untouched.

**One distinction worth separating, because it bears on "not easy to proceed".** Inline resolve is a response-path fix. Your markers are in MCP results and python stdout, which get compressed on the *request* path. Resolving the response stops the raw marker reaching your client, but the model itself still reads a marker it has no tool to redeem. If the blocking symptom is the model losing the script output rather than you seeing marker text, the response-path fix alone is not sufficient: the request side needs either a retrieve tool the caller can actually invoke, or markers suppressed for that hop.

Follow-ups I can pick up, kept out of this PR to keep it scoped:

* callback: implement `async_post_call_success_hook` and `async_post_call_streaming_iterator_hook` on `HeadroomCallback`, gated on the same `HEADROOM_CCR_INLINE_RESOLVE`.
* proxy: streaming-path resolution inside the buffered-CCR operation.

If you can say which of the two shapes you are on, plus a redacted snippet of your LiteLLM config, I will target that one first.

JerrettDavis (approved):
Re-reviewed at `40abc302`. After normalizing diff metadata, the current marker-resolution contribution is byte-for-byte equivalent to the previously approved revision; the follow-up commits restore the same settings registration and handler placement after the main refresh.

PR #1490, opened 2026-06-27
feat: add deterministic runtime rollout controls

JerrettDavis:
## Description

Establish one centrally resolved, observable, deterministic, versioned runtime rollout-control mechanism for Headroom. Runtime rollout controls which behaviors an already-built artifact may expose; it does not select or qualify a Headroom release/version.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Bug fix (non-breaking change that fixes rollout enforcement regressions)
- [x] Documentation update
- [x] Code refactoring (no functional changes)

## Changes Made

- Added `RolloutChannel`, `HEADROOM_ROLLOUT_CHANNEL`, `--rollout-channel`, and a versioned immutable `RolloutSnapshot` shared by Python configuration boundaries.
- Added schema/policy versions, canonical registry and snapshot SHA-256 identities, per-feature decision reasons, disable precedence, unsafe qualification poisoning, strict CLI validation, and fail-closed environment handling.
- Added `headroom rollout status --json`, Python `/stats.rollout`, and Rust `/rollout/status` runtime provenance.
- Added equivalent Rust snapshot semantics and shared Python/Rust policy vectors while retaining language-specific feature registries.
- Enforced rollout policy at alternate Python server composition roots so `HEADROOM_READ_MATURATION=1` cannot bypass its beta gate.
- Preserved typed rollout snapshots across multi-worker serialization with schema, policy, registry, snapshot-digest, type, and feature-name validation.
- Made loopback runtime output-shaper updates replace the immutable snapshot atomically for request readers, retain explicit request/disable provenance, preserve channel and kill-switch precedence, invalidate cached stats, and return the effective rollout decision.
- Made `headroom learn --verbosity --apply` report a channel-blocked update instead of claiming the shaper is live.
- Made explicit CLI feature flags fail loudly when their current channel blocks them.
- Made persistent interceptor installation select canary automatically, or reject an explicitly insufficient channel unless the break-glass override is set.
- Updated architecture, proxy, rollout, learn, and output-shaper documentation with required channels and hot-reload semantics.

## Testing

- [x] Unit tests pass
- [x] Linting passes (`ruff check .` and `ruff format --check .`)
- [x] Type checking passes (`mypy headroom --ignore-missing-imports`)
- [x] New regression tests added for every corrected behavior
- [x] Rust tests and production-target Clippy pass
- [x] Documentation build passes

### Test Output

```text
Focused rollout coverage suite
[...]

PR #2524, opened 2026-07-24
fix(proxy): preserve established frozen prefix from output-shaper system mutation

nangsontay:
## Description

The output shaper appends the verbosity-steering tail to `body["system"]` with no frozen-prefix awareness. The system prompt is the head of the provider's cache prefix, so shaping a conversation whose frozen prefix was established without the tail invalidates the entire cached prefix (tools + system + frozen messages). Conversation-stable arm assignment protects steady state, but not conversations that start unshaped and gain the tail mid-flight (shaper enabled via env or live override while sessions are in progress).

A related test-isolation gap: proxy construction re-applies file-backed settings, so a developer's real `~/.headroom/settings.json` leaks live knobs into pytest and makes `test_memory_context_avoids_system_mutation_when_prefix_frozen` machine-dependent.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: skip `shape_request` when the tracker reports an established frozen prefix that predates the shaping tail; once a conversation has been shaped, replay the level it was established at. The pin is read *before* the enablement and arm gates, so turning the shaper off (or drifting into control) mid-conversation replays the established system bytes instead of dropping them. The replay path is narrower than the treatment path: no experiment label (attribution stays with the live assignment) and no effort routing (`output_config.effort` / `thinking` never enter the cached prefix). Explicit request bypass still wins.
- `headroom/cache/prefix_tracker.py`: new `PrefixCacheTracker.output_shaping_level` sticky pin (`int | None`) carrying the per-conversation level the frozen prefix was built at — `None` means never shaped, which is distinct from a real level `0`.
- `tests/conftest.py`: point `HEADROOM_SETTINGS_PATH` at a per-test path in the autouse env-scrub fixture so the developer's real settings file cannot leak into assertions.
- `tests/test_cli_proxy_env.py`: unset `HEADROOM_SETTINGS_PATH` in the one test that intentionally resolves the settings file through `HEADROOM_WORKSPACE_DIR`.
- `tests/test_proxy_anthropic_cache_stability.py`: new parametrized regression `test_output_shaper_respects_established_frozen_prefix` (skip when frozen-and-unshaped / shape fresh conversations / keep shaping shaped ones); `test_output_shaper_pins_the_level_the_frozen_prefix_was_built_at` (level `0` and drift in both directions); and `test_output_shaper_replays_pinned_tail_when_shaping_stops` (L2 frozen then shaper disabled, and L2 frozen then holdout drift into control — both asserting the exact established system bytes and the absence of a treatment label).

## Testing

### Test Output

```text
$ pytest tests/test_proxy_anthropic_cache_stability.py tests/test_cross_turn_cache_safety.py \
    tests/test_compression_cache.py tests/test_proxy/test_settings_store.py \
    tests/test_proxy_settings_endpoints.py tests/test_cli_proxy_env.py \
    tests/test_output_savings.py tests/test_output_savings_policy.py \
    tests/test_openai_responses_output_shaper.py tests/test_rollout.py -q
================== 295 passed, 1 skipped, 1 warning in 12.04s ==================

$ ruff check headroom tests
All checks passed!

$ mypy headroom
Success: no issues found in 519 source files
```

## Real Behavior Proof

- Environment: macOS (Darwin 25.4), Python 3.13 venv; real `headroom proxy` process built from this branch (`headroom proxy --port 8788 --mode token --no-optimize --no-ccr`), forwarding to a local mock Anthropic upstream (`ANTHROPIC_TARGET_API_URL=http://127.0.0.1:9999`) that logs the forwarded `system` field and returns canned responses with `cache_read_input_tokens: 4000`, so the prefix tracker genuinely freezes the conversation.
- Exact command / steps: (1) two `curl /v1/messages` turns on conversation A with the shaper **off** — establishes a real frozen prefix without the tail; (2) enable the shaper live: `curl -X POST /admin/runtime-env -d '{"HEADROOM_OUTPUT_SHAPER":"1"}'`; (3) send turn 3 on conversation A; (4) send turn 1 of a fresh conversation B.
- Observed result: `system` as received by the upstream, per request (see below):

```text
req1 n_msgs=1 system='proof base system'                        # shaper off
req2 n_msgs=3 system='proof base system'                        # frozen prefix now established
req3 n_msgs=5 system='proof base system'                        # shaper ON, frozen conv -> tail withheld
req4 n_msgs=1 system=[{"type": "text", "text": "fresh conversation system"},
                      {"type": "text", "text": "<headroom_output_shaping>..."}]   # fresh conv -> shaped
```

  Proxy log for the same run:

```text
[hr_1784879972_000003] OutputShaper: skipped — frozen prefix (4 messages) predates the shaping tail; preserving provider cache
[hr_1784879972_000004] OutputShaper(L2/learned): ['output_shaper:verbosity:L2']
```

  Same before/after at test level: reverting only the handler to main (`git checkout upstream/main -- headroom/proxy/handlers/anthropic.py`) makes the new regression fail `[1-False-False]` (tail appended onto a frozen, never-shaped conversation) and `[0-False-True]` (shaped state never recorded on the tracker); with the guard restored, all 3 parametrized cases pass.

  Reverse direction (the requested change), verified at test level on the same handler-revert method — with only the pre-fix handler in place, both new cases fail on the exact system bytes, showing the established L2 tail being dropped rather than replayed:

```text
$ pytest tests/test_proxy_anthropic_cache_stability.py -q -k replays_pinned   # pre-fix handler
E   AssertionError: assert 'base system' == [{'text': 'base system', 'type': 'text'},
E     {'text': '<headroom_output_shaping>\nSkip preamble and postamble; ...', 'type': 'text'}]
FAILED ...::test_output_shaper_replays_pinned_tail_when_shaping_stops[shaper_disabled]
FAILED ...::test_output_shaper_replays_pinned_tail_when_shaping_stops[arm_control]

$ pytest tests/test_proxy_anthropic_cache_stability.py -q -k output_shaper    # with the fix
6 passed, 28 deselected, 1 warning in 2.60s
```

  A first attempt at the disabled case silently stayed in treatment: since #1490 landed on `main`, enablement is a rollout snapshot resolved once at proxy construction, so a live `HEADROOM_OUTPUT_SHAPER=0` no longer reaches the handler as a per-request env read. The regression now re-resolves the gate with the legacy alias falsey — the same route a channel demotion takes — and the missing treatment label in `x-headroom-transforms` proves the drift actually took effect rather than the treatment path having replayed the pin.

- Not tested: streaming responses and the live Anthropic API on this exact branch (mock upstream, non-streaming only); multi-worker deployments. The reverse-direction proof above is test-level only — the live mock-upstream proxy run was performed for the forward direction, not re-run for the replay path. The OpenAI handler was checked and does not call `shape_request`, so no equivalent change is needed there.

## Runtime Rollout Safety

- Rollout-managed feature(s): `proxy_output_shaper` (legacy alias `HEADROOM_OUTPUT_SHAPER`).
- Minimum rollout channel: `beta` — unchanged by this PR. Shaping still only runs when the resolved gate says enabled.
- Stable/default behavior changed: No. On `stable` the shaper never shapes, so no conversation is ever pinned (`output_shaping_level` stays `None`) and the new replay branch is unreachable.
- Kill switch / disable path: unchanged — `HEADROOM_DISABLE_FEATURES=proxy_output_shaper`, the falsey legacy alias, or a channel demotion. Deliberate caveat introduced here: disabling stops shaping new conversations immediately, but a conversation whose prefix already froze with a tail keeps replaying those exact system bytes until its prefix thaws. That is the fix — the alternative is busting the provider cache on every in-flight session at disable time. It is bounded by the session's prefix lifetime and carries no experiment attribution.
- Qualification impact: None — no change to `HEADROOM_UNSAFE_ALLOW_UNSTABLE_FEATURES` handling or to qualification eligibility.
- Rollback path: revert this PR's commits. No persisted state to unwind — `PrefixCacheTracker.output_shaping_level` is an in-memory per-session attribute that dies with the proxy process.

## Review Readiness

## Checklist

N/A — proxy-internal behavior; evidence is in the mock-upstream and proxy-log excerpts above.

## Additional Notes

- No linked issue; discovered while debugging prompt-cache hit-rate degradation in token mode.
- Documentation checklist item is N/A: the shaper remains opt-in with unchanged configuration surface; this only constrains when the tail may first be applied.
- The one skipped test in the suite run is a pre-existing external-model skip, unrelated to this change.


JerrettDavis (changes requested):
Found one cache-safety issue in the new sticky flag logic.

In `headroom/proxy/handlers/anthropic.py`, the code sets `prefix_tracker.output_shaping_applied = True` immediately after `shape_request(...)`, before checking what actually changed. That flag is described and used as “the system prompt already carried the output-shaping tail”, but `shape_request` can return without adding the Anthropic system steering tail: for example if the resolved verbosity level is `0`, or if only the effort router mutates `output_config` / `thinking`. In that case a fresh conversation can freeze with no steering tail, but the tracker is now marked shaped. If a later runtime/learned verbosity level becomes positive, the `frozen_message_count > 0 and not _shaped_before` guard will not fire, and the handler will append the tail to an already-established frozen prefix, recreating the cache invalidation this PR is intended to prevent.

Please make the sticky flag reflect the system-tail state specifically, e.g. set it only when the Anthropic system field contains `STEERING_SENTINEL` after shaping or when the verbosity label/tail was actually applied, and add a regression for level `0` or effort-only shaping followed by a frozen prefix.

Local verification run before this finding:
- `pytest tests/test_proxy_anthropic_cache_stability.py tests/test_cross_turn_cache_safety.py tests/test_compression_cache.py tests/test_proxy/test_settings_store.py tests/test_proxy_settings_endpoints.py tests/test_cli_proxy_env.py -q` (after copying the local compiled `headroom._core` extension into the detached worktree): `186 passed, 1 skipped`
- `ruff check headroom/cache/prefix_tracker.py headroom/proxy/handlers/anthropic.py tests/conftest.py tests/test_cli_proxy_env.py tests/test_proxy_anthropic_cache_stability.py`

JerrettDavis (approved):
Reviewed the latest commit `f41a1914`; this resolves my prior requested change.

The new `output_shaping_level` pin correctly distinguishes “never shaped” from level `0`, so a conversation that freezes without a steering tail will not later gain one, and a conversation that freezes with an L2 tail keeps that exact tail even if the live verbosity level drifts. The added regression covers the level `0` case and both directions of level drift.

Local verification:
- `uv run --extra dev python -m pytest tests/test_proxy_anthropic_cache_stability.py tests/test_cross_turn_cache_safety.py tests/test_compression_cache.py tests/test_proxy/test_settings_store.py tests/test_proxy_settings_endpoints.py tests/test_cli_proxy_env.py -q` -> `187 passed, 1 skipped, 1 warning`
- `ruff check headroom/cache/prefix_tracker.py headroom/proxy/handlers/anthropic.py tests/conftest.py tests/test_cli_proxy_env.py tests/test_proxy_anthropic_cache_stability.py` -> `All checks passed!`

I did not find a remaining blocking correctness issue in this diff.

JerrettDavis (approved):
Reviewed the synchronized latest head. The tracker state distinguishes never-shaped from a real level-0 pin, prevents all three cache-busting transitions (add, remove, or change the tail), and applies only within the stable treatment arm. The tests cover frozen/unshaped, fresh/shaped, already-shaped, and live level drift in both directions; the ambient settings isolation change is narrowly compensated in the settings-path test. Current substantive checks are green.

JerrettDavis (changes requested):
The level pin fixes live L0/L2/L3 drift while shaping remains enabled and in treatment, but the reverse prefix-bust path is still open. The pinned state is consulted only inside `if _shaper_settings.enabled:` and `if _arm == "treatment":`. A conversation can establish/freeze an L2 system tail, then a live `HEADROOM_OUTPUT_SHAPER=0` override makes the whole block skip and forwards the client’s unshaped system prompt; similarly, changing the live holdout can move the deterministic score into control and skip the pinned tail. In both cases the established system-prefix bytes change, contradicting the PR’s “tail must never flip mid-conversation in either direction” invariant and invalidating the whole provider cache.

Please make an established frozen `output_shaping_level` authoritative before current enablement/arm gating: replay that pinned level whenever the frozen prefix exists, while still respecting explicit request bypass semantics if bypass is intentionally allowed to sacrifice the cache. Do not record a new experiment treatment label when merely replaying a previously established tail under a now-disabled/control configuration; keep behavioral shaping and experiment attribution distinct. Add regressions for (1) L2 frozen then shaper disabled, and (2) L2 frozen then holdout/control drift, asserting the exact established system bytes remain unchanged. The existing level-drift and never-shaped-prefix cases otherwise look correct, and the ambient-settings test isolation is a useful cleanup.

nangsontay:
Fixed in `47146738` (plus `9f4dbb80` for the test, and a merge of current `main`).

The pin is now read **before** both gates and replayed whenever a frozen prefix carries one:

```python
_pinned = getattr(prefix_tracker, "output_shaping_level", None)
_pinned_established = frozen_message_count > 0 and _pinned is not None

_shaper_settings = OutputShaperSettings.from_env(enabled=...)
_arm = "control"
if _shaper_settings.enabled:
    ...  # holdout / arm / stratum label — unchanged
if _shaper_settings.enabled and _arm == "treatment":
    ...  # unchanged treatment path
elif _pinned_established:
    if _pinned > 0 and apply_verbosity_steering(body, _pinned):
        body_mutation_tracker.mark_mutated("output_shaper")
```

On the two points you raised about the replay's scope:

- **No new experiment label.** The replay branch sits outside `if _shaper_settings.enabled:`, so it never reaches `stratum_label(...)`. Attribution stays tied to the live assignment: a disabled conversation records nothing, and a control-drifted one records the *control* label it genuinely earned. A cache replay is never counted as a treatment observation.
- **Behavioral shaping stays distinct.** The replay deliberately calls `apply_verbosity_steering` rather than `shape_request`, so it reproduces the system bytes and nothing else — no effort routing, since `output_config.effort` / `thinking` are top-level fields that never enter the cached prefix. Replaying those would be behavioral shaping on a conversation that is no longer in treatment.
- **Bypass semantics untouched.** The whole block still sits under the `if not _bypass:` guard, so a client explicitly asking for an unshaped body still gets one and knowingly pays the cache cost.

Regressions added as `test_output_shaper_replays_pinned_tail_when_shaping_stops[shaper_disabled|arm_control]`, both asserting the exact established system bytes and the absence of a treatment label.

One correction worth flagging on the disabled case: my first version of that test silently stayed in treatment and still passed. Since #1490 landed on `main`, enablement is a rollout snapshot resolved once at proxy construction, so a live `HEADROOM_OUTPUT_SHAPER=0` no longer reaches the handler as a per-request env read — the byte assertion was being satisfied by the treatment path replaying the pin, not by the new branch. The test now re-resolves the gate with the legacy alias falsey (the same route a channel demotion takes), and the missing treatment label is what proves the drift actually took effect.

Verification, against current `main`:

```text
$ pytest tests/test_proxy_anthropic_cache_stability.py -q -k replays_pinned   # pre-fix handler
E   AssertionError: assert 'base system' == [{'text': 'base system', 'type': 'text'},
E     {'text': '<headroom_output_shaping>\nSkip preamble and postamble; ...', 'type': 'text'}]
FAILED ...[shaper_disabled]
FAILED ...[arm_control]

$ pytest tests/test_proxy_anthropic_cache_stability.py tests/test_cross_turn_cache_safety.py \
    tests/test_compression_cache.py tests/test_proxy/test_settings_store.py \
    tests/test_proxy_settings_endpoints.py tests/test_cli_proxy_env.py \
    tests/test_output_savings.py tests/test_output_savings_policy.py \
    tests/test_openai_responses_output_shaper.py tests/test_rollout.py -q
295 passed, 1 skipped, 1 warning in 12.04s

$ ruff check headroom tests   -> All checks passed!
$ mypy headroom               -> Success: no issues found in 519 source files
```

Also merged current `main` and picked up the beta-channel pairing the shaper cache-stability tests need post-#1490 (`828c813c`) — without it the shaper resolves `blocked_by_channel` on the default stable channel and those tests die on `AttributeError` before asserting anything. The PR body now carries the `Runtime Rollout Safety` section the template gained, including the one deliberate caveat: disabling the shaper stops shaping new conversations immediately, but an already-frozen conversation keeps replaying its established tail until its prefix thaws.

JerrettDavis (approved):
Re-reviewed exact head f1bf8810. The requested fix is present: an established frozen-prefix pin is consulted before current enablement/arm gates, its exact verbosity tail is replayed when shaping is disabled or drifts to control, replay does not create treatment attribution, and explicit bypass remains the intentional cache-sacrifice boundary. I reran all 34 cache-stability tests plus targeted Ruff checks on Windows; all passed.

PR #2517, opened 2026-07-23
fix: Vertex model pricing shows $0.00 for versioned model names and vertex:anthropic provider

yossiovadia:
## Description

Two bugs cause `$0.00` cost display for Vertex AI users in headroom's dashboard:

1. **Model name resolution** — Vertex appends `@YYYYMMDD` version tags at runtime (e.g. `claude-haiku-4-5@20251001`). LiteLLM's database stores bare names without version suffixes, so every versioned model missed the lookup.

2. **Prefix cache savings** — the provider match checks `provider == "anthropic"` but Vertex traffic is tagged `provider == "vertex:anthropic"`, so cache read savings computed as $0.00. This bug is **not** addressed by #2516.

Fixes #2515

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `headroom/pricing/litellm_model_resolution.py`: strip `@YYYYMMDD` suffix before lookup; add `vertex_ai/` to `MODEL_PREFIX_RULES` for Claude models; apply prefix rules to both original and bare names
- `headroom/proxy/cost.py`: extend provider match to include `vertex:anthropic` alongside `anthropic` for prefix cache savings
- `tests/test_pricing_litellm_model_resolution.py`: 4 new tests covering suffix stripping, versioned model resolution, pricing lookup, and end-to-end resolve

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ python -m pytest tests/test_pricing_litellm_model_resolution.py -v
collected 10 items

tests/test_pricing_litellm_model_resolution.py::test_prefix_rule_matches_case_insensitively PASSED
tests/test_pricing_litellm_model_resolution.py::test_resolution_candidates_try_bare_then_matching_prefix_then_alias PASSED
tests/test_pricing_litellm_model_resolution.py::test_pricing_lookup_candidates_include_provider_prefixes_and_aliases PASSED
tests/test_pricing_litellm_model_resolution.py::test_retired_claude_3_sonnet_aliases_to_sonnet_tier_not_haiku PASSED
[...]

PR #2530, opened 2026-07-24
fix(proxy/anthropic): forward the URL-path model to the backend body (#2363)

### LINKED ISSUE #2363: [BUG] litellm-vertex backend silently defaults to claude-3-5-sonnet-20241022 for all Claude Code + ANTHROPIC_VERTEX_BASE_URL requests (model_override never propagated to body)

abhay-codes07:
## Description

Fixes #2363. Claude Code's Vertex-native rawPredict/streamRawPredict requests carry the model as a **URL path segment**, not a JSON body field. `handle_anthropic_messages` resolves it via `model_override`:

```python
raw_model = body.get("model") or model_override or "unknown"
model = sanitize_anthropic_model_id(raw_model) if isinstance(raw_model, str) else raw_model
body_model = body.get("model")
if isinstance(body_model, str) and model != body_model:   # only when body already has a string model
    body["model"] = model
    ...
```

Vertex rawPredict requests never had a `body["model"]`, so `body_model` is `None`, `isinstance(body_model, str)` is `False`, and the correctly-resolved model is never written back. On the `--backend litellm-vertex` (or bedrock) path the backend then reads:

```python
original_model = body.get("model", "claude-3-5-sonnet-20241022")
```

so it falls through to the hardcoded default and silently sends **every** request as `claude-3-5-sonnet-v2@20241022`, regardless of the model actually requested (confirmed in the issue with live proxy logs and a deliberately-nonsense URL model name).

## Fix

Extract the persist decision into a pure `_model_to_persist_in_body(body_model, resolved_model, *, has_backend)` helper and also write `body["model"]` when there is **no** body model and a backend will consume the body. The existing behavior (sanitizing an existing string body model in place) is unchanged. The write is gated on `self.anthropic_backend is not None`: on a native passthrough (no backend) the URL carries the model and the raw body is forwarded upstream to Vertex, which must not gain a body model it would reject, so that path is left untouched. The `"unknown"` sentinel (no model anywhere) is never stamped.

Note: this is the primary bug — the model never reached the lookup. The separate `_VERTEX_MODEL_MAP` staleness the issue also mentions is out of scope here (and partly depends on an upstream `litellm` bump).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: add the pure `_model_to_persist_in_body` helper; use it in `handle_anthropic_messages` so a URL-path (`model_override`) model with no body model is persisted into `body["model"]` on the backend path.
- `tests/test_anthropic_body_model_persist.py` (new): unit tests for each branch — in-place sanitize (backend and native), unchanged model, URL-only model persisted on the backend path, URL-only model left alone on a native passthrough, `"unknown"` not stamped, non-string resolved model left alone.

## Testing

### Test Output

```text
$ python -m pytest tests/test_anthropic_body_model_persist.py tests/test_vertex_claude_compression.py tests/test_provider_vertex_runtime.py -q
24 passed

$ uvx ruff@0.15.17 check headroom/proxy/handlers/anthropic.py tests/test_anthropic_body_model_persist.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/handlers/anthropic.py
```

## Real Behavior Proof

- Exact command / steps: exercised `_model_to_persist_in_body` for every branch, including the #2363 case `(body_model=None, resolved_model="claude-opus-4@20250514", has_backend=True)`; and ran the existing Vertex compression/runtime suites which route through `handle_anthropic_messages`.
- Observed result: the URL-only model on the backend path now returns `"claude-opus-4@20250514"` (so `body["model"]` is set and the backend forwards the real model); the same input with `has_backend=False` returns `None` (native passthrough untouched); an in-place ANSI-styled body model still sanitizes; `"unknown"` and non-string models are left alone. The old inline code returned no-write for the `body_model=None` case, which is exactly the `$0`/wrong-model path the issue reported.
- Not tested: a live `--backend litellm-vertex` request end to end against a real Vertex deployment.

## Review Readiness

## Checklist


JerrettDavis (approved):
Reviewed both changes in this branch: Vertex @YYYYMMDD pricing candidates and backend-bound Anthropic body model persistence. The body model write is correctly gated to backend paths, preserving native Vertex passthrough bodies, and the pricing candidate expansion keeps non-suffixed models' candidate lists unchanged.

Local verification:
- pytest tests/test_anthropic_body_model_persist.py tests/test_vertex_claude_compression.py tests/test_provider_vertex_runtime.py tests/test_pricing_litellm_model_resolution.py -q (after copying the local compiled headroom._core extension into the detached worktree)
- ruff check headroom/proxy/handlers/anthropic.py headroom/pricing/litellm_model_resolution.py tests/test_anthropic_body_model_persist.py tests/test_pricing_litellm_model_resolution.py

JerrettDavis (approved):
Re-reviewed at `e53c96da`. The upstream resolution preserves backend-only URL-model persistence and correctly composes the Vertex suffix candidates with the newer gateway-unwrapping pricing logic. Native passthrough remains body-faithful, regression coverage is intact, and the branch is mergeable with no failing checks.

JerrettDavis:
This branch is now fully covered by two earlier, approved PRs and can be closed as superseded. The #2363 backend-only URL-model persistence is implemented with stronger end-to-end/native-passthrough coverage in #2367, while the bundled Vertex `@YYYYMMDD` pricing-resolution change is contained in #2517 together with the related `vertex:anthropic` cache-savings fix. Keeping #2530 open would duplicate both independently maintained patches.

abhay-codes07:
Makes sense, closing as superseded. The #2363 backend-only URL-model persistence is covered with stronger end-to-end and native-passthrough tests in #2367, and the bundled Vertex @YYYYMMDD pricing change lives in #2517 alongside the related vertex:anthropic cache-savings fix. No reason to keep this open and duplicate two independently maintained patches. Thanks JerrettDavis.

PR #2540, opened 2026-07-24
fix(onnx): stop ONNX thread pools from spinning idle cores (#2495)

### LINKED ISSUE #2495: [BUG] tokensave using 100% of all cores
## Description

When using the claude extension in vscode, tokensave starts to use nearly100% of my machines cpu after a period of time.  I am not sure what triggers it.

## To Reproduce

Install and configure headroom
Use claude extension in vscode.

## Expected Behavior

tokensave not using nearly 100% of all cores

## Actual Behavior

tokensave using nearly 100% of all cores, everything slows down

## Environment

- **Headroom version**: 0.32.1
- **Python version**: 3.12.12
- **OS**: Ubuntu 26.04
- **LLM Provider**: Anthropic

## Additional Context

Claude running in vscode

Image

abhay-codes07:
## Description

Fixes #2495 (tokensave / the proxy using ~100% of all cores). ONNX Runtime's intra-op (and inter-op) thread pools **spin-wait on every core between inferences** by default. Headroom is a long-lived process that keeps ONNX models loaded — the kompress code compressor ("tokensave"), the image technique/SigLIP routers, and the memory embedder — so once a model is loaded, its idle thread pool keeps every core busy even when no compression is running. That matches the report exactly: CPU climbs to ~100% of all cores "after a period of time" and the whole machine slows down, with no obvious trigger.

`create_cpu_session_options` (the shared factory every CPU ONNX session goes through) configured threads and the memory arena but never touched spinning, so ORT's default (spin enabled) was in effect everywhere.

## Fix

Disable intra-op and inter-op thread spinning in `create_cpu_session_options` so idle ORT threads block instead of spin-waiting. This applies to every ONNX session built through the factory (kompress + the image routers). It:

- is **best-effort per key** (wrapped in try/except) so an older ORT build that doesn't recognize a config key still creates a session;
- is **overridable** via `HEADROOM_ONNX_ALLOW_SPINNING=1` for a dedicated/batch box that wants ORT's peak-throughput spinning;
- does not change active-inference throughput meaningfully — blocking threads wake on new work with only microsecond-scale latency, which is the recommended setting for a server/proxy with idle periods.

The memory embedder already builds its own options with `intra_op_num_threads=1`; this change is orthogonal and additionally quiets its idle spinning if it were ever routed through the factory.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/onnx_runtime.py`: add `ONNX_ALLOW_SPINNING_ENV` + `onnx_thread_spinning_enabled()`; disable `session.intra_op.allow_spinning` / `session.inter_op.allow_spinning` in `create_cpu_session_options` unless spinning is explicitly re-enabled.
- `tests/test_onnx_runtime.py`: spinning is disabled by default (both keys), `HEADROOM_ONNX_ALLOW_SPINNING=1` re-enables it, an explicit `0` disables it, and a config key an older ORT rejects doesn't break session creation.

## Testing

### Test Output

```text
$ python -m pytest tests/test_onnx_runtime.py -q
11 passed

# with the fix reverted the new symbols don't exist, so the spinning tests
# fail at import — the pre-fix factory left ORT's spinning at its (enabled) default

$ uvx ruff@0.15.17 check headroom/onnx_runtime.py tests/test_onnx_runtime.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/onnx_runtime.py
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12, project venv (`uv sync --extra proxy`, onnxruntime 1.23.2 installed), `uvx ruff@0.15.17` / `uvx mypy@1.20.2`, pytest in the venv.
- Exact command / steps: built a real `onnxruntime.SessionOptions` via `create_cpu_session_options(ort)` and read back `session.intra_op.allow_spinning` / `session.inter_op.allow_spinning`; repeated with `HEADROOM_ONNX_ALLOW_SPINNING=1`.
- Observed result: by default both keys read back `"0"` (spinning disabled); with `HEADROOM_ONNX_ALLOW_SPINNING=1` neither key is set (ORT's default spinning restored). Against a real ORT the pre-fix factory set neither key, so ORT's default (spinning enabled) applied — the idle all-cores burn. Ran against the actual module and real onnxruntime.
- Not tested: a live multi-hour VS Code + Claude session measuring CPU before/after (the spinning-disable is the documented ORT remedy for idle-CPU in a long-lived process; the config change itself is verified end to end against real ORT).

## Review Readiness

## Checklist


JerrettDavis (approved):
Reviewed the ONNX session-options change and its tests. Disabling intra/inter-op spinning in the shared CPU session factory covers the kompress, image-router, and memory-embedding call sites while preserving an explicit HEADROOM_ONNX_ALLOW_SPINNING opt-in and best-effort behavior for older ORT builds.

Local verification:
- pytest tests/test_onnx_runtime.py -q
- ruff check headroom/onnx_runtime.py tests/test_onnx_runtime.py

JerrettDavis (approved):
Re-reviewed at `75464d54`. The only contribution delta from the previously approved revision is import/blank-line formatting in the runtime test; the ONNX idle-spin fix and its assertions are otherwise unchanged. The branch is mergeable and current checks have no failures.

Issue #2195, opened 2026-07-14
[BUG] Proxy memory is non-functional for Claude Code clients on 0.31.0 - empty retrieval query (fix attached), search never reaches any backend, and no write path


dinesh01ameh:
# [BUG] Proxy memory is non-functional for Claude Code clients on 0.31.0 — empty retrieval query (fix attached), search never reaches any backend, and no write path

## Environment
- headroom-ai **0.31.0** via `pip install "headroom-ai[proxy,memory]==0.31.0"` (sentence-transformers present and working; findings 1–2 also reproduced on 0.30.0 where noted)
- Proxy in Docker (`python:3.13-slim`), Windows host: `headroom proxy --host 0.0.0.0 --port 8787 --lossless --no-rate-limit --memory --memory-db-path /data/memory/memory.db`
- Client: Claude Code 2.1.205 (headless `claude -p` and interactive), routed via `ANTHROPIC_BASE_URL`
- `/health` reports `memory: {"enabled": true, "ready": true, "status": "healthy", "backend": "local", "initialized": true, "native_tool": false, "bridge_enabled": false}`

## TL;DR
End to end, memory is a silent no-op for Claude Code clients on 0.31.0 — three independent causes, each verified in isolation:

| # | Finding | Effect |
|---|---|---|
| 1 | `MemoryQuery.from_messages` drops user text for Anthropic list-shaped user messages (**fix attached**) | retrieval query is always empty |
| 2 | Even with a correct query, `search_and_format_context` never reaches any backend DB; the exception is swallowed and logged nowhere visible | context injection never fires |
| 3 | 0.31.0 injects no memory tools into these sessions (0.30.0 injected five custom ones) | no write path either |

---

## Finding 1 — `from_messages` drops user text for list-shaped user messages (fix attached)

`headroom/proxy/memory_query.py::MemoryQuery.from_messages` never sets `user_text` when the latest user message's `content` is a **list of blocks** — the standard Anthropic / Claude Code shape. The `isinstance(content, list)` branch only harvests `tool_result` blocks; plain `{"type": "text"}` blocks are ignored. Only the `isinstance(content, str)` (OpenAI-style) branch sets `latest_user`:

```python
if role == "user":
    if isinstance(content, list):
        for block in content:
            if isinstance(block, dict) and block.get("type") == "tool_result":
                ...  # collected
        # plain text blocks: never harvested  <-- BUG
    elif isinstance(content, str):
        if not latest_user:
            latest_user = content
```

[...]

PR #2541, opened 2026-07-24
fix(memory): skip <system-reminder> blocks when building the retrieval query (#2195)

abhay-codes07:
## Description

Addresses #2195 Finding 1. `extract_memory_query_sources` (the memory retrieval query builder) was extended to harvest text blocks from Anthropic list-shaped user turns — the standard Claude Code shape — but it joins **every** text block in the turn. Claude Code appends `<system-reminder>` harness blocks to essentially every user turn, so those get concatenated into the embedding input alongside the real question.

Per the reporter's measurements (`all-MiniLM-L6-v2`): the clean question scored top cosine **0.748** against a stored memory; the same question wrapped in harness boilerplate scored **0.232**. The default `memory_min_similarity` floor is **0.3**, so the diluted query falls under the floor and **nothing is retrieved** — memory silently no-ops for Claude Code clients. The reporter explicitly warned that a naive "concatenate all text blocks" harvest would still retrieve nothing, which is exactly the current behavior.

## Fix

Filter out text blocks whose text starts with `<system-reminder` when building `user_text`, so the retrieval query keys on the substantive question and the embedding isn't diluted by harness boilerplate. A turn that is only a system-reminder yields no `user_text` (as before). All other harvesting (tool_result blocks, OpenAI string content, assistant/tool context) is unchanged.

Note: the reporter also asked to expose `memory_min_similarity` as an env var / CLI flag (it lives on `ProxyConfig` with no surface today). That is a sensible companion but is a separate config-plumbing change; I kept this PR focused on the retrieval-query bug so it stays easy to review, and I'm happy to follow up with the env/CLI surface.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/memory_query_policy.py`: in `extract_memory_query_sources`, skip `<system-reminder>` text blocks when assembling the user query from a list-shaped Anthropic user turn.
- `tests/test_memory_query_policy.py`: regressions that a system-reminder block is excluded (real question kept) and that a reminder-only turn yields no user text.

## Testing

### Test Output

```text
$ python -m pytest tests/test_memory_query_policy.py -q
7 passed

# with the fix reverted, the two new tests fail: the system-reminder text is
# concatenated into user_text (the diluted-query behavior)

$ uvx ruff@0.15.17 check headroom/proxy/memory_query_policy.py tests/test_memory_query_policy.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/memory_query_policy.py
```

## Real Behavior Proof

- Exact command / steps: called `extract_memory_query_sources` with a Claude Code-shaped user turn (real question text block + an appended `<system-reminder>` text block), and with a reminder-only turn; then reverted the source and re-ran.
- Observed result: with the fix `user_text` is exactly `"how do I add caching to the auth handler?"` (no `system-reminder` substring), and a reminder-only turn yields `""`; with the fix reverted `user_text` includes the full `<system-reminder>...</system-reminder>` text (the diluted embedding input). Ran against the actual module.
- Not tested: an end-to-end embedding + backend retrieval against a live memory DB measuring the cosine recovery (the dilution figures are the reporter's; this change removes the boilerplate from the query text that produces them).

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/memory_query_policy.py
+++ b/headroom/proxy/memory_query_policy.py
@@ -63,11 +63,18 @@ def extract_memory_query_sources(
                     # Anthropic user turns carry the actual prompt as text blocks
                     # ({"type":"text","text":...}), not a plain string. Capture it
                     # so the memory retrieval query keys on the user's question and
-                    # not just any tool_result blocks in the same turn.
+                    # not just any tool_result blocks in the same turn. Skip Claude
+                    # Code's <system-reminder> harness blocks: they are appended to
+                    # the same turn and, concatenated into the embedding input,
+                    # dilute the user's question below the similarity floor so
+                    # nothing is retrieved (#2195). Filtering them keeps the query
+                    # on the substantive question.
                     user_text = "\n".join(
-                        b.get("text", "")
+                        text
                         for b in content
                         if isinstance(b, dict) and b.get("type") == "text"
+                        for text in (str(b.get("text", "")).strip(),)
+                        if text and not text.startswith("<system-reminder")
                     ).strip()
                     if user_text:
                         latest_user = user_text
--- a/tests/test_memory_query_policy.py
+++ b/tests/test_memory_query_policy.py
@@ -69,6 +69,42 @@ def test_extract_sources_captures_anthropic_user_text_blocks() -> None:
     assert user_text == "help me refactor auth"
 
 
+def test_extract_sources_skips_system_reminder_blocks() -> None:
+    """Claude Code appends <system-reminder> harness blocks to the user turn.
+    Concatenated into the embedding input they dilute the real question below the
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed the memory query extraction change. Filtering Claude Code <system-reminder> text blocks at extraction time keeps the embedding query focused on the substantive user text while preserving the existing tool-output and assistant-context collection behavior. The reminder-only regression also covers the empty-query edge case.

Local verification:
- pytest tests/test_memory_query_policy.py -q
- ruff check headroom/proxy/memory_query_policy.py tests/test_memory_query_policy.py

PR #2553, opened 2026-07-25
fix(proxy/hooks): only run stream-safe turn hooks on the OpenAI Responses path (#2549)

abhay-codes07:
## Description

Follow-up to #2549, which closed the turn-hook gap on the streamed OpenAI **chat** path. The same seam is still open on the OpenAI **Responses** path.

All three Responses entry points — `handle_openai_responses` (HTTP `/v1/responses`), `handle_openai_responses_ws` (the WS transport), and `_maybe_compress_passthrough_responses` — run turn hooks through `_compress_openai_responses_payload`, which called:

```python
run_request_hooks(_req_ctx)   # unfiltered: every hook's on_request runs
```

But **none** of the Responses paths ever call `run_response_hooks` — only `handle_openai_chat` does. So a buffered-only hook (one that re-drives the model in `on_response`: defer a tool, then reload it when the model asks for it) had its `on_request` applied on the Responses path with **no `on_response` to follow**, leaving the turn half-transformed (e.g. a tool deferred and never restored). This is the same class of bug #2549 fixed for streamed chat, where an on_request without a possible on_response is unsafe.

## Fix

Restrict the Responses `on_request` to stream-safe (fold-only) hooks — those whose `on_request` stands alone with no `on_response` re-drive:

```python
run_request_hooks(_req_ctx, stream_safe_only=True)
```

This is exactly the guard #2549 added for the streamed chat path (`stream_safe_only=stream`); the Responses path uses `True` unconditionally because it never runs `on_response` at all. Fold-only hooks (e.g. the lossless-guard fold) declare `stream_safe = True` and are unaffected; only a re-drive hook is now correctly skipped on Responses instead of being half-applied. No behavior change until a non-stream-safe hook is registered.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/openai.py`: pass `stream_safe_only=True` to `run_request_hooks` in `_compress_openai_responses_payload`, with a comment explaining the Responses path never runs `on_response`.
- `tests/test_openai_responses_compression_units.py`: register a stream-safe fold hook and a buffered-only re-drive hook, run the Responses payload compressor, and assert only the fold hook's `on_request` was applied.

## Testing

### Test Output

```text
$ python -m pytest tests/test_openai_responses_compression_units.py -q
38 passed

# with the fix reverted, the new test fails: the buffered-only hook's
# on_request ("redrive_ran") is applied on the Responses path

$ uvx ruff@0.15.17 check headroom/proxy/handlers/openai.py tests/test_openai_responses_compression_units.py
All checks passed!
```

Note: `tests/test_turn_hooks.py` has 6 `@pytest.mark.asyncio` `on_response` tests that fail locally because pytest-asyncio isn't configured in this environment; they fail identically on clean `main` and are unrelated to this change (it doesn't touch `turn_hooks.py`).

## Real Behavior Proof

- Exact command / steps: registered a `stream_safe=True` fold hook and a default (buffered-only) re-drive hook, both mutating `ctx.tools` in `on_request`; called `_compress_openai_responses_payload` on a Responses payload; then reverted the source and re-ran.
- Observed result: with the fix the returned payload's tools contain the fold hook's marker but not the re-drive hook's; with the fix reverted both markers appear (the re-drive hook is half-applied, since the Responses path never runs its `on_response`). Ran against the actual handler.
- Not tested: a live `/v1/responses` request end to end with a registered re-drive hook.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -2384,7 +2384,14 @@ class OpenAIHandlerMixin:
                 tools=working.get("tools"),
                 config=getattr(self, "config", None),
             )
-            run_request_hooks(_req_ctx)
+            # The Responses path (HTTP /v1/responses, WS, and passthrough) never
+            # runs on_response hooks — only handle_openai_chat does. So a
+            # buffered-only hook that re-drives the model in on_response (defer a
+            # tool, reload it when asked) would have its on_request applied here
+            # but its on_response never fire, leaving the turn half-transformed.
+            # Restrict to stream-safe (fold-only) hooks, whose on_request stands
+            # alone — the same guard the streamed chat path uses (#2549).
+            run_request_hooks(_req_ctx, stream_safe_only=True)
             if _req_ctx.tools is not working.get("tools"):
                 working["tools"] = _req_ctx.tools
             # A hook may also fold the messages (replace or in-place). Write back a
--- a/tests/test_openai_responses_compression_units.py
+++ b/tests/test_openai_responses_compression_units.py
@@ -27,6 +27,46 @@ def _handler_with_router(router: ContentRouter) -> OpenAIHandlerMixin:
     return handler
 
 
+def test_responses_payload_runs_only_stream_safe_hooks_on_request():
+    """The Responses path never runs on_response hooks, so a buffered-only
+    (re-drive) hook must not have its on_request applied there — it would leave
+    the turn half-transformed. Only fold-only (stream_safe) hooks run, matching
+    the streamed chat path (#2549)."""
+    from headroom.proxy.turn_hooks import clear_turn_hooks, register_turn_hook
+
+    class Fold:
+        name = "fold"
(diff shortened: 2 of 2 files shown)


abhay-codes07:
Closing this — it's the wrong fix. CI (test (3)) correctly flagged that the existing `test_responses_turn_hook_message_fold_is_applied_and_counted` registers a fold-only hook (on_request only, no on_response) that does NOT set `stream_safe`, and the Responses path is intended to run it. My `stream_safe_only=True` over-filters: it would disable legitimate fold-only hooks that simply didn't set the flag, not just re-drive hooks.

The underlying observation still holds (the Responses path never runs on_response, so a re-drive hook's on_request stands alone there), but the right remedy — if any — would be to run on_response on the Responses path, not to filter on_request. That's a much larger change and clearly out of scope for a drive-by, so I'll leave the current design as-is. Apologies for the noise.

PR #2556, opened 2026-07-25
feat(proxy/cache): prefix-stability defaults, per-TTL write pricing, and segment-level cache-miss attribution

chopratejas:
## Description

Cache **writes** dominate Claude Code's input spend, and most of that comes from a small number of full-prefix invalidations — not from large tool outputs. Measured across 35 local Claude Code sessions (23,018 turns, 8,985M prompt tokens):

| | tokens | share | cost-equiv (read 0.1x, 5m write 1.25x) |
|---|---|---|---|
| cache reads | 8,425M | 93.8% | 843 units |
| cache writes | 569M | 6.3% | **711 units** |
| uncached input | ~2M | 0.02% | 2 units |

Writes are ~46% of input spend from 6.3% of tokens. Drilling in: warm turns (<=5 min after the previous turn) have a p50 write of 1,661 tokens, but **714 of them each write >100K**, carrying **321M tokens = 83% of all warm-path writes ~= 26% of total input spend**. 708 of those are *fully cold* (cache_read <50K, avg 452K written) — tools/system-tier invalidation, not TTL expiry (gap <5 min) and not model switches (3 events). For contrast, the entire tool_result corpus is ~0.4% of prompt tokens (p99 = 8,677 chars, well under Claude Code's 25K/30K caps).

So: **attribute** those invalidations so they stop being unexplained, **price** cache writes correctly, and offer an **opt-in** way to prevent the leading suspected cause.

Builds on #2555 (merged) — rebased onto `main` at cb8f4b64.

Closes #

## Type of Change

- [x] Performance improvement

## Changes Made

**1. `feat(proxy/cache)` + **2.** `fix(proxy/cache)` — name WHICH cache-key segment changed**
- `reason=prefix_change` said the bytes moved but never where. Now reports `tools` / `system` / `messages` — three different causes, three different fixes.
- **Fixes a real misclassification:** `prefix_changed` was computed only from forwarded *messages*, but the cache key is the whole prefix. A changed tool schema or system prompt invalidates it while `messages` stays byte-identical, so every tools/system change was filed as `unknown` — and tools render at position 0, making that simultaneously the class that invalidates the most and the one least explained.
- Wired on **all four** Anthropic-side call sites: non-streaming (x2), direct SSE, and Bedrock streaming. Claude Code streams by default, so the streaming sites are the ones that carry the traffic.
- Segment hashes are **dropped** when a call can't supply segments, so a mixed session can't report a change on the wrong turn. A gap in a diagnostic beats a false positive.

**3. `fix(proxy/cost)` — price cache writes per TTL bucket**
- Anthropic bills a 1h-TTL cache write at **2.0x** base input; 5m costs 1.25x. `_CACHE_ECONOMICS` carried a single flat 1.25x, so any session on `ENABLE_PROMPT_CACHING_1H` had its write premium under-reported by 60%.
- New `cache_write_premium_usd()` prices each bucket at its own rate; unbucketed tokens bill at 5m, so a provider reporting no split behaves exactly as before. Bucket clamped to the total.
- Reporting accuracy only — this does **not** change what TTL we request; the proxy still preserves the client's marker verbatim.

**4. `feat(wrap/claude)` + **5.** `fix(wrap/claude)` — opt-in prefix-stability, and close the claude env gap**
- `wrap claude` does not route through `_launch_tool`, so it never received the reduce-at-source env defaults every other wrapped tool gets. #2548's commit message claimed a single chokepoint; the flagship launcher was the exception. Fixed.
- `HEADROOM_WRAP_PREFIX_STABLE=1` (**opt-in, default off**) sets `CLAUDE_CODE_SYNC_SKILLS`/`SYNC_PLUGINS` so skills/plugins load at startup instead of entering the prompt mid-session. Ordering only — nothing removes a capability. Off by default because the cost is certain (slower launch; CC's own budgets there are 30s install / 60s download-stall) while the benefit is still inferred by elimination. The segment attributor is what should promote it to a default.
- `HEADROOM_WRAP_TRIM_PROMPT=1` (opt-in) for the system-prompt trims. These hide real capabilities from the model — the class the quiet-CLI defaults deliberately refuse to touch — so they are never implicit.
- Both flags registered in `settings_store.SETTINGS` (57 -> 59) so they are discoverable in the settings UI/API. `HEADROOM_WRAP_QUIET` is the cautionary precedent — still unregistered and invisible.
- Vertex/Foundry: note that Headroom's server-side tool-search injection is first-party only, so on those upstreams `ENABLE_TOOL_SEARCH` governs only Claude Code's *local* deferral. The old banner implied both halves were active.

**Net behavior change for someone who opts into nothing:** more accurate cache-miss attribution in logs/metrics, and corrected write-premium reporting. No request-path behavior change.

## Cross-harness / cross-provider scoping

Deliberate, since `classify_cache_miss` is **shared** by the Anthropic and OpenAI handlers (Codex / Kimi / GLM / OpenCode ride the latter), while `normalize_message_cache_control` is Anthropic-only.

| Change | Scope | Effect elsewhere |
|---|---|---|
| `CLAUDE_CODE_*` env defaults | Claude Code launcher only, opt-in | none — Codex/OpenCode/Kimi untouched; symbols named for the client |
| `write_multiplier_1h` | keyed by **provider** (anthropic/bedrock) | openai/gemini have no key and fall back to their 5m rate; their responses carry no `ephemeral_1h` field, so the bucket is 0 and numbers are unchanged |
| segment attribution | provider-**neutral** mechanism | `segments=` is an open ordered mapping; each handler owns its own adapter. Only the Anthropic adapter is wired; omitting `segments` preserves prior behavior, so OpenAI/Kimi paths are unchanged (and now explicitly clear their baseline rather than hold a stale one) |

`anthropic_cache_segments` lives in `proxy/helpers.py` (which already hosts the Anthropic/Claude-Code-shaped helpers and is imported by both handlers) — no new module, no import cycle between sibling mixins, and `prefix_tracker` stays provider-neutral. `segment_fingerprint()` strips `cache_control` (clients re-mark the breakpoint every turn; Headroom re-places message markers too) and sorts keys, so marker movement and key ordering never read as content changes. Hashes only — no second copy of tool schemas retained. Values are the **post-mutation forwarded** body, so Headroom's own tool injection / system compaction is attributed to Headroom, not blamed on the client.

## Testing

### Test Output

```text
$ python -m pytest tests/test_cache_miss_segment_attribution.py tests/test_cache_write_premium_ttl.py \
    tests/test_wrap_prefix_stability.py tests/test_wrap_quiet_cli.py tests/test_cache/ \
    tests/test_proxy_cache_ttl_metrics.py tests/test_backend_streaming_cache_metrics.py \
    tests/test_bedrock_streaming_input_tokens.py tests/test_issue_746_tool_search.py \
    tests/test_proxy_settings_endpoints.py -q
373 passed, 2 skipped, 2 warnings in 28.76s

$ python -m pytest tests/test_backend_streaming_cache_metrics.py tests/test_backend_nonstreaming_cache_metrics.py \
    tests/test_bedrock_streaming_input_tokens.py tests/test_bedrock_tool_result_cache_and_streaming_stats.py \
    tests/test_issue_746_tool_search.py tests/test_openai_streaming_backend.py \
    tests/test_openai_tool_search_deferral.py tests/test_anthropic_beta_session_sticky.py \
    tests/test_cache_control_move_bust.py -q
123 passed, 3 skipped in 13.54s

$ ruff check headroom/ --exclude headroom/dashboard
All checks passed!

$ mypy headroom/settings_store.py headroom/cli/wrap.py headroom/cache/prefix_tracker.py \
    headroom/proxy/helpers.py headroom/proxy/handlers/streaming.py headroom/proxy/handlers/anthropic.py
```

42 new tests across 3 files (15 segment attribution + 8 write premium + 9 wrap, plus 12 pre-existing wrap/quiet cases re-run).

## Real Behavior Proof

- **Environment:** macOS 25.4.0, Python 3.12 (`.venv`), branch `tejas/cache-prefix-stability` on `main` @ cb8f4b64 (#2555 squash-merged). All test runs performed in an isolated `git worktree` with `PYTHONPATH` pinned to it, so nothing resolved against the primary checkout.
- **Exact command / steps:**
  1. Measured real traffic: parsed 35 `~/.claude/projects/**/*.jsonl` transcripts (23,018 turns with `usage`), bucketing `cache_creation_input_tokens` by inter-turn gap and prompt size.
  2. Verified no import cycle by importing `proxy.helpers`, `handlers.streaming`, `handlers.anthropic` together.
  3. Ran the suites above.
- **Observed result:**
  - 97.6% of turns are <5 min apart; 1.8% are 5m–1h; 0.6% >1h.
  - 714 warm turns wrote >100K each = 321M tokens (83% of warm-path writes); 708 fully cold at avg 452K.
  - 0 of 32,152 messages had >20 content blocks (so the 20-block cache lookback window is not a factor here).
  - Segment attribution unit-verified for: tools change, system change, earliest-segment-wins when both move, messages fallback, first-sighting cannot blame, segment-blind call drops baseline, TTL-expiry names nothing, hit names nothing, OpenAI tools-only shape, moved `cache_control` not a change, key reordering not a change, unserializable value does not raise.
- **Not tested:**
  - **No live API call anywhere in this PR.** The write-premium arithmetic is unit-tested, not validated against a real Anthropic invoice. Partner pricing on bedrock (and vertex/foundry, which fall through to the anthropic entry) inherits the pre-existing first-party assumption and is unverified.
  - Segment attribution is verified at the unit level and wired at all four Anthropic call sites, but **not yet observed firing against a live session** — the value of `changed_segment` on real traffic is the next thing to look at, and is what would justify promoting the prefix-stability flag to a default.
  - The 708 fully-cold events are attributed to the tools/system tier **by elimination**; transcripts carry no `tools`/`system`, so the specific mutating field is unproven. That is precisely what this PR instruments.
  - `CLAUDE_CODE_SYNC_SKILLS`/`SYNC_PLUGINS` are verified to be injected into the child env; the resulting reduction in mid-session mutation is not measured end-to-end. The added startup latency is reasoned from Claude Code's documented timeouts, not benchmarked.
  - OpenAI/Kimi/Gemini segment adapters deliberately not wired, so segment attribution is unexercised on those paths (they keep prior behavior).

## Review notes (self-review found two defects, both fixed in-branch)

1. Segment attribution was originally wired only on the **non-streaming** sites. Claude Code streams by default, so the headline feature was dormant on the real path — and hashes going stale across segment-blind turns could have reported a change on the wrong turn. Commit `d35606c6` wires all four sites and drops the baseline rather than holding a stale one.
2. The prefix-stability defaults originally shipped **default-on**, putting a guaranteed startup-latency cost behind a benefit that is still inferred by elimination. Commit `9981733d` makes them opt-in and registers both flags for discoverability.

## Review Readiness

## Checklist


JerrettDavis (approved):
I reviewed the prefix tracker baseline/update logic, the Anthropic segment adapter wiring on non-streaming and streaming paths, the per-TTL write-premium math, and the opt-in wrapper/settings registration. The behavior is scoped as described: no default prefix-stability env injection unless HEADROOM_WRAP_PREFIX_STABLE is truthy, and OpenAI/Gemini keep the flat write behavior.

Local verification passed: new/focused cache-cost-wrap suites (37 passed), provider/cache streaming subset (98 passed, 3 skipped), settings/cache TTL/prefix/provider cache subset (147 passed), and ruff on the touched files. The broader tests/test_cache/ run hit a Windows-only SQLite temp path failure in test_client_integration.py (`C:`Users...), which is outside the touched paths; GitHub CI is green.

JerrettDavis (approved):
Resolved the current-main conflict by retaining the PR’s cache-segment imports while accepting main’s removal of the obsolete subprocess helper. Validation: 382 targeted tests passed (2 skipped), Ruff passed, and the lockfile/diff checks were clean.

PR #2559, opened 2026-07-25
Keep Responses memory tools working under store=false (Codex/ChatGPT backend)

sangemaru:
## Problem

Responses memory tools are skipped whenever a request carries `store=false`
(`_allow_responses_memory_tools`, `_responses_request_allows_memory_tool_continuation`).
That is **every** Codex turn against the chatgpt.com backend, which rejects
`store=true` outright with:

```
HTTP 400 {"detail":"Store must be set to false"}
```

So on that backend, memory is effectively off.

## Why the restriction existed

It came from the *continuation shape*, not from memory. After executing an
injected memory tool, Headroom returned the result through a
`previous_response_id` continuation, which requires the originating response to
be stored. `store=false` and `previous_response_id` are mutually exclusive, so
memory tools were disabled rather than left broken.

## Change

Add a stateless continuation for that case. Instead of addressing a stored
response, resend the history: the already-compressed `body["input"]`, the prior
output echoed verbatim, then the `function_call_output` items.

- **Compression is preserved** — the compressed payload is exactly what gets
  resent, so this is not a fallback to uncompressed history.
- **No server-side state needed**, so `store=false` is honoured as sent.
- This mirrors the existing CCR Responses continuation, which already does
  `{**body, "input": items}` + `pop("previous_response_id")`.
- Prior output is echoed **in order**, because a `function_call` replayed
  without its preceding `reasoning` item is rejected by reasoning models.

The WebSocket turn loop still continues via `previous_response_id`, so it keeps
the API-key-only rule. The new `stateless_continuation` flags mark exactly which
call sites resend history, so the two strategies stay explicit rather than
implicit.

Also guards continuation adoption on HTTP 200. Previously a failed
continuation's error body replaced the response wholesale; now it is logged and
the injected memory `function_call` items are stripped, so the client never
receives an unresolved call for a tool it cannot see.

## Contract tests updated

Two tests asserted the old behaviour and now assert the new one:

- `test_handle_openai_responses_chatgpt_auth_skips_memory_tools` →
  `..._keeps_memory_tools`
- `test_handle_openai_responses_api_auth_store_false_skips_memory_tools` →
  `..._keeps_memory_tools`

Both still assert `store is False` is honoured — that part of the contract is
unchanged and load-bearing.

## Tests

7 new unit tests cover both continuation shapes, the string-`input` shorthand,
the no-`response_id` fallback, and that `store` is never forced to `true` when
the continuation is stateless.

`pytest -k "openai or codex or memory or responses"` → **2192 passed**, 89
skipped. The one failure,
`test_cli/test_recover_codex.py::test_recovery_records_sockets_and_secures_both_backups`,
**also fails on clean `main`** and is unrelated.


JerrettDavis (changes requested):
I found one blocker in the new failed-continuation fallback. When the memory continuation returns a non-200, the handler logs and removes injected memory function_call items from resp_json (headroom/proxy/handlers/openai.py:5453), but the normal non-streaming return path still sends response.content from the original upstream response (headroom/proxy/handlers/openai.py:5586). That means the client can still receive the unresolved Headroom-injected memory call even though the PR explicitly tries to strip it. Please rebuild response from the sanitized resp_json on this path, or return a synthesized JSON response directly, and add a regression where the initial Responses reply contains a memory tool call, the continuation returns 500, and the final client body no longer contains that function_call.

Focused local verification otherwise passed after copying the built Rust extension into the worktree: pytest tests/test_openai_responses_context_compaction.py tests/test_openai_codex_routing.py -q (41 passed) and ruff on the touched files.

JerrettDavis:
Exact-head follow-up at d81ebab1: the refreshed PR now has zero changed files against main, so the earlier continuation-sanitization finding no longer applies to a submitted delta. This branch has been fully absorbed and should be closed as obsolete rather than merged.

JerrettDavis (approved):
Reviewed exact head d81ebab1fb02e16885e9f957dc224b2941e87048. This head is now an ancestor of current main (0 commits ahead; the PR diff is empty), so there is no remaining change to merge or test. The PR can be closed as already integrated/superseded.

JerrettDavis:
Correction to my approval wording: this head is not a graph ancestor of current main; it is two commits ahead of its merge base. However, its merge-base-to-head content diff is empty (GitHub also reports zero changed files), so the substantive conclusion is unchanged: there is no patch left for this PR to merge, and it should be closed as already integrated/superseded.

JerrettDavis:
Thanks for working through the Responses memory-tool behavior. I checked the current branch again: GitHub reports zero changed files, and there’s no remaining patch against its merge base. Closing this as already absorbed so it doesn’t keep taking up a review slot.

PR #2534, opened 2026-07-24
perf(code_compressor): reduce AST compression latency

inix-x:
## Description

Reduce AST compression latency for large source files while preserving the observed compressed output, token counts, syntax status, and symbol scores. Repeated Python tree walks and per-symbol source scans made this path expensive as files grew.

This isolates the AST processing improvement from a broader latency investigation. Model fallback, concurrency tuning, dashboard changes, and production monitoring remain outside this PR.

## Type of Change

- [x] Performance improvement

## Changes Made

- Collect definitions and identifiers in tree-sitter, preserving source ordering and nested/duplicate symbol behavior.
- Reuse source lines and qualified-call lookups during extraction; use tree-sitter's propagated syntax-error flag.

## Testing

- [x] Unit tests pass (160 targeted code-compressor tests; full suite not claimed)
- [x] Linting passes (`uv run ruff check .`)
- [ ] Type checking passes (`uv run mypy headroom`; existing LangChain error noted below)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ uv run pytest -q tests/test_transforms/test_code_compressor.py tests/test_code_aware_brace_comment_regressions.py tests/test_code_compressor_language_alias.py tests/test_code_compressor_bash.py tests/test_code_compressor_syntax_breaker.py tests/test_code_compressor_thread_safety.py tests/test_transforms/test_code_compressor_cjk.py
160 passed in 34.21s
$ uv run ruff check .
All checks passed!
$ uv run ruff format --check .
3383 files already formatted
$ cargo test --workspace  # ORT_DYLIB_PATH points to installed ONNX Runtime 1.29.0
1493 passed across Rust test binaries; 3 ignored
$ make ci-precheck-python
175 passed, 4 skipped, 1 warning in 5.35s
$ .venv/bin/python -m pytest -q tests/test_code_aware_regressions.py
5 passed in 2.24s
$ npx --yes --package=@commitlint/cli --package=@commitlint/config-conventional -- commitlint --from upstream/main --to HEAD --config .commitlintrc.json
(exit 0)
```

The Unicode regression fails against pristine upstream `f302a38e`: `_slice_code_bytes` is called 20 times instead of zero. It passes after the change. Installed commit hooks, including their isolated mypy environment and commitlint, pass.

## Real Behavior Proof

[...]

PR #2529, opened 2026-07-24
fix(pricing): resolve Vertex model names with an @YYYYMMDD version suffix (#2515)

abhay-codes07:
## Description

Fixes #2515. When Claude Code routes through Vertex AI, model names arrive with a date-stamped version suffix: `claude-opus-4@20250514`, `claude-haiku-4-5@20251001`, `claude-opus-4-1@20250805`. `resolution_candidates()` and `pricing_lookup_candidates()` in `litellm_model_resolution.py` only tried the verbatim name plus the `anthropic/` / `openai/` / ... provider prefixes. None of those key into LiteLLM's cost DB (which stores `claude-haiku-4-5`, `vertex_ai/claude-opus-4@20250514`, ...), so every versioned Vertex model resolved to nothing and priced at **$0.00**:

- `cost.stats()` `per_model` shows `input_cost_usd: $0.00` for all versioned Vertex models
- the dashboard shows `Net savings: $0.00` even though `savings_usd` is computed correctly

## Fix

Strip the `@YYYYMMDD` suffix and extend the candidate lists (in both resolution and pricing lookup) with:

- `vertex_ai/<original-with-suffix>` (Vertex keeps the suffix on its own key, e.g. `vertex_ai/claude-opus-4@20250514`),
- the version-stripped **bare** name (`claude-haiku-4-5`),
- the provider-prefixed bare name (`anthropic/claude-haiku-4-5`),
- `vertex_ai/<bare>`.

When the model has no `@YYYYMMDD` suffix, `_strip_version_suffix` returns it unchanged and none of the extra candidates are added, so ordinary (non-Vertex) names keep their exact prior candidate lists. Resolution matches the issue's expected outcomes:

```
claude-haiku-4-5@20251001 -> claude-haiku-4-5
claude-opus-4@20250514    -> vertex_ai/claude-opus-4@20250514
claude-opus-4-1@20250805  -> claude-opus-4-1
claude-sonnet-4-6         -> claude-sonnet-4-6   (no suffix, unchanged)
```

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

[...]

PR #1800, opened 2026-07-04
fix(proxy): subtract cache write premiums from net savings

rodboev:
## Description

Cache stats already calculate both prompt-cache read savings and cache-write premium cost, but the exported `net_savings_usd` field used gross read savings alone. That made cache-heavy token-mode workloads look profitable even when extra cache writes offset or exceeded the read discount. This updates existing cache cost accounting so provider and total `net_savings_usd` subtract write premiums while keeping gross savings and write premium fields visible. Refs #327.

The scope follows doublefx's controlled measurement in https://github.com/headroomlabs-ai/headroom/issues/327#issuecomment-4683604089, which showed token-mode compression increasing cache write volume and billed cost while dashboard token savings looked positive.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Subtract cache write premiums from provider-level cache `net_savings_usd`.
- Subtract aggregate cache write premiums from total cache `net_savings_usd`.
- Keep gross `savings_usd` and `write_premium_usd` visible for dashboard and telemetry consumers.
- Add focused regressions for provider net, total net, and zero-write-premium preservation.
- Update the dashboard cache TTL fixture to match the corrected net value.

## Testing

- [x] Unit tests pass (`uv run pytest tests/test_proxy_cache_ttl_metrics.py tests/test_dashboard_cache_ttl_playwright.py tests/test_proxy_dashboard_stats_cache.py -q`)
- [x] Linting passes (`uv run ruff check headroom/proxy/cost.py tests/test_proxy_cache_ttl_metrics.py tests/test_dashboard_cache_ttl_playwright.py tests/test_proxy_dashboard_stats_cache.py`)
- [ ] Type checking passes (`uv run mypy headroom`)
- [x] New tests added for new functionality when applicable
- [x] Manual testing performed

### Test Output

```text
uv run pytest tests/test_proxy_cache_ttl_metrics.py tests/test_dashboard_cache_ttl_playwright.py tests/test_proxy_dashboard_stats_cache.py -q
28 passed, 2 skipped, 1 warning in 32.75s

uv run pytest tests/test_proxy_cache_ttl_metrics.py -q -k keeps_net_equal_without_write_premium
1 passed, 16 deselected in 0.15s

[...]

PR #2567, opened 2026-07-26
ci(parity): make the parity harness a real per-PR gate

chopratejas:
## Description

Three independent hardening steps on the Rust-vs-Python parity harness, ordered
smallest-blast-radius first. Together they make `make test-parity` a real gate
instead of a nightly advisory, and un-blind 6 fixtures that were being silently
ignored.

Context: the Rust port relies on the parity harness as its safety net, but the
harness was running once a night with `continue-on-error: true`, so a Rust
comparator that started diverging would not have failed anything.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- **`ef00772e` — drop the dead maturin/venv step.** `headroom-parity` has no
  `pyo3` dependency; its comparators link `headroom-core` directly and never
  cross into the interpreter (the crate manifest says as much: *"Phase 0 does not
  invoke Python from Rust"*). Yet `make test-parity` refused to run without an
  activated venv and rebuilt `headroom._core` via `maturin develop` first, and
  the CI job additionally provisioned Python 3.11 + pip + maturin +
  `pip install -e .`. All removed. `MATURIN` stays defined — `build-wheel` uses it.
- **`d9cf8ddb` — register a `text_crusher` comparator.** 6 fixtures have sat in
  `tests/parity/fixtures/text_crusher/` since the recorder landed, but
  `text_crusher` was never added to `builtin_comparators()`. `parity-run` only
  walks directories it has a comparator for, so those fixtures were not
  skipped — they were invisible, and nothing reported them missing. The API maps
  1:1, so the comparator is a straight adapter.
- **`6a1c6ab4` — promote parity to a blocking per-PR gate.** Drops
  `if: github.event_name == 'schedule'` and `continue-on-error: true`. The
  existing `paths` filter is already the right trigger set. The nightly cron
  stays as a backstop for toolchain/dependency drift that lands without touching
  a filtered path.

Deliberately **not** widening the path filter to Python paths: the fixtures are
frozen recordings of Python output and the harness never invokes Python, so it
measures Rust-vs-snapshot. A Python-side edit cannot move the result, so running
the job on Python paths would spend CI on a check that is structurally blind to
the diff. Detecting fixture staleness needs re-recording — a different mechanism,
and its own change.

## Testing

No Python source changed, so `pytest`/`mypy` are N/A for this diff. Rust gates
below are the relevant ones.

### Test Output

```text
$ cargo fmt --all -- --check
fmt OK

(no warnings, rc=0)

$ cargo test -p headroom-parity
test tests::harness_reports_diff_for_divergent_comparator ... ok
test tests::stub_comparators_skip_rather_than_panic ... ok
test tests::missing_transform_dir_yields_empty_report ... ok
test tests::harness_reports_match_for_agreeing_comparator ... ok
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ make test-parity
[log_compressor  ] total=20 matched=0  skipped=20 diffed=0
[diff_compressor ] total=27 matched=27 skipped=0  diffed=0
[cache_aligner   ] total=20 matched=0  skipped=20 diffed=0
[tokenizer       ] total=40 matched=40 skipped=0  diffed=0
[ccr             ] total=25 matched=0  skipped=25 diffed=0
[smart_crusher   ] total=17 matched=17 skipped=0  diffed=0
[content_detector] total=21 matched=21 skipped=0  diffed=0
[text_crusher    ] total=6  matched=6  skipped=0  diffed=0

total=176 matched=111 skipped=65 diffed=0   → exit 0
```

## Real Behavior Proof

- **Environment:** macOS 25.4.0 (darwin arm64), rustc 1.95.0, no venv activated.
- **Exact command / steps:**
  1. `env -u VIRTUAL_ENV make test-parity` — proves the harness runs from a bare
     checkout with no Python toolchain and no built extension module.
  2. `make test-parity` — full fixture sweep.
  3. `python3 -c "yaml.safe_load(open('.github/workflows/rust.yml'))"` — confirms
     the `parity` job has neither an `if` key nor `continue-on-error`, and that
     `push`/`pull_request`/`schedule` triggers all survive.
  4. `grep -rn parity-nightly . --exclude-dir=target --exclude-dir=.git` — empty,
     so the job rename leaves no dangling reference; no job uses `needs:`.
- **Observed result:** identical report before and after removing the maturin
  step (105 matched / 65 skipped / 0 diffed), plus 6 newly-visible
  `text_crusher` fixtures matching on the first run — so the Rust `TextCrusher`
  is already byte-identical to the recorded Python output, including the
  `unicode` fixture (CJK context, which routes through the separate CJK BM25
  branch) and `redundant` (Python's dedup fires, taking a requested 0.9 ratio
  down to an actual 0.44). Harness exit code 0.
- **Not tested:** `cargo test --workspace` was not run to completion locally
  (disk pressure on this box). It is unaffected by design — `headroom-parity` is
  a leaf crate that appears only in workspace `members`, with no `[dependencies]`
  entry anywhere, so no other crate's tests can observe this diff. CI runs the
  full workspace suite on this PR. The `cache_aligner` and `ccr` stubs are
  untouched and still report Skipped.

## Review Readiness

## Checklist

N/A — CI and harness output only.

## Additional Notes

**This gate blocks merges only once `parity` is added to the branch-protection
required-checks list for `main`.** Until then a failure shows red on the PR but
does not prevent merging. That is a repo-settings change, not a code change.

**Why it is safe to harden now rather than after the stubs are filled in:**
`parity-run` sets `any_diffs` only inside the diffed loop and exits non-zero
solely on that flag, so the 65 fixtures still served by `stub_comparator!`
(`log_compressor`, `cache_aligner`, `ccr`) report as `Skipped` and cannot turn
the job red. Measured on this branch: 111 matched / 65 skipped / 0 diffed.

**Documentation checkbox left unchecked:** `REALIGNMENT/11-phase-I-test-infra.md`
describes PR-I6 as blocked on I5. The exit-code logic above shows it was not.
I left the phase doc alone rather than editing planning history in a CI PR;
happy to follow up if you'd prefer it corrected.

Follow-ups, not in this PR:
- `log_compressor` comparator (20 fixtures) — in progress on a branch.
- `codex_openai_contracts/` (1 fixture) is still uncovered; it needs a real
  comparator, not just registration.
- `cache_aligner` needs `volatile_detector.rs`, which lives in `headroom-proxy`
  while `headroom-parity` depends only on `headroom-core`. `ccr` fixtures compare
  `ccr_retrieve` tool-definition injection, which has no Rust port at all.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -164,18 +164,9 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: 1.95.0
-      - uses: actions/setup-python@v6
-        with:
-          python-version: '3.11'
       - uses: Swatinem/rust-cache@v2
-      - name: Install deps
-        run: |
-          python -m venv .venv
-          source .venv/bin/activate
-          pip install --upgrade pip
-          pip install maturin
-          pip install -e .
+      # No Python toolchain: headroom-parity links headroom-core directly and
+      # never crosses into the interpreter, so the venv + maturin + `pip
+      # install -e .` setup this job used to do was pure overhead.
       - name: Run parity harness
-        run: |
-          source .venv/bin/activate
-          make test-parity
+        run: make test-parity
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,7 +21,9 @@ on:
       - 'Makefile'
       - '.github/workflows/rust.yml'
   schedule:
-    # Nightly parity run at 07:17 UTC (weekdays only). Phase 0 allows failure.
+    # Nightly parity run at 07:17 UTC (weekdays only). Redundant with the
+    # per-PR gate below, but catches drift from toolchain/dependency updates
+    # that land without touching any filtered path.
     - cron: '17 7 * * 1-5'
 
 concurrency:
@@ -154,11 +156,19 @@ jobs:
         continue-on-error: true
         run: cargo deny check licenses
 
-  parity-nightly:
-    name: parity (nightly, allowed to fail during Phase 0)
-    if: github.event_name == 'schedule'
+  # Blocking on every PR that touches Rust. Safe to harden now because the
+  # harness fails only on a Diff — `parity-run` sets `any_diffs` inside the
+  # diffed loop alone, so the 65 fixtures still served by `stub_comparator!`
+  # report as Skipped and cannot turn this red. Measured on main today:
+  # 111 matched / 65 skipped / 0 diffed.
+  #
+  # What it protects: the recorded fixtures are frozen Python output, so this
+  # gate catches the Rust side drifting away from that snapshot — exactly the
+  # failure mode the ongoing port produces. It cannot detect the Python side
+  # drifting away from the fixtures; that needs re-recording, not this job.
+  parity:
+    name: parity
     runs-on: ubuntu-latest
-    continue-on-error: true
     steps:
       - uses: actions/checkout@v7
       - uses: dtolnay/rust-toolchain@stable
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ FIXTURES ?= tests/parity/fixtures
 help:
 	@echo "Headroom Rust targets:"
 	@echo "  make test               - cargo test --workspace"
-	@echo "  make test-parity        - maturin develop + parity-run against fixtures"
+	@echo "  make test-parity        - parity-run against recorded fixtures"
 	@echo "  make bench              - cargo bench --workspace"
 	@echo "  make build-proxy        - release build + strip headroom-proxy, print size"
 	@echo "  make build-wheel        - release wheel for headroom-py"
@@ -36,12 +36,12 @@ help:
 test:
 	$(CARGO) test --workspace
 
+# headroom-parity has no pyo3 dependency — its comparators call headroom-core
+# directly, so this target needs neither a venv nor a built extension module.
+# (See crates/headroom-parity/Cargo.toml: "Phase 0 does not invoke Python from
+# Rust.") Dropping the `maturin develop` step keeps the harness runnable from a
+# bare checkout and takes the Python toolchain off the CI parity job.
 test-parity:
-	@if [ -z "$$VIRTUAL_ENV" ]; then \
-		echo "error: activate a venv first (e.g. source .venv/bin/activate)"; \
-		exit 1; \
-	fi
-	$(MATURIN) develop -m crates/headroom-py/Cargo.toml
 	$(CARGO) run -p headroom-parity -- run --fixtures $(FIXTURES)
 
 bench:
--- a/crates/headroom-parity/src/lib.rs
+++ b/crates/headroom-parity/src/lib.rs
@@ -466,6 +466,51 @@ impl TransformComparator for ContentDetectorComparator {
     }
 }
 
+/// Real comparator for the `text_crusher` transform. The fixture `input` is an
+/// object rather than a bare string, mirroring the recorder's three arguments
+/// (`tests/parity/record_text_crusher.py`), and `config` is always `null` — the
+/// recorder drove the Python default config, so the Rust side uses
+/// `TextCrusherConfig::default()` to match.
+pub struct TextCrusherComparator;
+
+impl TransformComparator for TextCrusherComparator {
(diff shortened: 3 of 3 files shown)


chopratejas:
### `test-dashboard-ui` is red, and it is not this diff — but it is a real finding

Two Playwright tests fail: `test_dashboard_cache_lifetime_playwright.py::test_card_renders_lifetime_cache_reads_after_zero_traffic_restart` and `test_dashboard_cache_ttl_playwright.py::test_dashboard_per_project_setup_url_uses_current_origin`, both `AssertionError: Locator expected to be visible / element(s) not found`.

**Why this PR triggered a job other PRs skip.** The `dashboard` paths-filter is `headroom/dashboard/**` *plus* `.github/workflows/**` (`ci.yml:85-86`). This PR edits `.github/workflows/rust.yml`, so `changes.outputs.dashboard == 'true'` and `test-dashboard-ui` runs. On PRs that touch neither, the job is `SKIPPED` — e.g. #2534, #2529, #2502.

**Why the diff cannot be the cause.** The job installs the wheel built from this branch, then runs `pytest tests/test_dashboard_*_playwright.py` (`ci.yml:419`) against mocked routes. This PR changes three files: `Makefile`, `.github/workflows/rust.yml`, and `crates/headroom-parity/src/lib.rs`.

- `headroom-parity` is **not** in the wheel. The wheel is built from `crates/headroom-py`, and `headroom-parity` appears nowhere as a `[dependencies]` entry — only in workspace `members`. So the installed wheel is behaviourally identical to main's for anything the dashboard touches.
- The job never invokes the `Makefile` or reads `rust.yml`.

There is no code path from a Rust parity comparator to dashboard DOM text.

**What I think actually happened.** These tests have been effectively unexercised for weeks. Per the comment at `ci.yml:417-418`, the main pytest shards skip them via `importorskip` because playwright is not installed there — this dedicated job is the only place they run, and it only fires on a `headroom/dashboard/**` or `.github/workflows/**` change. Across the last 30 `ci.yml` runs, `test-dashboard-ui` reached a real conclusion exactly once: this PR. The other two were `cancelled` release-please runs.

Both test files were last touched on 2026-07-08 (#1665, #1800), and the dashboard has changed under them since — #2198 (persist lifetime proxy metrics) and e5b3a634 (dedupe Codex WS request logging) both landed after. The likely story is straightforward rot that nothing was running to catch.

I did not get a clean before/after on main: playwright is not installed locally, and no recent run gives a baseline. If you want that confirmed empirically, a throwaway branch off `main` touching only a workflow comment would trigger the same job with none of this PR's content — happy to push one, or someone can re-run the job on a main-based branch.

**Worth noting:** this is structurally the same bug this PR fixes. A test suite that only runs under a rare path-filter match is a test suite that silently rots — the parity harness was nightly + `continue-on-error`, these are behind a filter that almost never matches. I have deliberately not touched it here to keep the diff to one concern. Two candidate follow-ups: fix the two assertions, and either widen the trigger or install playwright in the main shards so these run on dashboard-affecting changes.

chopratejas:
### Baseline confirmed: `test-dashboard-ui` is pre-existing red on main

Ran the control in #2569 (now closed, branch deleted): a three-comment-line edit to `rust.yml`, `yaml.safe_load` verified **equal** to `upstream/main`'s, branched off `b121223e` with every other file untouched — carrying the paths-filter trigger and none of this PR's content.

It fails identically:

```
tests/test_dashboard_cache_lifetime_playwright.py::test_card_renders_lifetime_cache_reads_after_zero_traffic_restart FAILED
tests/test_dashboard_cache_ttl_playwright.py::test_dashboard_per_project_setup_url_uses_current_origin FAILED
========================= 2 failed, 9 passed in 20.78s =========================
```

Same two tests, same `AssertionError: Locator expected to be visible`, same 2/9 split as here and on #2568. So the failure is pre-existing on main and unrelated to this PR — which also matches the dependency argument above (`headroom-parity` is not in the wheel, and the job never reads the `Makefile` or `rust.yml`).

`parity` is green on both #2567 and #2568.

Happy to file a separate issue for the two rotted assertions plus the structural gap that let them rot — the job only fires on a `headroom/dashboard/**` or `.github/workflows/**` match, and the main pytest shards skip these via `importorskip` because playwright is not installed there (`ci.yml:417-418`). Same failure mode this PR fixes for the parity harness, different subsystem.

PR #2579, opened 2026-07-26
fix(memory): bound the TrafficLearner pending-pattern accumulator (memory leak)

abhay-codes07:
## Description

`TrafficLearner` (the memory/learning subsystem that accumulates patterns from proxy traffic) has an unbounded in-memory accumulator.

`_pattern_counts` maps `content_hash -> (pattern, count)`. A pattern is added on first sighting, its count is bumped on each re-sighting, and it is **removed only when it reaches `min_evidence`** (default 5), at which point it is promoted and its hash moves to `_saved_hashes`:

```python
if h in self._pattern_counts:
    existing, count = self._pattern_counts[h]
    count += 1
    self._pattern_counts[h] = (existing, count)
else:
    self._pattern_counts[h] = (pattern, 1)
    return  # first sighting — wait for more evidence
...
if count >= self._min_evidence:
    del self._pattern_counts[h]          # only removal path
    self._saved_hashes.add(h)
    if len(self._saved_hashes) > self._dedup_window:  # sibling IS trimmed
        self._saved_hashes.pop()
```

A pattern seen **once but never corroborated** — the common case for one-off traffic (a unique error string, an ad-hoc shell command, a distinct file path) — never reaches `min_evidence`, so it is **never removed**. Over a long-lived proxy processing varied traffic, `_pattern_counts` grows without bound and RSS climbs. The sibling `_saved_hashes` is explicitly trimmed to `dedup_window` ("prevent unbounded growth"); `_pattern_counts` was missed.

Reproduced directly: feeding 500 distinct one-off patterns leaves 500 entries in `_pattern_counts` (one per pattern, forever).

## Fix

Make `_pattern_counts` an LRU-ordered `OrderedDict` capped at a new `max_pending_patterns` (default 2048):

- On each corroboration, `move_to_end(h)` so an actively-accumulating pattern stays "fresh" and is never evicted before it can be promoted.
- On a first sighting when the accumulator is full, evict the least-recently-corroborated pending entry (`popitem(last=False)`).

Evicting a stale one-off is safe: if it recurs it simply restarts accumulation (delayed promotion at worst) — the same tradeoff `_saved_hashes` already makes. Promotion at `min_evidence` is unchanged, and the cap (2048) is generous enough that any pattern receiving repeat sightings within a normal window reaches `min_evidence=5` long before eviction.

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Made

- `headroom/memory/traffic_learner.py`: `_pattern_counts` becomes a capped LRU `OrderedDict`; add `max_pending_patterns` (default 2048); `move_to_end` on corroboration and evict-oldest on overflow.
- `tests/test_memory/test_traffic_learner.py`: a regression that 500 one-off patterns keep the accumulator at its cap, and one that a corroborated pattern still promotes into `_saved_hashes` (both sync via `asyncio.run` so they run without the pytest-asyncio plugin).

## Testing

### Test Output

```text
$ python -m pytest tests/test_memory/test_traffic_learner.py -q
35 failed, 109 passed

# the 35 failures are pre-existing @pytest.mark.asyncio tests that need
# pytest-asyncio (not configured in this environment); they fail identically
# on clean main (35 failed, 107 passed) and pass in CI. My two new tests are
# synchronous and pass; they add +2 passing with no new failures.

# with the fix reverted, test_pending_accumulator_is_bounded fails
# (the accumulator holds all 500 one-off patterns)

$ uvx ruff@0.15.17 check headroom/memory/traffic_learner.py tests/test_memory/test_traffic_learner.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/memory/traffic_learner.py
```

## Real Behavior Proof

- Exact command / steps: built a `TrafficLearner(backend=None, min_evidence=5, max_pending_patterns=8)` and drove `_accumulate` with 500 distinct one-off `ExtractedPattern`s; separately corroborated one pattern to `min_evidence`; then reverted the source and re-ran.
- Observed result: with the fix `len(_pattern_counts)` stays at the cap (8) after 500 one-offs, the corroborated pattern is removed from pending and present in `_saved_hashes`, and an actively-bumped pattern survives LRU eviction; with the fix reverted the accumulator holds all 500 one-off entries (the unbounded leak). Ran against the actual module.
- Not tested: a live multi-day proxy run measuring RSS (the leak is inferred from the removed unbounded-growth path; the accumulator bound is verified directly).

## Review Readiness

## Checklist

### DIFF
--- a/headroom/memory/traffic_learner.py
+++ b/headroom/memory/traffic_learner.py
@@ -26,6 +26,7 @@ import os
 import re
 import sqlite3
 import time
+from collections import OrderedDict
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from enum import Enum
@@ -406,6 +407,7 @@ class TrafficLearner:
         max_history: int = 20,
         dedup_window: int = 100,
         min_evidence: int = 5,
+        max_pending_patterns: int = 2048,
     ) -> None:
         """Initialize the traffic learner.
 
@@ -424,12 +426,19 @@ class TrafficLearner:
         self.agent_type = agent_type
         self._max_history = max_history
         self._min_evidence = min_evidence
+        self._max_pending_patterns = max_pending_patterns
 
         # Recent tool call history for error→recovery matching
         self._tool_history: list[dict[str, Any]] = []
 
-        # Pattern accumulator: hash → (pattern, count)
-        self._pattern_counts: dict[str, tuple[ExtractedPattern, int]] = {}
+        # Pattern accumulator: hash → (pattern, count). LRU-ordered and capped:
+        # a pattern that is seen once but never reaches ``min_evidence`` would
+        # otherwise linger here forever, so this dict grew unbounded over a
+        # long-lived proxy's traffic (the sibling ``_saved_hashes`` is trimmed
+        # to ``dedup_window`` for the same reason; this one was missed). Evicting
+        # the least-recently-corroborated pending pattern is safe: if it recurs
+        # it simply restarts accumulation.
+        self._pattern_counts: OrderedDict[str, tuple[ExtractedPattern, int]] = OrderedDict()
 
         # Dedup: hashes of patterns already saved to DB
         self._saved_hashes: set[str] = set()
@@ -1196,7 +1205,13 @@ class TrafficLearner:
             existing, count = self._pattern_counts[h]
             count += 1
             self._pattern_counts[h] = (existing, count)
+            # Mark as most-recently-corroborated so it survives LRU eviction.
+            self._pattern_counts.move_to_end(h)
         else:
+            # Bound the pending accumulator so one-off patterns can't grow it
+            # without limit; drop the least-recently-corroborated pending entry.
+            if len(self._pattern_counts) >= self._max_pending_patterns:
+                self._pattern_counts.popitem(last=False)
             self._pattern_counts[h] = (pattern, 1)
             return  # First sighting — wait for more evidence
 
--- a/tests/test_memory/test_traffic_learner.py
+++ b/tests/test_memory/test_traffic_learner.py
@@ -360,6 +360,56 @@ class TestTrafficLearner:
         stats = learner.get_stats()
         assert stats["patterns_extracted"] >= 3
 
+    def test_pending_accumulator_is_bounded(self):
+        """One-off patterns that never reach ``min_evidence`` must not grow the
+        pending ``_pattern_counts`` accumulator without bound — the sibling
+        ``_saved_hashes`` is already trimmed to ``dedup_window`` and this one was
+        missed, so a long-lived proxy leaked memory across varied traffic. It is
+        now LRU-capped at ``max_pending_patterns``.
+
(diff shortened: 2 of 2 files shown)


abhay-codes07:
Both red checks on this PR are pre-existing on `main` (commit `904bc675`, "fix(cli): warn when Headroom proxy URL leaks into the shell after unwrap claude" #2571), not caused by this change. This branch only touches `headroom/memory/traffic_learner.py` and its test, so it just inherits them.

**1. `test (1)` (9 failures in `tests/test_cli/test_unwrap_claude.py`)**

This is the important one, because it is a real break of the `unwrap claude` command, not only a test problem. #2571 placed the new `def _warn_if_proxy_env_leaked` directly beneath the `unwrap claude` decorator stack:

```python
@unwrap.command("claude")
@click.option("--port", "-p", ...)
@click.option("--no-stop-proxy", is_flag=True, ...)
@click.option("--keep-mcp", is_flag=True, ...)
@click.option("--keep-rtk", is_flag=True, ...)
def _warn_if_proxy_env_leaked(port: int) -> None:   # <- decorators bind here now
    ...

def unwrap_claude(port, no_stop_proxy, keep_mcp, keep_rtk) -> None:   # <- now undecorated, orphaned
    ...
```

On the parent commit `fd6abac8` that same decorator stack sat directly on `def unwrap_claude`. Inserting `_warn_if_proxy_env_leaked` in between means the `claude` command now binds to `_warn_if_proxy_env_leaked`, so running `headroom unwrap claude` makes click call it with the command options and it raises `TypeError: _warn_if_proxy_env_leaked() got an unexpected keyword argument 'no_stop_proxy'` (and `'keep_mcp'`). `unwrap_claude` below it is never reached.

**2. `lint` (`ruff format --check`)**

Separately, the `click.echo(...)` call added inside `_warn_if_proxy_env_leaked` (around `wrap.py:5289`) is written across three lines, but ruff `0.15.17` (the pinned version) collapses it onto one line, so `ruff format --check` reports `Would reformat: headroom/cli/wrap.py` on every branch off `main`. I opened #2582 for that one line.

Neither issue is from this PR. Happy to send a single follow-up that moves `_warn_if_proxy_env_leaked` back below `unwrap_claude` (restoring the command) and applies the format fix, so `main` goes green again. Once it lands I will rebase this PR and its checks should clear.

JerrettDavis (approved):
Reviewed the pending-pattern accumulator cap. The `OrderedDict` LRU preserves actively corroborated patterns via `move_to_end`, promotes at `min_evidence` as before, and bounds one-off traffic growth. Focused local verification passed: `python -m pytest tests/test_memory/test_traffic_learner.py::TestTrafficLearner::test_pending_accumulator_is_bounded tests/test_memory/test_traffic_learner.py::TestTrafficLearner::test_pending_accumulator_lru_still_promotes_corroborated_pattern -q` (2 passed) and `python -m ruff check headroom/memory/traffic_learner.py tests/test_memory/test_traffic_learner.py`. CI is green on the refreshed branch.

PR #2597, opened 2026-07-27
fix(compression): report source-line span in CCR compression marker

### LINKED ISSUE #2586: [BUG] Compressed tool output silently drops whole lines
## Description

When tool output is compressed, whole lines are removed and replaced by a marker that identifies neither *which* content was dropped nor a unit that maps to lines. The consumer cannot distinguish "this line was compressed away" from "this line was not in the output", so absence of a line reads as evidence of absence — and in my case produced a materially wrong conclusion that was acted on.

The marker looks like this:

```
[122 items compressed to 27. Retrieve more: hash=c00eb437e5e570170bd38825]
```

"122 items" for what was 5 lines of command output. The unit is not lines, so the number gives the reader no way to reason about how much of a line-structured output is missing.

## To Reproduce

1. Run headroom in front of a coding agent (here: Claude Code, `--mode token`).
2. Have the agent run a shell command whose output is line-structured and where each line is independently meaningful — an environment dump, a key listing, a `grep -n` result.
3. Read what the agent receives.

## Expected Behavior

Either of:

- No line is dropped silently — a line is preserved, or its removal is individually marked in place; or
- The marker identifies what was dropped in a unit that maps onto the source (line ranges, keys, or field names), so the reader can tell whether the thing they are looking for was present and elided, versus absent.

## Actual Behavior

Three reproductions from a single session.

**1. Line deleted, decision corrupted.** A 5-line command output rendered as one line plus the marker above. The needed line — line 2 of 5, `SSH_AUTH_SOCK=...` — was gone. Because absence was indistinguishable from "the variable is not set", the reading taken was "no SSH agent in this environment". That was wrong; the variable was set. The conclusion was acted on.

**2. Half of an environment dump silently gone.**

```
$ systemctl --user show-environment
HOME=/home/tobias LANG=de_DE.UTF-8 LOGNAME=tobias PATH=... SHELL=/usr/bin/fish ...
[52 items compressed to 27. Retrieve more: hash=c00eb437e5e570170bd38825]
```

52 newline-separated `KEY=value` pairs arrived as a single run-on line with roughly half missing, and no way to know which keys. For an environment dump this is the worst case: the question being asked is almost always "is variable X set?", and the compressed form cannot answer it either way.

**3. Function words deleted inside retained lines.** This one is separate from line-dropping and arguably worse, because there is no marker at all — the text simply reads as if written that way. Retained prose came through as:

- `slim glossary infrastructure terms under split`
  — actual source: *slim glossary **to** infrastructure terms under **the** split*
- `minutes Nix build + bats validate nothing relevant change`
  — actual source: *minutes **of** Nix build + bats **that** validate nothing relevant **to the** change*
- `real hazard is habit trains — reaching --no-verify, then also skips gate on pushes genuinely change Nix code`

The result is still readable enough to be believed, which is the hazard: it looks like terse writing rather than damaged text.

## What did NOT go wrong

Included deliberately, since it narrows where the problem is:

- **Negations were preserved.** I tested for this specifically, expecting a "not" to be dropped somewhere and invert a meaning. It never happened across the session.
- **No hex identifier was ever corrupted.** Commit shas, run IDs and the marker's own `hash=` token all came through intact and usable.

So the damage is concentrated in low-information-weight tokens and in whole-line elision, not in identifiers or logical polarity. That is a much better failure profile than it could be — it just needs to stop being silent.

## Impact

The failure is silent and self-confirming. A dropped line produces confident, wrong reasoning rather than an error or a visible gap, and nothing in the output signals that a check is warranted. `Retrieve more: hash=...` is offered, but a reader only retrieves when they suspect something is missing — and the whole problem is that they cannot.

## Environment

- headroom-ai **0.27.0** (installed via `uv tool`)
- Client: Claude Code, `--mode token`
- OS: Linux 6.8 (Ubuntu), Python 3.12
- Frequency: always, whenever output exceeds the compression threshold

Noting the version explicitly: 0.27.0 is behind current, and I have not verified this against 0.32.x. If the marker semantics changed since — particularly whether it now reports dropped content in a line-mapped unit — please close this out; I could not confirm either way from the installed version.

## Related

- #2421 (*Read outliner can present a client-truncated file prefix as a complete definition list*) — same family: truncated content presented as complete, with nothing to signal the truncation. That issue is the Read outliner; this is the compression marker on tool output.

tenderdeve:
## Description

The compression marker read `[N items compressed to M. Retrieve more: hash=...]`, where `items` counts whitespace-split **words**, not lines. So five lines of tool output could show as `[122 items compressed to 27...]`. A reader can't map "items" to lines and can't tell "this line was compressed away" from "this line was never in the output" — absence reads as evidence of absence, which per the report led to a materially wrong conclusion.

Closes #2586

## Type of Change

## Changes Made

- Annotate the marker with the source line count — `[N items compressed to M (from L source lines). Retrieve more: hash=...]` — at both marker sites: `KompressCompressor.compress` / `compress_batch` (`kompress_compressor.py`) and the remote path (`kompress_remote.py`).
- The machine-parsed `Retrieve more: hash=` token is left byte-for-byte unchanged, so CCR detection/retrieval is unaffected.

Scope note: I intentionally kept the existing `items compressed to` phrasing rather than reword the unit, to avoid churning the marker format that's referenced across ~12 test fixtures and the `config.py` template. This is the minimal honesty fix; happy to go further (e.g. line-unit counts or unifying with the `config.py` template) if you'd prefer — see the issue thread where I asked about wording.

## Testing

### Test Output

```text
$ uv run pytest tests/test_compression_units.py tests/test_compression_batches.py \
    tests/test_ccr_marker_policy.py tests/test_ccr_tool_injection.py tests/test_session_probes.py -q
92 passed

$ uv run pytest tests/test_ccr_marker_policy.py -q
8 passed   # incl. new test_source_line_span_marker_is_still_detected

$ uv run ruff check headroom/transforms/kompress_compressor.py headroom/transforms/kompress_remote.py tests/test_ccr_marker_policy.py
All checks passed!
```

## Real Behavior Proof

- Environment: headroom @ main, Python 3.14, uv
- Exact command / steps: added a marker in the new enriched format and ran it through the CCR marker detector.
- Observed result: the retrieval hash is still detected from `[122 items compressed to 27 (from 5 source lines). Retrieve more: hash=...]`; existing compression/CCR suites unchanged.
- Not tested: mypy not run locally; the full model-backed compress() marker path isn't unit-exercised (needs a real backend), so the new test targets the parser boundary instead.

## Review Readiness

## Checklist

## Additional Notes

Wording is adjustable per the issue discussion. The `config.py` marker template (a different code path with `Omitted`/`Expires` fields) is left untouched to keep this focused on the Kompress marker the report hit.

### DIFF
--- a/headroom/transforms/kompress_compressor.py
+++ b/headroom/transforms/kompress_compressor.py
@@ -1452,8 +1452,14 @@ class KompressCompressor(Transform):
                 cache_key = self._store_in_ccr(ccr_source, compressed, ccr_source_tokens)
                 if cache_key:
                     result.cache_key = cache_key
+                    # Report the source line span so a reader can tell content was
+                    # compressed away rather than absent — "items" counts words, which
+                    # does not map to lines and reads as evidence of absence (#2586).
+                    source_lines = ccr_source.count("\n") + 1
+                    line_word = "line" if source_lines == 1 else "lines"
                     result.compressed += (
-                        f"\n[{n_words} items compressed to {compressed_count}."
+                        f"\n[{n_words} items compressed to {compressed_count}"
+                        f" (from {source_lines} source {line_word})."
                         f" Retrieve more: hash={cache_key}]"
                     )
 
@@ -1809,8 +1815,14 @@ class KompressCompressor(Transform):
                 cache_key = self._store_in_ccr(ccr_source, compressed, ccr_source_tokens)
                 if cache_key:
                     result.cache_key = cache_key
+                    # Report the source line span so a reader can tell content was
+                    # compressed away rather than absent — "items" counts words, which
+                    # does not map to lines and reads as evidence of absence (#2586).
+                    source_lines = ccr_source.count("\n") + 1
+                    line_word = "line" if source_lines == 1 else "lines"
                     result.compressed += (
-                        f"\n[{n_words} items compressed to {compressed_count}."
+                        f"\n[{n_words} items compressed to {compressed_count}"
+                        f" (from {source_lines} source {line_word})."
                         f" Retrieve more: hash={cache_key}]"
                     )
 
--- a/headroom/transforms/kompress_remote.py
+++ b/headroom/transforms/kompress_remote.py
@@ -129,9 +129,14 @@ class RemoteKompressCompressor:
             cache_key = store_kompress_in_ccr(content, compressed, result.original_tokens)
             if cache_key:
                 result.cache_key = cache_key
+                # Report the source line span so a reader can tell content was
+                # compressed away rather than absent (#2586).
+                source_lines = content.count("\n") + 1
+                line_word = "line" if source_lines == 1 else "lines"
                 result.compressed += (
                     f"\n[{result.original_tokens} items compressed to "
-                    f"{result.compressed_tokens}. Retrieve more: hash={cache_key}]"
+                    f"{result.compressed_tokens} (from {source_lines} source {line_word})."
+                    f" Retrieve more: hash={cache_key}]"
                 )
 
         return result
--- a/tests/test_ccr_marker_policy.py
+++ b/tests/test_ccr_marker_policy.py
@@ -27,6 +27,13 @@ def test_has_new_ccr_markers_filters_replayed_forwarded_markers() -> None:
     )
 
 
+def test_source_line_span_marker_is_still_detected() -> None:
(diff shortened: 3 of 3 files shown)


chopratejas:
A side effect of this is an increase in tokens - albeit too little though.

PR #2607, opened 2026-07-27
fix(ccr): treat the CCR TTL as an idle window, not a wall clock (#2604)

### LINKED ISSUE #2604: CCR entries expire mid-session — 30-min wall-clock TTL is shorter than a normal multi-agent burst, causing headroom_retrieve "Entry not found" on live results
Package: headroom-ai 0.32.1, `headroom mcp serve`

## Bug
`DEFAULT_CCR_TTL_SECONDS = 1800` (cache/compression_store.py:51) is a fixed wall-clock TTL,
session-oblivious. A burst of parallel sub-agents whose bulky tool results get compressed on
arrival, followed by any later step reaching back >30 min later, hits an expired entry.

## Repro (observed)
7 apply sub-agents dispatched 22:48–22:49, results landing (and ccr-compressing) 22:52–23:12.
First `"Entry not found (CCR TTL: 1800 seconds)"` at 23:41 — ~29 min after the last result
landed, ~49 min after the first. A second cluster hit on next-morning session resume, reaching
back to entries by then hours old.

## Ask
Session-scoped retention (keep entries for the life of the MCP session / drop on session end)
instead of a fixed wall-clock guess, and/or raise the default well above 30 min.

## Possible secondary issue
Worth checking whether there's a proxy-vs-`headroom_retrieve` cross-process store split — i.e.
entries compressed by one process invisible to the retrieve tool called from another. Happy to
provide a minimal compress→retrieve round-trip repro if that's the actual mechanism rather than
pure TTL expiry.

env: HEADROOM_CCR_TTL_SECONDS unset (default in effect)

Parideboy:
Closes #2604

## Description

CCR entries expired while the agent was still using them. `DEFAULT_CCR_TTL_SECONDS` was a wall-clock countdown from `created_at`, never refreshed, so a burst of parallel sub-agents whose bulky results were compressed on arrival hit `Entry not found` as soon as any later step reached back more than 30 minutes. The `<<ccr:...>>` marker was still in context but the original was gone, which silently converts "lossless with retrieval" into lossy.

The TTL is now an **idle** window. `retrieve()` already recorded `last_accessed` for the feedback path; expiry is measured from it, so an entry the session keeps touching survives arbitrarily long. An absolute `max_lifetime` ceiling (8x the TTL) keeps originals from sitting on disk indefinitely.

I also checked the reporter's secondary hypothesis — a proxy vs `headroom_retrieve` cross-process store split — and it is **not** the mechanism. Both `ccr/response_handler.py` and `ccr/mcp_server.py` resolve through `get_compression_store()` to the same SQLite file under `workspace_dir()`. Verified by round-tripping an entry between two store instances over one DB file. The only genuinely isolated store is `UniversalCompressor._store_in_ccr` (bare `CompressionStore()` = in-memory), which is the public library API, not the proxy or MCP path.

Session-scoped retention was requested in the issue but is not implemented literally: the SQLite store is cross-process with no session ownership, the proxy compresses with no MCP session attached, and the reporter's second cluster was a next-morning resume that session-scoped retention would have dropped. A sliding idle window with a longer default gives the same practical behaviour without new lifecycle machinery.

## Type of Change

- [ ] Breaking change
- [ ] Documentation update

## Changes Made

- `CompressionEntry.is_expired()` measures age from `last_accessed or created_at`; new `expiry_reason()` returns `idle` / `max_lifetime` / `None`
- New `max_lifetime` field with a `DEFAULT_CCR_MAX_LIFETIME_MULTIPLIER = 8` default, so a repeatedly-touched entry still has a hard ceiling
- `DEFAULT_CCR_TTL_SECONDS` 1800 -> 3600, matching the MCP server's `MCP_SESSION_TTL` (which proxy-compressed entries did not previously share); `MCP_SESSION_TTL` now derives from the store default, and `CCRConfig.store_ttl_seconds` plus the Rust `DEFAULT_TTL` move in lockstep
- Eviction heap re-keyed on last access instead of creation; `retrieve()` re-pushes on access and the existing `_stale_heap_entries` counter absorbs the duplicate tuples, so capacity pressure drops the coldest entry rather than the one the agent is actively retrieving
- Bounded tombstone ring (512 entries) records why a hash was removed, so misses distinguish `Entry expired (CCR idle TTL: ...)`, `Entry expired (CCR max lifetime: ...)`, `Entry evicted for capacity (CCR store limit: ...)` and `Entry not found`. Previously `retrieve()` deleted an expired entry before the status lookup ran, making expiry indistinguishable from a hash that was never stored, and the printed TTL was the global default rather than the entry's own — that is exactly the `Entry not found (CCR TTL: 1800 seconds)` the reporter saw
- `HEADROOM_CCR_TTL_SECONDS` registered in `settings_store` so the knob is discoverable
- SQLite backend gains an `expires_at` column via an ALTER-on-open migration; rows predating it fall back to the legacy `created_at + ttl` purge predicate instead of leaking (NULL comparisons are never true)
- The tombstone ring is best-effort by construction and now says so: reasons are process-local, and `SQLiteBackend._open()` / `_maybe_purge()` sweep expired rows themselves, so a restart or a sibling worker's purge leaves none. A bare miss reports `Entry not found - never stored, or expired/evicted with no reason recorded` rather than claiming the hash was never stored. Persisting reasons in the backend is filed as #2956
- Dropped a stale note above the Rust `DEFAULT_TTL` claiming that store measures TTL from insertion; it contradicted the doc block directly below it, and all three Rust backends implement the idle window plus ceiling
- Docs updated; new tests for the sliding window, the ceiling, eviction ordering, miss-reason reporting, and the reasonless-miss path across a simulated restart

## Testing

- [x] Unit tests pass locally
- [x] Manual end-to-end verification

Targeted suites:

```
$ python -m pytest tests/test_ccr.py tests/test_ccr_sqlite_backend.py tests/test_proxy_ccr.py tests/test_ccr_mcp_server.py tests/test_compression_store.py tests/test_ccr_batch_store.py tests/test_proxy_settings_endpoints.py -q
================= 179 passed, 1 skipped, 3 warnings in 39.55s =================
```

Adversarial benchmark, unchanged from the pre-change baseline of 24/36, with the TTL boundary case still passing:

```
$ python benchmarks/adversarial_ccr_tests.py
    PASS  TTL Exact Boundary
  TOTAL: 24/36 tests passed
```

Manual script covering the legacy-DB migration, cross-process visibility, and the sliding window:

```
$ python ccr_e2e.py
after migration+purge: ['legacy_live']
other process retrieve: True
alive 3s after creation (touched): True
expired when idle: True
miss status: expired_and_purged
```

Review round (`a7079999`, best-effort miss reasons + the Rust comment fix):

```
$ uv run pytest tests/test_ccr.py tests/test_ccr_sqlite_backend.py tests/test_proxy_ccr.py tests/test_compression_store.py -q
135 passed, 1 skipped, 2 warnings in 47.77s
$ cargo check -p headroom-core
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 37.78s
```

The new regression test fails on the old wording, so it is a real guard:

```
$ git stash push headroom/cache/compression_store.py && uv run pytest tests/test_ccr_sqlite_backend.py -q -k backend_purge_leaves_no_reason
tests\test_ccr_sqlite_backend.py:216: in test_backend_purge_leaves_no_reason_and_the_miss_says_so
    assert "expired/evicted with no reason recorded" in detail
E   AssertionError: assert 'expired/evicted with no reason recorded' in 'Entry not found (CCR idle TTL: 3600 seconds)'
1 failed, 21 deselected in 0.91s
```

Lint and types:

```
$ python -m ruff check headroom tests
All checks passed!
$ python -m mypy headroom --ignore-missing-imports
(no error: lines)
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13, local checkout at upstream/main 4a8157fa, headroom._core built locally
- Exact command / steps: created a legacy-shaped ccr_entries table with no expires_at column (one row already past its TTL, one live), opened it with SQLiteBackend to trigger the ALTER-on-open migration and purge, then stored an entry through a CompressionStore with default_ttl=2 backed by a second SQLiteBackend over the same file, retrieved it from a separate CompressionStore instance after 1.5s, retrieved it again from the first at 3.0s (past the old creation-based deadline), then left it idle for 2.5s and checked get_entry_status
- Observed result: only the genuinely expired legacy row was purged (`['legacy_live']`); the second store instance retrieved the entry written by the first, proving no cross-process split; the entry was still alive 3s after creation because retrieval restarted the idle window; it expired once left idle and reported status `expired_and_purged` instead of a bare miss
- Not tested: a real multi-hour agent session against a live provider, and the Redis backend end-to-end against a live Redis. The Rust store is not exercised by the Python run above, but it is not Python-only either: `in_memory.rs`, `redis.rs` and `sqlite.rs` all track `last_accessed` and enforce the idle window plus the max-lifetime ceiling, covered by `crates/headroom-core/tests/ccr_backends.rs` (`in_memory_get_refreshes_idle_ttl`, `sqlite_get_refreshes_idle_ttl`, and the two `max_lifetime_caps_sliding_window` cases)

## Runtime Rollout Safety

- Rollout-managed feature(s): none. CCR is not in the `headroom.rollout` feature registry (`tool_result_interceptors`, `proxy_output_shaper`, `read_maturation` only)
- Stable/default behavior changed: yes. Default CCR idle TTL changes from 1800s to 3600s, eviction is now keyed on last-access instead of creation, and a miss now reports why (idle/max-lifetime/capacity/unknown) instead of a bare "not found". All defaults, no channel gating.
- Kill switch / disable path: `HEADROOM_CCR_TTL_SECONDS` restores any prior duration; there is no flag to revert to wall-clock (age-from-creation) semantics, only a code revert
- Qualification impact: none. CCR entry lifecycle only, no routing/caching/backend-selection change
- Rollback path: revert this commit. The SQLite `expires_at` column is added via an ALTER-on-open migration and ignored by pre-change code, so a store already migrated stays readable after downgrade

## Review Readiness


JerrettDavis (changes requested):
Thanks for the thorough CCR retention pass. I found one blocker in the new miss-reason path: `get_metadata()` still deletes expired entries without recording a tombstone (`headroom/cache/compression_store.py:599`). That method is used by the proxy marker-refresh/context path, so an expired entry can be purged there before `/v1/retrieve` or `get_entry_status()` sees it. The user then gets the generic `Entry not found` message instead of the new `expired_and_purged`/idle-TTL reason this PR is adding. Please record the same `expired:<reason>` tombstone here before deletion, and add a regression where `get_metadata()` sees an expired entry and a later `get_entry_status()` / formatted miss still reports expiry rather than missing.

Parideboy:
Good catch — fixed in 641e12fa.

`get_metadata()` now purges through the same `expiry_reason()` + `_record_tombstone(hash_key, f"expired:{reason}", entry)` path the other removal sites use, so an entry the proxy marker-refresh path purges still reports `expired_and_purged` (with its `expiry_reason`) on a later `get_entry_status()` instead of the generic `Entry not found`.

While wiring the reason through I also noticed `format_retrieval_miss_detail()` hardcoded the idle TTL in the `expired_and_purged` branch, so a max-lifetime purge was reported as an idle-TTL expiry. That branch now honours `expiry_reason` the same way the live-`expired` branch does.

Two regressions added in `tests/test_ccr.py`:

- `test_get_metadata_purge_still_reports_expiry` — idle-TTL entry purged via `get_metadata()`, later `get_entry_status()` returns `expired_and_purged` / `idle` and the formatted miss says `Entry expired and was purged (CCR idle TTL: ...)`.
- `test_get_metadata_purge_reports_max_lifetime` — same flow under a max-lifetime ceiling, asserting the message names the max lifetime rather than the idle TTL.

```
$ python -m pytest tests/test_ccr.py tests/test_ccr_sqlite_backend.py tests/test_proxy_ccr.py tests/test_ccr_mcp_server.py tests/test_compression_store.py tests/test_ccr_batch_store.py tests/test_proxy_settings_endpoints.py -q
================= 181 passed, 1 skipped, 3 warnings in 48.45s =================
```

Ready for re-review.

JerrettDavis (approved):
Re-reviewed after 641e12fa. The previous blocker is addressed: get_metadata() now records the same expired:<reason> tombstone before deleting, and the formatted expired-and-purged message preserves max_lifetime versus idle TTL. The added regressions cover both paths.

Validated locally:
- uv run --extra dev python -m pytest tests/test_ccr.py tests/test_ccr_sqlite_backend.py tests/test_proxy_ccr.py tests/test_ccr_mcp_server.py tests/test_compression_store.py tests/test_ccr_batch_store.py tests/test_proxy_settings_endpoints.py -q (180 passed, 2 skipped)
- uv run --extra dev ruff check headroom/cache/compression_store.py headroom/cache/backends/sqlite.py headroom/ccr/mcp_server.py headroom/config.py headroom/settings_store.py tests/test_ccr.py tests/test_ccr_sqlite_backend.py tests/test_compression_store.py tests/test_proxy_ccr.py
- cargo test -p headroom-core --lib -- ccr (36 passed)
- cargo fmt --all -- --check

Parideboy:
Rebased onto current main to clear the conflicts. Two spots needed hand-merging, both additive:

- `compression_store.py` — the bare-marker guard from #2694 landed in `store()` right where this PR computes `effective_ttl`/`effective_max_lifetime`. Kept both, guard first, since it's an early return and there's no point computing a TTL for a value we're about to throw away.
- `ccr/mcp_server.py` — import collision only. Folded `DEFAULT_CCR_TTL_SECONDS` and `format_retrieval_miss_detail` into one parenthesized import.

Worth flagging the current state of #2604 since a few PRs have moved around it: #2631 landed the Rust-core half (sliding idle window + max-lifetime ceiling across the three backends) and #2669 tightened the SQLite boundary comparison on top of that. #2631's description points at this PR for the Python side, so the two halves are complementary rather than overlapping — but as things stand main has the Rust semantics and not the Python ones, which is a slightly awkward place to sit. This PR closes that.

CCR suite passes locally on the rebased branch (`test_ccr.py`, `test_ccr_sqlite_backend.py`, `test_compression_store.py`, `test_proxy_ccr.py` — 134 passed, 1 skipped), ruff clean.

Parideboy:
`cargo test --workspace` was failing on `sqlite_max_lifetime_caps_sliding_window`. It's not from this PR — the only Rust change here is the `DEFAULT_TTL` constant in `ccr/mod.rs`, and that test builds its store with explicit `open_with_ttls(2, 3)`, so the constant never enters into it. The test file itself is byte-identical to main on this branch.

What actually happened is a collision between two merged PRs. #2669 made the purge boundary strict:

```sql
WHERE last_accessed + ttl_seconds < ?1
   OR created_at + ?2 < ?1
```

so an entry is still valid at exactly `created_at + max_lifetime`. The test's margins were written for the older `<=` behaviour and weren't retuned — its own header comment still quotes `last_accessed + ttl <= now`. It sleeps 500ms then touches 4x at 700ms against a 3s ceiling, which is 3.3s of real time, so apparent age is `floor(t0 + 3.3) - floor(t0)` = 3s or 4s depending on where t0 falls in its second. At 3s the entry now survives and the final assertion blows up with `left: Some("capped sqlite"), right: None`. Fails roughly 70% of the time, on main too — this branch just drew the short straw.

40e29fc6 takes the loop to 6 touches so apparent age is 4s or 5s for any phase of t0, and fixes the stale comparison in the comment. Test only, no production code. Passes 5/5 locally plus a full `ccr_backends` run (12 passed).

Happy to pull this into its own PR against main if you'd rather keep #2607 focused — it's an independent flake and someone else will hit it.

JerrettDavis (approved):
Resolved the current-main conflict in the Rust max-lifetime boundary test by preserving the PR’s stronger six-iteration timing margin. Validation: 181 targeted Python tests passed (1 skipped), and the focused Rust CCR backend test passed.

JerrettDavis (changes requested):
The idle-window/max-lifetime mechanics and access-based heap re-keying look sound, but the new miss-reason contract is not correct for the production persistent path. `_tombstones` is process-local and is populated only when this `CompressionStore` instance performs the deletion. `SQLiteBackend._open()` and `_maybe_purge()` delete expired rows directly, without a tombstone, and every process restart loses the in-memory ring. A genuinely stored-and-expired hash can therefore return `status=missing` / “Entry not found” after startup or another backend purge. That contradicts the new CCR docs (“Entry not found means the hash was never stored by this store”) and the PR’s claim that misses now distinguish the cause.

Please either persist removal reasons in the backend (including startup/opportunistic purge) or explicitly make this diagnostic best-effort: soften the docs and response text so bare “not found” also covers an expiration/eviction whose tombstone is unavailable, and add a regression test that closes/reopens a SQLite-backed store after expiry. Also refresh the PR description’s Rust caveat: the current rebased Rust backends already track `last_accessed` and enforce an idle window plus max lifetime; the statement that Rust still measures from insertion is stale. The core Python sliding-TTL change otherwise looks well covered.

Parideboy:
Thanks, both points land. Pushed `a7079999`.

**Miss-reason contract.** You're right that it is wrong for the persistent path, and the failure is a little broader than the ring being lost on restart: `SQLiteBackend._open()` sweeps at startup *and* `_maybe_purge()` sweeps on write, so even inside one long-lived process a sibling worker's purge produces a reasonless miss.

I took the best-effort option rather than persisting, because persistence would not actually make the guarantee absolute. Redis and MongoDB backends load by entry point, so a tombstone table can only ever be an optional backend capability; the table itself needs a bound; and cross-worker purges race. The honest wording is a prerequisite for persistence, not an alternative to it, so it goes in either way.

Concretely:

- `format_retrieval_miss_detail()` now returns `Entry not found - never stored, or expired/evicted with no reason recorded (CCR idle TTL: N seconds)`. The `Entry not found` prefix is preserved deliberately, since `tests/test_ccr.py` and `tests/test_proxy_ccr.py` assert on it.
- The `_tombstones` comment states the best-effort scope where the ring is defined and names the two backend sweeps that bypass it, so the next reader does not have to rediscover this from the miss text.
- `docs/content/docs/ccr.mdx` drops the "never stored by this store" claim and says a bare not-found covers a backend sweep or an earlier process too.

Persistence is filed as #2956 with the design sketch (bounded `ccr_tombstones` table, INSERT-then-DELETE at both sweep sites, `max_lifetime` promoted to a column so SQL can tell `idle` from `max_lifetime`, `getattr`-based optional backend capability, `_missing_status` consulting the backend before the ring).

**Regression test.** `tests/test_ccr_sqlite_backend.py::TestMultiWorkerSafety::test_backend_purge_leaves_no_reason_and_the_miss_says_so` stores an expired entry, drops the store, and rebuilds `SQLiteBackend` + `CompressionStore` against the same file. Verified it fails on the old wording:

```
tests\test_ccr_sqlite_backend.py:216: in test_backend_purge_leaves_no_reason_and_the_miss_says_so
    assert "expired/evicted with no reason recorded" in detail
E   AssertionError: assert 'expired/evicted with no reason recorded' in 'Entry not found (CCR idle TTL: 3600 seconds)'
```

**Rust caveat.** Stale in the code as well as the description, and I've fixed both. The PR had added this above `DEFAULT_TTL` in `crates/headroom-core/src/ccr/mod.rs`:

```rust
/// Note: this store measures TTL from insertion; the Python store treats
/// it as an idle window that restarts on every retrieval.
```

which contradicts the doc block immediately below it. `backends/in_memory.rs`, `backends/redis.rs` and `backends/sqlite.rs` all track `last_accessed` and enforce the idle window plus max lifetime, so those two lines are just wrong. Removed, and the PR description updated.

Local verification: `tests/test_ccr.py tests/test_ccr_sqlite_backend.py tests/test_proxy_ccr.py tests/test_compression_store.py` -> 135 passed, 1 skipped (POSIX-only permissions test, on Windows). `ruff check .` clean, `mypy headroom --ignore-missing-imports` unchanged (12 pre-existing errors, none in the touched files), `cargo check -p headroom-core` clean.

JerrettDavis (approved):
Re-reviewed the latest source after the diagnostic-contract fix and current-main refresh. The prior blocker is resolved: a reasonless persistent-backend miss is now documented and reported honestly as ambiguous, the stale Rust insertion-TTL statement is gone, and the restart regression covers the production SQLite failure mode.

I merged current `main` into the branch (clean merge) and validated that exact integration: 182 focused CCR/proxy/settings tests passed with 1 skip, focused Ruff passed, and `git diff --check` passed. The full prior-head CI was green. Ready from my review.

JerrettDavis (approved):
Revalidated after the upstream/main merge. The merge introduces no contributor-logic delta beyond the previously reviewed idle-TTL implementation, current main is fully contained, the branch is clean and conflict-free, and all checks on the merged head are green.

PR #2608, opened 2026-07-27
fix(read-lifecycle): don't mark a Read stale on a failed edit

abhay-codes07:
## Description

`ReadLifecycleManager` marks a Read output STALE when a later Edit/Write targets the same file. The staleness check looks only at tool *calls*, never at whether the edit actually succeeded:

```python
is_stale = self.config.compress_stale and any(
    e.msg_index > read_op.msg_index for e in edits
)
```

`edits` is built purely from assistant tool-use blocks (`_build_file_operation_index` never consults the tool result). So a **failed** Edit still counts. Failed edits are common in agentic coding: Claude Code frequently gets `String to replace not found in file` when an `old_string` no longer matches, and the tool result comes back with `is_error: true`. The file is unchanged, so the earlier Read is still accurate, yet it gets replaced with a stale marker plus a CCR hash. That both misleads the model (the marker says the file was edited when it was not) and forces an unnecessary retrieval to get content that was correct all along.

`compress_stale` and `enabled` are both on by default (`ReadLifecycleConfig`), so this happens in normal Claude Code sessions.

## Fix

Before building the file-operation index, collect the tool-call IDs whose Anthropic tool_result carries `is_error: true`, and skip those operations. A failed edit no longer triggers staleness, and a failed read (no real content) can no longer supersede an earlier read. A later *successful* edit still marks the Read stale.

The Anthropic `is_error` flag is the reliable signal (it is what Claude Code sets). OpenAI-format tool results carry no structured error flag, so they are left exactly as today (no behavior change on that path).

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Made

- `headroom/transforms/read_lifecycle.py`: add `_failed_tool_call_ids()` (scans tool_result blocks for `is_error: true`); `_build_file_operation_index()` takes a `failed_ids` set and skips those tool calls; `apply()` threads it through.
- `tests/test_transforms/test_read_lifecycle.py`: `make_anthropic_tool_result` gains an `is_error` flag; add `test_failed_edit_does_not_make_read_stale` and `test_failed_then_successful_edit_still_makes_read_stale`.

## Testing

### Test Output

```text
$ python -m pytest tests/test_transforms/test_read_lifecycle.py -q
25 passed

$ uvx ruff@0.15.17 check headroom/transforms/read_lifecycle.py tests/test_transforms/test_read_lifecycle.py
All checks passed!
$ uvx ruff@0.15.17 format --check headroom/transforms/read_lifecycle.py tests/test_transforms/test_read_lifecycle.py
2 files already formatted
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/transforms/read_lifecycle.py
```

## Real Behavior Proof

- Exact command / steps: built a `ReadLifecycleManager(ReadLifecycleConfig(enabled=True, compress_stale=True))` and drove `apply()` on an Anthropic message list of Read(/src/app.py) then Edit(/src/app.py) whose tool_result was `is_error=True`; then repeated with a success result and with a failed-then-successful edit.
- Observed result: before the fix the read was classified STALE (`reads_stale=1`, transform `read_lifecycle:stale:/src/app.py` emitted) even though the edit failed; after the fix the same failed-edit case leaves it FRESH (`reads_stale=0`, `reads_fresh=1`, no transform), a successful edit still yields STALE, and a failed-then-successful edit still yields STALE. Ran against the actual module.
- Not tested: a live end-to-end Claude Code session (the classifier is exercised directly instead).

## Review Readiness

## Checklist

### DIFF
--- a/headroom/transforms/read_lifecycle.py
+++ b/headroom/transforms/read_lifecycle.py
@@ -131,7 +131,12 @@ class ReadLifecycleManager:
 
         # Phase 1: Build tool metadata and file operation index
         tool_metadata = self._build_tool_metadata(messages)
-        file_ops = self._build_file_operation_index(messages, tool_metadata)
+        # A tool call whose result reported an error did not actually run: a
+        # failed Edit/Write left the file unchanged (so it must not mark a prior
+        # Read stale), and a failed Read produced no content (so it can't
+        # supersede an earlier one). Drop those ops before classifying.
+        failed_ids = self._failed_tool_call_ids(messages)
+        file_ops = self._build_file_operation_index(messages, tool_metadata, failed_ids)
 
         # Phase 2: Classify each Read
         classifications = self._classify_reads(file_ops)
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest commit cb38a6a. The fix correctly removes failed Anthropic tool calls from the read/edit file-operation index before classification, so failed edits no longer mark prior reads stale while later successful edits still do. OpenAI-format results are unchanged because they lack the same structured error signal. I ran the focused test file on the PR worktree after supplying the local compiled extension artifact: tests/test_transforms/test_read_lifecycle.py passed (25 passed).

PR #2609, opened 2026-07-27
fix(tool-schema): preserve $defs/definitions subschemas named like annotations

abhay-codes07:
## Description

`compact_tool_schema_value` (the Layer 1 tool-schema compactor, run as a turn hook on outgoing tool definitions) drops JSON Schema annotation keys such as `title`, `example`, `deprecated`, `readOnly`. To avoid deleting a real field that happens to be *named* like an annotation, it guards on the parent key:

```python
if _parent_key != "properties" and key in TOOL_SCHEMA_DROP_KEYS:
    continue
```

The guard only protects names inside a `properties` object. But `properties` is not the only JSON Schema keyword whose children are names rather than annotations. Reusable subschemas live under `$defs` (2020-12) and `definitions` (draft-07), and are referenced by `$ref`. So a subschema named `title` (i.e. `$defs.title`, referenced by `$ref: "#/$defs/title"`) is dropped as if it were an annotation, which leaves a **dangling `$ref`** and corrupts the tool schema. Strict providers reject a schema with an unresolvable `$ref` with a 400.

Reproduced directly against the module:

```python
schema = {
    "type": "object",
    "properties": {"field": {"$ref": "#/$defs/title"}},
    "$defs": {"title": {"type": "string", "enum": ["a", "b"]}},
}
compact_tool_schema_value(schema)["$defs"]   # -> {}  (the $ref now dangles)
```

## Fix

Introduce `TOOL_SCHEMA_NAME_CONTAINERS` (`properties`, `$defs`, `definitions`, `patternProperties`, `dependentSchemas`) and protect drop-key names under any of them, not just `properties`:

```python
if _parent_key not in TOOL_SCHEMA_NAME_CONTAINERS and key in TOOL_SCHEMA_DROP_KEYS:
    continue
```

This only ever preserves keys that were previously dropped, so it cannot over-compress or change any schema that already worked. Annotations *inside* each subschema are still dropped (e.g. a `title` annotation on the `$defs.title` subschema is removed; its `enum` and `type` are kept), so real compression is unaffected.

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Made

- `headroom/proxy/tool_schema_compaction.py`: add `TOOL_SCHEMA_NAME_CONTAINERS`; `compact_tool_schema_value` protects drop-key names under any name container, not only `properties`.
- `tests/test_tool_schema_compaction.py`: add `test_preserves_defs_subschema_named_like_annotation` and `test_preserves_definitions_subschema_named_like_annotation`.

## Testing

### Test Output

```text
$ python -m pytest tests/test_tool_schema_compaction.py -q
29 passed

# with the fix reverted, the two new tests fail (the $defs/definitions
# subschema is dropped, leaving a dangling $ref):
$ git stash push headroom/proxy/tool_schema_compaction.py && \
    python -m pytest tests/test_tool_schema_compaction.py -k named_like_annotation -q
2 failed

$ uvx ruff@0.15.17 check headroom/proxy/tool_schema_compaction.py tests/test_tool_schema_compaction.py
All checks passed!
$ uvx mypy@1.20.2 --ignore-missing-imports headroom/proxy/tool_schema_compaction.py
```

## Real Behavior Proof

- Exact command / steps: called `compact_tool_schema_value` on a schema whose `properties.field` is `{"$ref": "#/$defs/title"}` with a matching `$defs.title` subschema (that also carried a `title` annotation and an `enum`); then reverted the source and re-ran the two new tests.
- Observed result: with the fix the `$defs.title` subschema is preserved (so the `$ref` resolves), the inner `title` annotation is dropped, and `enum`/`type` are kept; a field literally named `deprecated` under `properties` still survives; with the fix reverted the whole `$defs.title` subschema is dropped (`$defs` becomes `{}`) and the `$ref` dangles. Ran against the actual module.
- Not tested: a live request to a provider that 400s on an unresolvable `$ref` (the corruption is verified at the transform boundary instead).

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/tool_schema_compaction.py
+++ b/headroom/proxy/tool_schema_compaction.py
@@ -56,6 +56,22 @@ TOOL_SCHEMA_DROP_KEYS: frozenset[str] = frozenset(
     }
 )
 
+# JSON Schema keywords whose direct children are *names* (property names,
+# reusable-subschema names, dependency triggers), NOT schema annotations. A
+# child key under one of these must never be dropped even when it happens to
+# equal a name in TOOL_SCHEMA_DROP_KEYS: e.g. a `$defs` entry named "title" is a
+# reusable subschema (referenced by `$ref: #/$defs/title`), and dropping it
+# leaves a dangling reference that breaks the tool schema.
+TOOL_SCHEMA_NAME_CONTAINERS: frozenset[str] = frozenset(
+    {
+        "properties",
+        "$defs",
+        "definitions",
+        "patternProperties",
+        "dependentSchemas",
+    }
+)
+
 # Parameter names that are self-explanatory.  When ``description``
 # matches the name (case-insensitive prefix), it can be stripped.
 _SEMANTIC_PARAM_NAMES: frozenset[str] = frozenset(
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest commit 409ab88. The change correctly treats JSON Schema name containers beyond properties as places where child keys are schema names rather than annotations, while still dropping annotations inside those subschemas. This prevents dangling $ref targets without reducing compaction elsewhere. I ran the focused test file on the PR worktree: tests/test_tool_schema_compaction.py passed (29 passed).

PR #2619, opened 2026-07-27
fix(wrap): set ANTHROPIC_HOST so `wrap goose` actually proxies Anthropic

filipkujawa:
## Description

`headroom wrap goose` never proxied Anthropic traffic. Goose's Anthropic provider reads `ANTHROPIC_HOST`, not `ANTHROPIC_BASE_URL`, so the wrapper started the proxy, printed the banner, and then sent every request straight to `api.anthropic.com` — silently, with no error and zero compression.

`ANTHROPIC_BASE_URL` appears in no `.rs` file in the goose repo. The provider reads `ANTHROPIC_HOST` at [`crates/goose/src/providers/anthropic_def.rs#L38-L40`](https://github.com/block/goose/blob/main/crates/goose/src/providers/anthropic_def.rs#L38-L40) and appends `/v1/messages` itself, so `_claude_proxy_base_url(port)` is already the correct value.

Note the README currently advertises `headroom wrap goose -- --provider anthropic` as an example — that is the exact invocation that bypasses the proxy. The OpenAI path was unaffected: goose does honour `OPENAI_BASE_URL`.

## Type of Change

## Changes Made

- Set `ANTHROPIC_HOST` alongside the existing vars in `wrap goose` (`headroom/cli/wrap.py`), and show it in the launch banner.
- Extend `test_wrap_goose_sets_provider_envs` to assert the value. This test **fails before** the change with a `KeyError`.
- Update the command docstring.

## Testing

### Test Output

```text
# Test A - the env vars `wrap goose` sets today (ANTHROPIC_HOST explicitly unset)
$ env -u ANTHROPIC_HOST ANTHROPIC_API_KEY=*** \
    OPENAI_BASE_URL=http://127.0.0.1:8787/v1 \
    OPENAI_API_BASE=http://127.0.0.1:8787/v1 \
    ANTHROPIC_BASE_URL=http://127.0.0.1:8787 \
    goose run --no-session -t "Reply with exactly: PONG"

● new session · anthropic claude-haiku-4-5
PONG
===== LISTENER HITS =====
(hit count:        0)

# Test B - same, plus ANTHROPIC_HOST (this change)
(hit count: 7)
2026-07-27T12:13:09.381232 HIT POST /v1/messages host=127.0.0.1:8787
2026-07-27T12:13:09.382744 HIT POST /v1/messages host=127.0.0.1:8787
2026-07-27T12:13:10.245477 HIT POST /v1/messages host=127.0.0.1:8787
...

# Test C - OpenAI path, unchanged env (confirms scope is Anthropic-only)
(hit count: 8)  POST /v1/chat/completions
```

## Real Behavior Proof

- Environment: macOS 15 arm64, goose v1.44.0 built from source, headroom main @ e530de5, provider anthropic / claude-haiku-4-5, isolated goose config via GOOSE_PATH_ROOT, stub HTTP listener on 127.0.0.1:8787 standing in for the proxy.
- Exact command / steps: Ran `goose run --no-session -t "Reply with exactly: PONG"` three times against the stub listener — (A) with the env vars `wrap goose` sets today and ANTHROPIC_HOST explicitly unset, (B) identical but with ANTHROPIC_HOST added, (C) OpenAI provider with unchanged env. Full commands and output in the Test Output block above.
- Observed result: A got 0 requests at the listener and goose returned a real completion from api.anthropic.com, so the bypass is silent rather than a failure; B got 7 requests at the listener, proving the fix; C got 8 requests on /v1/chat/completions, confirming the OpenAI path already worked and the scope is Anthropic-only.
- Not tested: pytest, ruff and mypy were not run locally because my corporate network blocks files.pythonhosted.org so the dev environment will not install — those three boxes are deliberately left unchecked and CI should be the authority on them; the new assertion mirrors the adjacent ANTHROPIC_BASE_URL one against the same `anthropic_base` value.

## Review Readiness

## Checklist

## Additional Notes

- Folded the assertion into the existing env-fan-out test rather than adding a near-duplicate, to match the file's stated organisation.
- On port fallback, `_launch_tool` rewrites the child environment (including the new var) but not `env_vars_display`, so the printed value can show the requested port. Pre-existing for the other displayed URLs; not addressed here.
- Possible follow-ups, happy to file separately: `wrap aider` sets `OPENAI_API_BASE` but not `OPENAI_BASE_URL`; and `wrap goose` does not call `with_project_prefix()`, so goose traffic isn't attributed per project.


JerrettDavis (approved):
Reviewed the latest commit 6e85f45. The fix correctly adds ANTHROPIC_HOST using the Anthropic proxy base (without the OpenAI /v1 suffix) while preserving the existing OpenAI envs, and the test now locks that behavior. I ran the focused Goose wrapper tests on the PR worktree: tests/test_cli/test_wrap_goose.py passed (2 passed).

JerrettDavis (approved):
Re-reviewed at `e14190d4`. The surviving change is limited to wiring Goose’s documented `ANTHROPIC_HOST` endpoint override alongside the existing compatibility variables, with an exact-value regression test. The branch is conflict-free, `git diff --check` is clean, and `uv run pytest tests/test_cli/test_wrap_goose.py -q` passes (2 tests).

Issue #2616, opened 2026-07-27
CCR entries expire mid-session — 30-min wall-clock TTL is shorter than a normal multi-agent burst, causing headroom_retrieve "Entry not found" on live results


JohnSchweden:
# headroom Issue Draft

**Repo:** https://github.com/chopratejas/headroom/issues
**Package:** `headroom-ai 0.32.1`
**Area:** `headroom mcp serve` — `cache/compression_store.py`

---

## Title

CCR entries expire mid-session — 30-min wall-clock TTL is shorter than a normal multi-agent burst, causing `headroom_retrieve` "Entry not found" on live results

### Summary

`DEFAULT_CCR_TTL_SECONDS = 1800` (30 minutes, `cache/compression_store.py:51`) is a fixed wall-clock TTL that is session-oblivious. When parallel sub-agents produce large tool results that get CCR-compressed on arrival, any downstream step that reaches back to those compressed entries more than 30 minutes later finds them expired. The `headroom_retrieve` tool returns `"Entry not found (CCR TTL: 1800 seconds)"`, losing the original data.

### Root cause

The TTL clock starts on compression (when the large tool result lands in the parent context). In a multi-agent orchestration flow:

1. A burst of parallel sub-agents finishes over a ~25-minute window.
2. Their results are CCR-compressed as they arrive (first result at T+0 min, last at T+25 min).
3. The orchestrator continues processing (user review, verification steps, follow-up agents).
4. At ~T+40–49 min (relative to the first result), the orchestrator reaches back to examine those compressed artifacts → they've expired.
5. `retrieve` fails silently or returns "Entry not found".

The TTL is measured from each entry's compression time, not from session start or last access. A burst of any duration >30 min from first-to-last result to last access will trigger this.

### Measured evidence (production session)

- **7 sub-agents** dispatched in parallel, finishing over a **~25 min window** (22:52–23:12).
- **First expired-entry error** at 23:41 — **~29 min after the last result landed, ~49 min after the first**. The earliest-compressed entries were 40–49 min old, well past the 30-min TTL.
- **5 misses** in a ~1-minute cluster (23:41–23:42).
- **2 more misses** on session resume the next morning (entries hours old).
- Config: `mcpServers.headroom.env` = `{}` — `HEADROOM_CCR_TTL_SECONDS` was never set.

This is not an edge case limited to long-running sessions: a single **~1-hour burst** of parallel sub-agents already triggers it.

### Expected behavior

[...]

PR #2631, opened 2026-07-28
fix(ccr): sliding idle-window TTL with max-lifetime ceiling in the Rust core backends (#2604)

lifeodyssey:
## Description

Rust-core counterpart of the CCR mid-session expiry fix. #2604 (and its duplicate #2616) report that the 30-minute wall-clock TTL kills entries in the middle of a normal multi-agent burst: the clock starts at compression time and never refreshes, so an entry the session keeps touching still dies.

#2607 fixes this on the Python side by turning the TTL into an idle window that restarts on every successful retrieval, bounded by an absolute max lifetime (8x the idle TTL) — but it explicitly notes the caveat that the Rust core still measures TTL from insertion. This PR closes that gap: all three Rust CCR backends (`InMemoryCcrStore`, `SqliteCcrStore`, `RedisCcrStore`) now use the same sliding idle-window + max-lifetime-ceiling semantics as the Python `CompressionStore`.

Scoped to the Rust core only; it deliberately does not touch `DEFAULT_TTL`'s value (1800), which #2607 bumps to 3600 — happy to rebase in lockstep whichever lands first.

Refs #2604, #2616. Complements #2607.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `crates/headroom-core/src/ccr/mod.rs`: `DEFAULT_MAX_LIFETIME_MULTIPLIER = 8` + `max_lifetime_for()` helper; documents the idle-window semantics.
- `in_memory.rs`: entries track `last_accessed`; a hit refreshes it under the shard write lock (`get_mut`), expiry checks idle window OR max lifetime, and the existing `remove_if` TOCTOU protection now uses the same predicate. New `with_capacity_and_ttls` constructor for independent control of window and ceiling.
- `sqlite.rs`: new `last_accessed` column (legacy DBs migrated in place via `ALTER TABLE`, backfilled from `created_at` so old rows keep their original expiry baseline); lazy purge and the lookup honour both bounds; a hit touches the row under the same connection mutex as the read. New `open_with_ttls` constructor.
- `redis.rs`: a hit re-arms the key's expiry, capped by a companion `{prefix}:{hash}:born` key whose remaining TTL marks the absolute ceiling; entries written by pre-sliding builds (no born key) are backfilled rather than dropped.
- `tests/ccr_backends.rs`: 6 new tests — sliding-window survival and max-lifetime cap for in-memory and SQLite, legacy-schema migration, and a gated Redis sliding test.

No public API is broken: existing constructors keep their signatures and derive the ceiling as 8x the idle TTL.

## Testing

- [x] Unit tests pass (`cargo test -p headroom-core`)
- [x] Linting passes (`cargo clippy -p headroom-core --all-features`, `cargo fmt --check`)
- [ ] Type checking passes (`mypy headroom`) — N/A, no Python files touched

### Test Output

```text
$ cargo test -p headroom-core --test ccr_backends
test result: ok. 12 passed; 0 failed; 0 ignored (8.12s)

$ cargo test -p headroom-core ccr          # all ccr-named tests across suites
38 passed, 949 filtered out (13 suites)

$ cargo test -p headroom-core --test ccr_roundtrip --test live_zone_ccr
18 passed (2 suites)

$ cargo check -p headroom-core --features redis   # cfg-gated backend compiles
Finished `dev` profile in 25.09s

$ cargo clippy -p headroom-core --all-features
No issues found
```

## Real Behavior Proof

- Environment: macOS (Darwin 25.5), local checkout at upstream `main` (57bf720d), `cargo test`.
- Exact command / steps: dropped a proof test file (`ccr_sliding_ttl_proof.rs`, uses only APIs present on both main and this branch) into `crates/headroom-core/tests/`, ran it against unpatched `main` src, then against this branch. The in-memory case touches an entry every 60ms with a 120ms TTL; the SQLite case touches at t+2s with a 3s TTL and reads again at t+4s — i.e. the issue's "session keeps using the entry" timeline scaled down.
- Observed result: on unpatched `main` both proof tests fail (in-memory: "entry vanished on touch #2 despite constant access"; SQLite: "entry expired at t+4s even though the session touched it at t+2s"); on this branch the same tests pass 2/2. Full output:

  Before (main, wall-clock TTL):

  ```text
  ---- proof_in_memory_entry_survives_while_session_keeps_touching_it stdout ----
  panicked: entry vanished on touch #2 despite constant access

  ---- proof_sqlite_entry_survives_while_session_keeps_touching_it stdout ----
  panicked: entry expired at t+4s even though the session touched it at t+2s (wall-clock TTL)

  test result: FAILED. 0 passed; 2 failed
  ```

  After (this branch, sliding idle window):

  ```text
  test result: ok. 2 passed; 0 failed (4.01s)
  ```

- Not tested: the Redis backend against a live Redis (the new `redis_get_refreshes_idle_ttl` test self-skips without `HEADROOM_TEST_REDIS_URL`, same as the existing gated tests; it compiles under `--features redis` and runs in the CI redis matrix).

## Review Readiness

## Checklist

## Additional Notes

- Docs: `docs/content/docs/ccr.mdx` TTL wording is being updated by #2607; not duplicated here to avoid conflicting hunks.
- The SQLite migration is intentionally in-place and idempotent (`pragma_table_info` check → `ALTER TABLE ADD COLUMN` → backfill), so a proxy restarting onto an existing `ccr.sqlite` keeps its rows.
- If #2607 lands first I will rebase; the only expected overlap is the doc comment around `DEFAULT_TTL`.

### DIFF
--- a/crates/headroom-core/src/ccr/backends/in_memory.rs
+++ b/crates/headroom-core/src/ccr/backends/in_memory.rs
@@ -15,13 +15,16 @@ use std::time::{Duration, Instant};
 
 use dashmap::DashMap;
 
-use crate::ccr::{CcrStore, DEFAULT_CAPACITY, DEFAULT_TTL};
+use crate::ccr::{max_lifetime_for, CcrStore, DEFAULT_CAPACITY, DEFAULT_TTL};
 
 /// In-memory CCR store backed by [`DashMap`] for sharded concurrent
 /// access.
 ///
-/// - **TTL**: 30 minutes by default. Entries past their TTL are dropped
-///   on the next `get` (lazy expiry — no background reaper thread).
+/// - **TTL**: 30 minutes by default, treated as an **idle window** —
+///   every successful `get` restarts the entry's clock (#2604), bounded
+///   by an absolute max lifetime of 8x the idle TTL measured from
+///   insertion. Entries past their window are dropped on the next `get`
+///   (lazy expiry — no background reaper thread).
 /// - **Capacity**: 1000 entries by default. When `put` would push us
 ///   past capacity, the oldest entry (per insertion order) is evicted.
 /// - **Concurrency**: gets and puts on distinct keys do not contend.
@@ -36,6 +39,7 @@ pub struct InMemoryCcrStore {
     /// they actually evict a real entry.
     order: Mutex<VecDeque<String>>,
     ttl: Duration,
+    max_lifetime: Duration,
     capacity: usize,
 }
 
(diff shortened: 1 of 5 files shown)


JerrettDavis (approved):
Reviewed the sliding TTL changes across the Rust CCR backends. The in-memory and SQLite paths now refresh an idle window on successful reads while still enforcing the max-lifetime ceiling, and the SQLite legacy migration backfills `last_accessed` from `created_at` as expected. Redis uses the companion born key to cap re-arms; the non-Redis test target still compiles and exercises the unsupported path locally.

Local verification on `8999900f`:
- `cargo test -p headroom-core --test ccr_backends` -> 12 passed
- `cargo fmt --all -- --check` -> passed

Issue #2578, opened 2026-07-26
Pricing lookup for internal `passthrough:*` model names spams "LLM Provider NOT provided" warnings via LiteLLM


w3a11y:
## Description

For requests routed through internal passthrough handlers (e.g. Anthropic's
`/v1/messages/count_tokens` endpoint), headroom assigns a synthetic model
name of the form `passthrough:<endpoint_name>` for dashboard/accounting
purposes:

https://github.com/headroomlabs-ai/headroom/blob/main/headroom/proxy/handlers/openai.py#L360

```python
def _passthrough_model_from_path(path: str, endpoint_name: str) -> str:
    ...
    return f"passthrough:{endpoint_name}"
```

This synthetic name is then passed straight into `litellm.cost_per_token()`
for cost estimation:

https://github.com/headroomlabs-ai/headroom/blob/main/headroom/proxy/cost.py#L709-L731

```python
input_cost, output_cost = litellm.cost_per_token(
    model=resolved_model,
    ...
)
```

Since `passthrough:count_tokens` isn't a real provider/model string, LiteLLM
raises `BadRequestError`, which is caught and logged as a `WARNING` on every
single call:

```
headroom.proxy - WARNING - Failed to get pricing for model passthrough:count_tokens: litellm.BadRequestError: LLM Provider NOT provided. Pass in the LLM provider you are trying to call. You passed model=passthrough:count_tokens
Pass model as E.g. For 'Huggingface' inference endpoints pass in `completion(model='huggingface/starcoder',..)` Learn more: https://docs.litellm.ai/docs/providers
```

Because `count_tokens` calls happen frequently (Claude Code calls this
endpoint before nearly every request to estimate context size), this warning
fires dozens of times per session and floods the proxy log / terminal
output.

It's also surfaced on the dashboard as a misleading pseudo-model row
(`passthrough:count_to...`) with 0 requests/tokens/savings that have no
real meaning to the user.

## Steps to Reproduce

1. Run `headroom proxy --port <port>`.
2. Point an Anthropic-compatible client (e.g. Claude Code) at the proxy.
3. Let the client make normal requests (which trigger `/v1/messages/count_tokens` calls).
4. Observe repeated `Failed to get pricing for model passthrough:count_tokens` warnings in `~/.headroom/logs/proxy.log`, and a `passthrough:count_to...` row on the dashboard's "Per-Model Token Savings" table.

## Expected Behavior

- No warning/log spam for internal passthrough accounting labels — cost
[...]

Issue #2515, opened 2026-07-23
bug: Vertex model names with @YYYYMMDD suffix show $0.00 cost (pricing lookup fails)


yossiovadia:
## Summary

When Claude Code routes through a proxy via Vertex AI, the model names include date suffixes like `claude-haiku-4-5@20251001`, `claude-opus-4@20250514`, `claude-opus-4-1@20250805`. The `resolution_candidates()` function in `headroom/pricing/litellm_model_resolution.py` tries these verbatim, but LiteLLM's database doesn't have entries with `@YYYYMMDD` suffixes — resulting in `$0.00` cost for all affected models.

## Affected models (Vertex-routed)

| Model name received | LiteLLM key needed | Cost |
|---|---|---|
| `claude-haiku-4-5@20251001` | `claude-haiku-4-5` | \$1.00/MTok |
| `claude-opus-4@20250514` | `vertex_ai/claude-opus-4` | \$15.00/MTok |
| `claude-opus-4-1@20250805` | `claude-opus-4-1` | \$15.00/MTok |
| `claude-sonnet-4-6` | `claude-sonnet-4-6` | \$3.00/MTok ✓ (no suffix, works) |

## Root cause

`resolution_candidates()` and `pricing_lookup_candidates()` in `litellm_model_resolution.py` never strip the `@YYYYMMDD` version suffix before lookup, and never try the `vertex_ai/` prefix (only `anthropic/`, `openai/`, etc.).

```python
# Current — tries "anthropic/claude-haiku-4-5@20251001" which doesn't exist
candidates = [model]
candidates.extend(rule.candidate_for(model) for rule in MODEL_PREFIX_RULES ...)

# Needed: also try bare name (suffix stripped) and vertex_ai/ prefix
bare = re.sub(r'@\d{8}$', '', model)  # "claude-haiku-4-5"
candidates.append(bare)
candidates.append(f'vertex_ai/{bare}')
```

## Fix

In `resolution_candidates()` and `pricing_lookup_candidates()`:

1. Strip `@YYYYMMDD` suffix and add the bare name as a candidate
2. Apply prefix rules to both the original and bare name
3. Try `vertex_ai/` prefix for Claude models

## Verified locally

```python
# Before fix
resolve('claude-haiku-4-5@20251001') -> $0.00  # fails
resolve('claude-opus-4@20250514')    -> $0.00  # fails

# After fix
[...]

Issue #2222, opened 2026-07-15
[BUG] headroom wrap copilot cannot use GPT-5.4 with GitHub Copilot CLI, while native Copilot CLI works


Jeff2Code:
## Summary

`headroom wrap copilot` cannot successfully use GPT-5.4 with GitHub Copilot CLI, even though the same GitHub account can use GPT-5.4 normally when running the native `copilot` command.

The issue appears to be specific to Headroom's wrapping/proxy layer rather than GitHub Copilot itself.

---

## Environment

- Headroom: 0.31.0
- GitHub Copilot CLI: 1.0.70
- macOS
- GitHub Copilot Enterprise

---

## Native Copilot CLI

Running the native CLI:

```bash
copilot
```

works correctly.

Inside the CLI, `/model` shows (among others):

- GPT-5.6 Sol (default)
- GPT-5.6 Terra
- GPT-5.6 Luna
- GPT-5.5
- GPT-5.4
- GPT-5.3-Codex
- GPT-5 mini
- Kimi K2.7 Code

GPT-5.4 is enabled in the GitHub Copilot settings and can be selected and used successfully.

---

## Headroom authentication

Authentication succeeds.

```bash
headroom copilot-auth status
```

returns:

```text
Status: logged in
```

---

## Reproduction

Launch Copilot through Headroom:

```bash
headroom wrap copilot -- --model gpt-5.4
```

Headroom starts successfully.

Sometimes it prints:

```text
Model "gpt-5.4" works best with wireApi: "responses".
Set COPILOT_PROVIDER_WIRE_API=responses to enable full model capabilities.
```

After sending any prompt, the request fails with:

```text
400 The requested model is not available for integrator "copilot-language-server".
```

---

## Additional observation

Trying a newer model:

```bash
headroom wrap copilot -- --model gpt-5.6
```

produces:

```text
Model "gpt-5.6" is not in the built-in catalog
```

This suggests that Headroom's built-in model catalog may not yet include the latest Copilot models.

---

## Troubleshooting already performed

Verified the following:

- ✅ Native `copilot` works correctly.
- ✅ GPT-5.4 is enabled for the GitHub account.
- ✅ GPT-5.4 can be selected and used from the native Copilot CLI.
- ✅ `headroom copilot-auth status` reports "logged in".
- ✅ `headroom wrap copilot` starts successfully.
- ✅ Setting

```bash
export COPILOT_PROVIDER_WIRE_API=responses
```

does **not** resolve the issue.

[...]

Issue #1910, opened 2026-07-09
[BUG] wrap copilot: 400 'model not supported' for Enterprise/Business tenants without BYOK key (claude-sonnet-5)


FixHarDeZ:
## Summary

`headroom wrap copilot -- --model claude-sonnet-5` always fails with `400 The requested model is not supported` for GitHub Enterprise / Copilot Business tenants when no BYOK provider key (`ANTHROPIC_API_KEY` / `OPENAI_API_KEY`) is set and `--subscription` cannot be used (Enterprise accounts are blocked from that path).

## Environment

- headroom version: 0.30.0
- OS: macOS 26.5.2 (BuildVersion 25F84)
- Copilot CLI version: 1.0.69
- Account type: GitHub Enterprise / Copilot for Business (`copilot_for_business_seat_quota`)
- Command: `headroom wrap copilot -- --model claude-sonnet-5`

## Steps to reproduce

```
headroom wrap copilot --verbose -- --model claude-sonnet-5 -p "say hi in one word"
```

## Observed behavior

Copilot CLI itself prints:

```
! Model "claude-sonnet-5" is not in the built-in catalog. Using defaults for: prompt tokens
  (COPILOT_PROVIDER_MAX_PROMPT_TOKENS), output tokens (COPILOT_PROVIDER_MAX_OUTPUT_TOKENS). Run `copilot
  help providers` for configuration details.

✗ 400 The requested model is not supported.
```

`~/.headroom/logs/proxy.log` shows:

```
apply_copilot_api_auth: incoming token not suitable (kind=gho_***), will replace
event=outbound_request forwarder=streaming method=POST path=https://api.githubcopilot.com/chat/completions body_bytes=97619 body_mutated=true mutation_reasons= source=canonical
Forwarding upstream streaming error status=400 url=https://api.githubcopilot.com/chat/completions
```

## Root cause (as far as I can trace it)

1. Because no `ANTHROPIC_API_KEY`/`OPENAI_API_KEY` is set, `_should_use_copilot_oauth()` in `headroom/cli/wrap.py` detects only a plain GitHub OAuth token (`gho_...`) and forces `COPILOT_PROVIDER_TYPE=openai` (BYOK mode) with `wire_api=completions`.
[...]

PR #2643, opened 2026-07-29
feat(copilot): drive model routing and discovery from the live model catalog

### LINKED ISSUE #1904: OpenAI chat proxy should normalize max_tokens for gpt-5.4-mini
## Summary
OpenAI-compatible clients can send `max_tokens` to the chat/completions proxy, but GPT-5 / o-series chat models reject that field and require `max_completion_tokens`.

## Reproduction
1. Send a chat/completions request with `model=gpt-5.4-mini` and `max_tokens` set.
2. The upstream OpenAI API responds with: `Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.`

## Expected
The proxy should translate legacy `max_tokens` to `max_completion_tokens` before forwarding the request so OpenAI-compatible clients continue to work unchanged.

## Notes
I added a regression test in the cloned repo to pin the normalization behavior.

msoleimanip:
## Description

A wrapped Copilot session could not reliably use more than one model. The model was effectively frozen at launch, and every mechanism that tried to compensate guessed from the model *name* — guesses the live API contradicts. This makes per-session and per-subagent model choice work across every vendor the account is served, driven by the provider's published capabilities instead of hardcoded tables.

Started as a one-directional routing fix (#1745 follow-up). Live testing against a real subscription kept surfacing further defects, each with wire evidence, so the change grew to cover the actual root cause: **Headroom was guessing four facts the upstream already publishes** — which wire API serves a model, whether it accepts `reasoning_effort` (and which values), its non-streaming output cap, and whether it is a chat model at all.

`GET /models` on the Copilot host returns all four, and is reachable with the headers Headroom already builds (`Copilot-Integration-Id: vscode-chat`).

Related: #1745, #1785. Refs #2222, #1910, #1904, #77.

## Type of Change

- [x] New feature (non-breaking change that adds functionality) — `headroom models`, catalog-driven discovery
- [x] Documentation update — BYOK single-model constraint in `TESTING-copilot-subscription.md`
- [ ] Performance improvement

## Changes Made

**Defects proven on the wire** (each reproduced against `api.githubcopilot.com`, then fixed):

| Defect | Evidence |
|---|---|
| `mai-code-1-flash-picker` is `/responses`-only but doesn't match `gpt-5*/o1*/o3*`, so name-based routing sent it to `/chat/completions` | **400** `unsupported_api_for_model` — the heuristic was wrong in the *unsafe* direction |
| `reasoning_effort` is a per-model **value set**, not a flag | `claude-opus-4.6`+`xhigh` → **400**, +`max` → 200; `gpt-5.4`+`max` → **400**, +`xhigh` → 200. The previous unconditional strip silently downgraded Opus-class subagents |
| `gpt-5.4` rejects `max_tokens` | Requires `max_completion_tokens`, accepted by all 10 probed models (closes #1904) |
| litellm returns a **dict** for `reasoning_effort` when `reasoning.summary` is present | Latent 400 the moment retention becomes conditional |
| litellm fabricates all-zero `usage` when upstream omits it | Zeros beat the fallback and wrote **zero tokens** into metrics, cost, savings ledger and request log |
| Model normalization only guarded `/v1/responses` | `/chat/completions` is the **default** wire API, so the motivating case (an agent sending `"Claude Opus 4.8"`) still 400'd on the common path |
| Normalization ran *before* the bypass gate | Explicit `x-headroom-bypass` requests were rewritten anyway |
| BYOK is single-model, so a wrapped agent had no authoritative model list | It scraped names from repo files and reported `claude-fable-5`, `claude-opus-4.8-fast`, `grok-4.5` as available — none are Copilot models |

**New modules**

- `headroom/models/copilot_catalog.py` — per-credential model catalog. Two semantics matter: **absent `supported_endpoints` means no constraint**, never "nothing works" (17 of 40 live models omit it — the whole `gpt-4*`/`gpt-3.5*` tail), and **absent `policy` means no gate**, never denied (17 models, including `gpt-5.3-codex`). Keyed by `(base_url, integration_id, token_fingerprint)` because entitlements differ per token (`copilot_auth.py:1263-1293`). Fails open to the name heuristic at every level; `HEADROOM_MODEL_CATALOG=off` disables it.
- `headroom/proxy/transport_planner.py` — selects the upstream path by intersecting published endpoints with the bridges that **exist**, never by list position. Published order is not semantic: `claude-opus-4.6` lists `/v1/messages` first, `claude-sonnet-4.6` lists `/chat/completions` first. Refuses to emit an unexecutable plan (asserted across all 40 fixture models × both inbound APIs). Clamps `reasoning_effort` to the nearest accepted rung, ties resolving *downward* since effort is spend.

**Fixes**

- Identity resolution gated on the resolved upstream being Copilot, run **after** the bypass gate, wired into **both** inbound handlers.
- Bridge: emit `max_completion_tokens`; coerce dict-valued `reasoning_effort`; stop fabricated zero `usage` from overwriting real accounting; clamp output to the non-streaming cap (16k on every `claude-*` vs a 64k streaming ceiling).
- Launcher picks the session wire API from the model's published endpoints instead of its name — this is what unblocks `/responses`-only models as the *main* model.
- Removed `gemini-2.5-pro`, `gpt-5.2`, `o1-experimental` from the static fallback. None exist live, so keeping them made a hallucinated name look *recognized*.
- Version-separator resolution against the live catalog (`claude-sonnet-4-6` → `claude-sonnet-4.6`), only on a unique match.
- `wrap` now rewrites `env_vars_display` on port fallback — the banner printed the old port next to the new one and read as a routing failure.

**New capability**

- `headroom models [--provider] [--vendor] [--tier] [--json]` — live enumeration from each provider's own endpoint. Nothing hardcoded: a provider without a credential is reported as *not enumerated* with the remedy, never replaced by a built-in list.
- The live model list is written into `.github/copilot-instructions.md` at launch (marker-guarded, refreshed in place, removed by `unwrap copilot`, `--no-model-list` opts out), restoring the model knowledge BYOK strips away.

## Testing

- [x] Type checking passes (`mypy` on every new/changed module)
- [x] Manual testing performed (real CLIs + live API, see Real Behavior Proof)

### Test Output

```text
$ .venv/Scripts/python.exe -m pytest tests/*openai*.py tests/*copilot*.py       tests/*model*.py tests/*cli*.py tests/*wrap*.py -q
3 failed, 914 passed, 19 skipped, 3 warnings in 105.53s

The 3 failures are PRE-EXISTING ON MAIN, not caused by this branch:
  tests/test_wrap_code_memory.py::test_serena_dashboard_disabled_flips_existing_config
  tests/test_wrap_code_memory.py::test_serena_dashboard_disabled_repairs_config_missing_projects
  tests/test_wrap_code_memory.py::test_serena_dashboard_disabled_preserves_registered_projects

Proven by checking out origin/main in a clean worktree and running that file
alone -- the same 3 fail with the identical assertion
(`assert parsed["web_dashboard_open_on_launch"] is False` -> `assert True is False`):
  $ git worktree add --detach /tmp/main-check origin/main
  $ cd /tmp/main-check && pytest tests/test_wrap_code_memory.py -q
  3 failed, 9 passed, 1 skipped
They date from 759209cf ("stop creating serena_config.yml"), whose tests were
not updated. Worth a separate fix; out of scope here.

$ .venv/Scripts/python.exe -m mypy headroom/models/copilot_catalog.py       headroom/proxy/transport_planner.py headroom/cli/models.py       headroom/proxy/handlers/openai.py headroom/cli/wrap.py       headroom/providers/copilot/wrap.py

$ .venv/Scripts/python.exe -m ruff check .
All checks passed!

$ .venv/Scripts/python.exe -m ruff format --check headroom/ tests/
1258 files already formatted
```

New tests: 47 catalog/planner (driven off a committed 40-model live capture), 14 bridge-defect regressions, 9 instruction-injection, 7 `headroom models` CLI.

## Real Behavior Proof

- **Environment:** Windows 11, Python 3.13.7, GitHub Copilot CLI 1.0.75 → 1.0.77, Claude Code 2.1.220, real Copilot subscription against `https://api.githubcopilot.com`, litellm 1.88.1.

- **Exact command / steps:**
  1. Direct wire probes for each defect (`/responses` with a `messages` array; `mai-code-1-flash-picker` on both endpoints; `reasoning_effort` value matrix; `max_tokens` vs `max_completion_tokens` across 10 models).
  2. Proxy-level: `headroom proxy --port 8793 --openai-api-url https://api.githubcopilot.com`, 12 scenarios covering all bridge quadrants, streaming, and a non-Copilot custom upstream.
  3. **Real CLI:** `headroom wrap copilot --subscription --port 8870 -- --model <M>` for `gpt-5.4`, `claude-opus-4.8`, `gemini-3.1-pro-preview`, `mai-code-1-flash-picker`; and `headroom wrap claude --port 8871`.
  4. Asked a wrapped session to enumerate the models it can use for a subagent, before and after the instruction fix.

- **Observed result:**
  - Proxy-level: **12/12** scenarios pass. Claude/Gemini/Kimi bridged from `/responses` return real completions with correct usage; `mai-code` held on `/responses`; `gpt-5.4` native; casual labels normalized on **both** wires; buffered stream replayed as SSE (`response.completed` present); a custom upstream with `model: "Sol"` was **not** retargeted to Copilot.
  - Real CLI, wire API chosen per model: `gpt-5.4`→`responses`, `claude-opus-4.8`→`completions`, `gemini-3.1-pro-preview`→`completions`, `mai-code-1-flash-picker`→`responses`. All four return real completions. **`mai-code` returned `400 not accessible via the /chat/completions endpoint` before the launcher fix** — that regression was only reachable through the real CLI, not through proxy-level tests, because it originates in the launcher.
  - `wrap claude` unaffected (Anthropic path untouched).
  - Model enumeration by a wrapped agent — **before:** 25 names including `claude-fable-5`, `claude-opus-4.8-fast`, `grok-4.5`, assembled by shelling out to a regex over repo files (363.5k input tokens, 1m24s). **After:** 22 names, **exact match** against the live catalog, zero hallucinated, zero missed (29.1k tokens, 16s).

- **Not tested:** the full repository test suite was not run end to end (the targeted 900+ test selection above was); an interactive in-session model switch, because **it is not achievable under BYOK** — see Additional Notes. The Codex WS `/responses` surface is untouched. CCR-retrieval / memory-tool continuation on a bridged connection remains the known gap disclosed in the original PR.

## Review Readiness

## Checklist

N/A — routing, discovery, and CLI output; no UI.

## Additional Notes

**@JerrettDavis's blocking review is addressed.** `normalize_copilot_model_id()` no longer runs before the upstream is resolved. It is now behind `resolve_copilot_model_id(..., upstream_base_url=...)`, gated on `is_copilot_api_url()`, and runs after the bypass gate. Regression test as requested: `test_non_copilot_upstreams_are_never_rewritten` asserts `"Sol"`, `"Claude Opus 4.8"`, `"GPT Sol"` and `"Gemini 3.1 Pro (Preview)"` pass through untouched for direct-OpenAI, a custom `x-headroom-base-url` gateway, and a local Ollama-style URL — plus a live proxy check that a custom upstream is not retargeted to Copilot. Thank you, it was a real bug.

**Scope limit worth knowing before merge: in-session model switching is not achievable under BYOK.** `wrap copilot` must use the CLI's provider override (its only interposition hook), and `copilot help providers` is explicit: BYOK *replaces* Copilot's model routing and requires exactly one model (`COPILOT_MODEL` / `COPILOT_PROVIDER_MODEL_ID` / `--model`, all singular — there is no list form). The CLI also never requests `/v1/models` from the provider (zero such requests across 7.6MB of `proxy.log`), so Headroom cannot populate the picker even by serving a catalog. Switching therefore means relaunching. What this PR delivers is that **any** available model now works per session and per subagent, chosen correctly — which is what the model list injection and `headroom models` exist to support. Upstream feature request (BYOK accepting a model list, or querying the provider's `/v1/models`) would be the real fix.

**Deliberately out of scope:** Anthropic-wire cross-vendor routing for Claude Code. The transport works — Copilot serves `claude-opus-4.8` on `/v1/messages`, verified **200**, zero translation — but it needs a credential lane that forwards the client's own token, `/v1/messages/count_tokens` interception (no Copilot equivalent), and an audit of model-string consumers (`models/registry.py:562-580`, `memory_tool_adapter.py:599-605`, `output_effort_policy.py:44`, pricing/ledger). Should adopt #1279's `HEADROOM_UPSTREAM_ROUTES` API rather than a new prefix scheme. Shipping it half-built would be worse than scoping it out.

**Known gaps carried forward:**
- Only the `responses→chat` bridge is implemented. A `completions`-pinned session switching to a `/responses`-only model forwards unchanged and fails loudly rather than attempting a bridge that does not exist (asserted in `test_chat_inbound_never_attempts_the_unimplemented_bridge`). The launcher fix makes this rare, since the main model now gets its native wire. `LiteLLMResponsesTransformationHandler.transform_request` is a viable path for the reverse direction — verified it produces a clean wire body after stripping two plumbing keys — but it needs a chat-completions chunk synthesizer, so it is honestly new machinery rather than reuse.
- Bridged requests are buffered (`stream: false`) and replayed as synthetic SSE, so they inherit the non-streaming output cap. Now clamped and logged rather than silently truncated.
- 13 of 40 live Copilot model IDs have no litellm pricing entry, so their cost records as unknown. Pre-existing (#2578, #2515), but this PR makes those models easier to reach, so it is worth a follow-up.

**Static tables:** the remaining hardcoded ID set is now strictly a last-resort fallback for label normalization when discovery is unreachable, with retired entries removed. The live catalog is authoritative everywhere else.

**Review provenance:** the design was reviewed by three independent models on separate lenses (adversarial correctness; the litellm bridge, streaming and accounting; end-to-end integration). All three returned BLOCK or MAJOR-REWORK, and every blocking claim was independently re-verified against the live API or source before being accepted — including one incorrect claim of my own that a reviewer caught. Several of the fixes above exist only because of that pass.


JerrettDavis (changes requested):
I found one blocking scope issue.

`normalize_copilot_model_id()` is applied immediately after parsing the `/v1/responses` body, before the handler resolves whether this request is actually going to GitHub Copilot. That means every Responses request gets the Copilot alias table applied, including direct OpenAI and `x-headroom-base-url` custom upstreams. For example, a custom upstream that intentionally accepts `model: "Sol"` or `model: "Claude Opus 4.8"` would now receive `gpt-5.6-sol` / `claude-opus-4.8` instead. This is especially surprising because the helper and PR rationale are Copilot-specific.

Please gate normalization on the resolved upstream being Copilot, e.g. after `resolved_upstream_base_url` is known and `is_copilot_api_url(resolved_upstream_base_url)` is true, then recompute routing from the normalized model. A regression test with a non-Copilot `x-headroom-base-url` and `model: "Sol"` would catch this.

Local verification I ran:

- `.venv`Scripts`python.exe -m pytest tests/test_proxy_openai_responses_copilot_chat_bridge.py tests/test_proxy_copilot_auth_hooks.py tests/test_openai_model_id_normalization.py -q` -> 26 passed
- `.venv`Scripts`python.exe -m pytest <all tests/**/*openai*.py and tests/**/*copilot*.py python files> -q` -> 527 passed, 18 skipped
- `.venv`Scripts`python.exe -m ruff check headroom/proxy/handlers/openai.py tests/test_proxy_openai_responses_copilot_chat_bridge.py tests/test_openai_model_id_normalization.py tests/test_proxy_copilot_auth_hooks.py` -> passed

msoleimanip:
> I found one blocking scope issue.
> 
> `normalize_copilot_model_id()` is applied immediately after parsing the `/v1/responses` body, before the handler resolves whether this request is actually going to GitHub Copilot. That means every Responses request gets the Copilot alias table applied, including direct OpenAI and `x-headroom-base-url` custom upstreams. For example, a custom upstream that intentionally accepts `model: "Sol"` or `model: "Claude Opus 4.8"` would now receive `gpt-5.6-sol` / `claude-opus-4.8` instead. This is especially surprising because the helper and PR rationale are Copilot-specific.
> 
> Please gate normalization on the resolved upstream being Copilot, e.g. after `resolved_upstream_base_url` is known and `is_copilot_api_url(resolved_upstream_base_url)` is true, then recompute routing from the normalized model. A regression test with a non-Copilot `x-headroom-base-url` and `model: "Sol"` would catch this.
> 
> Local verification I ran:
> 
> * `.venv\\Scripts\\python.exe -m pytest tests/test_proxy_openai_responses_copilot_chat_bridge.py tests/test_proxy_copilot_auth_hooks.py tests/test_openai_model_id_normalization.py -q` -> 26 passed
> * `.venv\\Scripts\\python.exe -m pytest <all tests/**/*openai*.py and tests/**/*copilot*.py python files> -q` -> 527 passed, 18 skipped
> * `.venv\\Scripts\\python.exe -m ruff check headroom/proxy/handlers/openai.py tests/test_proxy_openai_responses_copilot_chat_bridge.py tests/test_openai_model_id_normalization.py tests/test_proxy_copilot_auth_hooks.py` -> passed
> * `git diff --check upstream/main...HEAD` -> passed

msoleimanip:
The Copilot CLI was not working properly. We were unable to switch between AI models, and it was not possible to use different sub-agents with different AI models.

I fixed both issues and added a **native mode**, which allows users to access all available models supported by Copilot.

The Copilot CLI had several other issues that prevented it from functioning correctly. I thoroughly tested it and resolved those problems, and it is now working as expected.

Before my changes, the Copilot CLI also could not handle token reduction correctly. That issue has been fixed as well. Please let me know who to fix  ci failing. I am not able to join to Headroom Labs in Vercel with my gmail.  @JerrettDavis @SvenMeyer @Djabx  I have addressed Many major issues and now everything is working great in copilot cli. my plan is to implement Something in VS code to be able to use Headroom inside of the copilot vs code

msoleimanip:
> * #### ret scan (gitleaks)Exp

Hi @JerrettDavis @SvenMeyer @Djabx , I am not able to fix this issue as I am not able to use [vercel]. can you help for review? (https://github.com/apps/vercel)

chopratejas:
@msoleimanip  - thank you for having so many fixes for this.

@JerrettDavis  just pushed a PR which gets Copilot + VSCode working - I am wondering if that PR (which is merged to main) solves some of the issues you have also tried to solve here?

msoleimanip:
> @msoleimanip - thank you for having so many fixes for this.
> 
> @JerrettDavis just pushed a PR which gets Copilot + VSCode working - I am wondering if that PR (which is merged to main) solves some of the issues you have also tried to solve here?

I have checked all open pull requests and they are not fixing Subagents and ability to switch between models in the session.  We should be able to switch between available models and would be able to see all available models by /model and also we should be able to use different ai models in the middle of the process in the subagents. each subagent can be from different vendors and different ai models and should be dynamic

JerrettDavis (changes requested):
The original custom/non-Copilot upstream blocker is addressed: normalization/planning is now gated on the resolved Copilot target and bypass planning stays verbatim. However, the latest marker repair still has a reproducible user-data-loss case, so I cannot clear the changes request yet.

`_find_marked_block()` uses the first END marker and the last START before it. If a user documents a complete marker pair before Headroom's real generated block, it selects the user's pair. I reproduced this on `be51e957` with:

```text
User documentation:
<!-- headroom:available-models -->
DO NOT DELETE MY RULE
<!-- /headroom:available-models -->

<!-- headroom:available-models -->
## Available models
- old
<!-- /headroom:available-models -->
```

The returned span contains `DO NOT DELETE MY RULE`; `_inject_copilot_models_instructions()` will replace that span while printing a successful refresh. The current regression covers a stray START, but not the equally plausible case of documentation quoting both markers.

Please make ownership unambiguous before mutating. At minimum, validate the generated block's structure and refuse/warn on multiple or ambiguous complete pairs; a durable ownership sidecar/state is stronger if exact marker text can legitimately appear in user prose. Add regressions for a complete user pair both before and after the real block, plus a lone user pair with no generated block, proving refresh/unwrap never delete it. Invalid UTF-8 should also fail closed rather than pass through `errors="replace"` into a whole-file rewrite.

This is intentionally scoped to the destructive file mutation. The per-request transport-plan return, catalog single-flight/credential keying/stale fallback, and bypass/custom-upstream corrections are materially better on the latest source.

msoleimanip:
> @msoleimanip - thank you for having so many fixes for this.
> 
> @JerrettDavis just pushed a PR which gets Copilot + VSCode working - I am wondering if that PR (which is merged to main) solves some of the issues you have also tried to solve here?

Regarding VS code , We should be able to have the same as what I mentioned in the above message.

msoleimanip:
## Review response + CI status

### The marker data-loss finding is fixed — and the diagnosis was right in a deeper way than reported

Reproduced the exact case (`36a9a895`), then found the approach itself was unsound rather than the details. It is not only a *complete quoted pair* that breaks it: prose merely **describing** the feature was spliced away.

```markdown
# Team rules

Headroom keeps a `<!-- headroom:available-models -->` block in this file.
The block's heading is `# Available AI models (Headroom)`.

NEVER COMMIT SECRETS.

It closes with `<!-- /headroom:available-models -->` -- do not hand-edit it.
```

One launch printed `Refreshed available-models list` and deleted three lines including the security rule. Also reproduced: markers inside a fenced code block (the canonical way anyone documents a marker-guarded block) lost the entire fenced body, and the signature test was not line-anchored, so `## Available AI models (Headroom)` matched too.

**So ownership is no longer inferred from file content at all.** `640d9b87` records the SHA-256 of the exact block written, out of band in Headroom's workspace, and splices a span only when its bytes hash to that value. Nothing a user writes can forge it. Anything unproven is treated as theirs and the block is **appended** instead — a visible duplicate is recoverable; cutting a span out of someone's instructions file is not.

Two deliberate consequences:

- **`unwrap` now removes every proven copy** rather than refusing when several exist. Refusing wedged the file permanently and meant `unwrap copilot` could not undo its own durable setup. Safe because each copy is byte-identical to what we wrote.
- **The unwrap call site is guarded.** The write can raise on a read-only file, and an uncaught error aborted `unwrap` *before* it stopped the proxy — leaving the proxy running, the opposite of undoing setup.

Invalid UTF-8 now fails closed, as requested. `errors="replace"` followed by a whole-file rewrite would have substituted U+FFFD for every undecodable byte.

### The test-quality criticism was correct

5 of the 7 previous parametrisations passed against the buggy code they claimed to guard — they were tautological. Verified by `git archive` of the parent commit plus only the new test file: the suite now **fails 5 tests** against the old implementation (all four forgeable-content shapes plus the wedge), so these are genuine regressions.

Coverage now includes, each asserting the user's line survives **both** a refresh and an unwrap: prose describing the feature, markers in a fenced code block, non-line-anchored signature, markers with trailing whitespace, complete user pair before / after / both sides, lone user pair, stray START, stray END, reversed markers, and the block quoted verbatim. Plus ambiguity behaviour, UTF-8 fail-closed, byte-identity round-trips for LF and CRLF, and a normal inject → refresh → remove test so this cannot over-correct into append-only.

### CI status — both remaining items are environmental, not code

**`Vercel` — fails, not fixable from this branch.** `Authorization required to deploy`: Vercel will not hand deployment credentials to a fork PR without an explicit authorize click. This PR changes **no** docs-site files (nothing under `docs/`, no `mkdocs`, no Vercel config), so there is nothing for it to deploy. Needs a maintainer to authorize or merge past it.

**Security / CVE scan — was failing, now resolved.** Six CVEs in `aiohttp` (3.14.1) and `cryptography` (48.0.1). Not introduced here — this PR touches **zero** dependency files. The branch was 18 commits behind `main`, which had already fixed it in `0221e7f2`. Merged `main` (`658a829b`, clean, no conflicts); now on `aiohttp 3.14.3` / `cryptography 50.0.0`.

`label` and `template` pass.

### Verification

- **1017 tests pass.** The 3 failures in `tests/test_wrap_code_memory.py::test_serena_dashboard_disabled_*` are **pre-existing on `main`** — confirmed by checking out `origin/main` in a clean worktree and running that file alone (same 3 fail, identical assertion). They date from `759209cf`; out of scope here.
- `ruff check .` and `ruff format --check` clean; `mypy` clean on every changed module.
- Live end-to-end against a real Copilot subscription: BYOK re-verified unchanged (`--subscription --model claude-opus-4.8`), and `--native` verified with no `--model`, with `claude-opus-4.8`, and with `gpt-5.4`.
- Compression measured in a real native session — **87,738 tokens saved (15.3%)** across 10 requests, 572,817 → 485,088, read from the proxy's own PERF log.

### Not verified (stated rather than implied)

- An **interactive `/model` switch mid-conversation** in `--native`. All automated runs use non-interactive `-p`, so the picker is proven *populated* and any model works when selected at launch; switching mid-session needs a real TUI session.
- `ws:/responses`. The transport exists in the shipped CLI 1.0.77 but is gated behind two GitHub-controlled flags and nothing selects it today; `/responses` is deliberately **not** registered as a websocket route so it cannot silently acquire the unverified Codex WS relay. Documented limitation.
- `COPILOT_API_URL` is **undocumented** by GitHub. If a future CLI drops it the failure is silent — the CLI would talk straight to GitHub and Headroom would see nothing — so launch warns when the installed bundle does not reference it, and names how to confirm via the dashboard.

msoleimanip:
> ## Review response + CI status
> ### The marker data-loss finding is fixed — and the diagnosis was right in a deeper way than reported
> Reproduced the exact case (`36a9a895`), then found the approach itself was unsound rather than the details. It is not only a _complete quoted pair_ that breaks it: prose merely **describing** the feature was spliced away.
> 
> ```
> # Team rules
> 
> Headroom keeps a `<!-- headroom:available-models -->` block in this file.
> The block's heading is `# Available AI models (Headroom)`.
> 
> NEVER COMMIT SECRETS.
> 
> It closes with `<!-- /headroom:available-models -->` -- do not hand-edit it.
> ```
> 
> One launch printed `Refreshed available-models list` and deleted three lines including the security rule. Also reproduced: markers inside a fenced code block (the canonical way anyone documents a marker-guarded block) lost the entire fenced body, and the signature test was not line-anchored, so `## Available AI models (Headroom)` matched too.
> 
> **So ownership is no longer inferred from file content at all.** `640d9b87` records the SHA-256 of the exact block written, out of band in Headroom's workspace, and splices a span only when its bytes hash to that value. Nothing a user writes can forge it. Anything unproven is treated as theirs and the block is **appended** instead — a visible duplicate is recoverable; cutting a span out of someone's instructions file is not.
> 
> Two deliberate consequences:
> 
> * **`unwrap` now removes every proven copy** rather than refusing when several exist. Refusing wedged the file permanently and meant `unwrap copilot` could not undo its own durable setup. Safe because each copy is byte-identical to what we wrote.
> * **The unwrap call site is guarded.** The write can raise on a read-only file, and an uncaught error aborted `unwrap` _before_ it stopped the proxy — leaving the proxy running, the opposite of undoing setup.
> 
> Invalid UTF-8 now fails closed, as requested. `errors="replace"` followed by a whole-file rewrite would have substituted U+FFFD for every undecodable byte.
> 
> ### The test-quality criticism was correct
> 5 of the 7 previous parametrisations passed against the buggy code they claimed to guard — they were tautological. Verified by `git archive` of the parent commit plus only the new test file: the suite now **fails 5 tests** against the old implementation (all four forgeable-content shapes plus the wedge), so these are genuine regressions.
> 
> Coverage now includes, each asserting the user's line survives **both** a refresh and an unwrap: prose describing the feature, markers in a fenced code block, non-line-anchored signature, markers with trailing whitespace, complete user pair before / after / both sides, lone user pair, stray START, stray END, reversed markers, and the block quoted verbatim. Plus ambiguity behaviour, UTF-8 fail-closed, byte-identity round-trips for LF and CRLF, and a normal inject → refresh → remove test so this cannot over-correct into append-only.
> 
> ### CI status — both remaining items are environmental, not code
> **`Vercel` — fails, not fixable from this branch.** `Authorization required to deploy`: Vercel will not hand deployment credentials to a fork PR without an explicit authorize click. This PR changes **no** docs-site files (nothing under `docs/`, no `mkdocs`, no Vercel config), so there is nothing for it to deploy. Needs a maintainer to authorize or merge past it.
> 
> **Security / CVE scan — was failing, now resolved.** Six CVEs in `aiohttp` (3.14.1) and `cryptography` (48.0.1). Not introduced here — this PR touches **zero** dependency files. The branch was 18 commits behind `main`, which had already fixed it in `0221e7f2`. Merged `main` (`658a829b`, clean, no conflicts); now on `aiohttp 3.14.3` / `cryptography 50.0.0`.
> 
> `label` and `template` pass.
> 
> ### Verification
> * **1017 tests pass.** The 3 failures in `tests/test_wrap_code_memory.py::test_serena_dashboard_disabled_*` are **pre-existing on `main`** — confirmed by checking out `origin/main` in a clean worktree and running that file alone (same 3 fail, identical assertion). They date from `759209cf`; out of scope here.
> * `ruff check .` and `ruff format --check` clean; `mypy` clean on every changed module.
> * Live end-to-end against a real Copilot subscription: BYOK re-verified unchanged (`--subscription --model claude-opus-4.8`), and `--native` verified with no `--model`, with `claude-opus-4.8`, and with `gpt-5.4`.
> * Compression measured in a real native session — **87,738 tokens saved (15.3%)** across 10 requests, 572,817 → 485,088, read from the proxy's own PERF log.
> 

I have manually tested the functionality completely. Native mode will help us to be dynamic to use different ai models in each sessions

chopratejas:
@JerrettDavis  - whats your take on this PR? Given you tested the Copilot mode recently.

msoleimanip:
> > ## Review response + CI status
> > ### The marker data-loss finding is fixed — and the diagnosis was right in a deeper way than reported
> > Reproduced the exact case (`36a9a895`), then found the approach itself was unsound rather than the details. It is not only a _complete quoted pair_ that breaks it: prose merely **describing** the feature was spliced away.
> > ```
> > # Team rules
> > 
> > Headroom keeps a ``<!-- headroom:available-models -->`` block in this file.
> > The block's heading is ``# Available AI models (Headroom)``.
> > 
> > NEVER COMMIT SECRETS.
> > 
> > It closes with ``<!-- /headroom:available-models -->`` -- do not hand-edit it.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > One launch printed `Refreshed available-models list` and deleted three lines including the security rule. Also reproduced: markers inside a fenced code block (the canonical way anyone documents a marker-guarded block) lost the entire fenced body, and the signature test was not line-anchored, so `## Available AI models (Headroom)` matched too.
> > **So ownership is no longer inferred from file content at all.** `640d9b87` records the SHA-256 of the exact block written, out of band in Headroom's workspace, and splices a span only when its bytes hash to that value. Nothing a user writes can forge it. Anything unproven is treated as theirs and the block is **appended** instead — a visible duplicate is recoverable; cutting a span out of someone's instructions file is not.
> > Two deliberate consequences:
> > 
> > * **`unwrap` now removes every proven copy** rather than refusing when several exist. Refusing wedged the file permanently and meant `unwrap copilot` could not undo its own durable setup. Safe because each copy is byte-identical to what we wrote.
> > * **The unwrap call site is guarded.** The write can raise on a read-only file, and an uncaught error aborted `unwrap` _before_ it stopped the proxy — leaving the proxy running, the opposite of undoing setup.
> > 
> > Invalid UTF-8 now fails closed, as requested. `errors="replace"` followed by a whole-file rewrite would have substituted U+FFFD for every undecodable byte.
> > ### The test-quality criticism was correct
> > 5 of the 7 previous parametrisations passed against the buggy code they claimed to guard — they were tautological. Verified by `git archive` of the parent commit plus only the new test file: the suite now **fails 5 tests** against the old implementation (all four forgeable-content shapes plus the wedge), so these are genuine regressions.
> > Coverage now includes, each asserting the user's line survives **both** a refresh and an unwrap: prose describing the feature, markers in a fenced code block, non-line-anchored signature, markers with trailing whitespace, complete user pair before / after / both sides, lone user pair, stray START, stray END, reversed markers, and the block quoted verbatim. Plus ambiguity behaviour, UTF-8 fail-closed, byte-identity round-trips for LF and CRLF, and a normal inject → refresh → remove test so this cannot over-correct into append-only.
> > ### CI status — both remaining items are environmental, not code
> > **`Vercel` — fails, not fixable from this branch.** `Authorization required to deploy`: Vercel will not hand deployment credentials to a fork PR without an explicit authorize click. This PR changes **no** docs-site files (nothing under `docs/`, no `mkdocs`, no Vercel config), so there is nothing for it to deploy. Needs a maintainer to authorize or merge past it.
> > **Security / CVE scan — was failing, now resolved.** Six CVEs in `aiohttp` (3.14.1) and `cryptography` (48.0.1). Not introduced here — this PR touches **zero** dependency files. The branch was 18 commits behind `main`, which had already fixed it in `0221e7f2`. Merged `main` (`658a829b`, clean, no conflicts); now on `aiohttp 3.14.3` / `cryptography 50.0.0`.
> > `label` and `template` pass.
> > ### Verification
> > 
> > * **1017 tests pass.** The 3 failures in `tests/test_wrap_code_memory.py::test_serena_dashboard_disabled_*` are **pre-existing on `main`** — confirmed by checking out `origin/main` in a clean worktree and running that file alone (same 3 fail, identical assertion). They date from `759209cf`; out of scope here.
> > * `ruff check .` and `ruff format --check` clean; `mypy` clean on every changed module.
> > * Live end-to-end against a real Copilot subscription: BYOK re-verified unchanged (`--subscription --model claude-opus-4.8`), and `--native` verified with no `--model`, with `claude-opus-4.8`, and with `gpt-5.4`.
> > * Compression measured in a real native session — **87,738 tokens saved (15.3%)** across 10 requests, 572,817 → 485,088, read from the proxy's own PERF log.
> 
> I have manually tested the functionality completely

@chopratejas Give me a few minutes; I am addressing the comments properly and doing functional tests
@JerrettDavis @chopratejas @skorokithakis Thanks for the comments. Now it is ready for your review. Everything is now working dynamically.

msoleimanip:
## Update: CI failures explained + `--native` mode (not described in the body above)

Pushed `691f5405`. Four separate CI signals, and they are **not** all the same kind of thing — two were real bugs of mine, two are environmental. Detail below so a reviewer does not have to triage them.

---

### 1. `test (1)` / `test (3)` shards — REAL bug, mine, fixed

Four failures in `tests/test_cli/test_wrap_persistent.py`:

```
test_ensure_proxy_restarts_ephemeral_proxy_for_openai_api_url_mismatch   assert 8788 == 8787
test_ensure_proxy_restarts_persistent_deployment_for_feature_mismatch    assert 8788 == 8787
test_ensure_proxy_restarts_recovered_persistent_for_openai_api_url_mismatch   "ephemeral proxy should not start"
test_ensure_proxy_recovered_persistent_deployment_checks_feature_mismatch     "ephemeral proxy should not start"
```

**Why I missed it locally:** my test selection used the glob `tests/*cli*.py`, which matches top-level files but **not the `tests/test_cli/` package**. My mistake, and CI was the only thing that caught it. I now run `tests/test_cli/` explicitly.

**Root cause:** I had added a pre-emptive upstream-conflict check that took a *dedicated port*. This codebase already has an established contract for a mismatched upstream — land it in `missing`, kill the proxy, restart on the **same** port — and those four tests encode exactly that. I invented a second mechanism instead of using the existing one.

**Fix:** removed the pre-emptive check and the now-unused helper, and instead made the existing comparison **symmetric**. The real hole was that `if openai_api_url:` only fired when the caller *pinned* an upstream, so a caller wanting the **default** never noticed a proxy pinned elsewhere. That is precisely the `wrap copilot --native` → `wrap claude` case.

### 2. A regression the first fix introduced — caught in review, fixed

Routing the conflict through `missing` reintroduced the original hazard in the **attached-wrapper** case: a non-empty `missing` with other wrappers attached falls into

> `Leaving it running to avoid disrupting them; this session will use the existing proxy as-is.`

Correct for a missing *feature* (restarting would drop other sessions' in-flight requests) — **wrong for an upstream**, because sharing sends this session's traffic to a different vendor. Confirmed by driving `_ensure_proxy` directly: `wrap claude` was handed the Copilot-pinned proxy on 8787.

An upstream conflict now takes a dedicated port in that case — neither shared nor disrupting the attached session — and the isolation flag **vetoes the reuse return**, which it previously did not. Regression test added (`test_pinned_proxy_is_not_shared_even_when_another_wrapper_is_attached`).

**Why this matters beyond a test failure:** `apply_copilot_api_auth` does not forward `sk-ant-*` bearers, so a `wrap claude` session sharing a Copilot-pinned proxy has its Anthropic credential **stripped and replaced** by the Copilot token — silently, because the proxy reports healthy.

### 3. CodeQL "weak hashing algorithm on sensitive data (password)" — false positive, no code change

Flags `token_fingerprint()` in `headroom/copilot_auth.py` (SHA-256). That function is **pre-existing on `main`**; this PR added a new call site, which is why the alert surfaces here.

CodeQL's premise is password *verification*. This is not that:

- The input is an **OAuth bearer** (`gho_` / `ghs_` / `tid_`), not a user password.
- The digest is **never compared to authenticate anything** — there is no equality check on it anywhere in the tree.
- It is truncated to **12 hex chars**, so it could not verify a credential even in principle.
- It is used as an in-memory dict key (per-credential catalog isolation) and as a display ID in `headroom copilot-auth status`.

A computationally expensive KDF would be the wrong tool: this needs to be fast and stable per process, and it guards nothing. Suggest dismissing as *"used in tests / not a vulnerability"* — that needs repo-admin rights, which I do not have (`gh` returns 403 on the code-scanning API).

### 4. `Vercel` — environmental, not fixable from a fork

`Authorization required to deploy`. Vercel will not give deployment credentials to a fork PR without an explicit authorize click. **This PR changes no docs-site files** (nothing under `docs/`, no `mkdocs`, no Vercel config), so there is nothing to deploy. Needs a maintainer to authorize or merge past it.

Separately, the earlier **security/CVE scan** failure is resolved: six CVEs in `aiohttp` 3.14.1 / `cryptography` 48.0.1, not introduced here (this PR touches zero dependency files) — the branch was behind `main`, which had already fixed it in `0221e7f2`. Now on `aiohttp 3.14.3` / `cryptography 50.0.0`.

---

## `--native` mode — new since the PR body was written

The description above predates this and does not cover it.

**Problem.** BYOK is the Copilot CLI's provider override, and it *replaces* Copilot's model routing: it accepts exactly one model (13 strictly singular env vars, no list form), so `/models` collapses to the launch model and the **main** agent cannot be switched in-session. Subagents already worked; the main model did not.

**Mechanism.** The CLI resolves its entire native Copilot-API surface through `COPILOT_API_URL`. Redirecting *that* at the proxy leaves Copilot's own routing intact — the CLI still fetches `GET /models`, so the picker stays populated and `/model` switches the main agent. Found empirically (a fake HTTP catcher saw `GET /models` ×4 and the CLI failed with *"no fallback model could be resolved"* on an empty catalog), then confirmed in the shipped `app.js`, which resolves the base URL via a helper checking `process.env.COPILOT_API_URL` first.

**Opt-in and additive.** BYOK is untouched — a different auth lane, and still the only path for genuine third-party keys. `anthropic_api_url` is `None` outside native mode, pinned by a regression test.

**One non-obvious requirement:** the native CLI drives **both** wires off that single base URL — Claude models over `/v1/messages`, OpenAI models over `/responses` or `/chat/completions` — so both proxy upstreams must resolve to the Copilot host, or `/v1/messages` lands on `api.anthropic.com` and 401s. That was the first live failure and is pinned by a test. It is also what makes the proxy-sharing hazard above possible, hence the isolation work.

**Compression required a fix too.** The native base URL has no `/v1`, so the CLI's bare `/chat/completions` and `/responses` matched only the generic passthrough route — which forwards unchanged and does not compress by default. Half the picker was silently uncompressed. Both bare paths are now registered on the real handlers; `OPENAI_RESPONSES_WEBSOCKET_PATHS` was de-aliased so `/responses` does not silently acquire the unverified Codex WS relay. Measured after: `/chat/completions` 70,175 tokens saved, `/responses` 80,937 on a realistic multi-item agent payload.

**Safety guard.** `COPILOT_API_URL` is undocumented. If a future CLI drops it the failure is invisible — the CLI would talk straight to GitHub and Headroom would see nothing — so launch warns when the installed bundle does not reference it, and names how to confirm via the dashboard. (Warns rather than refuses: the probe reads an undocumented needle out of a shipped bundle and can false-negative, and refusing a CLI that works is the worse outcome.)

---

## Verification

- **582 passed** in `tests/test_cli/` — the shard CI flagged, now green. 79 across the three touched test files. `ruff` and `mypy` clean.
- **`_ensure_proxy` matrix** driven directly: pinned+attached → dedicated port; pinned+unattached → restart on the same port; plain and matching proxies → still reused, so normal launches do not regress.
- **Blast radius checked against the pre-change commit.** `wrap grok` / `wrap kimi` (which pin upstreams) behaved identically before and after — they already took a dedicated proxy against a plain one. Only the previously-broken direction changed.
- **Live, real CLI, real subscription:** `--native` with no `--model`, with `claude-opus-4.8`, and with `gpt-5.4`; BYOK re-verified unchanged; `wrap claude` unaffected. Compression in a real native session: **87,738 tokens saved (15.3%)**, 572,817 → 485,088 across 10 requests, read from the proxy's PERF log.
- Manually confirmed by the author in interactive sessions, including in-session `/model` switching.

### Known, disclosed

- `tests/test_cli/test_wrap_opencode.py` **hangs on `origin/main` too** (verified via `git archive` of `origin/main`), so it is excluded from local runs and is not from this branch.
- `tests/test_wrap_code_memory.py::test_serena_dashboard_disabled_*` (3) and `tests/test_adapter_hooks.py::...::test_sqlite_scheme` also fail/error on `origin/main`. Out of scope here.
- `ws:/responses` is not routed. The transport exists in CLI 1.0.77 but is gated behind two GitHub-controlled flags and nothing selects it today.
- Trailing-newline normalization on block removal, and the "not valid UTF-8" wording when the target path is a directory, are both left as-is on review advice — no user content is lost, and the action is correct in both cases.

JerrettDavis (approved):
Re-reviewed current head 691f5405 after the destructive marker-management changes. The prior custom-upstream normalization blocker remains fixed, and instruction-file ownership is now established out of band by an exact block digest rather than forgeable in-file structure. Ambiguous/unowned pairs fail closed or append; invalid UTF-8/NUL-bearing encodings are left byte-identical; refresh preserves line endings; unwrap removes only proven generated copies. The regression matrix covers complete user pairs before/after the real block, quoted blocks, ambiguous duplicates, lone pairs, and invalid encodings. This resolves the remaining data-loss finding. CI is not authorized beyond policy checks, and I am not asking the contributor to trigger it. No remaining source blocker.

msoleimanip:
> Re-reviewed current head [691f540](https://github.com/headroomlabs-ai/headroom/commit/691f54056ba49a0e7cbd31a5a884aa92acdc4db6) after the destructive marker-management changes. The prior custom-upstream normalization blocker remains fixed, and instruction-file ownership is now established out of band by an exact block digest rather than forgeable in-file structure. Ambiguous/unowned pairs fail closed or append; invalid UTF-8/NUL-bearing encodings are left byte-identical; refresh preserves line endings; unwrap removes only proven generated copies. The regression matrix covers complete user pairs before/after the real block, quoted blocks, ambiguous duplicates, lone pairs, and invalid encodings. This resolves the remaining data-loss finding. CI is not authorized beyond policy checks, and I am not asking the contributor to trigger it. No remaining source blocker.

@JerrettDavis Thank you for giving me the opportunity on this. What will be the next step till I see this feature in the main branch? @chopratejas @skorokithakis Can you please help to prioritize this? I do not know what I have to do next. I do not have access to [Vercel](https://vercel.com/git/authorize?team=Headroom%20Labs&slug=headroom-labs&teamId=team_pRDLAiu3q2uEZYYNOVU4n8K6&type=github&job=%7B%22headInfo%22%3A%7B%22sha%22%3A%22691f54056ba49a0e7cbd31a5a884aa92acdc4db6%22%7D%2C%22id%22%3A%22QmWj4QBqXgXAK6KubFqfvkj1CCFj2aVkQGWyU8zHGpvPrC%22%2C%22org%22%3A%22headroomlabs-ai%22%2C%22prId%22%3A2643%2C%22repo%22%3A%22headroom%22%7D). Also I have another question. I want to integrate Headroom with VS Code so that we would be able to use it for Copilot chat with agents. Do you have any concerns or ideas? I am going to enhance it. Can you please help with this @chopratejas , @JerrettDavis ?

msoleimanip:
@chopratejas @JerrettDavis @skorokithakis can you please let me know what I have to do for next step?

msoleimanip:
@chopratejas, @JerrettDavis , @skorokithakis  can you please advise?

msoleimanip:
@JerrettDavis are you going to consider this pull request?

msoleimanip:
@chopratejas can you please help?

JerrettDavis (changes requested):
The new CAPI path has one settings-ownership/update blocker before this is safe to ship.

`enable_capi_override()` treats the mere presence of `_CAPI_MARKER_START` anywhere in `settings.json` as proof that Headroom owns a complete block. If the end marker is missing (for example after a partial Settings Sync merge), or a user has quoted/copied that marker in a comment, the update path calls `_remove_settings_block()` (which returns `False`), then `_append_settings_block()` (which returns `"already set"` because the start marker still exists), discards both return values, and reports `"updated"`. The stale CAPI URL remains active even though the CLI tells the user the new proxy/project URL was installed. With a complete user-authored marker pair, `disable_capi_override()` can instead delete their content.

Please give the settings blocks the same out-of-band/exact-content ownership proof already added for `chatLanguageModels.json`, and fail closed when a marker pair is incomplete or the current block no longer matches the recorded bytes/digest. Please cover at least a lone start marker and a user-authored complete marker pair in tests; neither case should be mutated or reported as updated.

msoleimanip:
> The new CAPI path has one settings-ownership/update blocker before this is safe to ship.
> 
> `enable_capi_override()` treats the mere presence of `_CAPI_MARKER_START` anywhere in `settings.json` as proof that Headroom owns a complete block. If the end marker is missing (for example after a partial Settings Sync merge), or a user has quoted/copied that marker in a comment, the update path calls `_remove_settings_block()` (which returns `False`), then `_append_settings_block()` (which returns `"already set"` because the start marker still exists), discards both return values, and reports `"updated"`. The stale CAPI URL remains active even though the CLI tells the user the new proxy/project URL was installed. With a complete user-authored marker pair, `disable_capi_override()` can instead delete their content.
> 
> Please give the settings blocks the same out-of-band/exact-content ownership proof already added for `chatLanguageModels.json`, and fail closed when a marker pair is incomplete or the current block no longer matches the recorded bytes/digest. Please cover at least a lone start marker and a user-authored complete marker pair in tests; neither case should be mutated or reported as updated.

Done

msoleimanip:
> The new CAPI path has one settings-ownership/update blocker before this is safe to ship.
> 
> `enable_capi_override()` treats the mere presence of `_CAPI_MARKER_START` anywhere in `settings.json` as proof that Headroom owns a complete block. If the end marker is missing (for example after a partial Settings Sync merge), or a user has quoted/copied that marker in a comment, the update path calls `_remove_settings_block()` (which returns `False`), then `_append_settings_block()` (which returns `"already set"` because the start marker still exists), discards both return values, and reports `"updated"`. The stale CAPI URL remains active even though the CLI tells the user the new proxy/project URL was installed. With a complete user-authored marker pair, `disable_capi_override()` can instead delete their content.
> 
> Please give the settings blocks the same out-of-band/exact-content ownership proof already added for `chatLanguageModels.json`, and fail closed when a marker pair is incomplete or the current block no longer matches the recorded bytes/digest. Please cover at least a lone start marker and a user-authored complete marker pair in tests; neither case should be mutated or reported as updated.

Now with this change, we are able to handle dynamic switching of AI models in VS Code-Copilot CLI  and Claude code. Also in subagents. I have tested it manually, and it looks good

JerrettDavis (changes requested):
The incomplete-marker path is fixed, but the ownership proof still fails closed only for user blocks containing extra content. When no provenance record exists, _owns_settings_block() structurally adopts any intact pair whose body is exactly the expected single key. A user-authored/copied pair containing only github.copilot.advanced.debug.overrideCapiUrl is therefore treated as ours; enable rewrites their gateway and disable deletes their setting. The new test named user-authored pair uses unrelated keys, so it does not cover the indistinguishable shape. This also contradicts the requested regression that a user-authored complete marker pair must not be mutated. Please do not infer ownership from an unrecorded in-file pair. A legacy migration needs an explicit user-confirmed/adoption path or independent evidence; absent that, preserve the bytes and report not ours. Add the exact single-key user-authored pair regression for both enable and disable.

msoleimanip:
Fixed

PR #2249, opened 2026-07-15
perf(proxy): token-mode hot-path: cut per-turn CPU and restore provider prefix-cache hit rate

nangsontay:
## Description

Two related token-mode workstreams on the Anthropic (Claude Code) path, landed together because they touch the same request handler and cache seam:

**A. Restore provider prompt-cache hit rate in `--mode token`.** Rebased onto `main`, which now carries #2193's per-conversation prefix trackers (`resolve_tracker`) — this change's earlier session-id work has been dropped. Three root causes of token-mode prompt-cache busts:

- **OpenAI-path recording (#1850):** the OpenAI chat paths (both non-streaming branches and the streaming finalizer) omitted `original_messages` on `PrefixCacheTracker.update_from_response`, so the tracker recorded the **forwarded (compressed)** bytes as "originals". The next turn's append-only check compared raw client bytes to compressed ones, diverged on the first compressed message, and never replayed the cached prefix — silently busting the provider cache every turn for OpenAI-format clients.
- **CCR injection double-inject (#2186):** proactive-expansion and memory-context injection ran unconditionally in token mode, appending to the live-zone tail even on same-messages re-requests that `overlay_cached_prefix` had already replayed byte-identically — busting the final segment and growing it unboundedly. No per-session dedup, so content from an old session resurfaced into a fresh post-`/compact` session.
- **Anthropic system-instruction 400:** with `ccr_inject_system_instructions=True`, Anthropic-native requests hit the injector's prepend branch, which inserted a `role:"system"` message the Messages API rejects outright — a 400 on every marker-bearing turn.

**B. Cut per-turn CPU on the token-mode hot path (compressor optimization).** The path was O(full transcript) per turn where it should be O(delta): every request re-deepcopied, re-counted, re-hashed, and re-parsed bytes byte-identical to the previous turn. Since the frozen prefix is byte-identical across turns by design, that work is pure overhead. Root causes:

- **Redundant deepcopies (F1):** `get_last_original/forwarded_messages()` deepcopied the full transcript per call, and the response path double-copied — 4-6 full-transcript deepcopies per turn.
- **Repeated hash walks (F4):** `CompressionCache` extracted + MD5-hashed every `tool_result` 3-4× per request, and swapped messages under the session lock.
- **Redundant token counts (F2):** up to 6 full-transcript `count_messages` per request, two synchronous on the event loop.
- **Telemetry re-parse on the critical path (F5):** waste-signal detection re-parsed the transcript inline on most compressing turns.
[...]

Issue #2186, opened 2026-07-14
CCR proactive expansion re-injects old compact summaries, increasing token usage [workaround available]


tadayoshimiura:
When using `headroom wrap claude`, the `ccr_proactive_expansion` feature caches Claude Code's `/compact` summaries and re-injects them into new sessions, increasing token usage and causing cache busts.

**Steps to reproduce:**
1. Use `headroom wrap claude` for a long session
2. Run `/compact` to summarize the conversation
3. Start a new session — Headroom re-injects the compact summary via `<headroom_proactive_expansion>` tags

**Impact:**
- Old compact summaries injected into new sessions add unnecessary tokens
- Causes cache busts (605 observed) because the injected content changes the conversation prefix
- Increases cost

**Workaround:**
```bash
export HEADROOM_NO_CCR_PROACTIVE_EXPANSION=1
```

**Expected behavior:**
Claude Code's compact summaries should be excluded from CCR proactive expansion.

**Environment:**
- Headroom v0.31.0
- Claude Code with Anthropic backend

Nelsonochoam:
I have been experiencing something similar. I haven't dug into the code but what I started seeing is that when I start a new session wrapped with headroom and send a single message all of a sudden the context jumps to 150% utilization without claude even replying

PR #2648, opened 2026-07-29
perf(proxy): token-mode hot path: dedup CCR proactive expansion, cut per-turn CPU, restore prefix-cache hit rate

nangsontay:
## Description

Supersedes #2249, re-rooted onto current `main` as two clean commits. This PR carries both token-mode workstreams from that branch **plus the guard wiring that branch had lost**: #2249's head was missing the Anthropic call sites for the CCR expansion guards (they were dropped in an upstream sync merge on that branch), so as pushed it still re-injected the same proactively-expanded contexts on every turn.

**1. CCR proactive-expansion re-injection loop (the bug).** Proactive expansion injects retrieved context into the **forwarded body only** — the client transcript never carries it. The next turn looks identical to the tracker (markers still present, query still relevant), so the same compressed-content hashes are expanded and re-injected on **every turn**: each request gains the full retrieval payload again, net savings degrade to zero, and the transient tail append busts the provider prefix cache from that position on. Related: #2186 (closed with a workaround — this addresses the general mechanism, not just compact summaries).

**2. Restore provider prompt-cache hit rate in `--mode token`.** The OpenAI chat paths omitted `original_messages` on `PrefixCacheTracker.update_from_response`, so the tracker recorded forwarded (compressed) bytes as "originals" and the append-only check diverged every turn (#1850). Anthropic-native requests with `ccr_inject_system_instructions=True` hit a prepend branch inserting a `role:"system"` message the Messages API 400s on.

**3. Cut per-turn CPU on the token-mode hot path.** The path was O(full transcript) per turn where it should be O(delta): redundant full-transcript deepcopies (4–6/turn), repeated tool_result hash walks, up to 6 full-transcript token counts per request, telemetry re-parse on the critical path, and repr-based router counting.

Closes #1850.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

**Commit 1 — `fix(proxy/ccr)`: expansion dedup + double-inject guards**

- `headroom/proxy/ccr_session_tracker.py`: new `SessionExpansionDedupTracker` (bounded LRU) — a hash is proactively expanded at most once per session; explicit `headroom_retrieve` calls are unaffected.
- `headroom/proxy/handlers/anthropic.py`: pass `session_id` to `track_compression`/`analyze_query`; filter recommendations through `filter_new()`; `record_injected()` only after a *real* append (the append helper is a no-op when the latest turn is frozen or has no eligible text block — a later turn may retry); gate the proactive-expansion append and the memory-context append behind `injection_target_already_forwarded()`.
- `headroom/proxy/handlers/openai.py`: same gate on the memory-context append.
- `headroom/proxy/helpers.py`: shared `injection_target_already_forwarded()` + dedup-tracker singleton.
- `headroom/cache/prefix_tracker.py`: `is_append_only_extension()` — the same append-only condition `overlay_cached_prefix` uses.
- `headroom/ccr/context_tracker.py`: entries record their session id; `analyze_query` skips contexts tracked under a different session (post-`/compact` sessions don't resurface pre-compact content). Legacy/no-session callers keep workspace-only scoping.
- `tests/test_injection_cache_safety.py` (new): unit coverage for the three guards + a tripwire asserting the anthropic handler keeps them wired (the failure mode is silent — compression still runs, savings just net to zero).

**Commit 2 — `perf(proxy)`: squash-import of #2249's workstreams**

- **openai/streaming:** pass raw client messages as `original_messages` to `update_from_response` in both non-streaming branches and the streaming finalizer.
- **anthropic system:** append CCR system instructions to top-level `body["system"]` via a static, hash-free `create_stable_system_instructions` block (idempotent, preserves prior `cache_control` breakpoints) — never a `role:"system"` message.
- **cache/prefix_tracker:** getters return internal references under a documented read-only contract; single defensive deepcopy stays at record time.
- **cache/compression_cache:** `prepare_turn` does one extract+hash walk (replacing `compute_frozen_count`/`mark_stable_from_messages`/`apply_cached`); shallow-copy swaps; hashing outside the lock.
- **transforms/pipeline + agent_savings + observability/metrics + proxy/outcome:** `defer_waste_signals` moves the telemetry-only re-parse off the critical path (Anthropic route only; background task collected done-only by the outcome funnel).
- **cache/token_count_memo (new) + tokenizers + server:** per-session `TokenCountMemo` for recount sites, guarded by an `ADDITIVE_COUNTS` tokenizer capability flag; estimator regex uses `finditer`.
- **transforms/content_router:** no `str()`-serializing of block lists; consumes `tokens_before_hint`.
- **dead code:** removed the disabled `CacheAligner` stage and `should_defer_compression`/`should_force_compress`.
- **tests:** `test_openai_cross_turn_cache_safety.py`, `test_token_count_memo.py` (new); extended anthropic cache-stability, CCR deferred-injection, cross-turn mutation-guard, pipeline waste-signal, handler-helper, and content-router suites.

## Testing

### Test Output

```text
$ HOME=$(mktemp -d) pytest <29 suites: injection_cache_safety, ccr_context_tracker, ccr_session_tracker,
    proxy_ccr, ccr_tool_injection, token_count_memo, compression_cache, cross_turn_cache_safety,
    openai_cross_turn_cache_safety, proxy_{anthropic,openai}_cache_stability, proxy_handler_helpers,
    cold_start_fast_pass, agent_savings, anthropic_stage_timings, canonical_pipeline,
    transforms_content_router, pipeline_waste_signal_limit, websearch_tool_result_protection,
    anthropic_pre_upstream_backpressure, content_router_*, config, cache/prefix_tracker,
    proxy/anthropic_ccr_deferred_injection, memory_injection_*> -q
595 passed, 4 warnings in 23.33s

$ ruff check headroom/ tests/
All checks passed!

$ mypy headroom
Found 1 error in 1 file (checked 511 source files)
# headroom/observability/metrics.py:324 — pre-existing on main (code verbatim on 1588f5e0,
# get_meter(name, version: str | None)); untouched by this PR.
```

## Real Behavior Proof

- Environment: macOS 15 (darwin 25.4.0), Python 3.13, headroom proxy 0.33.0-dev started as `headroom proxy --port 8787 --mode token --backend anthropic`, Claude Code as the client, `savings_profile=coding`.
- Exact command / steps: run a long agentic Claude Code session through the proxy so the transcript accumulates CCR compression markers, then keep asking questions whose keywords match the tracked compressed contexts, so `ContextTracker.analyze_query` recommends the same hashes on every turn. Watch `~/.headroom/logs/proxy.log` for `Proactively expanded` and compare `tok_before`/`tok_after`/`tok_saved` on the `PERF` line of each turn.
- Observed result: the two recommended hashes were re-expanded and re-injected on every subsequent turn. `tok_saved` was 1,740 on the turn before the first expansion and **0 on every turn after it**; from msgs=27 on, the forwarded request was *larger* than the client's original (`tok_after > tok_before` by +2,214 → +6,865 tokens) — the proxy added more than compression removed. One turn also logged `CACHE-BUST ... tokens_lost=8,500`, caused by the tail double-inject. Log excerpt from that session (build without these guards, i.e. the code path commit 1 changes):

  ```text
  11:47:32 PERF model=claude-opus-5 msgs=19 tok_before=31521 tok_after=29781 tok_saved=1740
  11:47:32 CCR: Proactively expanded 2 context(s) based on query relevance     # hashes A+B, 340+801 items
  11:47:37 PERF model=claude-opus-5 msgs=21 tok_before=34210 tok_after=34447 tok_saved=0
  11:47:54 headroom_retrieve ... hash=A ... hash=B                             # same hashes, next turn
  11:47:59 PERF model=claude-opus-5 msgs=27 tok_before=35893 tok_after=38107 tok_saved=0
  11:48:25 PERF model=claude-opus-5 msgs=33 tok_before=36932 tok_after=43797 tok_saved=0
  ```

- Not tested: no live-traffic run of *this* branch yet — the reproduction above is from the unguarded build, and the guarded behavior is so far only verified by the unit suites listed under Testing (dedup filter, session-scoped `analyze_query`, `injection_target_already_forwarded`, handler-wiring tripwire). The post-fix log signature to confirm on a live run is `CCR: skipping proactive expansion append — tail position already forwarded last turn` plus a `tok_saved` that stays positive across turns. Also not exercised live: the OpenAI `original_messages` recording fix (#1850) and the per-turn CPU reductions (no before/after latency profile captured). Manual testing stays unchecked and the PR stays in draft until that run is done.

## Review Readiness

## Checklist

N/A (proxy-internal behavior; log evidence above).

## Additional Notes

- **Supersedes #2249** (closed in favor of this PR). Rationale: that branch's pushed head had lost the anthropic guard wiring in an upstream sync merge (helpers/tracker/tests survived, the call sites didn't — silent because the unit tests only cover the helpers), and its history carried several main-sync merges. This PR re-roots the same content onto current `main` as two clean commits, with the guards complete. The 8 reviews on #2249 apply to content that is byte-similar here; deliberate deltas: `record_injected()` fires only after a real append, the expansion payload does not carry an unused `tokens` field, and the handler tripwire test is new.
- **Overlap with #2500:** touches the same three files (`anthropic.py`, `openai.py`, `helpers.py`) but disjoint hunks (tool-search deferral / tool_reference recovery); either merge order is trivial.
- Docs checklist item is N/A: no user-facing configuration or documented behavior changes.
- The mypy finding above is pre-existing on `main` and untouched here (see Test Output).

### DIFF
--- a/headroom/agent_savings.py
+++ b/headroom/agent_savings.py
@@ -279,6 +279,12 @@ def proxy_pipeline_kwargs(config: object) -> dict[str, object]:
     compression knobs should be consistent across Claude, Codex, and Cursor.
     """
 
+    # NOTE: defer_waste_signals is intentionally NOT set here. This dict feeds
+    # every provider handler (Anthropic, OpenAI, Gemini, batch endpoints), but
+    # only the Anthropic messages route has the off-path background task that
+    # compensates for deferral; deferring everywhere silently dropped
+    # waste-signal telemetry on the other routes. The Anthropic handler passes
+    # defer_waste_signals=True explicitly at its covered call sites.
     kwargs: dict[str, object] = {}
     profile_name = getattr(config, "savings_profile", None)
     if profile_name:
--- a/headroom/cache/compression_cache.py
+++ b/headroom/cache/compression_cache.py
@@ -6,12 +6,10 @@ Maps original content hashes to their compressed versions.
 
 from __future__ import annotations
 
-import copy
 import hashlib
 import json
 import logging
 import threading
-import time
 from collections import OrderedDict
 from dataclasses import dataclass
 
@@ -77,16 +75,22 @@ def _extract_tool_result_content(msg: dict) -> str | None:
 
 
 def _swap_tool_result_content(msg: dict, new_content: str) -> dict:
-    """Deep copy msg and replace tool result content with new_content."""
-    new_msg = copy.deepcopy(msg)
+    """Shallow-copy msg and replace tool result content with new_content.
+
+    Only the message dict, its content list, and the swapped tool_result
+    block are copied; sibling blocks stay shared refs with the input. Safe
+    because nothing downstream mutates message/block dicts in place (Phase 2
+    mutation audit) — a deepcopy here bought no safety, only cost, and ran
+    inside the cache lock (perf review F4).
+    """
     # OpenAI format
-    if new_msg.get("role") == "tool":
-        new_msg["content"] = new_content
-        return new_msg
+    if msg.get("role") == "tool":
+        return {**msg, "content": new_content}
     # Anthropic format
-    content = new_msg.get("content")
+    content = msg.get("content")
     if isinstance(content, list):
-        for block in content:
+        new_content_list = list(content)
+        for i, block in enumerate(new_content_list):
             if isinstance(block, dict) and block.get("type") == "tool_result":
                 inner = block.get("content")
                 if isinstance(inner, list):
@@ -95,11 +99,13 @@ def _swap_tool_result_content(msg: dict, new_content: str) -> dict:
                     # multiple text blocks would produce a different joined
                     # output on re-extraction (first text block replaced,
                     # remaining text blocks still joined).
-                    block["content"] = [{"type": "text", "text": new_content}]
+                    new_block_content: str | list = [{"type": "text", "text": new_content}]
                 else:
-                    block["content"] = new_content
+                    new_block_content = new_content
+                new_content_list[i] = {**block, "content": new_block_content}
                 break
-    return new_msg
+        return {**msg, "content": new_content_list}
+    return dict(msg)
 
 
 class CompressionCache:
@@ -116,7 +122,7 @@ class CompressionCache:
         # session (Claude Code background tools, parallel agents, etc.) into
         # `asyncio.to_thread` workers — without this, two concurrent
         # requests for the same `session_id` race on `_cache`,
-        # `_stable_hashes`, `_first_seen`, and `_total_tokens_saved`. The
+        # `_stable_hashes`, and `_total_tokens_saved`. The
         # observable failures are (a) lost-update on `_total_tokens_saved`,
         # (b) `OrderedDict mutated during iteration` in `apply_cached`, and
         # (c) lost stable-hash records that drop the next-turn cache lookup.
@@ -136,7 +142,6 @@ class CompressionCache:
         # `proxy/handlers/anthropic.py`) and `update_from_result`'s
         # "unchanged content" tracking.
         self._stable_hashes: set[str] = set()
-        self._first_seen: dict[str, float] = {}
         self._hits: int = 0
         self._misses: int = 0
         self._total_tokens_saved: int = 0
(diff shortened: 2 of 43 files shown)


JerrettDavis (changes requested):
I found two integration gaps that leave the advertised hot-path CPU reductions inactive:

1. `defer_waste_signals` is never passed by the Anthropic handler, and no `RequestOutcome` is populated with `waste_signals_task`. The new `_start_waste_signal_task()` helper is currently only referenced by tests, so waste-signal parsing stays inline wherever `TransformPipeline.apply()` runs with the default `defer_waste_signals=False`.

2. `TokenCountMemo` is added and `_get_token_count_memo()` is available on the proxy, but no production call site uses `count_messages_memoized()` or `_get_token_count_memo()`. The existing recount sites still call `tokenizer.count_messages(...)`, so the "up to 6 full-transcript token counts per request" part of the perf fix is not actually applied.

The correctness fixes and cache-stability guards look solid in the areas I checked, but these are material for a PR titled as a hot-path perf cut. Please either wire these into the Anthropic/proxy paths, or narrow the PR description/title so it does not claim these CPU reductions.

Local verification I ran:

- `.venv`Scripts`python.exe -m pytest tests/test_token_count_memo.py tests/test_compression_cache.py tests/test_cache/test_prefix_tracker.py -q` -> 116 passed
- `.venv`Scripts`python.exe -m pytest tests/test_injection_cache_safety.py tests/test_openai_cross_turn_cache_safety.py tests/test_cross_turn_cache_safety.py -q` -> 17 passed
- `.venv`Scripts`python.exe -m pytest tests/test_proxy_handler_helpers.py tests/test_transforms/test_pipeline_waste_signal_limit.py tests/test_agent_savings.py -q` -> 91 passed
- `.venv`Scripts`python.exe -m pytest tests/test_proxy_anthropic_cache_stability.py tests/test_proxy/test_anthropic_ccr_deferred_injection.py tests/test_cold_start_fast_pass.py -q` -> 44 passed
- `.venv`Scripts`python.exe -m ruff check <changed python files>` -> passed

nangsontay:
Closing in favour of #2706, which splits this PR along the line @JerrettDavis identified in review.

**Carried over into #2706** — the correctness half (`a0bc1ace`): the per-session expansion dedup, the session filter on `ContextTracker` entries, `injection_target_already_forwarded()`, and `tests/test_injection_cache_safety.py`. Cherry-picked onto current `main` (applies cleanly, unchanged), with all perf claims dropped from the title and description. #2706 additionally adds a tool-provenance filter to `analyze_query`, since the entries this was re-injecting turned out to be compressed instruction text (`tool_name=None`) rather than tool output.

**Not carried over** — the hot-path perf half (`34fb273a`): `defer_waste_signals` / `_start_waste_signal_task()` and `TokenCountMemo` / `count_messages_memoized()`. The review was right that neither is reachable from production code, and I confirmed the gaps are real rather than a packaging artifact of this PR: `defer_waste_signals` is never passed by the Anthropic handler, and `count_messages_memoized` has no production call site anywhere. Wiring them is unwritten work rather than a rebase, so it should come back as its own PR with the wiring included and the CPU claims measured — not ride along with a correctness fix.

Thanks for the careful review — splitting on that boundary was the right call.

PR #2657, opened 2026-07-29
feat(integrations): add provider-independent Pi/OMP extension

### LINKED ISSUE #2655: [FEATURE] Add a provider-independent Headroom context extension for Pi and OMP
## Problem Statement

The current `headroom wrap omp` integration redirects only OMP's built-in `anthropic` provider. It injects `providers.anthropic.baseUrl` into `~/.omp/agent/models.yml`; the command's own help says that OpenAI-direct, Gemini, and other providers keep their normal endpoints.

The current source makes this boundary explicit:

- `headroom/providers/omp/runtime.py` writes only `providers.anthropic.baseUrl`;
- the `headroom wrap omp` help in `headroom/cli/wrap.py` says other providers retain their normal endpoints.

This matters when OMP selects a non-Anthropic model. My initial reproduction used:

- Headroom 0.32.1
- OMP 17.1.8
- `openai-codex/gpt-5.6-sol`

OMP completed the request, but no corresponding inference request appeared in the Headroom proxy log. The generic wrapper output still said `Launching OMP (API routed through Headroom)`, so the launch looked fully wrapped even though that selected provider did not use the injected Anthropic endpoint.

I independently rechecked the routing mechanism before filing this issue:

- the active Headroom-managed `models.yml` contained an Anthropic localhost `baseUrl` override and no OpenAI/Codex `baseUrl` override;
- that Anthropic localhost endpoint was unreachable;
- `omp --no-session --model openai-codex/gpt-5.6-sol --no-tools --max-time 120 -p "Reply with exactly HEADROOM_OMP_PROVIDER_PROBE"` still returned `HEADROOM_OMP_PROVIDER_PROBE`.

This establishes that the tested GPT request did not depend on the Headroom-managed Anthropic endpoint. It does **not** establish how every OMP provider routes, and I am not making that broader claim.

### Reproduce the current wrapper limitation

Prerequisites: Headroom and OMP installed, plus an already authenticated OMP OpenAI Codex account. The probe spends one small model request. Use an unused port and a disposable log path.

Terminal 1:

```bash
rm -f /tmp/headroom-omp-repro.jsonl
headroom proxy \
  --host 127.0.0.1 \
  --port 8797 \
  --log-file /tmp/headroom-omp-repro.jsonl
```

Confirm the proxy is ready:

```bash
curl -fsS http://127.0.0.1:8797/readyz
```

Terminal 2:

```bash
touch /tmp/headroom-omp-repro.jsonl
before="$(wc -l < /tmp/headroom-omp-repro.jsonl)"

headroom wrap omp \
  --no-proxy \
  --no-context-tool \
  --port 8797 \
  -- \
  --no-session \
  --model openai-codex/gpt-5.6-sol \
  --no-tools \
  --max-time 120 \
  -p 'Reply with exactly HEADROOM_OMP_PROVIDER_PROBE'

after="$(wc -l < /tmp/headroom-omp-repro.jsonl)"
printf 'Headroom request-log lines: before=%s after=%s\n' "$before" "$after"
grep -n -A2 -B1 'anthropic:' ~/.omp/agent/models.yml
```

Expected observation with the current wrapper:

1. OMP prints `HEADROOM_OMP_PROVIDER_PROBE`.
2. The Headroom request-log line count does not increase for that probe.
3. The managed OMP configuration shows the Headroom `baseUrl` under `providers.anthropic`; it does not add an OpenAI Codex endpoint override.

Cleanup:

```bash
headroom unwrap omp
# Stop the Terminal 1 proxy with Ctrl-C.
```

This reproduces the tested OpenAI Codex bypass only. Repeat provider-specific controls before drawing conclusions about another provider.

## Why This Is Required

OMP can switch model providers without changing its extension lifecycle. Endpoint-based wrapping, however, must handle each provider's protocol, endpoint registry, and authentication behavior separately. Today, Headroom processing therefore disappears for the tested GPT path unless the user creates and maintains a custom provider configuration.

A native Pi/OMP extension provides a narrower, provider-independent capability: prepare eligible tool results through Headroom and substitute only validated, already-prepared content when the host constructs model context. It does not need to redirect inference traffic or handle provider credentials.

This would give Pi and OMP users automatic Headroom tool-result compression when their selected model is not on the wrapper's Anthropic path. It would not claim full proxy coverage for that model.

## Proposed Solution

Add a publishable Pi package, usable by both Pi and OMP, for example:

```text
@headroomlabs/pi-extension-headroom
```

Expected installation after publication:

```bash
pi install npm:@headroomlabs/pi-extension-headroom
omp plugin install npm:@headroomlabs/pi-extension-headroom
```

The proposed implementation:

1. Observes eligible text tool results and prepares them asynchronously through Headroom's `/v1/compress` endpoint.
2. Keeps the synchronous host `context` callback cache-only: no network, disk, or subprocess work occurs on that path.
3. Substitutes only a previously validated compressed result; otherwise it leaves the original result unchanged.
4. Preserves the host session transcript byte-for-byte. Only the model-facing context copy changes.
5. Keeps protected tools and the newest configured tool results raw.
6. Registers `headroom_retrieve` so a CCR marker can recover its exact source content from a bounded local cache or `/v1/retrieve`.
7. Fails open when Headroom is unavailable, unhealthy, slow, or returns an invalid/unsuitable result.
8. Exposes `/headroom status`, `/headroom stats`, `/headroom health`, `/headroom on`, and `/headroom off` commands.
9. Distinguishes prepared-entry savings from substitutions actually applied to the latest model-facing context.
10. Defaults to a loopback proxy. A remote endpoint requires explicit enablement and an exact hostname allowlist.

## Scope and Non-Goals

This proposal does **not**:

- route the selected model's inference HTTP request—including OpenAI Codex, Grok/xAI, Gemini, or another non-Anthropic provider—through the Headroom proxy;
- make that model-inference request appear in Headroom's inference-proxy request or savings statistics;
- replace `headroom wrap omp` for full request-level proxy processing on its supported Anthropic path;
- modify provider endpoints, models, API keys, OAuth sessions, or other provider credentials;
- claim support for every future Pi/OMP version or every real upstream provider without testing.

These non-goals apply only to **model-inference routing**. They do not make the extension Anthropic-only. The extension's eligibility logic does not filter by provider, so it can compress eligible tool-result context while Pi or OMP uses Grok/xAI, OpenAI, OpenAI Codex, Gemini, or another provider. In those sessions, the two paths remain separate:

- **Model inference:** Pi/OMP sends the prompt to the selected provider's normal endpoint; that request bypasses Headroom and does not appear in Headroom's inference-proxy statistics.
- **Tool-result compression:** the extension sends eligible tool results to Headroom's `/v1/compress`, substitutes validated prepared content into the model-facing context, and reports its own preparation metrics through `/headroom stats`.

Non-text or mixed-content results remain raw, as do results below policy thresholds. Actual upstream-provider accounts still require provider-specific verification; the local host gate proves the provider-independent lifecycle with deterministic test providers.

## Use Case

I use OMP with models from more than one provider, including `openai-codex/gpt-5.6-sol` or `xai-oauth/grok-4.5`. I want large tool outputs prepared and compressed without changing model-provider configuration whenever I switch models.

The value is deterministic context reduction and reversible retrieval on the tool-result path. The synthetic benchmark below measures that path; it is not a production workload, provider cost benchmark, or model-quality benchmark.

### Reopen #2645 and make `wrap omp` provider-aware

#2645 requests inference routing for additional OMP providers and was closed as `not planned`. That remains a separate architecture with provider-specific endpoint and authentication requirements. This proposal does not satisfy #2645's inference-routing acceptance criteria and should not reopen it implicitly.

### Continue #1163's transient provider-routing extension

#1163 proposes a transient Pi extension that overrides Anthropic, Google, OpenAI, and OpenAI Codex provider endpoints. It may address provider routing if completed, but it changes a different boundary. This proposal uses the host context lifecycle, does not rewrite endpoints, supports both Pi and OMP, and limits its claim to tool-result compression.

### Manual custom providers in `models.yml`

This can route selected protocols today, but it requires users to maintain provider-specific configuration and does not provide a package-level, automatic context policy.

### Documentation or warning only

The wrapper should describe provider coverage accurately, but clearer output alone does not provide Headroom tool-result compression for bypassing providers.

## Verification Performed

Fresh local verification on macOS arm64 on 2026-07-29:

- Unit suite: **8 files, 65 tests passed**.
- TypeScript: `tsc --noEmit` passed.
- Build: ESM bundle and declaration generation passed.
- Live Headroom contract: **3 tests passed** against a real local Headroom proxy:
  - real `/v1/compress` plus CCR `/v1/retrieve` validation;
  - offline fail-open followed by recovery;
  - a 1 MiB context-path p95 gate below 10 ms.
- Packed-host lifecycle:
  - **Pi 0.82.1 passed**;
  - **OMP 17.1.8 passed**.
- The packed-host test uses real Pi and OMP processes with deterministic test providers. It verified:
  - three tool results reached the next model turn;
  - the first eligible result was compressed;
  - the newest two results remained byte-exact and raw;
  - the stored session transcript retained all raw results byte-for-byte;
  - a provider/model switch did not invalidate prepared content.
- Publish dry run: **14 intended package entries**, 18,704 bytes packed and 68,028 bytes unpacked; test drivers and fixtures were excluded.
- Production dependency audit: **0 known vulnerabilities** reported by `npm audit --omit=dev`.

### Measured extension-runtime benchmark

I also ran a deterministic synthetic workload through the production `HeadroomRuntime` and `HeadroomClient` from the proposed extension against the real local Headroom 0.33.0-dev proxy on an Apple M2 Max. The model label was `openai-codex/gpt-5.6-sol`, but the benchmark sent no model-inference request; it measures only extension-side tool-result preparation and cache-only substitution.

Workload and policy:

- 8 unique, compressible JSON tool results from an unprotected `bash` tool;
- 1,149,848 original bytes total; 143,731 bytes per result on average;
- `minContextTokens=20000`, `minResultChars=4000`;
- newest 2 tool results protected raw;
- 64 MiB bounded cache;
- 100 repeated cache-only context transformations after preparation.

Observed once on 2026-07-29:

| Metric | Result |
|---|---:|
| Candidates / accepted / rejected / dropped | 8 / 8 / 0 / 0 |
| Preparation wall time | 15,484 ms |
| Prepared tokens before / after | 213,791 / 74,235 |
| Prepared tokens saved | 139,556 (65.3%) |
| Prepared bytes saved | 1,018,528 (88.6%) |
| Results substituted in sampled transform / protected raw | 6 / 2 |
| Cache-only transform median / p95 / max | 0.743 / 1.122 / 2.980 ms |

The token counts are values returned by Headroom's compression response, not a locally estimated provider bill. The token and byte savings are cumulative across accepted prepared entries. They include the two newest entries that remained raw in the sampled transform, so they represent prepared savings potential rather than realized provider-prompt or billing reduction. The byte reduction compares original result bytes with validated compressed entries. Preparation wall time includes local compression work. Transform timings exclude preparation and model/network latency. The fixture is deliberately compressible, so these savings must not be generalized to production traffic.

A separate CCR spot check parsed the first `<<ccr:...>>` marker, called `/v1/retrieve`, received 880 bytes, and confirmed that returned content occurred byte-for-byte in the original fixture.

### How to run and observe the proposed extension

From a source checkout containing the extension:

```bash

cd integrations/pi-extension
npm ci
npm run build

# In another terminal:
omp --extension "$PWD/src/index.ts"
```

Inside Pi or OMP:

```text
/headroom health
/headroom status
/headroom stats
```

`/headroom stats` currently reports session-local `candidates`, `accepted`, `rejected`, `dropped`, prepared-entry `bytes saved`, prepared-entry `tokens saved`, and `retrievals`. The savings counters advance when a compression result is accepted, not when it is substituted, so they are potential context reduction rather than provider billing data. The proposed package should label that distinction and add latest-transform applied-substitution metrics. A result becomes eligible only when it is text-only, at least 4,000 characters, from an unprotected tool, and the selected model context window meets the 20,000-token threshold. The newest two tool results remain raw. Generate at least three eligible tool results, wait for background preparation, make another model turn, and rerun `/headroom stats`.

The measured run produced:

```text
Headroom -65% · 139,556 saved
candidates 8
accepted 8
rejected 0
dropped 0
bytes saved 1,018,528
tokens saved 139,556
retrievals 0
```

The proxy log should show extension calls to `/v1/compress` and, when retrieval is requested, `/v1/retrieve`. It still will not show the selected model's inference request when that provider bypasses the proxy.

Deterministic verification commands:

```bash
cd integrations/pi-extension
HEADROOM_LIVE_BASE_URL=http://127.0.0.1:8787 npm run test:live
PI_BIN="$(command -v pi)" npm run test:hosts
```

The proposed PR should include a seeded `benchmark:live` command that emits the workload definition and metrics as JSON so maintainers can rerun the exact measurement without relying on an interactive model.

### Evidence boundaries

The deterministic host test validates the extension lifecycle without spending or depending on a real model provider. I did **not** use it to claim successful compressed conversations against every real Anthropic, OpenAI Codex, Gemini, or other account. The real GPT probe above validates the current wrapper bypass symptom, not end-to-end extension compression with GPT. Windows and Linux host runs were not part of this local verification.

## Acceptance Criteria

- Pi and OMP can install the same published extension package.
- Eligible tool results are prepared asynchronously and substituted only from validated cache entries.
- The synchronous context callback performs no network, disk, or subprocess I/O.
- Raw session history remains byte-for-byte unchanged.
- Provider/model switches do not require endpoint rewriting and do not invalidate otherwise eligible prepared entries.
- Proxy failures and invalid responses leave model context unchanged.
- CCR retrieval returns the exact source content associated with an available marker and an actionable miss when unavailable.
- Default configuration is loopback-only; remote endpoints require explicit allowlisting.
- Documentation distinguishes host-level tool-result compression from inference-proxy routing and lists the tested host versions.
- `/headroom stats` labels prepared-entry savings separately from latest-transform applied substitutions; documentation defines eligibility, session-local scope, and the distinction from inference-proxy statistics and provider billing.
- A seeded live benchmark command emits its fixture definition, policy, environment, and raw metrics as machine-readable output.
- Package, unit, live-contract, and real-host gates run in CI.

## Additional Context

Related work:

- #1149 — original OMP wrapper request
- #1811 — merged Anthropic-only OMP wrapper
- #2645 — closed proposal for provider-aware OMP inference routing
- #1163 — draft transient Pi provider-routing extension

I am willing to contribute this feature. A working implementation and the tests described above are ready in a local fork; I would open a PR after maintainers confirm that this package-level integration belongs in the repository.

gurkanguray:
The release dry-run exposed a Node 20 test compatibility issue in `Promise.withResolvers`. The test now uses a standard deferred promise; all 91 Pi tests pass under Node 20, and TypeScript checking passes. Current-head CI is pending.

Refreshed from current main: resolved README drift while preserving the Pi/OMP integration rows and durable setup instructions. The plugin source is unchanged by the merge. Version alignment and 144 lifecycle/provider/release tests pass. The wrap E2E installer also uses the tested npm 11.16.0 pin. Current-head CI is pending.

## Description  Add one provider-independent Headroom extension for Pi and Oh My Pi. The extension prepares eligible large tool results asynchronously, substitutes only validated cached output in the copied model-facing context, and preserves the raw host transcript.  This is a local-install fleet path: one loopback proxy plus `headroom-pi`. It coexists with the existing Anthropic inference-proxy wrapper. Pi wrap does **not** rewrite inference URLs. OMP wrap still owns `models.yml`; durable OMP setup never touches that file.  Closes #2655  ## Type of Change  - [ ] Bug fix (non-breaking change that fixes an issue) - [x] New feature (non-breaking change that adds functionality) - [ ] Breaking change (fix or feature that would cause existing functionality to change) - [x] Documentation update - [x] Performance improvement - [ ] Code refactoring (no functional changes)  ## Changes Made  - Added `headroom-pi` at `plugins/pi` (same layout as OpenClaw/OpenCode). One package loads in Pi and OMP. - Background tool-result preparation with bounded concurrency, backoff, cancellation, and fail-open behavior. - Copy-on-write substitution that keeps message order, tool identity, protected/recent results, and raw session history. - `/v1/compress` now sends `config.mode: "ccr"` so prepare/retrieve matches the current Headroom CCR API (review [4921293436](https://github.com/headroomlabs-ai/headroom/pull/2657#pullrequestreview-4921293436)). - Typed compression outcomes: accepted / `{status:"skipped",reason:"noop"}` / `{status:"rejected",reason}`. No-ops are skipped, not rejected. `lastError` clears after a later accept. - Session footer is `Headroom saved N tokens this session`, persisted next to the Pi session file across `/reload` and `/resume`. `/new` resets it. - Durable global-only lifecycle: `headroom init -g pi` / `headroom init -g omp` pins the exact unpublished-safe release version after npm publish. `headroom wrap pi` sets `HEADROOM_PI_BASE_URL` only. - Ownership-aware install/remove and config live in `headroom/providers/pi_extension.py`. There is no `providers/pi/` inference rewrite. - `/metrics` scrape failures return 500 with `# scrape_error ...` and `headroom_metrics_scrape_errors_total`. `headroom doctor` probes `/metrics` and the native pin. - Release gate publishes `headroom-pi` before PyPI. CI matrix: Pi `0.80.10` / `0.82.1` / `0.84.1`, OMP `17.1.8`. Workflow still uses `uv sync --frozen --extra proxy`. - Operator docs: `plugins/pi/README.md`, `wiki/cli.md`, `wiki/pi-omp-runbook.md`.  ## Testing  - [x] Unit tests pass (`pytest`) — focused Python suites below. - [x] Linting / typecheck on changed Python — full configured hook suite passed after the current-main merge, including Ruff and mypy. - [x] TypeScript unit tests pass (`npm test`). - [x] TypeScript type checking passes (`npm run typecheck`). - [x] New tests added for new functionality. - [x] Manual testing performed against a real local Headroom proxy.  ### Test Output  ```text # plugins/pi npm test Test Files  9 passed (9) Tests       91 passed (91)  npm run typecheck > tsc --noEmit  # live CCR contract against loopback proxy :8787 npm run test:live Test Files  1 passed (1) Tests       3 passed (3)  # focused Python after merging origin/main pytest tests/test_cli/test_init_pi_omp_lifecycle.py \        tests/test_cli/test_wrap_pi.py \        tests/test_cli/test_wrap_omp.py \        tests/test_cli_doctor.py \        tests/test_providers/test_pi_extension.py \        tests/test_install/test_health.py \        tests/test_release_workflows.py 219 passed  python3 scripts/verify-versions.py All versions aligned at 0.35.0 ```  ## Real Behavior Proof  - Environment: macOS/Darwin arm64, Node `v24.12.0`, checkout Headroom `0.35.0-dev` after merge with `origin/main`, loopback proxy on `:8787`. - Exact command / steps: `HEADROOM_PI_BASE_URL=http://127.0.0.1:8787 npm run test:live` in `plugins/pi`. Payload includes `config.mode: "ccr"`. Both live cases from the review (`validates real compression and retrieval...` and `fails open while offline and resumes...`) pass. - Observed result: prepared entry is present, CCR hashes retrieve as substrings of the original tool result, offline fail-open then recovery substitution works. - Earlier packed-host evidence (pre-`plugins/pi` move): Pi `0.80.10` / `0.82.1` and OMP `17.1.8` loaded the packed package; raw history unchanged; model switch reused the prepared result. That matrix was not re-run after the path rename + main merge. - Not tested: published-registry `headroom init -g` (`headroom-pi` is still npm 404); real external model accounts; non-loopback Headroom; upstream fork workflows (need maintainer approval).  ## Runtime Rollout Safety  - Rollout-managed feature(s): none. This is a local opt-in extension, not a hosted rollout flag. - Minimum rollout channel: N/A. - Stable/default behavior changed: no. Unwrapped Pi/OMP sessions are unchanged. Wrapped/durable hosts fail open if the loopback proxy is down. - Kill switch / disable path: `headroom init -g remove pi` and/or `headroom init -g remove omp`. Or stop the local proxy. - Unsafe override required: no. - Qualification impact: none beyond local Pi/OMP extension users. - Rollback path: install the previous released `headroom-ai` and rerun `headroom init -g pi` / `headroom init -g omp` so the pin matches that CLI. Do not install `latest`.  ## Review Readiness  - [x] I have performed a self-review - [x] This PR is ready for human review  Maintainer gates still pending: core-maintainer approval of #2655, permission to run fork workflows, and first `headroom-pi` npm publish before durable `init -g` can pin a released version.  ## Checklist  - [x] My code follows the project's style guidelines. - [x] I have performed a self-review of my code. - [x] I have commented code where the safety invariant or host compatibility is not obvious. - [x] I have made corresponding changes to the documentation. - [x] My changes generate no new warnings. - [x] I have added tests that prove the feature works and fails open safely. - [x] New and existing extension tests pass locally with my changes. - [x] I did **not** edit `CHANGELOG.md` — the PR title supplies the release-please entry.  ## Screenshots (if applicable)  N/A — terminal extension and host lifecycle change.  ## Additional Notes  - Review [4921293436](https://github.com/headroomlabs-ai/headroom/pull/2657#pullrequestreview-4921293436) is addressed in `plugins/pi/src/bridge.ts`: `buildCompressPayload` sends `config: { compress_user_messages: false, mode: "ccr" }`. The workflow still uses `uv sync --frozen --extra proxy`. - Pi wrap sets `HEADROOM_PI_BASE_URL` only. Package install/ownership is `headroom/providers/pi_extension.py`, not a `providers/pi/` inference analog. - Compression requests contain one synthetic tool call/result pair with empty arguments. They exclude user prompts, assistant prose, system prompts, provider payloads, and credentials. - Do not close #2655 until merge + `headroom-pi` publish.


JerrettDavis (approved):
Reviewed the Pi/OMP extension end-to-end after the latest updates. The context transformation path stays copy-on-write, retrieval is hash/substring guarded, protected/recent/error tool results remain fail-open, and the runtime/client/cache boundaries look appropriately conservative for a host extension.

I pushed one maintainer cleanup commit (`chore(pi-extension): fix whitespace`) to clear the local `git diff --check` failure. No behavior changes in that commit.

Local validation:
- `npm test` in `integrations/pi-extension` -> 69 passed
- `npm run typecheck` in `integrations/pi-extension` -> passed
- `npm run build` in `integrations/pi-extension` -> passed
- `npm audit --omit=dev --json` in `integrations/pi-extension` -> 0 vulnerabilities
- `npm pack --dry-run --json` -> package includes only the intended runtime/docs files
- `git diff --check upstream/main...HEAD` -> passed after the whitespace cleanup

GitHub is still showing only the PR governance checks queued/blocked on the fork workflow side, so this still needs the normal maintainer merge-gate handling.

gurkanguray:
Thanks for the review

Rishang:
+1

JerrettDavis (approved):
Re-approved after the CI-only repair at 36f78154. The previously reviewed feature tree is byte-identical to the approved head outside `.github/workflows/pi-extension.yml`; the workflow now installs `--extra proxy`, which supplies FastAPI before starting `headroom proxy`. The initial commit message was reworded only to wrap its overlong body line for commitlint.

Verification: `git diff 08ea9fdfc..36f78154 -- :!.github/workflows/pi-extension.yml` is empty, and all rewritten commit-message lines are within the repository’s 200-character limit.

JerrettDavis (changes requested):
The workflow/commitlint repair is correct, but the newly enabled live matrix now exposes a real product-contract blocker in both supported Pi versions.

`npm run test:live` fails identically for Pi 0.80.10 and 0.82.1:
- `validates real compression and retrieval before model-independent substitution`: no prepared entry is found (`entry` is undefined at `e2e/live-contract.test.ts:121`).
- `fails open while offline and resumes prepared substitution after recovery`: `waitFor` times out at line 190 after ~60 seconds.

This is no longer an installation/infrastructure failure—the canonical proxy starts successfully with `[proxy]` installed, and the hot-path test passes. Please reconcile the extension’s prepare/retrieve assumptions with the current Headroom CCR API, then prove both live cases pass for both matrix versions. Keep the workflow’s `uv sync --frozen --extra proxy` fix.

gurkanguray:
@JerrettDavis Live contract should be fixed — /v1/compress now sends config.mode: "ccr". Local test:live is 3/3. Please re-run the Pi matrix.

JerrettDavis (approved):
The prior live-contract blocker is resolved on the latest source. I re-ran the extension unit suite (78 tests), typecheck, build, package dry-run, a real proxy-backed live contract (3 tests), and the focused Python lifecycle/release suites (122 tests); all passed. I also audited the native-host transaction/rollback and exact-artifact release gates.

I made one small release-metadata cleanup on the branch: the npm banner now uses the durable upstream `main` URL instead of the contributor feature branch. The package gates remain green after that change. This is ready from my review.

dio16:
Design warning: strict-prefix provider cache destruction via per-request outbound rewriting

Context: I maintain an OMP setup that used the third-party `omp-headroom` plugin (DarkPhilosophy/omp-headroom). Last week its provider cache hit rate collapsed from ~98% to ~40% overnight. Root cause, fully traced with per-request usage data and a live provider test, applies directly to any extension that rewrites the model-facing conversation prefix between requests.

## The failure mechanism (in omp-headroom)

Its "session archive" transform re-ran on **every** `before_provider_request`:

- `createSessionCompaction` computes a cut = body.length − live_messages, then checks for an existing archive marker **only inside `body[cut:]`** (the last N messages). A marker inserted on a previous request sits near the head of the body (inside the prefix), so it is never detected → the transform re-fires every request.
- The marker hash is `sha256(full messages list)` → since the list grows, **every request produces a new hash and rewrites the entire history region** of the outbound payload.
- Observed signature: `cache_read` stuck at ~37k tokens (system prompt only), `input` growing monotonically with context (~33M input tokens over 492 requests). The provider's per-request usage showed the history never hitting cache again.
- Live proof of provider semantics: two identical-prefix requests hit ~98.7%; changing **one token** anywhere in the prefix collapsed the hit to 0% (strict prefix caching). Full details: https://github.com/DarkPhilosophy/omp-headroom/issues/2

## Why this is relevant to this PR

The new provider-independent extension promises to "substitute only previously prepared results in the model-facing context, and preserve the host session transcript byte-for-byte" — that is the right shape, and it avoids the *whole-history* rewrite bug by construction. The remaining risk is narrower but still real:

1. **Byte-stability of substituted content across requests.** Any transform whose output for a given message depends on global context (message count, session state, token budget, adaptive ratio) will rotate the prefix and destroy caching on strict-prefix providers (DeepSeek-family, and others with prompt-cache discounting). omp-headroom's proxy-side compressor was deterministic (content-hash cached); the instability came entirely from the archive layer recomputing per request. Make byte-stability an explicit invariant + test (same input messages ⇒ identical outbound payload region, regardless of appended turns).
2. **One-shot substitution.** Once a tool result is replaced with a prepared/cached form, never re-decide it. Any "re-archive with a new hash" pattern reintroduces the bug.
3. **Recommend a regression test** that replays a real session transcript twice (with appended turns) and asserts the outbound payload prefix is byte-identical between the two replays. That single test would have caught the omp-headroom bug immediately.

I'd be happy to provide the full measured dataset (per-request cache_read curves for healthy vs broken windows) if useful for the PR's test design.

chopratejas (commented):
Holding this one out of the current merge batch — not on quality, but because there is an open bug filed directly against this branch: **#3056, "Pi extension double-compresses RTK-truncated tool results."**

The reporter ran this PR end to end and measured the failure: 262,369 chars → 12,000 via RTK → 10,698 via Headroom. The candidate policy in `plugins/pi/src/policy.ts` treats an already-truncated `toolResult` as compressible because the pruning-metadata check doesn't look at the RTK markers. Compressing a compaction is not just wasted work — it degrades content the upstream optimizer already decided to keep.

Concretely, the metadata check needs to recognise both shapes the reporter observed:

```
details.rtkCompaction.truncated
details.metadata.rtkCompaction.truncated
```

They also specified the regression test they'd want, which is worth taking as written.

Everything else here looks ready, so I'd rather land it once with the guard than land it and immediately follow with a fix. Once #3056's case is covered and there's a test pinning it, this is good to go from my side.

Separately: CI on this branch has only run 2 checks, which is well short of the ~35 a full run produces. Worth pushing an empty commit to re-trigger so the merge decision rests on a complete run.

gurkanguray:
Thanks @chopratejas #3056 is fixed and covers both RTK metadata paths, with regression tests. Post-sync validation also passed with 91 plugin tests, 276 Python tests, build/typecheck, audit, and the full Pi/OMP matrix.  About CI, GitHub still shows only two governance checks after my commits. I see `Expected — Waiting for status to be reported` on pending checks; it seems you need to approve or run them

gurkanguray:
Hi @dio16, can you please provide the dataset? Also, I've been using this pr for a while and haven't faced this issue yet. I need to observe this more closely by reproducing it locally

dio16:
Hi @gurkanguray — here's the dataset from the omp-headroom investigation (referenced in DarkPhilosophy/omp-headroom#2). Full per-request curves + README: https://gist.github.com/dio16/e1a9b1b5bc1eba21d1a3ef8c9e44c16e

Same binary and provider/model across both windows (`opencode-go` / `deepseek-v4-flash`, `messages` API, loopback proxy :8787, Windows 11).

**Broken window** — session `01a003cc`, 2026-08-15, 214 requests (archive stats: 212 archives for 214 requests, `charsBefore` 228 MB):
- `cache_read` pinned at 37,248 tokens (= system prompt) for 93.5% of requests; p50 = p90 = 37,248. Never grows with the conversation.
- `input` grows monotonically ~70k → ~165k; 18.7M uncached input tokens; effective hit 31.7%.
- Representative rows (input / cache_read): req 5 `12,989 / 37,120`, req 100 `82,210 / 37,248`, req 150 `116,835 / 37,248`, req 210 `147,275 / 37,248`.

**Healthy window** — 08-02..08-08, 1,082 deepseek-v4-flash requests:
- `cache_read` grows with the conversation (08-04 session: 8,448 → 311,552; avg first quarter 121,459 → last quarter 279,332), `input` stays small. Per-session hit 96.3–99.3%; window total 99.1%.

**Post-fix A/B** (fresh process, `session_archive: false`): req1 `35,243 / 640` → req2 `55 / 35,840` — full history served from cache. Also 12-turn session `01a004f3`: after the cold request, every turn is `input < 200` with `cache_read ≈ 35,712–35,840`.

**Strict-prefix sensitivity** (session `01a0052c`): fresh `33,316 / 2,176` → prefix variant `27,895 / 7,936` → stable prefix `160 / 35,712`.

**Why you likely haven't seen it with this PR**: the rewrite lived in omp-headroom's *archive layer*, which recomputed a new `sha256(prefix)` marker on every request and replaced the whole history region — exactly the whole-history rewrite the copy-on-write, one-shot design here avoids by construction. That's why I framed it as a byte-stability invariant + regression-test suggestion rather than a bug in this PR.

**Reproduction recipe**: `deepseek-v4-flash` on `messages`, then per request rewrite one byte anywhere in the history region of the outbound payload (e.g. re-hash-and-replace the prefix, as `createSessionCompaction` did). Observable signature: after the first request, `cache_read` drops to system-prompt size and stays flat while `input` grows by ~full-history each turn. That exact signature is in `broken-window-01a003cc.csv`; it doubles as a fast pass/fail check for any future transform.

gurkanguray:
Thanks for sharing the dataset @dio16 . The failure came from `omp-headroom`’s per-request whole-history rewrite, while this PR substitutes individual prepared results copy-on-write; I haven’t reproduced the same behavior here. Based on current evidence, no PR change is needed, but the signature is useful for future regression testing.

JerrettDavis (approved):
Re-reviewed and refreshed the current head at eb5e1ba2. The RTK guard now recognizes both documented metadata paths (including mixed direct/nested values) and keeps malformed/non-truncated metadata eligible, so it avoids double-compressing an upstream-truncated result without broad false positives. I merged current main, resolved release metadata by keeping the 0.36.0 fleet and adding headroom-pi at 0.36.0, and aligned the Pi package/lock versions. A Windows-only mypy defect in the POSIX fcntl branch was fixed with a typing-only Any cast matching the existing msvcrt pattern.

Validation: 91 extension tests, TypeScript typecheck, 53 passed/1 skipped Pi lifecycle Python tests, production npm audit (0 vulnerabilities), version alignment, Ruff, and full commit hooks including mypy all pass. No remaining source blocker; fork CI may still require workflow authorization.

JerrettDavis (approved):
Re-reviewed and refreshed exact head 85578d22 against current upstream main. The merge conflicts were semantic but additive: release metadata now carries headroom-pi at the current 0.37.0 fleet version, while doctor retains both the Pi/OMP native-extension diagnostics and main's Codex auth diagnostic. I also aligned the Pi npm manifest and lockfile to 0.37.0 so the repository version gate remains valid. No feature behavior was removed in the integration.\n\nValidation: Pi TypeScript typecheck and declaration build pass; all 91 Pi unit tests pass; 309 focused Python tests pass with 1 skipped and only the Unix-specific live-SIGTERM test deselected on Windows; scripts/verify-versions.py reports all packages aligned at 0.37.0; and the full configured hook suite passes, including Ruff, formatting, and mypy. The remaining queued checks are fork governance/authorization gates rather than a local source failure.

JerrettDavis (changes requested):
The provider-independent extension integration remains coherent, and typechecking/build pass. The session restoration check exposed a persistence race that needs correction.

[P2] plugins/pi/src/runtime.ts:116-121 / 224-226 starts the session-state write from stop() without awaiting it, while the next runtime reads the file only once during resume. A deterministic barrier around the write reproduces 8,593 saved tokens on disk after completion but zero restored in the resumed runtime. The ordinary suite also caught this once (90 passed, 1 failed), then passed in isolation. Please make shutdown persistence awaitable and await it before resume (or use an equivalent serialized lifecycle), with a delayed-write regression proving the new instance restores the completed snapshot. Atomic replacement should also prevent partially written state from being observed.

JerrettDavis (changes requested):
The current head only merges main and still leaves the session-persistence race from fd8e3629 unchanged. HeadroomRuntime.stop() remains synchronous, calls persistSessionSavings(), and that method still fire-and-forgets saveSessionSavings(); a resumed runtime can therefore perform its one-time load before the final snapshot reaches disk. Please serialize/await the shutdown write (with atomic replacement) and add the delayed-write lifecycle regression requested previously, proving a new runtime restores the completed snapshot after shutdown.

PR #2642, opened 2026-07-29
fix(mcp): pin mcp dependency to <2.0.0 to prevent server startup crash

robert-schorr:
## Description

The MCP Python SDK recently released version `2.0.0`, which introduced breaking changes to the high-level server interface (removing `.list_tools()` and `.call_tool()` decorators on `mcp.server.Server`). Because `headroom-ai` specified `mcp>=1.28.1` without an upper bound, installing or upgrading `headroom-ai` pulled in `mcp 2.0.0`. When `headroom mcp serve` was started by an MCP client (such as OpenCode or Claude Code), the server crashed immediately on startup with `AttributeError: 'Server' object has no attribute 'list_tools'`, resulting in the connection closing error (`headroom MCP error -32000: Connection closed`).

This PR pins the `mcp` dependency to `<2.0.0` (`mcp>=1.28.1,<2.0.0`) in `pyproject.toml` so compatible 1.x SDK releases (e.g. `1.29.0`) are used.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Pinned `mcp` dependency to `"mcp>=1.28.1,<2.0.0"` under both `proxy` dependencies and the `mcp` extra in `pyproject.toml`.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [ ] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ ruff check pyproject.toml headroom/ccr/mcp_server.py
All checks passed!

$ pytest tests/test_ccr_mcp_server.py tests/test_cli/test_mcp.py tests/test_cli/test_mcp_status.py
============================== 45 passed in 1.03s ==============================
```

## Real Behavior Proof
- Environment: macOS (Darwin arm64), Python 3.14.5, uv
- Exact command / steps: Executed uv sync --all-extras and sent stdio JSON-RPC initialize and tools/list requests to .venv/bin/headroom mcp serve.
[...]

PR #2658, opened 2026-07-29
feat(mcp)!: port MCP servers to SDK 2.x

chopratejas:
## Description

MCP SDK 2.0.0 removed the `Server.list_tools()` / `call_tool()` decorators and `Server.request_context`, replacing them with constructor callbacks and a per-call context argument. #2642 pinned `mcp<2.0.0` to stop the resulting startup crash; this ports both MCP servers to the 2.x interface so the pin can eventually be lifted.

**Draft — blocked on an upstream dependency, not on review.** See "Blocker" below. The code is complete and verified against a real `mcp` 2.0.0 install; it cannot merge until `strands-agents` supports MCP 2.x.

Follow-up to #2642.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Blocker: `strands-agents` caps `mcp<2.0.0`

`uv lock --upgrade-package mcp` fails to resolve:

```text
And because strands-agents>=1.2.0,<=1.26.0 depends on mcp>=1.11.0,<2.0.0
and mcp>=1.23.0,<2.0.0, we can conclude that strands-agents>=0.1.0
depends on mcp>=1.8.0,<2.0.0.
And because headroom-ai[strands] depends on strands-agents>=0.1.0
and headroom-ai[all] depends on mcp>=2.0.0, we can conclude that
headroom-ai[all] and headroom-ai[strands] are incompatible.
```

`strands-agents` 1.50.2 (latest) declares `mcp<2.0.0,>=1.23.0`, and the `strands` extra (`pyproject.toml:215`) is included in `all`. So `uv.lock` still pins `mcp 1.28.1` in this PR and **CI is expected to fail** until one of:

1. `strands-agents` relaxes its `mcp` cap upstream (preferred — no action here);
2. `strands` is dropped from the `all` extra;
3. Headroom straddles both majors behind a compat shim (doubles the MCP test surface; not recommended).

Since #2642's pin already restored working behaviour, there is no urgency — this branch can sit until option 1 lands.

## Changes Made

- `headroom/ccr/mcp_server.py`, `headroom/memory/mcp_server.py`: register handlers as `on_list_tools` / `on_call_tool` constructor callbacks, return `ListToolsResult` / `CallToolResult`, and read `params.name` / `params.arguments`. In the memory server the `Server(...)` construction moved to the end of the factory, since 2.x needs the handler closures to exist first.
- `_current_client()` reads the per-request context from a `ContextVar` instead of the removed `Server.request_context`. This is deliberately the same mechanism 1.x used internally (`mcp/server/lowlevel/server.py:109` in 1.28.1); an instance attribute would race across concurrent tool calls.
- `Tool(inputSchema=)` → `input_schema=` (6 sites). The camelCase alias still resolves at runtime via pydantic's `populate_by_name`, but mypy rejects it.
- Normalise `params.arguments or {}` — 2.x types `arguments` as optional while every `_handle_*` helper calls `.get()` on it.
- `tests/_mcp_stub.py` mirrored the **1.x** decorator API (`list_tools()` returning `lambda fn: fn`), so every stubbed MCP test stayed green while the real server crashed on startup. Moved to the 2.x shape with the decorator surface removed, so a caller still on decorators now fails loudly.
- `tests/test_mcp_sdk2_contract.py` (new): drives the real SDK rather than the stub, including a canary asserting `Server` has no `list_tools` / `call_tool` / `request_context`.
- `pyproject.toml`: `mcp>=1.28.1,<2.0.0` → `mcp>=2.0.0` (both the `proxy` deps and the `mcp` extra).

`is_error` is deliberately left unset on error payloads, matching 1.x behaviour where the SDK wrapped a returned list as a success. Flagging error payloads as `is_error=True` is a behaviour change and belongs in its own PR.

## Testing

### Test Output

Because the lock cannot be regenerated (see Blocker), the 2.x runs used a scratch venv with `mcp==2.0.0`; the stubbed suite runs in the repo venv.

```text
$ # real mcp 2.0.0
$ PYTHONPATH=. python -m pytest tests/test_mcp_sdk2_contract.py -q
7 passed, 1 warning in 0.83s

$ PYTHONPATH=. python -m pytest tests/test_cli/test_mcp.py -q
19 passed, 1 warning in 0.98s

$ # stubbed suite, repo venv
$ python -m pytest tests/test_ccr_mcp_server.py tests/test_memory/test_mcp_server.py tests/test_mcp_stub.py -q
37 passed, 20 warnings in 6.82s

$ ruff check <changed files> && ruff format --check <changed files>
All checks passed!
6 files already formatted

$ PYTHONPATH=. mypy headroom/ccr/mcp_server.py headroom/memory/mcp_server.py
(no errors in either file)
```

The new contract test is a real guard, not a no-op — it fails against 1.28.1 and passes against 2.0.0:

```text
$ # mcp 1.28.1
$ python -m pytest tests/test_mcp_sdk2_contract.py -q
7 failed in 0.58s
```

## Real Behavior Proof

- **Environment:** macOS (Darwin arm64), Python 3.12, scratch venv with `mcp==2.0.0` + `httpx`; repo venv at `mcp==1.28.1` left untouched.
- **Exact command / steps:** constructed `HeadroomMCPServer(proxy_url=...)` and drove `_on_list_tools` / `_on_call_tool` directly against the real SDK, plus `create_initialization_options()` for the handshake shape.
- **Observed result:**
  - `ListToolsResult` with `['headroom_compress', 'headroom_retrieve', 'headroom_stats']`
  - `capabilities.tools: list_changed=False` — matches the handshake output in #2642's proof
  - `CallToolResult` returned for an unknown tool, and for `arguments` omitted entirely
  - `_current_client()` resolves to `claude-code` inside a request and `unknown` outside it
  - ContextVar is empty after a call returns (no cross-request leak)
- **Not tested:** an end-to-end `headroom mcp serve` stdio handshake driven over a real subprocess (the manual check done in #2642); a live MCP client session against Claude Code or OpenCode; Windows/Linux.

## Review Readiness

- [ ] This PR is ready for human review — **draft: blocked on `strands-agents`, and CI cannot pass until `uv.lock` can be regenerated**

## Checklist

## Additional Notes

**Why the stubbed suite could not have caught the 2.0.0 crash.** `tests/_mcp_stub.py` replaces the SDK in `sys.modules` with a `DummyServer` that hardcoded the 1.x decorator shape. Combined with `uv.lock` fixing `mcp` at 1.28.1 and `security.yml`'s `uv export --frozen` reading that lock, nothing in CI ever resolves a new upstream major. The 45 tests cited in #2642 were green throughout.

**The durable fix is not in this PR.** A scheduled job that resolves dependencies *unpinned* would have caught `mcp` 2.0.0 before users did, and would catch the next breaking major too. The contract test added here only helps once something actually resolves 2.x. Worth filing separately.

**Docs not updated** — no user-facing behaviour changes, but the `mcp` extra's version requirement should be revisited if option 2 or 3 above is chosen.

### DIFF
--- a/headroom/ccr/mcp_server.py
+++ b/headroom/ccr/mcp_server.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 import argparse
 import asyncio
 import contextlib
+import contextvars
 import json
 import logging
 import os
@@ -48,11 +49,12 @@ try:
 except ImportError:
     _HAS_FCNTL = False
 
-# Try to import MCP SDK
+# Try to import MCP SDK (2.x: handlers are constructor callbacks, results are
+# typed ``*Result`` models rather than bare lists).
 try:
     from mcp.server import Server
     from mcp.server.stdio import stdio_server
-    from mcp.types import TextContent, Tool
+    from mcp.types import CallToolResult, ListToolsResult, TextContent, Tool
 
     MCP_AVAILABLE = True
 except ImportError:
@@ -87,6 +89,14 @@ _READ_ENABLED = os.environ.get("HEADROOM_MCP_READ", "off").lower().strip() in (
 )
 
 DEFAULT_PROXY_URL = os.environ.get("HEADROOM_PROXY_URL", "http://127.0.0.1:8787")
+
+# MCP 2.x removed ``Server.request_context``; the per-request context is now only
+# handed to the handler as an argument. ``_current_client()`` needs it three frames
+# below the handler, so we stash it here instead of threading it through every
+# ``_handle_*`` signature. A ContextVar rather than an instance attribute because
+# concurrent tool calls on one server would race on ``self`` — this is the same
+# mechanism 1.x used internally to back ``request_context``.
+_request_ctx: contextvars.ContextVar[Any] = contextvars.ContextVar("headroom_mcp_request_ctx")
 
 # How often the parent-death watchdog polls os.getppid() (seconds). When the
 # launching MCP client is SIGKILLed, stdin EOF may never arrive and the SDK's
@@ -383,8 +393,14 @@ class HeadroomMCPServer:
         if not MCP_AVAILABLE or Server is None:
             raise ImportError("MCP SDK not installed. Install with: pip install mcp")
 
-        self.server: Server = Server("headroom")
-        self._setup_handlers()
+        # MCP 2.x registers handlers as constructor callbacks. Bound methods are
+        # resolved at call time, so passing them here does not require the rest of
+        # ``__init__`` to have run.
+        self.server: Server = Server(
+            "headroom",
+            on_list_tools=self._on_list_tools,
+            on_call_tool=self._on_call_tool,
+        )
 
     def _get_local_store(self) -> Any:
         """Get the shared compression store singleton (lazy init).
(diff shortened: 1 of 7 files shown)


JerrettDavis (changes requested):
`uv lock --check` is currently unsatisfiable on this branch. The PR raises both the `proxy` and `mcp` extras to `mcp>=2.0.0`, but the existing `strands` extra still depends on `strands-agents>=0.1.0`; every published `strands-agents` version uv considered requires `mcp<2.0.0`. Because the repo supports installing `headroom-ai[all]` alongside `headroom-ai[strands]`, the resolver now concludes those requirements are incompatible. This is the same root cause as the red Dev Containers checks: `.devcontainer/post-create.sh` runs `uv sync --frozen --extra dev` (and memory-stack adds one more extra), and uv aborts with `headroom-ai[all] and headroom-ai[strands] are incompatible`.

Please resolve the extra compatibility before this leaves draft: for example by excluding one of these extras from combinations that install together, gating `strands` until it supports MCP 2.x, or otherwise changing the declared extras so `uv lock --check` and the devcontainer `uv sync --frozen` path have a valid solution again.

Local checks:
- `uv lock --check` -> fails with the `strands-agents` / `mcp>=2.0.0` conflict above
- `.venv`Scripts`python.exe -m pytest tests/test_mcp_sdk2_contract.py tests/test_mcp_stub.py tests/test_memory/test_mcp_server.py -q` -> fails locally because the current venv still has the 1.x MCP SDK and the branch cannot be synced while the dependency graph is unsatisfiable

PR #2662, opened 2026-07-30
feat(proxy): default /v1/compress to marker-free output

chopratejas:
## Description

`POST /v1/compress` emits CCR retrieval markers by default. A marker is only useful to a caller that **also** injects the `headroom_retrieve` tool **and** can reach `/v1/retrieve`. Neither holds for the callers this route exists for:

- Tool injection lives in the provider request handlers (`handlers/anthropic.py:1894`, `should_inject_ccr_tool`). `handle_compress` never touches `tools` — it reads `messages` + `model` and returns messages.
- Every `/v1/retrieve*` route is `Depends(_require_loopback)` (`server.py:4422, 4470, 4749, 4781`) with no remote opt-in. `HEADROOM_COMPRESS_ALLOW_REMOTE` drops the loopback dependency on `/v1/compress` **only** (`server.py:4898`), so a remote gateway can compress but physically cannot rehydrate.
- LiteLLM's `headroom` guardrail — the main consumer of this route — swaps `messages` in `pre_call` and forwards (https://docs.litellm.ai/docs/proxy/headroom). It never injects a tool and never calls back.

So today a gateway forwards a `Retrieve more: hash=…` pointer the model cannot follow, and the proxy pays a CCR store write nobody will ever read.

This makes the route marker-free by default. `config.mode="ccr"` opts back in for callers that do run the retrieve loop.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `_no_ccr_pipeline()` — new default for `/v1/compress`: the live router config with `ccr_inject_marker=False, ccr_enabled=False`. Nothing else differs, so compressors and aggressiveness are unchanged.
- `config.mode` dispatch in `handle_compress`: unset → marker-free (default), `"ccr"` → the existing `openai_pipeline` with markers + store writes, `"lossy_inline"` / `"lossless_then_lossy"` → unchanged.
- `_derived_compress_pipeline(key, **overrides)` — one cached helper replacing the copy-pasted pipeline-derivation block; `_lossy_inline_pipeline` now uses it too. Net simplification despite the added mode.
- Docstring on `handle_compress` documents all three modes.
- Two tests updated to send `config: {"mode": "ccr"}`, which is now the explicit opt-in they were implicitly relying on.

### Behavior change

A caller that relied on default markers now gets none. The only in-tree caller that can resolve them is the TypeScript SDK (`sdk/typescript/src/client.ts:398 retrieve`, `:422 handleToolCall`); it needs `config: {"mode": "ccr"}` to keep today's behavior. Recommend landing that SDK default in the same release so SDK behavior is unchanged and only gateway callers — for whom markers were never resolvable — see the difference.

Deliberately not adding a `HEADROOM_COMPRESS_DEFAULT_MODE` compat env: a flag nobody sets becomes permanent debt, and the wire-level opt-in (`mode`) already covers the one caller that needs it.

## Testing

### Test Output

```text
$ python -m pytest tests/test_proxy_compress_endpoint.py tests/test_ccr_row_drop_store_bridge.py -q
24 passed in 17.13s

$ python -m pytest tests/test_gateway_sidecar_ports.py tests/test_compress_api.py -q
29 passed

$ ruff check headroom/proxy/handlers/openai.py
All checks passed!

$ ruff format --check headroom/proxy/handlers/openai.py
1 file already formatted

$ mypy headroom/proxy/handlers/openai.py
```

No new tests: the two updated tests now pin both sides of the switch — `test_v1_compress_then_v1_retrieve_resolves_marker_hash` proves `mode:"ccr"` still produces a marker the store resolves, and the rest of `test_proxy_compress_endpoint.py` exercises the new default path.

## Real Behavior Proof

- **Environment:** macOS arm64, Python 3.12.6, `.venv`, proxy built from `_proxy_config_from_env()` with the default `coding` profile, Kompress offloaded to a remote `kompress-v2-base` endpoint, `HEADROOM_COMPRESSION_TIMEOUT_SECONDS=300`.
- **Exact command / steps:** `POST /v1/compress` over `TestClient` with an OpenAI-shaped payload (system + RAG user blob + assistant `tool_calls` + a 200-row JSON tool result + a 300-line log tool result + assistant text), counting `Retrieve …: hash=` / `<<ccr:` occurrences in the returned messages. Compared: default route, `--no-ccr` at the config level, `config.mode="lossy_inline"`, `--lossless`.
- **Observed result:**

  | Config | tokens | saved |
  |---|---|---|
  | default (before this PR: markers on) | 37,791 → 24,415 | 35.4% |
  | markers off (this PR's default) | 37,791 → 24,415 | **35.4% — identical** |
  | `mode="lossy_inline"` | 37,791 → 25,129 | 33.5% |
  | `--lossless` | 37,791 → 35,100 | 7.1% |

  Turning markers off costs nothing on gateway-shaped traffic. `lossy_inline` is slightly worse than the new default because it also flips `lossless_then_lossy`, which reorders the search fold — worth knowing for anyone who reached for that mode as the "safe for a sidecar" option.
- **Not tested:** `tests/test_transforms_content_router.py` does not complete on this machine (wedges on its 5th test; passes in 5.7s alone). Verified pre-existing — the clean tree stalls identically at the same test, including under `HF_HUB_OFFLINE=1 HEADROOM_OFFLINE=1`. Unrelated to this diff, which does not touch the router. Suspected #575.

## Review Readiness

## Checklist

## Additional Notes

Docs unchecked — follow-ups worth doing in the same release:

- `docs/content/docs/litellm.mdx` and `wiki/proxy.md` should document the `config.mode` values.
- The TS SDK default (`mode:"ccr"`) described above.
- Operational note for anyone running this behind a gateway with remote ML inference: `COMPRESSION_TIMEOUT_SECONDS` defaults to 30 (`helpers.py:687`), and a remote Kompress endpoint makes one sequential call per unit. On a large payload it trips the executor timeout and the handler fails open — returning `compression_skipped: true` with `tokens_before: 0`, which reads as "nothing to save" rather than "we gave up". Separate PR, but the zeroed counters in that response are misleading and worth fixing.

Companion PR: #2661 (lossless provider on the general path). Independent — different file, no shared tests.

### DIFF
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -8252,21 +8252,20 @@ class OpenAIHandlerMixin:
             with contextlib.suppress(Exception):
                 await websocket.close()
 
-    def _lossy_inline_pipeline(self) -> Any:
-        """Cached pipeline for ``/v1/compress`` ``config.mode="lossy_inline"``.
+    def _derived_compress_pipeline(self, key: str, **overrides: Any) -> Any:
+        """Cached ``/v1/compress`` pipeline derived from the live OpenAI router.
 
-        Runs the lossless byte/data fold first, then Kompresses the folded
-        remainder (``lossless_then_lossy``). ``ccr_inject_marker=False`` makes
-        every compressor (Kompress, SmartCrusher, search/log/config) emit inline
-        lossy output with NO ``<<ccr:…>>`` / ``Retrieve more: hash=`` marker and
-        NO CCR store write, so the result is safe to forward straight to a
-        provider with no retrieval round-trip. Derived once from the live OpenAI
-        router's config and reused read-only across requests.
+        ``overrides`` are applied to the live ContentRouter's config, so the
+        derived pipeline inherits every operator setting (Kompress on/off,
+        exclusions, profile knobs) and differs only in what the caller needs.
 
         ponytail: a first-request race just builds it twice — both are
         equivalent and Kompress weights are cached at module level, so no lock.
         """
-        cached = getattr(self, "_lossy_inline_pipeline_cache", None)
+        cache = getattr(self, "_compress_pipeline_cache", None)
+        if cache is None:
+            cache = self._compress_pipeline_cache = {}
+        cached = cache.get(key)
         if cached is not None:
             return cached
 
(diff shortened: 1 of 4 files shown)


JerrettDavis (approved):
Reviewed the `/v1/compress` mode split. The default marker-free path is derived from the live router config, `config.mode="ccr"` preserves the previous marker/store path for retrieve-aware callers, and `lossy_inline` stays a separate lossless-then-lossy posture. That behavior matches the stated gateway use case.

I pushed `6f602344` to align `test_v1_compress_success_reports_actual_metrics` with the new `mode="ccr"` opt-in; before that, CI and the local focused run failed because the test patched `proxy.openai_pipeline` while the new default uses the derived no-CCR pipeline.

Local validation after the fix:
- `.\.venv\Scripts\python.exe -m pytest tests/test_platform_stabilization_functional.py::test_v1_compress_success_reports_actual_metrics -q` -> 1 passed
- `.\.venv\Scripts\python.exe -m pytest tests/test_proxy_compress_endpoint.py tests/test_ccr_row_drop_store_bridge.py -q` -> 24 passed
- `git diff --check` -> passed

GitHub CI is rerunning on the new head, and the PR body still needs the Real Behavior Proof fields the governance bot called out.

chopratejas:
Superseded by #2691, which contains this change plus the review fixes for it, rebased onto latest main.

Reviewing them together rather than landing this and patching after: the fixes are inseparable from the code they fix, and this repo squash-merges, so a follow-up branch carrying cherry-picked copies would have conflicted with the squashed commits.

PR #2668, opened 2026-07-30
fix: protect file reads in chained shell commands

gglucass:
## Description

`_is_read_command` decides whether a shell command's output is file content the agent will patch from. Output classified as a read is passed verbatim; everything else is eligible for lossy Kompress compression.

It inspected only the **first program**, after peeling `cd`/`sudo`/`rtk` wrappers. Agents routinely batch a read behind other work, so every chained read fell through to the compression path despite `HEADROOM_PROTECT_READS=1`:

```
wc -l a.py b.py && sed -n '1,60p' c.py     # parsed as `wc`   -> not a read
grep -n X f.py | head -40; head -80 g.py   # parsed as `grep` -> not a read
ls -la || cat f.py                         # parsed as `ls`   -> not a read
```

Kompress operates on words and rejoins the survivors, so file content arrives with function words missing. That is precisely the failure read protection exists to prevent: the agent either re-reads to recover detail (turn inflation) or builds an exact-match `old_string` from text that no longer matches the file on disk.

Reported by a Headroom desktop user who saw source content garbled while the `Read` tool path stayed byte-exact — which is what pointed at the shell-command classifier rather than the tool exclusion list.

No linked issue; reported directly.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `_is_read_command` now scans every `;` / `&&` / `||` segment and protects if **any** segment is a read, instead of matching only the first program.
- Extracted the existing single-command logic into `_segment_is_read`, unchanged in behavior for a non-chained command.
- Added `_command_segments`, which splits on `;` / `&&` / `||` and reduces each segment to its first pipeline stage.
- Pipelines are deliberately **not** split: a downstream stage consumes the previous stage's output rather than a file, so `grep … | head -40` stays compressible while `cat f.py | head -40` is still caught via its first stage.
- The whole-string write/heredoc guard is unchanged, so a heredoc body containing `;` cannot be split into a bogus read segment.
- Two latent false positives close as a consequence of per-segment evaluation: a `-n` flag in a sibling segment no longer qualifies a bare `sed`, and the lockfile carve-out now applies to the segment naming the lockfile rather than the whole command.
- New test file `tests/test_transforms/test_read_command_detection.py`.

## Testing

- [ ] Linting passes (`ruff check .`) — see Additional Notes

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_transforms/ \
    tests/test_observed_wire_shapes.py tests/test_content_router_exclude_tools.py -q
======================= 408 passed, 62 skipped in 13.46s =======================

$ uv run --frozen --extra dev pytest tests/test_transforms/test_read_command_detection.py -q \
    --cov=headroom.transforms.content_router --cov-report=term-missing --cov-branch
30 passed in 0.45s
# no uncovered lines or partial branches remain in the changed range (614-654)

$ uvx ruff check headroom tests --output-format concise
All checks passed!

$ uvx ruff format --check headroom/transforms/content_router.py \
    tests/test_transforms/test_read_command_detection.py
2 files already formatted

$ uv run --frozen --extra dev mypy headroom
Success: no issues found in 512 source files
```

## Real Behavior Proof

- Environment: macOS (Darwin 24.6.0), Python 3.10.18, worktree branched from `upstream/main` @ `28aa53dc`; plus a live Headroom desktop proxy running `headroom-ai` 0.32.1 with Claude Code pointed at `127.0.0.1:6767`.
- Exact command / steps: (1) pre-fix on the live proxy, ran `grep -n "Read" transforms/compression_policy.py | head -40; echo "=== POLICY HEAD ==="; head -80 transforms/compression_policy.py` through Claude Code's Bash tool; (2) ran `ContentRouter().apply(...)` with `HEADROOM_PROTECT_READS=1` over a `role: tool` bash observation produced by `wc -l app/services/inventory.py && sed -n '1,200p' app/services/inventory.py` (11,007 chars of plain Python source, aged behind 20 follow-up turns so the positional "recent code" guard no longer covers it), against both the pre-fix and post-fix `content_router.py`.
- Observed result: step 1 returned the file's own docstring with function words dropped and a retrieval marker appended; step 2 went from no protection transform at all (pre-fix — the observation enters the compression path) to `router:read_protected` (post-fix). Verbatim output below.
- Not tested: byte-level Kompress corruption inside the repo venv (the ML model is not downloaded there — `Kompress model not ready; requests will not be compressed` — so step 2 shows the routing change and step 1 shows the corruption); `magika` unavailable in that venv (`no pip onnxruntime native library was found`) so detection fell through to the regex tier during step 2, which affects which compressor a released read routes to but not whether a read is protected; Windows shell command strings; fleet-scale traffic replay, so the savings impact of newly-protected reads is unquantified.

### Proof detail

Step 1 — marker appended to the returned file content:

```
[478 items compressed to 305. Retrieve more: hash=5d959a193f8641579c741ad5]
```

On disk:

> Hand-mirrored port **of** `headroom_core::compression_policy::CompressionPolicy` (Rust). **The** Rust crate **is the** source **of** truth; **this** module exists **so the** Python proxy's `TransformPipeline` (which still runs `CacheAligner` **and other** detector-only transforms) **can** read **the** same per-mode flags **the** Rust dispatcher does.

As received by the model:

> Hand-mirrored port `headroom_core::compression_policy::CompressionPolicy` (Rust). Rust crate source truth; module exists Python proxy's `TransformPipeline` (which still runs `CacheAligner` detector-only transforms) read same per-mode flags Rust dispatcher does.

Step 2 — routing branch taken by the tool observation:

```text
##### BEFORE (upstream/main content_router.py) #####
command:               wc -l app/services/inventory.py && sed -n '1,200p' app/services/inventory.py
input chars:           11007
file content verbatim: True
transforms:            ['router:protected:user_message', ... x10]      <- no read protection

##### AFTER (this PR) #####
command:               wc -l app/services/inventory.py && sed -n '1,200p' app/services/inventory.py
input chars:           11007
file content verbatim: True
transforms:            ['router:read_protected', 'router:protected:user_message', ... x10]
```

Before the fix the observation carries no protection transform at all and enters the compression path. After the fix it is `router:read_protected`.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A, no user-facing docs describe this internal classifier; its docstrings are updated

N/A — no UI surface.

## Additional Notes

- `ruff check .` reports 4 errors, all pre-existing on `upstream/main` and all in `plugins/headroom-oauth2/` (`RUF022`, `BLE001` x2, `S110`). Untouched by this PR, so the box is left unchecked rather than claimed. `ruff check headroom tests` is clean.
- Trade-off: the write/heredoc guard stays whole-string rather than per-segment. That keeps `cat a.py && echo x > b` unprotected, which is the pre-existing behavior — chosen so a heredoc body containing `;` can never be split into a segment that looks like a read. Narrowing it further is a separate change.
- Follow-up for the reporter's second issue (not in scope here): passive traffic learning has no off switch, and the desktop starts the proxy with `--learn` unconditionally.

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -586,14 +586,47 @@ def _is_read_command(command: str) -> bool:
     Excludes writes: a redirect (``>``/``>>``), ``tee``, or heredoc (``<<``) means the
     command WRITES a file (e.g. ``cat > f <<EOF``), and a bare ``sed`` (without ``-n``)
     is a stream edit — neither is a read.
+
+    CHAINED commands are scanned segment by segment: agents routinely batch a read
+    with other work (``wc -l a.py && sed -n '1,60p' a.py``, ``grep … ; head -80 b.py``),
+    and matching only the FIRST program left every such read unprotected — the file
+    content was lossy-compressed despite read protection being enabled.
     """
     if not command or not isinstance(command, str):
         return False
     # strip leading `cd <dir> && ` chains (agents prefix reads with a cd)
     c = _strip_cd_prefix(command)
-    # a write / append / tee / heredoc anywhere => not a pure read
+    # a write / append / tee / heredoc anywhere => not a pure read. Kept
+    # whole-string (not per segment) so a heredoc body containing `;` can never
+    # be split into a bogus segment that looks like a read.
     if re.search(r"(^|\s)(>>?|tee\b|<<)", c):
         return False
+    return any(_segment_is_read(seg) for seg in _command_segments(c))
+
+
+# Separators that start a NEW command. A pipeline (`|`) deliberately does NOT:
+# downstream stages consume the previous stage's output, not a file, so
+# `grep -n x a.py | head -40` is derived search output (compressible), while
+# `cat a.py | head -40` is a file read (caught via the first stage).
+_CMD_SEPARATOR_RE = re.compile(r"\|\||&&|;")
+
+
+def _command_segments(command: str) -> list[str]:
+    """Split a shell string into the independently-executed commands in it.
+
+    Splits on ``;``, ``&&`` and ``||`` — never on a single ``|``. Each segment is
+    reduced to its FIRST pipeline stage, which is the one that touches a file.
+    """
+    segments = []
+    for seg in _CMD_SEPARATOR_RE.split(command):
+        stage = seg.split("|", 1)[0].strip()
+        if stage:
+            segments.append(stage)
+    return segments
+
+
+def _segment_is_read(c: str) -> bool:
+    """:func:`_is_read_command` for ONE command (no ``;``/``&&``/``||`` chaining)."""
     # Parse the real program with the SAME structural parser the search-fold uses
     # (_bash_program peels sudo/env/timeout/rtk wrappers + env assignments), so
     # `sudo cat f`, `timeout 30 cat f`, `rtk cat f` are recognized as reads, not
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -585,7 +585,9 @@ def _is_read_command(command: str) -> bool:
 
     Excludes writes: a redirect (``>``/``>>``), ``tee``, or heredoc (``<<``) means the
     command WRITES a file (e.g. ``cat > f <<EOF``), and a bare ``sed`` (without ``-n``)
-    is a stream edit — neither is a read.
+    is a stream edit — neither is a read. Redirects and ``tee`` are judged PER SEGMENT,
+    so a sibling write (``cat a.py && echo done > marker``) does not strip protection
+    from the read segment next to it.
 
     CHAINED commands are scanned segment by segment: agents routinely batch a read
     with other work (``wc -l a.py && sed -n '1,60p' a.py``, ``grep … ; head -80 b.py``),
@@ -596,10 +598,11 @@ def _is_read_command(command: str) -> bool:
         return False
     # strip leading `cd <dir> && ` chains (agents prefix reads with a cd)
     c = _strip_cd_prefix(command)
-    # a write / append / tee / heredoc anywhere => not a pure read. Kept
-    # whole-string (not per segment) so a heredoc body containing `;` can never
-    # be split into a bogus segment that looks like a read.
-    if re.search(r"(^|\s)(>>?|tee\b|<<)", c):
+    # A heredoc is the one WHOLE-STRING bailout: its body can contain `;`/`&&`,
+    # which would split into bogus segments that look like reads. Redirects and
+    # `tee` are per-segment (see `_segment_is_read`) — a sibling write must not
+    # unprotect the read next to it.
+    if re.search(r"(^|\s)<<", c):
         return False
     return any(_segment_is_read(seg) for seg in _command_segments(c))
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
Thanks for tightening this path; the segment-by-segment scan fixes the common `wc && sed` / `ls || cat` cases and the focused tests are good. I found one remaining correctness gap that I think should be fixed before merge.

`_is_read_command()` still applies the write/heredoc guard to the whole command before segment evaluation:

```py
if re.search(r"(^|\s)(>>?|tee\b|<<)", c):
    return False
```

That means a read segment is left unprotected whenever a sibling segment writes, even though the command output can still include exact file content the agent may patch from:

```text
False  cat foo.py && echo done > marker
False  echo done > marker && cat foo.py
```

Those cases still route the `cat foo.py` bytes into the compressible path under `HEADROOM_PROTECT_READS=1`, so the PR does not fully close the chained-read failure mode. The existing heredoc concern is valid, but we can preserve it more narrowly: keep `<<` as a whole-command bailout, then evaluate redirects/`tee` per segment before deciding whether that segment is a read. That would also keep `cat foo.py | tee bar.py` unprotected while allowing `cat foo.py && echo done > marker` to protect the read segment.

Local verification on the branch:
- `.venv\Scripts\python.exe -m pytest tests/test_transforms/test_read_command_detection.py` -> 30 passed
- Manual classifier smoke shows the sibling-write false negatives above.

JerrettDavis (approved):
Re-reviewed latest head `d75d0ec`. The previous blocker is fixed: redirects and `tee` are now evaluated per command segment while heredoc remains a whole-command bailout, so sibling writes no longer strip protection from exact file-read output.

Local validation:
- `.venv\Scripts\python.exe -m pytest tests/test_transforms/test_read_command_detection.py -q` -> 35 passed
- `.venv\Scripts\python.exe -m ruff check headroom/transforms/content_router.py tests/test_transforms/test_read_command_detection.py` -> passed
- Manual classifier smoke now returns `True` for `cat foo.py && echo done > marker`, `echo done > marker && cat foo.py`, and `sed -n '1,60p' foo.py; touch out >> log`; and still returns `False` for `cat foo.py | tee bar.py`, `cat foo.py > copy.py && echo done`, and `grep -n x foo.py | head -40`.

CI is green and the branch is conflict-free. Looks ready for final human review.

gglucass:
@chopratejas @JerrettDavis @DevanshiVyas — merge ping. Approved by @JerrettDavis on 2026-07-30 after the requested change was pushed.

Re-verified against `main` @ 7c9b0465:

- **Still needed:** `_command_segments` is absent from `main`, so read protection is still matched against the first program only. Anything an agent batches — `wc -l a.py && sed -n '1,60p' a.py`, `grep ... ; head -80 b.py` — leaves the actual read unprotected and the file content gets lossy-compressed even with read protection enabled.
- **Still clean:** `MERGEABLE` / `CLEAN`, all checks green.

+140/-3, confined to `headroom/transforms/content_router.py` plus tests.

gglucass:
Merge ping @chopratejas @DevanshiVyas — approved by @JerrettDavis on 2026-07-30, previously pinged 2026-08-03. Re-verified against `main` @ 941c25d3 today: `headroom/transforms/content_router.py` still has no chained-command read protection, so file reads behind `&&`/`;`/`|` remain unprotected. Merge state is CLEAN, no conflicts.

gglucass:
Still open, still unfixed on current `main` (`e67b3c8a`), and re-verified end to end today.

`main`'s `_is_read_command` has been improved since this branch forked (it now peels wrappers via `_bash_program`), but both root causes this PR addresses are untouched: the redirect/`tee` bailout is still whole-string (`content_router.py:654`), and only the FIRST program is parsed. Probing the merged result of `main` + this PR against `main` alone:

| command | `main` | with this PR |
|---|---|---|
| `cat a.py` | True | True |
| `wc -l a.py && sed -n '1,60p' a.py` | **False** | True |
| `grep -n x a.py ; head -80 b.py` | **False** | True |
| `cat a.py && echo done > marker` | **False** | True |
| `cd /tmp && cat a.py` | True | True |
| `grep -n x a.py \| head -40` | False | False |

The three `False` rows on `main` are reads that silently lose read protection, so the file content gets lossy-compressed with protection nominally enabled. The last row is the deliberate non-regression: a pipeline is derived search output, not a file read, and stays compressible.

Worth noting how routine those shapes are. Agent harnesses actively teach the chained form for token efficiency (locate with `grep -n`, then read the range with `sed -n 'A,Bp'` in the same call), so the unprotected path is closer to the common case than the exception.

Verification on `main` + this PR merged, not on the stale branch head:

- Merge: `git merge --no-commit --no-ff` into `e67b3c8a` auto-merges, no conflicts, despite 347 commits of drift.
- Tests: `pytest tests/test_transforms/test_read_command_detection.py` -> 35 passed.

Happy to rebase and squash the history if that helps review.

PR #2670, opened 2026-07-30
feat(cache): add the headroom-cache-ttl offline TTL estimator the docs reference

gglucass:
## Description

Adds the missing offline estimator half of the cache-TTL learning seam: a `headroom-cache-ttl` console script that reads `cache_ttl_observations.jsonl` (written by the proxy under `HEADROOM_CACHE_TTL_LEARN=1`) and writes the `cache_ttl_learned.json` table that `resolve_learned_ttl()` already consults.

The docs and `headroom/cache/ttl_observations.py`'s docstring both reference this estimator as the thing that turns observations into learned TTLs — but the tool does not ship:

```
$ git grep -l "headroom-cache-ttl" v0.33.0
docs/content/docs/configuration.mdx
headroom/cache/ttl_observations.py    # docstring reference only
$ headroom-cache-ttl
command not found
```

So today `HEADROOM_CACHE_TTL_LEARN=1` collects observations that nothing ever reads, and `HEADROOM_COLD_RECOMPACT` cold detection for providers that don't expose a TTL (OpenAI/Codex, Kimi) stays on the static 300s guess forever — exactly the aggressive-in-the-wrong-direction value the docs warn about (a resume in the 6–60 min band can recompact a still-warm OpenAI cache).

Filed as an implementation of the tool the docs already describe rather than a new design; happy to move/rename if you'd prefer a `headroom` subcommand over a second console script.

Closes #

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- New `headroom/cache/ttl_estimator.py` with the `headroom-cache-ttl` CLI (`[project.scripts]` entry in `pyproject.toml`, named exactly as the docs reference it). Stdlib only, no new dependencies.
- **Interval estimation, per the contract documented in `ttl_observations.py`:** hit idles bound the TTL from below (cache proven alive), `ttl_expiry` miss idles bound it from above (cache proven dead). The estimator emits the interval's **upper** end — smallest observed death-idle beyond the largest observed life-idle — because the consumer treats a turn as cold only when `idle > ttl`: overestimating merely skips a recompaction, underestimating would bust a warm cache.
- Keys with no death-evidence beyond their last observed life are **skipped** (no safe upper bound → static default stays). Sample floors: ≥3 hits and ≥3 `ttl_expiry` misses per key (`--min-hits` / `--min-expiry-misses`).
- Emits `provider/model` keys **only** — no bare-`provider` aggregate. A `ttl_expiry` miss for model B says nothing about model A's TTL, so a pooled per-provider interval could hand the consumer a TTL below a model's real one (the costly direction). Unseen models fall back to the consumer's static default; extra metadata fields (`max_hit_idle`, sample counts, `updated_at`) ride along and are ignored by the reader.
- Atomic tmp+rename write, **merged** with the existing table so observation-window rotation doesn't un-learn old keys. On every write, bare-provider aggregates left behind by earlier revisions of this tool (recognized by their generated metadata shape, e.g. `max_hit_idle`) are purged; operator-written provider fallbacks (`{"ttl_seconds": N}` or a plain number, which `resolve_learned_ttl()` still supports) are preserved.
- Offline/out-of-process by design (zero live-feedback risk on the hot path); the proxy picks up a rewritten table by mtime, no restart needed.
- Docs: "Running the estimator" section added under the cold-prefix hook in `configuration.mdx`.

## Testing

New `tests/test_cache/test_ttl_estimator.py` — 23 tests: upper-bound selection, overlapping bounds (variable eviction) skipping, sample floors, non-`ttl_expiry` misses ignored, malformed/non-finite rows and lines skipped, one model's death evidence cannot close another's interval (no provider pooling), merge-preserves-existing-keys, legacy-aggregate purge with manual-fallback preservation (upgrade regression, both directions), corrupt-existing-file recovery, and CLI end-to-end through `resolve_learned_ttl()` (env paths, explicit flags, `--dry-run`).

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_cache/
======================= 277 passed, 2 skipped in 16.54s ========================

$ uvx ruff check headroom/cache/ttl_estimator.py tests/test_cache/test_ttl_estimator.py
All checks passed!
$ uvx ruff format --check headroom/cache/ttl_estimator.py tests/test_cache/test_ttl_estimator.py
2 files already formatted
$ uv run --frozen --extra dev mypy headroom/cache/ttl_estimator.py
```

## Real Behavior Proof

- Environment: macOS 15 (Darwin 24.6.0), Python 3.12 via `uv run --frozen`, branch `feat/cache-ttl-estimator` on top of `main`.
- Exact command / steps: wrote a synthetic observation log in the exact row shape `record_cache_observation` emits (3 hits at idle 120/300/480s, 3 `ttl_expiry` misses at 600/900/1200s for `openai/gpt-5.5`), plus a pre-existing `learned.json` containing a legacy `openai` estimator aggregate and a hand-written `kimi` fallback, then (at ec6e1eef):

  ```
  $ uv run --frozen headroom-cache-ttl --obs /tmp/obs.jsonl --out /tmp/learned.json
  headroom-cache-ttl: wrote 1 estimate(s) (2 total) to /tmp/learned.json
  ```

- Observed result: `ttl_seconds: 600` under `openai/gpt-5.5` only — the smallest observed death-idle (600s) beyond the largest observed life-idle (480s); the legacy `openai` aggregate was purged and the manual `kimi` fallback survived:

  ```json
  {
   "kimi": {"ttl_seconds": 3600},
   "openai/gpt-5.5": {"ttl_seconds": 600, "max_hit_idle": 480, "hits": 3, "ttl_expiry_misses": 3, "updated_at": "..."}
  }
  ```

  and `resolve_learned_ttl("openai", "gpt-5.5") == 600`, `resolve_learned_ttl("openai", "unseen-model") is None` (static default, no provider pooling), `resolve_learned_ttl("kimi", "anything") == 3600` (manual fallback), verified in the end-to-end tests.

- Not tested: a multi-day live observation window from real Codex traffic (my desktop deployment enables `HEADROOM_CACHE_TTL_LEARN` for Codex users; I can report a real-traffic estimate on this PR once a window accumulates).

## Review Readiness

## Checklist

N/A (CLI-only change; output shown above).

## Additional Notes

- No linked issue: this fills a gap the merged docs (#2557) already describe, so there was no separate design to discuss. Happy to open one retroactively if you want the discussion tracked.
- `write_learned` merge semantics mean a stale key learned from an old window persists until re-estimated; acceptable since estimates only ever bound from the safe side, and `--dry-run` lets an operator inspect before writing.

### DIFF
--- a/docs/content/docs/configuration.mdx
+++ b/docs/content/docs/configuration.mdx
@@ -350,6 +350,24 @@ and the **prefix** itself once its prompt cache has lapsed. All flags below are
   conservative default until the learner (`HEADROOM_CACHE_TTL_LEARN` + the
   `headroom-cache-ttl` plugin) fills in the real value from observed hits/misses.
 
+**Running the estimator.** With `HEADROOM_CACHE_TTL_LEARN=1` set on the proxy,
+run the batch estimator periodically (cron / launchd — it is deliberately
+out-of-process so it can never touch the request path):
+
+```bash
+headroom-cache-ttl            # reads cache_ttl_observations.jsonl, writes cache_ttl_learned.json
+headroom-cache-ttl --dry-run  # print the estimate table without writing
+```
+
+It emits a TTL for a `provider` (and `provider/model`) only when the
+observations bound it on both sides — hits at some idle prove the cache was
+still alive (lower bound), `ttl_expiry` misses at a larger idle prove it was
+dead (upper bound) — and it writes the upper end of that interval: overestimating
+a TTL only skips a recompaction, while underestimating one would bust a warm
+cache. Keys without death-evidence beyond their last observed hit are skipped
+and keep the conservative static default. The proxy picks up a rewritten table
+by mtime; no restart needed.
+
 **Can these be on by default?**
 
 - `HEADROOM_THINKING_COMPACT` — **stays opt-in.** It rewrites model inputs
--- a/docs/content/docs/configuration.mdx
+++ b/docs/content/docs/configuration.mdx
@@ -359,14 +359,17 @@ headroom-cache-ttl            # reads cache_ttl_observations.jsonl, writes cache
 headroom-cache-ttl --dry-run  # print the estimate table without writing
 ```
 
-It emits a TTL for a `provider` (and `provider/model`) only when the
-observations bound it on both sides — hits at some idle prove the cache was
-still alive (lower bound), `ttl_expiry` misses at a larger idle prove it was
-dead (upper bound) — and it writes the upper end of that interval: overestimating
-a TTL only skips a recompaction, while underestimating one would bust a warm
-cache. Keys without death-evidence beyond their last observed hit are skipped
-and keep the conservative static default. The proxy picks up a rewritten table
-by mtime; no restart needed.
+It emits a TTL for a `provider/model` only when the observations bound it on
+both sides — hits at some idle prove the cache was still alive (lower bound),
+`ttl_expiry` misses at a larger idle prove it was dead (upper bound) — and it
+writes the upper end of that interval: overestimating a TTL only skips a
+recompaction, while underestimating one would bust a warm cache. Estimates are
+never pooled across models: one model's expiry says nothing about another's
+TTL, so a cross-model aggregate could close one model's interval with another's
+death evidence. Keys without death-evidence beyond their last observed hit are
+skipped, as is any model with no estimate of its own; both keep the
+conservative static default. The proxy picks up a rewritten table by mtime; no
+restart needed.
 
 **Can these be on by default?**
 
--- a/headroom/cache/ttl_estimator.py
+++ b/headroom/cache/ttl_estimator.py
@@ -143,7 +143,12 @@ def write_learned(table: dict[str, dict[str, Any]], path: str) -> dict[str, Any]
     """Merge ``table`` into the learned file at ``path`` (atomic tmp+rename).
 
     Merging (not replacing) keeps keys learned from an earlier, since-rotated
-    observation window instead of silently un-learning them.
+    observation window instead of silently un-learning them — but only
+    ``provider/model`` keys. Bare-provider aggregates written by earlier
+    versions of this tool are exactly the unsound cross-model intervals the
+    module docstring rules out; left in place they would keep feeding
+    ``resolve_learned_ttl``'s provider fallback forever, so they are dropped
+    on every write.
     """
     merged: dict[str, Any] = {}
     try:
@@ -154,6 +159,7 @@ def write_learned(table: dict[str, dict[str, Any]], path: str) -> dict[str, Any]
     except (OSError, json.JSONDecodeError):
         pass
     merged.update(table)
+    merged = {k: v for k, v in merged.items() if "/" in k}
     os.makedirs(os.path.dirname(path) or ".", exist_ok=True)
     tmp = f"{path}.tmp.{os.getpid()}"
     with open(tmp, "w", encoding="utf-8") as f:
@@ -207,7 +213,10 @@ def main(argv: list[str] | None = None) -> int:
     if args.dry_run:
         print(json.dumps(table, indent=2, sort_keys=True))
         return 0
-    if not table:
+    # With nothing to add AND no existing file there is nothing to do; if the
+    # file exists we write even an empty table so the legacy-aggregate purge in
+    # write_learned() runs regardless of whether today's window was estimable.
+    if not table and not os.path.exists(out_path):
         print(
             f"headroom-cache-ttl: no estimable keys in {obs_path} "
             f"({len(rows)} rows); nothing written"
--- a/headroom/cache/ttl_estimator.py
+++ b/headroom/cache/ttl_estimator.py
@@ -1,0 +1,194 @@
+"""Offline cache-TTL estimator — the `headroom-cache-ttl` CLI.
+
+The other half of the seam in :mod:`headroom.cache.ttl_observations`: the proxy
+appends per-turn cache outcomes to ``cache_ttl_observations.jsonl`` (when
+``HEADROOM_CACHE_TTL_LEARN`` is set); this batch tool reads those rows and
+writes the ``cache_ttl_learned.json`` table that
+:func:`~headroom.cache.ttl_observations.resolve_learned_ttl` consults, so cold
+detection (``HEADROOM_COLD_RECOMPACT``) stops guessing for providers that don't
+expose their TTL (Kimi / OpenAI / Codex).
+
+Estimation is interval-based, per the contract in ``ttl_observations``: idle
+gaps that still HIT bound the TTL from below (the cache was alive at that
+idle), and ``ttl_expiry`` misses bound it from above (the cache was dead by
+that idle). We emit the UPPER end of the interval — the smallest observed
+death-idle beyond the largest observed life-idle — because the consumer treats
+a turn as cold only when ``idle > ttl``: overestimating the TTL merely skips a
+recompaction (lost savings), while underestimating it would recompact a
+still-warm prefix and bust the cache (net cost). A key with no death evidence
+beyond its last observed life is skipped entirely for the same reason.
+
+Runs out-of-process on purpose (zero live-feedback risk on the hot path); the
+proxy re-reads the learned table by mtime, so a periodic cron/launchd run is
+enough. Never imports anything request-scoped.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import time
+from collections import defaultdict
+from typing import Any
+
+from headroom.cache.ttl_observations import _learned_path, _obs_path
+
+DEFAULT_MIN_HITS = 3
+DEFAULT_MIN_EXPIRY_MISSES = 3
+
+
+def _iter_rows(path: str) -> list[dict[str, Any]]:
+    """Parse the observation JSONL, skipping malformed lines (best-effort)."""
+    rows: list[dict[str, Any]] = []
+    try:
+        with open(path, encoding="utf-8") as f:
+            for line in f:
+                line = line.strip()
+                if not line:
+                    continue
+                try:
+                    row = json.loads(line)
+                except json.JSONDecodeError:
+                    continue
+                if isinstance(row, dict):
+                    rows.append(row)
+    except OSError:
+        pass
+    return rows
+
+
+def estimate_ttls(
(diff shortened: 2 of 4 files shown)


JerrettDavis (approved):
Reviewed the latest head (`eb97b7d`) against the existing TTL observation seam and cold-prefix lookup path.

What I checked:
- The estimator uses the same `provider` / `provider/model` lookup shape that `resolve_learned_ttl()` reads.
- Hit rows and `ttl_expiry` miss rows are separated correctly; non-TTL misses do not influence the learned upper bound.
- Missing or malformed observation files/rows are best-effort and do not put the request path at risk.
- The learned table write is atomic and preserves existing keys, which is the right behavior with rotated observation windows.
- The docs and console-script entry point line up with the implementation.

Verification:
- GitHub CI is green and the PR is clean against `main`.
- Local focused test run passed: `.venv\Scripts\python.exe -m pytest tests/test_cache/test_ttl_estimator.py` -> 15 passed.

One local note: `uv run --frozen ...` could not rebuild the editable wheel on my Windows checkout because an existing Python process was holding `headroom/_core.pyd` open. I used the existing venv for the focused pure-Python test instead; CI already covers the package build path successfully.

JerrettDavis (changes requested):
The provider aggregate is not a valid TTL interval when models have different cache policies. Life and death evidence must describe the same cache population: a `ttl_expiry` miss for model B does not upper-bound model A's TTL. For example, model A can have hits at 900/950/1000s and a real 3600s TTL, while model B supplies expiry misses at 1200/1300/1400s. The aggregate emits `provider.ttl_seconds=1200`; `resolve_learned_ttl(provider, an-unseen-model)` then treats 1200 as authoritative and can recompact a cache that is still warm. That contradicts the PR's load-bearing claim that estimates can only overestimate and skip savings, never underestimate and bust a warm prefix.

Please keep estimates scoped to `provider/model`, or derive a provider fallback only from evidence that is conservatively valid across every contributing model (and document/prove that invariant). Do not let one model's death evidence close another model's life interval. Add a mixed-model regression with materially different TTLs and exercise the consumer fallback path.

Separately, `float("nan")` and infinities pass the `idle <= 0` filter; they can poison `max()`/comparisons and `int(...)` then raises `ValueError` or `OverflowError`, aborting the whole offline run on one malformed JSONL row. Please require `math.isfinite(idle)` and add NaN/±Infinity rows to the malformed-input coverage. Validating positive CLI sample floors would also prevent accidental zero/negative thresholds from disabling the evidence requirement.

gglucass:
Thanks @JerrettDavis — both findings are right, and the first one is worse than the writeup suggests: the previous test suite had `test_provider_aggregate_spans_models` asserting the defect as intended behaviour ("the mini model has no expiry evidence of its own -> no per-model key, but its hits still feed the provider aggregate"). That is exactly your model A / model B case, blessed by a passing test. Fixed in f91c832b.

**Provider aggregate — removed rather than repaired.**

I took your first option. On the second one (a fallback "conservatively valid across every contributing model"): I don't think that invariant exists to be proven. Safety here means erring *high*, and a provider fallback is consumed for models we have never observed, so it would have to upper-bound a TTL we have no evidence about. Any pooled value can be busted by a model whose real TTL is longer than everything in the pool. So the estimator now emits `provider/model` keys only, and rows without a model are dropped instead of being pooled.

Consequence, deliberately: an unseen model now gets `None` from `resolve_learned_ttl()` and `is_cold_prefix` keeps its static default. That is the same "no safe upper end -> skip the key" rule the PR already applied to keys without death evidence, just extended to models without evidence of their own. The `provider` step in `resolve_learned_ttl()`'s lookup order is untouched and still works for a hand-written table — this tool simply never writes those keys.

Regression added, per your request: `test_one_models_death_cannot_close_anothers_life_interval` builds your scenario (gpt-5.5 alive to 1000s and never observed dead, mini dying at 1200/1300/1400s), asserts only mini is estimable, and the end-to-end test asserts the consumer falls through for an unseen model.

**Non-finite idles.**

Confirmed and fixed with `math.isfinite()`. Worth recording that this was reachable from our own recorder, not just hand-edited files: `json.dumps()` emits bare `NaN`/`Infinity` and `json.loads()` reads them straight back, so `record_cache_observation` will happily write one. The existing coverage was also weaker than it looked — `{"idle_seconds": "nan-ish"}` is a *string*, caught by the `except ValueError`, so it never exercised the float path at all.

The failure mode is order-dependent, which is why the new tests pin the ordering: `max([nan, 5])` is `nan` but `max([5, nan])` is `5`, so a leading NaN propagates to `max_hit_idle` and takes the run down in `int()`, while a trailing one silently skews the estimate. There are now NaN/±Inf rows at unit level and a JSONL-level test with the NaN first.

**Sample floors.**

Also right, and it was a crash rather than just a disabled check: `--min-hits 0` makes `len(key_hits) < 0` unsatisfiable, so a key with no hits at all reached `max([])`. Clamped to >= 1 inside `estimate_ttls()` so the invariant holds for library callers too (an interval needs one life and one death sample to exist), plus an argparse type that rejects sub-1 values at the CLI with a real error message instead of silently clamping.

All six new/updated tests fail against the previous estimator; 19 pass on the new one, `ruff check` + `ruff format --check` + `mypy` clean, docs updated to drop the aggregate claim.

gglucass:
CI on `f91c832b`: 48 pass, 14 skipped. The single red check is the `Vercel` preview deploy — "Authorization required to deploy", which is the fork-PR authorization prompt on the Headroom Labs Vercel team, not a test failure. It fires because this PR touches `docs/content/docs/configuration.mdx`; it needs a maintainer click and there is nothing I can do about it from the fork side.

JerrettDavis (changes requested):
The cross-model aggregation and non-finite input issues are fixed for newly generated estimates, but existing unsafe provider aggregates survive upgrades. `write_learned()` deliberately starts with every key from the prior learned file and then overlays the new model-only table. Any file produced by the earlier branch/version can therefore retain an `"openai"` (or other bare-provider) aggregate indefinitely; `resolve_learned_ttl(provider, unseen_model)` will keep consuming it, preserving the exact warm-cache underestimation this revision is meant to eliminate. Please remove/migrate legacy bare-provider records when writing model-scoped output (while preserving valid `provider/model` keys), and add an upgrade regression that seeds an old provider aggregate, runs `main()`, and proves the aggregate is gone and the consumer fallback for an unseen model returns `None`. The new model scoping, finite checks, and positive sample floors otherwise address the prior findings cleanly.

gglucass:
Addressed in 926f15ef @JerrettDavis : `write_learned()` now drops every non-`provider/model` key on write, so legacy bare-provider aggregates from the earlier revision are purged instead of merged forward. `main()` also runs the write (and thus the purge) even when the current observation window yields no estimates, as long as a learned file already exists — otherwise an upgrade with a quiet window would leave the aggregate in place.

Regressions added:
- `test_upgrade_purges_legacy_provider_aggregate`: seeds an old `"openai"` aggregate, runs `main()`, asserts the aggregate is gone, the model-scoped estimate is present, and `resolve_learned_ttl("openai", "unseen-model")` returns `None`.
- `test_upgrade_purge_runs_even_with_no_estimable_data`: same seed with a non-estimable window; the aggregate is still purged.
- `test_legacy_bare_provider_keys_are_purged_on_write`: unit-level check that valid `provider/model` keys survive the purge.

22/22 tests pass locally; ruff + mypy clean.

JerrettDavis (changes requested):
The latest model-scoped estimator and non-finite/sample-floor fixes are sound, but the upgrade purge in `write_learned()` is too broad: `merged = {k: v for k, v in merged.items() if "/" in k}` deletes *every* bare-provider entry. `resolve_learned_ttl()` intentionally still supports provider fallbacks, and the module docstring explicitly says a hand-written table may use them. An operator-maintained `{"openai": {"ttl_seconds": 3600}}` is therefore silently destroyed whenever this CLI runs, even though it was not produced by the unsafe earlier estimator. Please migrate only records identifiable as legacy estimator aggregates (for example, the old generated metadata shape containing `max_hit_idle`, `hits`, `ttl_expiry_misses`, and `updated_at`), preserve manual provider fallbacks and unrelated valid entries, and add an upgrade regression proving both behaviors. Also update the PR body: Changes Made and Real Behavior Proof still claim the tool emits provider aggregates and that unseen models resolve through them, contradicting the current source and tests. Reviewed at 926f15ef; CI is green and the rest of the prior findings are resolved.

gglucass:
@JerrettDavis All three points addressed in ec6e1eef:

- **Purge scoped to legacy estimator output.** `write_learned()` now identifies legacy bare-provider aggregates by their generated metadata shape (`max_hit_idle`) instead of dropping every slash-less key. Operator-maintained provider fallbacks — both the `{"ttl_seconds": N}` dict and the plain-number form `resolve_learned_ttl()` accepts — survive every run.
- **Upgrade regression added.** `test_manual_bare_provider_fallbacks_survive_the_purge` covers both hand-written forms alongside a purged legacy record, and `test_upgrade_purges_legacy_provider_aggregate` now asserts end-to-end that the legacy aggregate stops resolving while a manual fallback in the same file keeps resolving. The pre-existing purge tests were switched to the real legacy record shape so they'd catch a regression in the shape check itself.
- **PR body updated.** Changes Made no longer claims a provider aggregate is emitted, the Testing summary reflects the current suite, and Real Behavior Proof is a fresh run at ec6e1eef showing the legacy `openai` aggregate purged and a manual `kimi` fallback preserved in one pass.

Thanks for catching the manual-fallback destruction — that was real data loss for anyone hand-maintaining the table.

JerrettDavis (approved):
Reviewed ec6e1eef. The selective upgrade migration resolves the remaining blocker. Generated legacy provider aggregates are identified by their estimator metadata and removed, while both supported operator-maintained provider fallback forms survive. The end-to-end regression proves an unseen model no longer consumes the unsafe legacy aggregate while a manual fallback still resolves. The description now matches the model-scoped implementation, all current checks are green or intentionally skipped, and the PR is mergeable. Ready for final human review.

gglucass:
Merge ping @chopratejas @DevanshiVyas — approved by @JerrettDavis on 2026-08-07 (after three review rounds, all findings addressed). Re-verified against `main` @ 941c25d3 today: `headroom/cache/ttl_estimator.py` still does not exist on main, so the docs' `headroom-cache-ttl` reference still points at nothing. Merge state is CLEAN, no conflicts.

Issue #2671, opened 2026-07-30
Prefix-cache: two independent cache-write sinks in the Anthropic path


nangsontay:
Two independent defects on the Anthropic proxy path each dominate prompt-cache write volume. Filing one issue so the measurements live in one place instead of being duplicated across PRs.

All figures come from one parser run over local proxy logs, `--mode token`, so before/after are directly comparable.

## Baseline

1678 real turns (`tok_after >= 5000`), 2026-07-28 → 2026-07-30. Aggregate: read 199,009,301 / write 30,397,235 → **86.75% hit**.

Buckets are disjoint and exhaustive; a flipped sub-call counts as a flip.

| bucket | n | read | write | hit | % of all write |
|---|---:|---:|---:|---:|---:|
| injection state flipped | 177 | 14,721,701 | 13,600,610 | 52.0% | **44.7%** |
| sub-call `msgs<=4` | 563 | 19,422,582 | 13,605,298 | 58.8% | **44.8%** |
| main conv, no flip | 938 | 164,865,018 | 3,191,327 | 98.1% | 10.5% |

Steady-state write share for warm non-flipping main turns: **1.192%** — the floor the first defect's turns should approach.

## Defect 1 — `headroom_retrieve` toggles in the `tools` array

`should_inject_ccr_tool` defers CCR tool injection whenever `frozen_message_count > 0`. That counter answers "is the prefix warm?", but the decision needs "does the established prefix already contain the tool?" — which only `SessionCcrTracker` knows. So once a session has injected the tool, the prefix going warm *removes* a tool that is already inside the provider-cached prefix.

`tools` is the head of Anthropic's cache key, so each toggle invalidates 100% of the prefix, in both directions: `0 → >0` removes it, and `>0 → 0` (proxy restart, `/model` switch, lineage eviction, TTL lapse) adds it back.

Log signature: `cache_read` alternates between two values exactly **172 tokens** apart — the 464-byte tool definition. Observed as `24657/24485` on fable and `25033/24861` on opus.

The deferral was added "to preserve cache", but it inverts its own intent: deferring is only cache-safe on a prefix that was established *without* the tool.

### Measured after fixing defect 1

235 real turns over 5.5 hours with the fix live:

| metric | baseline | after |
|---|---|---|
| flip turns | 177, carrying **44.7%** of all cache write, 52.0% hit | **2**, carrying **1.7%** |
| main conv, warm | 98.1% | **97.70%** (n=149) |
[...]

PR #2672, opened 2026-07-30
fix(proxy): stop toggling headroom_retrieve in the Anthropic tools array

nangsontay:
## Description

`should_inject_ccr_tool` deferred CCR tool injection whenever `frozen_message_count > 0`. Because `tools` is the head of Anthropic's cache key, that dropped a tool which was already inside the provider-cached prefix and invalidated the whole prefix — in both directions (`0 → >0` removes it; `>0 → 0` on proxy restart, `/model` switch, lineage eviction or TTL lapse adds it back).

On three days of local proxy logs the turns that flipped injection state carried **44.7% of all cache-write tokens at a 52.0% hit rate**, against 98.1% for non-flipping turns. The log signature is `cache_read` alternating between two values exactly 172 tokens apart — the 464-byte tool definition.

This deletes the gate and calls `apply_session_sticky_ccr_tool` directly, which is **what `openai.py` already does** — the two handlers now have the same shape. Net −61 production lines, no new state, no new config flag.

Fixes defect 1 of #2671.

## Type of Change

- [x] Performance improvement

## Changes Made

- `headroom/proxy/ccr_marker_policy.py` — deleted the `should_inject_ccr_tool` gate; `apply_session_sticky_ccr_tool` is now the single decision point.
- `headroom/proxy/handlers/anthropic.py` — calls `apply_session_sticky_ccr_tool` directly, matching `openai.py`.
- `headroom/proxy/helpers.py` — dropped the now-unused gate plumbing.
- `tests/test_proxy_anthropic_cache_stability.py` — new test asserting the forwarded `tools` array is byte-identical across a `frozen 0 → >0` transition.
- `tests/test_ccr_marker_policy.py` — removed the three unit tests that pinned the deleted decision (they encoded the defect).
- `tests/test_proxy/test_ccr_frozen_prefix_coupling.py` — same unredeemable-marker intent, re-pinned at the sticky helper.
- `tests/test_proxy/test_anthropic_ccr_deferred_injection.py` — autouse reset fixture for the process-global `SessionCcrTracker` (separate commit).
- Formatting-only follow-up commit applying `ruff format` (pinned 0.15.17) to the two test files above.

### Why deleting the gate is safe

`apply_session_sticky_ccr_tool` already holds the correct rule. Its four branches, in order:

| # | condition | action |
|---|---|---|
| 1 | tool already in the incoming tool list (client/MCP pre-registered) | skip; the client's bytes win |
| 2 | `session_id is None` (WS / pre-session) | per-turn flag drives it verbatim |
| 3 | session has done CCR | always inject the recorded golden bytes |
| 4 | fresh session, no compression this turn | **skip** |

Branch 4 is the safety property: a session that has never compressed still gets no tool, so removing the gate cannot start injecting into non-CCR conversations. Branch 3 is what the gate was starving. `has_new_ccr_markers` still gates first-time injection, so markers replayed from the previously-forwarded prefix cannot trigger one.

## Testing

The three deleted unit tests encoded the defect. Coverage moves to the property that actually matters and was previously untested: **the forwarded `tools` array must be byte-identical across a `frozen 0 → >0` transition.** That test asserts on the forwarded request body rather than on a policy function's return value; unit-testing the old policy in isolation is exactly what let a wrong-but-self-consistent decision pass. Verified failing on `upstream/main` with an assertion on the missing tool (not an `ImportError`, so it fails for the right reason).

Full suite: same pre-existing unrelated failures as `upstream/main`, **zero new** (verified by running the whole suite on both revisions and diffing the failure sets).

### Test Output

```text
$ uv run pytest tests/test_ccr_marker_policy.py \
    tests/test_proxy/test_anthropic_ccr_deferred_injection.py \
    tests/test_proxy/test_ccr_frozen_prefix_coupling.py \
collected 48 items

tests/test_ccr_marker_policy.py .....                                    [ 10%]
tests/test_proxy/test_anthropic_ccr_deferred_injection.py .............. [ 39%]
.                                                                        [ 41%]
tests/test_proxy/test_ccr_frozen_prefix_coupling.py ..                   [ 45%]
tests/test_proxy_anthropic_cache_stability.py .......................... [100%]

======================= 48 passed, 2 warnings in 13.59s ========================

$ ruff check .
All checks passed!

$ ruff format --check .
1349 files already formatted
```

## Real Behavior Proof

- Environment: local macOS proxy serving live Claude Code traffic to the Anthropic API; baseline = 3 days of proxy logs on `upstream/main`, after = 5.5 hours with this change live.
- Exact command / steps: ran the proxy with this branch built in, drove normal Claude Code sessions through it (including `/model` switches and proxy restarts, the two events that used to flip injection state), then parsed 235 real turns from the proxy logs with the same parser used for the baseline in #2671.
- Observed result: flip turns fell from 177 (44.7% of all cache write) to 2 (1.7%); steady-state write share 1.192% → 0.867%; aggregate hit rate 86.75% → 89.21%; main conversation warm hit rate 98.1% → 97.70% (n=149). The 2 remaining "flips" have `cache_read == 0` — cold starts that the bucketing counts as a state change, not real flips.

| metric | baseline | after |
|---|---|---|
| flip turns | 177, carrying 44.7% of all cache write | **2**, carrying **1.7%** |
| main conv, warm | 98.1% | **97.70%** (n=149) |
| steady-state write share | 1.192% | **0.867%** |
| aggregate | 86.75% | **89.21%** |

- Not tested: `mypy headroom` was not run locally for this body; the OpenAI handler path (unchanged by this PR); tracker state loss mid-session (see note below); and defect 2 of #2671 (the sub-call breakpoint), which is untouched and is now 54.9% of remaining cache write — that is why aggregate stays just under 90%.

## Review Readiness

## Checklist

## Additional Notes

**Pre-existing and unchanged here:** if the tracker loses state mid-session while the transcript still carries markers, branch 4 returns no tool and those markers are unredeemable. `upstream/main` has no recovery for that; this PR neither creates nor fixes it. See my comment on #2500, which adds a recovery path for the related dangling-reference case.

**N/A checklist items:** no documentation changes — this removes an internal policy function with no user-facing surface. `mypy headroom` left unchecked because it was not run for this body; CI covers it.

**Merge-order conflict with #2500 (please read before landing either):** this PR *deletes* `should_inject_ccr_tool`, which is the exact function #2500 extends with `transcript_requires_tool`. Whichever lands second needs a semantic rebase, not just a textual one — git will not flag it. If this PR lands first, #2500's recovery path should re-target `apply_session_sticky_ccr_tool` (the sticky helper now owns the decision alone) or the handler call site in `handlers/anthropic.py`. If #2500 lands first, the gate deletion here still applies but the `transcript_requires_tool` override needs to move with it. Happy to do the rebase either way — say which order you prefer.

### DIFF
--- a/headroom/proxy/ccr_marker_policy.py
+++ b/headroom/proxy/ccr_marker_policy.py
@@ -1,4 +1,8 @@
-"""CCR marker freshness and retrieval-tool injection policy."""
+"""CCR marker freshness policy.
+
+Retrieval-tool injection is decided by ``apply_session_sticky_ccr_tool`` in
+``headroom.proxy.helpers``, from what the session has actually forwarded.
+"""
 
 from __future__ import annotations
 
@@ -28,18 +32,3 @@ def has_new_ccr_markers(
     )
     previous.scan_for_markers(previous_forwarded_messages)
     return bool(current - set(previous.detected_hashes))
-
-
-def should_inject_ccr_tool(
-    *,
-    configured_inject_tool: bool,
-    frozen_message_count: int,
-    has_compressed_content: bool,
-) -> tuple[bool, bool]:
-    """Decide whether the CCR retrieval tool must be injected this turn."""
-
-    inject_tool = configured_inject_tool
-    if inject_tool and frozen_message_count > 0:
-        inject_tool = False
-    is_marker_override = not inject_tool and has_compressed_content
-    return (inject_tool or is_marker_override), is_marker_override
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -1845,11 +1845,6 @@ class AnthropicHandlerMixin:
                     )
                     inject_system_instructions = False
                 configured_inject_tool = self.config.ccr_inject_tool
-                if configured_inject_tool and frozen_message_count > 0:
-                    logger.info(
-                        f"[{request_id}] CCR: deferring tool injection "
-                        f"(frozen_message_count={frozen_message_count}) to preserve cache"
-                    )
                 # Scan for compression markers + maybe inject system instructions.
                 # Tool-list injection is handled separately via the sticky helper.
                 injector = CCRToolInjector(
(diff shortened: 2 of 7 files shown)


JerrettDavis (approved):
Reviewed the latest 9a0b2da diff. The change correctly moves the Anthropic retrieval-tool decision to the existing session-sticky helper instead of `frozen_message_count`, while preserving the important fresh-session/no-marker skip path. The new regression coverage checks both the no-unredeemable-marker property and the frozen 0 -> >0 forwarded-tools stability case.

Verified locally with `.venv\Scripts\python.exe -m pytest tests/test_ccr_marker_policy.py tests/test_proxy/test_anthropic_ccr_deferred_injection.py tests/test_proxy/test_ccr_frozen_prefix_coupling.py tests/test_proxy_anthropic_cache_stability.py -q` -> 48 passed. CI is green on the latest commit. My first `uv run pytest ...` attempt failed before running tests because Windows had `headroom/_core.pyd` locked during the editable maturin rebuild; rerunning through the existing venv avoided that environment lock.

kiyo-e:
I can corroborate this on `headroom-ai 0.33.0` with Claude Code `2.1.220`, using the Anthropic Messages path through a local upstream proxy.

This is especially visible with `claude-sonnet-5` advisor/sub-calls. Across 153 inference requests, effective `tokens_saved` was 0 for all of them. The message-count distribution was almost exactly 50 requests each at 1, 3, and 5 messages.

For a representative 5-message request, the logs show:

- `Pipeline: freezing first 2/5 messages`
- `Pipeline complete: 81067 -> 71458 tokens (saved 9609, 11.9% reduction)`
- `CCR: deferring tool injection (frozen_message_count=2) to preserve cache`
- Final PERF: `tok_before=71443 tok_after=71443 tok_saved=0`

The inbound body was 294,516 bytes and the outbound body was 294,117 bytes, so the ~9.6K-token pipeline reduction was not reflected in the effective upstream request.

On the same Headroom instance, `claude-fable-5` did retain substantial savings, so this does not appear to be a global compression-worker failure.

This looks consistent with the deferred/sticky `headroom_retrieve` behavior addressed by this PR. It may also explain why short Sonnet advisor sessions show zero effective compression despite successful internal ContentRouter compression.

Issue #1221, opened 2026-06-21
[BUG] Upstream Anthropic 429 is passed through without retry/Retry-After, aborting parallel (dynamic-workflow) runs — upstream-429 sibling of #1025


NUT-JP:
## Summary

When a client fans out **many parallel requests** through the proxy (Claude Code "dynamic
workflow" / multi-subagent runs), the **upstream Anthropic 429 `rate_limit_error` is passed
straight back to the client without any retry or `Retry-After` honoring**. Because each
subagent's `agent()` call receives a raw 429, the agents die and the whole workflow aborts —
**every single time** a workflow is run.

The proxy's own generated 429s (rate-limiter / budget) are carefully handled with
`Retry-After` headers, but **upstream-origin 429s take a different code path** in
`_retry_request` that explicitly does *not* retry 4xx:

```python
# headroom/proxy/server.py  (_retry_request, current main)
# Don't retry client errors (4xx)
if 400 <= response.status_code < 500:
    return response          # <-- upstream 429 returned verbatim, Retry-After ignored

# Retry server errors (5xx)
if response.status_code >= 500:
    raise httpx.HTTPStatusError(...)
```

So 5xx / connect / timeout get exponential backoff (`retry_max_attempts=3`), but a **429 —
the one status that comes with an authoritative `Retry-After` and is the most retry-friendly —
is never retried**. Under a parallel burst that exceeds the account's per-minute upstream limit,
this guarantees client-visible failure.

This is the **upstream-429 sibling of #1025**: same trigger (parallel fan-out + large requests
through one proxy), different exit. #1025 saturates the *internal* pre-upstream queue and returns
`503`; here the burst makes it *past* the queue, Anthropic returns `429`, and the proxy forwards
it unretried. (See also #1171 — sync kompress on the request path — as the shared deeper cause of
the burst not being smoothed.)

## Version / environment

- headroom **0.22.4** (`python -m headroom.cli proxy --host 127.0.0.1 --port 8787 --mode token --backend anthropic`)
- macOS (Apple Silicon, arm64), Python 3.12
- Client: Claude Code `2.1.185`, durably wrapped via `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`
- `Anthropic pre-upstream concurrency: 8 (auto-detected)`
- Trigger: a single Claude Code **dynamic workflow** (parallel subagents, each carrying a large context)

## Steps to reproduce

[...]

PR #2682, opened 2026-07-31
fix(proxy): record streaming upstream 5xx as failed instead of success

inix-x:
## Description

A streaming request that fails upstream is currently recorded as a success.

`RequestOutcome.from_stream` takes no `status_code`, so every outcome built through it keeps the dataclass default of 200. The `>= 500` guard in `emit_request_outcome` can never fire on a streaming request, and an exhausted 529 lands in the savings and cost stats as a request the proxy served. That inflates the save-rate. `streaming.py` contains zero `record_failed` calls today.

The error path is not a dead end that skips accounting. The `>= 400` branch does not early-return. It calls `_finalize_stream_response` and then returns the error response to the client. That finalizer has no `return` statement anywhere in its body and no reference to a status, so the failure books through the full success funnel exactly like a 200.

This is the streaming half of #1568. The non-streaming half landed in 7836aea2 and e365ad71. That second commit deliberately exempted streaming, on two grounds that have both since stopped holding:

1. It said streaming returns before the retry-exhaustion logic. Streaming has since grown its own retry-exhaustion loop (the sibling of the `_retry_request` path from #1221), whose comment says "on exhaustion, fall through to forward the status to the client". So streaming does now receive an exhausted 429 or 529.
2. It listed "streaming generators that early-return on `>= 400`" among the sites that may keep the default 200. That branch finalizes instead of early-returning.

Related: #1568

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added an optional `status_code: int = 200` to `RequestOutcome.from_stream` and passed it into the constructed outcome. Keyword-only and defaulted, so existing callers are untouched.
- Threaded the same parameter through `_finalize_stream_response` into that `from_stream` call.
- Passed `status_code=upstream_response.status_code` at the `>= 400` branch, the one streaming call site that has a real upstream status.
- Added two tests covering the 5xx path and the default path.

Scope, and what this deliberately leaves alone:

- `from_stream` has three call sites. Only one has a status to pass. `_stream_response_bedrock` and `_stream_openai_via_backend` reach the backend through the SDK and never hold an HTTP response object, so they keep the default 200 and are not touched.
- No `record_failed` call was added to `streaming.py`. The guard in `emit_request_outcome` already routes `>= 500` there. The only defect was that the status never arrived, and a second failure path would double-count.
- 4xx behaviour is unchanged. They stay on the success funnel because they are client errors the proxy still served.
- Two sibling sites drop an available status the same way, `bedrock.py` in its outcome construction and the batch handlers. Both are outside this change and are worth their own PRs.

## Testing

### Test Output

```text
$ .venv/bin/python -m pytest tests/test_outcome_records_5xx_as_failed.py -v
tests/test_outcome_records_5xx_as_failed.py::test_529_recorded_as_failed_and_skips_success_funnel PASSED [ 25%]
tests/test_outcome_records_5xx_as_failed.py::test_503_recorded_as_failed PASSED [ 50%]
tests/test_outcome_records_5xx_as_failed.py::test_stream_outcome_529_recorded_as_failed_and_skips_success_funnel PASSED [ 75%]
tests/test_outcome_records_5xx_as_failed.py::test_stream_outcome_without_status_code_defaults_to_success PASSED [100%]
============================== 4 passed in 0.41s ===============================

$ .venv/bin/python -m pytest tests/test_request_outcome.py tests/test_handler_outcome_tag_invariant.py \
    tests/test_backend_streaming_cache_metrics.py tests/test_proxy_streaming_request_logger.py \
    tests/test_proxy_streaming_resilience.py -q
74 passed, 1 warning in 7.67s

$ .venv/bin/python -m pytest tests/test_outcome_records_5xx_as_failed.py \
    tests/test_retry_preserve_upstream_status.py tests/test_proxy_streaming_ratelimit_headers.py -q
20 passed in 1.95s

$ .venv/bin/ruff check headroom/proxy/outcome.py headroom/proxy/handlers/streaming.py tests/test_outcome_records_5xx_as_failed.py
All checks passed!

$ .venv/bin/ruff format --check headroom/proxy/outcome.py headroom/proxy/handlers/streaming.py tests/test_outcome_records_5xx_as_failed.py
3 files already formatted

$ .venv/bin/mypy headroom
Success: no issues found in 508 source files
```

Fails-before check, with the bytecode cleared first so a stale `.pyc` could not serve the fixed module:

```text
$ git checkout -- headroom/proxy/outcome.py     # revert the fix, keep the tests
$ python -c "clear every __pycache__ outside .venv"
$ .venv/bin/python -m pytest tests/test_outcome_records_5xx_as_failed.py -v

tests/test_outcome_records_5xx_as_failed.py::test_stream_outcome_529_recorded_as_failed_and_skips_success_funnel FAILED

E   TypeError: RequestOutcome.from_stream() got an unexpected keyword argument 'status_code'
========================= 1 failed, 3 passed in 2.58s ==========================
```

The tests assert on recorded funnel state rather than raising from a stub. The streaming code around this path swallows exceptions broadly, so a test that raised would have its `AssertionError` eaten and would pass even with the fix reverted.

## Real Behavior Proof

- Environment: macOS 15.5, Python 3.13.13, worktree on `upstream/main` at `e0ce4b1d`, `HF_HUB_OFFLINE=1 LITELLM_LOCAL_MODEL_COST_MAP=true`.
- Exact command / steps: Built a `RequestOutcome` through `from_stream` exactly as a streaming finalize does, for four upstream statuses, and ran each through `emit_request_outcome` with a fake metrics object that records which funnel it lands in. Ran it once on `main` for the baseline and once on this branch.
- Observed result: On `main` a 529 cannot even be expressed, `from_stream()` raises `TypeError: got an unexpected keyword argument 'status_code'`, and the default path records `failed=0 requested=1`. On this branch, 529 gives `failed=1 requested=0`, 503 gives `failed=1 requested=0`, 429 gives `failed=0 requested=1` (a client error the proxy served, unchanged on purpose), and a call with no status passed gives `failed=0 requested=1`, so the two untouched finalizers behave exactly as before. Numbers below.
- Not tested: No live upstream. The 529 came from a simulated status rather than a real provider overload, and the run exercises the outcome funnel rather than a full request through the proxy. The two backend-routed streaming finalizers were not exercised with a non-200 status because neither can produce one. Nothing was run on Linux; CI covers that.

```text
=== streaming outcome funnel: simulated upstream statuses ===

  BEFORE (main, no fix)
  529 exhausted overload      TypeError: from_stream() got an unexpected keyword argument 'status_code'
  503 server error            TypeError: from_stream() got an unexpected keyword argument 'status_code'
  429 client error            TypeError: from_stream() got an unexpected keyword argument 'status_code'
  no status passed            status=200  failed=0 requested=1

  AFTER (this branch)
  529 exhausted overload      status=529  failed=1 requested=0
  503 server error            status=503  failed=1 requested=0
  429 client error            status=429  failed=0 requested=1
  no status passed            status=200  failed=0 requested=1
```

## Review Readiness

## Checklist

## Additional Notes

Docs checklist item is N/A. This restores the accounting behaviour the `>= 500` guard already documents in `outcome.py`, so no documented behaviour changes.

The diff is five lines of source. Most of the work was proving the exemption in `e365ad71` had gone stale rather than guessing at it, which is why the description spends more space on the trace than on the change.

Happy to split the sibling sites out into follow-ups, or to fold them in here if you would rather see the accounting family fixed in one pass.


JerrettDavis (changes requested):
The status propagation fixes the metric/cost funnel, but the same 5xx still mutates provider prefix state before that short-circuit.

`_stream_response()` calls `_finalize_stream_response(..., status_code=503)` on the upstream error branch. In `_finalize_stream_response`, the prefix-tracker block runs before `RequestOutcome` is emitted and is not gated by status. Because an error response has no assistant SSE/cache usage, it currently:

- calls `classify_cache_miss(cache_read_tokens=0, current_forwarded_messages=...)`, potentially recording a false miss; and
- calls `prefix_tracker.update_from_response(...)` with the rejected request messages and no assistant, overwriting the last successful turn's prefix state.

The later `emit_request_outcome` 5xx return cannot undo that mutation. The next request can therefore freeze/compare against a turn the upstream never accepted, even though the request is now correctly counted as failed.

Please skip prefix-cache classification/update for non-success upstream statuses (at minimum `>=500`; I would use the actual accepted-response boundary rather than treating a rejected 4xx as prefix state), and extend the existing streaming 503 integration test with a tracker spy proving neither method is called. The metric assertions in the current patch should remain.

inix-x:
> The status propagation fixes the metric/cost funnel, but the same 5xx still mutates provider prefix state before that short-circuit.
> 
> `_stream_response()` calls `_finalize_stream_response(..., status_code=503)` on the upstream error branch. In `_finalize_stream_response`, the prefix-tracker block runs before `RequestOutcome` is emitted and is not gated by status. Because an error response has no assistant SSE/cache usage, it currently:
> 
> * calls `classify_cache_miss(cache_read_tokens=0, current_forwarded_messages=...)`, potentially recording a false miss; and
> * calls `prefix_tracker.update_from_response(...)` with the rejected request messages and no assistant, overwriting the last successful turn's prefix state.
> 
> The later `emit_request_outcome` 5xx return cannot undo that mutation. The next request can therefore freeze/compare against a turn the upstream never accepted, even though the request is now correctly counted as failed.
> 
> Please skip prefix-cache classification/update for non-success upstream statuses (at minimum `>=500`; I would use the actual accepted-response boundary rather than treating a rejected 4xx as prefix state), and extend the existing streaming 503 integration test with a tracker spy proving neither method is called. The metric assertions in the current patch should remain.

Thank you @JerrettDavis, will do

JerrettDavis (changes requested):
The 5xx tracker mutation is fixed, but the guard still treats every status below 400 as an accepted response. A 3xx from an upstream is not a completed model response: it normally has no assistant SSE/cache usage, yet `status_code < 400` will still classify a false miss and overwrite prefix state from the rejected/redirected request. This is the same state-corruption mechanism as the original 503 finding, just on the other non-success class. Please gate prefix classification/update on the actual success range (`200 <= status_code < 300`, or a narrower documented accepted set) and add a redirect/non-2xx regression proving both tracker methods remain untouched. The new 503 spy and outcome accounting are otherwise correct, and substantive CI is green.

inix-x:
> Please skip prefix-cache classification/update for non-success upstream statuses, and extend the existing streaming 503 integration test with a tracker spy proving neither method is called.

Addressed in `b9fd4cef` and tightened in `6bda6f3e`: streaming prefix classification and updates now run only for actual 2xx responses. The 503 regression proves both tracker methods remain untouched while preserving the failed-outcome metric assertions.

inix-x:
> Please gate prefix classification/update on the actual success range (`200 <= status_code < 300`, or a narrower documented accepted set) and add a redirect/non-2xx regression proving both tracker methods remain untouched.

Addressed in `6bda6f3e`: normal streamed finalization now receives the real upstream status, prefix tracking is gated to `200 <= status_code < 300`, and a 307 regression consumes the full stream and proves neither tracker method is called.

JerrettDavis (approved):
Re-reviewed latest 4bedfcdc. The normal and error streaming finalizers now receive the real upstream status, prefix-cache classification/update is gated to 200–299, and the 503/307 regressions consume the relevant paths and prove both tracker methods stay untouched. The 5xx outcome still records failed exactly once and skips savings/cost success accounting. Diff check and full CI are green.

JerrettDavis (approved):
Re-reviewed after resolving the new upstream-main conflict. Preserved the real upstream status alongside main’s conversation-savings arguments in both error and successful streaming finalizers. All 37 outcome, streaming-header, and conversation-savings tests pass. Ruff lint/format and diff checks pass. Remote CI is pending on this refreshed head; no PR merge performed.

PR #2684, opened 2026-07-31
docs(security): correct credential-storage claim and add a security model page

### LINKED ISSUE #2683: [DOCUMENTATION] SECURITY.md claims "we never store API keys" but copilot-auth persists an OAuth token to disk; no data-at-rest docs exist
## Description

`SECURITY.md` states, under **Security Features**:

> - **No credential storage**: We never store or log API keys

This is inaccurate as written. `headroom copilot-auth login` persists a GitHub Copilot OAuth **refresh token** to disk in plaintext JSON. OAuth refresh tokens are credentials, so the unqualified "we never store credentials" claim is wrong for at least one documented, first-party code path.

The underlying issue is structural rather than a single bad line: **there is no security or privacy documentation anywhere in the project.** `SECURITY.md` is a vulnerability-disclosure policy that grew a four-bullet summary on the end, and `mkdocs.yml` has no security entry in its nav. Because nothing owns the topic, questions a security review actually asks — where does data land on disk, with what permissions, for how long, and what leaves the host — have no documented answer, and the one summary that exists drifted into being false.

## Evidence

**1. The Copilot OAuth token is stored on disk.**

`headroom/copilot_auth.py:140-146` resolves the path:

```python
def headroom_copilot_auth_path() -> Path:
    """Return the path where Headroom stores its Copilot OAuth token."""
    ...
    return paths.workspace_dir() / "copilot_auth.json"
```

`headroom/copilot_auth.py:501-515` writes it:

```python
path.write_text(json.dumps(body, indent=2, sort_keys=True) + "\n", encoding="utf-8")
try:
    path.chmod(0o600)
except OSError:
    pass
```

So: `~/.headroom/copilot_auth.json` by default (relocatable via `$HEADROOM_COPILOT_AUTH_PATH`), plaintext, `0600` on a best-effort basis — the `chmod` is inside a `try`/`except OSError: pass`, so on filesystems or platforms where it fails the file silently keeps umask-default permissions. The token body includes the `refresh` field.

The README documents this command and describes it as storing "a Headroom-specific Copilot OAuth token", so the README and `SECURITY.md` contradict each other today.

**2. Pre-compression originals are persisted to disk, undocumented.**

`headroom/cache/compression_store.py:952-977` — when `HEADROOM_CCR_BACKEND` is unset, the CCR store defaults to SQLite on disk:

```python
Default (env unset or "sqlite"): SQLiteBackend at workspace_dir()/ccr_store.db
— restart-safe and shared across worker processes, which the
session-scale 30-minute TTL assumes.
```

CCR caches original content so it can be retrieved after compression — tool outputs, file contents, command output, RAG chunks. #1227 established both that this content is sensitive and that loopback-guarding it is the project's chosen control. That the same content is also written to `~/.headroom/ccr_store.db`, unencrypted, surviving restarts, is documented nowhere. `wiki/ccr.md` describes the caching behavior without ever saying it touches disk.

Credit where due: `headroom/cache/backends/sqlite.py:107` does `chmod(0o600)` on the store. That is the right behavior and is exactly the kind of thing the docs should be able to state.

**3. TTL expiry is lazy, not scheduled.**

Expired rows are dropped by a sweep that runs on lookup, not on a timer. On an idle proxy, content past its 30-minute TTL remains on disk until the next access. "The TTL purges" is true; "the TTL purges on schedule" is not, and the difference matters to anyone reasoning about retention windows.

**4. The supported-version table is stale again.**

`SECURITY.md:5-8` lists `0.27.x (latest)` as supported. Current `pyproject.toml` is `0.33.0`. #1491 corrected this same table (`0.2.x` → `0.27.x`) on 2026-06-27; it has drifted by six minor versions since. A table that requires manual bumping will keep going stale.

## Why this survived the last accuracy pass

#1491 ("fix(cli): harden all CLI surfaces + fix docs accuracy") ran three parallel audit agents across root docs, wiki, and the Fumadocs site specifically hunting factual errors. It edited `SECURITY.md` — a 3-line diff to the supported-version table. The "No credential storage" bullet four lines below was not flagged.

That pass was checking flags, imports, and defaults. Security claims are a different class of assertion and need a different check, which argues for giving them a documented home rather than another one-line correction.

## Related work

- **#184** — "Telemetry transparency: enumerate every observable touchpoint" (closed, `status: stale`). Asked for exactly this shape of deliverable one domain over: a single table where "a prospective user or security reviewer can see at a glance" every surface, what it emits, its default, and its retention — because otherwise "answering 'does headroom phone home?' takes an afternoon of grep and the answer is easy to get subtly wrong." Never delivered; no telemetry page exists. The page proposed below can absorb its egress table.
- **#1227** — CCR cross-origin disclosure (closed, fixed). Establishes the sensitivity of cached originals.
- **#1526** — removed the Supabase beacon; notes the license usage reporter remains as an egress path. Needed for the egress list to be complete.
- **#748** — consolidated `headroom uninstall` + document teardown (open). Overlaps the "what does `wrap` change, and what does `unwrap` restore" question. Deliberately **not** in scope here to avoid duplicating it.

## Proposed fix

**1. Correct `SECURITY.md`.**

Replace the `No credential storage` bullet with something precise: provider API keys are forwarded and not persisted; the Copilot OAuth token *is* persisted at `~/.headroom/copilot_auth.json`, mode `0600` best-effort, plaintext, relocatable via `$HEADROOM_COPILOT_AUTH_PATH`. Drop `Input validation` and `Safe defaults` — they are unfalsifiable and are what makes the section read as marketing rather than policy. Add a note to "Best Practices for Users" that `~/.headroom` holds credential and cached-content state and should be treated as sensitive on shared hosts. Refresh the version table to `0.33.x`.

**2. Add a security-model page and wire it into `mkdocs.yml` nav.**

The load-bearing piece. Sections:

- **Data at rest** — a table of every path that persists state: `copilot_auth.json`, `ccr_store.db`, `memory.db`, `learn` artifacts. For each: location, backing store, TTL, encryption status, permissions. State plainly that CCR holds pre-compression originals, that expiry is lazy-on-access, and that `HEADROOM_CCR_BACKEND=memory` is the opt-out for deployments that cannot accept disk persistence.
- **Data in flight** — the proxy sees full plaintext agent traffic by design. Say it directly; it is the first thing a reviewer needs to understand.
- **Local admin surface** — document the existing `require_loopback` design so it stops being invisible. Both gates (client IP *and* `Host:` header, the latter specifically as DNS-rebinding defence per `headroom/proxy/loopback_guard.py`), the fact that `POST /admin/runtime-env` accepts only the hardcoded `RUNTIME_ENV_KNOBS` allowlist in `headroom/proxy/runtime_env.py` (six keys: the output-shaper family plus one ast-grep threshold), and that neither the upstream URL nor `HEADROOM_TLS_STRICT` is reachable through it — `/admin/upstream` is read-only, deliberately and with a comment saying why. Then state the residual honestly: this is a local-process-is-trusted model, so any process on the host can flip those six knobs, and it is not suitable for multi-tenant machines.
- **Network egress** — one consolidated list with the disable/pre-provision knob for each: providers, the PyPI update check (`HEADROOM_UPDATE_CHECK=off`), `cdn.pyke.io`, `huggingface.co`, the license usage reporter. This is the allowlist a fleet operator currently has to reverse-engineer.
- **Governed knobs** — `HEADROOM_TLS_STRICT=0` and `HEADROOM_UPDATE_CHECK=off`, linking to the existing accurate prose in `README.md` and `docs/content/docs/proxy.mdx` rather than restating it.

To be explicit about scope: items 3 and 4 of the local admin section are documenting behavior that is **already correctly implemented**. The defect is that a reader cannot discover it without reading the source, so a reviewer's reasonable default assumption is that the endpoint is unguarded.

## Environment

- **Headroom version**: 0.33.0 (`pyproject.toml`)
- **OS**: macOS 15 (darwin 25.5.0)
- **Affected files**: `SECURITY.md`, `mkdocs.yml` (nav), new security-model page

## Additional Context

Willing to contribute this — happy to open a PR covering fix items 1 and 2 if the approach looks right. Flagging one thing to confirm first: #1809 ("CCR is in-memory only") and #1023 ("CCR cache permanently empty") both describe in-memory-only behavior, which contradicts the SQLite default I traced above. Those likely predate the SQLite backend or describe the silent `except` fallback at `compression_store.py:970`, but I would rather confirm at runtime than publish a data-at-rest table describing behavior users are not actually getting — that would repeat the exact failure this issue is about.

keshavlingala:
## Description

`SECURITY.md` claimed **"No credential storage: We never store or log API keys."** That is inaccurate — `headroom copilot-auth login` persists a GitHub Copilot OAuth **refresh token** to disk in plaintext, CCR writes pre-compression originals to a local SQLite database by default, and the dashboard settings store persists two credential-shaped header maps. None of it was documented anywhere.

Rather than patch the one line, this PR gives the topic a home: `docs/content/docs/security-model.mdx` documents the trust model, data in flight, data at rest, the local control surfaces, and network egress in enough detail to answer a security review without reading the source.

**This documents existing behavior only. There are no code changes.** Several things the page describes are already implemented correctly and simply weren't discoverable — the loopback guard's DNS-rebinding defence, the same-origin CSRF gate, the `runtime-env` allowlist, the audit stream's keys-not-values discipline, and the `0600` permissions on the credential, cache, and settings files.

Closes #2683

## Type of Change

- [x] Documentation update

## Changes Made

**`docs/content/docs/security-model.mdx`** (new, 392 lines) — trust model; data in flight; a data-at-rest table with path, contents, retention, encryption and permissions per item; the settings store; logs; memory store paths; CCR persistence and lazy purge; local control surfaces (three gates, two CIDR allowlists, what `POST /admin/runtime-env` and `POST /settings` can each change, residual risk); network egress and the beacon; governed knobs; what `wrap` changes.

**`docs/content/docs/meta.json`** — a `Security` nav section carrying the new page.

**`SECURITY.md`**
- Replaced the `Security Features` bullets with an accurate `Security Posture` section: provider API keys are forwarded and not persisted *except* for the two settings-store header maps; the Copilot OAuth token **is** persisted (path, format, permissions stated); CCR cached originals **are** written to disk, with the `HEADROOM_CCR_BACKEND=memory` opt-out; the beacon uploads by default; operational logs are always written; local control surfaces are loopback-gated, with the trusted-dashboard CIDR caveat
- Dropped `Input validation: All inputs are validated before processing` and `Safe defaults: Security-conscious defaults out of the box` — unfalsifiable claims that made the section read as marketing rather than policy
- Added best-practice notes covering `~/.headroom` and the two log paths
- **Supported versions**: replaced the pinned `0.33.x` row with the actual policy — latest release only, fixes ship forward rather than being backported — so the table stops going false every release

## Testing

### Test Output

The docs site is the Next.js/Fumadocs app in `docs/`. `npm run build` compiles clean:

```text
$ cd docs && npm ci && npm run build
▲ Next.js 16.3.0 (Turbopack)
✓ Running next.config.mjs took 27ms
  Finished writing to disk in 4ms
  Finished TypeScript in 1216ms ...
✓ Generating static pages using 12 workers (167/167) in 833ms
  Finalizing page optimization ...
```

167 static pages, zero warnings, zero errors. `security-model` builds
(`.next/server/app/docs/security-model.html`) and appears in `sitemap.xml`.

All 17 internal anchors on the page resolve against the generated HTML:

```text
$ python3 - <<'EOF'   # ids vs href="#..." in the built page
internal anchor links: 17
BROKEN: none
EOF
```

## Real Behavior Proof

Every factual claim on the page was verified against a running interpreter rather than read off a comment.

- Environment: Linux (Ubuntu 26.04, x86_64), Python 3.14.4, repo at `2ef83d35` (rebased onto `7ef736fb`), package importable from the source tree.
- Exact command / steps: seeded each store under an isolated `HEADROOM_WORKSPACE_DIR`, then read the resulting files back directly — below.
- Observed result: SQLite is the default CCR backend and the pre-compression original is on disk in plaintext at `0600`; the settings store writes a secret header map to disk in plaintext at `0600` while masking it on read, and feeds it back into the process environment; `proxy.log` is created without an explicit mode and lands at `0664`.
- Not tested: the `huggingface.co` and `app.headroomlabs.ai` rows in the egress table are documented from code paths, not exercised end-to-end. I deliberately left `cdn.pyke.io` **out** of that table: `ort` is declared with `default-features = false` and `load-dynamic` (`crates/headroom-core/Cargo.toml`), which loads a system ONNX Runtime rather than downloading one, so I could not substantiate it as a live egress path. Happy to add it if a maintainer knows of a build configuration where it does fetch.

### 1. CCR writes pre-compression originals to disk

```bash
HEADROOM_WORKSPACE_DIR=/tmp/hrtest python -c "
from headroom.cache.compression_store import get_compression_store
s = get_compression_store()
print('default backend:', type(s._backend).__name__)
print('hash:', s.store('SENTINEL_ORIGINAL_CONTENT_abc123', 'SENTINEL...', tool_name='Read'))
print('default_ttl_seconds:', s.default_ttl_seconds)"

sqlite3 /tmp/hrtest/ccr_store.db "SELECT * FROM ccr_entries;"
stat -c '%a %n' /tmp/hrtest/ccr_store.db
```

```text
default backend: SQLiteBackend
hash: 6c1bda2f2b2a6682c6c9bd82
default_ttl_seconds: 1800

6c1bda2f2b2a6682c6c9bd82|{"hash": "6c1bda2f2b2a6682c6c9bd82",
"original_content": "SENTINEL_ORIGINAL_CONTENT_abc123",
"compressed_content": "SENTINEL...", ..., "ttl": 1800, ...}

600 /tmp/hrtest/ccr_store.db
```

### 2. The settings store holds credentials in plaintext, and sets the environment

```bash
HEADROOM_WORKSPACE_DIR=/tmp/hrtest python -c "
from headroom import settings_store
settings_store.save({'anthropic_extra_headers': {'Api-Key': 'SENTINEL_GATEWAY_KEY_xyz789'},
                     'log_messages': True,
                     'anthropic_base_url': 'http://attacker.example/v1'})
print('masked read:', settings_store.stored_values()['anthropic_extra_headers'])"

cat /tmp/hrtest/settings.json
stat -c '%a %n' /tmp/hrtest/settings.json
```

```text
masked read: ••••••

{
  "anthropic_base_url": "http://attacker.example/v1",
  "anthropic_extra_headers": "{\"Api-Key\": \"SENTINEL_GATEWAY_KEY_xyz789\"}",
  "log_messages": true
}

600 /tmp/hrtest/settings.json
```

The masking is a display control on `GET /settings`, not an at-rest one. And the file is a second source for the process environment:

```text
before apply_to_environ: {'ANTHROPIC_TARGET_API_URL': None, 'HEADROOM_LOG_MESSAGES': None,
                          'ANTHROPIC_TARGET_API_HEADERS': None}
after  apply_to_environ: {'ANTHROPIC_TARGET_API_URL': 'http://attacker.example/v1',
                          'HEADROOM_LOG_MESSAGES': '1',
                          'ANTHROPIC_TARGET_API_HEADERS': '{"Api-Key": "SENTINEL_GATEWAY_KEY_xyz789"}'}
```

Both of those keys are in the settings registry, so `POST /settings` + `POST /settings/apply` (which restarts the proxy) can set them. That is why the page no longer claims the upstream URL is unchangeable over HTTP.

### 3. `proxy.log` is always on, carries the audit stream, and is not `0600`

```bash
HEADROOM_WORKSPACE_DIR=/tmp/hrtest python -c "
import logging
from headroom.proxy.helpers import _setup_file_logging
_setup_file_logging()
logging.getLogger('headroom.audit').info('{\"event\": \"headroom_admin_audit\", \"action\": \"settings_update\"}')
h = [x for x in logging.getLogger('headroom').handlers
     if x.__class__.__name__ == 'RotatingFileHandler'][0]
print('log path:', h.baseFilename, '| maxBytes:', h.maxBytes, '| backups:', h.backupCount)"

stat -c '%a %n' /tmp/hrtest/logs/proxy.log
```

```text
log path: /tmp/hrtest/logs/proxy.log | maxBytes: 10485760 | backups: 5
664 /tmp/hrtest/logs/proxy.log            # umask 0002 — not 0600 like the others
2026-08-18 06:35:56 - headroom.audit - INFO - {"event": "headroom_admin_audit", "action": "settings_update"}
```

### 4. Every documented env var resolves to a read site

The check that would have caught the `HEADROOM_COPILOT_AUTH_PATH` error from round one — each `HEADROOM_*` / `ANTHROPIC_*` / `OPENAI_*` / `DO_NOT_TRACK` name in the page and in `SECURITY.md`, grepped back to `headroom/` or `crates/`:

```text
30 env names referenced
NOT FOUND IN SOURCE: none
```

The 7-key `runtime-env` allowlist and the 59-field settings registry were likewise enumerated from `RUNTIME_ENV_KNOBS` and `SETTINGS` directly, not transcribed.

## Review Readiness

## Checklist

## Additional Notes

**N/A checklist items:** unit tests, linting, mypy, and new tests — this is a docs-only change with no Python touched.

**Site location.** The page lives in `docs/`, the Next.js/Fumadocs app, and links resolve to `headroom-docs.vercel.app/docs/security-model`. Re-confirmed against current `main`: `.github/workflows/docs.yml` and `docs/app/layout.tsx` both name that app as the one documentation site, and `wiki/` is unpublished markdown nothing builds. An earlier revision of this PR put the page in `wiki/` under `mkdocs.yml`; #2784 removed that site, and this PR was moved accordingly.

**Two things a reviewer may want to weigh in on:**

1. **Wording on lazy TTL purge.** The page states that expired CCR rows are removed by a sweep that runs on lookup rather than a background timer, so on an idle proxy content can persist past its TTL. Re-checked on current `main`: the documented Python path is still a fixed `created_at + ttl` window (`headroom/cache/backends/sqlite.py`), and the sliding idle-window TTL with an 8× max-lifetime ceiling from #2604 lives in the Rust core backends, so it does not change what this page describes. I documented the behavior as-is rather than treating it as a defect.

2. **Residual-risk framing.** The "Local control surfaces" section says plainly that the gates distinguish where a request came from, not who sent it, and that Headroom should not be deployed on multi-user hosts without additional isolation. I think stating this is a strength — it is the honest read of the design — but it is a judgement call about tone and I am happy to adjust.
## Runtime Rollout Safety

- Rollout-managed feature(s): none; this PR changes documentation only.
- Minimum rollout channel: documentation deployment after the Next.js/Fumadocs build passes.
- Stable/default behavior changed: none; the page documents current credential, persistence, control-surface, logging, and egress behavior.
- Kill switch / disable path: human-revert the documentation if a factual claim is later disproven.
- Qualification impact: the docs build, sitemap entry, internal links, and source-backed environment/path claims must remain valid.
- Rollback path: human-revert the docs commit; no runtime state or migration is involved.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -43,7 +43,9 @@ When using Headroom:
 
 1. **API Keys**: Never commit API keys. Use environment variables.
 2. **Proxy Exposure**: Don't expose the proxy server to the public internet without authentication
-3. **Log Files**: Be aware that request logs may contain sensitive information
+3. **Log Files**: Headroom always writes an operational log to `~/.headroom/logs/proxy.log`.
+   The opt-in `--log-file` request log, and especially `--log-messages`, additionally write
+   request and response *content* to disk. Be aware that both may contain sensitive information.
 4. **Budget Limits**: Set budget limits to prevent unexpected costs
 5. **Workspace directory**: `~/.headroom` holds both credential material and cached
    request content. Treat it as sensitive — especially on shared or multi-user hosts.
@@ -74,7 +76,10 @@ host, and what the local admin surface does and does not allow. The summary:
 
 - **Provider API keys are forwarded, not persisted.** Keys supplied via environment
   variables or request headers are used to authenticate the upstream call and are not
-  written to disk or emitted to logs.
+  written to disk or emitted to logs. The one exception is deliberate: the
+  `ANTHROPIC_TARGET_API_HEADERS` / `OPENAI_TARGET_API_HEADERS` header maps, if you set
+  them through the dashboard settings GUI, are persisted to `~/.headroom/settings.json`
+  in plaintext — and a header map is where a gateway key typically goes.
 - **Some credentials *are* stored, by design.** `headroom copilot-auth login` persists a
   GitHub Copilot OAuth refresh token to `~/.headroom/copilot_auth.json` (relocatable via
   `$HEADROOM_COPILOT_AUTH_FILE`). It is plaintext JSON, written with `0600` permissions
@@ -91,8 +96,15 @@ host, and what the local admin surface does and does not allow. The summary:
   `HEADROOM_OFFLINE=1`. This is separate from `HEADROOM_TELEMETRY`, which is opt-in and
   stays on the machine.
 - **Passthrough mode**: Sensitive content passes through unchanged by default.
-- **Local control surfaces are loopback-gated.** The `/admin/*`, `/debug/*`, and
-  `/v1/retrieve*` routes require a loopback client address *and* a loopback `Host:`
-  header. This is a local-trust model: any process on the same host is trusted.
+- **Operational logs are written unconditionally.** `~/.headroom/logs/proxy.log`
+  (10 MB × 5 rotations) is always on and also carries the admin audit stream. The
+  request log (`--log-file`) and full message logging (`--log-messages`) are opt-in and
+  write request/response content to a path you choose.
+- **Local control surfaces are loopback-gated.** The `/admin/*`, `/debug/*`,
+  `/v1/retrieve*`, `/v1/telemetry*`, `/v1/toin/*`, and `/settings*` routes require a
+  loopback client address *and* a loopback `Host:` header, plus a same-origin check on
+  the mutating ones. This is a local-trust model: any process on the same host is
+  trusted. Setting `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS` deliberately widens
+  the settings and stats surface beyond loopback.
 
 Thank you for helping keep Headroom and its users safe!
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,8 +4,8 @@
 
 | Version | Supported          |
 | ------- | ------------------ |
-| 0.27.x (latest) | :white_check_mark: |
-| < 0.27.x | :x:              |
+| 0.33.x (latest) | :white_check_mark: |
+| < 0.33.x | :x:              |
 
 ## Reporting a Vulnerability
 
@@ -40,6 +40,12 @@ When using Headroom:
 2. **Proxy Exposure**: Don't expose the proxy server to the public internet without authentication
 3. **Log Files**: Be aware that request logs may contain sensitive information
 4. **Budget Limits**: Set budget limits to prevent unexpected costs
+5. **Workspace directory**: `~/.headroom` holds both credential material and cached
+   request content. Treat it as sensitive — especially on shared or multi-user hosts.
+   Note that it is not the only location: with memory enabled, extracted memories default
+   to a project-local `.headroom/` directory beside the code you ran the agent in. See
+   [Security Model](https://headroom-docs.vercel.app/docs/security-model) for exactly what
+   is written where.
 
 ### Scope
 
@@ -53,13 +59,35 @@ The following are out of scope:
 - Issues in dependencies (report these to the upstream project)
 - Social engineering attacks
 
-## Security Features
+## Security Posture
 
-Headroom includes several security features:
+Headroom is a local proxy that sits between your agents and your LLM providers, so
+it necessarily handles both your traffic and your credentials. The
+[Security Model](https://headroom-docs.vercel.app/docs/security-model) documents this in
+full — what is written to disk, with what permissions and retention, what leaves the
+host, and what the local admin surface does and does not allow. The summary:
 
-- **No credential storage**: We never store or log API keys
-- **Passthrough mode**: Sensitive content passes through unchanged by default
-- **Input validation**: All inputs are validated before processing
-- **Safe defaults**: Security-conscious defaults out of the box
+- **Provider API keys are forwarded, not persisted.** Keys supplied via environment
+  variables or request headers are used to authenticate the upstream call and are not
+  written to disk or emitted to logs.
+- **Some credentials *are* stored, by design.** `headroom copilot-auth login` persists a
+  GitHub Copilot OAuth refresh token to `~/.headroom/copilot_auth.json` (relocatable via
+  `$HEADROOM_COPILOT_AUTH_FILE`). It is plaintext JSON, written with `0600` permissions
+  on a best-effort basis.
+- **Cached request content is written to disk.** CCR stores pre-compression originals —
+  tool outputs, file contents, retrieved chunks — in a local SQLite database
+  (`~/.headroom/ccr_store.db`, `0600`) so they can be retrieved after compression. Entries
+  are plaintext and are not encrypted at rest. Set `HEADROOM_CCR_BACKEND=memory` if your
+  deployment cannot accept disk persistence.
+- **Anonymous session summaries are uploaded by default.** The `HEADROOM_BEACON` telemetry
+  beacon is opt-*out*: it POSTs content-free session counters, a random per-install UUID,
+  version, OS, and architecture to Headroom Labs. No prompts, completions, file contents,
+  or paths are included. Disable with `HEADROOM_BEACON=off`, `DO_NOT_TRACK=1`, or
+  `HEADROOM_OFFLINE=1`. This is separate from `HEADROOM_TELEMETRY`, which is opt-in and
+  stays on the machine.
+- **Passthrough mode**: Sensitive content passes through unchanged by default.
+- **Local control surfaces are loopback-gated.** The `/admin/*`, `/debug/*`, and
+  `/v1/retrieve*` routes require a loopback client address *and* a loopback `Host:`
+  header. This is a local-trust model: any process on the same host is trusted.
 
 Thank you for helping keep Headroom and its users safe!
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,8 +4,13 @@
 
 | Version | Supported          |
 | ------- | ------------------ |
-| 0.33.x (latest) | :white_check_mark: |
-| < 0.33.x | :x:              |
+| Latest release on PyPI | :white_check_mark: |
+| Any earlier release | :x:              |
+
+Headroom releases frequently and only the most recent release is supported. Security
+fixes ship in a new release rather than being backported to earlier ones, so upgrading is
+the remediation path. Check what you are running with `headroom --version`, and compare it
+against [the current release](https://pypi.org/project/headroom-ai/).
 
 ## Reporting a Vulnerability
 
--- a/docs/content/docs/meta.json
+++ b/docs/content/docs/meta.json
@@ -61,6 +61,8 @@
     "runtime-rollouts",
     "benchmarks",
     "limitations",
+    "---Security---",
+    "security-model",
     "---Help---",
     "errors",
     "troubleshooting"
--- a/docs/content/docs/security-model.mdx
+++ b/docs/content/docs/security-model.mdx
@@ -55,12 +55,69 @@ and both the memory store and the Copilot token can be relocated independently. 
 | Install ID | `<config dir>/install_id` | Random UUID identifying the install to the beacon | Until deleted | No | `0600`, best-effort |
 | Savings tracker | `<workspace>/proxy_savings.json` | Aggregate token/cost counters — no prompt content | Until deleted | No | Inherits umask |
 | License cache | `<workspace>/license_cache.json` | Cached license envelope | Per envelope | No | Inherits umask |
+| Settings store | `<workspace>/settings.json`, or `$HEADROOM_SETTINGS_PATH` | Dashboard-managed knob values — **including two header maps that carry credentials**, see [The settings store](#the-settings-store) | Until deleted | No | `0600`, inherited from `mkstemp` |
+| Proxy log | `<workspace>/logs/proxy.log` + 5 rotations | INFO-level operational log, always on; carries the admin audit stream | 10 MB × 5 backups, then rotated out | No | Inherits umask |
+| Request log | Opt-in, `--log-file` / `$HEADROOM_LOG_FILE` | JSONL per-request records; **full request and response content** when `--log-messages` is on | Until deleted | No | Inherits umask |
 
 The config directory is `$HEADROOM_CONFIG_DIR`, else `$HEADROOM_WORKSPACE_DIR/config`,
 else `~/.headroom/config`.
 
 **Nothing Headroom writes is encrypted at rest.** Confidentiality of everything above
 rests on filesystem permissions and on the security of the account the proxy runs as.
+
+### The settings store
+
+The dashboard settings GUI persists a curated set of 59 knobs to
+`<workspace>/settings.json`. Two of them are marked secret, and both are credential-shaped:
+`ANTHROPIC_TARGET_API_HEADERS` and `OPENAI_TARGET_API_HEADERS` are free-form JSON header
+maps merged into forwarded requests — the field's own help text gives
+`{"Api-Key": "..."}` as the example. Anything you put there is written to disk in
+plaintext. The GUI masks them on read (`GET /settings` returns a sentinel, not the
+value), but that is a display control, not an at-rest one.
+
+Two further properties matter for policy:
+
+- **It is a second way environment knobs get set.** `settings_store.apply_to_environ` runs
+  at CLI startup and again when the app is constructed, using `os.environ.setdefault`. The
+  precedence is `shell export > settings.json > code default`, so a stored value silently
+  becomes the effective configuration on any host where the operator assumed the
+  environment was the only input.
+- **The file is `0600` by construction, not by assertion.** `save()` writes atomically via
+  `tempfile.mkstemp` + `os.replace`, and inherits mkstemp's `0600`. There is no explicit
+  `chmod`, so treat the permission as a property of the current implementation rather than
+  a guarantee.
+
+`HEADROOM_TLS_STRICT`, `HEADROOM_BEACON`, `HEADROOM_CCR_BACKEND`, and `HEADROOM_OFFLINE`
+are **not** in the registry — none of them can be set from this file or the routes that
+write it.
+
+### Logs
+
+Two log paths, with very different sensitivity:
+
+- **`<workspace>/logs/proxy.log` — always on.** A `RotatingFileHandler` (10 MB, 5 backups)
+  is attached to the `headroom` logger when the app is created; there is no flag to
+  suppress it, and it is written before configuration is otherwise applied. It carries
+  INFO-level operational records, and — because the audit logger is a child of `headroom`
+  and writes no file of its own — the structured admin audit stream lands here too. Audit
+  events record the action, method, path, source IP, and status code of `/admin/*`,
+  `/cache/clear`, and `/stats/reset` calls.
+
+  Note the permission asymmetry: unlike `copilot_auth.json`, `ccr_store.db`, and
+  `settings.json`, this file gets no explicit mode. Under a common `0002` umask it lands
+  at `0664` — readable by every account on the host.
+- **The request log — opt-in, and content-bearing when you ask for it.** `--log-file`
+  (`$HEADROOM_LOG_FILE`) writes one JSON object per request. By default those objects hold
+  metadata only. `--log-messages` (`$HEADROOM_LOG_MESSAGES`) adds full request and response
+  message content to the same file *and* serves it on the live feed endpoint. Its own help
+  text warns that it may log sensitive data; on this page's terms, it writes your prompts
+  and completions to disk unencrypted, for as long as you keep the file.
+
+Both `HEADROOM_LOG_FILE` and `HEADROOM_LOG_MESSAGES` are in the settings registry, which
+means full message logging can be switched on — and its destination chosen — from the
+settings surface described under
+[Local control surfaces](#local-control-surfaces). `--stateless` suppresses the request
+log entirely (`log_file` is forced to `None`); it does not suppress `proxy.log`.
 
 ### Memory store paths
 
@@ -118,8 +175,9 @@ memory only.
 
 ## Local control surfaces
 
-The proxy exposes HTTP routes beyond the provider-shaped ones. All of the sensitive
-routes — `/admin/*`, `/debug/*`, `/stats/reset`, and the content-bearing `/v1/retrieve*`
+The proxy exposes HTTP routes beyond the provider-shaped ones. The sensitive ones —
+`/admin/*`, `/debug/*`, `/cache/clear`, `/stats/reset`, `/settings*`, `/transformations/feed`,
+`/v1/feedback*`, `/v1/telemetry*`, `/v1/toin/*`, and the content-bearing `/v1/retrieve*`
 family — are guarded by `require_loopback`, which enforces two independent conditions:
 
 1. The client address must be a loopback IP.
@@ -130,6 +188,31 @@ attacker-controlled hostname to `127.0.0.1` so the browser connects locally whil
 page origin stays remote. The IP check alone passes in that scenario; the `Host:` check
 does not. Guarded routes return **404**, not 403, so they are indistinguishable from
 absent routes to a scanner.
+
+A third gate, `require_same_origin`, sits on the mutating routes (`/admin/runtime-env`,
+`/cache/clear`, `/stats/reset`, `/settings`, `/settings/apply`). It defends a different
+attack: a plain CSRF, where a remote page's script posts a "simple" request straight at
+`http://127.0.0.1:<port>`. Both loopback gates pass in that case — the browser really is
+connecting to loopback and really does send a loopback `Host:` — but the `Origin:` header
+still names the attacker's page, and that is what this gate rejects. Requests with no
+`Origin` at all (CLI tools, curl) pass, which is the intended behaviour for a
+locally-driven surface.
+
+### The two CIDR allowlists
+
+Two environment variables can extend this beyond loopback. Both default to empty, so the
+default posture is the one described above — but an operator who sets either has changed
+the boundary and should say so in their own threat model.
+
+| Variable | Effect |
+|---|---|
+| `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS` | Peers in these CIDRs reach `/settings`, `/settings/schema`, `/dashboard/settings`, and the settings *writes* **without being on loopback** — and additionally see the fields `/stats` and `/stats-lifetime` otherwise withhold from network callers (per-request ids, providers, models, errors, project names, backend config). Added so the settings UI works behind a reverse proxy. |
+| `HEADROOM_PROXY_TRUSTED_GATEWAY_CIDRS` | Peers in these CIDRs are trusted to supply `X-Forwarded-*` headers, which is what makes the client IP the *forwarded* one rather than the gateway's. |
+
+Both parse strictly: a malformed CIDR raises at startup rather than silently emptying the
+allowlist. Note the interaction — trusting a gateway for `X-Forwarded-For` means the
+dashboard CIDR check is applied to a client-supplied header, so the two must be configured
+together and pointed at infrastructure you control.
 
 ### What `POST /admin/runtime-env` can change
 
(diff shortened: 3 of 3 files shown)


JerrettDavis (changes requested):
This security model is materially inaccurate against current `main` in two operator-facing places:

1. The Network egress section says anonymous third-party telemetry was removed and that `HEADROOM_TELEMETRY` is local-only. While the latter is true, `headroom/telemetry/beacon.py` now defines a separate `HEADROOM_BEACON` that is **enabled by default**, and `headroom/telemetry/session.py` POSTs session summaries to `https://headroom-beacon.headroom-beacon.workers.dev/v1/logs`. The page omits both that destination and its opt-out controls (`HEADROOM_BEACON=off`, `DO_NOT_TRACK=1`, or offline mode), then makes the stronger false claim that nothing is sent off-host. A security review relying on this page would therefore miss default network egress. Please document the beacon separately from local telemetry, including its default, payload class, destination, and disable controls.

2. Both `SECURITY.md` and the new page name `$HEADROOM_COPILOT_AUTH_PATH`, but the implementation reads `HEADROOM_COPILOT_AUTH_FILE` in `headroom/copilot_auth.py::headroom_copilot_auth_path()`. As written, an operator following the security guidance cannot relocate the plaintext refresh token. Please use the implemented variable and add a doc/source consistency check if practical.

Please also revalidate the opening assertion that every data-at-rest path is workspace-relative: the active memory CLI can select a project-local `{cwd}/.headroom/memory.db` or an explicit `--memory-db-path`, so the table should describe those alternatives instead of implying the global workspace is exhaustive. After updating from current main, a focused source-backed pass over every path, default, and egress statement would keep this page suitable for enterprise security review.

keshavlingala:
Thanks — all three were valid, and two of them were drift I should have caught: the branch was 54 commits behind `main`. Rebased onto `d0a86d40` and pushed as `0db5577a`. Point by point:

### 1. The beacon

Confirmed and fixed. `HEADROOM_BEACON` is opt-*out* (`beacon.py:56`, `BEACON_DEFAULT_ON = True`) and posts to `https://headroom-beacon.headroom-beacon.workers.dev/v1/logs` (`session.py:75`). The page's "telemetry was removed / nothing off-host" claim was true at my branch point and false on current `main` — you're right that a security review relying on it would have missed default egress.

There is now a dedicated `The beacon is on by default` section documenting the destination, the default, the payload class, and the disable controls, kept separate from `HEADROOM_TELEMETRY` as you asked. `HEADROOM_TELEMETRY` gets its own subsection stating that the two switches are independent in both directions.

On payload class, from `_Session.payload()` and `resource_attributes()`: cumulative session counters, a random session id and seq, token totals and derived ratios, transform/skip tallies, plus a random per-install UUID, version, OS type, arch, and — when known — install mode and stack. Provider and model names are included; prompts, completions, file contents, and paths are not.

Two things I added beyond what you flagged, both of which seem load-bearing for an operator writing policy:

- **The beacon is fail-open while `HEADROOM_TELEMETRY` is fail-closed.** `beacon.py:76-79` says so explicitly — an unrecognised `HEADROOM_BEACON` value uploads. Anyone disabling it through configuration management needs to assert one of the exact off-values rather than assume a typo fails safe.
- **`HEADROOM_TELEMETRY_ENDPOINT` redirects the destination** (`session.py:601`), so terminating reports at a receiver you control is an option alongside blocking them.

### 2. `HEADROOM_COPILOT_AUTH_FILE`

Straightforwardly my error, not drift — `copilot_auth.py:143` reads `HEADROOM_COPILOT_AUTH_FILE` and has for a while. Fixed in both `SECURITY.md` and the page.

On the consistency check: I'd rather not grow a docs-only PR into tooling, but I agree this bug argues for it and I'm happy to open a separate issue.

### 3. Data-at-rest paths

Confirmed, and it's broader than the page implied. `--memory-storage` defaults to `project` (`cli/proxy.py:682`), and an unset `--memory-db-path` resolves to `{cwd}/.headroom/memory.db` (`server.py:1219-1225`) — so the default is project-local, not workspace-relative.

The preamble now says plainly that the workspace is not exhaustive, the `Path` column states where each item actually resolves, and a new `Memory store paths` subsection covers the three partitioning modes, the `HEADROOM_MEMORY_DB_PATH` env var, and the per-request scoping headers. I also added the `install_id` row (`<config dir>/install_id`, `0600`), which the beacon work surfaced.

### One structural change worth your attention

The rebase pulled in #2784 ("one documentation site, not two"), which **deleted `mkdocs.yml`**. Since `.github/workflows/docs.yml` now states that `wiki/` is unpublished markdown that nothing builds, a new page there would never have been published — which defeats the purpose of a page meant to stand up to a security review.

So the page moved to `docs/content/docs/security-model.mdx` with a `Security` entry in `meta.json`, and the `SECURITY.md` links now point at `headroom-docs.vercel.app/docs/security-model`. Happy to reconsider if you'd rather it live somewhere else. I also added a beacon bullet to the `SECURITY.md` summary, since that section claims to cover what leaves the host and would otherwise have reproduced the original bug in a new place.

### Re-verification

I redid the source-backed pass over every path, default, and egress statement against post-rebase `main`. Unchanged and still accurate: the loopback guard (both gates, 404-not-403), all seven `RUNTIME_ENV_KNOBS`, the CCR TTL / backend selection / silent-fallback behaviour, and the `0600` permissions.

`npm run build` in `docs/` compiles clean — 161 pages, no errors or warnings, `security-model` renders and appears in the sitemap, and all internal anchors resolve.

Still documented from code paths rather than exercised end-to-end, as in the original description: the `huggingface.co` and `app.headroomlabs.ai` rows.

JerrettDavis (changes requested):
The substantive security corrections from the prior review are now present and source-backed. Two freshness issues remain before this can serve as the enterprise-facing policy it claims to be: (1) `SECURITY.md` labels `0.33.x` as the latest supported version, but current main is already 0.34.0 and the generated release PR stages 0.35.0. Please align this with the project’s actual support policy (ideally wording that does not become false every release, if only latest is supported). (2) The PR body still says the new page is `wiki/security-model.md`, mentions `mkdocs.yml`, and reports `mkdocs build`; the current diff instead adds `docs/content/docs/security-model.mdx` and `docs/content/docs/meta.json`. Please update Changes Made, Testing/Test Output, and Real Behavior Proof to describe and validate the actual Next.js/Fumadocs artifact. Reviewed at 0db5577a; CI is green and the earlier beacon/Copilot/memory-path blockers are otherwise resolved.

keshavlingala:
Both points addressed, and the "focused source-backed pass" you asked for turned up more than I expected. Rebased onto `7ef736fb` (the branch was 183 behind) and pushed as `2ef83d35`.

### 1. Supported versions

Agreed that re-pinning a minor every release is a step nobody remembers, so I went with the policy rather than a number:

> | Latest release on PyPI | :white_check_mark: |
> | Any earlier release | :x: |
>
> Headroom releases frequently and only the most recent release is supported. Security fixes ship in a new release rather than being backported to earlier ones, so upgrading is the remediation path.

If there is in fact a supported line older than latest, tell me and I will write that instead — I inferred "latest only" from the absence of any maintenance branch and from release-please cutting straight off `main`.

### 2. PR body

Rewritten. Changes Made, Testing, and Real Behavior Proof now describe `docs/content/docs/security-model.mdx` + `meta.json` and validate with `npm run build` (167 pages, zero warnings, `security-model` in the sitemap, all 17 internal anchors resolving against the generated HTML). Every `wiki/` and `mkdocs` reference is gone, and I added a note recording *why* the page moved so the next reader does not have to reconstruct it from the thread.

While there: re-confirmed the site target against current `main` rather than assuming it survived the rebase — `.github/workflows/docs.yml` and `docs/app/layout.tsx` both still name the Vercel app as the one documentation site, so the `headroom-docs.vercel.app/docs/security-model` links are correct.

### The re-verification pass, and one claim it contradicts

Unchanged and re-confirmed against post-rebase `main`: the beacon default and destination, `HEADROOM_COPILOT_AUTH_FILE`, all seven `RUNTIME_ENV_KNOBS`, the memory-path defaults, the CCR TTL and backend selection, the 404-not-403 loopback guard, and the `0600` permissions.

Five things had drifted. Four are gaps; the last one is a claim the page made that is now false.

**a. `settings.json` was missing from the data-at-rest table** — and it is not a minor omission. The dashboard settings GUI persists 59 knobs, two of them marked secret: `ANTHROPIC_TARGET_API_HEADERS` and `OPENAI_TARGET_API_HEADERS`, free-form header maps whose own help text gives `{"Api-Key": "..."}` as the example. Proof in the PR body: the sentinel key is on disk in plaintext while `GET /settings` returns `••••••`, so the masking is a display control, not an at-rest one. It is also a second source for the process environment (`apply_to_environ`, `setdefault`), which matters to anyone asserting fleet policy by inspecting exported variables.

**b. `logs/proxy.log` was missing too.** Always on, 10 MB × 5, and it carries the admin audit stream because the audit logger is a child of `headroom` and writes no file of its own. It also gets no explicit mode — `0664` under a common umask, against `0600` for every other sensitive file Headroom writes. The opt-in request log is now documented alongside it, including that `--log-messages` writes prompts and completions to disk.

**c. The "two gates" framing was incomplete.** There is a third — `require_same_origin` on the mutating routes — and it defends a genuinely different attack (plain CSRF at a known loopback port, where both loopback checks pass and only `Origin` gives it away). Also documented: `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS`, which admits non-loopback callers to the settings routes and to the sensitive fields of `/stats`, and `HEADROOM_PROXY_TRUSTED_GATEWAY_CIDRS`, which governs who may supply the forwarded headers that check then reads. Both default empty, so default posture is unchanged — but the two have to be configured together, and an operator who sets one has moved the boundary.

**d. The page's claim that the upstream URL cannot be changed over HTTP is no longer true.** This is the one I would most want a second opinion on. `/admin/upstream` is still read-only, and the comment there still explains that this is deliberate — but `ANTHROPIC_TARGET_API_URL` and `OPENAI_TARGET_API_URL` are in the settings registry, and `POST /settings/apply` persists *and restarts the proxy* to apply. So a caller on the settings surface can repoint credential-bearing traffic and have it take effect, which is the exact outcome the `/admin/upstream` design note says it exists to prevent. `HEADROOM_LOG_MESSAGES` is reachable the same way.

I have documented it as current behavior rather than filing it, because I cannot tell from the outside whether the settings surface is intended to be equivalent in privilege to `/admin/*` — under the local-trust model both are loopback-gated and it is arguably consistent. But the CIDR allowlist widens the settings surface and not the admin one, and that asymmetry looks unintentional. If you read it as a defect I will open an issue.

**e. Beacon payload re-enumerated** after #2796, #2815, #2818 and #2853. Per-strategy breakdown, all-layers rates, and failures-by-status are new. The no-content claim holds and now says why: the tally keys are internal transform slugs truncated at the first colon, and sources are one of two literals, so no free text reaches the payload.

Also checked and *not* changed: #2604's sliding idle-window TTL with the 8× max-lifetime ceiling is in the Rust core backends. The documented Python path is still a fixed `created_at + ttl` window, so the page's lazy-purge wording stands as written.

### On the consistency check

I said last round I would rather not grow a docs-only PR into tooling but that the bug argued for it. I have run the check by hand here — every `HEADROOM_*` / `ANTHROPIC_*` / `OPENAI_*` / `DO_NOT_TRACK` name in the page and in `SECURITY.md` grepped back to a read site in `headroom/` or `crates/`, 30 names, none missing — and I will open the issue for automating it as a follow-up rather than folding it in.

keshavlingala:
Follow-up issue for the doc/source consistency check is open: #3103.

JerrettDavis (approved):
The two prior review rounds are resolved on the current head. The security model now separately documents default-on beacon egress and opt-outs, uses HEADROOM_COPILOT_AUTH_FILE, inventories project/explicit memory paths plus settings and operational logs, describes all three local-control gates and CIDR boundary widening, and validates the actual Fumadocs page rather than the retired wiki. SECURITY.md now states a durable latest-release-only support policy instead of a stale version literal. I also added the missing docs-only Runtime Rollout Safety section. Diff whitespace is clean; no remaining factual blocker found.

keshavlingala:
Fixed the two `Real Behavior Proof` field labels — they were pluralized (`Exact commands / steps`, `Observed results`), and `scripts/pr-governance.py` matches them verbatim, so the check was failing on the two `s`es rather than on missing content. Governance is green now.

@JerrettDavis thanks for the approve. One label is stale from the earlier review rounds — `status: needs author action` cleared on its own with the body edit, but `status: code changes requested` isn't managed by the workflow and I don't have label permission. Could someone drop it? No conflicts against `main`, all checks passing.

keshavlingala:
Gentle nudge on this one — it's been approved since Aug 20, all checks are green, and there are no conflicts against `main`. Two small things are keeping it from going in:

1. **Stale label.** `status: code changes requested` is left over from the earlier review rounds and was cleared by the Aug 20 approval. It isn't managed by the governance workflow and I don't have label permission — could a maintainer drop it? Merge state is currently `BLOCKED`, and this looks like the remaining piece.

2. **Branch age.** The branch is now 62 commits behind `main` (`d12ea501`). It still merges cleanly and the change is docs-only, but given that drift against `main` was the substance of both earlier review rounds, I'm happy to rebase and re-run the source-backed verification pass against current `main` before merge — just say the word and I'll push it.

Otherwise this is ready to go from my side. @JerrettDavis thanks again for the three rounds of review on it.

keshavlingala:
Rebased onto `d12ea501` (the branch was 62 behind) and pushed as `dac4a25b`. The rebase itself was clean — docs-only, no conflicts — but the verification pass that came with it turned up drift worth flagging, so this is not a no-op force-push.

### What re-verified unchanged

- All documented env names resolve to a read site in `headroom/` or `crates/` (now 37, up from 30).
- `RUNTIME_ENV_KNOBS` is still exactly the seven the page tables.
- The settings registry is still 59 fields, with exactly two `secret=True` header maps.
- `/admin/upstream` is still read-only + loopback; the `/settings` and `/settings/apply` gates are unchanged.

### What had drifted

Four security-relevant modules landed after my last verification pass, and the page was silent on all four. One of them made a standing claim inaccurate:

**`proxy/identity.py` (WEB-02)** — the page described `x-headroom-user-id` as a per-request scoping header without qualification. `resolve_memory_identity` now honors it **only for loopback callers**; every other caller is bound to a hash of `HEADROOM_PROXY_TOKEN` or the server's OS user, and the check fails closed when peer metadata is absent. It's wired into all three handlers. Corrected, and the `user` row in the partition table now points at the note instead of naming the header.

**`proxy/upstream_guard.py` (WEB-01)** — `x-headroom-base-url` lets a client pick the destination per request, which the page didn't mention existed. Now documented: the private/loopback/link-local rejection, the NAT64 unwrap, that a rejected override is ignored rather than fatal, and `HEADROOM_ALLOWED_BASE_URLS` with its bare-host vs. exact-origin distinction.

**`proxy/upstream_trust.py`** — this one *narrows* a risk the page raises. The two secret `*_extra_headers` maps now travel only to operator-designated hosts (`HEADROOM_UPSTREAM_ALLOWED_HOSTS`), matched on the parsed hostname rather than the URL string. Documented next to the guard, with a note that it bounds the blast radius without changing how those values are stored.

**`proxy/ssl_context.py`** — added the CA-bundle semantics: `SSL_CERT_FILE`/`REQUESTS_CA_BUNDLE` replace the trust store, `NODE_EXTRA_CA_CERTS` adds to it.

I deliberately did **not** assert anything about whether a client's own `Authorization` reaches an overridden upstream on every path. There's an `OPENAI_API_KEY` env fallback on the WebSocket path (`handlers/openai.py`) that I couldn't trace to a confident conclusion, and this page's standard is source-backed or omitted. Flagging it here in case it's worth a separate look — it may be nothing.

### Build

`npm run build` clean: 167 static pages, zero warnings, zero errors. All 18 internal anchors on the page resolve against the generated HTML (17 before, +1 for the new cross-reference).

@JerrettDavis sorry to spend your approval — the identity claim was inaccurate as written and I didn't want to merge it that way. The diff since your approve is one commit, `dac4a25b`, docs-only. The stale `status: code changes requested` label is still there if a maintainer can drop it.

JerrettDavis (approved):
Re-reviewed the post-approval commit at dac4a25b. The added security-model sections accurately cover the client-selectable upstream boundary, SSRF and credential-forwarding guards, CA-bundle precedence/semantics, and the loopback-only identity hint. I cross-checked the claims against the current upstream guard, upstream trust, identity, and SSL context implementations; the documented defaults and escape hatches match the implementation. The docs-only checks are green. Approval restored for final human review.

PR #1848, opened 2026-07-06
fix(litellm): surface Bedrock cache token usage in non-streaming responses

dspv:
## Description

Non-streaming `complete_message()` builds the Anthropic-shape usage from `prompt_tokens`/`completion_tokens` only. LiteLLM's `prompt_tokens` includes cached tokens, so when Bedrock prompt caching is active a non-streaming client sees `input_tokens` equal to the full prompt and no cache fields. That looks identical to the cache being broken (#1345), and the savings tracker never credits the hits. The streaming and OpenAI paths already map these fields.

Related: #1390 — that PR makes the markers reach Bedrock; this one makes the result visible in non-streaming responses.

Closes # (contributes to #1345 together with #1390; not closing it alone)

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Extract `_anthropic_usage_from_litellm()` in `headroom/backends/litellm.py`: maps `cache_read_input_tokens` / `cache_creation_input_tokens` (with `prompt_tokens_details` fallback) into the Anthropic-shape usage and reports `input_tokens` without the cached portion, matching what Anthropic returns.
- Use it in `complete_message()` instead of the inline `prompt_tokens`/`completion_tokens` dict.
- Add `tests/test_litellm_nonstream_cache_usage.py` (5 cases: plain usage, cache read, cache write, `prompt_tokens_details` fallback, negative clamp).

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ pytest tests/test_litellm_nonstream_cache_usage.py -q
5 passed, 1 warning in 2.13s

$ ruff check headroom/backends/litellm.py tests/test_litellm_nonstream_cache_usage.py
All checks passed!
```

mypy not run locally: my environment fails on unrelated numpy stubs (`numpy/__init__.pyi: Type statement is only supported in Python 3.12+`); relying on CI for the mypy gate.

## Real Behavior Proof

[...]

PR #2688, opened 2026-07-31
fix(backends/litellm): align streaming usage tokens with the non-streaming path

abhay-codes07:
## Description

On the LiteLLM/Bedrock streaming path, the terminal `message_delta` usage does not match what the non-streaming path reports for the same response. Two token buckets are wrong, both because the streaming code does not reuse the logic the non-streaming path already has in `_anthropic_usage_from_litellm`. This PR fixes both so the streaming usage object is a conformant Anthropic usage object.

**1. `input_tokens` double-counts the cached prefix.** LiteLLM's `prompt_tokens` is the total prompt size, inclusive of cache-read and cache-write tokens (`AmazonConverseConfig._transform_usage` takes Bedrock's raw `inputTokens` and adds `cacheReadInputTokens + cacheWriteInputTokens`, on both the streaming and non-streaming paths). Anthropic's `input_tokens` must exclude both, since the response also carries `cache_read_input_tokens` and `cache_creation_input_tokens` and clients treat the three buckets as disjoint (they sum them for the total prompt and price each differently: cache-read 0.1x, cache-write 1.25x). The non-streaming path subtracts (added in #1848 for #1345); the streaming path added in #2144 emitted the inclusive `prompt_tokens` straight through. So on every cached streaming turn the client counts the cached prefix twice, once at the discounted cache rate and again inside `input_tokens` at the full input rate.

**2. `output_tokens` reports an SSE chunk count, not real tokens.** The streaming loop increments `output_tokens` once per `content_block_delta`. That is a chunk count, and it undercounts the true output token total several-fold because each delta carries multiple tokens. The trailing usage chunk (requested via `stream_options.include_usage`) already carries the provider's real `completion_tokens`, but `stream_message` only read prompt and cache tokens from it and discarded `completion_tokens`. The non-streaming path uses `completion_tokens` directly. This undercount reaches both the client and Headroom's own accounting: the Bedrock streamer copies `output_tokens` into the `RequestOutcome`, which feeds the `RequestLog`, the `PERF tok_out` field, and the verbosity savings estimate.

Both fixes mirror `_anthropic_usage_from_litellm`: subtract the cache tokens from `input_tokens`, and prefer the real `completion_tokens` for `output_tokens` (falling back to the delta count only when no usage chunk arrives).

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/backends/litellm.py` (`stream_message`): terminal `message_delta` `input_tokens` is now `max(prompt_tokens - cache_read - cache_write, 0)` instead of the inclusive `prompt_tokens`; `output_tokens` now uses the trailing usage chunk's `completion_tokens` when present, falling back to the per-delta count only when no usage chunk arrives. Both match `_anthropic_usage_from_litellm` on the non-streaming path.
- `tests/test_bedrock_tool_result_cache_and_streaming_stats.py`: `test_terminal_message_delta_carries_real_cache_stats` updated (its `input_tokens == 1500` assertion encoded the double-count: 1500 inclusive with 1200 read + 300 write left 0 uncached) to a 1600-token prompt with 100 uncached tokens and a real-`completion_tokens` output assertion; added `test_terminal_message_delta_input_tokens_excludes_cache_tokens` (disjoint-bucket invariant), `test_output_tokens_uses_real_completion_tokens_not_delta_count`, and `test_output_tokens_falls_back_to_delta_count_without_usage_chunk`. The `_mock_stream_with_final_usage` helper now carries `completion_tokens` to mirror the real trailing chunk.

## Testing

### Test Output

```text
# Fail-before (source fix stashed, updated tests kept):
  assert final_usage["input_tokens"] == 100   -> assert 1600 == 100
  assert final_usage["input_tokens"] == 500   -> assert 5000 == 500
  assert final_usage["output_tokens"] == 50    -> assert 1 == 50
  assert final_usage["output_tokens"] == 137   -> assert 1 == 137

# Pass-after (fix applied):
tests/test_bedrock_tool_result_cache_and_streaming_stats.py ................
16 passed in 6.21s

# Related suites (streaming cache metrics, bedrock streaming input tokens,
# nonstream cache usage, litellm cache stats, backend bugs, streaming resilience):
all passed

```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.11, litellm installed in the project venv, pytest 9.1.1 with pytest-asyncio 1.4.0 (asyncio_mode=auto per pyproject).
- Exact command / steps: verified litellm inclusivity and the completion_tokens source by reading `litellm/llms/bedrock/chat/converse_transformation.py` (`_transform_usage` does `input_tokens += cacheReadInputTokens; input_tokens += cacheWriteInputTokens; prompt_tokens=input_tokens; completion_tokens=output_tokens`) and `invoke_handler.py` (the streaming trailing chunk calls the same `_transform_usage`); then fail-before with `git stash push headroom/backends/litellm.py` and `python -m pytest tests/test_bedrock_tool_result_cache_and_streaming_stats.py` (the four value assertions fail, emitting inclusive 1600 / 5000 for input and the delta count 1 for output), and pass-after with `git stash pop` and rerunning the file (16 passed).
- Observed result: on the fixed streaming path the terminal `message_delta.usage` reports `input_tokens` as the genuinely-uncached count with cache tokens only in `cache_read_input_tokens` and `cache_creation_input_tokens` (prompt_tokens=1600, read=1200, write=300 gives input_tokens=100 and 100+1200+300=1600, versus the old inclusive 1600 that a client sums with the cache fields to 3100), and reports `output_tokens` as the provider's `completion_tokens` (50 or 137 in the tests) rather than the single content-delta count. Both now match the non-streaming response shape.
- Not tested: a live Bedrock Converse streaming call against real AWS credentials (no Bedrock access in this environment). The litellm usage shape is exercised through a mock that mirrors `stream_options.include_usage`'s trailing chunk, and the litellm inclusivity and completion_tokens fields were verified against the installed litellm source rather than assumed.

## Review Readiness

## Checklist

## Additional Notes

Docs and manual testing are N/A: this is an internal usage-mapping correctness fix with no API surface or documented behavior change. The streaming path already read the top-level `cache_read_input_tokens` / `cache_creation_input_tokens` attrs, so the input subtraction uses the same values the response emits; the OpenAI nested `prompt_tokens_details` dialect on streaming is a separate gap and out of scope here.


JerrettDavis (approved):
The streaming Anthropic usage now mirrors the established buffered conversion: cached buckets are disjoint from input_tokens, real completion_tokens win over chunk counts, and the no-usage fallback remains intact. Reviewed and validated on the exact current-main tree; 16 focused tests and Ruff pass.

PR #2689, opened 2026-07-31
fix(backends/litellm): map Anthropic tool_choice none to OpenAI none

abhay-codes07:
## Description

`_convert_tool_choice` in the LiteLLM backend converts an Anthropic `tool_choice` into the OpenAI shape. It handled `{"type": "auto"}`, `{"type": "any"}` and `{"type": "tool", "name": ...}`, but there was no branch for Anthropic's `{"type": "none"}`, so it fell through to the `return "auto"` default at the end of the function.

That inverts the instruction. A client that sends `tool_choice: {"type": "none"}` is saying "do not use any tool this turn" (tools stay defined, but the model must not call one). When such a request is routed through the LiteLLM backend (Bedrock, Vertex, OpenRouter, and the other providers this backend serves), the choice was rewritten to `"auto"`, which tells the model it may call a tool. So the model could invoke a tool the client explicitly forbade.

The string form already passed through unchanged (`_convert_tool_choice("none") == "none"`, covered by an existing test); only the Anthropic dict form, which is what Anthropic clients actually send, was mishandled. OpenAI's equivalent is the string `"none"`, so this maps `{"type": "none"}` to `"none"`.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/backends/litellm.py` (`_convert_tool_choice`): added a `{"type": "none"}` to `"none"` branch and updated the docstring shapes.
- `tests/test_backend_bugs.py`: added `test_none_dict_not_inverted_to_auto` asserting the dict form maps to `"none"` (it previously returned `"auto"`).

## Testing

### Test Output

```text
tests/test_backend_bugs.py::TestConvertToolChoice::test_none_dict_not_inverted_to_auto FAILED
  assert _convert_tool_choice({"type": "none"}) == "none"
  AssertionError: assert 'auto' == 'none'

# Pass-after (fix applied):
tests/test_backend_bugs.py ...........................................
43 passed in 4.83s

```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.11, litellm installed in the project venv, pytest 9.1.1.
- Exact command / steps: fail-before with `git stash push headroom/backends/litellm.py` and `python -m pytest tests/test_backend_bugs.py -k TestConvertToolChoice` (the new assertion fails, `_convert_tool_choice({"type": "none"})` returns `"auto"`), then pass-after with `git stash pop` and rerunning the file (43 passed).
- Observed result: `_convert_tool_choice({"type": "none"})` now returns `"none"` instead of `"auto"`, so an Anthropic client that forbids tool use keeps that instruction when the request is routed through the LiteLLM backend. The existing auto / any / tool / string cases are unchanged.
- Not tested: a live end-to-end call to a provider with a real "none" tool_choice (no provider credentials in this environment). The conversion is a pure function verified directly, and the downstream provider is responsible for honoring OpenAI's `"none"`.

## Review Readiness

## Checklist

## Additional Notes

Docs and manual testing are N/A: this is a pure conversion-correctness fix with no API surface change. The fallback `return "auto"` is kept for genuinely unknown future types; only the known "none" type is now mapped explicitly.


JerrettDavis (approved):
The Anthropic dict form of tool_choice none now maps to the OpenAI none sentinel instead of falling through to auto, preserving an explicit no-tool instruction while leaving existing mappings unchanged. Reviewed and validated on the exact current-main tree; 43 backend tests and Ruff pass.

PR #2690, opened 2026-07-31
fix(backends/litellm): report requested model in OpenAI streaming chunks

abhay-codes07:
## Description

`stream_openai_message` in the LiteLLM backend forwards each LiteLLM chunk to the client with `chunk.model_dump()` untouched, so the `model` field carries the LiteLLM-mapped provider slug the backend was called with (for example `openrouter/qwen3`, or `bedrock/us.anthropic.claude-...`) rather than the model the client requested.

The non-streaming sibling `send_openai_message` already rewrites the response `model` to `original_model` (the requested name). So the two paths disagree on the same request: an OpenAI-format client that keys cost or telemetry on the `model` field sees the requested name for a buffered reply, but an unrecognized provider slug for every streamed chunk. A client that prices or attributes by model name then reads that slug as an unknown model on streamed turns, while real OpenAI streaming reports the requested model in every chunk.

The fix rewrites the chunk `model` to `original_model` before serializing, matching the non-streaming path and real OpenAI streaming behavior.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/backends/litellm.py` (`stream_openai_message`): set `chunk_dict["model"] = original_model` (when the chunk carries a model field) before yielding, mirroring `send_openai_message`.
- `tests/test_litellm_openai_passthrough.py`: added `test_streaming_chunks_report_requested_model_not_mapped_slug`, which streams a chunk tagged `openrouter/qwen3` and asserts the client sees `qwen3`.

## Testing

### Test Output

```text
tests/test_litellm_openai_passthrough.py::test_streaming_chunks_report_requested_model_not_mapped_slug FAILED
  assert payload["model"] == "qwen3"
  AssertionError: assert 'openrouter/qwen3' == 'qwen3'

# Pass-after (fix applied):
tests/test_litellm_openai_passthrough.py .....
5 passed in 4.94s

# Related streaming suites:
tests/test_backend_anyllm.py tests/test_backend_streaming_cache_metrics.py
20 passed

```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.11, litellm installed in the project venv, pytest 9.1.1 with pytest-asyncio 1.4.0 (asyncio_mode=auto per pyproject).
- Exact command / steps: confirmed the non-streaming path rewrites the model by reading `send_openai_message` (`"model": original_model` in the response dict), then fail-before with `git stash push headroom/backends/litellm.py` and `python -m pytest tests/test_litellm_openai_passthrough.py -k requested_model` (the chunk reaches the client as `openrouter/qwen3`), and pass-after with `git stash pop` and rerunning the file (5 passed).
- Observed result: streamed OpenAI chunks now report the requested model (`qwen3`) rather than the LiteLLM-mapped slug (`openrouter/qwen3`), so the streaming and non-streaming responses agree on the model field.
- Not tested: a live streamed call through a real provider (no provider credentials in this environment). The mapped-slug behavior is reproduced with a mock chunk that mirrors what LiteLLM tags chunks with, and the non-streaming rewrite it matches is existing, tested code.

## Review Readiness

## Checklist

## Additional Notes

Docs and manual testing are N/A: this aligns the streaming model field with the already-correct non-streaming path, no API surface change. The rewrite is guarded on the chunk actually carrying a `model` field, so chunks that omit it (if any) are left untouched.


JerrettDavis (approved):
Streaming OpenAI chunks now expose the model requested by the client, matching the buffered adapter and avoiding provider-slug leakage into client accounting. Reviewed and validated on the exact current-main tree; 5 focused passthrough tests and Ruff pass.

PR #2696, opened 2026-07-31
fix(openai): skip Responses tool-search deferral for clients that cannot execute it

### LINKED ISSUE #2660: [BUG] OpenCode receives unavailable tool_search from OpenAI Responses deferral
## Summary

Headroom's OpenAI Responses tool-search deferral can inject a `tool_search` tool into an OpenCode request, but OpenCode does not register or handle that injected tool locally. When the model selects it, OpenCode records an unavailable-tool error instead of completing the turn.

## Environment

- Headroom `0.32.1`
- Python `3.14`
- macOS
- OpenCode `1.18.9`
- `headroom-opencode` plugin, proxy at `http://127.0.0.1:8787`
- OpenAI-compatible provider
- Affected models observed: `gpt-5.6-luna`, `gpt-5.6-sol`, `gpt-5.6-terra`

## Evidence

OpenCode records errors like:

```text
Model tried to call unavailable tool 'tool_search'
```

The local OpenCode tool registry does not contain `tool_search`.

Headroom's raw performance records for the last 24 hours contained 334 occurrences of:

```text
openai:responses:tool_search_deferral
```

Breakdown:

- `gpt-5.6-luna`: 226
- `gpt-5.6-sol`: 100
- `gpt-5.6-terra`: 8

The installed source shows `headroom/proxy/handlers/openai.py` calling `inject_tool_search_deferral_openai(...)` for Responses requests. The helper in `headroom/proxy/helpers.py` prepends `{"type": "tool_search"}` when the model looks like `gpt-5.4+` and the request contains at least 12 tools.

## Reproduction

1. Run Headroom proxy with the OpenCode plugin and an OpenAI-compatible Responses provider.
2. Use a GPT-5.4+ model with OpenCode's normal tool surface.
3. Allow the proxy to apply `openai:responses:tool_search_deferral`.
4. Continue a multi-turn tool session until the model selects the injected tool.

Actual:

```text
Model tried to call unavailable tool 'tool_search'
```

Expected: OpenCode either handles the provider-native tool-search flow, or Headroom does not inject the tool for clients that cannot handle it.

## Root Cause

The proxy changes the outbound Responses `tools` array, but OpenCode's local tool registry and tool-call parser are unaware of the injected tool. The model can therefore return a `tool_search` call that OpenCode rejects as unknown.

`HEADROOM_TOOL_SEARCH=0` does not gate this OpenAI Responses path. The current model regex override can disable it with `HEADROOM_OPENAI_TOOL_SEARCH_MODELS=^$`, but this workaround is not obvious from the CLI help.

## Workaround

Start the proxy with:

```bash
HEADROOM_OPENAI_TOOL_SEARCH_MODELS='^$' headroom proxy ...
```

Then restart the persistent proxy.

## Suggested Fix

- Add an explicit OpenAI tool-search opt-in/opt-out setting.
- Do not inject `{"type":"tool_search"}` unless the client declares support for handling the resulting tool-search events.
- Document how to disable the OpenAI Responses deferral path.
- Consider recognizing OpenCode as a client that requires local tool registration or implementing the complete OpenCode-compatible flow.

No credentials or private request payloads are included here.

rodboev:
## Description

OpenCode rejects the proxy-injected Responses `tool_search` tool because it resolves tool calls against its local registry. This PR now uses the shared client policy from current `main` and leaves OpenCode tools resident, alongside the existing Codex exclusion. Other clients retain tool-search deferral.

Closes #2660.

## Type of Change

- [ ] Documentation update

## Changes Made

- Add `opencode` to the shared exact-match unsupported-client set in `headroom.proxy.helpers`.
- Carry the already-classified `client` through native HTTP, WebSocket, and custom-base Responses paths.
- Preserve `main`'s compatibility loop, which retries only exact unsupported `client` or `timing` keyword errors and re-raises internal `TypeError`s.
- Add focused helper, compressor, HTTP, passthrough, and WebSocket coverage.

## Testing

- [x] Unit tests pass
- [x] Ruff check and format pass
- [x] New tests added
- [ ] Live OpenCode session tested

```text
uv run --extra dev pytest tests/test_openai_tool_search_deferral.py tests/test_proxy_openai.py -q
57 passed

uv run --extra dev ruff check headroom/proxy/handlers/openai.py headroom/proxy/helpers.py tests/test_openai_tool_search_deferral.py tests/test_proxy_openai.py
All checks passed
```

## Real Behavior Proof

The focused route tests classify OpenCode from both `User-Agent` and `X-Client`, verify its tools remain untouched, and verify the decision reaches all three Responses ingresses. Supported clients continue to receive deferral. Codex remains excluded by the policy already on `main`.

Not tested: a live OpenCode instance; the incompatibility itself remains based on the reporter's reproduction in #2660.

## Review Readiness

- [x] Updated from current upstream `main`
- [x] Merge conflicts resolved
- [x] Focused tests pass locally
- [x] Ready for human review

## Additional Notes

No user configuration or documentation change is required. Vercel authorization failures are external integration noise, not a source check.


rodboev:
Codecov caught the untested propagation arm in the Responses compressor compatibility fallback. I added a focused regression in `tests/test_proxy_openai.py` that makes an override accept the optional arguments and then raise its own `TypeError`, confirming `_compress_openai_responses_payload_in_executor` propagates that error instead of treating it as an old-signature mismatch and retrying. This is test coverage only; `headroom/proxy/handlers/openai.py` and request behavior stay unchanged.

JerrettDavis (changes requested):
The per-client gate is well targeted, but the legacy-signature fallback now catches any `TypeError` whose message contains `unexpected keyword argument`, not specifically rejection of the optional `timing` / `client_can_tool_search` arguments. A genuine bug inside `_compress_openai_responses_payload` can raise that same standard Python message (for example, an internal callee receiving an unexpected `foo=`); this wrapper will then silently retry the entire compression without the optional flags. That both masks the defect and can repeat shaping/compressor side effects, and the current internal-TypeError test only uses the message `internal compressor failure`, so it does not exercise the false-positive shape. Please avoid exception-text-wide retry: inspect the override signature before calling, or verify the rejected keyword is one of the two wrapper-supplied options and that the TypeError originates at the call boundary. Add a regression where the compressor body raises `TypeError("... unexpected keyword argument internal ...")` and prove it is propagated with one invocation. The branch is also currently conflicting with main, whose tool-search client gating has evolved; resolve that by preserving both Codex and OpenCode exclusions rather than choosing one during the merge.

JerrettDavis (approved):
Re-reviewed the refreshed head after the `main` merge. The OpenCode exclusion now uses the shared exact-match client policy, preserves the existing Codex safeguard, and keeps `main`’s narrow compatibility retry (only exact unsupported `client`/`timing` keywords; internal `TypeError`s still propagate). Focused HTTP, WebSocket, passthrough, helper, and compressor coverage passes locally (57 tests), and Ruff check/format are clean. Ready for final human review.

PR #2697, opened 2026-07-31
feat(proxy): route Gemini plugin traffic through native transforms

### LINKED ISSUE #2602: Format adapters for plugin-routed providers (Gemini first): compress, not just forward
## Summary

Traffic routed through the proxy by the OpenCode transport plugin (`x-headroom-base-url`) is forwarded verbatim when the wire format is neither Anthropic Messages nor OpenAI Chat/Responses. The proxy attributes and counts it, but applies no compression - `PERF ... tok_saved=0 transforms=none`.

Request: format adapters for plugin-routed providers, Gemini first.

## Current behavior (verified against headroom-ai 0.32.1)

- The transport plugin reroutes every provider inside OpenCode to the proxy, tagging the real upstream via `x-headroom-base-url`.
- Anthropic/OpenAI-format requests hit their handlers and get the full transform pipeline.
- Anything else (e.g. `POST /v1beta/models/gemini-2.5-flash:streamGenerateContent`, @ai-sdk/google's native format) lands in the catch-all route (`providers/proxy_routes.py` custom-base branch), which forwards bytes unmodified.

Reproduced end-to-end with opencode 1.18.5 + the standalone plugin bundle (#2601): Gemini traffic routes correctly to `generativelanguage.googleapis.com` and is logged with `client=opencode`, but `transforms=none`.

## Proposal

Per-format adapters that parse the foreign request schema onto the transform pipeline's internal representation and re-encode the (streaming) response, plus tokenizer + pricing entries for honest counts. Gemini (`contents` schema, `:streamGenerateContent` SSE) is the highest-value first target - it is the most common non-anthropic/openai provider in OpenCode configs.

Alternative worth weighing for Google specifically: translate the endpoint choice instead of the format - Google exposes an OpenAI-compatible endpoint (`/v1beta/openai/`), so the plugin (or proxy) could rewrite Gemini-native calls onto the OpenAI-compatible surface and reuse the existing OpenAI pipeline. Smaller lift, Google-only.

## Context

Headroom Desktop ships the transport plugin to OpenCode users (vendored until #2601 lands), so plugin-routed third-party traffic exists in the wild now; today it is visibility-only.

rodboev:
## Description

Gemini traffic that OpenCode routes through the proxy is forwarded without transforms, and to the wrong host. The transport plugin tags the real upstream with `x-headroom-base-url` and preserves the upstream pathname, so a native Gemini call arrives as `POST /v1beta/models/{model}:streamGenerateContent`. Route matching is path-based, so that request reaches `handle_gemini_stream_generate_content`; the route did not pass the tagged base to the handler. Direct native `:streamGenerateContent` also used a duplicate forwarding path that skipped transforms, while `:generateContent?alt=sse` already used the transform-capable handler.

`handle_gemini_generate_content` already accepts an `upstream_base_url`, runs the OpenAI pipeline over converted `contents`, and handles streaming by URL shape. `handle_gemini_count_tokens` already accepts one too. This adds an opt-in capability on the route spec, forwards the custom base at the route seam for the three Gemini model endpoints, and sends both direct and plugin-routed native streaming through the shared transform-capable handler. Native parts, incoming query parameters, and the SSE selector are preserved. Wire formats with no native handler keep the existing custom-base passthrough branch.

Closes #2602.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `ProviderHandlerRoute` gains `supports_custom_base_url`, defaulting to `False` so every existing handler route keeps its current behavior
- `GEMINI_HANDLER_ROUTES` opts in all three model endpoints: `:generateContent`, `:streamGenerateContent`, and `:countTokens`. Counting has to follow the same upstream as generating, or the client gets counts from a host that is not serving the request
- `_register_provider_handler_route` reads `x-headroom-base-url` and forwards it as `upstream_base_url` for routes that opted in
- `handle_gemini_stream_generate_content` delegates unconditionally to `handle_gemini_generate_content`, which runs the transform pipeline and builds the SSE URL from the request path for both direct and plugin-routed traffic
- regressions for direct and plugin production-composed routes, mixed native parts, raw query preservation with one `alt=sse`, and adjacent paths that stay on passthrough

## Testing

### Test Output

`uv run pytest tests/test_proxy/test_gemini_plugin_routed_streaming.py tests/test_provider_route_specs.py tests/test_tokenizer_count_offload.py -q`

```text
34 passed, 1 warning
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.13, no live Google credentials; `create_app` composed with a fake upstream
- Exact command / steps: `uv run pytest tests/test_proxy/test_gemini_plugin_routed_streaming.py tests/test_provider_route_specs.py tests/test_tokenizer_count_offload.py -q`, with production-composed `create_app` routes and fake upstream callbacks
- Observed result: direct and plugin-routed `:streamGenerateContent` requests reach the same transform pipeline, pass compressed text and preserved `functionCall` and `inlineData` parts to `_stream_response`, select the correct Google or tagged URL, retain raw query segments, add one `alt=sse`, and strip `x-headroom-*` headers. `:countTokens` follows the same tagged upstream, and adjacent custom-base routes remain passthrough.
- Not tested: live Google endpoint

## Review Readiness

## Checklist

## Additional Notes

Scope note: this covers Gemini's three model endpoints, including direct and plugin-routed `:streamGenerateContent`. Gemini's other routes still go to `generativelanguage.googleapis.com` regardless of the tag, and I left them alone deliberately rather than by oversight. The batch routes would need their handlers to accept an upstream base first, and the passthrough routes (`/v1beta/models` listing, `:embedContent`, `:batchEmbedContents`, `/v1beta/cachedContents`) resolve their target through a different registrar. `cachedContents` remains separate because a tagged client could create a cache entry at Google and reference it from a `:generateContent` request that goes to the gateway. That can be handled in a follow-up or a separate slice.

The tokenizer and pricing entries the issue also asks for, and adapters for other non-Anthropic/non-OpenAI wire formats, are not here.


Hitaishkg:
Verified against this branch - ran a large compressible `contents` payload through `create_app` with a fake upstream. Plugin-routed `:streamGenerateContent` now reaches the pipeline: `transforms=[router:smart_crusher:0.34]`, body 32,377 → 12,045 chars, outbound URL rebuilt from the tagged base. Clean fix for the plugin case.

One scope question, not a blocker. The delegation is gated on `if upstream_base_url:`, so direct (non-plugin) `:streamGenerateContent` still forwards uncompressed:

```
:streamGenerateContent, baseline 32,377 chars

plugin header  -> transforms=[router:smart_crusher:0.34]   12,045   compressed
direct         -> transforms=[]                            32,377   not compressed
```

That leaves `:streamGenerateContent` inconsistent with `:generateContent?alt=sse`, which compresses direct streaming today. Since `handle_gemini_generate_content` already builds the default no-custom-base SSE URL (`gemini.py:428-431`), delegating unconditionally would cover the direct case too.

Was scoping to the plugin path deliberate, or worth folding direct streaming into the same delegation? Flagging so it's an explicit choice either way.

rodboev:
Good catch. I folded direct native streaming into the same delegation.

1. `handle_gemini_stream_generate_content` now delegates to `handle_gemini_generate_content` unconditionally, passing the optional tagged upstream through when present. That removes the separate direct branch that always forwarded with an empty transform list.
2. The shared generate-content path still distinguishes direct and plugin-routed URLs. Direct requests use the public Gemini `:streamGenerateContent` URL, plugin requests rebuild the same pathname under `x-headroom-base-url`, and both add `alt=sse` once.
3. The regressions now cover both routes reaching compression, plus exact URL selection, SSE bytes, and native function-call and inline-data parts. The token-counting wiring check also follows the shared generate-content owner.

This makes `:streamGenerateContent` consistent with `:generateContent?alt=sse` while keeping the tagged-upstream behavior you verified.

JerrettDavis (approved):
Reviewed the Gemini plugin routing change and current-main refresh at `90e6ee51`. The custom upstream is opted into only for the three native Gemini model endpoints, stripped from outbound headers, and consistently threaded through generate, stream, and countTokens. Native streaming now shares the non-streaming transform/conversion path, preserving non-text Gemini parts, while query rebuilding retains raw non-`alt` parameters and emits exactly one `alt=sse`. All non-Gemini handler specs remain opted out.

The remote merge tree exactly matches the locally tested tree. Validation: 34 focused plugin-routing/route-spec/token-offload tests passed, Ruff passed on the touched source/tests, and `git diff --check` is clean. The prior head's full CI was green; replacement CI is pending on this refreshed, cleanly mergeable head.

PR #2698, opened 2026-07-31
fix(compression): honor qualified CCR names across integrations

### LINKED ISSUE #2656: SmartCrusher.apply() bare-name CCR guard misses MCP-qualified headroom_retrieve; langgraph/strands .crush() calls unguarded
## Description

PR #2654 fixes `ContentRouter` recompressing `headroom_retrieve` output, but code review on
that PR (by an independent reviewer, empirically verified) found the same bug class still
reachable through two other call paths that #2654 does not touch.

### 1. `SmartCrusher.apply()`'s own #1077 guard is not alias-aware

`headroom/transforms/smart_crusher.py:1283` and `:1313` guard against recompressing
`headroom_retrieve` output with a bare comparison:

```python
if tool_names_by_id.get(msg.get("tool_call_id") or "") == CCR_TOOL_NAME:
```

This only matches the literal string `headroom_retrieve`. Claude Code (and other MCP
clients) send the qualified form `mcp__headroom__headroom_retrieve`, which this comparison
does not match. Verified empirically: calling
`SmartCrusher(config=SmartCrusherConfig(min_tokens_to_crush=0)).apply(...)` on a
`headroom_retrieve` result named `mcp__headroom__headroom_retrieve` produces
`transforms_applied = ['smart_crush:1:mcp__headroom__headroom_retrieve', ...]` and
non-byte-identical content -- i.e. it recompresses it today.

This is a real, production-reachable entry point:
`headroom/integrations/mcp/server.py:304` (`HeadroomMCPCompressor.compress()`, used by
the MCP client-wrapping utility for third-party upstream servers) calls
`crusher.apply()` directly.

Suggested fix (matches the `is_tool_excluded()` pattern #2654 uses for `ContentRouter`):

```python
from ..config import DEFAULT_VERBATIM_EXCLUDE_TOOLS, is_tool_excluded

# smart_crusher.py:1283, :1313
- if tool_names_by_id.get(msg.get("tool_call_id") or "") == CCR_TOOL_NAME:
+ if is_tool_excluded(tool_names_by_id.get(msg.get("tool_call_id") or "", ""), DEFAULT_VERBATIM_EXCLUDE_TOOLS):
```

### 2. Two other direct `.crush()` call sites have no guard at all

- `headroom/integrations/langchain/langgraph.py:263` (`compress_tool_messages`)
- `headroom/integrations/strands/hooks.py:361` (`_compress_tool_result`; its
  `_should_skip_compression` at `hooks.py:424-446` never checks `tool_name`)

Both call `SmartCrusher.crush()` directly with no tool-name-based skip check at all --
not even the bare-name check in (1). `docs/content/docs/strands.mdx:107` documents this
publicly ("The hook provider uses SmartCrusher directly for fast JSON compression"), so
it's a user-visible gap, not a theoretical one.

Suggested fix: add an `is_tool_excluded(tool_name, DEFAULT_VERBATIM_EXCLUDE_TOOLS)` check
before calling `.crush()` in both `_should_skip_compression` (strands) and `_should_skip`
(langgraph.py:156-176), analogous to what #2654 added to `ContentRouter`.

## Why a separate issue

Both gaps are pre-existing (not introduced by #2654) and live in call paths
(`HeadroomMCPCompressor`, LangGraph/Strands integrations) that are narrower and separate
from the primary `ContentRouter` proxy pipeline #2654 fixes. Bundling them into #2654
would have expanded that PR's diff beyond its stated single fix; tracking them here keeps
each PR reviewable as one logical change.

## Type of Change

- [x] Bug fix

rodboev:
## Description

Three compression consumers compare tool names against the bare literal `headroom_retrieve`, so the qualified forms MCP clients actually send (`mcp__Headroom__headroom_retrieve`, `mcp_Headroom_headroom_retrieve`) slip past the guard and get recompressed. `SmartCrusher.apply` has the bare comparison at both its OpenAI `role=tool` site and its Anthropic `tool_result` block site; the LangGraph compressor and the Strands hook have no tool-name check at all. Recompressing already-retrieved CCR content mints a new `<<ccr:hash>>` marker the agent cannot redeem.

`headroom.config.is_tool_excluded` already owns alias resolution, including the MCP wrapper forms. This routes all three consumers through it instead of adding a second name matcher. Closes #2656.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `SmartCrusher.apply` routes both its `role=tool` and its Anthropic `tool_result` guards through `is_tool_excluded`
- `_should_skip` in the LangGraph compressor takes the tool name and skips excluded tools; tool-call names are indexed by id so a `ToolMessage` without a copied `name` is still classifiable
- `_should_skip_compression` in the Strands hook takes the tool name and skips excluded tools, recording `tool_excluded`
- regressions for the qualified and bare names across all three consumers, the Anthropic block shape, the MCP wrapper entry point, and a near-match name that must still compress
- a LangGraph regression for incomplete tool-call metadata that continues to a later qualified call

## Testing

### Test Output

`pytest tests/test_smart_crusher.py tests/integrations/test_langgraph.py tests/integrations/test_strands tests/test_transforms/test_smart_crusher_ccr_retrieve_exemption.py -q`

```text
tests\test_smart_crusher.py ............                                     [ 10%]
tests\integrations\test_langgraph.py .....                                   [ 15%]
tests\integrations\test_strands\test_ccr_exclusion.py .....                  [ 19%]
tests\integrations\test_strands\test_hooks.py sssssssss                      [ 27%]
tests\integrations\test_strands\test_hooks_unit.py ssssssssssssssssssssssssssssssssss [ 57%]
tests\integrations\test_strands\test_model.py ssssssssssssssss               [ 71%]
tests\integrations\test_strands\test_model_unit.py sssssssssssssssssssssssssss [ 95%]
tests\test_transforms\test_smart_crusher_ccr_retrieve_exemption.py .....      [100%]

28 passed, 86 skipped in the focused invariant suite
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.13, `headroom._core` built
- Exact command / steps: `uv run pytest tests/test_smart_crusher.py tests/integrations/test_langgraph.py tests/integrations/test_strands -q`, and the same suite run against the pre-change implementation with the new tests in place
- Observed result: before the change, five regressions fail. `SmartCrusher` returns non-byte-identical content for a `mcp__Headroom__headroom_retrieve` result, the LangGraph compressor replaces the message content, and the Strands hook returns `"compressed"` in place of the tool output. After the change all three preserve the content byte-for-byte, incomplete LangGraph tool-call metadata is ignored while the later qualified call remains indexed, the Strands hook records `tool_excluded` and never calls the crusher, and `HeadroomMCPCompressor.compress` returns the payload unchanged. `mcp__Headroom__headroom_retrieve_extra` still compresses in all three, and the Kompress and ContentRouter suites are unchanged.
- Not tested: the optional Strands package, so the additions to `tests/integrations/test_strands/test_hooks_unit.py` skip locally

## Review Readiness

## Checklist

## Additional Notes

One deliberate divergence from the issue: the suggested snippet passes `DEFAULT_VERBATIM_EXCLUDE_TOOLS` to `is_tool_excluded`, but that constant holds only `WebSearch`, `WebFetch`, `web_search`, `web_fetch`. Applied literally it would drop `headroom_retrieve` from the comparison entirely and delete the #1077 guard these two SmartCrusher sites exist to enforce. This passes `(CCR_TOOL_NAME,)` so each guard keeps doing the one thing it documents. If you'd rather these paths also honor the verbatim-exclude set, the tuple can become `(CCR_TOOL_NAME, *DEFAULT_VERBATIM_EXCLUDE_TOOLS)` — the CCR name has to stay in it either way.

Adjacent work: PR #2654 covers `ContentRouter` only.

### DIFF
--- a/headroom/integrations/langchain/langgraph.py
+++ b/headroom/integrations/langchain/langgraph.py
@@ -47,6 +47,8 @@ except ImportError:
     BaseMessage = object  # type: ignore[misc,assignment]
     ToolMessage = object  # type: ignore[misc,assignment]
 
+from headroom.ccr.tool_injection import CCR_TOOL_NAME
+from headroom.config import is_tool_excluded
 from headroom.transforms.smart_crusher import SmartCrusher, SmartCrusherConfig
 
 logger = logging.getLogger(__name__)
@@ -156,6 +158,7 @@ def _get_crusher(min_tokens: int) -> SmartCrusher:
 def _should_skip(
     content: str,
     config: CompressToolMessagesConfig,
+    tool_name: str | None = None,
 ) -> str | None:
     """Check if a ToolMessage should skip compression.
 
@@ -163,6 +166,9 @@ def _should_skip(
     """
     if not content:
         return "empty_content"
+
+    if tool_name and is_tool_excluded(tool_name, (CCR_TOOL_NAME,)):
+        return "tool_excluded"
 
     tokens = _estimate_tokens(content)
     if tokens < config.min_tokens_to_compress:
@@ -174,6 +180,18 @@ def _should_skip(
                 return "error_content_preserved"
 
     return None
+
+
+def _tool_names_by_id(messages: list[BaseMessage]) -> dict[str, str]:  # type: ignore[type-arg]
+    """Index tool-call names so results without a copied name remain classifiable."""
+    names: dict[str, str] = {}
+    for message in messages:
+        for tool_call in getattr(message, "tool_calls", ()) or ():
+            tool_call_id = tool_call.get("id")
+            tool_name = tool_call.get("name")
+            if tool_call_id and tool_name:
+                names[tool_call_id] = tool_name
+    return names
 
 
 def compress_tool_messages(
@@ -223,6 +241,7 @@ def compress_tool_messages(
     crusher = _get_crusher(config.min_tokens_to_compress)
     result_messages: list[BaseMessage] = []
     metrics: list[ToolMessageCompressionMetrics] = []
+    tool_names = _tool_names_by_id(messages)
 
     for msg in messages:
         if not isinstance(msg, ToolMessage):
@@ -233,7 +252,8 @@ def compress_tool_messages(
         request_id = str(uuid4())
 
         # Check if we should skip
-        skip_reason = _should_skip(content, config)
+        tool_name = getattr(msg, "name", None) or tool_names.get(getattr(msg, "tool_call_id", ""))
+        skip_reason = _should_skip(content, config, tool_name)
         if skip_reason:
             result_messages.append(msg)
             tokens = _estimate_tokens(content)
--- a/headroom/integrations/strands/hooks.py
+++ b/headroom/integrations/strands/hooks.py
@@ -430,6 +430,8 @@ class HeadroomHookProvider(HookProvider):  # type: ignore[misc]
 
         Args:
             result: The tool result to check.
+            tool_name: The name of the tool that produced the result, used to
+                honor the shared tool-exclusion set.
 
         Returns:
             Skip reason string if should skip, None if should compress.
--- a/headroom/integrations/strands/hooks.py
+++ b/headroom/integrations/strands/hooks.py
@@ -48,6 +48,8 @@ except ImportError:
     ToolResult = dict  # type: ignore[misc,assignment]
 
 from headroom import HeadroomConfig
+from headroom.ccr.tool_injection import CCR_TOOL_NAME
+from headroom.config import is_tool_excluded
 from headroom.transforms.smart_crusher import SmartCrusher, SmartCrusherConfig
 
 logger = logging.getLogger(__name__)
@@ -314,7 +316,7 @@ class HeadroomHookProvider(HookProvider):  # type: ignore[misc]
         tool_use_id = event.tool_use.get("toolUseId", "unknown")
 
         # Check if compression should be skipped
-        skip_reason = self._should_skip_compression(result)
+        skip_reason = self._should_skip_compression(result, tool_name)
         if skip_reason:
             self._record_metrics(
                 request_id=request_id,
@@ -421,7 +423,9 @@ class HeadroomHookProvider(HookProvider):  # type: ignore[misc]
                 tokens_before,
             )
 
-    def _should_skip_compression(self, result: ToolResult) -> str | None:
+    def _should_skip_compression(
+        self, result: ToolResult, tool_name: str | None = None
+    ) -> str | None:
         """Check if compression should be skipped for this result.
 
         Args:
@@ -433,6 +437,9 @@ class HeadroomHookProvider(HookProvider):  # type: ignore[misc]
         # Skip if compression is disabled
         if not self.compress_tool_outputs:
             return "compression_disabled"
+
+        if tool_name and is_tool_excluded(tool_name, (CCR_TOOL_NAME,)):
+            return "tool_excluded"
 
         # Skip error results if preserve_errors is True
         if self.preserve_errors and result.get("status") == "error":
--- a/headroom/transforms/smart_crusher.py
+++ b/headroom/transforms/smart_crusher.py
@@ -52,7 +52,7 @@ from dataclasses import dataclass
 from typing import Any
 
 from ..ccr.tool_injection import CCR_TOOL_NAME
-from ..config import CCRConfig, TransformResult
+from ..config import CCRConfig, TransformResult, is_tool_excluded
 from ..tokenizer import Tokenizer
 from ..utils import compute_short_hash, create_tool_digest_marker, deep_copy_messages
 from .base import Transform
@@ -1280,7 +1280,10 @@ class SmartCrusher(Transform):
                 # unresolvable retrieval loop.
                 # ponytail: ceiling is tool_call_id lookup; if the id is missing we
                 # compress (conservative: unknown tool names don't get a free pass).
-                if tool_names_by_id.get(msg.get("tool_call_id") or "") == CCR_TOOL_NAME:
+                if is_tool_excluded(
+                    tool_names_by_id.get(msg.get("tool_call_id") or "") or "",
+                    (CCR_TOOL_NAME,),
+                ):
                     continue
                 content = msg.get("content", "")
                 if isinstance(content, str):
@@ -1310,7 +1313,10 @@ class SmartCrusher(Transform):
                     # would produce a new <<ccr:hash>> marker the agent cannot
                     # redeem (infinite retrieval loop).
                     # ponytail: ceiling is tool_use_id lookup; unknown ids pass through.
-                    if tool_names_by_id.get(block.get("tool_use_id") or "") == CCR_TOOL_NAME:
+                    if is_tool_excluded(
+                        tool_names_by_id.get(block.get("tool_use_id") or "") or "",
+                        (CCR_TOOL_NAME,),
+                    ):
                         continue
                     tool_content = block.get("content", "")
                     if not isinstance(tool_content, str):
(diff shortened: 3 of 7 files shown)


rodboev:
Thanks for flagging the partial LangGraph branch. I added a composed regression with incomplete tool-call entries before a valid qualified CCR call, then verified the matching nameless `ToolMessage` is still preserved with `tool_excluded`. This covers the index continuation path through `compress_tool_messages` without changing production behavior.

warelik:
**Confirmed on `v0.33.0` (current release)** — the bare-literal comparison this PR fixes is exactly what makes #1077 reproducible in production for any MCP-client harness. Measured repro + a config-only workaround below.

### Harness / Deployment

`kimi-code` CLI → headroom proxy (`HEADROOM_MODE=token`) → upstream gateway. `headroom` MCP server over stdio. Like every MCP client, kimi-code sends the **qualified** tool name: `mcp__headroom__headroom_retrieve`. Both Anthropic `/v1/messages` and OpenAI `/v1/chat/completions` paths affected.

### What happens (measured)

`headroom_compress` (~2 KB nonce text) → `headroom_retrieve(hash)` → model receives a nested `<<ccr:HASH,KIND,SIZE>>` marker instead of bytes; the stored original in `ccr_store.db` is intact (sqlite-verified). The `#1077` guard in [`smart_crusher.py:1283`](https://github.com/headroomlabs-ai/headroom/blob/v0.33.0/headroom/transforms/smart_crusher.py#L1283) never fires because `tool_names_by_id[...] == CCR_TOOL_NAME` compares the bare literal against the qualified name — your diagnosis in the PR description matches the production behavior exactly.

### Workaround that works today (config only)

```
HEADROOM_PROTECT_TOOL_RESULTS=headroom_retrieve
```

Works because the protect-list path uses `is_tool_excluded()`'s alias matching ([`config.py:248-298`](https://github.com/headroomlabs-ai/headroom/blob/v0.33.0/headroom/config.py#L248-L298), from #870), which strips `mcp__<server>__` before comparing — i.e., the only consumer that already honors qualified names is the user-config protect list. After enabling: retrieve returns full original bytes on both protocol paths; double-processing clean; 78 KB MCP payload round-trip byte-equal (120/120 symbols vs `ccr_store.db`).

### Coverage notes for reviewers

- The default proxy pipeline's `ContentRouter` has **no** retrieve exemption at all — bare *or* qualified — so this PR needs #2654 to land alongside it (or the qualified-name fix there) for the loop to actually close on the main path.
- Rust core is already correct: [`live_zone.rs:2362-2383`](https://github.com/headroomlabs-ai/headroom/blob/v0.33.0/crates/headroom-core/src/transforms/live_zone.rs#L2362-L2383) uses `ends_with("__headroom_retrieve")`, with a test at `:2889`.
- #1176 (chat-completions path) is the third piece of the same bug class.

### Related

#1077 (original report; #1323 fixed only the bare-name SmartCrusher path), #2654, #1176, #870.

chopratejas:
## Review: approve — one shape still slips through

Good PR, and on the axis I criticised in #2654 this one is the better of the two: it routes through the **single existing matcher** (`is_tool_excluded`) and references the canonical **`CCR_TOOL_NAME`** constant instead of adding another string literal. That's the right instinct and I'd like #2654 to adopt it (see the cross-PR note at the end).

### Verified: integration coverage is actually complete

The description says "three compression consumers", which undersells it. There are eight integration surfaces (`agno`, `autogen`, `crewai`, `langchain`, `litellm_callback`, `mcp`, `strands`, `asgi`), and I checked each rather than assuming:

- **Only `langgraph` and `strands` had their own tool-name/skip logic.** Both fixed here.
- **`crewai` delegates**: `crewai/agents.py:40` imports `compress_tool_result` from `integrations/mcp`, which builds a conversation and calls `crusher.apply(...)` at `integrations/mcp/server.py:304` — i.e. the exact entry point this PR fixes.
- I also confirmed the *shape* lines up, which is the part that could have silently failed. `compress_tool_result` synthesizes an assistant message carrying `tool_calls[].function.name = tool_name` plus `{"role": "tool", "content": ..., "tool_call_id": "call_1"}`. That is precisely what the fixed guard's `tool_names_by_id[tool_call_id]` lookup can resolve — so a host application calling `compress_tool_result(content, tool_name="mcp__Headroom__headroom_retrieve")` is now correctly skipped. Worth stating in the PR, because it's a non-obvious win.
- **`litellm_callback`** goes through `headroom.compress()` → `ContentRouter`, which is #2654's half.
- `agno`, `autogen`, `asgi` carry no tool-name logic at all.

So coverage is complete across the two PRs together. Please say so explicitly — right now a reader can't tell whether `crewai` or a bare MCP host app is covered.

The conservative default (`ponytail: if the id is missing we compress — unknown tool names don't get a free pass`) is the right direction for genuinely unknown tools.

---

### Gap: no `name`-field fallback for legacy `role:"function"`

Both new guards resolve the name **only** through the id map:

```python
if is_tool_excluded(
    tool_names_by_id.get(msg.get("tool_call_id") or "") or "",
    (CCR_TOOL_NAME,),
):
```

Legacy `role:"function"` messages (OpenAI's pre-parallel-tool-calls shape) carry **no `tool_call_id`** — the tool name sits on the message itself as `name`. So a `headroom_retrieve` result in that shape resolves to `""`, fails the check, and gets recompressed — the exact unredeemable-marker loop this PR closes, still live for one wire shape.

#2654 hits this same case in `ContentRouter` and handles it explicitly:

> *"Legacy `role:"function"` messages carry no call id in that shape, so the tool name is read directly off the message's `name` field instead of through the id-keyed `tool_name_map`."*

Net effect if only this PR merges: the same retrieve result is **protected** through `ContentRouter` and **recompressed** through `SmartCrusher.apply`.

Note the existing `ponytail:` justification doesn't cover this case. "Unknown tool names don't get a free pass" is fair when the name is genuinely unavailable — but here it *is* available, on the message, and simply isn't consulted. Mirroring #2654:

```python
name = tool_names_by_id.get(msg.get("tool_call_id") or "") or (
    msg.get("name") or "" if msg.get("role") == "function" else ""
)
if is_tool_excluded(name, (CCR_TOOL_NAME,)):
    continue
```

The Anthropic `tool_result` site doesn't need this — that shape always carries `tool_use_id`.

---

### Cross-PR: this and #2654 are two halves of one fix

Neither PR mentions the other, and they close the same failure mode (#1077 / #2656) at different entry points:

| | entry point | reaches |
|---|---|---|
| **#2654** | `ContentRouter` | proxy + default pipeline, `litellm_callback` |
| **#2698** (this) | `SmartCrusher.apply` | integrations, MCP hosts, `crewai`, langgraph, strands |

**Merging only one leaves a live path.** They touch disjoint files (`content_router.py` + `config.py` vs `smart_crusher.py` + integrations) so there's no conflict and either order works — but they should land as a pair.

One divergence to reconcile while both are open: this PR uses `CCR_TOOL_NAME`; #2654 adds five bare `"headroom_retrieve"` literals. If both merge as-is the codebase carries two patterns for the same check, and the literals are a silent-drift risk (there are already two independent `CCR_TOOL_NAME` definitions). I've asked #2654 to align on your approach.

Disclosure: reviewed statically and from CI (23 pass / 0 fail); I could not run the suite locally.

chopratejas (approved):
Approving. Best-practice on the axis that matters here: routes through the single existing matcher (`is_tool_excluded`) and references `CCR_TOOL_NAME` rather than adding another literal. I verified integration coverage end-to-end rather than taking the "three consumers" framing — `crewai` reaches this via `integrations/mcp.compress_tool_result` → `crusher.apply()` at `mcp/server.py:304`, and the message shape it synthesizes is exactly what the fixed id→name lookup resolves. Worth stating in the description; it's a non-obvious win.

One gap to close, ideally here: legacy `role:"function"` messages carry no `tool_call_id`, so the name resolves to `""` and the result still recompresses. #2654 handles that shape by reading `message["name"]` — copy it. Note the existing `ponytail: unknown tool names don't get a free pass` comment doesn't cover this case: the name isn't unknown, it's on the message and simply not consulted.

Merge as a pair with #2654.

PR #2699, opened 2026-08-01
fix(proxy/cache): reuse the prefix-cache lineage across same-message block appends

### LINKED ISSUE #2671: Prefix-cache: two independent cache-write sinks in the Anthropic path

rodboev:
## Description

This is defect 2 from #2671, but not the fix the issue sketches. The issue reads the symptom as the breakpoint sitting on the newest block; the actual cause is one layer up, in lineage selection.

`SessionTrackerStore.resolve_tracker` picks the recorded chain whose canonicalized form is an exact prefix of the incoming history. A message that grows block by block is not an exact prefix of anything recorded, so every turn allocated a fresh tracker. That is the empty previously-forwarded state the issue reports, and it explains the measurement directly: with no prior forwarded prefix there is nothing for `overlay_cached_prefix` to replay, so the compressed bytes we forward differ turn to turn and the prefix hash never matches. `cache_read` pinned at system+tools only, `cache_write` growing.

So this teaches the lineage lookup that a single message growing by whole trailing blocks is still the same conversation, and fixes the replay merge to carry the previously forwarded blocks plus the new tail.

The breakpoint itself is left where it was. Anthropic writes a cache entry only at the breakpoint and, on lookup, walks backward up to 20 blocks looking for a prior write. The newest block is therefore the only position that both reads last turn's entry and writes a new entry covering this turn's growth; anchoring it backward would re-write an already-cached prefix and leave the appended blocks out of the cache entirely. The issue's own shape grows one to five blocks a turn, well inside the lookback window.

Closes #2671 for defect 2. The `tools`-array toggle is defect 1 and is not in this branch.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `classify_append_only_prefix` is the one place that answers "does this history append-only-extend that one", and it reports the block frontier when exactly one existing message grew by whole trailing blocks
- lineage selection routes through it, so a block append reuses its tracker instead of starting a fresh one. It calls the classifier's inner form directly, since the snapshot and the recorded chain are already canonical projections and re-canonicalizing them once per recorded lineage is the expensive part
- a message that grew blocks while any field outside `content` changed is a replacement, not an append, so the replay never forwards last turn's role or tool ids under this turn's blocks
- `overlay_cached_prefix` merges the previously forwarded blocks with this turn's appended tail. It bails when the forwarded and original block counts disagree, which is the block-level analog of its existing message-count guard, and it re-establishes the split against the raw previous originals rather than the classifier's index, since the classifier works on a projection that drops pure cache-directive blocks
- focused regressions cover the defensive classifier and overlay fallback paths reported by Codecov; production behavior is unchanged

## Testing

### Test Output

`pytest tests/test_cache/test_prefix_tracker.py tests/test_cache_control_move_bust.py tests/test_proxy_anthropic_cache_stability.py tests/test_cache_prefix_overlay.py tests/test_cache_mode_delta_marker.py tests/test_proxy_openai_cache_stability.py tests/test_cache_ttl_preserved.py tests/test_cross_turn_cache_safety.py -q`

```text
tests\test_cache\test_prefix_tracker.py ...................................  [ 47%]
tests\test_cache_control_move_bust.py ..........                             [ 54%]
tests\test_proxy_anthropic_cache_stability.py .................................  [ 77%]
tests\test_cache_prefix_overlay.py .........                                 [ 83%]
tests\test_cache_mode_delta_marker.py ..........                             [ 90%]
tests\test_proxy_openai_cache_stability.py ......                            [ 94%]
tests\test_cache_ttl_preserved.py .....                                      [ 97%]
tests\test_cross_turn_cache_safety.py ...                                    [100%]

152 passed, 1 warning in 43.47s
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.13, no Anthropic credentials; `create_app` composed with a fake upstream
- Exact command / steps: two consecutive `POST /v1/messages` turns through the composed app in cache mode, where the first message grows from two blocks to three between turns and the assistant reply is resent. Nothing about the tracker is stubbed: the session id, the lineage lookup, and the prior forwarded state all come from the real `SessionTrackerStore`, and turn one's state is recorded by the handler's own `update_from_response`. The test spies on `resolve_tracker` rather than replacing it, so it can assert which tracker each turn landed on.
- Observed result: on the pre-change implementation turn two allocates a second tracker and its recorded forwarded messages are empty, which is the inert state the issue describes. After the change both turns resolve the same tracker, the recorded prefix is there to replay, and the leading blocks the provider hashed last turn are forwarded byte-identical. The single breakpoint stays on the last block of the last block-style message.
- Not tested: live Anthropic cache-hit metrics

## Review Readiness

## Checklist

## Additional Notes

I did try the relocation the issue describes and then backed it out, so it's worth recording why in case someone picks it up again: moving the breakpoint back to the prior frontier writes a cache entry for a prefix that was already written last turn and never writes the appended blocks, so it loses ground every turn rather than gaining it. In a history where the grown message is not the last block-style message it is worse still, because the anchor moves backward across whole messages.

Whole-message appends, TTL payloads, divergent histories, and parallel sessions keep their current behavior. The negative cases cover removed, reordered, inserted, and replaced blocks, and a block append whose message role changed.

One caveat on the replay path: the recorded prior history includes the assistant turn reconstructed from the response body, and the classifier requires every prior message to match, so a client that resends an assistant turn in a different shape than we reconstructed will drop the replay for that turn. Lineage selection is unaffected, since it stamps the request messages rather than the reconstruction.


rodboev:
Added focused regression coverage for the Codecov paths in `prefix_tracker.py`, including defensive classifier rejection and overlay fallback cases. Production code is unchanged; the focused suite now passes 152 tests.

chopratejas:
@rodboev — this is a cross-post; the same comment is on #2702. Your counterpart PR is **#2702** (@nangsontay), which fixes the same `resolve_tracker` cause and additionally relocates the breakpoint. Please take a look at it — the analysis below argues your breakpoint reasoning is right for *your* shape and theirs is right for a different one.

## Heads-up: #2699 and #2702 overlap, conflict, and disagree

These two PRs are solving the same issue (#2671 defect 2) in the same functions, by different means. **@rodboev and @nangsontay — please read each other's PR before either lands.** I don't think either should merge as-is, and a merge-order race here risks landing a half-fix that quietly degrades cache behaviour.

### Both are individually green, and that's part of the problem

| | #2699 | #2702 |
|---|---|---|
| CI | 22 pass / 0 fail | 22 pass / 0 fail |
| Merges into current main (`08fce29b`) | clean | clean |
| **Merge into each other** | **conflict in `headroom/cache/prefix_tracker.py`** | same |

Verified by trial merge. Both touch the same four functions:

- `overlay_cached_prefix`
- `normalize_message_cache_control`
- `_lineage_snapshot`
- `resolve_tracker`

### You agree on the root cause

Both descriptions independently land on the same diagnosis: `resolve_tracker` matches a conversation to its tracker by requiring the recorded messages to be a whole-**message** prefix of the incoming history, so a message that grows by **blocks** reads as diverged on every turn → fresh tracker → frozen prefix stuck at 0 → nothing for `overlay_cached_prefix` to replay → the forwarded bytes differ turn to turn and the prefix hash never matches. Two people arriving at that separately is good corroboration.

### You disagree on the breakpoint — and you're each right about a different shape

- **#2699:** leaves the breakpoint on the newest block, arguing that *"anchoring it backward would re-write an already-cached prefix and leave the appended blocks out of the cache entirely."*
- **#2702:** moves it to the end of the stable prefix, citing Anthropic's guidance for varying tails.

Both hold, for different histories:

| Shape | Is the current tail still there next turn? | Correct breakpoint |
|---|---|---|
| Blocks **appended**, earlier blocks stable (#2699's shape) | yes | **newest block** — the write covers the new tail |
| Last blocks **rewritten** every turn (#2702's shape) | no | **end of the stable run** — the newest block never repeats, so nothing ever matches |

Neither PR handles the other's shape.

### A specific bug in #2702's discriminator

`_breakpoint_index` tries to separate the two cases:

```python
run = _stable_leading_block_run(content, previous_blocks)
# "Only anchor backwards when the stable run is real (the message diverged
#  before its end) and covers most of the message."
if 1 <= run < len(content) and run * 2 >= len(content):
    return run - 1
```

The comment describes the right rule, but the guard compares `run` against **`len(content)`** — this turn's block count — rather than against `len(previous_blocks)`. A pure append satisfies it:

> Previous message has `N` stable blocks; `k` new blocks are appended (`k` = 1–5, per #2699's shape). Then `run == N` and `len(content) == N + k`.
> `run < len(content)` ✓, and `run * 2 >= len(content)` ⟺ `N >= k` — always true once the `_MIN_BLOCKS_FOR_RELOCATION = 20` gate is met.

So #2702 relocates the breakpoint on **append-only** histories too, excluding the freshly appended blocks from that turn's cache write. That is exactly #2699's objection, and it's why the two PRs read as contradictory.

The intent is right; one comparison implements it:

```python
if 1 <= run < len(previous_blocks) and run * 2 >= len(content):
```

- pure append → `run == len(previous_blocks)` → falls through to `newest` (#2699's shape, correct)
- rewritten tail → `run < len(previous_blocks)` → relocates (#2702's shape, correct)

With that, both of your claims are simultaneously true.

### Suggested convergence

1. **Take #2699's lineage fix as the base.** It's the more surgical of the two — an explicit `AppendOnlyClassification` / `block_frontier` threaded through `extract_cache_stable_delta` and `overlay_cached_prefix` — versus #2702's broader `_lineage_snapshot` rewrite.
2. **Layer #2702's breakpoint relocation on top, with the corrected discriminator**, and keep its `HEADROOM_STABLE_BOUNDARY_BREAKPOINT` kill switch. An env rollback for something on the cache-key path of every Anthropic request is good judgement, and #2699 has no equivalent.
3. Whoever lands second: **re-run CI after resolving the conflict.** Both currently passed against a base that does not contain the other.

### Note on test coverage

CI would not have caught the append-relocation issue: it needs a message of **≥20 blocks** that grows by **pure append**, and I don't believe either PR's tests construct that. Worth adding as part of the merge — it's the case that separates the two designs.

Caveat on my side: I reviewed this statically and from CI. I could not run the suite locally, and the `cache_read pinned at system+tools while cache_write grows` measurements in both descriptions are from your own logs and I haven't reproduced them.

JerrettDavis (approved):
Approved at `667a1a6a` for the append-only growth case this PR claims.

I traced the canonical/raw boundary through lineage resolution and `overlay_cached_prefix`: non-content metadata changes are rejected, directive-only canonicalization cannot hand raw callers shifted indices, and replay only occurs when the prior forwarded/raw block counts and current raw prefix prove the splice is safe. The unchanged newest-block breakpoint is correct for pure appends because it both finds the prior write within Anthropic's lookback and writes the newly appended tail.

Local verification: 76 prefix-tracker tests + 33 composed Anthropic cache-stability tests passed; Ruff passed on the changed production/test files. The head also merges cleanly with current `main`, and CI is green.

#2702 remains a competing/conflicting implementation for rewritten-tail histories. This approval is for #2699's scoped append-only fix, not approval to merge both independently; the final human should choose/converge the pair so the second PR cannot overwrite this behavior.

rodboev:
Added the 20+ block pure-append regression @chopratejas and @JerrettDavis flagged: commit `ad72b1d1` pins newest-block placement in `tests/test_cache_control_move_bust.py`, so the append-only shape has explicit coverage on this side before the #2702 relocation rebases onto it. The two shapes stay complementary, and the corrected `_breakpoint_index` discriminator (`run < len(previous_blocks)`) belongs to #2702's rework.

Heads-up on check status: the only failing check on this head is Vercel's deploy-preview check (`Authorization required to deploy`), an integration-authorization gate rather than a code failure; the Actions checks pass. This PR remains the append-only lineage base and can merge as-is, with #2702 carrying the relocation and the discriminator fix.

JerrettDavis (approved):
Re-reviewed current head ad72b1d1. The only change since my prior approval is a focused regression pinning newest-block breakpoint placement for pure appends beyond the provider lookback window. It strengthens the exact scoped invariant previously reviewed and does not alter production behavior. The earlier caveat remains: #2702 is a competing implementation and the final human should converge the pair rather than independently landing behavior that can overwrite the same normalization path. No new blocker.

chopratejas:
Thank you for isolating and testing the pure block-append path. I opened #2917 as a consolidated successor after the latest #2671 trace showed that we need to support both this append shape and #2702's rewritten-tail shape, plus keep different tool profiles on separate trackers.

#2917 retains this PR's key behavior: replay the exact previously forwarded blocks and keep pure appends on the newest breakpoint. It also makes the discriminator explicit (`stable_prefix_blocks == previous_block_count` for append, strictly less for rewrite), preventing the append path from being accidentally routed through stable-boundary relocation.

I credited this PR's contribution in #2917. Please take a look at the combined classifier and cache-oracle tests; they are intended to preserve the behavior proven here while avoiding three competing fixes for the same production incident.

JerrettDavis:
Superseded by #2917, merged to main as 1a04c957. The merged implementation covers block-growing lineages and breakpoint stability with stricter shared relation classification and regression coverage, so retaining this conflicting older branch would duplicate the shipped fix.

PR #2702, opened 2026-08-01
fix(cache): anchor the message breakpoint to the static prefix when a message grows in place

nangsontay:
## Description

`normalize_message_cache_control` re-places its single message-level breakpoint on the **newest block** of the last block-style message. That is correct for histories that grow by appending messages: the provider walks back up to 20 content-block boundaries from a breakpoint looking for a previous write, so it still finds last turn's.

It is wrong for a message that grows **in place with a varying tail**. One caller shape (`msgs=2`, sonnet) packs a transcript into a single block-style message and rewrites its last few blocks every turn. The breakpoint then rides a block that never repeats, no stable prefix is ever found, and the message re-writes in full forever — `cache_read` pinned at the system+tools constant while `cache_write` grows with the transcript. Anthropic's documented guidance covers this case exactly: place the breakpoint at the end of the static prefix rather than on the varying block.

Fixing that alone is a no-op, which #2671 recorded without identifying why. The cause is upstream of the breakpoint: `SessionTrackerStore.resolve_tracker` matches a conversation to its tracker by requiring the recorded messages to be a whole-**message** prefix of the incoming history (`snap[:len(chain)] == chain`), assuming client histories only ever append messages. This shape appends **blocks to its last message**, so the recorded last message never re-compares equal and the conversation reads as diverged on every turn — a fresh tracker each time, frozen prefix permanently 0, nothing for `overlay_cached_prefix` to replay, and one leaked lineage per turn until the family hits `max_lineages_per_session`. Reproduced in isolation: 5 turns of the captured shape give **5 distinct trackers**; with this change, **1**.

Hence two commits — the lineage match has to be repaired before any breakpoint logic can see the previous turn.

Fixes defect 2 of #2671. Defect 1 is #2672; #2671 can close once both land.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `resolve_tracker` now also matches a lineage whose recorded last message **grew in place**: the messages before it must still compare equal, and the last one must share the same role, have block-style content that did not shrink, and keep a leading run of byte-stable blocks covering most of the recorded version. Extracted to `_match_lineage` / `_chain_matches` / `_is_message_continuation`, which keeps `resolve_tracker`'s complexity below where it started.
- Strict whole-message matches are tried first, so an exact continuation is never displaced by a looser one.
- Rewritten heads (client-side compaction), shrunk messages, and templated fan-outs that diverge from the first block of their newest message all still start a fresh lineage — the concurrent-conversation split from #2085 is unchanged.
- `normalize_message_cache_control` takes an optional `previous_forwarded_messages` and picks its breakpoint via a new `_breakpoint_index`, which compares the target message against **its own counterpart from last turn** (same position, same role) and anchors to the end of that message's byte-stable leading run.
- Relocation requires all of: a run that is real (the message diverged before its end), a run covering most of the message, and a message longer than the 20-block lookback window. A main conversation therefore keeps newest-block placement untouched.
- Comparison goes through the existing `_canonicalize_for_prefix_compare`, so last turn's `cache_control` marker does not shorten the run.
- `HEADROOM_STABLE_BOUNDARY_BREAKPOINT=0` restores newest-block placement unconditionally, without a redeploy.
- Handler reads the previously-forwarded prefix once and shares it between the overlay replay and the breakpoint decision, rather than deep-copying the transcript twice per request.

Commit 3 narrows the continuation match after review found it merged conversations that lineages exist to keep apart. All three narrowings are pinned by tests that fail without them:

- The **final block must be unchanged.** Two sub-call streams over one parent transcript — a summarizer and a titler, say — share nearly every leading block and differ only in the instruction appended at the end. They traded one tracker back and forth every turn and never recovered, since neither ever gains a message to tell them apart. The real growing shape keeps a fixed suffix pinned at the end while the blocks before it churn, so the tail is exactly what separates the two cases.
- The stable run must clear an **absolute floor**, not just a proportion. Injected boilerplate can bracket a message on both sides, leaving two conversations agreeing on their first blocks and their last while differing in the only part that is theirs.
- The loose pass requires an **unchanged message count.** It previously ignored everything after the recorded chain, so a two-message chain could match any longer history whose second message shared half its blocks. A history that gained a message and rewrote an older one was rewritten, not grown.

Worth stating plainly for reviewers: no wrong content could be forwarded in either case. The per-message canonical guards in `overlay_cached_prefix` and `extract_cache_stable_delta` stop a mismatched tracker from replaying another conversation's bytes; the damage was cache thrash and polluted frozen counts.

## Testing

New tests drive the real sequence — `resolve_tracker` → `normalize_message_cache_control` → `update_from_response` — rather than hand-building previous-turn state. That distinction is the point: the earlier attempt at this fix passed its unit tests precisely *because* they supplied the previous-turn state that production never delivered. These fail if that state stops arriving.

Covered: the in-place-growth shape; the chaining property (each breakpoint reachable from the previous write); main conversations keeping newest-block placement; short messages and mostly-rewritten messages declining relocation; the kill switch; marker-count and content invariants. Guard tests pin the lineage splits that must not regress — diverging fan-outs, rewritten heads, shrinking messages, strict-match precedence.

Both new test groups are red on the base commit for the right reason (`AssertionError: turn 2 switched trackers`).

Every guard in this PR was mutation-tested — each one reverted in turn, confirming a test goes red. Without this, a guard is just an assertion nobody checks:

| mutation | caught by |
|---|---|
| handler drops `previous_forwarded_messages` | `test_handler_passes_previous_forwarded_messages_to_breakpoint_placement` |
| drop the stable-tail requirement | `test_sibling_sub_calls_over_one_transcript_do_not_ping_pong` |
| drop the absolute run floor | `test_boilerplate_on_both_ends_is_not_enough_to_share_a_lineage` |
| drop the message-count guard | `test_added_message_plus_rewritten_older_one_is_not_in_place_growth` |
| breakpoint back to newest block | 2 tests |
| lineage loose pass disabled | 4 tests |

### Test Output

```text
$ python -m pytest -q
11 failed, 9662 passed, 557 skipped, 5889 warnings in 357.44s (0:05:57)

$ python -m ruff check .
All checks passed!

$ python -m mypy headroom

# Red on base (28aa53dc) for the right reason, before the fix:
$ python -m pytest tests/test_cache/test_prefix_tracker.py -q \
    -k test_in_place_message_growth_keeps_its_lineage
    assert tracker is first, f"turn {turn} switched trackers"
E   AssertionError: turn 2 switched trackers
FAILED tests/test_cache/test_prefix_tracker.py::TestConversationLineageResolution::test_in_place_message_growth_keeps_its_lineage
```

The 11 failures are pre-existing and unrelated — the identical set fails on the base commit with the same command, so this branch introduces zero regressions. They are dashboard Playwright tests, `readyz` kompress-state tests, an ONNX image-classification test, and two CLI wrap/recover tests:

```text
tests/test_cli/test_recover_codex.py::test_recovery_records_sockets_and_secures_both_backups
tests/test_cli/test_unwrap_claude.py::test_unwrap_claude_stops_claude_owned_persistent_deployment
tests/test_dashboard/test_live_feed.py::test_live_feed_button_exists
tests/test_dashboard/test_live_feed.py::test_live_feed_drawer_opens
tests/test_dashboard/test_live_feed.py::test_live_feed_fetches_and_displays
tests/test_dashboard/test_live_feed.py::test_live_feed_shows_empty_state
tests/test_image_compression.py::TestOnnxRouter::test_full_classify_with_image
tests/test_proxy_health.py::test_readyz_excludes_kompress_from_aggregate_readiness
tests/test_proxy_health.py::test_readyz_keeps_pending_kompress_unloaded
tests/test_proxy_health.py::test_readyz_kompress_state_matrix[null-None-False-expected0]
tests/test_proxy_health.py::test_readyz_never_calls_lazy_kompress_getters
```

The cache and proxy suites this PR touches are green:

```text
$ python -m pytest tests/test_cache/ tests/test_cache_control_move_bust.py \
    tests/test_cross_turn_cache_safety.py tests/test_cache_ttl_preserved.py \
314 passed, 2 skipped, 1 warning in 14.26s
```

## Real Behavior Proof

- Environment: local Headroom proxy on macOS (darwin 25.4.0), Python 3.13.14, serving real Claude Code traffic against `claude-sonnet-5` and `claude-opus-5`, running this branch's two commits applied to the daily-driver checkout.
- Exact command / steps: apply both commits, restart the proxy (`python -m headroom.cli proxy`), use it normally for ~9 minutes, then run the frozen parser from the #2671 baseline unmodified over the post-restart log slice — `python parse_perf.py post-fix.log`.
- Observed result: the relocation fires and chains as designed, the sub-call `cache_read` pin at 36,952 is broken, per-turn `cache_write` for that shape drops from ~28–48k to 42–109, and the `msgs<=4` bucket goes from 43.1% hit / 54.9% of all cache write to 92.4% hit / 14.0% — details below.
- Not tested: the OpenAI handler (unchanged — `normalize_message_cache_control` has a single call site, in the Anthropic handler, so non-Anthropic providers are unaffected); durability beyond a 41-turn / 9-minute window, which is far smaller than the 1678-turn baseline and the 235-turn soak behind #2672, so the magnitude below is directional while the mechanism is confirmed; load, concurrency, and multi-workspace behaviour beyond normal interactive use.

The relocation chains with the anchor tracking one turn behind the growth, the gap staying well inside the 20-block lookback:

```text
stable run / blocks:  131/134 → 131/135 → 132/136 → 133/137 → 134/138
                    → 135/139 → 136/140 → 138/141 → 138/142 → 139/143 → 140/144
```

The pin is broken. Consecutive turns of one affected sub-call conversation:

| tok_before | cache_read | cache_write | hit |
|---:|---:|---:|---:|
| 40,292 | 0 | 65,362 | 0% (cold) |
| 40,405 | 36,952 ← the pinned constant | 28,323 | 57% |
| 40,497 | 65,275 | 0 | 100% |
| 40,567 | 65,275 | 42 | 100% |
| 40,645 | 65,317 | 42 | 100% |
| 40,814 | 65,359 | 109 | 100% |

`cache_read` now tracks the conversation upward instead of pinning, and per-turn `cache_write` falls from ~28–48k to 42–109. `frozen=1` also starts appearing on `msgs=2` requests that were always `frozen=0` before — direct evidence the tracker survives across turns instead of being recreated.

Bucket totals from the same parser, against the figures in #2671:

| sub-call bucket (`msgs<=4`) | hit | share of all cache write |
|---|---:|---:|
| baseline (1678 turns) | 58.8% | 44.8% |
| after defect 1 fixed in #2672 (235 turns) | 43.1% | 54.9% |
| this change (41 turns) | **92.4%** | **14.0%** |

Aggregate over the window: 91.06%. No 400s and no `cache_control` errors were observed.

## Review Readiness

## Checklist

## Additional Notes

- **Documentation:** N/A — no user-facing surface changes. The behaviour is described in the two functions' docstrings, including why the newest-block default stays correct for ordinary conversations.
- **Why three commits:** the first two are separable defects and each stands on its own (the lineage fix also stops the per-turn tracker and lineage churn regardless of breakpoints), but the breakpoint fix is inert without the lineage fix, so shipping only the second would repeat the earlier no-op. The third narrows the continuation predicate in response to review; it is kept separate so the reasoning behind each guard stays readable, and can be squashed on merge.
- **Measurement predates commit 3.** The live numbers above were taken with the looser predicate. Commit 3 only *rejects* matches, so it can reduce how often the win applies but cannot inflate it; the affected shape (runs of 100+ blocks behind a pinned suffix) still matches. Re-measuring after the narrowing is the obvious follow-up and I am happy to do it before merge if you would rather not take that on trust.
- **Correction to #2671:** the issue body describes this caller as growing "append-only". It does not — it rewrites its tail behind a fixed 2-block suffix. That is precisely why the anchor has to be the *stable leading run* rather than an append-only assumption, which would have missed it.
- **Rollback:** `HEADROOM_STABLE_BOUNDARY_BREAKPOINT=0` disables the relocation and restores the previous placement. The lineage change has no switch; it is a match-predicate widening whose failure mode is the pre-existing behaviour (a fresh lineage).

### DIFF
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -21,6 +21,7 @@ import hashlib
 import itertools
 import json
 import logging
+import os
 import time
 from collections import OrderedDict
 from dataclasses import dataclass
@@ -384,8 +385,81 @@ def _stable_leading_block_run(
     return k
 
 
+# Env kill switch for the stable-boundary breakpoint placement below. Default on;
+# set to 0/false/no/off to restore the newest-block placement unconditionally.
+# This sits on the cache-key path of every Anthropic request, so it needs a
+# rollback that does not require shipping a new build.
+_STABLE_BOUNDARY_ENV = "HEADROOM_STABLE_BOUNDARY_BREAKPOINT"
+
+# Below this many blocks a message cannot benefit from relocation: the provider
+# walks back up to 20 content-block boundaries from the breakpoint looking for a
+# previous write, so a short message's prefix is still found from the newest
+# block, and anchoring backwards would only shrink what gets cached.
+_MIN_BLOCKS_FOR_RELOCATION = 20
+
+
+def _stable_boundary_enabled() -> bool:
+    return os.environ.get(_STABLE_BOUNDARY_ENV, "").strip().lower() not in (
+        "0",
+        "false",
+        "no",
+        "off",
+    )
+
+
+def _breakpoint_index(
+    content: list[Any],
+    msg: dict[str, Any],
+    msg_idx: int,
+    previous_forwarded_messages: list[dict[str, Any]] | None,
+) -> int:
+    """Index within ``content`` that should carry the single message breakpoint.
+
+    Default is the newest block, which is right whenever the provider's 20-block
+    lookback can still reach last turn's write from there — a cold turn, or a
+    conversation that grows by appending messages.
+
+    It is wrong for a message that grows IN PLACE with a varying tail: the
+    breakpoint then rides a block that never repeats, so no stable prefix is ever
+    found and the whole message re-writes every turn. For that shape the
+    breakpoint belongs at the end of the static prefix, which is the previous
+    turn's counterpart message compared block by block.
+
+    The counterpart is looked up by position AND role: the recorded forwarded
+    list is last turn's, so a growing conversation only lines up where the shape
+    really is stable, and any mismatch falls back to the newest block.
+    """
+    newest = len(content) - 1
+    if not previous_forwarded_messages or not _stable_boundary_enabled():
+        return newest
+    if len(content) < _MIN_BLOCKS_FOR_RELOCATION or msg_idx >= len(previous_forwarded_messages):
+        return newest
+    counterpart = previous_forwarded_messages[msg_idx]
+    if not isinstance(counterpart, dict) or counterpart.get("role") != msg.get("role"):
+        return newest
+    previous_blocks = counterpart.get("content")
+    if not isinstance(previous_blocks, list):
+        return newest
+    run = _stable_leading_block_run(content, previous_blocks)
+    # Only anchor backwards when the stable run is real (the message diverged
+    # before its end) and covers most of the message. A short run would cache
+    # less than the newest-block placement writes, which is a worse trade even
+    # though it reads.
+    if 1 <= run < len(content) and run * 2 >= len(content):
+        logger.debug(
+            "cache breakpoint anchored to the stable run of %d/%d blocks in message %d "
+            "(its newest block varies turn over turn)",
+            run,
+            len(content),
+            msg_idx,
+        )
+        return run - 1
+    return newest
+
+
 def normalize_message_cache_control(
     messages: list[dict[str, Any]],
+    previous_forwarded_messages: list[dict[str, Any]] | None = None,
 ) -> list[dict[str, Any]]:
     """Own message-level cache_control placement so breakpoints stay bounded.
 
@@ -396,13 +470,29 @@ def normalize_message_cache_control(
     so on a long conversation the accumulation eventually 400s.
 
     Fix: strip EVERY message-level cache_control and re-place a **single**
-    ephemeral breakpoint on the last block of the last block-style message. One
-    breakpoint caches the whole message prefix up to it, and — because the
-    provider's cache key is message CONTENT, not marker presence (moving the
-    breakpoint forward is the documented client pattern and it hits) — stripping
-    and re-placing markers never busts. system/tools breakpoints live outside
-    ``messages`` and are left untouched (they still count toward the 4 limit, so
-    holding messages to one breakpoint leaves room for them).
+    ephemeral breakpoint. One breakpoint caches the whole message prefix up to
+    it, and — because the provider's cache key is message CONTENT, not marker
+    presence (moving the breakpoint forward is the documented client pattern and
+    it hits) — stripping and re-placing markers never busts. system/tools
+    breakpoints live outside ``messages`` and are left untouched (they still
+    count toward the 4 limit, so holding messages to one breakpoint leaves room
+    for them).
+
+    WHERE that one breakpoint goes is the last block of the last block-style
+    message — except when that message grew IN PLACE since last turn. Sub-call
+    shapes pack a transcript into one block-style message and rewrite its tail
+    each turn, so a breakpoint on its newest block can never match next turn and
+    pins ``cache_read`` at the system+tools constant forever, however stable the
+    rest of the message is. When ``previous_forwarded_messages`` shows that the
+    same message diverged partway through, the breakpoint is anchored to the end
+    of its byte-stable leading run instead: that boundary IS cached, so the run
+    reads from cache and the breakpoint advances one turn behind the growth.
+
+    Relocation only fires when the stable run covers most of the message
+    (otherwise anchoring backwards would cache less than it saves) and only for
+    the same message position and role, so a main conversation — whose newest
+    message is genuinely new each turn — keeps the newest-block placement.
+    ``HEADROOM_STABLE_BOUNDARY_BREAKPOINT=0`` restores it unconditionally.
 
     Headroom owns WHERE the breakpoint goes; the client still owns WHAT it says:
     the re-placed marker reuses the newest client marker verbatim, so an explicit
@@ -442,7 +532,8 @@ def normalize_message_cache_control(
         msg = out[last_block_idx]
         content = list(msg["content"])
         marker = dict(last_marker) if last_marker else {"type": "ephemeral"}
-        content[-1] = {**content[-1], "cache_control": marker}
+        bp_idx = _breakpoint_index(content, msg, last_block_idx, previous_forwarded_messages)
+        content[bp_idx] = {**content[bp_idx], "cache_control": marker}
         out[last_block_idx] = {**msg, "content": content}
         changed = True
     return out if changed else messages
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -359,6 +359,31 @@ def overlay_cached_prefix(
     return list(prev_fwd[:k]) + list(optimized_messages[k:])
 
 
+def _stable_leading_block_run(
+    current_blocks: list[Any],
+    previous_blocks: list[Any] | None,
+) -> int:
+    """Length of the longest leading run of content blocks that canonicalize-
+    equal the previous turn's blocks.
+
+    Uses :func:`_canonicalize_for_prefix_compare`, which drops ``cache_control``
+    and other non-semantic keys, so a moved breakpoint or per-turn annotation
+    churn does not shorten the run. Two blocks canonicalize-equal iff their
+    forwarded content is the same, which is exactly what the provider's prefix
+    cache keys on — so this run is the part of a message that can still be read
+    from cache.
+    """
+    if not previous_blocks:
+        return 0
+    limit = min(len(current_blocks), len(previous_blocks))
+    k = 0
+    while k < limit and _canonicalize_for_prefix_compare(
+        current_blocks[k]
+    ) == _canonicalize_for_prefix_compare(previous_blocks[k]):
+        k += 1
+    return k
+
+
 def normalize_message_cache_control(
     messages: list[dict[str, Any]],
 ) -> list[dict[str, Any]]:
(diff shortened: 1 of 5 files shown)


chopratejas:
@nangsontay — this is a cross-post; the same comment is on #2699. Your counterpart PR is **#2699** (@rodboev), which fixes the same `resolve_tracker` cause but deliberately leaves the breakpoint on the newest block. Please take a look at it — and note the discriminator issue flagged below in `_breakpoint_index`.

## Heads-up: #2699 and #2702 overlap, conflict, and disagree

These two PRs are solving the same issue (#2671 defect 2) in the same functions, by different means. **@rodboev and @nangsontay — please read each other's PR before either lands.** I don't think either should merge as-is, and a merge-order race here risks landing a half-fix that quietly degrades cache behaviour.

### Both are individually green, and that's part of the problem

| | #2699 | #2702 |
|---|---|---|
| CI | 22 pass / 0 fail | 22 pass / 0 fail |
| Merges into current main (`08fce29b`) | clean | clean |
| **Merge into each other** | **conflict in `headroom/cache/prefix_tracker.py`** | same |

Verified by trial merge. Both touch the same four functions:

- `overlay_cached_prefix`
- `normalize_message_cache_control`
- `_lineage_snapshot`
- `resolve_tracker`

### You agree on the root cause

Both descriptions independently land on the same diagnosis: `resolve_tracker` matches a conversation to its tracker by requiring the recorded messages to be a whole-**message** prefix of the incoming history, so a message that grows by **blocks** reads as diverged on every turn → fresh tracker → frozen prefix stuck at 0 → nothing for `overlay_cached_prefix` to replay → the forwarded bytes differ turn to turn and the prefix hash never matches. Two people arriving at that separately is good corroboration.

### You disagree on the breakpoint — and you're each right about a different shape

- **#2699:** leaves the breakpoint on the newest block, arguing that *"anchoring it backward would re-write an already-cached prefix and leave the appended blocks out of the cache entirely."*
- **#2702:** moves it to the end of the stable prefix, citing Anthropic's guidance for varying tails.

Both hold, for different histories:

| Shape | Is the current tail still there next turn? | Correct breakpoint |
|---|---|---|
| Blocks **appended**, earlier blocks stable (#2699's shape) | yes | **newest block** — the write covers the new tail |
| Last blocks **rewritten** every turn (#2702's shape) | no | **end of the stable run** — the newest block never repeats, so nothing ever matches |

Neither PR handles the other's shape.

### A specific bug in #2702's discriminator

`_breakpoint_index` tries to separate the two cases:

```python
run = _stable_leading_block_run(content, previous_blocks)
# "Only anchor backwards when the stable run is real (the message diverged
#  before its end) and covers most of the message."
if 1 <= run < len(content) and run * 2 >= len(content):
    return run - 1
```

The comment describes the right rule, but the guard compares `run` against **`len(content)`** — this turn's block count — rather than against `len(previous_blocks)`. A pure append satisfies it:

> Previous message has `N` stable blocks; `k` new blocks are appended (`k` = 1–5, per #2699's shape). Then `run == N` and `len(content) == N + k`.
> `run < len(content)` ✓, and `run * 2 >= len(content)` ⟺ `N >= k` — always true once the `_MIN_BLOCKS_FOR_RELOCATION = 20` gate is met.

So #2702 relocates the breakpoint on **append-only** histories too, excluding the freshly appended blocks from that turn's cache write. That is exactly #2699's objection, and it's why the two PRs read as contradictory.

The intent is right; one comparison implements it:

```python
if 1 <= run < len(previous_blocks) and run * 2 >= len(content):
```

- pure append → `run == len(previous_blocks)` → falls through to `newest` (#2699's shape, correct)
- rewritten tail → `run < len(previous_blocks)` → relocates (#2702's shape, correct)

With that, both of your claims are simultaneously true.

### Suggested convergence

1. **Take #2699's lineage fix as the base.** It's the more surgical of the two — an explicit `AppendOnlyClassification` / `block_frontier` threaded through `extract_cache_stable_delta` and `overlay_cached_prefix` — versus #2702's broader `_lineage_snapshot` rewrite.
2. **Layer #2702's breakpoint relocation on top, with the corrected discriminator**, and keep its `HEADROOM_STABLE_BOUNDARY_BREAKPOINT` kill switch. An env rollback for something on the cache-key path of every Anthropic request is good judgement, and #2699 has no equivalent.
3. Whoever lands second: **re-run CI after resolving the conflict.** Both currently passed against a base that does not contain the other.

### Note on test coverage

CI would not have caught the append-relocation issue: it needs a message of **≥20 blocks** that grows by **pure append**, and I don't believe either PR's tests construct that. Worth adding as part of the merge — it's the case that separates the two designs.

Caveat on my side: I reviewed this statically and from CI. I could not run the suite locally, and the `cache_read pinned at system+tools while cache_write grows` measurements in both descriptions are from your own logs and I haven't reproduced them.

JerrettDavis (changes requested):
The rewritten-tail lineage work is thoughtful, but the breakpoint discriminator currently regresses the pure-append shape, so this should not land as-is.

`_breakpoint_index` tests `run < len(content)`. For a pure block append, `run == len(previous_blocks)` and `len(content) == len(previous_blocks) + appended`, so that condition is true whenever anything was appended. With 20 previous blocks plus one new block, I reproduced `_breakpoint_index(...)=19` while the correct newest index is `20`. The relocation rewrites the already-cached prefix and excludes the newly appended block from this turn's cache write—the exact behavior #2699 correctly avoids.

The discriminator needs to establish that the prior message itself diverged, e.g. `run < len(previous_blocks)`, not merely that the current message has blocks after the stable run. Please add a >=20-block pure-append regression asserting newest-block placement. This PR also conflicts in `prefix_tracker.py` with #2699, which now has an independently verified append-only lineage/replay fix; converge the two implementations (or explicitly sequence/rebase and preserve both shapes) rather than allowing either merge order to overwrite the other.

The current head is otherwise green and merges cleanly with `main`; this request is specifically about the cache-write correctness of the latest `c1d81e42` source.

chopratejas:
Thank you for finding the missing lineage state and validating stable-boundary relocation against live traffic. I opened #2917 as a consolidated successor after the latest #2671 trace exposed an additional tool-profile collision and review of the two open approaches found that pure block appends need different breakpoint behavior.

#2917 carries forward this PR's conservative rewritten-tail requirements and kill switch, while separating:

- pure append: the prior blocks are an exact prefix, so replay them and advance to the newest block;
- rewritten tail: divergence occurs before the previous end and a fixed suffix remains, so anchor at the stable leading boundary;
- different provider cache-key profiles: never share a tracker even if messages are identical.

It also adds a real-handler wiring test and a provider-cache oracle showing the rewritten sequence converge to stable reads with zero repeated writes. I credited the root-cause work and live validation from this PR in #2917; review of the combined safety guards would be very welcome.

JerrettDavis:
Superseded by the cache-lineage and rewritten-tail breakpoint implementation now on main (including shared history-relation classification, affinity-aware/ambiguity-safe lineage matching, scalar-block fallback, and regression coverage in tests/test_issue_2671_block_growth_cache.py and tests/test_proxy_anthropic_cache_stability.py). Merging this older implementation would duplicate and weaken the guards now shipped, so closing rather than resolving conflicts.

PR #2704, opened 2026-08-01
feat(proxy): add an opt-in pre-forward context budget guard

### LINKED ISSUE #2649: [FEATURE] Pre-upstream context limit guard — hard-cap input tokens to avoid upstream 400 + client retry storm
## Problem Statement

Headroom compresses tokens before they reach the LLM, but has no **hard upper bound** on total request input tokens. On a long agentic session, input (system prompt + tool definitions + conversation history) can grow past the upstream model's context window. Headroom forwards the oversized request unchanged; the upstream returns a bare `400 input tokens exceed context limit`; and the client (Claude Code) retries the *same* oversized prompt — producing a burst of repeated 400s (attempt N/10) until it gives up.

The user-visible symptom is "dozens of 400 errors" in one session — really 1–2 root over-limit events amplified by client retries. Existing `--code-aware` / `--intercept-tool-results` reduce tokens but are best-effort, not a hard ceiling, so a long enough session still blows past the window.

## Proposed Solution

A configurable pre-upstream guard in the `/v1/messages` forwarding path, after the request body is finalized and before any forward:

1. Tokenize the request body (input tokens).
2. Compare against a per-model threshold (`context_limit − safety_margin`, optionally reserving `max_tokens`).
3. On exceed:
   - **truncate**: drop the oldest non-essential history until under threshold, inject a system note that history was trimmed; forward the trimmed request.
   - **reject**: return a structured 400 describing the overage, instead of passing the opaque upstream 400 through.

This turns an opaque upstream 400 + retry storm into a single deterministic local decision.

## Use Case

- Proxying Claude Code → a provider with a fixed context window (here: Step Fun `step-router-v1`, 256K / 262144 tokens). Long sessions (large files read, big tool outputs) accumulate past 256K.
- A hard cap at the proxy means the client never needs to know the upstream window, and an over-limit request never reaches the provider.
- Saves wasted/billable upstream calls (each retry is a full-size request that 400s) and eliminates the retry storm.

## Alternatives Considered

- **Client-side limit** (`CLAUDE_CODE_MAX_CONTEXT_TOKENS=200000`): works, but pushes the policy to every client and can't cover clients you don't control. (This is the workaround I run today, plus a watchdog that restarts headroom when `/livez` stops responding during these bursts.)
- **Rely on existing compression**: `--code-aware` / `--intercept-tool-results` help but aren't a guarantee — observed real sessions still exceed the window.
- **Upstream truncation**: the provider hard-validates `input ≤ 262144` and returns 400 with no truncation/fallback — no help upstream.

The guard is the missing "hard ceiling" layer and fits headroom's existing position (it already sees the full body and already mutates requests for compression).

## Example API (Optional)

Config-driven, **not** hardcoded — a hardcoded per-model table is fragile and can reject the very long sessions the proxy runs on (learned this the hard way):

```toml
[context_limit]
mode = "truncate"            # or "reject"
safety_margin = 12000

[context_limit.models]
"step-router-v1" = 262144
```

```python
# in the /v1/messages handler, after body is final, before forward
threshold = model_limit[model] - max(safety_margin, max_tokens + slack)
input_tokens = count_tokens(body)
if input_tokens <= threshold:
    forward(body)                                   # pass
elif mode == "truncate" and truncate_history(body, threshold)[1]:
    forward(inject_note(body))                      # truncated
else:
    return structured_400(overage=input_tokens - threshold)  # reject
```

## Notes

- I have a working tokenizer-based implementation of this guard (pass / truncate / reject, with atomic history truncation that preserves `tool_use`/`tool_result` pairs) and am happy to open a PR — but I'd like to align on design first (truncate-vs-reject default, config shape, model-limit sourcing).
- Deliberately flagging the footgun: a **hardcoded** model-limit table can reject legitimate long sessions (I hit this in practice — a too-low limit rejected the very sessions the proxy was serving). Prefer config-driven / usage-calibrated limits over a baked-in table.

rodboev:
## Latest branch refresh

Refreshed from upstream main at `e67b3c8a2`, retaining both context-budget settings and the current lossless-compaction setting in the configuration guide. The only conflict was documentation. All 38 context-budget policy/handler tests pass after the refresh. The shared npm installer pin is included to avoid the reproduced OpenClaw CI installation failure. New-head CI is running.

## Description

Proxying Claude Code at a provider with a fixed window (the issue's case is Step Fun `step-router-v1` at 262144) means long sessions eventually exceed it. The provider hard-validates and returns 400 with no truncation, the client retries the same oversized body, and the loop burns round trips.

This adds an opt-in budget guard at the point where the request body is final and about to be forwarded. It counts the finalized input, compares it against a context window the operator declared for that model, and either logs the overage or refuses locally before the request leaves the proxy.

Two things are deliberate. The limit comes only from operator configuration, `HEADROOM_MODEL_LIMITS` or `models.json`, never from `get_context_limit`'s inference chain: that chain falls back to LiteLLM's `max_tokens` (an output cap for many entries) and then to a flat 128000 for anything non-Claude, which would reject `step-router-v1` at half its real window. The guard cannot key on a number Headroom guessed. And the default mode is `observe`: it logs and forwards. Refusing requires setting `HEADROOM_CONTEXT_LIMIT_MODE=reject`, because the token count is a local estimate. Claude model IDs use the tokenizer registry's `o200k_base` proxy, while unknown IDs such as `step-router-v1` use the registry's generic estimator. That is the operator's call to make against their own declared number, not a default the proxy should impose.

With neither the limits config nor the mode set, which is every install today, nothing changes.

Closes #2649.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- add `headroom/proxy/context_budget_policy.py`: a pure-logic `evaluate()` plus `HEADROOM_CONTEXT_LIMIT_MODE` and `HEADROOM_CONTEXT_LIMIT_SAFETY_MARGIN` resolvers, with no I/O and no message inspection
- add `AnthropicProvider.get_operator_context_limit()`, which reads only operator-declared limits and returns `None` otherwise; `__init__` now keeps those separate from the built-in table before merging, so `get_context_limit` behaves exactly as before
- add one guard block at the pre-forward seam in the Anthropic messages handler: short-circuits on `x-headroom-bypass`, degrades to observe when a `context-1m` beta is forwarded without a matching declaration, logs an overage, and returns a local 400 only in `reject` mode
- document both env vars next to the existing model-limits configuration

The threshold is `declared_limit - max(safety_margin, max_tokens)`, following the formula in the issue. A non-positive threshold, where the reserved output alone consumes the declared window, counts as over budget: `reject` mode refuses the request locally and `observe` mode logs and forwards. A request that cannot fit by construction is never silently forwarded past the guard.

Truncation is not implemented. The existing pipeline leaves message history intact, and this guard preserves that behavior at the transport boundary. A separate change would be needed to introduce history truncation.

## Testing

### Test Output

```text
uv run pytest tests/test_proxy_context_budget.py -q
38 passed, 1 warning

uv run pytest tests/test_anthropic_pre_upstream_backpressure.py tests/test_anthropic_stage_timings.py tests/test_cold_start_fast_pass.py tests/test_proxy/test_anthropic_ccr_deferred_injection.py -q
46 passed, 1 warning

uv run ruff check headroom/proxy/context_budget_policy.py headroom/providers/anthropic.py headroom/proxy/handlers/anthropic.py tests/test_proxy_context_budget.py
All checks passed!

uv run ruff format --check headroom/proxy/context_budget_policy.py headroom/providers/anthropic.py headroom/proxy/handlers/anthropic.py tests/test_proxy_context_budget.py
4 files already formatted
```

## Real Behavior Proof

- Environment: Windows, FastAPI TestClient with a stubbed upstream transport.
- Exact command / steps: `uv run pytest tests/test_proxy_context_budget.py -q`; the production-route test posts an oversized finalized Anthropic request with an operator-declared 262144-token limit and reject mode enabled; the degenerate-threshold tests post a request whose `max_tokens` or configured safety margin equals or exceeds the declared limit in reject mode.
- Observed result: the proxy returns a local 400 and makes zero upstream calls both for a normal over-limit request and for a degenerate-threshold request; observe mode forwards the request and logs the declared limit, reserve, threshold, counted tokens, and overage.
- Not tested: live provider token accounting; client behavior after a local 400.

## Review Readiness

## Checklist

## Additional Notes

CHANGELOG updates are handled by release automation.

Scope is Anthropic `/v1/messages` on the initial forward. CCR and memory continuation bodies, plus the batch endpoint, remain outside this guard.

On the two open design questions from the issue: truncate-vs-reject default is answered as observe-by-default with opt-in reject and no truncation, for the reasons above; config shape reuses the existing `{"context_limits": {...}}` block in `HEADROOM_MODEL_LIMITS` / `models.json` rather than introducing a second one. A `[1m]`-suffixed model id can carry its own declared window, since `wrap --1m` uses that suffix to trigger the `context-1m` beta.


JerrettDavis (changes requested):
`evaluate()` fails open when the reserved output consumes the entire declared window (`threshold <= 0`). For example, `declared_limit=200_000`, `max_output_tokens=200_000`, `counted_tokens=1`, and `mode="reject"` returns `reason="degenerate_threshold"` with `should_reject=False`, so the handler forwards a request whose input + requested output cannot fit by construction. Larger safety margins produce the same bypass. This makes reject mode least effective for the most clearly impossible requests and contradicts the documented promise to refuse over-limit finalized requests before upstream. Please treat a non-positive threshold as over budget in reject mode (while observe mode may continue to log/forward), report a meaningful overage, and add policy plus handler regression tests for both `max_tokens >= declared_limit` and `safety_margin >= declared_limit`.

rodboev:
Thanks, agreed. The `threshold <= 0` branch in `context_budget_policy.evaluate()` treated a reserve that consumes the whole window as an inert, forward-it case, so reject mode could send a request the provider would reject by construction.

The rework makes that branch mode-aware: in reject mode a non-positive threshold is now over budget and returns the local 400 with zero upstream calls, and `overage` is no longer forced to 0 (it is now `counted_tokens - threshold`, the same formula the over-threshold branch uses). Observe mode keeps logging and forwarding. `reason` stays `degenerate_threshold` so the two states remain distinguishable.

The handler change is limited to logging: the pre-forward WARNING now also fires for `degenerate_threshold`, so an observe-mode degenerate request logs its overage instead of forwarding silently. The 400 path already keyed on `should_reject` and needed no change.

Regression coverage now includes both degenerate forms at both layers: `max_tokens >= declared_limit` and `safety_margin >= declared_limit` each get policy and handler tests, with reject mode asserting the local 400 and zero upstream calls and observe mode asserting forward plus WARNING.

JerrettDavis (approved):
Re-reviewed current head cb9101fe after the degenerate-threshold fix. Reject mode now treats both output reservation and safety margin consuming the declared window as impossible locally, reports a consistent positive overage, and makes zero upstream calls; observe mode remains intentionally non-blocking and observable. Policy and handler regressions cover the requested boundary cases, and substantive CI—including builds, full test shards, native wrappers, docs validation, and E2E—is green. The external Vercel authorization status is repository integration noise, not a code failure. No remaining blockers.

JerrettDavis (approved):
Reviewed the current context-budget guard, endpoint wiring and bypass/limit behavior against the earlier feedback. All 38 focused guard tests pass; no new blocking source finding.

JerrettDavis (approved):
Re-reviewed the refreshed head after merging current main. The PR-specific logic is unchanged from the previously approved revision, the complete hosted matrix is green, and the branch is current and conflict-free.

PR #2705, opened 2026-08-01
fix(proxy): make Responses memory continuations ZDR-safe

### LINKED ISSUE #2675: [BUG] does not work in enterprise environment with Zero Data Retention (ZDR)
## Description
Headroom proxy causes 400 Bad Request (unsupported_parameter: previous_response_id) in VS Code Copilot Chat during multi-step tool calls in a Zero Data Retention (ZDR) Enterprise environment. The proxy appears to mask ZDR negotiation headers, causing the VS Code client to incorrectly send previous_response_id when returning from a tool execution.

## To Reproduce
1. Install headroom with `uv tool install --python 3.13 "headroom-ai[all]"`
2. Run this command: `headroom proxy --port 8787`
3. In a new terminal, run: `$env:COPILOT_PROXY_URL = "http://localhost:8787"` and then `code .`
4. With Copilot configured in an Enterprise account with ZDR enabled, run any prompt that triggers a tool call (e.g., searching a workspace or reading a file).

## Expected Behavior
Tools are routed through the Headroom proxy and return compressed output back to Copilot without issue, gracefully handling the lack of server-side history.

## Actual Behavior
The tool continuation request fails with a `400 Bad Request (unsupported_parameter: previous_response_id)`, crashing the chat session.

## Error Output
```text
Sorry, your request failed. Please try again.
Client Request Id: [REDACTED]
Reason: Request Failed: 400 { "error": { "message": "Previous response cannot be used for this organization due to Zero Data Retention.", "type": "invalid_request_error", "param": "previous_response_id", "code": "unsupported_parameter" } }: Error: Request Failed: 400 { "error": { "message": "Previous response cannot be used for this organization due to Zero Data Retention.", "type": "invalid_request_error", "param": "previous_response_id", "code": "unsupported_parameter" } } at qG._provideLanguageModelResponse (c:\Program Files\Microsoft VS Code\4fe60c8b1c\resources\app\extensions\copilot\dist\extension.js:1690:14362) at process.processTicksAndRejections (node:internal/process/task_queues:104:5) at async qG.provideLanguageModelResponse (c:\Program Files\Microsoft VS Code\4fe60c8b1c\resources\app\extensions\copilot\dist\extension.js:1690:15327)
Environment
OS: Windows 11

LLM Provider: GitHub Copilot Enterprise (with ZDR enforced)

### PR DESCRIPTION
**Pull request by rodboev**
## Description

Headroom currently forces `store=true` and sends `previous_response_id` when continuing an injected Responses memory tool call. That continuation fails for ZDR organizations.

Closes #2675

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Made

- Keep the client’s explicit or omitted `store` field unchanged.
- Apply the same `store=false` gate to HTTP and WebSocket memory injection.
- Build the memory continuation from normalized input, every replayable first response output, and Headroom’s function outputs. Reasoning is replayed only when it carries encrypted content, and WebSocket continuations preserve the request’s model, tools, instructions, include, and store fields.
- Remove the rejected Codex responses-lite marker from generated WebSocket memory continuations while retaining unrelated client metadata.
- On the WebSocket path, transparent memory execution now happens only while no output frame has streamed to the client. A memory call that appears after the first visible output frame, and the remainder of that response, pass through to the client unexecuted, so a response is never both partially exposed and transparently executed by Headroom.
- Remove `previous_response_id` from this Headroom-owned continuation.
- Add production-path regressions covering HTTP and WebSocket memory continuations, reasoning-item replay, the late-memory-call pass-through lifecycle, and existing storage/auth preservation cases.

## Testing

Focused HTTP and WebSocket production-path tests, Ruff, format, and diff checks pass. The base/head comparison and live GitHub Copilot Enterprise ZDR round trip are not available in this environment.

- [ ] Type checking (`mypy headroom`) was not run
- [x] New tests added for the reported failure
- [ ] Manual testing was not available

### Test Output

```text
uv run pytest tests/test_openai_codex_routing.py tests/test_openai_codex_ws_lifecycle.py -q: 73 passed in 1.20s
uv run pytest tests/test_openai_codex_ws_lifecycle.py -q -k "test_ws_late_memory_call_after_streamed_message_passes_through or test_ws_memory_continuation_continues_pre_stream_and_passes_late_call or test_ws_memory_continuation_replays_history_without_previous_response_id or test_ws_memory_enabled_non_memory_response_streams_completion": 7 passed
uv run pytest tests/test_codex_ws_per_frame_memory.py tests/test_codex_ws_compression_scheduler.py tests/test_openai_codex_ws_timings.py -q: 17 passed, 1 skipped
uv run ruff check headroom/proxy/handlers/openai.py tests/test_openai_codex_routing.py tests/test_openai_codex_ws_lifecycle.py: All checks passed!
uv run ruff format --check headroom/proxy/handlers/openai.py tests/test_openai_codex_routing.py tests/test_openai_codex_ws_lifecycle.py: 3 files already formatted
git diff --check: pass
```

## Real Behavior Proof

The regressions execute the real OpenAI Responses HTTP and WebSocket handlers. The head uses stateless continuations, preserves encrypted reasoning items, omits `previous_response_id`, and strips the rejected responses-lite marker from generated WebSocket continuations while retaining unrelated metadata. A production-path WebSocket regression drives `message added/done -> memory function_call added/done -> response.completed` through the real handler and proves the client receives one coherent complete response, that no continuation is issued for a response whose frames were already exposed, and that the memory tool never executes for such a response. The base/head comparison was not obtained in this run. Official OpenAI documentation supports the storage and continuation contract.

- Environment: Windows, CPython 3.13, the Headroom proxy test environment.
- Exact command / steps: run the focused HTTP and WebSocket production-path tests listed in `Test Output`.
- Observed result: the handlers return the stateless continuation shape, preserve encrypted reasoning, omit `previous_response_id`, remove the rejected generated-frame marker, retain unrelated metadata, deliver a coherent complete WebSocket response when a memory call follows streamed output, and pass 73 focused tests.
- Not tested: a live GitHub Copilot Enterprise tenant with ZDR enabled; a base/head comparison.

## Review Readiness

- The diff is limited to the OpenAI Responses memory continuation and its tests.
- Explicit storage policy, ChatGPT routing, CCR, general WebSocket transport, and other providers are covered as preservation cases.

## Checklist

- [x] Closes #2675
- [x] Tests cover the reported failure and preservation cases
- [x] No `CHANGELOG.md` change
- [x] No tenant or user-agent ZDR heuristic

## Screenshots

Not applicable.

## Additional Notes

The live tenant remains a maintainer or reporter validation step. The base/head comparison remains unmeasured in this run.


rodboev:
Thanks for flagging the coverage gap. I extended the existing HTTP and WebSocket handler tests across the uncovered `include` preservation, frame-shape fail-open, ordinary non-memory completion, malformed tool-argument, and unavailable-backend paths. The generated WebSocket memory continuation now also applies the existing responses-lite metadata scrub before sending upstream, and the regression keeps unrelated client metadata intact. I also covered the stateless-output type guard, then refreshed the focused test and coverage results in the PR body.

JerrettDavis (changes requested):
The stateless replay direction is sound, but the new WS state machine has a client-visible late-tool corruption path.

Once a non-reasoning output item arrives, the code flushes `event_buffer` and streams the response normally. If a memory function call appears later, this block then flips `suppress_response = True` and clears only the now-local buffer:

```python
if decided and not suppress_response and event_type in {...}:
    ...
    if item is a memory function call:
        suppress_response = True
        event_buffer.clear()
```

Frames already sent to the client cannot be retracted. The remainder of that response, including its completion, is suppressed; Headroom executes the tool and sends a new continuation upstream. The client therefore sees the start/message portion of response A with no matching completion, followed by response B. That is exactly the orphaned-response condition the buffering design is meant to prevent.

Please choose one safe invariant:

- buffer the entire response until completion whenever transparent memory handling is enabled, then suppress/replay only if any memory call exists; or
- after any visible frame has been flushed, never switch to transparent suppression. Pass a later memory call and the rest of that response through for the client to handle normally.

Add a WS lifecycle regression with `message added/done -> memory function_call added/done -> response.completed`. It should prove the client receives one coherent complete response under the chosen policy and that Headroom does not both partially expose and transparently execute the same response.

The current branch is clean and CI is green; this is a source-level state-machine gap not exercised by the new tests.

rodboev:
Thanks, that trace is exactly right. The post-stream branch could retract a response that was already mid-flight: the client saw response A start with no completion, then response B.

I'm taking the second invariant you offered. Once any visible frame has flushed, the loop never switches to transparent suppression; a later memory call and the rest of that response pass through to the client unexecuted. The fix removes that toggle from `handle_openai_responses_ws` in `headroom/proxy/handlers/openai.py` (the `if decided and not suppress_response and event_type in {...}` branch). After removal, `suppress_response` can only become true while `not decided`, which by construction means nothing has reached the client yet: reasoning frames stay buffered, and the first non-reasoning `output_item.added` is what triggers the flush. Pre-stream calls still suppress and replay statelessly, unchanged.

The new regression, `test_ws_late_memory_call_after_streamed_message_passes_through`, runs `message added/done -> memory_search function_call added/done -> response.completed` through the real handler and asserts the client receives the full coherent stream, no continuation is issued, and the memory tool never executes for a response whose frames were already exposed. The repeated-call test now covers both halves of the policy: the first response (call before any output) continues statelessly, and the second (call after a streamed message) passes through.

Pre-stream transparent continuation, the standalone item history, the responses-lite metadata scrub, and the HTTP path are unchanged.

JerrettDavis (approved):
Re-reviewed current head efcfd6f2 after the requested WS lifecycle fix. Once a non-reasoning item makes the response visible, the state machine now remains in pass-through mode: late memory calls and the matching completion are forwarded coherently and are not transparently executed. Pre-stream memory calls still take the suppress/execute/continue path, and the new mixed lifecycle regression covers both transitions. All substantive GitHub checks are green; the only reported failure is the repository-wide Vercel authorization integration, unrelated to this source. No remaining blocking finding.

rodboev:
Superseded by https://github.com/headroomlabs-ai/headroom/pull/2997, which merged the complete Responses ZDR and stateless-continuation fix for issue #2675.

PR #2706, opened 2026-08-01
fix(ccr): tighten proactive-expansion eligibility

nangsontay:
## Description

Proactive CCR expansion is firing on contexts it should never consider, and once it fires the cost is permanent. This tightens *which tracked contexts are eligible* — three filters in the same `analyze_query` loop plus the tail guard that keeps an append from being replayed twice.

Supersedes the correctness half of #2648 (rebased onto current `main`, unchanged), and adds a tool-provenance filter on top. Refs #2186, which was closed with a workaround rather than a fix.

The two filters are one concern deliberately kept in one PR: both answer "may this tracked context be expanded into this request", they sit on adjacent lines of the same loop, and reviewing them apart hides the shared story.

### What goes wrong today

Expansion appends the retrieved original into the forwarded body. The client transcript never carries it, so the next turn looks identical to the tracker and the same hashes get expanded again. On agentic sessions that is 2–8k tokens per request, and the transient append also busts the provider prefix cache at the tail position on every turn.

Worse, the entries it picks are usually not tool output at all. Compressed instruction/system text (agent rules, injected reminders) is tracked with `tool_name=None` and matches generic queries easily. Re-expanding it is strictly loss-making: the append lands *alongside* the still-compressed copy, so the request carries both. Measured on real traffic, instruction blocks compress ~18%, so the pair exceeds the uncompressed baseline for the same content — worse than never compressing it. And because `overlay_cached_prefix` replays the mutated tail byte-identically, one firing keeps costing on every later turn of the conversation.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `SessionExpansionDedupTracker` (bounded LRU): a hash is proactively expanded at most once per session. The handler filters `analyze_query` recommendations through it and records hashes only after a real append. Explicit `headroom_retrieve` calls are unaffected.
- `ContextTracker` entries carry a session id, and `analyze_query` skips contexts tracked under a different session — content compressed before `/compact` must not resurface in the fresh session.
- `injection_target_already_forwarded()`: shared guard, used by both the proactive-expansion and memory-context append paths, that skips the append when the position the append helper will mutate already carries the bytes we forwarded there last turn. The overlay replayed those bytes including last turn's injection; appending again would double-inject and bust the cache from that point.
  - The guard asks about the **exact** index each helper mutates, not `messages[-1]`. Both index rules live in one finder each — `latest_user_chat_message_index()` (OpenAI scans backward for the latest *user* message, so an assistant prefill or trailing `role="tool"` tail makes the target an earlier position) and `latest_non_frozen_user_turn_index()` (Anthropic mutates only the final message, and only outside the frozen prefix). The append helpers use the same finders, so guard and mutation cannot disagree.
  - `position_already_forwarded()` performs the replay test on the post-overlay bytes: is this position `previous_forwarded[i]`, or does its content start with those blocks (the block-append merge shape). That covers both shapes `overlay_cached_prefix` actually produces — it replays the longest leading *run*, not an all-or-nothing full-prefix match — without re-deriving the overlay's trigger conditions.
- **New:** `analyze_query` skips contexts with no tool provenance (`tool_name is None`). Filtering happens on read, not on write — the entry stays tracked and retrievable on demand via `headroom_retrieve`, which bills only what the model actually asks for. Only the proxy-initiated guess is gated.
- `SessionExpansionDedupTracker.filter_new()` refreshes LRU recency, so a session that is only *queried* for a while cannot age out behind other sessions and have its already-seen hashes expanded a second time.
- `tests/test_injection_cache_safety.py` covers the three cache guards at unit level plus tripwires asserting both handlers keep them wired — including that each passes its own append target. It adds handler-shaped regressions driven through the real `overlay_cached_prefix` → finder → guard → append sequence: an Anthropic assistant-prefill tail and an OpenAI trailing-`tool` tail both keep exactly ONE copy of the already-injected user turn, while a genuinely fresh eligible user tail behind the same history still receives injection. Also covered: the block-append merge shape, the Anthropic frozen boundary, and the LRU read refresh. `TestToolProvenanceGate` covers the new filter, including that it does not consume a `max_proactive_expansions` slot and does not purge the entry.

`test_exact_substring_match_bonus` now tracks with `tool_name="Glob"`. It had been relying on non-tool eligibility to reach `analyze_query`; with provenance it still exercises the substring-bonus path it is named for.

## Testing

### Test Output

Re-run after the rebase onto current `main` and the review follow-up:

```text
$ pytest tests/test_injection_cache_safety.py tests/test_ccr_context_tracker.py \
         tests/test_proxy_handler_helpers.py tests/test_proxy_byte_faithful_forwarding.py \
         tests/test_cache/test_prefix_tracker.py tests/test_cache_prefix_overlay.py \
         tests/test_ccr_session_tracker.py -q
======================= 214 passed, 2 warnings in 13.56s =======================

$ pytest tests/ -q -k "ccr or proactive or expansion or cache_stability or prefix or injection or helpers or context_tracker or memory"
= 2272 passed, 40 skipped, 8771 deselected, 5495 warnings in 163.36s (0:02:43) =

$ ruff check headroom tests/test_injection_cache_safety.py
All checks passed!

$ ruff format headroom/cache/prefix_tracker.py headroom/proxy/helpers.py \
              headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py \
              headroom/proxy/ccr_session_tracker.py tests/test_injection_cache_safety.py
1 file reformatted, 5 files left unchanged

$ mypy headroom/cache/prefix_tracker.py headroom/proxy/helpers.py \
       headroom/proxy/ccr_session_tracker.py headroom/proxy/handlers/anthropic.py \
       headroom/proxy/handlers/openai.py --ignore-missing-imports
Success: no issues found in 5 source files
```

No failures remain in that selection. The two `tests/test_cli_proxy_env.py::TestCLICompressionOnlyFlags` failures reported on the pre-rebase branch no longer reproduce — they pass on current `main` under the same unmodified `HOME` (`4 passed in 0.29s`), so the test-isolation issue was resolved upstream in the meantime.

### A/B against `main`, same inputs, real proxy

- Environment: macOS 15 (arm64), Python 3.13.14. A proxy booted from this branch and, for the control, from unmodified `main` — `headroom proxy --mode token --backend anthropic --anthropic-api-url http://127.0.0.1:<stub>`, each with an isolated `HOME` so neither run could touch a developer's live logs/store, and `HF_HOME` pointed at the cached kompress weights so the lossy+CCR-marker path is actually exercised.
- Exact command / steps: three requests over one conversation, driven through the real HTTP path with `x-headroom-cwd: /tmp/proof` (workspace resolution is fail-closed without it) and `user-agent: claude-code/1.4.2`:
  1. a user turn carrying the real `~/.claude/rules/*.md` agent-rules text (~8.2k tokens) — still the newest message, so protected;
  2. `assistant` + a short user turn, so that block is no longer newest and becomes compressible → kompress emits a CCR marker → tracked with `tool_name=None`;
  3. a follow-up whose leading text block shares vocabulary with the rules text, pushing `analyze_query` over `relevance_threshold`.

  A stub upstream recorded every forwarded body and whether it contained a `headroom_proactive_expansion` block.

- Observed result: on identical input the filter flips the turn from a 3,635-token loss to an 886-token saving, and the expansion block never reaches the wire —

  | turn 3, identical input | `main` (no filter) | this branch |
  |---|---|---|
  | expansion fired | `f0cf4efb42373ec225f57725` (1417 items), twice in the run | **never** |
  | expansion block reached upstream | `has_expansion_block: true` | **`false`** |
  | `tok_before` → `tok_after` | 8245 → **11880** | 8245 → **7359** |
  | `tok_saved` | **0** | **886** |
  | forwarded body upstream | 54,130 B | **33,918 B** |

  So on `main` the request leaves 3,635 tokens *larger* than it arrived and reports `tok_saved=0`; with the filter the same request saves 886 tokens and sends 20,212 fewer bytes. The expanded entry was `tool_name=None` instruction text, i.e. exactly the class this filter excludes.

### Field data that identified the problem

From a live `--mode token` proxy on Claude Code traffic (`claude-opus-4-8`, session `5ff4cb7f21a8535f`): savings ramped 1105 → 1854 → 1948 → **1997** tokens/turn and held flat, then on the turn expansion fired (`msgs=28`) the delta inverted to **+2684** and stayed at +2684/+2643 for **19 consecutive turns** — the session never saved again. Inbound vs outbound bytes agree: last healthy turn 243,268 → 235,583 (−7,685 B), first broken turn 254,499 → 268,317 (+13,818 B). Both expanded hashes were `tool_name=None`; stored originals were 20,537 B against 16,870 B compressed, so compression gained 18% while expansion re-added 100% on top. Across four rotated logs, **148 of 150** CCR retrievals were proxy-initiated (99,307 items) against **2** agent-initiated (150 items).

- Not tested: the claim/release reservation is covered by a deterministic interleaved-call test (request A claims and pauses, request B attempts the same hash), not by a live multi-threaded concurrent-turn run against a real proxy — the handler holds no lock between calls, so interleaved calls *are* the race. The session filter and the injection guard are covered by unit and handler-shaped tests but were not isolated in the live A/B above — that run demonstrates the provenance filter, since a single-session scenario never exercises cross-session resurfacing. The A/B was also measured before the rebase onto current `main` and before the guard rework; the provenance filter it exercises is byte-identical since, but the numbers were not re-collected. Streaming responses were not exercised (the stub replies non-streaming). No multi-hour session was run, so the 300 s `max_context_age_seconds` boundary is untested end-to-end.

## Runtime Rollout Safety

- Rollout-managed feature(s): None — CCR proactive expansion is not registered in `headroom/rollout.py`; it is a plain config/env flag.
- Stable/default behavior changed: Yes — with `ccr_proactive_expansion` on (the default), fewer tracked contexts are now eligible (tool provenance, same session, exact-target guard) and a hash is proactively expanded at most once per session. Explicit `headroom_retrieve` calls are unchanged.
- Kill switch / disable path: `--no-ccr-proactive-expansion` / `HEADROOM_NO_CCR_PROACTIVE_EXPANSION=1` disables the whole proactive path; nothing here adds a new flag.
- Qualification impact: None — no qualification gates read these paths.
- Rollback path: Revert this PR; all state is in-process (`ContextTracker`, `SessionExpansionDedupTracker`), so there is no migration or persisted-format change.

## Review Readiness

## Checklist

## Additional Notes

- Supersedes the correctness half of #2648, which is now closed. The hot-path perf half of that PR (`defer_waste_signals`, `TokenCountMemo`) is deliberately **not** carried over: as review found, neither is reachable from production code, and wiring them is unwritten work rather than a rebase. It should return as its own PR with the CPU claims measured.
- #2707 and #2708 touch the same area but are independent of this PR — verified `conflicts=0` against this branch via `git merge-tree`, and they can merge in any order.
- Rebased onto current `main` (was 87 commits behind and conflicting). The only conflict was `headroom/cache/prefix_tracker.py`, where `main`'s stable-boundary breakpoint helpers (#2917) and this branch's new helper landed in the same spot — resolved by keeping both. `git merge-tree upstream/main HEAD` is now clean.
- Review follow-up in `fix(proxy): guard the exact index each append helper mutates`: exact-target-index guard, `position_already_forwarded()` replacing `is_append_only_extension()`, `filter_new()` LRU recency refresh, handler-shaped regressions for non-user tails, and realigned test docstring section labels.
- Review follow-up in `fix(ccr): make proactive-expansion dedup an atomic claim`, addressing both points of the second review:
  - **Check-then-act race.** `filter_new()` read the per-session seen-set and released the lock; the handler then executed expansions, checked the cache-mode and already-forwarded guards, appended, and only then recorded. Two concurrent requests for one session could both pass selection and both append the same hash. Selection is now `claim()`: filter and record under one lock hold, so a claimed hash is hidden from other requests immediately. The claim is the commit; a request that does not append hands the hashes back with `release()` from a `finally` covering every skip path — expansion returning nothing **or a subset**, cache-mode gating, the exact-target guard, an ineligible tail, and an exception raised inside the block. A dropped release costs one missed optimization; a dropped claim would cost a permanent per-turn tax, so the failure direction is deliberate.
  - **Content-addressed session ownership.** `ContextTracker._contexts` is keyed by hash alone, so two sessions compressing identical content share one entry, and a single `session_id` field let session B's re-track overwrite session A's ownership — A then lost eligibility for a context it still held. `CompressedContext` now carries `session_ids: frozenset[str]`, `track_compression` unions the new owner into the existing set, and `analyze_query` tests membership instead of equality. An empty set still means legacy/no-session tracking with workspace-only scoping, so no caller signature changed.
  - Tests: deterministic interleaving (exactly one of A/B proceeds), release-makes-eligible-again, partial expansion releasing only the unused subset, release of an unknown session, and same-content-in-two-sessions keeping both owners eligible. The interleaving assertions were also run against a local copy of the pre-fix `filter_new`/`record_injected` shape and fail there (`A=['hash1'] B=['hash1'] -> winners=2`), so they are a regression test rather than a restatement of the new code. The handler tripwire now asserts `.claim(` / `.release(`.
  - Also added the `Runtime Rollout Safety` section this template now requires; the governance comment above predates it.


JerrettDavis (changes requested):
`injection_target_already_forwarded()` assumes the mutation target is `messages[-1]`, but neither caller actually has that contract: Anthropic uses `_append_context_to_latest_non_frozen_user_turn()` and OpenAI uses `append_text_to_latest_user_chat_message()`, both of which scan backward for the latest eligible user message. This matters for valid assistant-prefill/history shapes. If the prior turn forwarded `[user]` (including an injected block) and the current append-only request is `[user, assistant]`, `target_index` becomes 1, `previous_forwarded` has length 1, and this guard returns false; the caller then appends the same context again to user index 0 even though overlay replayed that position byte-for-byte. The duplicate-injection/cache-bust bug therefore remains for non-user tail messages, and message-count-changing transforms can create similar index mismatches.

Please make the guard identify the exact index that the corresponding append helper will mutate (including Anthropic's frozen boundary), and compare that index against the replayed prefix. Add handler-level regressions for Anthropic assistant prefill and OpenAI histories whose newest message is non-user, proving the old injected user content is not appended twice while a genuinely fresh eligible user tail still receives injection.

nangsontay:
Addressed in `fix(proxy): guard the exact index each append helper mutates`, and the branch is rebased onto current `main` (was 87 behind and conflicting; `git merge-tree upstream/main HEAD` is now clean).

**@JerrettDavis — guard targets the exact mutated index.** You were right about OpenAI: `append_text_to_latest_user_chat_message()` scans backward, so on `[user, assistant]` / `[user, tool]` the target is index 0 while the old guard tested `len-1`, missed, and appended a second copy into the position the overlay had just replayed. Both index rules now live in one finder each — `latest_user_chat_message_index()` and `latest_non_frozen_user_turn_index()` (which carries the frozen boundary) — the append helpers were refactored onto those same finders, and callers pass the resulting index to the guard, so the two cannot drift apart. `target_index < 0` (no eligible target at all) makes the guard a no-op.

One correction on the Anthropic half: `_append_context_to_latest_non_frozen_user_turn()` does not scan backward — it only ever mutates `messages[-1]`, and returns the input unchanged when that is not a non-frozen user turn. So an assistant prefill there could not double-inject; the new test asserts that explicitly (helper returns the same object, replayed turn keeps one copy) rather than claiming a fix that was not needed.

**The replay test is exact now.** `is_append_only_extension()` is gone. As Copilot noted, its docstring was wrong about the overlay: `overlay_cached_prefix` replays the longest leading *run*, plus a block-append merge shape — not an all-or-nothing full-prefix match. `position_already_forwarded()` replaces it and asks the question directly of the post-overlay bytes: is this position `previous_forwarded[i]`, or does its content begin with those blocks. That covers both replay shapes without re-deriving the overlay's trigger conditions, and it is deliberately conservative in the safe direction (a coincidentally-identical position skips the append rather than risking a double).

**Also from Copilot:** `filter_new()` now refreshes LRU recency, so an actively-queried session cannot be evicted and have its already-seen hashes expanded twice; and the test docstring section labels are realigned with the module's (a)/(b)/(c) breakdown.

**New regressions**, driven through the real `overlay_cached_prefix` → finder → guard → append sequence: OpenAI trailing-`tool` tail (target is the earlier user message, guard fires, and the unguarded path is shown producing two copies), Anthropic assistant-prefill tail, a genuinely fresh user tail behind the same already-injected history still receiving injection, the block-append merge shape, the frozen boundary, and the LRU read refresh. Plus an OpenAI wiring tripwire alongside the existing Anthropic one, each asserting the handler passes its own append target.

`214 passed` on the directly-affected suites, `2272 passed / 0 failed` on the broad CCR+cache+injection+memory selection, ruff and mypy clean. The two `test_cli_proxy_env.py` failures reported earlier no longer reproduce — they pass on current `main` under the same unmodified `HOME`, so that isolation issue was fixed upstream in the meantime.

JerrettDavis (approved):
Re-reviewed the latest ba62fadc head against my prior exact-target-index concern. Both handlers now derive the same target index their append helper mutates, the post-overlay guard handles exact replay and block-append replay, the Anthropic frozen boundary is preserved, and the new handler-shaped regressions cover assistant-prefill/trailing-tool histories plus a genuinely fresh eligible turn. Full CI is green; the prior blocker is resolved.

JerrettDavis (approved):
Re-reviewed the post-approval head. The only change after ba62fadc is the clean merge from current main; the feature delta and exact-target injection guards are unchanged, the merge resolves without conflict, and CI on the refreshed head is green.

JerrettDavis (changes requested):
The eligibility filters and exact-target replay guard are well motivated, but the per-session dedup has a blocking check-then-act race. `filter_new()` only reads the session set and releases its lock; the handler then executes expansions, checks cache/injection guards, appends, and only afterwards calls `record_injected()`. Two concurrent requests for the same session can both pass `filter_new()` before either records and both append the same hash—the precise duplicate this PR is intended to prevent. Shared-session concurrent agent turns are a normal production path.

Please make selection an atomic reservation/claim operation. A claimed hash must be hidden from other requests immediately, then committed after a real append or released when expansion execution, cache-mode gating, target guarding, or append eligibility prevents injection. Add a deterministic interleaving test where request A pauses after claiming and request B attempts the same hash; exactly one may proceed, and a failed/skipped A must make the hash eligible again.

Please also cover the content-addressed/session ownership case while touching the tracker: `ContextTracker._contexts` is keyed only by hash, so tracking identical content in session B overwrites session A’s `CompressedContext.session_id`; session A then loses eligibility for its own still-valid context. Keying eligibility by `(workspace, session, hash)` or storing a set of owning sessions avoids making same-content parallel sessions evict each other. The provenance and exact-index guards otherwise look substantially improved.

JerrettDavis (approved):
The latest source resolves both remaining blockers. Proactive-expansion selection is now an atomic per-session claim: concurrent requests cannot select the same hash, and every path that does not append releases exactly the unused claim subset in `finally`. The deterministic overlap and release regressions cover the check/act race. Content-addressed contexts now retain a set of owning sessions, so tracking identical content in session B no longer removes session A’s eligibility while unrelated sessions remain excluded.

I merged current `main` into the maintainer-editable contributor branch; the only conflict was additive helper imports, retaining both this PR’s exact-target guards and `main`’s system-message relocation. `git diff --check`, focused Ruff, and 119 CCR/injection/context/handler tests pass locally. New exact-head CI is running.

PR #2707, opened 2026-08-01
refactor(ccr): read the expansion query through extract_user_query

nangsontay:
## Description

The CCR proactive-expansion path hand-rolled its own "most recent user question" scan instead of calling `headroom.utils.extract_user_query`, which every other consumer on this handler already uses. Two copies of the same read meant any fix to how the query is identified had to be made twice — and only one copy would get it.

No behaviour change intended for the normal case. This is groundwork: there is a real defect in how that query is identified (the helper does not strip injected `<system-reminder>` boilerplate, so under harnesses that inject context the "query" is the boilerplate rather than the user's question). Fixing that belongs in its own PR because it moves SmartCrusher's keep/drop scoring on every provider path and needs a before/after quality measurement. Collapsing to one call site first means that fix lands in one place.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: replace the inline 12-line reversed-scan in the proactive-expansion block with `extract_user_query(messages, latest_user_turn_only=True)`.
- `headroom/utils.py`: add `latest_user_turn_only` to `extract_user_query`.

The flag exists to preserve behaviour exactly. `extract_user_query` walks back to *earlier* user turns when the newest one yields no text; the expansion path must not. On a `tool_result` continuation turn the newest user turn carries no text at all, and scoring relevance against an older turn's question would expand context for a stale intent — i.e. without the flag this refactor would make expansion fire on turns where it previously stayed silent.

One deliberate difference remains: within the chosen turn the helper skips empty text blocks rather than committing to the first block it sees, so an empty leading text block no longer suppresses expansion outright. That is a no-op whenever the first text block has content, which is the normal case.

## Testing

### Test Output

```text
$ pytest tests/test_utils.py tests/test_proxy_ccr.py -q
======================== 25 passed, 3 warnings in 5.08s ========================

$ pytest tests/ -q -k "ccr or proactive or expansion or utils or memory_query or extract"
=== 2 failed, 782 passed, 22 skipped, 9409 deselected, 43 warnings in 43.49s ===

$ ruff check headroom/utils.py headroom/proxy/handlers/anthropic.py tests/test_utils.py
All checks passed!

$ ruff format --check headroom/utils.py headroom/proxy/handlers/anthropic.py tests/test_utils.py
3 files already formatted

$ mypy headroom/utils.py --ignore-missing-imports
```

The 2 failures are `tests/test_cli_proxy_env.py::TestCLICompressionOnlyFlags::{test_ccr_defaults_on,test_no_ccr_flag}` and are **pre-existing and unrelated** — they fail identically with this change stashed, and pass under a clean `HOME`. The real `~/.headroom/settings.json` leaks into those config tests.

Two new tests pin the flag: that it returns `""` on a `tool_result` continuation turn where the default walks back to `"the original question"`, and that it agrees with the default whenever the newest user turn does carry text.

## Real Behavior Proof

The contract for a refactor like this is "the proxy makes the same decisions". So it was verified by running the real proxy on both `main` and this branch and diffing every observable.

- Environment: macOS 15 (arm64), Python 3.13.14. `headroom proxy --mode token --backend anthropic --anthropic-api-url http://127.0.0.1:<stub>`, isolated `HOME` per run so neither could touch a developer's live logs/store, `HF_HOME` pointed at cached kompress weights so the lossy+CCR-marker path is exercised and expansion can actually arm.
- Exact command / steps: three requests over one conversation through the real HTTP path (`x-headroom-cwd: /tmp/proof`, `user-agent: claude-code/1.4.2`): a user turn carrying the real `~/.claude/rules/*.md` text (~8.2k tokens); then `assistant` + a short user turn so that block becomes compressible and gets tracked as a CCR entry; then a follow-up whose leading text block shares vocabulary with it so `analyze_query` scores above threshold. A stub upstream recorded every forwarded body. Identical request sequence against both builds.
- Observed result: identical on every axis — this branch resolves the same query and therefore reaches the same expansion decision.

  | observable | `main` | this branch |
  |---|---|---|
  | expanded hash | `f0cf4efb42373ec225f57725` | `f0cf4efb42373ec225f57725` |
  | items expanded | 1417 | 1417 |
  | expansion firings in run | 2 | 2 |
  | `tok_before` → `tok_after` per turn | 8170→9553, 8184→9567, 8245→11880 | 8170→9553, 8184→9567, 8245→11880 |
  | forwarded body bytes upstream | 43,507 / 43,637 / 54,130 | 43,507 / 43,637 / 54,130 |

  (The growth visible in those numbers is the pre-existing expansion behaviour on `main`, unchanged here by design — it is what #2706 addresses. This PR only proves it does not *move*.)

- Not tested: the `tool_result` continuation turn — the case `latest_user_turn_only` exists for — is covered by unit test rather than by the live run above, since the scenario needs a turn whose newest user message has no text block. The remaining intentional delta (empty leading text block) was likewise not exercised against real traffic; it cannot trigger while the first text block has content. Streaming responses were not exercised (the stub replies non-streaming), and the OpenAI/Gemini/batch call sites of `extract_user_query` were not run — this change adds a keyword-only parameter with a default that preserves their behaviour, but they were verified by test suite only.

## Review Readiness

## Checklist

## Additional Notes

- Independent of #2706 despite touching the same feature area — verified `conflicts=0` via `git merge-tree`; either can merge first.
- The follow-up this unblocks: `extract_user_query` does not strip injected `<system-reminder>` boilerplate, so under harnesses that inject context the resolved "query" is the boilerplate rather than the user's question. That defect is shared by every consumer of the helper (SmartCrusher relevance scoring on the Anthropic, OpenAI, Gemini and batch paths), which is exactly why it wants a dedicated PR with a before/after compression-quality measurement rather than riding along here.


JerrettDavis (approved):
Reviewed the latest source and the mainline merge result. The new opt-in helper mode correctly stops at the newest user turn for tool-result continuations while preserving existing default callers, and the Anthropic proactive-expansion path now uses the shared extraction logic without reviving stale intent. Focused utility tests and Ruff pass on the merged tree.

JerrettDavis (approved):
Reviewed the updated head (`0d17fad`) against the former inline scan and every helper behavior branch. `latest_user_turn_only=True` preserves the load-bearing stop at the newest user turn, so tool-result continuations cannot revive stale intent; the keyword-only default leaves all existing provider consumers unchanged. String/list content, whitespace, empty blocks, and no-user cases are handled consistently, with the one documented empty-leading-block improvement covered by tests. I found no blocking correctness issue. The two failing matrix shards are unrelated existing WebSocket timing and tokenizer-state failures.

JerrettDavis (approved):
Re-reviewed the latest merged head. No feature source follows the previously approved helper/handler implementation: latest_user_turn_only remains keyword-only and opt-in, the proactive CCR call preserves the newest-turn boundary, and existing consumers retain default behavior. Current main introduces no conflicting query-extraction contract, the branch is clean, and the diff has no whitespace errors. No blocking finding.

PR #2708, opened 2026-08-01
fix(savings): surface request growth the tok_saved clamp swallows

nangsontay:
## Description

`tokens_saved` is clamped at zero, so a request the proxy forwards **larger** than it arrived is indistinguishable in the PERF line from one it simply could not compress. Both read `tok_saved=0`.

That ambiguity hides real regressions. Anything that appends to the body after compression — proactive context expansion, memory injection — can outweigh the compression it sits on top of and still look like a neutral turn. On the session that prompted this, a request went from 55,161 tokens in to 57,845 out and reported `tok_saved=0`, for 19 consecutive turns, with nothing in the logs distinguishing it from a turn with nothing left to compress.

This reports the swallowed amount as `tok_inflated`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `RequestOutcome.tokens_inflated`: `max(0, optimized_tokens - original_tokens)`, derived from two counts the outcome already carries — **no new plumbing at any of the emit sites**.
- Added `tok_inflated=` to the PERF log line, next to `tok_saved=`.

Diagnostic only, deliberately. It does **not** feed `tokens_saved` or `attempted_input_tokens`, for two reasons:

1. `attempted_input_tokens = optimized_tokens + tokens_saved` is a *size*, not a signed delta. Letting the second term go negative makes it smaller than the bytes actually forwarded, corrupting the active-savings denominator.
2. Injection paths already book their own cost through the retrieval-drawback channel. A negative landing in `tokens_saved` as well would count the same loss twice.

So the clamp stays and the hidden number surfaces beside it. Worth noting there is already a revert-on-inflation guard *before* compression's own inflation can escape (`anthropic.py`, "Optimization inflated tokens … reverting to original messages") — it is only growth added *after* that point which the clamp was silently absorbing.

## Testing

### Test Output

```text
$ pytest tests/test_request_outcome.py tests/test_cli_perf_format.py -q
============================== 54 passed in 1.56s ==============================

$ pytest tests/ -q -k "outcome or perf or savings or stats"
======= 510 passed, 33 skipped, 9674 deselected, 542 warnings in 40.37s ========

$ ruff check headroom/proxy/outcome.py tests/test_request_outcome.py
All checks passed!

$ ruff format --check headroom/proxy/outcome.py tests/test_request_outcome.py
2 files already formatted

$ mypy headroom/proxy/outcome.py --ignore-missing-imports
```

Four new tests pin the distinction that was missing: shrank (0), no-op compression (0, and `tok_saved` also 0 — the two cases that used to look identical), grew (reports 2684 while `tok_saved` stays 0), and that `attempted_input_tokens` / `savings_pct` keep their unsigned semantics.

`tests/test_cli_perf_format.py` parses hand-written PERF fixtures by field name, so adding a field does not disturb it — verified green above.

### The field catching a real inflating request

- Environment: macOS 15 (arm64), Python 3.13.14. A proxy booted from this branch: `headroom proxy --mode token --backend anthropic --anthropic-api-url http://127.0.0.1:<stub>`, isolated `HOME` so the run could not touch a developer's live logs/store, `HF_HOME` pointed at cached kompress weights so the lossy+CCR-marker path is exercised and proactive expansion can actually arm.
- Exact command / steps: three requests over one conversation through the real HTTP path (`x-headroom-cwd: /tmp/proof`, `user-agent: claude-code/1.4.2`): a user turn carrying the real `~/.claude/rules/*.md` text (~8.2k tokens); then `assistant` + a short user turn so that block becomes compressible and gets tracked as a CCR entry; then a follow-up whose leading text block shares vocabulary with it, so proactive expansion fires and appends the original — which is how a request ends up leaving larger than it arrived. PERF lines read from the isolated `~/.headroom/logs/proxy.log`.
- Observed result: real PERF output from that run —

  ```text
  msgs=1 tok_before=8170 tok_after=9553  tok_saved=0 tok_inflated=1383 ... transforms=router:text_block:mixed
  msgs=3 tok_before=8184 tok_after=9567  tok_saved=0 tok_inflated=1383 ... transforms=router:text_block:mixed
  msgs=5 tok_before=8245 tok_after=11880 tok_saved=0 tok_inflated=3635 ... transforms=router:text_block:mixed
  ```

  Correlated from the same run: `CCR Tracker: Proactively expanded f0cf4efb42373ec225f57725 (1417 items)`, and the stub upstream confirms the block reached the wire (`has_expansion_block: true`, forwarded body 54,130 B on the third request).

  Every one of those turns reports `tok_saved=0`. Before this change that is all the log said, and it is the same thing it says when there was simply nothing left to compress. `tok_inflated=3635` is the number that was missing.

  For contrast, the same scenario run against a build where the request genuinely shrinks reported `tok_before=8245 tok_after=7359 tok_saved=886` — that build predates this field, so it does not print `tok_inflated`; the point is only that the inflating and shrinking cases are the two states the field has to separate, and on `main` today both render as `tok_saved=0` whenever the growth path is taken. The `tok_inflated=0` case on a shrinking request is covered by unit test.

### Scale of what was hidden

From a live `--mode token` proxy on Claude Code traffic across four rotated logs: **305 of 3,743 requests (8%)** had `tok_after > tok_before` while every one reported `tok_saved=0` — 192,829 tokens of growth rendered as "nothing to compress". The worst single session held +2,643/turn for 19 consecutive turns.

- Not tested: the `headroom perf` CLI was not run against a real log file containing the new field — it parses by field name and `tests/test_cli_perf_format.py` is green, but that is test-level rather than end-to-end evidence. Streaming responses were not exercised (the stub replies non-streaming), so the streaming emit path carries the new field on the strength of sharing `emit_request_outcome` rather than by observation. No dashboard or Prometheus consumer was re-run.

## Review Readiness

## Checklist

N/A — one added field on an existing log line; no user-facing surface.

## Additional Notes

- Independent of #2706 and #2707 — verified `conflicts=0` via `git merge-tree`; mergeable in any order.
- Related but deliberately out of scope: `main` has no producer for retrieval-cost accounting (`record_savings_event` takes no `kind`/`tokens_retrieved`, and nothing writes `tokens_retrieved` anywhere), so proactive expansion's cost is not booked into net savings at all. Adding that channel is a cross-cutting accounting change and belongs in its own PR; this one only makes the growth visible in the log.

### DIFF
--- a/headroom/proxy/outcome.py
+++ b/headroom/proxy/outcome.py
@@ -186,6 +186,28 @@ class RequestOutcome:
             return 0.0
         return self.tokens_saved / self.original_tokens * 100.0
 
+    @property
+    def tokens_inflated(self) -> int:
+        """Tokens the forwarded request grew by, if it ended up larger.
+
+        ``tokens_saved`` is clamped at zero, so a request that leaves the
+        proxy *bigger* than it arrived is indistinguishable from one the
+        proxy simply could not compress: both report ``tok_saved=0``. That
+        ambiguity hides real regressions — anything that adds to the body
+        after compression (proactive context expansion, memory injection)
+        can outweigh the compression it sits on top of and still look like
+        a neutral turn.
+
+        Report the swallowed amount alongside it so the two cases are
+        distinguishable. This is diagnostic only: it deliberately does not
+        feed ``tokens_saved`` or ``attempted_input_tokens``, because
+        ``attempted_input_tokens = optimized_tokens + tokens_saved`` is a
+        size, not a signed delta, and because injection paths already book
+        their own cost through the retrieval-drawback channel — letting a
+        negative land here too would count it twice.
+        """
+        return max(0, self.optimized_tokens - self.original_tokens)
+
     @classmethod
     def from_stream(
         cls,
@@ -493,6 +515,7 @@ async def emit_request_outcome(handler: Any, outcome: RequestOutcome) -> None:
         f"model={outcome.model} msgs={outcome.num_messages} "
         f"tok_before={outcome.original_tokens} tok_after={outcome.optimized_tokens} "
         f"tok_saved={outcome.tokens_saved} "
+        f"tok_inflated={outcome.tokens_inflated} "
         f"tool_saved={tool_saved} "
         f"cache_read={outcome.cache_read_tokens} cache_write={outcome.cache_write_tokens} "
         f"cache_hit_pct={outcome.cache_hit_pct} "
--- a/tests/test_request_outcome.py
+++ b/tests/test_request_outcome.py
@@ -431,6 +431,7 @@ async def test_funnel_emits_perf_log_with_canonical_shape(
     assert "tok_before=1000" in line
     assert "tok_after=300" in line
     assert "tok_saved=700" in line
+    assert "tok_inflated=0" in line
     assert "cache_read=200" in line
     assert "cache_write=100" in line
     assert "cache_hit_pct=67" in line  # 200/(200+100) * 100 = 67
@@ -651,3 +652,44 @@ def test_from_stream_threads_waste_signals_for_openai_via_backend_site() -> None
         waste_signals={"skipped_units": 3, "applied_units": 7},
     )
     assert o.waste_signals == {"skipped_units": 3, "applied_units": 7}
+
+
+# ── tokens_inflated: distinguishing "could not compress" from "grew" ──
+
+
+def test_tokens_inflated_is_zero_when_request_shrank() -> None:
+    """A normally-compressed request reports no inflation."""
+    o = _outcome(original_tokens=1000, optimized_tokens=300, tokens_saved=700)
+    assert o.tokens_inflated == 0
+
+
+def test_tokens_inflated_is_zero_when_compression_was_a_no_op() -> None:
+    """Nothing compressible is not the same as growth — both keep tok_saved=0."""
+    o = _outcome(original_tokens=1000, optimized_tokens=1000, tokens_saved=0)
+    assert o.tokens_inflated == 0
+    assert o.tokens_saved == 0
+
+
+def test_tokens_inflated_reports_growth_the_clamp_swallows() -> None:
+    """A request forwarded larger than it arrived is no longer indistinguishable.
+
+    tokens_saved stays clamped at 0 (its consumers treat it as a size, and
+    injection paths book their own cost separately), so the grown amount has
+    to surface as its own number or the regression is invisible.
+    """
+    o = _outcome(original_tokens=55161, optimized_tokens=57845, tokens_saved=0)
+    assert o.tokens_saved == 0
+    assert o.tokens_inflated == 2684
+
+
+def test_tokens_inflated_does_not_disturb_derived_sizes() -> None:
+    """attempted_input_tokens and savings_pct keep their unsigned semantics."""
+    o = _outcome(
+        original_tokens=55161,
+        optimized_tokens=57845,
+        tokens_saved=0,
+        attempted_input_tokens=57845,
+    )
+    assert o.attempted_input_tokens == 57845
+    assert o.savings_pct == 0.0
+    assert o.tokens_inflated == 2684


chopratejas:
## Review: approve — with two coverage gaps and one scale caveat

Good PR. The restraint is the best part of it, and I want to be specific about why before the nitpicks.

### The problem is real, and it shows up in production data

The log that prompted our recent `/v1/compress` and savings-reporting work carried **9 × `Optimization inflated tokens (63587 -> 63588), reverting to original messages`** — visible only because a *separate* warning fired. Growth added after the revert guard (proactive CCR expansion, memory injection) is invisible in the PERF line today, exactly as this PR describes.

The framing is also correct on the ordering: `anthropic.py`'s revert guard (~1656) runs *before* the final recount (~2485), so what the clamp absorbs is strictly post-guard growth. Verified.

### Not feeding `tokens_saved` / `attempted_input_tokens` is the right call

I checked the consumer, and this is load-bearing rather than stylistic. `prometheus_metrics.py:722` does:

```python
self.attempted_input_tokens_total += max(0, int(attempted_input_tokens))
```

…and `server.py` divides the savings rate by that total. A negative term would silently shrink the denominator and **inflate the reported savings percentage** — the opposite of the honesty this PR is after. A less careful version of this change would have "fixed" the clamp and corrupted the headline number. Reason 1 in the description is exactly right that `attempted_input_tokens` is a size, not a signed delta.

Also confirmed: merges clean with current `main` (including the `total_saved=` field), the resulting PERF line orders sensibly (`tok_saved` -> `tok_inflated` -> `tool_saved` -> `total_saved`), and `_parse_kv` in `perf/analyzer.py` splits arbitrary `k=v` so no log parser breaks on the new field. CI 22 pass / 0 fail.

The three unit tests are well chosen — they pin shrank / no-op / grew as three distinct states, and the **no-op** case is precisely the ambiguity the PR exists to remove.

---

### 1. Scale caveat on the OpenAI path

On the OpenAI path `RequestOutcome.optimized_tokens` is the **provider's** `usage.prompt_tokens`:

```python
total_input_tokens = _usage_int(usage.get("prompt_tokens")) or optimized_tokens
...
optimized_tokens=total_input_tokens,   # -> RequestOutcome
```

…while `original_tokens` is a **local tokenizer** count. So `optimized_tokens - original_tokens` mixes two scales there, and any provider counting *above* our estimator would print a bogus `tok_inflated` on every request. The Anthropic path is safe — both endpoints are `tokenizer.count_messages`.

I measured this rather than guessing, against a real 722-request opencode/chat-completions session:

```
after == before : 595
after <  before : 127
after >  before :   0      <- zero false positives
max(after - before) = 0
```

So it does not misfire in practice for that provider (nano-gpt counted at or below our estimate). But the margin is zero, so any provider counting above the local estimator flips it — Kimi is the known case where our estimator undercut by ~20%.

**Suggestion:** one docstring line noting the value is only trustworthy where both endpoints share a tokenizer. No code change needed.

### 2. Structurally blind on the Codex WS path

```python
original_tokens  = max(0, input_delta) + max(0, saved_delta),
optimized_tokens = max(0, input_delta),
```

`original >= optimized` by construction, so `tokens_inflated` is **always 0** on that path and can never report inflation even when it happens. Not a false positive — just uncovered. Fine as a follow-up, worth knowing.

### 3. Log-only, so the reach is narrower than the motivation

`tok_inflated` lands in the PERF line and nowhere else: not in `headroom perf` (`PERF_RECORD_FIELDS` is a fixed list and omits it), not in `/api/stats`, not on the dashboard. The motivating story is "19 consecutive turns with nothing in the logs distinguishing it" — but a number that lives only in raw logs still requires already suspecting the problem to go find it.

**Suggestion:** add `tok_inflated` to `PERF_RECORD_FIELDS` and `build_perf_summary` so `headroom perf` surfaces it. Few lines, and it's what makes this diagnostic actually get used.

---

None of the three block merging. (1) is a comment, (2) is a follow-up, (3) is small enough to fold in here if you'd like.

Disclosure: I reviewed statically and from CI — I could not run the suite locally.

chopratejas (approved):
Approving. Correct, well-scoped, and the restraint is the point: not feeding `tokens_saved`/`attempted_input_tokens` is load-bearing — a negative there would shrink the savings denominator in `prometheus_metrics.py` and *inflate* the reported rate. Full review above.

Three non-blocking follow-ups, in priority order:
1. Add `tok_inflated` to `PERF_RECORD_FIELDS` + `build_perf_summary` so `headroom perf` surfaces it — otherwise it lives only in raw logs, which undercuts the "19 turns and nothing in the logs" motivation.
2. Docstring note that the value is only trustworthy where both endpoints share a tokenizer (the OpenAI path takes `optimized_tokens` from provider `usage.prompt_tokens`). Measured 0 false positives on a real 722-request session, but the margin was exactly 0.
3. Codex WS path can't express inflation at all (`original = input_delta + saved_delta`, `optimized = input_delta`) — follow-up, not this PR.

JerrettDavis (changes requested):
The new signal is useful and the clamp/denominator invariants are handled correctly, but there is one correctness issue to address before this becomes an enterprise diagnostic.

On both OpenAI chat HTTP outcome sites, `original_tokens` is the local pre-transform tokenizer count while `optimized_tokens` is replaced with the provider's `usage.prompt_tokens` (`openai.py` around 3891/3957-3958 and 4232/4367-4368). `tokens_inflated = optimized_tokens - original_tokens` therefore compares different counting systems. A provider whose tokenizer reports above our estimator will emit `tok_inflated>0` for an unchanged or genuinely compressed request. That is especially problematic because the field is presented as actual post-compression body growth and is intended to identify regressions.

Please derive inflation from same-scale local before/after counts, or explicitly carry a same-scale inflation value/validity through `RequestOutcome` instead of deriving it from fields whose semantics vary by producer. Add a regression covering an OpenAI response where provider `prompt_tokens` exceeds the local original count while the local optimized request did not grow. The Anthropic same-tokenizer path and the new shrink/no-op/grow unit cases otherwise look sound; current CI is green.

PR #2710, opened 2026-08-01
perf(proxy): wire the token-mode hot-path optimisations into the Anthropic handler

nangsontay:
## Description

This carries the token-mode hot-path work from #2648 and adds the production wiring it was missing.

Review of #2648 found that both advertised optimisations were unreachable from production code, so the CPU reduction it claimed was inert while every test still passed. That finding was correct, and it was not a packaging artifact of that PR — the wiring had never been written anywhere:

- `TransformPipeline.apply()` accepts `defer_waste_signals`, but no handler ever passed it, and `_start_waste_signal_task()` was referenced only by tests. Waste-signal parsing stayed inline on the response path.
- `TokenCountMemo` / `count_messages_memoized()` had no production call site. Every recount site still called `tokenizer.count_messages(...)`.

#2648 has been closed; its correctness half is in #2706. This PR is the other half, now actually connected.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- Pass `defer_waste_signals=True` at the four response-path `pipeline.apply` sites and start the compensating background task, attaching it to the outcome as `waste_signals_task`.
- Route all thirteen full-transcript recount sites in the Anthropic handler through a per-session `TokenCountMemo` instead of raw `tokenizer.count_messages`.
- Add `tests/test_proxy/test_perf_hot_path_wiring.py` covering both wirings, so a helper with no production caller fails the suite.

Detail on each below.

**Waste-signal deferral.** The four response-path `pipeline.apply` sites — token fast-pass, token main, non-cache, cache-delta — now pass `defer_waste_signals=True`. When the pipeline hands back a provider instead of parsed signals, the handler starts the compensating background task and attaches it to the outcome as `waste_signals_task`, which the funnel already collects done-only. Inline capture is still attempted first, so a non-deferred result is unaffected.

The background-compression enqueue deliberately keeps inline detection: its result feeds the compression cache and never reaches `emit_request_outcome`, so deferring there would drop the signals rather than move them off the response path. This is the distinction the `proxy_pipeline_kwargs` note already described — the flag belongs at covered call sites, not in the shared kwargs dict that every provider handler consumes.

**Token-count memoisation.** The handler recounts the full transcript at up to six points per request (pre-hook snapshot, post-hook fold, the single-tokenizer consistency recount, and the token/cache-mode replay branches). Each raw count re-tokenises messages that have not changed since the previous turn. All thirteen recount sites now route through a per-session `TokenCountMemo` via a single local helper.

Exact, not approximate: `count_messages_memoized` verifies the tokenizer declares `ADDITIVE_COUNTS` and otherwise falls straight through to a plain `count_messages`, and the memo binds to the tokenizer instance so a mid-session swap resets it rather than serving counts computed under a different tokenizer. The memo is keyed per session, with the same eviction contract as `_get_compression_cache`.

**Tests.** `tests/test_proxy/test_perf_hot_path_wiring.py` pins the wiring itself — that the response-path sites defer, that not every site defers, that the compensating task is started and reaches the funnel, that inline capture is tried first, and that no raw `tokenizer.count_messages` survives in the handler. Plus behavioural coverage of `_start_waste_signal_task`: no provider → no task, provider → runs on the background runner and yields the signals dict, failing provider → swallowed. This exists so "helper ships with no caller" cannot regress silently again.

## Testing

### Test Output

```text
$ pytest tests/test_proxy/test_perf_hot_path_wiring.py -q
============================== 9 passed in 0.33s ===============================

$ pytest tests/test_token_count_memo.py tests/test_transforms/test_pipeline_waste_signal_limit.py \
         tests/test_proxy_handler_helpers.py tests/test_agent_savings.py -q
======================== 110 passed, 1 warning in 3.69s ========================

$ pytest tests/ -q -k "anthropic or outcome or pipeline or waste or memo or token_count or perf or wiring or savings or cache"
= 2644 passed, 79 skipped, 7545 deselected, 5401 warnings in 120.86s (0:02:00) =

$ ruff check headroom tests/test_proxy/test_perf_hot_path_wiring.py
All checks passed!

$ ruff format --check headroom/proxy/handlers/anthropic.py tests/test_proxy/test_perf_hot_path_wiring.py
2 files already formatted

$ mypy headroom/proxy/handlers/anthropic.py headroom/cache/token_count_memo.py --ignore-missing-imports
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), Python 3.13.14. Two checks — a proxy booted from this branch (`headroom proxy --mode token --backend anthropic --anthropic-api-url http://127.0.0.1:<stub>`, isolated `HOME` so the run could not touch a developer's live logs/store, `HF_HOME` pointed at cached kompress weights) to confirm the wiring is behaviour-neutral on a real request path, plus a direct measurement of the memo against the handler's real access pattern.
- Exact command / steps: for the proxy run, three requests over one conversation through the real HTTP path (`x-headroom-cwd: /tmp/proof`, `user-agent: claude-code/1.4.2`) carrying ~8.2k tokens of text, with a stub upstream recording every forwarded body; the same sequence was run against unmodified `main` for comparison. For the measurement, a counting wrapper around the real `EstimatingTokenCounter` over a transcript growing one turn per request, recounted six times per request as the handler does, comparing raw `count_messages` against `count_messages_memoized`.
- Observed result: the proxy run is byte-identical to `main` — `tok_before`/`tok_after` of 8170→9553, 8184→9567, 8245→11880 on both, same forwarded body sizes (43,507 / 43,637 / 54,130 B), same compression transforms, no errors — which is what a perf change should look like. The memo measurement over 30 requests: **5,580 per-message tokenisations before, 60 after (98.9% avoided)**, and per request at 60 messages it drops from 360 to 2, since only the two new messages need tokenising. Exactness was asserted in the same run: `raw=5033 memoized=5033 equal=True`.
- Not tested: no wall-clock CPU profile — the measurement counts tokenisation work avoided, which is the mechanism, not a latency number, and the proxy run above is too small to time meaningfully. The waste-signal deferral was verified by unit test and by source-level wiring assertions rather than by observing OTel counters from a live run, so the claim is "the task is started and handed to the funnel", not "telemetry was captured end-to-end under load". Streaming responses were not exercised (the stub replies non-streaming). The non-Anthropic handlers are untouched and keep inline detection by design.

## Review Readiness

## Checklist

## Additional Notes

- This is deliberately titled as a perf change and now actually delivers one; the earlier title over-claimed relative to what was wired, which is what review objected to.
- Independent of the three open PRs in this area — `conflicts=0` against #2706, #2707 and #2708 via `git merge-tree`, so it can merge in any order.
- The diff is large (40 files) because it carries the inherited hot-path work unchanged and adds the wiring on top. If you would rather review the wiring alone, I can rebase it onto #2648's original perf commit and open the wiring as a follow-up — say which you prefer.


JerrettDavis (changes requested):
The production wiring is now present, but I found two correctness blockers in the optimization primitives.

1. `TokenCountMemo.message_hash()` uses only the first 16 hex characters of MD5 (64 bits) as the sole cache key. Message content is client-controlled, and a collision returns the first message's token count for a different message, directly changing compression thresholds/count accounting. The `# nosec` annotation suppresses the warning but does not make collision-based cache aliasing safe. Please use a collision-resistant full digest (for example full SHA-256/BLAKE2) or store/verify the canonical serialized message alongside the hash. Add a regression that forces the digest/key collision path and proves unequal canonical messages are never served as the same cached entry.

2. The deferred waste provider closes over live mutable lists:

```python
deferred_input = waste_messages or messages
deferred_compressed = current_messages
```

The handler starts the task immediately after `pipeline.apply()` but continues processing/mutating the returned optimized messages (cache-control ownership, CCR/tool/system handling, overlays/extensions) before forwarding. Depending on executor scheduling, `_deferred_waste_parse()` can therefore observe either the pipeline-final snapshot or later handler mutations. That violates the stated parity with the old inline parse and makes telemetry nondeterministic. Snapshot both inputs when constructing the provider (a targeted deep copy or immutable normalized representation), then add a test that mutates the original/result lists after `apply()` and proves the deferred provider still parses the exact pipeline-time values.

The current head is also `DIRTY` against `main`; please resolve that after these fixes so the final handler integration can be reviewed on the actual mergeable tree. CI is currently green, but these paths are not covered by its existing assertions.

JerrettDavis (approved):
Re-reviewed after merging current main and resolving both correctness blockers. TokenCountMemo now uses full SHA-256 keys and verifies canonical message bytes within each digest bucket, with a forced-collision regression proving unequal messages cannot alias. Deferred waste parsing snapshots both inputs at pipeline return time, with a mutation-after-apply regression. Current handler/tool-token accounting is preserved. Validation: Ruff and format pass; 178 focused memo, pipeline, wiring, handler, outcome, cache-stability, backpressure, and timing tests pass.

JerrettDavis (approved):
Resolved against current main while preserving both original-message tracking and per-request backend routing. All 517 changed test files pass; Ruff and formatting pass; uv.lock matches main.

JerrettDavis (approved):
Re-approved on exact maintained head `c2b2c93d71969765ba3a9c15548b955f947dfeb9`. The #2852 conflict combined both intended optimizations: the consistency recounts stay off the event loop via `asyncio.to_thread`, and execute through this PR’s per-session `_count_msgs` memo rather than reverting to raw tokenizer calls. Local validation passed 123 focused hot-path/memo/pipeline/recount/handler/savings tests, Ruff check, Ruff format check, and diff check. The complete refreshed GitHub rollup is green, including smoke and all four test shards; the PR is mergeable and clean.

JerrettDavis (changes requested):
The production wiring is now present, but the new token memo has a blocking cross-request race. `count_messages_memoized()` calls `bind_or_reset(tokenizer)` once, then releases the memo lock and performs each `get()` / tokenizer count / `put()` under separate lock acquisitions. If two requests sharing the session memo use different tokenizer instances (for example, one request takes the count fail-open replacement while another is already counting), B can rebind and clear for tokenizer B, then A can insert tokenizer-A counts into the cache now marked as bound to B. B may subsequently consume those counts. That breaks the stated exactness guarantee and can feed a wrong total into compression/context-window decisions.

Please make binding plus the entire lookup/populate operation atomic with respect to tokenizer identity, or include a stable tokenizer identity/generation in every cache key and reject puts from an obsolete generation. Add a deterministic concurrency test that pauses A after binding, lets B rebind, resumes A’s put, and proves B never receives A’s counts. The per-session store makes this a real production interleaving, not only a unit-level concern.

The waste-signal deferral wiring and non-additive fallback otherwise look thoughtfully scoped. Given this PR also carries substantial inherited cache/prefix behavior beyond the headline wiring, please keep the regression focused on the exact cross-tokenizer race rather than weakening the tokenizer-swap guarantee.

JerrettDavis (approved):
The cross-tokenizer race is fixed on the latest source. Every memo read and write now revalidates tokenizer identity under the binding lock; after a concurrent rebind, stale reads miss and stale writes are dropped, so each request can only return counts produced by its own tokenizer. The deterministic two-thread test covers the reported bind/miss/rebind/populate/stale-put interleaving and the symmetric stale-read direction.

I also merged current `main` into the contributor branch with maintainer edits enabled. The conflict resolution preserves this PR’s CCR system-body mutation and resolved deferred waste signals while retaining `main`’s CCR ownership validation, SSE detection, savings attribution, and helper coverage. `git diff --check`, focused Ruff, and 96 memo/hot-path/pipeline/handler/cross-turn tests pass locally. New exact-head CI is running.

JerrettDavis (approved):
The exact-head shard failure was a test-fixture integration miss after current `main` introduced CCR marker ownership verification: the two new system-injection tests hand-authored a hash without seeding the compression store, so production correctly classified it as foreign and withheld retrieval instructions. I updated only the fixtures to seed and clean up the explicit hash; the production ownership guard remains unchanged. The complete affected cache-stability file now passes (30 tests), with focused Ruff, formatting, and diff checks green. Renewing approval on this corrected head.

Issue #2504, opened 2026-07-23
Deduplicate pricing lookup warnings for custom/unknown models


pgjh:
## Problem

When using Headroom proxy with a custom or unknown model (e.g. `glm-5.2` routed through an OpenAI-compatible provider), LiteLLM cannot resolve pricing for that model. This triggers a `WARNING` log on **every single request**:

```
WARNING - headroom.proxy - Failed to get pricing for model glm-5.2: litellm.BadRequestError: LLM Provider NOT provided. Pass in the LLM provider you are trying to call. You passed model=glm-5.2
```

In an active session this floods `proxy.log` with hundreds of identical warnings — one per API call. While `proxy.log` has rotation, the log still grows rapidly and makes it harder to spot genuinely useful warnings.

## Location

`headroom/proxy/cost.py`, line ~731:

```python
except Exception as e:
    logger.warning(f"Failed to get pricing for model {model}: {e}")
    return None
```

## Suggested Fix

Deduplicate the warning per model so it only fires once per unique model name per process lifetime:

```python
except Exception as e:
    if not getattr(logger, "_warned_models", set()):
        logger._warned_models = set()
    if model not in logger._warned_models:
        logger._warned_models.add(model)
        logger.warning(f"Failed to get pricing for model {model}: {e}")
    return None
```

This is a minimal, non-invasive change — no new dependencies, no config flags, just a `set` on the logger object to track which models have already been warned about.

## Environment

- Headroom version: 0.32.1
- Model: `glm-5.2` via OpenAI-compatible backend (`--backend anyllm --anyllm-provider openai`)
- LiteLLM: installed (cannot resolve custom model names, as expected)

## Impact

Low severity — does not affect functionality (pricing returns `None`, requests succeed). But the log spam is a real operational annoyance, especially for long-running proxy sessions.

Parideboy:
Fixed in #2505. Dedupes the pricing-lookup warning per unresolvable model (module-level set, one warning per model per process lifetime) instead of logging on every request.

abhay-codes07:
Opened #2535. It tracks already-warned models in a module-level set and warns once per process for both the pricing-failure and the LiteLLM-unavailable cases (kept the cost result unchanged at `None`). I used a module-level set rather than an attribute on the logger object so it doesn't rely on mutating the shared logger, but the effect is the same one-warning-per-model behavior you suggested.

PR #2725, opened 2026-08-02
fix(proxy/cost): mark estimated-basis budget records and add an enforcement policy (#2713)

### LINKED ISSUE #2713: [BUG] Budget enforcement silently uses token estimate when provider omits usage breakdown
## Description

`CostTracker.check_budget` is supposed to be a hard spend limit: once the configured budget is hit, requests get refused. But when a provider response comes back without an input-token breakdown (`uncached` / `cache_read` / `cache_write` all zero/absent), `record_tokens` substitutes headroom's **own token estimate** (`tokens_sent`) for the real input count and feeds that into the cost — and therefore into the budget — as if it were measured.

In `headroom/proxy/cost.py`, inside `CostTracker.record_tokens`:

```python
input_tokens = uncached_tokens
if not (uncached_tokens or cache_read_tokens or cache_write_tokens):
    input_tokens = tokens_sent   # headroom's own estimate, not provider-reported
```

Nothing on this path flags the figure as estimated: no field on the record, no log line, no distinction in the budget ledger. A budget decision made from an estimate is indistinguishable from one made from provider-reported usage.

The fallback itself is reasonable (you don't want input cost silently dropped from the budget when a breakdown is missing). The problem is that the resulting cost record is indistinguishable from a measured one, and the estimate can under-count just as easily as over-count — so a budget check can pass after real spend has already gone over.

## To Reproduce

```bash
git clone https://github.com/headroomlabs-ai/headroom
cd headroom
git checkout 28aa53d
pip install -e ".[proxy]"   # pulls litellm, which provides the pricing used below
```

Then save this as `repro.py` and run `python repro.py`:

```python
from headroom.proxy.cost import CostTracker

def booked_cost(tokens_sent, uncached=0, cache_read=0, cache_write=0, out=1000):
    t = CostTracker(budget_limit_usd=100.0, budget_period="daily")
    t.record_tokens(
        model="gpt-4o-mini",        # litellm-priced, so rates are deterministic
        tokens_saved=0,
        tokens_sent=tokens_sent,    # headroom's own count
        uncached_tokens=uncached,
        cache_read_tokens=cache_read,
        cache_write_tokens=cache_write,
        output_tokens=out,
    )
    return t.get_period_cost()

# NOTE: 50k / 30k are SYNTHETIC values chosen to make the basis difference visible.
# Same tracker, same output tokens, only the breakdown differs.

no_breakdown  = booked_cost(tokens_sent=50_000, uncached=0)          # estimate used as input basis
with_breakdown = booked_cost(tokens_sent=50_000, uncached=30_000)    # provider-reported basis

print(f"no breakdown   -> ${no_breakdown:.6f}")
print(f"with breakdown -> ${with_breakdown:.6f}")
print(f"ratio          -> {no_breakdown / with_breakdown:.2f}x")
assert no_breakdown != with_breakdown, "basis differs => estimate was substituted"
```

## Expected Behavior

When the provider omits a usage breakdown, the budget record should be marked as estimated (or at minimum logged), so that spend derived from headroom's estimate is separable from spend derived from provider-reported usage. A hard control like `check_budget` shouldn't act on a guess as if it were a measurement.

## Actual Behavior

The estimate is substituted silently and the cost is booked identically to a measured record. With the synthetic inputs above, the no-breakdown call books exactly `50,000 * input_rate + 1,000 * output_rate`, while the with-breakdown call books exactly `30,000 * input_rate + 1,000 * output_rate` — i.e. the estimate becomes the budget basis with no annotation. The numbers are constructed extremes for visibility; on real traffic the estimate is usually close to the real count, but it is unflagged and can drift in either direction.

## Code Sample

See the repro above — it drives the real `CostTracker.record_tokens` and `get_period_cost`; it does not reimplement any of headroom's arithmetic.

## Error Output

```
no breakdown   -> $0.008100
with breakdown -> $0.005100
ratio          -> 1.59x
```

(The assertion passes: the two bookings differ, confirming the estimate was substituted into the budget basis.)

## Environment

- **Headroom version**: main @ `28aa53d` (pyproject `0.33.0`)
- **Python version**: 3.13
- **OS**: macOS
- **LLM Provider**: provider-agnostic — triggers on any route whose response carries no input-token breakdown; `gpt-4o-mini` is used in the repro only because litellm prices it deterministically.

## Additional Context

This is an internal accounting-basis change in headroom's own budget ledger, not a provider billing issue. One way to make it actionable without changing the fallback behavior:

```python
if not (uncached_tokens or cache_read_tokens or cache_write_tokens):
    input_tokens = tokens_sent
    # mark and surface it, then let the operator decide the policy for estimated records
    logger.warning("budget basis estimated: no usage breakdown from provider for %s", model)
```

A `basis = "estimated"` field on the cost record would be enough to keep estimated and measured spend separable in the ledger.

Parideboy:
## Description

`CostTracker.check_budget()` is a hard spend control — the Anthropic handler refuses the request with a 429 once the period budget is gone. The ledger that control reads could not tell a measured dollar from a guessed one.

When a provider response carries no input-token breakdown, `record_tokens()` substitutes Headroom's own `tokens_sent` estimate for the input count so input cost isn't silently dropped from the budget. That fallback is the right call, but the resulting record was byte-identical to a provider-measured one: no field, no log line, no separation in `/stats`. `RequestOutcome.uncached_input_tokens` defaults to `0`, so any route whose response omits usage lands on this branch in production. An estimate can drift in either direction, so a budget check could pass after real spend had already gone over — with nothing saying the decision rested on an estimate.

This keeps the fallback and makes it visible, then lets operators decide what an estimate is allowed to do to a hard limit.

Closes #2713

## Type of Change

## Changes Made

- New `headroom/proxy/budget_basis_policy.py` (pure policy module, matching the existing `*_policy.py` convention): the `measured`/`estimated` basis constants, the `count`/`ignore`/`block` policy values, and `resolve_estimated_basis_policy()` (explicit value → `HEADROOM_BUDGET_ESTIMATED_BASIS` → `count`; an unknown value warns once and falls back rather than failing proxy startup).
- `headroom/proxy/cost.py`: ledger entries are now `CostEntry(timestamp, cost_usd, basis)` instead of a bare tuple; `record_tokens()` marks the fallback branch `estimated` and logs one WARNING per model (deduped the same way pricing warnings are, per #2504 — an unguarded warning on this path fires once per request for a provider that never reports usage); new `period_cost_breakdown()` and an optional `basis` filter on `get_period_cost()`; new `budget_denial_detail()` builds the 429 body where the ledger lives; `check_budget()` honors the policy while keeping its `(allowed, remaining)` signature.
- `stats()` gains `budget_estimated_basis` (the active policy) and `budget_basis` (the period split: `total_usd`, `measured_usd`, `estimated_usd`, `estimated_pct`, `records`, `estimated_records`). `merge_cost_stats()` already spreads `**cost_stats`, so both reach `/stats["cost"]` with no extra plumbing.
- Operator knob wired through every config layer: `ProxyConfig.budget_estimated_basis` (`models.py`), the Click `--budget-estimated-basis` option with `envvar=` (`cli/proxy.py`), the argparse `--budget-estimated-basis` flag (`server.py`, `default=None` so the env var stays reachable), and a `SettingField` in the `Budget` group (`settings_store.py`).
- `headroom/proxy/handlers/anthropic.py`: the 429 body now comes from `budget_denial_detail()`, which names how much of the period's spend was booked from an estimate and distinguishes "you overspent" from "I refuse to enforce a hard limit on a guess".
- `headroom/cli/doctor.py`: the budget check stays **PASS** and appends the estimated share (and the policy, when it isn't the default). No new WARN state — a provider that never reports usage would otherwise sit at a permanent WARN. Every new read is `.get()` + type-guarded so `doctor` still works against an older running proxy.
- `docs/content/docs/metrics.mdx`: a "Measured vs Estimated Spend" subsection with the `/stats` shape and the three policy values.
- Tests: new `tests/test_cost_budget_basis.py` (20 tests) plus 4 new `doctor` tests; `tests/test_anthropic_pre_upstream_backpressure.py`'s cost-tracker double gained `budget_denial_detail()` to match the handler's duck-typed contract.

### Policy values

| `HEADROOM_BUDGET_ESTIMATED_BASIS` | Effect on the hard limit |
|---|---|
| `count` (default) | Unchanged behavior — estimated spend consumes the budget. |
| `ignore` | Booked and reported, but only measured spend enforces. |
| `block` | Fail closed — refuse rather than enforce a hard limit on a guess. |

Default enforcement is unchanged. `CHANGELOG.md` is untouched.

## Testing

- [x] Unit tests pass (`pytest`) — every test covering the changed modules; see `Not tested` for this machine's pre-existing environment failures
- [x] Type checking passes (`mypy headroom`) — clean on every file this PR touches

### Test Output

```text
$ python -m pytest tests/test_cost_budget_basis.py tests/test_cost_tracker_counterfactual.py tests/test_cost_pricing_warning_dedup.py -q
31 passed

$ python -m pytest tests/test_cli_doctor.py -q
72 passed

$ python -m pytest tests/test_anthropic_pre_upstream_backpressure.py -q
25 passed

$ python -m pytest tests/test_proxy_settings_endpoints.py tests/test_proxy/test_settings_store.py -q
50 passed

# full suite (see "Not tested" below for the excluded modules and the pre-existing failures)
$ python -m pytest -q
...
tests\test_cost_budget_basis.py ....................                     [ 25%]
tests\test_cost_pricing_warning_dedup.py ...                             [ 25%]
tests\test_cost_tracker_counterfactual.py ........                       [ 25%]
...
217 failed, 8807 passed, 657 skipped, 5318 warnings, 59 errors in 716.30s (0:11:56)

# same failing files re-run on clean upstream/main with the change stashed -> identical count
$ git stash push -u -- headroom tests docs
$ python -m pytest tests/test_log_compressor.py tests/test_cache/test_client_integration.py \
    tests/test_fsutil.py tests/test_savings_ledger.py tests/test_ccr_mcp_http.py \
    tests/test_router_registry_dispatch.py tests/test_proxy_savings_history.py \
    tests/test_text_compressors.py tests/test_builtin_compressor_adapters.py \
    tests/test_cli_proxy_env.py -q
73 failed, 182 passed in 34.82s     # 40+16+2+2+1+1+1+4+3+3 = 73, matching the run above

# 12 errors, all in release_version.py / ccr/mcp_server.py / memory/mcp_server.py
# (stale local `mcp` stubs) — none in any file this PR touches

$ python -m ruff check headroom/proxy/budget_basis_policy.py headroom/proxy/cost.py headroom/proxy/server.py headroom/proxy/models.py headroom/proxy/handlers/anthropic.py headroom/cli/proxy.py headroom/cli/doctor.py headroom/settings_store.py tests/test_cost_budget_basis.py tests/test_cli_doctor.py tests/test_anthropic_pre_upstream_backpressure.py
All checks passed!

$ python -m ruff format --check <same 11 files>
11 files already formatted
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.11, branch `fix/budget-estimated-basis-2713` off `upstream/main` @ `232fb49c`, `PYTHONPATH` pointed at the working tree so the repo copy of `headroom` is imported rather than the installed one.
- Exact command / steps: ran the repro script from the issue body verbatim, then extended it to print `stats()["budget_basis"]` for both trackers, to construct the same tracker with `estimated_basis_policy="block"` and with `"ignore"`, and to record twice against the same model to check the warning dedup. Separately drove `headroom doctor`'s `check_budget` against stub `/stats` payloads (mixed basis, all-measured, non-default policy, and an older proxy that omits the new keys).
- Observed result: the issue's two figures are unchanged, so the fallback still works — no breakdown `$0.008100`, with breakdown `$0.005100`, ratio `1.59x`. The two are now separable: the no-breakdown tracker reports `{'total_usd': 0.0081, 'measured_usd': 0.0, 'estimated_usd': 0.0081, 'estimated_pct': 100.0, 'records': 1, 'estimated_records': 1}` and the with-breakdown tracker reports `estimated_usd: 0.0, estimated_pct: 0.0, estimated_records: 0`. One `WARNING headroom.proxy: budget basis estimated: no usage breakdown from provider for gpt-4o-mini — input cost booked from Headroom's own token count` fires across repeated records, not one per request. With `policy=block`, `check_budget()` returns `(False, 0.0)` and the 429 detail reads `Budget enforcement blocked for daily period: $0.0081 of $0.0081 was booked from Headroom's own token estimate because the provider returned no usage breakdown, and HEADROOM_BUDGET_ESTIMATED_BASIS=block refuses to enforce a budget on an estimate. Set it to 'count' or 'ignore' to serve these requests.` With `policy=ignore`, `check_budget()` returns `(True, 0.0001)` while the spend is still booked and reported (`0.7506`). `doctor` prints `pass $10.0/daily budget enforced — 62% of period spend ($1.2400) booked from Headroom token estimates`, appends `— estimated-basis policy: block` for a non-default policy, and degrades to the plain `$10.0/daily budget enforced` against a proxy that doesn't report the new fields. `--budget-estimated-basis [count|ignore|block]` shows in `headroom proxy --help`; the argparse path resolves the env var when the flag is absent and an explicit flag wins over the env.
- Not tested: no live end-to-end run against a real provider that omits usage in its response — the estimated basis was exercised through `record_tokens()` directly, which is the single funnel `emit_request_outcome()` uses. The `settings_store` field was not exercised through the settings UI. The full-suite run above excludes three things this machine cannot run, none of which touch the changed files: `tests/test_hermes_passthrough_compression.py` (`respx` not installed), `tests/test_memory/test_embedder_mps_serialization.py` (`sentence_transformers` pins `tokenizers<=0.23.0`, local has `0.23.1`), and `tests/test_cli/` (its subprocess-spawning tests wedge against a leftover local proxy on :8787; each file passes in isolation, e.g. `test_wrap_bridge.py` 7/7). Its 217 failures are all pre-existing environment breakage — a stale local Rust `_core` build (`test_log_compressor.py`, `test_text_compressors.py`, `test_builtin_compressor_adapters.py`, `test_cli_proxy_env.py`, the `test_transforms*` files) and the broken `sentence_transformers` install (`tests/test_memory/*`, `test_memory_system.py`, `test_sqlite_graph_store.py`) — with zero overlap with the modules this PR changes; the stashed baseline above reproduces them 1:1. CI is the authority for a green full suite.

## Review Readiness

## Checklist

- [x] New and existing unit tests pass locally with my changes — the only local failures are pre-existing and reproduce with the change stashed

## Additional Notes

The estimated-basis WARNING is deduped per model rather than emitted per request, following the precedent set by #2504 for pricing warnings — the whole point of this code path is that it fires on every request for a provider that never reports usage, so an unguarded `logger.warning` would flood `proxy.log`.

`headroom doctor` deliberately stays PASS. A WARN would be permanent, not actionable, for anyone whose provider simply doesn't report usage; the note tells them the number, and the `block` policy is there for operators who want the hard failure.

`check_budget()` keeps its `(allowed, remaining)` signature and its default `count` semantics, so `tests/test_cost_tracker_counterfactual.py` — including `test_budget_input_cost_counted_without_usage_breakdown`, the contract that the fallback keeps working — passes unmodified.

### DIFF
--- a/docs/content/docs/metrics.mdx
+++ b/docs/content/docs/metrics.mdx
@@ -252,6 +252,44 @@ headroom proxy --budget 10.00
 
 When the budget is exceeded, requests return a budget exceeded error, the `/stats` endpoint shows budget status, and logs indicate the budget state.
 
+### Measured vs Estimated Spend
+
+Every cost record carries a *basis* — where its input-token count came from. When a provider response includes a usage breakdown, the basis is `measured`. When it doesn't, Headroom substitutes its own `tokens_sent` count so input cost isn't dropped from the budget, and the record's basis is `estimated`. Headroom logs one warning per model the first time this happens.
+
+`/stats` keeps the two separable under `cost.budget_basis`:
+
+```json
+{
+  "cost": {
+    "budget_limit_usd": 10.0,
+    "budget_period": "daily",
+    "budget_estimated_basis": "count",
+    "budget_basis": {
+      "total_usd": 3.1400,
+      "measured_usd": 2.9000,
+      "estimated_usd": 0.2400,
+      "estimated_pct": 7.6,
+      "records": 412,
+      "estimated_records": 31
+    }
+  }
+}
+```
+
+An estimate can drift in either direction, so you choose what it does to the hard limit:
+
+```bash
+headroom proxy --budget 10.00 --budget-estimated-basis count   # default
+```
+
+| Value | Effect |
+|-------|--------|
+| `count` | Estimated spend consumes the budget like measured spend. The default; matches historical behavior. |
+| `ignore` | Estimated spend is still booked and reported, but only provider-reported spend consumes the budget. |
+| `block` | Refuse requests once the period holds any estimated spend, rather than enforcing a hard limit against a guess. |
+
+Env: `HEADROOM_BUDGET_ESTIMATED_BASIS`. `headroom doctor` reports the estimated share alongside the budget check.
+
 ## Key Metrics to Monitor
 
 | Metric | What It Tells You | Target |
--- a/headroom/cli/doctor.py
+++ b/headroom/cli/doctor.py
@@ -426,7 +426,36 @@ def check_budget(stats: dict[str, Any] | None) -> CheckResult:
             hint="set one: headroom proxy --budget 10 (env: HEADROOM_BUDGET)",
         )
     period = cost.get("budget_period", "daily")
-    return CheckResult(name=name, status=PASS, summary=f"${limit}/{period} budget enforced")
+    summary = f"${limit}/{period} budget enforced"
+    return CheckResult(name=name, status=PASS, summary=summary + _estimated_basis_note(cost))
+
+
+def _estimated_basis_note(cost: dict[str, Any]) -> str:
+    """Describe how much of the period's spend was booked from a token estimate.
+
+    Informational, never a WARN: a provider that simply never reports a usage
+    breakdown would otherwise sit at a permanent warning. Every read is
+    defensive so `doctor` still works against a proxy predating these fields.
+    """
+    note = ""
+
+    basis = cost.get("budget_basis")
+    if isinstance(basis, dict):
+        estimated_usd = basis.get("estimated_usd")
+        estimated_pct = basis.get("estimated_pct")
+        if isinstance(estimated_usd, (int, float)) and estimated_usd > 0:
+            pct = f"{estimated_pct:.0f}% " if isinstance(estimated_pct, (int, float)) else ""
+            note += (
+                f" — {pct}of period spend (${estimated_usd:.4f}) "
+                "booked from Headroom token estimates"
+            )
+
+    # Reported independently of the breakdown: a non-default policy changes how
+    # the budget is enforced and should surface even if the split is missing.
+    policy = cost.get("budget_estimated_basis")
+    if isinstance(policy, str) and policy and policy != "count":
+        note += f" — estimated-basis policy: {policy}"
+    return note
 
 
 def check_deployments(manifests: list[Any], probe: Any = probe_json) -> CheckResult | None:
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -534,6 +534,20 @@ def dashboard(port: int, no_open: bool) -> None:
         "Env: HEADROOM_BUDGET_PERIOD."
     ),
 )
+@click.option(
+    "--budget-estimated-basis",
+    type=click.Choice(["count", "ignore", "block"]),
+    default="count",
+    envvar="HEADROOM_BUDGET_ESTIMATED_BASIS",
+    help=(
+        "What to do with spend booked from Headroom's own token estimate, which "
+        "happens when a provider response carries no input-token breakdown. "
+        "count: it consumes the budget like measured spend (default). "
+        "ignore: only provider-reported spend consumes the budget. "
+        "block: refuse requests rather than enforce a hard limit on an estimate. "
+        "Env: HEADROOM_BUDGET_ESTIMATED_BASIS."
+    ),
+)
 # Code-aware compression (AST-based, requires `pip install headroom-ai[code]`).
 # Pair of flags so users can override the env-var default in either direction.
 # We resolve HEADROOM_CODE_AWARE_ENABLED in the body (not via Click's envvar=),
@@ -941,6 +955,7 @@ def proxy(
     codex_wire_debug_dir: str | None,
     budget: float | None,
     budget_period: str,
+    budget_estimated_basis: str,
     code_aware_flag: bool | None,
     disable_kompress: bool,
     disable_kompress_fallback: bool,
@@ -1235,6 +1250,7 @@ def proxy(
         or os.environ.get("HEADROOM_LOG_MESSAGES", "").lower() in ("true", "1", "yes", "on"),
         budget_limit_usd=budget,
         budget_period=cast(Literal["hourly", "daily", "monthly"], budget_period),
+        budget_estimated_basis=cast(Literal["count", "ignore", "block"], budget_estimated_basis),
         # Code-aware compression resolution:
         # 1. Explicit --code-aware / --no-code-aware always wins.
         # 2. Otherwise read HEADROOM_CODE_AWARE_ENABLED (truthy = on).
--- a/headroom/proxy/budget_basis_policy.py
+++ b/headroom/proxy/budget_basis_policy.py
@@ -1,0 +1,81 @@
+"""Policy for how estimate-based cost records feed budget enforcement.
+
+Every cost record the proxy books carries a *basis*: where the input-token
+count came from. ``measured`` means the provider reported a usage breakdown;
+``estimated`` means the response carried none and Headroom substituted its own
+``tokens_sent`` count so input cost wasn't silently dropped from the budget
+(#2713).
+
+The substitution is the right default — dropping input cost entirely would
+under-enforce far worse — but a budget is a hard control, and an estimate can
+drift in either direction. This module names the two bases and the three
+policies an operator can pick for what estimated spend does to the budget:
+
+* ``count``  — estimated spend counts toward the limit (the historical
+  behavior, and still the default; nothing changes except that the record is
+  now marked, logged, and separable in ``/stats``).
+* ``ignore`` — the record is still booked to the ledger and reported, but only
+  provider-measured spend consumes the budget. Fails open on the estimate.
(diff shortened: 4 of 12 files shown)


EvolveAegis:
Independently reported this bug in #2713. Our runtime evidence confirms the mechanism: when the provider's usage breakdown is all zeros, `cost.py` silently falls back to `tokens_sent` (headroom's own estimate) for `input_tokens`, and `check_budget` enforces against that estimate with no flag or marker. Our per-trial measurement showed the estimate drifts both ways (0.89 to 1.66x vs real usage), so the budget can silently over- or under-enforce. This PR's basis distinction (`measured` vs `estimated`) plus the operator policy choice (`count`/`ignore`/`block`) directly addresses both concerns: the silent fallback is now marked, and operators can decide how estimated spend interacts with the budget. That is the right shape for a hard-control-on-soft-estimate problem.

PR #2727, opened 2026-08-03
fix(proxy): cap decompressed request body size to prevent decompression-bomb DoS

Chukwuebuka-2003:
## Description

`_read_request_body_bytes` in `headroom/proxy/helpers.py` decompresses
`Content-Encoding` request bodies (zstd/gzip/deflate/br) with uncapped
one-shot calls: `gzip.decompress`, `zlib.decompress`, `brotli.decompress`,
and `stream_reader().read()`. A small compressed body can therefore expand
into an unbounded in-memory buffer and OOM the proxy process.

This PR decompresses incrementally in 64 KiB slices against a new
`MAX_DECOMPRESSED_BODY_BYTES` cap (defaults to `MAX_REQUEST_BODY_SIZE`,
100 MiB) and raises `RequestBodyTooLarge` — a `ValueError` subclass — the
moment the running total exceeds the cap. gzip now goes through zlib's
bounded `decompressobj` API instead of `gzip.decompress`.

Brotli (`br`) is **rejected outright** with a `ValueError`: the Python
`brotli` package's `Decompressor.process()` has no output-bounding API,
so a single 64 KiB input slice can materialize beyond the cap before our
check runs. Brotli-encoded requests that previously decompressed unbounded
will now fail with a clean `ValueError` instead.

Cap violations return the **configured body-too-large status** (413 by
default, overridable via `HEADROOM_PROXY_BODY_TOO_LARGE_STATUS`), not 400
`invalid_json`. The `/v1/compress` endpoint (both normal and bypass paths)
catches `RequestBodyTooLarge` explicitly to preserve this distinction.

No associated issue — bug-fix PR per CONTRIBUTING.md (no feature issue required).

## Type of Change

- [x] Bug fix (non-breaking change that fixes a bug)

## Changes Made

- `headroom/proxy/helpers.py`: bounded incremental decompression for zstd,
  gzip, and deflate; Brotli rejected (no output-bounded API); new
  `MAX_DECOMPRESSED_BODY_BYTES` constant and `RequestBodyTooLarge` exception
- `headroom/proxy/handlers/openai.py`: `handle_compress` catches
  `RequestBodyTooLarge` explicitly in both normal and bypass branches,
  returning the configured body-too-large status instead of 400
- `tests/test_proxy_request_decompression_cap.py`: 23 tests — round-trips
  for zstd/gzip/deflate, bomb rejection per format, at-cap accepted,
  corrupt/truncated/unsupported error paths, gzip multi-member accounting,
  trailing garbage, slice-boundary edge cases, `/v1/compress` integration
  tests for both normal and bypass paths

## Testing

### Test Output

```
$ python -m pytest tests/test_proxy_request_decompression_cap.py -v
23 passed, 3 skipped
```

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -533,8 +533,10 @@ class AnthropicHandlerMixin:
         from headroom.proxy.helpers import (
             MAX_MESSAGE_ARRAY_LENGTH,
             MAX_REQUEST_BODY_SIZE,
+            RequestBodyTooLarge,
             _get_image_compressor,
             compute_turn_id,
+            get_body_too_large_status,
             read_request_json_with_bytes,
         )
         from headroom.proxy.modes import is_cache_mode, is_token_mode
@@ -683,6 +685,18 @@ class AnthropicHandlerMixin:
             try:
                 async with stage_timer.measure("read_request_json"):
                     body, original_body_bytes = await read_request_json_with_bytes(request)
+            except RequestBodyTooLarge as e:
+                await _finalize_pre_upstream()
+                return JSONResponse(
+                    status_code=get_body_too_large_status(),
+                    content={
+                        "type": "error",
+                        "error": {
+                            "type": "request_too_large",
+                            "message": f"{e!s}",
+                        },
+                    },
+                )
             except (json.JSONDecodeError, ValueError) as e:
                 await _finalize_pre_upstream()
                 return JSONResponse(
@@ -3819,7 +3833,12 @@ class AnthropicHandlerMixin:
         from fastapi.responses import JSONResponse, Response
 
         from headroom.ccr import CCRToolInjector
-        from headroom.proxy.helpers import MAX_REQUEST_BODY_SIZE, _read_request_json
+        from headroom.proxy.helpers import (
+            MAX_REQUEST_BODY_SIZE,
+            RequestBodyTooLarge,
+            _read_request_json,
+            get_body_too_large_status,
+        )
         from headroom.proxy.modes import is_cache_mode
         from headroom.utils import extract_user_query
 
@@ -3843,6 +3862,17 @@ class AnthropicHandlerMixin:
         # Parse request
         try:
             body = await _read_request_json(request)
+        except RequestBodyTooLarge as e:
+            return JSONResponse(
+                status_code=get_body_too_large_status(),
+                content={
+                    "type": "error",
+                    "error": {
+                        "type": "request_too_large",
+                        "message": f"{e!s}",
+                    },
+                },
+            )
         except (json.JSONDecodeError, ValueError) as e:
             return JSONResponse(
                 status_code=400,
--- a/headroom/proxy/handlers/batch.py
+++ b/headroom/proxy/handlers/batch.py
@@ -53,7 +53,12 @@ class BatchHandlerMixin:
         from fastapi.responses import JSONResponse, Response
 
         from headroom.ccr import CCRToolInjector
-        from headroom.proxy.helpers import MAX_REQUEST_BODY_SIZE, _read_request_json
+        from headroom.proxy.helpers import (
+            MAX_REQUEST_BODY_SIZE,
+            RequestBodyTooLarge,
+            _read_request_json,
+            get_body_too_large_status,
+        )
         from headroom.utils import extract_user_query
 
         start_time = time.time()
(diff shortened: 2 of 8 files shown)


chopratejas:
Great fix. Can you address the lint failures?

Chukwuebuka-2003:
> Great fix. Can you address the lint failures?

Fixed:  ran `ruff format` on `headroom/proxy/helpers.py`; `ruff check` and `ruff format --check` both pass locally with the pinned version (0.15.17). CI will pick it up once the workflow run is approved.

JerrettDavis (changes requested):
Thanks for taking on this security boundary and for expanding the multi-member/corruption coverage. I found two blockers on the latest SHA:

1. The Brotli implementation is not actually memory-bounded. `Decompressor.process(data[i:i + 64 KiB])` returns all output produced from that input slice before `_enforce_decompression_cap` runs. A highly compressible stream smaller than 64 KiB can therefore materialize well beyond the configured cap in one call—the exact allocation this PR intends to prevent. Please use a Brotli strategy/API that bounds output before allocation (or isolate/reject the encoding when that guarantee cannot be made), and add a regression that demonstrates peak-producing input cannot cross the boundary inside one `process()` call.

2. `RequestBodyTooLarge` is currently a `ValueError`, so Anthropic/OpenAI translate it to `400 invalid_json`; the Bedrock adapter catches it under `Exception` and fail-opens by forwarding the compressed bomb verbatim. A size-policy rejection must be distinguished from malformed JSON: return the configured body-too-large status (normally 413), never fail open or forward on this exception, and cover the affected handler paths with integration tests. Please audit the batch best-effort catch for the same exception as well.

The zlib/zstd incremental approach, cumulative gzip-member accounting, and corruption tests otherwise look well structured. These changes are needed before the code can claim a process-wide decompression-bomb boundary.

Chukwuebuka-2003:
@JerrettDavis both blockers from your review are addressed on the latest SHA (`b4b3a02`):

**1. Brotli is now memory-bounded by rejecting the encoding.**

The Python brotli bindings (verified against brotli 1.2.0 and brotlicffi) expose no output-bounded streaming API — `Decompressor.process(input_slice)` returns all output produced from that slice before returning, so a highly compressible stream smaller than any feed slice can expand well past the cap inside a single call. Since that guarantee cannot be made, `br` request bodies are rejected outright with a clear client error (gzip/deflate/zstd are all bounded and supported), and the unsafe `_decompress_brotli_capped` helper is removed entirely so the boundary holds by construction — nothing is ever decompressed.

Regression test `test_brotli_peak_input_never_reaches_decompressor` demonstrates peak-producing input is refused before any decompression happens (and never surfaces as `RequestBodyTooLarge`, which would imply a decompressor ran).

**2. `RequestBodyTooLarge` is now a distinct size-policy rejection.**

Every handler path that reads a request body returns `get_body_too_large_status()` (default 413) with a `request_too_large` error code instead of the generic `400 invalid_json`:
- OpenAI: `/v1/chat/completions`, `/v1/responses`, compress bypass
- Anthropic: `/v1/messages`, batch create
- Gemini: generateContent, code-assist, streamGenerateContent, batch
- Batch: OpenAI create, Google create

The Bedrock adapter no longer catches it under `Exception` — it returns the configured 413 and never fail-opens or forwards the expanding body. The batch passthrough best-effort read re-raises it so the handler's 413 path sees it instead of swallowing it into a canonical re-serialization.

Integration tests in `test_proxy_decompression_cap_status.py` cover each affected handler path (including an assertion that the Bedrock forwarder is never invoked on a cap rejection), the configurable status override, and the batch best-effort audit. All 34 tests in the two decompression-cap files pass locally, along with the neighboring handler suites.

JerrettDavis (changes requested):
The two original blockers are substantially addressed (Brotli is rejected rather than unboundedly decoded, and the main provider/batch paths now preserve `RequestBodyTooLarge` as the configured status), but one request-body call site was missed. In `OpenAIHandlerMixin.handle_compress`, the non-bypass path still wraps `_read_request_json(request)` in `except Exception` and returns a generic 400. A gzip/zstd/deflate bomb sent to `/v1/compress` therefore loses the configured body-too-large status, while the bypass branch immediately above correctly catches `RequestBodyTooLarge`. Please add the same explicit catch to the normal branch and cover `/v1/compress` in the status integration tests. Also update the PR body before readiness: it still says Brotli is incrementally decompressed, all four formats round-trip, and existing callers remain 400 unchanged; current head instead rejects Brotli and intentionally returns 413/configured status for cap violations.

JerrettDavis (approved):
Re-reviewed the updated handler path and tests at the latest head. Preserving `RequestBodyTooLarge` through `/v1/compress` closes the remaining status-mapping gap without weakening the decompressed-byte cap, including on the bypass path. I also applied the two mechanical import-order fixes permitted by maintainer edits. Verification on the final head: Ruff passes for the changed files; `tests/test_proxy_request_decompression_cap.py` reports 23 passed and 3 skipped. No remaining code-review blockers found.

Chukwuebuka-2003:
> Re-reviewed the updated handler path and tests at the latest head. Preserving `RequestBodyTooLarge` through `/v1/compress` closes the remaining status-mapping gap without weakening the decompressed-byte cap, including on the bypass path. I also applied the two mechanical import-order fixes permitted by maintainer edits. Verification on the final head: Ruff passes for the changed files; `tests/test_proxy_request_decompression_cap.py` reports 23 passed and 3 skipped. No remaining code-review blockers found.

Thanks a lot 👍🏾

Chukwuebuka-2003:
@JerrettDavis can this be merged?

PR #2743, opened 2026-08-03
fix(compress): resolve the /v1/compress tokenizer per model, and document the real contract

chopratejas:
## Description

`/v1/compress` does no format conversion — callers send whichever wire shape they already use — but the pipeline pinned **one provider's token counter for the whole route**.

`OpenAITokenCounter.count_message` walks list content for `text` and `image_url` only and has **no else branch**, so Anthropic content blocks contributed literally zero. A 599-token `tool_result` scored 8. A request that really removed 235 characters reported `tokens_saved: 0` — so a caller gating on `tokens_saved > 0` concludes compression is broken while it is working.

Prompted by a Kong integration question ("do you support the Anthropic native format?"). The answer is that we already did — we just reported zeros for it, and the docs said otherwise.

Closes #

## Type of Change

- [x] Documentation update

### Tokenizer resolution (no hardcoded lists)

Build the derived pipelines with `provider=None` so `TransformPipeline` resolves the tokenizer from the **per-model registry**. Every registry tokenizer derives from `BaseTokenizer`, whose `_count_content_parts` ends in a serialize-and-count catch-all, which means:

- No block type counts as zero, and there is **no per-provider block-type list to keep in sync**. An enumerated set was the first thing I tried and it already missed `mcp_tool_result`, `web_search_tool_result`, `document`, and `thinking`.
- Gemini / Mistral / DeepSeek / Kimi stop defaulting to a tiktoken count when the registry already has a calibrated counter for them.
- Gateway aliases matching no vendor pattern still count correctly.

`mode="ccr"` now runs a derived pipeline too, for the same reason — sharing `openai_pipeline` pinned its provider. Costs that mode its own cold compression cache; correct metrics win.

### Tokenizer selection stays separate from context-limit resolution

Deliberately not welded together. `model_limit` feeds `context_pressure -> min_ratio`, so letting a tokenizer decision pick the limit table changes compression aggressiveness: `gpt-4-32k` answered by the Anthropic table is **8,192 instead of 32,768**, a 4× under-estimate. `test_tokenizer_choice_does_not_move_the_context_limit` pins the independence.

### Docs, rewritten from the code

- **`proxy.mdx`** — the loopback-only default and **404-not-403** behavior, previously undocumented *anywhere* in `docs/` despite shipping in #2458 explicitly for gateway sidecars; `HEADROOM_COMPRESS_ALLOW_REMOTE`; all four request fields; the whole `config` object including every `mode` value and `frozen_message_count`; `transforms_summary`; the 400/401/404/503 contract; and the timeout fail-open shape (`compression_skipped` / `skip_reason`).
- **Corrected "never calls an LLM"** — accurate about *generative* provider requests, misleading for a sidecar operator. Kompress (a ModernBERT **encoder**, classification not generation) and Magika run **in-process**, and `HEADROOM_KOMPRESS_ENDPOINT` offloads inference over HTTP — **real egress**. Now stated explicitly, with `HEADROOM_DISABLE_KOMPRESS=1` as the structural-only option.
- **Both wire formats documented as accepted**, and removed `anthropic-sdk.mdx`'s claim that OpenAI format is "the compression engine's native format" — the exact misconception that prompted this work. The SDK's conversion is now framed as an SDK choice, not an API requirement.
- **`litellm.mdx`** had no mention of the endpoint at all, despite the code naming LiteLLM's guardrail as its primary consumer. Added the HTTP deployment path, the `HEADROOM_COMPRESS_ALLOW_REMOTE` requirement, and why to leave `config.mode` unset.
- **`index.mdx`** printed `compressionRatio * 100` labelled "Saved …%", so a 77% saving displayed as **23%**. `api-reference.mdx` already defined it correctly, so the docs contradicted each other.
- `openai-sdk.mdx`, `wiki/proxy.md`, `wiki/typescript-sdk.md` — same corrections; dropped "any HTTP client", "Cloud", and a CacheAligner claim (it is detector-only).

## Testing

### Test Output

```text
$ .venv/bin/ruff check headroom/ tests/ --exclude headroom/dashboard/templates
All checks passed!

$ .venv/bin/mypy headroom/
Success: no issues found in 511 source files

$ python -m pytest tests/test_compress_route_tokenizer_by_model.py \
    tests/test_proxy_compress_endpoint.py tests/test_compress_api.py \
    tests/test_platform_stabilization_functional.py tests/test_proxy_eager_preload_bind.py -q
99 passed, 2 warnings in 47.15s
```

Broader sweep (`-k "compress or litellm or gateway or guardrail"`): **1625 passed, 4 failed** — all 4 pre-existing, verified by stashing this diff and re-running on clean `main` (2 strands hook tests, 1 codex WS semaphore-tail timing test, 1 unrelated local WIP test).

## Real Behavior Proof

- **Environment:** macOS 26.4 arm64, Python 3.12.6, repo `.venv`, branch rebased on `upstream/main`.

**(1) Before → after, same request** (60-line grep payload in an Anthropic `tool_result`):

| model | before | after |
| --- | --- | --- |
| `claude-sonnet-4-6` | `before=28 saved=0` | `before=1223 saved=58` |
| `bedrock/anthropic.claude-3-5-sonnet` | `saved=0` | `before=1037 saved=59` |
| `my-gateway/big-model` (alias) | `saved=0` | `before=1037 saved=59` |
| `gemini-2.5-pro` | `saved=0` | `before=1036 saved=59` |
| `gpt-4o` + OpenAI shape | `before=1225 saved=58` | `before=1225 saved=58` (unchanged) |

All three `config.mode` values verified for each. Response shape preserved: `type=tool_result`, `tool_use_id` intact.

**(2) Counter-level root cause**, 6.8 KB body, `count_message()`:

```text
OpenAITokenCounter    string-content -> 1406    tool_result block -> 5
registry (BaseTokenizer) claude       tool_result=408  thinking=418  mcp_tool_result=421
                                      web_search_tool_result=421  document=422
```

**(3) Every documented behavior asserted against the running app** — 13 checks, all PASS: 400s for missing `messages`/`model`, invalid `config.mode`, and all four invalid `frozen_message_count` forms; 200 for valid ones; non-dict `config` ignored; bypass and empty-messages omit `transforms_summary`; success returns exactly the 8 documented keys.

- **Not tested:** the docs site was not built (`docs/node_modules` absent) — MDX was checked for balanced `<Callout>` tags only, so a reviewer with the site running should eyeball rendering. No live gateway/Kong request; verification is via `TestClient` against the real ASGI app.
- **Note:** `HEADROOM_DISABLE_KOMPRESS` is read into `ProxyConfig` at `server.py:4919` and by the CLI, not by `create_app(ProxyConfig(...))` directly — I confirmed `disable_kompress=True` does reach the derived pipeline.

## Review Readiness

## Checklist

### DIFF
--- a/docs/content/docs/anthropic-sdk.mdx
+++ b/docs/content/docs/anthropic-sdk.mdx
@@ -41,7 +41,7 @@ Every call to `client.messages.create()` compresses messages first. The response
 
 `withHeadroom()` returns a proxy around your Anthropic client that intercepts `messages.create()`:
 
-1. Converts Anthropic-format messages to OpenAI format (the compression engine's native format)
+1. Converts Anthropic-format messages to OpenAI format
 2. Sends them to the Headroom proxy's `/v1/compress` endpoint
 3. Converts the compressed messages back to Anthropic format
 4. Forwards the request to Anthropic as normal
@@ -57,6 +57,10 @@ The adapter handles the full Anthropic message format including content blocks:
 | `{ type: "tool_result", tool_use_id, content }` | `{ role: "tool", tool_call_id, content }` |
 
 This conversion is lossless. Your request and response behave identically to an unwrapped client.
+
+<Callout type="info" title="The conversion is an SDK choice, not an API requirement">
+`POST /v1/compress` does no format conversion and compresses Anthropic content blocks natively — see [Message format](/docs/proxy#message-format). If you are calling the endpoint directly (from a gateway, or LiteLLM's `headroom` guardrail), send Anthropic-shaped messages as-is; you get the same shape back. Only this TypeScript adapter converts, because it normalises on OpenAI types internally.
+</Callout>
 
 ## Options
 
--- a/docs/content/docs/index.mdx
+++ b/docs/content/docs/index.mdx
@@ -17,7 +17,8 @@ const messages = [
 ];
 
 const result = await compress(messages, { model: 'gpt-4o' });
-console.log(`Saved ${result.tokensSaved} tokens (${(result.compressionRatio * 100).toFixed(0)}%)`);
+// compressionRatio is tokensAfter / tokensBefore, so savings is 1 - ratio.
+console.log(`Saved ${result.tokensSaved} tokens (${((1 - result.compressionRatio) * 100).toFixed(0)}%)`);
 ```
 </Tab>
 <Tab value="Python">
--- a/docs/content/docs/litellm.mdx
+++ b/docs/content/docs/litellm.mdx
@@ -114,4 +114,4 @@ Leave `config.mode` unset. The default pipeline is marker-free, which is what a 
 
 Because LiteLLM passes model names through, send the real one — `claude-sonnet-4-6`, `bedrock/anthropic.claude-3-5-sonnet`, `gemini-2.5-pro` — so Headroom resolves the right tokenizer and context limit. Anthropic-shaped messages need no conversion; see [Message format](/docs/proxy#message-format).
 
-For multi-turn agent loops, set `config.frozen_message_count` to the number of messages the provider has already cached so compression cannot rewrite the cached prefix.
+For multi-turn agent loops, set `config.frozen_message_count` to the number of messages the provider has already cached, **and send back the messages you previously forwarded rather than the pristine originals**. Getting this wrong silently destroys the provider's prefix cache — see [Multi-turn usage](/docs/proxy#multi-turn-usage-keeping-the-prefix-cache) for the loop.
--- a/docs/content/docs/litellm.mdx
+++ b/docs/content/docs/litellm.mdx
@@ -94,3 +94,24 @@ app.add_middleware(CompressionMiddleware)
 ```
 
 Response headers include `x-headroom-compressed: true` and `x-headroom-tokens-saved: 1234`.
+
+## Over HTTP (guardrail / gateway)
+
+The options above run Headroom **in** the LiteLLM process. If instead LiteLLM runs as its own proxy and you want it to call Headroom over the network — the guardrail deployment — point it at [`POST /v1/compress`](/docs/proxy#post-v1compress). LiteLLM swaps `messages` for the compressed result and forwards to the provider.
+
+Two things this deployment needs:
+
+```bash
+# Headroom is loopback-only by default and answers remote callers with 404.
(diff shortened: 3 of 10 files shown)


chopratejas:
## Added: multi-turn prefix-cache guidance (2a06e828)

Follow-up commit documenting how to actually *use* this endpoint in an agent loop. #2718 plumbed `config.frozen_message_count`, but the usage detail that makes it work was undocumented — and getting it wrong is silent.

**The trap:** the provider caches the bytes you **forwarded**, which compression already changed. `frozen_message_count` pins *whatever you hand it*, so feeding back pristine originals returns pristine originals — precisely the prefix the provider does not have. You pay for compression **and** a cache miss, and nothing in the response says so.

Compounding it: compression is not uniform over a conversation. An older tool result can fall outside the recent-read protection window as the conversation grows and be compressed harder than last turn — so re-compression is not guaranteed to reproduce earlier output either.

**Measured on this branch:**

```
naive: re-send pristine originals each turn
  1 turn  (3 msgs): msg[2] = 1534 chars
  2 turns (5 msgs): msg[2] = 1534 chars
  4 turns (9 msgs): msg[2] = 1389 chars   <- cached bytes changed
  8 turns (17 msgs): msg[2] = 1389 chars

documented loop: feed back previous output + frozen_message_count
  turn 2: cached prefix of 3 msgs byte-identical: True
  turn 3: cached prefix of 5 msgs byte-identical: True
  turn 4: cached prefix of 7 msgs byte-identical: True
  turn 8: cached prefix of 9 msgs byte-identical: True
```

Also documented that this endpoint runs **no CacheAligner and no prefix tracker**, unlike the proxy's own request path:

```
live request path : [CacheAligner, ContentRouter]
/v1/compress path : [ContentRouter]
```

And that `protect_recent` is not a substitute — it guards the newest messages; `frozen_message_count` guards the oldest, which is the cached end.

Two tests pin both halves so the guidance cannot rot. Note I deliberately did **not** assert 'compression always changes with conversation length' — that is conditional on crossing the protection window and can be masked by the result cache. The test asserts the robust invariant instead: the frozen prefix replays what you passed in, not what you forwarded.

Docs touched: `proxy.mdx` (new *Multi-turn usage* section + warning callout), `litellm.mdx` (cross-link where a gateway operator lands), `wiki/proxy.md`.

ruff + mypy clean (511 files); 97 compress tests pass.

PR #2756, opened 2026-08-04
fix(telemetry): stop mixing tokenizer scales in RequestOutcome, and fix the overhead framing

chopratejas:
## Description

Two defects found by reading real beacon payloads, not by inspection.

### 1. `eligible_pct: 120`

A gpt-4o-mini session shipped this:

```json
"tokens": {"original": 10, "attempted": 12, "input": 12, "saved": 0},
"rates":  {"eligible_pct": 120}
```

120% is structurally impossible — you cannot attempt to compress more than arrived. And nothing had grown.

`original_tokens` is our **local tokenizer** count. `optimized_tokens` on the OpenAI path carried the **provider's** `usage.prompt_tokens`. Our estimator undercounted gpt-4o-mini by 2 tokens on a 10-token request, and every quantity derived from that pair inherited the mismatch:

- `attempted_input_tokens = optimized + saved` → 12, exceeding `original` → `eligible_pct` 120, `yield_pct` contaminated
- `tokens_inflated` (added in #2708) → reported **2 tokens of phantom growth**

`optimized_tokens` was dual-purpose by design — *"post-compression bytes actually forwarded, for `input_tokens` and `tok_after`"*. Billing wants the provider's count; deltas need the same ruler as `original_tokens`. Those are different jobs sharing one field.

This is the same class of bug as #2743, on the request path instead of `/v1/compress`, and it is the exact false positive I flagged as theoretical when reviewing #2708 — where I measured the margin on a real 722-request log as **exactly zero**, so any provider counting above our estimator would flip it. gpt-4o-mini does, and it is now in production telemetry.

### 2. `overhead_pct` was documented as wall-clock, and is not

A 1393-turn session shipped `latency_ms_total: 16396565.8` against `duration_s: 8904` — **4.55h of latency inside a 2.47h session, 1.84× over wall clock.**

Both terms are sums over turns, and turns run concurrently (parallel tool calls, subagents, several clients per proxy), so each sum over-counts elapsed time. `overhead_pct` is still a meaningful latency-weighted per-request share, but the comment claiming *"what fraction of wall-clock did Headroom itself add?"* was wrong, and shipping `latency_ms_total` beside `duration_s` invites a comparison that yields nonsense.

Closes #

## Type of Change

## Changes Made

**Scale split (`outcome.py`, `handlers/openai.py`)**

- `optimized_tokens` is now always the **local** count — same tokenizer as `original_tokens`, so every delta built from the pair is coherent.
- New optional `provider_input_tokens` carries the provider's own count. Defaults to `0`, so the other emit sites need no change.
- Cost and volume totals read `provider_input_tokens or optimized_tokens`, so **billing is unchanged** wherever a provider reports usage, and falls back exactly as before where it doesn't.
- Removes a band-aid: one of the three OpenAI sites already computed `effective_original_tokens = max(original_tokens, optimized + saved)`, inflating `original` upward so `attempted` could not exceed it. That hid the symptom at one site while the other two shipped the impossible ratio.

**Overhead framing (`telemetry/session.py`)**

- Corrects the wall-clock comment on `overhead_pct` and states what it actually is.
- Documents that `latency_ms_total` is a sum of per-request durations, not elapsed time, and why it can exceed `session.duration_s`.
- Adds `overhead_ms_per_turn` and `latency_ms_per_turn` — unambiguous under concurrency and comparable across sessions.
- Field names kept for schema-v1 consumers; **additive only**, no schema bump.

## Testing

- [x] Unit tests pass (new file)
- [x] Linting passes (`ruff check` + `format --check`, pinned 0.15.17)

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/proxy/outcome.py headroom/proxy/handlers/openai.py headroom/telemetry/session.py
All checks passed!

$ uvx ruff@0.15.17 format --check <same three>
3 files already formatted

$ pytest tests/test_outcome_token_scale.py -q
5 passed in 0.24s
```

## Real Behavior Proof

- **Environment:** macOS 26.4 arm64, isolated git worktree off `upstream/main`, throwaway venv (see caveat).

Replayed both reported payloads through the corrected arithmetic:

```text
ENTRY 1 (gpt-4o-mini, the eligible_pct:120 case)
  BEFORE: original=10 (local)  optimized=12 (PROVIDER)  saved=0
          attempted = 12+0 = 12   eligible_pct = 120.0   <- impossible
          tok_inflated = max(0, 12-10) = 2               <- phantom
  AFTER : original=10 (local)  optimized=10 (local)  provider_input=12
          attempted = 10+0 = 10   eligible_pct = 100.0
          tok_inflated = 0
          billed_input = 12  -> cost/cache math unchanged

ENTRY 2 (1393 turns, the overhead case)
  latency_ms_total = 16397s vs duration 8904s -> 1.84x wall clock
  NEW overhead_ms_per_turn = 333.5   latency_ms_per_turn = 11770.7
  wall clock per turn      = 6392.0  -> per-turn latency exceeds it => concurrency
```

Genuine post-compression growth still surfaces: `55,161 → 57,845` reports `tokens_inflated = 2,684` (pinned as a test), so the fix doesn't mute what #2708 exists to show.

- **Not tested locally beyond the new file.** The repo venv currently has no `pytest`, no `ruff`, and no compiled `headroom._core`, so I used a throwaway venv. The outcome/telemetry suites fail there for missing deps — `click` (12/12), then `headroom._core` (10/10) — with **zero assertion failures**, and an **identical failure set on this branch and on clean `upstream/main`** in the same env. So they are environment-only, not regressions. CI on this PR is the authoritative signal for the full suite.
- Unrelated: `Wrap E2E / docker-wrap-e2e` is currently red on `main` from a quay.io CDN `tls: internal error` pulling the manylinux base image — infrastructure, transient, and green on the three prior runs.

## Review Readiness

## Checklist

## Follow-ups not in this PR

- **`cache_write` and `uncached` double-count** on inferred-cache providers. The same beacon entry shows `input: 12, cache_read: 0, cache_write: 12, uncached: 12` — both fields describe the identical 12 tokens, because `_infer_openai_cache_write_tokens` and `uncached_input_tokens` are computed the same way (`input − cache_read`). Any consumer summing them gets 2×. The payload also carries no `cache_inferred` flag, so a reader can't distinguish an inferred write from an Anthropic-reported one.
- **No skip reason for compression itself.** That entry records `memory_skip:no_handler` but nothing explains why compression didn't fire; "below the size floor" and "compressor failed" are indistinguishable — the same ambiguity #2708 just removed for inflation.
- **`eligible_pct` can still legitimately exceed 100** when a request genuinely grows after compression (memory injection, proactive expansion), since `attempted = optimized + saved` uses the forwarded size. Left alone deliberately: clamping would hide real inflation, and `tokens_inflated` now expresses it properly.

### DIFF
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -4015,10 +4015,15 @@ class OpenAIHandlerMixin:
                             provider=self.anthropic_backend.name,
                             model=model,
                             original_tokens=original_tokens,
-                            optimized_tokens=total_input_tokens,
+                            # Local count, same tokenizer as original_tokens, so
+                            # every delta built from the pair is coherent. The
+                            # provider's own number rides along separately for
+                            # billing — see RequestOutcome's field docs.
+                            optimized_tokens=optimized_tokens,
+                            provider_input_tokens=total_input_tokens,
                             output_tokens=output_tokens,
                             tokens_saved=tokens_saved,
-                            attempted_input_tokens=total_input_tokens + tokens_saved,
+                            attempted_input_tokens=optimized_tokens + tokens_saved,
                             cache_read_tokens=cache_read_tokens,
                             cache_write_tokens=cache_write_tokens,
                             uncached_input_tokens=uncached_input_tokens,
@@ -4425,10 +4430,12 @@ class OpenAIHandlerMixin:
                         model=model,
                         status_code=response.status_code,
                         original_tokens=original_tokens,
-                        optimized_tokens=total_input_tokens,
+                        # Same-tokenizer pair; provider count carried separately.
+                        optimized_tokens=optimized_tokens,
+                        provider_input_tokens=total_input_tokens,
                         output_tokens=output_tokens,
                         tokens_saved=tokens_saved,
-                        attempted_input_tokens=total_input_tokens + tokens_saved,
+                        attempted_input_tokens=optimized_tokens + tokens_saved,
                         cache_read_tokens=cache_read_tokens,
                         cache_write_tokens=cache_write_tokens,
                         uncached_input_tokens=uncached_input_tokens,
@@ -5417,13 +5424,15 @@ class OpenAIHandlerMixin:
                     )
                     uncached_input_tokens = max(0, total_input_tokens - cache_read_tokens)
 
-                    effective_optimized_tokens = (
-                        total_input_tokens if total_input_tokens > 0 else optimized_tokens
-                    )
-                    effective_original_tokens = max(
-                        original_tokens,
-                        effective_optimized_tokens + tokens_saved,
-                    )
+                    # Was: optimized := provider count, then original := max(original,
+                    # optimized + saved) to stop `attempted` exceeding `original`.
+                    # That paper over the symptom by inflating `original` upward,
+                    # which is why the beacon still shipped `eligible_pct > 100`
+                    # from the other emit sites that lacked the same fudge. The
+                    # real cause was mixing tokenizer scales; keep the local pair
+                    # coherent and hand the provider count over separately.
+                    effective_optimized_tokens = optimized_tokens
+                    effective_original_tokens = original_tokens
 
                     _resp_log_tags = {
                         **(tags or {}),
@@ -5446,6 +5455,7 @@ class OpenAIHandlerMixin:
                             status_code=response.status_code,
                             original_tokens=effective_original_tokens,
                             optimized_tokens=effective_optimized_tokens,
+                            provider_input_tokens=total_input_tokens,
                             output_tokens=output_tokens,
                             tokens_saved=tokens_saved,
                             attempted_input_tokens=attempted_input_tokens,
--- a/headroom/proxy/outcome.py
+++ b/headroom/proxy/outcome.py
@@ -57,8 +57,24 @@ class RequestOutcome:
 
     # ── Tokens (required — every site has these) ──────────────────────
     # original_tokens: pre-compression request size, for `tok_before`
-    # optimized_tokens: post-compression bytes actually forwarded, for
-    #     ``input_tokens`` and ``tok_after``
+    # optimized_tokens: post-compression size actually forwarded, for
+    #     ``tok_after``. MUST be counted with the SAME tokenizer as
+    #     ``original_tokens`` — every derived quantity is a delta between the
+    #     two (``tokens_saved``, ``tokens_inflated``, ``attempted_input_tokens``,
+    #     and the beacon's ``eligible_pct`` / ``yield_pct``), so mixing scales
+    #     silently corrupts all of them. Handlers used to pass the provider's
+    #     ``usage.prompt_tokens`` here because it also fed billing; that made
+    #     ``tok_after`` a provider count against a locally-estimated
+    #     ``tok_before``. On a gpt-4o-mini turn where our estimator undercounted
+    #     by 2 tokens that shipped as ``eligible_pct: 120`` — a structurally
+    #     impossible ratio — plus a phantom ``tok_inflated``. Provider-reported
+    #     input now lives in ``provider_input_tokens``.
+    # provider_input_tokens: the provider's own prompt-token count for this
+    #     request, when it reported one (0 otherwise). This is the billed
+    #     quantity, so cost and volume totals use it in preference to
+    #     ``optimized_tokens``. Kept separate precisely because it is on the
+    #     provider's tokenizer scale and must never be differenced against
+    #     ``original_tokens``.
     # output_tokens: response tokens from upstream
     # tokens_saved: original - optimized (or 0 if compression bypassed)
     # attempted_input_tokens: denominator for active-savings-percent.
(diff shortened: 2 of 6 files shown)


JerrettDavis (changes requested):
The field split is the right fix, but two downstream paths still use the wrong side of it, so the latest source does not yet satisfy its own payload/accounting proof.

1. `record_outcome(outcome)` runs before `billed_input_tokens` is selected, and `telemetry.session._fold` still does `sess.input_tokens += optimized_tokens`. For the documented 10-local / 12-provider example, the post-fix beacon will therefore emit `tokens.input=10`, not the claimed billed input `12`. If `tokens.input` is the volume/billing field described here, `_fold` must prefer `provider_input_tokens or optimized_tokens`, with an aggregator/payload regression—not only dataclass assertions.

2. `emit_request_outcome` now passes `billed_input_tokens` into `metrics.record_request` together with locally-counted `tokens_saved`. `PrometheusMetrics.record_request` feeds that pair into the savings tracker and reconstructs the durable ledger event as `tokens_before=input_tokens + tokens_saved`, `tokens_after=input_tokens` (around 850). That reintroduces the exact mixed-ruler delta this PR is meant to remove. Example: local 10→6 saved 4, provider reports 8; the ledger becomes 12→8 instead of 10→6. Cost/volume should use the provider count, but reduction/yield/ledger math needs the local optimized count (or explicit local original/after fields).

Please thread both semantics through the funnel rather than replacing one overloaded value with another overloaded value at `metrics.record_request`. Add an end-to-end funnel regression proving, for one mismatched provider count: billed/volume input uses 12, local delta fields use 10, and a saved case records the local before/after pair.

The three OpenAI emit-site changes and overhead documentation look sound. The new 5 tests and Ruff pass locally on `6c525d49`, but they do not exercise either affected consumer above. CI is still running and should remain a merge gate.

chopratejas:
@JerrettDavis both points were correct, and both are fixed in 18713f8b. Thank you — the field split was only half the job and I'd left each consumer reading whichever value happened to be there.

**1. `_fold` / `tokens.input`** — you're right that this would have shipped `tokens.input=10` for the documented example, contradicting the PR's own claim. And the reason it matters is sharper than the claim being wrong: `tokens.input` sits beside `output`/`cache_read`/`cache_write`/`uncached`, all provider-reported, and it is exactly what a reader sums the cache buckets against. Making it local would have put one local number in a dict of provider numbers. Now prefers `provider_input_tokens` with a local fallback.

**2. The ledger delta** — confirmed at `prometheus_metrics.py:851-852`. Pairing the billed input with a locally-counted `tokens_saved` reproduces precisely the bug this PR is meant to remove; local 10→6 with the provider at 8 recorded 12→8. `record_request` now takes `local_input_tokens` (default `None` → falls back to `input_tokens`, so every other caller is unchanged). Volume and cost keep the billed figure; anything deriving a delta uses the local one.

On your point about threading both semantics rather than swapping one overloaded value for another — that's the right framing and it's what I've done, rather than picking a winner at `record_request`.

**The regression you asked for**, driving the real code:

- `_fold` on one mismatched outcome: `tokens.input == 12` (billed) while `original_tokens == 10` and `tokens_saved == 4` stay local.
- the real `metrics.record_request` with `savings_ledger.record_savings_event` captured: ledger records **10 → 6**, plus an explicit assertion that it is *not* the 12 → 8 mixed shape.
- the paired fallback, omitting `local_input_tokens`, records **8** — so the two paths demonstrably differ rather than coinciding by accident.
- `_fold` with no provider count falls back to local.

Worth admitting: my first draft of the ledger test hand-computed the operands and asserted against a fake ledger, which proved nothing. Rewritten to call `record_request` directly.

`tests/test_request_outcome.py` + `tests/test_proxy_savings_history.py`: identical 4 failed / 73 passed on this branch and on clean `upstream/main` (pre-existing). ruff check + format clean. Agreed CI stays the merge gate — I'll wait for it rather than merge on green-so-far.

chopratejas:
Rebased onto current `main` (`0e1d6bfa`) and re-pushed as `ffd3d369`.

Worth noting why: the branch was **9 commits behind**, and those 9 are the tokenizer/pricing fixes from this same audit — #2758, #2759, #2760, #2761, #2762, #2776, #2778, #2779. Several touch the very surfaces this PR changes (provider tokenizers, cost basis, token units). The green CI on the old base had therefore never tested the combination, so merging on that signal would have been merging on a stale one.

Rebase was clean (no conflicts). Local verification on the new base:

```text
$ pytest tests/test_outcome_dual_ruler_funnel.py tests/test_outcome_token_scale.py \
         tests/test_request_outcome.py tests/test_proxy_savings_history.py \
         tests/test_savings_ledger.py tests/test_cost_budget_basis.py -q
125 passed, 2 skipped

$ pytest tests/test_*{pricing,cost,provider,token,netcost,outcome,savings,telemetry,prometheus,metric}*.py -q
3 failed, 1237 passed, 38 skipped
# all 3 pre-existing on main (2x deepseek needs `transformers`, 1x bundled_tools_savings)
```

@JerrettDavis — both of your points are in and unchanged by the rebase:

1. `telemetry/session.py:490` — `_fold` now prefers `provider_input_tokens`, so `tokens.input` stays on the provider ruler alongside the other provider-reported cache fields instead of silently becoming a local number.
2. `prometheus_metrics.py:693/703/866-867` — `record_request` takes `local_input_tokens` (default `None`, so other callers are untouched) and the ledger delta is built from it, rather than pairing billed input with a local `tokens_saved` and reproducing the mixed-ruler bug this PR exists to remove.

Merging at the repo owner's direction now that CI is re-running on the correct base. Happy to follow up in a separate PR if you want the naming taken further — the deeper issue your review pointed at is that these fields don't carry their ruler in the name, which is what makes cross-ruler subtraction look correct in the first place.

PR #2757, opened 2026-08-04
fix(cost): send litellm the total prompt so --budget stops seeing $0

chopratejas:
## Description

**`--budget` has been silently inert on any cache-warm request** — which is the normal case in an agent session. This is a disabled control, not a metrics bug.

`record_tokens` passed only the **uncached slice** as litellm's `prompt_tokens`. Measured, `litellm.cost_per_token` charges:

```
  (prompt_tokens - cache_read - cache_creation) * input_rate
+ cache_read     * read_rate
+ cache_creation * write_rate
```

So `prompt_tokens` is the **whole** prompt and litellm removes the cached parts itself. Handing it the uncached slice drives the input term **negative** as soon as anything is cached. `estimate_cost` ends with `float(total) if total > 0 else None`, so it returned `None`, no `CostEntry` was appended, and `check_budget()` saw **$0**.

| model | 100k prompt, 80k cached — old call | booked |
| --- | --- | --- |
| `gpt-5` | **-$0.065000** | None |
| `gpt-4o-mini` | **-$0.003000** | None |
| `claude-sonnet-4-5` | **-$0.156000** | None |

Closes #

## Type of Change

## Changes Made

Fixing the total alone would **over-charge OpenAI**, because two bugs are entangled here.

OpenAI exposes no cache-write counter, so `_infer_openai_cache_write_tokens` uses the uncached portion as a write proxy. At two of the four inference sites (`openai.py:4304`, `openai.py:5413`) `uncached_input_tokens` is derived by subtracting **only** `cache_read`, so `cache_write_tokens` and `uncached_tokens` are **the same tokens**. Summing all three would double-count the prompt and charge a write premium OpenAI does not have. (The other two sites — `openai.py:3969` and `streaming.py:2024` — subtract both, so their buckets are genuinely disjoint; those are left alone.)

- `cost.py` — `record_tokens` now passes `uncached + cache_read + cache_write` as the prompt total.
- `cost.py` — new `cache_inferred: bool = False` parameter. When set, the inferred write is excluded from **both** the prompt total and the write premium. The default preserves behaviour for every provider that reports disjoint buckets.
- `outcome.py` — plumbs `outcome.cache_inferred` through. The field already existed on `RequestOutcome` for the dashboard; it just never reached cost.
- `handlers/openai.py` — sets `cache_inferred=True` at the two outcome sites whose buckets genuinely duplicate.

## Testing

- [x] Unit tests pass (new file)
- [x] Linting passes (`ruff check` + `format --check`, pinned 0.15.17)
- [x] New tests added

### Test Output

```text
$ uvx ruff@0.15.17 check headroom/proxy/cost.py headroom/proxy/outcome.py headroom/proxy/handlers/openai.py
All checks passed!
$ uvx ruff@0.15.17 format --check <same three>
3 files already formatted

$ pytest tests/test_cost_budget_total_prompt.py -q
5 passed in 0.26s
```

The 5 new tests assert on the **arguments handed to `estimate_cost`** rather than on dollar values, so they pin the contract that broke without depending on litellm's pricing tables — or on litellm being installed at all.

## Real Behavior Proof

- **Environment:** macOS 26.4 arm64, isolated git worktree off `upstream/main`.

**litellm's actual formula**, established by probe rather than assumed:

```text
rates (claude-sonnet-4-5): input=3e-06 read=3e-07 write=3.75e-06
total=100k, 80k read, 5k write  ->  $0.087750
  hypothesis (p-r)*ir   + r*rr + w*wr = $0.102750   x
  hypothesis (p-r-w)*ir + r*rr + w*wr = $0.087750   <- matches
```

**Before → after:**

```text
Anthropic, disjoint: uncached=900 read=48000 write=1500
  OLD prompt=900     raw=-0.125775  booked=None   <-- BUDGET BLIND
  NEW prompt=50400   raw= 0.022725  booked=0.022725

OpenAI, inferred write == uncached: uncached=20000 read=80000 write=20000
  OLD prompt=20000                  raw=-0.090000  booked=None   <-- BUDGET BLIND
  NEW prompt=100000, write excluded raw= 0.035000  booked=0.035   truth=0.035000
```

The OpenAI "after" equals the hand-computed truth `20,000*input + 80,000*read_rate` exactly.

- **Not fully tested locally.** The cost/outcome suites show an **identical 10-failure set** on this branch and on clean `upstream/main` in the same throwaway env — all `ModuleNotFoundError: headroom._core`, the compiled Rust extension this machine cannot currently build. So they are environment-only, not regressions. One of them, `test_funnel_passes_canonical_record_tokens_shape`, covers the `record_tokens` call shape this PR changes, so **CI is the authoritative check for that one**.

## Review Readiness

## Checklist

## Context

Found during a tokenizer-consistency audit that also turned up: HuggingFace-routed models counting a 6,000-char message as **2 tokens**, `gpt-5`/`o4-mini` falling to a char estimator, and the Kompress size gate missing its own cap by 24%. Those are separate PRs — different subsystems, different risk.


JerrettDavis (dismissed):
The implementation direction is sound, but the branch currently breaks the repository's canonical outcome-funnel contract test.

On current head `5581dcbe`, this command:

`pytest tests/test_request_outcome.py tests/test_cost_budget_basis.py tests/test_cost_pricing_warning_dedup.py tests/test_cost_tracker_counterfactual.py tests/test_litellm_nonstream_cache_usage.py -q`

produces 75 passes and one failure: `tests/test_request_outcome.py::test_funnel_passes_canonical_record_tokens_shape`. `emit_request_outcome()` now correctly supplies `cache_inferred=False`, but the expected `record_tokens` kwargs were not updated to include that new canonical field. Please update that assertion (and keep the new focused tests); the branch's focused cost tests otherwise pass 5/5, Ruff passes, and `diff --check` passes.

chopratejas:
Thanks — this was exactly right, and it was the failure I had flagged in the PR body as "CI is the authoritative check here." Your repro command named it precisely.

Fixed in `497b5935`: `test_funnel_passes_canonical_record_tokens_shape` now includes `cache_inferred=False` in the expected `record_tokens` kwargs. The focused cost tests are untouched.

Your repro command on current head:

```
tests/test_request_outcome.py tests/test_cost_budget_basis.py tests/test_cost_pricing_warning_dedup.py
tests/test_cost_tracker_counterfactual.py tests/test_litellm_nonstream_cache_usage.py
-> 76 passed
```

CI `test (4)` is green (24/24 checks pass). Re-requesting review since the block is on the pre-fix commit.

Issue #2765, opened 2026-08-04
[FEATURE] ModelRoute: a condition on the request's own `max_tokens` (max_output_tokens)


ferrastas:
## Environment

- `headroom-ai` from `main` at `0cb72f45` (self-reports 0.33.0), pipx, Python 3.14.6, macOS.
- Feature: the opt-in cost-aware model router (#1706).

## Request

`ModelRoute` can match on `max_input_tokens`, `min_input_tokens`, `require_no_tools`
and `from_models`. There is no condition on the request's own `max_tokens`.

That value is the single best "this turn is trivial" signal available at routing time:
it is the caller's own declaration of how much output it expects. Unlike input size it
cannot be inflated by a long system prompt, and unlike tool presence it distinguishes a
one-word answer from an essay.

## Why input size alone is not enough

With this rule:

```json
[{"name":"trivial-opus->haiku","from_models":["claude-opus-4-8"],
  "max_input_tokens":4096,"require_no_tools":true,"to_model":"claude-haiku-4-5"}]
```

this request routes to Haiku:

```json
{"model":"claude-opus-4-8","max_tokens":4000,
 "messages":[{"role":"user","content":"Reply with the single word OK."}]}
```

The prompt is tiny, so `max_input_tokens` matches — but the caller asked for up to 4000
output tokens, i.e. it may well want a long, considered answer from the strong model.
There is currently no way to express "downshift only turns that are cheap on *both*
sides".

(Verified through `uvicorn headroom.proxy.server:create_app_from_env --factory`, since
the CLI entrypoint does not reach the router at all — that is #2764, filed separately.
`require_no_tools` and `from_models` both behaved as documented in the same run.)

## Proposed shape

A `max_output_tokens: int | None` field on `ModelRoute`, matching when the request's
`max_tokens` is `<= N`, following the existing conventions: `None` means "ignore this
condition", all set conditions AND together, first matching rule wins.

```json
[{"name":"trivial-opus->haiku","from_models":["claude-opus-4-8"],
  "max_input_tokens":4096,"max_output_tokens":512,
  "require_no_tools":true,"to_model":"claude-haiku-4-5"}]
```

Open question worth deciding explicitly: how to treat a request with **no** `max_tokens`
field. Anthropic's `/v1/messages` requires it, but the OpenAI paths (#2354) do not.
Treating "absent" as "does not match" is the conservative reading — an unbounded
response is not a trivial turn — and mirrors what a local implementation of this gate
would do.

## Context

[...]

PR #2395, opened 2026-07-18
feat(proxy/model-router): add shadow mode (HEADROOM_MODEL_ROUTER_LOG_ONLY)

abhay-codes07:
## Description

Adds a shadow mode to the cost-aware model router (#1706) so operators can validate their routes against real traffic before switching routing on for real.

Enabling routing today immediately rewrites the upstream model. That is a big first step to take on faith: an operator wants to see which requests each rule would catch (and confirm nothing important gets downgraded) before letting it change live traffic. Shadow mode provides exactly that dry run.

## How it works

Set `HEADROOM_MODEL_ROUTER_LOG_ONLY=1` (alongside the usual `HEADROOM_MODEL_ROUTER_ENABLED` / `HEADROOM_MODEL_ROUTES`). Rules are evaluated normally and the intended route is logged, but the outgoing model is left unchanged.

It is implemented entirely inside the pure `ModelRouter.select()` and needs no handler changes: in `log_only` mode a matched rule returns a `ModelDecision` whose `routed_model` equals the original model, so `changed` is `False` and every caller (the Anthropic and OpenAI routing helpers) already leaves the model untouched. The decision `reason` carries the shadow intent, e.g.

```
[shadow] would route gpt-5.4 -> gpt-5.4-mini via rule 'small' (input_tokens=1200, has_tools=False); log_only, model unchanged
```

which the handlers already log at INFO, so the would-be routing is fully observable.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `headroom/proxy/model_router.py`: add `ModelRouterConfig.log_only` (parsed from `HEADROOM_MODEL_ROUTER_LOG_ONLY` in `from_env`), a `ModelRouter.log_only` property, and the shadow branch in `select()` that returns a matched-but-unchanged decision with a `[shadow]` reason.
- `headroom/proxy/server.py`: pass `HEADROOM_MODEL_ROUTER_LOG_ONLY` into `from_env`.
- `tests/test_proxy/test_model_router.py`: cover env parsing, the default-off case, that a matched rule leaves the model unchanged in shadow mode (with the shadow reason), and that shadow off still routes.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
[...]

PR #2354, opened 2026-07-17
feat(proxy/openai): extend cost-aware model routing to the OpenAI paths

abhay-codes07:
## Description

Extends the opt-in cost-aware model router (#1706) to the OpenAI paths.

Today the router only rewrites the upstream model on the Anthropic `/v1/messages` path (`AnthropicHandlerMixin._maybe_route_model`); the CHANGELOG even notes it is "currently applied on the Anthropic `/v1/messages` path". OpenAI-compatible clients that route through `/v1/chat/completions` (GitHub Copilot CLI, opencode, OpenAI SDKs) or `/v1/responses` (Codex over HTTP) get no routing, so they cannot send small/tool-free turns to a cheaper model the way Anthropic clients can.

## Fix

Add `OpenAIHandlerMixin._maybe_route_openai_model`, the OpenAI counterpart of the Anthropic helper, and call it from `handle_openai_chat` and `handle_openai_responses` right after bypass detection (before compression, so the routed model drives tokenizer/limit selection):

- Reuses the shared `ModelRouter` and `estimate_input_tokens` and the same `HEADROOM_MODEL_ROUTER_ENABLED` / `HEADROOM_MODEL_ROUTES` config — no new routing surface.
- chat/completions carries the turn in `messages` (+ `tools`); Responses carries it in `input` (+ `tools`, with the top-level `instructions` folded into the size estimate). `has_tools` comes from `body.get("tools")` in both.
- Opt-in and disabled by default, fail-closed when no `model_router` is present (alternate mixin hosts / test doubles never set it), and skipped under `x-headroom-bypass` so a byte-faithful request is never rewritten.
- On the chat path, mutating `body["model"]` in place is enough — the outbound request is built from `body`. On the Responses path, the forwarder gates re-serialization on the body-mutation tracker, so a model rewrite marks it (`model_router`) or an otherwise-unmutated request would be forwarded byte-faithfully and drop the routed model.

Scope: this covers the OpenAI HTTP paths (chat + Responses). The Codex `/v1/responses` **WebSocket** path is left as a follow-up (it rewrites per-frame and would need the tracker equivalent per frame).

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

[...]

PR #2769, opened 2026-08-04
fix(proxy): read model-router env vars on the CLI entrypoints

### LINKED ISSUE #2764: [BUG] HEADROOM_MODEL_ROUTER_ENABLED / HEADROOM_MODEL_ROUTES are silently ignored by `headroom proxy` (single-worker CLI path)
## Environment

- `headroom-ai` installed from `main` at commit `0cb72f45` (self-reports 0.33.0),
  pipx, Python 3.14.6, macOS.
- Feature under test: the Unreleased "opt-in cost-aware model routing (#1706)".
- Proxy started exactly as a normal user would: `headroom proxy --host 127.0.0.1
  --port <p>`, no `--workers` flag (so the default single-worker path).

## Summary

Setting `HEADROOM_MODEL_ROUTER_ENABLED=1` and a valid `HEADROOM_MODEL_ROUTES`
has no effect when the proxy is started with `headroom proxy`. No routing
happens, and — the part that costs debugging time — **nothing is logged**: not
the `"model routing decision: ..."` line, not the "enabled but no valid routes"
warning, not a "router disabled" note. The operator's only signal is that spend
does not change.

The same env vars, same JSON, same request work correctly when the app is built
through `create_app_from_env()`.

## Repro

Rules used in both runs (identical):

```json
[{"name":"trivial-opus->haiku","from_models":["claude-opus-4-8"],
  "max_input_tokens":4096,"require_no_tools":true,"to_model":"claude-haiku-4-5"}]
```

Request used in both runs (identical):

```json
{"model":"claude-opus-4-8","max_tokens":12,
 "messages":[{"role":"user","content":"Reply with the single word OK."}]}
```

**A. CLI (broken)**

```bash
HEADROOM_MODEL_ROUTER_ENABLED=1 HEADROOM_MODEL_ROUTES="$ROUTES" \
```

→ response `model` is `claude-opus-4-8` (not routed).
→ `/stats` shows `"router": {"route_counts": {}}` and no `model_router` entry
  under `config`.
→ no routing line anywhere in the proxy log.

**B. env factory (works)**

```bash
HEADROOM_MODEL_ROUTER_ENABLED=1 HEADROOM_MODEL_ROUTES="$ROUTES" \
  python -m uvicorn headroom.proxy.server:create_app_from_env --factory \
  --host 127.0.0.1 --port 8788
```

→ response `model` is `claude-haiku-4-5`. Routed, as documented.

Gate semantics on path B behave as specified: a request carrying `tools` is not
routed, and a source model outside `from_models` (`claude-sonnet-5`) is not
routed.

## Cause

`ProxyConfig` has a `model_router` field, but the env vars are only read in
`_proxy_config_from_env()` / `create_app_from_env()`. `headroom/cli/proxy.py`
builds its own `ProxyConfig(...)` and never mentions `model_router` or
`HEADROOM_MODEL_ROUTES`:

```
$ python -c "import inspect; from headroom.cli import proxy as cp; \
    src=inspect.getsource(cp); print('model_router' in src, 'HEADROOM_MODEL_ROUTES' in src)"
False False
```

And `run_server()` only reaches the env factory on the multi-worker branch —
`workers > 1` sets `app_target = "headroom.proxy.server:create_app_from_env"`,
while the default single-worker path calls `create_app(config)` with the
CLI-built config. So the feature is reachable only under a deployment mode the
codebase itself warns against for cache-sensitive setups (per-process prefix
tracker, CCR retrieval busts, partial `/stats` per worker).

## Prior art

This is not the first env var to fall down this exact hole. **#948** ("`HEADROOM_EXCLUDE_TOOLS`
env var ignored by click CLI (proxy.py)", closed) reported the same shape: the value
was parsed in `server.py` but `cli/proxy.py` built its `ProxyConfig` without passing it,
so the setting was silently dropped. That report noted other env-only options had each
been wired into the click path individually — which is exactly why a new one can be
missed. (In `0cb72f45`, `cli/proxy.py` does now carry an `exclude_tools` parameter,
though the `HEADROOM_EXCLUDE_TOOLS` name itself still does not appear in that module;
`model_router` and `HEADROOM_MODEL_ROUTES` appear in neither form.)

Worth noting for two open PRs that extend the same surface: **#2395**
(`HEADROOM_MODEL_ROUTER_LOG_ONLY` shadow mode) and **#2354** (routing on the OpenAI
paths) both inherit this path. Shadow mode is hit hardest — its entire purpose is to
observe routing decisions, and on the CLI entrypoint no decision is ever made or logged.

**#279** already documents the `python -m uvicorn` ASGI launcher as the way to reach
env-driven configuration, so the CLI/factory divergence is known; the gap is that a
feature documented purely in terms of env vars is unreachable from the documented CLI.

## Suggested fix

Given #948, the general form looks more durable than another per-variable patch:
have `ProxyConfig` default `model_router` to `ModelRouterConfig.from_env(...)` so
every entrypoint inherits it, rather than adding two more names to the click CLI
that the next env-configured feature will also have to remember. Either way,
please also log the router's state once at startup — "model router: enabled,
N rules" / "disabled" — so a misconfiguration is visible without diffing spend. `from_env` already fails open to disabled with a warning on
malformed JSON; that warning simply never runs today on the CLI path.

Filing the missing `max_tokens` condition separately as a feature request.

ferrastas:
## Description

`HEADROOM_MODEL_ROUTER_ENABLED` and `HEADROOM_MODEL_ROUTES` are only consumed by
`_proxy_config_from_env()`, which `run_server()` reaches only on the multi-worker branch
(`workers > 1` swaps in the `create_app_from_env` factory). Both entrypoints that build
their own `ProxyConfig` — the click `headroom proxy` command and the argparse server main —
omitted `model_router` entirely, so cost-aware routing (#1706) is unreachable from the
documented way to run the proxy.

Nothing surfaces the misconfiguration: with the router absent no decision line is logged and
`/stats` carries no router config, so setting both variables looks identical to "no traffic
matched a rule". The only remaining path to the feature is `--workers > 1`, which the
codebase itself warns against for cache-sensitive deployments (per-process prefix tracker,
CCR retrieval busts, partial `/stats` per worker).

Closes #2764

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cli/proxy.py`: build `model_router` from `ModelRouterConfig.from_env(...)`, the
  same call the factory path uses. Matches how the surrounding env-only options are already
  read in that `ProxyConfig(...)` (`HEADROOM_COMPRESS_USER_MESSAGES`, `HEADROOM_MIN_TOKENS`, …).
- `headroom/proxy/server.py` (argparse main): same wiring — this entrypoint was missing it too.
- `HeadroomProxy.__init__`: log the resolved state once, `model router: enabled, N rule(s)`
  or `model router: disabled`, so a future entrypoint that forgets to wire it — or a rule set
  that failed to parse — is visible at startup rather than only in the bill.
- `tests/test_proxy/test_model_router_wiring.py`: cover the CLI entrypoint, not just
  `_proxy_config_from_env`.

`from_env` keeps failing open to disabled on malformed JSON; no behaviour changes when the
env vars are unset.

I wired the two call sites rather than defaulting the `ProxyConfig.model_router` field from
env, to keep the change in the shape those functions already use and avoid giving a dataclass
an implicit environment dependency. Happy to switch to the field-default form if you'd prefer
one place every future entrypoint inherits — say the word and I'll push that instead.

## Testing

- [ ] Unit tests pass (`pytest`) — see note below: the full suite has pre-existing failures on `main`

Four new tests, all of which fail on the unpatched tree:

- `test_cli_proxy_reads_router_from_env` — CLI config has the router enabled, rule parsed
- `test_cli_proxy_router_disabled_by_default` — no env, disabled
- `test_cli_proxy_router_disabled_on_malformed_routes` — fails open, does not crash
- `test_router_state_is_logged_at_startup` — both log states

The log test attaches a handler to the `headroom.proxy` logger instead of using `caplog`:
`create_app` reconfigures logging while it runs and drops pytest's root handler mid-call, so
a record emitted during app construction never reaches `caplog`.

### Test Output

```text
# targeted suite, with the fix
$ uv run pytest tests/test_proxy/test_model_router_wiring.py -q
17 passed, 1 warning in 2.31s

# same file, with the two source changes stashed (the four new tests are the regression)
FAILED tests/test_proxy/test_model_router_wiring.py::test_cli_proxy_reads_router_from_env
FAILED tests/test_proxy/test_model_router_wiring.py::test_cli_proxy_router_disabled_by_default
FAILED tests/test_proxy/test_model_router_wiring.py::test_cli_proxy_router_disabled_on_malformed_routes
FAILED tests/test_proxy/test_model_router_wiring.py::test_router_state_is_logged_at_startup
4 failed, 13 passed, 1 warning in 2.12s

$ uv run --with ruff ruff check .
All checks passed!

$ uv run mypy headroom/cli/proxy.py headroom/proxy/server.py

# full suite (tests/test_memory_eval.py excluded: needs litellm, not in dev+proxy extras)
$ uv run pytest tests/ -q --ignore=tests/test_memory_eval.py
38 failed, 9794 passed, 538 skipped, 5913 warnings, 2 errors in 601.28s
```

On the full-suite failures: re-running just that failed set with my two source changes
stashed still fails 35 of the 38 on an otherwise clean `origin/main`, and the handful that
flip between runs are the live/network ones (`*_live*`, the langchain live errors) — the same
set fails with and without this patch, so they are pre-existing and flaky rather than
introduced here. Nothing in the failing set touches the model router or CLI config.

## Real Behavior Proof

- Environment: macOS 15 (Darwin 25.6.0), Python 3.14.6, `headroom-ai` installed with pipx
  from `main` at `0cb72f45`, `--mode cache` (default), upstream an Anthropic-compatible
  endpoint on loopback.

- Exact command / steps: same rules and same request on every run.

  ```json
  [{"name":"trivial-opus->haiku","from_models":["claude-opus-4-8"],
    "max_input_tokens":4096,"require_no_tools":true,"to_model":"claude-haiku-4-5"}]
  ```

  ```json
  {"model":"claude-opus-4-8","max_tokens":12,
   "messages":[{"role":"user","content":"Reply with the single word OK."}]}
  ```

  1. Before the patch, the documented entrypoint:
     `HEADROOM_MODEL_ROUTER_ENABLED=1 HEADROOM_MODEL_ROUTES="$ROUTES" headroom proxy --host 127.0.0.1 --port 8787`
  2. Before the patch, the factory (control, to prove the router itself works):
     `HEADROOM_MODEL_ROUTER_ENABLED=1 HEADROOM_MODEL_ROUTES="$ROUTES" python -m uvicorn headroom.proxy.server:create_app_from_env --factory --host 127.0.0.1 --port 8788`
  3. After the patch, step 1 again.

- Observed result: before the patch the documented entrypoint did not route at all (response
  `model` came back `claude-opus-4-8`, nothing logged), while the factory path routed the
  identical request to `claude-haiku-4-5`; after the patch the CLI entrypoint routes to
  `claude-haiku-4-5` too and logs `model router: enabled, 1 rule(s)` at startup. Per step:
  1. Response `model` was `claude-opus-4-8` — not routed. `/stats` showed
     `"router": {"route_counts": {}}` and no `model_router` under `config`. No routing line
     anywhere in the proxy log.
  2. Response `model` was `claude-haiku-4-5` — routed. In the same run a request carrying
     `tools` stayed on `claude-opus-4-8`, and a `claude-sonnet-5` source model was left
     alone, so `require_no_tools` and `from_models` behave as documented.
  3. Response `model` was `claude-haiku-4-5`, matching the factory path, and startup logged
     `model router: enabled, 1 rule(s)`. With the env vars unset it logs
     `model router: disabled` and the request is untouched.

- Not tested: the OpenAI/Gemini/Bedrock paths (routing is Anthropic-only today, and
  #2354 is extending that separately); Windows and Linux supervised runners; multi-worker
  startup, which already worked and is unchanged.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A, the documented env vars
      were already correct; this makes them work on the documented entrypoint
- [x] I did **not** edit `CHANGELOG.md` — untouched, as is `uv.lock`

## Additional Notes

- No dependency changes. `uv.lock` deliberately left unmodified.
- Prior art for the same class of bug: #948, where `HEADROOM_EXCLUDE_TOOLS` was parsed in
  `server.py` but never passed by `cli/proxy.py`'s `ProxyConfig(...)`. That is the argument
  for the field-default variant mentioned above, if you want to stop this recurring.
- The missing `max_tokens` condition on `ModelRoute` that I hit while testing this is filed
  separately as #2765 and is not part of this PR.
- Two open PRs extend the same env surface and inherit this path: #2395 (shadow mode) and
  #2354 (OpenAI paths). Shadow mode is affected most directly, since its whole purpose is
  observing decisions that on the CLI entrypoint were never made.

### DIFF
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -14,6 +14,7 @@ from headroom.providers.registry import (
     resolve_api_targets,
     resolve_extra_headers,
 )
+from headroom.proxy.model_router import ModelRouterConfig
 from headroom.proxy.modes import PROXY_MODE_CACHE, normalize_proxy_mode
 
 from .main import main
@@ -1194,6 +1195,13 @@ def proxy(
         else frozenset(),
         tool_profiles=_parse_tool_profiles([]) or None,
         smart_crusher_with_compaction=_get_env_bool_optional("HEADROOM_SMART_CRUSHER_COMPACTION"),
+        # Cost-aware model routing (#1706) is configured purely by env, and this
+        # is the entrypoint `headroom proxy` uses, so it has to read it here —
+        # `_proxy_config_from_env()` only runs on the multi-worker/factory path.
+        model_router=ModelRouterConfig.from_env(
+            os.environ.get("HEADROOM_MODEL_ROUTER_ENABLED"),
+            os.environ.get("HEADROOM_MODEL_ROUTES"),
+        ),
         savings_profile=os.environ.get("HEADROOM_SAVINGS_PROFILE") or "coding",
         target_ratio=target_ratio,
         compress_system_messages=_get_env_bool_optional("HEADROOM_COMPRESS_SYSTEM_MESSAGES"),
--- a/headroom/proxy/server.py
+++ b/headroom/proxy/server.py
@@ -806,8 +806,18 @@ class HeadroomProxy(
         self.metrics = PrometheusMetrics(cost_tracker=self.cost_tracker, stateless=config.stateless)
 
         # Cost-aware model routing (issue #1706). Disabled unless configured, so
-        # the default request path is unchanged.
+        # the default request path is unchanged. Logged once at startup: routing
+        # is invisible when off (no per-request decision line is emitted), so an
+        # entrypoint that never wired the config, or a rule set that failed to
+        # parse, would otherwise show up only as an unexplained bill.
         self.model_router = ModelRouter(config.model_router)
+        if self.model_router.enabled:
+            logger.info(
+                "model router: enabled, %d rule(s)",
+                len(config.model_router.routes) if config.model_router else 0,
+            )
+        else:
+            logger.info("model router: disabled")
 
         # Initialize transforms based on routing mode.
         #
@@ -5621,6 +5631,12 @@ if __name__ == "__main__":
             _get_env_bool("HEADROOM_SMART_CRUSHER_COMPACTION", False)
             if "HEADROOM_SMART_CRUSHER_COMPACTION" in os.environ
             else None
+        ),
+        # Cost-aware model routing (#1706): env-only config, so every entrypoint
+        # that builds its own ProxyConfig has to read it (see also cli/proxy.py).
+        model_router=ModelRouterConfig.from_env(
+            os.environ.get("HEADROOM_MODEL_ROUTER_ENABLED"),
+            os.environ.get("HEADROOM_MODEL_ROUTES"),
         ),
         cache_enabled=cache_enabled,
         cache_ttl_seconds=_get_env_int("HEADROOM_CACHE_TTL", args.cache_ttl),
--- a/tests/test_proxy/test_model_router_wiring.py
+++ b/tests/test_proxy/test_model_router_wiring.py
@@ -2,6 +2,11 @@
 
 Covers env -> ProxyConfig, ProxyConfig -> live proxy, and the presence of the
 routing block in the Anthropic request handler.
+
+The env -> ProxyConfig half is checked once per entrypoint that builds its own
+ProxyConfig, not just via ``_proxy_config_from_env``: that helper only runs on
+the factory/multi-worker path, so testing it alone let #2764 through — the
+router was unreachable from ``headroom proxy``, the documented way to run it.
 """
 
 from __future__ import annotations
@@ -9,15 +14,24 @@ from __future__ import annotations
 import inspect
 import json
 import logging
-from unittest.mock import AsyncMock, MagicMock
+from collections.abc import Iterator
+from contextlib import contextmanager
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import httpx
 import pytest
+from click.testing import CliRunner
 from fastapi.testclient import TestClient
 
+from headroom.cli.main import main
 from headroom.proxy.handlers.anthropic import AnthropicHandlerMixin
 from headroom.proxy.model_router import ModelRoute, ModelRouter, ModelRouterConfig
 from headroom.proxy.server import ProxyConfig, _proxy_config_from_env, create_app
+
+ROUTES_JSON = (
+    '[{"name":"small","max_input_tokens":4000,"require_no_tools":true,'
+    '"to_model":"claude-haiku-4-5"}]'
+)
 
 MESSAGES = "/v1/messages"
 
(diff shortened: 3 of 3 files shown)


ferrastas:
One more data point for the field-default question in the PR description, found while
debugging an unrelated cache problem on the same tree.

`prefix_freeze_session_ttl` is in the same position `model_router` was: a `ProxyConfig`
field (`proxy/models.py:316`, default `600`) that is consumed —
`proxy/server.py:1017` passes it into `PrefixFreezeConfig` — but that no entrypoint ever
sets. There is no env var and no CLI flag; the only writer in the tree is
`testing/harness.py:848`. Same for `prefix_freeze.enabled` and `min_cached_tokens`.

The effect is quieter than the router one, since the default is functional rather than
absent: the freeze tracker forgets a session after 10 minutes of inactivity, while the
provider cache it exists to protect can live for an hour. An operator who wants those to
agree has no way to say so.

I mention it only because it is the third instance of this shape in one afternoon
(`HEADROOM_EXCLUDE_TOOLS` in #948, `model_router` here, this), and because two of the
three would have been covered for free by defaulting the fields on `ProxyConfig` rather
than wiring each entrypoint. Still happy to push either variant of this PR — just noting
that the per-knob approach has a visible tail.

JerrettDavis (approved):
Reviewed the current head (72d517dc) across both direct ProxyConfig construction paths, the existing factory path, malformed/disabled env behavior, startup observability, and targeted regression tests. The wiring matches the established from_env semantics, defaults remain disabled, and no routing behavior changes when the variables are unset. The description is complete and candid about suite noise and non-goals. No blocking findings.

PR #2771, opened 2026-08-04
fix(proxy): keep the longest cache_control TTL when deferring tools

### LINKED ISSUE #2767: inject_tool_search_deferral downgrades a 1h cache_control marker to 5m, causing 400 invalid_request_error
### Summary

`inject_tool_search_deferral` keeps only the **last** `cache_control` marker it strips from deferred tools. When a bare (5m) marker is stripped after a `ttl: "1h"` one, the 1h marker is lost and the tools prefix goes upstream at 5m, while the client's message breakpoints are still 1h. The API rejects the request:

```
400 invalid_request_error
messages.4.content.2.cache_control.ttl: a ttl='1h' cache_control block must not come
after a ttl='5m' cache_control block. Note that blocks are processed in the following
order: `tools`, `system`, `messages`.
```

### Environment

- headroom-ai 0.33.0, `--mode token --backend anthropic`
- Claude Code 2.1.221 (sends `anthropic-beta: ...prompt-caching-scope-2026-01-05...`, puts cache breakpoints on **tools**)
- Deferral very active: `tool_search_deferral:13tools:18589tok client=claude-code`

### Root cause

`proxy/helpers.py`, in `inject_tool_search_deferral`:

```python
_dropped = new_tool.pop("cache_control", None)
if _dropped is not None:
    dropped_cache_control = True
    # Keep the marker itself, not just the fact of it: re-placing a bare
    # ephemeral would downgrade a 1h breakpoint to the 5m default.
    if isinstance(_dropped, dict):
        dropped_marker = _dropped
```

The comment states the intent correctly, but last-write-wins defeats it: any 5m marker dropped after the 1h one silently replaces it, producing exactly the downgrade the comment set out to prevent.

Related: the docstring nearby says the transform is safe because cache breakpoints "sit on messages (never tools), so defer_loading cannot collide with cache_control". That holds for the client it was validated against, but Claude Code does put a breakpoint on tools, so the collision is reachable in practice.

### Reproduction

13 tools, 3 core (resident) and 10 deferrable, where a non-core tool carries `{"type": "ephemeral", "ttl": "1h"}` and a **later** non-core tool carries bare `{"type": "ephemeral"}`.

```
before fix: markers kept = [{'type': 'ephemeral'}]                  # 1h lost
after fix:  markers kept = [{'type': 'ephemeral', 'ttl': '1h'}]     # ordering stays legal
```

### Suggested fix

Keep the longest TTL among dropped markers instead of the last one:

```python
if dropped_marker is None or (
    _dropped.get("ttl") == "1h" and dropped_marker.get("ttl") != "1h"
):
    dropped_marker = _dropped
```

There is a second, narrower hole on the same path: when `resident_has_cache_control` is already true at 5m and a 1h marker gets dropped, the 1h is discarded outright. I did not patch that one since I could not trigger it, flagging it in case it matters.

Happy to open a PR.

crb8:
## Description

`inject_tool_search_deferral` strips `cache_control` off every deferred tool and re-places a single marker on the last resident tool. It kept the last marker it stripped, so a bare (5m) marker on a later deferred tool overwrote a `ttl='1h'` one, which is exactly the downgrade the existing comment set out to prevent.

The tools prefix then goes upstream at 5m while the client's message breakpoints are still 1h, and the API rejects the whole request.

Closes #2767

## Type of Change

## Changes Made

- `inject_tool_search_deferral` now keeps the longest TTL among the stripped `cache_control` markers instead of the last one.
- New test `tests/test_tool_search_deferral_ttl.py` pinning the invariant.

## Testing

### Test Output

```text
# new test, before the fix
E   AssertionError: the 1h breakpoint was downgraded to 5m by a later bare marker,
    which makes the client's 1h message breakpoints illegal upstream.
    markers=[{'type': 'ephemeral'}]
1 failed

# after the fix
tests/test_tool_search_deferral_ttl.py .                                 [100%]
1 passed

# related suites
tests/test_tool_search_deferral_ttl.py tests/test_cache_control_move_bust.py
tests/test_proxy/ tests/test_ccr_tool_always_on.py
tests/test_issue_728_empty_tools_injection.py
265 passed, 2 failed

# both remaining failures are pre-existing on main, unrelated to this change:
#   test_openai_backend_path.py::test_litellm_vertex_backend_path_... (fails on clean main)
#   test_cc_switch_reconciler.py::test_switching_provider_recaptures (flaky: passed
#   1/3 runs on clean main with this change stashed; does not touch this code path)

ruff check  -> All checks passed!
ruff format -> 2 files already formatted
```

## Real Behavior Proof

- Environment: headroom-ai 0.33.0, `--mode token --backend anthropic`, persistent-task deployment on Windows 11 / Python 3.14, Claude Code 2.1.221 (which places cache breakpoints on `tools`). Deferral active at `tool_search_deferral:13tools:18589tok client=claude-code`.
- Exact command / steps: observed in production logs first, then reduced to the unit test above (13 tools: 3 resident core, a `ttl='1h'` marker on a non-core tool, a bare `{"type":"ephemeral"}` marker on a later non-core tool).
- Observed result: upstream returned `400 invalid_request_error` with `messages.4.content.2.cache_control.ttl: a ttl='1h' cache_control block must not come after a ttl='5m' cache_control block`, three times in a row (client retries) before failing the turn. After patching the same change into the deployed 0.33.0 build, that error has not recurred.
- Not tested: the narrower sibling hole on the same path, where `resident_has_cache_control` is already true at 5m and a 1h marker is dropped outright. I could not construct a client request that reaches it, so I left it alone and flagged it in #2767 rather than guess.

## Review Readiness

## Checklist


JerrettDavis (approved):
Reviewed the current head (c7cb9e49) against the cache-control ordering invariant and the surrounding deferral/move logic. Preferring an explicit 1h marker over later bare/5m markers is narrowly scoped, preserves marker shape, and directly covers the reported regression. The documented resident-marker sibling case is appropriately left out rather than broadened without a valid request shape. No blocking findings.

JerrettDavis:
Superseded by merged PR #2941, which fixes the same #2767 failure at the final outbound seam and includes an end-to-end regression for the deferral shape. Closing this narrower implementation to avoid overlapping fixes.

PR #2777, opened 2026-08-04
fix(providers): stop pricing gpt-4.1 as legacy gpt-4

chopratejas:
## Description

`_get_pricing` matches by prefix in **plain dict order**, so the first *inserted* prefix won rather than the most specific one. `gpt-4.1` fell into the `gpt-4` entry and was priced at the legacy $30/$60.

Audited every entry against litellm's `model_cost`:

| model | ours (before) | actual | error |
|---|---|---|---|
| `gpt-4.1` | $30.00 / $60.00 | $2.00 / $8.00 | **15×** |
| `gpt-4.1-mini` | $30.00 / $60.00 | $0.40 / $1.60 | **75×** |
| `gpt-4.1-nano` | $30.00 / $60.00 | $0.10 / $0.40 | **300×** |
| `gpt-5` | $2.50 / $10.00 | $1.25 / $10.00 | 2× |
| `gpt-5-mini` | $2.50 / $10.00 | $0.25 / $2.00 | 10× |
| `gpt-5-nano` | $2.50 / $10.00 | $0.05 / $0.40 | **50×** |
| `o4-mini` | $2.50 / $10.00 | $1.10 / $4.40 | 2× |
| `o3` | $10.00 / $40.00 | $2.00 / $8.00 | 5× (stale) |

Two distinct causes: `gpt-4.1*` was *shadowed* by `gpt-4`; `gpt-5*` and `o4-mini` had no entry and fell to the unknown-model default (GPT-4o tier). `o3` is a third thing — a pre-existing entry still carrying its pre-June-2025 rate.

`gpt-4o`, `gpt-4`, `gpt-4-turbo`, `gpt-3.5-turbo`, `o1` and `o3-mini` already matched litellm exactly, which is what makes litellm a trustworthy oracle for the rest.

**Scope — this is narrower than it looks.** Unlike `get_context_limit`, `_get_pricing` has **no litellm lookup in front of it**, so this table is the only source for:

- `client.py:793-794` — `cost_before` / `cost_after` for SDK users
- `reporting/generator.py`
- `evals/cost_tracker.py`

The **proxy cost path is unaffected** — `proxy/cost.py` calls `litellm.cost_per_token` directly, and `savings_ledger.estimate_cost_usd` also goes through litellm. So the dashboard's cost-avoided figures were already correct; this fixes the SDK and report numbers.

## Type of Change

## Changes Made

- `_get_pricing` iterates `sorted(..., key=len, reverse=True)` — longest prefix wins. Same root-cause fix already applied to the context-limit and encoding lookups in #2762, so all three resolvers now agree on precedence.
- Added `gpt-4.1` (+`-mini`/`-nano`), `gpt-5` (+`-mini`/`-nano`), `o4-mini`.
- Corrected `o3` to $2/$8 in both `_PRICING` and `_PATTERN_DEFAULTS`.
- Rewrote the comment above `_PRICING`, which claimed "LiteLLM is preferred source" — there is no litellm lookup in this path, and believing otherwise is precisely why a stale table went unnoticed.
- Stamped `_PRICING_LAST_UPDATED` to today. It read `2025-01-14` against a 60-day staleness window, so the "pricing data is N days old" warning had been firing for ~18 months, which trains people to ignore it.

## Testing

### Test Output

```text
$ pytest tests/test_openai_pricing_resolution.py -q
18 passed

# on main, 16 of those 18 fail
```

Targeted run across everything that reads pricing or provider metadata:

```text
$ pytest tests/test_utils.py tests/test_reporting.py tests/test_cost_pricing_warning_dedup.py \
         tests/test_pricing.py tests/test_pricing_litellm.py tests/test_provider_model_fallback.py \
         tests/test_models.py tests/test_savings_ledger.py \
         tests/test_openai_model_table_resolution.py tests/test_provider_tokenizer_one_ruler.py -q
153 passed, 2 skipped
```

```text
$ ruff check headroom/providers/openai.py tests/test_openai_pricing_resolution.py
All checks passed!
```

Deferring the full suite to CI — no maturin/Rust core in this environment, so the native shards can't run locally.

## Real Behavior Proof

- **Environment:** macOS, Python 3.13.7, isolated worktree at `upstream/main` (`0cb72f45`), litellm installed as the oracle.
- **Exact command / steps:** for every key in `_PRICING` plus versioned variants, compare `provider._get_pricing(m)` against `litellm.get_model_info(m)` input/output cost per token × 1e6.
- **Observed result — after:**

```text
model                       in     out
gpt-3.5-turbo              0.5     1.5   ok
gpt-4                     30.0    60.0   ok
gpt-4-turbo               10.0    30.0   ok
gpt-4.1                    2.0     8.0   ok
gpt-4.1-2025-04-14         2.0     8.0   ok
gpt-4.1-mini               0.4     1.6   ok
gpt-4.1-nano               0.1     0.4   ok
gpt-4o                     2.5    10.0   ok
gpt-4o-2024-08-06          2.5    10.0   ok
gpt-4o-mini               0.15     0.6   ok
gpt-5                     1.25    10.0   ok
gpt-5-mini                0.25     2.0   ok
gpt-5-nano                0.05     0.4   ok
o1                        15.0    60.0   ok
o3                         2.0     8.0   ok
o3-mini                    1.1     4.4   ok
o4-mini                    1.1     4.4   ok

mismatches: 0
```

Entries litellm cannot price (`gpt-4-32k`, `o1-mini`, `o1-preview`, `gpt-5-mini-2025`) resolve via prefix inheritance and are unchanged by this PR.

## Deliberately not included

`gpt-4-32k` inherits `gpt-4`'s $30/$60 by prefix; the real historical rate was $60/$120. litellm has no entry for it, so I would be guessing at a retired model's price — left alone rather than assert a number I can't verify.

### DIFF
--- a/headroom/providers/openai.py
+++ b/headroom/providers/openai.py
@@ -23,7 +23,7 @@ from .base import Provider, TokenCounter
 logger = logging.getLogger(__name__)
 
 # Pricing metadata for transparency
-_PRICING_LAST_UPDATED = date(2025, 1, 14)
+_PRICING_LAST_UPDATED = date(2026, 8, 4)  # every _PRICING entry verified vs litellm
 _PRICING_STALE_DAYS = 60  # Warn if pricing data is older than this
 
 # Warning tracking
@@ -129,21 +129,35 @@ _CONTEXT_LIMITS: dict[str, int] = {
     "deepseek-coder-v2": 128_000,
 }
 
-# Fallback pricing - LiteLLM is preferred source
-# OpenAI pricing per 1M tokens (input, output)
-# NOTE: These are ESTIMATES. Always verify against actual OpenAI billing.
-# Last updated: 2025-01-14
+# USD per 1M tokens, (input, output). NOTE: these are ESTIMATES -- always verify
+# against actual OpenAI billing.
+#
+# Unlike get_context_limit, _get_pricing has NO litellm lookup in front of it, so
+# this table is authoritative for every consumer (client.py cost_before /
+# cost_after, reporting, evals). Every entry below was checked against litellm's
+# model_cost; keep the newer families explicit, because these are matched by
+# prefix and "gpt-4.1" would otherwise fall into "gpt-4" and be priced at the
+# legacy $30/$60 -- 15x its real rate, and 300x for gpt-4.1-nano.
 _PRICING: dict[str, tuple[float, float]] = {
     "gpt-4o": (2.50, 10.00),
     "gpt-4o-mini": (0.15, 0.60),
+    "gpt-4.1": (2.00, 8.00),
+    "gpt-4.1-mini": (0.40, 1.60),
+    "gpt-4.1-nano": (0.10, 0.40),
+    "gpt-5": (1.25, 10.00),
+    "gpt-5-mini": (0.25, 2.00),
+    "gpt-5-nano": (0.05, 0.40),
     "gpt-4-turbo": (10.00, 30.00),
     "gpt-4": (30.00, 60.00),
     "gpt-3.5-turbo": (0.50, 1.50),
     "o1": (15.00, 60.00),
     "o1-preview": (15.00, 60.00),
     "o1-mini": (3.00, 12.00),
-    "o3": (10.00, 40.00),
+    # o3 was cut from $10/$40 to $2/$8 in June 2025; the old rate
+    # overstated every o3 cost estimate by 5x.
+    "o3": (2.00, 8.00),
     "o3-mini": (1.10, 4.40),
+    "o4-mini": (1.10, 4.40),
 }
 
 # Pattern-based defaults for unknown models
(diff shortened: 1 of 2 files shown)


chopratejas:
Superseded — the content of this PR landed on `main` via #2779, which was rebased onto this branch and so carried commit `29bfd69c` in its diff. GitHub did not auto-close this because the squash merge produced a new SHA, so the commit is not literally on `main` even though every change is.

Verified on `main` (`0e1d6bfa`):

```text
$ git show upstream/main:headroom/providers/openai.py | grep -E "gpt-4.1|o3.*\(|sorted\(self._pricing"
146:    "gpt-4.1": (2.00, 8.00),
148:    "gpt-4.1-nano": (0.10, 0.40),
160:    "o3": (2.00, 8.00),
707:        for model_prefix in sorted(self._pricing, key=len, reverse=True):
```

So the corrected table values, the longest-prefix-first matching, and the `o3` price fix are all in. Closing as merged-via-#2779, not abandoned.

PR #2780, opened 2026-08-04
fix(policy): price net-cost mutations with the 1h cache-write tier

### LINKED ISSUE #2773: [BUG] net_mutation_gain hardcodes the 5m cache-write multiplier — 1h-TTL clients are gated ~40% pro-mutation
## Description

`net_mutation_gain` prices a prompt-cache write at a hardcoded `1.25x` — Anthropic's
**5-minute** TTL tier. A client on the **1-hour** tier writes at `2.0x`. The constant is
the write penalty `w` in the #856 formula, so for 1h-TTL traffic the penalty is
under-stated by ~40% and the gate leans pro-mutation, in the same direction as the bug
just fixed in #906.

`headroom/transforms/compression_policy.py:68` and the Rust mirror
`crates/headroom-core/src/compression_policy.rs:136`:

```python
#: Anthropic prompt-cache write multiplier: a ``cache_creation`` token
#: costs 1.25x a plain input token (5-minute TTL tier).
CACHE_WRITE_MULTIPLIER: float = 1.25
```

The source comment already identifies this precisely, and defers it:

> ponytail: hardcoded to the 5m tier. A client on Anthropic's 1h cache
> (ENABLE_PROMPT_CACHING_1H / cache_control.ttl="1h", which Headroom preserves) writes at
> 2.0x, so its mutations are gated with a ~40% under-stated write penalty. Harmless while
> the net-cost gate stays default-off (HEADROOM_NET_COST_POLICY); thread the TTL from
> cold_prefix.anthropic_cache_ttl_seconds through ContentRouter -> net_mutation_gain if
> that gate is ever turned on.

Filing because "harmless while default-off" expires when #907 / #905 land. Claude Code on
the 1h tier is not an exotic configuration, and Headroom preserves the header that selects
it.

## To Reproduce

No repro needed — it is a constant. But the codebase's own golden scenario inverts.

`crates/headroom-core/src/compression_policy.rs:439-448`, mirrored in
`tests/test_compression_policy.py`, shaves 50K under a 10K warm suffix at R=3:

```
w=1.25 (5m):  50000·(1.25 + 0.1·2) − 1.0·1.15·60000 = 72500 − 69000 = +3500  → mutate
```

The comment calls it *"tight but positive — consistent with the 2.3-read break-even."*

Substituting the 1h write price, same scenario, nothing else changed:

```
w=2.00 (1h):  50000·(2.00 + 0.1·2) − 1.0·1.90·60000 = 110000 − 114000 = −4000  → do not mutate
```

The decision flips. Break-even moves from **R = 2.3 to R = 3.8 remaining reads** — the
#856 anchor is 65% off for these clients.

## Expected Behavior

`w` reflects the TTL the request actually selected: `1.25` for 5m, `2.0` for 1h.

## Actual Behavior

`w` is always `1.25`, so 1h-TTL mutations clear the gate at a ~40% discount on their true
cost.

## Code Sample

```python
from headroom.transforms.compression_policy import policy_for_mode
from headroom.proxy.auth_mode import AuthMode

p = policy_for_mode(AuthMode.PAYG)
print(p.net_mutation_gain(50_000, 10_000, 3.0, 1.0))   # 3500.0 -> mutate
print(p.should_mutate_deep(50_000, 10_000, 3.0, 1.0))  # True

# The same turn on the 1h tier is worth -4000: w=2.0 gives
#   50000*(2.0 + 0.1*2) - 1.0*(2.0-0.1)*60000 = -4000
# There is no parameter to express that.
```

## Error Output

```
n/a — silent mispricing, no error raised
```

## Environment

- **Headroom version**: 0.33.0 (installed from `main` at `0cb72f45`)
- **Python version**: 3.14.6
- **OS**: macOS 15 (Darwin 25.6.0)
- **LLM Provider**: Anthropic, via `headroom proxy --mode cache`

## Additional Context

The TTL is already resolved on this path — `cold_prefix.anthropic_cache_ttl_seconds(model,
messages)` returns `3600` / `300` from the request's own `cache_control`, and
`handlers/anthropic.py:1141` already calls it. So the fix the comment describes is
threading a value that exists a few frames away, not deriving a new one. A `w` derived
from that (`2.0` when `ttl == 3600`, else `1.25`) keeps the parity goldens intact for the
5m cases and needs one new golden per tier.

Worth noting that Headroom does record the split, which makes the affected population
measurable rather than hypothetical. From `/stats` on one working session of Claude Code
traffic through the proxy:

```json
"cache_write_5m_tokens": 0,
"cache_write_1h_tokens": 12873033,
"cache_write_1h_requests": 548
```

Every cache write in that window was on the tier the constant does not describe.

Related: #904 (tracking), #906 (the warm-cache penalty correction — same direction of
error, different term of the formula), #907 / #905 (P2 consumption, which is what makes
this reachable).

axelray-dev:
## Description

This fixes the net-cost mutation gate for requests using Anthropic's 1-hour prompt-cache TTL.

The gate previously hardcoded the 5-minute cache-write multiplier of 1.25x. A 1-hour cache write costs 2.0x, so the old calculation understated the true write penalty and could incorrectly recommend mutation for 1-hour clients.

Closes #2773

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added TTL-aware cache-write multiplier selection for 5-minute and 1-hour tiers.
- Threaded the resolved TTL through the content router and compression policy helpers.
- Preserved the existing 5-minute behavior as the default.
- Added Python and Rust regression coverage for the 1-hour tier.
- Retuned the netcost gate fixtures so the 1-hour write tier flips the decision in the full ContentRouter path.
- Did not edit CHANGELOG.md.

## Testing

- [x] Unit tests pass (pytest)
- [x] Linting passes (ruff check .)
- [ ] Type checking passes (mypy headroom)

### Test Output

```text
pytest tests/test_compression_policy.py -q
20 passed

cargo test -p headroom-core --lib compression_policy -- --nocapture
14 passed

pytest tests/test_netcost_gate.py -q
27 passed

Ruff checks and formatting passed.
git diff --check passed.
```

## Real Behavior Proof

- Environment: Linux x86_64 contributor checkout with Python and Rust test environments.
- Exact command / steps:
  - Ran the Python compression policy test suite.
  - Ran the Rust compression policy unit tests.
  - Ran the netcost gate suite, including the 1-hour env and request-marker cases.
  - Exercised the new 1-hour TTL golden case alongside the existing 5-minute cases.
- Observed result: The 1-hour case uses the 2.0x write multiplier and skips the same candidate that still mutates under 5-minute pricing. Existing 5-minute behavior remains covered and passing.
- Not tested: A live Anthropic request through the proxy and production traffic.

## Review Readiness

## Checklist

- [x] I did **not** edit CHANGELOG.md - it is generated by release-please from my Conventional Commit PR title (a CI guard enforces this)

Not applicable for this backend policy fix.

## Additional Notes

Ready for review. CI is green on the current tip.

### DIFF
--- a/crates/headroom-core/src/compression_policy.rs
+++ b/crates/headroom-core/src/compression_policy.rs
@@ -135,10 +135,23 @@ pub(crate) const MAX_LOSSY_RATIO_SUBSCRIPTION: f32 = 0.25;
 /// net-cost mutation formula (#856).
 pub const CACHE_WRITE_MULTIPLIER: f32 = 1.25;
 
+/// Anthropic prompt-cache write multiplier for the 1-hour TTL tier.
+pub const CACHE_WRITE_MULTIPLIER_1H: f32 = 2.0;
+
 /// Anthropic prompt-cache read multiplier: a `cache_read` token costs
 /// 0.1× a plain input token. Input to the net-cost mutation formula
 /// (#856).
 pub const CACHE_READ_MULTIPLIER: f32 = 0.1;
+
+/// Return the cache-write multiplier for a prompt-cache TTL tier.
+///
+/// Invalid, missing, and non-positive values retain the 5-minute default.
+pub fn cache_write_multiplier_for_ttl(ttl_seconds: Option<f32>) -> f32 {
+    match ttl_seconds {
+        Some(ttl) if ttl.is_finite() && ttl >= 3_600.0 => CACHE_WRITE_MULTIPLIER_1H,
+        _ => CACHE_WRITE_MULTIPLIER,
+    }
+}
 
 /// Per-auth-mode policy that downstream compression stages consult.
 ///
@@ -269,7 +282,26 @@ impl CompressionPolicy {
         expected_reads: f32,
         p_alive: f32,
     ) -> f32 {
-        let w = CACHE_WRITE_MULTIPLIER;
+        self.net_mutation_gain_with_write_multiplier(
+            delta_t,
+            suffix_tokens,
+            expected_reads,
+            p_alive,
+            None,
+        )
+    }
+
+    /// Variant of [`Self::net_mutation_gain`] with an explicit cache-write
+    /// multiplier. `None` uses the 5-minute default.
+    pub fn net_mutation_gain_with_write_multiplier(
+        &self,
+        delta_t: u32,
+        suffix_tokens: u32,
+        expected_reads: f32,
+        p_alive: f32,
+        write_multiplier: Option<f32>,
+    ) -> f32 {
+        let w = write_multiplier.unwrap_or(CACHE_WRITE_MULTIPLIER);
         let r = CACHE_READ_MULTIPLIER;
         // f32::max ignores NaN (returns the other operand), so NaN reads
         // land on 0.0; clamp would propagate NaN, so guard alive explicitly.
@@ -299,7 +331,32 @@ impl CompressionPolicy {
         expected_reads: f32,
         p_alive: f32,
     ) -> bool {
-        self.net_mutation_gain(delta_t, suffix_tokens, expected_reads, p_alive) > 0.0
+        self.should_mutate_deep_with_write_multiplier(
+            delta_t,
+            suffix_tokens,
+            expected_reads,
+            p_alive,
+            None,
+        )
+    }
+
+    /// Variant of [`Self::should_mutate_deep`] with an explicit cache-write
+    /// multiplier. `None` uses the 5-minute default.
+    pub fn should_mutate_deep_with_write_multiplier(
+        &self,
+        delta_t: u32,
+        suffix_tokens: u32,
+        expected_reads: f32,
+        p_alive: f32,
+        write_multiplier: Option<f32>,
+    ) -> bool {
+        self.net_mutation_gain_with_write_multiplier(
+            delta_t,
+            suffix_tokens,
+            expected_reads,
+            p_alive,
+            write_multiplier,
+        ) > 0.0
     }
 
     /// Remaining-read count at which a warm-cache (P_alive = 1)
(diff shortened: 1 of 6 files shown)


JerrettDavis (changes requested):
The current draft does not yet fix ordinary 1-hour requests as the title/description claim. `ContentRouter` derives the multiplier only from `_net_cost_cache_ttl_seconds()`, whose default is 300 and whose only override is the separate `HEADROOM_NET_COST_CACHE_TTL_SECONDS` variable. The new integration test likewise has to set that variable to 3600. Meanwhile the Anthropic handler already resolves the actual model/request tier through `cold_prefix.anthropic_cache_ttl_seconds(...)` (including request `cache_control.ttl="1h"` and Claude Code env settings), but that value is not threaded into this gate. Consequently a real 1-hour request without the net-cost-specific override still uses 1.25x. Please pass the already-resolved request TTL (or an equivalent authoritative per-request value) through to `ContentRouter`, retain the env value only as an explicit fallback if desired, and add an integration test where a request-level 1h marker changes the decision without setting `HEADROOM_NET_COST_CACHE_TTL_SECONDS`.

axelray-dev:
Implemented the requested per-request TTL plumbing. The Anthropic handler now resolves the authoritative prompt-cache TTL with `anthropic_cache_ttl_seconds(...)` and passes it through every compression pipeline path, including batch requests. `ContentRouter` uses that TTL when present and retains `HEADROOM_NET_COST_CACHE_TTL_SECONDS` as the fallback for other callers. Added a regression test where a request-level `cache_control.ttl: "1h"` marker selects the 1-hour pricing decision without setting the net-cost TTL environment variable. The PR remains a draft while the updated checks run.

JerrettDavis (approved):
Re-reviewed latest 58c389c6 after the requested fix. The handler now resolves the authoritative request/model cache TTL once and passes it through every Anthropic compression path (including batch), while ContentRouter uses the env only as a fallback. The request-level 1h regression proves the decision changes without HEADROOM_NET_COST_CACHE_TTL_SECONDS. Rust compression-policy tests pass (14/14) and git diff --check is clean. The red checks were GitHub Actions infrastructure failures before checkout (`Service Unavailable` resolving actions), so I reran the failed workflows.

JerrettDavis (approved):
Reviewed 045764f4. The only change since the approved implementation is a test-fixture retune after mainline pricing inputs moved. The new fixture still proves the intended invariant: the same candidate is admitted under five-minute pricing and rejected under the one-hour write tier, including the request-level one-hour path without the fallback environment override. Production source is unchanged from the previously approved head, all current checks pass or are intentionally skipped, and the PR is mergeable. Ready for final human review.

PR #2746, opened 2026-08-03
ci(docs): deploy Pages on wiki changes, drop the never-working Vercel job

chopratejas:
## Description

Two independent bugs in `.github/workflows/docs.yml`.

**1. Pages went stale because the push filter watched the wrong directory.**

`mkdocs.yml` sets `docs_dir: wiki`, but the push filter listed `docs/**` and not `wiki/**`. So a merge touching only `wiki/` never triggered the workflow and the published Pages site silently went stale, while a `docs/**`-only change triggered a Pages rebuild whose sources mkdocs doesn't even read.

Push now filters on `wiki/**` + `mkdocs.yml`. **Pull requests keep `docs/**`**, so `validate-nextjs` still catches a broken MDX change before merge.

**2. `deploy-vercel` never worked — it wasn't a working path that went stale.**

It ran `npx vercel deploy --prod --token=${{ secrets.VERCEL_TOKEN }}`, but those secrets don't exist:

```text
$ gh api repos/headroomlabs-ai/headroom/actions/secrets --jq '.secrets[].name' | grep -i vercel
(nothing)
$ gh api orgs/headroomlabs-ai/actions/secrets  --jq '.secrets[].name' | grep -i vercel
(nothing)
```

So it invoked the CLI with an empty `--token=` and exited 1, every time:

```text
30 failed runs on main, 2026-07-14 .. 2026-08-04

most recent:
  30874173333  13a310a0  failure
  30845343100  6422a80a  failure
  30810614546  007446c7  failure

per-job on 30874173333:
  failure  Deploy Vercel Docs
  success  Deploy GitHub Pages     <- same run
  skipped  Validate mkdocs build
  skipped  Validate Next.js build
```

`deploy-github-pages` succeeded in those same runs, so this job contributed nothing but a red X on every merge to main.

The Next.js site is published by **Vercel's own Git integration**, which is what has actually been deploying it. Removing this job leaves exactly one deploy path per site: Pages via mkdocs here, Vercel via its Git integration.

## Why this PR previously said the opposite

I opened this as "drop the redundant Vercel job", then converted it to draft and posted a correction saying the deletion was **wrong** — because at that point the Vercel site had been stale since mid-June and this looked like the only deploy path. That correction was itself based on a wrong assumption: with no `VERCEL_TOKEN` configured, this job could never have deployed anything. It wasn't the deploy path; it was a job that had always failed. The site was stale because *nothing* was publishing it until the Git integration was connected.

[...]

PR #2781, opened 2026-08-04
docs(configuration): document LiteLLM as the source for limits and prices

chopratejas:
## Description

The ten token/pricing fixes merged this session changed documented behaviour, and `configuration.mdx` / `wiki/configuration.md` were left describing the old model. This is docs-only.

**The main problem:** both trees documented resolution as

```
1. Built-in defaults
2. models.json
3. HEADROOM_MODEL_LIMITS
4. SDK constructor arguments
```

with **no mention of LiteLLM at all**. That is misleading in a way that costs users real work: it reads as though editing `models.json` is the only way to correct a wrong limit or price. In fact public models resolve from LiteLLM and stay current on their own — pinning one in config means signing up to maintain a number by hand forever.

Restated as an explicit first-match-wins list covering **both** limits and pricing:

```
1. explicit config (models.json / env / SDK args)   <- a decision, beats everything
2. LiteLLM                                          <- live source for public models
3. built-in tables -> pattern -> generic default    <- fallback only
```

…and noted that step 3 exists for installs where LiteLLM is unavailable (the dependency is gated `python_version < '3.14'`) and for models LiteLLM doesn't know.

## Changes Made

- **Fixed the example overrides.** Both trees used `"gpt-5": 256000` and `"gpt-5": [5.00, 15.00]`. Both numbers are wrong (272K input; $1.25/$10.00), and gpt-5 is now a *known* model — so the example was teaching readers to override a correct, self-updating value with a stale hand-written one. Swapped for `ft:gpt-4o:my-org` and `my-private-deployment`, which are what custom configuration is actually for, plus a callout saying so.
- **Documented that limits are an INPUT budget.** They come from LiteLLM's `max_input_tokens`, so `gpt-5` is 272K of prompt room, not 400K — the total window is 272K in + 128K out. I conflated these myself while writing #2762 and had to correct it, so it's worth stating for anyone configuring a limit.
- **Updated pattern inference** with `gpt-4.1*`, `gpt-5*`, `o4*`, and stated that matching prefers the **most specific** prefix — so `gpt-4.1` isn't treated as `gpt-4` (which is exactly the bug #2762 fixed: an 8K limit and legacy `gpt-4` pricing).

Kept in sync across both doc trees, since they carried identical stale text.

## Type of Change

- [x] Documentation update

## Checked and deliberately left alone

- `savings.mdx:50` — already correct: it says cost-avoided uses LiteLLM list pricing for proxy traffic. That path (`savings_ledger`) was always LiteLLM-backed; it was the *provider* tables that weren't.
- `opencode.mdx:62` — "GPT-4.1, 1M context, 32K output" is right. Worth noting the docs had the correct figure all along while the code resolved 8192.
- `wiki/metrics.md` — uses `tokens_input_before/after`, a different surface from the beacon's `tokens.input` that #2756 touched. No change needed.

## Testing

- [x] Documentation update

No code changed, so no unit tests. `<Callout type="info">` matches existing un-imported usage in `savings.mdx`, `ccr.mdx` and `benchmarks.mdx`; the docs workflow runs `validate-mkdocs` and `validate-nextjs` (`npm run build`) on any `docs/**` / `wiki/**` PR, which is the real check for both trees here.

## Review Readiness

## Checklist

- [x] My changes are documentation-only
- [x] I have kept both documentation trees in sync

### DIFF
--- a/docs/content/docs/configuration.mdx
+++ b/docs/content/docs/configuration.mdx
@@ -425,8 +425,11 @@ Configure context limits and pricing for new or custom models:
   },
   "openai": {
     "context_limits": {
-      "gpt-5": 256000,
-      "ft:gpt-4o:my-org": 128000
+      "ft:gpt-4o:my-org": 128000,
+      "my-private-deployment": 200000
+    },
+    "pricing": {
+      "my-private-deployment": [5.00, 15.00]
     }
   }
 }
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The central resolution-order claim does not match the current provider implementations yet. Both `AnthropicProvider` and `OpenAIProvider` initialize `_context_limits` with their built-in tables, merge user overrides into that same mapping, and return any exact/prefix match from it before attempting LiteLLM (`headroom/providers/anthropic.py::get_context_limit`, `headroom/providers/openai.py::get_context_limit`). This means known public models present in the built-in tables do not currently resolve as “explicit config → LiteLLM → built-ins.” Also, LiteLLM’s `model_cost`/model metadata ships with the installed LiteLLM version; calling it a “live source” implies a network-refreshed database that the code does not use. Please either align the docs with the actual per-provider precedence and describe LiteLLM as the installed model database, or pair this with the code changes and tests needed to make the stated unified order true. The custom-model examples and input-budget clarification are useful; this is focused on the new core precedence claim.

chopratejas:
You are right on both counts, and the first one is a real error in this PR. Thank you — I would have shipped a page describing a precedence the code does not implement.

**1. Built-in tables beat LiteLLM for context limits.** Confirmed empirically. I stubbed `_get_litellm_module` to return a distinctive `max_input_tokens=999999` and asked the provider:

```text
gpt-4o             ->  128000   BUILT-IN TABLE WON (litellm never consulted)
gpt-4              ->    8192   BUILT-IN TABLE WON
gpt-4.1            -> 1047576   BUILT-IN TABLE WON
gpt-5              ->  272000   BUILT-IN TABLE WON
gpt-4o-2024-08-06  ->  128000   BUILT-IN TABLE WON
some-unknown-model ->  999999   litellm used
gpt-4.1-preview    ->  999999   litellm used
```

Exactly as you said: `self._context_limits` is `{**_CONTEXT_LIMITS}` merged with user overrides, so the `model in self._context_limits` check at `openai.py:505` / `anthropic.py:598` cannot distinguish "user configured this" from "we shipped this", and any exact table entry short-circuits before LiteLLM.

The claim happens to be true for **pricing** — #2779 introduced a separate `_pricing_overrides` map precisely so config could outrank LiteLLM — but I never did the equivalent for limits and then documented both as one unified order. Worth noting `AnthropicProvider.get_context_limit`'s own docstring makes the same wrong claim ("4. LiteLLM model info / 5. Built-in ANTHROPIC_CONTEXT_LIMITS"), so this predates my page.

**2. "Live source" is wrong.** `litellm.model_cost` is a static dict shipped with the installed package; nothing refreshes it at runtime. Reworded throughout to "the installed LiteLLM model database", with an explicit note that it is as current as your installed LiteLLM version and updates when you upgrade.

**On your second suggested option — I deliberately did not take it.** You offered "or pair this with the code changes needed to make the stated unified order true." I audited both tables against LiteLLM before attempting that:

```text
openai _CONTEXT_LIMITS:   36 entries — agree=25  disagree=0  no-litellm-entry=11
ANTHROPIC_CONTEXT_LIMITS: 21 entries — agree=11  disagree=1  no-litellm-entry=9
    claude-sonnet-4-20250514    table=200000   litellm=1000000
```

For OpenAI, flipping precedence is a provable no-op — zero disagreements. For Anthropic it would change `claude-sonnet-4-20250514` from 200K to 1M, and that 1M window is an **opt-in beta**. LiteLLM reports maximum *capability*; Headroom needs the *effective default*. Raising the assumed budget to 1M for clients that have not enabled the beta would build prompts the upstream rejects.

So the built-in-first behaviour for limits is defensible rather than accidental, and forcing the unified order would be a regression. I have documented the asymmetry and why it exists instead — including the Sonnet 4 case as the concrete reason — rather than flatten it into a tidier claim that would be wrong.

Pushed as a rebase onto current `main` (this branch predated #2746). The custom-model examples and the input-budget clarification you called useful are unchanged.

Two follow-ups I would rather do separately than widen a docs PR:
- fix `AnthropicProvider.get_context_limit`'s docstring, which states the precedence backwards
- decide whether `claude-sonnet-4-20250514` should track the beta window when a client signals it, rather than being pinned to 200K

JerrettDavis (approved):
The revision addresses the requested correctness issue: it now distinguishes context-limit precedence from pricing precedence, accurately describes LiteLLM as the installed package database, and explains why built-in effective limits can intentionally outrank LiteLLM capability maxima. I rechecked both doc trees against the provider behavior; the custom-model examples and prefix-order guidance remain clear. Current docs, security, and governance checks are green.

Issue #790, opened 2026-06-09
[BUG] Optimization mode prevents proxy from binding on Windows (v0.24.0, Python 3.13)


gunturajip:
## Description

When running Headroom Proxy with optimization enabled (default behavior), the startup banner is displayed successfully, but the proxy never binds to the configured port and no HTTP listener is created.

The issue appears to be isolated to the optimization pipeline because the proxy starts and serves requests correctly when optimization is disabled using `--no-optimize`.

No error message or stack trace is shown.

## To Reproduce

1. Install Headroom 0.24.0 on Windows 11 with Python 3.13.
2. Start the proxy:

```powershell
headroom proxy --port 8787 --no-telemetry
```

3. Observe that the startup banner is displayed:

```text
Starting proxy server...
URL:          http://127.0.0.1:8787
Mode:         token
Optimization: ENABLED
Caching:      ENABLED
Rate Limit:   ENABLED
Memory:       DISABLED
Context Tool: rtk
```

4. Verify whether the port is listening:

```powershell
netstat -ano | findstr :8787
```

5. Attempt to access the proxy:

```powershell
curl http://127.0.0.1:8787/health
```

## Expected Behavior

The proxy should bind to port 8787 and become reachable through endpoints such as:

```text
http://127.0.0.1:8787/health
http://127.0.0.1:8787/livez
http://127.0.0.1:8787/readyz
```

`netstat` should show a listener on port 8787.

## Actual Behavior

The startup banner is displayed, but no listener is created.

```powershell
netstat -ano | findstr :8787
```

returns:

```text
(no output)
```

Requests fail:

```powershell
curl http://127.0.0.1:8787/health
```

returns:

```text
curl: (7) Failed to connect to 127.0.0.1 port 8787
```

## Code Sample

Not applicable. The issue occurs during proxy startup before any client requests are processed.

Startup command:

```powershell
headroom proxy --port 8787 --no-telemetry
```

## Error Output

No error output is displayed.

Additional observations:

* Cache is not the issue.
* Rate limiting is not the issue.
* Port binding works when optimization is disabled.

The following command starts successfully:

```powershell
headroom proxy ^
  --port 8787 ^
  --no-optimize ^
  --no-cache ^
  --no-rate-limit ^
  --stateless ^
  --no-telemetry
```

This suggests the failure is isolated to the optimization subsystem.

## Environment

* **Headroom version**: 0.24.0
* **Python version**: 3.13.13
* **OS**: Windows 11
[...]

PR #2799, opened 2026-08-05
fix(proxy): warm deferred kompress model so /health stops reporting backend=null

### LINKED ISSUE #2730: Proxy always reports kompress backend=null while the compressor works standalone (0.33.0, macOS arm64)
## Summary

The proxy always reports `kompress: unhealthy` / `backend: null`, even though the Kompress model, tokenizer, and ONNX runtime are all present and the compressor works correctly when invoked directly as a library. The kompress transform appears never to be wired into the proxy pipeline, so all compression falls through to SmartCrusher.

**Version:** headroom 0.33.0 (`uv tool install --python 3.13 "headroom-ai[all]"`)
**Platform:** macOS, Apple Silicon (arm64), Python 3.13

## Observed

`/readyz` and `/health`, on every start:

```json
"kompress": {"enabled": true, "ready": false, "status": "unhealthy", "backend": null}
```

The startup banner lists `Code-Aware: ENABLED (AST-based)` but never prints a Kompress line.

## The compressor itself works

Invoked directly against the tool venv's interpreter, it loads and compresses successfully:

```python
from headroom.transforms.kompress_compressor import KompressCompressor, KompressConfig
c = KompressCompressor(KompressConfig())
r = c.compress(text)   # -> KompressResult, tokens_saved = 27
```

So the model, tokenizer, and ONNX backend are all functional. The failure is confined to the proxy.

## Environment checks (all pass)

- ML deps import cleanly in the tool venv: `torch 2.13.0`, `transformers 5.14.1`, `onnxruntime 1.28.0`, `tokenizers 0.22.2`
- `chopratejas/kompress-v2-base` fully cached (`onnx/kompress-int8-wo.onnx`, ~274 MB blob, snapshot symlink intact)
- `answerdotai/ModernBERT-base` cached (fetched lazily on first direct `.compress()` call)

## Ruled out

- **Not mode-dependent.** Reproduced with both `--mode cache` (default) and `--mode token`, on separate ports. Identical `backend: null`.
- **Not the backend selector.** `HEADROOM_KOMPRESS_BACKEND` is unset, so it resolves to `auto`.
- **Not disabled.** `--disable-kompress` is not passed.
- **Not a preload timeout.** The proxy binds and reports ready in ~10s, far inside the 120s `HEADROOM_EAGER_PRELOAD_TIMEOUT_SECONDS`, and no "Eager preload exceeded" warning appears in the log.

## Suspected cause

In `headroom/proxy/server.py`:

- L1679 initialises `self._kompress_status = "not installed"`
- L~1711 only overwrites it via `if eager_status.get("kompress") in {"enabled", "deferred"}`
- `_eager_preload_transforms()` (L1574) builds `eager_status` purely from transforms that expose `eager_load_compressors()`

Since preload does run (`Optimization: ENABLED`, "Pre-loading compressors and parsers..." is logged) yet no `kompress` key is ever produced, it looks like no transform in either pipeline contributes one — i.e. the kompress transform isn't present in the pipeline, so the status is left at its `"not installed"` default regardless of the model actually being installed and working.

## Impact

All compression falls through to SmartCrusher. That still works well on uniform tabular JSON (measured 29–39% on record arrays), but the ML path for prose/code never engages, so the documented ~15–20% for coding agents isn't reachable via the proxy.

## Expected

With the model cached and the ML extras installed, the proxy should report `kompress: healthy` with a concrete backend (e.g. `onnx_cpu`), and route prose/code through Kompress.

## Suggested

Separately from the wiring fix, it would help a lot if `"not installed"` were distinguishable from "installed but not wired into the pipeline" — the current status makes a pipeline-wiring problem look like a missing dependency, which sends you debugging the wrong thing.

Parideboy:
## Description

On an idle proxy with the ML extras installed and the Kompress model cached, `/health` and `/readyz` reported `kompress: {"ready": false, "status": "unhealthy", "backend": null}` forever, and the startup banner logged `Kompress: not installed`.

The issue diagnosed this as "the kompress transform is not wired into the pipeline". That is not accurate — `enable_kompress` defaults to `True` and `ContentRouter` sits in both the Anthropic and OpenAI pipelines. The real chain is:

1. `ContentRouter.eager_load_compressors()` deliberately does not call `preload()`, so a native model load can never run before the port binds (#790 hang on Windows, #1908 segfault on RHEL/CentOS 7). It returns `kompress = "deferred"`.
2. `WarmupRegistry.merge_transform_status` only promotes `enabled` / `ready` / `loaded*`, so `"deferred"` leaves the slot at `status="null"`.
3. `/health` reads that slot, hence `ready=false, backend=null`.
4. `_reconcile_kompress_health()` can promote the slot, but only once a compressor reports `is_ready()`. Nothing warms the model after startup — the only loader kick, `ensure_background_load()`, lives in the compression request path. No kompress-routed request means unhealthy indefinitely.

Closes #2730

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `HeadroomProxy._start_kompress_background_warm()`: when startup reports `kompress = "deferred"`, a daemon thread (`kompress-startup-warm`) calls `compressor.preload(allow_download=False)` and promotes `warmup.kompress` with the returned backend. Cache-only means no network at startup, so a cold cache raises `KompressModelNotCached` and stays deferred exactly as today, and the port still binds immediately. Disable with `HEADROOM_KOMPRESS_BACKGROUND_WARM=0`.
- Router/compressor lookup lifted out of `create_app` into module-level `_kompress_routers()` / `_kompress_compressors()` so startup and the health endpoint share one implementation.
- Honest status reporting: the startup banner logs `Kompress: DEFERRED (wired into pipeline; model loads in background)`, and the kompress health check gains a `detail` field — `loaded` / `warming` / `model not cached` / `not installed` / the warm error. `enabled`, `ready` and `backend` semantics are unchanged, so the `/readyz` aggregate exclusion still holds.
- Tests: pinned the existing awaited-startup deferral test with the kill switch, added warm-cache promotion and cold-cache coverage, and updated the health/log expectations.

## Testing

### Test Output

```text
$ uv run --no-sync pytest tests/test_proxy_health.py tests/test_proxy_warmup.py tests/test_kompress_preload_deferral.py tests/test_proxy_eager_preload_bind.py -q
45 passed, 1 failed in 9.4s

FAILED tests/test_kompress_preload_deferral.py::test_proxy_startup_does_not_enter_cached_kompress_native_loader
E     - deferred
E     + unavailable

$ uv run --no-sync ruff check headroom tests
All checks passed!

$ uv run --no-sync ruff format --check headroom/proxy/server.py tests/test_proxy_health.py tests/test_kompress_preload_deferral.py tests/test_proxy_eager_preload_bind.py
4 files already formatted

$ uv run --no-sync mypy headroom/proxy/server.py --ignore-missing-imports --python-version 3.13
(no errors reported for headroom/proxy/server.py)
```

The single local failure is pre-existing on `upstream/main` in my environment and is not caused by this change — verified by swapping in the unmodified `main` copies of `headroom/proxy/server.py` and `tests/test_kompress_preload_deferral.py`, which fail identically. It comes from a second ContentRouter in the pipelines whose real `_get_kompress()` returns `None` on this machine (ML extras not fully installed), so `merge_transform_status` overwrites `source_status` with `unavailable`. CI passes this test, so it is environment-specific; worth a separate look.

`ruff check .` also reports 4 pre-existing errors in `plugins/headroom-oauth2/`, untouched by this PR.

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.11, pytest 9.1.1, branch rebased on `upstream/main` d0a86d4
- Exact command / steps: `uv run --no-sync pytest tests/test_kompress_preload_deferral.py tests/test_proxy_health.py tests/test_proxy_warmup.py tests/test_proxy_eager_preload_bind.py -q`, exercising the two new tests `test_startup_background_warm_promotes_deferred_kompress` (joins the warm thread, asserts `preload_calls == [False]`, slot `loaded`, backend `onnx`) and `test_startup_background_warm_leaves_cold_cache_deferred` (slot stays `null`, `detail == "model not cached"`)
- Observed result: the warm thread promotes the deferred slot with no request at all, and a cold cache stays deferred with zero network calls; the DEFERRED startup line replaces the misleading `not installed` line. Full CI on this branch is green.
- Not tested: a live proxy run against a real cached ModernBERT model on macOS arm64 (the reporter's platform). The ML extras are not fully installed in this environment, so the warm path is covered by stubs only.

## Review Readiness

## Checklist

N/A — no user-facing UI change.

## Additional Notes

- "Manual testing performed" is unchecked: no live proxy run against a real cached model, for the reason under Not tested above.
- Documentation is unchanged; the new `HEADROOM_KOMPRESS_BACKGROUND_WARM` kill switch is an escape hatch rather than a documented knob. Happy to add it to the env-var reference if maintainers prefer.

### DIFF
--- a/headroom/proxy/kompress_health_detail_policy.py
+++ b/headroom/proxy/kompress_health_detail_policy.py
@@ -1,0 +1,70 @@
+"""Policy for the public ``detail`` field on the kompress health check.
+
+``/health`` and ``/readyz`` are auth-exempt, so everything they serialize is
+unauthenticated public data. The kompress check reports ``detail`` because
+``ready=false`` on its own is ambiguous — the model may be warming in the
+background, missing from the local cache, or the ML extras may not be installed
+at all (#2730).
+
+That diagnostic value must not come at the cost of leaking internals. Native
+loader, model and cache failures routinely embed absolute filesystem paths,
+model identifiers, endpoint URLs and configuration values in their exception
+text, and a warmup slot's ``error`` string is populated from exactly those
+failures. So the public field draws from a closed vocabulary: anything not in
+it degrades to ``unavailable`` rather than being echoed.
+
+The precise cause is not lost — it stays in the proxy log and in the warmup
+slot, which ``/debug/warmup`` serializes behind authentication.
+"""
+
+from __future__ import annotations
+
+# The model is loaded and serving.
+KOMPRESS_DETAIL_LOADED = "loaded"
+# The background warm thread is still running.
+KOMPRESS_DETAIL_WARMING = "warming"
+# Cache-only warm found no local artifacts; no network call was made.
+KOMPRESS_DETAIL_NOT_CACHED = "model not cached"
+# The warm attempt raised. The cause is in the log, not here.
+KOMPRESS_DETAIL_WARM_FAILED = "warm failed"
+# The ML extras are absent, or no compressor was ever constructed.
+KOMPRESS_DETAIL_NOT_INSTALLED = "not installed"
+
+# ``ContentRouter.eager_load_compressors()`` source statuses. These are fixed
+# literals in the router, never interpolated, so they are safe to surface.
+KOMPRESS_DETAIL_ENABLED = "enabled"
+KOMPRESS_DETAIL_DEFERRED = "deferred"
+KOMPRESS_DETAIL_UNAVAILABLE = "unavailable"
+
+#: Every value ``/health`` and ``/readyz`` may report for ``kompress.detail``.
+KOMPRESS_HEALTH_DETAILS = frozenset(
+    {
+        KOMPRESS_DETAIL_LOADED,
+        KOMPRESS_DETAIL_WARMING,
+        KOMPRESS_DETAIL_NOT_CACHED,
+        KOMPRESS_DETAIL_WARM_FAILED,
+        KOMPRESS_DETAIL_NOT_INSTALLED,
+        KOMPRESS_DETAIL_ENABLED,
+        KOMPRESS_DETAIL_DEFERRED,
+        KOMPRESS_DETAIL_UNAVAILABLE,
+    }
+)
+
+#: What an out-of-vocabulary value collapses to.
+DEFAULT_KOMPRESS_HEALTH_DETAIL = KOMPRESS_DETAIL_UNAVAILABLE
+
+
+def public_detail(value: str | None) -> str:
+    """Return ``value`` if it is a known detail, else the bounded fallback.
+
+    This is the single choke point for the field. Route *every* candidate
+    through it — including strings that look safe at the call site, such as a
+    warmup slot's ``source_status`` — so a future code path cannot introduce a
+    leak by writing free-form text into the slot.
+    """
+    if not isinstance(value, str):
+        return DEFAULT_KOMPRESS_HEALTH_DETAIL
+    normalized = value.strip().lower()
+    if normalized in KOMPRESS_HEALTH_DETAILS:
+        return normalized
+    return DEFAULT_KOMPRESS_HEALTH_DETAIL
--- a/headroom/proxy/server.py
+++ b/headroom/proxy/server.py
@@ -142,6 +142,14 @@ from headroom.proxy.helpers import (
     resolve_display_provider,
     retry_after_ms,
 )
+from headroom.proxy.kompress_health_detail_policy import (
+    KOMPRESS_DETAIL_LOADED,
+    KOMPRESS_DETAIL_NOT_CACHED,
+    KOMPRESS_DETAIL_NOT_INSTALLED,
+    KOMPRESS_DETAIL_WARM_FAILED,
+    KOMPRESS_DETAIL_WARMING,
+    public_detail,
+)
 from headroom.proxy.loop_callback_failure_policy import is_known_websocket_callback_failure
 from headroom.proxy.loopback_guard import is_loopback_host
 from headroom.proxy.memory_handler import MemoryConfig, MemoryHandler
@@ -1751,11 +1759,14 @@ class HeadroomProxy(
                 try:
                     backend = compressor.preload(allow_download=False)
                 except not_cached:
-                    slot.info["detail"] = "model not cached"
+                    slot.info["detail"] = KOMPRESS_DETAIL_NOT_CACHED
                     continue
-                except Exception as exc:
-                    slot.info["detail"] = f"warm failed: {exc}"
-                    logger.debug("Kompress background warm failed: %s", exc)
+                except Exception:
+                    # The detail lands on the auth-exempt /health payload, so it
+                    # stays a bounded token; loader failures embed paths, model
+                    # ids and URLs. The real cause goes to the log only.
+                    slot.info["detail"] = KOMPRESS_DETAIL_WARM_FAILED
+                    logger.warning("Kompress background warm failed", exc_info=True)
                     continue
                 if not backend:
                     continue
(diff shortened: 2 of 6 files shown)


JerrettDavis (changes requested):
Blocking: the new unauthenticated /health and /readyz detail field can expose arbitrary exception text. _warm() stores `f"warm failed: {exc}"` in `slot.info["detail"]`, and _kompress_detail() also returns `slot.error` verbatim. Native/model/cache failures commonly include absolute filesystem paths, model identifiers, URLs, or configuration details, so this turns internal diagnostics into public response data. Please keep the precise exception in logs/debug-only state and map the public health detail to a bounded value such as `warm failed` / `unavailable`, with a regression test using an exception that contains a sentinel secret/path and asserting it is absent from both public endpoints. Reviewed at 44ff8f2d.

Parideboy:
Thanks @JerrettDavis — the finding is correct and I've fixed it in 05ee081a. `/health`, `/healthz`, `/livez` and `/readyz` are all in `_AUTH_EXEMPT_PATHS`, so the new `detail` field was unauthenticated public data, and both paths you identified could put raw exception text in it.

**What changed**

- New `headroom/proxy/kompress_health_detail_policy.py` (pure module, matching the existing `*_policy.py` convention) defines `KOMPRESS_HEALTH_DETAILS` — the closed vocabulary the field may report: `loaded`, `warming`, `model not cached`, `warm failed`, `not installed`, plus the three fixed `eager_load_compressors()` source statuses `enabled` / `deferred` / `unavailable`. `public_detail()` passes those through and collapses anything else to `unavailable`.
- The background warm thread now stores the bounded token `"warm failed"` and logs the real cause with `logger.warning(..., exc_info=True)` instead of interpolating the exception into the slot.
- `_kompress_detail()` no longer returns `slot.error` verbatim. Every candidate — including `source_status` — goes through `public_detail()`, so the allowlist is enforced at a single choke point and a future code path can't reintroduce the leak by writing free-form text into the slot.

The precise exception is not lost: it is in the proxy log, and `slot.error` still serializes in full via `/debug/warmup`, which is not auth-exempt.

**Regression tests**

- `tests/test_proxy_health.py`: `test_public_health_does_not_leak_warm_failure_text` and `test_public_health_does_not_leak_slot_error_text`, both parametrized over `/health` and `/readyz`. They plant `/opt/secret-path/model.onnx token=SENTINEL-LEAK-9f3a` in `slot.info["detail"]` and in `slot.error` respectively, then assert `detail == "unavailable"` and that neither `SENTINEL-LEAK-9f3a` nor `secret-path` appears anywhere in the response body.
- `tests/test_kompress_preload_deferral.py::test_startup_background_warm_failure_detail_is_bounded` drives the real warm thread with a compressor that raises the same sentinel-bearing message and asserts the slot ends up with exactly `"warm failed"`.
- `tests/test_kompress_health_detail_policy.py` covers the policy directly: vocabulary membership, path/URL/token-bearing strings collapsing, `None`/empty/non-string input, and normalization.

`enabled`, `ready` and `backend` semantics are untouched, so the `/readyz` aggregate exclusion still holds.

Local: `ruff check headroom tests` clean, `ruff format --check` clean, mypy reports no errors for `server.py` or the new module, and the five affected test files pass apart from the pre-existing environment-specific failure in `test_proxy_startup_does_not_enter_cached_kompress_native_loader` already documented in the PR body (ML extras not fully installed here; green in CI).

Ready for re-review.

JerrettDavis (approved):
Reviewed 05ee081a. The public health-detail leak is resolved on this head. Every candidate, including raw warmup slot errors and source status, now passes through a closed vocabulary; background exceptions are logged without being serialized by the auth-exempt health endpoints. The sentinel regressions cover both health and ready endpoints and both prior leak paths. Current checks are green and the PR is mergeable. Ready for final human review.

JerrettDavis (approved):
Re-reviewed the rebased two-commit series with `git range-diff` against the previously approved head. The public health-detail hardening remains unchanged: all raw warmup errors/source status pass through the closed-vocabulary policy, with sentinel regressions covering both `/health` and `/readyz`. The rebase only integrates `main`’s optional-component health schema, correctly reporting deferred Kompress as `degraded`/`optional` without contradicting aggregate readiness.

The exact current head is mergeable and its full applicable CI rollup is green.

PR #2677, opened 2026-07-30
fix: remove rtk and lean-ctx CLI context tools

chopratejas:
## Description

Removes both third-party CLI context tools — **rtk** and **lean-ctx** — and with them the context-tool selector itself. Headroom no longer downloads, installs or configures either one, and there is no replacement.

The previous pass (#2344) gated only three entry points inside `headroom/cli/wrap.py`. That left the feature reachable in practice:

| Gap | Effect |
|---|---|
| `scripts/install.sh:1544`, `install.ps1:1681` | Ran `rtk init --global --auto-patch` from bash/PowerShell, **bypassing the Python gate entirely** — `curl \| sh` still wrote a Claude Code `PreToolUse` hook regardless of `HEADROOM_RTK` |
| `wrap.py` `_setup_context_tool_for_agent` | **`wrap openhands` was broken by default**: `rtk_required=True` met a gate returning `None` → `SystemExit(1)`. Invisible because all 8 openhands tests patched `_ensure_rtk_binary` to a fake path |
| `proxy/helpers.py`, `subscription/tracker.py` | Proxy shelled out to `rtk gain` from `/stats`, the dashboard and `headroom perf`; the tracker polled it per contribution (`_RTK_WIRING_DEFAULT = "enabled"`) |
| No cleanup path | Nothing removed artifacts an earlier default had installed, so a machine that once ran the old default kept rtk in the loop forever (#1669, #1955) |

Also worth noting: the rtk binary download had **no SHA or signature verification** — only `rtk --version` as a smoke test.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring (no functional changes)

## Changes Made

[...]

PR #1698, opened 2026-07-02
fix(rtk): link managed rtk onto PATH instead of mutating the hook

tenderdeve:
## Description

`headroom wrap claude` / `headroom update` patched `~/.claude/hooks/rtk-rewrite.sh` after `rtk init --global --auto-patch` wrote it. `rtk` bakes the expected SHA-256 of the canonical hook into itself, so the post-write mutation trips its integrity guard — `rtk verify` reports `hook integrity check FAILED … RTK will not execute` and rtk hard-refuses to run. The patch also only absolutized the `rtk` inside the hook, but `rtk rewrite` emits a bare `rtk` on stdout at runtime that still needs PATH resolution, so the original silent-no-op (#487) was never actually fixed. This leaves the hook untouched and instead links the managed binary onto PATH.

Closes #1631

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Removed `_patch_rtk_hook_absolute_path` (mutated the canonical hook → broke rtk's SHA-256 integrity guard).
- Added `_ensure_rtk_on_path`: symlinks the Headroom-managed `rtk` into a PATH dir (prefers `~/.local/bin`) so the bare `rtk` that `rtk rewrite` emits resolves, leaving the hook byte-for-byte as `rtk init` wrote it.
- No-op when a `rtk` already resolves on PATH, on Windows, or when no writable PATH dir exists; never clobbers an existing real file or foreign binary.
- Rewrote the test module (`test_wrap_rtk_on_path.py`) for the new behavior.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ .venv/bin/python -m pytest tests/test_cli/test_wrap_rtk_on_path.py -q
collected 7 items
tests/test_cli/test_wrap_rtk_on_path.py .......                          [100%]
============================== 7 passed in 0.25s ===============================

$ .venv/bin/ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_rtk_on_path.py
All checks passed!
```

## Real Behavior Proof

[...]

Issue #487, opened 2026-05-20
[BUG] RTK hook script fails because RTK not in path


jessetan:
## Summary

When headroom wrap claude registers RTK hooks, the generated hook script (~/.claude/hooks/rtk-rewrite.sh) fails silently because it searches for rtk via PATH lookup instead of using an absolute path. Since ~/.headroom/bin is not automatically added to PATH, Claude never actually calls RTK, defeating token compression.

## Steps to Reproduce

1. Run headroom wrap claude on a fresh system
2. Check ~/.claude/hooks/rtk-rewrite.sh — it contains command -v rtk (PATH search)
3. Verify ~/.headroom/bin is NOT in your $PATH
4. Use Claude Code — RTK commands are not executed
5. Check Claude's settings: cat ~/.claude/settings.json | grep rtk-rewrite
  - Hook path is absolute: /Users/username/.claude/hooks/rtk-rewrite.sh ✓
  - But the script inside uses command -v rtk (PATH lookup) ✗

## Expected Behavior

The hook script should be able to locate and execute the RTK binary that Headroom just downloaded, regardless of PATH.

## Actual Behavior

The hook script fails the command -v rtk check (line 18-20) because RTK isn't in PATH, then silently exits. Token compression never occurs.

## Root Cause

When rtk init --global --auto-patch is invoked with an absolute path (e.g., /Users/username/.headroom/bin/rtk), the resulting hook script should embed that absolute path. Instead, RTK creates a generic hook that depends on PATH lookup.

## Possible solutions
- Modify Headroom to add ~/.headroom/bin to PATH when wrapping agents
- Post-process the RTK hook script to replace rtk calls with the absolute path after rtk init --global runs
- File an issue with RTK to support adding absolute paths in hook script
- Manual workaround: add export PATH="$HOME/.headroom/bin:$PATH" to shell profile, or install RTK system-wide via Homebrew

## Environment

Headroom: v0.22.2
RTK: v0.28.2 (bundled)
Claude Code: any version

warnes:
Same issue, but with copilot.

dmitriid:
The bug is still there

```
❯ uname -a
Linux dmitriid-cachy 7.0.3-1-cachyos #1 SMP PREEMPT Fri, 01 May 2026 16:30:15 +0000 x86_64 GNU/Linux

❯ headroom -v
headroom, version 0.22.2
```

[...]

PR #2828, opened 2026-08-06
fix(context-tool-cleanup): delete only the rtk / lean-ctx artifacts Headroom installed

### LINKED ISSUE #2817: [BUG] headroom wrap deletes a user's own rtk / lean-ctx install (MCP entry, hook scripts, hook entries)
## Description

`headroom wrap` deletes a lean-ctx installation the user set up themselves, not just the one
older Headroom versions installed.

Every `headroom wrap` and `headroom unwrap` run calls `purge_context_tool_artifacts()`
(`headroom/context_tool_cleanup.py`, added in `e0ce4b1d`, #2677). Its MCP step matches by name
and nothing else:

```python
# headroom/context_tool_cleanup.py:75
_MCP_SERVER_NAMES = ("lean-ctx", "lean_ctx", "rtk")

# headroom/context_tool_cleanup.py:254, in _purge_mcp_entries()
removed = [name for name in _MCP_SERVER_NAMES if name in servers]
```

The entry disappears from `~/.claude.json` whoever wrote it. A registration from the user's own
`lean-ctx init` looks exactly like Headroom's retired one, so Headroom drops it.

That breaks the promise the module makes about itself:

```
* only files Headroom installed (or caused a context tool to install) are
  deleted;
* ``~/.local/bin/{rtk,lean-ctx}`` is unlinked only when it is a symlink into
  Headroom's own bin directory — a user's own build is never touched;
```

The binary path keeps that promise. The MCP path never checks:

```python
# headroom/context_tool_cleanup.py:281, _remove_managed_path_link()
if managed_dir not in target.parents:
    return []   # a user's own build is left alone
```

Two neighbouring steps delete by name as well, and they hit a user-owned install just as hard:

- step 2 (lines 96-99) unlinks `~/.claude/hooks/lean-ctx-rewrite.sh`, `lean-ctx-redirect.sh` and
  the `-native` variants. `lean-ctx init --agent` writes those files, whoever runs it.
- step 1 (`_HOOK_COMMAND_MARKERS`, lines 44-50) strips any hook whose command mentions
  `lean-ctx hook`, `lean-ctx-rewrite` or `lean-ctx-redirect` from `~/.claude/settings.json` and
  `~/.cursor/hooks.json`.

Reinstalling lean-ctx does not survive, because the purge runs on every launch. The next
`headroom wrap claude` tears it down again.

## To Reproduce

1. Install headroom (`pipx install headroom-ai`) at a version containing `e0ce4b1d`.
2. Install lean-ctx yourself and register it: `lean-ctx init` writes `mcpServers["lean-ctx"]`
   into `~/.claude.json` and drops its hook scripts in `~/.claude/hooks/`.
3. Check that the entry is there.
4. Run `headroom wrap claude`.
5. Read `~/.claude.json` again. The `lean-ctx` entry is gone, and stderr carried one line:
   `Retired CLI context tool cleanup: removed the lean-ctx MCP server entry from ~/.claude.json`

## Expected Behavior

Headroom removes what Headroom installed. A lean-ctx MCP registration, hook script or hook entry
the user created survives `headroom wrap`, the way `~/.local/bin/lean-ctx` already survives when
it is not a symlink into Headroom's managed bin directory.

## Actual Behavior

Every `headroom wrap` and `headroom unwrap` run removes:

- `mcpServers["lean-ctx"]` from `~/.claude.json`, and `mcp["lean-ctx"]` from `opencode.json`,
  whatever the entry's `command` points at
- `~/.claude/hooks/lean-ctx-*.sh` and the `-native` scripts
- any `~/.claude/settings.json` hook whose command mentions lean-ctx

The binary stays on disk, so nothing fails at the shell. The tool is simply gone from the
harness, and the agent stops seeing it.

## Code Sample

```bash
# a lean-ctx registration the user made, unrelated to Headroom
lean-ctx init
python - <<'PY'
import json, pathlib
cfg = json.loads((pathlib.Path.home() / ".claude.json").read_text())
print("before:", sorted(cfg.get("mcpServers", {})))
PY

headroom wrap claude

python - <<'PY'
import json, pathlib
cfg = json.loads((pathlib.Path.home() / ".claude.json").read_text())
print("after: ", sorted(cfg.get("mcpServers", {})))
PY
```

Here, `lean-ctx` vanished from the user-scope `mcpServers` while `codegraph`, `serena` and
`headroom` stayed. That is the exact-name match above, doing what it was written to do.

## Error Output

```
Retired CLI context tool cleanup: removed the lean-ctx MCP server entry from /home/<user>/.claude.json
```

No error, no exit code. The purge reports on stderr and keeps going, so the line scrolls past
behind the wrapped tool's own output.

## Environment

- **Headroom version**: 0.33.0 (pipx `headroom-ai`); the code is in that build, in `v0.34.0`, and
  on current `main` (`e0ce4b1d`)
- **Python version**: 3.13 (headroom venv), 3.14.6 system
- **OS**: Arch Linux, kernel 7.1.5
- **LLM Provider**: Anthropic (`headroom wrap claude`)

## Additional Context

The fix that matches the existing design: before dropping an MCP entry, resolve its `command`
and remove the entry only when it points into `paths.bin_dir()`. That is the same guard
`_remove_managed_path_link` already applies to the PATH symlink. Hook scripts and hook commands
want the same treatment, keyed on whether the script body references Headroom's managed binary.

A smaller change would also work: run the purge once, record that it ran, and skip it afterwards.
Cleaning up leftovers from a retired integration is a migration. It does not need to happen on
every launch, and doing it on every launch is what makes a fresh lean-ctx install impossible to
keep.

Happy to send a PR for either approach, with tests covering a user-owned entry surviving the
purge. Say which one you prefer and I will open it.

---

**Correction (see the comment below):** the two approaches above are not alternatives. Each
covers what the other cannot, and the naive form of the guard described here reproduces the
bug on a sibling directory such as `~/.headroom/binaries/`. The comment has the details, along
with what rtk changes about any provenance check.

davdittrich:
## Description

`purge_context_tool_artifacts()` deleted rtk and lean-ctx artifacts by name, so it removed an install the user had set up themselves, and it did so on every `wrap` and `unwrap` invocation. This makes deletion conditional on provenance, stamps the machine-global migration once it completes, and continues checking invocation-scoped project/config roots idempotently.

Closes #2817

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

Breaking in one direction only: artifacts that cannot be proved to be Headroom's are now kept. See "Behaviour the maintainer is accepting" below.

## Changes Made

- An artifact counts as Headroom's only when it names a path inside `paths.bin_dir()` — the same rule `_remove_managed_path_link` already applied to the PATH symlink, now applied to the three steps that matched by name: MCP entries, hook scripts, and hook entries in `settings.json` / `hooks.json`.
- The match lands on a directory boundary and normalizes the candidate first, so `~/.headroom/binaries/lean-ctx` and `~/.headroom/bin/../evil/lean-ctx` are not Headroom's. It scans raw text rather than pre-splitting on whitespace, so a path containing a space survives, and it tolerates quoting, `VAR=` prefixes and `PATH=…:` joins.
- Hook scripts are classified once, up front; both the entry-pruning and script-deletion steps read the same verdicts, so neither depends on the other's success.
- A hook entry inherits the verdict of the script it names, matched by absolute path rather than basename — a user's own `~/mytools/lean-ctx-rewrite.sh` no longer inherits ours. A relative command is not recognised: the harness resolves it against the project cwd, not home.
- `.rtk-hook.sha256` holds a hex digest and can never name a path, so it follows `rtk-rewrite.sh`'s verdict; `<name>.lean-ctx.bak` follows `<name>`.
- A script that exists but cannot be read is unprovable: deleted by nothing, named in the report.
- The first run that settles machine-global cleanup stamps `.context-tools-purged` beside the managed bin directory. Later runs skip that global half but continue checking the current project, `CODEX_HOME`, `OPENCODE_HOME`, OpenCode MCP config, and Continue config. An unresolved global artifact prevents the stamp; scoped cleanup remains retryable independently.

## Behaviour the maintainer is accepting

- **rtk artifacts from after #1698 now survive permanently.** `140cb05f` stopped patching the absolute managed path into `rtk-rewrite.sh` three weeks before #2677 retired the tools, so a hook from that window execs a bare `rtk` and is indistinguishable from one a user wrote. It survives while step 3 still removes the binary, leaving a hook that silently no-ops (#487). Hooks written before #1698 carry the path and are still cleaned.
- **A lean-ctx install that predated Headroom on `PATH` survives.** `get_lean_ctx_path()` checked `PATH` first, so Headroom drove the user's own binary and the config it wrote is byte-identical to theirs.
- **Only machine-global cleanup stops after it completes.** Project- and config-directory-scoped fenced guidance still runs idempotently because later invocations can use a different cwd, `CODEX_HOME`, or `OPENCODE_HOME`.

## Known limitations

- A hook *entry* whose command quotes a script path containing a space is not recognised, because that path match tokenizes on whitespace and boundary punctuation. Script *bodies* and MCP commands are unaffected — those scan raw text. `shlex.split` would close it; it is a behaviour change and belongs in its own commit.
- A script body that only mentions the managed path in a comment counts as Headroom's. Marked in the source with its upgrade path.
- A `<name>.lean-ctx.bak` whose principal was already deleted has no verdict to inherit and is left in place.

## Testing

### Test Output

```text
$ uv run pytest tests/test_context_tool_cleanup.py tests/test_cli/test_unwrap_claude.py -q
41 passed in 0.21s

$ uv run ruff check .
All checks passed!

$ uv run mypy headroom/context_tool_cleanup.py
```

Full suites on the fork, against this branch: CI, Security, Wrap E2E, Wrap Native E2E, Init E2E, Init Native E2E, Install Native E2E, Validate Docs, Dev Containers.

## Real Behavior Proof

- Environment: Arch Linux kernel 7.1.5, Python 3.13 in the headroom venv, headroom installed with pipx as `headroom-ai`.
- Exact command / steps: ran `headroom wrap claude` on the reporter's machine before the fix, then read `~/.claude.json`; afterwards ran `uv run pytest tests/test_context_tool_cleanup.py tests/test_cli/test_unwrap_claude.py -q` against the branch, and dispatched all nine fork workflows on commit 923cc2a8.
- Observed result: before the fix, `mcpServers["lean-ctx"]` was gone from `~/.claude.json` while `codegraph`, `serena` and `headroom` survived — the exact-name match this PR replaces. After the fix, 41 tests pass, covering a user-owned MCP entry, hook script and hook entry surviving a purge while an artifact naming `paths.bin_dir()` is still removed; Wrap E2E is green on the fork.
- Not tested: no manual `headroom wrap` run against the fixed build on the development machine, because that spawns a proxy on the port a live session already holds; `docker-wrap-e2e` covers the wrap path instead. The `rust`, `Merge Conflicts`, `opencode-plugin` and `Changelog Guard` workflows are `pull_request`-only, so they have not run anywhere yet — they need maintainer approval on this PR.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -750,7 +750,7 @@ _RETIRED_CONTEXT_TOOL_MESSAGE = (
     "rewrote shell commands through a third-party binary Headroom no longer "
     "manages. Drop --context-tool / --no-context-tool and unset "
     f"{_RETIRED_CONTEXT_TOOL_ENV}; `headroom wrap` uninstalls what they left "
-    "behind on first run."
+    "behind automatically."
 )
 
 
@@ -810,8 +810,10 @@ def _report_context_tool_purge() -> None:
     default: the Claude ``PreToolUse`` hook, the vendored binaries and the
     injected hint-file guidance are all durable on disk. Running this once per
     ``wrap`` / ``unwrap`` invocation is what actually makes the tools go away.
-    Silent when there is nothing to do, which is the steady state after the first
-    run, and never fatal — a cleanup failure must not block launching the tool.
+    Silent when there is nothing to do — the common case once the machine-global
+    half is stamped done, though the project- and config-directory-scoped half
+    still runs every launch — and never fatal: a cleanup failure must not block
+    launching the tool.
 
     Reports on **stderr**: some subcommands (``wrap/unwrap openclaw
     --prepare-only``) emit machine-readable JSON on stdout as their entire
--- a/headroom/context_tool_cleanup.py
+++ b/headroom/context_tool_cleanup.py
@@ -16,14 +16,37 @@ remove what earlier versions installed.
 Everything here is idempotent, best-effort and deliberately conservative:
 
 * only files Headroom installed (or caused a context tool to install) are
-  deleted;
+  deleted — an MCP entry's ``command``, a hook entry's ``command``, or a hook
+  script's body counts as Headroom's only when it names a path inside
+  :func:`paths.bin_dir`, compared on normalized forms
+  (:func:`_references_managed_bin`): the text is split on whitespace into
+  path-shaped tokens, each is lexically normalized, and a token counts only
+  on exact equality with the managed directory or a path separator right
+  after it — a sibling like ``bin-backup`` and a ``bin/../evil`` traversal
+  never match;
+* ``.rtk-hook.sha256`` is never read for its own provenance (it holds a hex
+  digest, not a path) and instead inherits ``rtk-rewrite.sh``'s
+  classification; a ``<name>.lean-ctx.bak`` backup inherits ``<name>``'s
+  (:func:`_classify_hook_scripts`);
+* a hook script that exists but cannot be read is classified unknown —
+  deleted by nothing, and named in the report so the user can remove it by
+  hand;
 * ``~/.local/bin/{rtk,lean-ctx}`` is unlinked only when it is a symlink into
   Headroom's own bin directory — a user's own build is never touched;
 * a JSON config that does not parse is reported and **skipped**, never
   overwritten (a hand-edited typo must not cost the user their settings);
 * the tools' own backups of *config* files (``~/.claude.json.lean-ctx.bak`` and
   friends) are left in place — they hold the user's real settings history. Only
-  backups of the hook scripts being deleted are cleaned up.
+  backups of the hook scripts proven to be Headroom's are cleaned up;
+* one case cannot be decided at all, and is accepted as a limit rather than
+  fixed: ``get_lean_ctx_path`` used to check ``PATH`` before Headroom's own
+  bin directory, so on a machine that already had ``lean-ctx`` on ``PATH``,
+  the tool that ran was the user's own, and the config it wrote looks exactly
+  like config the user wrote by hand. That leftover survives the purge — it
+  still points at a binary that exists, so nothing dangles;
+* the marker-fenced guidance block is the one step with no provenance check
+  to make — ``<!-- headroom:rtk-instructions -->`` is Headroom's own fence,
+  and no third party writes it.
 """
 
 from __future__ import annotations
--- a/headroom/context_tool_cleanup.py
+++ b/headroom/context_tool_cleanup.py
@@ -78,6 +78,13 @@ Everything here is idempotent, best-effort and deliberately conservative:
 * the marker-fenced guidance block is the one step with no provenance check
   to make — ``<!-- headroom:rtk-instructions -->`` is Headroom's own fence,
   and no third party writes it.
+
+Removing the retired integration is a one-time migration, so the first
+completed run stamps ``.context-tools-purged`` beside the managed bin
+directory and every later run returns immediately. Without it this would keep
+rewriting the user's own config files on every ``wrap`` invocation forever,
+and a user who installs one of these tools *after* the migration would have
+Headroom auditing their files at each launch for a leftover that cannot exist.
 """
 
 from __future__ import annotations
@@ -144,6 +151,10 @@ def purge_context_tool_artifacts() -> list[str]:
     empty list means there was nothing to do, which is the steady state after
     the first run.
     """
+    marker = _purge_marker()
+    if marker is not None and marker.exists():
+        return []
+
     home = Path.home()
     project = Path.cwd()
     report: list[str] = []
@@ -193,7 +204,26 @@ def purge_context_tool_artifacts() -> list[str]:
         report += _purge_fenced_block(hint_file)
     report += _purge_continue_system_messages(project / ".continue" / "config.json")
 
+    if marker is not None:
+        try:
+            marker.parent.mkdir(parents=True, exist_ok=True)
+            marker.touch()
+        except OSError:
+            pass  # Unwritable workspace: the purge simply runs again next time.
+
     return report
+
+
+def _purge_marker() -> Path | None:
+    """Path of the "already migrated" stamp, or ``None`` if it cannot be located.
+
+    Derived from :func:`paths.bin_dir` rather than ``workspace_dir`` so the
+    stamp always lands beside the binaries this module governs.
+    """
+    try:
+        return paths.bin_dir().parent / ".context-tools-purged"
+    except OSError:
+        return None
 
 
 def _instruction_files(home: Path, project: Path) -> list[Path]:
--- a/headroom/context_tool_cleanup.py
+++ b/headroom/context_tool_cleanup.py
@@ -35,12 +35,15 @@ Everything here is idempotent, best-effort and deliberately conservative:
   ``_HOOK_SCRIPTS`` files under ``~/.claude/hooks`` the classification map
   already marked ``True`` (:func:`_names_a_managed_script`): it inherits that
   script's verdict rather than being checked again. The path match tolerates
-  a wrapper (``bash <script>``), quoting, a redundant ``./``, and a form
-  written relative to home — but not an unexpanded ``$HOME``-style variable,
-  which is simply not recognised (unprovable, so kept). The map is built
-  from ``~/.claude/hooks`` only, so a Cursor ``hooks.json`` entry naming a
-  script under ``~/.cursor`` can never inherit a verdict this way — it is
-  still caught if its ``command`` names the managed bin directory directly;
+  a wrapper (``bash <script>``), quoting, and a redundant ``./`` segment, but
+  matches the script's absolute path only — never a relative form, since a
+  relative hook command resolves against the harness's project cwd, not
+  against home, and this module has no business touching a project-local
+  script; nor an unexpanded ``$HOME``-style variable. Both are simply not
+  recognised (unprovable, so kept). The map is built from ``~/.claude/hooks``
+  only, so a Cursor ``hooks.json`` entry naming a script under ``~/.cursor``
+  can never inherit a verdict this way — it is still caught if its
+  ``command`` names the managed bin directory directly;
 * ``.rtk-hook.sha256`` is never read for its own provenance (it holds a hex
   digest, not a path) and instead inherits ``rtk-rewrite.sh``'s
   classification; a ``<name>.lean-ctx.bak`` backup inherits ``<name>``'s
@@ -239,6 +242,11 @@ def _opencode_home(home: Path) -> Path:
 # joiners) and `()` (subshells) all end a path the same way whitespace does.
 _PATH_BOUNDARY_CHARS = frozenset("\"'=:;,()|")
 
+# Splits a command into path-shaped tokens on whitespace plus the same
+# boundary punctuation above — used by _names_a_managed_script, which (unlike
+# _references_managed_bin's needle-anchored scan) tokenizes the whole command.
+_PATH_TOKEN_SPLIT = re.compile(r"[\s" + re.escape("".join(_PATH_BOUNDARY_CHARS)) + r"]+")
+
 
 def _references_managed_bin(text: str) -> bool:
     """Whether ``text`` names a path inside Headroom's managed bin directory.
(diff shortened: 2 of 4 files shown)


chopratejas:
Hi, thanks for the change - can you fix the linting issues etc.

davdittrich:
> Hi, thanks for the change - can you fix the linting issues etc.

The failing CI checks are all "Error: Service Unavailable" and not actually failing checks. GitHub is experiencing issues affecting GitHub Actions at the moment, and all three failures died before checkout, on `Failed to resolve action download info` — the runner could not download `actions/checkout@v7` itself. No step ran, so nothing was linted or tested:

- `changes` (CI) and `commitlint` (CI) — both `##[error]Service Unavailable`
- `Dependency audit (pip-audit)` (Security) — same
- `rust-changes` cancelled; every downstream job SKIPPED because `changes` gates them

githubstatus.com confirms it: Actions `major_outage`, incident opened 15:22 UTC, our jobs started 15:43.

I ran the equivalents locally on the same SHA (923cc2a8) to be sure:

```text
$ ruff check .
All checks passed!

$ ruff format --check .
1370 files already formatted

Success: no issues found in 514 source files

$ commitlint --config .commitlintrc.json --from $(git merge-base origin/main HEAD) --to HEAD
found 0 problems, 0 warnings   (every commit in the branch)

$ pytest tests/test_context_tool_cleanup.py tests/test_cli/test_unwrap_claude.py -q
41 passed
```

These same jobs also passed on my fork on this exact SHA before the outage (CI and Security workflows, all green). `commitlint` was the one check the fork run could not cover — the job is gated on `github.event_name == 'pull_request'`, so a `workflow_dispatch` skips it — which is why I ran it locally above.

One genuine finding, unrelated to this PR: `pip-audit` now reports `h2 4.3.0` / `CVE-2026-71554` (GHSA-6hr6-w5qg-qmwg), fixed in 4.4.1. The advisory was published today at 21:53 UTC, after the runs above. This PR touches three Python files and leaves `uv.lock` and `pyproject.toml` byte-identical to `main`, so the audit gate will fail on `main` and on every open PR until h2 is bumped. Worth its own change rather than folding into this one — happy to open it if that helps.

Once Actions recovers, a re-run should clear these. I do not have the rights to trigger one on this repo.

JerrettDavis (changes requested):
The provenance direction is necessary, but I found two destructive-safety blockers on the latest head.

1. The global `.context-tools-purged` marker is written after scanning only the current `Path.cwd()` project plus the currently selected CODEX/OPENCODE homes. If an upgraded user has retired fenced guidance in projects A and B, the first wrap in A stamps the machine and every later wrap in B returns before inspecting B. The same applies when CODEX_HOME or OPENCODE_HOME changes. This contradicts the claim that a completed run means no leftover can exist. Please separate global/home artifacts from invocation-scoped project/config roots (for example, keep project cleanup running idempotently or use scope-specific stamps), and add a test that purges project A, changes cwd to project B containing an old fenced block, and proves B is still cleaned.

2. `_names_managed_dir` validates only the character after the matched needle. It does not require a path boundary before the occurrence. Therefore a user-owned absolute path such as `/prefix<absolute-managed-bin>/lean-ctx` contains the managed path as a substring and is classified as Headroom-owned, even though it resolves outside the managed directory. For code that deletes files/config, both boundaries are required. Please require start-of-string or a valid leading delimiter/path boundary, and add negative tests for prefixed absolute paths in script bodies and MCP commands.

The unreadable-state deferral and exact managed-script inheritance otherwise look carefully designed, but these two cases need to be closed before the one-time migration is safe.

davdittrich:
Both blockers are closed on `afa65b93`.

**Blocker 2 — leading path boundary (`f0a5be92`)**

`_names_managed_dir` now counts an occurrence only when it starts at offset 0 or is preceded by whitespace or a `_PATH_BOUNDARY_CHARS` delimiter. A rejected occurrence `continue`s the scan rather than returning, so a prefixed lookalike cannot mask a genuine reference later in the same string. `_PATH_BOUNDARY_CHARS` deliberately excludes `/`, which is what makes `/prefix<managed-bin>` reject while `PATH=<dir>`, `"<dir>/rtk"`, `$PATH:<dir>` and `exec <dir>/rtk` still match. It is the single shared guard behind all three deletion paths (`_classify_hook_scripts`, `_references_context_tool`, `_purge_mcp_entries`), so one condition covers script bodies, hook registrations and MCP commands.

Tests that failed before the change:
- `test_leaves_a_hook_script_whose_body_has_the_bin_dir_as_a_path_segment_alone` — script body
- `test_leaves_an_mcp_entry_whose_command_has_the_bin_dir_as_a_path_segment_alone` — MCP command
- `test_leaves_a_hook_entry_whose_command_has_the_bin_dir_as_a_path_segment_alone` — hook registration (carries a `_HOOK_COMMAND_MARKERS` token, otherwise `_references_context_tool` returns before reaching the guard and the test would be vacuous)

Plus `test_removes_a_hook_script_whose_body_has_a_lookalike_prefix_before_a_genuine_managed_path`, which pins `continue` over `return False`.

Deliberate narrowing worth naming: a managed path preceded by a non-boundary character (`file:///…`, `exec </dir/x`, `${X}/dir/x`, `#!/dir/x`) is now classified user-owned. That matches the token rule `_names_a_managed_script` already applies, and on a code path that deletes files the false negative — a leftover survives — is the recoverable direction.

**Blocker 1 — stamp scope (`afa65b93`)**

Split the entry point. The machine-global half (hook registrations, hook scripts, PATH symlinks, managed binaries, Claude Code's `~/.claude.json` MCP) stays gated by `.context-tools-purged`. The invocation-scoped half — OpenCode's `opencode.json`, the hint files under `Path.cwd()` / `CODEX_HOME` / `OPENCODE_HOME`, and Continue's `.continue/config.json` — moved into `_purge_invocation_scoped` and runs on every call. Those steps are read-only when nothing matches: `_purge_fenced_block`, `_purge_continue_system_messages` and `_purge_mcp_entries` each return before any write when their target is absent or unchanged, so the steady-state cost is a handful of reads per `wrap` process launch.

The stamp is now written *before* the scoped steps run, so it certifies only the half it actually covered: a scoped deferral no longer withholds it, while a global deferral still does. That is the one semantic change, and it is pinned by `test_a_scoped_deferral_does_not_withhold_the_global_stamp`.

Tests:
- `test_a_completed_purge_still_cleans_a_different_project` — the case you asked for: purge to completion in A, `chdir` to B holding a fenced `AGENTS.md`, assert B's fence is gone and the surrounding prose survives
- `test_a_completed_purge_still_inspects_a_repointed_codex_home` — `CODEX_HOME` repointed after the stamp
- `test_a_completed_purge_never_runs_again` unchanged and still green, pinning that the global half stays one-time

You were also right that the claim itself was wrong, not just the code. It lived in five places, all corrected in the same commit: the module docstring (twice), `purge_context_tool_artifacts`'s docstring, `_report_context_tool_purge`, and the user-facing `_RETIRED_CONTEXT_TOOL_MESSAGE` — a stamped run in another project now legitimately reports removals, so "nothing left after the first run" no longer holds. The stamp hangs off `paths.bin_dir().parent`, so the wording is now "once per workspace" rather than "per machine".

The branch also merges current `main`, which brings `h2 4.4.1` and clears the `pip-audit` finding that was failing on every open PR.

CI on `afa65b93` (dispatched on my fork, since fork-PR workflows sit in `action_required` here) — all green: CI including the unwrap CLI tests, Security, Wrap E2E, Init E2E, Validate Docs. `commitlint` is gated on `event_name == 'pull_request'` so a dispatch can't exercise it; I ran it locally over the branch range instead, clean.

JerrettDavis (approved):
Reviewed afa65b93. Both destructive-safety blockers are resolved on this head. The global stamp no longer suppresses cleanup in a later project or repointed agent home, and the provenance matcher now requires boundaries on both sides while continuing past rejected lookalikes to find a later genuine managed path. The new tests cover project and CODEX_HOME scope changes plus prefixed lookalikes across scripts, hooks, and MCP commands. I also corrected the PR description to match the split global versus invocation-scoped lifecycle. The PR is mergeable; CI has not run beyond governance, which is not a code-review blocker. Ready for final human review.

JerrettDavis:
Superseded by current-main consolidated install-safety PR #2980. The full provenance-guard commit series, authorship, documentation, and regression suite are preserved there, combined with the related install blocker to reduce stale PR count.

PR #2831, opened 2026-08-06
test(proxy): assert bounded concurrency structurally in the pre-upstream gate

Parideboy:
## Description

`tests/test_anthropic_pre_upstream_backpressure.py::test_n_plus_one_contention_only_waiter_has_nonzero_wait` is timing-flaky and fails CI shards on unrelated PRs (seen on #2607, run 31056617633, job `test (1)`):

```
FAILED tests/test_anthropic_pre_upstream_backpressure.py::test_n_plus_one_contention_only_waiter_has_nonzero_wait - AssertionError: [0.020465000005742695, 156.20862800000168, 353.85536299997966]
assert 156.20862800000168 < 25.0
```

The test drove 3 concurrent requests through the Unit 4 pre-upstream semaphore with `Semaphore(2)` and a 150 ms fake upstream, then asserted the two non-waiters each measured `pre_upstream_wait < 25 ms`. That is a wall-clock proxy for the property, not the property. The shard runs sequentially under `pytest --cov` on a shared runner; when the loop schedules the three coroutines serially, a request that acquired the semaphore with no queueing still records a 156 ms wait, and the assertion fails even though `headroom/proxy/handlers/anthropic.py` behaved correctly.

Per review feedback, the test no longer contains any millisecond threshold, absolute or relative. Contention is now produced deterministically with coordination primitives, so the assertions hold under arbitrary scheduling pauses on a loaded runner.

Closes # 

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `_DummyAnthropicHandler` gains an optional `upstream_release: asyncio.Semaphore`. When set, a request inside the upstream section blocks on a permit instead of sleeping for `upstream_delay_s`, so one `release()` frees exactly one holder. The existing sleep path is untouched for the other tests that use it
- The handler also exposes an `upstream_entered` condition, notified on each entry, so a test can await "N requests are inside the upstream section" without polling a duration
- `test_n_plus_one_contention_only_waiter_has_nonzero_wait` is replaced by `test_n_plus_one_contention_blocks_until_slot_frees`, which: pins two requests inside the upstream section, asserts both slots are held (`sem._value == 0`), starts the third, drains a bounded number of event-loop turns and asserts it has **not** entered, frees exactly one holder and asserts it enters only then
- Post-run assertions are structural: 3 enters / 3 exits, peak occupancy `<= 2` via the `_max_concurrent` event sweep (exits sort before enters at equal timestamps so a handed-over slot is not double-counted), the waiter's entry ordered after the first exit, and `sem._value == 2` for clean release
- The only timing-derived assertion left is `max(pre_upstream_wait) > 0` — no upper bound, no ratio. A comment records that wall-clock bounds must not come back in either form
- The test is parametrized `range(10)` as a stress invocation; it stays fast because every wait is event-driven rather than sleep-driven
- No production code touched; `test_happy_path_single_request_negligible_wait` and `test_concurrency_one_serializes_requests` are unchanged
- The scenario is wrapped in `anyio.fail_after(10)`, below the 15 s `anthropic_pre_upstream_acquire_timeout_seconds`, so a genuine deadlock fails the test rather than silently exercising the fail-open path

## Testing

- [ ] Type checking passes (`mypy headroom`) — N/A, tests-only change

### Test Output

```text
$ python -m pytest tests/test_anthropic_pre_upstream_backpressure.py -q
34 passed in 2.98s

$ python -m ruff check tests/test_anthropic_pre_upstream_backpressure.py
All checks passed!

$ python -m ruff format --check tests/test_anthropic_pre_upstream_backpressure.py
1 file already formatted
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.11, pytest 9.1.1, local checkout branched from upstream/main 7940c05e
- Exact command / steps: ran the full file (34 tests, including the 10 parametrized repeats of the rewritten test), then ran two mutation checks against the rewritten test — (1) widened the gate to `asyncio.Semaphore(3)` in the test, (2) removed `pre_upstream_sem.release()` from `_release_pre_upstream_sem` in `headroom/proxy/handlers/anthropic.py`. Both mutations were reverted afterwards. Separately verified the `_max_concurrent` sweep returns 3 for three overlapping intervals and 1 for a direct hand-over
- Observed result: 34 passed in 2.98s on the unmodified tree. With the gate widened to 3, all 10 repeats failed on the "third request has not entered" assertion. With the semaphore release removed, all 10 repeats failed. The helper sweep returned 3 and 1 as expected. So the test rejects unbounded concurrency, a leaked semaphore, and double-counted hand-overs, while containing no millisecond threshold at all
- Not tested: reproducing the original failure on a GitHub runner; the fix changes what is asserted, not proxy behaviour

## Review Readiness

## Checklist

## Additional Notes

Unblocks #2607, whose only red check is this flake. Screenshots N/A.

### DIFF
--- a/tests/test_anthropic_pre_upstream_backpressure.py
+++ b/tests/test_anthropic_pre_upstream_backpressure.py
@@ -431,14 +431,21 @@ async def _await_entered(handler: _DummyAnthropicHandler, count: int) -> None:
         await handler.upstream_entered.wait_for(lambda: len(handler.upstream_enter_times) >= count)
 
 
-async def _drain_scheduler(iterations: int = 100) -> None:
-    """Give every runnable task ample opportunity to progress.
+def _queued_waiters(sem: asyncio.Semaphore) -> int:
+    """Number of tasks parked inside ``sem.acquire()``."""
+    return sum(1 for waiter in (sem._waiters or ()) if not waiter.cancelled())
 
-    A bounded number of event-loop turns, not a duration: a request that is
-    *able* to enter the upstream section will have done so long before this
-    returns, no matter how slow or how loaded the machine is.
+
+async def _await_queued_waiters(sem: asyncio.Semaphore, count: int) -> None:
+    """Block until ``count`` tasks are queued on ``sem``.
+
+    This is the handshake that makes the "has not entered yet" assertion
+    meaningful: it establishes *where* the (N+1)th request is parked rather
+    than merely giving it chances to run. Yielding until the waiter appears
+    has no iteration budget to get wrong; the enclosing ``anyio.fail_after``
+    turns a genuine deadlock into a failure.
     """
-    for _ in range(iterations):
+    while _queued_waiters(sem) < count:
         await asyncio.sleep(0)
 
 
@@ -448,8 +455,10 @@ def test_n_plus_one_contention_blocks_until_slot_frees(_iteration, stage_log_cap
 
     Contention is driven by coordination primitives, never by sleeps or
     millisecond thresholds: the two admitted requests are pinned inside the
-    upstream section until the test hands out a permit. Every assertion is
-    therefore structural (occupancy, ordering, semaphore state) and stays
+    upstream section until the test hands out a permit, and the third is
+    observed queued on the gate before its non-entry is asserted. Every
+    assertion is therefore structural (occupancy, ordering, semaphore
+    queue depth and final value) and stays
     valid under arbitrary scheduling pauses on a loaded CI runner. Do not
     reintroduce wall-clock bounds here, absolute *or* relative to another
     request's measured wait -- both fail when the loop delays all three
@@ -482,8 +491,12 @@ def test_n_plus_one_contention_blocks_until_slot_frees(_iteration, stage_log_cap
                 assert sem._value == 0
 
                 # The third request cannot enter while both slots are held.
+                # Wait until it is provably parked *inside* the gate before
+                # asserting non-entry, so the assertion cannot pass merely
+                # because the task had not reached ``acquire()`` yet.
                 tg.start_soon(handler.handle_anthropic_messages, reqs[2])
-                await _drain_scheduler()
+                await _await_queued_waiters(sem, 1)
+                assert _queued_waiters(sem) == 1
                 assert len(handler.upstream_enter_times) == 2, handler.upstream_enter_times
                 assert len(handler.upstream_exit_times) == 0, handler.upstream_exit_times
 
--- a/tests/test_anthropic_pre_upstream_backpressure.py
+++ b/tests/test_anthropic_pre_upstream_backpressure.py
@@ -119,6 +119,7 @@ class _DummyAnthropicHandler(AnthropicHandlerMixin):
         *,
         anthropic_pre_upstream_sem: asyncio.Semaphore | None = None,
         upstream_delay_s: float = 0.0,
+        upstream_release: asyncio.Semaphore | None = None,
         raise_during_critical: bool = False,
         security: Any = None,
         upstream_status: int = 200,
@@ -206,6 +207,13 @@ class _DummyAnthropicHandler(AnthropicHandlerMixin):
         self._raise_during_critical = raise_during_critical
         self.upstream_enter_times: list[float] = []
         self.upstream_exit_times: list[float] = []
+        # Deterministic hold: when set, a request inside the upstream section
+        # blocks until the test hands it a permit, instead of sleeping for
+        # ``upstream_delay_s``. One ``release()`` frees exactly one holder, so
+        # a test can prove that the (N+1)th request enters only once a slot is
+        # actually returned -- no wall-clock thresholds involved.
+        self._upstream_release = upstream_release
+        self.upstream_entered = asyncio.Condition()
 
     async def _run_compression_in_executor(self, fn, *, timeout):  # noqa: ANN001
         # Mirror of ``HeadroomProxy._run_compression_in_executor`` for the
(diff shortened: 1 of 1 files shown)


JerrettDavis (changes requested):
The occupancy helper is a good direction, but this does not yet remove the scheduling flake class the PR targets.

The test still requires `waits[2] > 75` and `waits[0] < waits[2] / 2`. The latter can fail under the same loaded-runner scheduling jitter described in the PR: if all coroutines are delayed around acquisition (for example waits near 200/300/350 ms), bounded concurrency and clean release can both be correct while the ratio fails. It is still a wall-clock proxy, just a relative one.

Please make the contention deterministic with coordination primitives rather than sleeps/ratios: hold the first two admitted requests behind an event/barrier, prove both slots are occupied, start the third and prove it has not entered until one holder is released, then assert peak occupancy, entry order/count, and final semaphore value. That directly proves N+1 blocking without any millisecond threshold and will stay stable under arbitrary runner pauses. A repeated/stress invocation of the rewritten test would also be useful.

The event sweep itself correctly orders exit before enter at equal timestamps.

Parideboy:
Thanks — you're right, and the ratio check is gone. `waits[2] > 75.0` and `waits[0] < waits[2] / 2.0` were the same flake class as the absolute bounds; if the loop delays all three coroutines around acquisition, bounded concurrency and clean release can both be correct while the ratio fails.

The test is now driven by coordination primitives with no millisecond threshold anywhere (d026f5d):

- `_DummyAnthropicHandler` takes an optional `upstream_release: asyncio.Semaphore`. A request inside the upstream section blocks on a permit instead of sleeping, so one `release()` frees exactly one holder. An `upstream_entered` condition, notified on each entry, lets the test await "N requests are inside" without polling a duration. The old `upstream_delay_s` path is untouched for the other tests.
- Flow: pin two requests inside the section → assert both slots held (`sem._value == 0`) → start the third → drain a bounded number of event-loop turns (turns, not time) → assert it has **not** entered → free exactly one holder → assert it enters only then.
- Post-run: 3 enters / 3 exits, peak occupancy `<= 2` via the event sweep, the waiter's entry ordered after the first exit, `sem._value == 2`. The only timing-derived assertion left is `max(pre_upstream_wait) > 0` — no upper bound, no ratio.
- Parametrized `range(10)` as the stress invocation you asked for; it stays fast because every wait is event-driven.
- The scenario is wrapped in `anyio.fail_after(10)`, below the 15 s `anthropic_pre_upstream_acquire_timeout_seconds`, so a genuine deadlock fails the test instead of quietly exercising the fail-open path.

Mutation-checked both directions: widening the gate to `Semaphore(3)` fails all 10 repeats on the "third has not entered" assertion, and removing `pre_upstream_sem.release()` from `_release_pre_upstream_sem` fails all 10 too. Both reverted. Full file: 34 passed in 2.98s.

JerrettDavis (changes requested):
The permit-based rewrite removes the prior wall-clock flake, but the test still does not prove that request 3 is blocked at the gate before the first release. A fixed 100 calls to sleep(0) only gives it opportunities to run; if it has not reached the semaphore yet, len(upstream_enter_times) == 2 passes vacuously, and request 3 can first reach the gate after release and let the rest of the test pass. Please add an explicit handshake for the third acquire attempt or wait structurally until the semaphore has one queued waiter before asserting it has not entered. Since the test already inspects semaphore internals for _value, a small helper that waits for exactly one waiter is sufficient; an instrumented gate is even cleaner. Then the release-and-entry ordering proves the intended N+1 behavior without scheduler assumptions. The holder permits, final occupancy checks, and repeated run otherwise address the earlier timing-ratio concern well.

Parideboy:
Good catch — the drain was a scheduler assumption, not a proof. `_drain_scheduler(100)` only handed the third request opportunities to run; if it had not reached `pre_upstream_sem.acquire()` yet, `len(upstream_enter_times) == 2` passed vacuously and the request could have first arrived at the gate *after* the release while everything downstream still passed.

Replaced with an explicit handshake on the gate's queue (77f03c9):

- `_queued_waiters(sem)` counts non-cancelled futures in `sem._waiters`; `_await_queued_waiters(sem, 1)` yields until the gate reports that waiter. The test then asserts `_queued_waiters(sem) == 1` *before* asserting non-entry, so the request is proven parked inside `acquire()` rather than merely un-run.
- No iteration budget to get wrong: it is a wait-until, not a fixed number of turns. A genuine deadlock still surfaces through the enclosing `anyio.fail_after(10)`, which sits below the 15 s `anthropic_pre_upstream_acquire_timeout_seconds`.
- The semaphore's internals were already inspected here for `_value`, so this adds no coupling that was not present.
- `_drain_scheduler` is gone; it had no other call site.

Mutation-checked all three directions, each failing 10/10 repeats and then reverted:

- widen the gate to `asyncio.Semaphore(3)` — fails on "the third has not entered"
- move `release.release()` ahead of the waiter assertions — fails, confirming the assertion is no longer vacuous
- drop `pre_upstream_sem.release()` from `_release_pre_upstream_sem` in `headroom/proxy/handlers/anthropic.py` — fails

Clean run: `34 passed in 2.88s`, `ruff check` clean.

JerrettDavis (approved):
The new queued-waiter handshake resolves the outstanding concern: request 3 is now proven to be parked inside `Semaphore.acquire()` before the non-entry assertion, so the release/entry ordering cannot pass vacuously due to scheduling. The test remains free of timing thresholds, is bounded by `fail_after`, exercises the scenario repeatedly, and the latest head is green. Approved.

PR #2832, opened 2026-08-06
fix(cli/install): resolve the deployment profile instead of dead-ending on default

### LINKED ISSUE #2811: [BUG] `headroom init` deploys profile "init-user" but every `headroom install` lifecycle command defaults to `--profile default` — the documented commands all fail
### Summary

`headroom init` installs its persistent deployment under the profile name `init-user`, but every `headroom install <lifecycle>` subcommand hardcodes `--profile default`. The docs show those commands *without* `--profile`, so on a machine set up by `headroom init` every documented lifecycle command fails:

```console
$ headroom install status
Error: No deployment profile named 'default' is installed.
$ echo $?
1

$ headroom install status --profile init-user
Profile:    init-user
Preset:     persistent-task
Runtime:    python
Supervisor: none
Scope:      user
Port:       8787
Status:     running
Healthy:    yes
```

The error names neither the installed profile(s) nor the `--profile` flag, so there is nothing to lead the user to `init-user` — that name exists only as an internal constant. This cost me time twice: the command failed while the real deployment was running fine, and the message gave me nothing to act on, so I ended up reading the source to find the profile name.

### Environment

- headroom 0.32.0, `--mode token --backend anthropic` (same code paths verified against `main` today)
- preset `persistent-task`, scope `user`, supervisor `none`, port 8787
- Linux, Python 3.12, exactly one installed profile: `init-user`

### Where it comes from (`main`)

- `headroom/cli/init.py:53` — `_GLOBAL_PROFILE = "init-user"`; `_runtime_profile()` (`:123`) returns it for global-scope installs, i.e. the primary onboarding path names the profile, not the user.
- `headroom/cli/install.py` — `@click.option("--profile", default="default", show_default=True, ...)` on every lifecycle command: lines 433, 602, 720 (`status`), 748 (`start`), 763 (`stop`), 775 (`restart`), 789 (`remove`), 818, 830.
- `headroom/cli/install.py:68-75` — `_require_manifest()` raises `No deployment profile named '<p>' is installed.` with no list of what *is* installed.
- `headroom/install/state.py:102` — `list_manifests()` already enumerates installed manifests (globs `*/manifest.json`), but is never called from `cli/install.py`.
- `headroom/install/runtime.py:421` — `restart_current_deployment()` already resolves "the current deployment", so the concept exists internally; the CLI just doesn't use it.

Docs that show the bare form:

- `docs/content/docs/persistent-installs.mdx:31,40,67,69,70`
- `wiki/persistent-installs.md:25,52,54,55`
- `wiki/cli.md:677-706` (`headroom install status`, `start --profile default`, `stop`, `restart`)

### Proposal

When `--profile` is omitted, resolve it instead of assuming the literal `"default"`:

1. `HEADROOM_DEPLOYMENT_PROFILE` when set (the runtime already exports it),
2. else the single installed manifest when `list_manifests()` returns exactly one,
3. else fail, listing the installed profiles.

And make the error self-service:

```
Error: No deployment profile named 'default' is installed.
Installed profiles: init-user
Try: headroom install status --profile init-user
```

Either half removes the dead end; (2) also makes the documented copy-paste commands work right after `headroom init`.

abhay-codes07:
## Description

`headroom init` installs its persistent deployment under a non-`default` profile name (`init-user` for a global-scope install), but every `headroom install <lifecycle>` subcommand hardcodes `--profile default`. The docs show those commands without `--profile`, so on a machine set up by `headroom init` every documented lifecycle command fails while the real deployment is running fine:

```console
$ headroom install status
Error: No deployment profile named 'default' is installed.

$ headroom install status --profile init-user
Status:  running
Healthy: yes
```

The error named neither the installed profile nor the `--profile` flag, so there was nothing to lead the user to `init-user`, which exists only as an internal constant.

When the requested profile is not installed, `_require_manifest` now resolves the real target instead of dead-ending on a name the user never chose:

1. an explicit `HEADROOM_DEPLOYMENT_PROFILE` (which the runtime already exports) wins;
2. otherwise, when `--profile` was left at its `default` default and exactly one deployment is installed, that one is used;
3. when it still cannot decide, the error lists the installed profiles and points at `--profile`.

This changes only the not-found path. An installed `default` still loads exactly as before, and an explicit typo'd `--profile` still fails, now with a helpful list.

Fixes #2811

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cli/install.py` (`_require_manifest`): on a manifest miss, resolve via `HEADROOM_DEPLOYMENT_PROFILE`, then a single installed deployment when the request is the bare `default`, and otherwise raise an error that lists installed profiles and points at `--profile`. Imported `list_manifests` (already present in `headroom.install.state`) for the enumeration.
- `tests/test_cli/test_install_cli.py`: added `test_require_manifest_resolves_single_profile_when_default_missing`, `test_require_manifest_honors_env_profile`, and `test_require_manifest_lists_installed_profiles_when_ambiguous`.

## Testing

### Test Output

```text
# Before/after on the exact reported scenario (one installed profile "init-user"):
# ORIGINAL: _require_manifest("default")  -> raises "No deployment profile named 'default' is installed."
# FIXED:    _require_manifest("default")  -> resolves to "init-user"

# Pass-after, install suites:
tests/test_cli/test_install_cli.py         33 passed
tests/test_install/                        174 passed, 1 skipped, 1 pre-existing failure
# the 1 failure is tests/test_install/test_native_installers.py::
#   test_powershell_native_installer_supports_persistent_docker_lifecycle, which
#   runs scripts/install.ps1 and fails identically on clean main with these changes
#   stashed (an environment-specific PowerShell exit, unrelated to this diff).

# uvx mypy@1.20.2 headroom/cli/install.py -> Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: read `detect`/`_require_manifest` and the lifecycle command options (`--profile default` at install.py:720/748/763/775/789/818/830) to confirm the mismatch with `init.py`'s `_GLOBAL_PROFILE = "init-user"`, then demonstrated before/after by monkeypatching `load_manifest`/`list_manifests`: on the original code `_require_manifest("default")` raises "No deployment profile named 'default' is installed."; with the fix it returns the single installed manifest (`init-user`). Fail-before via `git stash push headroom/cli/install.py` and a direct call; pass-after with `git stash pop` and the install suites (33 passed in the CLI file, 174 passed in test_install with one pre-existing environment failure).
- Observed result: a bare lifecycle command on an init'd machine now targets the running deployment instead of failing, matching the `--profile init-user` command the issue reporter confirmed works. An explicit `HEADROOM_DEPLOYMENT_PROFILE` selects the target, and an ambiguous multi-profile machine gets an error naming the installed profiles and the `--profile` flag.
- Not tested: a full end-to-end `headroom init` then `headroom install status` on a fresh host (that flow spawns a real deployment and supervisor). The resolution logic is a pure function verified directly, and the manifest loading it calls is existing, tested code.

## Review Readiness

## Checklist

## Additional Notes

The resolution deliberately only triggers on the not-found path and only auto-selects when a single deployment is installed or an explicit env profile names one, so it never silently picks the wrong deployment on a multi-profile host. The docs that show the bare commands (`docs/content/docs/persistent-installs.mdx`, `wiki/persistent-installs.md`, `wiki/cli.md`) become correct again without needing a `--profile` on every line.

### DIFF
--- a/headroom/cli/install.py
+++ b/headroom/cli/install.py
@@ -36,6 +36,7 @@ from headroom.install.runtime import (
 from headroom.install.state import (
     ManifestError,
     delete_manifest,
+    list_manifests,
     load_manifest,
     save_manifest,
 )
@@ -70,9 +71,35 @@ def _require_manifest(profile: str) -> DeploymentManifest:
         manifest = load_manifest(profile)
     except ManifestError as e:
         raise click.ClickException(str(e)) from None
-    if manifest is None:
-        raise click.ClickException(f"No deployment profile named '{profile}' is installed.")
-    return manifest
+    if manifest is not None:
+        return manifest
+
+    # The requested profile isn't installed. `headroom init` installs under a
+    # non-"default" profile name (e.g. init-user), while every lifecycle command
+    # defaults --profile to "default" -- so on an init'd machine the documented
+    # bare commands (`headroom install status`, etc.) would all dead-end (#2811).
+    # Resolve the real target instead of failing with a name the user never
+    # chose: an explicit HEADROOM_DEPLOYMENT_PROFILE (which the runtime exports)
+    # wins; otherwise, when --profile was left at its "default" default and
+    # exactly one deployment is installed, use that one.
+    installed = list_manifests()
+    env_profile = os.environ.get("HEADROOM_DEPLOYMENT_PROFILE", "").strip()
+    if env_profile and env_profile != profile:
+        try:
+            resolved = load_manifest(env_profile)
+        except ManifestError:
+            resolved = None
+        if resolved is not None:
+            return resolved
+    if profile == "default" and len(installed) == 1:
+        return installed[0]
+
+    if installed:
+        names = ", ".join(sorted(m.profile for m in installed))
+        hint = f" Installed: {names}. Select one with --profile <name>."
+    else:
+        hint = " No deployments are installed; run `headroom init` or `headroom install apply`."
+    raise click.ClickException(f"No deployment profile named '{profile}' is installed.{hint}")
 
 
 def _start_deployment(manifest: DeploymentManifest, *, assume_start_lock: bool = False) -> None:
--- a/headroom/cli/install.py
+++ b/headroom/cli/install.py
@@ -11,6 +11,7 @@ from copy import deepcopy
 from dataclasses import dataclass
 
 import click
+from click.core import ParameterSource
 
 from headroom._subprocess import run
 from headroom.install.health import probe_json, probe_ready
@@ -66,6 +67,22 @@ def install() -> None:
     """Install and manage persistent Headroom deployments."""
 
 
+def _profile_selection_was_explicit() -> bool:
+    """True when the current command received an explicit ``--profile``.
+
+    An explicit selection must be honored verbatim or rejected, never redirected
+    to ``HEADROOM_DEPLOYMENT_PROFILE`` or a lone installed deployment: silently
+    operating ``stop``/``restart``/``remove`` on a different profile than the one
+    the user typed is dangerous. Only a defaulted (omitted) ``--profile`` is
+    eligible for the recovery fallback. Outside a Click command context (direct
+    calls / unit tests) there is no explicit selection to protect.
+    """
+    ctx = click.get_current_context(silent=True)
+    if ctx is None:
+        return False
+    return bool(ctx.get_parameter_source("profile") == ParameterSource.COMMANDLINE)
+
+
 def _require_manifest(profile: str) -> DeploymentManifest:
     try:
         manifest = load_manifest(profile)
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The default-profile recovery is useful, but the latest implementation violates the stated explicit-selection boundary.

`_require_manifest(profile)` has no information about whether `profile` came from the Click default or an explicit `--profile`. Consequently, whenever the requested manifest is absent, lines 88-95 let `HEADROOM_DEPLOYMENT_PROFILE` override any requested name. For example, with `HEADROOM_DEPLOYMENT_PROFILE=init-user`, `headroom install status --profile typo` operates on `init-user` instead of rejecting the explicit typo. That is especially risky for stop/restart/remove lifecycle commands.

Please preserve explicit CLI selection: only apply env/single-installed fallback when Click reports that `--profile` came from the default (for example via parameter-source handling or an explicit resolver flag), and add a `CliRunner` regression test proving an explicit missing profile fails even when the env profile exists. The current direct helper tests cannot exercise that distinction because they bypass Click option parsing.

The installed-profile enumeration and improved ambiguous error otherwise look good.

abhay-codes07:
Good catch, thanks. You are right that the explicit-vs-default distinction was the whole safety boundary and the helper had no way to see it.

Fixed in b4cbc7fe: the fallback is now gated on Click parameter-source. `_require_manifest` only resolves an alternative deployment when `--profile` was left at its default; an explicit `--profile` is honored or rejected verbatim, so `headroom install status --profile typo` with `HEADROOM_DEPLOYMENT_PROFILE=init-user` now fails and names the typo rather than acting on `init-user`. Outside a Click command context (the direct helper tests) there is no explicit selection to protect, so those still exercise the fallback path.

Added the two CliRunner tests you asked for, which go through real option parsing:
- `test_install_status_explicit_missing_profile_is_not_redirected_to_env`: `--profile typo` with the env profile set exits non-zero and never renders init-user status (fails without the gate).
- `test_install_status_omitted_profile_resolves_env_deployment`: the bare command still resolves the env deployment.

35 tests pass, ruff and mypy clean.

JerrettDavis (changes requested):
The command-line explicitness fix is good, but the environment-profile path still has a dangerous fallthrough. If `HEADROOM_DEPLOYMENT_PROFILE` is set to a missing/stale profile and exactly one different manifest exists, `load_manifest(env_profile)` returns `None` (or raises), then the code falls through to `len(installed) == 1` and silently operates on that different deployment. For lifecycle commands such as `stop`, `restart`, and `remove`, an explicit environment selection should be honored or rejected just like an explicit `--profile`, not redirected.

Please make a non-empty `HEADROOM_DEPLOYMENT_PROFILE` authoritative: return it when installed, otherwise raise an actionable error naming the missing env-selected profile and installed choices. Only use the lone-manifest fallback when neither CLI nor environment supplied a selection. Add a `CliRunner` regression test with env=`missing` plus one installed `init-user` proving the command fails without invoking the lifecycle operation.

abhay-codes07:
Good catch again, thanks. You are right that a stale env selection was still falling through to the lone-manifest path.

Fixed in 77619192: a non-empty `HEADROOM_DEPLOYMENT_PROFILE` is now authoritative, exactly like an explicit `--profile`. It is honored when installed, otherwise the command fails with an actionable error naming the env-selected profile (and that it came from `HEADROOM_DEPLOYMENT_PROFILE`) plus the installed choices. The lone-manifest fallback only applies when neither the CLI nor the environment supplied a selection.

Added the regression test you asked for, `test_install_status_stale_env_profile_is_not_redirected_to_lone_manifest`: env=`missing` with one installed `init-user` exits non-zero and never renders init-user status. It fails on the previous revision (exit 0, redirected). 36 tests pass, ruff and mypy clean.

JerrettDavis (changes requested):
The stale `HEADROOM_DEPLOYMENT_PROFILE` fallthrough is now correctly rejected and the new lifecycle regression test addresses the production-safety concern. The head is not merge-ready yet because the required `commitlint` check is still failing; please repair the commit range so that gate passes.

abhay-codes07:
Thanks. Repaired the commit range: the second follow-up header was 102 chars, over commitlint header-max-length of 100, so I reworded it to `fix(cli/install): honor HEADROOM_DEPLOYMENT_PROFILE, don't redirect a stale env selection` (89 chars). All three headers are now within the limit and body lines are well under the 200 cap. No code change.

JerrettDavis (approved):
The rewritten head preserves the authoritative environment-profile fix and its regression coverage, and commitlint now passes. Content review is resolved; remaining CI jobs are still in progress.

PR #2835, opened 2026-08-06
perf(kompress): coalesce concurrent remote calls into one batch request

chopratejas:
## Description

A remote Kompress endpoint was receiving a burst of single-content POSTs per turn, because the pipeline already compresses blocks concurrently. That is the worst possible request shape for it: the model has **one** execution slot, and a caller that misses it is *shed*, not queued — it gets its content back uncompressed, logged only as `Kompress execution saturated after %.2fms; skipping chunk`. Nothing about that looks like a failure from the outside. It looks like fast, healthy service that happens to compress less.

Measured, 16-way concurrency cost ~61% of the compression.

This folds calls that overlap within a 5ms window into a single `/compress_batch` request, so the endpoint sees one request instead of N and nothing gets shed.

Closes #

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `RemoteKompressCompressor` coalesces overlapping `compress()` calls into one `/compress_batch` request. The first caller for a target ratio becomes the leader: it waits out the window, claims the queue, and fires one request for everyone; the rest park on an `Event`.
- Calls wanting **different** target ratios never ride together — the batch route applies one ratio to the whole request.
- Every failure path falls back to the single call, so a missing or broken batch route costs latency, not compression: batch HTTP error, a short result list, a malformed payload, or a follower timing out.
- Extracted `_result_from_payload` and `_store_ccr` so the batch and single paths cannot drift in how they coerce metadata or attach the CCR marker.
- `_compress_mixed` gains matching section-level overlap via `ThreadPoolExecutor.map` (order-preserving). It exists **purely** so sections arrive together for the fold — it stays serial unless `ready_backend() == "remote"`, because local compression has the same single slot and overlap there only buys contention.
- Two escape hatches: `HEADROOM_KOMPRESS_ENDPOINT_BATCH_PATH=""` disables coalescing for an endpoint with no batch route; `HEADROOM_MIXED_SECTION_WORKERS` tunes the fan-out.

No behaviour change when `HEADROOM_KOMPRESS_ENDPOINT` is unset, which is the default.

## Testing

### Test Output

```text
$ python -m pytest tests/test_kompress_coalesce.py tests/test_transforms_content_router.py \
    tests/test_kompress_remote_endpoint.py tests/test_content_router_single_item_deadline.py \
    tests/test_content_router_user_blocks.py tests/test_kompress_failsafe.py \
    tests/test_router_registry_dispatch.py -q
138 passed, 1 warning in 9.71s

$ ruff check headroom/transforms/kompress_remote.py headroom/transforms/content_router.py tests/test_kompress_coalesce.py
All checks passed!

$ mypy --python-version 3.12 headroom/transforms/kompress_remote.py headroom/transforms/content_router.py
```

`--python-version 3.12` is needed only because the installed numpy stubs use a `type` statement that the configured target rejects; it is unrelated to this change and reproduces on a clean tree.

The new `tests/test_kompress_coalesce.py` uses a fake client that counts requests per route, so "did N calls become one request" is directly observable: 8 concurrent calls produce exactly one fold of 8, mixed ratios split into two folds, `max_batch` caps the fold, a broken batch route still compresses via the single route, and an empty `batch_path` never touches it.

## Real Behavior Proof

- **Environment:** macOS, Python 3.12.6, live Kompress endpoint on Modal (kompress-v2-base, ONNX-INT8 CPU), `HEADROOM_KOMPRESS_ENDPOINT` set.
- **Exact command / steps:** 35 README sections at `target_ratio=0.6`, run twice through `RemoteKompressCompressor` — once serial with `batch_path=""`, once with 32 workers and coalescing on. **Both arms cache-cold**: the endpoint memoizes by `(model, ratio, content)`, and without a per-arm nonce it silently replays the other arm's answers and hides any regression. Then the same workload end to end through `anthropic_pipeline.apply` with `httpx.Client.post` monkeypatched to count and time every call.
- **Observed result:**

  | arm | wall | tokens |
  |---|---|---|
  | serial, no batch route | 13.33s | 2,691 |
  | 32 workers, coalesced | 7.17s | 2,665 |

  1.9x faster, and **0 of 35 sections differ by more than 2 tokens** — the fold does not change what any caller gets. On a real turn, Modal requests drop from 35 to 3 (folds of 8, 24 and 3 blocks).

  At the full-pipeline level both arms are noisy (the baseline alone spans 9,191 / 9,942 / 9,472 tokens across three cold runs), but coalesced medians land lower: **8,447 vs 9,472 tokens**. The baseline's faster walls are it shedding work rather than doing it.

- **Not tested:** GPU endpoints — on CPU/ONNX the server's `compress_batch` falls back to sequential, so this PR buys the round-trips, not a batched forward pass. Endpoints that expose a batch route with different semantics than `{"contents": [...], "target_ratio": r} -> {"results": [...]}`; those should set `HEADROOM_KOMPRESS_ENDPOINT_BATCH_PATH=""`. No load test with multiple concurrent proxies against one endpoint.

## Review Readiness

## Checklist

## Additional Notes

**Docs:** N/A for user-facing docs — the two new env vars are escape hatches for a non-default deployment path, and the reasoning is documented at length in the `kompress_remote` module docstring, which is where someone debugging this will actually look.

**Why the window is 5ms:** long enough that sections dispatched by one `ThreadPoolExecutor.map` land in the same window, short enough to be invisible against a ~470ms round-trip. A single un-overlapped call pays it once.

**Trap for anyone re-measuring this:** the two failure modes here are both silent. The endpoint's result cache will replay a previous arm's answers unless every input is nonced, and the client fails open, so a regression shows up as *less compression*, never as an error. Both cost me a wrong conclusion before I nonced the inputs.

**Follow-up, not in this PR:** the server side needs the matching fix — `/compress_batch` was itself gathering its items (so N-1 of N were shed) and `@modal.concurrent(max_inputs=16)` packed a burst into one container. Both are fixed and deployed on the endpoint. With that in place, residual shedding on concurrent singles is 4/24 instead of a majority, and 0/24 through the batch route.

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -1120,6 +1120,11 @@ def _create_content_signature(
 # session-tracker cleanup TTL (``PrefixFreezeConfig.session_ttl_seconds``).
 _NET_COST_CACHE_TTL_SECONDS = 300.0
 
+# Sections compressed at once when Kompress runs remotely. Sized to be comfortably
+# above a typical section count so one turn folds into one request; the client's
+# max_batch is the real ceiling.
+_REMOTE_SECTION_WORKERS = 32
+
 
 def _net_cost_cache_ttl_seconds() -> float:
     """Provider cache TTL (seconds) used to decay P_alive from idle time.
@@ -2388,6 +2393,25 @@ class ContentRouter(Transform):
 
         return strategy
 
+    def _mixed_section_workers(self, section_count: int) -> int:
+        """How many sections to compress at once.
+
+        One, unless Kompress is running remotely. Local compression has a single
+        execution slot, so overlap there buys nothing and costs contention. A remote
+        endpoint is ~470ms of round-trip per call regardless of size, and its client
+        coalesces concurrent calls into one request — so sections have to *arrive*
+        together for that fold to happen. This is the only reason the overlap exists.
+        """
+        try:
+            kompress = self._get_kompress()
+            if kompress is None or kompress.ready_backend() != "remote":
+                return 1
+            override = os.environ.get("HEADROOM_MIXED_SECTION_WORKERS")
+            limit = int(override) if override else _REMOTE_SECTION_WORKERS
+            return max(1, min(limit, section_count))
+        except Exception:
+            return 1
+
     def _compress_mixed(
         self,
         content: str,
@@ -2422,10 +2446,8 @@ class ContentRouter(Transform):
                 strategy_used=CompressionStrategy.PASSTHROUGH,
             )
 
-        compressed_sections: list[str] = []
-        routing_log: list[RoutingDecision] = []
-
-        for i, section in enumerate(sections):
+        def _compress_section(item: tuple[int, ContentSection]) -> tuple[str, RoutingDecision]:
+            i, section = item
             # Get strategy for this section
             strategy = self._strategy_from_detection_type(section.content_type)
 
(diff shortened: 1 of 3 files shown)


JerrettDavis (changes requested):
The batching direction and cache-cold measurement are strong, but I found two concurrency blockers on the latest head:

1. `content_router.py:2456-2483` parallelizes the entire `_apply_strategy_to_content` path whenever a remote Kompress instance exists. Mixed sections can select code/search/log/diff/config/external strategies, and this path also touches shared lazy compressor fields, the registry, CCR/TOIN state, and fallback chains. Those components were previously serialized and are not made thread-safe by this PR. Please restrict fan-out to the remote-Kompress operation (or establish/test thread safety for every reachable strategy). A mixed-content regression test should prove non-Kompress strategies remain serialized while Kompress-bound sections still coalesce.

2. `kompress_remote.py:344-362` treats `Event.wait()` timeout as a failed batch and lets the caller proceed to `/compress`, even though the leader may still be executing `/compress_batch`. Scheduling/cleanup around the httpx timeout is enough for that race, and it breaks the stated invariant that concurrent requests must never reach the model. The same issue occurs in the "another thread claimed this queue" branch. Please make batch ownership/completion explicit so followers cannot launch singles while a batch is still live, and add a delayed fake-client test that crosses the follower wait boundary and asserts no overlapping batch/single requests.

Once those boundaries are tightened, the rest of the extraction and per-ratio grouping looks well structured.

Issue #2804, opened 2026-08-05
[BUG] persistent-docker sets --user $(id -u):$(id -g), which makes every bind-mount write fail under rootless Podman


rorar:
## Summary

`build_runtime_command` unconditionally adds `--user <uid>:<gid>` on non-Windows hosts. That is correct for Docker, where container UIDs equal host UIDs, but wrong for rootless Podman, where the host user is already mapped to container UID 0. Passing the host UID selects a subordinate UID with no access to the bind-mounted host directories, so every write into `~/.headroom` fails. The proxy still starts and reports healthy — the failure only surfaces once a request touches a write path.

## Environment

| | |
|---|---|
| headroom CLI | 0.34.0 |
| Image | `ghcr.io/headroomlabs-ai/headroom:latest` (`b68ae77c1863`) |
| Container runtime | Podman 5.8.2, **rootless**, invoked through a `docker` shim (`/run/current-system/sw/bin/docker → podman`) |
| Host | NixOS 26.05 (Yarara), Linux 6.18.39 x86_64 |
| ID mapping | Default rootless userns; host user is a regular non-root account with a standard 65536-wide `/etc/subuid` range |
| Verified against | `main` @ `64e20393`; zero occurrences of `podman` anywhere in the package |

## Root cause

`headroom/install/runtime.py:138-142`:

```python
if not _is_windows():
    getuid = getattr(os, "getuid", None)
    getgid = getattr(os, "getgid", None)
    if callable(getuid) and callable(getgid):
        command.extend(["--user", f"{getuid()}:{getgid()}"])
```

Under rootless Podman's default user namespace the host user is mapped to container UID 0, and the `/etc/subuid` range is mapped to container UIDs 1 and above. Passing `--user $(id -u):$(id -g)` therefore selects a container UID that is backed by a subordinate host UID owning nothing. The bind-mounted `~/.headroom` appears inside the container as `root:root` and is unwritable.

Docker maps container UIDs to host UIDs one-to-one, which is why the same flag is correct there and wrong here.

## Evidence

Account IDs redacted below; the mapping, not the specific values, is the point.

```console
$ podman run --rm -v "$HOME/.headroom:/tmp/headroom-home/.headroom" \
    --user "$(id -u):$(id -g)" <image> \
    sh -c 'id; ls -ldn /tmp/headroom-home/.headroom; touch /tmp/headroom-home/.headroom/x'
uid=<host uid> gid=<host gid>
[...]

PR #2845, opened 2026-08-07
fix(install): stop baking the host memory DB path into a container deployment

### LINKED ISSUE #2803: [BUG] persistent-docker + --memory: planner passes a host path as --memory-db-path, so the container can never become ready
## Summary

`headroom deploy --memory` on the `persistent-docker` preset can never succeed. The planner resolves the memory DB path on the **host** and appends it verbatim to `proxy_args`; the runtime translates only the leading `--host` pair, so the container receives a path that does not exist in its filesystem. Memory init fails, `/readyz` stays 503, and `_start_deployment`'s 45 s `wait_ready` expires. The CLI then rolls back, so the failure presents as a timeout rather than as a path bug.

## Environment

| | |
|---|---|
| headroom CLI | 0.34.0 |
| Image | `ghcr.io/headroomlabs-ai/headroom:latest` (`b68ae77c1863`), reports 0.34.0 |
| Preset | `persistent-docker`, port 8787, mode `cache`, backend `anthropic` |
| Host | NixOS 26.05 (Yarara), Linux 6.18.39 x86_64 |
| Verified against | `main` @ `64e20393` — `install/runtime.py` and `install/planner.py` byte-identical to the installed 0.34.0 |

## Reproduction

```bash
headroom deploy --memory
```

## Actual

```
Error: Deployment 'default' did not become ready after start.
Restoring previous deployment 'default'...
```

Container log:

```
Memory: backend initialization failed (startup continues): unable to open database file
```

`/health` reports `memory: {"enabled": true, "ready": false, "initialized": false}`, which keeps `/readyz` at 503 for the full 45 s window.

## Root cause

`headroom/install/planner.py:184` resolves the path against the **host** home:

```python
if memory_enabled:
    proxy_args.extend(["--memory", "--memory-db-path", str(_paths.memory_db_path())])
```

yielding `--memory-db-path /home/<user>/.headroom/memory.db`.

`headroom/install/runtime.py:159-167` rewrites only the first host/port pair and passes everything after it through unchanged:

```python
command.extend([
    manifest.image,
    "--host",
    CONTAINER_BIND_HOST,
    *manifest.proxy_args[_PROXY_ARGS_HOST_PAIR_LEN:],
])
```

The container's `HOME` is `/tmp/headroom-home` (runtime.py:104) and `$HOME/.headroom` is mounted at `/tmp/headroom-home/.headroom` (runtime.py:124-125). `/home/<user>/.headroom/` does not exist inside the container, so SQLite cannot create the file. The same applies on macOS with `/Users/<user>/…`.

## Suggested fix

Do not emit `--memory-db-path` when `runtime_kind == docker`. `MemoryConfig.memory_db_path` already defaults to `""` (`headroom/proxy/models.py:369` — *"Empty = auto: {cwd}/.headroom/memory.db"*), and the container `--workdir` is exactly `/tmp/headroom-home` (runtime.py:114), so the auto-default lands inside the mounted volume with no translation needed.

Alternatively, translate the path in `build_runtime_command` the way `--host` already is.

## Verified workaround

Isolated container on port 8788, same image, `--memory` passed without `--memory-db-path`:

```
Memory: Project-scoped DB
/health → memory: enabled=true ready=true initialized=true
```

The auto-default additionally yields a project-scoped DB instead of legacy global mode, which appears to be the intended behaviour.

## Related

Same bug class as #2432 (`--port` honored by the proxy, ignored by the healthcheck and `doctor`): a host-resolved value crossing into container context without translation.

---

**How this was written**

Drafted with Claude Opus 5 in Claude Code. It is an AI-assisted report, so treat the reasoning as something to check rather than to take on trust.

abhay-codes07:
## Description

`headroom deploy --memory` on the `persistent-docker` preset can never become ready. The planner resolves the memory DB path against the **host** home and appends it verbatim to `proxy_args`:

```python
# headroom/install/planner.py
proxy_args.extend(["--memory", "--memory-db-path", str(_paths.memory_db_path())])
# -> --memory-db-path /home/<user>/.headroom/memory.db
```

The docker runtime passes everything after the leading `--host` pair through unchanged, and the container's `HOME` is `/tmp/headroom-home` with the host's `~/.headroom` bind-mounted at `/tmp/headroom-home/.headroom`. The host path `/home/<user>/.headroom/memory.db` does not exist inside the container, so SQLite cannot open the DB:

```text
Memory: backend initialization failed (startup continues): unable to open database file
```

`/health` then reports `memory.ready = false`, `/readyz` stays 503 for the full `wait_ready` window, and `_start_deployment` times out and rolls back, so the failure presents as "did not become ready" rather than a path bug. The same applies on macOS with `/Users/<user>/...`.

The fix omits `--memory-db-path` for a container (docker) runtime. When the flag is absent the proxy resolves the DB under its own cwd (`.headroom/memory.db`), and the container's workdir is `/tmp/headroom-home` (the bind mount), so the DB lands in exactly the same host file the explicit path intended. The host (python) runtime still passes the resolved host path, which is correct there.

Fixes #2803

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/install/planner.py` (`build_manifest`): append `--memory` always, but add `--memory-db-path <host path>` only when `runtime_kind != RuntimeKind.DOCKER.value`. Imported `RuntimeKind` from `.models`.
- `tests/test_install/test_planner.py`: extended `test_build_manifest_for_persistent_docker_sets_expected_defaults` to assert `--memory-db-path` is absent for the docker runtime, and added `test_build_manifest_python_runtime_keeps_explicit_memory_db_path` asserting it is still present for the python runtime.

## Testing

### Test Output

```text
# Fail-before (source fix stashed, updated tests kept):
tests/test_install/test_planner.py::test_build_manifest_for_persistent_docker_sets_expected_defaults FAILED
  assert "--memory-db-path" not in manifest.proxy_args
  AssertionError: assert '--memory-db-path' not in ['--host', '127.0.0.1', ...]

# Pass-after (fix applied):
tests/test_install/test_planner.py  19 passed

# Broader install suites:
tests/test_install/  141 passed, 1 skipped, 2 unrelated pre-existing/flaky failures
#   - test_native_installers.py::test_powershell_native_installer_supports_persistent_docker_lifecycle
#     runs scripts/install.ps1 and fails identically on clean main (environment-specific).
#   - test_runtime.py::test_runtime_status_survives_winerror87_systemerror passes in isolation
#     and in its own file; it only failed under cross-file ordering in the broad run, and is
#     untouched by this diff (planner.py only).

# uvx mypy@1.20.2 headroom/install/planner.py -> Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: traced the path from `planner.py` (`--memory-db-path str(_paths.memory_db_path())`, host home) through `runtime.py` (`build_runtime_command` passes `proxy_args[_PROXY_ARGS_HOST_PAIR_LEN:]` through, container HOME `/tmp/headroom-home`, `~/.headroom` bind-mounted) and confirmed via `server.py` that an empty `memory_db_path` resolves to `Path.cwd()/.headroom/memory.db` (the container workdir, hence the mount). Fail-before with `git stash push headroom/install/planner.py` and `python -m pytest tests/test_install/test_planner.py -k persistent_docker` (host path present in proxy_args), pass-after with `git stash pop` and rerunning (19 passed).
- Observed result: for the docker runtime, `manifest.proxy_args` now carries `--memory` without `--memory-db-path`, so the container resolves the DB to `/tmp/headroom-home/.headroom/memory.db` (the bind mount to host `~/.headroom/memory.db`) and can open it, instead of receiving a nonexistent host path. The python runtime still carries the explicit host path.
- Not tested: a live `headroom deploy --memory` against a running Docker daemon (no container runtime in this environment). The manifest construction is verified directly, and the container-side resolution it relies on is existing server behavior (`empty memory_db_path -> cwd/.headroom/memory.db`) confirmed by reading `server.py`.

## Review Readiness

## Checklist

## Additional Notes

The DB persistence location is unchanged: both the old host path and the new container-cwd resolution point at the host's `~/.headroom/memory.db` (directly on the host, or through the bind mount inside the container), so existing memory DBs are picked up either way. This is the memory-path half of the persistent-docker issues; the separate rootless-Podman `--user` bind-mount problem (#2804) is left for its own fix.

### DIFF
--- a/headroom/install/planner.py
+++ b/headroom/install/planner.py
@@ -15,6 +15,7 @@ from .models import (
     DeploymentManifest,
     InstallPreset,
     ProviderSelectionMode,
+    RuntimeKind,
     SupervisorKind,
     ToolTarget,
 )
@@ -181,7 +182,19 @@ def build_manifest(
     ]
     proxy_args.append("--telemetry" if telemetry_enabled else "--no-telemetry")
     if memory_enabled:
-        proxy_args.extend(["--memory", "--memory-db-path", str(_paths.memory_db_path())])
+        proxy_args.append("--memory")
+        # `_paths.memory_db_path()` resolves against the HOST home. A container
+        # runtime cannot use it: the container's HOME is /tmp/headroom-home and
+        # the host's ~/.headroom is bind-mounted there, so a host path like
+        # /home/<user>/.headroom/memory.db does not exist inside the container,
+        # SQLite fails to open the DB, /readyz stays 503, and the deployment
+        # times out and rolls back (#2803). Omit the flag for a container runtime:
+        # the proxy then resolves the DB under its own cwd (.headroom/memory.db),
+        # which is the container's workdir and therefore the bind mount, landing
+        # in the same host file the explicit path intended. On the host (python)
+        # runtime the resolved host path is correct, so keep passing it.
+        if runtime_kind != RuntimeKind.DOCKER.value:
+            proxy_args.extend(["--memory-db-path", str(_paths.memory_db_path())])
     if anyllm_provider:
         proxy_args.extend(["--anyllm-provider", anyllm_provider])
     if region:
--- a/tests/test_install/test_planner.py
+++ b/tests/test_install/test_planner.py
@@ -46,6 +46,33 @@ def test_build_manifest_for_persistent_docker_sets_expected_defaults() -> None:
     assert manifest.tool_envs["claude"]["ANTHROPIC_BASE_URL"] == "http://127.0.0.1:8787"
     assert manifest.tool_envs["copilot"]["COPILOT_PROVIDER_TYPE"] == "anthropic"
     assert "--memory" in manifest.proxy_args
+    # A container runtime must NOT carry the host memory DB path: it does not
+    # exist inside the container and would keep /readyz at 503 (#2803). The proxy
+    # resolves the DB under its own cwd, which is the bind-mounted ~/.headroom.
+    assert "--memory-db-path" not in manifest.proxy_args
+
+
+def test_build_manifest_python_runtime_keeps_explicit_memory_db_path() -> None:
+    manifest = build_manifest(
+        profile="default",
+        preset=InstallPreset.PERSISTENT_SERVICE.value,
+        runtime_kind="python",
+        scope="user",
+        provider_mode="manual",
+        targets=["claude"],
+        port=8787,
+        backend="anthropic",
+        anyllm_provider=None,
+        region=None,
+        proxy_mode="token",
+        memory_enabled=True,
+        telemetry_enabled=False,
+        image="ghcr.io/headroomlabs-ai/headroom:latest",
+    )
+
+    # On the host the resolved path is correct, so it is still passed explicitly.
+    assert "--memory" in manifest.proxy_args
+    assert "--memory-db-path" in manifest.proxy_args
 
 
 def test_build_manifest_uses_provider_slice_env_builders_for_all_supported_targets() -> None:


JerrettDavis (approved):
Reviewed 8b9557bd end to end through manifest construction, the container workdir contract, the existing home bind mount, and the proxy empty-path fallback. Omitting the host-only path for Docker resolves to the same persisted file while preserving the explicit path for native Python installs. Focused coverage is appropriate and CI is green. Ready for final human review.

Issue #2810, opened 2026-08-05
Consistency re-count runs two count_messages passes on the event loop (v0.33.0)


itsumonotakumi:
## Summary

In v0.33.0 the proxy's "consistency re-count" runs `tokenizer.count_messages()` twice **directly on the event loop**, unconditionally, for every request. Combined with #2543 (which moved Claude token counting from a character estimate to a real BPE), this blocks the event loop for a measurable amount of time on large requests — up to roughly **1 second on a 2.3 MB request body** in our measurements.

For a single-user proxy this is mostly invisible. For a shared deployment it is not: while one request is being re-counted, every other in-flight request on the same process is stalled.

We noticed this because the same release also contains #2498, which deliberately moved the *other* tokenizer counting off the event loop. The new re-count path appears to have simply been missed rather than intentionally left synchronous — hence this report.

## Where

`headroom/proxy/handlers/anthropic.py`, inside `handle_anthropic_messages`:

```python
# Consistency: report tok_before/tok_after with ONE tokenizer. ...
# ... Runs unconditionally.
try:
    _orig_snapshot = original_client_messages
    original_tokens = tokenizer.count_messages(_orig_snapshot)
    optimized_tokens = tokenizer.count_messages(optimized_messages)
```

`headroom/proxy/handlers/openai.py`, inside `handle_openai_chat`, has the same shape:

```python
original_tokens = tokenizer.count_messages(original_client_messages)
optimized_tokens = tokenizer.count_messages(body["messages"])
```

Both are `async def`, and both call the tokenizer synchronously. The class already has `_count_tokens_offloaded()` (used elsewhere in the same files), so the offload mechanism exists — these two sites just don't use it.

### Method

Two images built from the same Dockerfile, differing only in the pinned version (`headroom-ai[proxy]==0.32.0` vs `==0.33.0`). Run sequentially on the same host, `--cpus=2 --memory=1536m`, with:

```
headroom proxy --host 127.0.0.1 --port 8787 \
  --no-telemetry --no-ccr --stateless --no-memory-context --no-memory-tools
HEADROOM_MODE=cache
HEADROOM_SAVINGS_PROFILE=coding
HEADROOM_COMPRESSION_TIMEOUT_SECONDS=60
```

[...]

Issue #2768, opened 2026-08-04
MemoryError in the original-body verification re-parse aborts the request on 1M-context payloads


crb8:
### Summary

A best-effort verification re-parse of the original request body can raise `MemoryError`, which is not caught, escapes, and aborts the inbound request:

```
ERROR - event=proxy_inbound_request_aborted id=inbound-... method=POST
        path=/v1/messages reason=MemoryError duration_ms=10605.91
```

14 aborted requests across 8 days of logs, clustered during heavy 1M-context sessions.

### Environment

- headroom-ai 0.33.0, `--mode token --backend anthropic`
- Claude Code 2.1.221 with `context-1m-2025-08-07`
- Host has 32 GB RAM with 16 GB free at the time, so this is **not** host memory exhaustion
- `resolved_concurrency: 8`

### Root cause

`proxy/handlers/anthropic.py`:

```python
if not body_mutation_tracker.mutated and original_body_bytes is not None:
    try:
        parsed_original = json.loads(original_body_bytes)
        if parsed_original != body:
            body_mutation_tracker.mark_mutated("structural_diff_vs_original")
    except (json.JSONDecodeError, ValueError):
        body_mutation_tracker.mark_mutated("original_unparseable")
```

At that point the process holds `original_body_bytes`, the already-parsed `body`, and a second full parse in `parsed_original`, then deep-compares them. On 1M-context payloads with 8 concurrent workers that is a large transient spike, and `json.loads` raises `MemoryError`. `MemoryError` is not a subclass of `ValueError`, so it is not caught here and it kills a request that was otherwise fine.

Traceback tail:

```
proxy_routes.py, in anthropic_messages
proxy/handlers/anthropic.py, in handle_anthropic_messages
json/__init__.py, line 347, in loads
MemoryError
```

### Suggested fix

The block is a safety net, so it should never be able to fail the request. The existing fallback is already the safe outcome (it forces canonical re-serialization):

```python
except (json.JSONDecodeError, ValueError, MemoryError, RecursionError):
    body_mutation_tracker.mark_mutated("original_unparseable")
```

A larger improvement would be to skip the double parse above a body-size threshold and mark mutated directly, avoiding the spike instead of surviving it.

Happy to open a PR.

JerrettDavis:
[...]

PR #2852, opened 2026-08-08
perf(proxy): bound upstream calls and hot-path costs

chopratejas:
Seven commits from one week of load testing: one hang, two request-path
correctness fixes, and four hot-path costs that only show up in production.

## Reliability

**Bound every upstream call.** The litellm backend had no timeout at all, so a
request the upstream never answered blocked its caller forever. Observed under
load on 2026-08-07: four agent workers on ESTABLISHED connections for 36+
minutes while `/readyz` answered in 0.11s. No error, no retry, no log line —
indistinguishable from slow work, which is the worst shape a failure can take.

A float rather than an `httpx.Timeout`, deliberately: litellm expands a float
across all four httpx phases, so on a streaming call it becomes the maximum gap
*between chunks*, not a cap on total generation. A long answer streaming
steadily is never cut off; a stalled one dies. Default 600s via
`HEADROOM_UPSTREAM_TIMEOUT`; 0, negative, and junk fall back to the default
rather than meaning "no timeout".

**Keep the consistency re-count off the event loop.** It ran
`tokenizer.count_messages` twice directly on the loop. Since Claude counting
moved to a real BPE that is CPU-bound work stalling every other in-flight
request — ~1s on a 2.3 MB body, with `/healthz` gaps tracking body size.
Offloaded via `asyncio.to_thread` on the same tokenizer instance, so reported
values are unchanged. (#2810)

**Survive a re-parse MemoryError.** `MemoryError` is not a `ValueError`, so on
1M-context payloads the byte-faithful forwarder's verification re-parse escaped
the handler and aborted an otherwise-fine request — 14 aborts across 8 days of
reporter logs. (#2768)

## Performance

All four are measured, not guessed. Each degrades with something a short
benchmark does not vary: uptime, content shape, or process age.

| fix | before | after |
|---|---|---|
| Cost-record walk per request (at 100k records) | 13.6 ms | bounded by model count |
| JSON-block scan, JS-style object logs (1200 lines) | 4643 ms | 183 ms |
| JSON-block scan, truncated JSONL | 3737 ms | 116 ms |
| Lazy imports inside user requests | multi-second | paid at startup |
| `count_text` (80% of local CPU) | — | memoised |

Two worth calling out:

- **The cost walk degrades with proxy *uptime*, not load.** A freshly started
  proxy pays ~0.01 ms; a month-old one pays 4–13 ms on every request, on the
  event loop, holding the metrics lock. Deliberately not a TTL cache over
  `stats()`: those values feed `check_budget()` when `--budget` is set, and a
  stale reading under-enforces the budget. The fix is to stop computing what
  the caller discards.
- **The JSON-block memo is built only *after* a scan fails to balance.** That
  ordering is load-bearing, not an optimisation — caching from the start made
  pretty-printed JSON ~2x slower, since content that balances on the first scan
  has nothing to reuse and just pays the per-line dict traffic. Still a
  constant-factor fix, not an asymptotic one.

## Tests

+1202 lines, 20 files. Each fix is pinned by a test that fails on the
unmodified code: the re-count test asserts no `count_messages` pass runs with a
live event loop in its thread; the re-parse test drives a `MemoryError` through
the real request path and expects a 200; `totals()` equality with `stats()` is
asserted across model counts, request volumes, and both pricing branches. The
timeout test is structural rather than a mock — the failure mode is a dispatch
path someone adds later without a guard, which mocking the existing four cannot
catch.


JerrettDavis (changes requested):
This contains several well-measured improvements, but I found two blockers on the current head:

- `headroom/backends/litellm.py` adds `_caller_key_fits_target()` and changes credential forwarding in all four dispatch paths. That is a substantive authentication/routing policy change outside the stated timeout/performance fixes, and the prefix heuristic is not safe for generic Bearer credentials or Anthropic-compatible/custom gateways: any non-`sk-ant-` credential is dropped for an Anthropic-classified model and silently replaced by the process environment credential. In a multi-user or gateway deployment that can both break valid upstream auth and route a request under the wrong account. Please split this behavior into a separately reviewed change with explicit provider/custom-upstream semantics and coverage, or remove it from this PR.

- The branch is not format-clean: `uv run ruff format --check .` reports `tests/test_litellm_upstream_timeout.py` would be reformatted. The current PR title/final commit also fails the repository Conventional Commit policy (`commitlint`). Please format that test and use a scoped Conventional Commit title/message.

I reproduced the format failure locally; `ruff check .` otherwise passes. The bounded timeout, off-loop recount, MemoryError fallback, bounded totals path, and memoization changes look directionally sound, but the credential-policy change needs isolation before this is merge-ready.

JerrettDavis (changes requested):
The corrective commits resolve both earlier blockers: the caller-key/target heuristic is removed, the timeout test is format-clean, and I verified 311 focused timeout/recount/cost/mixed-content/token-cache/preload tests plus Ruff and format checks locally. I also updated the PR title to a scoped Conventional Commit form so governance can reevaluate it.

One merge-scope blocker remains on current head `917a0203`: the diff adds `dashboard-cache-ttl-main.png`, `docs/context-mode-integration-analysis.md` (378 lines), and `run-all-plugins.sh` (171 lines). These are unrelated to the bounded upstream call, recount safety, cost walk, JSON scan, import, and token-count changes; they are not described or validated by this PR and materially broaden the review surface. Please remove them from this branch or submit them separately with their own rationale and tests. Once the diff contains only the stated reliability/performance work and CI is green, the code changes themselves look ready for final review.

chopratejas:
Done — `_caller_key_fits_target()` and all four call-site changes are reverted here (`a986cbe4`), so `git diff main -- headroom/backends/litellm.py` is now purely the timeout work. This PR is back on its stated scope.

The credential change moved to #2883, reworked rather than re-filed. Your objection was the real one: the old `not provider.startswith("anthropic")` dropped **any** non-`sk-ant-` credential against an Anthropic-class target, so a plain Bearer token against a compatible or custom gateway lost a key that was working. #2883 refuses on positive evidence only — `sk-ant-` against an explicitly listed non-Anthropic provider, nothing else — with 12 tests, including each unclassifiable credential shape asserted against both target families. Those fail against the version you rejected.

Also fixed the `ruff format` failure on `test_litellm_upstream_timeout` (`4df6c940`) and merged `main` in. 285 tests pass; `ruff check` and `ruff format --check` clean.

Only remaining red is commitlint.

PR #2853, opened 2026-08-08
Beacon: hourly R2 compaction, per-strategy savings, and a stack that reports

chopratejas:
Three beacon changes bundled because they are one story: the corpus got too
slow to query, and then too coarse to answer the question it was collected for.

## 1. Hourly compaction (`deploy/beacon`)

The beacon writes one ~1 KB object per heartbeat — **64,987 on 2026-08-06** and
climbing. A full analysis `pull` was ~100k HTTPS round trips for 95 MB: minutes
of pure per-object latency. Listing the bucket alone took 88 seconds.

Moving the query server-side does not help — R2 SQL reads only Iceberg tables,
and a pile of tiny files is the pathological case for every query engine.
Compaction is the fix, and it is Iceberg's own answer to the same problem.

An hourly cron collapses each **complete** hour of `sessions/` into one
`rollup/dt=…/hh=…/data.ndjson`, keeping the highest-`seq` heartbeat per
`(install, session)`.

| measured on `dt=2026-08-06/hh=14` | before | after |
|---|---|---|
| objects | 3,938 | **1** |
| rows | 3,938 | **1,061** |
| analysis `pull` | minutes | **seconds** |

Hourly rather than daily because every R2 binding call is a subrequest: a day
is ~65k, an hour is ~4k. Newest-first, so a backlog drains from the present
backwards and live data never starves behind it; a failing hour is logged and
skipped rather than blocking every older hour behind it. **Raw objects are
never deleted**, so any rollup is rebuildable by deleting it.

Backfill runs to the **oldest surviving raw day**, not a fixed window. A fixed
lookback strands everything older than it the moment analysis stops reading
`sessions/`: the raw objects are still there, but nothing would ever compact
them, so they disappear from every report. `oldestRawDay()` finds that floor in
one delimited LIST, and the rollup listing starts from it — so the work is
bounded by retention rather than by total history.

Three failure modes the tests pin down, because each one is silent:

- A **failed `get`** is transient, so the hour throws and writes nothing. A
  rollup is built once and trusted forever, so a short read would quietly
  become the permanent record.
- A **corrupt record** loses only itself. This Worker wrote that content with
  `JSON.stringify`; it will never become valid, so blocking on it strands the
  hour instead of the record.
- An **empty hour** writes an `empty` marker. Without one the hour stays
  "missing" and is re-listed on every run forever.

`test-rollup.mjs` asserts the Worker's dedup picks exactly the same rows as the
analysis-side `QUALIFY`. If those two ever disagree the reports go quietly
wrong rather than loudly broken, which is why that check exists. Its stub
paginates at 3 keys so the list cursor loop — load-bearing at the real ~4,000
objects/hour — runs in every case.

## 2. Per-strategy savings (`compression.by_strategy`)

`compression.transforms` counts *invocations*, which cannot distinguish a
compressor that saved 60% from one that ran constantly and saved nothing. The
fleet's top transform by count contributes an unknown share of `tokens.saved`.

It is worse than that in practice. Transform labels are slugged with
`split(":", 1)[0]`, so every `router:<strategy>:<detail>` label collapses into a
single `router` bucket. On 2026-08-08 that bucket held **19.1 M of the day's
transform counts, across 8,528 of 9,616 sessions** — the compressors that do
most of the work are indistinguishable from each other, by name as well as by
yield:

| transform | n | sessions |
|---|---|---|
| `router` | 19,108,118 | 8,528 |
| `anthropic` | 688,124 | 4,734 |
| `output_shaper` | 548,017 | 1,120 |

No question about which strategy is earning its keep can be answered from that,
which is what this field is for.

The measurement already existed. `PrometheusMetrics.record_compression` is the
configured `CompressionObserver` and already accumulates
`tokens_saved_by_strategy` on the hot path — the numbers just never left the
process. This forwards from that one chokepoint rather than adding a second
observer and a second measurement pass. The paths that have **no** observer
configured (MCP server, LangGraph, Strands hooks, the transform pipeline) get
`BeaconCompressionObserver` passed directly.

Compression runs on the executor thread *before* that request's outcome reaches
`record()`, so events are **staged** into module state and drained by the next
outcome. Staging is what makes two things true at once:

- The first turn of a session still reports its numbers — otherwise every
  session's opening turn, and any session short enough to be one turn, would
  report nothing.
- A compression event **never opens a session**. An abandoned request would
  otherwise emit a phantom `turns=0` row with all-zero tokens, inflating fleet
  session and install counts.

Staging takes a dedicated mutex the request path never touches, so the fan-out
stays off the aggregator's lock and `record_compression` keeps its
"synchronous + lock-free" contract.

```json
"by_strategy": [
  {"strategy": "code_aware",    "n": 1, "tokens_in":  800, "tokens_out": 800},
  {"strategy": "smart_crusher", "n": 2, "tokens_in": 1500, "tokens_out": 700}
]
```

A **list of records, sorted by strategy** — not an object keyed by strategy.
Keyed shapes change type as keys accumulate: DuckDB infers a STRUCT under ~24
keys and a MAP over it, so the analysis query breaks on the day the fleet
picks up a 25th strategy. Sorted so heartbeats are byte-comparable.

**These do not sum to `tokens.saved`,** and the field comment says so:
strategies compose (the router routes, a strategy runs inside it) so the same
text is measured more than once. A row means "of what this strategy was handed,
it removed this much" — a per-strategy yield, not a share of the total. A
strategy that saved nothing still appears; dropping it would make every
strategy look effective.

## 3. `headroom.stack`

`resource_attributes()` was called with no arguments at its one call site, so
`headroom.stack` was absent from **all 24,040 sessions** in the corpus while
`detect_stack` sat unused — dead code on both ends of a wire nobody connected.
The fleet was unsegmentable by agent, which is the question the corpus is asked
most often.

Environment detection alone is not enough. It answers `wrap_claude` only under
`headroom wrap`; every install that points an agent at a persistent proxy — the
common deployment — reports `proxy`, which segments nothing. The per-request
`X-Headroom-Stack` slugs are the only signal that names the harness there, so
`record_stack()` stages them the same way and feeds `detect_stack`'s `by_stack`
branch:

```
9x wrap_claude, 1x wrap_cursor -> wrap_claude   (dominant harness wins)
5x wrap_claude, 5x wrap_cursor -> mixed
no per-request signal          -> proxy         (environment fallback)
junk slug                      -> dropped before staging
```

## Privacy

The strategy string is slugged through the same `_safe_slug` as skip reasons
and capped at `MAX_STRATEGIES`, because the observer protocol takes a free
string and an extension could otherwise invent keys per request. Stack slugs
are normalized and capped the same way.

**Deliberately not collected:** the tool names in `smart_crush:<count>:<names>`.
Those are user-defined MCP identifiers and can name internal tooling
(`acme_deploy_prod`). They stay stripped by the existing `split(":", 1)[0]`,
and this PR does not widen it. No new key was needed in `worker.js` —
`by_strategy` nests under the already-allowlisted `compression`, and
`headroom.stack` was already in `ALLOWED_RESOURCE`.

Nothing here deletes or rewrites existing data: the Worker only ever writes,
`sessions/` is never pruned by it, and readers merge old and new shapes with
`union_by_name`, so pre-change heartbeats keep reading with the new fields null.

## Verification

- `python -m headroom.telemetry.session` — self-check covers staging, the
  no-phantom-session case, drain exhaustiveness, a 0%-yield strategy staying
  visible, the cardinality cap, slug safety, and dominant/mixed/junk stack
  resolution
- `node test-rollup.mjs /tmp/hr` — 3,938 real corpus objects → 1,061 sessions
  in 1 object, plus the pagination, partial-read, corrupt-record, empty-hour
  and `oldestRawDay` cases
- 51 passing in `test_compression_observability`, `test_prometheus_obs_counters`,
  `test_telemetry_context`, `test_compression_strategy_outcomes`
- Consumer side exists and is checked: `beacon.sh by_strategy` in
  headroom-beacon-stats reads the field end to end (sessions, installs,
  invocations, tokens in/out, yield %), verified against a synthetic parquet for
  the cases that matter — aggregation across installs, a 0%-yield strategy
  staying visible, pre-field sessions dropping out rather than erroring. It is
  guarded on the column, so it prints an instruction instead of a binder error
  until a release carrying this PR reaches the fleet.
- Deployed and running against the live corpus on the `5 * * * *` trigger

### DIFF
--- a/deploy/beacon/package.json
+++ b/deploy/beacon/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "headroom-beacon",
+  "private": true,
+  "type": "module"
+}
--- a/deploy/beacon/test-rollup.mjs
+++ b/deploy/beacon/test-rollup.mjs
@@ -1,0 +1,109 @@
+/**
+ * Self-check for scheduled()'s hourly compaction.  node test-rollup.mjs [dir]
+ *
+ * The one thing that must never drift: rollupHour() and the QUALIFY in
+ * headroom-beacon-stats/beacon.sh have to agree on which heartbeat wins. If
+ * they disagree the reports get quietly wrong rather than loudly broken, so
+ * this asserts the JS picks exactly the max-seq row per (install, session).
+ *
+ * Point it at a directory of real beacon objects to check against the corpus:
+ *   aws s3 sync s3://headroom-telemetry/sessions/dt=.../hh=.../ /tmp/hr/ ...
+ *   node test-rollup.mjs /tmp/hr
+ * With no argument it runs on a small fixture and needs no network.
+ */
+import { readdirSync, readFileSync } from 'node:fs';
+import assert from 'node:assert/strict';
+import { rollupHour } from './worker.js';
+
+/** The slice of the R2 binding rollupHour uses, backed by a plain object. */
+function stubBucket(files) {
+  const written = {};
+  return {
+    written,
+    list: async ({ prefix }) => ({
+      objects: Object.keys(files)
+        .filter((k) => k.startsWith(prefix))
+        .map((key) => ({ key })),
+      truncated: false,
+    }),
+    get: async (key) => ({ text: async () => files[key] }),
+    put: async (key, body) => {
+      written[key] = body;
+    },
+  };
+}
+
+const beacon = (install, id, seq) =>
+  JSON.stringify({ resource: { 'headroom.install_id': install }, session: { id, seq } });
+
+async function run(files, part) {
+  const CORPUS = stubBucket(files);
+  const read = await rollupHour({ CORPUS }, part);
+  const out = CORPUS.written[`rollup/${part}/data.ndjson`];
+  return { read, rows: out ? out.split('\n').map((l) => JSON.parse(l)) : [] };
+}
+
+const PART = 'dt=2026-08-06/hh=14';
+
+// 1. Highest seq wins, out-of-order input, one row per (install, session).
+{
+  const files = {
+    [`sessions/${PART}/a.json`]: [beacon('i1', 's1', 3), beacon('i1', 's2', 1)].join('\n'),
+    [`sessions/${PART}/b.json`]: beacon('i1', 's1', 9),
+    [`sessions/${PART}/c.json`]: beacon('i1', 's1', 7),
+    // Same session id under a different install must not collapse together.
+    [`sessions/${PART}/d.json`]: beacon('i2', 's1', 2),
+  };
+  const { read, rows } = await run(files, PART);
+  assert.equal(read, 4, 'reads every object in the hour');
+  assert.equal(rows.length, 3, 'one row per (install, session)');
+  const seq = Object.fromEntries(
+    rows.map((r) => [`${r.resource['headroom.install_id']} ${r.session.id}`, r.session.seq])
+  );
+  assert.deepEqual(seq, { 'i1 s1': 9, 'i1 s2': 1, 'i2 s1': 2 });
+}
+
+// 2. An unparseable object loses itself, never the rest of the hour, and an
+//    empty hour writes nothing rather than a zero-byte object.
+{
+  const files = {
+    [`sessions/${PART}/a.json`]: '{ this is not json',
+    [`sessions/${PART}/b.json`]: `\n${beacon('i1', 's1', 4)}\n`,
+  };
+  const { rows } = await run(files, PART);
+  assert.deepEqual(rows.map((r) => r.session.seq), [4], 'survives a corrupt object');
+
+  const empty = await run({}, PART);
+  assert.deepEqual(empty.rows, [], 'empty hour writes no object');
+}
+
+// 3. Against real objects, if a directory was given: same answer as the QUALIFY
+//    in beacon.sh, which is `count(DISTINCT install||session)` rows, each
+//    carrying that pair's max seq.
+const dir = process.argv[2];
+if (dir) {
+  const files = {};
+  for (const f of readdirSync(dir).filter((f) => f.endsWith('.json'))) {
+    files[`sessions/${PART}/${f}`] = readFileSync(`${dir}/${f}`, 'utf8');
+  }
+  const { read, rows } = await run(files, PART);
+
+  const expected = new Map();
+  for (const text of Object.values(files)) {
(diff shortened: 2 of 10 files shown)


JerrettDavis (changes requested):
This is a thoughtful measurement/compaction design, but the latest head is not deploy-ready yet. I found one production correctness blocker in addition to the two red checks:

1. `scheduled()` calls the Workers R2 binding as `CORPUS.list({ prefix: "rollup/", startAfter: ..., cursor })`. `startAfter` is not a supported `R2ListOptions` field for the Workers binding; the official API exposes `prefix`, `cursor`, `delimiter`, `limit`, and `include` only: https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#r2listoptions. The local stub destructures only selected fields and silently ignores `startAfter`, so it cannot prove the claimed oldest-raw-day lower bound. In production this listing starts at the beginning of `rollup/`, making the scan grow with total rollup history—the exact unbounded behavior the comments say is avoided. Please implement the lower bound using supported cursor/prefix semantics (or explicitly filter with a tested bounded listing strategy) and add a scheduled-handler test that fails when old rollups are enumerated.

2. `lint` is genuinely red: mypy reports the `SmartCrusherConfig` type mismatch at `headroom/integrations/mcp/server.py:272`. The ignore is attached to the call opener, not the incompatible `config=` argument, so it does not suppress that diagnostic. Please resolve the config-type seam or place a narrowly justified suppression on the reported expression.

3. `commitlint` is genuinely red because the first commit subject starts with `Beacon:` rather than an allowed conventional-commit type. Both commits in the PR range need to satisfy the repository gate.

The rollup tests and telemetry staging checks otherwise cover several important failure modes well; fixing these three items would make the platform behavior and CI evidence match the description.

JerrettDavis (changes requested):
The mypy suppression is now attached to the reported argument, but the other two blockers from the prior review remain on this head: `scheduled()` still passes unsupported `startAfter` to the Workers R2 binding, while the stub silently ignores it, so production rollup enumeration remains unbounded by the claimed oldest-day floor; and commitlint remains red because the first commit subject (`Beacon: ...`) is not a valid conventional-commit subject. Please replace the unsupported lower-bound mechanism with a supported, production-faithful implementation and test, and repair the commit history so the required gate passes.

PR #2854, opened 2026-08-08
fix(opencode): send x-headroom-project so per-project savings can attribute traffic

### LINKED ISSUE #2847: [BUG] OpenCode plugin never sends x-headroom-project — Per-Project Savings dashboard can't attribute OpenCode traffic
## Summary

The OpenCode transport plugin (`headroom/providers/opencode/_dist/entry.opencode.js`) sets `HEADROOM_PROJECT` as a spawned-subprocess env var only (`shell.env` hook, derived from `input.project.id`), but the `fetch`/`http`/`https` wrapping installed by `installHeadroomTransport` only ever sets `x-headroom-base-url` / `x-headroom-original-path` headers on routed requests. It never sets `x-headroom-project`.

Server-side, `headroom/proxy/project_policy.py` (`classify_project`) only attributes savings via the `x-headroom-project` header or a `/p/<name>` URL-path prefix. Since the OpenCode plugin never sends either on the actual proxied HTTP requests, every OpenCode request routed through Headroom is unattributed — the dashboard's "Per-Project Savings" panel permanently shows `0 project(s)` for OpenCode users, even with `--memory --memory-storage project` running correctly (that's a separate, cwd-based attribution path used only for the memory DB, not for cost/savings attribution).

## Environment

- headroom-ai 0.34.0
- OpenCode (current, plugin loaded via packaged `_dist/entry.opencode.js`)
- macOS arm64

## Reproduction

1. Run `headroom proxy --memory --memory-storage project` (or any profile with the OpenCode plugin active).
2. Use OpenCode with the Headroom transport plugin loaded (`OPENCODE_CONFIG_CONTENT` → `plugin: [...entry.opencode.js]`).
3. Make several requests through the proxy.
4. Open the Headroom dashboard → "Per-Project Savings" panel stays at `0 project(s)` / "No per-project data yet." regardless of traffic volume.

## Expected

OpenCode traffic routed through the proxy should be attributable per-project, same as Claude Code (`/p/<name>` base URL) or any client sending `x-headroom-project`.

## Suggested fix

Resolve `input.project.id` (falling back to `pluginOptions.project` / `input.directory`) once at plugin init in `HeadroomPlugin`, thread it through `installHeadroomTransport`'s state, and set an `x-headroom-project` header alongside the existing `x-headroom-base-url` header in both header-construction seams:

- `mergeFetchHeaders` (used by the wrapped `fetch`)
- `headersForNodeRequest` (used by the wrapped `http.request`/`https.request`)

I have a working patch against the packaged `_dist/entry.opencode.js` (adds a `project` parameter threaded through `installHeadroomTransport` → `wrapRequest`/`routedNodeOptions` → `headersForNodeRequest`, and through the `fetch` wrapper → `withRoutedFetchInput` → `mergeFetchHeaders`) and can share a diff/PR if useful.

## Related but distinct issues

- #2355 — `/p/<project>` WebSocket 403 for `headroom wrap codex` (a routing bug on an already-attributed path, not this gap)
- #2678 — packaged OpenCode plugin references a missing `hook-shim/handler.js` file (separate bundling bug, unrelated to project attribution)

abhay-codes07:
## Description

The OpenCode transport plugin sets `HEADROOM_PROJECT` only as a spawned-subprocess env var (the `shell.env` hook, derived from `input.project.id`), but the `fetch` / `http` / `https` wrapping that `installHeadroomTransport` installs only ever sets `x-headroom-base-url` and `x-headroom-original-path` on routed requests. It never sets `x-headroom-project`.

Server-side, `headroom/proxy/project_policy.py` (`classify_project`) attributes savings only via the `x-headroom-project` header or a `/p/<name>` URL-path prefix. Since the OpenCode plugin sends neither on the actual proxied HTTP requests, every OpenCode request is unattributed and the dashboard's "Per-Project Savings" panel permanently shows `0 project(s)`, even with per-project memory storage running (that path is cwd-based and only feeds the memory DB, not cost/savings attribution).

The fix resolves the project once at plugin init (`project` option -> `input.project.id` -> `input.directory`), threads it into the transport state, and sets `x-headroom-project` alongside `x-headroom-base-url` in both header-construction seams:

- `mergeFetchHeaders` (used by the wrapped `fetch`)
- `headersForNodeRequest` (used by the wrapped `http.request` / `https.request`)

The same resolved value now backs the `shell.env` `HEADROOM_PROJECT` export, so the header and the child-process env agree. When no project resolves, the header is simply omitted (behavior unchanged for that case).

Fixes #2847

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `plugins/opencode/src/transport.ts`: added a `PROJECT_HEADER` constant, a `project` field on `InstallOptions`/`TransportState`, and set `x-headroom-project` in `mergeFetchHeaders` and `headersForNodeRequest` when a project is configured.
- `plugins/opencode/src/plugin.ts`: resolve the project once via `resolveProject(...)`, pass it to `installHeadroomTransport`, and reuse it for the `shell.env` `HEADROOM_PROJECT` export.
- `plugins/opencode/src/transport.test.ts`: added three tests: the header is present on a routed `fetch` and on a routed Node `https.request` when a project is configured, and absent when it is not.
- `headroom/providers/opencode/_dist/entry.opencode.js`: rebuilt the committed wheel bundle via `npm run build:standalone` (the CI `opencode-plugin.yml` guard byte-compares this against the source build).

## Testing

- [x] Unit tests pass (`npm test` / vitest)
- [x] Type checking passes (`npm run typecheck` / `tsc --noEmit`)
- [x] Committed wheel bundle rebuilt and byte-matches the standalone build

### Test Output

```text
# Fail-before (transport.ts source stashed, new tests kept):
src/transport.test.ts (15 tests | 2 failed)
  × sets x-headroom-project on routed fetch when a project is configured (#2847)
  × sets x-headroom-project on routed Node https.request when a project is configured (#2847)

# Pass-after (fix applied):
npm run typecheck  -> tsc --noEmit (clean)
npm test           -> Test Files 2 passed (2), Tests 17 passed (17)

# Committed wheel bundle rebuilt + CI guard reproduced:
npm run build:standalone
cmp dist-standalone/entry.opencode.js ../../headroom/providers/opencode/_dist/entry.opencode.js  -> match
# committed blob is LF-only (0 CR), so the Linux CI byte-compare matches
```

## Real Behavior Proof

- Environment: Windows 11, Node v24.11.0, npm 11.5.2, vitest 4.1.10, tsup 8.5.1 / esbuild 0.28.1 (matching the committed `package-lock.json`).
- Exact command / steps: traced the gap (transport wrapping sets `x-headroom-base-url`/`x-headroom-original-path` but never `x-headroom-project`, while `classify_project` reads only `x-headroom-project` or `/p/<name>`), added the header in both seams sourced from transport state, resolved the project once in the plugin. Fail-before with `git stash push plugins/opencode/src/transport.ts` and `npm test` (the two project-header tests fail: header not set), pass-after with `git stash pop` and rerunning (`17 passed`), then `npm run build:standalone` and `cmp` against the committed bundle (match).
- Observed result: a routed `fetch` and a routed Node `https.request` now carry `x-headroom-project: <project>` when the plugin resolved a project, so `classify_project` can attribute the traffic and the Per-Project Savings panel populates for OpenCode. With no project resolved the header is omitted, unchanged.
- Not tested: a live OpenCode session against a running proxy dashboard (no OpenCode client in this environment). The header emission is verified at both transport seams with the real wrapped `fetch` and a real loopback Node `https.request` captured by an in-test HTTP server.

## Review Readiness

## Checklist

## Additional Notes

Scoped to project attribution on the routed HTTP requests, which is the reported gap. The child-process `HEADROOM_PROJECT` env var (which reaches spawned subprocesses but not the proxied requests) is preserved and now shares the single resolved value. The `package-lock.json` was left untouched: this change adds no dependencies, and the only local churn was Windows dropping optional non-Windows `@emnapi/*` entries during `npm install`, which I reverted.

### DIFF
--- a/headroom/providers/opencode/_dist/entry.opencode.js
+++ b/headroom/providers/opencode/_dist/entry.opencode.js
@@ -12487,6 +12487,7 @@ var childProcess = nodeRequire("node:child_process");
 var fs = nodeRequire("node:fs");
 var BASE_URL_HEADER = "x-headroom-base-url";
 var ORIGINAL_PATH_HEADER = "x-headroom-original-path";
+var PROJECT_HEADER = "x-headroom-project";
 var PROXY_ENV = "HEADROOM_OPENCODE_TRANSPORT_PROXY_URL";
 var STATE_KEY = /* @__PURE__ */ Symbol.for("headroom.opencode.transport");
 function getState() {
@@ -12642,6 +12643,10 @@ function mergeFetchHeaders(input, init, upstream, originalPath = void 0) {
   }
   if (upstream) {
     headers.set(BASE_URL_HEADER, upstream.origin);
+    const project = getState()?.project;
+    if (project) {
+      headers.set(PROJECT_HEADER, project);
+    }
     headers.delete("host");
   }
   if (originalPath) {
@@ -12706,6 +12711,10 @@ function urlFromRequestOptions(options) {
 function headersForNodeRequest(options, upstream, originalPath) {
   const headers = new Headers(options.headers);
   headers.set(BASE_URL_HEADER, upstream.origin);
+  const project = getState()?.project;
+  if (project) {
+    headers.set(PROJECT_HEADER, project);
+  }
   if (originalPath) {
     headers.set(ORIGINAL_PATH_HEADER, originalPath);
   }
@@ -12795,6 +12804,7 @@ function installHeadroomTransport(options) {
     existing.refs += 1;
     existing.proxyUrl = options.proxyUrl;
     existing.debug = Boolean(options.debug);
+    existing.project = options.project ?? existing.project;
     installProcessEnv(options.proxyUrl);
     return () => uninstallHeadroomTransport();
   }
@@ -12802,6 +12812,7 @@ function installHeadroomTransport(options) {
     refs: 1,
     proxyUrl: options.proxyUrl,
     debug: Boolean(options.debug),
+    project: options.project,
     originalFetch: globalThis.fetch,
     originalHttpRequest: http.request,
     originalHttpGet: http.get,
@@ -12868,13 +12879,18 @@ function resolveProxyUrl(options) {
     options?.proxyUrl ?? process.env.HEADROOM_PROXY_URL ?? process.env.HEADROOM_BASE_URL ?? getDefaultProxyUrl()
   );
 }
+function resolveProject(options, input) {
+  return options.project ?? input.project?.id ?? input.directory;
+}
 var HeadroomPlugin = async (input, options = {}) => {
   const pluginOptions = options;
   const proxyUrl = resolveProxyUrl(pluginOptions);
+  const project = resolveProject(pluginOptions, input);
   const retrieveTool = createHeadroomRetrieveTool({ proxyBaseUrl: proxyUrl });
   const uninstallTransport = installHeadroomTransport({
     proxyUrl,
-    debug: pluginOptions.debug
+    debug: pluginOptions.debug,
+    project
   });
   return {
     dispose: async () => {
@@ -12894,7 +12910,9 @@ var HeadroomPlugin = async (input, options = {}) => {
     "shell.env": async (_input, output) => {
       output.env.HEADROOM_ACTIVE = "1";
       output.env.HEADROOM_PROXY_URL = proxyUrl;
-      output.env.HEADROOM_PROJECT = pluginOptions.project ?? input.project.id ?? input.directory;
+      if (project) {
+        output.env.HEADROOM_PROJECT = project;
+      }
       if (pluginOptions.backend) {
         output.env.HEADROOM_BACKEND = pluginOptions.backend;
       }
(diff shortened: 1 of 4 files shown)


JerrettDavis:
This now overlaps functionally with #2868, which covers the same issue and both routed transport seams and has been reviewed on its latest head. Please coordinate on a single implementation so we do not land two competing changes to the same transport state/header paths. I recommend keeping one PR open and closing the other as superseded; if this one is retained, reconcile its `existing.project = options.project ?? existing.project` behavior with #2868’s explicit clearing behavior and add a test for repeated installs so the intended singleton-state semantics are unambiguous.

abhay-codes07:
Agreed, thanks for flagging. #2868 covers the same issue (#2847) across both transport seams and is already approved on its latest head, so closing this as superseded to avoid landing two competing changes to the same transport state and header paths. Happy to review or help on #2868 if useful.

PR #2830, opened 2026-08-06
fix(cli/update): let install ownership win over bare /.dockerenv so venv installs self-update

abhay-codes07:
## Description

`headroom update` refuses to self-update for any install that happens to run inside a container, including a plain `pip install` into a venv, because `detect_install_method` checks `_in_docker()` before the pipx / uv-tool / venv / user-site branches. The guidance it prints does not apply: there is no Headroom image in the picture, the container is the environment and Headroom was pip-installed into a venv inside it.

```console
$ headroom update --check
Update available: 0.32.0 -> 0.34.0
Running inside a container - pull a newer Headroom image instead of self-updating.
```

`_in_docker()` is purely environmental (`/.dockerenv` exists, or `HEADROOM_IN_DOCKER` is set), with no reference to how the package was installed, so `/.dockerenv` alone shadows a venv that clearly owns the install. This hits devcontainers, GitHub Codespaces, docker/LXC self-hosting, and dev images.

The fix splits the check by intent. An EXPLICIT `HEADROOM_IN_DOCKER` (which the official image can set) is a deliberate opt-out and still refuses up front, even over a venv, so the real-image behavior is preserved. The bare `/.dockerenv` heuristic now runs after ownership detection, so a venv / pipx / uv / user-site install self-updates and only a container whose own system interpreter owns the install still gets the pull-a-new-image guidance.

Fixes #2816

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `headroom/cli/update.py` (`detect_install_method`): replaced the up-front `_in_docker()` refusal with an explicit `os.environ.get("HEADROOM_IN_DOCKER")` refusal (the official image opt-out), and added the bare `_in_docker()` refusal after the pipx / uv-tool / venv / user-site branches so ownership wins over environment. Updated the resolution-order docstring.
- `tests/test_update_helpers.py`: added `test_venv_inside_bare_dockerenv_still_self_updates` (the fix), `test_explicit_headroom_in_docker_still_refuses_over_venv` (image opt-out preserved), and `test_bare_dockerenv_without_owner_refuses` (system-interpreter container still refuses).
[...]

PR #2866, opened 2026-08-08
fix(update): skip docker guard when install is a venv/pip inside a container

### LINKED ISSUE #2816: [BUG] `headroom update` refuses any pip/pipx install that merely runs inside a container: `_in_docker()` is checked before the venv/pipx branches, so the venv path is unreachable
### Summary

`headroom update` refuses to self-update for **any** install that happens to run inside a container, including a plain `pip install` into a venv, because the Docker check runs before the pipx / uv / venv checks. The guidance it prints ("pull a newer Headroom image") does not apply: there is no Headroom image in the picture — the container is the *environment*, and Headroom was pip-installed into a venv inside it.

```console
$ headroom update --check
Checking PyPI for the latest Headroom release...
Update available: 0.32.0 → 0.34.0
Running inside a container — pull a newer Headroom image instead of self-updating.
```

There is no other supported route offered, so the documented self-update path is a dead end for this whole class of setups. Doing by hand exactly what the venv branch would have done works fine:

```console
$ /path/to/venv/bin/pip install -U headroom-ai==0.34.0     # -> 0.34.0
$ headroom install restart --profile init-user
Restarted deployment 'init-user'.
# /health -> ready:true, version 0.34.0; 24x 200, 0x 400 on the next requests
```

### Where it comes from (`main`, `headroom/cli/update.py`)

Resolution order — the docstring lists Docker as step 3 and `venv / virtualenv / conda → sys.executable -m pip install -U` as step 6, and first match wins:

- `:339` `if _is_source_checkout():`
- `:345` `if _is_editable_install():`
- `:354` `if _in_docker():` ← refuses here
- `:392` `if _in_virtualenv():` ← never reached in a container

And the Docker test is purely environmental, with no reference to how the package was installed:

```python
def _in_docker() -> bool:
    return Path("/.dockerenv").exists() or bool(os.environ.get("HEADROOM_IN_DOCKER", "").strip())
```

So `/.dockerenv` alone decides, even though `sys.prefix != sys.base_prefix` proves a venv owns the install and pipx / uv-tool ownership is likewise detectable a few lines below.

This hits devcontainers, GitHub Codespaces, docker/LXC-based self-hosting and dev images — anywhere the *environment* is a container but the *install* is pip / pipx / uv.

### Environment

- headroom 0.32.0 → 0.34.0, Linux, Python 3.12
- `/.dockerenv` present; `sys.prefix=/config/.headroom-venv`, `sys.base_prefix=/usr` → `_in_virtualenv()` is `True`
- deployment: preset `persistent-task`, runtime `python`, supervisor `none`, scope `user`

### Proposal

Let ownership win over environment: move the `_in_docker()` refusal **after** the pipx / uv-tool / venv detection, or gate it on nothing else claiming the install:

```python
if _in_docker() and not (_in_virtualenv() or _owned_by_pipx(haystack) or _owned_by_uv(haystack)):
    ...refuse with the image guidance...
```

The real shipped image doesn't need the `/.dockerenv` heuristic anyway — `_in_docker()` already honours `HEADROOM_IN_DOCKER`, which the image can set explicitly, keeping "pull a newer image" advice for exactly the case where it is true.

gingeekrishna:
## Problem

`headroom update` refuses to self-update when `/.dockerenv` exists, with the message *"Running inside a container — pull a newer Headroom image instead."* But this blocks a legitimate and common pattern: pip/pipx/venv installs *inside* a container that are not Headroom-managed Docker images.

Fixes #2816.

## Root cause

`detect_install_method()` checks `_in_docker()` before `_in_virtualenv()`. Any container with `/.dockerenv` present (which is every Docker container, regardless of how Headroom was installed) hits the docker guard and refuses.

## Solution

Add `and not _in_virtualenv()` to the docker guard. The resolution order is already documented in `detect_install_method()`'s docstring — pip/venv installs should rank higher than the container heuristic.

When running in Docker but inside a venv, the function falls through to the normal pip/virtualenv branch (step 6), which is exactly right: users activated a venv, so `pip install -U headroom-ai` is appropriate.

The updated guidance message also mentions the pip path so users are never left wondering what to do.

## Changes

- `headroom/cli/update.py` — `detect_install_method()`: guard becomes `if _in_docker() and not _in_virtualenv()`.
- `tests/test_cli_update.py`:
  - `test_detect_docker`: explicitly mocks `_in_virtualenv=False` (was relying on fall-through order; now needs explicit pin after fix).
  - `test_detect_docker_in_venv_bypasses_docker_refuse` (new): asserts that `_in_docker=True, _in_virtualenv=True` resolves to `kind="pip", can_self_update=True`.

## Testing

```
$ python -m pytest tests/test_cli_update.py -v --no-header
...
PASSED tests/test_cli_update.py::test_detect_docker
PASSED tests/test_cli_update.py::test_detect_docker_in_venv_bypasses_docker_refuse
... (all 20 tests pass)
```

### DIFF
--- a/headroom/cli/update.py
+++ b/headroom/cli/update.py
@@ -351,12 +351,14 @@ def detect_install_method(extras: str | None = None) -> InstallMethod:
                 "reinstall with `pip install -U --force-reinstall .`."
             ),
         )
-    if _in_docker():
+    if _in_docker() and not _in_virtualenv():
         return InstallMethod(
             kind="docker",
             can_self_update=False,
             guidance=(
                 "Running inside a container — pull a newer Headroom image instead of self-updating."
+                " If Headroom was pip/pipx-installed into a venv inside this container,"
+                " activate that venv and run `pip install -U headroom-ai` directly."
             ),
         )
 
--- a/headroom/cli/update.py
+++ b/headroom/cli/update.py
@@ -357,8 +357,6 @@ def detect_install_method(extras: str | None = None) -> InstallMethod:
             can_self_update=False,
             guidance=(
                 "Running inside a container — pull a newer Headroom image instead of self-updating."
-                " If Headroom was pip/pipx-installed into a venv inside this container,"
-                " activate that venv and run `pip install -U headroom-ai` directly."
             ),
         )
 
--- a/tests/test_cli_update.py
+++ b/tests/test_cli_update.py
@@ -41,8 +41,22 @@ def test_detect_editable(monkeypatch):
 
 def test_detect_docker(monkeypatch):
     monkeypatch.setattr(up, "_in_docker", lambda: True)
+    monkeypatch.setattr(up, "_in_virtualenv", lambda: False)
     m = up.detect_install_method()
     assert m.kind == "docker" and m.can_self_update is False
+
+
+def test_detect_docker_in_venv_bypasses_docker_refuse(monkeypatch):
+    """A pip/venv install inside a container must not be blocked by the docker guard.
+
+    The docker check is there for Headroom-managed images; it must not prevent
+    self-update for a plain ``pip install`` into a venv that happens to run inside
+    a container (devcontainer, Codespace, self-hosted LXC, …).
+    """
+    monkeypatch.setattr(up, "_in_docker", lambda: True)
+    monkeypatch.setattr(up, "_in_virtualenv", lambda: True)
+    m = up.detect_install_method()
+    assert m.kind == "pip" and m.can_self_update is True
 
 
 def test_detect_pipx_by_path(monkeypatch):


JerrettDavis (approved):
Reviewed the install-method precedence and the updated regression coverage. The explicit managed-image signal still wins earlier in detection, while a real venv inside a generic container now reaches the existing pip/pipx classification. Verified the latest head locally: all 24 `tests/test_cli_update.py` tests pass.

JerrettDavis (approved):
The new delta only removes guidance that could imply direct pip updating is appropriate for the managed-image refusal path; the previously reviewed venv/container precedence and regression coverage remain intact. Current CI is green.

JerrettDavis:
This overlaps #2830 on the same issue and install-detection path. #2830 carries the more complete current contract: explicit `HEADROOM_IN_DOCKER` remains an up-front managed-image opt-out, bare `/.dockerenv` is evaluated only after all owned-install methods (pipx, uv tool, venv, and user site), and the broader 95-test update suite covers the precedence. Since #2830 is approved and actively maintained, this narrower duplicate has been superseded and can be closed in favor of #2830.

PR #2833, opened 2026-08-06
fix(settings): accept documented HEADROOM_* env names as settings keys

axelray-dev:
## Description

Settings validation only accepted short JSON/API keys, so documented HEADROOM_* env names were rejected as unknown. Users following the docs (for example HEADROOM_LOSSLESS) hit SettingsValidationError / PUT /settings 400 even though those names are already on each registry field.

This normalizes known env aliases to their short keys before validate/save, keeps existing short-key behavior, and rejects conflicting env+key pairs for the same field.

Closes #2812

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Added _BY_ENV and _normalize_values() in settings_store to map documented env names to short keys
- Call normalization at the start of validate() and save() so clear/retain paths also accept env aliases
- Reject payloads that supply both an env alias and its short key with different values
- Add unit coverage for accept/clear/conflict/same-value paths and update registry monkeypatches to rebuild _BY_ENV

## Testing

- [x] Unit tests pass (pytest)
- [x] Linting passes (ruff check .)
- [ ] Type checking passes (mypy headroom)
- [x] New tests added for new functionality
- [ ] Manual testing performed

### Test Output

```text
python -m pytest tests/test_proxy/test_settings_store.py -q -k "env_alias or validate_accepts or save_rejects or same_env or conflicting or save_accepts or env_alias_clear or anthropic_extra_headers_retain or TestValidation"
23 passed, 11 deselected

ruff check headroom/settings_store.py tests/test_proxy/test_settings_store.py tests/test_proxy_settings_endpoints.py
All checks passed!

ruff format --check headroom/settings_store.py tests/test_proxy/test_settings_store.py tests/test_proxy_settings_endpoints.py
3 files already formatted
```

## Real Behavior Proof

- Environment: Linux x86_64, Python 3.14.5 via contributor venv, worktree of headroom main at 7940c05e plus commit e4c87edf
- Exact command / steps: pytest tests/test_proxy/test_settings_store.py focused selection; ruff check and ruff format --check on the three touched files; settings_store.validate({"HEADROOM_LOSSLESS": True}) returns {"lossless": True}
[...]

PR #2867, opened 2026-08-08
fix(settings): accept HEADROOM_* env-name aliases in validate() and save()

### LINKED ISSUE #2812: [BUG] Settings registry rejects the documented `HEADROOM_*` env names as unknown keys — no alias, no "did you mean"
### Summary

Every knob in the settings registry carries two names — `SettingField.env` (the `HEADROOM_*` variable) and `SettingField.key` (the JSON/API key) — but validation indexes only by `key`. Passing the env name that the docs and issue threads use is rejected as an unknown key, with no hint that a short key exists or what it is:

```console
$ python -c 'from headroom import settings_store as s; s.save({"HEADROOM_LOSSLESS": True})'
SettingsValidationError: settings validation failed: unknown=['HEADROOM_LOSSLESS'] errors={}

# what actually works — discoverable only by reading the registry:
$ python -c 'from headroom import settings_store as s; s.save({"lossless": True})'
```

Same on the HTTP surface: `PUT /settings` 400s with `unknown_keys` (covered by `tests/test_proxy_settings_endpoints.py:123 test_unknown_key_400`).

This bites because the env name is the only name most users ever see: `docs/content/docs/proxy.mdx` documents the knobs as `HEADROOM_TOOL_SEARCH`, `HEADROOM_LOSSLESS`, … and issue threads discuss them the same way. Guessing the mapping happens to work for some knobs (`HEADROOM_LOSSLESS` → `lossless`) and the error gives no signal when it doesn't.

### Environment

headroom 0.32.0, Linux, Python 3.12 — verified against `main` today.

### Where it comes from (`main`)

- `headroom/settings_store.py:36-60` — `SettingField(env=..., key=..., ...)`: both names are present for every knob.
- `:699` — `_BY_KEY = {f.key: f for f in SETTINGS}`; there is no `_BY_ENV`.
- `:789-799` — `validate()`: `unknown = [key for key in values if key not in _BY_KEY]` → `SettingsValidationError(unknown, field_errors)`.
- `:702` — the error carries `unknown_keys` only; the API layer maps it to a 400.

### Proposal

The registry already has both halves of the mapping:

```python
_BY_ENV = {f.env: f for f in SETTINGS}
```

Then either

- **accept env aliases** — normalize `HEADROOM_X` → its `key` at the top of `validate()`/`save()`. Idempotent, no behavior change for existing callers; or
- **at minimum, hint** — add `did_you_mean` to `SettingsValidationError` and the 400 body when an unknown key matches a known `env` (or a case-insensitive / underscore-normalized variant of a `key`).

The first makes `s.save({"HEADROOM_LOSSLESS": True})` and `PUT /settings {"HEADROOM_TOOL_SEARCH": "0"}` just work, which is what anyone arriving from the docs will type.

gingeekrishna:
## Problem

Every setting has two names: a short canonical `key` (e.g. `"lossless"`) and an env-var name (e.g. `"HEADROOM_LOSSLESS"`). The docs and issue threads use the env name — that's the only name most users ever see. But `validate()` / `save()` / `PUT /settings` only indexed by `key`, so:

```
$ python -c 'from headroom import settings_store as s; s.save({"HEADROOM_LOSSLESS": True})'
SettingsValidationError: settings validation failed: unknown=['HEADROOM_LOSSLESS'] errors={}
```

Fixes #2812.

## Solution

Two minimal additions to `settings_store.py`:

1. `_BY_ENV: dict[str, SettingField] = {f.env: f for f in SETTINGS}` — mirror of `_BY_KEY` indexed by env name.
2. `_normalize_keys(values)` — rewrites any `HEADROOM_*` key found in `_BY_ENV` to its canonical `key`. Keys already in canonical form, and genuinely unknown keys, pass through unchanged.

`_normalize_keys` is called at the top of both `validate()` and `save()`, so the fix covers:
- `s.validate({"HEADROOM_LOSSLESS": True})` → works, returns `{"lossless": True}`
- `s.save({"HEADROOM_RPM": 42})` → works, stores `{"rpm": 42}`
- `s.save({"HEADROOM_RPM": None})` → works, clears the key
- `PUT /settings {"HEADROOM_TOOL_SEARCH": "0"}` → works (goes through `validate()`)
- `s.save({"totally_bogus": 1})` → still raises `SettingsValidationError(unknown_keys=["totally_bogus"])`

No schema or file-format change — stored settings always use the canonical key.

## Changes

- `headroom/settings_store.py` — add `_BY_ENV`, `_normalize_keys()`, call it at the top of `validate()` and `save()`.
- `tests/test_settings_store.py` (new) — 12 unit tests covering `_normalize_keys`, `validate`, and `save` for both env aliases and canonical keys. These tests import `settings_store` directly (no `headroom._core` dependency) and run locally.

## Testing

```
$ python -m pytest tests/test_settings_store.py -v
...
12 passed in 1.54s
```

### DIFF
--- a/headroom/settings_store.py
+++ b/headroom/settings_store.py
@@ -700,21 +700,6 @@ _BY_KEY: dict[str, SettingField] = {f.key: f for f in SETTINGS}
 _BY_ENV: dict[str, SettingField] = {f.env: f for f in SETTINGS}
 
 
-def _normalize_keys(values: dict[str, Any]) -> dict[str, Any]:
-    """Rewrite any HEADROOM_* env-name key to its canonical registry key.
-
-    Keys already in canonical form, or genuinely unknown keys, pass through
-    unchanged so callers can still surface an ``unknown_keys`` error for them.
-    """
-    out: dict[str, Any] = {}
-    for key, value in values.items():
-        if key not in _BY_KEY and key in _BY_ENV:
-            out[_BY_ENV[key].key] = value
-        else:
-            out[key] = value
-    return out
-
-
 class SettingsValidationError(Exception):
     """Raised when a settings payload has unknown keys or invalid values.
 
@@ -728,6 +713,49 @@ class SettingsValidationError(Exception):
         super().__init__(
             f"settings validation failed: unknown={unknown_keys} errors={field_errors}"
         )
+
+
+def _normalize_keys(values: dict[str, Any]) -> dict[str, Any]:
+    """Rewrite any HEADROOM_* env-name key to its canonical registry key.
+
+    Keys already in canonical form, or genuinely unknown keys, pass through
+    unchanged so callers can still surface an ``unknown_keys`` error for them.
+
+    Raises :class:`SettingsValidationError` when both the canonical key and its
+    env-name alias appear in the same payload with different values (e.g.
+    ``rpm=10`` and ``HEADROOM_RPM=20`` simultaneously).  Same-value duplicates
+    are silently deduplicated to the canonical key.
+    """
+    out: dict[str, Any] = {}
+    for key, value in values.items():
+        if key not in _BY_KEY and key in _BY_ENV:
+            canonical = _BY_ENV[key].key
+            if canonical in out and out[canonical] != value:
+                raise SettingsValidationError(
+                    [],
+                    {
+                        canonical: (
+                            f"conflicting values for {canonical!r} via canonical key"
+                            f" and env alias {key!r}: {out[canonical]!r} vs {value!r}"
+                        )
+                    },
+                )
+            out[canonical] = value
+        else:
+            if key in _BY_KEY:
+                env_alias = _BY_KEY[key].env
+                if env_alias in values and values[env_alias] != value:
+                    raise SettingsValidationError(
+                        [],
+                        {
+                            key: (
+                                f"conflicting values for {key!r} via canonical key"
+                                f" and env alias {env_alias!r}: {value!r} vs {values[env_alias]!r}"
+                            )
+                        },
+                    )
+            out[key] = value
+    return out
 
 
 def _coerce(field: SettingField, value: Any) -> Any:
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The alias support is useful, but the normalization currently makes conflicting input order-dependent. For example, `{"HEADROOM_RPM": 10, "rpm": 20}` produces 20, while reversing those two keys produces 10, because both normalize to `rpm` and silently overwrite each other. JSON object order should not decide a persisted enterprise setting.

Please define and test a deterministic collision policy. Rejecting a request that supplies both names with different values is the safest option (same-value duplicates could either be accepted or rejected consistently). The policy should cover both `validate()` and `save()`, including `None`/clear and masked-secret cases.

gingeekrishna:
Thanks for the catch — agreed that silent last-write-wins is the wrong behaviour for a persisted enterprise setting.

**What changed** (commit `0589c5d5`):

- `SettingsValidationError` moved above `_normalize_keys` so the function can raise it directly.
- `_normalize_keys` now detects when the same canonical key is supplied via both names:
  - **different values** → `SettingsValidationError` with a `field_errors` entry naming the key and both conflicting values, regardless of which form appears first in the dict
  - **same value** → silently deduplicates to the canonical key (intent is unambiguous)
- Both orderings (`{"rpm": 20, "HEADROOM_RPM": 10}` and `{"HEADROOM_RPM": 10, "rpm": 20}`) are tested.
- Also fixed the two Copilot-flagged save/load tests: added `monkeypatch.delenv("HEADROOM_SETTINGS_PATH", raising=False)` so they're hermetic on machines that set that env var.

16/16 tests pass locally.

JerrettDavis (approved):
The new normalization policy deterministically rejects differing canonical/env-alias values in either input order and deduplicates identical values. The added validation tests cover the reported collision, and the save tests now isolate the settings path. Current CI is green.

JerrettDavis:
This duplicates #2833 for #2812. #2833 already implements the same deterministic alias normalization/collision policy in the established settings-store test suite, includes the endpoint registry-fixture update needed when `_BY_ENV` is introduced, has full focused validation, and is approved on the current head. This PR has therefore been superseded and can be closed in favor of #2833.

PR #2869, opened 2026-08-08
fix(memory): sync FTS5 and vector indexes on CLI delete/edit/prune/purge

### LINKED ISSUE #2856: [BUG] CLI memory delete/edit bypass HierarchicalMemory — orphaned vector and FTS index entries keep deleted memories in memory_search results
## Description

`headroom memory delete` and `headroom memory edit` operate on the raw `SQLiteMemoryStore` and never touch the search indexes. Deleted memories remain in both the sqlite-vec vector index (`memory_vectors.db`) and the FTS5 text index (`memory_fts` in `memory.db`), so the memory MCP server keeps returning them from `memory_search` — even after a full server restart, because startup only re-indexes memories missing embeddings and never removes orphans. Edited memories keep serving their pre-edit content from both indexes.

Root cause (v0.34.0):

- `cli/memory.py:64` — `get_store()` returns a bare `SQLiteMemoryStore`, not a `HierarchicalMemory`.
- `cli/memory.py:777` — `delete` calls `store.delete_batch(resolved_ids)` (`memory/adapters/sqlite.py:387`), which only deletes the primary rows.
- `cli/memory.py:629` — `edit` calls `store.save(mem)`, which only rewrites the primary row.
- Also affected: `prune` (`cli/memory.py:899`, same `store.delete_batch`) and `purge` (`cli/memory.py:954`, `store.clear_all()` — `memory/adapters/sqlite.py:846`, a bare `DELETE FROM memories`).
- The index removal/re-add logic lives exclusively in `HierarchicalMemory`, which the CLI never instantiates: `memory/core.py:643` `delete` calls `_vector_index.remove` / `_text_index.remove` at lines 657–658; `memory/core.py:447` `update` re-indexes changed content at lines 498–501; `memory/core.py:510` `supersede` removes the old entry from both indexes at lines 572–573.

Line numbers are from the installed 0.34.0 release and may have shifted on `main`.

## To Reproduce

1. Save a memory via the memory MCP server (`memory_save`) so it gets embedded and FTS-indexed.
2. Run `headroom memory delete <id> --force` — prints `Success: Deleted 1 memory(ies).`
3. Restart every `headroom.memory.mcp_server` process.
4. Call `memory_search` with a query matching the deleted memory.
5. The deleted memory is returned. Row counts confirm the orphans: after deleting 16 of 46 memories, `memories` had 30 rows while `vec_metadata` and `memory_fts` both still had 46.

## Expected Behavior

CLI delete/edit keep the primary store, vector index, and FTS index consistent — the same contract as `HierarchicalMemory.delete()`/`.update()`. Deleted memories must not be retrievable via `memory_search` after the server reloads.

## Actual Behavior

Only the primary `memories` table is updated. Orphaned index entries persist indefinitely; deleted content stays searchable and edited content is served stale.

## Code Sample

```bash
headroom memory delete <id> --force
# then, from any MCP client after restarting the memory server:
# memory_search("<query matching the deleted memory>")  -> deleted memory still returned
```

Verifying the orphans directly:

```sql
-- memory.db
SELECT COUNT(*) FROM memories;    -- 30
SELECT COUNT(*) FROM memory_fts;  -- 46
-- memory_vectors.db
SELECT COUNT(*) FROM vec_metadata;  -- 46
```

## Error Output

```
(no error — delete reports "Success: Deleted 1 memory(ies)." while the indexes silently retain the entries)
```

## Environment

- **Headroom version**: 0.34.0 (`headroom-ai` from PyPI)
- **Python version**: 3.14.6
- **LLM Provider**: Anthropic (Claude Code) + Codex CLI sharing a single store at `~/.headroom/memory.db`

## Additional Context

- Related but distinct from #2296: that issue covers stale vector metadata on metadata-only updates through `TrafficLearner`/`HierarchicalMemory.update`; this one is the CLI bypassing index maintenance entirely for delete/edit.
- Manual cleanup that worked: delete orphan rowid pairs from `vec_embeddings` + `vec_metadata` (sqlite-vec extension loaded, matching `vec_metadata.memory_id` against `memories.id`), delete orphan `memory_fts` rows by `memory_id`, then restart the memory MCP servers. FTS `integrity-check` passes afterwards.
- Suggested fix direction: route CLI delete/edit through `HierarchicalMemory`, or add an index-reconciliation pass at server startup (remove index entries whose `memory_id` no longer exists).

gingeekrishna:
## Problem

`headroom memory delete`, `prune`, `edit`, and `purge` all operate on the bare `SQLiteMemoryStore` — they update the primary `memories` table but never touch the FTS5 full-text index (`memory_fts` in `memory.db`) or the vector index (`vec_metadata` / `vec_embeddings` in `memory_vectors.db`). The index maintenance path lives in `HierarchicalMemory.delete()` / `.update()`, which the CLI never instantiates.

**Symptoms (from #2856):**
```sql
-- After deleting 16 of 46 memories via CLI:
SELECT COUNT(*) FROM memories;    -- 30
SELECT COUNT(*) FROM memory_fts;  -- 46  ← orphans
-- memory_vectors.db
SELECT COUNT(*) FROM vec_metadata;  -- 46  ← orphans
```
Deleted memories keep surfacing in `memory_search` results even after a full server restart, because server startup only re-embeds memories whose `embedding IS NULL` — it never removes orphaned index entries.

Fixes #2856.

## Solution

Add two best-effort helpers to `headroom/cli/memory.py` that use **direct SQLite** (no `sqlite-vec` extension, no embedder):

- **`_remove_from_search_indexes(db_path, memory_ids)`**: removes specific IDs from `memory_fts` and from `vec_metadata` / `vec_embeddings`. Skips silently if an index doesn't exist.
- **`_clear_all_search_indexes(db_path)`**: truncates both indexes completely (for purge).

Wire these up in four commands:
| Command | Change |
|---|---|
| `delete` | `_remove_from_search_indexes` after `store.delete_batch()` |
| `prune` | `_remove_from_search_indexes` after `store.delete_batch()` |
| `purge` | `_clear_all_search_indexes` after `store.clear_all()` |
| `edit` | If content changed: remove stale entries, clear `embedding` (server re-embeds on next startup), re-add FTS5 entry with new content immediately |

The edit path re-adds the FTS5 entry right away so keyword search reflects the new content without requiring a server restart. Vector search is deferred to the next startup re-embed cycle (same as what the server already does for missing embeddings).

## Changes

- `headroom/cli/memory.py` — two new helpers; four command call sites
- `tests/test_cli_memory_index_sync.py` (new) — 9 unit tests covering both helpers with FTS5 and a stub vector DB. No `sqlite-vec` or embedder required; tests run locally.

## Testing

```
$ python -m pytest tests/test_cli_memory_index_sync.py -v
...
9 passed in 2.38s
```

### DIFF
--- a/headroom/cli/memory.py
+++ b/headroom/cli/memory.py
@@ -30,6 +30,8 @@ from ._utils.formatting import (
 from ._utils.parsers import parse_duration
 from .main import main
 
+_REINDEX_PAGE_SIZE = 1_000
+
 
 def _default_db_path() -> str:
     """Resolve the memory DB the proxy/install actually use.
@@ -66,6 +68,15 @@ def get_store(db_path: str) -> SQLiteMemoryStore:
     return SQLiteMemoryStore(db_path)
 
 
+def _sqlite_table_exists(conn: Any, table_name: str) -> bool:
+    """Return whether a SQLite table or virtual table has been initialized."""
+    row = conn.execute(
+        "SELECT 1 FROM sqlite_master WHERE type = 'table' AND name = ?",
+        (table_name,),
+    ).fetchone()
+    return row is not None
+
+
 def _remove_from_search_indexes(db_path: str, memory_ids: list[str]) -> bool:
     """Remove specific memories from FTS5 and vector search indexes.
 
@@ -89,14 +100,15 @@ def _remove_from_search_indexes(db_path: str, memory_ids: list[str]) -> bool:
     # FTS5 table lives in the same memory.db file (built-in, no extension needed).
     try:
         with sqlite3.connect(str(db)) as conn:
-            for i in range(0, len(memory_ids), 500):
-                chunk = memory_ids[i : i + 500]
-                placeholders = ",".join("?" * len(chunk))
-                conn.execute(
-                    f"DELETE FROM memory_fts WHERE memory_id IN ({placeholders})",
-                    chunk,
-                )
-            conn.commit()
+            if _sqlite_table_exists(conn, "memory_fts"):
+                for i in range(0, len(memory_ids), 500):
+                    chunk = memory_ids[i : i + 500]
+                    placeholders = ",".join("?" * len(chunk))
+                    conn.execute(
+                        f"DELETE FROM memory_fts WHERE memory_id IN ({placeholders})",
+                        chunk,
+                    )
+                conn.commit()
     except Exception as exc:
         print_warning(f"FTS5 index cleanup incomplete: {exc}")
         ok = False
@@ -109,16 +121,21 @@ def _remove_from_search_indexes(db_path: str, memory_ids: list[str]) -> bool:
         return ok
 
     try:
-        import sqlite_vec
-    except ImportError:
-        print_warning(
-            "sqlite-vec is not installed; stale vector index entries may remain. "
-            "Run 'headroom memory reindex' after installing sqlite-vec to repair."
-        )
-        return False
-
-    try:
         with sqlite3.connect(str(vector_db)) as conn:
+            # A sibling database may exist before the optional vector index has
+            # ever been initialized. That is a valid no-op, not a sync failure.
+            if not _sqlite_table_exists(conn, "vec_metadata"):
+                return ok
+
+            try:
+                import sqlite_vec
+            except ImportError:
+                print_warning(
+                    "sqlite-vec is not installed; stale vector index entries may remain. "
+                    "Run 'headroom memory reindex' after installing sqlite-vec to repair."
+                )
+                return False
+
             conn.enable_load_extension(True)
             sqlite_vec.load(conn)
             conn.enable_load_extension(False)
@@ -155,8 +172,9 @@ def _clear_all_search_indexes(db_path: str) -> bool:
 
     try:
         with sqlite3.connect(str(db)) as conn:
-            conn.execute("DELETE FROM memory_fts")
-            conn.commit()
+            if _sqlite_table_exists(conn, "memory_fts"):
+                conn.execute("DELETE FROM memory_fts")
+                conn.commit()
     except Exception as exc:
         print_warning(f"FTS5 index cleanup incomplete: {exc}")
         ok = False
@@ -166,16 +184,19 @@ def _clear_all_search_indexes(db_path: str) -> bool:
         return ok
 
     try:
-        import sqlite_vec
-    except ImportError:
-        print_warning(
-            "sqlite-vec is not installed; stale vector index entries may remain. "
-            "Run 'headroom memory reindex' after installing sqlite-vec to repair."
-        )
-        return False
-
-    try:
         with sqlite3.connect(str(vector_db)) as conn:
+            if not _sqlite_table_exists(conn, "vec_metadata"):
+                return ok
+
+            try:
+                import sqlite_vec
+            except ImportError:
+                print_warning(
+                    "sqlite-vec is not installed; stale vector index entries may remain. "
+                    "Run 'headroom memory reindex' after installing sqlite-vec to repair."
+                )
+                return False
+
             conn.enable_load_extension(True)
             sqlite_vec.load(conn)
             conn.enable_load_extension(False)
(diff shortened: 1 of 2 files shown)


JerrettDavis (dismissed):
Thanks for tackling the stale-index path. There is one production-specific gap that needs to be resolved before this is safe:

The vector cleanup is tested with a normal `vec_embeddings` table, but production creates it as a `vec0` virtual table and explicitly loads `sqlite_vec` on every connection in `SQLiteVectorIndex._get_conn()`. These new helpers open a bare `sqlite3` connection and then execute against `vec_embeddings`; with the real schema that operation cannot reliably run without loading the extension. Because the broad `except Exception: pass` swallows the failure (and the metadata delete is after the virtual-table delete), the CLI can report success while leaving the stale vector rows this PR is intended to remove.

Please exercise the real `SQLiteVectorIndex` schema in the regression test and use its extension-aware connection/deletion seam (or equivalently load sqlite-vec here). Also avoid silently converting a failed synchronization into success: the CLI should surface an actionable failure or use a design that makes the primary/index update recoverably consistent.

gingeekrishna:
Good catch — the review is correct on both counts.

**Root cause**: `vec_embeddings` is a `vec0` virtual table. A bare `sqlite3.connect()` without loading the `sqlite-vec` extension has no `vec0` module registered, so any DML on that table raises. The original `except Exception: pass` swallowed the failure, and the CLI reported success while leaving stale rows.

**What changed** (commit `3b291e11`):

**`_remove_from_search_indexes` and `_clear_all_search_indexes`**:
- Now `import sqlite_vec` explicitly before opening the vector DB, mirroring `SQLiteVectorIndex._create_conn()` exactly:
  ```python
  conn.enable_load_extension(True)
  sqlite_vec.load(conn)
  conn.enable_load_extension(False)
  ```
- If `sqlite_vec` is not installed, `print_warning()` is emitted with an actionable message ("install with: pip install sqlite-vec") instead of silently skipping.
- Other exceptions during vector cleanup are now surfaced as warnings too — no more silent success-with-stale-entries.
- FTS5 cleanup is unchanged (FTS5 is built-in, no extension needed).

**Tests**:
- `_seed_vector()` now uses `SQLiteVectorIndex` directly to create the real `vec0` schema, so the extension-aware connection path is actually exercised rather than tested against a plain table stand-in.
- All vector index tests are gated with `@requires_sqlite_vec` (skipped locally where sqlite-vec isn't installed; run in CI where it is).
- Added `test_remove_from_search_indexes_no_vector_rows_to_delete_is_noop`, `test_delete_command_removes_from_vector_index`, and `test_purge_command_clears_vector_index` using the real schema.
- FTS5 tests remain unconditional.

Locally: 7 passed, 5 skipped (sqlite_vec not installed in this dev env — the 5 skipped are the vec0 tests that will run in CI).

JerrettDavis (dismissed):
Loading sqlite-vec and exercising the real `vec0` schema resolves the production-schema gap, but failed synchronization is still converted into a successful CLI operation. Delete/prune/purge mutate the primary store first and then only `print_warning` if FTS/vector cleanup fails, leaving stale searchable rows; edit also still wraps FTS re-indexing in bare `except Exception: pass`. Please make index failure affect command success or introduce a durable/recoverable rebuild path so the CLI cannot report a fully successful mutation while indexes remain inconsistent, and add failure-path coverage.

gingeekrishna:
Thanks — the second review is correct. `print_warning` without changing exit code means the command silently reports success even when indexes are inconsistent.

**What changed** (commit `f27831d6`):

**`_remove_from_search_indexes` / `_clear_all_search_indexes`** now return `bool` (True = fully synced, False = any partial failure). Callers decide what to do with the result; the helpers no longer hide failures by returning None.

**`delete` / `prune` / `purge`**: check the return value.
- `True` → `print_success(...)` and exit 0 as before
- `False` → `print_warning("... Run 'headroom memory reindex' to repair.")` then `sys.exit(1)`

The primary store mutation is reported accurately in both branches — the message names the correct count regardless.

**`edit`**: the bare `except Exception: pass` on FTS re-indexing is gone. Both the `_remove_from_search_indexes` result and the re-index result are tracked; if either fails, `print_warning` + `sys.exit(1)` after the store `save` (so the primary update is never lost, only the index is inconsistent).

**`headroom memory reindex`** (new command): wipes and rebuilds FTS5 from the primary store; removes orphaned `vec_metadata`/`vec_embeddings` entries when sqlite-vec is installed. All failure messages in helpers and commands now point users here. Exits 1 if any step fails.

**Tests** — 4 new failure-path tests:
- `test_remove_from_search_indexes_fts_failure_returns_false`: patches `sqlite3.connect` (global, since it's imported locally inside the helper) to fail on the first call → asserts False
- `test_remove_from_search_indexes_sqlite_vec_missing_returns_false`: `patch.dict(sys.modules, {"sqlite_vec": None})` → asserts False
- Same two patterns for `_clear_all_search_indexes`

Locally: 11 passed, 5 skipped (the 5 sqlite_vec-dependent vec0 schema tests run in CI).

JerrettDavis (dismissed):
The new repair command can silently destroy valid index coverage on larger stores. `reindex_memories()` loads the primary store with `MemoryFilter(limit=100_000)`, then unconditionally deletes the entire FTS table and rebuilds only that capped result set. Every primary memory after the first 100,000 is therefore left absent from FTS. The same truncated `primary_ids` set is used to identify vector orphans, so valid vector rows for memories beyond the cap are deleted as if their primary records no longer existed.

Please iterate/page through the complete primary store (or use an uncapped store API) before clearing/rebuilding or deciding which vectors are orphaned, and add a regression test with a deliberately small page size/cap proving records beyond the first page remain indexed.

chopratejas (approved):
Latest-head maintainer review complete. Reindex no longer has a destructive record cap, orphan detection uses the complete primary ID set, absent optional indexes are valid no-ops, real vec0 access remains extension-aware, and actual delete, edit, prune, and purge commands surface partial synchronization with nonzero exits. Merge after full CI is green.

PR #2872, opened 2026-08-09
fix(proxy/openai): run response hooks on Responses, and bill their re-drives

chopratejas:
The Responses path runs `run_request_hooks` but never `run_response_hooks` —
only `handle_openai_chat` does. So a turn hook can shrink a Responses turn and
then never be asked to resolve what the model did about it: the model's injected
tool call goes straight to a client that has no such tool.

That asymmetry is why tool-belt deferral has to be disabled wholesale on the
Responses API, which is the surface Codex uses.

## 1. Wire the response side

Mirrors the chat-completions block. **Buffered path only**, for the same reason
CCR already forces `stream:false` when it needs to intercept: you cannot
re-drive a turn whose bytes are already flowing.

## 2. Honour `stream_safe_only` on the Responses request path

It was the one hook call site that ignored the flag. A re-driving hook would run
its shrink on a streamed turn and then have no response side to finish it —
latent until (1) lands, live afterwards.

`stream` is not a parameter of `_compress_openai_responses_payload`, but the
payload it is compressing carries the flag. It is read **before** CCR may force
`stream:false` further down, so this is the client's request rather than the
effective one — conservative in the safe direction: at worst a CCR-buffered turn
misses a saving, never a stranded tool call.

Fold-only hooks that declare `stream_safe = True` are unaffected.

## 3. Bill what the re-drives cost

Both handlers read usage from the **final** upstream response, so every
intermediate call a hook made was free as far as Headroom was concerned.

For a token-saving feature that is not a rounding error. A tool-search reload is
a whole extra model call; counting only the last one lets the feature hide its
own overhead behind the saving it is claiming, and the numbers come out better
than the truth.

`TurnHookUsage` accumulates input/output/cached across re-drives; both HTTP
paths fold it into their totals. The two surfaces report the same three
quantities under different names (`prompt_tokens` vs `input_tokens`), so the key
pair is passed in.

Expect measured cost to go **up** and savings percentage to go **down** on any
deployment running a re-driving hook. That is the correction, not a regression.

## Also: restore the body after the hooks

A re-drive rewrites `body[input]` / `body[messages]` / `body[tools]` so the next
upstream call carries the hook's turn. Everything downstream — CCR's
`_responses_input_to_items(body["input"])`, usage accounting, observability — is
describing the request the *client* made, not the proxy's internal detour.

Without the restore, a turn that both reloaded a tool and hit CCR retrieval
hands CCR the proxy's synthetic items. The chat path had the same leak
(`body["messages"]` stayed rewritten); both are fixed the same way.

## Known gap

A re-drive on the custom backend path (`send_openai_message`) is still not
folded into that request's accounting — its usage is recorded elsewhere.
Commented at the call site rather than silently skipped.

## Blast radius

**Inert unless a turn hook is registered**, so no behaviour change for a stock
OSS proxy. `TurnHookUsage` starts at zero and stays there on every path that
does not re-drive.

## Verification

- `tests/test_turn_hook_usage.py` — 5 new tests: per-surface key names,
  accumulation across rounds, negative counts floored not subtracted, and that
  an unreadable shape still counts the call (a silent zero there looks exactly
  like "the hook cost nothing")
- 434 passing across `turn_hook`, `extension`, `tool_search`, `responses` and
  `openai_chat` suites
- `ruff check` + `ruff format` clean

### DIFF
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -187,6 +187,55 @@ def _header_get(headers: dict[str, str], name: str) -> str | None:
     return None
 
 
+class TurnHookUsage:
+    """Tokens spent by a turn hook's extra upstream calls.
+
+    A hook that re-drives the model (``call_model``) makes real, billed requests
+    that the handler's own accounting never sees: usage is read from the FINAL
+    response, so a turn that re-drove twice was charged for three calls and
+    reported as one. That is not a rounding error for a token-saving feature —
+    it undercounts the exact cost the saving has to beat.
+
+    Both OpenAI surfaces report the same three quantities under different names,
+    so :meth:`add` takes the key pair and the handlers just add the totals in.
+    """
+
+    __slots__ = ("input_tokens", "output_tokens", "cache_read_tokens", "calls")
+
+    def __init__(self) -> None:
+        self.input_tokens = 0
+        self.output_tokens = 0
+        self.cache_read_tokens = 0
+        self.calls = 0
+
+    def add(
+        self,
+        payload: Any,
+        *,
+        input_key: str,
+        output_key: str,
+        details_key: str,
+    ) -> None:
+        """Fold one re-drive response's usage in. Never raises: a hook must not
+        be able to 500 a request by returning an odd shape."""
+        self.calls += 1
+        usage = payload.get("usage") if isinstance(payload, dict) else None
+        if not isinstance(usage, dict):
+            return
+
+        def _int(value: Any) -> int:
+            try:
+                return max(int(value), 0)
+            except (TypeError, ValueError):
+                return 0
+
+        self.input_tokens += _int(usage.get(input_key))
+        self.output_tokens += _int(usage.get(output_key))
+        details = usage.get(details_key)
+        if isinstance(details, dict):
+            self.cache_read_tokens += _int(details.get("cached_tokens"))
+
+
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The response-hook wiring and body restoration are valuable, but the new usage accounting currently charges the wrong call set whenever a hook re-drives:

`_hook_call_model` / `_resp_hook_call_model` adds every re-drive response to `TurnHookUsage`. If the hook returns that re-drive as its final response, the handler replaces `response` with it; the normal usage block then reads that final response, and afterward the code adds `TurnHookUsage` again. For one re-drive, this counts `re-drive + re-drive` while omitting the original upstream response. For multiple rounds it counts `B + C + final C` and still omits original A. The desired billed total is `A + B (+ C ...)`.

Please preserve/account the original response usage before response replacement and ensure each upstream call contributes exactly once (an alternative is to have the accumulator represent only responses that the normal final-response accounting will not read). Add handler-level regression tests with deliberately different usage values for original and re-drive responses on both Chat Completions and Responses; the current `TurnHookUsage` unit tests cannot detect this integration-level double-count/omission.

JerrettDavis (approved):
The revised accumulator records the original and re-drive responses, subtracts exactly the response handled by normal accounting, and adds only the remaining billed calls. The new Chat Completions and Responses handler tests use distinct usage values and directly cover the prior double-count/omission. CI is green.

PR #2873, opened 2026-08-09
harden: the deployment script interpolates the $bucket ... in query.sh

anupamme:
## Description

Adds input validation for the \`R2_BUCKET\` environment variable in
\`deploy/beacon/query.sh\`, an operator-run local shell script (not a web
service or request handler — it is never on a network path). The script is
invoked manually or from a cron job by someone who already holds Cloudflare R2
credentials on their workstation.

## Type of Change

- [x] Bug fix / security hardening
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Vulnerability

| Field | Value |
|-------|-------|
| ID | V-001 |
| Severity | HIGH |
| Scanner | multi_agent_ai |
| File | \`deploy/beacon/query.sh\` |

Description: Without this guard, a shell special character in \`R2_BUCKET\`
(e.g. from a misconfigured env file or an untrusted CI/CD variable) would
reach the DuckDB path string \`r2://\${BUCKET}/sessions//.json\` unvalidated.

## Threat model

\`query.sh\` is an operator/admin script — it is not a CGI handler and is not
reachable over a network. The threat scenario is: a workstation env file that
has \`R2_BUCKET\` set to a value containing shell-special characters (copied from
a non-R2 source), or a CI/CD pipeline that interpolates \`R2_BUCKET\` from an
untrusted variable. The fix is defensive hardening appropriate to that context.

## Changes Made

- Add an `R2_BUCKET` allowlist guard before environment loading, credential checks, and DuckDB invocation.
- Add a dependency-free executable Bash regression covering invalid and valid bucket names.

\`deploy/beacon/query.sh\` — one-line regex guard at line 23, running before
env-file sourcing, credential checks, and DuckDB invocation:

\`\`\`bash
[[ "\$BUCKET" =~ ^[a-zA-Z0-9_-]+\$ ]] || {
  echo "Invalid R2_BUCKET value — must contain only alphanumeric, hyphen, or underscore characters" >&2
  exit 1
}
\`\`\`

## Testing

- [x] Regression tests pass
- [x] New executable test added

### Test Output

```text
PASS: 7 adversarial R2_BUCKET values rejected
PASS: 5 representative R2_BUCKET values accepted
All 12 tests passed.
```

\`deploy/beacon/test_query_bucket_validation.sh\` — pure bash, no external
dependencies, no DuckDB or Cloudflare credentials needed. Run:

\`\`\`bash
bash deploy/beacon/test_query_bucket_validation.sh
\`\`\`

Rejected values (exit 1 + \`"Invalid R2_BUCKET"\` in stderr):

| Value | Threat |
|---|---|
| \`../etc/passwd\` | path traversal |
| \`bucket; rm -rf /\` | shell injection |
| \`bucket\$(whoami)\` | command substitution |
| \`bucket with spaces\` | word-splitting |
| \`bucket!@#\` | special characters |
| \`bucket.name\` | dot (outside grammar) |
| \`bucket/subpath\` | slash (outside grammar) |

Accepted values (pass validation, then fail at credential check — proving
the guard was cleared without real credentials):

| Value | Notes |
|---|---|
| \`headroom-telemetry\` | production default |
| \`headroom_backup\` | underscore variant |
| \`my-bucket-123\` | alphanumeric + hyphens |
| \`MyBucket\` | mixed case |
| \`a\` | single character |

Accepted cases use a temporary empty file for \`HEADROOM_ENV_FILE\`. The script
clears bucket validation and then fails with \`"R2_ACCOUNT_ID not set"\` — a
distinct error that confirms validation was passed.

## Behavior preservation

Valid bucket names (alphanumeric, hyphens, underscores) are unaffected. The
guard rejects only values that contain characters outside that grammar.

## Real Behavior Proof

- Environment: pure Bash test on the current PR head; no DuckDB or Cloudflare credentials required.
- Exact command / steps: ran bash deploy/beacon/test_query_bucket_validation.sh, which invokes query.sh with seven adversarial and five representative valid R2_BUCKET values.
- Observed result: all 12 cases passed; invalid values were rejected before credential setup and valid values reached the expected missing-credential check.
- Not tested: live DuckDB/R2 access with production credentials.

## Review Readiness

### DIFF
--- a/deploy/beacon/query.sh
+++ b/deploy/beacon/query.sh
@@ -20,6 +20,7 @@
 set -euo pipefail
 
 BUCKET="${R2_BUCKET:-headroom-telemetry}"
+[[ "$BUCKET" =~ ^[a-zA-Z0-9_-]+$ ]] || { echo "Invalid R2_BUCKET value — must contain only alphanumeric, hyphen, or underscore characters" >&2; exit 1; }
 _repo_env="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)/.env"
 ENV_FILE="${HEADROOM_ENV_FILE:-$HOME/env.txt}"
 [ -f "$ENV_FILE" ] || ENV_FILE="$_repo_env"
--- a/deploy/beacon/test_query_bucket_validation.sh
+++ b/deploy/beacon/test_query_bucket_validation.sh
@@ -1,0 +1,109 @@
+#!/usr/bin/env bash
+# Regression test for R2_BUCKET validation in query.sh (commit 381fdb9d).
+#
+# Run from any directory:
+#   bash deploy/beacon/test_query_bucket_validation.sh
+#
+# No external dependencies — pure bash, no DuckDB or Cloudflare credentials
+# required.  Accepted-value cases use a temporary empty env file so the script
+# reaches the credential check ("R2_ACCOUNT_ID not set"), which is the positive
+# signal that bucket validation was passed.
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+QUERY_SH="${SCRIPT_DIR}/query.sh"
+
+FAKE_ENV="$(mktemp)"
+trap 'rm -f "$FAKE_ENV"' EXIT
+
+PASS=0
+FAIL=0
+ERRORS=()
+
+# assert_rejected LABEL VALUE
+# Asserts: exit code 1 AND "Invalid R2_BUCKET" in stderr.
+assert_rejected() {
+  local label="$1"
+  local bucket_value="$2"
+  local stderr_out exit_code=0
+
+  stderr_out="$(R2_BUCKET="$bucket_value" HEADROOM_ENV_FILE="$FAKE_ENV" \
+    bash "$QUERY_SH" 2>&1 >/dev/null)" || exit_code=$?
+
+  local ok=1
+  if [[ $exit_code -ne 1 ]]; then
+    ok=0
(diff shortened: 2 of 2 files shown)


JerrettDavis (changes requested):
The allowlist is a reasonable hardening direction, but the claimed regression coverage is not part of this patch, and the test shown in the PR body sets `BUCKET` even though the script reads `R2_BUCKET`; it therefore would not exercise this change.

Please add an executable regression test that invokes the script with adversarial `R2_BUCKET` values and asserts rejection occurs before DuckDB/credential setup, plus representative accepted values. That will also pin the intended bucket-name grammar. Please update the PR description to reflect the actual test and avoid describing this operator script as a remotely exposed web request handler.

anupamme:
Review comments addressed. Pls review.

JerrettDavis (approved):
Approved on exact head `bb483a8e2cde6771d27ad8d34234c3748d2306dc`. This directly addresses the prior review: the executable harness invokes `query.sh` with `R2_BUCKET` (not the unused `BUCKET` variable), proves seven adversarial values are rejected before credential/DuckDB setup, and proves five representative values clear the guard and reach the distinct credential error. I ran all 12 cases locally plus `bash -n` on both scripts. The PR description now accurately frames this as defensive hardening for an operator-run script.

JerrettDavis (approved):
Re-reviewed the current head (). The requested executable regression is now part of the patch, uses  (the variable the script actually reads), and proves seven adversarial values are rejected before credential/DuckDB setup while five representative valid names clear the guard. I ran the script locally: all 12 cases passed. The PR body now matches the actual operator-script threat model and test. The requested changes are satisfied.

Issue #2805, opened 2026-08-05
[BUG] Server-side tool_search deferral poisons Claude Code transcripts: prompt-type Stop hook evaluator 400s with "Tool reference 'AskUserQuestion' not found in available tools"


myagizmaktav:
## Summary

When `HEADROOM_TOOL_SEARCH` is active, `inject_tool_search_deferral()` adds a `tool_search_tool_*` tool + `defer_loading` to **Claude Code** traffic. Anthropic then answers with `server_tool_use` / `tool_search_tool_result` blocks whose payload is a list of `tool_reference` entries, and Claude Code writes those blocks into its session transcript **permanently** — while still believing tool search is off (`ENABLE_TOOL_SEARCH=false`).

That is fine for the main loop, because the proxy re-injects the deferred tools array on every main-loop request. It breaks the moment Claude Code makes a **side-request that carries a different tools array**. The one I hit reliably is the *prompt-type Stop hook* evaluator (the hook flavour `/loop` installs): the proxy passes that request through untouched (`transforms=none`), so the transcript references `AskUserQuestion` but the request's `tools` array does not contain it:

```
Ran 2 stop hooks
  ⎿ Stop hook error: Hook evaluator API error:
    API Error: 400 Tool reference 'AskUserQuestion' not found in available tools
```

This is the same failure class as #2766 / #2440 (`headroom_retrieve`), but for the tool-search tool, and triggered by a client side-request rather than a frozen CCR marker.

## Environment

- headroom `0.32.0` (pip `headroom-ai`), Python 3.12, Linux
- Proxy: `python -m headroom.cli proxy --host 127.0.0.1 --port 8787 --mode cache --backend anthropic --telemetry --no-http2`
- Deployment: `install` profile, preset `persistent-task`, runtime `python`, supervisor `none`
- Upstream: first-party `https://api.anthropic.com` (no Bedrock/Vertex)
- Clients: Claude Code `2.1.221` and `2.1.222` via `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`, `ENABLE_TOOL_SEARCH=false` in `settings.json`
- ~500 tool definitions in the request (several MCP servers), model `claude-opus-5`
- Also checked against current `main`: the gate (`handlers/anthropic.py` ~L2420) and `inject_tool_search_deferral()` are unchanged in the relevant way, and `tool_reference` appears nowhere in the anthropic handler — so this should still reproduce on 0.34.0.

## Mechanism

[...]

Issue #2766, opened 2026-08-04
400 "Tool reference 'headroom_retrieve' not found in available tools" when a frozen turn replays an existing marker


crb8:
### Summary

On a frozen-prefix turn that replays an **existing** `<<ccr:hash>>` marker, the proxy does not inject the `headroom_retrieve` tool, so the agent has a marker it cannot redeem. When it tries, the Anthropic API rejects the whole request:

```
API Error: 400 Tool reference 'headroom_retrieve' not found in available tools
```

This is user-visible in Claude Code and was the most frequent failure in my logs.

### Environment

- headroom-ai 0.33.0, `--mode token --backend anthropic`, persistent-task deployment
- Claude Code 2.1.221, Windows 11, Python 3.14
- Long sessions (180 msgs avg, 734 max), compression very active

### Root cause

Two gates have to pass for the tool to be injected, and both count **only markers that are new this turn**:

1. `proxy/handlers/anthropic.py`, the `should_inject_ccr_tool(...)` call passes `has_compressed_content=has_new_compressed_content`.
2. `proxy/helpers.py`, `apply_session_sticky_ccr_tool(...)` receives `has_compressed_content_this_turn=has_new_compressed_content`, and for a session the tracker considers fresh it returns early:

```python
# Fresh session — only inject when this turn produced compressed content.
if not has_compressed_content_this_turn:
    ... decision="skip" ...
    return tools_out, False
```

The #1850 change narrowed the #1006 override to new markers only, to protect the tools cache segment. But a replayed marker is exactly as unredeemable as a fresh one: the agent redeems hashes it was handed turns ago, which is the point of a marker. Anything that instructs the model to retrieve later (a project instruction saying "call `headroom_retrieve` with the hash before asserting an exact value") lands on this path by construction.

The `SessionCcrTracker` is per-process, so **every proxy restart makes all live sessions look fresh again**, re-arming the bug mid-conversation. My deployment restarts often, which is why this reproduced constantly.

### Evidence

`CCR: deferring tool injection` across three rotated logs: 1626 + 1863 + 1842 occurrences. `decision=skip` appears steadily alongside them.

### Suggested fix

Count markers present in the request, not just markers created this turn, at both gates:

```python
has_compressed_content=(has_new_compressed_content or injector.has_compressed_content)
```

[...]

PR #2533, opened 2026-07-24
fix(ccr): re-inject headroom_retrieve when history references it on the sessionless path (#2440)

abhay-codes07:
## Description

Fixes #2440. `apply_session_sticky_ccr_tool` bypasses the `SessionCcrTracker` when `session_id` is `None` (WS / pre-session paths) and drives injection purely off the per-turn `has_compressed_content_this_turn` flag:

```python
if not session_id:
    if not has_compressed_content_this_turn:
        ...  # skip: tool NOT re-declared
        return tools_out, False
    ...
```

If an earlier turn emitted a `headroom_retrieve` tool_use into history but the current turn produced no fresh compression marker, the tool definition is not re-declared in `tools`, while the forwarded history still references it. The provider then rejects the whole request:

```
API Error: 400 Tool reference 'headroom_retrieve' not found in available tools.
```

Without a session the tracker can't remember the earlier turn's CCR, so this is unique to the sessionless path.

## Fix

Add `history_references_ccr_tool(messages)` which detects an existing `headroom_retrieve` call in the forwarded messages — both the Anthropic assistant `tool_use` content block and the OpenAI assistant `tool_calls[].function.name` shapes, fully null-guarded. On the sessionless path, injection now fires when `has_compressed_content_this_turn` **or** history already references the tool, so the definition is re-declared and the request validates. The decision is logged as a new `inject_history_reference` outcome. Both handlers pass the signal computed from `optimized_messages` (the bytes actually forwarded). Behavior with a real `session_id` (the sticky tracker path) is unchanged.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `headroom/proxy/helpers.py`: add `history_references_ccr_tool`; add a `history_has_ccr_reference` parameter to `apply_session_sticky_ccr_tool` and OR it into the sessionless injection decision.
- `headroom/proxy/tool_injection_logging.py`: add the `inject_history_reference` decision literal.
- `headroom/proxy/handlers/anthropic.py`, `headroom/proxy/handlers/openai.py`: pass `history_references_ccr_tool(optimized_messages)` into the sticky-tool call.
[...]

PR #2876, opened 2026-08-09
fix(proxy/anthropic): repair headroom_retrieve history references the tools array cannot support

### LINKED ISSUE #2814: [BUG] Passthrough side-requests (prompt-type Stop hook evaluator, /compact) can 400 on `headroom_retrieve` history references — injection-side fixes (#2440/#2533) may not cover this class; mirror #2807's history repair for CCR
### Summary

#2805 / #2807 established the mechanism: Claude Code replays **one** transcript across requests that carry **different** `tools` arrays, and Anthropic validates every history reference against the array of the request at hand. #2807 fixed that for tool-search blocks by *repairing history* (`strip_unsupported_tool_search_blocks`) rather than trying to predict the client's tool set.

The same mechanism applies to CCR's `headroom_retrieve`, but every fix proposed for it so far (#2766, #2440, PR #2533) is **injection-side** — they decide *when to re-declare the tool*. That covers turns the proxy transforms; it does not obviously cover the class where the proxy forwards a request **untouched**.

### Evidence (my logs, 0.32.0, `--mode token --backend anthropic`, preset `persistent-task`)

The prompt-type Stop hook evaluator's request reaches the proxy and is forwarded verbatim:

```
[hr_1785948639_009559] ... transforms=none body_mutated=false source=passthrough
[hr_1785948639_009559] Forwarding upstream streaming error status=400
```

…while the main-loop request in the same second (carrying `router:tool_search_deferral`) returns 200. 61 such 400s across 11 days of logs, all on hook/loop triggers.

**Honest limit of my evidence:** in my case the rejected reference was the *tool_search* variant (`Tool reference 'AskUserQuestion' not found`), i.e. #2805. I run `HEADROOM_LOSSLESS=1`, which disables the CCR marker and the tool injection entirely, so the CCR variant cannot occur on my box — `headroom_retrieve` appears 0 times in my current logs. I am reporting the *class*, not a fresh CCR repro: the transcript-vs-tools-array mismatch is tool-agnostic, a history `tool_use` naming `headroom_retrieve` is validated the same way, and hook/`/compact` side-requests are precisely where the proxy cannot know which tools the client will send.

### Why the injection-side fixes may not close it (`main` today)

- `apply_session_sticky_ccr_tool` (`headroom/proxy/helpers.py:1715`) still gates on `has_compressed_content_this_turn` (`:1772`, `:1837`).
- PR #2533 (open, for #2440) adds a `history_references_ccr_tool` check — the right direction. It would cover a passthrough request **if** the CCR path actually runs on `source=passthrough` turns and adding a tool to an otherwise-untouched request is acceptable. Worth stating explicitly either way, because that is the whole question for this class.
- #2807's repair only recognizes `tool_search_tool_result` / `server_tool_use` blocks (`helpers.py:2409`). A `tool_use` / `tool_result` pair naming `headroom_retrieve` is not covered.

### Proposal

Add the symmetric repair next to #2807's: when the outbound `tools` array does not declare `headroom_retrieve`, drop (or neutralize) the history blocks that reference it — the `tool_use` and its paired `tool_result` — instead of depending on re-injection succeeding for every request shape. Same properties as the existing helper: deterministic (prefix cache intact), stateless, and self-healing for transcripts poisoned before the fix. Sequencing matters the same way too: run it *after* injection so main-loop turns strip nothing.

This is belt-and-braces with #2533, not a replacement — #2533 keeps the tool available where it belongs, the repair makes the 400 structurally impossible where it cannot.

### Environment

headroom 0.32.0 (code paths above checked against `main` today), Linux, Python 3.12, Claude Code CLI 2.1.221/2.1.222, single profile `init-user`, port 8787.

abhay-codes07:
## Description

#2805 / #2807 established the mechanism: Claude Code replays one transcript across requests that carry different `tools` arrays, and Anthropic validates every history reference against the array of the request at hand. #2807 fixed it for tool-search blocks by repairing history (`strip_unsupported_tool_search_blocks`) rather than trying to predict the client's tool set.

The same mechanism applies to CCR's `headroom_retrieve`, and it is tool-agnostic. A passthrough side-request (the prompt-type Stop hook evaluator, `/compact`) that the proxy forwards without declaring `headroom_retrieve` still carries a historical `tool_use` naming it, and Anthropic 400s on the dangling reference. The injection-side fixes (#2766 / #2533) decide *when to re-declare the tool*; this makes the 400 *structurally impossible* where the tool is intentionally absent. It is belt-and-braces with them, not a replacement.

The fix adds the symmetric repair next to #2807's. When the outbound `tools` array does not declare `headroom_retrieve`, it replaces each `headroom_retrieve` `tool_use` and its paired `tool_result` with a text block, so no dangling reference survives.

It **neutralizes** (replaces in place) rather than **drops**, which is the one deliberate difference from #2807: CCR's `tool_use` lives in an assistant turn and its `tool_result` in the next user turn, i.e. two different messages. Dropping a whole message could leave two same-role messages adjacent and break Anthropic's strict user/assistant alternation, turning one 400 into another. Replacing blocks in place keeps every message and role intact, and preserves the retrieved text the model already saw. #2807's server-tool blocks both live in the same assistant turn, so dropping was safe there.

It runs after CCR tool injection, so on the main loop -- where the tool IS injected (a present marker) -- it neutralizes nothing and the prompt-cache prefix is untouched, mirroring #2807's placement and sequencing.

Fixes #2814

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/helpers.py`: added `strip_unsupported_ccr_retrieve_blocks(messages, tools)` (and a small `_ccr_result_as_text` helper). No-ops (returning the original object by identity) when `headroom_retrieve` is declared or no such history exists; otherwise neutralizes the `tool_use` and its paired `tool_result` to text.
- `headroom/proxy/handlers/anthropic.py`: call the repair right after the tool-search history repair (which is after CCR tool injection), guarded on it actually changing anything, tagged `router:ccr_retrieve_repair:Nblocks`.
- `tests/test_ccr_retrieve_history_repair.py`: 5 unit tests (no-op when declared, no-op without retrieve history, neutralize + preserve result text + keep alternation, leave foreign tool_use untouched, placeholder when the result has no text).

## Testing

### Test Output

```text
tests/test_ccr_retrieve_history_repair.py  5 passed

# Broader CCR / tool-search / handler suites (unchanged behavior):
tests/test_ccr_retrieve_history_repair.py tests/test_proxy/test_anthropic_ccr_deferred_injection.py
tests/test_issue_746_tool_search.py                                      71 passed

# uvx mypy@1.20.2 headroom/proxy/helpers.py -> Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: confirmed the injection point (`apply_session_sticky_ccr_tool`) and the tool-search repair placement in `handlers/anthropic.py`, confirmed `body["tools"]` reflects the CCR injection before the repair call site (`body["tools"] = tools` is written well upstream and the adjacent tool-search repair already relies on it), then drove the helper over a transcript with a `headroom_retrieve` tool_use + paired tool_result: with the tool declared it returns the original object unchanged; with the tool absent it neutralizes both blocks, preserves the result text, and keeps the message roles/count identical.
- Observed result: a forwarded request that would 400 with "Tool reference 'headroom_retrieve' not found in available tools" now carries text blocks in place of the retrieve `tool_use`/`tool_result`, so there is no reference for Anthropic to reject, and user/assistant alternation is preserved. The main loop (tool present) is a no-op.
- Not tested: a live multi-turn Claude Code session hitting a Stop-hook/`/compact` side-request against a real provider (no live provider here). The repair is a pure function verified directly over the exact block shapes Anthropic validates, and it mirrors the already-merged tool-search repair's mechanism and wiring.

## Review Readiness

## Checklist

## Additional Notes

The issue reporter noted their own logs show the tool-search variant of this 400 (61 across 11 days) but zero `headroom_retrieve` occurrences, because they run `HEADROOM_LOSSLESS=1` which disables CCR entirely. This PR fixes the CCR variant of the same, proven, tool-agnostic mechanism rather than a fresh CCR repro. The neutralize-vs-drop choice is the one place I departed from #2807, for the alternation reason above; if you would rather it drop (accepting the alternation handling that implies), I am happy to switch it.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -2737,6 +2737,28 @@ class AnthropicHandlerMixin:
                     _ts_stripped,
                 )
 
+            # CCR retrieve history repair (#2814). Run beside the tool-search
+            # repair and after both normal CCR injection and turn hooks, so it
+            # validates against the final outbound tools array. A side-request
+            # that does not declare headroom_retrieve cannot retain historical
+            # tool_use/tool_result references that Anthropic would reject.
+            from headroom.proxy.helpers import strip_unsupported_ccr_retrieve_blocks
+
+            _ccr_repaired, _ccr_neutralized = strip_unsupported_ccr_retrieve_blocks(
+                body.get("messages"), body.get("tools")
+            )
+            if _ccr_neutralized:
+                body["messages"] = _ccr_repaired
+                optimized_messages = _ccr_repaired
+                body_mutation_tracker.mark_mutated("ccr_retrieve_history_repair")
+                transforms_applied.append(f"router:ccr_retrieve_repair:{_ccr_neutralized}blocks")
+                logger.info(
+                    "[%s] CCR: neutralized %d headroom_retrieve history block(s) "
+                    "(tools array does not declare the tool this turn)",
+                    request_id,
+                    _ccr_neutralized,
+                )
+
             # Consistency: report tok_before/tok_after with ONE tokenizer. The pipeline
             # and the handler use different token estimators, and cache-mode branches
             # can leave original_tokens (handler, line ~1049) and optimized_tokens
--- a/headroom/proxy/helpers.py
+++ b/headroom/proxy/helpers.py
@@ -3027,6 +3027,126 @@ def strip_unsupported_tool_search_blocks(messages: Any, tools: Any) -> tuple[Any
     return (out, removed) if changed else (messages, 0)
 
 
+def _ccr_result_as_text(block: dict[str, Any]) -> str:
+    """Flatten a ``tool_result`` block's content to plain text, preserving what
+    the model already saw. Falls back to a short placeholder when there is no
+    textual content to keep."""
+    content = block.get("content")
+    if isinstance(content, str) and content.strip():
+        return content
+    if isinstance(content, list):
+        parts = [
+            b.get("text", "")
+            for b in content
+            if isinstance(b, dict) and b.get("type") == "text" and b.get("text")
+        ]
+        joined = "\n".join(part for part in parts if part)
+        if joined.strip():
+            return joined
+    return "[headroom_retrieve result omitted]"
+
+
+def strip_unsupported_ccr_retrieve_blocks(messages: Any, tools: Any) -> tuple[Any, int]:
+    """Neutralize ``headroom_retrieve`` history references the outbound ``tools``
+    array cannot support.
+
+    Claude Code replays one transcript across requests that carry different
+    ``tools`` arrays, and Anthropic validates every history ``tool_use`` against
+    the array of the request at hand. A passthrough side-request (the prompt-type
+    Stop hook evaluator, ``/compact``) that the proxy forwards without declaring
+    ``headroom_retrieve`` then 400s on a historical ``tool_use`` that names it --
+    the CCR sibling of the tool-search history repair (#2814 / #2807). This is
+    belt-and-braces with the injection-side fixes: they keep the tool available
+    where it belongs; this makes the 400 structurally impossible where it cannot.
+
+    When the request does NOT declare ``headroom_retrieve``, replace each
+    ``headroom_retrieve`` ``tool_use`` block and its paired ``tool_result`` with a
+    text block, so no dangling reference survives. Neutralize rather than drop:
+    CCR's ``tool_use`` (an assistant turn) and its ``tool_result`` (the next user
+    turn) live in DIFFERENT messages, so removing a message could leave two
+    same-role messages adjacent and break Anthropic's user/assistant alternation.
+    Replacing blocks in place keeps every message and role intact, and preserves
+    the retrieved text the model already saw.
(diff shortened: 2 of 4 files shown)


JerrettDavis (approved):
The repair is gated on the tool being absent, preserves message roles and existing text, neutralizes both the matching tool use and result, leaves foreign tools untouched, and is wired after normal CCR injection. Current CI is green.

JerrettDavis:
Maintainer reconciliation: rebased this PR onto current main, preserving the MCP v1 cap and the later #2888 ordering rule. The CCR repair now runs after turn hooks beside the final tool-search repair, so it validates against the actual outbound tools array. Added handler-level regressions for both cases: a hook removes headroom_retrieve and the pair is neutralized; an inert hook keeps it and structured history survives. Focused CCR, tool-search, handler, and inherited MCP suites: 76 passed; Ruff and formatting clean.

PR #2879, opened 2026-08-09
fix(proxy): tune macOS libmalloc and trim allocator pages so long-lived RSS stays bounded

### LINKED ISSUE #2820: [BUG] macOS: long-lived proxy RSS ratchets to ~11 GB under large concurrent request bodies — libmalloc retains freed pages; verified fix + patch included
## Description

On macOS, a long-lived `headroom proxy` serving coding-agent traffic (Claude Code) ratchets RSS upward until it destabilizes the host. Observed in production: two proxies reached **~11 GB RSS each** (lifetime peaks ~12.8 GB) within ~8 hours on a 24 GB Mac mini; the machine hit a jetsam "System is unhealthy" event (compressor at 17.1 GB) and had to be restarted. This happened twice within three days before diagnosis.

This is **allocator-level memory bloat, not a Python reference leak** — `leaks` measured only ~16 MB of true leaks at ~600 MB RSS, and tracemalloc-visible Python growth was ~10% of RSS growth. The evidence below isolates the mechanism, falsifies two plausible-but-wrong hypotheses, and verifies a fix A/B. A tested patch is included at the bottom (applies cleanly to 0.33.0 and 0.34.0).

## Root cause

The proxy's request path allocates a large transient working set per in-flight request: each 0.5–1 MB Anthropic messages body is JSON-parsed into ~10–20 MB of Python objects, deep-copied (the `original_client_messages` snapshot), re-serialized, and tokenized. Under concurrency these bursts drive **macOS libmalloc** to a high-water mark it never returns:

- freed large allocations sit in libmalloc's "death row" cache (`MallocLargeCache`), and
- freed pages are not madvised back to the OS,

so `vmmap` shows whole **`MALLOC_LARGE (empty)` / `MALLOC_SMALL (empty)` regions held resident indefinitely**. Real agent conversations grow to 600+ messages and retries re-send them, so every biggest-yet burst permanently raises the floor — hence the multi-day ratchet into double-digit GB on long-lived proxies.

## To Reproduce

1. Isolated `HOME`, mock upstream via `--anthropic-api-url` (no network dependency), `headroom proxy` with default flags.
2. Drive 5 sessions of Claude-Code-shaped bodies (~600–740 KB each) at 4-way concurrency, 30 requests total.
3. Watch RSS and `vmmap --summary <pid>`.

Result: the first concurrent burst permanently commits ~220 MB (~44 MB per concurrent request); afterwards ~0.6 MB/request of further creep; **nothing is released after going idle**. `--memory` was bisected out — identical growth without it (it only adds a ~180 MB constant baseline for lazily-loaded models).

## Expected Behavior

RSS returns toward baseline after bursts; a proxy serving one agent session stays in the hundreds of MB indefinitely.

## Actual Behavior

RSS only ratchets upward; `vmmap` accumulates resident-but-empty MALLOC regions; long-lived proxies reach ~11 GB and trigger jetsam.

## Hypotheses falsified on the way (so you don't re-chase them)

1. **Request-logger retention** — plausible from source, but `log_full_messages` defaults to False and the deque held 8 KB for 74 entries on the live proxy.
2. **`malloc_zone_pressure_relief()` alone as the fix** — a periodic trim task by itself did *not* purge the empty regions on macOS 26.6 (identical RSS to stock through 7+ trim ticks). The launch-time knobs below are what actually work; the trim task is kept as cheap defense in depth.

## The fix, verified A/B

Two libmalloc environment knobs, which must be set **before `main()`** (libmalloc reads them pre-main, so the proxy re-execs itself once to apply them):

- `MallocAggressiveMadvise=1` — madvise freed pages back to the OS eagerly
- `MallocLargeCache=0` — no death-row cache for freed large allocations

plus a periodic `gc.collect()` + `malloc_zone_pressure_relief(NULL, 0)` trim task (glibc `malloc_trim(0)` on Linux) as defense in depth.

Same load harness for all runs (macOS 26.6, CPython 3.13.14):

| Run | Config | Settled RSS | Dirty pages | Empty MALLOC regions |
|---|---|---|---|---|
| R1 | stock, `--memory` | 570.3 MB | 516 MB | 51.6 MB LARGE + 17.4 MB SMALL |
| R2 | stock, no `--memory` | 401.0 MB | 353 MB | (same pattern, smaller baseline) |
| R6 | trim task only | 572.1 MB | 518 MB | unchanged → falsified |
| R7 | env knobs by hand | 507.0 MB | 395.8 MB | **0 dirty** |
| R8 | patch end-to-end (self re-exec) | **507.0 MB** | **395.1 MB** | **0 dirty** |

Startup RSS also dropped 335 → 280 MB. Post-burst creep decelerates (+7.7, +5.2, +1.0, +0.6, +0.6 MB per 5 requests) consistent with bounded LRU caches filling, not a leak. Under a real agent session, the patched proxy shows **no `MALLOC_LARGE (empty)` regions at all** and 176 KB (vs 17–28 MB stock) of empty-small dirty pages. The knobs convert never-returned high water into continuously-purged pages, so the multi-day ceiling becomes bounded caches + concurrency working set instead of ~11 GB.

The trim task's `gc.collect()` holds the GIL for the duration of a collection, so its pause lands between requests just as any other collection would; the madvise cost of the knobs lands on free paths. (I did not benchmark throughput explicitly — no latency change was apparent in the harness or in production use, but treat that as anecdotal.)

## Patch

Four small, self-contained changes (diff against 0.34.0 below; applies cleanly to 0.33.0 too — same anchors):

1. `headroom/cli/proxy.py` — `_reexec_with_malloc_tuning()`, called first thing in the `proxy` command: on darwin, unless `HEADROOM_MALLOC_TUNING=0` or the operator already set the vars, set the two knobs and `execv` once (`_HEADROOM_MALLOC_TUNED=1` guards against loops; the PID is preserved so supervisor/pid-file setups keep working; `--help`/`--version` never reach it).
2. `headroom/proxy/malloc_trim.py` (new) — `trim()` and `trim_periodically()`.
3. `headroom/proxy/server.py` — lifespan starts the trim task per worker (allocator state is per-process), plus env plumbing in the standalone config path.
4. `headroom/proxy/models.py` — `periodic_malloc_trim_enabled: bool = True`, `malloc_trim_interval_seconds: int = 60` on `ProxyConfig`.

New env surface (all opt-out): `HEADROOM_MALLOC_TUNING` (default on, darwin only), `HEADROOM_MALLOC_TRIM` (default on), `HEADROOM_MALLOC_TRIM_INTERVAL_SECONDS` (default 60).

Happy to open this as a PR if you'd take it — contributed under the project's Apache-2.0 license. Suggestions welcome on naming/placement; the re-exec is the only part with any subtlety (env must be set pre-main, and in-process setting is impossible for libmalloc).

<details>
<summary><b>Full diff (apply with <code>patch -p1</code> from <code>site-packages</code> or the repo's python package root)</b></summary>

```diff
diff -ruN a/headroom/cli/proxy.py b/headroom/cli/proxy.py
--- a/headroom/cli/proxy.py	2006-07-23 20:21:28
+++ b/headroom/cli/proxy.py	2026-08-05 19:38:46
@@ -80,6 +80,32 @@
     return _get_env_bool(name, False)
 
 
+# libmalloc reads these before main() runs, so they cannot be set from inside
+# the current process — the proxy re-execs itself once to apply them. Without
+# them, freed pages from large concurrent request bodies stay resident
+# (``vmmap`` shows whole "MALLOC_LARGE (empty)" regions) and long-lived proxy
+# RSS only ratchets upward. Vars the operator already set are left untouched;
+# HEADROOM_MALLOC_TUNING=0 disables the re-exec entirely.
+_MALLOC_TUNING = {
+    "MallocAggressiveMadvise": "1",  # madvise freed pages back to the OS eagerly
+    "MallocLargeCache": "0",  # no death-row cache for freed large allocations
+}
+
+
+def _reexec_with_malloc_tuning() -> None:
+    if sys.platform != "darwin":
+        return
+    if not _get_env_bool("HEADROOM_MALLOC_TUNING", True):
+        return
+    if os.environ.get("_HEADROOM_MALLOC_TUNED") == "1":
+        return
+    missing = {k: v for k, v in _MALLOC_TUNING.items() if k not in os.environ}
+    os.environ["_HEADROOM_MALLOC_TUNED"] = "1"
+    if not missing:
+        return
+    os.environ.update(missing)
+    os.execv(sys.executable, [sys.executable, "-m", "headroom.cli", *sys.argv[1:]])
+
 def _get_env_int_optional(name: str) -> int | None:
     val = os.environ.get(name)
     if val is None or val == "":
@@ -1017,6 +1043,8 @@
     Usage with OpenAI-compatible clients:
         OPENAI_BASE_URL=http://localhost:8787/v1 your-app
     """
+    _reexec_with_malloc_tuning()
+
     # Import here to avoid slow startup
     try:
         from headroom.proxy.server import (
@@ -1186,6 +1214,8 @@
         rate_limit_requests_per_minute=rpm if rpm is not None else 60,
         rate_limit_tokens_per_minute=tpm if tpm is not None else 100_000,
         compress_user_messages=_get_env_bool("HEADROOM_COMPRESS_USER_MESSAGES", False),
+        periodic_malloc_trim_enabled=_get_env_bool("HEADROOM_MALLOC_TRIM", True),
+        malloc_trim_interval_seconds=_get_env_int("HEADROOM_MALLOC_TRIM_INTERVAL_SECONDS", 60),
         min_tokens_to_crush=_get_env_int("HEADROOM_MIN_TOKENS", 500),
         max_items_after_crush=_get_env_int("HEADROOM_MAX_ITEMS", 50),
         exclude_tools=_parse_exclude_tools(None) or None,
diff -ruN a/headroom/proxy/malloc_trim.py b/headroom/proxy/malloc_trim.py
--- a/headroom/proxy/malloc_trim.py	1969-12-31 18:00:00
+++ b/headroom/proxy/malloc_trim.py	2026-08-05 19:38:46
@@ -0,0 +1,96 @@
+"""Return freed-but-retained allocator pages to the OS on long-lived proxies.
+
+Large concurrent Anthropic bodies (0.5-1 MB of JSON parsed, deep-copied and
+re-serialized per in-flight request) drive libmalloc and pymalloc to a
+high-water mark that is never returned to the OS: after a burst the malloc
+zones keep entire regions resident but empty (``vmmap`` lists them as
+``MALLOC_LARGE (empty)`` / ``MALLOC_SMALL (empty)``), so process RSS only
+ratchets upward. Over a multi-day proxy lifetime under Claude Code traffic
+this reaches double-digit GB and starves the host.
+
+Neither runtime returns these pages on its own. macOS exposes
+``malloc_zone_pressure_relief(NULL, 0)`` to purge every zone's free pages;
+glibc has ``malloc_trim(0)``. ``trim()`` runs a full ``gc.collect()`` first so
+obmalloc can unmap empty arenas and freed CPython blocks reach the allocator
+free lists before the purge.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import ctypes
+import gc
+import logging
+import sys
+import time
+
+logger = logging.getLogger(__name__)
+
+# Lazily resolved (platform_tag, foreign_function | None). ``None`` function
+# means the platform has no supported trim call and trim() is a no-op.
+_relief: tuple[str, object | None] | None = None
+
+
+def _resolve() -> tuple[str, object | None]:
+    global _relief
+    if _relief is not None:
+        return _relief
+    try:
+        libc = ctypes.CDLL(None)
+        if sys.platform == "darwin":
+            fn = libc.malloc_zone_pressure_relief
+            fn.argtypes = [ctypes.c_void_p, ctypes.c_size_t]
+            fn.restype = ctypes.c_size_t
+            _relief = ("darwin", fn)
+        else:
+            fn = libc.malloc_trim
+            fn.argtypes = [ctypes.c_size_t]
+            fn.restype = ctypes.c_int
+            _relief = ("glibc", fn)
+    except (OSError, AttributeError):
+        _relief = ("unsupported", None)
+    return _relief
+
+
+def trim() -> int:
+    """Collect garbage and return allocator free pages to the OS.
+
+    Returns the number of bytes freed on macOS (glibc's ``malloc_trim``
+    reports only success, so 0 is returned there and on unsupported
+    platforms).
+    """
+    kind, fn = _resolve()
+    if fn is None:
+        return 0
+    gc.collect()
+    if kind == "darwin":
+        return int(fn(None, 0))  # type: ignore[operator]
+    fn(0)  # type: ignore[operator]
+    return 0
+
+
+async def trim_periodically(interval_seconds: int = 60) -> None:
+    """Background task that periodically returns allocator free pages to the OS.
+
+    Runs in every worker process (allocator state is per-process). The
+    ``gc.collect()`` inside ``trim()`` holds the GIL for the duration of the
+    collection, so the pause lands between requests just as any other
+    collection would.
+
+    Args:
+        interval_seconds: How often to trim (default: 60 seconds).
+    """
+    while True:
+        await asyncio.sleep(interval_seconds)
+        try:
+            start = time.perf_counter()
+            freed = trim()
+            elapsed_ms = (time.perf_counter() - start) * 1000
+            log = logger.info if freed >= (16 << 20) else logger.debug
+            log(
+                "MallocTrim: returned %.1f MB to OS in %.0f ms",
+                freed / 1048576,
+                elapsed_ms,
+            )
+        except Exception as e:
+            logger.debug("MallocTrim failed: %s", e)
diff -ruN a/headroom/proxy/models.py b/headroom/proxy/models.py
--- a/headroom/proxy/models.py	2006-07-23 20:21:28
+++ b/headroom/proxy/models.py	2026-08-05 19:38:46
@@ -425,6 +425,14 @@
     # Env: HEADROOM_PERIODIC_TOIN_STATS=0.
     periodic_toin_stats_enabled: bool = True
 
+    # Periodic allocator trim. Long-lived proxies processing large concurrent
+    # request bodies ratchet RSS through freed-but-retained allocator pages;
+    # this returns them to the OS (malloc_zone_pressure_relief on macOS,
+    # malloc_trim on glibc). Envs: HEADROOM_MALLOC_TRIM=0,
+    # HEADROOM_MALLOC_TRIM_INTERVAL_SECONDS.
+    periodic_malloc_trim_enabled: bool = True
+    malloc_trim_interval_seconds: int = 60
+
     # Stateless mode — disable all filesystem writes for read-only / container deployments
     stateless: bool = False
 
diff -ruN a/headroom/proxy/server.py b/headroom/proxy/server.py
--- a/headroom/proxy/server.py	2006-07-23 20:21:28
+++ b/headroom/proxy/server.py	2026-08-05 19:38:46
@@ -144,6 +144,7 @@
 )
 from headroom.proxy.loop_callback_failure_policy import is_known_websocket_callback_failure
 from headroom.proxy.loopback_guard import is_loopback_host
+from headroom.proxy.malloc_trim import trim_periodically
 from headroom.proxy.memory_handler import MemoryConfig, MemoryHandler
 
 # Data models (extracted to headroom/proxy/models.py for maintainability)
@@ -2529,6 +2530,12 @@
                 await proxy.startup()
                 if config.periodic_toin_stats_enabled:
                     asyncio.create_task(_log_toin_stats_periodically())
+                # Per-worker on purpose: allocator state is per-process, so
+                # every worker must trim its own zones (no beacon-owner gate).
+                if config.periodic_malloc_trim_enabled:
+                    asyncio.create_task(
+                        trim_periodically(config.malloc_trim_interval_seconds)
+                    )
                 if proxy.usage_reporter:
                     await proxy.usage_reporter.start(proxy)
                 if proxy.traffic_learner:
@@ -4929,6 +4936,8 @@
         http2=_get_env_bool("HEADROOM_HTTP2", True),
         http_proxy=os.environ.get("HEADROOM_HTTP_PROXY") or None,
         periodic_toin_stats_enabled=_get_env_bool("HEADROOM_PERIODIC_TOIN_STATS", True),
+        periodic_malloc_trim_enabled=_get_env_bool("HEADROOM_MALLOC_TRIM", True),
+        malloc_trim_interval_seconds=_get_env_int("HEADROOM_MALLOC_TRIM_INTERVAL_SECONDS", 60),
         proxy_token=os.environ.get("HEADROOM_PROXY_TOKEN") or None,
         offline=_get_env_bool("HEADROOM_OFFLINE", False),
         # Default mode is CACHE (Headroom's coding posture): delta-only compression
```

</details>

## Environment

- headroom-ai 0.33.0 and 0.34.0 (both affected; no fix found in the 0.34.0 tree)
- macOS 26.6, Apple Silicon, 24 GB RAM
- CPython 3.13.14 (uv-managed tool install)
- Traffic: Claude Code via the Anthropic `/v1/messages` route (production) and Anthropic-shaped bodies against a mock upstream (harness). The mechanism is request-body size + concurrency, so other providers' large-body routes are plausibly affected too, but I only measured the Anthropic path

abhay-codes07:
## Summary

Fixes #2820.

Prevents long-lived macOS proxies from retaining every largest transient request-body allocation in libmalloc. The reporter’s production A/B isolated the allocator behavior and verified the two pre-main libmalloc knobs; this PR applies them through a one-time Darwin-only re-exec and adds periodic per-worker pressure relief.

- `MallocAggressiveMadvise=1` returns freed pages eagerly.
- `MallocLargeCache=0` disables the large-allocation death-row cache.
- Operator-set allocator variables are preserved; `HEADROOM_MALLOC_TUNING=0` is the kill switch.
- Periodic trim defaults on only for macOS, runs off the event loop, performs no forced Python GC, validates its interval, and is retained/cancelled through the app lifecycle.
- Non-Darwin behavior remains unchanged unless explicitly enabled.
- Semantically rebased onto current `main`, retaining startup dependency validation, MCP SDK v1 compatibility, and all newer proxy behavior.

## Verification

- 147 proxy CLI/config/malloc/MCP-contract tests pass; 1 platform skip.
- Ruff check and formatting clean; `git diff --check` clean.
- The reporter’s macOS A/B reduced dirty empty malloc regions to zero and lowered steady/startup RSS; the control flow and shutdown lifecycle are covered locally.

## Safety

The re-exec is Darwin-only, PID-preserving, loop-guarded, and opt-out. The trim task is per worker because allocator state is per process, and shutdown cancels it explicitly.

### DIFF
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -113,6 +113,35 @@ def _get_env_bool_optional(name: str) -> bool | None:
     return _get_env_bool(name, False)
 
 
+# libmalloc reads these before main() runs, so they cannot be set from inside
+# the current process — the proxy re-execs itself once to apply them. Without
+# them, freed pages from large concurrent request bodies stay resident
+# (``vmmap`` shows whole "MALLOC_LARGE (empty)" regions) and long-lived proxy
+# RSS only ratchets upward (#2820). Vars the operator already set are left
+# untouched; HEADROOM_MALLOC_TUNING=0 disables the re-exec entirely.
+_MALLOC_TUNING = {
+    "MallocAggressiveMadvise": "1",  # madvise freed pages back to the OS eagerly
+    "MallocLargeCache": "0",  # no death-row cache for freed large allocations
+}
+
+
+def _reexec_with_malloc_tuning() -> None:
+    if sys.platform != "darwin":
+        return
+    if not _get_env_bool("HEADROOM_MALLOC_TUNING", True):
+        return
+    if os.environ.get("_HEADROOM_MALLOC_TUNED") == "1":
+        return
+    missing = {k: v for k, v in _MALLOC_TUNING.items() if k not in os.environ}
+    # Set the loop guard before the re-exec so the replacement process (which
+    # inherits this environment) skips this path instead of re-execing forever.
+    os.environ["_HEADROOM_MALLOC_TUNED"] = "1"
+    if not missing:
+        return
+    os.environ.update(missing)
+    os.execv(sys.executable, [sys.executable, "-m", "headroom.cli", *sys.argv[1:]])
+
+
 def _get_env_int_optional(name: str) -> int | None:
     val = os.environ.get(name)
     if val is None or val == "":
@@ -1065,6 +1094,7 @@ def proxy(
     Usage with OpenAI-compatible clients:
         OPENAI_BASE_URL=http://localhost:8787/v1 your-app
     """
+    _reexec_with_malloc_tuning()
     ensure_proxy_dependencies()
 
     # Import here to avoid slow startup
@@ -1261,6 +1291,8 @@ def proxy(
         rate_limit_requests_per_minute=rpm if rpm is not None else 60,
         rate_limit_tokens_per_minute=tpm if tpm is not None else 100_000,
         compress_user_messages=_get_env_bool("HEADROOM_COMPRESS_USER_MESSAGES", False),
+        periodic_malloc_trim_enabled=_get_env_bool("HEADROOM_MALLOC_TRIM", True),
+        malloc_trim_interval_seconds=_get_env_int("HEADROOM_MALLOC_TRIM_INTERVAL_SECONDS", 60),
         min_tokens_to_crush=_get_env_int("HEADROOM_MIN_TOKENS", 500),
         max_items_after_crush=_get_env_int("HEADROOM_MAX_ITEMS", 50),
         exclude_tools=_parse_exclude_tools(None) or None,
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -1291,7 +1291,9 @@ def proxy(
         rate_limit_requests_per_minute=rpm if rpm is not None else 60,
         rate_limit_tokens_per_minute=tpm if tpm is not None else 100_000,
         compress_user_messages=_get_env_bool("HEADROOM_COMPRESS_USER_MESSAGES", False),
-        periodic_malloc_trim_enabled=_get_env_bool("HEADROOM_MALLOC_TRIM", True),
+        periodic_malloc_trim_enabled=_get_env_bool(
+            "HEADROOM_MALLOC_TRIM", sys.platform == "darwin"
+        ),
         malloc_trim_interval_seconds=_get_env_int("HEADROOM_MALLOC_TRIM_INTERVAL_SECONDS", 60),
         min_tokens_to_crush=_get_env_int("HEADROOM_MIN_TOKENS", 500),
         max_items_after_crush=_get_env_int("HEADROOM_MAX_ITEMS", 50),
--- a/headroom/proxy/malloc_trim.py
+++ b/headroom/proxy/malloc_trim.py
@@ -26,6 +26,12 @@ import time
 
 logger = logging.getLogger(__name__)
 
+# Interval bounds for the periodic trim task. A non-positive interval would make
+# ``asyncio.sleep`` return immediately and spin a continuous collect/trim loop,
+# so anything below the minimum falls back to the default.
+_DEFAULT_TRIM_INTERVAL_SECONDS = 60
+_MIN_TRIM_INTERVAL_SECONDS = 1
+
 # Lazily resolved (platform_tag, foreign_function | None). ``None`` function
 # means the platform has no supported trim call and trim() is a no-op.
 _relief: tuple[str, object | None] | None = None
@@ -72,19 +78,40 @@ def trim() -> int:
 async def trim_periodically(interval_seconds: int = 60) -> None:
     """Background task that periodically returns allocator free pages to the OS.
 
-    Runs in every worker process (allocator state is per-process). The
-    ``gc.collect()`` inside ``trim()`` holds the GIL for the duration of the
-    collection, so the pause lands between requests just as any other
-    collection would.
+    Runs in every worker process (allocator state is per-process). The blocking
+    trim (``gc.collect()`` plus ``malloc_zone_pressure_relief``/``malloc_trim``,
+    which can be slow on a large heap) runs via ``asyncio.to_thread`` so it never
+    blocks the event loop, and the task exits immediately on platforms with no
+    supported trim call, so it is a true no-op there.
 
     Args:
-        interval_seconds: How often to trim (default: 60 seconds).
+        interval_seconds: How often to trim (default: 60 seconds). A value below
+            ``_MIN_TRIM_INTERVAL_SECONDS`` (which would busy-loop) falls back to
+            the default.
     """
+    _, fn = _resolve()
+    if fn is None:
+        # No supported allocator-trim call on this platform (Windows, musl, ...);
+        # do not spin a wakeup task that can only ever no-op.
+        logger.debug("MallocTrim: no supported trim on %s; task disabled", sys.platform)
+        return
+
+    if interval_seconds < _MIN_TRIM_INTERVAL_SECONDS:
+        logger.warning(
+            "MallocTrim: interval %ss is below the %ds minimum; using default %ds",
+            interval_seconds,
+            _MIN_TRIM_INTERVAL_SECONDS,
+            _DEFAULT_TRIM_INTERVAL_SECONDS,
+        )
+        interval_seconds = _DEFAULT_TRIM_INTERVAL_SECONDS
+
     while True:
         await asyncio.sleep(interval_seconds)
         try:
             start = time.perf_counter()
-            freed = trim()
+            # Off the event-loop thread: the gc + C-level purge can pause for a
+            # while on a large heap, and that pause must not stall proxy traffic.
+            freed = await asyncio.to_thread(trim)
             elapsed_ms = (time.perf_counter() - start) * 1000
             log = logger.info if freed >= (16 << 20) else logger.debug
             log(
--- a/headroom/proxy/malloc_trim.py
+++ b/headroom/proxy/malloc_trim.py
@@ -1,0 +1,96 @@
+"""Return freed-but-retained allocator pages to the OS on long-lived proxies.
+
+Large concurrent Anthropic bodies (0.5-1 MB of JSON parsed, deep-copied and
+re-serialized per in-flight request) drive libmalloc and pymalloc to a
+high-water mark that is never returned to the OS: after a burst the malloc
+zones keep entire regions resident but empty (``vmmap`` lists them as
+``MALLOC_LARGE (empty)`` / ``MALLOC_SMALL (empty)``), so process RSS only
+ratchets upward. Over a multi-day proxy lifetime under Claude Code traffic
+this reaches double-digit GB and starves the host.
+
+Neither runtime returns these pages on its own. macOS exposes
+``malloc_zone_pressure_relief(NULL, 0)`` to purge every zone's free pages;
+glibc has ``malloc_trim(0)``. ``trim()`` runs a full ``gc.collect()`` first so
+obmalloc can unmap empty arenas and freed CPython blocks reach the allocator
+free lists before the purge.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import ctypes
+import gc
+import logging
+import sys
+import time
+
+logger = logging.getLogger(__name__)
+
+# Lazily resolved (platform_tag, foreign_function | None). ``None`` function
+# means the platform has no supported trim call and trim() is a no-op.
+_relief: tuple[str, object | None] | None = None
+
+
+def _resolve() -> tuple[str, object | None]:
+    global _relief
+    if _relief is not None:
+        return _relief
+    try:
+        libc = ctypes.CDLL(None)
+        if sys.platform == "darwin":
+            fn = libc.malloc_zone_pressure_relief
+            fn.argtypes = [ctypes.c_void_p, ctypes.c_size_t]
+            fn.restype = ctypes.c_size_t
+            _relief = ("darwin", fn)
+        else:
+            fn = libc.malloc_trim
+            fn.argtypes = [ctypes.c_size_t]
+            fn.restype = ctypes.c_int
+            _relief = ("glibc", fn)
+    except (OSError, AttributeError):
+        _relief = ("unsupported", None)
+    return _relief
+
+
+def trim() -> int:
+    """Collect garbage and return allocator free pages to the OS.
+
+    Returns the number of bytes freed on macOS (glibc's ``malloc_trim``
+    reports only success, so 0 is returned there and on unsupported
+    platforms).
+    """
+    kind, fn = _resolve()
+    if fn is None:
+        return 0
+    gc.collect()
+    if kind == "darwin":
+        return int(fn(None, 0))  # type: ignore[operator]
+    fn(0)  # type: ignore[operator]
+    return 0
+
+
+async def trim_periodically(interval_seconds: int = 60) -> None:
+    """Background task that periodically returns allocator free pages to the OS.
+
+    Runs in every worker process (allocator state is per-process). The
+    ``gc.collect()`` inside ``trim()`` holds the GIL for the duration of the
+    collection, so the pause lands between requests just as any other
+    collection would.
+
+    Args:
+        interval_seconds: How often to trim (default: 60 seconds).
+    """
+    while True:
+        await asyncio.sleep(interval_seconds)
+        try:
+            start = time.perf_counter()
+            freed = trim()
+            elapsed_ms = (time.perf_counter() - start) * 1000
+            log = logger.info if freed >= (16 << 20) else logger.debug
+            log(
+                "MallocTrim: returned %.1f MB to OS in %.0f ms",
+                freed / 1048576,
+                elapsed_ms,
+            )
+        except Exception as e:
+            logger.debug("MallocTrim failed: %s", e)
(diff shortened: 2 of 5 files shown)


JerrettDavis (changes requested):
The macOS re-exec mechanism is carefully gated, but the periodic trim path introduces an unconditional cross-platform event-loop regression. `periodic_malloc_trim_enabled` defaults true on every platform, and the lifespan task calls synchronous `trim()` directly: `gc.collect()` holds the GIL and `malloc_trim`/zone relief run on the async event-loop thread every 60 seconds, including Linux despite the PR saying non-Darwin is unaffected. With a large live heap this can pause all proxy traffic; an interval of 0 or a negative value also makes `asyncio.sleep()` return immediately and creates a continuous collect/trim loop. Please platform-gate the default/behavior consistently, validate a positive bounded interval, and move the blocking trim work off the event-loop thread where possible. Add a concurrency regression proving a deliberately slow trim does not stop unrelated async work, plus config tests for invalid intervals.

abhay-codes07:
Good catch, thanks. All three addressed in 7a5f783d:

1. Off the event loop: the periodic task now runs the blocking trim via `asyncio.to_thread(trim)`, so `gc.collect()` and the C-level purge never execute on the loop thread. New test asserts the trim runs on a worker thread (not the loop thread), and a second test parks a deliberately slow trim in that thread and proves an unrelated coroutine keeps ticking while it runs.

2. Interval validation: a value below the minimum (including 0 or negative, which made `asyncio.sleep` return immediately and busy-loop) now falls back to the default 60s with a warning. Parametrized test covers 0 and -5.

3. Consistent platform behavior: the task now resolves the platform once up front and returns immediately when there is no supported trim call (Windows, musl), so it is a true no-op there rather than a 60s spinner. Where a call exists (glibc `malloc_trim`, macOS `malloc_zone_pressure_relief`) it runs, but always off-thread. To be precise about my earlier wording: the darwin-only part is the pre-main re-exec of the libmalloc knobs (the actual fix for the RSS ratchet); the trim task is cross-platform defense-in-depth on glibc + macOS and now cannot block the loop anywhere. Happy to gate the trim darwin-only as well if you would rather keep it strictly to the reported platform.

14 tests pass, ruff and mypy clean.

JerrettDavis (changes requested):
Thanks for addressing the busy-loop case and moving the allocator C call onto a worker. The latest head passes its 14 focused tests and Ruff locally, but two parts of the original blocking concern remain:

1. `periodic_malloc_trim_enabled` still defaults to true in `ProxyConfig`, both env builders, and therefore on Linux. This PR is scoped and evidenced as a macOS RSS fix, yet it introduces a new once-per-minute full collection/trim policy for every glibc deployment. Please make Darwin the default-enabled platform and require an explicit opt-in elsewhere, or provide Linux-specific evidence and a separately justified policy.

2. `asyncio.to_thread(trim)` does not make the full operation non-blocking: `trim()` begins with `gc.collect()`, whose cyclic-GC traversal runs while holding the GIL. The new concurrency test replaces `trim` with `threading.Event.wait()`, which releases the GIL, so it proves thread dispatch but not that the real full-heap collection leaves the event loop responsive. The docstring/comment saying it “never blocks the event loop” is therefore too strong. Please remove the unconditional full collection from this periodic path (allocator pressure relief can still run off-thread), or redesign/measure it so request-latency impact is bounded; update the regression to exercise the relevant GIL-holding behavior rather than an Event wait.

The invalid-interval fallback and unsupported-platform early exit look good. Once the default platform scope and GIL pause are resolved, this should be much safer for a long-lived proxy.

abhay-codes07:
Thanks, both remaining points addressed in 9a0221a5.

1. Platform-scoped default. `periodic_malloc_trim_enabled` now defaults on only where `sys.platform == "darwin"` and is an explicit opt-in (`HEADROOM_MALLOC_TRIM=1`) everywhere else. Changed in the `ProxyConfig` field (`default_factory`) and both env builders (`cli/proxy.py`, `server.py`). A glibc deployment no longer takes on the once-a-minute purge unless it asks for it. New test asserts the default is true on darwin and false on linux.

2. Dropped the unconditional `gc.collect()` from the periodic path. You are right that `to_thread` did not fix this: `gc.collect()` runs a whole-heap cyclic traversal while holding the GIL. The periodic trim now runs only the allocator pressure-relief C call (`malloc_zone_pressure_relief` / `malloc_trim`), which releases the GIL for its duration; reclaiming cyclic garbage is left to CPython's own automatic collection rather than forced on this off-thread task. So the periodic path no longer holds the GIL for a full-heap collection at all. The docstrings no longer say "never blocks the event loop" and instead state the actual behavior (a GIL-releasing C call dispatched off-loop). New regression asserts `trim()` never calls `gc.collect()`; I updated the concurrency test's framing accordingly (it now honestly models the GIL-releasing purge, not a stand-in for GIL-held work).

The re-exec of the libmalloc knobs (the primary macOS fix) is unchanged. 15 tests pass, ruff and mypy clean.

JerrettDavis (approved):
Approved on exact head `9a0221a58ec243c7e22b3e8cfd94ba68a5ae37c1`. The correction scopes default-on periodic pressure relief to Darwin, keeps glibc trimming opt-in, and removes forced `gc.collect()` from the periodic path. I ran the 15 focused allocator/config tests plus Ruff locally; the complete GitHub check rollup is now green, including all four test shards and Docker-native e2e.

JerrettDavis:
CI status note: this PR is not review-ready under the all-green requirement. Test shard 4 was cancelled by the repository's 30-minute job ceiling without a test assertion result, independently reproducing the infrastructure issue addressed by #2996. After that non-skipping CI correction is reviewed and human-merged, this PR must be refreshed onto current `main` and its complete applicable suite rerun green. No merge or auto-merge action has been taken.

abhay-codes07:
Thanks for the clear CI note. #2996 merged earlier today, so I have refreshed this branch onto current main (clean rebase over the three malloc-tuning commits, force-pushed) and the full applicable suite is rerunning now under the non-skipping correction. No behavior change in the refresh. I will confirm once shard 4 lands green.

JerrettDavis (approved):
Re-reviewed the rebased series against the previously approved implementation using `git range-diff`. The allocator policy and its safety fixes are unchanged: default-on remains Darwin-only, glibc is opt-in, the periodic path does not force `gc.collect()`, interval validation remains in place, and the allocator call stays off the event loop. The only substantive rebase integration is lifecycle cleanup: the periodic trim task is stored on app state, cancelled, and awaited during shutdown, consistent with `main`’s newer periodic-task handling.

The exact current head is mergeable and its full applicable CI rollup is green across all Python shards, native wrappers, and Docker checks.

Issue #2797, opened 2026-08-05
[BUG] Headroom becomes unhealthy with Max outbound streams is 100, 100 open when used with Claude Code


kuwabaky:
# Headroom becomes unhealthy after prolonged use with Claude Code (`Max outbound streams is 100, 100 open`)

## Summary

After using Headroom as an Anthropic proxy for Claude Code, the proxy eventually becomes unhealthy and stops responding.

Claude Code starts returning:

```
API Error: 502
An error occurred while processing your request.
If it persists, check your inference gateway (localhost:8787).
```

At the same time, the health endpoint reports:

```json
"upstream": {
  "status": "unhealthy",
  "error": "Max outbound streams is 100, 100 open"
}
```

Eventually, requests to `/health` may also hang.

---

## Environment

- macOS (Apple Silicon)
- Headroom 0.33.0 (also reproduced on 0.32.1)
- RTK 0.44.2
- Claude Code 2.1.221
- Backend: Anthropic
- Persistent service (`headroom install`)
- Runtime: Python

---

## Reproduction

The issue is not immediately reproducible.

Typical sequence:

1. Start Headroom as a persistent service.
2. Configure Claude Code to use Headroom (`ANTHROPIC_BASE_URL=http://127.0.0.1:8787`).
3. Use Claude Code normally for an extended period (multiple conversations and tool calls).
4. Eventually, Claude Code starts returning HTTP 502 errors.

---

## Health endpoint

Before failure:

```
Status: running
Healthy: yes
```

After failure:

```
Status: running
Healthy: no
```

Health response includes:

```json
{
  "service": "headroom-proxy",
  "status": "unhealthy",
  "checks": {
    "upstream": {
      "status": "unhealthy",
      "error": "Max outbound streams is 100, 100 open"
    }
  }
}
```

Sometimes `/health` itself stops responding and hangs.

---

## Expected behavior

Headroom should continue forwarding requests without exhausting upstream HTTP/2 streams.

---

## Actual behavior

- Claude Code receives HTTP 502 errors.
- Headroom becomes unhealthy.
- Upstream check reports:
  - `Max outbound streams is 100, 100 open`
- Restarting the Headroom service temporarily fixes the issue, but it eventually returns.

---

## Additional observations

- Connecting Claude Code directly to Anthropic (without Headroom) works normally.
- The issue only occurs when requests go through Headroom.
- Upgrading from Headroom 0.32.1 to 0.33.0 did not prevent the issue from recurring.
[...]

PR #2882, opened 2026-08-09
fix(proxy): close the upstream stream when a streaming body is never consumed

abhay-codes07:
## Description

A long-lived proxy serving Claude Code eventually goes unhealthy with:

```json
"upstream": {"status": "unhealthy", "error": "Max outbound streams is 100, 100 open"}
```

and clients start getting `502`s until the service is restarted (#2797).

Root cause is a leaked upstream HTTP/2 stream on the streaming path. In `_stream_response_inner` the upstream response is opened *before* the body generator, so its headers are available to build the `StreamingResponse` (ratelimit forwarding):

```python
upstream_response = await self.http_client.send(_upstream_req, stream=True)
...
return StreamingResponse(generate(), media_type="text/event-stream", headers=forwarded_headers)
```

The only thing that closes `upstream_response` on the success path is the generator's own `contextlib.aclosing`:

```python
async def generate():
    ...
    async with contextlib.aclosing(upstream_response) as response:
        async for chunk in response.aiter_bytes():
            ...
```

That cleanup runs only if the body is iterated. When a client disconnects before Starlette starts sending the body, the generator is never started, its `aclosing` never runs, and nothing else closes the stream. This is routine with Claude Code, which cancels or supersedes in-flight turns (ESC, a newer request, `/compact` side-requests). Each abandoned request leaks one open h2 stream.

Leaked streams accumulate on the pooled upstream connection until it reaches the server's `SETTINGS_MAX_CONCURRENT_STREAMS` (100 for Anthropic). After that httpcore cannot open a new stream on it, the upstream health probe surfaces `Max outbound streams is 100, 100 open`, and the proxy stays unhealthy until restart. It builds up slowly, which matches the "after prolonged use" reports.

The outer `_stream_response` already anticipates exactly this disconnect-mid-setup case for the session key:

```python
# any exception here - including asyncio.CancelledError from a client disconnect
# mid-setup - must still release session_key ...
except (Exception, asyncio.CancelledError):
    self._cleanup_mid_turn_stream(session_key)
    raise
```

The opened upstream stream never got the same guarantee.

## Fix

Attach an idempotent close as the `StreamingResponse` `background` task. Starlette runs a response's background task after the body finishes *and* after an early client disconnect, so the stream is released in both cases. On the normal path the generator has already closed it, so `aclose()` is a no-op (`httpx.Response.aclose()` is idempotent).

```python
async def _release_upstream_stream() -> None:
    with contextlib.suppress(Exception):
        await upstream_response.aclose()

return StreamingResponse(
    generate(),
    media_type="text/event-stream",
    headers=forwarded_headers,
    background=BackgroundTask(_release_upstream_stream),
)
```

The already-correct paths are untouched: the `>= 400` error path and the retry path close the stream explicitly, and the transport-error path returns before a stream is held.

## Type of Change

## Changes Made

- `headroom/proxy/handlers/streaming.py` (`_stream_response_inner`): attach `background=BackgroundTask(_release_upstream_stream)` to the success `StreamingResponse` so the upstream stream is closed even when the body generator is never iterated. Imports `starlette.background.BackgroundTask`.
- `tests/test_proxy_streaming_ratelimit_headers.py`: add `test_upstream_stream_closed_when_body_never_consumed` - builds the streaming response, does not consume the body (asserts `aclose` not yet called), then runs the response's background task and asserts the upstream `aclose` is awaited.

## Testing

- [x] New test added

### Test Output

```text
# Fail-before (background= line removed, new test kept):
tests/test_proxy_streaming_ratelimit_headers.py::...::test_upstream_stream_closed_when_body_never_consumed FAILED
  assert result.background is not None -> AssertionError: streaming response must carry a cleanup task

# Pass-after (fix in place):
tests/test_proxy_streaming_ratelimit_headers.py  12 passed
tests/test_proxy_streaming_ratelimit_headers.py tests/test_h2_stream_reset_retry.py
tests/test_proxy_streaming_request_logger.py
tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py   34 passed

# uvx mypy@1.20.2 headroom/proxy/handlers/streaming.py -> Success: no issues found in 1 source file
```

## Real Behavior Proof

- Steps: traced the streaming forwarder from `send(stream=True)` to the `StreamingResponse` return and confirmed the only close on the success path is the body generator's `contextlib.aclosing`, which does not run if the body is never iterated. Reproduced that condition in the existing streaming unit harness (mock upstream whose `aclose` is an `AsyncMock`), built the response, left the body unconsumed, and observed `aclose` was never awaited before the fix; the response also carried no `background` task. With the fix, the response carries a background task that awaits the upstream `aclose`.
- Observed result: an abandoned streaming request now releases its upstream stream instead of leaking it, so streams no longer accumulate toward the 100-stream ceiling on the pooled connection.
- Not tested: an end-to-end multi-hour Claude Code session reproducing the 100-open exhaustion against a live Anthropic connection (no live long-running proxy here). The leak condition and its release are verified directly at the streaming handler with the same plumbing the proxy uses.

## Review Readiness

## Checklist

## Additional Notes

`aclose()` is idempotent, so the background close is safe alongside the generator's own cleanup on the normal path - it simply becomes the sole closer when the generator never runs. The background task runs only after the response lifecycle completes, so it never races an in-flight stream.

### DIFF
--- a/headroom/proxy/handlers/streaming.py
+++ b/headroom/proxy/handlers/streaming.py
@@ -1098,6 +1098,7 @@ class StreamingMixin:
     ) -> Response | StreamingResponse:
         """Actual streaming implementation, guarded by _stream_response's cleanup wrapper."""
         from fastapi.responses import Response, StreamingResponse
+        from starlette.background import BackgroundTask
 
         from headroom.proxy.helpers import MAX_SSE_BUFFER_SIZE
 
@@ -1656,10 +1657,29 @@ class StreamingMixin:
                     )
                     yield f"event: headroom_pending_messages\ndata: {pending_event}\n\n".encode()
 
+        async def _release_upstream_stream() -> None:
+            # Guarantee the upstream HTTP/2 stream is released even when the
+            # body generator above is never iterated — the client disconnected
+            # before Starlette started sending the response body (routine when a
+            # harness like Claude Code cancels or supersedes an in-flight turn),
+            # so ``generate()`` never entered its own ``aclosing`` and nothing
+            # else closes ``upstream_response``. Each such request otherwise
+            # leaks one open h2 stream; they accumulate on the pooled upstream
+            # connection until it reaches SETTINGS_MAX_CONCURRENT_STREAMS (100)
+            # and no new stream can open ("Max outbound streams is 100, 100
+            # open"), and the proxy goes unhealthy until restart (#2797).
+            # Starlette runs a response's ``background`` task after the body
+            # finishes *and* after an early client disconnect, so this fires in
+            # both cases. ``aclose()`` is idempotent, so on the normal path —
+            # where the generator already closed the stream — this is a no-op.
+            with contextlib.suppress(Exception):
+                await upstream_response.aclose()
+
         return StreamingResponse(
             generate(),
             media_type="text/event-stream",
             headers=forwarded_headers,
+            background=BackgroundTask(_release_upstream_stream),
         )
 
     async def _stream_response_bedrock(
--- a/tests/test_proxy_copilot_auth_hooks.py
+++ b/tests/test_proxy_copilot_auth_hooks.py
@@ -52,11 +52,22 @@ def _load_handler_module(monkeypatch: pytest.MonkeyPatch, module_name: str, rela
     responses_mod = types.ModuleType("fastapi.responses")
 
     class Response:
-        def __init__(self, content=None, status_code: int = 200, headers=None, media_type=None):
+        def __init__(
+            self,
+            content=None,
+            status_code: int = 200,
+            headers=None,
+            media_type=None,
+            background=None,
+        ):
             self.content = content
             self.status_code = status_code
             self.headers = headers or {}
             self.media_type = media_type
+            # The streaming forwarder attaches a background task that releases the
+            # upstream stream when the body is never consumed (#2882); the double
+            # must accept and store it so the real StreamingResponse call works.
+            self.background = background
 
     class StreamingResponse(Response):
         pass
@@ -228,6 +239,9 @@ def test_streaming_response_applies_copilot_auth(monkeypatch: pytest.MonkeyPatch
     assert sent_headers["Authorization"] == "Bearer upstream-token"
     assert sent_headers["content-type"] == "application/json"
     assert response.status_code == 200
+    # The Copilot auth hook and the #2882 upstream-stream cleanup coexist: the
+    # streaming response still carries its background release task.
+    assert response.background is not None
 
 
 def test_openai_chat_routes_copilot_requests_per_model(monkeypatch: pytest.MonkeyPatch) -> None:
--- a/tests/test_proxy_streaming_ratelimit_headers.py
+++ b/tests/test_proxy_streaming_ratelimit_headers.py
@@ -513,6 +513,56 @@ class TestStreamingRatelimitHeaderForwarding:
         mock_response.aclose.assert_awaited()
 
     @pytest.mark.asyncio
+    async def test_upstream_stream_released_over_asgi_lifecycle_on_disconnect(self):
+        """Driving the real ASGI response through an early disconnect releases the stream.
+
+        Rather than calling ``result.background()`` directly, this exercises the
+        Starlette response lifecycle with a client that disconnects immediately,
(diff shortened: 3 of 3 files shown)


JerrettDavis (changes requested):
The cleanup mechanism is directionally sound, and the new focused test plus the existing streaming suites pass locally. One repository integration is broken on the current head, however: `tests/test_proxy_copilot_auth_hooks.py::test_streaming_response_applies_copilot_auth` replaces FastAPI’s response classes with a minimal `Response` test double whose constructor does not accept `background`. `_stream_response_inner()` now unconditionally passes that keyword, so the real Copilot-auth streaming regression fails with `TypeError` both locally and in CI shard 3.

Please update that response double to accept/store the background task and extend the assertion so the existing auth hook and cleanup coexist. Ideally exercise the response lifecycle (including a simulated early disconnect) rather than only calling `result.background()` directly; that would prove Starlette invokes the close in the condition this PR targets, not just that a callable was attached. CI shard 4 is also red on the native-tls release-tree guard; that appears unrelated to these two changed files, but the head still needs a clean required run before approval.

abhay-codes07:
Thanks, fixed in b5253790.

The minimal `Response` test double in `test_proxy_copilot_auth_hooks.py` now accepts and stores the `background` keyword, so `test_streaming_response_applies_copilot_auth` no longer TypeErrors when the forwarder passes the cleanup task. That test also now asserts the Copilot auth hook and the cleanup task coexist on the response (`response.background is not None`).

On the stronger-proof point: I added `test_upstream_stream_released_over_asgi_lifecycle_on_disconnect` to `test_proxy_streaming_ratelimit_headers.py`. Instead of calling `result.background()` directly, it drives the real Starlette `StreamingResponse.__call__` with a receive channel that returns `http.disconnect` immediately (client gone before the body is consumed) and asserts the upstream `aclose()` was awaited by the end of the lifecycle -- proving Starlette actually invokes the release in the condition this PR targets. The existing direct-invocation test is kept as the focused unit check. Full streaming suite + the copilot-auth streaming test pass; ruff clean.

Re the CI shard 4 native-tls release-tree guard: agreed it looks unrelated to these two files; I will keep an eye on the required re-run.

ashishpatel26:
## Question: sequential idempotency vs. concurrent safety on mid-stream disconnect

Nice diagnosis — the before-body-iteration leak and its fix both check out. One gap I don't see covered:

### Scenario
Not the case this PR targets (client disconnects *before* `generate()` starts iterating), but a **mid-stream** disconnect — the client goes away while `generate()` is actively inside `async for chunk in response.aiter_bytes()`. Two cleanup paths become live at once:

1. Starlette unwinds the generator on disconnect → `contextlib.aclosing.__aexit__` → `upstream_response.aclose()`
2. Starlette also runs the response's `background` task after the send loop ends → `_release_upstream_stream()` → `upstream_response.aclose()`

### The actual question
`aclose()` being **idempotent** (`if self.is_closed: return`) protects against a *second* call after the *first* has already completed and set its flag — i.e., sequential double-close. It does **not** by itself protect against two coroutines both reading `is_closed=False` **before either has set it** — a TOCTOU race, not a sequencing guarantee.

So: does Starlette strictly *sequence* "generator teardown fully awaited" **before** "background task starts," or can the two run concurrently on a mid-stream disconnect? If it's the latter:

- Does `httpx.Response.aclose()` (or the underlying `h2`/`httpcore` stream-close path) hold an internal lock that makes concurrent entry safe, or are these two independent close operations racing on the same stream object?
- Is there a test that drives disconnect **mid-iteration** (not before-body-consumption) to prove both cleanup paths can't land concurrently? `test_upstream_stream_released_over_asgi_lifecycle_on_disconnect` proves the target bug (disconnect-before-iteration) is fixed, but doesn't appear to exercise the case where both paths are simultaneously live.

If Starlette's scheduling here is concurrent rather than sequential, that's a second, subtler failure mode — a double-close race, not a leak — that "idempotent" alone doesn't rule out.

JerrettDavis (approved):
The revised head resolves the requested integration gap: the Copilot response double now accepts and retains the cleanup task, and the new ASGI lifecycle test drives an immediate disconnect through Starlette rather than merely invoking the callback directly. The upstream response is closed on both normal-generator and early-disconnect paths, and all current checks are green.

PR #2883, opened 2026-08-09
fix(litellm): don't forward a caller key the target cannot accept

chopratejas:
Split out of #2852 at review request: that PR is bounded upstream calls plus
measured hot-path costs, and this is an authentication/routing change that
belongs on its own scope. #2852 now carries only the timeout work.

## The bug

A routing extension can rewrite the model across families mid-request
(`claude-opus-5` → `gpt-5-mini`). The caller's key does not travel with that
rewrite, so the proxy forwards `sk-ant-...` to OpenAI and earns a guaranteed
401. Downstream that is indistinguishable from *"the cheap model failed the
task"* — it scores as a quality regression against the router, not as a bug.

Dropping the `api_key` kwarg instead lets litellm fall back to the target
provider's own env credential, which is the only key that can work.

## Why this cut is different from the one that was rejected

The first version returned `not provider.startswith("anthropic")`, so **any**
non-`sk-ant-` credential was dropped against an Anthropic-class target — a
plain Bearer token against an Anthropic-compatible or custom gateway lost its
key and fell back to an env credential that may not exist.

That direction is the dangerous one. A false refusal breaks a deployment that
was working; a missed refusal just leaves today's 401. So this refuses on
**positive evidence only**:

| credential | target | forwarded? |
|---|---|---|
| `sk-ant-…` | `openai` / `azure` / `gemini` | **no** — cannot possibly authenticate |
| `sk-ant-…` | anthropic | yes |
| `sk-ant-…` | unrecognised / unclassifiable model | yes — pass-through |
| anything else | anything | yes — pass-through, unchanged |

`sk-ant-` is Anthropic's documented vendor-specific prefix, which is what makes
it classifiable. `sk-` is not: a dozen vendors mint that shape. Everything the
string cannot settle keeps main's behaviour.

The reject list is explicit rather than inverted (`not anthropic`) because an
unrecognised provider is usually a compatible or self-hosted gateway. Marked in
the code as a hand-kept tuple with the registry-lookup upgrade path noted.

Bedrock / Vertex / SageMaker are unaffected — all four dispatch sites already
skip credential forwarding for them entirely (env-based auth).

## Verification

`tests/test_litellm_caller_key.py`, 12 cases — the refusal, the Anthropic
target, the unknown provider, `get_llm_provider` raising, and each
unclassifiable credential shape asserted against **both** target families.
Those last ones fail against the rejected version.

Applied at all four dispatch sites (Anthropic non-stream/stream, OpenAI
non-stream/stream).

### DIFF
--- a/headroom/backends/litellm.py
+++ b/headroom/backends/litellm.py
@@ -447,6 +447,47 @@ def _upstream_timeout() -> float:
     return v if v > 0 else DEFAULT_UPSTREAM_TIMEOUT
 
 
+# Providers that cannot possibly accept an Anthropic `sk-ant-` credential.
+#
+# Explicit, rather than the inverse "anything not Anthropic": an unrecognised
+# provider is usually a compatible or self-hosted gateway, and guessing wrong
+# there drops a key that WAS working. Bedrock/Vertex are absent because the
+# dispatch sites already skip them entirely (env-based auth).
+#
+# ponytail: a hand-kept tuple; grow it as targets are confirmed. A registry
+# lookup would be the upgrade if this ever outgrows a handful of entries.
+_REJECTS_ANTHROPIC_KEY = ("openai", "azure", "gemini")
+
+
+def _caller_key_travels_to(model: str, key: str) -> bool:
+    """Can this inbound credential authenticate the provider we are about to call?
+
+    The caller authenticates to the PROXY. A routing extension may then rewrite
+    the model across families mid-request (claude-opus-5 -> gpt-5-mini), and the
+    caller's key does not travel with that rewrite: we forward `sk-ant-...` to
+    OpenAI and earn a guaranteed 401, which reads downstream as "the cheap model
+    failed the task" rather than as the routing bug it is.
+
+    Only an unambiguous mismatch is refused. `sk-ant-` is Anthropic's documented
+    vendor-specific prefix, so it cannot authenticate one of the providers above.
+    Every other credential -- a plain Bearer token, an OpenAI-style `sk-` that a
+    dozen vendors also mint, anything aimed at a compatible or custom gateway --
+    is unclassifiable from the string alone and keeps the pass-through.
+
+    Returning False drops the api_key kwarg, so litellm falls back to the target
+    provider's own env credential: the only key that can work.
+    """
+    if not key.startswith("sk-ant-"):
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The positive-evidence direction is much safer, and the 12 focused tests plus Ruff pass locally on `a32ef0e2`. One cross-provider hole remains before this is safe to merge: all four dispatch sites decide the env-auth bypass from `self.provider`, but the scenario this PR addresses is specifically one where `litellm_model` has been rewritten to a different target provider. If an Anthropic-backed request is rewritten to `bedrock/...`, `vertex_ai/...`, `vertex_ai_beta/...`, or `sagemaker/...`, `self.provider` is still Anthropic, `_caller_key_travels_to()` returns true because those targets are absent from `_REJECTS_ANTHROPIC_KEY`, and the caller `sk-ant-...` is forwarded despite the comments saying those targets must use AWS/Google env auth. Please classify the rewritten target once and use that target provider both for the env-auth skip and the mismatch decision, with dispatch-level tests for at least one Anthropic→Bedrock/Vertex rewrite (not only helper tests). The branch also currently conflicts with `main` after #2852 landed, so the corrected head will need a clean rebase/merge and green CI.

JerrettDavis (changes requested):
Re-reviewed exact rebased head `4ae687273ee5154aabc74811ea249c08c99b44fa`. The 12 helper tests and Ruff/format checks pass locally, but the requested dispatch fix is still absent: all four sites test `self.provider` against `_env_auth_providers`, even though this PR’s scenario is a routing extension rewriting `litellm_model` to a different target. For an Anthropic-origin request rewritten to `bedrock/...`, `vertex_ai/...`, `vertex_ai_beta/...`, or `sagemaker/...`, the original-provider gate permits forwarding and `_caller_key_travels_to()` returns true because those target providers are intentionally absent from `_REJECTS_ANTHROPIC_KEY`; the caller `sk-ant-...` therefore still overrides/contaminates an environment-auth target despite the comments claiming those targets are skipped. The rebase only incorporated current main and added no target-rewrite dispatch test. Please classify the rewritten target once, use it for the env-auth bypass as well as mismatch logic, and add a dispatch-level Anthropic→Bedrock or Vertex regression test.

PR #2885, opened 2026-08-10
fix(image): cap the technique-router ONNX intra-op thread pool

probepark:
## Description

`OnnxTechniqueRouter._load_classifier` builds its session with a bare `create_cpu_session_options(ort)`, so ORT sizes the intra-op thread pool from the host core count:

```python
self._classifier_session = ort.InferenceSession(
    model_path,
    create_cpu_session_options(ort),
    providers=["CPUExecutionProvider"],
)
```

The classifier is a MiniLM INT8 graph over a fixed `[1, 64]` input — the tokenizer two lines below pins it with `enable_truncation(max_length=64)` / `enable_padding(length=64)`. A graph that small saturates after a handful of threads, so on a many-core host the extra threads cost CPU without buying latency.

`headroom/memory/adapters/embedders.py` already pins its own ONNX session (`intra_op_num_threads=1, inter_op_num_threads=1`) with a comment about keeping the pool small. This brings the two call sites in line.

Follow-up to @JerrettDavis' review: the cap is now sized from the CPUs this
process may run on, not from the host's core count.

Closes #

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `headroom/image/onnx_router.py`: build the classifier session with
  `create_cpu_session_options(ort, intra_op_num_threads=..., inter_op_num_threads=1)`
  instead of the bare call, capped at `_CLASSIFIER_MAX_INTRA_OP_THREADS = 4`.
- `headroom/image/onnx_router.py`: add `_available_cpu_count()`, which prefers
  `len(os.sched_getaffinity(0))`, falls back to `os.process_cpu_count()`
  (Python 3.13+) and only then to `os.cpu_count()`. `os.cpu_count()` reports the
  host's cores inside a cgroup-restricted cpuset, so a container pinned to two
  CPUs on a 64-core box would have been handed the full four threads — the exact
  oversubscription this PR exists to remove.
- `tests/test_onnx_router_threads.py`: 7 tests covering the session options, the
  container case (host cores > process affinity), both fallback layers, and the
  untouched retention defaults.

## Testing

### Test Output

```text
$ uv run pytest tests/test_onnx_router_threads.py tests/test_onnx_runtime.py -q
21 passed in 1.18s

$ uv run ruff check headroom/image/onnx_router.py tests/test_onnx_router_threads.py
All checks passed!

$ uv run ruff format --check headroom/image/onnx_router.py tests/test_onnx_router_threads.py
2 files already formatted

$ uv run mypy headroom/image/onnx_router.py
```

Before this follow-up, `test_thread_cap_follows_process_affinity_not_host_cores`
fails on the old helper:

```text
E   assert 4 == 2
E    +  where 4 = _classifier_intra_op_threads()
```

## Real Behavior Proof

- Environment: macOS 27.0 arm64 (20 cores), Python 3.12.13, onnxruntime
  1.23.2, headroom @ `main` (2f2950a) + this branch.
- Exact command / steps: ran the focused ONNX router/runtime tests and the CPU-affinity probe shown below.

```text
$ uv run python -c "
import os
from headroom.image import onnx_router as r
print('host cpu_count      :', os.cpu_count())
print('available (real)    :', r._available_cpu_count())
print('intra-op (real)     :', r._classifier_intra_op_threads())
os.sched_getaffinity = lambda pid: {3, 11}   # emulate a 2-CPU cpuset
os.cpu_count = lambda: 64
print('available (cpuset 2):', r._available_cpu_count())
print('intra-op (cpuset 2) :', r._classifier_intra_op_threads())
"
```

- Observed result: the real process selects four intra-op threads, while an emulated two-CPU cpuset selects two; focused tests and static checks pass.

```text
host cpu_count      : 20
available (real)    : 20
intra-op (real)     : 4
available (cpuset 2): 2
intra-op (cpuset 2) : 2
```

- Not tested: no run inside a real Docker cgroup, and cgroup CPU quota
  (`--cpus=1.5`, `cpu.max`) is still invisible to both `sched_getaffinity` and
  `os.cpu_count()` — that is a separate gap from the cpuset case raised in
  review, and nothing else in the repo reads cgroup limits today.

### Thread-count benchmark (unchanged from the original submission)

Environment: macOS 27.0 arm64 (20 physical cores, 16P + 4E), Python 3.13, onnxruntime 1.26.0, headroom @ `main` (2f2950a).

Sessions were built through `create_cpu_session_options`, not a bare `ort.SessionOptions()`, so the arena and `allow_spinning` entries from #2540 are in effect and the numbers reflect what the proxy actually runs. Median of 60 calls after 5 warmup calls, `[1, 64]` int64 inputs.

### technique router (`model_quantized.onnx`)

| intra-op threads | median ms | p90 ms | CPU ms/call |
| --- | --- | --- | --- |
| 1 | 12.72 | 13.08 | 12.65 |
| 2 | 9.96 | 10.55 | 16.34 |
| 4 | 8.57 | 9.29 | 20.76 |
| 8 | 9.56 | 10.10 | 27.17 |
| 20 | 15.45 | 19.55 | 69.87 |
| ORT default (today) | 10.07 | 10.57 | 33.65 |

4 threads is 15% faster and 38% cheaper in CPU than the current default. It is strictly dominant here — there is no latency/CPU tradeoff to argue about.

### SigLIP encoder — measured, deliberately unchanged

| intra-op threads | median ms | CPU ms/call |
| --- | --- | --- |
| 4 | 56.48 | 180.28 |
| 8 | 42.55 | 209.63 |
| 20 | 50.86 | 370.17 |
| ORT default (today) | 41.82 | 242.42 |

ORT's default is already the encoder's best latency point, so this PR leaves the SigLIP session alone. Capping it to 8 would trade 2% latency for 14% CPU, which is a judgement call rather than a bug fix; say the word and I will add it.

## Review Readiness

## Checklist

## Additional Notes

Documentation is unchanged: the cap is an internal ONNX session detail with no
user-facing knob. The SigLIP session is still left at ORT's default on purpose
(see the benchmark above).

### DIFF
--- a/headroom/image/onnx_router.py
+++ b/headroom/image/onnx_router.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 import io
 import logging
 import math
+import os
 from pathlib import Path
 from typing import Any
 
@@ -26,6 +27,18 @@ logger = logging.getLogger(__name__)
 
 _TECHNIQUE_ROUTER_REPO = "chopratejas/technique-router-onnx"
 _SIGLIP_ENCODER_REPO = "chopratejas/siglip-image-encoder-onnx"
+
+# The query classifier is a MiniLM INT8 graph over a fixed [1, 64] input (the
+# tokenizer pads to exactly 64 below), so it saturates after a handful of
+# threads. Left at ORT's default the pool scales with the core count and the
+# per-call cost grows without buying latency -- on a 20-core box that measured
+# 10.07 ms / 33.65 CPU-ms versus 8.57 ms / 20.76 CPU-ms at 4 threads.
+# Capped rather than fixed so smaller hosts keep ORT's own choice.
+_CLASSIFIER_MAX_INTRA_OP_THREADS = 4
+
+
+def _classifier_intra_op_threads() -> int:
+    return max(1, min(_CLASSIFIER_MAX_INTRA_OP_THREADS, os.cpu_count() or 1))
 
 
 # ImageSignals, RouteDecision, Technique imported from trained_router
@@ -64,7 +77,11 @@ class OnnxTechniqueRouter:
         model_path = hf_hub_download_local_first(_TECHNIQUE_ROUTER_REPO, "model_quantized.onnx")
         self._classifier_session = ort.InferenceSession(
             model_path,
-            create_cpu_session_options(ort),
+            create_cpu_session_options(
+                ort,
+                intra_op_num_threads=_classifier_intra_op_threads(),
+                inter_op_num_threads=1,
+            ),
             providers=["CPUExecutionProvider"],
         )
 
--- a/headroom/image/onnx_router.py
+++ b/headroom/image/onnx_router.py
@@ -37,8 +37,30 @@ _SIGLIP_ENCODER_REPO = "chopratejas/siglip-image-encoder-onnx"
 _CLASSIFIER_MAX_INTRA_OP_THREADS = 4
 
 
+def _available_cpu_count() -> int:
+    """CPUs this process may actually run on.
+
+    ``os.cpu_count()`` reports the host's cores even when a cgroup cpuset pins
+    the process to a subset, so a two-CPU container on a 64-core box would be
+    handed the full four-thread cap -- exactly the oversubscription the cap
+    exists to avoid. Prefer the affinity-aware counts where the platform has
+    them.
+    """
+    sched_getaffinity = getattr(os, "sched_getaffinity", None)
+    if sched_getaffinity is not None:
+        return max(1, len(sched_getaffinity(0)))
+
+    process_cpu_count = getattr(os, "process_cpu_count", None)  # Python 3.13+
+    if process_cpu_count is not None:
+        usable: int | None = process_cpu_count()
+        if usable:
+            return usable
+
+    return max(1, os.cpu_count() or 1)
+
+
 def _classifier_intra_op_threads() -> int:
-    return max(1, min(_CLASSIFIER_MAX_INTRA_OP_THREADS, os.cpu_count() or 1))
+    return min(_CLASSIFIER_MAX_INTRA_OP_THREADS, _available_cpu_count())
 
 
 # ImageSignals, RouteDecision, Technique imported from trained_router
--- a/tests/test_onnx_router_threads.py
+++ b/tests/test_onnx_router_threads.py
@@ -19,6 +19,7 @@ import pytest
 from headroom.image.onnx_router import (
     _CLASSIFIER_MAX_INTRA_OP_THREADS,
     OnnxTechniqueRouter,
+    _available_cpu_count,
     _classifier_intra_op_threads,
 )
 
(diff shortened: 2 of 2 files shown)


JerrettDavis (changes requested):
The measured four-thread ceiling is useful, but `_classifier_intra_op_threads()` does not actually preserve the stated smaller-host/container guarantee. `os.cpu_count()` commonly reports the host CPU count inside a cgroup or restricted cpuset, so a process limited to one or two CPUs can still be explicitly assigned four ORT intra-op threads. That can introduce the exact oversubscription this change is intended to prevent, and it may override a better affinity-aware ORT default.

Please base the minimum on CPUs available to this process (for example `len(os.sched_getaffinity(0))` where supported, with a portable fallback such as `os.process_cpu_count()` when available and then `os.cpu_count()`), and add a regression test that models host count > process affinity. The existing 2-core test only patches host count and does not cover the container case.

probepark:
@JerrettDavis you're right — thanks, that was a real hole. `os.cpu_count()` reports the host's cores inside a cgroup-restricted cpuset, so a container pinned to 1–2 CPUs on a big box got the full 4 intra-op threads and I'd have shipped the exact oversubscription this PR is supposed to remove.

Fixed in ccda45f. The cap is now sized from the CPUs the *process* may run on:

```python
def _available_cpu_count() -> int:
    sched_getaffinity = getattr(os, "sched_getaffinity", None)
    if sched_getaffinity is not None:
        return max(1, len(sched_getaffinity(0)))

    process_cpu_count = getattr(os, "process_cpu_count", None)  # Python 3.13+
    if process_cpu_count is not None:
        usable: int | None = process_cpu_count()
        if usable:
            return usable

    return max(1, os.cpu_count() or 1)


def _classifier_intra_op_threads() -> int:
    return min(_CLASSIFIER_MAX_INTRA_OP_THREADS, _available_cpu_count())
```

`getattr` rather than a version check because `requires-python = ">=3.10"`: `sched_getaffinity` is Linux-only and `process_cpu_count` is 3.13+. `process_cpu_count()` returning `None` (undeterminable) falls through to `os.cpu_count()` instead of clamping to 1.

Tests — the old 2-core test only patched `os.cpu_count`, which on Linux wouldn't have caught this at all, so the helper `_fake_cpu_topology()` now pins every layer explicitly and the suite is deterministic on macOS and Linux alike:

- `test_thread_cap_follows_process_affinity_not_host_cores` — host reports 64, affinity is `{3, 11}` → 2 threads (the regression you asked for)
- `test_classifier_session_honors_affinity_on_many_core_host` — same topology asserted on the built session, not just the helper
- `test_thread_cap_uses_process_cpu_count_without_affinity_api` — no `sched_getaffinity` (macOS/Windows), `process_cpu_count() == 2` → 2
- `test_thread_cap_falls_back_to_host_cores` — neither API usable → host count
- `test_classifier_thread_cap_never_exceeds_available_cores` — rewritten to delete both APIs so it really tests the host-count path

```
$ uv run pytest tests/test_onnx_router_threads.py tests/test_onnx_runtime.py -q
21 passed in 1.18s
```

One gap I deliberately did **not** close: cgroup CPU *quota* (`docker --cpus=1.5`, `cpu.max`) is invisible to `sched_getaffinity` and `os.cpu_count()` alike, so a quota-limited (rather than cpuset-limited) container still sees the host count. Nothing in the repo reads cgroup limits today, and it affects every `os.cpu_count()` call site here (`proxy/server.py`, `cli/learn.py`, `memory/adapters/embedders.py`), so it feels like its own change rather than something to bolt onto this one. Happy to do it here if you'd rather it land together.

PR body also rewritten to satisfy the governance template.

JerrettDavis (approved):
Re-reviewed the changes since c8e8885. The revised helper now prefers process affinity, uses Python 3.13 process-aware counting as the portable second choice, and keeps a safe host-count fallback. The added regression directly models a 64-core host with a two-CPU affinity set and verifies both the helper and constructed ORT session. This resolves my prior correctness concern.

PR #2848, opened 2026-08-07
fix(proxy/anthropic): inject headroom_retrieve whenever a CCR marker is present, not only for new markers

abhay-codes07:
## Description

On a frozen-prefix turn that replays an existing `<<ccr:hash>>` marker, the proxy did not inject the `headroom_retrieve` tool, so the agent held a marker it could not redeem. When it tried, the Anthropic API rejected the whole request:

```text
API Error: 400 Tool reference 'headroom_retrieve' not found in available tools
```

This was a frequent, user-visible failure in Claude Code.

### Root cause

The sticky tool-injection gate in `handlers/anthropic.py` was driven by `has_new_ccr_markers(...)` -- markers created THIS turn only:

```python
has_new_compressed_content = has_new_ccr_markers(
    current_detected_hashes=injector.detected_hashes,
    previous_forwarded_messages=prefix_tracker.get_last_forwarded_messages(),
    provider="anthropic",
)
tools, ccr_tool_injected = apply_session_sticky_ccr_tool(
    ...,
    has_compressed_content_this_turn=has_new_compressed_content,
)
```

`apply_session_sticky_ccr_tool` returns early with `decision="skip"` for a session it considers fresh when `not has_compressed_content_this_turn`. A marker replayed from the frozen prefix is "historical" (already in `previous_forwarded_messages`), so `has_new_ccr_markers` returns `False`, and on a fresh session the tool is skipped even though the request carries a redeemable marker. The `SessionCcrTracker` is per-process, so every proxy restart makes live sessions look fresh again and re-arms the failure mid-conversation. Anything that instructs the model to retrieve later (a project instruction saying "call `headroom_retrieve` with the hash before asserting an exact value") lands on this path by construction.

### Fix

Drive the gate from `injector.has_compressed_content` -- whether the forwarded request carries ANY CCR marker, new or replayed -- instead of new-markers-only. `#1850` narrowed the first-time gate to new markers to avoid arming a session that never compressed, but a present marker means the session HAS compressed, and a replayed marker is exactly as unredeemable as a fresh one. Since a new marker is also a present marker, `has_new_compressed_content or injector.has_compressed_content` collapses to `injector.has_compressed_content`, so the now-redundant `has_new_ccr_markers` call is removed.

[...]

PR #2507, opened 2026-07-23
fix(proxy/anthropic): rescue sessions bricked by stale tool-search references

gglucass:
## Description

Anthropic's server-side tool search records `tool_reference` entries inside `tool_search_tool_result` blocks in the assistant turn. The client replays them verbatim on every subsequent request, and the API validates each replayed reference against the *current* request's tools array. If one referenced tool is missing, the request is rejected with `400 Tool reference '<name>' not found in available tools` - and since the client replays the same history every turn, the session is permanently bricked.

Headroom itself can strand such a reference: the proxy-injected CCR `headroom_retrieve` tool is visible to tool search, but injection can stop on a later turn - the frozen-prefix deferral in `should_inject_ccr_tool`, or `SessionCcrTracker` state lost to a proxy restart / LRU eviction - leaving the history referencing a tool that only headroom ever provided. References can also go stale through no fault of headroom (e.g. a client plugin tool flapping out of the tools array mid-session); today the proxy forwards those doomed requests as-is.

Observed in the wild on 2026-07-23: a live Claude Code session through headroom had `tool_references: ["headroom_retrieve", "mcp__headroom__headroom_compress", "mcp__headroom__headroom_retrieve"]` recorded in its transcript; once injection stopped, every request in that session failed with the 400 above until the transcript was hand-edited.

Closes #

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- New `headroom/proxy/tool_reference_sanitizer.py`: `sanitize_stale_tool_references()` drops replayed `tool_reference` entries whose tool is absent from the final tools array (copy-on-write; the input is returned unchanged, same object identity, when nothing is stale), and `collect_tool_reference_names()` gathers referenced names from replayed messages. Handles the GA wire shape (`content` is a dict holding `tool_references`) plus a list-shaped `content` defensively; malformed blocks and non-reference entries are left untouched.
[...]

PR #2889, opened 2026-08-10
fix(proxy/anthropic): run tool-search history repair after turn hooks

### LINKED ISSUE #2888: tool-search history repair runs before turn hooks, which may mutate the tools array
## Summary

`strip_unsupported_tool_search_blocks` (#2807) runs *before* the turn-hooks block in `headroom/proxy/handlers/anthropic.py`, but a registered turn hook may rewrite the outbound `tools` array. If a hook removes a tool that a replayed `tool_reference` names, the repair has already run against a stale view and upstream rejects the request with:

```text
400 Tool reference '<name>' not found in available tools
```

## Detail

The repair is invoked at the site tagged `router:tool_search_repair`, immediately after the tool-search deferral/injection block. Its correctness argument is that it validates the transcript against exactly the `tools` array upstream will see — which is true at that point in the pipeline, and stops being true as soon as a later stage mutates `tools`.

The turn-hooks block runs directly afterwards and is documented as "a registered hook may inspect or rewrite the outbound tools/messages before we send upstream". That is the mutation that can invalidate the repair.

## Impact

Latent. No turn hook ships in-tree today, so this cannot fire on a default install; it is reachable only by a third-party registered hook that drops tools. Filing so the ordering constraint is written down rather than rediscovered.

## Suggested fix

Move the `strip_unsupported_tool_search_blocks` call to after the turn-hooks block (and after any other stage that can mutate `tools`), so it validates against the final array. The function is copy-on-write and returns the original `messages` by identity when nothing is removed, so relocating it does not affect the main-loop no-op or the prompt-cache prefix.

A regression test would register a turn hook that removes a referenced tool and assert the forwarded body carries no unresolvable `tool_reference`.

## Context

Surfaced while confirming that #2807 and #2848 supersede #2507 (closed as such).

gglucass:
## Description

`strip_unsupported_tool_search_blocks` (#2807) validates every replayed `tool_reference` in the transcript against the request's `tools` array. It ran *before* the turn-hooks block in `handlers/anthropic.py`, and a registered turn hook may rewrite that array — the hook surface is documented as "a registered hook may inspect or rewrite the outbound tools/messages before we send upstream".

So a hook that drops a tool named by a replayed reference leaves the repair having validated against a stale view, and upstream rejects the request:

```text
400 Tool reference 'X' not found in available tools
```

The repair's correctness argument is that it validates against exactly the `tools` array upstream will see. That was true at the old call site and stopped being true one block later.

### Fix

Move the repair to after the turn-hooks block, so it is the last stage that can invalidate a reference:

- It still runs **after** the deferral injection, so the tool just injected counts as present — the main loop strips nothing and the frozen prefix stays byte-identical.
- Nothing past the new call site mutates `body["tools"]` on the outbound path. (The two later `continuation_body["tools"]` assignments build a *derived* body from the already-repaired `body`, so they inherit the repair.)
- It still runs before the consistency token re-count, so `tok_after` continues to reflect the repaired messages.
- It remains unconditional (not gated on `HEADROOM_TOOL_SEARCH`, not gated on `_bypass`), so transcripts poisoned before the flag was turned off still recover.

`strip_unsupported_tool_search_blocks` is copy-on-write and returns the original `messages` object by identity when nothing is removed, so relocating the call does not change the no-op path.

### Severity

Latent. No turn hook ships in-tree, so this cannot fire on a default install — it is reachable only through a third-party registered hook that shrinks the tools array. Filing the fix now so the ordering constraint is enforced by a test rather than rediscovered.

Closes #2888

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: the tool-search history repair block moves from just after the deferral injection to just after the turn-hooks block. The comment now states the ordering constraint in both directions (after injection, after hooks) so the next person to add a stage knows where the boundary is. No logic change.
- `tests/test_proxy/test_tool_search_repair_after_turn_hooks.py` (new): two handler-level regressions. Ordering is the whole property under test, so a unit test of the helper cannot see it — these drive the real handler through `TestClient` and assert on the forwarded body.

## Testing

- [ ] Manual testing performed — no in-tree turn hook exists to exercise this against a live API key; the handler-level test below is the substitute, see Not tested.

### Test Output

```text
$ uv run --extra dev pytest tests/test_proxy/ -q
======================= 241 passed, 1 warning in 35.82s ========================

$ uvx ruff check headroom tests
All checks passed!

$ uv run --extra dev mypy headroom
Success: no issues found in 515 source files
```

## Real Behavior Proof

- Environment: macOS 15 (darwin 24.6.0), Python 3.10.18, pytest 9.0.3, in a worktree off `upstream/main` at 2f2950a6. No live Anthropic key: the proxy handler is driven end to end through `fastapi.testclient.TestClient` with `_retry_request` stubbed, so the assertion is on the exact body that would have been sent upstream.
- Exact command / steps: (1) on the branch as submitted, `uv run --extra dev pytest tests/test_proxy/test_tool_search_repair_after_turn_hooks.py -q` -> 2 passed; (2) revert ONLY the handler ordering change while keeping the new tests, `git checkout HEAD~1 -- headroom/proxy/handlers/anthropic.py`, and re-run the same command. The request under test carries a `tool_search_tool_result` referencing `Grep`, a `tools` array containing `Grep`, and a registered turn hook that removes `Grep`.
- Observed result: with the fix reverted the primary test fails on exactly the shape upstream 400s on, because the forwarded body still carries a `tool_reference` naming a tool the turn hook had already removed from `tools`. Restoring the handler change turns it green. The second test passes in both states by design: it pins the converse (a hook that leaves `tools` alone must not cause over-stripping), so the fix cannot regress into "strip always". Verbatim output of the reverted run:

```text
$ git checkout HEAD~1 -- headroom/proxy/handlers/anthropic.py   # revert ONLY the ordering fix
$ uv run --extra dev pytest tests/test_proxy/test_tool_search_repair_after_turn_hooks.py -q
collected 2 items
tests/test_proxy/test_tool_search_repair_after_turn_hooks.py F.          [100%]
=================================== FAILURES ===================================
____________ test_repair_sees_the_tools_array_the_hook_left_behind _____________
tests/test_proxy/test_tool_search_repair_after_turn_hooks.py:166: in test_repair_sees_the_tools_array_the_hook_left_behind
    assert _referenced_tool_names(forwarded) == []
E   AssertionError: assert ['Grep'] == []
E
E     Left contains one more item: 'Grep'
FAILED tests/test_proxy/test_tool_search_repair_after_turn_hooks.py::test_repair_sees_the_tools_array_the_hook_left_behind
```

- Not tested: no live-API reproduction of the 400 itself, since triggering it needs a third-party turn hook that removes a tool and none ships in-tree (the assertion above is on the forwarded body, which is the input that produces the 400); no streaming-path variant, since the repair mutates `body` upstream of the stream/buffered split so both inherit it but only the buffered path is asserted; no performance measurement, since the change moves an existing call ~60 lines later in the same function and adds no work.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A, no user-facing or configuration surface changes

## Additional Notes

- **CI:** `test (4)` failed on `tests/test_tokenizer_count_offload.py::test_count_tokens_offloaded_keeps_loop_responsive` with `assert 0 >= 5`. That is an event-loop-responsiveness timing assertion under a shared runner, and it is unrelated to this diff — nothing here touches the tokenizer or the offload path. It passes locally (`10 passed in 1.43s`). I do not have rerun permission on this fork PR (`gh run rerun` → `cannot be rerun`), so a maintainer rerun is needed to clear it.
- **Codecov:** reports "All modified and coverable lines are covered by tests". The accompanying warning is the repo-level "install the Codecov app" notice, not a finding against this PR.
- Surfaced while confirming that #2807 and #2848 supersede #2507, which is now closed as such.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -2477,35 +2477,6 @@ class AnthropicHandlerMixin:
                         f"{_ts_saved_tokens}tok"
                     )
 
-            # Tool-search history repair (#2805). Once deferral is on, the client
-            # stores Anthropic's server_tool_use / tool_search_tool_result blocks in
-            # its transcript forever, and upstream validates every tool_reference in
-            # that history against THIS request's tools array. Claude Code replays
-            # the same transcript on side-requests carrying a different, smaller
-            # tools array (the prompt-type Stop hook evaluator, /compact), which the
-            # proxy cannot predict — so upstream 400s with "Tool reference 'X' not
-            # found in available tools". Drop the blocks such a request cannot
-            # support. Runs AFTER the injection above so the tool we just added
-            # counts as present: on the main loop nothing is stripped and the prefix
-            # is untouched. Unconditional (not gated on the flag) so transcripts
-            # poisoned before the flag was turned off still recover.
-            from headroom.proxy.helpers import strip_unsupported_tool_search_blocks
-
-            _ts_repaired, _ts_stripped = strip_unsupported_tool_search_blocks(
-                body.get("messages"), body.get("tools")
-            )
-            if _ts_stripped:
-                body["messages"] = _ts_repaired
-                optimized_messages = _ts_repaired
-                body_mutation_tracker.mark_mutated("tool_search_history_repair")
-                transforms_applied.append(f"router:tool_search_repair:{_ts_stripped}blocks")
-                logger.info(
-                    "[%s] Tool search: dropped %d unsupportable history block(s) "
-                    "(tools array cannot resolve their tool_reference entries)",
-                    request_id,
-                    _ts_stripped,
-                )
-
             # Turn hooks (opt-in extensions): a registered hook may inspect or
             # rewrite the outbound tools/messages before we send upstream — the
             # extensible counterpart to the built-in deferral above. A single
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
The relocation preserves the existing repair behavior while making it validate against the final post-hook tools array. The handler-level tests cover both the stale-reference failure and the no-op/converse path, the description accurately scopes the latent third-party-hook case, the branch is mergeable, and the current checks are green.

chopratejas (approved):
Deep maintainer review complete. The repair now validates replayed tool references against the final post-hook tools array, while retaining the required post-deferral ordering. The handler-level regressions prove both stale-reference removal and the no-op path. Full CI is green.

PR #2890, opened 2026-08-10
fix(backends/anyllm): stream tool_use blocks and map finish_reason on the streaming path

abhay-codes07:
## Description

On the any-llm backend (`--backend anyllm`), the streaming path drops tool calls entirely.

`stream_message` forwards `tools` / `tool_choice` upstream:

```python
if "tools" in body:
    kwargs["tools"] = body["tools"]
if "tool_choice" in body:
    kwargs["tool_choice"] = body["tool_choice"]
```

but its event loop only ever inspected `delta.content`. It pre-opened a single text block at index 0 and never looked at `delta.tool_calls`:

```python
yield StreamEvent(event_type="content_block_start", data={... "index": 0, "content_block": {"type": "text", ...}})
...
async for chunk in stream_response:
    if hasattr(chunk, "choices") and chunk.choices:
        delta = chunk.choices[0].delta
        if hasattr(delta, "content") and delta.content:
            ...  # text only
```

So when the model streamed a tool call back, the client (Claude Code and other agent harnesses) saw an empty assistant turn -- no `tool_use` block -- and, because `stop_reason` was hardcoded to `end_turn`, no signal that a tool was even called. The agent loop stalls with nothing to execute. `stop_reason` being hardcoded also mislabels a `length` truncation as `end_turn`.

The non-streaming path on this same backend (`_to_anthropic_response`) already handles tool calls and maps `finish_reason` correctly; only the streaming path was missing it.

## Fix

Open content blocks lazily as deltas arrive (so a response can carry `tool_use` blocks, not just one pre-opened text block) and handle `delta.tool_calls`:

- Emit a `tool_use` `content_block_start` per provider tool index, carrying the tool `id`/`name`.
- Stream the arguments as `input_json_delta` deltas.
- Derive `stop_reason` from the OpenAI `finish_reason` (`tool_calls` -> `tool_use`, `length` -> `max_tokens`, `stop` -> `end_turn`).

This mirrors the non-streaming `_to_anthropic_response` on the same backend and the LiteLLM streamer, so the transports agree. The text-only path is unchanged in output: the first text delta opens the block lazily at index 0, producing the same event sequence as before.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/backends/anyllm.py` (`stream_message`): lazy block opening + `tool_calls` handling (tool_use `content_block_start` / `input_json_delta`) + `finish_reason` -> `stop_reason` mapping. Removes the unconditional pre-opened text block.
- `tests/test_backend_anyllm.py`: added `test_stream_message_emits_tool_use_blocks` (a streamed tool call surfaces as a tool_use block with reassembled arguments and `stop_reason=tool_use`) and `test_stream_message_maps_length_finish_reason` (`length` -> `max_tokens`). The existing text-only streaming test is unchanged and still passes.

## Testing

- [x] New tests added for the tool-call and finish_reason behaviour

### Test Output

```text
# Fail-before (original source, new tests kept):
test_stream_message_emits_tool_use_blocks
  -> event sequence diff at index 2: 'content_block_stop' != 'content_block_delta'  (tool call dropped)
test_stream_message_maps_length_finish_reason
  -> assert 'end_turn' == 'max_tokens'  (hardcoded stop_reason)

# Pass-after (fix in place):
tests/test_backend_anyllm.py  18 passed

# uvx mypy@1.20.2 headroom/backends/anyllm.py -> Success: no issues found in 1 source file
```

## Real Behavior Proof

- Steps: drove `stream_message` with a fake any-llm async stream yielding OpenAI-style `delta.tool_calls` chunks (id + name, then two argument fragments) followed by a `finish_reason=tool_calls` chunk, and collected the emitted `StreamEvent`s. Before the fix the tool call produced no `tool_use` block and the turn ended `end_turn`; after the fix it emits `content_block_start{type:tool_use,id,name}`, two `input_json_delta` deltas that reassemble to `{"city":"paris"}`, a `content_block_stop`, and `message_delta` with `stop_reason=tool_use`. A separate text stream with `finish_reason=length` now reports `stop_reason=max_tokens`.
- Observed result: an any-llm streaming turn that calls a tool now reaches the client as a valid Anthropic `tool_use` block instead of an empty turn, so agent loops using this backend can execute the tool.
- Not tested: a live any-llm provider call (no third-party keys here). The conversion is exercised directly against the streaming shape any-llm normalizes to (OpenAI-compatible `choices[0].delta.tool_calls`).

## Review Readiness

## Checklist

## Additional Notes

Output-token accounting on this path is still the per-delta chunk count (unchanged here); that is a separate concern from the dropped tool calls and is left as-is to keep this fix focused.

### DIFF
--- a/headroom/backends/anyllm.py
+++ b/headroom/backends/anyllm.py
@@ -326,16 +326,22 @@ class AnyLLMBackend(Backend):
 
             stream_response = await self.llm.acompletion(**kwargs)
             output_tokens = 0
-            # Open content blocks lazily as deltas arrive so a response can carry
-            # tool_use blocks (not just text). The previous version pre-opened a
-            # single text block at index 0 and only inspected ``delta.content``,
-            # so a tool call streamed back over any-llm was silently dropped even
-            # though ``tools``/``tool_choice`` are forwarded above — the client
-            # saw an empty turn and the agent loop stalled. Mirrors the
-            # non-streaming ``_to_anthropic_response`` and the LiteLLM streamer.
+            # Stream text immediately in a single text block, but BUFFER tool
+            # calls and emit them as complete blocks at the end. OpenAI streams
+            # parallel tool calls interleaved by index (index 0 and 1 introduced
+            # together, then a fragment for 0, then for 1), while Anthropic
+            # requires each content block to be fully emitted — start, deltas,
+            # stop — before the next opens. Reassembling per index and flushing
+            # complete blocks keeps every delta inside its own block's start/stop
+            # for any interleaving. (The previous version pre-opened one text
+            # block and dropped tool calls entirely; a naive open-on-new-index
+            # instead mis-sequenced parallel calls, emitting a fragment for an
+            # already-stopped block.)
             current_block_index = -1
-            active_block_type: str | None = None  # "text" or "tool_use"
-            tool_block_map: dict[int, int] = {}  # provider tc.index -> SSE block index
+            text_block_open = False
+            # provider tool index -> {"id", "name", "arguments"}, first-seen order
+            tool_calls: dict[int, dict[str, Any]] = {}
+            tool_order: list[int] = []
             stop_reason = "end_turn"
 
             async for chunk in cast(AsyncIterator[Any], stream_response):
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The single-tool path is well covered, but the current block lifecycle breaks parallel tool calls. When a chunk first introduces index 0 and index 1, opening index 1 closes index 0. On the next provider chunk, an argument fragment for index 0 is still emitted with `index: tool_block_map[0]`, even though that Anthropic content block has already received `content_block_stop`; the same issue occurs whenever fragments for multiple tool indices are interleaved. That produces an invalid Anthropic event sequence for a standard OpenAI parallel-tool stream.

Please buffer/reassemble provider tool-call state (or otherwise serialize complete Anthropic tool blocks) so every delta is emitted between that block’s start and stop, and add a regression test with two tool-call indices whose argument fragments arrive across at least two chunks. The test should assert each index has exactly one start/stop and no delta occurs after its stop.

abhay-codes07:
Good catch, thanks. Fixed in 72a2be9b.

The block lifecycle now buffers tool-call fragments per provider index (id, name, accumulated arguments) during the stream and flushes each as a complete, self-contained block after the loop: `content_block_start`, one `input_json_delta` with the reassembled arguments, `content_block_stop`. Since nothing is emitted for a tool block until the stream is fully drained, every delta falls inside its own block's start/stop regardless of how the provider interleaves indices, so the parallel-stream sequence you described can no longer produce a delta after a stop. Text still streams incrementally in a single leading block.

Added `test_stream_message_handles_parallel_tool_calls`: two tool indices introduced in one chunk, then argument fragments for each across later chunks. It asserts each block index has exactly one start and one stop, the arguments reassemble per index (`{"a":1}` / `{"b":2}`), and no `content_block_delta` occurs after that index's `content_block_stop`. I confirmed it fails on the previous head with exactly `delta for block 0 after its stop`, and passes now. The single-tool test was updated since arguments now arrive as one reassembled delta. 19 tests pass, ruff and mypy clean.

JerrettDavis (approved):
The revised implementation resolves the requested parallel-call lifecycle issue: it accumulates state by provider tool index, preserves first-seen order, and emits each Anthropic tool block as a complete start/delta/stop sequence. The new two-index interleaving regression directly covers the previously invalid shape. Finish-reason mapping remains covered, and all current checks are green.

PR #2891, opened 2026-08-10
fix(telemetry): stop pretty-printing and pre-materializing TOIN saves

### LINKED ISSUE #2886: TOIN store grows without bound and re-serializes 14.5 GB every 10 minutes
## Summary

`ToolIntelligenceNetwork._patterns` has no eviction, and `FileSystemTOINBackend.save` serializes the whole store as one `json.dumps(..., indent=2)` string in memory. On a long-lived proxy this compounds until the autosave itself becomes the dominant cost.

Observed on one host after ~2 months of continuous use:

```
-rw-------  14,485,506,631  ~/.headroom/toin.json      # 14.5 GB
```

- proxy RSS **15 GB**, `sample` reported physical footprint peak **42 GB**
- swap **20.97 GB of 22.5 GB** used, 4.7M swapouts
- `TOIN: 278704 patterns, 733924 compressions, 111 retrievals, 0.0% retrieval rate`

A `sample` of the proxy during autosave shows the encoder recursion:

```
encoder_listencode_obj  (in _json.cpython-313-darwin.so)
  encoder_encode_key_value
    encoder_listencode_obj
      _steal_accumulate / _PyUnicodeWriter_WriteStr
```

## Why it grows unbounded

Every inner field is carefully bounded — `strategy_success_rates` to 20, `field_semantics` to 100, `commonly_retrieved_fields` to 20, `query_pattern_frequency` to `max_query_patterns * 2`, `_seen_instance_hashes` to 100. The `_patterns` dict that holds them is not:

```python
if key not in self._patterns:
    self._patterns[key] = ToolPattern(...)
```

The key is `(auth_mode, model_family, tool_signature.structure_hash)`. `structure_hash` is content-shaped, so a proxy seeing varied tool output mints new keys indefinitely. On the host above the distribution was:

```
patterns=278,767   >=50 observations: 820 (0.29%)   <=2 observations: 237,217 (85.1%)   mean 2.63
```

85% of the store is patterns seen twice or less, and 99.71% sit below `DEFAULT_MIN_OBSERVATIONS_TO_PUBLISH = 50`, so they can never produce a `toin publish` recommendation row. The bytes are almost entirely noise that will never be used.

## Why save is expensive

`FileSystemTOINBackend.save` materializes the complete JSON text before it opens the temp file:

```python
json_data = json.dumps(data, indent=2)
fd, tmp_path = tempfile.mkstemp(dir=self._path.parent, prefix=".toin_", suffix=".tmp")
```

With `auto_save_interval = 600` that is a full serialization of the entire store every 10 minutes, on top of `export_patterns()` having already built a dict of every `ToolPattern.to_dict()`. `indent=2` roughly doubles the output for no machine consumer. `_load_from_backend` pays the mirror-image cost at every proxy start.

## Suggested directions

Happy to implement whichever you prefer — flagging first since eviction policy is a product decision, not a mechanical fix.

1. **Bound `_patterns`.** Evict by `last_updated` / `sample_size` at a configurable ceiling. Patterns under `DEFAULT_MIN_OBSERVATIONS_TO_PUBLISH` that have not been touched in N days cannot become recommendations, so they are the natural first eviction class.
2. **Stop pretty-printing.** `indent=2` on a machine-read artifact is pure size. Dropping it is a one-liner.
3. **Stream the write.** `json.dump(data, f)` into the temp file instead of building the string first would cut the peak footprint substantially even before eviction lands.
4. **Prune on load.** Even without runtime eviction, dropping sub-threshold stale patterns during `_load_from_backend` would stop a store from carrying old noise forward across restarts.

(2) and (3) are safe mechanical wins and I can send them as a small PR immediately if useful. (1) and (4) need a policy call from you.

## Environment

macOS 27.0 arm64, Python 3.13, headroom-ai 0.34.0, filesystem TOIN backend at the default path. The same code is present on `main` (2f2950a) — `headroom/telemetry/toin.py` and `headroom/telemetry/backends/filesystem.py` are byte-identical between the 0.34.0 wheel and `main`.

Related: #2884 (`HEADROOM_TOIN_BACKEND=none` not honored) — that is the escape hatch an operator would reach for when hitting this, which is how I found it.

ashishpatel26:
## Description

`FileSystemTOINBackend.save()` built the full JSON string in memory via `json.dumps(data, indent=2)` before writing to the temp file. On a long-lived proxy the TOIN store can grow into the tens of GB (reporter observed 14.5GB after ~2 months of continuous use, 42GB peak physical footprint during autosave), and this pattern doubles both peak memory and on-disk size for every autosave (every 10 minutes by default) — for a file with no human reader.

Closes #2886

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `headroom/telemetry/backends/filesystem.py` (`FileSystemTOINBackend.save`): drop `indent=2` (compact JSON — no machine consumer needs it pretty-printed) and stream directly into the temp file via `json.dump(data, f)` instead of building the full string first via `json.dumps(...)` then `f.write(...)`.
- `tests/test_adapter_hooks.py`: new regression test asserting `save()` output is single-line compact JSON, not pretty-printed.

These are the two "safe mechanical" fixes the reporter explicitly flagged as not requiring a policy call — the reporter also suggested bounding `_patterns` (eviction) and pruning stale patterns on load, both left as separate follow-ups since they're product/policy decisions, not mechanical ones.

## Testing

- [ ] Linting passes (`ruff check .`) — not run locally, will confirm via CI
- [ ] Type checking passes (`mypy headroom`) — not run locally, will confirm via CI

### Test Output

```text
$ .venv/Scripts/python -m pytest tests/test_adapter_hooks.py -q
30 passed, 2 failed (pre-existing, unrelated — see below)

$ .venv/Scripts/python -m pytest tests/test_toin*.py tests/test_stateless_toin.py tests/test_compression_policy_toin_gate.py tests/test_smart_crusher_toin_attachment.py -q
109 passed, 2 failed (pre-existing, unrelated), 22 skipped
```

The 4 failures (`TestStorageEntryPointLoading::test_sqlite_scheme`, `::test_jsonl_scheme`, `TestTOINDefaultStoragePath::test_toin_default_storage_path_exists`, `::test_empty_env_var_uses_default`) reproduce identically on a clean, unmodified `main` — confirmed via `git stash`. They're Windows-specific path-separator/URI-parsing issues in unrelated storage backends, not caused by this change.

## Real Behavior Proof

- Environment: Windows 11, Python 3.14.5, local venv
- Exact command / steps: built a synthetic TOIN store matching the issue's shape (20,000 patterns, each with a nested `strategy_success_rates` dict) and compared `json.dumps(data, indent=2)` size against `FileSystemTOINBackend.save()`'s actual on-disk output, then reloaded and diffed for correctness
- Observed result: compact output was 31.4% smaller than the old pretty-printed output (4,846,712 bytes → 3,326,702 bytes) and reloaded byte-identical to the original data — `old (indent=2): 4,846,712 bytes`, `new (compact): 3,326,702 bytes`, `reduction: 31.4%`, `round-trip correct: True`
- Not tested: have not reproduced the reporter's actual 14.5GB store or measured real peak-RSS during autosave on a long-lived process — that requires the exact multi-month usage pattern from the issue, which isn't reproducible in this session. The size-reduction and streaming-write mechanism are verified directly at the unit level instead.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A — internal storage-format detail, not user-facing)

## Additional Notes

Scope note: this PR intentionally does **not** address the unbounded `_patterns` dict growth itself (the reporter's suggestions 1 and 4) — that needs an eviction policy decision (by `last_updated`? `sample_size`? a day-based staleness window?) that's a product call, not something to guess at in a mechanical PR. Happy to take that on as a follow-up once there's a steer on the policy.

### DIFF
--- a/headroom/telemetry/backends/filesystem.py
+++ b/headroom/telemetry/backends/filesystem.py
@@ -60,12 +60,16 @@ class FileSystemTOINBackend:
         try:
             self._path.parent.mkdir(parents=True, exist_ok=True)
 
-            json_data = json.dumps(data, indent=2)
-
             fd, tmp_path = tempfile.mkstemp(dir=self._path.parent, prefix=".toin_", suffix=".tmp")
             try:
+                # Stream directly into the temp file rather than building the
+                # full JSON string in memory first (json.dumps(...) then
+                # f.write(...)) — on a large store that doubled peak RSS
+                # during every autosave. No `indent`: this file has no human
+                # reader, and pretty-printing roughly doubles its size for a
+                # multi-GB store (issue #2886).
                 with open(fd, "w") as f:
-                    f.write(json_data)
+                    json.dump(data, f)
                 Path(tmp_path).replace(self._path)
             except Exception:
                 try:
--- a/tests/test_adapter_hooks.py
+++ b/tests/test_adapter_hooks.py
@@ -148,6 +148,18 @@ class TestTOINBackendProtocol:
         assert loaded == {"version": "2.0", "new": True}
         assert "old" not in loaded
 
+    def test_save_writes_compact_json_not_pretty_printed(self, fs_backend, tmp_toin_path):
+        """Regression for issue #2886: pretty-printing (indent=2) roughly
+        doubles the on-disk size of a store that has no human reader and
+        can grow into the tens of GB. save() must write compact JSON —
+        one line, no indentation.
+        """
+        fs_backend.save({"version": "1.0", "patterns": {"a": 1, "b": 2}})
+        with open(tmp_toin_path) as f:
+            raw = f.read()
+        assert "\n" not in raw
+        assert json.loads(raw) == {"version": "1.0", "patterns": {"a": 1, "b": 2}}
+
 
 class TestCustomTOINBackend:
     """Verify any dict-based backend satisfies the protocol."""


JerrettDavis (approved):
`json.dump` streams encoder chunks directly to the already-atomic temporary file and compact encoding removes the avoidable indentation overhead without changing the loaded data contract. The regression pins compact valid JSON, while the existing overwrite/round-trip coverage protects backend behavior. All current checks are green.

ashishpatel26:
Closing as superseded — this exact fix (compact JSON + streaming write) landed on main as part of commit 8cd13803 ("fix(toin): bound private query and pattern retention"), which also addresses the retention/eviction follow-up (issue #2886's items 1 and 4) that I'd deliberately left out of scope here as a separate policy decision.

Thanks @JerrettDavis for the review — appreciated, even though it's moot now that main has moved past this.

PR #2900, opened 2026-08-10
fix(openclaw): persist context engine compaction

### LINKED ISSUE #2893: OpenClaw compact reports success without compacting or persisting the session
### Summary

The OpenClaw `ContextEngine` advertises `ownsCompaction=true`, but its `compact()` implementation does not read, compress, or write the supplied `sessionFile`. It increments `stats.compactions` and returns `{ ok: true, compacted: true }` even though the compaction work is left as a TODO.

The `assemble()` implementation also appears to promise RollingWindow/history dropping, but the Python pipeline has retired `RollingWindow`, and `/v1/compress` currently uses `token_budget` only as a context-limit input.

### Reproduction

1. Invoke:

```text
compact({ sessionFile: <path>, tokenBudget: 1000 })
```

2. Compare the `sessionFile` before and after the call.
3. Invoke `assemble()` with a conversation whose token count exceeds the supplied budget.

### Expected behavior

* `compact()` should either:

  * perform the documented compaction and persist the resulting state to `sessionFile`, or
  * truthfully report that the compaction work is deferred/not implemented.
* `tokenBudget` should enforce the documented context-limit contract.
* `assemble()` should drop or otherwise compact old conversation history when the conversation exceeds the applicable budget.

### Actual behavior

* `sessionFile` remains unchanged after `compact()`.
* `compact()` reports `compacted: true` despite performing no compaction.
* `stats.compactions` is incremented even though no state was compacted.
* Old conversation messages are not dropped when the conversation exceeds the budget.
* The current Python compression pipeline does not appear to implement the RollingWindow behavior described by the OpenClaw `assemble()` comments.

### Relevant code

* `plugins/openclaw/src/engine.ts:23`
* `plugins/openclaw/src/engine.ts:104`
* `plugins/openclaw/src/engine.ts:152`
* `plugins/openclaw/src/engine.ts:177-192`
* `headroom/proxy/handlers/openai.py:8834-8860`
* `headroom/transforms/pipeline.py:95`

### Related issue

Open issue [[#1266](https://github.com/headroomlabs-ai/headroom/issues/1266)](https://github.com/headroomlabs-ai/headroom/issues/1266) covers the broader OpenClaw no-compression/stats problem, but I could not find an existing issue or PR covering this specific `compact()` hook no-op and its incorrect success reporting.

### Impact

This can cause OpenClaw to believe that compaction has occurred when the underlying session state has not changed. As a result, callers relying on `compacted=true` or `stats.compactions` may incorrectly assume that context reduction and persistence have taken place.

### Suggested fix

Either implement the `compact()` hook so that it actually performs and persists the documented compaction behavior, including enforcing `tokenBudget`, or change the result/statistics reporting so that deferred/unimplemented compaction is not reported as successfully completed.

abhinavkr26104:
## Description

Fixes #2893

OpenClaw context engines must not replace the append-only session transcript themselves. The engine now declares `ownsCompaction=false` and delegates `compact()` to OpenClaw's public `delegateCompactionToRuntime()` helper whenever the current runtime does not provide transcript-maintenance rewrites. This keeps /compact and overflow-recovery compaction on the host-owned persistence path.

Headroom's guarded transcript compression is also exposed through `maintain()`, where current OpenClaw supplies `runtimeContext.rewriteTranscriptEntries()`. The host serializes that rewrite with concurrent transcript appends. No plugin-owned temporary file or direct file replacement is used.

Malformed transcripts, no-op compression, message-count changes, runtime rewrite failures, concurrent appends, and concurrent compaction attempts fail closed without unsafe plugin-side persistence.

## Type of Change

- [ ] Breaking change (bug fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added the OpenClaw public compaction delegate as an optional peer-loaded host integration.
- Set `ownsCompaction=false` so OpenClaw remains responsible for durable compaction transactions.
- Added the supported `maintain()` lifecycle path for serialized Headroom transcript rewrites.
- Added regression coverage for delegation, persistence safety, metadata preservation, malformed input, no-op compression, unsafe message-count changes, concurrent appends, permissions, rewrite failures, and concurrent compaction.

## Testing

- [x] Unit tests pass (npm test in plugins/openclaw: 86 passed)
- [ ] Linting passes (ruff check .)
- [ ] Type checking passes (mypy headroom)

### Test Output

    plugins/openclaw: npm test
    Test Files  6 passed (6)
    Tests  86 passed (86)

    plugins/openclaw: npm run typecheck
    tsc --noEmit passed

    plugins/openclaw: npm run build
    tsup and prepare-dist passed

    git diff --check passed

    Full Python pytest was not rerun; the prior Windows checkout could not build the Rust headroom._core extension and lacked tomlkit. ruff and mypy results remain as previously reported.

## Real Behavior Proof

- Environment: Windows PowerShell, Node 22.14.0, npm 11.2.0, Python 3.12.
- Exact command / steps: Ran the OpenClaw plugin unit suite, typecheck, and production build against the PR head plus commit `4c2167af`.
- Observed result: 6 test files and 86 tests passed; the delegation contract path, host-owned maintenance rewrite path, malformed/no-op/count-change/concurrency/failure regressions, typecheck, and build all passed.
- Not tested: A live OpenClaw Gateway or production Headroom proxy.

## Review Readiness

## Checklist

- [x] New and existing plugin unit tests pass locally
- [x] I did not edit CHANGELOG.md; it is generated by release-please

## Additional Notes

The plugin never rewrites the transcript directly. When OpenClaw provides its serialized `maintain()` rewrite capability, Headroom only replaces stable message payloads through that host API; otherwise `compact()` delegates to OpenClaw's native compaction implementation.


JerrettDavis (dismissed):
The current read/transform/rename sequence can silently lose transcript data. OpenClaw’s JSONL is described here as append-only, but this code reads a snapshot, awaits a network compression call, then replaces the live path. Any record appended during that interval is absent from `outputLines` and is discarded by `rename`. Please coordinate with OpenClaw’s session-file lock if one is available, or implement a compare/retry/fail-closed protocol that proves the source has not changed immediately before replacement (and add a regression that appends while `compress()` is pending and verifies no record is lost).

The replacement also creates a new file with default mode instead of preserving the source mode; a 0600 transcript can become 0644 depending on umask. Use a securely created unique temp file in the same directory, preserve the original mode (and any other required metadata), then atomically replace it. Add a permissions regression alongside the concurrent-append case. Until both are covered, compaction risks losing or exposing enterprise session data.

chopratejas (changes requested):
The data-loss and permissions hazards are fixed on the latest head: Headroom no longer writes or replaces the transcript directly and the new regressions prove fail-closed behavior. One compatibility blocker remains before this can close #2893: current OpenClaw exposes its serialized rewrite capability to maintain(), not compact(), so the normal compact call receives no rewriteTranscriptEntries function and this implementation correctly returns compacted=false. Wire the host-owned rewrite capability into compact(), or move this behavior into the supported maintain lifecycle, and add an integration test against the real host contract. Do not restore direct file replacement.

JerrettDavis (changes requested):
The unsafe direct file replacement is gone, but this revision now depends on a `runtimeContext.rewriteTranscriptEntries` host capability that is not implemented or integrated anywhere in this PR (the only non-test occurrence is the optional callback read in `engine.ts`). The tests manufacture that callback themselves, so they do not prove that a supported OpenClaw version ever supplies it. On an actual host that does not know this new contract, every compaction now fails closed and `ownsCompaction: true` still advertises a capability that cannot work.

Please either wire this to a real, versioned OpenClaw host API and add an integration/contract test using the host-provided runtime context, or gate/disable `ownsCompaction` until that API exists. Also update the PR description and behavior proof: they still claim the plugin atomically replaces the JSONL file and report a successful test without the runtime callback, which no longer describes the current head. The concurrency and mode-preservation regressions are useful, but their test-only rewrite helper cannot establish production availability of the capability.

JerrettDavis (changes requested):
Re-reviewed exact head 34e1d209 after the main sync. The production contract blocker remains unchanged: ownsCompaction is still advertised, while rewriteTranscriptEntries exists only as an optional callback consumed by engine.ts and manufactured by tests; no supported host integration supplies it. Please resolve the previously requested host capability integration/gating and refresh the behavior proof.

abhinavkr26104:
Resolved the host-contract blocker. The engine now declares ownsCompaction=false and delegates compact() to OpenClaw's public delegateCompactionToRuntime() whenever the runtime does not provide a rewrite callback, so /compact and overflow recovery remain functional on current OpenClaw. Headroom's guarded transcript transform is also exposed through maintain(), where OpenClaw supplies rewriteTranscriptEntries and serializes writes with appends. The direct file replacement path is never used. Added the optional SDK declaration, a delegation contract regression, and kept malformed/no-op/count-change/concurrent-append/permission/failure coverage. Verified locally: plugins/openclaw npm test (86 passed), npm run typecheck, npm run build, and git diff --check. Let me know if you need any additional host-version integration coverage or a behavior-proof artifact.

JerrettDavis (changes requested):
The upstream APIs now exist, but maintain() cannot safely be implemented as an unconditional alias of compact(). OpenClaw invokes maintenance after bootstrap, successful turns, and compaction. If runtimeContext.rewriteTranscriptEntries is absent for any lifecycle/version/runtime, this alias enters compact()'s fallback and calls delegateCompactionToRuntime, potentially running a full built-in compaction during ordinary post-turn maintenance. The intended fallback delegation belongs only to an actual /compact/overflow compact() call; maintenance without the rewrite capability should no-op/fail closed, never initiate host compaction. Please split the guarded rewrite operation from compact(): maintain() should use it only when the host rewrite seam is present, while compact() delegates when it is absent. Add a test calling maintain() without runtime context and assert the delegate is not invoked.

The production contract also remains unversioned. peerDependencies.openclaw is *, the manifest declares no compat.minGatewayVersion/plugin API floor, and tests supply both host APIs through a virtual: true mock plus a handwritten declaration. Older accepted hosts can therefore load this build but fail when the new SDK export or maintenance context is needed. Declare the minimum OpenClaw/plugin API version that provides delegateCompactionToRuntime, acceptedHostParams, maintain, and rewriteTranscriptEntries, narrow the peer range accordingly, and add a contract/type/build test against that real pinned host package rather than only the virtual module. The official OpenClaw context-engine documentation confirms these are supported current APIs; the blocker is lifecycle separation and enforceable compatibility, not the earlier direct-file approach.

PR #2908, opened 2026-08-10
fix(ccr): verify a scanned marker's hash before advertising it

### LINKED ISSUE #2836: [BUG] CCR marker scanner adopts foreign compression markers by shape, injecting an unredeemable retrieve tool
## Description

CCR decides whether a compression marker is Headroom's own by its *shape*. Any bracket marker carrying a 24-hex hash counts, so a marker emitted by a different context tool is adopted too. The generic fallback pattern is the widest door:

```python
# headroom/ccr/tool_injection.py
re.compile(r"\[.*?compressed.*?hash=([a-f0-9]{24})\]", re.IGNORECASE),
```

Other context tools emit exactly that shape. Once a foreign hash is scanned, `CCRToolInjector` reports compressed content, `has_new_ccr_markers` returns True, and the sticky `headroom_retrieve` tool plus the "Available hashes" system instruction go into the request. The model is told `headroom_retrieve` is the only tool that expands markers containing `hash=`, so it calls it with a hash this proxy never stored, gets a miss, and follows the recovery hint by re-running the command it already ran. Two wasted turns, and the instruction advertises hashes that cannot be redeemed.

## To Reproduce

1. Install headroom 0.34.0.
2. Run the code below, which feeds the scanner a marker produced by another context tool.
3. Watch a hash Headroom never created get adopted.

## Expected Behavior

Only hashes Headroom actually created drive retrieve-tool injection. A marker this proxy did not emit is ignored, whatever its format.

## Actual Behavior

The foreign hash lands in `detected_hashes`, `has_compressed_content` is True, and `process_request` injects the retrieve tool.

## Code Sample

```python
from headroom.ccr.tool_injection import CCRToolInjector

# Marker emitted by a different context tool, copied verbatim.
foreign = (
    "[374 items compressed to 267 (from 65 source lines). "
    "Retrieve more: hash=ddc3d69afad7bc53fbee11e2]"
)

injector = CCRToolInjector(provider="anthropic")
injector.scan_for_markers([{"role": "user", "content": foreign}])

print(injector.has_compressed_content)  # True  -- expected False
print(injector.detected_hashes)         # ['ddc3d69afad7bc53fbee11e2']
```

## Error Output

What the model gets back after it redeems the adopted hash:

```
{
  "error": "Content not found. It may have expired or the hash may be incorrect.",
  "hash": "ddc3d69afad7bc53fbee11e2",
  "hint": "To recover: if the compression marker references a file Read, re-read that file (the path is in the marker; disk is the source of truth). If it was command output, re-run the command."
}
```

## Environment

- **Headroom version**: 0.34.0
- **Python version**: 3.13.14
- **OS**: Arch Linux (Linux 7.1.5)
- **LLM Provider**: Anthropic

## Additional Context

**How much this bites today.** I found it by reading code while chasing an unrelated question, not from a user report. Across 2876 local Claude Code transcripts I counted 27 foreign markers and **zero** `headroom_retrieve` calls carrying a foreign hash. It is latent here, not an active outage. Two reasons I still think it is worth closing: the marker format is public and unversioned, so the collision costs nothing to trigger on purpose; and the failure is quiet, because the proxy hands the model a handle, the model spends a turn redeeming it, and the hint sends it back to re-run work it already did.

**A fix I built and then backed away from.** I gated detection on the compression store instead of the regex: keep a hash only when `get_entry_status()` recognizes it. It works, and the tests pass. I am not opening it as a PR, because it rests on an assumption I cannot make from inside the proxy.

Treating a `"missing"` answer as "not ours" is only sound when the store is one every process can see. Under `HEADROOM_CCR_BACKEND=memory` another worker may hold the entry, so I fell back to trusting `"missing"` only from a non-in-memory backend. Two reviewers took that apart, and they were right: multi-pod SQLite on local disk is still classified as shared while being as isolated as the in-process dict, entry-point adapters are assumed shared with nothing enforcing it, and a wrapper around `InMemoryBackend` is misclassified. When that verdict is wrong the proxy suppresses the retrieve tool for a marker that was still redeemable, which trades a rare bug for a quieter one. Whether "SQLite means shared" holds is a fact about your deployment topology, not mine to assume. It also puts a synchronous store lookup on the scan path, once per matched hash.

**What I would suggest instead.** Make Headroom's own markers self-identifying, so ownership is readable from the marker itself: namespace the hash (`hash=hr_<hex>`) or the marker, and treat anything unnamespaced as foreign after a release of grandfathering. That needs no store lookup, no hot-path I/O, and no assumption about how the store is deployed, and it stays correct when the next tool picks the same bracket syntax. It is a wire-format change though, which is your call rather than mine — hence an issue and not a patch.

Happy to open the store-gating version anyway if you would rather have that shape, or to implement the namespacing if you tell me which form you want.

ashishpatel26:
## Description

`CCRToolInjector.scan_for_markers()` decides whether a compression marker is Headroom's own by *shape* alone — any bracket marker carrying a 24-hex hash counts, per the generic fallback pattern (`\[.*?compressed.*?hash=([a-f0-9]{24})\]`). Other context tools emit exactly that shape. Once a foreign hash is scanned, `has_compressed_content` flips true and the retrieve tool + "Available hashes" system instruction get injected for a hash this proxy never stored — the model calls `headroom_retrieve`, gets a guaranteed miss, and re-does work it already had. Two wasted turns per adopted foreign hash.

Closes #2836

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/ccr/tool_injection.py`: added `CCRToolInjector.verify_ownership()` — filters `detected_hashes` down to hashes the compression store actually recognizes, via the same `store.exists()` check the retrieve endpoint itself performs. Added a small `_HashOwnershipStore` Protocol (structural typing, not a hard dependency on the concrete `CompressionStore` class) and a `compression_store` constructor field for dependency injection/testing. `scan_for_markers()` itself is untouched — kept store-independent (pure regex) rather than baking the check into the scan loop, since that approach broke 24 existing tests that correctly test "does this shape match" in isolation.
- `headroom/proxy/handlers/anthropic.py`, `headroom/proxy/handlers/openai.py`: call `injector.verify_ownership()` right after `scan_for_markers()` — the two real per-request call sites.
- `verify_ownership()` is also called inside `process_request()` (the convenience wrapper `batch.py`'s Google path uses), so that path is covered without a separate call site edit.
- `tests/test_ccr_tool_injection.py`: new `TestVerifyOwnership` class (6 tests) — the exact issue repro, a real-hash-survives case, mixed own/foreign hashes, explicit store override, store-exception safety (must not raise), and no-op-on-empty-hashes.
- `tests/test_proxy_anthropic_cache_stability.py`: 3 pre-existing tests needed updating for the new (correct) behavior — two `_FakeInjector` test doubles needed a `verify_ownership()` stub added, and one real end-to-end test needed a genuine store entry seeded (via `explicit_hash`) for the hash its hand-typed marker references, instead of asserting on an unverified shape-only match.

## Testing

- [ ] Type checking passes (`mypy headroom`) — not run locally, will confirm via CI

### Test Output

```text
$ .venv/Scripts/python -m pytest tests/test_ccr_marker_policy.py tests/test_ccr_tool_always_on.py tests/test_ccr_tool_injection.py tests/test_proxy/test_ccr_frozen_prefix_coupling.py tests/test_proxy_anthropic_cache_stability.py tests/test_proxy_handlers_batch.py tests/test_proxy_handler_helpers.py -q
151 passed in 15.87s

$ .venv/Scripts/python -m pytest tests/test_proxy_ccr.py tests/test_proxy_openai_responses_stream_ccr.py tests/test_anthropic_ccr_workspace_unbound.py tests/test_compression_store.py tests/test_no_ccr_lossy.py tests/test_proxy_handlers_batch.py -q
121 passed in 20.73s

$ .venv/Scripts/ruff check . && .venv/Scripts/ruff format --check .   # touched files only
All checks passed / already formatted
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.14.5, local venv
- Exact command / steps: ran the issue's exact 3-line repro (`CCRToolInjector.scan_for_markers()` on the foreign marker text, then `verify_ownership()`) before and after the fix; separately verified a genuinely-Headroom-stored hash (via `store.store(..., explicit_hash=...)`) still survives verification and still drives injection
- Observed result: before the fix (scan only, no verify step exists yet) `has_compressed_content` is `True` for the foreign marker — matches the bug report exactly. After adding `verify_ownership()`: foreign marker → `detected_hashes == []`, `has_compressed_content is False`; real stored hash → `detected_hashes == [real_hash]`, `has_compressed_content is True`.
- Not tested: have not driven this through a live two-context-tool proxy session (e.g. Headroom alongside another CCR-shaped tool in the same conversation) — verified at the unit/integration level (the exact repro plus the real proxy handler call sites via `test_proxy_anthropic_cache_stability.py`'s end-to-end `TestClient` tests), not via a live multi-tool session.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A — internal CCR safety behavior, no user-facing docs reference the marker-adoption mechanism)

## Additional Notes

Design note on why `verify_ownership()` is a separate step rather than baked into `scan_for_markers()`: my first attempt did exactly that and broke 24 tests across `test_ccr_tool_injection.py`, `test_ccr_marker_policy.py`, `test_proxy_anthropic_cache_stability.py`, and `test_proxy_handler_helpers.py` — all of them legitimately testing "does the regex detect this marker shape" independent of any store state. Keeping the scan pure and adding an explicit, separately-testable verification step kept that test surface intact while still closing the real gap at the three places that actually decide whether to advertise the retrieve tool.

### DIFF
--- a/headroom/ccr/tool_injection.py
+++ b/headroom/ccr/tool_injection.py
@@ -16,10 +16,22 @@ from __future__ import annotations
 import json
 import re
 from dataclasses import dataclass, field
-from typing import Any
+from typing import Any, Protocol, runtime_checkable
 
 # Tool name constant - used for matching tool calls
 CCR_TOOL_NAME = "headroom_retrieve"
+
+
+@runtime_checkable
+class _HashOwnershipStore(Protocol):
+    """Structural type for verify_ownership()'s store dependency.
+
+    Only needs the existence check — matches CompressionStore.exists()
+    without coupling this module to the concrete cache implementation
+    (or requiring test doubles to subclass it).
+    """
+
+    def exists(self, hash_key: str, clean_expired: bool = False) -> bool: ...
 
 
 def create_ccr_tool_definition(
@@ -170,6 +182,11 @@ class CCRToolInjector:
     inject_tool: bool = True
     inject_system_instructions: bool = True
     retrieval_endpoint: str = "/v1/retrieve"
+    # Store used to verify a scanned marker's hash is actually ours before
+    # advertising it (issue #2836). None resolves lazily to
+    # get_compression_store() — request-scoped store if one is set, else the
+    # global singleton — matching how every other CCR call site resolves it.
+    compression_store: _HashOwnershipStore | None = None
 
     # Detected compression markers
     _detected_hashes: list[str] = field(default_factory=list)
@@ -281,7 +298,16 @@ class CCRToolInjector:
         return self._detected_hashes
 
     def _scan_text(self, text: str) -> None:
-        """Scan text for compression markers from any compressor."""
+        """Scan text for compression markers from any compressor.
+
+        Shape-only: this matches the bracket format any compressor (or,
+        as it turns out, any *other* context tool) can produce. Callers
+        that need to know the hash is actually ours — i.e. before
+        advertising it to the model via the retrieve tool — must call
+        :meth:`verify_ownership` afterward. Kept separate so this method
+        stays a pure, store-independent text scan (that's what the
+        existing marker-format test suite exercises).
+        """
         for pattern in self._marker_patterns:
             matches = pattern.findall(text)
             for match in matches:
@@ -292,6 +318,59 @@ class CCRToolInjector:
                     hash_key = match  # Single capture group (generic pattern)
                 if hash_key and hash_key not in self._detected_hashes:
                     self._detected_hashes.append(hash_key)
+
+    def verify_ownership(self, store: _HashOwnershipStore | None = None) -> list[str]:
+        """Drop any detected hash the compression store doesn't recognize.
+
+        The bracket-marker shape (``[... hash=...]``) is not unique to
+        Headroom — other context tools emit visually identical markers.
+        Matching shape alone (what :meth:`scan_for_markers` does) adopts
+        their hashes too: ``has_compressed_content`` goes true and the
+        retrieve tool + "Available hashes" instruction get injected for a
+        hash this proxy never stored. The model then calls
+        ``headroom_retrieve``, gets a guaranteed miss, and re-does the work
+        it already had (issue #2836).
+
+        Call this after :meth:`scan_for_markers` and before checking
+        ``has_compressed_content`` / injecting the tool. Uses the same
+        ``store.exists()`` check the retrieve endpoint itself performs, so
+        a hash that survives this filter is provably redeemable right now
+        (or, if it expires between this check and the model's next call,
+        fails the same way a genuinely-ours stale hash already would —
+        this only removes hashes that were never ours to begin with).
+
+        Args:
+            store: Compression store to verify against. Defaults to
+                ``get_compression_store()`` (request-scoped if set, else
+                the global singleton) — the same resolution every other
+                CCR call site uses.
+
+        Returns:
+            The filtered ``detected_hashes`` list (also updates
+            ``self.detected_hashes`` in place).
+        """
+        if not self._detected_hashes:
+            return self._detected_hashes
+        if store is None:
+            store = self.compression_store
+        if store is None:
+            from headroom.cache.compression_store import get_compression_store
+
+            store = get_compression_store()
+
+        def _safe_exists(hash_key: str) -> bool:
+            try:
+                return store.exists(hash_key)
+            except Exception:
+                # A store lookup failure must not make CCR verification
+                # blow up the request; treat as "not ours" (drop the
+                # marker) — the safe direction, since a dropped real
+                # marker just means the model can't use the retrieve tool
+                # for it this turn, the same failure mode as CCR being off.
+                return False
+
+        self._detected_hashes = [h for h in self._detected_hashes if _safe_exists(h)]
+        return self._detected_hashes
 
     def inject_tool_definition(
         self,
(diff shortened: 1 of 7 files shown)


JerrettDavis (approved):
The ownership check closes the false-positive marker path at every injection site while retaining shape-only scanning as a pure parser. It resolves the same request-scoped/global store used by retrieval, fails safely on backend errors, and preserves verified hashes. The tests cover foreign, owned, mixed, injected-store, exception, and empty cases; all current checks are green.

JerrettDavis (approved):
Reviewed the updated head (`48d1add`) across all production scan/injection paths, request-scoped/global store resolution, mixed ownership, expiration/lookup failure behavior, session policy interaction, and the rebased batch path. Direct Anthropic/OpenAI handlers verify before advertising, while `process_request` covers the shared Google/batch path; failures safely suppress an unusable retrieve tool rather than failing the request. The focused regression suites are substantive, and the only red shard is an unrelated external binary-download 503. I found no blocking correctness issue. Small non-blocking cleanup: please update `CCRToolInjector`’s class usage example in a later edit to show `verify_ownership()` after `scan_for_markers()`, since the current example still demonstrates the unsafe shape-only sequence this PR fixes.

PR #2912, opened 2026-08-10
fix(proxy): emit request log timestamps in UTC

### LINKED ISSUE #2910: [BUG] RequestLog.timestamp is timezone-naive (outcome.py:529), producing negative "Xs ago" in dashboard Recent Requests
## Description

`RequestLog.timestamp` is written as a **timezone-naive** local-time string, so the dashboard's Recent Requests panel renders relative times that are wrong by the container's UTC offset. In a Docker deployment (container `TZ` = UTC) viewed from a browser in a negative-offset zone, the ages go negative:

```
TIME              MODEL             INPUT    OUTPUT   SAVED   LATENCY
-17996s ago       claude-sonnet-5   457.5k   892      2%      16951ms
-17978s ago       claude-sonnet-5   455.8k   294      2%      16075ms
-17961s ago       claude-sonnet-5   453.8k   403      2%      14491ms
```

`-17996s` ≈ 5 hours, exactly the UTC−5 offset of the viewing browser.

The offending write is `proxy/outcome.py:529`:

```python
timestamp=datetime.now().isoformat(),
```

`datetime.now()` with no `tz` yields naive local time, and `.isoformat()` emits no offset:

```
$ curl -s http://127.0.0.1:8787/stats | python3 -c "import json,sys; print(repr((json.load(sys.stdin)['recent_requests'] or [{}])[0].get('timestamp')))"
'2026-08-10T18:36:54.803515'
```

Per ECMAScript, a date-time string with no offset is parsed as **local** time, so the browser interprets that `18:36` UTC value as `18:36` in its own zone — five hours in the future — and the computed age is negative.

This appears to be an inconsistency rather than a deliberate choice: the durable ledger writes offset-aware timestamps. `savings_events.jsonl` entries carry `+00:00`, and `proxy/savings_tracker.py:79` does the right thing:

```python
return datetime.now(timezone.utc)
```

So `RequestLog.timestamp` looks like the one straggler.

## To Reproduce

1. Run the proxy in Docker with no `TZ` set (container defaults to UTC).
2. Browse the dashboard from a host in any non-UTC zone — a negative offset makes the ages negative; a positive offset inflates them.
3. Send any request through the proxy and watch the Recent Requests TIME column.

Demonstrating the skew directly, container vs. host:

```
$ docker exec <container> python3 -c "import datetime,time; print(time.tzname, datetime.datetime.now().isoformat())"
('UTC', 'UTC') 2026-08-10T18:37:30.085561

$ date +"%Y-%m-%dT%H:%M:%S%z"
2026-08-10T13:37:30-0500
```

## Expected Behavior

An offset-aware timestamp, so any client can convert unambiguously regardless of where the proxy runs:

```python
timestamp=datetime.now(timezone.utc).isoformat(),
```

That yields `2026-08-10T18:36:54.803515+00:00`, which browsers parse as an absolute instant. It also matches what `savings_events.jsonl` and `savings_tracker.py` already emit.

Worth checking whether any dashboard-side code parses this field assuming naive-local, since adding an offset would shift it; from the outside the panel appears to do plain relative-time math, which the fix corrects.

## Actual Behavior

Naive local time is emitted, and the panel shows negative ages (`-17996s ago`) whenever the container's zone is ahead of the browser's.

## Workaround

Aligning the container clock to the viewer's zone hides it:

```
--env TZ=America/Chicago
```

This only works because it makes container-local equal browser-local; the string is still naive, so the skew returns for any viewer in a different zone. Noting it as a workaround, not a fix.

## Environment

- Docker Desktop on Windows 11; container `TZ` unset (UTC), host UTC−5 (CDT)
- Preset `persistent-docker`, mode `token`, backend `anthropic`
- Browser and host in the same zone (UTC−5)

SulimanAbdulrazzaq:
## Description

`RequestLog.timestamp` was serialized with `datetime.now().isoformat()`, which omits timezone information. Browsers then interpret the value as local time, so requests from a UTC container can display negative ages in non-UTC dashboards.

Closes #2910

## Type of Change

## Changes Made

- Emit request-log timestamps from `datetime.now(timezone.utc)` so the ISO-8601 value includes `+00:00`.
- Add a regression test that parses the emitted timestamp and requires a UTC offset.

## Testing

- [x] New tests added for the regression
- [x] `python -m compileall -q headroom/proxy/outcome.py tests/test_request_outcome.py`
- [x] `git diff --check`
- [ ] Unit tests pass (`pytest`) — the repository's Rust extension cannot build in this Windows environment because `link.exe` (MSVC) is unavailable; the focused test is included for CI.

### Test Output

```text
python -m compileall -q headroom/proxy/outcome.py tests/test_request_outcome.py
(pass)

git diff --check
(pass)

uv run pytest tests/test_request_outcome.py -q
blocked while building headroom-py: linker `link.exe` not found
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.11; proxy timestamps are generated in `headroom/proxy/outcome.py`.
- Exact command / steps: traced the Recent Requests write path and added a timestamp assertion in `tests/test_request_outcome.py` (CI will run with the project's Rust toolchain).
- Observed result: the production call now emits an ISO-8601 timestamp with `+00:00`; the regression assertion requires an offset-aware UTC value, preventing browser timezone skew.
- Not tested: full pytest suite locally because the MSVC linker is unavailable.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/outcome.py
+++ b/headroom/proxy/outcome.py
@@ -27,7 +27,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any
 
 from headroom.proxy.tool_schema_savings_policy import (
@@ -526,7 +526,10 @@ async def emit_request_outcome(handler: Any, outcome: RequestOutcome) -> None:
         request_logger.log(
             RequestLog(
                 request_id=outcome.request_id,
-                timestamp=datetime.now().isoformat(),
+                # Request logs are consumed by browsers in arbitrary time zones.
+                # Include the UTC offset so relative-age calculations represent
+                # the same instant regardless of where the proxy runs.
+                timestamp=datetime.now(timezone.utc).isoformat(),
                 provider=outcome.provider,
                 model=outcome.model,
                 input_tokens_original=outcome.original_tokens,
--- a/tests/test_request_outcome.py
+++ b/tests/test_request_outcome.py
@@ -15,6 +15,7 @@ import asyncio
 import contextlib
 import logging
 from dataclasses import FrozenInstanceError
+from datetime import datetime, timezone
 from typing import Any
 from unittest.mock import AsyncMock, MagicMock
 
@@ -323,6 +324,22 @@ async def test_funnel_logs_request_with_derived_cache_hit() -> None:
     assert len(h.logger.logs) == 1
     log_entry = h.logger.logs[0]
     assert log_entry.cache_hit is True
+
+
+@pytest.mark.asyncio
+async def test_funnel_logs_request_timestamp_with_utc_offset() -> None:
+    """Recent-request timestamps must identify an absolute instant.
+
+    A naive ISO timestamp is interpreted in the browser's local timezone,
+    which makes the dashboard show negative ages when the proxy and browser
+    use different timezone settings.
+    """
+    h = _FunnelHarness()
+    await h._record_request_outcome(_outcome())
+
+    timestamp = datetime.fromisoformat(h.logger.logs[0].timestamp)
+    assert timestamp.tzinfo is not None
+    assert timestamp.utcoffset() == timezone.utc.utcoffset(timestamp)
 
 
 @pytest.mark.asyncio


JerrettDavis (approved):
This makes request-log timestamps unambiguous absolute instants and aligns the surface with the existing UTC-aware durable ledger. The regression test parses the emitted value and verifies both timezone awareness and a zero UTC offset. The branch is mergeable; broader CI has not run, which is not a contributor action item.

chopratejas (approved):
Maintainer review complete: UTC-aware ISO timestamps align with the durable ledger and remove browser-local ambiguity; the regression parses the emitted value and verifies a zero UTC offset. Merge after full CI.

PR #2913, opened 2026-08-10
fix(opencode): keep Claude models off OpenAI provider

### LINKED ISSUE #2911: [BUG] Injected OpenCode provider advertises Claude models on the OpenAI-compatible path; requests are forwarded to platform.openai.com
**Bug:** the `headroom` provider that Headroom injects into OpenCode advertises Claude model ids on an OpenAI-compatible endpoint, so selecting one forwards the request to `platform.openai.com` and fails with `invalid_api_key`. `/stats` then reports it under `by_provider: openai` while `cost.per_model` records it as `claude-sonnet-4-6`, which reads like a mis-attribution but is actually the routing being wrong one layer earlier.

**Repo:** https://github.com/headroomlabs-ai/headroom
**Component:** `headroom/providers/opencode/config.py`, proxy upstream selection
**Version:** `headroom-ai==0.34.0` (uv tool), OpenCode `1.18.16` (from its own `User-Agent` on the proxied request), Linux
**Severity:** Medium — every Claude model Headroom itself advertises to OpenCode is unusable, and the dashboard label makes it look like an accounting problem

## Summary

`headroom_provider_entry()` declares the provider as `@ai-sdk/openai-compatible` pointed at `http://127.0.0.1:<port>/v1`, and `HEADROOM_OPENCODE_MODELS` lists `claude-sonnet-4-6`, `claude-opus-4-6`, `claude-haiku-4-5-20251001` alongside `gpt-4o` and `gpt-4.1`.

The proxy selects its upstream from the request path, not from the model. A request on `/v1/chat/completions` therefore goes to the configured OpenAI upstream regardless of the model id, so any of the three Claude models Headroom advertises is sent to OpenAI's API.

## Reproduction (no client needed)

```
$ curl -s http://127.0.0.1:8787/v1/chat/completions \
    -H 'Content-Type: application/json' -H 'Authorization: Bearer dummy' \
    -d '{"model":"claude-sonnet-4-6","messages":[{"role":"user","content":"ping"}],"max_tokens":8}'
{
    "error": {
        "message": "Incorrect API key provided: dummy. You can find your API key at https://platform.openai.com/account/api-keys.",
        "type": "invalid_request_error",
        "param": null,
        "code": "invalid_api_key"
    }
}
```

The upstream is named by OpenAI's own error page: an Anthropic model id was forwarded to `platform.openai.com`.

Via OpenCode the same thing happens through Headroom's own injected config:

```
$ opencode run --model headroom/claude-sonnet-4-6 "reply with exactly: ok"
Error: You didn't provide an API key. [...] https://platform.openai.com/account/api-keys
```

## What `/stats` shows afterwards

```
before: by_provider={'zen': 24, 'openai': 2}
after:  by_provider={'zen': 24, 'openai': 3}
        cost.per_model['claude-sonnet-4-6'] = {'requests': 3, 'tokens_sent': 27059, ...}
```

The `openai` label is correct for what `by_provider` measures — `resolve_display_provider()` only relabels counts whose internal provider is already `openai`, and it documents itself as a display label over the upstream, not the model vendor. So this is **not** an attribution bug, and I want to be explicit about that: the counter is honest, it is describing a request that really did go to OpenAI. The defect is that the request went there at all.

The consequence is still a reporting one, though. A dashboard reader sees Claude traffic filed under OpenAI and reasonably concludes the accounting is broken, which sends the investigation to the wrong layer. I spent an evening there.

## Expected

One of these, in order of preference:

1. Refuse at the edge: a model id that cannot be served by the resolved upstream returns a Headroom error naming both, rather than being forwarded to a vendor that will reject it.
2. Route by model when the id is unambiguous, so Anthropic ids on the OpenAI-compatible path reach the Anthropic upstream.
3. If neither is intended, stop advertising Claude ids in `HEADROOM_OPENCODE_MODELS`, since every one of them is a guaranteed failure for anyone who follows the generated config.

## Note

Reported from local reproduction on a machine where the proxy is otherwise healthy and serving `zen` traffic normally (24 requests routed and compressed in the same session as the failures above), so this is not a broken install.

SulimanAbdulrazzaq:
## Description

The injected `headroom` OpenCode provider uses `@ai-sdk/openai-compatible` and the proxy's `/v1/chat/completions` route. It currently advertises Claude model IDs in that provider, so OpenCode sends Claude requests to the OpenAI upstream and receives `invalid_api_key` errors. Keep Claude on OpenCode's native `anthropic` provider, which Headroom already redirects to the proxy.

Closes #2911

## Type of Change

- [ ] Breaking change (bug fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Remove Claude IDs from the injected OpenAI-compatible provider model map.
- Keep GPT models available through the `headroom/<id>` namespace.
- Add regression assertions that generated config never advertises Claude models on this endpoint.

## Testing

### Test Output

```text
python -m pytest tests/test_providers_opencode_config.py -q -k "not build_launch_env_with_project"
40 passed, 1 deselected

python -m ruff check headroom/providers/opencode/config.py tests/test_providers_opencode_config.py
All checks passed!

python -m compileall -q headroom/providers/opencode/config.py tests/test_providers_opencode_config.py
(pass)
```

The full config test module also exposes an unrelated pre-existing Windows path assertion failure in `test_build_launch_env_with_project`; the failure is caused by comparing a native `Path` string with JSON-escaped backslashes and is outside this change.

## Real Behavior Proof

- Environment: Windows 11, Python 3.11; no external API credentials used.
- Exact command / steps: `python -c "from headroom.providers.opencode.config import headroom_provider_entry; print(sorted(headroom_provider_entry(8787)['models']))"`
- Observed result: `['gpt-4.1', 'gpt-4o']`; the generated OpenAI-compatible provider no longer advertises any `claude-*` IDs.
- Not tested: live OpenCode request routing or a vendor API call, because they require external credentials. The regression suite verifies the generated configuration consumed by OpenCode.

## Review Readiness

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas (not needed; the provider routing rationale is documented inline)
- [ ] I have made corresponding changes to the documentation (the generated provider behavior is documented in code; existing docs describe the separate npm provider)
- [x] New and existing relevant unit tests pass locally with my changes

## Additional Notes

The native `anthropic` and `openai` provider entries both continue to point at the Headroom proxy, so this change only removes an invalid duplicate Claude route and does not affect native Claude traffic.

### DIFF
--- a/headroom/providers/opencode/config.py
+++ b/headroom/providers/opencode/config.py
@@ -31,24 +31,18 @@ _MCP_BLOCK_RE = re.compile(
 )
 HEADROOM_OPENCODE_PLUGIN = "headroom-opencode"
 
-# Models exposed by the injected `headroom` provider. OpenCode only resolves
-# `headroom/<id>` for ids listed in the provider's `models` map, so an empty
-# map means every documented `headroom/*` model fails with "Model not found".
-# Keep in sync with DEFAULT_MODELS in plugins/opencode/src/provider.ts and the
-# table in plugins/opencode/README.md.
+# Models exposed by the injected `headroom` provider. This provider uses
+# ``@ai-sdk/openai-compatible`` and the proxy's ``/v1/chat/completions`` path,
+# which is routed to the configured OpenAI upstream. Do not advertise Claude
+# models here: OpenCode would send them through the OpenAI endpoint and report
+# an ``invalid_api_key`` error instead of reaching Anthropic. Claude models are
+# available through OpenCode's native ``anthropic`` provider, whose base URL is
+# also redirected to Headroom by ``build_opencode_config_content``.
+#
+# OpenCode only resolves ``headroom/<id>`` for ids listed in this map, so an
+# empty map means every documented ``headroom/*`` model fails with "Model not
+# found".
 HEADROOM_OPENCODE_MODELS: dict[str, Any] = {
-    "claude-sonnet-4-6": {
-        "name": "Claude Sonnet 4.6",
-        "limit": {"context": 200000, "output": 16384},
-    },
-    "claude-opus-4-6": {
-        "name": "Claude Opus 4.6",
-        "limit": {"context": 200000, "output": 16384},
-    },
-    "claude-haiku-4-5-20251001": {
-        "name": "Claude Haiku 4.5",
-        "limit": {"context": 200000, "output": 8192},
-    },
     "gpt-4o": {
         "name": "GPT-4o",
         "limit": {"context": 128000, "output": 16384},
--- a/tests/test_providers_opencode_config.py
+++ b/tests/test_providers_opencode_config.py
@@ -199,7 +199,10 @@ def test_inject_provider_config_creates_file(
     assert config["provider"]["headroom"]["npm"] == "@ai-sdk/openai-compatible"
     # Bare model ids: OpenCode resolves them as "headroom/<id>" (#1657).
     models = config["provider"]["headroom"]["models"]
-    assert "claude-sonnet-4-6" in models
+    assert set(models) == {"gpt-4o", "gpt-4.1"}
+    # The injected provider is OpenAI-compatible. Claude models must remain on
+    # OpenCode's native Anthropic provider so they are not sent to OpenAI.
+    assert not any(model_id.startswith("claude-") for model_id in models)
     assert all(not model_id.startswith("headroom/") for model_id in models)
     assert "mcp" not in config
     assert "model" not in config  # headroom provider is a transparent pass-through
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
For the Python-injected OpenCode config, the `headroom` entry is explicitly OpenAI-compatible, so removing Claude IDs prevents a selectable route that cannot reach the native Anthropic provider. The native `anthropic` entry remains redirected through Headroom, and the tests pin both the positive GPT set and the absence of `claude-*`. The TypeScript helper is a separately documented npm surface and is not silently changed here. The branch is mergeable; broader CI has not run.

chopratejas (approved):
Maintainer review complete for the reported Python-injected OpenCode configuration: Claude IDs are removed from the OpenAI-compatible provider while the native Anthropic provider remains routed through Headroom, and the regression pins both model sets. Merge after full CI. The separate npm helper surface should be tracked independently.

PR #2914, opened 2026-08-10
fix(install): trust Docker bridge for dashboard metadata

### LINKED ISSUE #2909: [BUG] persistent-docker preset can't display its own dashboard: Recent Requests and Per-Project Savings silently gated by non-loopback bridge peer
## Description

A stock `headroom install apply --preset persistent-docker` produces a dashboard where **Recent Requests** and **Per-Project Savings** are permanently empty, even while traffic is flowing and being compressed correctly. No reverse proxy, no custom network, no hostname access — just the default preset on Docker's default bridge, browsed at `http://127.0.0.1:8787/dashboard`.

The cause is that Docker NATs host-originated traffic to the bridge gateway, so the proxy never sees a loopback peer:

```
$ grep -o 'path=/stats.*client=[0-9.]*' logs/proxy.log | grep -o 'client=[0-9.]*' | sort | uniq -c
   1293 client=172.17.0.1
```

1293 of 1293 `/stats` requests arrive from `172.17.0.1`, never `127.0.0.1`. `_request_can_view_dashboard_metadata()` (`proxy/server.py:2292`) therefore fails its loopback check, and with `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS` unset `peer_is_trusted_gateway()` returns `False`. The sensitive sub-payloads are then stripped:

- `/stats` pops `recent_requests` and `request_logs` (`server.py:~4206`)
- `/stats-lifetime` pops `projects` (`server.py:~4216`)

So the API withholds data that exists and is correct on disk.

The mechanism is understood and intentional — this is not a request to weaken the gate. The bug is that the **first-party Docker preset does not configure itself to pass its own gate**, so the out-of-box experience is a broken dashboard.

## To Reproduce

1. `headroom install apply --preset persistent-docker --runtime docker --port 8787 --backend anthropic --mode token`
2. Route an agent through the proxy and let it make several requests.
3. Open `http://127.0.0.1:8787/dashboard`.
4. Aggregate panels (Per-Model Token Savings, Subscription Window) populate correctly. **Recent Requests** shows *"No requests yet. Start using the proxy to see activity here."* and **Per-Project Savings** shows *"No per-project data yet — 0 project(s)"*.

Confirming the data exists but is gated — from the host:

```
# recent_requests / request_logs keys absent entirely
$ curl -s http://127.0.0.1:8787/stats | python3 -c "import json,sys; d=json.load(sys.stdin); print('recent_requests' in d, 'request_logs' in d)"
False False

# projects popped
$ curl -s http://127.0.0.1:8787/stats-lifetime | python3 -c "import json,sys; print(json.load(sys.stdin).get('projects'))"
None
```

After adding `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS=172.17.0.1/32`, the same two calls return the keys, `projects` returns 8 populated entries, and both panels render. Nothing else changed.

## Expected Behavior

`--preset persistent-docker` should render its own dashboard without manual network configuration. Options, roughly in order of preference:

1. Have `install apply --preset persistent-docker` default `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS` to the bridge gateway it is about to publish through (discoverable via `docker network inspect bridge`). Since the container publishes to `127.0.0.1:<port>` on the host, the gateway peer *is* the loopback boundary in that topology.
2. Treat a peer that matches the container's default-gateway address as loopback-equivalent when the published binding is host-loopback.
3. At minimum, make the failure legible — see below.

## Actual Behavior

Both panels are silently empty, and the empty-state copy actively misleads: *"No requests yet. Start using the proxy to see activity here."* asserts there was no traffic, when the real cause is authorization. This sent me down a long false trail (I suspected volume persistence and Windows bind-mount write semantics) before finding the gate. A distinguishable empty state — e.g. *"Per-request metadata withheld: client 172.17.0.1 is not loopback and not in HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS"* — would have made this a two-minute diagnosis. A one-line startup `WARN` when the listener is non-loopback and the CIDR allowlist is empty would do the same.

## Environment

- Docker Desktop on Windows 11, default bridge network (`172.17.0.0/16`, gateway `172.17.0.1`)
- Preset `persistent-docker`, scope `user`, mode `token`, backend `anthropic`
- Published as `127.0.0.1:8787->8787/tcp`; container binds `0.0.0.0:8787`

## Related

- #1959 — the reverse-proxy feature request that (as I read it) motivated `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS`. That knob works; this issue is that the first-party Docker preset doesn't set it.
- #2466 — same gate, custom network + reverse proxy, 404 on the settings route. This report is the narrower case: default preset, default bridge, no reverse proxy.

SulimanAbdulrazzaq:
## Summary

Closes #2909.

The `persistent-docker` installer now discovers Docker's default bridge gateway and passes the exact `/32` gateway CIDR to the proxy's dashboard metadata allowlist when no explicit `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS` value is configured. This keeps the existing metadata gate intact while allowing the first-party loopback-published container to see its own Recent Requests and Per-Project Savings data. Explicit user configuration continues to take precedence.

Both native wrappers (POSIX and PowerShell) use the same behavior, and installer integration coverage verifies the generated Docker command.

## Validation

- `python -m pytest tests/test_install/test_native_installers.py -q -k bash` (1 skipped on Windows because Bash is unavailable)
- PowerShell wrapper smoke test with the repository fake Docker shim: verified `docker network inspect bridge` is called and `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS=172.17.0.1/32` is passed to `docker run`
- Explicit allowlist smoke test: verified an existing `HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS` is preserved without adding a discovered default
- `git diff --check`

## Real behavior proof

Setup tested: Windows 11 host, PowerShell wrapper, repository fake Docker shim (Docker CLI is not installed in this environment).

Exact command: `headroom.ps1 install apply --profile smoke --port 18999 --image fake/headroom:test`.

Observed result: the generated Docker invocation included `docker network inspect bridge --format ...` and `--env HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS=172.17.0.1/32`, and the installer completed successfully.

Not tested: a live Docker daemon/dashboard request on this host.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -349,7 +349,9 @@ function Get-PersistentDockerArgs {
 function Add-DashboardGatewayEnv {
     param([System.Collections.Generic.List[string]]$ArgsList)
 
-    # A host request published through Docker's default bridge reaches the
+    # This default is safe only because the published dashboard port is bound
+    # to the host loopback interface below. A host request published through
+    # Docker's default bridge reaches the
     # container from the bridge gateway (for example, 172.17.0.1), not from
     # 127.0.0.1. Trust only that exact gateway by default so the dashboard's
     # metadata gate works for the first-party persistent Docker preset while
@@ -513,7 +515,7 @@ function Start-PersistentDockerInstall {
     docker rm -f $containerName | Out-Null 2>$null
 
     $dockerArgs = New-Object System.Collections.Generic.List[string]
-    $dockerArgs.AddRange([string[]]@('run','-d','--restart','unless-stopped','--name',$containerName,'-p',"$Port`:$Port"))
+    $dockerArgs.AddRange([string[]]@('run','-d','--restart','unless-stopped','--name',$containerName,'-p',"127.0.0.1`:$Port`:$Port"))
     $dockerArgs.AddRange((Get-PersistentDockerArgs))
     Add-DashboardGatewayEnv -ArgsList $dockerArgs
     $dockerArgs.AddRange([string[]]@(
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -346,6 +346,27 @@ function Get-PersistentDockerArgs {
     return ,$args.ToArray()
 }
 
+function Add-DashboardGatewayEnv {
+    param([System.Collections.Generic.List[string]]$ArgsList)
+
+    # A host request published through Docker's default bridge reaches the
+    # container from the bridge gateway (for example, 172.17.0.1), not from
+    # 127.0.0.1. Trust only that exact gateway by default so the dashboard's
+    # metadata gate works for the first-party persistent Docker preset while
+    # preserving an explicitly configured allowlist.
+    if (Test-Path Env:HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS) {
+        return
+    }
+
+    $gateway = (& docker network inspect bridge --format '{{(index .IPAM.Config 0).Gateway}}' 2>$null | Out-String).Trim()
+    if ($LASTEXITCODE -eq 0 -and $gateway) {
+        $ArgsList.Add('--env')
+        $ArgsList.Add("HEADROOM_PROXY_TRUSTED_DASHBOARD_CLIENT_CIDRS=$gateway/32")
+    } else {
+        Write-Warning 'Could not determine Docker bridge gateway; dashboard metadata remains restricted'
+    }
+}
+
 function Get-ManifestProxyArgs {
     param(
         [int]$Port,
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
The installer change scopes the default trust to the single bridge-gateway /32, preserves an explicitly configured allowlist (including an explicitly empty value), and applies the same behavior to both native wrappers. The integration shim verifies the generated Docker arguments on both paths. The branch is mergeable; broader CI has not run, which is not a contributor action item for this fork.

chopratejas (dismissed):
Security blocker: both native wrappers currently publish the container port on all host interfaces, then this PR trusts the Docker bridge gateway for dashboard metadata. On Docker Desktop and other proxying/NAT topologies, LAN traffic can arrive represented by that trusted gateway and gain access to recent-request and project metadata. Bind the published port to loopback in both wrappers (127.0.0.1:PORT:PORT) before adding gateway trust. Add Bash and PowerShell assertions for the exact mapping, plus coverage that explicit set and explicit-empty CIDR values are preserved.

chopratejas (approved):
Latest-head maintainer review complete. Both native installers now bind the published port to host loopback before trusting the exact bridge-gateway address. The harness verifies discovered, explicitly configured, explicitly empty, and failed-discovery behavior for both wrapper paths. Merge after Linux and Windows CI are green.

Issue #2375, opened 2026-07-18
[BUG] cache_control breakpoint consolidation silently drops ttl (1h caching downgrades to 5m after the first turn)


brentdunnaway:
## Summary

`normalize_message_cache_control()` in `headroom/cache/prefix_tracker.py` silently drops the `ttl` field (e.g. `"1h"`) when consolidating cache_control breakpoints. Any client using the 1-hour prompt-cache TTL (`cache_control: {"type": "ephemeral", "ttl": "1h"}`) silently gets downgraded to the 5-minute default on every turn after the first, with no error and no indication anything changed.

## Root cause

`normalize_message_cache_control()` (`headroom/cache/prefix_tracker.py:362-409`) exists to solve a real problem: Anthropic hard-errors past 4 `cache_control` blocks total, and this proxy's own `overlay_cached_prefix` replay mechanism (used for CCR) piles up markers turn over turn. The fix strips every message-level `cache_control` and re-places exactly one, on the last block of the last block-style message:

```python
stripped = [
    {k: v for k, v in b.items() if k != "cache_control"} if isinstance(b, dict) else b
    for b in content
]
...
content[-1] = {**content[-1], "cache_control": {"type": "ephemeral"}}
```

That second line is the bug — it's a brand-new literal dict, not a copy of whatever `cache_control` was actually on the stripped markers. Any `ttl` key is discarded; the re-placed breakpoint is unconditionally the bare 5-minute default.

## Reproduction

1. Run the proxy against `litellm-vertex` (or any backend) with a client that sets `ENABLE_PROMPT_CACHING_1H=1` (Claude Code) or otherwise sends `cache_control: {"type": "ephemeral", "ttl": "1h"}`.
2. Have a multi-turn conversation (2+ turns).
3. Inspect the request actually forwarded upstream (e.g. via `--log-messages`) — `request_messages` (what the client sent) shows `ttl: "1h"`; `compressed_messages` (what gets forwarded) shows only `{"type": "ephemeral"}`, no `ttl`, on every turn after the first.

Confirmed this is not backend-specific — reproduced identically in this codebase's own request logging regardless of which provider is configured, since `normalize_message_cache_control` runs unconditionally in `proxy/handlers/anthropic.py`'s pre-upstream pipeline.

## Suggested fix

Capture whichever `ttl` (if any) exists on the markers being stripped, and carry it onto the re-placed breakpoint:

```python
found_ttl: str | None = None
for i, msg in enumerate(messages):
[...]

PR #2919, opened 2026-08-11
fix(cache): mirror client cache_control positions instead of single-marker consolidation

gglucass:
## Description

`normalize_message_cache_control` strips every message-level `cache_control` marker and re-places a single one on the newest block. The premise (the provider's cache key is content, not markers) is correct, but Anthropic resolves each breakpoint with a **~20-content-block lookback** ([prompt caching docs](https://platform.claude.com/docs/en/build-with-claude/prompt-caching), "20-block lookback window"). Agentic clients such as Claude Code keep a marker on the previous turn's newest message (and on multiple blocks within one long message) precisely so a new turn's cache write can chain to the prior entry. Consolidating to one marker breaks that chain whenever a single turn adds more than 20 blocks - typical for MCP/tool-heavy turns - so the whole message history silently re-bills as cache creation, and a marker anchored short of the final block leaves the tail billing as fully uncached input.

This came out of a user field report (2026-08-11): cache miss rate 2-3x higher on 11 of 11 projects after enabling the proxy, plus 0.6-1.4M fully uncached input tokens/day, concentrated in MCP-heavy sessions. The consolidation never registers as a body mutation, so existing `mutation_reasons` logging could not implicate it, and `classify_cache_miss` counts partial reads as hits, so TTL telemetry showed a clean sheet throughout.

This PR fixes the consolidation and adds the diagnostics that found it:

1. **Mirror the client's marker positions** (block-level) instead of consolidating. Replay-leftover stripping is unchanged; accumulation stays bounded by the client's own 4-marker budget. Per-position marker values are preserved, so an explicit `ttl: "1h"` survives at every position (extends #2375). Falls back to the legacy single-marker behavior when client positions are unavailable, message counts mismatch, or the client marked nothing.
2. **`event=cache_breakpoints` log line per forwarded request**: client markers in vs forwarded out, per section, plus the last marker's distance from the tail. Escalates to WARNING when a marker is dropped or the last marker moves away from the tail - this class of bug becomes self-diagnosing from proxy.log alone.
3. **`HEADROOM_LOG_PAYLOAD_PREVIEW=0`** strips verbatim `payload_preview` content from `headroom_retrieve` log events (byte counts retained). The reporting user could not share logs because previews carry raw tool-result content; this makes proxy.log shareable in bug reports.
4. **Proactive-expansion guard**: never inject expansion text into a text block carrying the client's `cache_control` - the client re-sends the original bytes next turn, so mutating the breakpointed block busts the prefix at that message on every later turn.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cache/prefix_tracker.py`: `normalize_message_cache_control` gains a `client_messages` parameter and mirrors the client's (message, block) marker positions; new `_client_marker_positions` helper; legacy consolidation kept as fallback
- `headroom/proxy/handlers/anthropic.py`: pass `original_client_messages` into normalize; snapshot inbound breakpoints after request parse and log in/out comparison before forwarding; skip proactive-expansion injection into breakpointed text blocks
- `headroom/proxy/helpers.py`: `count_cache_breakpoints` (counts system/tools/messages markers incl. nested tool_result sub-blocks, tracks last-marker tail distance) and `log_cache_breakpoints` (INFO normally, WARNING on drop/tail-growth)
- `headroom/cache/compression_store.py`: `HEADROOM_LOG_PAYLOAD_PREVIEW` env switch for `_payload_for_retrieval_log`
- `tests/test_cache_control_move_bust.py`: 8 new tests for position mirroring (multi-marker messages, per-position ttl, boundedness across turns, count-mismatch/no-marker fallbacks, out-of-range block clamp, string-content skip)
- `tests/test_cache_breakpoint_diagnostics.py`: new file, 9 tests covering the counter, the WARNING escalation, the payload-preview switch, and the expansion guard

## Testing

### Test Output

```text
$ pytest tests/test_cache_control_move_bust.py tests/test_cache_ttl_preserved.py \
    tests/test_cache_breakpoint_diagnostics.py tests/test_cache/ \
================== 314 passed, 2 skipped, 1 warning in 19.91s ==================

$ ruff check .
Found 4 errors.   # all pre-existing in plugins/headroom-oauth2/, untouched by this PR
$ ruff check headroom/ tests/test_cache_control_move_bust.py tests/test_cache_breakpoint_diagnostics.py
All checks passed!

$ mypy headroom/cache/prefix_tracker.py headroom/proxy/helpers.py headroom/cache/compression_store.py
```

## Real Behavior Proof

- Environment: macOS (Darwin 24.6.0), Python 3.12, patch applied to a live proxy deployment serving real Claude Code traffic (model claude-fable-5, Anthropic Messages API), one full working day
- Exact command / steps: enabled the `event=cache_breakpoints` logging first, then deployed the fix in two stages; compared `grep "event=cache_breakpoints" proxy.log` counts per stage and per-turn `usage` from the client's session transcripts
- Observed result: with consolidation (pre-fix): 62 of 218 requests logged `dropped=true` (typically 3 client markers in, 1 out). Message-level mirroring: 17 of 132. Block-level mirroring (this PR): 0 of 38, with `in_messages == out_messages` on every request including the 3-marker intra-message shape; compression savings events continued unchanged throughout, and client-side cache miss rate held at ~3.3% with near-zero uncached input tokens
- Not tested: OpenAI-format clients (message-level `cache_control` is Anthropic-format only; those paths fall back to legacy behavior), Bedrock handler beyond the existing unit suite, and the reporting user's exact MCP-heavy workload (their confirmation pending upgrade)

## Review Readiness

## Checklist

N/A - log-line and request-shaping change; representative log output is in Real Behavior Proof.

## Additional Notes

- Documentation checklist item is unchecked because no user-facing docs cover message-level marker handling today; happy to add a section to the caching docs if maintainers want one (the docstring on `normalize_message_cache_control` carries the full rationale).
- The `ruff check .` findings are 4 pre-existing issues in `plugins/headroom-oauth2/` that this PR does not touch.
- Design tradeoff: the newest client marker keeps the stable-run anchoring from #2917 only when the client placed it on its message's final block; an explicit mid-message marker is honored verbatim. Out-of-range block indices (a transform merged/removed blocks) clamp to the message's last block - markers are not part of the provider's cache key, so a near-miss placement is harmless while a dropped marker is not.

### DIFF
--- a/headroom/cache/compression_store.py
+++ b/headroom/cache/compression_store.py
@@ -52,6 +52,10 @@ DEFAULT_CCR_TTL_SECONDS = 1800  # session-scale; override via HEADROOM_CCR_TTL_S
 CCR_TTL_SECONDS_ENV = "HEADROOM_CCR_TTL_SECONDS"
 
 _RETRIEVAL_LOG_PREVIEW_CHARS = 4096
+# Previews carry verbatim tool-result content (post-redaction), which makes
+# proxy.log too sensitive for users to share in bug reports. Set to
+# 0/false/no/off to log byte counts only.
+PAYLOAD_PREVIEW_ENV = "HEADROOM_LOG_PAYLOAD_PREVIEW"
 _SECRET_KEY_VALUE_RE = re.compile(
     r"(?i)\b([A-Z0-9_-]*(?:API[_-]?KEY|TOKEN|SECRET|PASSWORD|CREDENTIAL|AUTH)[A-Z0-9_-]*)"
     r"(\s*[:=]\s*)([\"']?)([^\"'\s,}]+)"
@@ -108,7 +112,21 @@ def _redact_retrieval_log_payload(payload: str) -> str:
     return _API_KEY_VALUE_RE.sub("sk-[REDACTED]", redacted)
 
 
+def _payload_preview_enabled() -> bool:
+    raw = os.environ.get(PAYLOAD_PREVIEW_ENV)
+    if raw is None:
+        return True
+    return raw.strip().lower() not in ("0", "false", "no", "off")
+
+
 def _payload_for_retrieval_log(payload: str) -> dict[str, Any]:
+    if not _payload_preview_enabled():
+        return {
+            "payload_chars": len(payload),
+            "payload_preview_chars": 0,
+            "payload_truncated": len(payload) > 0,
+            "payload_preview": "",
+        }
     redacted = _redact_retrieval_log_payload(payload)
     preview = redacted[:_RETRIEVAL_LOG_PREVIEW_CHARS]
     truncated = len(redacted) > len(preview)
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -635,23 +635,22 @@ def _breakpoint_index(
 
 def _client_marker_positions(
     client_messages: list[dict[str, Any]],
-) -> list[tuple[int, dict[str, Any]]]:
-    """Message indices where the CLIENT placed cache_control, with each marker.
+) -> list[tuple[int, int, dict[str, Any]]]:
+    """(message index, block index, marker) for every CLIENT cache_control.
 
-    One entry per marked message (the last marker in the message wins), in
-    message order. Only block-style content can carry markers.
+    Block-level, not one-per-message: clients mark multiple blocks within a
+    single long message (Claude Code does this on 1-2-message requests with a
+    large first message), and the ~20-block lookback applies within a message
+    just as it does across messages. Only block-style content carries markers.
     """
-    positions: list[tuple[int, dict[str, Any]]] = []
+    positions: list[tuple[int, int, dict[str, Any]]] = []
     for i, msg in enumerate(client_messages):
         content = msg.get("content") if isinstance(msg, dict) else None
         if not isinstance(content, list):
             continue
-        marker: dict[str, Any] | None = None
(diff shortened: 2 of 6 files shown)


gglucass:
Flagging this one for priority review @JerrettDavis @chopratejas 

It fixes a silent cost regression in normalize_message_cache_control that affects every Claude Code user on tool-heavy workloads. The single-marker consolidation breaks Anthropic's documented ~20-block breakpoint lookback, so large turns re-bill the whole message history as cache creation — we measured markers being dropped on 28% of requests on a live deployment, and a user field report independently measured 2–3× cache miss rates and ~1M uncached input tokens/day. The failure is invisible to existing telemetry (mutation_reasons never lists it; classify_cache_miss counts partial reads as hits), which is why the PR also adds the event=cache_breakpoints in/out logging that found it. Fix is fallback-safe, field-validated (drop rate 28% → 0%), and builds on #2375/#2917. Happy to adjust anything.

JerrettDavis (approved):
Reviewed the latest 144b7d1f head adversarially. The normalization now strips replay leftovers while restoring every current client block-level breakpoint (including per-position TTL), retains the stable-run anchor only for the newest final-block intent, and falls back safely when message alignment is unavailable. The diagnostics cover all request sections and warn on count/tail regressions; the payload-preview switch and breakpoint-safe proactive injection are independently tested. Full CI and patch coverage are green.

PR #2412, opened 2026-07-19
fix(proxy): time-cap the compression timeout-debt quarantine (#2360)

abhay-codes07:
## Description

Fixes #2360.

The proxy runs compression on a bounded thread-pool executor with a per-request deadline. Because Python cannot preempt a worker after its `asyncio.wait_for` times out, the code quarantines new compression while a timed-out worker is still running (`_compression_timed_out_in_flight > 0`), to avoid piling more work onto a saturating executor.

The gap: that counter only decrements when the worker finally exits. A worker that **never returns** — a hung or pathological compression of a large frame — keeps the counter above zero forever, so the quarantine stays open permanently and every subsequent compression raises `CompressionQuarantinedError`. On Codex WS this is exactly what #2360 reports: one 5s timeout, then Token Savings pinned at ~0% with no recovery, even though the machine is fine.

The "parity with direct upstream" nature of the accounting was correct; the only missing piece is an upper bound on how long a single stuck worker may hold the quarantine.

## Fix

Add a time cap on the quarantine:

- A deadline (`_compression_quarantine_deadline`) is (re)armed on every fresh timeout, to `now + HEADROOM_COMPRESSION_QUARANTINE_MAX_SECONDS` (default **60s**).
- The gate quarantines only while `timed_out_in_flight > 0` **and** `now < deadline`. Once the deadline lapses with no new timeouts, the worker is presumed leaked/abandoned and compression resumes. The release is counted once (a `"released"` quarantine metric + a warning), and the deadline is cleared so it is not re-counted on every later request.
- The bounded executor still caps thread growth, and any new timeout re-arms the quarantine, so ongoing genuine slowness keeps quarantining while a single hung worker cannot pin it forever.

This preserves the original protection (a burst of slow compressions still quarantines) while guaranteeing recovery.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `headroom/proxy/server.py`: add `_compression_quarantine_deadline` / `_compression_quarantine_max_seconds` (from `HEADROOM_COMPRESSION_QUARANTINE_MAX_SECONDS`, default 60s) and `_compression_quarantine_releases`; arm the deadline when timeout debt is recorded; release the quarantine (once) in the gate when the deadline lapses.
[...]

PR #2164, opened 2026-07-14
fix(proxy): give each Codex /v1/responses WS turn a unique request_id

rodboev:
## Description

After any Codex traffic, the dashboard "Recent Requests" table goes blank — including the unrelated Anthropic/Claude rows — even though the proxy is actively handling and compressing Codex `/v1/responses` WebSocket turns and aggregate counters keep moving. The feed isn't stale; it is being wiped client-side.

Root cause: the Codex WebSocket handler `OpenAIHandlerMixin.handle_openai_responses_ws` (`headroom/proxy/handlers/openai.py`) mints a single `request_id` per WebSocket **session** (`_next_request_id()` near the top of the handler) and reuses it for every per-turn `RequestOutcome` in `_record_ws_response_metrics`, the session-residual outcome, and the session-summary `RequestLog`. Those all flow through `emit_request_outcome` (`headroom/proxy/outcome.py`), which writes a `RequestLog` per outcome into the request logger that backs `/stats.recent_requests` and `/transformations/feed` — so one session with N turns produces N+ feed rows sharing one `request_id`. The dashboard renders that feed with `<template x-for="req in (stats.recent_requests || [])" :key="req.request_id">` (`headroom/dashboard/templates/dashboard.html:1298`); Alpine requires unique `:key`s, so duplicate ids abort the entire `x-for` render and blank the whole table. Anthropic/HTTP requests each get a unique incrementing id from `_next_request_id()` and are unaffected — which is why only Codex traffic triggers the blanking.

This PR gives each Codex WS feed emission a fresh unique id from the same authoritative `_next_request_id()` counter (per-turn, residual, and summary sites), restoring the "one unique id per feed row" invariant that Anthropic already satisfies. With unique ids the Alpine `:key`s no longer collide and the table renders Codex turns like any other request. Feed-row counts, per-turn token and savings values, ordering, and per-session metrics/cost bookkeeping are unchanged; the `[{session request_id}]` log prefixes still use the session id so a session's log lines stay greppable together.

Scope: this is the backend root-cause fix. Hardening the dashboard `:key` against duplicate/`null` keys is a separate render-robustness change and is deliberately left to a follow-up (`Refs #310`); once the backend guarantees unique ids, the collision that blanks the table is gone. The comment's secondary `savings_percent.toFixed(0)` concern is already resolved on `main` (the row uses `formatOptionalPercent`).

[...]

PR #2921, opened 2026-08-11
feat(savings): carry cumulative actual output_tokens in history checkpoints

gglucass:
## Description

History checkpoints carry cumulative `cache_read_tokens` (and `cache_savings_usd`), which lets a consumer diff consecutive checkpoints into per-day / per-hour cache series - we use exactly that in the Headroom desktop app to show window-scoped cache-hit rates. The output-shaping dimension cannot do the same: checkpoints carry cumulative `output_tokens_saved` but not cumulative ACTUAL output tokens, so there is no denominator and no per-bucket output-reduction rate (`saved / (saved + actual)`) can be derived from the payload. Dashboards are left showing an all-time estimator figure under per-day charts.

This PR adds one cumulative `output_tokens` counter, symmetric with the existing cache fields:

- `record_request` accepts the request's actual output tokens - the caller in `prometheus_metrics.py` already had them in hand (it passes them to `record_lifetime_request` two calls earlier).
- The lifetime aggregate accumulates them as a lazy key, matching `output_tokens_saved`'s legacy-tolerant shape (older state files keep parsing; salvage behavior is identical to the existing output fields).
- Every history checkpoint emits the cumulative, and `_normalize_history_entry` coerces it through reload so persisted history retains the field (entries without the key normalize to 0, same as the cache fields did when they were introduced).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/savings_tracker.py`: `output_tokens` parameter on `record_request`; lazy lifetime accumulation; checkpoint emission; `_normalize_history_entry` coercion
- `headroom/proxy/prometheus_metrics.py`: pass the already-available `output_tokens` through to `record_request`
- `tests/test_proxy_savings_history.py`: new test covering accumulation across requests, checkpoint values, reload survival, and legacy entries without the key; two existing exact-shape fixtures extended with the new field

## Testing

### Test Output

```text
$ pytest tests/test_proxy_savings_history.py tests/test_output_shaping_rollup.py \
    tests/test_persistent_metrics_integration.py -q
======================== 45 passed, 1 warning in 8.46s =========================
$ ruff check (changed files)
All checks passed!
$ mypy headroom/proxy/savings_tracker.py
```

## Real Behavior Proof

- Environment: macOS (Darwin 24.6.0), Python 3.10 dev venv of this branch
- Exact command / steps: unit-level - the new test drives two `record_request` calls with `output_tokens=200/300`, snapshots the tracker, reloads it from disk, and normalizes a legacy entry
- Observed result: lifetime cumulative reaches 500; the two checkpoints carry 200 and 500; a tracker reloaded from the persisted file still reports 500 on the newest checkpoint; a legacy entry without the key normalizes to 0 rather than erroring
- Not tested: a live proxy deployment of this exact commit. The consumer that motivated it (Headroom desktop) currently approximates the same series by polling `/stats` cumulatives and will switch to checkpoint diffs once a wheel ships this field.

## Runtime Rollout Safety

- Stable/default behavior changed: Yes - history checkpoints carry a new cumulative `output_tokens` field and output-only requests now emit a checkpoint; savings values are unchanged.
- Kill switch / disable path: Not applicable - the field is additive and `_normalize_history_entry` coerces missing values to 0.
- Qualification impact: None.
- Rollback path: Revert; readers ignore the extra key and legacy entries without it normalize to 0.

## Review Readiness

## Checklist

N/A - one new numeric field on an internal persistence schema.

## Additional Notes

- Documentation checklist item unchecked: the checkpoint schema is not documented outside the code today; happy to add a schema note if maintainers want one.
- Related: #2919 (cache-breakpoint diagnostics) consumes the cache half of this checkpoint pattern; this PR completes the output half. Independent - neither depends on the other.


JerrettDavis (changes requested):
The new cumulative counter is not sufficient for the windowed denominator as wired. `record_request` still appends a history checkpoint only when `delta_tokens_saved > 0`, `delta_cache_read_tokens > 0`, or `delta_output_tokens_saved > 0`. A normal request with `output_tokens > 0` and no savings updates `lifetime["output_tokens"]` but emits no checkpoint; if it is the final request in a window (or until some later savings event), consumers diffing history omit those actual tokens and overstate output reduction. The new test gives both requests `output_tokens_saved=5`, so it cannot catch this. Please make actual-output activity participate in checkpoint emission (or otherwise guarantee a terminal/current checkpoint) and add a regression with a savings checkpoint followed by an output-only request, including the persisted/reloaded history value.

gglucass:
@JerrettDavis addressed in aa101765.

- Emission gate: `record_request` now computes `delta_output_tokens` alongside the savings deltas and appends a checkpoint when it is positive, so an output-only request (no compression, no cache read, no output savings) that moves `lifetime["output_tokens"]` always lands in history. Window-closing output activity can no longer be silently absorbed into a later savings checkpoint.
- Regression: `test_output_only_request_emits_checkpoint` records a savings request (emits checkpoint 1), then an output-only request (`tokens_saved=0`, `output_tokens_saved=0`, `output_tokens=300`), and asserts a second checkpoint exists with the cumulative `output_tokens == 500` while `output_tokens_saved`/`total_tokens_saved` stay flat. It then reloads the tracker from disk and asserts the checkpoint and its cumulative survive `_normalize_history_entry`.
- Growth bound: checkpoints remain trimmed by `_max_history_age_days` / `_max_history_points` in `_trim_history_locked`, so emitting on output-bearing requests does not change the retention ceiling.

`tests/test_proxy_savings_history.py` passes (41 tests), plus the adjacent savings suites (ledger offload, zero-price, persistent metrics, output shaping rollup).

gglucass:
Note on the red X: the `test (3)` / `test (4)` failures on aa101765 are inherited from `main`, not from this PR. `main` itself fails the identical three tests (push runs 31566354029 and 31568129622) — merge skew between this morning's #2412 (time-capped quarantine vs the tokenizer offload tests) and #2164 (per-turn WS request ids vs its own PERF-prefix assertion). Bisect details and test-only fixes in #2940; once that lands, a rebase or re-run here should be green. This PR's own suites (`tests/test_proxy_savings_history.py` and the adjacent savings suites) pass on this branch.

JerrettDavis (approved):
Re-reviewed the current head (`aa101765`) after the requested fix. Actual-output-only requests now emit a checkpoint, so a terminal request with no savings still advances the cumulative denominator instead of being omitted from a window. The new regression covers a savings checkpoint followed by an output-only request and verifies both the cumulative values and persistence/reload behavior. The additive field remains backward-compatible for legacy history entries. The two red broad shards match the unrelated repository-wide WebSocket timing and tokenizer-state failures; I found no blocker in this PR.

JerrettDavis (approved):
Re-reviewed the exact current head after the main merge and its conflicts in savings_tracker/prometheus_metrics. The merge preserves the approved invariant: actual-output-only requests still participate in checkpoint emission, output_tokens remains normalized/persisted, and the metrics caller still forwards actual output tokens. Main's newer per-layer savings pricing/attribution is integrated without conflating or dropping this cumulative denominator. Full current-head CI is green; no new blocking findings.

PR #2932, opened 2026-08-11
fix(toin): address patterns by signature hash

### LINKED ISSUE #2928: sig_hash[:12] truncates the composite auth|model|hash key, so /v1/toin patterns share one id and detail lookup is arbitrary
## Summary

## Environment

## Reproduction Steps

## Expected Behavior

## Actual Behavior

```python
"hash": sig_hash[:12],
```

```
```

## Impact

**Low.**

## Fix Options

## Additional Context

chopratejas:
## Summary

Fixes #2928.

TOIN v2 serializes pattern keys as `auth|model|signature_hash`, but the API took the first 12 characters of that composite key. Patterns commonly appeared as `unknown|unknown`, making all identifiers collide and detail lookup arbitrary.

## Changes

- extract the signature-hash component from v2 composite keys before exposing `hash`
- make detail lookup match against that same signature component
- preserve legacy bare-hash export compatibility through the existing deserializer

## Validation

- added list/detail round-trip tests with multiple tenant/model slices
- 78 focused proxy/TOIN tests passed (6 skipped)
- Ruff check/format and `git diff --check` passed

## Review notes

The documented 12-character display-prefix contract is preserved. As with any prefix API, two hashes sharing the same first 12 characters remain theoretically ambiguous; the fix removes the total collision caused by the composite-key prefix and keeps the endpoint backward compatible.

### DIFF
--- a/headroom/proxy/server.py
+++ b/headroom/proxy/server.py
@@ -179,7 +179,7 @@ from headroom.subscription.tracker import (
 )
 from headroom.telemetry import get_telemetry_collector
 from headroom.telemetry.beacon import is_telemetry_enabled
-from headroom.telemetry.toin import get_toin
+from headroom.telemetry.toin import _deserialize_pattern_key, get_toin
 from headroom.transforms import (
     CacheAligner,
     CodeAwareCompressor,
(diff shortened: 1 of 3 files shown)


JerrettDavis (changes requested):
The composite-key parsing fixes the `unknown|unkn` failure, but the returned identifier still does not uniquely identify a v2 pattern. TOIN deliberately keys each record by `(auth_mode, model_family, signature_hash)`, so the same tool signature can—and commonly will—exist in multiple tenant/model slices with different counters and recommendations. Both rows would now expose the same 12-character `hash`, and the detail endpoint would still return whichever slice iterates first.

Please make the list/detail round trip preserve the slice identity (for example, return an opaque URL-safe identifier derived from the full serialized key, or expose the slice fields and require them in detail lookup), and add a regression test containing two records with the same signature hash but different auth/model values and different expected detail payloads. A 404/409 for an ambiguous legacy hash prefix would also be safer than returning an arbitrary record. This is especially important because the issue’s expected behavior says each entry must be distinguishable.

JerrettDavis:
I’m consolidating the TOIN identifier work in #3217, which addresses the same list/detail parsing problem through a shared helper. The review concern about distinguishing auth/model slices still needs to be resolved there; closing this duplicate does not mean that issue is fixed. Thanks for helping pinpoint the broken identifiers.

PR #2933, opened 2026-08-11
fix(cli): honor explicit Claude model with 1m

### LINKED ISSUE #2915: `--1m` silently loses to an explicit `--model` claude_arg (e.g. `--model opusplan`), session stays capped at 200k

chopratejas:
## Summary

Fixes #2915.

`headroom wrap claude --1m --model opusplan` previously set `ANTHROPIC_MODEL`, but Claude Code gave the explicit CLI model argument precedence. The banner claimed 1M was active while the launched session remained capped at 200k.

## Changes

- parse `--model VALUE`, `-m VALUE`, `--model=VALUE`, and `-m=VALUE` in forwarded Claude arguments
- append `[1m]` to the explicit model argument that Claude Code will actually resolve
- keep `ANTHROPIC_MODEL` aligned with that resolved model
- fail before launch for the multi-model `opusplan` alias instead of claiming success
- report whether the effective model came from the explicit argument or environment
- support concrete model IDs such as newer Opus releases without hardcoding a new default

## Validation

- 52 wrap-helper tests passed
- existing Claude wrap/remote-control accuracy tests passed (119 total in the focused suite)
- regression coverage includes separated/inline model flags, environment fallback, alias rejection, and malformed flags
- Ruff check/format and `git diff --check` passed

## Review notes

Only the effective/last explicit model selection is rewritten; unrelated forwarded arguments remain byte-for-byte unchanged. The existing default model remains unchanged because Claude model availability is provider-controlled; explicit concrete IDs are now supported correctly.

### DIFF
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -303,6 +303,66 @@ def _resolve_1m_model(current: str | None) -> str:
     return base if base.endswith(_CONTEXT_1M_SUFFIX) else f"{base}{_CONTEXT_1M_SUFFIX}"
 
 
+def _resolve_1m_model_selection(
+    claude_args: tuple[str, ...],
+    environment_model: str | None,
+) -> tuple[tuple[str, ...], str, bool]:
+    """Apply ``[1m]`` to the model Claude Code will actually resolve.
+
+    Claude Code gives an explicit ``--model``/``-m`` argument precedence over
+    ``ANTHROPIC_MODEL``. Rewrite that argument when present; otherwise retain
+    the existing environment-based behavior. Mode aliases such as
+    ``opusplan`` select multiple underlying models and cannot safely carry one
+    suffix, so fail before launching with a misleading success banner.
+    """
+    args = list(claude_args)
+    model_index: int | None = None
+    inline = False
+
+    index = 0
+    while index < len(args):
+        argument = args[index]
+        if argument in {"--model", "-m"}:
+            if index + 1 >= len(args) or not args[index + 1].strip():
+                raise ValueError(f"{argument} requires a model when used with --1m")
+            model_index = index + 1
+            inline = False
+            index += 2
+            continue
+        if argument.startswith("--model=") or argument.startswith("-m="):
+            value = argument.split("=", 1)[1].strip()
+            if not value:
+                raise ValueError(
+                    f"{argument.split('=', 1)[0]} requires a model when used with --1m"
+                )
+            model_index = index
+            inline = True
+        index += 1
+
+    if model_index is None:
+        return claude_args, _resolve_1m_model(environment_model), False
+
+    if inline:
+        selected_model = args[model_index].split("=", 1)[1].strip()
+    else:
+        selected_model = args[model_index].strip()
+
+    alias_model = selected_model.removesuffix(_CONTEXT_1M_SUFFIX).lower()
+    if alias_model in {"opusplan"}:
+        raise ValueError(
+            "--1m cannot be combined with Claude Code model alias "
+            f"{selected_model!r}; pass a concrete Claude model with --model"
+        )
+
+    resolved_model = _resolve_1m_model(selected_model)
+    if inline:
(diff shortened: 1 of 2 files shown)


chopratejas:
Superseded by merged PR #2922, which fixes the same Claude --1m explicit-model precedence issue and closes #2915.

chopratejas:
Superseded by merged PR #2922.

PR #2942, opened 2026-08-12
fix(proxy): break the compact-every-turn loop near the real context window

gglucass:
## Description

Compression makes the usage reported back to the client reflect the *forwarded* (compressed) request, not the client's raw transcript. Clients whose auto-compaction keys off reported usage (Claude Code) therefore defer compaction — the selling point — but with no floor: the raw transcript can grow until even the compressed request exceeds the model's real context window. From that point every turn fails with a 400 `prompt is too long`, the client force-compacts on each error, and the session degrades into a compact-every-other-prompt loop. This came from a churned-user field report (2026-08-12): "Headroom cut off my context and started compacting every two prompts."

A second, related bug: `get_context_limit()` resolves purely from the model id, so a session running with the `context-1m` beta is treated as 5x over budget by compression pressure (`context_pressure` pins at 1.0) and gets maximum-aggression lossy compression from early on — the "cut off my context" half of the report.

This PR adds `headroom/proxy/context_guard.py` with three cooperating pieces, all passthrough until the danger zone:

1. `effective_context_limit` — the window the forwarded request is really subject to: the registry limit, raised to 1M when the client sent a `context-1m` beta token, capped by any limit learned from a real `prompt is too long` error. Also used for compression pressure at the `context_limit` call site.
2. `note_prompt_too_long` — parses `prompt is too long: N tokens > M maximum` from upstream 400 bodies (streaming and buffered error paths) and records M keyed by (model, has-1m-beta). This is how a client that believes 1M but whose account is capped at 200k stops looping after exactly one error — and why the learned limit must not leak between beta shapes.
3. `MessageStartGuard` — an SSE transformer applied to the client-bound bytes only: when the `message_start` usage total reaches 90% of the effective limit, it inflates the reported `input_tokens` so the client's gauge reads 95% of the window the client believes in, which trips the client's own graceful compaction *before* the 400 wall. Inert after a single decision (steady-state overhead is one `if`); savings metrics and telemetry parse the original upstream bytes, never the nudged ones.

Kill switch: `HEADROOM_CONTEXT_GUARD=0`.

Out of scope, deliberately: the buffered (non-streaming) success path is not nudged — the affected clients (Claude Code) always stream, and rewriting buffered bodies would break response byte-fidelity for a path with no known symptom. The batch endpoint's `context_limit` site is also untouched (batch has no interactive compaction loop).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- New `headroom/proxy/context_guard.py`: limit resolution (`believed_context_limit`, `effective_context_limit`), 400-learning (`note_prompt_too_long`), and the `MessageStartGuard` SSE usage rewriter, plus the `HEADROOM_CONTEXT_GUARD` kill switch.
- `headroom/proxy/handlers/streaming.py`: build the guard for anthropic streams (inside try/except, passthrough on any setup failure), route yielded chunks through `guard.feed()`, flush held bytes at stream end, and learn from streaming 400 bodies.
- `headroom/proxy/handlers/anthropic.py`: compression `context_limit` now uses `effective_context_limit` (1m-beta aware, learned-cap aware); the existing UPSTREAM_ERROR block also feeds `note_prompt_too_long`.
- `tests/test_proxy/test_context_guard.py`: 25 tests — unit coverage of limits/learning/rewriter (chunk splits, pings, malformed JSON, buffer cap, never-deflate) and four integration tests driving `_stream_response` with a mocked upstream (nudge on the wire, byte-identical passthrough far from the limit, kill switch, 400 learning).

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_proxy/test_context_guard.py -q
25 passed in 0.53s

$ uv run --frozen --extra dev pytest tests/test_proxy/ -q
272 passed, 1 warning in 71.42s (0:01:11)

$ uv run --frozen --extra dev pytest tests/test_sse_thinking_blocks.py tests/test_h2_stream_reset_retry.py \
    tests/test_proxy_streaming_ratelimit_headers.py tests/test_request_outcome.py tests/test_proxy_retry_429.py -q
81 passed in 1.53s

$ uv run --frozen --extra dev mypy headroom/proxy/context_guard.py

$ uvx ruff check headroom/proxy/context_guard.py headroom/proxy/handlers/streaming.py \
    headroom/proxy/handlers/anthropic.py tests/test_proxy/test_context_guard.py
All checks passed!
```

## Real Behavior Proof

- Environment: macOS 15 (Darwin 24.6.0), Python 3.10 project venv via uv, repo at this branch; live runs used a sandboxed proxy instance (this module injected) with Claude Code 2.1.206 as the real client.
- Exact command / steps: ran the four integration tests in `TestStreamResponseIntegration`, which drive the real `_stream_response` code path with a mocked upstream httpx response: (a) a `message_start` carrying a 185k-token usage against a 200k model limit (above the 90% trigger, below the 95% target, so the assertion can only pass if the rewrite happened), (b) the same stream at 50k tokens, (c) a 190k stream with `HEADROOM_CONTEXT_GUARD=0`, (d) a 400 body containing `prompt is too long: 213021 tokens > 200000 maximum` with `anthropic-beta: context-1m-2025-08-07`.
- Observed result: (a) the client-received `message_start` reports `input_tokens` 185000 inflated to 190000 (the 95% target) while the `message_stop` bytes pass through unmodified; (b) the client-received bytes are byte-identical to the upstream bytes; (c) byte-identical passthrough with the kill switch set; (d) the 400 is forwarded unchanged and `effective_context_limit("claude-sonnet-4-5", 200000, "context-1m-2025-08-07")` returns 200000 afterwards.
- Live end-to-end (added after the initial PR): drove a real Claude Code 2.1.206 session through a proxy running this code, with a learned limit seeded to the state a real prompt-too-long 400 produces. Result: the client-visible usage total read exactly 190000 (95% of the believed 200k window) and the interactive TUI showed "100% context used". This run is also why the second commit exists: with only message_start rewritten, the client kept displaying the raw values because it merges the final cumulative-usage message_delta over message_start. Forcing a prompt-too-long 400 on the next turn produced a clean "Compacting conversation..." -> summarization request -> retried turn -> correct response, with no user-facing error.
- Not tested: an organic (unseeded) 180k+-token conversation against the production API; the buffered non-streaming success path is deliberately not nudged (see Description). Note for reviewers: Claude Code 2.1.206 compacts reactively at prompt-too-long (with a threshold-armed precomputed summary) rather than proactively pre-send, so the nudge's role there is an honest gauge/warnings and threshold arming; older clients and other harnesses that compact proactively off reported usage get the pre-wall compaction directly.

## Review Readiness

## Checklist

- [x] I did **not** edit `CHANGELOG.md` — it is generated by release-please from the Conventional Commit PR title (a CI guard enforces this)


gglucass:
Follow-up commit 99750ca after a live end-to-end run: driving a real Claude Code session through the guard showed the nudged message_start reaching the wire (proxy logged the rewrite) but the client's surfaced usage still carried the raw values -- the final message_delta now includes cumulative input/cache usage and the client merges it over message_start. StreamUsageGuard (renamed from MessageStartGuard) therefore rewrites both events when armed. Below the trigger the guard still goes inert on the first event, so steady-state traffic remains byte-identical. New unit + integration coverage for the delta rewrite (29 guard tests total).

JerrettDavis (approved):
Reviewed the current head (`99750ca`) across beta-aware limit resolution, learned-cap isolation, both 400-learning paths, SSE framing/chunk splits, cumulative usage arithmetic, final message_delta handling, telemetry separation, buffer bounds, and the kill switch. The guard is fail-open and does not alter far-from-limit traffic or internal accounting; it also never lowers an already-higher reported total. The focused and adjacent proxy coverage is substantive. I found no blocking correctness issue. The two failing matrix shards are unrelated existing WebSocket timing and tokenizer-state failures.

JerrettDavis (approved):
Re-reviewed the post-approval main merge. The feature commits remain unchanged: the near-limit guard is beta-aware, fail-open, bounded, and updates final cumulative usage without lowering higher provider totals or contaminating internal accounting. Current main's surrounding streaming changes compose cleanly, the branch is mergeable, and the diff has no whitespace errors. No blocking finding.

TerrysPOV:
Field report that matches this PR's description, including the `context-1m` half — with proxy logs.

**Environment**

- `headroom-ai` 0.36.5 (latest on PyPI at time of writing), installed via `uv tool`
- Claude Code 2.1.245, model `opus` **with the `context-1m` beta active** (`opus[1m]`)
- macOS, proxy started as `headroom proxy --port 8787`, `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`

**Symptom**

Auto-compaction sticks at ~95% and never completes. Claude Code shows `API error · Retrying in 0s · attempt 1/10`, and the cycle repeats every ~3 minutes indefinitely. Normal turns keep working; only compaction fails.

The status line shows the two sides disagreeing, which I think is the reported-usage divergence this PR describes:

```
ctx:17%     ...     0% until auto-compact
```

**Proxy logs**

Every compaction attempt is an identical 1.84 MB body, three internal retries, ~190s, then 502:

```
event=outbound_request forwarder=anthropic_messages method=POST path=/v1/messages
  body_bytes=1930860 body_mutated=true mutation_reasons=structural_diff_vs_original
WARNING - Request failed (attempt 1), retrying in 518ms: Server disconnected
WARNING - Request failed (attempt 2), retrying in 2931ms: Server disconnected
ERROR   - [hr_...] Request failed: RemoteProtocolError: Server disconnected
INFO    - event=proxy_inbound_response method=POST path=/v1/messages status=502 duration_ms=190000
```

Consecutive attempts, all identical in size:

```
10:44  1.84MB -> disconnected
10:47  1.84MB -> disconnected
10:54  1.84MB -> disconnected
10:58  1.84MB -> 502 after 189.6s
11:01  1.84MB -> 502 after 188.5s
11:04  1.84MB -> 502 after 190.0s
```

**Two things I checked that may narrow it down**

1. **Headroom is not inflating the payload.** Pairing `proxy_inbound_request content_length` against the subsequent `outbound_request body_bytes` gives a ratio of exactly 1.00 (1.84 MB in, 1.84 MB out) on every attempt. The oversized request is what the client sends, consistent with "the raw transcript can grow until even the compressed request exceeds the real window".
2. **Beta headers are forwarded intact** — `event=beta_header_merge provider=anthropic client_betas=12 sticky_betas=12 headroom_added=`. So the `context-1m` token reaches upstream; it is the internal limit resolution, not header stripping, that is at issue — matching the `get_context_limit()` half of this PR.

Stage timings show the proxy itself is fast and effectively all the time is spent waiting on the upstream connection:

```
total_pre_upstream ~100ms   compression_first_stage ~22ms   read_request_json ~18ms
upstream_connect   ~189000ms  -> Server disconnected
```

**Onset correlates with install**

First error in `proxy.log` is `2026-08-23 13:21`, three minutes after `headroom-ai` was installed at 13:18. Error counts by day: 08-23 ×1, 08-24 ×86, 08-25 ×11 (still climbing until I bypassed).

**Workaround**

Confirming what #3083 states — it does not self-heal, and pointing the client away from the proxy is the only recovery. Worth noting for others that removing `ANTHROPIC_BASE_URL` alone is not enough if the project has the `SessionStart` → `headroom wrap selfheal` hook installed, because that re-wraps the project on the next session start. Both have to be disabled.

Happy to test a build of this PR against the same 1M-context session if that would help.

TerrysPOV:
Following up on my field report above — I ran a full review over the diff (base `6262c28`, merge ref head `6e0fce5`) before vendoring this locally. Tests pass cleanly on my side first: **29/29** in `test_context_guard.py` and **280/280** across `tests/test_proxy` on a fresh `.[all]` install.

**Credential/body handling looks clean.** The guard reads only `anthropic-beta`, never touches `x-api-key`/`authorization`, doesn't persist headers or bodies, and logs integers plus `model`/`request_id`. Telemetry keeps parsing the unmodified upstream bytes, so savings accounting isn't polluted by the nudge. Flagging that explicitly since it's the thing most people would worry about when putting an unmerged proxy patch in front of their traffic.

Ten findings below. I've marked the three I verified directly against the source; the rest are from the review pass and I have not individually confirmed them.

### High

**1. `handlers/anthropic.py:3240` — `note_prompt_too_long` is called outside the `try` that built `err_msg`.**
On the JSON branch `err_msg = err_body.get("error", {}).get("message", "")`, which is `None` if an upstream sends `{"error":{"type":"...","message":null}}`, and a dict/list for gateways that nest error detail. `_PROMPT_TOO_LONG_RE.search(None)` raises `TypeError`, which propagates out of `_buffered_ccr_operation()` — a recoverable upstream 400 becomes a proxy exception, the client never sees the real body, and the diagnostic dump below never runs. `note_prompt_too_long` already normalises `bytes`, so the narrow fix is to return `None` for anything that isn't `str`/`bytes`.

**2. `context_guard.py:119` — the learned maximum has no sanity bounds.** *(verified against source)*
`maximum = int(match.group(2))` goes straight into `_learned_limits` with no floor, no ceiling, and no comparison to the believed limit. Via `anthropic.py:1340` that value becomes `model_limit` for the compression pipeline, so any upstream reporting a small maximum — an alias resolving to a small-context model, an Anthropic-compatible gateway with different wording, a repointed `ANTHROPIC_BASE_URL` — permanently pins e.g. `effective_context_limit = 4096`. The compressor is then asked to fit the transcript into 4096 tokens and the trigger fires on every request. Suggest ignoring values below ~8k and never above `believed_context_limit`.

**3. `context_guard.py:60` — `_learned_limits` is an unscoped process global.**
Keyed only on `(model, has_1m_beta)`: no per-key/session/tenant scoping, no TTL, no upward revision, and `reset_learned_limits` is documented as a test hook. On a shared proxy, one client's 400 permanently clamps every other client on that model — including a client whose key genuinely has 1M, which then gets compressed against the lower budget *and* nudged to ~95% early, i.e. force-compacted every turn. That's the loop this PR removes, reintroduced from a different direction. Recovery needs a process restart.

### Medium

**4. `context_guard.py:223` — the armed-state disarm is a substring match, not an event-type match.** *(verified against source)*

```python
if b"message_delta" not in event:
    return event
self._done = True
```

That tests the whole event body, so a `content_block_delta` whose *text payload* contains the literal `message_delta` sets `_done = True`. `_rewrite_message_delta` then returns the event untouched (`payload["type"] != "message_delta"`), and the real final cumulative-usage `message_delta` is never rewritten — the client merges the un-nudged delta over the nudged `message_start` and the nudge is silently lost. This is self-triggering for anyone whose session discusses SSE or this PR, and it fails in exactly the long-transcript sessions the guard is for, with no log line. `b"event: ping" in event` at line 203 and the `message_start` check at 210 are the same class. Parsing the `event:` line fixes all three.

**5. `handlers/anthropic.py:1340` — pre-learning 1M optimism is a regression on first encounter.**
Before anything is learned, `effective_context_limit` returns 1M for any session carrying a sticky `context-1m` token, cutting compression pressure ~5x. For an account not actually provisioned at 1M the forwarded request is *larger* than pre-PR, so the first turn past the real cap hits `prompt is too long` with less compression than the status quo — and the nudge can't help, since the trigger is 0.9 × 1M. Because the limit is memory-only this recurs after every restart, per `(model, beta)`. A persisted or config-backed limit, or a conservative default until confirmed, would avoid the guaranteed 400.

**6. `context_guard.py:94` + `streaming.py:1129` — non-streaming path gets the raised budget but no nudge.**
`StreamUsageGuard` is wired only into `_stream_response`, while the raised `effective_context_limit` also applies to non-streaming `/v1/messages`. A `stream: false` client with the 1m beta therefore gets less compression with no compensating nudge — strictly more likely to hit the wall than before this PR.

**7. `context_guard.py:205` — a pre-`message_start` SSE comment disarms the guard permanently.**
The "first data-bearing event must be `message_start`" scan treats anything else as end-of-scan. `event.strip() == b""` doesn't catch comment frames (`: keep-alive`) or a `retry:` field, so any intermediary that prepends one (nginx/CDN keepalive shims being the common case) makes the guard inert for the whole deployment, silently. Skip lines starting with `:` and keep scanning.

**8. `context_guard.py:256` — the nudged `input_tokens` isn't marked as synthetic.**
In a 1m session the guard reports ~950k where the forwarded request was ~190k. Nothing in the SSE distinguishes the gauge from the ledger, so every downstream consumer takes it at face value: Claude Code's usage display, LangChain/LangGraph token callbacks, OTel usage spans, client-side billing reconciliation. An `x-headroom-usage-nudged` response header or a custom SSE field would let consumers tell them apart.

### Low

**9. `streaming.py:1555` — `flush()` only runs on the normal completion path.** It sits inside the `try` before the `except` clauses at 1640+, so a mid-stream `httpx.ReadError`/`RemoteProtocolError` drops the held-back partial event. Moving it to `finally` covers that. Separately, the comment at 1588 ("may hold back bytes only until the first complete event") is no longer accurate once armed — the guard re-frames on `\n\n` boundaries until the final `message_delta`, and that invariant is load-bearing for reviewers.

**10. `handlers/anthropic.py:4229` — batch path bypasses `effective_context_limit`.** Still uses `get_context_limit(model)` directly as `model_limit`, so the same model has two budgets depending on endpoint. Probably deliberate given batch has no auto-compaction, but undocumented.

### On the arithmetic

The nudge itself checks out: `total >= target_total` short-circuits before any deflation, and `cache_read + cache_creation <= total < target_total` keeps the rewritten `input_tokens` non-negative in both rewrite paths. Event framing, ordering across the `_MAX_GUARD_BUFFER` bail-out, and the `_finish()` release path preserve byte order with no duplication or loss.

Happy to open a PR against your branch for #4 and #2 if useful — they're the two that would bite a single-user deployment, and #4 in particular defeats the fix silently.

gglucass:
The failing `test-agno` check is not from this branch: agno 3.0.0 (released 2026-08-24) removed `agno.models.metrics`, and the job installs the latest agno, so it is red on main and every open PR since yesterday. Fix in #3260 - after that merges, a rerun/refresh here should restore green. The feature commits are unchanged since the last approval.

gglucass:
Update: refreshed from main - `test-agno` passes now, but `lint` will fail on this run for the same main-inherited reason as everywhere else: #3260 and #3246 overlapped on the agno 3.0.0 compat fix and left a dead import that F401s. Cleanup in #3268; a final refresh after it merges should make this fully green. Feature commits still unchanged since approval.

JerrettDavis (approved):
Re-reviewed the substantive hardening commit and the refreshed head after updating from current main. The learned cap is now bounded, expiring, and credential-scoped; diagnostic parsing is fail-open; SSE framing no longer mistakes payload text or keepalives for protocol events; buffered and batch paths use the same effective-window policy; and held stream bytes are drained on all exits. The prior lint failure was solely the now-fixed Agno import on the old base. No remaining correctness blocker found; approval restored pending the newly triggered checks and final human review.

TerrysPOV:
Field report from someone running this branch, plus a status check.

**Still the only path to the fix.** `headroom/proxy/context_guard.py` does not exist on `main` — it is introduced here and nowhere else. v0.37.0 shipped eight proxy changes (session state #3261, sidecar unification #3271, Copilot attribution #3277, WebSocket tokens #3305, atomic savings flush #3231, protected reads #3238, Kimi routing #3223, Windows process tree #3313) and none of them addresses the compact-every-turn loop near the real context window. So there is no alternative route by which this got fixed, and anyone hitting the loop today still has to unset `ANTHROPIC_BASE_URL` to avoid it — which is what we have been doing since August.

**Current state, for whoever picks this up:** approved three times (12 Aug, 21 Aug, 26 Aug), `MERGEABLE`/`CLEAN`, last moved 26 August, now 28 commits behind `main`. The CI failures cited in the thread were both inherited from `main` rather than from this branch — the `agno` 3.0.0 breakage and the repo-wide lint issue — and the branch has since taken the fix for the former.

**What we are running.** A build from `6e0fce5` plus one local patch that parses the SSE `event:` line rather than substring-matching frames. Reading the 26 August approval, that same class of issue was fixed properly here, so our local patch is redundant against the current head and we will drop it on the next rebuild. Mentioning it only as evidence that the SSE framing concern was real in practice and not theoretical.

**The question:** is anything still blocking the merge, or is it waiting on a maintainer with the bit? Happy to rebase, re-test, or split it if the size is the obstacle — 1,305 lines across four files is a lot to land in one go, and the guard module plus its tests could go in ahead of the handler wiring if that helps it move.

Not chasing for the sake of it. It is approved, mergeable, and drifting a further week from `main` each time it waits, which eventually costs another rebase and another review of work that has already been reviewed three times.

gglucass:
Nothing is blocking from this side. Refreshed against today's `main` (a199488) - 74 commits, clean merge, no conflicts. `tests/test_proxy/` is 329 passed, `test_context_guard.py` 39 of those, ruff clean on the three touched files.

On your SSE patch being redundant: it is, and by construction rather than by luck. `StreamUsageGuard.feed` buffers until a `\n\n` event boundary, splits there, and reads the type off the `event:` field (`_sse_event_type`) - it never substring-matches into a frame, and it holds bytes only until the first complete event, so a `message_start` split across chunk boundaries is handled the same as one that arrives whole. Anything non-data-bearing (ping, comment, retry) passes through without arming or disarming it. You can drop the local patch on the next rebuild.

Your read of the state is right otherwise: three approvals, `MERGEABLE`/`CLEAN`, and the CI failures earlier in the thread were `main`'s (agno 3.0.0, repo-wide lint), not this branch's.

On splitting: happy to, but the 1,305 lines are 505 of `context_guard.py` plus 581 of its tests - 1,086 of the total is the module and its coverage, which has to land together to be reviewable at all. What could go separately is the two handler wire-ups (`anthropic.py` 135, `streaming.py` 87), landing behind the existing `context_guard_enabled()` gate so the module ships inert first. Say the word and I will split it that way; otherwise it is waiting on the maintainer bit.

JerrettDavis (approved):
Re-reviewed the current head after its main update. The contributor change remains sound, the focused context-guard suite passes locally (39 passed), and the current head is conflict-free with all attached checks complete and no failures. Approved.

PR #2943, opened 2026-08-12
fix(proxy): preserve merged session and quarantine contracts

JerrettDavis:
## Description

Forward-fixes two integration contracts exposed while auditing the large August 12 merge batch on `main`.

The Codex WebSocket request-ID hardening correctly gave every emitted dashboard/feed row a unique ID, but it also changed the human-readable `PERF` prefix from the stable WebSocket session ID to that per-emission ID. That broke operator correlation and the contract documented by the original merge. This PR separates storage identity from log correlation: rows remain unique, while `PERF` lines remain grouped under the session ID.

The same audit found two tokenizer quarantine tests still modeling the pre-time-cap behavior. Timeout debt no longer activates quarantine after its deadline expires. The tests now establish a live deadline and therefore continue to exercise the intended fail-open branch without weakening the production guard.

## Changes

- Add an optional `RequestOutcome.perf_request_id` correlation field, defaulting to the existing `request_id` behavior for all current callers.
- Set that field to the stable session ID for both per-turn and residual Codex WebSocket emissions.
- Strengthen the lifecycle regression test to prove the unique feed-row ID is not used as the `PERF` prefix.
- Update tokenizer quarantine tests to model an active, time-capped quarantine.

This is a forward fix; it does not revert the unique WebSocket request IDs or the time-capped quarantine behavior.

## Merge-batch audit context

- Audited 48 squash merges from `12149f74466c08b69be8d5fe751425be63c2fda4` through `941c25d31e6c6e0b436c307cbe212771ff76b45f`.
- Reviewed repeated-touch hotspots in the OpenAI/Anthropic handlers and proxy server.
- Confirmed the batch has no deleted or renamed paths and no revert/supersession commits.
- Reproduced the deterministic failures on the merged head and confirmed the corresponding pre-batch tokenizer tests passed; the new WebSocket assertion was introduced by the batch and exposed the dropped correlation contract.

## Validation

- `ruff check` — passed
- `ruff format --check` — passed
- `mypy headroom` — passed (517 source files)
- Focused proxy/outcome/tokenizer suite — 99 passed
- CI-equivalent Python suite in four `pytest-split` shards — 10,797 passed, 603 skipped
- `cargo fmt --all -- --check` — passed
- `cargo clippy --workspace -- -D warnings` — passed
- New beta-sticky integration suite — 9 passed
- Simulator-backed proxy E2E — 7 passed
- Real local Headroom process against a real local FastAPI upstream:
  - `/livez` healthy
  - `/readyz` ready (Kompress correctly optional/degraded in minimal passthrough mode)
  - OpenAI `/v1/chat/completions` round trip passed
  - Anthropic `/v1/messages` round trip passed
- Local `act` exercised the push workflow's change-detection job and entered the real lint job. `act` required local no-op cache/setup shims because its runner action post-hooks/pip pairing are incompatible with this macOS Docker environment; the actual Ruff/mypy validations above were run natively and passed.

## Risk

Low. The new field is optional and preserves existing behavior by default. Only Codex WebSocket outcome emitters opt into a separate log correlation ID; dashboard/feed row identity remains unchanged.

### DIFF
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -7922,6 +7922,9 @@ class OpenAIHandlerMixin:
                                 RequestOutcome(
                                     # Per-emission ids keep dashboard request-log keys unique.
                                     request_id=await self._next_request_id(),
+                                    # PERF remains grouped under the stable WS
+                                    # session id even though feed rows are unique.
+                                    perf_request_id=request_id,
                                     provider="openai",
                                     model=model_for_metrics,
                                     original_tokens=max(0, input_delta) + max(0, saved_delta),
@@ -8498,6 +8501,7 @@ class OpenAIHandlerMixin:
                     RequestOutcome(
                         # Per-emission ids keep dashboard request-log keys unique.
                         request_id=await self._next_request_id(),
+                        perf_request_id=request_id,
                         provider="openai",
                         model=model_name,
                         original_tokens=residual_input_tokens + residual_tokens_saved,
--- a/headroom/proxy/outcome.py
+++ b/headroom/proxy/outcome.py
@@ -87,6 +87,12 @@ class RequestOutcome:
     output_tokens: int
     tokens_saved: int
     attempted_input_tokens: int
+    # Optional correlation id for the human-readable PERF line. Most requests
+    # use ``request_id`` for both storage identity and log correlation. A
+    # long-lived WebSocket session is different: each emitted feed row needs a
+    # unique request id, while operators still need every line from the socket
+    # under one greppable session prefix.
+    perf_request_id: str | None = None
     # Optional so the 18 existing emit sites need no change: a handler that has
     # no provider count (or whose optimized_tokens is already provider-scaled)
     # leaves it 0 and billing falls back to optimized_tokens, exactly as before.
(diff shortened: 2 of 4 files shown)


Parideboy:
Independently reached the same fix in #2950 before spotting this one; closed mine as a duplicate since yours came first. Confirming the diagnosis matches on both counts, reproduced locally against `upstream/main` @ `941c25d3`:

- `test_count_tokens_offloaded_*` — #2412's guard is `timed_out_in_flight > 0 and now < self._compression_quarantine_deadline`, and the unarmed `0.0` deadline reads as already lapsed, so a real `HuggingFaceTokenizer(model='qwen2.5-coder', tokenizer='Qwen/Qwen2.5-7B')` resolves instead of the fail-open estimator.
- `test_ws_session_log_prefix_uses_session_id` — the PERF line logs as `[req-ws-2]` because the session consumes the first id and the per-turn outcome mints the second, while `emit_request_outcome` builds the prefix from the same field that keys the feed row.

Your extra `assert f"[{handler.logger.entries[0].request_id}] PERF" not in caplog.text` is a genuinely better test than what I had.

One thing worth surfacing: **#2940 is mutually exclusive with this PR**, not just overlapping. It asserts `"[req-ws-1] PERF" not in caplog.text` where this asserts `"[req-ws-1] PERF" in caplog.text` — opposite answers to whether the PERF prefix is the session id or the per-emission row id. Whichever merges first will break the other's test, so the two probably need settling together rather than in sequence. #2164's PR body ("the `[{request_id}]` PERF/log prefixes keep the session id for operator correlation") supports this PR's reading, but that's a maintainer call.

Issue #2754, opened 2026-08-04
[FEATURE] Ability to use pre-installed Serena


AlexeyRaga:
## Problem Statement

Currently Headroom uses `uv` to install and use `Serena`. 
But in systems that already have `Serena` installed (and configured as an MCP for `claude`) it is not needed. The executable is already in the path, everything is in place and can be used.

Currently it requires me to have `uv` available, then gives the following log:
```
  MCP retrieve tool: existing config differs (env {'HEADROOM_PROXY_URL': 'http://127.0.0.1:8788'} -> {}). To update: headroom mcp install --proxy-url http://127.0.0.1:8787 --force
  Serena MCP: existing config differs (command '/nix/store/qkhkmyil9hbq8w52sxkl8zm0zggl0rn3-serena-1.6.1/bin/serena' -> 'uvx'; args ['start-mcp-server', '--project-from-cwd', '--context', 'claude-code', '--open-web-dashboard', 'False'] -> ['--from', 'git+https://github.com/oraios/serena', 'serena', 'start-mcp-server', '--project-from-cwd', '--context', 'claude-code', '--open-web-dashboard', 'False']). To update: update or remove the existing serena MCP entry, then rerun headroom wrap
  Serena: project pre-indexed (symbol cache warmed)
```

and I am not sure if it is an error or a warning, or whether Headroom will use Serena correctly or not.

## Proposed Solution

Maybe detect if `serena` executable is available and use it?

AlexeyRaga:
Additionally, `headroom` seems to always use `serena` from `main`, so when there is a breaking change in `serena`, it brings it in and makes it unusable.

For example, the latest version of `serena` changed configuration from `languages` to `language_servers`.

When `headroom` starts, it tries to run `main` `serena`, which creates a new config file. 
Then it realises claude config already have Serena MCP registers and backs off with this message:
```
Serena MCP: existing config differs (command 'serena' -> 'uvx'; args ['start-mcp-server', '--project-from-cwd', '--context', 'claude-code', '--open-web-dashboard', 'False'] -> ['--from', 'git+https://github.com/oraios/serena', 'serena', 'start-mcp-server', '--project-from-cwd', '--context', 'claude-code', '--open-web-dashboard', 'False']). To update: update or remove the existing serena MCP entry, then rerun headroom wrap
```

But it makes existing `serena` fail because the config created by the `main` version one is incompatible.

Issue #2674, opened 2026-07-30
[BUG] Serena MCP -32000 with OpenCode


LunaticMuch:
## Description

Serena MCP seems not working right with opencode

```
2 MCP Servers
•headroom Connected
•serena MCP error -32000: Connection closed
No Formatters
1 Plugins
•/Users/stefano/.pyenv/versions/3.12.11/lib/python3.12/site-packages/headroom/providers/opencode/_dist/entry.opencode.js @latest
```

## To Reproduce

1. Install headroom `pip install "headroom-ai[all]"`
2. run `headroom wrap opencode`

## Expected Behavior

The MCP should be connected and green

## Actual Behavior

Serena MCP does not connect

## Environment

- **Headroom version**: 0.33.0
- **Python version**:  3.12.11
- **OS**: macOS 26.5.2
- **LLM Provider**: NanoGPT (via OpenAI compatible)

Issue #880, opened 2026-06-12
[BUG] headroom mcp serve never exits after stdio client disconnects — orphaned processes accumulate (3 alive >27h)


fuzemobi:
## Summary

`headroom mcp serve` (stdio transport) never exits when its MCP client disconnects. Each Claude Code / Claude Desktop session spawns a fresh server, and the old ones stay alive indefinitely. On my machine right now there are **5 concurrent orphaned `headroom mcp serve` processes, 3 of them alive for 27+ hours** after the sessions that spawned them ended.

```
  PID    RSS     ELAPSED COMMAND
27617  60640       21:10 .../Headroom/headroom/runtime/venv/bin/headroom mcp serve
40222   4144 01-03:19:56 .../Headroom/headroom/runtime/venv/bin/headroom mcp serve
 3028  15792    02:19:58 .../Headroom/headroom/runtime/venv/bin/headroom mcp serve
 4016   4144 01-04:33:06 .../Headroom/headroom/runtime/venv/bin/headroom mcp serve
41004   4144 01-03:18:29 .../Headroom/headroom/runtime/venv/bin/headroom mcp serve
```

(ELAPSED `01-03:19:56` = 1 day 3 h. Only one of these belongs to a live session.)

## Environment

- headroom-ai 0.21.39 (Headroom.app bundled runtime, Python 3.12.12); pipx 0.24.0 also installed
- macOS 26.3.1 ARM64 (Apple Silicon)
- Client: Claude Code / Claude Desktop, stdio transport
- Registration: `claude mcp` entry → `.../Headroom/headroom/runtime/venv/bin/headroom mcp serve` with `HEADROOM_PROXY_URL=http://127.0.0.1:6767`

## Steps to reproduce

1. Register headroom as a stdio MCP server in Claude Code.
2. Start a session (server spawns, connects fine: "Successfully connected (transport: stdio)").
3. End/kill the client session.
4. `pgrep -fl "headroom mcp serve"` — the server process is still running, and stays running for days.

## Expected

A stdio MCP server should detect EOF/close on stdin when the client goes away and exit. (Most MCP SDK servers exit on stdin EOF; something here keeps the event loop alive — possibly a background thread or the proxy HTTP client keeping the loop referenced.)

## Impact

Process/memory leak that grows with every session; users who restart sessions frequently accumulate dozens of zombie servers. Also makes "which server is my session actually talking to" undiagnosable.

fuzemobi:
[...]

Issue #615, opened 2026-06-05
Process leak: headroom init persistent-task spawns hundreds of orphaned headroom proxy processes


Happy2Git:
Environment

  - headroom version: 0.22.4 (installed via uv tool install headroom-ai[all])
  - OS: macOS 15.x (Darwin)
  - Python: 3.12
  - Agents configured: headroom init -g claude codex (both Claude Code and Codex)

  Problem

  Over a few days, headroom proxy processes accumulated to 350+ instances, consuming ~5GB of memory and measurable CPU. All are orphaned (parent PID = launchd/1).
  Only one process actually binds port 8787; the rest are stuck waiting or in a retry state.

  After a pkill -9, the count dropped from 350 to 3 (the current session's processes).

  Evidence

  From ~/.headroom/logs/proxy.log:
  - 377 "Headroom Proxy started" entries
  - 43 "SESSION SUMMARY" (clean shutdown)
  - 0 port bind errors — all instances that logged startup successfully bound the port

  From ~/.headroom/deploy/init-user/runner.log:
  - 60 occurrences of ERROR: [Errno 48] address already in use — these are the proxy instances that lost the port race and exited cleanly
  
  From ~/.claude/plugins/cache/headroom-marketplace/headroom/0.22.3/.in_use/:
  - 36 session marker files, 35 of which belong to dead sessions (never cleaned up)
  
  The math: 377 startups - 43 clean shutdowns ≈ 334 processes that never exited properly. Close to the 350 found running.

  Root Cause Analysis

  Race condition in _ensure_profile_running (headroom/cli/init.py:525):

  1. The proxy takes 3-7 seconds to start (loading Kompress ONNX model, Magika detector, etc.)
  2. The health check uses 1 second timeout (wait_ready(timeout_seconds=1))
  3. During the startup window, subsequent hook invocations see "not ready" and each spawns a new headroom install agent run via start_detached_agent()
  4. start_detached_agent() does NOT check if a previous runner is still alive before spawning a new one
  5. The winner binds port 8787; the losers crash with Errno 48

  Aggravated by multi-agent hooks (Claude Code + Codex):

  The manifest targets: ["claude", "codex"] means hooks fire on SessionStart AND PreToolUse for BOTH agents. When the proxy is down, multiple hooks fire
  simultaneously, each spawning a new proxy attempt.

  No cleanup of old processes:

[...]

PR #2945, opened 2026-08-12
fix(wrap): stop the Serena pre-index stalling the launch path for 300s

### LINKED ISSUE #2938: [BUG] wrap: serena project index blocks the launch path for 300s on a hidden interactive prompt, and leaks an orphaned process on timeout
## Summary

`headroom wrap <agent>` can stall for a full 300 seconds before the agent launches, because `_index_serena_project` runs `serena project index` synchronously on the launch path and that command blocks on an interactive prompt whose output Headroom has captured. The user sees no prompt, no progress and no error — only a wrapper that appears to hang. When the timeout finally expires, the process that was actually doing the work is not killed; it is reparented to PID 1 and survives indefinitely, so every stalled wrap leaks one process.

This affects any project where Serena has no `.serena/project.yml` yet and auto-detects more than one language, which is the normal state the first time a project is wrapped.

## Impact

- Every first `headroom wrap` in a mixed-language project costs 300 seconds of dead time before the agent starts.
- The delay is silent. Because `capture_output=True` swallows the prompt, there is nothing on screen explaining the wait.
- Each occurrence leaks a `serena project index` process that never exits.
- The pre-index never succeeds in this state, so the symbol cache it exists to warm is never warmed. The 300 seconds buy nothing.

## Reproduction

Any project directory with no `.serena/project.yml` and source files in more than one language (the case below has bash and TypeScript at 44.44% each, plus three other detected servers):

```python
# Mirrors cli/wrap.py:1700-1714 exactly, with timeout=20 instead of 300.
import subprocess, time
from subprocess import run

t0 = time.time()
try:
    r = run(["uvx", "--from", "git+https://github.com/oraios/serena",
             "serena", "project", "index"],
            capture_output=True, text=True, timeout=20, cwd="/path/to/project")
    print("returncode:", r.returncode, "elapsed:", round(time.time() - t0, 1))
except subprocess.TimeoutExpired as e:
    print("TimeoutExpired after", round(time.time() - t0, 1))
    print("stdout captured:", (e.stdout or b"")[-400:])
```

Output:

```
TimeoutExpired after 20.0 s
stdout captured: b"...Detected and enabled main language server 'bash' (44.44% of source files).
Additionally detected 4 other applicable language servers.
...
Which additional language servers do you want to enable?
Enable typescript (44.44% of source files)? [y/N] "
```

Then check for the leak, after `run()` has already returned:

```
$ ps -eo pid,ppid,etime,command | grep "serena project index"
27566     1       00:19  .../bin/serena project index
```

`PPID` is 1. `subprocess.run` killed the `uvx` wrapper it spawned, but the Python grandchild that `uvx` execs is in the same position and is never signalled.

## Evidence from a real wrap

Timings from a wrapped session on a 176-file, 1.0 MB project:

```
PID    STAT  STARTED   COMMAND
17637  S+    10:57:03  headroom wrap opencode        <- invoked here
17650  S+    10:57:04  serena project index          <- spawned 1s later, later PPID=1
18915  S+    11:02:03  opencode                      <- agent starts, +300s exactly
```

PID 17650 was still alive 19 minutes later having consumed 0.52 s of CPU — blocked, not working. No `.serena/` directory was ever written.

## Root cause

`cli/wrap.py:1700-1714` (v0.34.0):

```python
result = run(
    ["uvx", "--from", "git+https://github.com/oraios/serena",
     "serena", "project", "index"],
    capture_output=True,
    text=True,
    timeout=300,
    cwd=str(Path.cwd()),
)
```

`serena project index` auto-creates `project.yml` when none exists, and that auto-creation asks one `[y/N]` question per additionally-detected language server. Three things combine:

1. `stdin` is not redirected, so Serena believes it can prompt.
2. `capture_output=True` pipes stdout, so the question never reaches the terminal.
3. The call is synchronous on the launch path, so the agent waits out the entire 300-second timeout.

The call site is `_setup_serena_mcp` at `cli/wrap.py:1808`, which every wrap subcommand invokes before `_launch_tool` — for example `wrap opencode` at `cli/wrap.py:6959`.

Two comments in the source state the opposite of the observed behaviour and should be corrected along with the fix:

- `cli/wrap.py:1692-1694` — *"Timeout-guarded and best-effort: Serena also indexes lazily on demand, so a failure or timeout here never blocks the wrap."*
- `cli/wrap.py:1790` — *"Both are best-effort and non-fatal — neither blocks the wrap."*

Both are accurate about intent and wrong about effect: a timeout blocks the wrap for exactly as long as the timeout.

## Proposed fix

The three parts are independent and each is worth doing on its own.

**1. Never let a launch-path subprocess prompt.** Passing `stdin=subprocess.DEVNULL` is a one-line change and is sufficient to remove the 300-second stall. Verified with the reproduction above:

```
returncode: 1  elapsed: 1.4 s
stderr tail: Project configuration auto-generation failed after 0.014 seconds
             Error: EOF when reading a line
```

1.4 seconds and a clean non-zero exit, instead of 300 seconds of silence.

**2. Make the pre-index actually able to succeed.** With `stdin=DEVNULL` alone the index still does not run, so the feature stays dead. `serena project index` accepts `--language/--ls`, documented as *"Language server(s) (only used if auto-creating project.yml). Inferred if not specified."* Passing it explicitly avoids the interactive branch. Alternatively, extend `_serena_project_skip_reason` (`cli/wrap.py:1666`) to skip the pre-index when `.serena/project.yml` is absent, and let Serena index lazily on demand — which the existing docstring already relies on as the fallback.

**3. Kill the process group, not just the direct child.** Spawning with `start_new_session=True` and killing the group on `TimeoutExpired` would stop the orphan leak. This repository has fixed this class of bug before, in #615 and #880.

## Environment

- headroom-ai 0.34.0, Python 3.14.6, macOS 15 (arm64), pipx install
- Serena resolved from `git+https://github.com/oraios/serena` as hardcoded at `cli/wrap.py:1705` and `mcp_registry/install.py:63`
- Reproduced with `wrap opencode`; the defect is in shared code (`_setup_serena_mcp`), so every wrap subcommand that registers Serena is affected

## Related

- #615 — process leak, orphaned proxy processes
- #880 — `headroom mcp serve` never exits, orphaned processes accumulate
- #2754 — feature request to use a pre-installed Serena, which would sidestep the `uvx` layer this bug depends on

Parideboy:
## Description

`headroom wrap <agent>` could sit silently for a full 300 seconds before the agent launched, and leaked one orphaned process every time it did.

`_setup_serena_mcp` runs `serena project index` synchronously on the launch path, with `capture_output=True`, an inherited stdin and `timeout=300`. When a project has no `.serena/project.yml`, Serena auto-creates one — and that auto-creation asks one `[y/N]` question per additionally-detected language server. Three things then combine:

1. stdin was inherited, so Serena believed it could prompt.
2. stdout was captured, so the question never reached the terminal.
3. the call was synchronous, so the agent waited out the entire timeout.

The user saw no prompt, no progress and no error — only a wrapper that appeared to hang. The pre-index could never succeed in that state, so the 300 seconds bought nothing.

On top of that, `subprocess.run` kills only its direct child on timeout. `uvx` is a launcher that execs the real `serena` executable as a grandchild, which was never signalled: it reparented to PID 1 and survived indefinitely. Same class of bug as #615 and #880.

Closes #2938

## Type of Change

## Changes Made

- `_serena_project_skip_reason` (`headroom/cli/wrap.py`) now returns a skip reason when `.serena/project.yml` is absent, so the pre-index does not run in the one state where it cannot succeed.
- `_index_serena_project` passes `stdin=subprocess.DEVNULL`, so a subprocess that decides to prompt gets EOF and exits in about a second instead of blocking behind a captured pipe. This is deliberately kept as a second line of defence even though the skip above already avoids the known prompt.
- `_index_serena_project` now spawns via `subprocess.Popen` in its own process group (`start_new_session=True` on POSIX, `CREATE_NEW_PROCESS_GROUP` on Windows) instead of `run(...)`, so the whole tree can be signalled.
- New `_kill_serena_index_tree` helper kills that tree on timeout — `killpg(..., SIGKILL)` on POSIX, `taskkill /F /T /PID` on Windows — then reaps the child and closes the capture pipes. Best-effort throughout; it never raises.
- Corrected two comments that asserted the opposite of the observed behaviour ("a failure or timeout here never blocks the wrap", "neither blocks the wrap"). Both were accurate about intent and wrong about effect.
- Added `_SERENA_INDEX_TIMEOUT` (still 300) and a line announcing the pre-index, so a legitimately long index no longer looks like a hang.
- Tests in `tests/test_cli/test_wrap_serena_boost.py` rewritten for the `Popen` path and extended to cover the DEVNULL stdin, the process-group flag, the timeout tree-kill, the new skip reason, and the `_setup_serena_mcp` wiring on both a fresh project and one that already has `project.yml`.

### Behaviour change worth a reviewer's attention

**On a project with no `.serena/project.yml`, the pre-index no longer runs at all.** That is the first wrap of any project, so this is the common case.

I went this way rather than fixing the prompt because there is no way to fix it from Headroom's side without re-introducing something the project deliberately removed. Serena's `project index` command has no non-interactive switch: `ProjectCommands._create_project` calls `ProjectConfig.autogenerate(..., interactive=True)` with `interactive` hardcoded. The only path that skips the prompt is passing `--ls/--language` explicitly, which means Headroom guessing the project's languages again — exactly the hand-maintained extension-to-language map that was removed in #2674, with a comment in this same function explaining why Serena should own that job.

The cost of skipping is small and self-correcting. Serena's MCP server (`serena start-mcp-server --project-from-cwd`) generates `project.yml` itself, non-interactively, on first start, and indexes lazily on demand — which is the fallback the existing docstring already relied on. So the first wrap now launches immediately with lazy indexing, and every wrap after that pre-indexes for real. Previously the first wrap cost 300 seconds *and* still produced no index, so nothing of value is lost.

Happy to switch to passing `--ls` instead if maintainers would rather keep the pre-index on the first wrap and accept a language map; the other two changes stand either way.

## Testing

### Test Output

```text
$ python -m pytest tests/test_cli/test_wrap_serena_boost.py tests/test_cli/test_serena_migrate.py -q
collected 29 items

tests\test_cli\test_wrap_serena_boost.py .............s...........       [ 86%]
tests\test_cli\test_serena_migrate.py ....                               [100%]

======================== 28 passed, 1 skipped in 0.54s ========================

$ python -m ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_serena_boost.py
All checks passed!

$ python -m ruff format --check headroom/cli/wrap.py tests/test_cli/test_wrap_serena_boost.py
2 files already formatted

$ python -m mypy headroom/cli/wrap.py --ignore-missing-imports --python-version 3.13 --follow-imports=silent
```

The single skip is `test_kill_tree_signals_the_group_on_posix`, which is platform-gated; the Windows counterpart ran. I develop on Windows, so the POSIX `killpg` branch is covered by unit test only — the end-to-end tree-kill proof below is the Windows `taskkill` branch.

## Real Behavior Proof

- Environment: Windows 11 Pro 26200, Python 3.13.11, headroom checkout at 941c25d3 plus this branch, uvx resolving `serena-agent` from PyPI, throwaway project with bash + TypeScript + Python sources and no `.serena/`
- Exact command / steps: spawned the real `uvx --from serena-agent serena project index` against that project twice — once with stdin readable and never closed (a parent-held pipe, the faithful stand-in for the idle terminal the old code inherited; `communicate()` cannot be used here because with `input=None` it closes the child's stdin immediately and hands it the EOF the real bug never delivers), once with `stdin=subprocess.DEVNULL`. Then spawned it twice more, stopping it with a plain `proc.kill()` (the old timeout behaviour) versus the new `_kill_serena_index_tree`, counting survivors with `Get-CimInstance Win32_Process` filtered on the command line.
- Observed result: stall reproduced and fixed — `A/open-stdin: STILL BLOCKED after 30.0s (timeout hit)` versus `B/DEVNULL: exit=1 elapsed=1.1s` with stderr tail `Project configuration auto-generation failed after 0.000 seconds / Error: EOF when reading a line`. Leak reproduced and fixed — after spawning a 5-process tree, `after proc.kill(): still alive: [34468, 35044, 35808]` versus `after _kill_serena_index_tree: still alive: []`. The 30-second bound in the first experiment stands in for the shipped 300; the point is that the child never returns on its own.
- Not tested: the POSIX `killpg`/`start_new_session` branch end-to-end (no Linux or macOS host available, so it is unit-tested only — and it is the branch the issue reporter observed failing); a full `headroom wrap opencode` launch end-to-end, since this machine has no working local proxy (`tests/test_cli/test_wrap_opencode.py` fails identically with and without this branch for that reason); and the second-wrap pre-index success path against a live Serena run, which is covered by unit test instead.

## Review Readiness

## Checklist

## Additional Notes

- The documentation checkbox is unchecked because no user-facing doc describes the pre-index; the behaviour is explained in source docstrings, which this PR rewrites.
- `_kill_serena_index_tree` is deliberately total: every step is wrapped so cleaning up an already-dead child cannot turn a timeout into a crash on the launch path. There is a test for that.
- #2754 (use a pre-installed Serena) would remove the `uvx` layer this leak depends on, but not the prompt itself — the stdin guard here is still needed after that lands.

### DIFF
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -1713,6 +1713,18 @@ def _serena_project_skip_reason(root: Path) -> str | None:
     Serena's own ``~/.serena`` config directory. A linked git worktree (its
     top-level ``.git`` is a file, not a directory) is an ephemeral checkout that
     would pay for its own index at a path that soon disappears.
+
+    A project with no ``.serena/project.yml`` is skipped because the pre-index
+    cannot succeed there (#2938). ``serena project index`` auto-creates the file
+    when it is missing, and that auto-creation calls
+    ``ProjectConfig.autogenerate(interactive=True)``, which asks one ``[y/N]``
+    question per additionally-detected language server. The CLI has no
+    non-interactive switch; the only way to reach the silent branch is to pass
+    ``--ls/--language`` explicitly, which means Headroom guessing the project's
+    languages again — exactly the hand-maintained map removed below. Serena's
+    MCP server generates that file itself (non-interactively) on first start and
+    indexes lazily on demand, so the pre-index simply resumes from the next
+    wrap onwards.
     """
     try:
         resolved = root.resolve()
@@ -1723,24 +1735,108 @@ def _serena_project_skip_reason(root: Path) -> str | None:
         return "$HOME is not a project"
     if (resolved / ".git").is_file():
         return "linked git worktree"
+    if not (resolved / ".serena" / "project.yml").is_file():
+        return "no .serena/project.yml yet — Serena will create it and index on demand"
     return None
+
+
+#: Upper bound on the synchronous pre-index. The agent does not launch until
+#: this call returns, so the number is a stall budget, not just a safety net.
+_SERENA_INDEX_TIMEOUT = 300
+
+
+def _kill_serena_index_tree(proc: subprocess.Popen) -> None:
+    """Kill *proc* and everything it spawned (best-effort, never raises).
+
+    ``uvx`` is a launcher: it resolves the environment and then runs the real
+    ``serena`` executable as a grandchild. Killing only the direct child leaves
+    that grandchild alive and reparented to PID 1, so every timed-out pre-index
+    leaked one process that never exits (#2938 — the same failure mode as #615
+    and #880). The child is started in its own process group precisely so the
+    whole tree can be signalled here.
+    """
+    if sys.platform == "win32":
+        # Windows has no process groups to signal for an already-wedged child;
+        # ``taskkill /T`` walks the tree by parent PID instead. ``/F`` because a
+        # process blocked in a read will not act on a graceful close request.
+        try:
+            subprocess.run(
+                ["taskkill", "/F", "/T", "/PID", str(proc.pid)],
+                capture_output=True,
+                timeout=10,
+                check=False,
+            )
+        except Exception:
+            pass
+    else:
+        try:
+            os.killpg(os.getpgid(proc.pid), signal.SIGKILL)
+        except Exception:
+            pass
+    # Backstop: if the tree kill above did not land, at least the direct child
+    # goes. Then reap so the parent does not leave a zombie behind, and close
+    # the capture pipes we opened so the wrap does not carry stray fds into the
+    # agent it is about to exec.
+    try:
+        proc.kill()
+    except Exception:
+        pass
+    try:
+        proc.wait(timeout=10)
+    except Exception:
+        pass
+    for stream in (proc.stdout, proc.stderr, proc.stdin):
+        try:
+            if stream is not None:
+                stream.close()
+        except Exception:
+            pass
 
 
 def _index_serena_project(*, verbose: bool = False) -> None:
     """Warm Serena's symbol cache for the current project (non-fatal).
 
-    Runs ``serena project index`` (the same ``uvx --from git+…`` launch used to
-    start the MCP server) in the project directory so the first symbol query is
-    not paying for a cold index. Timeout-guarded and best-effort: Serena also
-    indexes lazily on demand, so a failure or timeout here never blocks the
-    wrap.
+    Runs ``serena project index`` (the same ``uvx --from serena-agent`` launch
+    used to start the MCP server) in the project directory so the first symbol
+    query is not paying for a cold index. Serena also indexes lazily on demand,
+    so any failure here is survivable.
+
+    This runs on the launch path, synchronously: the agent starts only once it
+    returns, so the timeout below is time the user spends staring at nothing.
+    Two guards keep that bounded (#2938):
+
+    * ``stdin`` is ``DEVNULL``. Serena prompts when it has to auto-create
+      ``project.yml``, and because stdout is captured the question never
+      reaches the terminal — an inherited stdin turned that into a silent,
+      full-timeout hang. EOF makes it fail in about a second instead.
+      ``_serena_project_skip_reason`` already keeps us out of that state; this
+      is the belt-and-braces half, and it covers any future Serena prompt too.
+    * The child gets its own process group so ``_kill_serena_index_tree`` can
+      take out the ``uvx`` grandchild on timeout rather than orphaning it.
     """
     if shutil.which("uvx") is None:
         if verbose:
             click.echo("  Serena: uvx not found — skipping pre-index")
         return
+
+    popen_kwargs: dict[str, Any] = {
+        "stdout": subprocess.PIPE,
+        "stderr": subprocess.PIPE,
+        "stdin": subprocess.DEVNULL,
+        "text": True,
+        # ``subprocess.Popen`` directly, so the encoding defaults that
+        # ``headroom._subprocess.run`` applies have to be repeated here.
+        "encoding": "utf-8",
+        "errors": "replace",
+        "cwd": str(Path.cwd()),
+    }
+    if sys.platform == "win32":
+        popen_kwargs["creationflags"] = subprocess.CREATE_NEW_PROCESS_GROUP
+    else:
+        popen_kwargs["start_new_session"] = True
+
     try:
-        result = run(
+        proc = subprocess.Popen(
             [
                 "uvx",
                 # PyPI (prebuilt wheels), not the git source that fails to build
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the current head (`d8ca4b1`) adversarially across the fresh-project decision, process creation, timeout/error cleanup, and both platform branches. Skipping the interactive pre-index when `project.yml` is absent is consistent with Serena’s `--project-from-cwd` activation path, while DEVNULL remains an appropriate second guard. The isolated process group, whole-tree kill, direct-child backstop, reap, and pipe closure address the leak without making launch cleanup fatal. The focused tests cover the branch wiring and failure modes. I found no blocking issue. The two failing matrix shards are unrelated existing WebSocket timing and tokenizer-state failures.

PR #2953, opened 2026-08-12
fix(proxy/anthropic): don't buffer a CCR stream when passthrough discards the stream flip

### LINKED ISSUE #2952: [BUG] Signed thinking blocks discard the buffered-CCR stream:false flip, so Claude Code gets `API returned an empty or malformed response (HTTP 200)`
## Description

Since `dc163bcd` ("fix(proxy): preserve signed Anthropic thinking blocks on outbound re-serialize", #2254), any Claude Code session with extended thinking on fails from the turn where the first signed `thinking` block enters history:

```
API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request
```

`select_outbound_body` (`headroom/proxy/body_forwarding.py:110-111`) forwards the client's original bytes whenever the body carries a signed `thinking` / `redacted_thinking` block — that branch is evaluated **before** `body_mutated`, so every edit the handler made is discarded silently. The load-bearing casualty is the buffered-CCR path in `AnthropicHandler`, which sets `body["stream"] = False` (`headroom/proxy/handlers/anthropic.py:3073-3078`) precisely so it can read one JSON document and resolve `headroom_retrieve` server-side. That flip never reaches the wire, upstream streams, and the handler then tries to parse SSE as JSON.

This is not the same failure as #2251 — that one was a 400 from Anthropic. This is a 200 with a body the client cannot read.

## To Reproduce

1. `headroom proxy --port 8787 --memory --code-aware --mode token` (CCR on, i.e. defaults)
2. `ANTHROPIC_BASE_URL=http://127.0.0.1:8787 claude` with extended thinking on (Claude Code default)
3. Drive the session past the first assistant turn that emits a `thinking` block, with `headroom_retrieve` injected (any turn where Headroom compressed a tool result)
4. The next streaming turn returns 200 with no usable body, and the retry after it is answered from the semantic cache with the same broken bytes

## Expected Behavior

The turn either streams normally or fails with an actionable error. A 200 whose body the client cannot parse is the worst of both.

## Actual Behavior

From `~/.headroom/logs/proxy.log`, the turn that breaks:

```
CCR: stream:true request has headroom_retrieve available; using buffered stream:false upstream request
event=outbound_request forwarder=anthropic_messages body_bytes=132017 body_mutated=true
  mutation_reasons=structural_diff_vs_original,ccr_streaming_retrieve_buffered_non_stream source=passthrough
event=proxy_inbound_response status=200 duration_ms=1175.40
PERF ... msgs=6 tok_saved=3575 tool_saved=16064 tok_out=0 total_ms=9623
```

`body_bytes` (132017) is byte-identical to the inbound `content_length` (132017) and the source is `passthrough` despite two recorded mutations. The turn immediately before it (4 messages, no thinking block yet) was `source=canonical` with `tok_out=341`.

And the next request, which is what the user actually sees the error on:

```
event=proxy_inbound_request content_length=132003 accept: application/json
PERF ... msgs=6 tok_before=0 tok_after=0 opt_ms=1 total_ms=1 transforms=none
event=proxy_inbound_response status=200 duration_ms=2.11
```

No upstream call: a `SemanticCache` hit serving the entry the broken turn had just written.

### Chain

1. Passthrough discards `body["stream"] = False`; upstream gets `"stream": true` and returns a 200 SSE stream.
2. `response.json()` fails, `resp_json = None`, logged at **DEBUG only** (`anthropic.py:3321-3327`) — invisible at the default log level, which is why nothing in the log looks wrong.
3. `if buffered_stream_ccr and response.status_code == 200 and resp_json:` (`anthropic.py:3818`) is therefore False, SSE resynthesis is skipped, and the code falls through to the plain `return Response(...)` at `anthropic.py:3904`.
4. `_BufferedCCRResponse` (`anthropic.py:3914-4033`) has already committed `content-type: text/event-stream` after its 1 s keepalive. A plain `Response` exposes no `body_iterator`, so it emits the hardcoded bare `event: error` at `anthropic.py:3994-4000`. `tok_out=0` follows because usage is only read from `resp_json`.
5. **Cache poisoning.** `anthropic.py:3655` caches any 200 body — here the raw SSE bytes plus the upstream `content-type: text/event-stream`. The cache key (`semantic_cache.py:38-53`) has no `stream` component, and the hit path (`anthropic.py:1020-1031`) pops only `content-encoding` / `content-length`, so Starlette keeps the stored content-type and ignores `media_type="application/json"`. A later non-streaming request with the same messages is answered with SSE bytes labelled SSE. TTL is 3600 s, so it keeps answering.

Secondary, same root cause: on every thinking-bearing turn the passthrough also discards compression, memory-tool injection and tool-schema compaction, while `PERF` still reports the savings (`tok_saved=3575 tool_saved=16064` above) as if they had shipped.

`/v1/responses` has the same `stream`-flip pattern (`openai.py:5479-5487`), but its bodies carry `input` rather than `messages`, so `has_signed_thinking_blocks` never fires there.

## Environment

- **Headroom version**: 0.35.0-dev (source checkout at `dc163bcd`+)
- **Python version**: Python 3.13.11
- **OS**: Windows 11 Pro 26200
- **LLM Provider**: Anthropic (`api.anthropic.com`), `claude-opus-5`, Claude Code CLI 2.1.228

## Workaround

Restart the proxy with `--no-ccr` so `headroom_retrieve` is never injected and the buffered path cannot engage (the restart also clears the poisoned cache entry). `HEADROOM_PROXY_PYTHON_FORWARDER_MODE=legacy_json_kwarg` is **not** an escape hatch — the signed-thinking branch overrides the legacy encoder too.

## Suggested Fix

1. Do not choose the buffered-CCR path when the outbound body is locked to the client's bytes; take the plain streaming path instead. The injected retrieve tool is itself a discarded mutation on those turns, so that is the coherent behaviour.
2. On the buffered path, relay a 200 SSE reply to the client verbatim instead of degrading to a bare error event, and log the format mismatch at WARNING rather than DEBUG.
3. Never cache a body that did not parse as JSON, and drop the stored `content-type` on the cache-hit path.
4. Report discarded mutations: have `select_outbound_body` surface what passthrough threw away and log it, so an operator does not read savings that never shipped.

Not addressed by the above, worth a separate issue: option (b) of #2251 — a canonical re-serialization that preserves signed blocks byte-for-byte so compression and tool injection survive a thinking-bearing history. #2251 reports a 400 even on a no-op re-encode whose only transform was `tool_search_deferral`, which hints the signature covers `tools` too, so that needs validation against the live API.

Parideboy:
## Description

Fixes #2952. Since `dc163bcd` (#2254), a Claude Code session with extended thinking on dies from the turn where the first signed `thinking` block enters history:

```
API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request
```

`select_outbound_body` forwards the client's original bytes whenever the body carries a signed `thinking` / `redacted_thinking` block, and it decides that **before** it looks at `body_mutated` — so every edit the handler made is discarded. The buffered-CCR path depends on exactly such an edit: it sets `body["stream"] = False` (`anthropic.py:3073-3078`) so the reply arrives as one JSON document it can scan for `headroom_retrieve` calls. With the flip discarded, upstream streams, `response.json()` fails, SSE resynthesis is skipped, and the client is handed a 200 it cannot read.

From the reporter's `proxy.log`, the turn that breaks — note `body_bytes` equals the inbound `content_length` byte for byte, and `source=passthrough` despite two recorded mutations:

```
CCR: stream:true request has headroom_retrieve available; using buffered stream:false upstream request
event=outbound_request forwarder=anthropic_messages body_bytes=132017 body_mutated=true
  mutation_reasons=structural_diff_vs_original,ccr_streaming_retrieve_buffered_non_stream source=passthrough
PERF ... msgs=6 tok_saved=3575 tool_saved=16064 tok_out=0 total_ms=9623
```

This is a different failure from #2251 (a 400 from Anthropic). Signed thinking blocks still leave as original bytes here, so that fix is untouched.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

Primary fix, `headroom/proxy/handlers/anthropic.py`:

- Gate `buffered_stream_ccr` on the outbound body actually being ours to change, via a new `outbound_body_is_client_bytes()` predicate that mirrors the passthrough branch. Those turns take the plain streaming path instead, which is the coherent outcome: the injected retrieve tool is itself a discarded mutation there, so the model was never going to see it. One INFO line records the choice.

Three follow-on defenses, each of which independently kept the failure alive or invisible:

- A non-JSON 200 on the buffered path now logs at WARNING (it was DEBUG, which is why nothing in `proxy.log` looked wrong) and the upstream SSE is relayed to the client verbatim, instead of falling through to a plain `Response` that `_BufferedCCRResponse` can only turn into a bare `event: error` once its 1 s keepalive has committed headers.
- The semantic cache no longer stores a body that did not parse as JSON, and drops the stored `content-type` on the hit path. The cache key has no `stream` component, so a cached SSE body was replayed to buffered callers for the full 3600 s TTL — that replay is the request the reporter actually saw the error on (a 2 ms `PERF ... transforms=none` cache hit).
- `select_outbound_body` now reports the mutations passthrough discarded (`dropped_mutations` / `dropped_mutation_reasons`), and `log_outbound_request` logs them as `event=outbound_body_mutations_dropped` at WARNING. Without it, `PERF` reports savings and tool injections that never reached the wire and nothing contradicts it.

`prepare_outbound_body_bytes` keeps its two-value shape, so existing callers are unchanged.

## Testing

New `tests/test_ccr_buffered_stream_signed_thinking.py` covers the gate (signed-thinking history takes the streaming path and still leaves as `stream: true`; the same request without thinking blocks still takes the buffered path with `stream: false`), the SSE relay, and the cache guards. The relay case is parametrized on upstream latency because the failure only reaches its worst form past the 1 s keepalive, where a plain `Response` has no `body_iterator` left to forward.

Verified red before green — with the source changes stashed and the tests in place:

```text
FAILED tests/test_ccr_buffered_stream_signed_thinking.py::test_signed_thinking_history_skips_the_buffered_ccr_path[True-True]
FAILED tests/test_ccr_buffered_stream_signed_thinking.py::test_buffered_ccr_relays_an_unexpected_sse_reply_and_does_not_cache_it[prompt]
FAILED tests/test_ccr_buffered_stream_signed_thinking.py::test_buffered_ccr_relays_an_unexpected_sse_reply_and_does_not_cache_it[past-keepalive]
FAILED tests/test_ccr_buffered_stream_signed_thinking.py::test_cache_hit_never_replays_a_foreign_content_type
========================= 4 failed, 1 passed in 8.88s =========================
```

With the fix applied:

```text
$ python -m pytest tests/test_proxy_byte_faithful_forwarding.py tests/test_ccr_buffered_stream_signed_thinking.py -q
55 passed in 11.40s

$ python -m pytest tests/test_compression_cache.py tests/test_ccr_inline_resolve_handlers.py \
    tests/test_ccr_sqlite_backend.py tests/test_anthropic_stage_timings.py \
    tests/test_backend_nonstreaming_cache_metrics.py tests/test_cache_mode_cold_start.py \
    tests/test_cache_breakpoint_diagnostics.py -q
87 passed, 1 skipped in 17.57s

$ python -m ruff check .
All checks passed!

Found 12 errors in 3 files (checked 517 source files)
# all 12 pre-existing in headroom/ccr/mcp_server.py and headroom/memory/mcp_server.py
# (local mcp package version); none in the four files this PR touches
```

## Real Behavior Proof

- Environment: headroom 0.35.0-dev source checkout, Python 3.13.11, Windows 11, Claude Code CLI 2.1.228 against api.anthropic.com (claude-opus-5), proxy run as `headroom proxy --port 8787 --memory --code-aware --mode token`
- Exact command / steps: reproduced from the reporter's `~/.headroom/logs/proxy.log` — request `hr_1786551079_000005` shows `source=passthrough` with `body_bytes` identical to the inbound `content_length` while `mutation_reasons` contains `ccr_streaming_retrieve_buffered_non_stream`, then `tok_out=0`; request `hr_1786551089_000006` is the 2 ms `transforms=none` semantic-cache hit that returned the poisoned SSE body to a caller asking for JSON. The preceding turn (`...000004`, no thinking block yet) was `source=canonical` with `tok_out=341`. Then: pytest suites above, with the red/green stash comparison
- Observed result: with the gate in place the thinking-bearing turn goes down `_stream_response` and the forwarded body still says `"stream": true`, matching the bytes passthrough will send; a buffered turn whose upstream answers with SSE reaches the client as a stream and writes nothing to the semantic cache; a cache entry can no longer hand a caller a content-type from a differently-shaped request
- Not tested: a live end-to-end Claude Code session against Anthropic with the patched proxy (the reproduction here is the reporter's proxy log plus handler-level tests); `/v1/responses`, which has the same `stream`-flip pattern (`openai.py:5479-5487`) but carries `input` rather than `messages`, so `has_signed_thinking_blocks` never fires there and I left it alone

## Review Readiness

## Checklist

## Additional Notes

Deliberately out of scope, worth a separate issue: option (b) of #2251 — a canonical re-serialization that preserves signed blocks byte-for-byte so compression and tool injection survive a thinking-bearing history rather than being silently dropped. #2251 reports a 400 even on a no-op re-encode whose only transform was `tool_search_deferral`, which hints the signature covers `tools` too, so getting it wrong would re-break every multi-turn thinking session. That needs validating against the live API, not guessing. Until then the new WARNING at least makes the dropped work visible.

Also unfixed by design: the savings accounting itself. A passthrough turn still books `tok_saved` / `tool_saved` in `PERF` for bytes that never shipped; correcting the numbers is a wider change than this bug needs.

### DIFF
--- a/headroom/proxy/body_forwarding.py
+++ b/headroom/proxy/body_forwarding.py
@@ -32,6 +32,15 @@ class OutboundBody:
 
     content: bytes
     source: OutboundBodySource
+    #: True when byte-faithful passthrough won over a mutated body, so every
+    #: edit the handler made to ``body`` was discarded before the wire. Callers
+    #: that gated a downstream decision on their own mutation (for example
+    #: flipping ``stream`` to False to buffer a reply) MUST consult this — the
+    #: request upstream actually sees is the client's original one.
+    dropped_mutations: bool = False
+    #: The ``BodyMutationTracker`` reasons discarded alongside it, when the
+    #: caller supplied them. Empty when the caller passed no reason list.
+    dropped_mutation_reasons: tuple[str, ...] = ()
 
 
 def get_python_forwarder_mode() -> PythonForwarderMode:
@@ -104,11 +113,23 @@ def select_outbound_body(
     original_body_bytes: bytes | None,
     body_mutated: bool,
     forwarder_mode: PythonForwarderMode | None = None,
+    mutation_reasons: list[str] | None = None,
 ) -> OutboundBody:
-    """Select the exact bytes to forward upstream."""
+    """Select the exact bytes to forward upstream.
+
+    ``mutation_reasons`` is optional and only used for reporting: when the
+    signed-thinking passthrough overrides a mutated body, the discarded reasons
+    are echoed back on the result so the call site can log what never made it
+    upstream instead of silently claiming the edit landed.
+    """
     mode = forwarder_mode if forwarder_mode is not None else get_python_forwarder_mode()
     if original_body_bytes is not None and has_signed_thinking_blocks(body):
-        return OutboundBody(content=original_body_bytes, source="passthrough")
+        return OutboundBody(
+            content=original_body_bytes,
+            source="passthrough",
+            dropped_mutations=body_mutated,
+            dropped_mutation_reasons=tuple(mutation_reasons or ()) if body_mutated else (),
+        )
 
     if mode == "legacy_json_kwarg":
         content = json.dumps(body, separators=(", ", ": "), ensure_ascii=True).encode("utf-8")
@@ -125,12 +146,38 @@ def prepare_outbound_body_bytes(
     original_body_bytes: bytes | None,
     body_mutated: bool,
     forwarder_mode: PythonForwarderMode | None = None,
+    mutation_reasons: list[str] | None = None,
 ) -> tuple[bytes, OutboundBodySource]:
-    """Compatibility tuple wrapper around :func:`select_outbound_body`."""
+    """Compatibility tuple wrapper around :func:`select_outbound_body`.
+
+    Keeps the two-value shape its existing callers unpack. Call
+    :func:`select_outbound_body` directly when you need the dropped-mutation
+    reporting.
+    """
     outbound = select_outbound_body(
         body=body,
         original_body_bytes=original_body_bytes,
         body_mutated=body_mutated,
         forwarder_mode=forwarder_mode,
+        mutation_reasons=mutation_reasons,
     )
     return outbound.content, outbound.source
+
+
+def outbound_body_is_client_bytes(
+    *,
+    body: dict[str, Any],
+    original_body_bytes: bytes | None,
+) -> bool:
+    """Return whether the wire body will be the client's original bytes.
+
+    A handler that changes ``body`` to steer its own upstream call — the
+    ``stream`` flip that buys a buffered reply is the load-bearing case — has to
+    know that the signed-thinking passthrough will throw that change away and
+    send the client's bytes verbatim. Asking before acting is cheaper than
+    discovering it from a reply in the wrong wire format.
+
+    Mirrors the first branch of :func:`select_outbound_body`; the forwarder mode
+    is deliberately not consulted because that branch overrides it too.
+    """
+    return original_body_bytes is not None and has_signed_thinking_blocks(body)
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -73,6 +73,20 @@ def _strip_index_from_content_blocks(content: Any) -> None:
             _strip_index_from_content_blocks(block.get("content"))
 
 
+def _looks_like_sse_response(response: httpx.Response) -> bool:
+    """Return whether an upstream reply is a Server-Sent Events stream.
+
+    Trusts the declared content-type first and falls back to sniffing the
+    leading bytes for an SSE field, because a gateway in front of Anthropic may
+    relay the stream under a vaguer type.
+    """
+    content_type = (response.headers.get("content-type") or "").lower()
+    if "text/event-stream" in content_type:
+        return True
+    head = response.content[:64].lstrip()
+    return head.startswith(b"event:") or head.startswith(b"data:")
+
+
 class AnthropicHandlerMixin:
     """Mixin providing Anthropic API handler methods for HeadroomProxy."""
 
@@ -1020,6 +1034,11 @@ class AnthropicHandlerMixin:
                     response_headers = dict(cached.response_headers)
                     response_headers.pop("content-encoding", None)
                     response_headers.pop("content-length", None)
+                    # Drop the stored content-type too. Starlette lets an
+                    # explicit header win over ``media_type``, so keeping the
+                    # producing request's type would let a cache entry hand this
+                    # caller a wire format it never asked for (#2952).
+                    response_headers.pop("content-type", None)
 
                     # Unit 4: release the pre-upstream semaphore on cache
                     # hit — no upstream call will happen.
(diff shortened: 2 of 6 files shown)


JerrettDavis (approved):
Reviewed the current head (`5fc6f57`) adversarially. The buffered-CCR gate now follows the same signed-thinking predicate as outbound byte selection, the unexpected-reply fallback is constrained to positively identified SSE, and the cache changes prevent a differently shaped response from poisoning later callers. The new tests cover the decision point, delayed relay behavior, cache write guard, and cache-hit content type. I found no blocking correctness issue in this change. The failing matrix shards are outside this diff (external binary download failures, the existing WebSocket timing assertion, and tokenizer state leakage); the targeted coverage and the remaining build, lint, and security checks are green.

gglucass:
Hit this in production too, and our proxy logs line up exactly with the diagnosis in #2952 — strongly encourage merging this.

@chopratejas @JerrettDavis @DevanshiVyas

We chased `API returned an empty or malformed response (HTTP 200)` on Claude Code 2.1.228 / claude-opus-5 against headroom-ai 0.35.0, and the `event=outbound_request` lines partition the buffered-CCR traffic cleanly by how the body was forwarded:

| buffered CCR requests (one machine, ~19h) | count | outcome |
|---|---|---|
| `source=canonical` | 285 | all fine |
| `source=passthrough` | **6** | **all failed** |

Six for six. The `source=passthrough` request IDs are exactly the turns that logged `tok_out=0 ttfb_ms=0` and preceded each client-visible failure. That is this bug: when byte-faithful passthrough forwards the original bytes, the `body["stream"] = False` flip is discarded, upstream returns an SSE stream, `response.json()` fails, `resp_json` is `None`, and the non-SSE fallback returns a 200 the client cannot parse or retry.

Two details that may help others land here:

- The sessions were all running signed thinking (`interleaved-thinking-2025-05-14`, `thinking-token-count-2026-05-13`), consistent with #2952's repro.
- It is invisible in the proxy log today. The `response.json()` failure is `logger.debug`, the fallback logs nothing, and the request still records `status=200` — so the only trace is `tok_out=0 ttfb_ms=0` next to a `source=passthrough` outbound line. (#2968 promotes that one log line to `warning`, which is orthogonal to this PR.)

Worth noting the blast radius is wider than the malformed response: because the status is 200, client-side retry and backoff never fire, so a recoverable turn becomes a hard session failure.

Issue #2827, opened 2026-08-06
[BUG] VS Code Copilot Chat 0.60.0 only proxies /models through Headroom; chat responses bypass the proxy


SpydrCS:
## Description

When running `headroom wrap vscode`, Headroom successfully receives GitHub Copilot Chat's `GET /models` requests, but no subsequent `/responses` or `/chat/completions` request is sent through the proxy when submitting a prompt in VS Code.

After the proxied `/models` response, the VS Code Copilot extension logs model-catalog errors. Copilot Chat can still return a response, but Headroom records no model-generation traffic and no token savings.

The same GitHub Copilot account and model work successfully through the standalone Copilot CLI using Headroom's subscription wrapper:

```powershell
headroom wrap copilot --subscription --port 8788 -- --model gpt-5.6-terra
```

## To Reproduce

1. Install Headroom 0.34.0 and authenticate it for GitHub Copilot.
2. Close all VS Code windows.
3. Run:

   ```powershell
   headroom wrap vscode
   ```

4. Open VS Code normally.
5. Select `GPT-5.6 Terra` in Copilot Chat.
6. Send a short prompt.
7. Open the Headroom dashboard at `http://127.0.0.1:8787/dashboard`.
8. Observe that the proxy logs `GET /models` requests from `GitHubCopilotChat/0.60.0`, but no `/responses` or `/chat/completions` request for the submitted prompt.

## Expected Behavior

VS Code Copilot Chat should send the selected model's generation request through Headroom after model discovery.

The dashboard should show a successful proxied `/responses` request and record the corresponding request/token metrics.

## Actual Behavior

Headroom receives `GET /models` requests and returns HTTP 200, but no model-generation request reaches the proxy.

After the model metadata requests, VS Code logs errors from the built-in Copilot extension. Copilot Chat still produces a response, but it appears to use a path that bypasses Headroom.

## Code Sample

```python
# No Python code is required.
# Reproduction uses the VS Code Copilot Chat extension:

# PowerShell:
# headroom wrap vscode

# Then open VS Code, select GPT-5.6 Terra, and submit a short chat prompt.
```

## Error Output

```text
2026-08-06 10:52:53,397 - headroom.proxy - INFO - event=proxy_inbound_request ... method=GET path=/models ...
user-agent: GitHubCopilotChat/0.60.0
editor-plugin-version: copilot-chat/0.60.0
editor-version: vscode/1.132.0

[...]

PR #2954, opened 2026-08-12
fix(copilot): restore VS Code 0.60 proxy settings

chopratejas:
## Description
Restore transparent Copilot routing for VS Code Copilot extension 0.60.0, which no longer reliably recognizes the flattened dotted override keys emitted by Headroom.

## Type of Change
- [x] Bug fix

## Changes Made
- Emit the canonical nested `github.copilot.advanced.debug` settings object.
- Preserve the proxy URL and token auth override.
- Keep marker-based safe removal and refuse to overwrite unmanaged advanced settings.

## Testing

- [x] Focused unit tests pass
- [x] Ruff, formatting, and diff checks pass
- [ ] Docker wrap E2E passes

### Test Output

```text
64 focused Copilot VS Code/wrap tests passed
Ruff, format, and diff checks passed
Docker wrap E2E: fails at the stale flattened-key assertion in verify_vscode_wrap
```
- 64 focused Copilot VS Code/wrap tests passed.
- Ruff, format, and diff checks passed.

## Real Behavior Proof

- Environment: current PR head, VS Code Copilot 0.60 settings writer and focused wrapper/config tests.
- Exact command / steps: generated a managed VS Code settings block, updated it, removed it, and ran 64 focused Copilot VS Code/wrap tests plus Ruff and formatting checks.
- Observed result: the unit-level writer emits the nested github.copilot.advanced.debug object and unwrap restores the original JSONC; Docker E2E still expects the former flattened key and currently fails.
- Not tested: a live VS Code Copilot 0.60 session successfully routing through the generated configuration.
The generated settings now match the object-valued configuration shape consumed by current Copilot Chat configuration code, while unwrap restores the original JSONC content.

## Review Readiness

### DIFF
--- a/e2e/wrap/run.py
+++ b/e2e/wrap/run.py
@@ -786,12 +786,12 @@ def verify_vscode_wrap(base_env: dict[str, str], project_dir: Path) -> None:
         project_prefix = f"/p/{quote(project_dir.name, safe='')}"
         configured = settings_path.read_text(encoding="utf-8")
         assert_true(
-            f'"github.copilot.advanced.debug.overrideProxyUrl": '
-            f'"http://127.0.0.1:{port}{project_prefix}"' in configured,
+            f'"github.copilot.advanced": {{' in configured
+            and f'"overrideProxyUrl": "http://127.0.0.1:{port}{project_prefix}"' in configured,
             "VS Code wrap should configure the project-scoped proxy URL",
         )
         assert_true(
-            '"github.copilot.advanced.debug.overrideAuthType": "token"' in configured,
+            '"overrideAuthType": "token"' in configured,
             "VS Code wrap should configure token auth",
         )
         assert_true(
--- a/e2e/wrap/run.py
+++ b/e2e/wrap/run.py
@@ -786,7 +786,7 @@ def verify_vscode_wrap(base_env: dict[str, str], project_dir: Path) -> None:
         project_prefix = f"/p/{quote(project_dir.name, safe='')}"
         configured = settings_path.read_text(encoding="utf-8")
         assert_true(
-            f'"github.copilot.advanced": {{' in configured
+            '"github.copilot.advanced": {' in configured
             and f'"overrideProxyUrl": "http://127.0.0.1:{port}{project_prefix}"' in configured,
             "VS Code wrap should configure the project-scoped proxy URL",
         )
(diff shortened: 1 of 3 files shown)


JerrettDavis (changes requested):
The source/unit change is internally consistent, but both required Docker wrap E2E jobs fail because e2e/wrap/run.py::verify_vscode_wrap still asserts the old flattened github.copilot.advanced.debug.overrideProxyUrl key. Please update the E2E verifier to parse/assert github.copilot.advanced.debug.overrideProxyUrl from the new nested object (and preserve the token auth assertion), then rerun the failed jobs. Also add a real VS Code/Copilot 0.60 behavior proof if available: the current evidence proves serialization and reversible removal, but not that the extension consumes this shape. Maintainer edits are disabled on this fork, so I cannot push the small E2E correction directly.

chopratejas:
Addressed the review blocker in cfff1a2d: updated e2e/wrap/run.py::verify_vscode_wrap to parse/assert the nested github.copilot.advanced.debug proxy URL and token auth fields. The focused 64 tests remain green; Docker E2E and full CI are rerunning now.

JerrettDavis (approved):

chopratejas:
Closing this — the premise turned out not to hold, and the real cause is fixed in #3077.

**The settings were never the problem.** This PR rewrites the flattened dotted keys into a nested `github.copilot.advanced` object on the theory that Copilot 0.60.0 no longer recognises the flattened form. Two pieces of evidence contradict that:

- #2827 was filed against 0.60.0 with only `overrideProxyUrl` set, and Headroom **received** that extension's `GET /models`. The flattened key was honoured.
- The reporter who originally raised the 0.60.0 configuration concern followed up after reading the logs: *"I could actually see that the proxy is picked up by copilot."* The overrides are being applied.

**What was actually broken** is where Headroom forwarded the traffic once it arrived. Inline completions land on `/v1/engines/<engine>/completions`, which matches no route, so the catch-all resolved an upstream from the auth headers alone — never looking at the path — and fell through to the OpenAI default:

```
https://api.openai.com/v1/engines/gpt-41-copilot/completions
```

OpenAI removed the Engines API years ago, so that could never have worked; on networks that allow GitHub Copilot but not OpenAI it fails as a blocked domain. #3077 routes that path to Copilot (using the completions host GitHub advertises in `endpoints.proxy`, with a `GITHUB_COPILOT_PROXY_URL` override) and stops `/v1` being stripped off a path Copilot built itself.

**Merging this as-is would carry risk.** The branch predates #2986, so its `_managed_block` has no `_CAPI_KEY` — merging would drop `overrideCapiUrl` and regress chat routing, which currently works. It would also change a settings shape that the evidence above says is working today.

If the nested object shape is worth adopting on its own merits, that is a separate change on top of current `main`, with a live 0.60.0 session confirming both forms behave as expected. Reopening is fine if someone reproduces the flattened keys being ignored on a current build.

PR #2959, opened 2026-08-13
fix(proxy): stop committing HTTP 200 before the buffered CCR result is known

nangsontay:
## Description

A `stream:true` request carrying `headroom_retrieve` is rewritten to a buffered `stream:false` upstream call so CCR can resolve retrievals server-side. `_BufferedCCRResponse` then committed `http.response.start` with `200 text/event-stream` on a 1s keepalive timer — before the outcome was known — and began emitting `event: ping`.

When the buffered operation later resolved to anything that could not become SSE (non-200 upstream, or a body that did not parse to `resp_json`), the fall-through returned a plain `Response`. With the status already committed there was no `body_iterator` to pump, so the client received pings plus a generic error event: an HTTP 200 whose body carries no `message_start`. Claude Code surfaces that as `API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request`. The real upstream status was discarded with it, so client-side 429/5xx backoff never fired.

The fix sends nothing until the operation resolves, then delegates to the real result so its status and headers survive. A crash still returns the sanitized 502. The same defect and fix apply to the OpenAI Responses twin.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/anthropic.py`: `_BufferedCCRResponse.__call__` awaits the buffered CCR operation and delegates to the resulting response instead of committing `200 text/event-stream` on a 1s keepalive timer and hand-pumping `body_iterator`.
- `headroom/proxy/handlers/openai.py`: same change to the `/v1/responses` twin.
- Removed the keepalive ping preamble in both. Clients budget minutes per turn (Claude Code sends `x-stainless-timeout: 600`), so nothing is gained by committing a status early.
- Tests: renamed the two `*_emits_keepalive_before_delayed_upstream` cases to `*_withholds_output_until_*_resolves` and asserted no ASGI message is sent for 1.1s (past the deleted deadline); the two `*_late_failure_emits_sanitized_error_event` cases now assert a real 502 rather than a downgraded 200.
- Tests: added `test_buffered_ccr_preserves_late_failure_status_and_headers` — a 429 landing after the old deadline must reach the client as a 429 with `retry-after` and the upstream message.
- Tests: the two success-path fakes' `receive()` now delivers the body once and then blocks, matching a real ASGI server. Delegating to `StreamingResponse` runs Starlette's `listen_for_disconnect`, and a `receive()` that returns instantly on every call spins that loop without yielding.

## Testing

### Test Output

```text
$ .venv/bin/pytest tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py \
      tests/test_proxy/test_openai_responses_ccr.py -q -p no:randomly
======================== 18 passed, 1 warning in 9.04s =========================

$ ruff check headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py \
      tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py \
      tests/test_proxy/test_openai_responses_ccr.py
All checks passed!
```

## Real Behavior Proof

- Environment: macOS 25.4.0 (arm64), Python 3.13.14, Headroom proxy on localhost:8787, Claude Code 2.1.228 wrapping through it against api.anthropic.com
- Exact command / steps: reproduced from a live session, not synthetically — `~/.headroom/logs/proxy.log` request `hr_1786547104_000092` logged `CCR: stream:true request has headroom_retrieve available; using buffered stream:false upstream request`, then `outbound_request ... mutation_reasons=structural_diff_vs_original,ccr_streaming_retrieve_buffered_non_stream`, then `proxy_inbound_response status=200 duration_ms=1259.84` (the +1.0s keepalive commit), then `PERF ... tok_out=0 cache_read=0 cache_write=0 total_ms=4809`; the matching Claude Code transcript entry at `2026-08-12T15:05:09.550Z` carries no `requestId`
- Observed result: before the fix the client saw `API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request` and the turn was lost; after the fix `test_buffered_ccr_preserves_late_failure_status_and_headers` shows a late 429 arriving as a real 429 with `retry-after: 7` and the upstream message in the body, and the full buffered-CCR suite is green (18 passed)
- Not tested: no end-to-end re-run against the live Anthropic API after the fix — the regression is covered by the new unit test rather than by reproducing the upstream failure again; `mypy headroom` was not run; the OpenAI Responses path is covered by its unit tests only, not by a live Codex session

## Runtime Rollout Safety

- Rollout-managed feature(s): None. The buffered CCR response path is not in the rollout registry — its only entries are `tool_result_interceptors`, `proxy_output_shaper`, and `read_maturation`. This path is gated by `CCRConfig` / `ccr_handle_responses`, not by a channel.
- Minimum rollout channel: stable — the fix ships unconditionally on the default channel, exactly like the code it replaces.
- Stable/default behavior changed: Yes. On `stable`, a buffered CCR turn no longer emits `event: ping` keepalives and no longer commits `200 text/event-stream` before the upstream result exists; a failed turn now reaches the client with its real status and headers instead of a 200.
- Kill switch / disable path: `--no-ccr` (settings `no_ccr: true`) sets `ccr_inject_tool=False`, so no `headroom_retrieve` tool is injected, `_has_headroom_retrieve_tool()` is false, and the buffered path is never entered — streaming falls back to plain passthrough. There is no separate flag for the response-commit behavior itself.
- Qualification impact: None. No rollout policy, registry entry, `registry_digest`, or `qualification_eligible` field is touched; `headroom/rollout.py`, `tests/test_rollout.py`, and `tests/fixtures/rollout_policy_vectors.json` are unchanged by this PR.
- Rollback path: revert the single commit on this branch. No migration, no persisted state, no config change — reverting restores the keepalive preamble as-is.

## Review Readiness

## Checklist

N/A — proxy behavior change, no UI surface.

## Additional Notes

Tradeoff worth a maintainer's eye: the keepalive pings were deliberate and tested, and removing them means a buffered CCR turn now sends zero bytes until the upstream resolves. That is fine for Claude Code and Codex, which allow minutes per turn, but an intermediary with a short idle-read timeout (nginx `proxy_read_timeout`, Cloudflare) would now cut long CCR turns that previously stayed alive on pings. The alternative — keep the pings and instead forward the upstream error body into an SSE `error` event — fixes the malformed-200 symptom but permanently loses the status code, so backoff still cannot work. I chose correctness of the status over idle-timeout tolerance; happy to switch if you would rather keep the pings.

Documentation was left unchanged: the keepalive was an implementation detail with no docs surface I could find.

No linked issue — found while debugging a live wrapped session, filed straight as a PR.


gglucass:
Hit this in production too — the early `200 text/event-stream` commit turned a recoverable upstream failure into an unretryable one for us, so I'd encourage merging this direction.

@chopratejas @JerrettDavis @DevanshiVyas

One caveat worth resolving before it lands, because I think the justification in the comment is load-bearing and not quite right:

> Clients budget minutes for a turn (Claude Code sends `x-stainless-timeout: 600`), so waiting is free.

`x-stainless-timeout` is the **total request** timeout, not the **stream idle** timeout — they're separate watchdogs. #2465 was specifically `Stream idle timeout - no chunks received`, and #2479 added this heartbeat to fix it. Removing the preamble entirely looks like it reintroduces #2465 for exactly the slow upstreams that issue reported (Foundry, long Opus turns).

Two data points from our logs, headroom-ai 0.35.0, Claude Code 2.1.228:

- Buffered CCR calls routinely ran 15-25s wall clock (`total_ms=19269`, `22899`, `15531`). They survived only because pings were flowing.
- That client sends `x-stainless-timeout: 300`, not 600, so the total-timeout headroom is smaller than assumed — and the idle timeout is the binding constraint anyway.

A middle option that keeps both properties: leave the heartbeat in place but make the pre-commit grace configurable instead of hard-coded at `1.0`, defaulting it to a few seconds. Anything that resolves inside the window — a fast 4xx, and importantly a 429/529 that only resolves after `_retry_request` has honored `Retry-After` and retried — passes through with its real status, while genuinely long turns still get a first byte before any client idle watchdog fires. Happy to send that as a follow-up if useful; I have it working locally but did not open a competing PR since this one covers both handlers and got here first.

JerrettDavis:
Thank you for isolating this failure mode and for the late-429 regression coverage. The core fix from this branch has since landed on `main` via #2997 (`fix(proxy): complete stateless Responses and buffered CCR lifecycle`), together with the surrounding handler/lifecycle changes and additional buffered-stream tests. That makes this branch a duplicate now; rebasing it would reintroduce the same implementation through substantial conflicts.

The concern raised above about client/intermediary stream-idle timeouts remains a useful follow-up consideration for the landed behavior: preserving the real pre-commit HTTP status and maintaining a long-running connection are competing constraints, and any heartbeat/grace-window design needs explicit tests for both. Marking this duplicate so the review queue points at the code that is actually on `main`.

chopratejas (commented):
Checks are green on this branch and it's mergeable, but it has no human review yet, so it isn't in today's merge batch — the batch was limited to approved + fully green.

Flagging it as queued rather than overlooked. A rebase on current `main` would be useful in the meantime; ten PRs landed today and several touch the proxy handlers.

gglucass:
@chopratejas This branch has nothing left to merge. Its diff against `main` is empty (0 files changed, 0 additions), because #2997 landed this exact implementation on 2026-08-14. `_BufferedCCRResponse.__call__` on `main` today is this branch's code in both handlers, comment text included. A rebase would produce an empty commit, so this is a close-as-absorbed, as @JerrettDavis flagged above on 08-14.

Closing it does not close the open item, though. #2997 also removed the `event: ping` preamble from both handlers (its diff drops `keepalive_deadline = loop.time() + 1.0` and `b'event: ping\ndata: {"type":"ping"}\n\n'` in `anthropic.py` and `openai.py`), and shipped the justification I questioned above verbatim:

> Clients budget minutes for a turn (Claude Code sends `x-stainless-timeout: 600`), so waiting is free.

`x-stainless-timeout` is the total-request budget; the stream-idle watchdog is a separate timer, and it is the one #2465 was about. #2479 added the heartbeat specifically to close #2465, and `main` is now back to #2465's stated condition of zero bytes for the entire buffered wait. #2465 is still marked closed.

Filed as #3079 against `main` so it is tracked on the code that actually shipped rather than on a duplicate PR, with the grace-window proposal and the one thing neither PR tests. Worth a look before 0.36.0 goes out (#3067).

JerrettDavis:
Exact-head follow-up at db990a4d: the post-review main sync leaves this PR with zero changed files, confirming the fix remains fully absorbed by current main. The prior recommendation to close this duplicate still applies; no further code review action is needed on this branch.

PR #1451, opened 2026-06-26
fix(proxy): handle streaming CCR retrieval

aivinay:
## Description

Fixes Anthropic-compatible streaming requests that can emit the internal `headroom_retrieve` CCR tool. When a `stream: true` request includes the CCR retrieve tool and response handling is enabled, Headroom now buffers the upstream call as `stream: false`, lets the existing CCR response handler retrieve and continue, and returns the final result as Anthropic SSE so streaming clients do not see the internal tool call.

Closes #1450

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Detect direct Anthropic-compatible `stream: true` requests where `headroom_retrieve` is available and CCR response handling is enabled.
- Route those requests through the existing buffered/non-stream CCR response handler, then convert the final response back to `text/event-stream`.
- Fail closed with a 502 SSE error if a buffered response still contains `headroom_retrieve` after CCR handling, instead of leaking the internal tool to the client.
- Preserve Anthropic `thinking`, `redacted_thinking`, signatures, and citations when converting response JSON back to SSE.
- Add regression coverage for handled CCR retrieval, unused CCR tool availability, normal streaming passthrough, mixed client/CCR tool fail-closed behavior, and SSE conversion preservation.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ rtk gh pr checks 1451 --repo headroomlabs-ai/headroom
CI Checks Summary:
  [ok] Passed: 20
  [FAIL] Failed: 0

Relevant CI commands from .github/workflows/ci.yml:
- ruff check .
- ruff format --check .
- mypy headroom --ignore-missing-imports
- pytest tests scripts/tests

$ rtk python3 -m py_compile headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/streaming.py tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_sse_thinking_blocks.py
# passed, no output

$ rtk pytest tests/test_sse_thinking_blocks.py -q
Pytest: 6 passed

[...]

PR #2968, opened 2026-08-13
fix(proxy): report the real upstream status on buffered CCR streams

gglucass:
## Description

When a `stream: true` request carries the injected `headroom_retrieve` tool, the proxy rewrites it to a buffered `stream: false` upstream call and re-synthesizes SSE (`buffered_stream_ccr`, #1451; keepalive wrapper #2479).

`_BufferedCCRResponse.__call__` commits `http.response.start` with **status 200 and `content-type: text/event-stream`** after a 1.0s keepalive grace, before the upstream outcome is known. Once committed, a buffered call that returns the non-streaming fallback — an unparseable body, or a real 429 / 500 / 529 — could only emit a single contextless `An error occurred while processing the request.` event, with no log line at any level.

The client receives HTTP 200 with no message events and reports it as an empty or malformed response. Because the status is 200 it does not retry, so a transient upstream overload becomes a hard session failure. On Claude Code this surfaces as:

```
API Error: API returned an empty or malformed response (HTTP 200) - check for a proxy or gateway intercepting the request
```

The failure was also close to undiagnosable: the `response.json()` parse failure that routes to this fallback is logged at `debug`, the fallback `return Response(...)` logs nothing, and the request still records `status=200`.

Closes #

## Type of Change

## Changes Made

- Report the real upstream status in the SSE error event and at `warning` level when it can no longer reach the client. A status code is exactly what the client would have received had the stream not been committed; the upstream body text is still not leaked (the existing sanitization test for the exception path is unchanged).
- Log the buffered-response JSON parse failure at `warning` (with status and body size) when `buffered_stream_ccr` is set, `debug` otherwise.
- Never end a committed stream with zero events: emit an error event instead. This one is defensive — `_response_to_sse` always emits `message_start` for the `anthropic` provider — but the failure mode it guards is silent and unretryable, so it is worth pinning.
- Hoist the SSE error builder to module scope as `_ccr_sse_error_event`, so the keepalive wrapper (which lives outside the resynthesis block) can reach it.

## Testing

Two regression tests added to `tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py`. Both drive the wrapper past its keepalive so the 200 is already committed, then release the upstream:

- `test_buffered_ccr_late_non_200_reports_upstream_status_in_band` — a 529 landing after the keepalive must name 529 in-band and in the log, without leaking the upstream body.
- `test_buffered_ccr_late_unparseable_body_is_logged_and_reported` — an unparseable 200 body must produce an SSE error and a warning, never a ping-only stream.

The existing `test_buffered_ccr_preserves_early_failure_status_and_headers` covers the complementary path: when the upstream resolves *before* the keepalive fires, the real status and headers still pass through untouched.

### Test Output

Both new tests fail on the unmodified handler and pass with the fix:

```text
$ git stash push headroom/proxy/handlers/anthropic.py
$ uv run --extra dev pytest tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py -q
FAILED tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py::test_buffered_ccr_late_non_200_reports_upstream_status_in_band
FAILED tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py::test_buffered_ccr_late_unparseable_body_is_logged_and_reported
=================== 2 failed, 9 passed, 1 warning in 10.48s ====================

$ git stash pop
$ uv run --extra dev pytest tests/test_proxy/ tests/test_sse_thinking_blocks.py -q
================== 268 passed, 1 warning in 74.75s (0:01:14) ===================

$ uvx ruff check headroom/proxy/handlers/anthropic.py tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py
All checks passed!

$ uvx ruff format --check headroom/proxy/handlers/anthropic.py tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py
2 files already formatted

$ uv run --extra dev mypy headroom/proxy/handlers/anthropic.py
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), headroom-ai 0.35.0 installed runtime, Claude Code 2.1.228, Anthropic upstream, claude-opus-5, ~140k-token session with the CCR tool injected.
- Exact command / steps: ordinary Claude Code usage through the proxy, then correlated the client error against `~/.headroom/logs/proxy.log`. Measured over ~19h of logs: 291 `CCR: stream:true request has headroom_retrieve available` activations, and on one day 131 of 349 non-streaming outbound requests carried `mutation_reasons=...,ccr_streaming_retrieve_buffered_non_stream`.
- Observed result: the client reported `API returned an empty or malformed response (HTTP 200)`. The buffered upstream calls for that conversation logged `tok_out=0 ttfb_ms=0` with `status=200`, and there was no WARNING or ERROR anywhere in the proxy log. The session recovered only when the proxy was restarted.
- Not tested: the production failure was not reproduced synthetically, so which of the fallback paths fired for that specific request is not determinable from today's logs — that opacity is part of what this PR fixes. The new tests exercise the paths directly rather than reproducing the original session. No manual end-to-end run against a live Anthropic endpoint was performed with this patch.

## Known follow-ups (not in this PR)

1. Once headers are committed, the real status cannot reach the client at all — this PR only makes it visible in-band and in the log. Issuing the buffered upstream call with `client.stream(...)` would surface response headers (and status) before the body, letting the wrapper commit the real status while still buffering the body for CCR handling. That is the complete fix for the status laundering and deserves its own PR.
2. `SemanticCache.set` (`headroom/proxy/semantic_cache.py`) stores any response body gated only on `status_code == 200`, with no validity check. One empty, unparseable, or error body is then replayed to every matching non-streaming request for the full TTL (default 1h), and only a proxy restart clears it. Observed alongside this bug: three `/v1/messages` replays served in 6-10ms from a single poisoned entry. Happy to send that as a sibling PR if wanted.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This changes an existing path (`buffered_stream_ccr`) that is already reached only when CCR tool injection is on; no new rollout flag is introduced.
- Minimum rollout channel: n/a, not rollout-managed.
- Stable/default behavior changed: yes, narrowly. On the buffered-CCR failure path the client now receives an SSE `error` event naming the upstream HTTP status instead of a fixed generic string, and two diagnostics move from `debug` to `warning`. The success path is unchanged and still covered by the existing tests in this module.
- Kill switch / disable path: `--no-ccr` / `HEADROOM_NO_CCR` disables CCR tool injection, which is what puts a request on this path at all. Requests that never enter the buffered path are untouched by this PR.
- Qualification impact: none. No new dependencies, no config or schema change, no wire-protocol change. SSE `error` events are already part of the streaming contract and are already emitted by the sibling branches in this same block.
- Rollback path: revert the commit. No persisted state, no migration, no cache or on-disk format touched.

## Review Readiness


gglucass:
Flagging overlap I missed before opening this: #2959 covers the same block, arrived first, and handles both the Anthropic and OpenAI handlers. I've commented there rather than competing.

@chopratejas @JerrettDavis @DevanshiVyas — sequencing suggestion, and I'm happy to trim or close this on request:

- **#2953 is the root cause** for the report that led me here. Our logs partition buffered-CCR traffic 285 `source=canonical` (all fine) vs 6 `source=passthrough` (all failed), six for six. Evidence posted on that PR. That one is worth merging first.
- **#2959 owns the commit-ordering half** of this PR (not committing 200 before the outcome is known). I've raised one caveat there about `x-stainless-timeout` being the total-request rather than stream-idle timeout, i.e. a possible #2465 regression.
- **What does not overlap and survives either way** is the diagnosability change here: the `response.json()` failure on the buffered path is `logger.debug`, the non-SSE fallback logs nothing, and the request still records `status=200`. That combination is why this took log archaeology to identify rather than showing up as an error. This PR promotes it to `warning` with status and body size.

There is also a residual gap that #2959 does not close, if I'm reading it right: with the preamble removed, a 200 upstream whose body cannot be converted to SSE is forwarded verbatim, so a client that asked for `stream: true` still receives a non-SSE 200. That is precisely the #2952 shape. Worth handling wherever you prefer — here, in #2959, or separately.

Happy to rebase this down to just the logging change once #2959 lands, or close it if you'd rather fold everything into that PR.

JerrettDavis:
Thanks for already identifying the overlap. I agree #2959 should remain the canonical implementation: it owns the status-commit ordering across both provider handlers, while this branch is now conflicted in the same block. I’ve labeled this PR `duplicate` rather than creating another rebase/CI cycle.

The useful non-overlapping pieces are the warning-level parse diagnostic, body-size/status metadata, and the “never finish with zero SSE events” regression. Please carry those focused diagnostics/tests into #2959 (or a small follow-up after it lands) and then close this duplicate when you are comfortable. No merge or branch rewrite was performed here.

gglucass:
Closing as absorbed. @JerrettDavis, the pieces you asked me to carry over are all on `main` already via #2997: the warning-level parse diagnostic on the buffered-stream path, the verbatim SSE relay when upstream streams instead of buffering, and the malformed-200 rejection that returns a real 502 instead of a zero-event 200. My diff against `main` is now empty, so there is nothing left here to rebase.

The remaining non-overlapping concern is the heartbeat removal that rode along with #2997, which I have filed separately as #3079 rather than reopening this branch.

PR #2972, opened 2026-08-13
fix(install/windows): stop the installer writing a sandboxed dir to the user PATH

### LINKED ISSUE #2970: [BUG] install.ps1 leaks pytest temp dirs into the persistent Windows user PATH
## Description

`scripts/install.ps1` persists the install directory to the user's PATH through
`Ensure-PathEntry`, which calls `[Environment]::SetEnvironmentVariable('Path', ..., 'User')`.
That value lives in the `HKCU\Environment` registry key, so it is **not** scoped by a
`HOME` / `USERPROFILE` override.

`tests/test_install/test_native_installers.py::test_powershell_native_installer_supports_persistent_docker_lifecycle`
executes that real installer against a `tmp_path` fake home. Every run therefore prepends the
test's throwaway shim directory to the developer's actual, persistent user PATH — and it stays
there after the test finishes.

The entries accumulate silently, one per run, ahead of the developer's real install directory. The
installer also drops `headroom.ps1` / `headroom.cmd` Docker wrappers into that directory, so once
the leak has happened, `headroom` on a fresh shell resolves to the leftover wrapper from a deleted
temp directory instead of the real executable.

## To Reproduce

On a Windows host with PowerShell available:

1. Note the current user PATH:
   `([Environment]::GetEnvironmentVariable('Path','User') -split ';').Count`
2. Run the installer test suite: `uv run pytest tests/test_install/test_native_installers.py -q`
3. Re-read the user PATH — it has gained an entry, permanently:
   `([Environment]::GetEnvironmentVariable('Path','User') -split ';') | Where-Object { $_ -like '*pytest-of-*' }`
4. Repeat step 2 a few times, then open a new shell and run `headroom --version`.

## Expected Behavior

Running the test suite leaves the developer's persistent environment untouched. The installer's
PATH mutation should be suppressible when it is installing into a sandboxed home.

## Actual Behavior

Each test run leaves one more `pytest-of-<user>\pytest-N\test_powershell_native_install0\home\.local\bin`
entry at the front of the persistent user PATH. On my machine 12 had built up, so `headroom`
resolved to the stale Docker wrapper and every invocation failed:

## Error Output

```
PS G:\...\headroom> headroom wrap claude --memory --learn
Missing required command: docker
At C:\Users\<user>\AppData\Local\Temp\pytest-of-<user>\pytest-92\test_powershell_native_install0\home\.local\bin\headroom
.ps1:9 char:5
+     throw $Message
+     ~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (Missing required command: docker:String) [], RuntimeException
    + FullyQualifiedErrorId : Missing required command: docker
```

`Get-Command headroom -All` confirmed the shadowing:

```
CommandType    Name          Source
-----------    ----          ------
ExternalScript headroom.ps1  C:\Users\<user>\AppData\Local\Temp\pytest-of-<user>\pytest-92\...\.local\bin
Application    headroom.cmd  C:\Users\<user>\AppData\Local\Temp\pytest-of-<user>\pytest-92\...\.local\bin
Application    headroom.exe  C:\Users\<user>\.local\bin\headroom.exe
```

## Why CI does not catch this

`.github/workflows/ci.yml` runs `pytest tests/test_install/test_native_installers.py -q` on
`windows-latest`. The runner's PATH is polluted the same way, but the runner is destroyed
afterwards, so the suite stays green while only real Windows contributors accumulate the damage.

## Environment

- **Headroom version**: 0.35.0
- **Python version**: 3.13.11
- **LLM Provider**: Anthropic

## Additional Context

`scripts/install.sh` is unaffected: it only writes shell profile files under `$HOME`, which the
fake home already sandboxes. The problem is specific to the Windows registry write.

Suggested fix: have `Ensure-PathEntry` honour an opt-out environment variable
(`HEADROOM_INSTALL_SKIP_PATH`), and set it in the test suite's environment builder, plus a
regression test asserting `HKCU\Environment\Path` is unchanged across an installer run. PR follows.

Parideboy:
## Description

`Ensure-PathEntry` in `scripts/install.ps1` persists the install directory through
`[Environment]::SetEnvironmentVariable('Path', ..., 'User')`, which writes `HKCU\Environment`.
That registry key is not scoped by a `HOME`/`USERPROFILE` override, so the native installer tests,
which run the real installer against a `tmp_path` fake home, prepended their throwaway shim
directory to the developer's persistent user PATH on every run, where it outlived the test.

The entries accumulate ahead of the real install directory, and the installer drops
`headroom.ps1` / `headroom.cmd` Docker wrappers into that directory. So after a few test runs
`headroom` in a new shell resolves to a wrapper inside a since-deleted temp directory and fails with
`Missing required command: docker`.

CI cannot catch this: `.github/workflows/ci.yml` runs that test file on `windows-latest`, whose
polluted PATH dies with the runner. Only real Windows contributors accumulate the damage.

`scripts/install.sh` is unaffected: it only writes shell profile files under `$HOME`, which the
fake home already sandboxes.

Closes #2970

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `scripts/install.ps1`: `Ensure-PathEntry` returns early when `HEADROOM_INSTALL_SKIP_PATH` is set, so callers installing into a throwaway home can opt out of the persistent registry write. Default behavior for real installs is unchanged.
- `tests/test_install/test_native_installers.py`: `_build_env` sets `HEADROOM_INSTALL_SKIP_PATH=1` next to the existing `HOME`/`USERPROFILE` overrides, so every installer test inherits the opt-out.
- `tests/test_install/test_native_installers.py`: new Windows-only regression test `test_powershell_native_installer_leaves_the_real_user_path_untouched` reads the raw `HKCU\Environment` PATH value with `winreg`, runs the installer, and asserts the value is byte-identical afterwards. It restores the value only when it actually changed, so a regressed guard cannot leave a contributor's PATH polluted by the test that detects it.

## Testing

### Test Output

```text
$ uv run pytest tests/test_install/test_native_installers.py -q
platform win32 -- Python 3.13.11, pytest-9.0.3, pluggy-1.6.0
collected 3 items

tests\test_install\test_native_installers.py s..                         [100%]

======================== 2 passed, 1 skipped in 10.08s ========================

$ uv run ruff check .
All checks passed!

$ uv run ruff format --check tests/test_install/test_native_installers.py
1 file already formatted

$ uv run mypy headroom --ignore-missing-imports
```

Fails before the fix. With `scripts/install.ps1` stashed, the new test catches the leak:

```text
$ git stash push -- scripts/install.ps1
$ uv run pytest tests/test_install/test_native_installers.py -q -k "leaves_the_real_user_path"
tests\test_install\test_native_installers.py:864: in test_powershell_native_installer_leaves_the_real_user_path_untouched
    assert str(home) not in (after[0] if after else "")
E   AssertionError: assert 'C:\\Users\\...stall0\\home' not in 'C:\\Users\\...l_build\\bin'
E     'C:\Users\...\pytest-97\test_powershell_native_install0\home' is contained here:
E       C:\Users\<user>\AppData\Local\Temp\pytest-of-<user>\pytest-97\test_powershell_native_install0\home\.local\bin;C:\Users\<user>\AppData\Local\Programs\OpenAI\Codex\bin;...

======================= 1 failed, 2 deselected in 4.24s =======================
```

## Real Behavior Proof

- Environment: Windows 11 Pro 10.0.26200, PowerShell 7, Python 3.13.11, headroom 0.35.0, provider Anthropic
- Exact command / steps: counted the persistent user PATH entries with `([Environment]::GetEnvironmentVariable('Path','User') -split ';' | Where-Object {$_}).Count`; ran `uv run pytest tests/test_install/test_native_installers.py -q` on the patched tree; re-counted; filtered for `*pytest-of-*`; then ran `Get-Command headroom -All` and `headroom --version` in a fresh shell
- Observed result: before this fix my user PATH had accumulated 12 `...\pytest-of-<user>\pytest-N\test_powershell_native_install0\home\.local\bin` entries ahead of `C:\Users\<user>\.local\bin`, and `headroom wrap claude --memory --learn` failed with `Missing required command: docker` thrown from `...\pytest-92\...\headroom.ps1:9`. After cleaning PATH (29 entries down to 17) and applying this patch, a full run of the installer test file leaves the count at 17 with zero `pytest-of-*` matches, `Get-Command headroom -All` returns only `C:\Users\<user>\.local\bin\headroom.exe`, and `headroom --version` prints `headroom, version 0.35.0`. Running the new test against the unpatched `install.ps1` reproduces the leak and then self-restores the registry value.
- Not tested: non-Windows hosts (the new test and the changed code path are Windows-only; the Linux/macOS `install.sh` path is untouched), `ConfigScope.SYSTEM` / Machine-scope installs, and elevated/admin installs

## Runtime Rollout Safety

- Rollout-managed feature(s): none (installer script change, not runtime-gated)
- Stable/default behavior changed: no; a real install with `HEADROOM_INSTALL_SKIP_PATH` unset behaves exactly as before
- Kill switch / disable path: leave `HEADROOM_INSTALL_SKIP_PATH` unset (the default)
- Qualification impact: none
- Rollback path: revert this commit; the installer returns to writing the user PATH unconditionally

## Review Readiness


JerrettDavis (approved):
Reviewed the latest head after updating it to current `main`, including the installer mutation boundary and the failure-path registry restoration. The hidden test/sandbox opt-out is narrower than exposing a general persistence target: production remains hard-wired to user scope, while every native-installer test inherits the no-persistence guard from one environment builder. The Windows regression records the raw value and registry kind and restores only on a detected regression.

Local exact-head installer coverage passes where applicable (1 passed, 2 Windows-only skips); the previous exact-head Windows run supplied the real registry proof. No blocking correctness findings. #2985 is now labeled duplicate so one canonical fix can proceed.

chopratejas (commented):
Thanks for this — the diagnosis was right and it was one of two correct fixes filed for #2970.

I've merged #2985 for that issue and I want to be explicit about why, because it isn't a quality judgement on this PR.

Both changes gate `Ensure-PathEntry` on an env var set by the test harness. The difference is what the tests then exercise:

- **This PR** (`HEADROOM_INSTALL_SKIP_PATH`) returns from the function early, so the installer tests no longer execute the PATH-writing code at all. The tested path diverges from the production path exactly where the bug lived.
- **#2985** (`HEADROOM_INSTALL_PATH_SCOPE=Process`) still runs the whole function and still writes a PATH entry — just to the ephemeral Process scope. The logic stays under test, and it additionally allow-lists `User`/`Process` and throws on `Machine`, which closes a real foot-gun if the variable is ever inherited by an elevated installer.

One thing this PR does better, and which #2985 does **not** carry over: your `_read_user_path_entry` / `_restore_user_path_entry` helpers assert against the real `HKCU\Environment` rather than trusting the env var to have worked. That's a stronger proof, and it would be a genuinely useful follow-up on top of the merged fix — it verifies the guard instead of assuming it.

Closing as superseded, but please do open that follow-up if you're willing. Sorry you both ended up writing the same fix; that's a triage failure on our side, not yours.

Parideboy:
Makes sense, and thanks for spelling out the reasoning rather than just closing it.

Agreed on the substantive point: returning early from `Ensure-PathEntry` means the installer tests stop exercising the write path exactly where the bug was, whereas redirecting to `Process` keeps the function under test. The `Machine` rejection in #2985 is a genuine extra, since the variable is inheritable by an elevated installer.

Follow-up opened as #3068. It ports the raw-registry helpers onto the merged fix and rewrites `test_powershell_installer_does_not_leak_into_user_path` rather than adding a third overlapping test. Beyond the count comparison it now catches two further cases: the .NET getter expands `%USERPROFILE%`-style references, so it cannot see a change of the registry value kind, and a count comparison also passes when an entry is replaced or reordered rather than appended. It restores the recorded value only when the value actually changed, so a passing run writes nothing to the registry and a regressed run cannot leave the pollution behind that #2970 was about.

Closing this as superseded. No hard feelings about the duplicate work.

PR #2600, opened 2026-07-27
fix(install): fall back to Task Scheduler on Windows

Inference1:
## Description

On Windows, `persistent-service` registered the Python command runner as a raw `sc.exe` service. That runner cannot participate in the Service Control Manager protocol, so `sc.exe start` always failed with error 1053. This change resolves `persistent-service` to the existing Task Scheduler-backed `persistent-task` implementation on Windows, which provides startup and periodic health recovery.

Closes #2552

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Resolve Windows `persistent-service` manifests to `persistent-task` before supervisor installation.
- Display an explicit Task Scheduler fallback warning in `headroom install apply`.
- Add planner and CLI regression coverage for the fallback and absence of `sc.exe` output.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes for changed files (`ruff check`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [ ] Manual testing performed

### Test Output

```text
pytest tests/test_install/test_planner.py tests/test_install/test_supervisors.py tests/test_cli/test_install_cli.py::test_install_apply_announces_windows_service_fallback -q --tb=short
47 passed, 1 warning in 0.60s

uvx ruff check headroom/install/planner.py headroom/cli/install.py tests/test_install/test_planner.py tests/test_cli/test_install_cli.py
All checks passed!

mypy headroom
Success: no issues found in 509 source files
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.1, pytest 9.1.1.
- Exact command / steps: Run `headroom install apply --preset persistent-service` on Windows; the manifest planner now resolves it to `persistent-task`, and the CLI announces the Task Scheduler fallback.
- Observed result: The focused Windows regression tests pass, the manifest selects `persistent-task`, and the CLI warning explicitly names Task Scheduler. The terminal capture below shows both focused tests passing and the generated warning text.
- Not tested: A full elevated Task Scheduler registration against a live proxy process.

## Review Readiness

- [x] I have performed a self-review
- [x] This PR is ready for human review

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
[...]

PR #1628, opened 2026-07-01
fix(wrap): verify proxy deps before mutating Codex config

syf2211:
## Description

\`headroom wrap codex\` now verifies that optional proxy dependencies (\`headroom-ai[proxy]\`) are installed before mutating Codex \`config.toml\`. If the check fails, the command exits with the same error message as \`headroom proxy\` and leaves Codex config untouched.

Fixes #1614 (Bug 1: config mutated before proxy dependency check).

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Extract \`ensure_proxy_dependencies()\` in \`headroom/cli/proxy.py\` (shared with \`headroom proxy\`)
- Call it at the start of \`wrap codex\` when \`not no_proxy\`, before config snapshot/injection
- Add regression tests for prepare-only abort, \`--no-proxy\` skip, and import failure messaging

## Testing

- [x] Unit tests pass (\`pytest\`)
- [x] Linting passes (\`ruff check .\`)
- [ ] Type checking passes (\`mypy headroom\`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

\`\`\`bash
pytest tests/test_cli/test_wrap_codex.py::test_wrap_codex_aborts_before_mutating_config_when_proxy_deps_missing \
       tests/test_cli/test_wrap_codex.py::test_wrap_codex_skips_proxy_dependency_check_with_no_proxy \
       tests/test_cli/test_wrap_codex.py::test_ensure_proxy_dependencies_exits_when_server_import_fails -q
# 3 passed
ruff check headroom/cli/wrap.py headroom/cli/proxy.py tests/test_cli/test_wrap_codex.py
ruff format --check headroom/cli/wrap.py headroom/cli/proxy.py tests/test_cli/test_wrap_codex.py
\`\`\`

## Real Behavior Proof

Environment: Linux (Ubuntu), Python 3.12, local checkout with \`PYTHONPATH\` pointed at patched sources.

Exact command / steps:
1. Created a temp \`~/.codex/config.toml\` with \`model_provider = "openai"\`.
2. Patched \`headroom.cli.wrap.ensure_proxy_dependencies\` to raise \`SystemExit(1)\` (simulating missing \`[proxy]\` extra).
[...]

PR #2980, opened 2026-08-13
fix(install): consolidate Windows fallback and cleanup safety

### LINKED ISSUE #2552: install apply --preset persistent-service: sc.exe create succeeds but sc.exe start always fails (1053) — wrapped script never implements the Service Control Protocol
Body:

## Summary

`headroom install apply --preset persistent-service` on Windows successfully registers the service (`sc.exe create` succeeds), but `sc.exe start` always fails with error 1053 ("The service did not respond to the start or control request in a timely fashion"). The installer then auto-rolls-back the failed deployment, removing the service and deploy directory.

This is a different failure mode than #1654 (which was a `sc.exe create` quoting bug that's since been fixed — thanks for that). This bug happens *after* `create` succeeds, at `start`.

**Environment**
- Windows 11 Home 10.0.26200
- `headroom-ai` 0.32.1 (installed via `uv tool install headroom-ai`)
- Elevated ("Run as Administrator") PowerShell 7.6.3
- Python runtime, `--scope user`

**Command run**
```powershell
headroom install apply --preset persistent-service --profile myprofile --port 8787
Output

[SC] CreateService SUCCESS
[SC] ChangeServiceConfig2 SUCCESS
[SC] StartService FAILED 1053:
The service did not respond to the start or control request in a timely fashion.
[SC] ControlService FAILED 1062:
The service has not been started.
Error: Cannot start deployment 'myprofile': command failed (sc.exe start headroom-myprofile)
Reproduced twice, consistent.

Root cause
Looking at headroom/install/supervisors.py (around line 333), the Windows service is created with:

f"sc.exe create {manifest.service_name} "
f'binPath= "cmd.exe /c \\"{run_cmd}\\"" start= auto'
binPath= points at cmd.exe /c "...run-headroom.cmd", which just launches the Python proxy as a normal console process. Windows Services are required to actively participate in the Service Control Protocol — the executable must call StartServiceCtrlDispatcher and then report SERVICE_RUNNING back to the SCM via SetServiceStatus. A plain cmd.exe/Python process never does this, so the SCM waits for a status update that never arrives and times out with 1053 every time. This isn't an environment quirk — any non-service-aware executable registered this way via raw sc.exe create will hit the same wall on any Windows machine.

Suggested fix
Raw sc.exe create pointed at an ordinary script can't work for this. Options, roughly in order of how "proper" they are:

Ship a small service-host shim that does speak the SCM protocol (e.g. via pywin32's win32serviceutil.ServiceFramework) and spawns/monitors the actual proxy process as a child — this is the standard pattern (similar to what NSSM/WinSW do).
Bundle or shell out to an existing battle-tested wrapper like NSSM or WinSW for the Windows service case specifically, rather than raw sc.exe.
At minimum, if a proper service host isn't feasible short-term: detect this failure mode and fall back to persistent-task automatically (with restart-on-failure settings — see below) instead of silently leaving the user with nothing installed.

### LINKED ISSUE #2817: [BUG] headroom wrap deletes a user's own rtk / lean-ctx install (MCP entry, hook scripts, hook entries)

### PR DESCRIPTION
**Pull request by JerrettDavis**
## Description

Consolidates two fully reviewed installation-safety fixes whose original PRs can no longer merge under current branch protection: Windows persistent-service deployments need a supported Task Scheduler fallback, and legacy context-tool cleanup must never delete user-owned RTK/lean-ctx artifacts.

Closes #2552
Closes #2817

Supersedes #2600 and #2828 while preserving their authors' commits and review-driven corrections.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- Convert Windows `persistent-service` plans to the supported `persistent-task` supervisor and make the fallback explicit in CLI output.
- Restrict context-tool cleanup to artifacts proven to live under Headroom's managed directory.
- Recognize wrapped, relative, and platform-specific managed commands without accepting prefixed/path-boundary lookalikes.
- Scope cleanup completion state correctly across projects and alternate agent homes.
- Stamp cleanup complete only after all managed remnants are settled.
- Preserve the original focused regression suites and behavior-proof artifact.

## Testing

### Test Output

```text
$ uv run pytest -q tests/test_install/test_planner.py tests/test_install/test_supervisors.py tests/test_cli/test_install_cli.py tests/test_context_tool_cleanup.py tests/test_cli/test_unwrap_claude.py
135 passed in 0.45s

$ uv run ruff check <changed Python and test files>
All checks passed!

$ uv run ruff format --check <changed Python and test files>
8 files already formatted
```

## Real Behavior Proof

- Environment: macOS arm64 for consolidated current-main validation; the Windows fallback source PR was independently validated on Windows and includes its captured verification artifact.
- Exact command / steps: run the planner, supervisor, install CLI, cleanup provenance, and unwrap suites on the rebased combined branch.
- Observed result: 135/135 focused tests pass. Windows service requests resolve to `persistent-task`; cleanup rejects user-owned and path-prefix lookalikes while removing managed artifacts.
- Not tested: a fresh privileged Windows host deployment in this local pass; #2600's accepted review contains the Windows-specific proof.

## Runtime Rollout Safety

- Rollout-managed feature(s): Install supervisor selection and one-time legacy cleanup.
- Minimum rollout channel: Stable/default; both prevent currently destructive or nonfunctional install paths.
- Stable/default behavior changed: Windows service requests use Task Scheduler; cleanup requires managed provenance.
- Kill switch / disable path: Select `persistent-task` explicitly; cleanup remains bounded by its completion stamp and provenance checks.
- Qualification impact: Windows native install and wrap/unwrap cleanup suites.
- Rollback path: Revert this PR, restoring the two pre-fix behaviors.

## Review Readiness

## Checklist

The Windows verification artifact from #2600 is retained at `.github/pr-images/issue-2552-windows-fallback-verification.png`.

## Additional Notes

This is intentionally an installation-safety batch rather than two replacement PRs. Original commit authorship is preserved, and the combined diff was applied cleanly to current `main` after #2832 and #1628 landed.


JerrettDavis:
CI follow-up: reproduced shard 1 locally (1 failed / 2,750 passed). The new cleanup completion stamp created `~/.headroom` on a pristine machine before `wrap opencode` could reject a missing binary. The stamp now writes only when Headroom state already exists: established installs retain the one-time fast path, while clean missing-tool failures remain read-only. Focused cleanup/OpenCode regressions: **7 passed**, Ruff + format clean.

JerrettDavis:
Semantic rebase audit: force-updated this branch onto current main after #2978 so the MCP v1 compatibility fix is part of the tested branch, not merely something GitHub would combine at merge time. Revalidated the inherited MCP dependency contract together with the Windows fallback and ownership-safe cleanup suites: 107 passed. No current-main behavior was dropped.

PR #2984, opened 2026-08-13
fix(proxy/toin): address TOIN patterns by tool hash, not the auth|model key head

### LINKED ISSUE #2928: sig_hash[:12] truncates the composite auth|model|hash key, so /v1/toin patterns share one id and detail lookup is arbitrary

abhay-codes07:
## Description

`/v1/toin/patterns` reports each pattern's identifier as `sig_hash[:12]`, documented as the "tool signature hash". But `sig_hash` is not a hash: it is the composite aggregation key encoded as `"auth|model|hash"` (`telemetry/toin._serialize_pattern_key`). So the first 12 characters are the `auth_mode|model_family` head, not the tool signature.

In a single-tenant deployment every pattern shares that head, so all patterns report the *same* id. The reporter saw all 1,797 patterns return `unknown|unkn`. `/v1/toin/pattern/{hash_prefix}` then matches with `sig_hash.startswith(hash_prefix)` on the full composite key and returns on the first hit, so with identical heads it returns an arbitrary pattern. The two diagnostic endpoints therefore cannot address a specific pattern (#2928).

Impact is low (diagnostic endpoints only, no effect on compression), but it defeats the stated purpose of the endpoints.

## Fix

Add `_toin_tool_hash(sig_hash)`, which returns the component after the last separator (`sig_hash.rsplit("|", 1)[-1]`, also correct for the legacy bare-hash format), and use it in both endpoints:

- `/v1/toin/patterns`: `"hash": _toin_tool_hash(sig_hash)[:12]` -- the real tool-signature hash, which varies per pattern.
- `/v1/toin/pattern/{prefix}`: match `_toin_tool_hash(sig_hash).startswith(prefix)` -- the same component the list advertises.

Each pattern now gets a distinct, addressable id.

Fixes #2928

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/server.py`: add `_toin_tool_hash`; use it for the `hash` field in `/v1/toin/patterns` and for the prefix match in `/v1/toin/pattern/{hash_prefix}`.
- `tests/test_proxy_loopback_gating.py`: two patterns in one tenant slice now report distinct hashes (not `unknown|unkn`); each id addresses exactly its pattern via the detail endpoint; a unit test for `_toin_tool_hash` (composite + legacy). Updated the existing detail test, which addressed by the composite-key head (`unknown`), to use the tool-hash prefix.

## Testing

- [x] New tests added

### Test Output

```text
# Fail-before (endpoint usages reverted to sig_hash[:12] / sig_hash.startswith):
test_toin_patterns_hash_is_the_tool_hash_not_the_key_head   -> hashes were {"unknown|unkn"}  (FAIL)
test_toin_pattern_detail_addresses_a_specific_pattern       -> detail lookup by tool-hash prefix 404s  (FAIL)

# Pass-after:
tests/test_proxy_loopback_gating.py  76 passed
# uvx mypy@1.20.2 headroom/proxy/server.py -> Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: drove the real endpoints via FastAPI `TestClient` (loopback) with `get_toin` mocked to two patterns sharing the `unknown|unknown` tenant slice but different tool hashes (`...aaaaaaaaaaaa1`, `...bbbbbbbbbbbb2`). `/v1/toin/patterns` now returns hashes `{aaaaaaaaaaaa, bbbbbbbbbbbb}` instead of both `unknown|unkn`, and `/v1/toin/pattern/aaaaaaaaaaaa` vs `/bbbbbbbbbbbb` return the correct distinct patterns.
- Observed result: the two diagnostic endpoints can address a specific TOIN pattern again.
- Not tested: a live proxy with a populated TOIN store (the store is populated by real traffic). The endpoint behaviour is verified directly against the export shape the endpoints read.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. `/v1/toin/patterns` and `/v1/toin/pattern/{hash}` are loopback-gated diagnostic endpoints, not rollout-channel-gated.
- Stable/default behavior changed: no functional proxy behavior. Only the `hash` field these diagnostic endpoints report is corrected (the tool-hash component instead of the aggregation-key head).
- Kill switch / disable path: N/A (diagnostic-only, loopback-gated).
- Qualification impact: none. Token accounting and routing are untouched.
- Rollback path: revert this PR; the endpoints return to their prior hash field with no state to migrate.

## Review Readiness

## Checklist

## Additional Notes

Two patterns that share the exact same tool signature but differ in auth/model would still share a tool-hash prefix; that is inherent to a "tool signature hash" and matches the endpoint's documented prefix semantics. It is a strict improvement over the previous behaviour, where the id was the constant auth|model head.


JerrettDavis (changes requested):
This reproduces the same unresolved identifier-collision issue already called out on #2932. Extracting the signature component fixes the `unknown|unkn` collision, but it still does not identify a TOIN v2 pattern. The store key is deliberately `(auth_mode, model_family, signature_hash)`, and the same tool signature commonly appears in multiple auth/model slices with independent counters and recommendations. Those rows will expose the same 12-character id, and the detail endpoint will still return whichever slice iterates first.

The current tests use two different signature hashes in one slice, so they cannot prove the promised “each pattern gets a distinct, addressable id.” Please add the adversarial case: identical signature hash, different auth/model slices, different counters. Then preserve full slice identity in an opaque URL-safe id (or require the slice fields), and return an explicit ambiguity response for legacy signature-only prefixes rather than arbitrary data.

Because #2932 implements the same partial approach and already has this exact requested change, please coordinate one canonical PR and close/supersede the duplicate rather than maintaining two competing fixes for #2928. I refreshed this branch to current `main` before reviewing; no merge action was taken.

abhay-codes07:
Agreed, and thanks for the review. #2932 by @chopratejas predates this (created 2026-08-11 vs this one on 2026-08-13) and is the canonical fix for #2928, so I am closing this as superseded to avoid two competing PRs. I will not maintain a parallel change.

On the substantive point you raised for the canonical PR: the adversarial case that matters is identical signature_hash across different (auth_mode, model_family) slices, which signature-only addressing cannot disambiguate. Preserving full slice identity in an opaque URL-safe id (or requiring the slice fields) and returning an explicit ambiguity response for legacy signature-only prefixes, rather than whichever slice iterates first, is the right shape. Happy to help land that on #2932 if useful.

PR #2985, opened 2026-08-13
fix(install): stop the PowerShell installer leaking temp dirs into the real user PATH

### LINKED ISSUE #2970: [BUG] install.ps1 leaks pytest temp dirs into the persistent Windows user PATH

abhay-codes07:
## Description

`scripts/install.ps1` persists the install directory to the user's PATH through `Ensure-PathEntry`, which calls `[Environment]::SetEnvironmentVariable('Path', ..., 'User')`. That value lives in the `HKCU\Environment` registry key, so it is **not** scoped by a `HOME` / `USERPROFILE` override.

`tests/test_install/test_native_installers.py::test_powershell_native_installer_supports_persistent_docker_lifecycle` runs that real installer against a `tmp_path` fake home. Every run therefore prepended the test's throwaway shim directory to the developer's actual, persistent user PATH -- and it stayed there after the test finished. The entries accumulate one per run, ahead of the real install dir; and since the installer also drops `headroom.ps1`/`headroom.cmd` into that dir, `headroom` in a fresh shell could then resolve to a leftover wrapper from a deleted temp directory (#2970).

## Fix

Make the persistence scope configurable via `HEADROOM_INSTALL_PATH_SCOPE`, defaulting to `'User'` so production behavior is unchanged:

```powershell
$scope = if ($env:HEADROOM_INSTALL_PATH_SCOPE) { $env:HEADROOM_INSTALL_PATH_SCOPE } else { 'User' }
$currentPath = [Environment]::GetEnvironmentVariable('Path', $scope)
...
[Environment]::SetEnvironmentVariable('Path', ($newPath -join ';'), $scope)
```

The installer tests (`_build_env`) set `HEADROOM_INSTALL_PATH_SCOPE=Process`, so the PATH update stays in the spawned PowerShell process (discarded when it exits) instead of writing to the registry.

Fixes #2970

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `scripts/install.ps1` (`Ensure-PathEntry`): read/write the PATH via `$env:HEADROOM_INSTALL_PATH_SCOPE` (default `'User'`).
- `tests/test_install/test_native_installers.py`: `_build_env` sets `HEADROOM_INSTALL_PATH_SCOPE=Process` for every installer invocation; add a Windows-only `test_powershell_installer_does_not_leak_into_user_path` asserting the real User PATH entry count is unchanged across an installer run.

## Testing

- [x] New test added

### Test Output

```text
tests/test_install/test_native_installers.py -k does_not_leak_into_user_path  1 passed
# uvx ruff@0.15.22 check tests/test_install/test_native_installers.py -> All checks passed!
```

## Real Behavior Proof

- Environment: Windows 11, Windows PowerShell 5.1, Python 3.12.11, project venv, pytest 9.1.1, ruff 0.15.22 via uvx.
- Exact command / steps: recorded the real user PATH entry count (`([Environment]::GetEnvironmentVariable('Path','User') -split ';').Count` = 27), ran the PowerShell installer test with the fix, then re-read the count: still 27 -- no leak. The new `test_powershell_installer_does_not_leak_into_user_path` formalizes this (before == after).
- Observed result: running the installer test suite no longer mutates the developer's persistent user PATH; production installs still persist to `'User'` as before.
- Not tested: the sibling `test_powershell_native_installer_supports_persistent_docker_lifecycle` fails on my Windows host on an unrelated `trusted_cidrs` dashboard-gateway assertion (it fails identically on `main` without this change, and the whole PowerShell suite is skipped on the Linux CI runners). This PR does not touch that path.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This is the native PowerShell installer script, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: no. Production installs still persist PATH to the `User` scope exactly as before; the new `HEADROOM_INSTALL_PATH_SCOPE` override defaults to `User` and is used only by the test suite to avoid mutating the developer's persistent PATH.
- Kill switch / disable path: leave `HEADROOM_INSTALL_PATH_SCOPE` unset (the default) for the normal `User` behavior.
- Qualification impact: none. Installer-only; no proxy runtime path is touched.
- Rollback path: revert this PR; the installer returns to writing the `User` PATH unconditionally.

## Review Readiness

## Checklist

## Additional Notes

The scope override defaults to `'User'`, so nothing changes for real installs. It doubles as an escape hatch for any environment (CI images, ephemeral containers) that must not touch the persistent user PATH.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -22,14 +22,22 @@ function Require-Command {
 function Ensure-PathEntry {
     param([string]$PathEntry)
 
-    $currentPath = [Environment]::GetEnvironmentVariable('Path', 'User')
+    # Persist to the User PATH by default. The 'User' scope lives in
+    # HKCU\Environment and is NOT redirected by a HOME/USERPROFILE override, so a
+    # caller that must not mutate the real persistent PATH (the installer test
+    # suite, which runs this against a throwaway fake home) sets
+    # HEADROOM_INSTALL_PATH_SCOPE=Process to keep the update ephemeral instead of
+    # leaking the temp shim dir into the developer's actual user PATH (#2970).
+    $scope = if ($env:HEADROOM_INSTALL_PATH_SCOPE) { $env:HEADROOM_INSTALL_PATH_SCOPE } else { 'User' }
+
+    $currentPath = [Environment]::GetEnvironmentVariable('Path', $scope)
     $parts = @()
     if ($currentPath) {
         $parts = $currentPath -split ';' | Where-Object { $_ }
     }
     if ($parts -notcontains $PathEntry) {
         $newPath = @($PathEntry) + $parts
-        [Environment]::SetEnvironmentVariable('Path', ($newPath -join ';'), 'User')
+        [Environment]::SetEnvironmentVariable('Path', ($newPath -join ';'), $scope)
     }
 }
 
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -28,7 +28,23 @@ function Ensure-PathEntry {
     # suite, which runs this against a throwaway fake home) sets
     # HEADROOM_INSTALL_PATH_SCOPE=Process to keep the update ephemeral instead of
     # leaking the temp shim dir into the developer's actual user PATH (#2970).
-    $scope = if ($env:HEADROOM_INSTALL_PATH_SCOPE) { $env:HEADROOM_INSTALL_PATH_SCOPE } else { 'User' }
+    #
+    # Only those two persistence modes are supported. The value is handed to
+    # .NET's EnvironmentVariableTarget, whose 'Machine' member would rewrite the
+    # SYSTEM-wide PATH if this variable were inherited by an elevated installer,
+    # and a typo would otherwise fail late with an opaque enum-conversion error.
+    # Normalize case-insensitively and allow-list 'User'/'Process', failing early
+    # and clearly for 'Machine' or anything else.
+    $scope = 'User'
+    if ($env:HEADROOM_INSTALL_PATH_SCOPE) {
+        switch ($env:HEADROOM_INSTALL_PATH_SCOPE.Trim().ToLowerInvariant()) {
+            'user' { $scope = 'User' }
+            'process' { $scope = 'Process' }
+            default {
+                throw "HEADROOM_INSTALL_PATH_SCOPE must be 'User' or 'Process' (got '$($env:HEADROOM_INSTALL_PATH_SCOPE)'); 'Machine' and other targets are not supported."
+            }
+        }
+    }
 
     $currentPath = [Environment]::GetEnvironmentVariable('Path', $scope)
     $parts = @()
--- a/tests/test_install/test_native_installers.py
+++ b/tests/test_install/test_native_installers.py
@@ -224,6 +224,11 @@ def _build_env(home: Path, tmp_path: Path) -> dict[str, str]:
     env["PATH"] = str(shim_dir) + os.pathsep + env.get("PATH", "")
     env["FAKE_DOCKER_STATE"] = str(tmp_path / "fake-docker-state.json")
     env["FAKE_DOCKER_LOG"] = str(tmp_path / "fake-docker.log")
+    # #2970: the PowerShell installer's Ensure-PathEntry persists to the 'User'
+    # PATH scope (HKCU\Environment), which a HOME/USERPROFILE override does not
+    # redirect. Keep the PATH update ephemeral (Process scope) so running these
+    # tests never leaks the throwaway shim dir into the developer's real PATH.
+    env["HEADROOM_INSTALL_PATH_SCOPE"] = "Process"
     return env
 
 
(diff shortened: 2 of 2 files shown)


JerrettDavis (changes requested):
The leak fix is directionally correct, but the new persistence boundary needs validation before this is safe for an enterprise installer.

`HEADROOM_INSTALL_PATH_SCOPE` is passed directly into the .NET `EnvironmentVariableTarget` conversion. That silently creates an undocumented `Machine` mode: if this environment variable is inherited by an elevated installer, it can rewrite the system-wide PATH rather than the user or ephemeral process PATH. Any typo also fails later with an opaque overload/enum conversion error. The PR only needs two supported semantics—default `User` and test/ephemeral `Process`.

Please normalize and explicitly allow only `User` and `Process` (case-insensitively if desired), fail early with a clear message for everything else, and add script-level coverage for both the accepted `Process` override and rejected `Machine`/invalid values. The existing before/after user-PATH test is useful, but validation should be pinned independently so a future caller cannot widen persistence accidentally.

The current `main` refresh was applied before this review; no merge action was taken.

abhay-codes07:
Thanks, fixed. HEADROOM_INSTALL_PATH_SCOPE is now normalized case-insensitively and allow-listed to exactly the two supported semantics, default User and test/ephemeral Process. Anything else, including Machine, throws early with a clear message ("must be 'User' or 'Process'... 'Machine' and other targets are not supported") before any PATH read or write, so an inherited or mistyped value can no longer reach EnvironmentVariableTarget and rewrite the system-wide PATH, and there is no opaque enum-conversion failure.

Validation is pinned independently of the before/after User-PATH test: two script-level tests AST-extract Ensure-PathEntry from install.ps1 and drive it directly, one asserting Process (any case) runs, the other asserting Machine/machine/system/bogus are all rejected before mutating PATH. Neither exercises User, so they never touch the real HKCU PATH.

JerrettDavis (approved):
The prior installer-safety blocker is resolved in the latest source: the scope is normalized and explicitly restricted to User/Process before any PATH access, with Machine and invalid values rejected clearly. I exercised the extracted real PowerShell function under `process`, `Process`, `Machine`, and an invalid value; accepted/rejected behavior matched the contract, and focused Ruff/diff checks pass.

Code-wise this follow-up is sound. The PR remains labeled duplicate because #2972 is the canonical already-approved fix for #2970; only one implementation should land.

Issue #2028, opened 2026-07-11
ENABLE_TOOL_SEARCH proxy injection breaks Claude Code VSCode webview rendering


premdon009:
## Summary

When `tool_envs.claude.ENABLE_TOOL_SEARCH` is `"true"` in a Headroom deploy manifest
(`~/.headroom/deploy/<profile>/manifest.json`), the proxy injects deferred-tool /
ToolSearch payloads into Claude Code's tool-list. Claude Code's VSCode extension webview
has no renderer for the resulting `server_tool_use` / `tool_search_tool_result` content
blocks and displays "unsupported content type" instead of the normal tool-list text.

## Environment

- Headroom Doctor v0.31.0, proxy port 8787
- Deploy profile: `persistent-service`, systemd user service `headroom-default`
- Claude Code running as the VSCode native extension (not the standalone CLI)
- Manifest: `~/.headroom/deploy/default/manifest.json`

## Repro

1. Deploy Headroom with the default `persistent-service` preset targeting `claude`
   (manifest has `tool_envs.claude.ENABLE_TOOL_SEARCH: "true"`).
2. Open a repo in VSCode with the Claude Code extension routed through the Headroom
   proxy (`ANTHROPIC_BASE_URL=http://127.0.0.1:8787`).
3. Trigger a tool call that goes through Headroom's ToolSearch/deferred-tool path.
4. The webview panel shows "unsupported content type" instead of rendering the tool
   list / tool-search result.

## Workaround

Set `tool_envs.claude.ENABLE_TOOL_SEARCH` to `"false"` in the manifest and restart the
service (`systemctl --user restart headroom-default.service`). This avoids triggering the
webview bug, at the cost of losing Headroom's tool-list compression for the `claude`
target.

## Expected behavior

Either Headroom should avoid emitting content-block shapes the target client's UI can't
render, or this incompatibility should be documented so users on the VSCode extension
know to set `ENABLE_TOOL_SEARCH: "false"`.

## Related

Anthropic's `claude-code` repo has a related-looking VSCode webview rendering issue for
the same content-block types: anthropics/claude-code#73499. That issue's root cause is in
the extension's own `webview/index.js` (no case for `server_tool_use` /
`tool_search_tool_result`), so Headroom emitting those blocks more frequently via
`ENABLE_TOOL_SEARCH` makes users hit it more often — worth cross-referencing, though the
fix is presumably on Anthropic's side.

JerrettDavis:
[...]

PR #2986, opened 2026-08-13
fix(vscode): persist compatible Claude modes and route Copilot CAPI

### LINKED ISSUE #2492: [BUG] Foundry: `wrap --tool-search false` / `ENABLE_TOOL_SEARCH=false` never reach Claude Code's conversation workers — `init`-persisted `true` in settings.json wins, so `tool_search_server` still 400s
## Summary

Disabling tool deferral via the shell env (`ENABLE_TOOL_SEARCH=false`) or the CLI
(`headroom wrap claude --tool-search false`) does **not** stop Claude Code from sending the
`tool_search_server` tool. The value Claude Code's **daemon-spawned conversation workers**
actually read lives in `~/.claude/settings.json`, where `headroom init claude` bakes
`ENABLE_TOOL_SEARCH=true` — and neither the shell env nor `wrap --tool-search` ever updates
that persisted value.

On Azure AI Foundry this is fatal, because Foundry can't provide `tool_search_server`:

```
API Error: 400 tool_search_server not supported in your workspace
```

The trap is that both the `--tool-search` help and the `init` code comment say a pre-set env
value is *respected*. The user sets it, sees it apparently take effect on the launched
process (the banner may even print "on-demand tool loading DISABLED per your setting"), and
still gets the 400 — with nothing pointing at the stale settings.json value the worker is
reading.

This is the **third** distinct blocker for running Claude Code on Foundry through headroom,
after the forced-tool-search 400 and the CCR buffered-stream idle timeout (companion issues).
Each fix is silently undone by the next layer, so Foundry doesn't work out of the box.

## Environment

- **headroom**: '0.32.0`
- **Setup**: `headroom init claude` was run (persistent hooks + settings.json), then
  `headroom wrap claude`
- **Mode**: `CLAUDE_CODE_USE_FOUNDRY=1` + `ANTHROPIC_FOUNDRY_RESOURCE=<resource>`
  (Anthropic-format Foundry endpoint)
- **Client**: Claude Code `<version>` (daemon-spawned conversation workers)
- **OS / arch**: `macOS arm64 / Ubuntu x64`

## Root cause

1. **`init` persists the default into the file workers read.**
   `headroom/cli/init.py :: _ensure_claude_hooks` (~178–184):

   ```python
   env_map["ANTHROPIC_BASE_URL"] = f"http://127.0.0.1:{port}"
   # GH #746 ... Keep deferral on; respect a user-set value.
   env_map.setdefault(TOOL_SEARCH_ENV, TOOL_SEARCH_DEFAULT)   # -> "ENABLE_TOOL_SEARCH": "true"
   payload["env"] = env_map
   ```

   After `init claude`, `~/.claude/settings.json` contains `"ENABLE_TOOL_SEARCH": "true"`.

2. **`wrap --tool-search` only sets the launched *process* env — never settings.json.**
   `headroom/cli/wrap.py :: _configure_tool_search_env` (~292–303) mutates the `env` dict only
   (the two write sites, lines 294 and 302, are the *only* places `wrap` writes the key, and
   both target the process env). `wrap`'s settings.json writer
   (`_write_claude_wrap_base_url`, ~1461–1467) writes **only** the base-url key into
   `.claude/settings.local.json`; it does not touch `ENABLE_TOOL_SEARCH`.

3. **Workers read settings.json, not the parent env.**
   Per Issue #951: "daemon-spawned conversation workers read settings.json fresh rather than
   inheriting the daemon's environment." So they honor the `init`-written `true`, not the
   process env `false`.

4. **`tool_search_server` is emitted by Claude Code, not headroom** (`grep` finds the string
   nowhere in the repo). Foundry then 400s because that server tool isn't provisioned in the
   workspace.

Net: the only value that changes worker behavior lives in settings.json, and **no documented
tool-search knob updates it after `init` has written the default**.

## To reproduce

1. `headroom init claude` → `~/.claude/settings.json` now has `"ENABLE_TOOL_SEARCH": "true"`.
2. `export CLAUDE_CODE_USE_FOUNDRY=1 ANTHROPIC_FOUNDRY_RESOURCE=<r>`
3. Try to disable it: `export ENABLE_TOOL_SEARCH=false` and/or
   `headroom wrap claude --tool-search false` (banner may print "…DISABLED per your setting").
4. Send a turn → `400 tool_search_server not supported in your workspace`.
5. `grep ENABLE_TOOL_SEARCH ~/.claude/settings.json` → still `"true"`.

## Expected

Disabling tool-search through any documented knob should change what the conversation workers
send. On Foundry (`CLAUDE_CODE_USE_FOUNDRY=1`), where `tool_search_server` is unsupported,
headroom should default it off and persist that where the workers read it.

## Actual

`false` applies only to the parent process; workers keep reading `init`'s `true` from
settings.json and keep sending `tool_search_server`; Foundry 400s. The divergence is silent.

## Workaround (for anyone landing here)

Edit the file the workers actually read, then restart Claude Code:

```bash
# find every copy (user, project, and project-local override):
grep -rn ENABLE_TOOL_SEARCH ~/.claude/settings.json ./.claude/settings*.json 2>/dev/null

# set in the "env" block of the effective settings file:
#   "ENABLE_TOOL_SEARCH": "false"

pkill -f claude    # workers re-read settings.json only after a restart
```

## Proposed fix (make Foundry first-class)

1. **Reconcile settings.json from `wrap --tool-search <mode>`.** Write the resolved
   `ENABLE_TOOL_SEARCH` into the effective settings file the workers read (the same place
   `wrap` already writes the base-url), so a `false` actually reaches them. Today it writes
   only the base-url.
2. **Make `init` and `wrap` Foundry-aware.** When `CLAUDE_CODE_USE_FOUNDRY=1` (or the upstream
   resolves to a Foundry `/anthropic` endpoint), `setdefault` `ENABLE_TOOL_SEARCH=false` —
   Foundry can't run `tool_search_server` regardless of #746.
3. **Warn on conflict in `doctor`/banner.** If the resolved process value differs from the
   effective settings.json value the workers will use, say so loudly and name the file to edit,
   instead of failing silently with a confusing upstream 400.
4. **Fix the help/comment.** "A pre-set `ENABLE_TOOL_SEARCH` env var is respected" is true only
   for the launched parent process; note that daemon-spawned conversation workers take the value
   from settings.json (#951), so changing worker behavior requires updating settings.json (via
   `init`, or the reconciled `wrap` above).

## Related

- **#746** — why `ENABLE_TOOL_SEARCH` is forced on for custom base URLs.
- **#951** — daemon workers read settings.json fresh (the mechanism that makes the env not
  reach the workers).
- **#1779** — banner accuracy for the tool-search state.
- **#1076 / #1133** — Foundry support (URL / auth wiring).
- Companion Foundry issues: the **forced-tool-search 400** and the **CCR buffered-stream
  "no chunks received" idle timeout**. This is the third in that set — fixing all three is
  what makes `CLAUDE_CODE_USE_FOUNDRY=1 && headroom wrap claude` work end-to-end.

JerrettDavis:
## Description

Fixes #2492, #2028, and #2827.

Claude daemon workers consume project settings rather than reliably inheriting wrapper environment state, while the Claude VS Code webview cannot render deferred-tool response blocks. Separately, recent Copilot Chat versions use the whole CAPI override for generation; the legacy proxy override alone only sends model discovery through Headroom.

This PR carries both integrations through to the actual consumers instead of only changing their launch-time surface configuration.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Build / CI

## Changes Made

- Persist the resolved Claude ENABLE_TOOL_SEARCH value into project settings for daemon workers and restore it transactionally after wrap exits.
- Use compatibility-safe Foundry and Claude VS Code defaults while preserving explicit user choices.
- Configure both Copilot overrideProxyUrl and overrideCapiUrl in the reversible managed VS Code settings block.
- Route Copilot unprefixed POST /chat/completions and HTTP /responses requests through the real compression handlers.
- Keep /responses out of the Codex WebSocket aliases because Copilot and Codex use different WebSocket wire protocols.
- Extend wrap E2E assertions for both the Claude webview mode and Copilot CAPI routing.

## Testing

- [x] 127 combined Claude, Copilot, route-integration, and MCP dependency-contract tests pass.
- [x] Ruff check passes on all changed Python files.
- [x] Ruff format check passes.
- [x] Python compilation and git diff --check pass.

## Runtime Safety

Standalone Claude CLI defaults remain unchanged. Explicit Claude tool-search values retain precedence, and project settings are restored through the existing cleanup path. Copilot model/session helper endpoints continue through generic passthrough, while only validated HTTP generation paths receive explicit compression routes. Existing Codex WebSocket behavior is unchanged.

## Review Readiness

- [x] Current main and MCP v1 compatibility retained
- [x] Worker-facing Claude persistence covered
- [x] Reversible Copilot and Claude settings behavior covered
- [x] Copilot generation routes covered at registration and proxy integration layers
- [x] Ready for review

### DIFF
--- a/e2e/wrap/run.py
+++ b/e2e/wrap/run.py
@@ -849,8 +849,8 @@ def verify_vscode_claude_wrap(base_env: dict[str, str], project_dir: Path) -> No
             "VS Code Claude wrap should configure the project-scoped Anthropic URL",
         )
         assert_true(
-            configured["env"]["ENABLE_TOOL_SEARCH"] == "true",
-            "VS Code Claude wrap should retain Claude Code tool deferral",
+            configured["env"]["ENABLE_TOOL_SEARCH"] == "false",
+            "VS Code Claude wrap should disable tool deferral for webview compatibility",
         )
         assert_true(configured["env"]["KEEP"] == "yes", "Existing Claude env must remain")
         assert_true(str(settings_path) in output, "Wrap output should identify Claude settings")
--- a/e2e/wrap/run.py
+++ b/e2e/wrap/run.py
@@ -791,6 +791,11 @@ def verify_vscode_wrap(base_env: dict[str, str], project_dir: Path) -> None:
             "VS Code wrap should configure the project-scoped proxy URL",
         )
         assert_true(
+            f'"github.copilot.advanced.debug.overrideCapiUrl": '
+            f'"http://127.0.0.1:{port}{project_prefix}"' in configured,
+            "VS Code wrap should route Copilot Chat generation through Headroom",
+        )
+        assert_true(
             '"github.copilot.advanced.debug.overrideAuthType": "token"' in configured,
             "VS Code wrap should configure token auth",
         )
--- a/headroom/cli/init.py
+++ b/headroom/cli/init.py
@@ -41,6 +41,7 @@ from headroom.install.runtime import (
 from headroom.install.state import ManifestError, load_manifest, save_manifest
 from headroom.install.supervisors import start_supervisor
 from headroom.providers.claude import TOOL_SEARCH_DEFAULT, TOOL_SEARCH_ENV
+from headroom.providers.claude.runtime import TOOL_SEARCH_FOUNDRY_DEFAULT
 from headroom.providers.codex.install import codex_uses_chatgpt_auth
 from headroom.providers.codex.threads import retag_to_headroom
 
@@ -181,7 +182,12 @@ def _ensure_claude_hooks(path: Path, profile: str, port: int) -> None:
     # all into its context window — overflowing it (breaks sub-agent spawns,
     # forces constant compaction). Keep deferral on; respect a user-set value.
     # Shares the TOOL_SEARCH_* constants with `wrap` and `install`.
-    env_map.setdefault(TOOL_SEARCH_ENV, TOOL_SEARCH_DEFAULT)
+    tool_search_default = (
+        TOOL_SEARCH_FOUNDRY_DEFAULT
+        if os.environ.get("CLAUDE_CODE_USE_FOUNDRY")
+        else TOOL_SEARCH_DEFAULT
+    )
+    env_map.setdefault(TOOL_SEARCH_ENV, tool_search_default)
     payload["env"] = env_map
 
     hooks = dict(payload.get("hooks") or {}) if isinstance(payload.get("hooks"), dict) else {}
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -5477,8 +5477,8 @@ def vscode_copilot(
 ) -> None:
     """Run Headroom for GitHub Copilot inside Visual Studio Code.
 
-    Transparently overrides Copilot's proxy endpoint, preserving the model
-    selected in VS Code. It does not edit Codex settings.
+    Transparently overrides Copilot's proxy and CAPI endpoints, preserving the
+    model selected in VS Code. It does not edit Codex settings.
     """
     resolution = _require_copilot_subscription_resolution()
     target_settings = settings_file or vscode_settings_path()
@@ -5497,6 +5497,9 @@ def vscode_copilot(
         click.echo("  Add these user settings to VS Code:")
         click.echo(
             f'  "github.copilot.advanced.debug.overrideProxyUrl": "{vscode_proxy_url(actual_port, _project_name_from_cwd())}",'
+        )
+        click.echo(
+            f'  "github.copilot.advanced.debug.overrideCapiUrl": "{vscode_proxy_url(actual_port, _project_name_from_cwd())}",'
         )
         click.echo('  "github.copilot.advanced.debug.overrideAuthType": "token"')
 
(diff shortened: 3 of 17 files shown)


JerrettDavis:
CI correction: docker-wrap-e2e still asserted the old VS Code value ENABLE_TOOL_SEARCH=true. The product change intentionally sets false only for the VS Code webview because that surface cannot render deferred-tool blocks; standalone Claude CLI remains default-on. Updated the end-to-end contract accordingly. Focused VS Code configuration/wrap plus inherited MCP contract: 21 passed; Ruff, format, and py_compile clean.

JerrettDavis:
Completed the #2827 path rather than stopping at the VS Code setting. Commit 99b11499 now manages both overrideProxyUrl and overrideCapiUrl, explicitly sends Copilot unprefixed POST /chat/completions and HTTP /responses through the compression handlers, and deliberately leaves /responses out of the Codex WebSocket aliases because the wire protocols differ. The combined Claude/Copilot/route/MCP regression set is 127 passed; Ruff, formatting, py_compile, and diff checks are clean.

PR #2988, opened 2026-08-13
fix(proxy/openai): propagate provider usage on the Responses WS->HTTP fallback

### LINKED ISSUE #2957: [BUG] OpenAI Responses WebSocket fallback does not propagate provider input token usage
## Description

When Codex uses the OpenAI Responses WebSocket endpoint through Headroom, the proxy can report zero or otherwise incorrect input tokens in the dashboard even though the request contains a large context and the upstream response contains usage data.

This can also produce invalid savings percentages, for example:

```text
INPUT: 0
OUTPUT: 246
SAVED: 33233
SAVINGS: 33233%
```

## To Reproduce

1. Install Headroom `0.34.0`.
2. Run Headroom as a proxy for Codex/OpenCode using the `/v1/responses` endpoint.
3. Use the `anyllm` proxy backend with `cache` mode.
4. Start a Codex session with a sufficiently large context.
5. Let Codex connect through the Responses WebSocket endpoint while the upstream Responses WebSocket connection is rejected and Headroom falls back to HTTPS POST/SSE.
6. Inspect the Headroom request log or dashboard.
7. Observe that the recorded input token count is zero or otherwise much smaller than the actual request context.

## Expected Behavior

When provider usage is available, Headroom should use the provider-reported usage as the authoritative wire-token count.

In particular:

* `input_tokens_optimized` should reflect the actual input tokens sent to the provider.
* `input_tokens_original` should be calculated consistently with the compression savings.
* `tokens_saved` should not exceed `input_tokens_original`.
* `savings_percent` should remain within a meaningful range.
* Usage from the SSE `response.completed` event should be propagated into the WebSocket session and request totals.
* The dashboard should not show `INPUT: 0` for a request with non-zero provider input usage.

## Actual Behavior

Before applying any local patch, the dashboard request log contained entries similar to:

```json
{
  "input_tokens_original": 3,
  "input_tokens_optimized": 0,
  "output_tokens": 246,
  "tokens_saved": 31052,
  "savings_percent": 33233.33333333333,
  "cache_write_tokens": 31052,
  "uncached_input_tokens": 31052
}
```

The request was not a three-token request. It contained a large Codex context, and the upstream provider response also reported substantial input usage.

However, the local request statistics used a very small local token count for `input_tokens_original` and `input_tokens_optimized`. The resulting calculation produced:

* `input_tokens_optimized = 0`;
* `tokens_saved > input_tokens_original`;
* savings percentages far above 100%;
* dashboard rows showing more saved tokens than input tokens.

## Code Sample

```python
# Minimal reproduction:
# Run Headroom as a proxy for a Codex/OpenCode Responses request
# using the /v1/responses endpoint and cache mode.

# The issue occurs in the Responses WebSocket -> HTTP/SSE
# fallback and token-accounting path.
```

## Error Output

```text
INPUT: 0
OUTPUT: 246
SAVED: 33233
SAVINGS: 33233%
```

Example request statistics:

```json
{
  "input_tokens_original": 3,
  "input_tokens_optimized": 0,
  "output_tokens": 246,
  "tokens_saved": 31052,
  "savings_percent": 33233.33333333333,
  "cache_write_tokens": 31052,
  "uncached_input_tokens": 31052
}
```

The upstream Responses WebSocket connection also returns:

```text
426 Upgrade Required
```

before Headroom falls back to HTTPS POST/SSE.

## Environment

* **Headroom version**: `0.34.0`
* **Python version**: `3.14`
* **OS**: macOS
* **Client**: Codex/OpenCode
* **LLM Provider**: OpenAI-compatible HTTPS Responses endpoint
* **Proxy backend**: `anyllm`
* **Proxy mode**: `cache`
* **Endpoint**: `/v1/responses`

## Additional Context

The affected transport path is:

1. Codex connects to Headroom using the Responses WebSocket endpoint.
2. Headroom attempts to connect to the upstream Responses WebSocket endpoint.
3. The upstream returns HTTP `426 Upgrade Required`.
4. Headroom falls back to HTTPS `POST /v1/responses?stream=true`.
5. The upstream returns SSE events, including `response.completed` with usage information.
6. The usage from the HTTP/SSE fallback is not consistently propagated into the WebSocket session and request statistics.
7. The dashboard therefore reports zero or otherwise incorrect input-token values.

The issue appears to be in the Responses WebSocket fallback and token-accounting logic, rather than TCP Fast Open.

### Local diagnostic patch

To investigate the issue, I applied a local patch to `headroom/proxy/handlers/openai.py` that:

1. Added the environment variable:

   ```bash
   HEADROOM_OPENAI_RESPONSES_UPSTREAM_TRANSPORT=http
   ```

   to force HTTPS transport instead of attempting the upstream WebSocket connection first.

2. Changed `_ws_http_fallback()` to return usage values extracted from the SSE `response.completed` event, including:

   ```text
   input_tokens
   output_tokens
   cache_read_tokens
   cache_write_tokens
   uncached_input_tokens
   ```

3. Added those values to the existing WebSocket session totals.

4. Changed HTTP Responses request accounting to use the provider-reported input token count instead of the small locally converted-message token count.

The environment variable above was introduced by this local patch and is **not available in the unpatched `headroom-ai==0.34.0` release**.

The patch makes the accounting behave correctly, which indicates that the problem is in the Responses transport fallback and usage propagation rather than in the token-saving calculation itself.

abhay-codes07:
## Description

When Codex uses the OpenAI Responses WebSocket endpoint through Headroom and the upstream WebSocket is rejected, Headroom falls back to HTTPS POST/SSE. On that fallback the dashboard reported zero or tiny input tokens for a large request, and invalid savings:

```json
{ "input_tokens_original": 3, "input_tokens_optimized": 0,
  "output_tokens": 246, "tokens_saved": 31052, "savings_percent": 33233.33 }
```

## Root cause

`_ws_http_fallback` (openai.py) relays the SSE `data:` events to the client but never parses the terminal `response.completed` event for usage. The non-fallback WS path accumulates `_extract_responses_usage(event)` into the session totals on every `response.completed` frame (openai.py ~8182); the fallback path did not. So `ws_input_tokens_total` stayed at the small local count, and the session-end RequestLog computed `optimized_tokens = residual_input_tokens = 0`, leaving `tokens_saved > input_tokens_original` and `savings_percent` far above 100%.

## Fix

`_ws_http_fallback` now parses each relayed `response.completed` line with the existing `_extract_responses_usage` and returns the accumulated `(input, output, cache_read, cache_write, uncached)` provider usage. The caller folds it into the WS session totals, so the session-end outcome uses the authoritative provider wire-token count -- bringing the fallback to parity with the non-fallback WS path. SSE relay behaviour is otherwise unchanged.

Fixes #2957

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/handlers/openai.py` (`_ws_http_fallback`): accumulate usage from `response.completed` SSE lines (both the main relay loop and the buffer flush) and return the `(input, output, cache_read, cache_write, uncached)` tuple from every exit path; the WS handler caller adds it to `ws_input_tokens_total` / `ws_output_tokens_total` / cache / uncached totals before the session-end RequestLog.
- `tests/test_ws_http_fallback.py`: the fallback returns the provider usage from a `response.completed` event (input/output/cache_read/uncached), and returns all-zeros when no completed event arrives.

## Testing

- [x] New tests added

### Test Output

```text
tests/test_ws_http_fallback.py  13 passed  (11 existing + 2 new)
# uvx mypy@1.20.2 headroom/proxy/handlers/openai.py -> Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: drove `_ws_http_fallback` with the existing WS/stream mocks, feeding an SSE `response.completed` carrying `usage.input_tokens=31055`, `output_tokens=246`, `input_tokens_details.cached_tokens=20000`. The method now returns `(31055, 246, 20000, ..., 11055)`; a stream with no completed event returns all zeros. The existing 11 relay/routing/retry tests are unchanged (they ignore the new return value).
- Observed result: the fallback surfaces the provider's real input usage, so the WS session-end outcome records the actual input tokens instead of 0, and savings percentages stay within a meaningful range.
- Not tested: a live Codex WS session that triggers the upstream-WS rejection and HTTP fallback end to end (needs a real upstream refusing the WS). The usage-propagation contract is verified at the fallback boundary with the same mocks the existing fallback tests use.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. The OpenAI Responses WS-to-HTTP fallback is always-on transport behavior, not rollout-channel-gated.
- Stable/default behavior changed: yes, as a bug fix. On the WS-to-HTTP fallback the session-end outcome now records the provider's real input/output/cache usage from `response.completed` instead of leaving `ws_input_tokens_total` at 0 (which produced >100% savings). SSE relay to the client is unchanged.
- Kill switch / disable path: N/A. This corrects accounting only; there is no behavioral toggle and no user-facing surface beyond the recorded outcome numbers.
- Qualification impact: fallback-path token accounting now matches the non-fallback WS path and the HTTP Responses path (all three use `_extract_responses_usage`); savings percentages return to a valid range.
- Rollback path: revert this PR; the fallback returns to reporting zero input usage on this path.

## Review Readiness

## Checklist

## Additional Notes

The fix reuses the already-present `_extract_responses_usage` (same parser the non-fallback WS path and HTTP Responses path use), so cache-read/write and uncached accounting stay consistent across all three transports.


JerrettDavis (approved):
Reviewed the latest head after updating it to current `main`. The fallback now extracts usage at the same provider-event boundary as the native WS path and folds it into the existing session residual totals exactly once. I checked fragmented-buffer flush behavior, early client-send exits, HTTP error exits, and the no-completion case; each returns the usage accumulated before termination without changing the relayed payload.

The exact-head fallback-focused suite passes locally (14/14). No blocking correctness findings.

PR #2992, opened 2026-08-13
fix(wrap): scope Serena's MCP registration to the wrapped project (#2787)

### LINKED ISSUE #2787: wrap claude registers Serena at user scope, so one broken MCP server degrades every Claude Code session on the machine
Split out of #2783, which reported a `serena_config.yml` crash. That crash is fixed on `main` by #2676, but the reporter identified a second, independent problem that the fix does not touch: **why a single bad MCP server was able to affect every project on the machine, not just the wrapped one.**

## Current behavior

`headroom wrap claude` registers Serena at **user scope**, deliberately:

```python
# headroom/mcp_registry/claude.py:128
cmd = [str(self._claude_cli), "mcp", "add", spec.name, "-s", "user"]
```

The file fallback does the same thing, writing into the top-level `mcpServers` map of `~/.claude.json`:

```python
# headroom/mcp_registry/claude.py:168
servers = config.get("mcpServers")
...
servers[spec.name] = _spec_to_entry(spec)
```

So the entry sits alongside a user's pre-existing global servers and loads for **every** Claude Code session in every directory — including sessions in projects that were never wrapped, and sessions started long after the wrapped one exited. `headroom unwrap claude` removes it, but nothing scopes it in the meantime.

## Why it matters

User scope is a reasonable default in itself — it is what makes Serena available without re-wrapping every project. The problem is the blast radius when the registered server cannot start. From #2783:

> Since the broken Serena MCP registration was global, every subsequent `claude` launch (in any directory, any project) was affected — this wasn't a "the wrapped session is misconfigured" problem, it was "Claude Code itself no longer starts cleanly anywhere on this machine."

That reporter's only recovery path was `headroom unwrap claude`. A user who does not connect the breakage to Headroom has no obvious way back — and the failure appears in projects that have nothing to do with Headroom.

#2676 removed *the* known trigger. It does not remove the amplifier: the next Serena-side breakage (an upstream schema change, a `uvx` resolution failure, a network outage during `uvx --from git+...`, a required key we do not know about yet) lands machine-wide again.

## Options

1. **Project scope by default** (`-s local`, i.e. `projects[cwd].mcpServers` in `~/.claude.json`) — matches "I wrapped *this* project", contains any failure to it. Cost: Serena is re-registered per project, and users lose the ambient availability that user scope gives them today. Possibly behind a flag (`--code-memory-scope user|project`) with project as the default.
2. **Verify before registering** — do a real JSON-RPC `initialize` handshake against the exact spec `build_serena_spec()` produces (`headroom/mcp_registry/install.py:45`) and refuse to write the entry if it fails, printing the reason. #2676's PR already contains such a probe as a manual verification tool; it would need a short timeout and an offline-safe skip to run at wrap time. Cost: adds latency to `wrap`, and needs care not to fail wrap on a slow network.
3. **Registration ledger already exists** — `~/.headroom/mcp_installs.json` records what we installed. A `headroom doctor` check could detect "we registered a server that no longer starts" and offer to remove it, which helps the user who never suspects Headroom.

These are not exclusive; (2) is the cheapest guard, (1) is the actual containment, (3) is the recovery path.

## Not doing

Nothing here changes whether Serena is registered by default — that is a separate discussion (see #716, #568) and the `--code-memory none` opt-out already exists.

Parideboy:
## Description

Fixes #2787.

Scopes Headroom-installed Serena to the wrapped Claude project by default, containing an upstream MCP failure to that project instead of degrading every Claude Code session on the machine. `--code-memory-scope user` / `HEADROOM_CODE_MEMORY_SCOPE=user` retains the prior machine-wide behavior when explicitly requested.

The implementation completes ownership semantics across multiple wrapped projects:

- Claude MCP registration, lookup, overwrite, and removal are scope-aware.
- Legacy ledger-owned user entries migrate safely; user-managed global Serena remains untouched.
- Each project gets an independent ledger ownership key, so wrapping project B no longer overwrites project A's ownership.
- Unwrapping project A removes only A plus an independently-owned legacy user entry; it cannot orphan B or delete an unowned global entry.
- The migration that retires a legacy machine-wide entry now clears the ledger ownership record along with the config entry (see Changes Made) — an ownership record can no longer outlive the entry it authorized.

## Type of Change

## Changes Made

- `headroom/mcp_registry/claude.py`: `ClaudeRegistrar` gained a `scope` argument and reads/writes both `local` (project) and `user` (machine-wide) Claude Code MCP scopes.
- `headroom/cli/wrap.py`: `_setup_serena_mcp` registers Serena at `local` scope by default; `--code-memory-scope` (env `HEADROOM_CODE_MEMORY_SCOPE`) restores `user` scope on request.
- `headroom/cli/wrap.py`: `_retire_user_scope_serena` migrates a pre-#2787 machine-wide Serena entry on the next wrap, but only when the ledger proves Headroom installed it.
- `headroom/cli/wrap.py`: fixed `_retire_user_scope_serena` to clear the ledger's ownership record after removing the migrated entry — previously the record survived the removal, so a user later installing the identical Serena command globally themselves could collide with the stale fingerprint and have their own entry deleted by a later wrap.
- `headroom/cli/wrap.py`: `_remove_headroom_installed_serena_mcp` now also clears a scope's ownership record when that scope's live entry is already absent, so a record can't outlive the entry it authorized.
- `headroom/mcp_registry/ledger.py`: ledger entries key off a project-scoped `ownership_key` in addition to the legacy global key, so multiple wrapped projects don't clobber each other's ownership records.
- Added/extended tests: `tests/test_cli/test_serena_scope.py`, `tests/test_mcp_registry/test_claude_registrar.py`, `tests/test_cli/test_unwrap_claude.py`.

## Testing

### Test Output

```text
pytest tests/test_cli/test_serena_scope.py -v
13 passed in 19.29s

pytest tests/test_mcp_registry/ tests/test_cli/test_unwrap_claude.py -q
137 passed, 2 failed in 2.16s
```

Both failures are pre-existing and unrelated to this change — they reproduce identically with this PR's diff to `headroom/cli/wrap.py` reverted, on a clean checkout of this branch. `test_unwrap_claude_stops_claude_owned_persistent_deployment` was already confirmed pre-existing on `main` by an independent reviewer on this PR (comment from PrakharSrivastav, 2026-08-13). `test_unwrap_claude_removes_mcp_purges_retired_hook_and_stops_proxy` fails locally on Windows over a hook-config path-comparison mismatch unrelated to Serena scoping; it is not exercised by this diff.

```text
ruff check .
All checks passed!

ruff format --check .
All files already formatted.

mypy headroom --ignore-missing-imports
Success: no issues found in 522 source files
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.11, project `.venv` (pytest 9.0.3, mypy 2.3.1, ruff).
- Exact command / steps: `pytest tests/test_cli/test_serena_scope.py -v`, then re-ran the two new regression tests (`test_rewrap_migration_clears_the_ledger_so_a_reinstall_is_never_deleted`, `test_remove_serena_clears_a_stale_ownership_record_with_no_live_entry`) with the `headroom/cli/wrap.py` fix reverted via `git stash`, to confirm they fail on the old code and pass on the fix.
- Observed result: both new tests fail with `assert not True` (stale ledger record still present) against the pre-fix code, and pass once the `clear_install(...)` calls are added. Full `test_serena_scope.py` suite: 13/13 passed post-fix. `mypy headroom` and `ruff check .` both clean.
- Not tested: a real Claude Code installation end-to-end (these tests drive `ClaudeRegistrar` against a `tmp_path` home directory, per the existing pattern in this file); real macOS/Linux CI runners, which is what the GitHub Actions matrix below covers.

## Runtime Rollout Safety

- Rollout-managed feature(s): `headroom wrap claude` Serena MCP registration scope only.
- Minimum rollout channel: normal patch release; no staged feature flag required.
- Stable/default behavior changed: Serena now registers at Claude Code's `local` (project) scope by default instead of `user` (machine-wide) scope; a pre-existing Headroom-installed machine-wide entry is migrated (removed) on the next wrap, with its ledger record now also cleared.
- Kill switch / disable path: `--code-memory-scope user` (or `HEADROOM_CODE_MEMORY_SCOPE=user`) restores the previous machine-wide registration.
- Qualification impact: `tests/test_cli/test_serena_scope.py`, `tests/test_mcp_registry/test_claude_registrar.py`, and `tests/test_cli/test_unwrap_claude.py` must stay green (module to the two pre-existing, environment-specific failures noted under Testing).
- Rollback path: human revert of this PR; `--code-memory-scope user` provides an immediate operational fallback to the old registration scope without a code rollback.

## Review Readiness

## Checklist

- [x] I have made corresponding changes to the documentation (`README.md`, `docs/content/docs/proxy.mdx`)
- [x] I did **not** edit `CHANGELOG.md` — it is generated by release-please from my Conventional Commit PR title

## Additional Notes

Addresses the code-review finding from JerrettDavis (2026-08-14): the ledger's stale ownership record after migration is now cleared, with a regression test that reproduces the exact fingerprint-collision scenario described in that review and fails on the pre-fix code.


PrakharSrivastav:
Verified on macOS 26.4.1 (arm64), Python 3.12 via `uv sync --extra dev`, against `d2abf62` with `1aa701a` (main) as the control. I hit this bug on my own machine — Serena sits in the top-level `mcpServers` map of my `~/.claude.json` alongside four unrelated servers — so I was interested in whether the fix holds and whether the migration is safe.

### The behaviour change, main vs this branch

Same script against both checkouts: register a Serena spec through `ClaudeRegistrar` with an isolated `home_dir`, then read back where it landed.

```
main (no scope param)
  top-level mcpServers  : ['serena']
  projects              : (none)

this branch (scope=local)
  top-level mcpServers  : []
  projects[myproject].mcpServers : ['serena']
```

That is the reported defect and its fix, isolated from everything else in the wrap path.

### Round trip

Register at local scope, then `unregister_server`:

```
after register   — projects: {'myproject': ['serena']}
                   get_server: uvx
after unregister — returned: True
                   projects: {'myproject': []}
                   top-level: []
```

No leftovers, and `get_server` reads back through the new scope without needing to be told where to look.

### The migration, both halves

This is the part I most wanted to check, because getting it wrong either strands the old entry or deletes something the user owns.

A Serena the *user* registered machine-wide themselves:

```
user-installed serena survived migration: True -> /usr/local/bin/my-own-serena
```

A Serena that *Headroom* installed machine-wide, recorded in the ledger:

```
before migration — top-level: ['serena']
  Serena MCP: removed the machine-wide entry an earlier wrap installed (it now loads only in this project)
after  migration — top-level: []
```

Correct on both sides. The ledger check in `_retire_user_scope_serena` is what makes the difference, and the `getattr(registrar, "scope", None) != SCOPE_LOCAL` guard plus the `TypeError` catch mean the function is a no-op for the registrars that have no scopes rather than raising on them.

### Test suite

`tests/test_cli/test_serena_scope.py` and `tests/test_mcp_registry/test_claude_registrar.py` pass in full.

One failure in `tests/test_cli/test_unwrap_claude.py`:

```
FAILED test_unwrap_claude_stops_claude_owned_persistent_deployment
  assert "Claude is no longer durably wrapped by Headroom." in result.output
```

**This is pre-existing, not caused by this PR.** The same test fails identically on `main` at `1aa701a` (1 failed, 13 passed there; 1 failed, 14 passed on this branch — this branch adds a test to that file). Worth a separate issue; it should not block this.

### Other agents

The scope argument reaches `ClaudeRegistrar` only (`wrap.py:4950`). `OpencodeRegistrar`, `CodexRegistrar` and `GrokRegistrar` are separate classes and are untouched, which matches the docstring's claim that this no-ops for registrars without scopes.

### One question, not an objection

`--code-memory-scope` flows through `HEADROOM_CODE_MEMORY_SCOPE` via a Click eager callback so no subcommand signature changes. That is a reasonable trade for a cross-cutting flag, but it does mean the setting is process-global: two wraps in one process would see the last one set. Almost certainly irrelevant for a CLI, and I mention it only because the env var is also a documented public surface once it ships.

Nothing here contradicts the PR. The reported defect reproduces on `main`, does not reproduce on this branch, and the migration is safe in both directions.

JerrettDavis:
CI status note: this PR is not review-ready under the all-green requirement. Test shard 4 was cancelled by the repository's 30-minute job ceiling, with no test assertion result. This independently reproduces the infrastructure issue tracked by #2996. After that non-skipping CI correction is reviewed and human-merged, this PR must be refreshed onto current `main` and its complete applicable suite rerun green. No merge or auto-merge action has been taken.

Parideboy:
Heads-up on the red check, since it is not what it looks like: `test (4)` did not fail a test. The job hit its 30-minute timeout and the runner cancelled it while pytest was still running.

The same shard hangs in the same place on untouched `main`, at 1aa701ad, which is the commit this branch is based on:

- this PR: https://github.com/headroomlabs-ai/headroom/actions/runs/31741722442/job/94589499138 — output stops after `tests/test_transforms_content_detection.py` at 87%, cancelled at 30m15s
- `main` @ 1aa701ad: https://github.com/headroomlabs-ai/headroom/actions/runs/31739223871/job/94586139355 — output stops after the same file at 87%, cancelled at 30m23s

Both logs also show the same two failures in `tests/test_transforms/test_detect_fallback_1123.py`, before the hang. Nothing in this PR touches transforms or content detection, shards 1-3 finish here in about 7 minutes each, and every other check on the PR is green.

Locally I ran the suites this change affects on the current head: `tests/test_cli/test_serena_scope.py`, `tests/test_mcp_registry/`, `tests/test_cli/test_unwrap_claude.py`, `tests/test_cli/test_serena_migrate.py` and `tests/test_cli/test_serena_disable.py` give 157 passed and 2 failed, and both failures reproduce at 1aa701ad in a clean worktree with none of this branch's changes applied.

Happy to rebase and rerun once the hang on `main` is sorted out, or to help track it down if that would be useful.

JerrettDavis (changes requested):
The project-scope behavior and per-project ownership keys are substantially improved, but the migration leaves a stale authority record that can later delete a user-owned global Serena entry.

`_retire_user_scope_serena()` removes a ledger-owned user-scope entry, but it never clears that legacy `claude/serena` ledger record. `_remove_headroom_installed_serena_mcp()` also leaves the record behind whenever the current user entry is absent, because it only clears records after a matching live entry is removed. A concrete sequence is: (1) upgrade and migrate Headroom’s old global Serena; (2) the global config entry is removed but its fingerprint remains in the ledger; (3) the user later installs an identical Serena command globally themselves; (4) a later project-scoped wrap sees the stale fingerprint as proof of Headroom ownership and deletes the user’s entry. That breaks the PR’s central “user-managed global Serena remains untouched” guarantee.

Please clear the legacy user-scope ownership record immediately after successful retirement, and add a regression test that migrates the old entry, verifies the user ownership record is gone, recreates the same global spec as a user-managed entry, then confirms a subsequent wrap leaves it intact. It would also be prudent for scoped cleanup to clear a stale ownership record when its scoped entry is already absent, so ownership cannot survive the object it authorized.

Separately, the current PR body still fails the repository governance template and says CI is rerunning even though that run completed. Please restore the required sections and report the exact current-head verification when updating this fix.

Parideboy:
@JerrettDavis Fixed in c7b80ea3 — thanks for the precise repro path.

The gap was real: `_retire_user_scope_serena()` removed the migrated machine-wide Serena entry but never cleared the ledger's ownership record, so the fingerprint from the *old* Headroom install stayed valid ledger-side indefinitely. If a user later reinstalled the identical Serena command globally themselves, a future wrap's `headroom_installed_matching()` check would treat that stale record as proof of Headroom ownership and delete their entry.

Fix:
- `_retire_user_scope_serena` now calls `clear_install(registrar.name, "serena")` right after the successful `unregister_server(..., scope=SCOPE_USER)`, so the ledger record is retired along with the config entry it authorized.
- `_remove_headroom_installed_serena_mcp`'s per-scope loop is hardened the same way: when a scope's live entry is already absent, it now clears that scope's ownership record instead of leaving it to `continue` past silently — so a record can no longer outlive the entry it authorized, per your second suggestion.

New regression test `test_rewrap_migration_clears_the_ledger_so_a_reinstall_is_never_deleted` in `tests/test_cli/test_serena_scope.py` reproduces exactly the sequence you described: migrate the old global entry, assert the ledger record is gone, reinstall the identical spec as a user-managed global entry, wrap a second project, and assert the reinstalled entry survives. I verified it fails with `assert not True` against the pre-fix code (via `git stash` on just the `wrap.py` change) and passes after the fix. Added a second, narrower test (`test_remove_serena_clears_a_stale_ownership_record_with_no_live_entry`) for the absent-entry hardening in `_remove_headroom_installed_serena_mcp` directly.

Also restored the PR body to the required governance template sections and rebased onto current `main` (no conflicts). Local gates on the current head: `ruff check .` / `ruff format --check .` clean, `mypy headroom --ignore-missing-imports` — 522 files, no issues, `pytest tests/test_cli/test_serena_scope.py` 13/13 passed. The full local run also shows 2 pre-existing failures in `tests/test_cli/test_unwrap_claude.py` that reproduce identically with this branch's `wrap.py` change reverted (one of them is the same pre-existing failure PrakharSrivastav already flagged on `main` on 2026-08-13) — details are in the updated PR body under Testing.

Ready for re-review whenever you get a chance.

JerrettDavis (approved):
The requested ownership-lifecycle fix is complete on the current head. Successful legacy user-scope retirement now clears the exact legacy ledger record that authorized deletion, and scoped cleanup removes stale per-scope records when no live entry remains. The new regression reproduces the dangerous sequence end-to-end—migrate, verify ownership is gone, reinstall an identical user-managed global spec, wrap another project, and preserve that entry—plus a focused absent-entry case. The body is restored to the governance template and the current-head CI suite is green (the superseded cancelled governance run is non-operative). No remaining blocking findings.

JerrettDavis (approved):
Re-reviewed the rewritten Serena scope/ownership implementation and the current-main merge at 905b5cfb. Project and user entries have distinct stable ledger keys; migration removes only a fingerprint-matching legacy global entry and clears that authority after successful deletion; per-scope unwrap independently verifies ownership and clears stale records only when the corresponding object is absent. The regression reproduces the prior dangerous migrate/reinstall sequence and preserves the user-managed replacement. Focused validation passed 96/97 on Windows; the one failure is the pre-existing retired-hook cleanup test using a native backslash path, outside this Serena change. The feature-specific suite is clean and no blocking issue remains.

JerrettDavis (changes requested):
The Serena migration and per-project ownership corrections hold. One adjacent cleanup caller still assumes unscoped removal means global-only.

[P2] headroom/mcp_registry/claude.py:198-215 now removes local and user entries, but _remove_headroom_installed_tokensave_mcp in wrap.py still authorizes removal with the legacy global ledger. With a Headroom-owned global tokensave and an identical, separately user-managed project-local entry, the real temporary-file cleanup deletes both; the local entry has no ownership record. Please restrict legacy tokensave retirement to user scope, or apply the same per-scope ownership checks used for Serena. A regression should preserve the user-managed local entry during automatic wrap/unwrap cleanup.

Parideboy:
@JerrettDavis Fixed in c7f8b41a — thanks, the finding was correct, and it turned out to cut both ways.

Took your first option (restrict retirement to user scope) rather than mirroring Serena's per-scope loop: Headroom does not register tokensave at all any more — it was retired *before* #2787 introduced project scope, and its ledger key is still the plain pre-scope `"tokensave"` — so a local-scope ownership record for it can never exist, and the local branch of a per-scope loop would be dead code.

`_remove_headroom_installed_tokensave_mcp` now pins both the ownership check and the delete to `SCOPE_USER`, behind the same `hasattr(registrar, "ownership_key")` capability check `_remove_headroom_installed_serena_mcp` uses, so Codex/Grok/OpenCode keep the unscoped call they can accept.

That closes the case you described and a second one with the same root cause:

1. **Data loss (yours).** Headroom-owned global + identical user-managed project entry: the project entry was read first (local-first `get_server`), matched the *global* entry's fingerprint, and the unscoped delete took both.
2. **Cleanup silently blocked.** A project entry with a *different* command shadowed the machine-wide one in that same local-first read, so the fingerprint did not match and Headroom's own stale global entry was never retired.

`_disable_tokensave_mcp`'s early-out is deliberately left unscoped: a project-only user-managed entry still reports "present but user-managed — leaving it in place", which is true and non-destructive.

### Regressions

New `tests/test_cli/test_tokensave_scope.py`, driving the real helpers against a real `ClaudeRegistrar` rooted at a tmp_path — the wrap entry point (`_disable_tokensave_mcp`) for your case, and the unwrap-style user-scope registrar for the other direction:

```
$ pytest tests/test_cli/test_tokensave_scope.py -q
5 passed in 2.57s
```

Three of the five fail against the pre-fix helper, which is the point:

```
$ git stash push headroom/cli/wrap.py && pytest tests/test_cli/test_tokensave_scope.py -q
FAILED test_retirement_spares_an_identical_user_managed_project_entry
FAILED test_a_user_managed_project_entry_does_not_shadow_the_global_one
  AssertionError: assert 'not_headroom_owned' == 'removed'
FAILED test_unwrap_leaves_a_user_managed_project_entry_alone
  AssertionError: assert None == ServerSpec(name='tokensave', command='tokensave', args=('mcp',), env={})
3 failed, 2 passed in 0.56s
```

### Other checks

`ruff check .` and `ruff format --check .` clean; `mypy headroom --ignore-missing-imports` reports nothing in `wrap.py`. `pytest tests/test_cli/test_tokensave_scope.py tests/test_cli/test_serena_scope.py tests/test_wrap_code_memory.py tests/test_mcp_registry/ -q` gives 162 passed, 3 failed — the three are `test_serena_dashboard_disabled_*` in `test_wrap_code_memory.py`, which fail identically on this branch with the `wrap.py` change stashed (local environment, untouched by this commit).

Ready for re-review whenever you get a chance.

JerrettDavis (approved):
Re-reviewed the current head after its main update. The scope-aware cleanup correctly pins the Claude legacy TokenSave removal to user scope while preserving project-level configuration, and the focused regression suite passes locally (5 passed). The current head is conflict-free and all attached checks are complete with no failures. Approved.

JerrettDavis (approved):
Reviewed the main refresh and its integration with project-scoped registration. The test conflict retained both the scope-isolation/cleanup regressions and main's plugin-discovery coverage; production changes merged without conflicts. All 123 focused Serena, TokenSave, unwrap and Claude registrar tests pass; focused Ruff/format checks pass. No new blocking finding. Hosted validation is still in progress.

PR #2995, opened 2026-08-13
fix(proxy): complete buffered CCR response lifecycle

JerrettDavis:
## Summary

Completes the buffered CCR transport fix across both Anthropic Messages and OpenAI Responses. This supersedes the partial alternatives in #2959 and #2968 for review: it preserves real upstream error status/headers and also rejects malformed successful responses instead of forwarding a broken HTTP 200 stream.

- Waits for the buffered operation before committing ASGI response headers, preserving real 429/5xx status and retry headers.
- Applies the lifecycle correction to both Anthropic and OpenAI Responses paths.
- Preserves valid JSON-to-SSE resynthesis and the existing valid upstream-SSE adaptation/relay behavior.
- Converts a non-JSON, non-SSE upstream 200 into a sanitized 502 `upstream_protocol_error` rather than a malformed successful turn.
- Initializes the OpenAI parsed-response state explicitly and contains JSON decoding at the usage boundary, so malformed input reaches the protocol guard rather than the generic exception wrapper.
- Semantically rebased onto current `main`, including the MCP SDK v1 compatibility cap and newer buffered-CCR fixes already merged there.

## Verification

- `uv run pytest tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_proxy/test_openai_responses_ccr.py -q -p no:randomly`: 20 passed.
- Ruff check passes for both handlers and both test modules.
- Ruff formatting and `git diff --check` pass.
- Dependency diff against current `main` is empty; this PR does not alter the `mcp>=1.28.1,<2.0.0` cap.

## Review notes

This PR is for human review only. No auto-merge is configured. #2968 explicitly identifies #2959 as the lifecycle owner but leaves malformed-200 handling as a residual gap; this branch incorporates that missing case into the complete transport correction.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -4136,122 +4136,60 @@ class AnthropicHandlerMixin:
 
                     class _BufferedCCRResponse(Response):
                         async def __call__(self, scope, receive, send):  # noqa: ANN001
-                            await asyncio.sleep(0)
-                            loop = asyncio.get_running_loop()
-                            keepalive_deadline = loop.time() + 1.0
-                            started = False
+                            # Send nothing until the buffered operation resolves.
+                            # The previous keepalive preamble committed
+                            # `200 text/event-stream` after 1s, i.e. before the
+                            # outcome was known: any upstream reply that then
+                            # failed to become SSE (non-200, unparseable body)
+                            # reached the client as a 200 whose body carried no
+                            # `message_start`, which Claude Code reports as "API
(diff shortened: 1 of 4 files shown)


JerrettDavis:
Superseded by #2997, which preserves this complete buffered CCR lifecycle fix and consolidates it with the directly overlapping stateless/ZDR Responses continuation work. #2997 is semantically based on current main, retains the MCP SDK v1 compatibility cap, and passes the combined 117-test focused suite. No code was merged from this PR.

PR #2997, opened 2026-08-13
fix(proxy): complete stateless Responses and buffered CCR lifecycle

JerrettDavis:
## Description

Consolidates the related OpenAI Responses ZDR/stateless continuation and buffered CCR response-lifecycle corrections on current main. It preserves client storage policy, makes Headroom-owned continuations stateless across HTTP and WebSocket, and prevents buffered streaming paths from committing a false HTTP 200 before the real upstream outcome is known.

Closes #2675

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Preserves explicit and omitted Responses `store` policy instead of forcing provider storage or disabling memory tools.
- Replays normalized input, replayable outputs, encrypted reasoning content, and Headroom function outputs without `previous_response_id`.
- Applies the same stateless continuation policy to HTTP and WebSocket.
- Prevents transparent memory execution after client-visible WebSocket output.
- Delays buffered CCR ASGI status/headers until the operation resolves for Anthropic Messages and OpenAI Responses.
- Preserves real 429/5xx status and retry headers.
- Converts malformed non-JSON/non-SSE upstream 200 replies to a sanitized 502 protocol error.
- Preserves valid JSON-to-SSE synthesis and existing SSE adaptation.
- Removes unreachable task cleanup left behind after replacing the old keepalive polling loop with a direct awaited operation.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
118 passed across the changed HTTP/WS ZDR, lifecycle, and both-provider CCR suites
11526 tests collected with no collection errors
ruff check .: All checks passed
ruff format --check .: 1411 files already formatted
mypy headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py:
Success: no issues found in 2 source files
```

Exact-head CI is entirely green on `cbc2739c0c633ea477727ec3eb2f8a3862fa08f3`.

## Real Behavior Proof

- Environment: macOS arm64/Python 3.13 locally; GitHub-hosted Ubuntu matrix pending.
[...]

PR #3013, opened 2026-08-14
fix(ccr): splice private retrieval streams by event

### LINKED ISSUE #1877: CCR streaming interception never fires on the OpenAI Responses API; long Anthropic streams can leak markers via early flush
## Summary

CCR's streaming interception (`headroom/ccr/response_handler.py`) detects `headroom_retrieve` tool calls by byte-substring matching against Anthropic Messages SSE framing. Two consequences:

1. **On the OpenAI Responses API (e.g. `headroom wrap codex`), interception is structurally impossible** — the retrieve tool call and `<<ccr:...>>` markers pass through to a client that never defined the tool.
2. **On the Anthropic path, long responses can leak** — the buffer's early-flush heuristic yields a half-message to the client before CCR handling re-streams a continuation, producing interleaved/malformed SSE.

Observed in practice: a Codex agent received a raw `<<ccr:hash,html,6.6KB>>` marker in its context and attempted to "resolve" it manually. The `--no-ccr` help text already hints at the limitation (*"also right for streaming / non-MCP clients that can't resolve an injected tool"*, `headroom/cli/proxy.py:294-302`) — this issue proposes closing the gap instead of opting out.

Related: #1876 reports corruption in the same buffered re-synthesis path (extended-thinking responses) — independent evidence that the byte-level buffered mode is fragile; the proposal below would address that class of problem structurally.

## Details (file/line refs to current main)

**1. Detection is Messages-API-shaped byte matching** — `StreamingCCRBuffer` matches `b'"type":"tool_use"'` + the tool name in accumulated raw bytes (`response_handler.py:528-529`, check at `:542`). Responses API streams emit `function_call` output items (`response.output_item.added` events) — the pattern can never match, so `detected_ccr` never becomes true for Codex traffic.

**2. End-of-stream detection** — `StreamingCCRHandler.process_stream` gates on `b'"stop_reason"' in accumulated` (`:615`). That field does not exist in Responses SSE; `detection_complete` stays false and the stream passes through raw.

**3. The `provider` parameter is accepted but unused in detection** (`StreamingCCRHandler.__init__`, `:570`).

**4. Early-flush heuristic** (`:629`): once the buffer exceeds 10 KB without a match, all buffered chunks are yielded and the buffer cleared. Anthropic responses emit text blocks before `tool_use` blocks — a response with >10 KB of text followed by a retrieve call has already streamed a partial message to the client when CCR handling kicks in and re-streams a continuation (the client sees an unterminated message followed by fresh framing).

## Why this matters

The design assumes markers only travel proxy→model and the retrieve call only travels model→proxy. In streaming reality there is no enforcement of that boundary: markers reach end users through (a) model echo in visible text, (b) the unintercepted tool call, and (c) the model copying compressed content into client-side tool arguments (`write`/`bash`) — where the marker persists into real files.

## Proposal

**A. Egress marker scrubber (defense-in-depth, independent of interception).** A client-bound streaming filter with a small rolling hold-back window (markers are bounded, ≤ ~64 bytes, so the window survives chunk-boundary splits). On match:
- in tool-call arguments → substitute the original content from the CCR store (lossless restore; the hash is authoritative)
- in prose/text deltas → replace with a human-readable placeholder (e.g. `[compressed content: html, 6.6KB]`)
- TTL-expired hash → readable placeholder

This makes "no marker ever crosses the client boundary" an enforced invariant rather than an expectation of model behavior.

**B. Event-level SSE parsing per API family.** Replace byte-substring detection with real event parsing behind a small adapter interface:
- Anthropic Messages: `content_block_start` where `content_block.type == "tool_use"` and `name == "headroom_retrieve"`; stream end via `message_delta.stop_reason`.
- OpenAI Responses: `response.output_item.added` where `item.type == "function_call"` and `name == "headroom_retrieve"`; stream end via `response.completed`.

Detection becomes deterministic and the unused `provider` parameter becomes load-bearing.

**C. Event-level stream splicing** to replace the 10 KB flush heuristic: stream text deltas to the client immediately; on detecting a retrieve tool call, suppress the tool_use block events and closing stop metadata, run the retrieval continuation, and splice the continuation's content block events into the still-open stream. No duplicated framing, no unbounded buffering — and re-synthesis bugs like #1876 stop being possible because there is no re-synthesis.

Happy to contribute A and B (and C if the splicing approach is acceptable) — opening this first to agree on direction.

JerrettDavis:
## Description

CCR streaming previously depended on buffering or a 10KB early-flush heuristic. A valid private `headroom_retrieve` call after a long text prefix could therefore arrive after bytes had already escaped to the client, making safe server-side continuation structurally impossible. This change intercepts complete provider SSE events, suppresses only Headroom-private retrieval events, and splices a server-side continuation without replaying stream lifecycle events.

Closes #1877

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- Added a bounded event-level interceptor for Anthropic Messages and OpenAI Responses SSE streams.
- Streams ordinary complete events immediately, including text prefixes larger than the historical 10KB threshold.
- Suppresses private retrieval items, indexed deltas, and terminal payload references at the client boundary.
- Reconstructs the private turn and makes a non-streaming server-side continuation only after a private call is observed.
- Splices continuation events without duplicate start/created events, while adjusting indices, sequence numbers, terminal output, and combined usage.
- Preserves client-owned tools in mixed private/client turns and returns control to the client without starting a private continuation.
- Applies context-sensitive marker handling: exact originals in client tool arguments, compact descriptors in visible prose, and safe unavailable placeholders for misses.
- Retains the merged buffered implementation behind an explicit qualification fallback, and removes the unsafe 10KB early flush from the compatibility handler.
- Preserves current `main`, including #2996, #2997, #2971, #3009, and the MCP dependency cap.

## Testing

### Test Output

```text
uv run pytest -q <all CCR tests> tests/test_openai_codex_ws_lifecycle.py tests/test_sse_thinking_blocks.py tests/test_proxy_streaming_resilience.py tests/test_h2_stream_reset_retry.py
563 passed, 2 skipped

uv run pytest -q tests/test_openai_codex_routing.py tests/test_output_shaper_responses.py tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py tests/test_proxy/test_openai_responses_ccr.py
101 passed

uv run pytest -q tests/test_ccr_stream_splice.py
5 passed

pytest tests scripts/tests --splits 4 --group N --tb=short -q
CI-shaped fresh-process groups completed in roughly 75-82 seconds locally; affected groups 3 and 4 passed with 2804 and 2737 tests respectively after the config-compatibility correction.

uv run ruff format --check .
1415 files already formatted
uv run ruff check .
All checks passed
uv run mypy headroom
Success: no issues found in 522 source files
```

### Earlier main conflict resolution

The earlier `8e96acc7` refresh was based on main `1390d897`; the current reviewed head is `b59ded54e51f81593a17f9433fdac808c9db0e4e`. Conflict resolution preserves the CCR egress/interception helpers together with `main`'s Copilot authentication, buffered Accept-header handling, redeemable-marker gate, provider-specific stream terminals, and truncated-stream flush. The event-level Anthropic fixture now carries a real redeemable marker, matching the current interception precondition.

```text
uv run --extra dev pytest -q \
  tests/test_ccr_egress.py \
  tests/test_ccr_marker_resolution.py \
  tests/test_ccr_response_handler_extra.py \
  tests/test_ccr_stream_splice.py \
  tests/test_proxy/test_anthropic_streaming_ccr_retrieve.py \
  tests/test_proxy/test_openai_responses_ccr.py \
  tests/test_buffered_ccr_salvage.py \
  tests/test_ccr_buffered_stream_signed_thinking.py \
  tests/test_no_ccr_disables_response_handling.py \
  tests/test_proxy/test_anthropic_copilot_upstream_auth.py
152 passed

uv run ruff check <changed CCR/handler/test paths>
All checks passed!
```

Exact-head GitHub CI completed successfully.

## Real Behavior Proof

- Environment: macOS arm64, Python 3.13, current `main` after the #2996 native-detector hang fix.
- Exact command / steps: fed split Anthropic and Responses SSE events containing long visible prefixes, private retrieval calls, mixed client tools, marker-bearing tool arguments, and continuation responses through both the interceptor and shared HTTP streaming loop.
- Observed result: ordinary events stream immediately; private calls never reach the client; pure private calls continue once with `stream:false`; mixed client tools remain visible; lifecycle events are not duplicated; usage is combined; all local CI-shaped groups complete far below 30 minutes.
- Not tested: live paid-provider traffic with production credentials; provider wire shapes and shared HTTP integration are covered with deterministic fixtures.

## Runtime Rollout Safety

- Rollout-managed feature(s): event-level CCR streaming interception for Anthropic Messages and OpenAI Responses.
- Minimum rollout channel: normal patch release after exact-head CI is entirely green.
- Stable/default behavior changed: eligible CCR streams remain `stream:true` upstream and use event interception instead of always flipping the initial request to buffered `stream:false`.
- Kill switch / disable path: set the internal qualification switch `ccr_event_streaming` to false to use the retained buffered fallback.
- Qualification impact: all four Python shards, HTTP/WS lifecycle tests, Docker paths, static checks, and security checks must be green.
- Rollback path: disable event streaming to use the already merged buffered lifecycle; a human revert is available if the fallback itself must be restored as the default. No data migration is involved.

## Review Readiness

## Checklist

- [x] I have made corresponding changes to the documentation — inline runtime/configuration documentation; no separate user guide is required

Not applicable; proxy streaming behavior only.

## Additional Notes

Human review only. No merge or auto-merge is configured. This PR deliberately fixes the streaming invariant and removes the unsafe threshold rather than extending CI or stream timeouts.


chopratejas (commented):
Checks are green on this branch and it's mergeable, but it has no human review yet, so it isn't in today's merge batch — the batch was limited to approved + fully green.

Flagging it as queued rather than overlooked. A rebase on current `main` would be useful in the meantime; ten PRs landed today and several touch the proxy handlers.

JerrettDavis:
Maintainer readiness refresh (2026-08-20): re-reviewed the current event-interception head against current main, including provider lifecycle filtering, private-event suppression, continuation splicing, mixed client/private tool handling, usage/index adjustment, and buffered fallback. The description and rollout notes accurately match the implementation. Ruff passed on the changed runtime files; the exact-head CI matrix completed fully green today, including all four Python shards and security checks. The branch is clean and mergeable. No merge performed.

JerrettDavis:
Exact-head readiness audit for `c45e6d93`: the conflict resolution combined the reviewed CCR stream-egress changes with current `main`'s Copilot/Anthropic authentication imports, the 93-test focused integration suite passed locally, and all 22 current GitHub checks completed successfully. The branch is mergeable/CLEAN. Human review is still required; no merge or auto-merge was performed.

JerrettDavis (commented):
[P1] Renumber visible Anthropic blocks when suppressing private retrieval blocks. EventLevelCCRInterceptor currently forwards each surviving upstream index unchanged. With headroom_retrieve at index 0 followed by a client-owned tool at index 1, the emitted stream starts its first visible block at index 1. Feeding that actual output into Anthropic.messages.stream().get_final_message() raises IndexError: list index out of range, so the client never receives its tool call. The existing mixed-tools regression uses this ordering but only inspects bytes and misses the client failure. Please maintain an upstream-to-visible index mapping for block start/delta/stop events, and offset continuation indexes from the visible block count. Add a regression that consumes the mixed-tool stream through an SDK-style accumulator, including private blocks before and between visible blocks. The 95 focused CCR/egress/handler/WebSocket tests pass, but this SDK reproduction fails on the current head.

JerrettDavis (commented):
Resolved the SDK failure in this new commit: visible Anthropic blocks now receive consecutive indexes across start/delta/stop events, and continuations use the visible index range. Three SDK-consumption regressions cover private blocks before and between client tools; all 98 focused tests and Ruff checks pass. The previously reported block-index failure is resolved. New-head CI is running separately; no merge performed.

JerrettDavis (commented):
Refreshed this branch from current upstream main and resolved the additive stream-parameter conflicts, preserving both CCR provider state and conversation savings propagation. Rechecked the affected egress, Responses, Anthropic, conversation, and Codex paths: 88 focused tests pass, with Ruff lint/format clean. Remote CI is running on this head; no merge performed.

PR #2983, opened 2026-08-13
feat(wrap/claude): make the --1m fallback model configurable via HEADROOM_1M_MODEL

abhay-codes07:
## Description

The model `headroom wrap claude --1m` falls back to (when no model is otherwise selected) was a hardcoded constant `claude-opus-4-8`, with no env var or config key to override it. So it goes stale with every new Opus release, and the only workaround is pinning `ANTHROPIC_MODEL` globally -- which also changes every non-`--1m` session and overrides Claude Code's own `/model` picker. The knob the user actually wants ("what should `--1m` default to") did not exist (#2937).

## Fix

Add a `HEADROOM_1M_MODEL` env override that `_resolve_1m_model` consults for its fallback default, and bump the built-in default to `claude-opus-5` (Opus 5 has shipped):

```python
_1M_MODEL_ENV = "HEADROOM_1M_MODEL"
_DEFAULT_1M_MODEL = "claude-opus-5"

def _resolve_1m_model(current: str | None) -> str:
    fallback = (os.environ.get(_1M_MODEL_ENV) or "").strip() or _DEFAULT_1M_MODEL
    base = (current or "").strip() or fallback
    return base if base.endswith(_CONTEXT_1M_SUFFIX) else f"{base}{_CONTEXT_1M_SUFFIX}"
```

Precedence is unchanged: an explicit `ANTHROPIC_MODEL` (or a pass-through `--model`, via the existing `_apply_1m_to_claude_args`) still wins. `HEADROOM_1M_MODEL` only supplies the fallback when nothing else is selected. The `[1m]` suffixing and idempotency are unchanged.

Fixes #2937

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made
- `docs/content/docs/configuration.mdx`: document `HEADROOM_1M_MODEL` (new "Claude 1M context window" subsection covering `--1m` resolution order and `[1m]` acceptance) and register it in the Environment Variables catalog with its current default.
- `tests/test_cli/test_wrap_helpers.py`: assert the knob stays documented and the documented default tracks `_DEFAULT_1M_MODEL`, so it cannot silently drift.

[...]

Issue #2937, opened 2026-08-12
Make the `--1m` fallback model configurable (HEADROOM_1M_MODEL) — hardcoded `claude-opus-4-8` goes stale each Opus release


HydratedBattle:
## Summary

The model that `headroom wrap claude --1m` falls back to is a hardcoded constant. There is no env var or config key to override it, so it goes stale with every new Opus release and can only be worked around by pinning `ANTHROPIC_MODEL` globally — which changes non-`--1m` sessions too.

Request: add a `HEADROOM_1M_MODEL` env override (and bump the built-in default to `claude-opus-5`).

## Current behavior

In `headroom/cli/wrap.py` (v0.34.0, latest on PyPI):

```python
# wrap.py:288-290
# Only used when no model is otherwise selected (no ANTHROPIC_MODEL set). The
# current default Opus; the suffix logic preserves any model the user did set.
_DEFAULT_1M_MODEL = "claude-opus-4-8"
```

```python
# wrap.py:294-303
def _resolve_1m_model(current: str | None) -> str:
    base = (current or "").strip() or _DEFAULT_1M_MODEL
    return base if base.endswith(_CONTEXT_1M_SUFFIX) else f"{base}{_CONTEXT_1M_SUFFIX}"
```

Applied at `wrap.py:4718-4719`. Grepping the package confirms `_DEFAULT_1M_MODEL` has exactly two references — its definition and that one use. No env var, no config key, no CLI flag reads it.

Observed on 0.34.0:

```
>>> _resolve_1m_model(None)
'claude-opus-4-8[1m]'
>>> _resolve_1m_model('claude-opus-5')
'claude-opus-5[1m]'
```

## Why this is a problem

Opus 5 has shipped, but a fresh install with no `ANTHROPIC_MODEL` set still silently opts `--1m` into the previous generation. The user gets a working 1M window on the wrong model and no signal that it happened.

The only workaround is setting `ANTHROPIC_MODEL=claude-opus-5` persistently. That works — `_resolve_1m_model` correctly preserves it and appends `[1m]` — but it is a blunt instrument: it also pins the model for every `wrap claude` session that does *not* pass `--1m`, overriding Claude Code's own `/model` picker. The knob the user actually wants ("what should `--1m` default to") does not exist.

Patching the constant in site-packages works but is wiped by `uv tool upgrade`.

## Proposed

1. Read an env override, e.g.:

```python
[...]

Issue #765, opened 2026-06-09
[BUG] Compression inserts a role:"system" message at messages[0] → Anthropic 400 "messages.0: use the top-level 'system' parameter" on large contexts


quinneydavid:
**Version:** headroom-ai 0.5.19, proxy mode (`headroom proxy`), Claude Code via `ANTHROPIC_BASE_URL`, model `claude-opus-4-8`.

## Summary

On requests large enough to trigger compression, the proxy emits an upstream request whose `messages[0]` has `role: "system"`. Anthropic's Messages API rejects any `system` role inside `messages[]`, returning:

```
400 invalid_request_error: "messages.0: use the top-level 'system' parameter for the initial system prompt"
```

The request fails outright. For Claude Code this manifests as an **intermittent** hard error — intermittent only because it requires a context large enough to compress.

## Root cause (from the proxy's own `debug_400` captures)

Across 22 captured failures, every one shows:

- `compression.was_compressed == true`
- `original_messages[0].role == "user"` — the original request correctly carries its system prompt in the **top-level `system` parameter**
- `messages_sent[0].role == "system"` (post-compression) — the harness system/skills block was relocated **into** the `messages` array

So the compression pass is moving top-level `system` content into `messages[0]` as a `system`-role message instead of leaving it in the top-level `system` parameter.

## Log signature

Every occurrence looks like this (small/uncompressed requests never fail):

```
UPSTREAM_ERROR status=400 error_type=invalid_request_error
error_msg="messages.0: use the top-level 'system' parameter for the initial system prompt"
model=claude-opus-4-8 compressed=yes
transforms=[ ... , 'intelligent_cap:1']
original_tokens=287679  optimized_tokens=176965  message_count=...  stream=False
```

Observed `original_tokens` on failures ranged ~287k–512k, compressing to ~176k–193k.

## Likely-related

The same compression pass also logs:

```
headroom.transforms.tag_protector - WARNING - Tag placeholder lost during compression, appending: <system-reminder>
```

i.e. system-reminder tag handling is lossy in this path too — possibly the same relocation/reinsertion logic.

## Reproduction

1. Route Claude Code through `headroom proxy` (default optimization enabled).
2. Drive a session until the context exceeds the compression trigger (~250k+ input tokens in our captures).
3. The first request that compresses returns the 400 above.

## Workaround

[...]

PR #3015, opened 2026-08-14
fix(proxy): align signed-thinking wire accounting

### LINKED ISSUE #2990: [BUG] Signed-thinking passthrough silently discards every handler mutation, while PERF/stats report the transforms as shipped (option (b) follow-up to #2251/#2952)
## Description

This is the follow-up #2952 asked for:

> Not addressed by the above, worth a separate issue: option (b) of #2251 — a canonical re-serialization that preserves signed blocks byte-for-byte so compression and tool injection survive a thinking-bearing history.

#2953 fixed the load-bearing casualty (the buffered-CCR `stream:false` flip). The underlying behaviour is unchanged: on any turn whose history carries a signed `thinking` / `redacted_thinking` block, `select_outbound_body` (`headroom/proxy/body_forwarding.py`) forwards the client's original bytes, so **every** handler mutation is discarded — compression, tool-schema compaction, cross-turn dedup, the `role:system` relocation guard.

The reporting problem is the part that makes this hard to notice: `PERF` and `/stats` count those transforms as delivered. An operator reads savings that never reached the wire.

## To Reproduce

1. `headroom proxy --port 8787 --no-cache --lossless`
2. `ANTHROPIC_BASE_URL=http://127.0.0.1:8787 claude` on `claude-opus-5` with extended thinking on (Claude Code default)
3. Drive the session past the first assistant turn that emits a signed `thinking` block
4. Every subsequent turn logs `body_mutated=true ... source=passthrough`, while `PERF` reports non-empty `transforms=`

`--lossless` is not a precondition — it only removes the #2952 failure so the discard is visible on its own.

## Actual Behavior

Ten consecutive requests on a patched build, all discarded:

```
event=outbound_request body_bytes=1287983 body_mutated=true mutation_reasons=system_role_relocated           source=passthrough
event=outbound_request body_bytes=1290574 body_mutated=true mutation_reasons=structural_diff_vs_original     source=passthrough
event=outbound_request body_bytes=1293230 body_mutated=true mutation_reasons=structural_diff_vs_original     source=passthrough
event=outbound_request body_bytes=1297162 body_mutated=true mutation_reasons=structural_diff_vs_original     source=passthrough
event=outbound_request body_bytes=1304585 body_mutated=true mutation_reasons=structural_diff_vs_original     source=passthrough
```

`PERF` for those same requests reports transforms as applied:

```
PERF model=claude-opus-5 msgs=175 tok_before=358539 transforms=anthropic:tool_schema_compaction
PERF model=claude-opus-5 msgs=177 tok_before=359445 transforms=router:cross_turn_dedup:1
PERF model=claude-opus-5 msgs=181 tok_before=362402 transforms=router:tool_result:lossless_search
```

Over one session (136 forwarded requests): **87 `source=passthrough` vs 49 `source=canonical`**, with 79 of the passthrough ones carrying `body_mutated=true`.

The same session's `/stats` reported `tool_schema_tokens_saved: 323999` and `total_tokens_saved_all_layers: 367039`. An unknown fraction of that never shipped, because the requests it was measured on were forwarded as client bytes.

## Expected Behavior

Either:

- the mutation reaches the wire (option (b) of #2251 — byte-faithful re-serialization that preserves signed blocks), or
- the mutation is reported as discarded and excluded from the savings ledger, so `/stats` and the dashboard describe what actually shipped.

The current combination — silently discard, then report as saved — is the one outcome that misleads.

## Suggested Fix

1. **Byte-faithful re-serialization (option (b) of #2251).** Preserve signed `thinking` / `redacted_thinking` blocks verbatim while allowing edits elsewhere in the body, so the byte-lock is no longer needed for the whole request. #2251 notes a 400 even on a no-op re-encode whose only transform was `tool_search_deferral`, which suggests the signature covers `tools` as well — worth validating against the live API before designing around it.
2. **Until then, don't count discarded work.** `select_outbound_body` already knows the outbound body is locked. Surface that to the outcome funnel so `PERF` / `/stats` / `savings_ledger` exclude transforms that were computed but not forwarded, and log the discarded `mutation_reasons` at WARNING. #2953 added `outbound_locked_to_client_bytes` at the handler level, so the signal is already in hand.
3. **Consider a `doctor` check** for a high passthrough-with-mutations ratio — this is invisible in normal operation and `headroom doctor` currently reports a clean bill while every thinking-bearing turn is discarding its transforms.

## Environment

- **Headroom version**: git `main` @ `6077e5a149ee6548edaff033f2cdffffce6ea0cf` (includes #2953 / `f1c34d33`)
- **Python version**: 3.13.9
- **OS**: macOS 26.5.2 (Darwin 25.5.0)
- **LLM Provider**: Anthropic (`api.anthropic.com`), `claude-opus-5`, Claude Code CLI 2.1.231

## Related

- #2251 — canonical re-serialization corrupts signed thinking blocks (the reason the byte-lock exists)
- #2952 / #2953 — the buffered-CCR consequence, fixed
- #1876 — CCR buffered stream re-synthesis corrupting thinking responses

### LINKED ISSUE #2991: [BUG] role=system relocation guard (#765) is unconditional, so it hoists Claude Code's valid mid-conversation system messages into top-level system
## Description

The #765 wire-contract guard relocates **every** `role="system"` entry out of `messages[]` into the top-level `system` parameter. It has no position check and no model check:

```python
# headroom/proxy/helpers.py:893  (relocate_system_messages_to_top_level)
system_indices = {
    i for i, m in enumerate(messages)
    if isinstance(m, dict) and m.get("role") == _ROLE_SYSTEM
}
```

Its docstring assumes Headroom is the only possible source:

> Internal transforms / pipeline extensions can leave a stray system message in the list (e.g. a relocated harness system block during compression).

That was true when #765 was written. It isn't now — **mid-conversation system messages are a supported Anthropic API feature** on Claude Opus 5, Opus 4.8, Fable 5, and Mythos 5 (no beta header), and Claude Code emits them. Only `messages[0]` is invalid; a `role:"system"` entry that follows a user turn and is either last or followed by an assistant turn is a legitimate request that the API accepts.

So the guard hoists valid client content, which is precisely what the feature exists to avoid.

## Actual Behavior

One request from a normal Claude Code session on `claude-opus-5`:

```
WARNING [hr_...] Relocated role=system message(s) out of messages[] into the
        top-level system parameter (Anthropic wire-contract guard, issue #765)

event=cache_breakpoints in_total=3 out_total=3 in_system=2 out_system=3
  in_tools=0 out_tools=0 in_messages=1 out_messages=0
  in_msg_count=192 out_msg_count=175
  in_last_tail=0 out_last_tail=-1 dropped=false tail_grew=true
```

**17 valid mid-conversation system messages hoisted in a single request** (192 → 175), and the message-level cache breakpoint removed (`in_messages=1 → out_messages=0`, `out_last_tail=-1`).

## Why it matters

1. **It defeats the feature's purpose.** Mid-conversation system messages exist so operator instructions can be added *without* editing top-level `system`. Editing `system` changes the prefix ahead of the entire conversation, so every cached turn is re-processed uncached. Hoisting 17 of them rewrites `system` on a 192-message history — the worst case for prompt-cache retention, in a proxy whose main value is cache and token efficiency.
2. **It changes semantics.** A mid-conversation system message applies from its position forward. Hoisted to top-level, it applies to the whole conversation, including turns that preceded it.
3. **It drops a client cache breakpoint.** `in_messages=1 → out_messages=0` above.

## Currently masked — and armed by the fix for the other bug

On this build the relocation is discarded before it reaches the wire, because the same request is forwarded as client bytes:

```
event=outbound_request body_mutated=true mutation_reasons=system_role_relocated source=passthrough
```

That is the signed-thinking byte-lock from #2251 / #2952 (filed separately as the option-(b) follow-up #2952 asked for). My prefix-cache hit rate stayed at 99–100% precisely because the mutation never shipped.

The two are coupled: **fixing the passthrough discard activates this.** Whoever lands byte-faithful re-serialization will turn this behaviour on without necessarily realising it, and the symptom then is a prompt-cache collapse on long Claude Code sessions rather than an error.

## Suggested Fix

Narrow the guard to the cases Anthropic actually rejects, instead of all `role:"system"` entries:

- **`messages[0]`** — always invalid, relocate (this is #765's actual case).
- **A system entry not preceded by a `user` message** (or an `assistant` message ending in server-tool use) — invalid, relocate.
- **A system entry that is neither last nor followed by an `assistant` turn** — invalid, relocate.
- **Everything else** — valid client content, leave in place.

Optionally gate on model support (`claude-opus-5`, `claude-opus-4-8`, `claude-fable-5`, `claude-mythos-5`; **not** `claude-sonnet-5`) and relocate for models that don't accept the role, since there a 400 is the alternative.

Whichever predicate is chosen, the `mark_mutated("system_role_relocated")` + WARNING should fire only when something was actually out of contract — right now it fires on every turn of a normal Opus 5 session and reads as a defect that isn't one.

## Environment

- **Headroom version**: git `main` @ `6077e5a149ee6548edaff033f2cdffffce6ea0cf`
- **Python version**: 3.13.9
- **OS**: macOS 26.5.2 (Darwin 25.5.0)
- **LLM Provider**: Anthropic (`api.anthropic.com`), `claude-opus-5`, Claude Code CLI 2.1.231

## Related

- #765 — the guard's original motivation (Headroom's own compression inserting `role:"system"` at `messages[0]`)
- #2251 / #2952 / #2953 — the signed-thinking byte-lock currently masking this

JerrettDavis:
## Description

Signed-thinking histories force byte-faithful passthrough because re-serializing signed Anthropic blocks can invalidate their signatures. Headroom correctly forwarded the original client bytes, but continued reporting mutations, transforms, savings, response headers, and prefix state from a different body that never reached the provider. Separately, the final Anthropic guard hoisted every `role: system` message into the top-level prompt, including valid mid-conversation system sections, changing their semantics and destroying the cached prefix if that mutation ever shipped.

This coupled fix makes downstream accounting use the actual wire body whenever the signed-thinking lock discards edits, and narrows system relocation to the current Anthropic model and placement contract.

Closes #2990
Closes #2991

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- Detects signed thinking in the original request as well as the mutated body, so a transform cannot remove the block and accidentally bypass the byte lock.
- Keeps the original-body signature probe best-effort under malformed, recursive, and `MemoryError` conditions.
- Carries discarded mutation reasons through the streaming forwarder and emits the existing structured warning on HTTP streaming paths too.
- When signed passthrough wins, resets message savings, tool-schema savings, attribution ledgers, transform labels, response headers, and prefix tracking to the original client wire body.
- Adds bounded public diagnostic tags naming/counting discarded mutation reasons without exposing body content.
- Preserves valid mid-conversation system sections on currently supported Claude models and official Anthropic, Bedrock, and parsed `*.googleapis.com` routes; hostname-boundary validation rejects lookalike and userinfo URLs.
- Preserves consecutive system sections and enforces documented predecessor/successor placement rules.
- Continues relocating initial, invalidly placed, unsupported-model, and conservative third-party-gateway system messages to avoid upstream 400s.
- Includes current `main`, including #2996, #2997, #2971, #3009, #3012, and the MCP dependency cap.

## Testing

### Test Output

```text
uv run pytest -q <wire/cache/savings/system focused suite>
379 passed

uv run pytest -q tests/test_proxy/test_anthropic_recount_and_reparse_safety.py tests/test_proxy_byte_faithful_forwarding.py tests/test_proxy_handler_helpers.py
99 passed

pytest tests scripts/tests --splits 4 --group N --tb=short -q
All four CI-shaped fresh-process groups passed locally after correcting the MemoryError regression; each completed in roughly 75-83 seconds.

Post-CodeQL correction: 91 focused tests passed; all four exact-head CI-shaped shards passed in roughly 82-95 seconds.

uv run ruff format --check .
1411 files already formatted
uv run ruff check .
All checks passed
uv run mypy headroom
Success: no issues found in 520 source files
```

## Real Behavior Proof

- Environment: macOS arm64, Python 3.13, branch rebased onto current `main`.
- Exact command / steps: sent a signed-thinking request whose tool schema is measurably compacted inside the handler, captured the exact upstream bytes, wrapped the real outcome funnel, and inspected response headers, aggregate metrics, attribution tags, transforms, and prefix-tracker state. Exercised valid, consecutive, invalid, initial, supported-model, and unsupported-model system placements.
- Observed result: upstream bytes remain byte-identical to the client; discarded edits contribute zero tokens, zero tool savings, no transform header, and no attribution while the prefix tracker stores the actual wire messages. Valid mid-conversation system sections remain in place; only out-of-contract sections relocate.
- Not tested: live paid Anthropic traffic with production credentials. The placement/model contract was verified against the current official documentation and wire behavior is covered with a byte-capturing transport.

## Runtime Rollout Safety

- Rollout-managed feature(s): signed-thinking wire-truth accounting and Anthropic mid-conversation system preservation.
- Minimum rollout channel: normal patch release after exact-head CI is entirely green.
- Stable/default behavior changed: discarded mutations no longer inflate savings; supported valid system sections are no longer hoisted into the top-level prompt.
- Kill switch / disable path: no unsafe runtime override; human revert restores the previous conservative relocation/accounting behavior.
- Qualification impact: all Python shards, byte-forwarding, cache-prefix, outcome/savings, signed-thinking, Anthropic handler, static, Docker, and security checks must remain green.
- Rollback path: fix forward through a human-reviewed corrective PR; no persisted data or configuration migration is involved.

## Review Readiness

## Checklist

- [x] I have made corresponding changes to the documentation — inline wire-contract documentation; no separate guide is required

Not applicable; proxy wire behavior and accounting only.

## Additional Notes

Human review only. No merge or auto-merge is configured. Current provider contract reference: https://platform.claude.com/docs/en/build-with-claude/mid-conversation-system-messages

### DIFF
--- a/headroom/proxy/body_forwarding.py
+++ b/headroom/proxy/body_forwarding.py
@@ -81,6 +81,16 @@ def has_signed_thinking_blocks(body: dict[str, Any]) -> bool:
     return False
 
 
+def _original_body_has_signed_thinking_blocks(original_body_bytes: bytes | None) -> bool:
+    if original_body_bytes is None:
+        return False
+    try:
+        original = json.loads(original_body_bytes)
+    except (json.JSONDecodeError, UnicodeDecodeError, ValueError, MemoryError, RecursionError):
+        return False
+    return isinstance(original, dict) and has_signed_thinking_blocks(original)
+
+
 class BodyMutationTracker:
     """Records whether a request body was mutated and why."""
 
@@ -123,7 +133,10 @@ def select_outbound_body(
     upstream instead of silently claiming the edit landed.
     """
     mode = forwarder_mode if forwarder_mode is not None else get_python_forwarder_mode()
-    if original_body_bytes is not None and has_signed_thinking_blocks(body):
+    if original_body_bytes is not None and (
+        has_signed_thinking_blocks(body)
+        or _original_body_has_signed_thinking_blocks(original_body_bytes)
+    ):
         return OutboundBody(
             content=original_body_bytes,
             source="passthrough",
@@ -180,4 +193,7 @@ def outbound_body_is_client_bytes(
     Mirrors the first branch of :func:`select_outbound_body`; the forwarder mode
     is deliberately not consulted because that branch overrides it too.
     """
-    return original_body_bytes is not None and has_signed_thinking_blocks(body)
+    return original_body_bytes is not None and (
+        has_signed_thinking_blocks(body)
+        or _original_body_has_signed_thinking_blocks(original_body_bytes)
+    )
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -3012,7 +3012,19 @@ class AnthropicHandlerMixin:
             # the top-level 'system' parameter ..."), so relocate it back to the
             # top-level ``system`` parameter as the last step before forwarding.
             relocated_messages, relocated_system, system_relocated = (
-                relocate_system_messages_to_top_level(body["messages"], body.get("system"))
+                relocate_system_messages_to_top_level(
+                    body["messages"],
+                    body.get("system"),
+                    (
+                        str(model)
+                        if (
+                            not upstream_base_url
+                            or getattr(self, "anthropic_backend", None) is not None
+                            or "googleapis.com" in str(upstream_base_url)
+                        )
+                        else None
+                    ),
+                )
             )
             if system_relocated:
                 body["messages"] = relocated_messages
@@ -3392,6 +3404,44 @@ class AnthropicHandlerMixin:
                         body["tools"] = _ttl_tools
                     tools = _ttl_tools
                     body_mutation_tracker.mark_mutated("cache_control_ttl_order")
+
+                # Signed thinking locks the request to the client's original
+                # bytes. Once all mutation sites have run, make every downstream
+                # observer use that same wire body and neutralize savings from
+                # edits that will not be sent (#2990). This covers PERF, /stats,
+                # durable savings, response headers, pipeline events, and the
+                # prefix tracker rather than fixing only one reporting surface.
+                if outbound_locked_to_client_bytes and body_mutation_tracker.mutated:
+                    discarded_reasons = body_mutation_tracker.reasons
+                    try:
+                        wire_body = json.loads(original_body_bytes or b"")
+                    except (json.JSONDecodeError, UnicodeDecodeError, ValueError, RecursionError):
+                        wire_body = None
+                    if not isinstance(wire_body, dict):
+                        raise ValueError(
+                            "signed-thinking passthrough could not reconstruct its client wire body"
+                        )
+
+                    from headroom.proxy.savings_attribution import SAVINGS_ATTRIBUTION_TAG
+                    from headroom.proxy.tool_schema_savings_policy import (
+                        TOOL_SCHEMA_SAVINGS_TAGS,
+                    )
+
+                    attribution = tags.get(SAVINGS_ATTRIBUTION_TAG)
+                    if isinstance(attribution, list):
+                        attribution.clear()
+                    for savings_tag in TOOL_SCHEMA_SAVINGS_TAGS:
+                        tags.pop(savings_tag, None)
+                    tags.pop("tool_search_deferred_tools", None)
+                    tags["wire_mutations_discarded"] = len(discarded_reasons)
+                    tags["wire_mutation_reasons"] = ",".join(discarded_reasons)
+
+                    body = wire_body
+                    optimized_messages = body.get("messages", [])
+                    tools = body.get("tools")
+                    optimized_tokens = original_tokens
+                    tokens_saved = 0
+                    transforms_applied = []
 
                 log_cache_breakpoints(
                     request_id=request_id,
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -14,6 +14,7 @@ import time
 import uuid
 from datetime import datetime
 from typing import TYPE_CHECKING, Any
+from urllib.parse import urlsplit
 
 from headroom.proxy.stage_timer import StageTimer, emit_stage_timings_log
 
@@ -48,6 +49,23 @@ from headroom.proxy.model_router import estimate_input_tokens
 from headroom.proxy.outcome import RequestOutcome
 
 logger = logging.getLogger("headroom.proxy")
+
+
+def _is_googleapis_endpoint(value: object) -> bool:
+    """Return whether *value* targets Google APIs by parsed hostname.
+
+    A substring check would also trust attacker-controlled hosts such as
+    ``googleapis.com.example.test``. URL parsing plus a label-boundary suffix
+    check accepts Google API subdomains without widening the route gate.
+    """
+    raw = str(value).strip()
+    if not raw:
+        return False
+    try:
+        hostname = (urlsplit(raw).hostname or "").rstrip(".").lower()
+    except ValueError:
+        return False
+    return hostname == "googleapis.com" or hostname.endswith(".googleapis.com")
 
 
 class _AnthropicTurnHookUsage:
@@ -3020,7 +3038,7 @@ class AnthropicHandlerMixin:
                         if (
                             not upstream_base_url
                             or getattr(self, "anthropic_backend", None) is not None
-                            or "googleapis.com" in str(upstream_base_url)
+                            or _is_googleapis_endpoint(upstream_base_url)
                         )
                         else None
                     ),
--- a/headroom/proxy/handlers/streaming.py
+++ b/headroom/proxy/handlers/streaming.py
@@ -1121,18 +1121,20 @@ class StreamingMixin:
         # bytes once before entering the connection-retry loop. When a
         # transform mutated the body we re-serialize canonically; otherwise
         # we forward the original client bytes verbatim.
-        from headroom.proxy.body_forwarding import prepare_outbound_body_bytes
+        from headroom.proxy.body_forwarding import select_outbound_body
         from headroom.proxy.helpers import (
             capture_codex_wire_debug,
             codex_wire_debug_enabled,
             log_outbound_request,
         )
 
-        outbound_bytes, outbound_source = prepare_outbound_body_bytes(
+        outbound = select_outbound_body(
             body=body,
             original_body_bytes=original_body_bytes,
             body_mutated=body_mutated,
+            mutation_reasons=list(mutation_reasons or []),
         )
+        outbound_bytes, outbound_source = outbound.content, outbound.source
         outbound_headers = {**headers, "content-type": "application/json"}
         log_outbound_request(
             forwarder="streaming",
@@ -1143,6 +1145,7 @@ class StreamingMixin:
             mutation_reasons=list(mutation_reasons or []),
             request_id=request_id,
             source=outbound_source,
+            dropped_mutation_reasons=outbound.dropped_mutation_reasons,
         )
         _codex_wire_debug = (
             codex_wire_debug_enabled() and provider == "openai" and "/responses" in url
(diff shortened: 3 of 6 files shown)


chopratejas (commented):
This is the follow-up #2990 explicitly asked for, and it's the piece the High-severity #2952 fix deliberately left open — so it matters more than its size suggests.

The core move is right: `select_outbound_body` and `outbound_body_is_client_bytes` now both consult `_original_body_has_signed_thinking_blocks(original_body_bytes)`, not just the parsed `body`. That closes the case where a handler mutation strips or reshapes the thinking blocks before the check runs, which is precisely how the old code could report a transform as shipped while passthrough discarded it.

Reporting `wire_mutations_discarded` and `wire_mutation_reasons` as tags is the correct fix for the honesty problem in #2990 — PERF/stats claiming transforms landed when they didn't. Naming the reasons rather than just counting them is a good call; a bare count would tell an operator something was dropped without telling them what.

**One structural concern, and it's the reason I'm commenting rather than approving:** this PR closes two issues. #2991 (the `role=system` relocation guard being unconditional) and #2990 (signed-thinking accounting) are related by file but not by cause — one is about hoisting valid mid-conversation system messages, the other about mutation bookkeeping under passthrough. If this PR stalls or needs splitting, both bugs revert to uncovered, and #2990 is load-bearing for a High-severity area. If the two halves are separable without churn, I'd rather review and land them independently.

Specific question on the #2991 half: `relocate_system_messages_to_top_level(body.get("system"), ...)` is now conditional — what's the condition, and does it distinguish "Claude Code sent a legitimate mid-conversation system message" from "a stray system role that genuinely needs hoisting"? #2991's complaint was that the guard couldn't tell those apart. I'd like to see the predicate stated in a comment at the call site, because the next person to touch #765's guard will need it.

No reviews on this since the 14th — apologies for the wait.

chopratejas (approved):
Read the diff properly this time, ran the predicate against the real cases, and ran the full suite on a rebase onto current `main`. **Approving** — and retracting the structural objection I raised earlier.

**I was wrong to ask for a split.** Having read it, #2990 and #2991 are the same defect seen from two sides: what actually goes on the wire versus what we believe went. They touch overlapping code in `body_forwarding.py` and the Anthropic handler, and separating them would create an artificial merge conflict for no reviewer benefit. Keep it as one.

**This is more urgent than its label suggests.** The #765 relocation guard (`9fde1275`) is *unreleased* — it ships to users for the first time in 0.36.0. So #2991 is not a bug in the field, it is a regression 0.36.0 would introduce to every Claude Code user on Opus 5. There is also a cost dimension the issue doesn't mention: growing the top-level system block changes the cached prefix, so the unconditional hoist busts the Anthropic prompt cache on every affected turn. That makes this a release blocker rather than backlog.

**Verified behaviour** — the predicate against the real message shapes:

| case | model | relocates? |
|---|---|---|
| `user, SYSTEM, assistant` | opus-5 | no ✓ |
| `user, SYSTEM` (section at end) | opus-5 | no ✓ |
| `SYSTEM` first (`messages[0]`) | opus-5 | **yes** ✓ |
| `assistant, SYSTEM` (prev not user) | opus-5 | **yes** ✓ |
| `user, SYSTEM, user` (next not assistant) | opus-5 | **yes** ✓ |
| `server_tool_use, SYSTEM, assistant` | opus-5 | no ✓ |
| `user, SYSTEM, assistant` | sonnet-4-5 | **yes** ✓ |
| `model=None` (unknown upstream) | — | **yes** ✓ |

Nine of nine as specified. Full suite on the rebase: **11,100 passed**, 3 failures, all three reproducing on `main` and unrelated (`test_graceful_shutdown` ordering, `test_learn/test_integration`, `test_release_workflows` needing cargo).

**Two things I liked enough to name:**

The call site only passes `model` when the upstream is known-Anthropic — default, native backend, or Vertex — and passes `None` otherwise, so an unknown custom gateway keeps the old conservative hoisting. That is the right default for a change that can only be validated against Anthropic's contract.

`_is_googleapis_endpoint` parses the URL and checks a label boundary rather than substring-matching, with the docstring naming `googleapis.com.example.test`. That is a genuine security consideration in a helper that could easily have been a one-line `in`.

**One thing to watch, not blocking:** `supports_mid_conversation` is a hardcoded family list. That is the same shape as #2937 — a hardcoded `claude-opus-4-8` that goes stale every Opus release — and #2983 just made that one configurable for exactly this reason. When the next model ships supporting mid-conversation system sections, this list silently hoists its valid content until someone remembers to edit it. It fails safe, so it is a maintenance note rather than a defect; worth a follow-up issue pointing at how #2983 solved it.

Merging so it makes 0.36.0.

PR #3024, opened 2026-08-14
fix(proxy): stop cached responses replaying the producing turn's wire framing

### LINKED ISSUE #3019: [BUG] Empty HTTP 200 returned after CCR buffered-stream conversion (proxy never contacts upstream, logs no error)
# Empty HTTP 200 returned after CCR buffered-stream conversion (proxy never contacts upstream, logs no error)

## Description

When CCR converts a `stream:true` client request into a buffered `stream:false` upstream request so it can
handle `headroom_retrieve` server-side, the **next** request on that session aborts inside the proxy between
`outbound_headers` and `beta_header_merge`, and the client receives **HTTP 200 with an empty body**. The proxy
never opens a connection to `api.anthropic.com` for that request.

There are two defects here:

1. **The abort itself.** The request dies silently after the CCR buffered round-trip on the same session.
2. **The reporting.** The failure emits **no `ERROR` line at all** in `proxy.log`, and answers `200` rather
   than a 5xx. A failure that returns success and logs nothing took several sessions to localise.

Correlation in a single `proxy.log`: **6 CCR buffered conversions, 6 empty-200 deaths, perfectly paired in
both directions across three separate client sessions.** No conversion without a death, no death without a
conversion.

The trigger appears to be `headroom_retrieve` being **present in the tool list at all**, not any particular
prompt. Note the proxy itself injects that tool
(`event=tool_injection_decision ... decision=inject_sticky_replay tool_definition_bytes_count=464`), so
removing the `headroom` MCP server client-side does not avoid it.

## To Reproduce

1. Install headroom: `uv tool install headroom-ai` (version 0.35.0).
2. Start the proxy standalone, default transforms, CCR enabled:

   ```powershell
   headroom proxy --port 8787
   ```

3. Run a Claude Code client through it:

   ```powershell
   headroom wrap claude
   ```

   (this injects `ANTHROPIC_BASE_URL=http://127.0.0.1:8787` per-process)

4. Work a long session until the proxy logs a CCR buffered conversion:

   ```text
   CCR: stream:true request has headroom_retrieve available; using buffered stream:false upstream request for server-side retrieval handling
   ```

5. The **next** request on that session dies. The client prints the error below and the session must be
   exited.

Observed on requests that are not small: five of the six deaths landed at `msgs=35`, the sixth at `msgs=140`
with `content_length=284080`. Never on small requests.

To localise a death in `proxy.log`, grep for the signature:

```text
PERF model=\S+ msgs=\d+ tok_before=0 tok_after=0 .* transforms=none
```

That pattern combined with **no `event=outbound_request`** matched exactly the six deaths and nothing else.

## Expected Behavior

Either the request is forwarded upstream normally, or — if the proxy cannot process it — it returns a 5xx
with a body, and logs an `ERROR` line naming the failure.

## Actual Behavior

The proxy returns `status=200` in 17 ms with an empty body, having never opened an upstream connection, and
writes no `ERROR` line. The client cannot parse the response and drops the connection.

## Code Sample

No Python API involved — this is the `headroom proxy` CLI in front of a Claude Code client.

```python
# Not applicable. Reproduced via CLI:
#   headroom proxy --port 8787
#   headroom wrap claude
# The failing request is generated by Claude Code (claude-cli/2.1.232), not by user code.
```

## Error Output

Client side (Claude Code):

```text
API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request
```

A subagent in the same session died with `Agent terminated early due to an API error`.

Proxy side — the **entire** log footprint of the failing request `hr_1786680242_000147`
(2026-08-13 23:04:02, `proxy.log` lines 6067–6071), client `127.0.0.1:61084`, `user-agent: claude-cli/2.1.232`,
POST `/v1/messages?beta=true`, `model=claude-opus-5`, `msgs=140`, `content_length=284080`:

```text
event=proxy_inbound_request  ... content_length=284080
event=outbound_headers forwarder=anthropic_messages stripped_count=1
PERF msgs=140 tok_before=0 tok_after=0 tok_saved=0 opt_ms=13 total_ms=13 tok_out=0 ttfb_ms=0 transforms=none client=claude-code
STAGE_TIMINGS "compression_first_stage": null, "memory_context": null, "upstream_connect": null, "upstream_first_byte": null, "total_pre_upstream": 14.97
event=proxy_inbound_response status=200 duration_ms=17.31
```

A healthy request on the same proxy logs `outbound_headers` → `beta_header_merge` → pipeline →
`tool_injection_decision` → `cache_breakpoints` → `outbound_request`. This one stops after `outbound_headers`.
`upstream_connect: null` and `compression_first_stage: null` place the bail-out after `read_request_json` +
`deep_copy` and **before** compression — roughly 15 ms in. That is the step that reads per-session sticky
beta / sticky tool state.

The immediately preceding request `hr_1786680217_000146` (23:03:37, `total_ms=24796`) **succeeded**, and is
the CCR conversion:

```text
CCR: stream:true request has headroom_retrieve available; using buffered stream:false upstream request for server-side retrieval handling
CACHE-BUST: expected_cached=98,632 actual_read=0 tokens_lost=98,632 tokens_saved=718
CACHE-MISS-ATTRIBUTION: reason=prefix_change idle=25s ttl=300s prefix_changed=True ttl_exceeded=False
```

The whole log file contains **zero `ERROR` lines** across its entire life. The only `WARNING`s are upstream
429s, the Kompress/Magika startup notices, and `Could not parse output_tokens from SSE, estimating 28 from
1144 bytes`.

## Environment

- **Headroom version**: 0.35.0 (installed via `uv` as `headroom-ai`)
- **Python version**: 3.14.6 (system interpreter; the `uv` tool environment may pin its own)
- **LLM Provider**: Anthropic (`api.anthropic.com`, `model=claude-opus-5`)
- **Client**: Claude Code, `claude-cli/2.1.232`
- **Proxy invocation**: `headroom proxy --port 8787`, standalone, no config file under `~/.headroom`,
  default transforms

## Additional Context

**Paired line numbers**, single `proxy.log`, three separate client sessions:

| CCR buffered conversion | empty-200 death | gap (lines) |
| --- | --- | --- |
| 1838 | 1848 | +10 |
| 1918 | 1927 | +9 |
| 1946 | 1955 | +9 |
| 2235 | 2275 | +40 |
| 2554 | 2594 | +40 |
| 6027 | 6069 | +42 |

Other `CCR:` lines (`workspace unresolved`, `Proactively expanded N context(s)`,
`skipping proactive expansion append`) are common and harmless — only the `stream:true` → buffered
`stream:false` conversion pairs with a death.

**Ruled out on our side:**

- Double-bound proxy port — verified single instance,
  `17432 (launcher) → 20032 → 24620 (python, owns port 8787)`.
- Version drift — installed `headroom-ai` matches the reported 0.35.0.
- Client-side hooks — they never touch the HTTP path; a blocked hook becomes ordinary `tool_result` content
  client-side, before any API call.

**Wrong levers, for the record**, in case they help narrow the code path:

- `--no-ccr-proactive-expansion` does not help — the conversion fires on `headroom_retrieve` being
  *available*, not on expansion occurring.
- `--no-ccr` avoids it but costs lossy compression with no recovery path.
- `--lossless` looks like the correct workaround: no CCR retrieval marker is emitted, so no
  `headroom_retrieve` is needed and the buffered `stream:false` path never runs. Not yet confirmed by a long
  session at the time of writing.

**Log excerpt available.** `proxy.log` lines 5996–6106 verbatim with original line numbers, 47 KB — covers
the last three healthy requests, the CCR/cache-bust predecessor, the failure, and two minutes of following
silence. `authorization` / `x-api-key` / `sk-ant-*` redacted. Happy to attach on request.

**Note for maintainers**: `proxy.log` carries verbatim prompt and tool-call content elsewhere in the file, so
it cannot be shared wholesale.

Parideboy:
## Description

Closes #3019

A response-cache hit could hand the client an HTTP 200 that the client could not read, and nothing in the logs marked the turn as anything other than normal.

Two separate problems combine to produce the reported failure.

**The unreadable 200.** A cache entry stores the producing upstream's response headers verbatim. When the entry is replayed, the Anthropic handler removed only `content-encoding`, `content-length` and `content-type` before handing those headers to a brand-new `Response`. Anything else describing how that *other* connection framed its body rode along — most damagingly `transfer-encoding: chunked`. RFC 9112 §6.1 makes `Transfer-Encoding` override `Content-Length`, so the client is told to parse a plain JSON body as chunked frames, finds no valid chunk-size line, and reads an empty body out of a 200. Every other response-forwarding site in the Python proxy already strips that header; the two cache-hit sites were the only ones that did not.

**How a CCR turn could put a foreign response in the cache.** On the Anthropic path, `cache.get` is gated on `not stream` but `cache.set` was not, and the cache key has no `stream` component. A CCR buffered-stream conversion takes a request the client sent with `stream: true`, forces `stream: false` upstream, and — unlike every other streaming turn, which returns via `_stream_response` and never touches the cache — falls through to the store site. The stored reply was shaped by that forced flip plus CCR tool injection, and the key cannot distinguish it from an ordinary non-streaming reply, so a later non-streaming caller could be served a response built for a request it never made. This is why the reporters saw the failures pair with CCR activity and stop under `--lossless` / `--no-ccr`.

**Why it was invisible.** The cache-hit block emitted no log line at all, and the `PERF` line rendered no field for `RequestOutcome.from_response_cache`. A cache-served turn contacts no upstream, so it has no `outbound_request` line, no upstream stage timings, and all-zero token counters — byte-for-byte what a turn that died would look like. That is why `headroom doctor` reported zero failures while turns were dying.

### Scope note

The header fix also lands on the OpenAI cache-hit site, which additionally never received the `content-type` fix from #2952. The `not stream` gate is added to the OpenAI store site too, where it is currently redundant — a streaming chat request returns via `_stream_response` long before that point — purely to state the invariant, since the Anthropic handler had exactly that shape until a buffered-CCR branch began falling through to it.

Because the strip list now lives in one shared helper, the OpenAI handler's other five forwarding sites strip the three added headers as well. That is a widening, so it is worth being explicit about: each of those sites builds a fresh fixed-length `Response` (or, at `openai.py:6122`, synthesises SSE) from `response.content`, so replaying the upstream's framing there was the same latent bug, just without a cache to make it outlive the request that produced it. The precedent is already in the file — `openai.py:9865` passes `"transfer-encoding", "connection"` as extra names by hand, which is exactly the gap this PR closes centrally. That call site keeps its now-redundant arguments; removing them is a cleanup for another PR.

## Type of Change

## Changes Made

- Added `sanitize_forwarded_response_headers` to `headroom/proxy/helpers.py`, promoting the private helper that already lived in `headroom/proxy/handlers/openai.py` and extending it with the remaining wire-framing headers (`transfer-encoding`, `connection`, `keep-alive`). Matching is now case-insensitive; surviving headers keep their original casing. `openai.py`'s `_sanitize_forwarded_response_headers` is now a thin alias so its six call sites and the Anthropic handler strip an identical set.
- `headroom/proxy/handlers/anthropic.py`: the response-cache hit now sanitises through that helper (passing `content-type` as an extra name, preserving #2952) instead of three hand-rolled `pop` calls.
- `headroom/proxy/handlers/openai.py`: the response-cache hit sanitises the same way, gains the `content-type` handling it was missing, and sets `media_type="application/json"` explicitly.
- `headroom/proxy/handlers/anthropic.py`: `cache.set` is now gated on `not stream`, mirroring the read gate. `stream` still holds the client's original flag at that point — the buffered-CCR conversion flips `body["stream"]`, never the local variable.
- `headroom/proxy/handlers/openai.py`: the same `not stream` gate on its store site, as an invariant guard.
- Both cache-hit sites now log `RESPONSE-CACHE-HIT: model=… bytes=… age_s=… hits=…`, following the existing `CACHE-MISS-ATTRIBUTION` line style.
- `headroom/proxy/outcome.py`: the `PERF` line appends `cached=1` on a response-cache hit. It is appended only on a hit, so every other PERF line is byte-identical to before and existing parsers are unaffected.
- `headroom/perf/analyzer.py`: `PerfRecord.from_response_cache` reads that field, so `headroom perf` can tell a cache-served turn from a dead one. It defaults to `False`, so older logs still parse. `PERF_RECORD_FIELDS` gains the name at the end of the list, which is what `headroom perf --format csv --raw` uses as its column set; appending keeps every existing column at its current position. `--format json --raw` gains the key too.
- `tests/test_anthropic_pre_upstream_backpressure.py`: its cache-hit double was a partial hand-rolled stand-in for `CacheEntry` carrying only a body and headers, so it broke once the hit path started reading the entry's age and hit count. It now constructs a real `CacheEntry`, which is what the cache actually returns.

## Testing

### Test Output

```text
$ python -m pytest tests/test_proxy_response_cache_replay.py -q
tests\test_proxy_response_cache_replay.py .........                      [100%]
============================== 9 passed in 4.22s ==============================

# Everything that mentions PERF, the sanitiser, cache.set, PerfRecord or
# response_headers, plus the whole proxy suite.
$ python -m pytest tests/test_proxy/ tests/test_proxy_compression_headers.py \
    tests/test_agent_savings.py tests/test_anthropic_pre_upstream_backpressure.py \
    tests/test_backend_nonstreaming_cache_metrics.py tests/test_backend_streaming_cache_metrics.py \
    tests/test_ccr_buffered_stream_signed_thinking.py tests/test_cli_perf_format.py \
    tests/test_codex_ws_compression_scheduler.py tests/test_handler_outcome_tag_invariant.py \
    tests/test_openai_codex_ws_lifecycle.py tests/test_provider_codex_images.py \
    tests/test_proxy_handlers_batch.py tests/test_proxy_passthrough_transient_retry.py \
    tests/test_proxy_response_cache_replay.py tests/test_proxy_semantic_cache_key.py \
    tests/test_proxy_streaming_request_logger.py tests/test_request_outcome.py \
    tests/test_savings_tool_search_aggregation.py -q
================== 555 passed, 1 skipped in 88.60s (0:01:28) ==================

# Full suite, 16 workers. See "Real Behavior Proof" below for how every
# failure here was traced to a pre-existing failure or a parallelism flake.
$ python -m pytest tests scripts/tests -n 16 -q -p no:randomly --timeout=300
83 failed, 10493 passed, 657 skipped, 80 errors in 437.00s (0:07:17)

$ ruff check .
All checks passed!

$ ruff format --check <the 7 changed files>
7 files already formatted

Found 12 errors in 3 files (checked 520 source files)
# All 12 are pre-existing MCP-SDK/tomllib drift in release_version.py,
# ccr/mcp_server.py and memory/mcp_server.py; identical count before and
# after this change, none in the files it touches.
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.11, pytest 9.1.1, ruff 0.16.2, branch based on `upstream/main` at `2d88e31a`.
- Exact command / steps: Two experiments. (1) Revert-and-rerun: I reverted both fixes in place (dropped the three framing headers from `FRAMING_RESPONSE_HEADERS`, restored `cache.set` to `if self.cache and response.status_code == 200 and resp_json is not None:`), ran `python -m pytest tests/test_proxy_response_cache_replay.py -q`, then restored the fixes and re-ran. (2) Regression sweep: ran the full suite on this branch, then checked out `upstream/main` into a second worktree and re-ran, in that worktree, exactly the tests that failed here and not there.
- Observed result: With the fixes reverted, 5 of 9 new tests fail and reproduce both halves of the bug. `test_buffered_ccr_turn_does_not_write_the_response_cache` fails with `AssertionError: Expected mock to not have been awaited. Awaited 1 times.` — a turn the client sent as `stream: true` really does reach `cache.set` through the buffered-CCR branch. `test_cache_hit_replays_a_body_the_client_can_actually_read` fails with `AssertionError: assert 'transfer-encoding' not in {'transfer-encoding': 'chunked', 'connection': 'keep-alive', 'request-id': ..., 'content-length': '228', ...}` — the replayed 200 carries the producing turn's chunked framing alongside a fresh `content-length`, which is the exact framing conflict a client cannot parse. With the fixes restored, all 9 pass, the replayed body arrives intact as `application/json`, and the run logs both `RESPONSE-CACHE-HIT` and a `PERF … cached=1` line. The full suite on this branch gives `83 failed, 10493 passed, 657 skipped, 80 errors`; 33 of those failures were not in my baseline list, so I ran those 33 in the `upstream/main` worktree and 20 failed there identically (Windows-specific: `sqlite:///C:\…` path handling, private-directory permissions, fsync, ONNX thread caps, serena config discovery). Re-running the remaining 13 serially on this branch gave `1 failed, 25 passed` — the other 12 were xdist parallelism flakes, including all four `tests/test_proxy/test_anthropic_ccr_deferred_injection.py` tests, which are the only ones in this change's blast radius and which pass serially. The one real serial failure, `tests/test_savings_ledger_offload.py::test_concurrent_requests_all_land_their_events` (`AssertionError: a concurrent append was lost / assert 23 == 24`), fails the same way on `upstream/main` run serially. The 80 errors are dashboard-template collection errors unrelated to the proxy. Net: no failure attributable to this change.
- Not tested: I could not reproduce against live upstream traffic, so I have not confirmed which upstream in the reporters' setups emits `transfer-encoding: chunked`. Anthropic direct is HTTP/2, where the header is forbidden, but any HTTP/1.1 hop (corporate proxy, third-party gateway, local relay) reintroduces it. I have also not measured whether the `not stream` gate reduces the cache hit rate in practice; by construction it can only drop entries that were unsafe to serve. A reporter running unmodified 0.35.0 with `headroom proxy --no-cache` would confirm the cache path is the one involved, and that flag is a lighter workaround than `--lossless` or `--no-ccr` because it keeps CCR and compression enabled.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — this is a correctness fix on the always-on response-cache path (`cache_enabled` defaults to `True`).
- Stable/default behavior changed: yes, in four ways. Replayed cached responses no longer carry the producing upstream's framing headers (or `server`, on the Anthropic side). Forwarded responses on the OpenAI handler's other five sanitiser call sites no longer carry `transfer-encoding`, `connection` or `keep-alive` either, since the strip list is now shared; all five build a fixed-length response from `response.content`, so none of them could legitimately replay that framing. A turn whose client asked for `stream: true` no longer writes the response cache on the Anthropic path. `PERF` lines gain a trailing `cached=1` on a response-cache hit only; all other PERF lines are unchanged.
- Kill switch / disable path: `headroom proxy --no-cache` disables the response cache entirely and bypasses every path this PR touches.
- Qualification impact: low. No public API, config key, CLI flag or wire format changes. Two additive output changes: the `cached=1` PERF field, which `_parse_kv` already handles the same way it handles the existing trailing `client=` field, and a `from_response_cache` column appended to `headroom perf --format csv --raw` (plus the matching key in `--format json --raw`). Anything consuming that CSV positionally keeps working because the column is last; anything reading it by name is unaffected.
- Rollback path: revert this commit. It is self-contained with no migration, no persisted state and no schema change; cache entries written before or after behave identically on read.

## Review Readiness

## Checklist

## Additional Notes

Documentation is marked N/A: no user-facing surface changes, and the new `cached=` PERF field is additive and self-describing.

Relationship to nearby open PRs, since several touch adjacent code:

- **#2953** (already merged, unreleased) added the `resp_json is not None` guard at the same Anthropic store site. That stops an SSE *body* being stored; it does not stop a JSON-bodied response storing chunked framing headers, and it does not add the `stream` gate. The two changes are complementary.
- **#2959** and **#2968** both touch the buffered-CCR response path but address when and how the status is committed. Neither reaches the cache-hit replay.
- **#3013** rewrites CCR into event-level stream splicing and keeps `buffered_stream_ccr` as a fallback, so the store site this PR gates remains reachable. If #3013 lands first I am happy to rebase.

`mypy headroom --ignore-missing-imports` reports 12 pre-existing errors in `headroom/release_version.py`, `headroom/ccr/mcp_server.py` and `headroom/memory/mcp_server.py` from MCP SDK version drift in my local environment. None are in the files this PR touches, and the count is identical before and after the change.

### DIFF
--- a/headroom/perf/analyzer.py
+++ b/headroom/perf/analyzer.py
@@ -167,6 +167,11 @@ class PerfRecord:
     ttfb_ms: float = 0.0
     stages: dict[str, float] = field(default_factory=dict)
     savings_breakdown: list[dict[str, object]] = field(default_factory=list)
+    # True when the proxy answered from its own response cache and never
+    # contacted the upstream. Such a turn has all-zero token counters and no
+    # upstream stage timings, so without this flag it reads as a turn that
+    # did nothing (#3019). Absent from pre-#3019 logs, hence the default.
+    from_response_cache: bool = False
 
 
 @dataclass
@@ -373,6 +378,7 @@ def parse_log_files(last_n_hours: float = 168.0) -> PerfReport:
                                 total_ms=float(kv.get("total_ms", 0)),
                                 tokens_out=int(kv.get("tok_out", 0)),
                                 ttfb_ms=float(kv.get("ttfb_ms", 0)),
+                                from_response_cache=kv.get("cached", "0") == "1",
                                 stages=stages_by_rid.get(m.group("rid"), {}),
                             )
                         )
@@ -765,6 +771,9 @@ PERF_RECORD_FIELDS = [
     "ttfb_ms",
     "stages",
     "savings_breakdown",
+    # Appended last so every existing CSV column keeps its position; a reader
+    # that indexes by name is unaffected either way.
+    "from_response_cache",
 ]
 
 
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -36,7 +36,11 @@ from headroom.proxy.auth_mode import (
 from headroom.proxy.compression_decision import CompressionDecision
 from headroom.proxy.forwarded_headers import resolve_client_ip
 from headroom.proxy.handlers._debug_dump import _debug_dump_mode, _redact_debug_value
-from headroom.proxy.helpers import extract_tags, relocate_system_messages_to_top_level
+from headroom.proxy.helpers import (
+    extract_tags,
+    relocate_system_messages_to_top_level,
+    sanitize_forwarded_response_headers,
+)
 from headroom.proxy.image_isolation import run_image_compression_isolated
 from headroom.proxy.memory_decision import MemoryDecision
 from headroom.proxy.memory_query import MemoryQuery
@@ -1176,15 +1180,32 @@ class AnthropicHandlerMixin:
                         )
                     )
 
-                    # Remove compression headers from cached response
-                    response_headers = dict(cached.response_headers)
-                    response_headers.pop("content-encoding", None)
-                    response_headers.pop("content-length", None)
-                    # Drop the stored content-type too. Starlette lets an
-                    # explicit header win over ``media_type``, so keeping the
-                    # producing request's type would let a cache entry hand this
-                    # caller a wire format it never asked for (#2952).
-                    response_headers.pop("content-type", None)
+                    # Strip the stored response's wire-framing headers. The
+                    # entry carries whatever the *producing* upstream sent,
+                    # and replaying that framing over a different connection
+                    # breaks the body: a stale ``transfer-encoding: chunked``
+                    # makes the client parse plain JSON as chunked frames and
+                    # read nothing out of an HTTP 200 (#3019). ``content-type``
+                    # goes too, because Starlette lets an explicit header win
+                    # over ``media_type`` and the producing request's type
+                    # would hand this caller a wire format it never asked
+                    # for (#2952).
+                    response_headers = sanitize_forwarded_response_headers(
+                        cached.response_headers,
+                        "content-type",
+                    )
+
+                    # A cache hit answers the client without touching the
+                    # upstream, so it emits no outbound_request line and no
+                    # upstream stage timings. Without this log a served-from-
+                    # cache turn is indistinguishable from a turn that died
+                    # silently, which is exactly how #3019 stayed invisible.
+                    logger.info(
+                        f"[{request_id}] RESPONSE-CACHE-HIT: model={model} "
+                        f"bytes={len(cached.response_body)} "
+                        f"age_s={(datetime.now() - cached.created_at).total_seconds():.0f} "
+                        f"hits={cached.hit_count}"
+                    )
 
                     # Unit 4: release the pre-upstream semaphore on cache
                     # hit — no upstream call will happen.
@@ -3991,7 +4012,24 @@ class AnthropicHandlerMixin:
                         # the key: the cache key has no ``stream`` component, so
                         # a buffered request would be answered with a stream it
                         # cannot read (#2952).
-                        if self.cache and response.status_code == 200 and resp_json is not None:
+                        #
+                        # ``not stream`` mirrors the read gate at the cache
+                        # lookup above. ``stream`` still holds the *client's*
+                        # original flag here — the buffered-CCR conversion
+                        # flips ``body["stream"]``, never this variable — so a
+                        # turn the client asked to stream is the one case that
+                        # can reach this store site with a buffered body. That
+                        # body was shaped by a forced ``stream: false`` flip
+                        # plus CCR tool injection, and the key cannot tell it
+                        # apart from an ordinary non-streaming reply, so
+                        # storing it lets a later caller be answered with a
+                        # response built for a request it never made (#3019).
+                        if (
+                            self.cache
+                            and not stream
+                            and response.status_code == 200
+                            and resp_json is not None
+                        ):
                             await self.cache.set(
                                 cache_lookup_messages,
                                 model,
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -28,6 +28,7 @@ from headroom.proxy.helpers import (
     _headroom_bypass_enabled,
     extract_tags,
     jitter_delay_ms,
+    sanitize_forwarded_response_headers,
 )
 from headroom.proxy.loopback_guard import is_loopback_host
 from headroom.proxy.stage_timer import StageTimer, emit_stage_timings_log
@@ -327,10 +328,10 @@ def _sanitize_forwarded_response_headers(
     headers: httpx.Headers | dict[str, str],
     *extra_names: str,
 ) -> dict[str, str]:
-    cleaned = dict(headers)
-    for name in ("content-encoding", "content-length", "server", *extra_names):
-        cleaned.pop(name, None)
-    return cleaned
+    # Thin alias kept for the many call sites in this module; the policy
+    # (and the list of framing headers) lives in one place so the Anthropic
+    # handler strips exactly the same set — see #3019.
+    return sanitize_forwarded_response_headers(headers, *extra_names)
 
 
 def _resolve_openai_handler_path(
@@ -3273,10 +3274,34 @@ class OpenAIHandlerMixin:
                     )
                 )
 
-                # Remove compression headers from cached response
-                response_headers = _sanitize_forwarded_response_headers(cached.response_headers)
+                # Strip the stored response's wire-framing headers, and its
+                # content-type: the entry carries whatever the *producing*
+                # upstream sent, and replaying that framing over a different
+                # connection breaks the body — a stale
+                # ``transfer-encoding: chunked`` makes the client parse plain
+                # JSON as chunked frames and read nothing out of an HTTP 200
+                # (#3019, same reasoning as #2952 on the Anthropic twin).
+                response_headers = _sanitize_forwarded_response_headers(
+                    cached.response_headers,
+                    "content-type",
+                )
 
-                return Response(content=cached.response_body, headers=response_headers)
+                # A cache hit answers without touching the upstream, so it
+                # emits no outbound_request line and no upstream stage
+                # timings. Log it, or a served-from-cache turn looks exactly
+                # like a turn that died silently (#3019).
+                logger.info(
+                    f"[{request_id}] RESPONSE-CACHE-HIT: model={model} "
+                    f"bytes={len(cached.response_body)} "
+                    f"age_s={(datetime.now() - cached.created_at).total_seconds():.0f} "
+                    f"hits={cached.hit_count}"
+                )
+
+                return Response(
+                    content=cached.response_body,
+                    headers=response_headers,
+                    media_type="application/json",
+                )
 
         # Token counting (offloaded off the event loop — GH #1701)
         tokenizer, original_tokens = await self._count_tokens_offloaded(model, messages)
@@ -4810,7 +4835,16 @@ class OpenAIHandlerMixin:
                 # Cache response under the SAME key it was looked up by:
                 # cache_lookup_messages is the raw pre-mutation snapshot, not
                 # the live (hooked) `messages` (#327).
-                if self.cache and response.status_code == 200:
+                #
+                # ``not stream`` mirrors the read gate at the cache lookup
+                # above. It is currently redundant here — a streaming chat
+                # request returns via ``_stream_response`` well before this
+                # point — but the Anthropic handler had the same shape until a
+                # buffered-CCR branch started falling through to its store
+                # site, which let a response built for a stream:true request
+                # answer a later non-streaming caller (#3019). Stating the
+                # invariant keeps that from being reintroduced silently.
+                if self.cache and not stream and response.status_code == 200:
                     await self.cache.set(
                         cache_lookup_messages,
                         model,
--- a/headroom/proxy/helpers.py
+++ b/headroom/proxy/helpers.py
@@ -317,6 +317,40 @@ def _headroom_bypass_enabled(headers: Any) -> bool:
     return bypass or passthrough
 
 
+# Response headers that describe how the *upstream* framed its body on the
+# wire, not what the payload means. Every one of them is invalid to replay:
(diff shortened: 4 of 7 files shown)


JerrettDavis (approved):
Reviewed the latest merged head across both provider cache paths, the shared header policy, and the PERF parser/export contract. The key invariant is now symmetric: a client `stream: true` request cannot populate a cache that is only read by non-streaming requests, including the buffered-CCR fallthrough. Rebuilding the response after stripping stale content/framing metadata is also correct for decoded `response.content`, and the additive `cached=1` parser path remains backward-compatible.

The focused replay suite passes locally on the exact head (9/9), including poisoned framing, buffered CCR, cache-hit logging, and raw PERF parsing. No blocking correctness findings.

Non-blocking follow-up: if this sanitizer becomes the general hop-by-hop boundary, consider also stripping headers named by the incoming `Connection` field (plus the remaining RFC hop-by-hop names). That is broader hardening and is not required for the reported stale `transfer-encoding` failure.

PR #3048, opened 2026-08-15
fix(savings): refresh stale price map instead of pricing new models at $0

mactonicaaron:
## Problem

`litellm.model_cost` is populated once, at import. A proxy that stays up for weeks therefore prices every request against the map it fetched at startup, and a model released after that date is simply absent from it. `_estimate_compression_savings_usd` read "absent" as "free":

```python
input_cost_per_token = info.get("input_cost_per_token")
if not input_cost_per_token:
    return 0.0
```

There is no log line, no counter, and no dashboard signal. The savings figure just stops moving for whichever model you have most recently adopted — usually the newest and most expensive one.

## Reproduction

On a proxy started 2026-07-01 and still running 2026-08-14, walking consecutive checkpoints in `~/.headroom/proxy_savings.json` and grouping the deltas by model:

| model | tokens saved | $ accrued | implied $/Mtok | checkpoints |
|---|---:|---:|---:|---:|
| `claude-opus-5` | 25,400,013 | **$0.00** | **0.000** | 3,962 |
| `claude-sonnet-5` | 10,276,174 | $30.83 | 3.000 | 468 |
| `claude-opus-4-7` | 335,472 | $1.68 | 5.000 | 94 |
| `claude-fable-5` | 118,264 | $1.18 | 10.000 | 38 |
| `gpt-5.6-sol` | 43,844 | **$0.00** | **0.000** | 105 |

Every one of the 3,962 `claude-opus-5` checkpoints priced at exactly zero, while other models in the same process priced normally. `claude-opus-5` did not exist in the price map that process fetched on July 1.

A second signal in the same data: that process priced `claude-sonnet-5` at `$3.000/Mtok`, while today's map says `$2.000/Mtok`. Same code, two different maps — which is the point.

## Real behavior proof

**Setup**

- macOS 26.5.1, arm64, Python 3.13.13
- `headroom-ai` 0.25.0 installed non-editable in a venv, `litellm` 1.88.1
- Proxy: `headroom proxy --port 8787`, `HEADROOM_MODE=token`, backend `anthropic`
- Traffic: Claude Code (`claude-cli/2.1.233`) and Codex, primary model `claude-opus-5`

**Steps run after the patch**

1. Applied the change to the installed copy and restarted the proxy.
2. Let real session traffic accrue.
3. Re-read `~/.headroom/proxy_savings.json` and grouped deltas by model again.
4. Read `/stats`.

**After-fix evidence — observed result**

Splitting the same savings file at the restart boundary, nothing changed but the process:

```
BEFORE   claude-opus-5   25,400,013 tok  ->  $0.00    $0.000/Mtok   n=3,962
AFTER    claude-opus-5    1,937,548 tok  ->  $9.69    $5.000/Mtok   n=270
```

`/stats` now carries the new field, so an unpriceable model is visible rather than silent:

```json
"unpriced_savings": {"total": 0, "by_model": {}}
```

And the pricing path itself, run against the live installed module:

```
$ cd / && python -c "from headroom.proxy import savings_tracker as st; \
    print(st.__file__); \
    print(st._estimate_compression_savings_usd('claude-opus-5', 1_000_000))"
/…/site-packages/headroom/proxy/savings_tracker.py
5.0
```

**What I did not test**

- The background refresh against a genuinely stale map in a long-running process. The TTL path is unit-tested with an injected fake; I have not left a proxy up long enough for a real model release to land mid-process.
- Non-Anthropic providers beyond the `gpt-5.6-*` rows visible in the data above.
- Behaviour when `litellm` is absent entirely (the existing `LITELLM_AVAILABLE` guard is untouched).

## The fix

A miss now schedules a bounded, TTL'd refresh of `litellm.model_cost` on a background thread and records the tokens it could not price. Deliberately off the calling thread: this runs on the request-recording path, which must not block on a network fetch. A refresh failure is caught and is never fatal. `PRICE_MAP_REFRESH_INTERVAL_SECONDS` defaults to one hour, so a genuinely unpriceable model does not refetch on every request.

`/stats` gains `unpriced_savings: {total, by_model}`.

## Also included

`formatCurrency` in the dashboard rendered anything over `$1000` with one decimal, so `$1,050`–`$1,149` all displayed `"$1.1k"`. A lifetime figure growing a few dollars a day could not visibly move for weeks, which reads as "compression stopped saving anything". Below `$10k` now shows whole dollars:

```
999.994 -> $999.99   1000 -> $1,000   1112.59 -> $1,113   9999.4 -> $9,999   10000 -> $10.0k
```

Anchored on the `formatCurrency` signature so the identical line inside `formatNumber()` is untouched — `441.4M` is correct for token counts.

I kept this in the same PR because it is the same reported symptom (the savings number appears frozen) with a second contributing cause. Happy to split it out if you would rather.

## Tests

`tests/test_savings_tracker_price_map_refresh.py`, 5 cases. Each claim was mutated to red and back rather than trusted green:

- Restoring the original two-line body → 4 of 5 fail; the baseline "known model still prices at $3/Mtok" correctly stays green.
- Setting `PRICE_MAP_REFRESH_INTERVAL_SECONDS = 0.0` → only the rate-limit test fails.

Full suite on this branch matches `main` exactly: 11 failed / 86 passed across the pre-existing failures, which are environment gaps in my checkout (`cargo` not installed, ast-grep interceptors, tree-sitter) and are identical on `main`. `ruff check` and `ruff format --check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- **Savings: a long-lived proxy priced newly-released models at $0.00.**
+  `litellm.model_cost` is populated once, at import, so a proxy left up for
+  weeks prices every request against the map it fetched at startup and a model
+  released after that date is simply absent. The lookup treated absent as free
+  and returned `0.0` with no signal anywhere. A miss now schedules a bounded,
+  TTL'd refresh of the map on a background thread (the request-recording path
+  never blocks on network I/O) and records the tokens it could not price;
+  `/stats` gains `unpriced_savings` so a $0 is visible rather than silent.
+  Dashboard `formatCurrency` also widened its k-band — everything over $1000
+  rendered with one decimal, so $1,050-$1,149 all displayed "$1.1k" and a
+  lifetime figure growing a few dollars a day could not visibly move.
+
 ### Features
 
 * **proxy:** per-project savings breakdown on the dashboard for all wrapped agents — Claude Code, Codex, aider, Copilot, and Cursor ([#802](https://github.com/chopratejas/headroom/issues/802)). `headroom wrap claude`/`codex` tag requests with an `X-Headroom-Project` header (launch-directory name); `wrap aider`/`copilot`/`cursor` — whose clients cannot send custom headers — use a `/p/<name>` base-URL prefix the proxy strips. Savings are aggregated per project (persisted, schema v3 with transparent v2 migration), exposed as `savings.per_project` in `/stats` and `projects` in `/stats-history`, and shown in a Per-Project Savings dashboard table.
--- a/headroom/dashboard/templates/dashboard.html
+++ b/headroom/dashboard/templates/dashboard.html
@@ -1841,7 +1841,12 @@
 
                 formatCurrency(n) {
                     if (n < 0) return '-' + this.formatCurrency(-n);
-                    if (n >= 1000) return (n / 1000).toFixed(1) + 'k';
+                    // Below $10k show whole dollars. One decimal in the k-band
+                    // meant $1,050-$1,149 all rendered "$1.1k", so a lifetime
+                    // figure growing a few dollars a day looked frozen for
+                    // weeks and read as "compression stopped saving anything".
+                    if (n >= 10000) return (n / 1000).toFixed(1) + 'k';
+                    if (n >= 1000) return Math.round(n).toLocaleString('en-US');
                     if (n >= 1) return n.toFixed(2);
                     if (n >= 0.01) return n.toFixed(3);
                     if (n > 0) return n.toFixed(4);
--- a/headroom/proxy/savings_tracker.py
+++ b/headroom/proxy/savings_tracker.py
@@ -13,6 +13,7 @@ import logging
 import os
 import tempfile
 import threading
+import time
 from csv import DictWriter
 from datetime import datetime, timedelta, timezone
 from io import StringIO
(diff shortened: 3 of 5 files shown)


mactonicaaron:
Closing — the headline bug is already fixed on `main`, and I missed it because my clone was ~1,021 commits stale when I branched. `_estimate_compression_savings_usd` now distinguishes `None` (unknown → `DEFAULT_FALLBACK_INPUT_COST_PER_TOKEN`) from a real `0.0` (legitimately free), so an unknown model no longer silently banks $0.00. That resolves what I hit. Sorry for the noise.

Two things from the investigation that `main` does not appear to cover yet, in case either is worth an issue:

1. **The map itself is an import-time snapshot.** `litellm.model_cost` is populated once at import, so a long-lived proxy never learns a price for a model released after it started — it now bills the blended fallback forever rather than the model's real rate. Measured on a proxy up since 2026-07-01: 25,400,013 `claude-opus-5` tokens across 3,962 checkpoints priced at the wrong rate purely because opus-5 postdated the process, while `claude-sonnet-5` priced normally in the same process. Corroborator: that process priced sonnet-5 at $3.000/Mtok while the current map says $2.000/Mtok. Restarting was the only way to learn a price.

2. **Fallback-priced tokens are invisible.** There is no counter or `/stats` field distinguishing "priced at the model's real rate" from "priced at the blended fallback", so a systematically mispriced model looks identical to a correctly priced one on the dashboard.

I have a working patch for both (TTL'd background refresh on a miss, plus an `unpriced_savings` `/stats` field), but per CONTRIBUTING that reads more like a feature than a bug fix, so I'd rather open an issue first than push a PR. Let me know if you'd like it.

PR #3049, opened 2026-08-15
fix(codex): match the User-Agents Codex actually sends

mactonicaaron:
## Problem

`CLIENT_UA_MAP` (in `headroom/proxy/auth_policy.py`) carries exactly one OpenAI needle:

```python
# OpenAI ecosystem
("codex-cli/", "codex"),
```

No shipped Codex build sends that User-Agent. Every Codex request therefore classifies as `None`, and the Codex fail-open in `decide_compression_failure_action()` — which keys on `client == "codex"` — never fires. A compression timeout returns 413, and Codex treats 413 as fatal, so the session dies at startup instead of degrading to an uncompressed forward.

That fail-open was added deliberately (`fix(codex): fail open for proxy compression timeout`). It has been unreachable via the UA path since.

## Reproduction

Counting the User-Agents on the wire across two `proxy.log` rotations on a box running Codex and Claude Code:

| User-Agent | requests | matched `codex-cli/`? |
|---|---:|---|
| `codex-tui/` | 3,167 | ❌ |
| `codex-browser-use/` | 1,338 | ❌ |
| `Codex Desktop/` | 368 | ❌ |
| `codex_cli_rs/` | 13 | ❌ |
| `codex-computer-use/` | 1 | ❌ |
| `claude-cli/` | 4,054 | n/a — already mapped |

4,887 Codex requests, zero matches.

Verbatim samples:

```
codex-tui/0.147.0 (Mac OS 26.5.1; arm64) Apple_Terminal/470.2 (codex-tui; 0.147.0)
codex-browser-use/0.148.0-alpha.9 (Mac OS 26.5.1; arm64) xterm-256color (codex-browser-use; 0.1.0)
Codex Desktop/0.148.0-alpha.9 (Mac OS 26.5.1; arm64) unknown (Codex Desktop; 26.810.41047)
codex_cli_rs/0.148.0-alpha.9 (Mac OS 26.5.1; arm64) unknown
codex-computer-use/0.147.0-alpha.6.5 (Mac OS 26.5.1; arm64) unknown
```

The error users see:

```
unexpected status 413 Payload Too Large: {"detail":{"error":{"type":"compression_refused",
"message":"headroom: compression timeout on a 199908-byte request ..."}}}
```

## Real behavior proof

**Setup**

- macOS 26.5.1, arm64, Python 3.13.13, `headroom-ai` 0.25.0 installed non-editable in a venv
- Proxy on `127.0.0.1:8787`, `HEADROOM_MODE=token`, backend `anthropic`
- Clients: Codex `0.147.0` / `0.148.0-alpha.9` (TUI, Desktop, browser-use) and Claude Code `2.1.233`

**Steps run after the patch**

Applied the change to the installed copy, then exercised `classify_client()` and the fail-open decision against it, from `/` so the repo working tree could not shadow site-packages:

```python
from headroom.proxy.auth_mode import classify_client
from headroom.proxy.helpers import decide_compression_failure_action
assert "site-packages" in am.__file__
for ua, count in OBSERVED:                      # the five UAs above, verbatim
    c = classify_client({"user-agent": ua})
    a = decide_compression_failure_action(asyncio.TimeoutError(), 199908, client=c)
```

**After-fix evidence — observed result**

Before the patch, against the live module:

```
MISS codex-tui/             -> None     (3,167 req)
MISS codex-browser-use/     -> None     (1,338 req)
MISS Codex Desktop/         -> None     (368 req)
MISS codex_cli_rs/          -> None     (13 req)
MISS codex-computer-use/    -> None     (1 req)
=> 0 classified / 4,887 missing
```

After:

```
✅ verified: all 5 shipped Codex UAs -> fail-open (4,887 requests); unknown client -> still refuses
```

`decide_compression_failure_action(...).refuse` is `False` for all five, and still `True` for `client=None`, so the guard is not disabled — only reachable for Codex, as intended.

**What I did not test**

- I did not measure how many of those 4,887 requests actually took the refuse branch in production. On my box `HEADROOM_WS_FAIL_OPEN_ON_COMPRESSION_FAILURE=1` and the `is_chatgpt_auth` path independently cover ChatGPT-authenticated `/v1/responses`, so I am not claiming a 413 count — only that classification was wrong for 100% of Codex traffic.
- Non-macOS Codex builds. Their UA segment may differ again; the needles are substring matches, so a Linux `codex-tui/` still matches.
- Cursor / aider / Copilot needles are untouched and unverified against live traffic.

## The fix

Adds the five observed needles. Note they must be **lowercase**: `classify_client()` lowercases the UA before matching, so a copy-pasted `"Codex Desktop/"` needle would never fire while still looking correct in review. `Codex Desktop/` is the only client UA containing a space.

I left `codex-cli/` in place rather than removing it, in case a build somewhere does send it.

## Tests

`tests/test_auth_mode_codex_user_agents.py`, 12 cases, using the log's verbatim UA strings.

Mutated back to the single `codex-cli/` needle → 10 fail, and the two negative controls stay green, proving the broadened needles do not over-match. Restored → 12/12.

Negative controls included deliberately:

- `Mozilla/5.0`, `curl/8.7.1` → still `None`
- `claude-cli/2.1.233 (external, cli)` → still `claude-code`
- `codex_cli_rs/0.147.0 ... (claude-code; 2.1.233)` → `codex`. This one is real traffic: the Codex binary reporting Claude Code as its originator. The Anthropic needles are checked first but require `claude-code/` with a slash, and this is a semicolon, so Codex correctly wins — Codex is the process that dies on the 413. Map order decides this, so it is asserted rather than assumed.

Existing `tests/test_auth_mode.py` stays 24/24 green. Full suite matches `main` exactly (11 failed / 86 passed on the pre-existing environment failures, identical on `main`). `ruff check` and `ruff format --check` clean.

---

*Rebased onto current `main`. The first push of this branch was built from a clone that was ~1,021 commits stale; the map has since moved from `auth_mode.py` to `auth_policy.py`, and the change is applied there. Re-verified against current `main`: reverting `auth_policy.py` to `main` fails 10 of the 12 new tests (the two negative controls stay green), and restoring it passes 12/12. `tests/test_auth_mode.py` and `tests/test_proxy_openai.py` stay green (59 passed together). Two sibling PRs I opened at the same time (#3048, #3050) turned out to be already fixed on `main` and I have closed them.*


JerrettDavis (commented):
The UA additions and production-derived regression cases look correctly scoped; the negative controls avoid an overly broad Codex match. Before moving this out of draft, please remove the manual CHANGELOG.md edit (the repository generates it and enforces that policy), convert the body to the current PR template, and update the now-stale comments/tests that say Codex Desktop is not a known UA (notably auth_mode.py, server.py, and test_codex_client_stamp.py). The code change itself looks sound, but leaving those statements behind would make the classification contract contradictory.

PR #3052, opened 2026-08-16
fix(proxy/anthropic): don't replay recorded prefix over live history (#3026)

### LINKED ISSUE #3026: [BUG] Claude Code session fails with 400 prompt_too_long on 0.35.0 when reading a large JSON file (works direct and on 0.33.0)
## Description

When Claude Code reads a moderately large JSON file (~220 KB, well within the model's context window) through the Headroom proxy on **0.35.0**, the session consistently fails with an upstream Anthropic API error `400 prompt_too_long` — even though the identical session completes successfully when connected directly to `api.anthropic.com`, and also completes (without compression) through Headroom **0.33.0**.

Interestingly, `/stats` shows that compression *did* fire on the failing requests (`lossless_json`, ~14–16k tokens removed per file-bearing request), so the payload sent upstream should have been smaller than the baseline that succeeds. This suggests the rewritten request body is malformed or inflated in a way that trips Anthropic's prompt-length validation, rather than a genuine context overflow.

Thank you for looking into this — happy to provide any additional logs or re-run the repro with debug flags if that helps.

## To Reproduce

1. Install: `pip install "headroom-ai[proxy]==0.35.0"` (fresh venv, Python 3.12.3)
2. Create a synthetic JSON dataset (array of dicts, one record per line):

```python
import json, random
random.seed(42)
regions = ["Hokkaido","Tohoku","Kanto","Chubu","Kansai","Chugoku","Shikoku","Kyushu"]
rows = [{
    "order_id": f"ORD-{100000+i}",
    "date": f"2026-{random.randint(1,7):02d}-{random.randint(1,28):02d}",
    "region": random.choice(regions),
    "product_id": f"PRD-{random.randint(1,80):04d}",
    "units": random.randint(1,50),
    "unit_price_jpy": random.choice([980,1480,2980,4980,9800,14800]),
    "channel": random.choice(["web","store","partner"]),
    "customer_segment": random.choice(["smb","enterprise","consumer"]),
} for i in range(1200)]
with open("data2.json","w") as f:
    f.write("[\n" + ",\n".join(json.dumps(r) for r in rows) + "\n]\n")
# ~220 KB, 1202 lines
```

3. Also create six tiny `notes1.txt` … `notes6.txt` files (3 lines each).
4. Start the proxy: `HEADROOM_TELEMETRY=0 headroom proxy --port 18790` (default options)
5. Run a headless Claude Code session through it:

```bash
ANTHROPIC_BASE_URL=http://127.0.0.1:18790 claude -p \
  "Do these steps strictly in order, one Read tool call per step: \
   1) Read data2.json in full. 2) Read notes1.txt. 3) Read notes2.txt. \
   4) Read notes3.txt. 5) Read notes4.txt. 6) Read notes5.txt. \
   7) Read notes6.txt. 8) Finally, tell me only what the key fact of notes6.txt is." \
  --model haiku --allowedTools "Read" --max-turns 14 --output-format json
```

## Expected Behavior

The session completes all 8 steps (as it does when `ANTHROPIC_BASE_URL` is unset), ideally with compression applied to the stale JSON tool_result.

## Actual Behavior

The session aborts on the request that carries the JSON tool_result:

- Result JSON: `"result": "Prompt is too long"`, `"api_error_status": 400`, `"terminal_reason": "prompt_too_long"`, `num_turns: 2`
- Reproduced **2 out of 2** attempts on 0.35.0, with both a 1,900-row (~350 KB) and a 1,200-row (~220 KB) file. The 1,200-row payload is roughly 60k tokens — nowhere near the model's 200k context window.
- Control runs with the identical prompt and files:
  - **Direct to api.anthropic.com** (no proxy): completes 8 turns normally.
  - **Headroom 0.33.0** (same task): completes 8 turns, though with 0 requests compressed (all skipped as `prefix_frozen`).

`/stats` after the failing 0.35.0 runs (cumulative, includes two `/v1/compress` probes):

```json
"compression": {
  "requests_compressed": 5,
  "avg_compression_pct": 8.0,
  "best_compression_pct": 23.2,
  "best_detail": "130,781 → 100,382 tokens",
  "total_tokens_removed": 60441
},
"uncompressed_requests": { "passthrough": 1, "no_compressible_content": 1 }
```

So compression fired on the file-bearing requests, yet the upstream rejected them as too long.

## Error Output

```
"result": "Prompt is too long"
"api_error_status": 400
"terminal_reason": "prompt_too_long"
```

(Claude Code surfaces the upstream Anthropic 400 as the final result; the proxy log shows no error at default verbosity.)

## Environment

- **Headroom version**: 0.35.0 (bug present) / 0.33.0 (no failure, but no compression either)
- **Python version**: 3.12.3
- **OS**: Ubuntu (Linux 6.8.0)
- **LLM Provider**: Anthropic (`claude-haiku-4-5-20251001`), Claude Code CLI 2.1.231, ChatGPT-style subscription OAuth (claude.ai login), default proxy options
- **Health at test time**: `kompress` backend reported `degraded` (rust_core loaded; SmartCrusher active)

## Additional Context

- The same 0.35.0 instance compresses the equivalent payload correctly via a direct `POST /v1/compress` probe (130,781 → 100,382 tokens, `lossless_json`), so the transform itself seems fine — the issue appears specific to the live `/v1/messages` rewrite path with Claude Code's cache_control-laden requests.
- Possibly related to interactions with Claude Code's `count_tokens` calls (observed as `passthrough` in stats) or prompt-cache breakpoint handling on the rewritten body.
- Happy to re-run with debug logging or a candidate fix.

rodboev:
## Description

A Claude Code session that reads a large tool result through `headroom proxy` can fail on turn 2 with Anthropic's `400 prompt_too_long`. The reporter's controlled comparison completed eight turns through 0.33.0 with 187,986 input tokens, while 0.35.0 failed after five requests with 753,077 input tokens. The local regression uses an actual prior optimized request to populate tracker state, then a decision-false bypass turn with Claude-shaped tool-result content. The old unconditional replay path substitutes the compressed prefix; the eligibility gate preserves the client's outbound body without claiming a live provider reproduction.

The Anthropic `/v1/messages` route computes whether a request should be compressed, but cached-prefix replay currently runs outside that decision. The replay helper also derives its prefix length from the original message list and applies that index to the optimized list without proving the two lists still align. A stale forwarded prefix can therefore be grafted onto the wrong positions and enlarge later requests.

This change limits replay to requests whose existing compression decision permits it and whose pre-upstream backpressure path is inactive. It also makes `overlay_cached_prefix()` decline misaligned or inflating candidates while preserving normal append-only replay.

Reported by @itsumonotakumi, whose controlled comparison isolated the failure from compression, headers, one-request serialization, memory, code graph, and CCR.

Closes #3026

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Gate Anthropic cached-prefix replay on the existing `CompressionDecision.should_compress` result and the existing pre-upstream backpressure state.
- Require positional alignment between optimized and original message arrays before replay.
- Reject replay candidates that would serialize larger than the current optimized messages.
- Add focused handler coverage for the decision-false tool-result regression, bypass and backpressure paths, and outbound optimize-on preservation.
- Add direct unit coverage for positional mismatch, no-inflation, and JSON sizing-failure bailouts.
- Update the moved-cache-control and pure-block-append regression fixtures to keep the no-inflation contract explicit.
- Run the unchanged OpenAI cache-stability preservation proof; no OpenAI production code was edited.

## Testing

- [x] Unit tests pass (153 focused proxy, helper, cache-control, block-append, cross-turn, byte-faithful, Anthropic, OpenAI, and backpressure tests)
- [x] Linting passes (Ruff check and format validation on the seven changed repository files)
- [x] Manual testing performed with the in-process proxy and local stub upstream

### Test Output

```text
python -m pytest tests/test_proxy/test_anthropic_no_optimize_history_passthrough.py tests/test_cache_prefix_overlay.py tests/test_cross_turn_cache_safety.py tests/test_cache_control_move_bust.py tests/test_proxy_byte_faithful_forwarding.py tests/test_proxy_anthropic_cache_stability.py tests/test_anthropic_pre_upstream_backpressure.py -q
python -m pytest tests/test_proxy_openai_cache_stability.py -q
python -m pytest tests/test_issue_2671_block_growth_cache.py::test_pure_append_replays_forwarded_blocks_and_advances_breakpoint -q
153 passed across focused invocations, exit code 0
optimize_off turn2_message_count=3 marker_count=1 outbound_compact_utf8_bytes=2293 client_compact_utf8_bytes=2293
optimize_on turn2_message_count=3 client_message_count=3 marker_count=0 outbound_compact_utf8_bytes=171 client_compact_utf8_bytes=182
python -m ruff check headroom/proxy/handlers/anthropic.py headroom/cache/prefix_tracker.py tests/test_proxy/test_anthropic_no_optimize_history_passthrough.py tests/test_cache_prefix_overlay.py tests/test_cache_control_move_bust.py tests/test_issue_2671_block_growth_cache.py tests/test_proxy_openai_cache_stability.py
All checks passed!, exit code 0

python -m ruff format headroom/proxy/handlers/anthropic.py headroom/cache/prefix_tracker.py tests/test_proxy/test_anthropic_no_optimize_history_passthrough.py tests/test_cache_prefix_overlay.py tests/test_cache_control_move_bust.py tests/test_issue_2671_block_growth_cache.py tests/test_proxy_openai_cache_stability.py --check
7 files already formatted, exit code 0

git diff --check
clean, exit code 0
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12 via `uv`, real Headroom proxy app with a local stub Anthropic upstream
- Exact command / steps: send an actual optimize-on first request through the in-process proxy with a deterministic production-pipeline seam, then send a decision-false bypass turn containing a large Claude-shaped `tool_result` with moved `cache_control`; separately send an aligned optimize-on turn with a new suffix
- Observed result: the exact base checkout fails with `AssertionError: assert 'compressed-tool-result' == 'large-tool-result-marker ...'`; the guarded path passes with the client marker present once and outbound compact JSON no larger than the client body. The optimize-on preservation run records `optimize_on turn2_message_count=3 client_message_count=3 marker_count=0 outbound_compact_utf8_bytes=171 client_compact_utf8_bytes=182`, proving the actual compressed prefix is outbound before the new suffix without turn-2 growth.
- Not tested: live Claude Code session against api.anthropic.com on this host

## Runtime Rollout Safety

- Stable/default behavior changed: cached-prefix replay now follows the existing compression and backpressure decision and rejects misaligned or inflating candidates.
- Kill switch / disable path: no new switch; the existing optimize and bypass controls remain available.
- Qualification impact: none.
- Rollback path: revert the implementation commit.

## Review Readiness

## Checklist

## Additional Notes

`CHANGELOG.md` is not modified because Headroom's release automation generates it from conventional commits.

This change does not add a context-limit guard or alter compression, streaming tracker provenance, outbound-body selection, OpenAI behavior, or provider limits. Local tests prove request-body ownership and replay bounds. The reporter's live Claude Code completion and Anthropic token acceptance remain external to this local proof.

### DIFF
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -548,15 +548,12 @@ def overlay_cached_prefix(
                     + [merged]
                     + list(optimized_messages[message_index + 1 :])
                 )
-                cache_control_only_change = current_original_messages[
-                    : message_index + 1
-                ] != prev_orig[: message_index + 1] and _strip_cache_control(
-                    current_original_messages[: message_index + 1]
-                ) == _strip_cache_control(prev_orig[: message_index + 1])
+                replayed_bytes = _compact_json_bytes(replayed)
+                optimized_bytes = _compact_json_bytes(optimized_messages)
                 if (
-                    len(_compact_json_bytes(replayed))
-                    > len(_compact_json_bytes(optimized_messages))
-                    and not cache_control_only_change
+                    replayed_bytes is None
+                    or optimized_bytes is None
+                    or len(replayed_bytes) > len(optimized_bytes)
                 ):
                     logger.debug("overlay: block replay inflated compact JSON — skipping")
                     return optimized_messages
@@ -607,20 +604,29 @@ def overlay_cached_prefix(
     # Replay the cached (compressed) prefix byte-identical up to the first
     # divergence; keep this turn's freshly-produced output for the rest.
     replayed = list(prev_fwd[:k]) + list(optimized_messages[k:])
-    cache_control_only_change = current_original_messages[:k] != prev_orig[
-        :k
-    ] and _strip_cache_control(current_original_messages[:k]) == _strip_cache_control(prev_orig[:k])
+    replayed_bytes = _compact_json_bytes(replayed)
+    optimized_bytes = _compact_json_bytes(optimized_messages)
     if (
-        len(_compact_json_bytes(replayed)) > len(_compact_json_bytes(optimized_messages))
-        and not cache_control_only_change
+        replayed_bytes is None
+        or optimized_bytes is None
+        or len(replayed_bytes) > len(optimized_bytes)
     ):
         logger.debug("overlay: replay inflated compact JSON — skipping cached-prefix replay")
         return optimized_messages
     return replayed
 
 
-def _compact_json_bytes(value: Any) -> bytes:
-    return json.dumps(value, separators=(",", ":"), ensure_ascii=False).encode("utf-8")
+def _compact_json_bytes(value: Any) -> bytes | None:
+    """Return compact JSON bytes, or ``None`` when sizing cannot be proved."""
+    try:
+        return json.dumps(
+            value,
+            separators=(",", ":"),
+            ensure_ascii=False,
+            default=str,
+        ).encode("utf-8")
+    except (TypeError, ValueError, OverflowError, UnicodeError):
+        return None
 
 
 _STABLE_BOUNDARY_ENV = "HEADROOM_STABLE_BOUNDARY_BREAKPOINT"
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -449,7 +449,7 @@ def overlay_cached_prefix(
     previous_original_messages: list[dict[str, Any]] | None,
     previous_forwarded_messages: list[dict[str, Any]] | None,
 ) -> list[dict[str, Any]]:
-    """Replay the previously-forwarded (cached, compressed) prefix byte-identical.
+    """Replay a positional, non-inflating cached prefix when it is safe.
 
     Provider-agnostic cache-safety guard for the freeze path. When a message is
     "frozen", the compression pipeline may emit the agent's ORIGINAL bytes for
@@ -467,13 +467,22 @@ def overlay_cached_prefix(
     ``optimized_messages`` unchanged (accept a possible bust rather than forward
     wrong content).
 
-    This makes freezing byte-identical in BOTH proxy modes, so the only remaining
-    difference between them is how large a mutable (still-compressible) tail each
-    leaves — not whether the frozen prefix busts the cache.
+    The optimized and current-original lists must be positionally aligned, and
+    compact UTF-8 JSON for the replayed result must not exceed the optimized
+    candidate. These bounds prefer a cache miss to corrupting or inflating a
+    client's live history.
     """
     prev_orig = previous_original_messages
     prev_fwd = previous_forwarded_messages
     if not prev_orig or not prev_fwd:
+        return optimized_messages
+    if len(optimized_messages) != len(current_original_messages):
+        logger.debug(
+            "overlay: optimized/current-original length mismatch (optimized=%d, current=%d) "
+            "— skipping positional cached-prefix replay",
+            len(optimized_messages),
+            len(current_original_messages),
+        )
         return optimized_messages
     n = len(prev_orig)
     # Positional 1:1 correspondence between prev_orig[i] and prev_fwd[i] holds
@@ -534,11 +543,24 @@ def overlay_cached_prefix(
                     len(current_content) - split,
                     message_index,
                 )
-                return (
+                replayed = (
                     list(prev_fwd[:message_index])
                     + [merged]
                     + list(optimized_messages[message_index + 1 :])
                 )
+                cache_control_only_change = current_original_messages[
+                    : message_index + 1
+                ] != prev_orig[: message_index + 1] and _strip_cache_control(
+                    current_original_messages[: message_index + 1]
+                ) == _strip_cache_control(prev_orig[: message_index + 1])
+                if (
+                    len(_compact_json_bytes(replayed))
+                    > len(_compact_json_bytes(optimized_messages))
+                    and not cache_control_only_change
+                ):
+                    logger.debug("overlay: block replay inflated compact JSON — skipping")
+                    return optimized_messages
+                return replayed
     # Append-only guard on CONTENT ONLY, message-by-message. Replay the
     # previously-forwarded (cached, compressed) bytes for the longest LEADING
     # run of messages that is byte-for-byte (content-canonical) identical to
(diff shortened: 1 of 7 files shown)


chopratejas (commented):
Reviewed the diff. The shape of this is right, and the reporter's controlled comparison in #3026 (187,986 input tokens on 0.33.0 vs 753,077 on 0.35.0, same session) makes a strong case that replay is the growth source.

What I like: the three bounds in `overlay_cached_prefix` are each independently sufficient to prevent the reported failure, and they fail *closed* — a cache miss instead of a corrupted or inflated prefix. Deriving the prefix length from one list and indexing the other was a real latent bug regardless of #3026, and `limit = min(n, len(current_original_messages))` is the correct floor.

Two things I'd like resolved before merge:

**1. The no-inflation bound is byte-length on compact JSON, not tokens.** `len(replayed_bytes) > len(optimized_bytes)` is a good proxy and cheap, but it isn't the quantity the 400 is raised on. A replay that is byte-smaller yet token-larger (different whitespace/unicode density between a compressed marker and the original text) would still pass this gate. Given the failure mode is specifically `prompt_too_long`, is there a reason not to bound on the tracker's own token count, with bytes as the fast pre-filter?

**2. Gating on `should_compress` couples two decisions that were previously independent.** Replay now only happens when the compression decision says yes. That's defensible — and it's what fixes the reported bypass turn — but it means any future change to `should_compress` silently changes cache-replay behavior too. A one-line comment at the call site naming that coupling would save the next reader a bisect.

Neither is a blocker if you'd rather land the bound now and refine it; the current state is strictly safer than `main`.

**On the test evidence** — I want to flag this plainly rather than let it pass as proven. The PR body is honest that there's no live-provider reproduction, and the local proof establishes request-body ownership and replay bounds, not that Anthropic accepts the result. That's the right claim to make, but it does mean the fix is verified against our model of the bug rather than the bug itself. Worth asking @itsumonotakumi to run this branch against their exact repro before we call #3026 closed.

Related: I merged #3024 for #3019, which touches the adjacent cache-hit path. Worth a rebase and a re-run before merge — the two changes are close enough to interact.

PR #3053, opened 2026-08-16
fix(plugin): resolve hook CLI through plugin-root launcher

### LINKED ISSUE #3039: [BUG] Plugin hooks invoke bare 'headroom', so PreToolUse/SessionStart fail with 127 on every tool call when the agent host has no login-shell PATH
## Summary

`plugins/headroom-agent-hooks/hooks/hooks.json` invokes the CLI by **bare name** (`headroom init hook ensure`) in both `SessionStart` and `PreToolUse`. Claude Code executes hook commands with `/bin/sh`, which does **not** inherit the interactive shell's `PATH`. When the agent process is started from a non-login context (a `tmux` server, a LaunchAgent/systemd unit, a CI runner), `$HOME/.local/bin` is not on `PATH`, so the hook fails on every single tool call:

```
PreToolUse:Bash [headroom] failed with non-blocking status code 127:
/bin/sh: headroom: command not found
```

Because the failure is non-blocking, tool calls still succeed and nothing appears broken — but **headroom never actually runs**, so the context compression the plugin exists to provide is silently lost for the entire session, and the repeated error banner drowns out genuine hook failures.

Verified on plugin **0.34.0**; the file is byte-identical on **v0.35.0**, so latest is still affected.

## Reproduction

Any environment where the agent is not launched from a login shell. Minimal reproduction of the hook's execution context:

```console
$ env -i HOME="$HOME" PATH=/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin \
    /bin/sh -c 'headroom init hook ensure'
/bin/sh: headroom: command not found
$ echo $?
127
```

Claude Code classifies that 127 as non-blocking, so the session carries on — with compression disabled and a banner on every tool call.

The binary is present and healthy the whole time — it just is not on that `PATH`:

```console
$ command -v headroom
~/.local/bin/headroom -> ~/.local/share/uv/tools/headroom-ai/bin/headroom
```

Same command, same shell, with the install dir on `PATH` — silent success, hook does its job:

```console
$ env -i HOME="$HOME" PATH="$HOME/.local/bin:/usr/bin:/bin" \
    /bin/sh -c 'headroom init hook ensure'
$ echo $?
0
```

So the only variable is `PATH` resolution of the bare name.

## Root cause

`plugins/headroom-agent-hooks/hooks/hooks.json` (v0.35.0, both entries):

```json
{ "type": "command", "command": "headroom init hook ensure", "timeout": 15 }
```

Bare-name resolution depends on an ambient `PATH` that hook execution is not guaranteed to have. The default `uv`/`pipx` install location (`$HOME/.local/bin`) is on almost nobody's `/bin/sh` default `PATH`.

Worth noting: `/usr/local/bin` is **not** a reliable escape hatch on Apple Silicon — Homebrew installs to `/opt/homebrew/bin` there, and a non-login `PATH` on such a machine commonly reads `/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin` with no `/usr/local/bin` at all.

## Suggested fix

Anchor the command instead of relying on `PATH`. Every other Claude Code plugin I have installed (`superpowers`, `vercel`, `claude-obsidian`) resolves its hook entrypoint through `${CLAUDE_PLUGIN_ROOT}` rather than a bare name — headroom is the outlier.

Options, roughly in order of robustness:

1. **Ship a small launcher in the plugin** (`plugins/headroom-agent-hooks/bin/headroom`) that locates the real CLI, and reference it as `"${CLAUDE_PLUGIN_ROOT}/bin/headroom" init hook ensure`. This matches the surrounding ecosystem convention and survives any `PATH`.
2. **Self-heal `PATH` in the hook command**, e.g.
   `sh -c 'command -v headroom >/dev/null 2>&1 || PATH="$HOME/.local/bin:$PATH"; exec headroom init hook ensure'`
   — a one-line change to `hooks.json` that covers the dominant `uv`/`pipx` install location.
3. At minimum, **document the `PATH` requirement** in the plugin README and make the hook fail loudly (or emit a one-time actionable message) rather than repeating `command not found` on every tool call.

Option 1 or 2 would fix this for all non-login-shell agent hosts without users having to discover the failure themselves.

## Environment

- headroom plugin 0.34.0 (installed via `headroom-marketplace`); `hooks.json` confirmed unchanged in v0.35.0
- CLI installed with `uv` at `$HOME/.local/bin/headroom`
- macOS (Apple Silicon), Claude Code, agent sessions hosted under a `tmux` server started from a non-login context

rodboev:
## Description

The `headroom` plugin invokes `headroom init hook ensure` through non-login `/bin/sh`, but both `SessionStart` and `PreToolUse` entries use a bare executable name. Agent hosts whose PATH omits the default uv/pipx user bin exit 127 on every matching event. Tool calls continue because the hook is non-blocking, while Headroom compression remains inactive and the failure banner repeats.

Both hooks now invoke one launcher bundled under `${CLAUDE_PLUGIN_ROOT}` through `sh`. The launcher honors `HEADROOM_BIN`, preserves PATH-resolved installations, covers standard uv and pipx prefixes, records the Homebrew candidates for macOS, and falls back to an importable Python module. On the launcher-routed path, an absent Headroom CLI emits one actionable diagnostic and exits 0. Hosts that do not expose a readable plugin root retain the existing bare-command tail, preserving the shared GitHub Copilot CLI manifest behavior.

Reported by [@donnydongchen](https://github.com/donnydongchen), whose exact non-login `env -i` reproduction, Apple Silicon path detail, and plugin-root suggestion shaped the fix.

Closes #3039

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Route both plugin hook entries through a shared `${CLAUDE_PLUGIN_ROOT}`-anchored `sh` launcher while preserving matchers, arguments, and 15-second timeouts.
- Require a non-empty plugin root before probing the launcher, so unset or empty roots always reach the existing bare-command tail.
- Resolve `HEADROOM_BIN`, PATH, standard uv/pipx/Homebrew locations, and an importable `python -m headroom.cli` fallback in a deterministic order.
- Preserve the current bare command for hosts without a readable plugin root, including the shared Copilot CLI manifest path.
- On the launcher-routed path, replace the absent-CLI status-127 banner with one actionable diagnostic and a non-blocking exit.
- Add real-shell regression coverage for the reported environment, resolution precedence, supported prefixes, Python fallback, and negative boundaries.

## Testing

- [x] Unit tests pass (`python -m pytest tests/test_plugin_manifests.py tests/test_plugin_hook_launcher.py tests/test_cli/test_init_cli.py -q`, `88 passed`)
- [x] Reproduction override passes (`HEADROOM_REPRO_ARTIFACT` can select the issue fixture, `python -m pytest tests/test_plugin_hook_launcher.py::test_manifest_command_recovers_reported_non_login_environment -q`, `1 passed`)
- [x] Linting passes (Ruff check and format validation on the changed Python test files)

### Test Output

```text
88 passed
All checks passed!
3 files already formatted
```

## Real Behavior Proof

- Environment: POSIX `sh` with the reporter's sanitized non-login PATH and a recording Headroom executable only at `$HOME/.local/bin/headroom`
- Exact command / steps: `python -m pytest tests/test_plugin_hook_launcher.py::test_manifest_command_recovers_reported_non_login_environment -q`; the test preserves the reported PATH and error evidence, executes both manifest commands with a recorder at `$HOME/.local/bin/headroom`, and asserts both `home-local init hook ensure` receipts.
- Observed result: `1 passed in 0.41s`; both hooks recovered under the reported non-login PATH and emitted no command-not-found banner.
- Not tested: live Claude Code plugin-hook execution on a macOS tmux host

## Runtime Rollout Safety

- Stable/default behavior changed: plugin hooks resolve Headroom through the bundled launcher when the plugin root is readable; invalid launcher roots retain the existing bare-command fallback.
- Kill switch / disable path: none; the existing bare-command fallback remains available when the plugin root is absent or unreadable.
- Qualification impact: none.
- Rollback path: revert the implementation commit.

## Review Readiness

## Checklist

## Additional Notes

The launcher does not source shell profiles, install packages, access the network, change deployment/profile selection, or modify plugin versions.

### DIFF
--- a/plugins/headroom-agent-hooks/README.md
+++ b/plugins/headroom-agent-hooks/README.md
@@ -2,10 +2,22 @@
 
 This plugin exposes lightweight startup hooks for Claude Code and GitHub Copilot CLI.
 
-The hooks call:
+The hooks resolve and call:
 
 ```bash
 headroom init hook ensure
 ```
 
 That hidden helper checks for a matching durable `headroom init` deployment and starts it if needed.
+
+Resolution order is `HEADROOM_BIN`, `headroom` on `PATH`, then `headroom` and
+`headroom.exe` in `$HOME/.local/bin`, `$HOME/.local/share/uv/tools/headroom-ai/bin`,
+`${PIPX_HOME:-$HOME/.local/pipx}/venvs/headroom-ai/bin`, `/opt/homebrew/bin`, and
+`/usr/local/bin`. If no executable is found, `HEADROOM_PYTHON`, `python3`, and
+`python` are checked for an importable `headroom` module and invoked as
+`python -m headroom.cli init hook ensure`.
+
+Set `HEADROOM_BIN` or `HEADROOM_PYTHON` to override discovery. An absent CLI emits
+one actionable diagnostic and exits successfully, so the host hook remains
+nonblocking. On hosts that do not provide `CLAUDE_PLUGIN_ROOT`, the manifest
+preserves the rootless `exec headroom init hook ensure` tail for Copilot CLI.
--- a/plugins/headroom-agent-hooks/bin/headroom-hook.sh
+++ b/plugins/headroom-agent-hooks/bin/headroom-hook.sh
@@ -1,0 +1,46 @@
+#!/bin/sh
+set -u
+
+if [ -n "${HEADROOM_BIN:-}" ] && [ -x "$HEADROOM_BIN" ]; then
+    exec "$HEADROOM_BIN" init hook ensure
+fi
+
+if command -v headroom >/dev/null 2>&1; then
+    exec headroom init hook ensure
+fi
+
+home=${HOME:-}
+for prefix in \
+    "$home/.local/bin" \
+    "$home/.local/share/uv/tools/headroom-ai/bin" \
+    "${PIPX_HOME:-$home/.local/pipx}/venvs/headroom-ai/bin" \
+    "/opt/homebrew/bin" \
+    "/usr/local/bin"
+do
+    for executable in headroom headroom.exe
+    do
+        candidate="$prefix/$executable"
+        if [ -x "$candidate" ]; then
+            exec "$candidate" init hook ensure
+        fi
+    done
+done
+
+try_python() {
+    candidate=$1
+    if command -v "$candidate" >/dev/null 2>&1; then
+        candidate=$(command -v "$candidate")
+    elif [ ! -x "$candidate" ]; then
+        return
+    fi
+    if "$candidate" -c 'import headroom' >/dev/null 2>&1; then
+        exec "$candidate" -m headroom.cli init hook ensure
+    fi
+}
+
+try_python "${HEADROOM_PYTHON:-}"
+try_python python3
+try_python python
+
+printf '%s\n' "headroom: CLI not found; install with 'uv tool install headroom-ai' or set HEADROOM_BIN; compression hooks are inactive." >&2
+exit 0
--- a/plugins/headroom-agent-hooks/hooks/hooks.json
+++ b/plugins/headroom-agent-hooks/hooks/hooks.json
@@ -7,7 +7,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "sh -c 'launcher=\"${CLAUDE_PLUGIN_ROOT}/bin/headroom-hook.sh\"; [ -r \"$launcher\" ] && exec sh \"$launcher\"; exec headroom init hook ensure'",
+            "command": "sh -c '[ -n \"${CLAUDE_PLUGIN_ROOT:-}\" ] && launcher=\"${CLAUDE_PLUGIN_ROOT}/bin/headroom-hook.sh\" && [ -r \"$launcher\" ] && exec sh \"$launcher\"; exec headroom init hook ensure'",
             "timeout": 15
           }
         ]
@@ -19,7 +19,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "sh -c 'launcher=\"${CLAUDE_PLUGIN_ROOT}/bin/headroom-hook.sh\"; [ -r \"$launcher\" ] && exec sh \"$launcher\"; exec headroom init hook ensure'",
+            "command": "sh -c '[ -n \"${CLAUDE_PLUGIN_ROOT:-}\" ] && launcher=\"${CLAUDE_PLUGIN_ROOT}/bin/headroom-hook.sh\" && [ -r \"$launcher\" ] && exec sh \"$launcher\"; exec headroom init hook ensure'",
             "timeout": 15
           }
         ]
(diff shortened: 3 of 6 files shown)


chopratejas (commented):
Read the diff. This is a clean fix for #3039 and the reproduction fixture is exactly the right way to pin it — `env -i` with the reporter's sanitized PATH is the actual failing condition, not an approximation of it.

Things this gets right that are easy to get wrong:

- **`exit 0` on the not-found path.** The hook is non-blocking by design, so a missing CLI must degrade to "compression inactive" rather than failing every tool call. Turning the repeated status-127 banner into one actionable line is the correct trade.
- **`HEADROOM_BIN` first, `PATH` second.** An explicit operator override should beat discovery, and it does.
- **Preserving the bare-command tail when the plugin root is unreadable**, which keeps the shared Copilot CLI manifest working.

Two questions, neither blocking:

**1. `/opt/homebrew/bin` is hardcoded, so this covers Apple Silicon but not Intel macOS**, where Homebrew installs to `/usr/local/bin`. The reporter is on Apple Silicon so the fixture passes, but an Intel user with a Homebrew-installed CLI and a sanitized PATH still lands on the diagnostic. Adding `/usr/local/bin` to the candidate list is one line and closes that gap.

**2. The `python -m headroom.cli` fallback.** Which interpreter does this resolve to when `HEADROOM_PYTHON` is unset — `python` or `python3`? On a host where `python` is absent (increasingly common) or points at a 2.x shim, the fallback silently does nothing and we're back to the diagnostic. Worth being explicit about the resolution order here the way you were for the binary.

The test coverage is genuinely good — resolution precedence, supported prefixes, the Python fallback, and negative boundaries is the right matrix, and using a real shell rather than mocking `command -v` means the test would catch a quoting bug. Happy to approve once the Intel Homebrew path is in or you've decided it's out of scope.

rodboev:
Thanks for the careful read. I checked and it seems both points you raised are already covered on the current `head`:

1. The launcher probes `/opt/homebrew/bin` and `/usr/local/bin in order` (`test_launcher_declares_absolute_standard_prefixes` pins both paths.)
2. Interpreter resolution is explicit as `HEADROOM_PYTHON, then `python3`, then `python`; each candidate must pass import headroom before it runs `-m headroom.cli init hook ensure`. The `README` documents the same order.

I don't see anything to do here but if you feel this needs further work please ping me.

JerrettDavis (approved):
Reviewed the current launcher and shell-level test matrix. It now covers both Homebrew prefixes, explicit HEADROOM_BIN/PATH/fixed-prefix precedence, import-checked HEADROOM_PYTHON/python3/python fallback, safe quoting, rootless manifest behavior, and a non-blocking missing-CLI outcome. The prior review questions are resolved on this head; looks good.

PR #3059, opened 2026-08-16
feat(proxy): add x-headroom-keep-last-turns header for per-request context trimming (issue #2858)

### LINKED ISSUE #2858: [FEATURE] Keyword-based previous context clearing?
## Problem Statement

On long running tasks when an agent implements multiple problems that are not necessarily linked to each other, then at least with Cloud Code, there is no automatic or simple way to purge the context. 

## Proposed Solution

I wonder if it would make sense to have a magic word or some keyword that would allow the agent to remove parts of its own context. So let's say if it's an agentic session that is long running and implementing multiple things, so then between the tasks the agent could say a special word and that would cause Headroom to never ever send the context that is before that ever again. 

Example:
```
contextoftask1
contextoftask1
contextoftask1
contextoftask1
contextoftask1
contextoftask1
// agent finishes, and says 
CLEAR_EVERYTHING_ABOVE
contextoftask2
contextoftask2
contextoftask2
contextoftask2
contextoftask2
// agent finishes, and says 
CLEAR_EVERYTHING_ABOVE
...
```

So after the first "CLEAR_EVERYTHING_ABOVE" headroom would just drop contextoftask1, on the second it would also drop contextoftask2, and so on.

Maybe it would also make sense that the user would be able to type this, So not just the output of the agent, but maybe the input of the agent could be the keyword as well. Because advanced automations might talk agent to agent, you know. I know someone who is running Claude in Tmux and other agents are checking on the Tmux sessions and stuff :)

## Use Case

- How would this feature help you?
I think it could save input token usage.

- How many tokens/cost would this save?
Well, it's hard to put in numbers, but on a long session, if you can purge the context that is unneeded when you start to work on something new, I think that makes sense. 

## Alternatives Considered

Yeah, I was wondering that maybe I should do gardening instead.

gingeekrishna:
## Description

Adds `x-headroom-keep-last-turns: N` request header to both the Anthropic and OpenAI proxy handlers. When present, trims the leading message history to keep only the last `N` complete conversational turns plus the entire current turn, preserving system/developer instructions and tool-call/result pairing before any compression is applied. Implemented as a header to avoid prompt-injection risk; fail-open on malformed values.

Closes #2858.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- **`headroom/proxy/helpers.py`** — new `apply_keep_last_turns(messages, n)` → `(trimmed_list, n_dropped)` helper; returns original list identity when nothing is dropped
- **`headroom/proxy/handlers/anthropic.py`** — reads `x-headroom-keep-last-turns` before `optimized_messages = messages`; calls helper; logs dropped count at INFO level
- **`headroom/proxy/handlers/openai.py`** — same pattern as Anthropic handler
- **`tests/test_keep_last_turns.py`** — 23 helper and handler tests covering turn boundaries, OpenAI/Anthropic tool pairing, preserved system/developer instructions, dropped-message counts, and no-op identity.

## Testing

### Test Output

```text
$ python -m pytest tests/test_keep_last_turns.py -q
23 passed, 1 warning in 1.96s

$ uvx ruff==0.16.3 check headroom/proxy/helpers.py tests/test_keep_last_turns.py
All checks passed!

$ uvx ruff==0.16.3 format --check headroom/proxy/helpers.py tests/test_keep_last_turns.py
2 files already formatted
```

Review validation used macOS, Python 3.13.14, and the existing native extension. The warning is an installed Starlette/httpx deprecation.

## Real Behavior Proof

- Environment: macOS, Python 3.13.14, headroom source with existing native extension
- Exact command / steps: `python -m pytest tests/test_keep_last_turns.py -v`
- Observed result: 23 tests pass, including real proxy handler requests with mocked upstream responses. System/developer instructions and retained tool turns survive trimming; only removed messages contribute to the dropped count.
- Not tested: live proxy request with real upstream model (manual step)

## Runtime Rollout Safety

- Rollout-managed feature(s): None — header-driven, opt-in per request
- Minimum rollout channel: N/A (additive, backwards-compatible)
- Stable/default behavior changed: No — absent header leaves existing behavior unchanged; malformed values are silently ignored (fail-open)
- Kill switch / disable path: Simply omit the `x-headroom-keep-last-turns` header
- Qualification impact: None
- Rollback path: Revert this PR; no schema or data migration involved

## Review Readiness

## Checklist

N/A — no UI changes.

## Additional Notes

Trimming finds genuine user-turn starts, treating Anthropic tool-result messages as continuations. It retains the last N prior turns plus the entire current turn, and preserves every system/developer instruction in original order. When nothing is removed, the original list is returned unchanged.

Signed-off-by: Radha Krishnan P <gingeekrishna@gmail.com>

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -989,6 +989,7 @@ class AnthropicHandlerMixin:
             # point on, `headers` is the upstream-bound copy.
             from headroom.proxy.helpers import (
                 _strip_internal_headers,
+                apply_keep_last_turns,
                 log_outbound_headers,
                 merge_extra_headers,
             )
@@ -1248,6 +1249,25 @@ class AnthropicHandlerMixin:
                     logger.debug(f"[{request_id}] pre_compress hook error: {e}")
             else:
                 _hook_ctx = None
+
+            # x-headroom-keep-last-turns: N — trim history before optimization.
+            # Consumed here (after bypass check, after _strip_internal_headers)
+            # so it never leaks upstream.  Fail-open: any malformed value is
+            # silently ignored and the full message list is used instead.
+            _klt_raw = request.headers.get("x-headroom-keep-last-turns", "").strip()
+            if _klt_raw and not _bypass:
+                try:
+                    _klt = int(_klt_raw)
+                    messages, _klt_dropped = apply_keep_last_turns(messages, _klt)
+                    if _klt_dropped:
+                        logger.info(
+                            "[%s] keep-last-turns=%d: dropped %d leading messages",
+                            request_id,
+                            _klt,
+                            _klt_dropped,
+                        )
+                except ValueError:
+                    pass  # malformed value — never break the request
 
             # Apply optimization
             transforms_applied = []
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -3113,6 +3113,7 @@ class OpenAIHandlerMixin:
         # upstream-bound copy.
         from headroom.proxy.helpers import (
             _strip_internal_headers,
+            apply_keep_last_turns,
             log_outbound_headers,
             merge_extra_headers,
         )
@@ -3292,6 +3293,25 @@ class OpenAIHandlerMixin:
                 _hook_biases = self.config.hooks.compute_biases(messages, _hook_ctx)
             except Exception as e:
                 logger.debug(f"[{request_id}] Hook error: {e}")
+
+        # x-headroom-keep-last-turns: N — trim history before optimization.
+        # Consumed here (after bypass check, after _strip_internal_headers)
+        # so it never leaks upstream.  Fail-open: any malformed value is
+        # silently ignored and the full message list is used instead.
+        _klt_raw = request.headers.get("x-headroom-keep-last-turns", "").strip()
+        if _klt_raw and not _bypass:
+            try:
+                _klt = int(_klt_raw)
+                messages, _klt_dropped = apply_keep_last_turns(messages, _klt)
+                if _klt_dropped:
+                    logger.info(
+                        "[%s] keep-last-turns=%d: dropped %d leading messages",
+                        request_id,
+                        _klt,
+                        _klt_dropped,
+                    )
+            except ValueError:
+                pass  # malformed value — never break the request
 
         # Optimization
         transforms_applied = []
--- a/headroom/proxy/helpers.py
+++ b/headroom/proxy/helpers.py
@@ -3293,3 +3293,30 @@ def inject_tool_search_deferral_openai(
     if deferred == 0:
         return tools  # nothing to defer → don't perturb the request / cache prefix
     return out
+
+
+def apply_keep_last_turns(
+    messages: list[dict[str, Any]],
+    n: int,
+) -> tuple[list[dict[str, Any]], int]:
+    """Trim ``messages`` to the last *n* conversation turns.
+
+    A "turn" is one user+assistant pair.  The trailing user message is
+    always kept regardless of *n* (it is the current request).
+
+    Returns ``(trimmed_messages, n_dropped)`` — the caller can log
+    *n_dropped* and append ``keep_last_turns:{n}:{n_dropped}_dropped``
+    to ``transforms_applied``.  When nothing is dropped (n_dropped == 0)
+    the original list is returned unchanged so callers can detect a no-op
+    with an identity check.
+
+    Invariants:
+    - n < 0 is treated as no-op (invalid, never trim).
+    - The result always contains at least one message (the final user msg).
+    """
+    if n < 0 or not messages:
+        return messages, 0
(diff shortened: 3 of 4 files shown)


chopratejas (commented):
Holding this out of the merge batch: `lint` is failing, and it has no human review yet.

`ruff check` / `ruff format --check` on the changed files locally should reproduce it — it's usually formatting rather than a real defect, so this is likely a quick fix. Once lint is green I'll do a proper read of the change itself.

The PR governance bot has also flagged missing template fields on this one; worth clearing both in the same push.

gingeekrishna:
Fixed both — lint errors resolved (removed unused `pytest` import, fixed import sort order) and PR description updated to match the template.

JerrettDavis (changes requested):
apply_keep_last_turns assumes every turn is exactly two list entries (user, assistant), which is not true for the OpenAI handler this is wired into. A common history is user -> assistant(tool_calls) -> tool -> assistant -> user; with n=1, the arithmetic slice keeps only the last three entries, orphaning the tool result from its assistant tool call and sending an invalid/incoherent history upstream. Multiple tool results make the boundary even less predictable. Please derive the cutoff from message roles/actual user-turn boundaries (while preserving all messages belonging to retained turns), and add handler-level regression coverage for OpenAI tool-call/tool-result history plus ordinary Anthropic history. The existing helper tests only generate perfectly alternating pairs, so they cannot catch this integration defect.

gingeekrishna:
@JerrettDavis Confirmed and fixed. Replaced the arithmetic slice with turn-boundary detection: a message only starts a new turn when its role is \`"user"\` AND it isn't itself a tool-result continuation of the previous turn. \`user -> assistant(tool_calls) -> tool -> tool -> assistant\` is now correctly treated as one turn regardless of how many messages it spans, instead of an assumed fixed size of 2.

Anthropic needed the same care for a different reason: it represents tool results as \`role="user"\` messages (\`content\` blocks of type \`tool_result\`), so a bare role check would misfire on ordinary Anthropic tool use too. Reused the existing \`_is_tool_result_message\` classifier (already trusted elsewhere in the codebase for exactly this OpenAI/Anthropic distinction) rather than writing a second, possibly-diverging check.

All existing helper tests (perfectly-alternating pairs) still pass unchanged — the new turn-boundary math produces identical cutoffs to the old formula for that case, so this is purely additive correctness for the cases that formula got wrong.

Added the handler-level regression coverage you asked for — real \`/v1/chat/completions\` and \`/v1/messages\` requests through a captured outbound call, not just the helper in isolation: OpenAI tool-call/tool-result history, ordinary Anthropic history, and Anthropic tool_use/tool_result history (bonus, since the fix generalizes there too). All four new turn-boundary test cases reproduce the exact orphaned-tool-message symptom you described against the pre-fix code — confirmed that before verifying they pass against the fix. 19/19 tests pass.

JerrettDavis (changes requested):
The role-based cutoff resolves the orphaned tool-result case, and all 19 keep-last-turns tests pass, including handler-level tool pairing. One remaining OpenAI integration issue needs correction: the slice also removes system and developer instructions, which are not historical conversation turns. With [system("Always answer in JSON"), developer("Use the required response schema"), user(old), assistant(old), user(new)] and n=0, apply_keep_last_turns returns only user(new), dropping both application instructions. The OpenAI handler applies this slice directly to its messages list before forwarding.

Please retain system/developer messages in their original order while trimming complete conversational turns, and add a handler-level regression with both roles and a tool-using retained turn. The dropped-message count should count only messages actually removed. This preserves the requested last-N-turn behavior without silently changing the application instructions.

gingeekrishna:
@JerrettDavis Confirmed and fixed with your exact repro as a test case. \`apply_keep_last_turns\` now treats \`system\`/\`developer\` messages as always-kept application instructions rather than conversational turns: never counted when finding turn boundaries, never dropped regardless of \`n\`, and kept in their original relative position (not hoisted to the front) alongside whichever conversational turns the existing logic decides to keep. \`n_dropped\` now counts only messages actually removed, per your note.

New tests cover: your exact repro, original-order preservation with an interleaved instruction message (not just two leading ones), backward compatibility with no instruction messages present, and — the specific combination you asked for — a handler-level test through the real \`/v1/chat/completions\` endpoint with both instruction roles *and* a tool-using turn that's retained (not dropped), confirming the forwarded request has instructions preserved, the old turn gone, and the tool_calls/tool pairing still intact. All four new tests confirmed to reproduce the exact silent-drop symptom against the pre-fix code before verifying the fix. 23/23 tests pass.

JerrettDavis (approved):
The latest change resolves the instruction-loss finding: system/developer messages remain in order, dropped counts reflect only removed messages, and retained tool turns stay intact. Re-reviewed the complete helper and both handler integrations; all 23 tests pass, including the OpenAI handler regression with both instruction roles and a retained tool-call/result turn. Ruff lint and formatting pass. Updated the PR description to reflect role-based trimming and the current tests. No remaining blocking code findings.

Issue #2409, opened 2026-07-19
[BUG] Opencode GitHub Copilot provider Claude models return 404 because /v1/messages is forwarded as /messages


mars-peng-lb:
Description
GitHub Copilot Claude requests routed through Headroom return 404 page not found.
GitHub Copilot advertises /v1/messages as a supported endpoint for Claude models. Direct requests to that endpoint succeed, but requests routed through Headroom fail.
Affected models include:
- claude-haiku-4.5
- claude-sonnet-4.6
GitHub Copilot models using /chat/completions, such as Kimi, continue to work.
To Reproduce
1. Install Headroom as a persistent service with the OpenCode target.
2. Enable the Headroom OpenCode integration.
3. In OpenCode, select GitHub Copilot as the provider.
4. Select claude-haiku-4.5 or claude-sonnet-4.6.
5. Send any prompt.
6. Observe the 404 page not found response.
Expected Behavior
Headroom should forward the request to:
https://api.githubcopilot.com/v1/messages
The Claude model should return a normal response.
Actual Behavior
The request reaches Headroom, but GitHub Copilot receives a request at /messages rather than /v1/messages.
GitHub Copilot returns HTTP 404.
Code Sample
Not applicable. The issue is reproducible through the OpenCode UI using the GitHub Copilot provider.
Error Output
UPSTREAM_ERROR status=404
error_type=parse_error
error_msg='404 page not found\n'
model=claude-sonnet-4.6
The same error occurs with claude-haiku-4.5.
Environment
- Headroom version: 0.32.0
- Python version: 3.14.0
- OS: macOS 26.5 , arm64
- OpenCode version: 1.18.3
- headroom-opencode version: 0.1.0
- LLM Provider: Opencode GitHub Copilot
- Affected models: claude-haiku-4.5, claude-sonnet-4.6
Additional Context
GitHub Copilot’s /models response advertises these endpoints for the affected Claude models:
/chat/completions
/v1/messages
Testing with the same account and token confirms:
https://api.githubcopilot.com/v1/messages → HTTP 200
https://api.githubcopilot.com/messages    → HTTP 404
Headroom currently normalizes Copilot paths beginning with /v1/ by removing that prefix. This is correct for some OpenAI-compatible Copilot routes, but not for the Anthropic Messages endpoint.
The /v1 prefix should be preserved for /v1/messages.

PR #3077, opened 2026-08-17
fix(copilot): route VS Code inline completions to Copilot, not OpenAI

### LINKED ISSUE #3076: [BUG] VS Code Copilot inline completions are forwarded to api.openai.com instead of the Copilot API
## Description

With `github.copilot.advanced.debug.overrideProxyUrl` pointing at Headroom, the VS Code Copilot extension sends its inline ("ghost text") completions to:

```
POST /v1/engines/<engine>/completions
```

Headroom registers no route for that path, so it falls through to the catch-all passthrough. `select_passthrough_base_url()` resolves an upstream from the **auth headers alone** and never looks at the path, so the request takes the default OpenAI branch and Headroom forwards it to:

```
https://api.openai.com/v1/engines/gpt-41-copilot/completions
```

This is wrong under every configuration — OpenAI removed the Engines API years ago, so that host cannot serve the request. It is also actively blocked on corporate networks that allow GitHub Copilot but not OpenAI, which is how it was reported.

The Copilot **CLI** is unaffected: it uses the CAPI shape (`/chat/completions`, `/models`), which resolves to `https://api.githubcopilot.com` correctly.

## To Reproduce

1. Point VS Code Copilot at Headroom (`headroom wrap vscode`, or set `overrideProxyUrl`/`overrideCapiUrl` by hand).
2. Type in an editor to trigger an inline completion.
3. Observe the outbound request host in the proxy logs.

## Expected Behavior

Inline completions reach the Copilot API (`https://api.githubcopilot.com`, or the configured enterprise host) with the path Copilot built, unchanged.

## Actual Behavior

They are sent to `api.openai.com`, which 404s or is blocked outright by network policy. Inline completions stop working for every user behind such a policy.

## Additional Context

There is a second half to this. `build_copilot_upstream_url()` strips the `/v1` prefix when the upstream is a Copilot host, because Copilot serves its OpenAI-compatible surface unprefixed (`/chat/completions`, `/models`). But `/v1/engines/<engine>/completions` is a path the Copilot extension built itself, so it is already exactly what Copilot serves — stripping `/v1` there rewrites a working path into a 404. This is the same class of bug as #2409, where `/v1/messages` needed its prefix preserved.

So both are needed: route the path to Copilot, and leave the path alone once there.

## Environment

- **Headroom version**: 0.35.x / main
- **LLM Provider**: GitHub Copilot subscription
- **Client**: VS Code Copilot extension (reported on 0.60.0)
- **Engine observed**: `gpt-41-copilot`

Related: #2827 (chat bypassing the proxy, fixed by #2986 adding `overrideCapiUrl`), #2409 (`/v1/messages` prefix).

chopratejas:
## Description

Fixes #3076.

When `github.copilot.advanced.debug.overrideProxyUrl` points at Headroom, the VS Code Copilot extension sends its inline ("ghost text") completions to `/v1/engines/<engine>/completions`. No route matches that path, so it falls into the catch-all passthrough — and `select_passthrough_base_url()` resolves an upstream from the **auth headers alone**, never looking at the path. Copilot sends none of the headers the earlier branches key on, so the request reached the final line (default to OpenAI) and Headroom forwarded editor keystrokes to:

```
https://api.openai.com/v1/engines/gpt-41-copilot/completions
```

Wrong under every configuration — OpenAI removed the Engines API years ago — and blocked outright on corporate networks that permit GitHub Copilot but not OpenAI, which is how it was reported. Inline completions stopped working for every user behind such a policy.

The Copilot **CLI** was unaffected: it speaks the CAPI shape (`/chat/completions`), which already resolved correctly. That is the exact asymmetry in the report.

## Type of Change
- [x] Bug fix

## Changes Made

**Routing.** `select_passthrough_base_url()` now takes the request path and sends this one path to Copilot. The shape identifies Copilot on its own, so the redirect is unambiguous. It is scoped to the OpenAI fall-through — the branch that is wrong here — because every other branch reflects an upstream the caller chose with its own auth headers.

**The destination is not hardcoded.** GitHub's token exchange advertises the completions host in `endpoints.proxy`, alongside the `endpoints.api` chat host Headroom already reads. It is now recorded at the single chokepoint every exchange passes through, and preferred. Resolution order:

1. `GITHUB_COPILOT_PROXY_URL` — operator override
2. `endpoints.proxy` from the last token exchange — GitHub's own answer
3. The Copilot API URL

No I/O on the request path, and GHE deployments keep their host. This matters: it means the destination is not an assumption about which host serves completions, and if it is wrong for a given network it is an env var rather than a release.

**Path preservation.** `build_copilot_upstream_url()` strips `/v1` when the upstream is a Copilot host, because Copilot serves its OpenAI-compatible surface unprefixed (`/chat/completions`, `/models`). But the extension built `/v1/engines/<engine>/completions` itself, so that path is already exactly what Copilot serves — stripping the prefix rewrites a working request into a 404. Preserved, the same carve-out `/v1/messages` needed in #2409. The rule: strip only for clients speaking generic-OpenAI at Copilot, never for Copilot's own paths.

## Testing

- [x] New suite: `tests/test_copilot_vscode_completions_routing.py` (30 tests) — path recognition and its near-misses, upstream selection, the `endpoints.proxy` resolution order, and URL construction in both directions
- [x] 286 passed across the Copilot, provider-routing and passthrough suites
- [x] Ruff check and format pass

### Real Behavior Proof

Environment: this branch, a `POST /v1/engines/gpt-41-copilot/completions` driven through the real app with `OPENAI_API_URL=https://api.openai.com` and the outbound HTTP client captured.

```
BEFORE (main):  https://api.openai.com/v1/engines/gpt-41-copilot/completions
AFTER  (this):  https://api.githubcopilot.com/v1/engines/gpt-41-copilot/completions
```

The "before" line reproduces the reported URL exactly.

**Not tested:** a live VS Code Copilot session confirming GitHub accepts the forwarded request. That needs a real Copilot account and editor. If the completions host turns out to differ, the `endpoints.proxy` lookup or `GITHUB_COPILOT_PROXY_URL` covers it without a code change.

## Review Readiness

### DIFF
--- a/headroom/copilot_auth.py
+++ b/headroom/copilot_auth.py
@@ -245,6 +245,65 @@ def _configured_api_url() -> str:
     return DEFAULT_API_URL
 
 
+def copilot_api_url() -> str:
+    """Return the configured Copilot API base URL without any network calls.
+
+    Resolves ``GITHUB_COPILOT_API_URL``, then the configured enterprise domain,
+    then ``api.githubcopilot.com``. Unlike :func:`resolve_copilot_api_url` this
+    performs no token exchange, so it is safe to call while routing a request.
+    """
+
+    return _configured_api_url()
+
+
+# GitHub's token exchange advertises the host that serves inline completions
+# under ``endpoints.proxy``, alongside the ``endpoints.api`` chat host. It is
+# recorded here when observed so completions routing uses GitHub's own answer
+# instead of an assumption about which host serves that endpoint (#3076).
+_observed_completions_base_url: str | None = None
+
+
+def _remember_completions_endpoint(payload: Any) -> None:
+    """Record the completions host advertised by a token-exchange payload."""
+
+    global _observed_completions_base_url
+    endpoints = payload.get("endpoints") if isinstance(payload, dict) else None
+    proxy_url = endpoints.get("proxy") if isinstance(endpoints, dict) else None
+    if isinstance(proxy_url, str) and proxy_url.strip():
+        _observed_completions_base_url = proxy_url.strip().rstrip("/")
+
+
+def reset_observed_completions_endpoint() -> None:
+    """Forget the advertised completions host (test isolation)."""
+
+    global _observed_completions_base_url
+    _observed_completions_base_url = None
+
+
+def copilot_completions_base_url() -> str:
+    """Return the base URL serving Copilot's inline-completions endpoint.
+
+    Resolution order, most authoritative first:
+
+    1. ``GITHUB_COPILOT_PROXY_URL`` — an explicit operator override, so a
+       network that fronts Copilot behind its own gateway (or a GitHub change
+       to this endpoint) is a config edit rather than a code change.
+    2. ``endpoints.proxy`` from the last Copilot token exchange — GitHub
+       telling us directly where completions go.
+    3. The Copilot API URL, which is where GitHub's consolidated surface
+       serves them.
+
+    Never performs I/O; step 2 only reads what a previous exchange recorded.
+    """
+
+    override = os.environ.get("GITHUB_COPILOT_PROXY_URL", "").strip()
+    if override:
+        return override.rstrip("/")
+    if _observed_completions_base_url:
+        return _observed_completions_base_url
+    return copilot_api_url()
+
+
 def _github_oauth_domain(domain: str | None = None) -> str:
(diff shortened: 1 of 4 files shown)


chopratejas:
@JerrettDavis - can you check if this works? 

Thanks :)

JerrettDavis:
@chopratejas Will do! Pulling and confirming.

PR #3096, opened 2026-08-18
fix(proxy): stop headroom logger from suppressing propagation to stdout/stderr

### LINKED ISSUE #3087: [BUG] Application logs never reach stdout: headroom logger sets propagate=False, silently losing all logs in containerized deployments
## Description

`_setup_file_logging()` sets `propagate = False` on the `headroom` logger, so **no `headroom.*` log record ever reaches stdout/stderr**. Every application log goes only to `~/.headroom/logs/proxy.log`.

On a workstation that's invisible but harmless. In any containerized deployment it is a silent, total loss of application logs: Docker, Kubernetes, and Azure Container Apps all collect stdout/stderr, and `~/.headroom/logs/proxy.log` is inside the container. The log pipeline works perfectly and still receives nothing but uvicorn's own lines.

`headroom/proxy/helpers.py:1506-1539`:

```python
        # Attach to the headroom root logger so all sub-loggers are captured.
        # Disable propagation to root to avoid duplicate writes when
        # wrap.py redirects stderr to the same log file.
        headroom_logger = logging.getLogger("headroom")
        headroom_logger.setLevel(logging.INFO)
        if not any(isinstance(h, RotatingFileHandler) for h in headroom_logger.handlers):
            headroom_logger.addHandler(handler)
        headroom_logger.propagate = False
```

Called unconditionally from `create_app()` (`headroom/proxy/server.py:2508`), so it applies to every proxy process, container or not.

The de-duplication motive is sound — the problem is that the fix is unconditional. It optimizes for the `headroom wrap` case (stderr redirected to the same file) and silently breaks the deployed-server case, where stdout *is* the log channel.

This also defeats `logging.basicConfig(level=logging.INFO)` at `headroom/proxy/server.py:469`, which looks like it configures application logging but cannot, because propagation is severed afterwards.

## To Reproduce

1. Run the proxy in any environment where you read stdout/stderr rather than the container filesystem:

```bash
docker run --rm -p 8787:8787 ghcr.io/headroomlabs-ai/headroom:latest
```

2. Send traffic that provokes an application-level log — e.g. point the upstream at a closed port so the streaming path logs a connection error, or let a compression timeout fire.
3. Read the container's logs: `docker logs <id>`.

Minimal local equivalent (no Docker), which is how I found it:

```python
import os, sys
os.environ["HEADROOM_LOG_LEVEL"] = "info"   # PR #3083; use log_level="info" on older builds
import uvicorn
from headroom.proxy.models import ProxyConfig
from headroom.proxy.server import create_app

# upstream pinned to a closed port -> real httpx.ConnectError -> logger.error(...)
config = ProxyConfig(host="127.0.0.1", port=8799,
                     anthropic_api_url="http://127.0.0.1:59999",
                     retry_enabled=True, retry_max_attempts=2,
                     retry_base_delay_ms=0, retry_max_delay_ms=0)
uvicorn.run(create_app(config), host="127.0.0.1", port=8799, log_level="info")
```

```bash
curl -s http://127.0.0.1:8799/v1/messages \
  -H 'content-type: application/json' -H 'x-api-key: k' -H 'anthropic-version: 2023-06-01' \
  -d '{"model":"claude-opus-5","max_tokens":64,"stream":true,"messages":[{"role":"user","content":"hi"}]}'
```

## Expected Behavior

Application logs appear on stdout/stderr, the standard log channel for a server process, so container platforms collect them. The file handler is an addition, not a replacement.

## Actual Behavior

stdout contains only uvicorn's own lines — **zero** `headroom.*` records:

```text
INFO:     Started server process [45940]
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8799 (Press CTRL+C to quit)
INFO:     127.0.0.1:62472 - "POST /v1/messages HTTP/1.1" 502 Bad Gateway
```

The records exist — they went to the file instead:

```text
$ grep "Connection error to upstream API" ~/.headroom/logs/proxy.log
2026-08-17 14:48:21,432 - headroom.proxy - WARNING - [hr_1786970896_000001] Connection error to upstream API (attempt 1/2): ConnectError('All connection attempts failed'); retrying in 0ms
2026-08-17 14:48:23,505 - headroom.proxy - ERROR - [hr_1786970896_000001] Connection error to upstream API: All connection attempts failed
```

Confirmed the cause is `propagate`, not uvicorn's `dictConfig` (my first suspicion — it is innocent; it has no `root` key and `disable_existing_loggers: False`, and root handlers survive it intact):

```text
BEFORE dictConfig: root.handlers = [<StreamHandler <stderr>>] level = 20
headroom.proxy - ERROR - MARKER-BEFORE: this should be visible
AFTER dictConfig:  root.handlers = [<StreamHandler <stderr>>] level = 20
headroom.proxy - ERROR - MARKER-AFTER: is this visible?      <- still visible
```

Logging only goes quiet once `_setup_file_logging()` sets `propagate = False`.

## Production impact

Observed on a self-hosted deployment (Azure Container Apps → Log Analytics) running `v0.35.0`. Over **3.5 days of uptime the container emitted zero application log rows** — only the boot banner, which is `print()`-based and therefore unaffected.

That deployment hit a compression timeout during the window. `/metrics` proved the code path ran:

```text
headroom_compression_failed_total{reason="timeout"} 1
headroom_compression_quarantine_total{event="activated"} 1
```

`handlers/anthropic.py` logs `logger.warning("[%s] Optimization failed: ...")` on exactly that path — so the warning was emitted and never reached the log pipeline. Diagnosing the incident required reading `/health` and `/metrics` counters, because the logs that would have explained it in one line were unreachable. The ingestion pipeline was healthy throughout.

Worth noting the failure mode is *silent*: `docker logs` returns something (uvicorn's lines), so it looks like logging works.

## Error Output

N/A — the defect is the absence of output.

## Environment

- **Headroom version**: `0.36.0-dev` (repro, `main`); production observation on `v0.35.0`
- **Python version**: 3.13.15 (local repro); 3.12 in the container image
- **OS**: Windows 11 (local repro); Linux container on Azure Container Apps (production)
- **LLM Provider**: Anthropic

## Additional Context

Possible directions, in rough order of preference — happy to send a PR for whichever you prefer:

1. **Keep propagation, dedupe at the source.** Only sever propagation when `wrap.py` has actually redirected stderr to the same file, since that is the case the comment describes. A module-level flag set by `wrap.py`, or comparing the stream handler's target to the log path, both work.
2. **Make it configurable.** e.g. `HEADROOM_LOG_TO_STDOUT=1` (or `HEADROOM_LOG_FILE=` disabling the file handler and restoring propagation). Least invasive, but leaves the container default broken.
3. **Detect containerized runtime** (`/.dockerenv`, `KUBERNETES_SERVICE_HOST`, no TTY) and keep propagation on there. Convenient, but heuristics tend to rot.

Whichever way it goes, `logging.basicConfig()` at `server.py:469` is currently misleading and could carry a comment noting propagation is severed later.

Related: #3083 adds `headroom_upstream_connection_errors_total` partly because a log line could not be relied on for this path. Not a duplicate — that PR does not touch this, and I did not fold a fix in because the right approach is a maintainer call.

connectsudhindra-gif:
## Description

`_setup_file_logging()` in `headroom/proxy/helpers.py` unconditionally set `headroom_logger.propagate = False`, on the theory that `wrap.py` redirects stderr into the same `proxy.log` file the rotating handler writes to, so propagation would double-write records. That premise is wrong: `wrap.py` redirects the proxy subprocess's stdout/stderr to a separate sibling file, `proxy-stdio.log` (see `_get_proxy_stdio_log_path()`), never `proxy.log`. Severing propagation bought no deduplication — it only stopped every `headroom.*` log record from reaching root's `logging.basicConfig()` handler (`proxy/server.py`), which is what writes to stderr/stdout. In containerized deployments (Docker/Kubernetes), stdout/stderr is the only channel the platform collects, so this silently dropped all application logs.

Closes #3087

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Removed `headroom_logger.propagate = False` from `_setup_file_logging()` in `headroom/proxy/helpers.py`.
- Updated the surrounding comment to explain why propagation is safe to leave on: `proxy.log` (file handler) and stderr/stdout (root's `basicConfig` handler) are two distinct destinations, so nothing is duplicated in either file.

## Testing

### Test Output

```text
$ ruff check .
All checks passed!

$ pytest tests/test_corrupt_golden_bytes_recovery.py -q
collected 9 items
tests/test_corrupt_golden_bytes_recovery.py .........                    [100%]
9 passed in 0.47s
```

`mypy headroom` was not run to a clean result: it fails before reaching this file with a pre-existing, unrelated error in the installed `numpy` stub (`Type statement is only supported in Python 3.12 and greater`) in this sandbox's dependency set. Not caused by this change (a 4-line pure-Python logging tweak).

The broader `pytest tests/` collection also hits ~170 pre-existing, unrelated errors in this sandbox: `headroom._core`, the compiled Rust extension, isn't built here (no `cargo`/`maturin` available), so every module importing it fails at collection. This change doesn't touch Rust code or `_core` and isn't implicated.

No new tests added: this restores an existing, already-tested code path (see `tests/conftest.py`'s autouse fixture and `tests/test_corrupt_golden_bytes_recovery.py`, both of which already assert/require `propagate=True` for `caplog` to work) to its intended default rather than adding new behavior.

## Real Behavior Proof

- Environment: local sandbox, Python 3.12.3, unable to build/import `headroom.proxy.server` directly (missing compiled `headroom._core`). Verified with an isolated script that copies `_setup_file_logging()`'s exact logic byte-for-byte plus the same `logging.basicConfig()` call `proxy/server.py` makes at import time.
- Exact command / steps: ran the fixed logic, then emitted `logging.getLogger("headroom.proxy").info("compression timeout: request abc123 exceeded 30s budget")`.
- Observed result: the `headroom.proxy` record was written to stderr AND to `proxy.log` simultaneously (full output below); re-running the same script with the old `headroom_logger.propagate = False` line restored reproduces the bug — the record disappears from stderr entirely and only lands in `proxy.log`.
  ```text
  propagate flag on 'headroom' logger: True
  --- emitting a headroom.proxy log record, watch for it below on stderr ---
  2026-08-17 21:17:39,545 - headroom.proxy - INFO - compression timeout: request abc123 exceeded 30s budget
  --- also present in proxy.log (file handler): OK ---
  ```
- Not tested: end-to-end through the actual compiled proxy binary/server process (blocked by missing `headroom._core` in this sandbox) and real container log collection (Docker/Kubernetes) — the isolated repro exercises the exact same logging configuration code the real process runs, so this is considered sufficient proxy proof for a pure logging-config change.

## Runtime Rollout Safety

- Rollout-managed feature(s): None — this is unconditional startup logging configuration, not behind a rollout flag.
- Stable/default behavior changed: Yes — `headroom.*` log records now also reach stderr/stdout (previously file-only). No change to request/response handling, compression, or any user-facing proxy behavior.
- Kill switch / disable path: None needed; this restores default Python logging propagation. If a deployment truly wants file-only logging, `HEADROOM_LOG_TO_STDOUT`-style suppression could be added later, but no such request exists today.
- Qualification impact: None
- Rollback path: Revert this commit to restore `propagate = False`.

## Review Readiness

## Checklist

N/A — logging configuration change, no UI impact.

## Additional Notes

- "New and existing unit tests pass locally" is left unchecked because the full `pytest tests/` collection can't complete in this sandbox (missing compiled `headroom._core` Rust extension, unrelated to this change) — see Test Output above for what was actually run and passed.
- "I have added tests" is left unchecked: existing tests (`tests/conftest.py` autouse fixture, `tests/test_corrupt_golden_bytes_recovery.py`) already assert the code path this restores; happy to add a focused regression test asserting `logging.getLogger("headroom").propagate is True` after `_setup_file_logging()` if maintainers want one.


JerrettDavis (approved):
The diagnosis matches Python logging behavior and the repository's separate proxy.log versus proxy-stdio.log destinations. Leaving propagation at its default restores container-visible stderr logging while retaining the dedicated rotating file handler; the handler guard still prevents repeated file-handler attachment. Looks good.

connectsudhindra-gif:
@JerrettDavis all good to merge ? Pls let me know if you need any adjustment thanks

connectsudhindra-gif:
@chopratejas and @DevanshiVyas  pls review the PR and merge. Let me know if you need any adjustment

connectsudhindra-gif:
@chopratejas  and @DevanshiVyas pls review and merge

JerrettDavis (approved):
Re-reviewed the refreshed head after merging current main. The PR-specific logic is unchanged from the previously approved revision, the complete hosted matrix is green, and the branch is current and conflict-free.

connectsudhindra-gif:
@JerrettDavis  ready to merge ?

PR #3097, opened 2026-08-18
fix(cli): honor HEADROOM_PORT in install apply and deploy commands

### LINKED ISSUE #3072: install: HEADROOM_PORT ignored by apply/deploy, and install status contradicts doctor
## Summary

Two install/CLI defects were reported as secondary observations inside #3040, whose headline bug is the CCR empty-HTTP-200 family. #3040 is being closed as a duplicate of #2952, so these are split out here to keep them from being lost — they are unrelated to the CCR work and neither is fixed.

Both reported by @vsg-prog in #3040.

## Bug 1 — `install apply` / `deploy` ignore an explicit `HEADROOM_PORT`

> `headroom install apply --preset persistent-service` and `headroom deploy` ignored an explicit `HEADROOM_PORT=8788` override and always configured port 8787.

Expected: an explicit `HEADROOM_PORT` is honoured by the install/deploy path the same way `headroom proxy --port` honours it.
Actual: the generated service configuration is hardcoded to 8787.

Impact: anyone running a second instance, or avoiding a port conflict, gets a silently wrong configuration. It is also a confusing failure, because `--port` works on the direct path, so the override appears supported.

## Bug 2 — `install status` contradicts itself and `doctor`

> `headroom install status` reported `Status: stopped` while simultaneously reporting `Healthy: yes`, and while `headroom doctor` showed the proxy as pass/running.

Expected: one truth about whether the proxy is running, consistent between `install status` and `doctor`.
Actual: `install status` prints a self-contradictory pair, and disagrees with `doctor` on the same machine at the same moment.

Impact: `status` is the command people run to decide whether something is wrong. Reporting `stopped` and `Healthy: yes` together trains people to distrust it, and it will send them debugging a proxy that is actually fine. Related in spirit to #2925, where `doctor` reported a pass that did not reflect reality.

## Notes

The reporter offered full `doctor` output, a `/stats` dump, and a minimal repro with `--log-messages`. Worth taking them up on it for bug 2, since the disagreement between the two commands is the interesting part.

connectsudhindra-gif:
## Description

`headroom proxy --port` already honors `HEADROOM_PORT` via Click's `envvar` support (see `headroom/cli/proxy.py`), but the `--port` options on `headroom install apply --preset persistent-service` and `headroom deploy` did not declare that `envvar`, so an explicit `HEADROOM_PORT=8788` was silently ignored — both commands always fell back to the hardcoded `8787` default and generated a persistent-service manifest bound to the wrong port. Users running multiple instances or avoiding a port conflict got a silently wrong config with no error.

This addresses bug 1 of the two independent bugs reported in #3072. Bug 2 (`install status` reporting "stopped" while `doctor` reports healthy) has a different root cause — `runtime_status()` in `headroom/install/runtime.py` only tracks liveness via a locally-written PID file, which is never written for supervisor-managed presets (systemd/launchd/Task Scheduler start the process directly) — and needs its own fix querying the actual supervisor, which is out of scope here.

Related to #3072

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `envvar="HEADROOM_PORT"` to the `--port` option on `headroom install apply` (`headroom/cli/install.py`).
- Added `envvar="HEADROOM_PORT"` to the `--port` option on `headroom deploy` (`headroom/cli/install.py`).
- Updated both options' `help=` text to note the env var, matching the existing `headroom proxy --port` help string.

## Testing

### Test Output

```text
$ ruff check .
All checks passed!

$ pytest tests/test_cli/test_install_cli.py -q
collected 37 items
tests/test_cli/test_install_cli.py .....................................  [100%]
37 passed in 0.28s
```

`mypy headroom` was not run to a clean result in this sandbox: it fails before reaching this file on a pre-existing, unrelated error in the installed `numpy` stub (`Type statement is only supported in Python 3.12 and greater`). Not caused by this change (a Click option kwarg addition).

No new tests added: the existing 37 `test_install_cli.py` tests already exercise `install apply`/`deploy` option parsing and all continue to pass; a small `envvar=` kwarg addition to two existing options doesn't warrant new test scaffolding beyond what's demonstrated live in Real Behavior Proof below. Happy to add a `CliRunner` regression test asserting `HEADROOM_PORT` resolution if maintainers want one pinned in the suite.

## Real Behavior Proof

- Environment: local sandbox, Python 3.12.3, Click 8.x, using `click.testing.CliRunner` against the exact option definition now used by both `install apply` and `deploy` (same `default=8787, type=..., envvar="HEADROOM_PORT"` kwargs).
- Exact command / steps: invoked a minimal Click command built from the same option kwargs three ways — no flag/no env, `HEADROOM_PORT=8788` env only, and `HEADROOM_PORT=8788` env plus an explicit `--port 9000` flag — then separately ran `install apply --help` / `deploy --help` through the real CLI to confirm the env var now appears in `--help` output.
- Observed result: `default:` printed `port=8787`; `env only:` printed `port=8788` (the fix — previously this printed `port=8787`, ignoring the env var entirely); `env + explicit flag:` printed `port=9000`, confirming an explicit `--port` still wins over `HEADROOM_PORT` exactly like `headroom proxy --port` does. `install apply --help` and `deploy --help` both now show `HEADROOM_PORT` in the port option's help text (`grep` confirmed `True`/`True`).
- Not tested: the full `install apply`/`deploy` command end-to-end (writing an actual systemd/launchd unit or Windows Task and starting it) — that requires a real supervisor (systemd/launchd/schtasks) and elevated permissions not available in this sandbox. The port value is proven to flow unchanged from the Click option into `_build_deployment_manifest(port=port, ...)` → `build_manifest(port=port, ...)` (read directly in `headroom/cli/install.py`), which is the same `port` field `headroom/install/planner.py` writes into `HEADROOM_PORT` in the generated service env — so the isolated proof above covers the actual code path, just not a live supervisor invocation.

## Runtime Rollout Safety

- Rollout-managed feature(s): None — this is CLI option parsing for a one-time install/deploy command, not a proxy runtime behavior behind a rollout flag.
- Stable/default behavior changed: Yes, but only when `HEADROOM_PORT` is exported and no explicit `--port` is passed — previously ignored, now honored (matches `headroom proxy --port`'s existing, documented behavior). No change to default behavior when `HEADROOM_PORT` is unset.
- Kill switch / disable path: None needed — pass `--port` explicitly to override, exactly as today.
- Qualification impact: None
- Rollback path: Revert this commit to restore the previous (env-ignoring) `--port` options.

## Review Readiness

## Checklist

N/A — CLI option change, no UI impact.

## Additional Notes

- "New and existing unit tests pass locally" and "I have added tests" are left unchecked for the same reasons detailed in Testing above: the full `pytest tests/` collection can't complete in this sandbox (missing compiled `headroom._core` Rust extension, unrelated to this change), and no new tests were added — see Testing and Real Behavior Proof for what was actually verified.
- This PR intentionally does not close #3072 — it only fixes bug 1 (`HEADROOM_PORT` ignored). Bug 2 (`install status` vs `doctor` disagreement) needs a separate PR that teaches `runtime_status()` to query the actual supervisor (systemd/launchd/Task Scheduler) instead of relying solely on a PID file that supervisor-managed presets never write.

### DIFF
--- a/headroom/cli/install.py
+++ b/headroom/cli/install.py
@@ -506,8 +506,9 @@ def _echo_installed(manifest: DeploymentManifest, *, prefix: str = "Installed pe
     "-p",
     default=8787,
     type=click.IntRange(1, 65535),
+    envvar="HEADROOM_PORT",
     show_default=True,
-    help="Persistent proxy port.",
+    help="Persistent proxy port. (env: HEADROOM_PORT)",
 )
 @click.option(
     "--backend",
@@ -682,7 +683,13 @@ def install_apply(
 @main.command("deploy")
 @click.option("--profile", default="default", show_default=True, help="Deployment profile name.")
 @click.option(
-    "--port", "-p", default=8787, type=int, show_default=True, help="Persistent proxy port."
+    "--port",
+    "-p",
+    default=8787,
+    type=int,
+    envvar="HEADROOM_PORT",
+    show_default=True,
+    help="Persistent proxy port. (env: HEADROOM_PORT)",
 )
 @click.option(
     "--backend",


JerrettDavis (changes requested):
The option wiring itself is correct, but this regression currently has no committed test: the existing install CLI tests exercise parsing generally but do not assert that HEADROOM_PORT reaches both install apply and deploy, or that an explicit --port wins. Please add focused CliRunner coverage for those two commands (env-only plus precedence); that is the durable proof this behavior will not silently regress again. The branch also currently conflicts with main and will need rebasing.

connectsudhindra-gif:
Closing as superseded: while rebasing to address the review feedback (add CliRunner coverage for `HEADROOM_PORT` resolution + explicit `--port` precedence on both `install apply` and `deploy`), the rebase conflict on `headroom/cli/install.py` surfaced that #3085 already merged the identical `envvar="HEADROOM_PORT"` fix on both commands, plus exactly the requested test coverage (`test_install_apply_honors_headroom_port_env`, `test_install_apply_explicit_port_overrides_env`, `test_deploy_honors_headroom_port_env` in `tests/test_cli/test_install_cli.py`).

#3085 was opened before this PR and merged shortly after, so neither PR saw the other coming. Diffing this branch against current `main` confirms the remaining delta is purely cosmetic (kwarg ordering, help-string punctuation) — no functional difference. Closing this one rather than pushing a no-op rebase.

PR #3100, opened 2026-08-18
fix(proxy): add --no-ccr-response-handling flag to bypass response buffering (#3082)

### LINKED ISSUE #3082: [BUG] `headroom wrap opencode`'s bundled plugin sends `headroom_retrieve` unconditionally — `--no-ccr`/`HEADROOM_NO_CCR` don't stop the CCR buffered-stream empty-response/api_error bug for OpenCode
**Environment**
- headroom-ai 0.35.0 (installed via `uv tool install headroom-ai[all]`)
- OpenCode 1.16.0 (`opencode` CLI, launched via `headroom wrap opencode`)
- Linux (WSL2), single local proxy instance
- Provider: `github-copilot` (Anthropic-compatible passthrough), model `claude-sonnet-5`

**Symptom**

Same family as #2952 / #3071 / #3079 / #3017: the turn immediately after a large tool-result (e.g. an agent loading an ~80k-token skill/context file via a tool call) intermittently returns either a silently empty completion (`tok_out=0`) or a generic client-visible error:

```
{"type":"api_error","message":"An error occurred while processing the request."}
```

confirmed in `proxy.log` at the exact same request as:

```
CCR: stream:true request has headroom_retrieve available; using buffered stream:false upstream request for server-side retrieval handling
```

**What's different from the existing CCR issues (why this needed a new report)**

I initially assumed this was the same bug covered by #2952/#3079/#3071/#3017, and tried the maintainer-recommended workaround from those threads (`HEADROOM_NO_CCR=1` / `headroom proxy --no-ccr`, and `--lossless`). None of them stop this for `headroom wrap opencode` specifically. Tracing the code:

1. The bundled OpenCode plugin (`headroom/providers/opencode/_dist/entry.opencode.js`, built from the `plugins/opencode` package) registers `headroom_retrieve` **unconditionally** from `HeadroomPlugin`'s return value:
   ```js
   const retrieveTool = createHeadroomRetrieveTool({ proxyBaseUrl: proxyUrl });
   ...
   return {
     ...
     tool: {
       headroom_retrieve: tool({ description: retrieveTool.description, ... })
     }
   };
   ```
   There is no plugin option / env var read anywhere in the bundle to suppress this. It is not gated by anything the proxy or `wrap` sets.

2. On the proxy side (`headroom/proxy/handlers/anthropic.py`), the buffered-stream conversion is gated by **two** independent conditions:
   ```python
   ccr_handler_config = getattr(self.ccr_response_handler, "config", None)
   ccr_response_handler_enabled = bool(
       self.ccr_response_handler and getattr(ccr_handler_config, "enabled", True)
   )
   buffered_stream_ccr = bool(
       stream
       and ccr_response_handler_enabled
       and self._has_headroom_retrieve_tool(tools if tools is not None else body.get("tools"))
   )
   ```
   `self.ccr_response_handler` is only `None` (i.e. `ccr_response_handler_enabled` False) when `config.ccr_handle_responses` is `False` (`headroom/proxy/server.py:1165-1174`).

3. `config.ccr_handle_responses` defaults to `True` (`headroom/proxy/models.py:181: ccr_handle_responses: bool = True`) and **has no CLI flag or environment variable anywhere in the codebase** — I grepped the full pinned source and `cli/proxy.py` only wires `no_ccr` to `ccr_inject_tool` / `ccr_inject_marker` (`cli/proxy.py` ~1206-1211, comment: *"CCR opt-out: --no-ccr disables both halves at once (markers in content AND the injected retrieve tool)"*) — this only controls whether **the proxy itself** injects the marker/tool into a request that didn't already have one. It never touches `ccr_handle_responses`.

Since the OpenCode plugin already puts `headroom_retrieve` in every request's tools array **independently of the proxy's own injection**, `_has_headroom_retrieve_tool()` is always `True` for OpenCode sessions, and `ccr_response_handler_enabled` is always `True` too (nothing sets `ccr_handle_responses=False`), so `buffered_stream_ccr` is always `True` — regardless of `--no-ccr` / `HEADROOM_NO_CCR` / `--lossless`.

`headroom wrap opencode --no-mcp` also does not help: `headroom_retrieve` here comes from the bundled plugin's own `tool:` hook, not from the separate "headroom" MCP server that `--no-mcp` controls (`_setup_headroom_mcp` in `cli/wrap.py`).

**Reproduction**

1. `headroom wrap opencode --port 8797 -- run --agent <any-agent-that-calls-a-tool-with-a-large-result> "<prompt>"`
2. Confirm in the OpenCode log that `headroom_retrieve` is present in `service=tool.registry`.
3. Send a prompt that makes the agent call a tool returning a very large result (~80k+ tokens cached) in turn 1, then let it proceed to turn 2.
4. `proxy.log` shows `CCR: stream:true request has headroom_retrieve available; using buffered stream:false upstream request for server-side retrieval handling` for turn 2, and the OpenCode session log shows either `tok_out=0` or:
   ```
   ERROR ... service=llm providerID=github-copilot modelID=claude-sonnet-5 ... error={"error":{"type":"api_error","message":"An error occurred while processing the request."}} stream error
   ```
5. Confirmed this is **not** fixed by any of:
   - `HEADROOM_NO_CCR=1` exported before `headroom wrap opencode` (env correctly propagates to the `headroom proxy` subprocess `wrap` spawns via `subprocess.Popen(cmd, env=proxy_env, ...)`, `proxy_env = os.environ.copy()` — verified this is not scrubbed).
   - Manually starting `headroom proxy --port 8797 --no-ccr` standalone, then `headroom wrap opencode --port 8797 --no-proxy -- ...` to attach `wrap` to that CCR-disabled proxy instance instead of letting `wrap` start its own. Confirmed via `ps`/`ss` that this externally-started `--no-ccr` proxy was the only proxy process alive and handling the request — CCR still triggered on it.
   - `--no-mcp` on `wrap opencode` (removes the separate "headroom" MCP entry, unrelated to the plugin's own tool).

**Suggested fix**

- Expose `ccr_handle_responses` via a CLI flag / env var (e.g. have `--no-ccr` / `HEADROOM_NO_CCR` also set `ccr_handle_responses=False`, not just `ccr_inject_tool` / `ccr_inject_marker`), so the response-handler-side buffered-stream conversion can actually be disabled independently of who put the tool in the request.
- And/or make the bundled OpenCode plugin's `headroom_retrieve` registration conditional on the same setting (e.g. `HEADROOM_NO_CCR` read client-side too, or a new `headroom wrap opencode --no-ccr` flag that's actually forwarded to both the proxy subprocess *and* the injected plugin options), so users who need to opt out of CCR for OpenCode have a supported way to do it without hand-patching the installed plugin bundle.

**Related**

- #2952 (canonical thread for the "buffered stream discards the flip → empty/malformed response" family; fixed there, but not via a CCR-disable path — see below)
- #3079 (heartbeat regression in the buffered path, still open/unfixed)
- #3071, #3017 (open reports of the same symptom)
- #2410 (closed, earlier OpenCode-specific instance of the empty-response symptom)

I ended up mitigating this locally by patching the installed `entry.opencode.js` bundle to stop registering `headroom_retrieve`, since none of the documented/official CCR opt-outs actually prevent this for `headroom wrap opencode`. Happy to share the exact patch or a minimal repro repo if useful.

glatinone:
## Description

`ccr_handle_responses` (`headroom/proxy/models.py`) gates whether the proxy buffers a streaming response server-side to run CCR retrieval-marker handling on it (`headroom/proxy/buffered_ccr_response.py`). It had no CLI or environment-variable wiring at all, and `--no-ccr`/`HEADROOM_NO_CCR` only ever toggled `ccr_inject_tool`/`ccr_inject_marker` (see the comment above that call site in `headroom/cli/proxy.py`). Any client that offers the `headroom_retrieve` tool itself — e.g. the bundled OpenCode plugin — unconditionally takes the buffered-stream path with no supported opt-out, which is the root cause of #3082 and the underlying issue behind the buffered-CCR bug family the team has been patching (#3091, #3092, #3094 landed in the ~24h before this PR).

This PR adds `--no-ccr-response-handling` / `HEADROOM_NO_CCR_RESPONSE_HANDLING`, wired independently of `--no-ccr`, so operators get an actual kill switch instead of another point patch on the next incident in this family.

Closes #3082

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cli/proxy.py`: new `--no-ccr-response-handling` / `HEADROOM_NO_CCR_RESPONSE_HANDLING` click option, and `ccr_handle_responses=not no_ccr_response_handling` wired into the `ProxyConfig(...)` construction, right next to the other `ccr_*` fields.
- `tests/test_cli_proxy_env.py`: new `TestCLINoCcrResponseHandlingFlag` class (4 tests) — default stays `True`, the flag sets `False`, the env var sets `False`, and it composes correctly alongside `--no-ccr` without either implying the other.

## Testing

- [x] Linting passes (`ruff format --check .`)

### Test Output

```text
$ ruff check .
All checks passed!

$ ruff format --check .
1510 files already formatted
```

`pytest tests/test_cli_proxy_env.py` could not be run in my dev environment — see "Not tested" below — but see "Real Behavior Proof" for an equivalent-coverage manual run of the actual changed code path.

## Real Behavior Proof

- Environment: Windows dev machine, no MSVC Build Tools installed (no admin rights available to install them), so `headroom._core` (the maturin/PyO3 Rust extension) could not be built. `headroom.proxy.server` transitively imports `headroom._core` via `headroom.transforms.content_router` → `error_detection`, so **every** test in `tests/test_cli_proxy_env.py` that does `patch("headroom.proxy.server.run_server", ...)` — not just the 4 I added — fails in this environment with `AttributeError: module 'headroom.proxy' has no attribute 'server'`. Confirmed this is pre-existing/environmental and not caused by this change (66 pre-existing tests in the same file fail identically on an unmodified checkout).
- Exact command / steps: Since the real `headroom.proxy.server` module can't load here, I stubbed `sys.modules["headroom.proxy.server"]` with the real `headroom.proxy.models.ProxyConfig` plus no-op fakes for `_parse_csv_tools` / `_parse_exclude_tools` / `_parse_tool_profiles` / `run_server` / `_get_code_aware_banner_status` (the only other names `headroom/cli/proxy.py`'s `proxy` command lazily imports from that module), then drove the real, unmodified `proxy` click command through `CliRunner.invoke(main, ["proxy", ...])`, exactly like the existing test suite does — same code path, same assertions, just without needing the native build.
- Observed result: `--no-ccr-response-handling` and `HEADROOM_NO_CCR_RESPONSE_HANDLING=1` both produce a `ProxyConfig` with `ccr_handle_responses=False` while leaving `ccr_inject_tool`/`ccr_inject_marker` at their defaults; passing neither leaves `ccr_handle_responses=True`; combining `--no-ccr` with `--no-ccr-response-handling` sets all three fields `False`; `--no-ccr` alone does **not** set `ccr_handle_responses=False` (confirms the two switches are independent, as intended). Also confirmed `--help` lists the new flag with `HEADROOM_NO_CCR_RESPONSE_HANDLING` documented.
- Not tested: The actual buffered-response code path in `headroom/proxy/buffered_ccr_response.py` / `server.py` that reads `config.ccr_handle_responses` at runtime — I could not build `headroom._core` to exercise the real proxy server end-to-end. This PR only changes how the flag is populated (CLI/env → `ProxyConfig`); the field itself and its runtime consumer (`server.py:1214`, `server.py:1883`) are unchanged.

## Runtime Rollout Safety

- Rollout-managed feature(s): None — this is a CLI/env flag, not a rollout-gated feature.
- Stable/default behavior changed: No. `ccr_handle_responses` defaults to `True` exactly as before; the new flag is opt-in and off by default.
- Kill switch / disable path: This PR *is* the kill switch — `--no-ccr-response-handling` / `HEADROOM_NO_CCR_RESPONSE_HANDLING=1` fully disables the buffered-response path.
- Qualification impact: None.
- Rollback path: Revert the two touched files; no state/schema/config migration involved.

## Review Readiness

## Checklist

## Additional Notes

- "Unit tests pass locally" and "type checking passes" are unchecked because I could not build the Rust extension in this environment (see Real Behavior Proof) or run `mypy` locally; CI builds the extension and should be the authoritative signal for both. Happy to address any failures CI surfaces.
- No docs currently enumerate `--no-ccr*` flags individually (e.g. README doesn't list `--no-ccr-proactive-expansion` either), so I didn't add a doc section for symmetry with the existing flags — let me know if you'd like one added.

### DIFF
--- a/headroom/cli/proxy.py
+++ b/headroom/cli/proxy.py
@@ -419,6 +419,18 @@ def dashboard(port: int, no_open: bool) -> None:
     ),
 )
 @click.option(
+    "--no-ccr-response-handling",
+    is_flag=True,
+    envvar="HEADROOM_NO_CCR_RESPONSE_HANDLING",
+    help=(
+        "Disable server-side buffering of streaming responses for CCR response "
+        "handling, independent of --no-ccr. A client that offers the "
+        "headroom_retrieve tool itself (e.g. the bundled OpenCode plugin) "
+        "otherwise always takes the buffered-stream path with no opt-out "
+        "(issue #3082). Env: HEADROOM_NO_CCR_RESPONSE_HANDLING."
+    ),
+)
+@click.option(
     "--proxy-extension",
     "proxy_extension",
     multiple=True,
@@ -1033,6 +1045,7 @@ def proxy(
     lossless: bool,
     ccr_inline_resolve: bool,
     no_ccr_proactive_expansion: bool,
+    no_ccr_response_handling: bool,
     proxy_extension: tuple[str, ...],
     compressor: tuple[str, ...],
     no_subscription_tracking: bool,
@@ -1339,6 +1352,10 @@ def proxy(
         ccr_resolve_markers_inline=ccr_inline_resolve,
         lossless=lossless,
         ccr_proactive_expansion=not no_ccr_proactive_expansion,
+        # Independent of --no-ccr: a client can offer headroom_retrieve itself
+        # (see the --no-ccr-response-handling help text), which unconditionally
+        # takes the buffered-stream path with no other opt-out (issue #3082).
+        ccr_handle_responses=not no_ccr_response_handling,
         # Flatten repeat-flag tuple AND any comma-separated values inside it.
         # `--proxy-extension a,b --proxy-extension c` and `HEADROOM_PROXY_EXTENSIONS=a,b,c`
         # both yield ["a", "b", "c"]. None when nothing was supplied.
--- a/tests/test_cli_proxy_env.py
+++ b/tests/test_cli_proxy_env.py
@@ -960,6 +960,89 @@ class TestCLICompressionOnlyFlags:
         assert cfg.ccr_inject_tool is False
 
 
+class TestCLINoCcrResponseHandlingFlag:
+    """--no-ccr-response-handling / HEADROOM_NO_CCR_RESPONSE_HANDLING must flip
+    ccr_handle_responses, independent of --no-ccr (issue #3082).
+
+    Before this flag existed, ccr_handle_responses had no CLI/env wiring at
+    all — a client that offers the headroom_retrieve tool itself (e.g. the
+    bundled OpenCode plugin) always took the buffered-stream path with no
+    supported opt-out, even with --no-ccr set.
+    """
+
+    def test_ccr_handle_responses_defaults_on(self, runner):
+        """Without the flag, ccr_handle_responses stays enabled (no behavior change)."""
+        captured_config = {}
+
+        def mock_run_server(config, **kwargs):
+            captured_config["config"] = config
+
+        with patch("headroom.proxy.server.run_server", mock_run_server):
+            result = runner.invoke(main, ["proxy"], catch_exceptions=False)
+
+        assert result.exit_code == 0, result.output
+        assert captured_config["config"].ccr_handle_responses is True
+
+    def test_no_ccr_response_handling_flag(self, runner):
+        """--no-ccr-response-handling disables ccr_handle_responses only."""
+        captured_config = {}
+
+        def mock_run_server(config, **kwargs):
+            captured_config["config"] = config
+
+        with patch("headroom.proxy.server.run_server", mock_run_server):
+            result = runner.invoke(
+                main, ["proxy", "--no-ccr-response-handling"], catch_exceptions=False
+            )
+
+        assert result.exit_code == 0, result.output
+        cfg = captured_config["config"]
+        assert cfg.ccr_handle_responses is False
+        # Unrelated CCR knobs stay on: this flag is independent of --no-ccr.
+        assert cfg.ccr_inject_tool is True
+        assert cfg.ccr_inject_marker is True
+
+    def test_no_ccr_response_handling_from_env(self, runner):
+        """HEADROOM_NO_CCR_RESPONSE_HANDLING env var disables ccr_handle_responses."""
+        captured_config = {}
+
+        def mock_run_server(config, **kwargs):
+            captured_config["config"] = config
+
+        with patch("headroom.proxy.server.run_server", mock_run_server):
+            result = runner.invoke(
+                main,
+                ["proxy"],
+                env={"HEADROOM_NO_CCR_RESPONSE_HANDLING": "1"},
+                catch_exceptions=False,
+            )
+
+        assert result.exit_code == 0, result.output
+        assert captured_config["config"].ccr_handle_responses is False
+
+    def test_no_ccr_response_handling_combines_with_no_ccr(self, runner):
+        """--no-ccr-response-handling composes with --no-ccr instead of being
+        implied by it: both can be set together for a fully unbuffered,
+        no-retrieval-path configuration."""
+        captured_config = {}
+
+        def mock_run_server(config, **kwargs):
+            captured_config["config"] = config
+
+        with patch("headroom.proxy.server.run_server", mock_run_server):
+            result = runner.invoke(
+                main,
+                ["proxy", "--no-ccr", "--no-ccr-response-handling"],
+                catch_exceptions=False,
+            )
+
+        assert result.exit_code == 0, result.output
+        cfg = captured_config["config"]
+        assert cfg.ccr_inject_tool is False
+        assert cfg.ccr_inject_marker is False
+        assert cfg.ccr_handle_responses is False
+
+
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
The CLI/env wiring is correct and intentionally independent from --no-ccr; defaults remain unchanged, and the tests cover default, flag, env, and composition behavior. Code review looks good. The PR currently has only governance checks, so the normal CI/security suite should still run successfully before merge.

PR #3104, opened 2026-08-18
feat(memory): persist TrafficLearner pending evidence across restarts + expose learner stats

gglucass:
## Description

Pending (sub-threshold) TrafficLearner evidence lives only in `_pattern_counts`, so every proxy restart zeroes the counters. A user who accumulates 4 sightings of a pattern loses all of them on the next app update or quit, which makes `min_evidence=5` effectively unreachable for short or restart-heavy sessions - auto-learning looks dead even though it is working. (Surfaced by a beta tester who used Claude Code ~6 hours across several days behind the proxy and saw zero learnings.)

This PR persists pending evidence across restarts via a `pending_patterns.json` sidecar next to `memory.db`, and exposes learner progress in the `/stats` payload so dashboards can show that learning is alive before the first pattern crosses the threshold.

Follow-up to #2579, which bounded the same accumulator (the LRU cap is respected on load).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `TrafficLearner._save_pending_state()`: snapshot `_pattern_counts` to `pending_patterns.json` (atomic tmp+rename), called on `stop()` and on each debounced flush cycle so a crash loses at most one debounce window.
- `TrafficLearner._load_pending_state()`: rehydrate on `start()` after `_hydrate_persisted_state()`; skips entries whose hash is already persisted (re-sightings bump the existing row instead), tolerates corrupt/unreadable sidecars, and respects `max_pending_patterns`.
- `get_stats()` now includes `min_evidence` so consumers can render progress like "seen 3 of 5 times".
- `/stats` payload gains a `traffic_learner` block (`proxy.traffic_learner.get_stats()`, `null` when learning is disabled).
- 3 new tests: restart survival end to end, corrupt sidecar ignored, already-saved hash not rehydrated.

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_memory/test_traffic_learner.py
============================= 157 passed in 2.82s ==============================

$ uvx ruff check headroom/memory/traffic_learner.py headroom/proxy/server.py tests/test_memory/test_traffic_learner.py
All checks passed!
$ uvx ruff format --check headroom/memory/traffic_learner.py headroom/proxy/server.py tests/test_memory/test_traffic_learner.py
3 files already formatted

$ uv run --frozen --extra dev mypy headroom/memory/traffic_learner.py
```

## Real Behavior Proof

- Environment: macOS 15.6 (Darwin 24.6.0), Python 3.12, branch at 8895e054
- Exact command / steps: scripted two learner "processes" against the same memory.db in a temp dir: process 1 accumulates 4 sightings of one ENVIRONMENT pattern (min_evidence=5) and stops; process 2 starts fresh, then sees the pattern a 5th time.
- Observed result: process 1 stats `{'patterns_extracted': 4, 'patterns_saved': 0, 'pending_patterns': 1, 'min_evidence': 5, ...}`; sidecar after stop contains the pattern with `"count": 4`; process 2 stats after restart show `'pending_patterns': 1` before any traffic; after the 5th sighting the DB row is saved with `evidence_count == 5`.
- Not tested: the full proxy path over live agent traffic (learner unit surface only); mem0/remote backends, where `_resolve_backend_db_path` returns None and persistence is a silent no-op (pre-existing behavior for hydrate as well).

## Runtime Rollout Safety

- Rollout-managed feature(s): none - not behind a rollout flag.
- Stable/default behavior changed: pending evidence now survives restarts (the intended change); a small `pending_patterns.json` appears next to memory.db. No change when learning is disabled or no local backend is configured.
- Kill switch / disable path: disabling the traffic learner (existing toggle) stops both writes and reads; deleting the sidecar restores per-process behavior.
- Qualification impact: none - `/stats` gains one additive key.
- Rollback path: revert the commit; stale sidecars are ignored dead weight for older builds (never read).

## Review Readiness

## Checklist

### DIFF
--- a/headroom/memory/traffic_learner.py
+++ b/headroom/memory/traffic_learner.py
@@ -525,6 +525,10 @@ class TrafficLearner:
         # Hydrate persisted dedup state before workers spin up so cross-session
         # re-sightings bump existing rows instead of creating duplicates.
         await self._hydrate_persisted_state()
+        # Rehydrate sub-threshold evidence from the previous process. Without
+        # this every restart zeroed pending counts, so users with short or
+        # restart-heavy sessions could never cross min_evidence.
+        await self._load_pending_state()
         if self._save_task is None or self._save_task.done():
             self._save_task = asyncio.create_task(self._save_worker())
         if self._flush_task is None or self._flush_task.done():
@@ -569,6 +573,9 @@ class TrafficLearner:
             except Exception:
                 break
 
+        # Snapshot sub-threshold evidence so it survives the restart.
+        await self._save_pending_state()
+
         # Final flush on shutdown — bypass debounce.
         await self.flush_to_file()
 
@@ -586,6 +593,9 @@ class TrafficLearner:
                 self._flush_dirty = False
                 self._last_flush_at = time.monotonic()
                 await self.flush_to_file()
+                # Same debounce cycle also snapshots pending evidence, so a
+                # crash (vs clean stop) loses at most one debounce window.
+                await self._save_pending_state()
             except asyncio.CancelledError:
                 break
             except Exception as e:
@@ -831,6 +841,7 @@ class TrafficLearner:
             "patterns_extracted": self._patterns_extracted,
             "patterns_saved": self._patterns_saved,
             "pending_patterns": len(self._pattern_counts),
+            "min_evidence": self._min_evidence,
             "history_size": len(self._tool_history),
         }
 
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
Reviewed the lifecycle hooks, sidecar format, bounded rehydration, and restart tests. Pending state is snapshotted atomically, remains best-effort for unsupported backends/filesystems, is loaded only after persisted hashes, and the additive stats field is safe. Looks good.

gglucass:
Nudge on this one: it has been green and conflict-free since Aug 18 (22 checks passing, `MERGEABLE`/`CLEAN`, no review yet).

Context on why it matters downstream, in case it helps prioritise:

- Headroom Desktop 0.9.x already ships the consumer. It parses `traffic_learner.pending_patterns`, `min_evidence` and `patterns_saved` out of `/stats` and renders "N patterns detected; these are saved once detected 5 times" in the auto-learning panel.
- On the current release (0.37.0) that key is absent from `/stats`, so the parse returns `None` and every user falls back to static copy instead.
- Symptom this was written for, from beta feedback: a tester ran roughly six hours of Claude Code, saw zero learnings, and read the silence as the feature being broken. The evidence gate is behaving exactly as designed; there is just nothing that makes the accumulating evidence visible, and restarts used to reset it.

Happy to rebase, or to split the pending-persistence half from the stats-exposure half if a smaller diff is easier to review.

PR #1115, opened 2026-06-18
feat(learn): write per-project learnings to CLAUDE.local.md by default

MrAshRhodes:
## Description

`headroom learn` wrote per-project learnings into the project's `CLAUDE.md`, which Claude Code treats as team-shared and git-tracked. That meant machine-specific absolute paths and tool-discovery byproducts polluted the shared file for every teammate. This switches the default to the personal, gitignored `CLAUDE.local.md`, adds a `--target` override, and migrates any stale block out of `CLAUDE.md`.

Closes #1072.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `ClaudeCodeWriter` now writes CONTEXT_FILE recommendations to `CLAUDE.local.md` by default instead of `CLAUDE.md` (the home-directory case still uses `~/.claude/CLAUDE.md`, which is personal global memory).
- Added a `--target` flag (Claude Code only) and `set_context_target()` to override the destination — e.g. `--target CLAUDE.md` to opt back into the shared file, or any relative/absolute path.
- On first run after upgrade, a stale Headroom block left in `CLAUDE.md` is moved into `CLAUDE.local.md` and stripped from `CLAUDE.md`, with a warning surfaced by the CLI. If `CLAUDE.md` held nothing but the block, the empty file is removed.
- `WriteResult` carries `warnings`; the `learn` CLI prints them.
- Updated docs (`failure-learning.mdx`) and `CHANGELOG.md`.

This implements the maintainer's stated preference order from the issue (default → `CLAUDE.local.md`, plus a `--target` flag), scoped to the Claude writer only — `AGENTS.md`/`GEMINI.md` have no `.local` convention and are untouched.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ pytest tests/test_learn/ tests/test_cli_learn.py -q
196 passed, 2 skipped in 17.80s

$ ruff check headroom/learn/writer.py headroom/cli/learn.py
All checks passed!

$ mypy headroom/learn/writer.py headroom/cli/learn.py
Success: no issues found in 2 source files
```

[...]

Issue #1072, opened 2026-06-17
[FEATURE] `learn` should write per-project learnings to CLAUDE.local.md, not CLAUDE.md


im47cn:
### Context
Claude Code has an established convention:
- `CLAUDE.md` — **team-shared**, checked into git, contains project instructions for all developers
- `CLAUDE.local.md` — **personal**, gitignored by default, for individual developer overrides

Reference: https://docs.claude.com/en/docs/claude-code/memory

### Observed behavior (headroom 0.24.0, macOS, claude agent)
`headroom learn --all --apply` writes learned shell-command patterns into `<project>/CLAUDE.md`. The content looks like:

```markdown
## Headroom Learned Patterns
### Learned: environment
- Working test command: `cd /Users/xxx/sources/etf-radar/backend && .venv/bin/python -m pytest ...`
- Working test command: `cd /Users/xxx/sources/etf-radar/backend && uv run --all-extras pytest ...`
```

This content contains:
1. **Absolute paths from my machine** (`/Users/xxx/...`) — meaningless for teammates
2. **Tool-discovery byproducts** (e.g. learning that `uv run` works) — personal context, not team instruction

### Problem
Either I:
- (a) Commit `CLAUDE.md` → pollute team-shared instructions with personal absolute paths and shell archaeology
- (b) Add `CLAUDE.md` to `.gitignore` → lose the ability to use `CLAUDE.md` for its intended purpose (team-shared instructions)

Both are bad. The third option — manually editing `CLAUDE.md` after every `headroom learn` to remove the auto-generated block — defeats the point of automation.

### Proposal
Three options, in order of preference:

1. **Default to `CLAUDE.local.md`** for per-project learnings. This matches Claude Code's gitignore convention exactly: learnings are personal-by-default, opt-in to share.
2. **Add `--target` flag**: `headroom learn --target=CLAUDE.local.md` (or `CLAUDE.md`, or a custom path). Backward-compat: keep current default but document the recommendation.
3. **Detect existing hand-written `CLAUDE.md`**: if the file exists and has no `## Headroom Learned Patterns` block, write to `CLAUDE.local.md` instead.

### Related
Possibly related to but distinct from #96 (home-dir cwd handling). That issue is about `~/CLAUDE.md` vs `~/.claude/CLAUDE.md`; this one is about project-level file selection.

### Environment
- headroom 0.24.0 (PyPI: headroom-ai)
- macOS 25.5.0 (Darwin)
- Agent: Claude Code

PR #3108, opened 2026-08-18
fix(learn): keep CLAUDE.local.md out of git via .git/info/exclude

### LINKED ISSUE #1070: Support CLAUDE.local.md for headroom learn output
## Problem

Currently, `headroom learn` writes corrections and learned patterns to `CLAUDE.md`, `AGENTS.md`, and `GEMINI.md`.

These files are typically committed to version control and shared across the team. However, the patterns learned by `headroom learn` are developer-specific and session-specific — they should not be part of the shared repository configuration.

## Expected Behavior

`headroom learn` should prefer writing to local variant files:
- `CLAUDE.local.md` (instead of `CLAUDE.md`)
- `AGENTS.local.md` (instead of `AGENTS.md`)  
- `GEMINI.local.md` (instead of `GEMINI.md`)

This follows the pattern of other configuration systems (e.g., `.env.local`, `.prettierrc.local.json`), where:
- **Shared files** (`CLAUDE.md`, etc.) contain team conventions and project-wide rules
- **Local files** (`CLAUDE.local.md`, etc.) contain developer-specific learnings and are gitignored

## Suggested Implementation

1. Modify `headroom learn` to write to `{PROVIDER}.local.md` files by default
2. Update the merge logic to read from both shared and local files (local takes precedence)
3. Add `*.local.md` to the gitignore template if distributed

## Benefits

- Prevents commit noise from personal learning sessions
- Allows each developer to maintain their own optimization profiles  
- Keeps shared CLAUDE.md files stable and team-focused
- Aligns with common DevOps/development configuration patterns

gglucass:
## Description

`headroom learn` writes per-project learnings to `CLAUDE.local.md` rather than the team-shared `CLAUDE.md` (#1115, for #1072). The premise is that `CLAUDE.local.md` is personal and gitignored — `writer.py` says so in its own docstring — but nothing makes that true. Git ships no default rule for the name, Claude Code does not add one, and no ignore rule is written at learn time. The file is therefore swept up by the next `git add -A`, and the machine-specific absolute paths inside it reach teammates anyway. The pollution #1072 set out to prevent just moved one filename over.

This is item 3 of #1070 ("Add `*.local.md` to the gitignore template"), the half that never shipped. The codebase already half-acknowledges the gap: `writer.py:317` builds a `gitignore_hint` telling users to add the rule by hand, but it is only attached to the legacy-`CLAUDE.md` migration warning, so users who never had a legacy block never see it.

Closes #1070

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `_ensure_git_ignored()`, which registers the written filename in `.git/info/exclude` after a non-dry-run write.
- Use `.git/info/exclude` rather than the repo's `.gitignore`: the exclude file is per-clone and untracked, so Headroom never leaves an unexpected diff in a team-shared file. Writing to `.gitignore` would trade one kind of pollution for another.
- Resolve the exclude path through `git rev-parse --git-common-dir` so linked worktrees share the entry.
- Gate on `git check-ignore` so a rule the user already has (or one we wrote on a previous run) is left alone and never duplicated.
- Emit a `WriteResult.warnings` entry instead when the file is **already tracked**: git honors no ignore rule for tracked paths, so only `git rm --cached` fixes it, and running that here would silently stage a deletion in the user's repo.
- Gate the whole thing on `target_path.name.endswith(".local.md")`, so an explicit `--target CLAUDE.md` (a deliberate opt-in to the shared file) and the home-directory `~/.claude/CLAUDE.md` case are untouched. The latter matters because `~/.claude` is often itself a dotfiles repo.
- Best-effort by design: no git on PATH, not a repo, or a read-only `.git` all no-op rather than failing the write.

The entry is deliberately unanchored (`CLAUDE.local.md`, not `/CLAUDE.local.md`). A project may be a subdirectory of a larger repo, so anchoring would require the path relative to the repo root; and a `CLAUDE.local.md` at any depth is personal by the same argument.

## Testing

New tests drive real `git` in a real temp repo rather than mocking it — the behavior under test *is* git's, and a mocked `check-ignore` would assert nothing.

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_learn/test_writer.py
tests/test_learn/test_writer.py::TestClaudeLocalMdStaysOutOfGit::test_apply_adds_exclude_entry PASSED [ 83%]
tests/test_learn/test_writer.py::TestClaudeLocalMdStaysOutOfGit::test_second_run_does_not_duplicate_the_entry PASSED [ 86%]
tests/test_learn/test_writer.py::TestClaudeLocalMdStaysOutOfGit::test_existing_gitignore_rule_is_left_alone PASSED [ 89%]
tests/test_learn/test_writer.py::TestClaudeLocalMdStaysOutOfGit::test_tracked_file_warns_instead_of_excluding PASSED [ 91%]
tests/test_learn/test_writer.py::TestClaudeLocalMdStaysOutOfGit::test_dry_run_does_not_touch_exclude PASSED [ 94%]
tests/test_learn/test_writer.py::TestClaudeLocalMdStaysOutOfGit::test_explicit_shared_target_is_never_excluded PASSED [ 97%]
tests/test_learn/test_writer.py::TestClaudeLocalMdStaysOutOfGit::test_outside_a_git_repo_is_a_no_op PASSED [100%]

============================== 37 passed in 2.53s ==============================

$ uvx ruff check headroom/learn/writer.py tests/test_learn/test_writer.py
All checks passed!

$ uvx ruff format --check headroom/learn/writer.py tests/test_learn/test_writer.py
2 files already formatted

$ uv run --frozen --extra dev mypy headroom/learn/writer.py
```

## Real Behavior Proof

- Environment: macOS 15.6 (darwin 24.6.0), Python 3.12, git 2.51.0, worktree off `upstream/main` @ 7ef736fb
- Exact command / steps: `git init` a scratch repo, then run `ClaudeCodeWriter().write([rec], proj, dry_run=False)` against it with `dry_run=False`, once with this patch stashed and once with it applied; then a third run against a repo where `CLAUDE.local.md` was already `git add`ed.
- Observed result: before the patch, `git status --short --untracked-files=all` in the learned-on repo reports `?? CLAUDE.local.md`, i.e. the file is staged by the next `git add -A`. After the patch, `git status` is empty, `.git/info/exclude` gains a comment line plus the bare name `CLAUDE.local.md`, and `ls` confirms the file is still on disk. In the already-tracked repo the write returns exactly one warning: ``/tmp/hr-proof/tracked/CLAUDE.local.md is tracked in git, so learned patterns (including absolute paths from this machine) are committed and shared with your team. Run `git rm --cached CLAUDE.local.md` to untrack it; the file itself stays on disk.``
- Not tested: Windows. The implementation shells out to `git` with no path separators of its own and falls back to a no-op on `OSError`/`FileNotFoundError`, so the failure mode there is "no ignore rule written", not a crash — but I have not run it on Windows.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — this is a `learn` write-path fix, not a proxy/runtime feature.
- Stable/default behavior changed: yes, but only additively — a line is appended to `.git/info/exclude` (never a tracked file) when `learn --apply` writes `CLAUDE.local.md` into a git repo that does not already ignore it. No change to what is written into the context file itself.
- Kill switch / disable path: `--target <name>` with any name not ending in `.local.md` skips the logic entirely. A user who wants the file tracked deletes the one line from `.git/info/exclude`; nothing rewrites it, because `check-ignore` is not consulted for tracked files and the entry is only added when absent.
- Qualification impact: none
- Rollback path: revert the commit — the only persistent artifact is the `.git/info/exclude` line, which is inert once the code is gone and removable by hand.

## Review Readiness

## Checklist

### Out of scope

`CodexWriter`, `GeminiWriter`, and `GrokWriter` still write straight to the shared `AGENTS.md` / `GEMINI.md` / `GROK.md` with no `.local.md` variant at all — the other half of #1070. That is a behavior change rather than a leak fix, so it belongs in its own PR. Happy to follow up if you want it.

### DIFF
--- a/headroom/learn/writer.py
+++ b/headroom/learn/writer.py
@@ -7,9 +7,12 @@ injection mechanism for each agent system (CLAUDE.md, .cursorrules, etc.).
 from __future__ import annotations
 
 import re
+import subprocess
 from abc import ABC, abstractmethod
 from datetime import datetime, timezone
 from pathlib import Path
+
+from headroom._subprocess import run
 
 from ._shared import claude_config_dir
 from .models import (
@@ -204,6 +207,79 @@ def _strip_marker_block(content: str) -> str:
     return cleaned + "\n" if cleaned else ""
 
 
+def _git(repo: Path, *argv: str) -> subprocess.CompletedProcess | None:
+    """Run a git command in ``repo``; None when git is absent or hangs."""
+    try:
+        return run(
+            ["git", *argv],
+            capture_output=True,
+            text=True,
+            cwd=repo,
+            timeout=5,
+        )
+    except (subprocess.TimeoutExpired, FileNotFoundError, OSError):
+        return None
+
+
+def _ensure_git_ignored(target_path: Path, dry_run: bool) -> str | None:
+    """Keep a personal context file out of git via ``.git/info/exclude``.
+
+    ``CLAUDE.local.md`` is only personal if git actually ignores it, and nothing
+    makes that true by default: git ships no rule for the name and neither does
+    Claude Code, so the file lands in the next ``git add -A`` and the machine-
+    specific absolute paths inside it reach teammates anyway -- the exact
+    outcome issue #1072 set out to prevent.
+
+    Writes to the per-clone exclude file rather than the repo's ``.gitignore``
+    because the latter is team-shared and committed: appending to it would leave
+    an unexpected diff in someone else's repo, trading one kind of pollution for
+    another. Returns a warning instead of acting when the file is already
+    tracked -- git honors no ignore rule for tracked files, so only
+    ``git rm --cached`` can fix that, and running it here would silently stage a
+    deletion in the user's repo.
+    """
+    repo = target_path.parent
+    name = target_path.name
+
+    common_dir = _git(repo, "rev-parse", "--git-common-dir")
+    if common_dir is None or common_dir.returncode != 0:
+        return None  # not a git repo, or no git on PATH: nothing to ignore
+
+    tracked = _git(repo, "ls-files", "--error-unmatch", "--", name)
+    if tracked is not None and tracked.returncode == 0:
+        return (
+            f"{target_path} is tracked in git, so learned patterns (including "
+            f"absolute paths from this machine) are committed and shared with "
+            f"your team. Run `git rm --cached {name}` to untrack it; the file "
+            f"itself stays on disk."
+        )
+
+    ignored = _git(repo, "check-ignore", "-q", "--", name)
+    if ignored is not None and ignored.returncode == 0:
+        return None  # already covered by .gitignore or a previous run
+
+    if dry_run:
+        return None
+
+    # Deliberately unanchored: a CLAUDE.local.md at any depth is personal, and
+    # anchoring would need the path relative to the repo root, which differs
+    # when the project is a subdirectory of a larger repo.
+    # git shares info/exclude across linked worktrees via the common dir.
+    exclude = Path(common_dir.stdout.strip() or ".git")
+    exclude = (repo / exclude / "info" / "exclude").resolve()
+    try:
+        exclude.parent.mkdir(parents=True, exist_ok=True)
+        prior = exclude.read_text(encoding="utf-8") if exclude.exists() else ""
+        prefix = "" if not prior or prior.endswith("\n") else "\n"
+        exclude.write_text(
+            f"{prior}{prefix}\n# Personal `headroom learn` output, not team-shared\n{name}\n",
+            encoding="utf-8",
+        )
+    except OSError:
+        return None  # read-only .git, exotic setup: the write is best-effort
+    return None
+
+
 # =============================================================================
 # Claude Code Writer
 # =============================================================================
@@ -248,6 +324,13 @@ class ClaudeCodeWriter(ContextWriter):
 
         if context_recs:
             target_path = self._resolve_context_path(project)
+            # Only ever auto-ignore a file whose name marks it personal: an
+            # explicit --target CLAUDE.md is a deliberate opt-in to the shared
+            # file, and ~/.claude/CLAUDE.md may sit in a dotfiles repo.
+            if target_path.name.endswith(".local.md"):
+                tracked_warning = _ensure_git_ignored(target_path, dry_run)
+                if tracked_warning:
+                    result.warnings.append(tracked_warning)
             # Migrate any stale block left in the team-shared CLAUDE.md by older
             # headroom versions into the new target, then strip it from CLAUDE.md
             # so the shared file is no longer polluted.
--- a/tests/test_learn/test_writer.py
+++ b/tests/test_learn/test_writer.py
@@ -1,5 +1,6 @@
 """Tests for recommendation writer — marker-based file updates."""
 
+import subprocess
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed the write-path behavior and real-Git tests. Using the common Git directory's info/exclude keeps personal learnings local without creating a tracked diff; existing ignore coverage, tracked files, linked worktrees, dry runs, and unavailable/read-only Git metadata are handled appropriately. Looks good.

gglucass:
Still open and still reproducible on current `main` (`e67b3c8a`).

Re-checked today:

- The gap is unchanged upstream. `headroom/learn/writer.py:218` still documents `CLAUDE.local.md` as "gitignored by default", and nothing writes an ignore rule. The only `gitignore_hint` (`writer.py:317`) is still attached solely to the legacy-`CLAUDE.md` migration branch, so a user who never had a legacy block never sees it.
- This repo demonstrates it: `git check-ignore CLAUDE.local.md` in a fresh clone of headroom returns nothing. If `headroom learn` runs against its own repo, the file it writes is picked up by the next `git add -A`, machine-specific absolute paths included.
- No drift: `main` has not touched `headroom/learn/writer.py` in the 108 commits since this branch forked, so the diff still applies as written.
- `pytest tests/test_learn/test_writer.py` on the PR head: 37 passed.

Happy to rebase if that helps, but there is nothing to resolve at the moment. Closes #1070 item 3.

JerrettDavis (changes requested):
The rebased test suite is not hermetic: TestClaudeLocalMdStaysOutOfGit.test_explicit_shared_target_is_never_excluded asserts that git check-ignore reports CLAUDE.md as unignored, but check-ignore also reads the user's global core.excludesFile. On a machine whose global ignore contains CLAUDE.md (a common setup, and reproduced here via C:\Users\jd\.gitignore), the focused suite fails even though this code correctly did not add a repository-local .git/info/exclude entry.

Please make this test verify the side effect owned by this PR (for example, assert that the repo's common-dir info/exclude does not gain CLAUDE.md), or isolate Git from global/system ignore configuration in the test helper. The production behavior remains correct; this request is specifically for deterministic validation across developer and CI environments.

gglucass:
Good catch, and the diagnosis was exactly right - thank you for the reproduction detail, it made this quick.

Pushed 4bda069. Two things, because the cause ran one layer deeper than the test helper:

**The writer shells out to `check-ignore` too.** My first attempt isolated only the test's own git calls, and two *other* tests then failed under your conditions: `writer.py` runs `git check-ignore` to decide whether a rule is already in force, and deliberately skips adding one when it is (that behaviour is codified in `test_existing_gitignore_rule_is_left_alone`). With a global ignore covering `CLAUDE.local.md` the writer correctly does nothing, so the tests asserting the entry *was* added fail. Isolating a helper could not fix that. So the isolation is now an autouse fixture that pins `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` at a non-existent path for the class, which covers the writer's subprocess as well as the test's.

**Plus the PR-owned assertion you asked for.** `test_explicit_shared_target_is_never_excluded` now asserts the repo's own `.git/info/exclude` gains no `CLAUDE.md` entry, before the `check-ignore` effect check.

Verified by reproducing your setup - a global `core.excludesFile` listing `CLAUDE.md` and `CLAUDE.local.md`:

```text
# before this commit, under that global ignore
4 failed, 35 passed
  FAILED ...::test_explicit_shared_target_is_never_excluded   <- the one you hit
  FAILED ...::test_tracked_file_warns_instead_of_excluding
  FAILED ...::test_apply_adds_exclude_entry
  FAILED ...::test_second_run_does_not_duplicate_the_entry

# after, same hostile global ignore   -> 39 passed
# after, via GIT_CONFIG_GLOBAL         -> 39 passed
# after, ordinary environment          -> 39 passed
```

Worth noting the first two failures were latent on your machine already - `git add` refuses an ignored path, so `test_tracked_file_warns_instead_of_excluding` would have broken too. Production behaviour is unchanged; this commit touches only the test file.

PR #3111, opened 2026-08-18
docs(readme): fix broken star history chart

Dessalines39394:
The star history chart in the README is broken: the stargazer data source it
was built on is no longer available, so the chart fails to render.

This change points the chart at a working source so it displays correctly again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,9 +217,9 @@ shows an **Output Tokens Saved** card next to input compression, labelled
 
 → Full write-up incl. the measurement methodology: [Output token reduction](https://headroom-docs.vercel.app/docs/savings)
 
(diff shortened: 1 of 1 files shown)


JerrettDavis (changes requested):
The new host is responsive, but both URLs still request the pre-organization repository slug (chopratejas/headroom). That endpoint currently returns a Repo not found error, so the README badge/image remains broken. Please change both the image URL and its link target to headroomlabs-ai/headroom and verify the rendered SVG before merging.

JerrettDavis:
Thanks for looking into the broken chart. The current README no longer contains the star-history block that this patch updates, so there isn’t a chart URL left to repair there. I’m closing this as obsolete after the README changes.

PR #3114, opened 2026-08-18
fix(proxy/toin): report the tool signature hash, not the composite key prefix

### LINKED ISSUE #2928: sig_hash[:12] truncates the composite auth|model|hash key, so /v1/toin patterns share one id and detail lookup is arbitrary

abhay-codes07:
## Description

`/v1/toin/patterns` reports each pattern's identifier as `sig_hash[:12]`, and the detail endpoint `/v1/toin/pattern/{hash_prefix}` matches with `sig_hash.startswith(hash_prefix)`. But `sig_hash` is not a hash: it is the composite aggregation key produced by `_serialize_pattern_key`, encoded as `"auth|model|hash"`. So the first 12 characters are the auth mode and part of the model family, not the tool signature hash the field name and docstring promise ("Truncated tool signature hash (12 chars)").

The failure is total when `auth_mode` and `model_family` are both `unknown` (the common case): every pattern reports the identical id `unknown|unkn`, and the detail endpoint returns whichever pattern iterates first. It is partial otherwise: any two patterns sharing an auth mode and model family collide on the first 12 characters.

Both endpoints now key on the tool signature hash (the third `|`-delimited field), via a small shared helper, so the reported id is the promised signature hash and the detail lookup addresses exactly that pattern.

Closes #2928.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/telemetry/toin.py`: added `signature_hash_from_serialized_key`, which returns the tool-signature-hash component (third field) of a serialized `"auth|model|hash"` key.
- `headroom/proxy/server.py`: `/v1/toin/patterns` now reports `signature_hash_from_serialized_key(sig_hash)[:12]`, and `/v1/toin/pattern/{prefix}` matches the prefix against that same signature hash rather than the composite key.
- `tests/test_proxy_loopback_gating.py`: added `test_toin_patterns_hash_is_signature_not_composite_prefix` (two patterns that share auth+model must report distinct ids and each must be addressable). Updated the existing detail-endpoint test to address the pattern by its signature-hash prefix.

## Testing

### Test Output

```text
tests/test_proxy_loopback_gating.py -k toin  ->  8 passed
uvx ruff@0.16.2 check headroom/telemetry/toin.py headroom/proxy/server.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/telemetry/toin.py  ->  Success: no issues found
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.12.11, project venv, pytest 9.1.1, FastAPI TestClient, ruff 0.16.2 and mypy 1.20.2 via uvx.
- Exact command / steps: with a fake TOIN exporting two patterns keyed `unknown|unknown|aaaaaaaaaaaa1111` and `unknown|unknown|bbbbbbbbbbbb2222`, reverted both endpoints to the composite-key behavior and ran `python -m pytest tests/test_proxy_loopback_gating.py -k "toin_patterns_hash_is_signature or pattern_detail_whitelists"` -> 2 failed (the listing reported `unknown|unkn` for both entries, and `GET /v1/toin/pattern/bbbbbbbbbbbb` did not match the composite string); restored the fix and re-ran (`-k toin` -> 8 passed).
- Observed result: before the fix both listed entries share the id `unknown|unkn` and the detail endpoint cannot address either; after the fix the entries report `aaaaaaaaaaaa` / `bbbbbbbbbbbb` and `GET /v1/toin/pattern/bbbbbbbbbbbb` returns that pattern's stats.
- Not tested: a long-running proxy accumulating real TOIN patterns; the endpoints are exercised end to end through the FastAPI TestClient over the exported-pattern shape they consume.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. TOIN is observation-only and these are loopback-gated diagnostic endpoints, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: the `hash` field on `/v1/toin/patterns` now carries the tool signature hash instead of the auth/model prefix, and `/v1/toin/pattern/{prefix}` matches on that signature hash. No change to compression behavior or to stored TOIN data.
- Kill switch / disable path: N/A (diagnostics only).
- Qualification impact: the two diagnostic endpoints can now inspect a specific pattern, which they previously could not.
- Rollback path: revert this PR; the endpoints return to slicing/matching the composite key.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A: the endpoint docstring already described the field as the signature hash, which the code now honors)

## Additional Notes

The composite key retains the auth/model dimensions for aggregation; only the externally reported identifier and its lookup change. Patterns that share a tool signature across different auth/model slices still resolve to that signature hash, which matches the documented "tool signature hash" contract.

### DIFF
--- a/headroom/proxy/server.py
+++ b/headroom/proxy/server.py
@@ -181,7 +181,7 @@ from headroom.subscription.tracker import (
 )
 from headroom.telemetry import get_telemetry_collector
 from headroom.telemetry.beacon import is_telemetry_enabled
-from headroom.telemetry.toin import get_toin
+from headroom.telemetry.toin import get_toin, signature_hash_from_serialized_key
 from headroom.transforms import (
     CacheAligner,
     CodeAwareCompressor,
@@ -4932,7 +4932,7 @@ def create_app(config: ProxyConfig | None = None) -> FastAPI:
 
             patterns_list.append(
                 {
-                    "hash": sig_hash[:12],
+                    "hash": signature_hash_from_serialized_key(sig_hash)[:12],
                     "compressions": total_compressions,
                     "retrievals": total_retrievals,
                     "retrieval_rate": f"{retrieval_rate:.1%}",
@@ -4968,9 +4968,12 @@ def create_app(config: ProxyConfig | None = None) -> FastAPI:
         exported = toin.export_patterns()
         patterns_data = exported.get("patterns", {})
 
-        # Search for pattern with matching hash prefix
+        # Search for pattern with matching hash prefix. Match against the tool
+        # signature hash (the identifier /v1/toin/patterns reports), not the
+        # whole "auth|model|hash" composite key — otherwise a prefix taken from
+        # the reported hash never matches the composite string (#2928).
         for sig_hash, pattern_dict in patterns_data.items():
-            if sig_hash.startswith(hash_prefix):
+            if signature_hash_from_serialized_key(sig_hash).startswith(hash_prefix):
                 # Keep this response aligned with /v1/toin/patterns while
                 # excluding query text, field semantics, and other internal
                 # learning state from the detail endpoint.
--- a/headroom/telemetry/toin.py
+++ b/headroom/telemetry/toin.py
@@ -156,6 +156,18 @@ def _deserialize_pattern_key(serialized: str) -> PatternKey:
     return (DEFAULT_AUTH_MODE, DEFAULT_MODEL_FAMILY, serialized)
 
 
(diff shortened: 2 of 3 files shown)


abhay-codes07:
Closing as a duplicate of #2932, which already addresses this by keying the endpoints on the tool signature hash. Apologies for the overlap.

PR #3116, opened 2026-08-18
fix(utils): normalize timezone-aware datetimes in format_timestamp

abhay-codes07:
## Description

`format_timestamp` appended a bare `"Z"` to `dt.isoformat()`:

```python
return dt.isoformat() + "Z"
```

That is correct only for a naive (assumed-UTC) datetime. For a timezone-aware datetime it produces an invalid ISO 8601 string, and for a non-UTC offset it mislabels the instant:

```
format_timestamp(datetime(2024,1,1,12, tzinfo=utc))       -> '2024-01-01T12:00:00+00:00Z'   # invalid; fromisoformat rejects it
format_timestamp(datetime(2024,1,1,12, tzinfo=utc+5h))    -> '2024-01-01T12:00:00+05:00Z'   # labels 07:00Z as "Z"
```

This is reachable: the shipped integrations construct `RequestMetrics(timestamp=datetime.now(timezone.utc))` (aware) - e.g. `headroom/integrations/agno/model.py`, `.../agno/hooks.py`, `.../autogen`, `.../crewai`, `.../langchain` - and the storage layer persists `format_timestamp(metrics.timestamp)` to JSONL/SQLite. So those rows carried malformed timestamps that a strict parser (including Python's own `datetime.fromisoformat`) rejects, and any non-UTC offset silently recorded the wrong instant.

The fix converts an aware datetime to UTC before formatting; a naive datetime is still assumed to be UTC. Output stays `...Z` and now round-trips through a strict parser.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `headroom/utils.py` (`format_timestamp`): convert timezone-aware datetimes to UTC (`astimezone(timezone.utc).replace(tzinfo=None)`) before appending `"Z"`; naive datetimes are unchanged. The `None` default now uses aware `datetime.now(timezone.utc)` and flows through the same normalization, producing the identical string as before.
- `tests/test_utils.py`: added `test_format_timestamp_normalizes_aware_datetimes` (aware UTC round-trips through `fromisoformat`; a `+05:00` offset is converted to the correct UTC time). The existing naive-datetime assertions are retained.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
[...]

PR #3117, opened 2026-08-18
fix(integrations): record metric timestamps in UTC, not naive local time

### LINKED ISSUE #2910: [BUG] RequestLog.timestamp is timezone-naive (outcome.py:529), producing negative "Xs ago" in dashboard Recent Requests

abhay-codes07:
## Description

Several framework integrations stamp their metric records with a naive local `datetime.now()`:

- `OptimizationMetrics` in `langchain/chat_model.py` (two sites)
- `ToolCompressionMetrics` in `autogen/agents.py`, `crewai/agents.py`, `langchain/agents.py`

Everywhere else in the codebase these metric timestamps are UTC: the SDK's core recorder `headroom/client.py` uses `datetime.now(timezone.utc).replace(tzinfo=None)`, and the agno and strands adapters build the same metric classes with `datetime.now(timezone.utc)`. `langchain/langsmith.py` then exports `metrics.timestamp.isoformat()` as an absolute telemetry field (`headroom.optimization_timestamp`).

A naive local `datetime.now()` records the local wall-clock time with no offset. Serialized and read downstream as UTC, it is wrong by the local UTC offset for every user not on UTC (and can produce negative "time ago" values, the same class as the previously fixed #2910). This aligns the metric sites to UTC.

The `start_time` / `end_time` calls in `langchain/chat_model.py` are deliberately left as naive `datetime.now()`: they are only ever subtracted from each other to compute a duration, where the offset cancels, so UTC is unnecessary there.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/integrations/autogen/agents.py`, `headroom/integrations/crewai/agents.py`, `headroom/integrations/langchain/agents.py`: `ToolCompressionMetrics(timestamp=...)` now uses `datetime.now(timezone.utc)`.
- `headroom/integrations/langchain/chat_model.py`: the two `OptimizationMetrics(timestamp=...)` sites now use `datetime.now(timezone.utc)`; the four `start_time`/`end_time` duration calls are unchanged.
- Added `timezone` to the `datetime` import in each of those four files.
- `tests/test_integration_metric_timestamps_utc.py`: new regression test driving the three framework-free recording seams (autogen's standalone `_record_metrics`; crewai and langchain via an unbound `_record_metrics` call with a duck-typed self) and asserting the recorded timestamp is UTC-aware.

## Testing

### Test Output

```text
tests/test_integration_metric_timestamps_utc.py  ->  3 passed
uvx ruff@0.16.2 check <changed files>  ->  All checks passed!
uvx mypy@1.20.2 <four integration modules>  ->  Success: no issues found in 4 source files
```

## Real Behavior Proof

- Exact command / steps: reverted the three tested sites to `datetime.now()` and ran `python -m pytest tests/test_integration_metric_timestamps_utc.py` -> 3 failed (`assert ts.tzinfo is not None` -> the recorded timestamp was `datetime.datetime(2026, 8, 18, 23, 42, 45, ...)` with `tzinfo=None`, i.e. naive local time); restored `datetime.now(timezone.utc)` and re-ran -> 3 passed. The adapters import their frameworks lazily, so the recording functions run without autogen/crewai/langchain installed.
- Observed result: after the fix, each adapter records a UTC-aware timestamp (`utcoffset() == timedelta(0)`); before, it recorded a naive local datetime.
- Not tested: the two `OptimizationMetrics` sites in `chat_model._optimize_messages` (they require a live pipeline + a wrapped LangChain model to reach); they are the identical one-line change and are the sites the LangSmith exporter serializes absolutely.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This is metric-timestamp construction in the framework adapters, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: integration metric timestamps are now UTC-aware instead of naive local time. Duration math (`start_time`/`end_time`) is unchanged.
- Qualification impact: LangSmith optimization timestamps (and any absolute use of these metrics) are now correct for non-UTC users.
- Rollback path: revert this PR; the metric sites return to naive `datetime.now()`.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A: internal metric fields, no user-facing docs)

## Additional Notes

I scoped this to the metric-record timestamps and intentionally left the duration-only `datetime.now()` calls naive, since converting them would add no value and the offset cancels in subtraction. Complements the writer fix in the sibling PR for `utils.format_timestamp` (aware-datetime handling); the two are independent.

### DIFF
--- a/headroom/integrations/autogen/agents.py
+++ b/headroom/integrations/autogen/agents.py
@@ -36,7 +36,7 @@ import asyncio
 import functools
 import logging
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any
 
 try:
@@ -226,7 +226,7 @@ def _record_metrics(
 
     metric = ToolCompressionMetrics(
         tool_name=tool_name,
-        timestamp=datetime.now(),
+        timestamp=datetime.now(timezone.utc),
         chars_before=chars_before,
         chars_after=chars_after,
         chars_saved=max(0, chars_saved),
--- a/headroom/integrations/crewai/agents.py
+++ b/headroom/integrations/crewai/agents.py
@@ -26,7 +26,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any
 
 try:
@@ -294,7 +294,7 @@ class HeadroomToolWrapper(BaseTool):  # type: ignore[misc]
 
         metric = ToolCompressionMetrics(
             tool_name=self.name,
-            timestamp=datetime.now(),
+            timestamp=datetime.now(timezone.utc),
             chars_before=chars_before,
             chars_after=chars_after,
             chars_saved=max(0, chars_saved),
--- a/headroom/integrations/langchain/agents.py
+++ b/headroom/integrations/langchain/agents.py
@@ -26,7 +26,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any
 
 # LangChain imports - these are optional dependencies
@@ -251,7 +251,7 @@ class HeadroomToolWrapper:
 
         metric = ToolCompressionMetrics(
             tool_name=self.name,
-            timestamp=datetime.now(),
+            timestamp=datetime.now(timezone.utc),
             chars_before=chars_before,
             chars_after=chars_after,
             chars_saved=max(0, chars_saved),
--- a/headroom/integrations/langchain/chat_model.py
+++ b/headroom/integrations/langchain/chat_model.py
@@ -33,7 +33,7 @@ import json
 import logging
 from collections.abc import AsyncIterator, Iterator, Sequence
 from dataclasses import dataclass
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any
 from uuid import UUID, uuid4
 
@@ -349,7 +349,7 @@ class HeadroomChatModel(BaseChatModel):
         # Create metrics
         metrics = OptimizationMetrics(
             request_id=request_id,
-            timestamp=datetime.now(),
+            timestamp=datetime.now(timezone.utc),
             tokens_before=result.tokens_before,
             tokens_after=result.tokens_after,
             tokens_saved=result.tokens_before - result.tokens_after,
@@ -876,7 +876,7 @@ class HeadroomRunnable:
         # Track metrics
         metrics = OptimizationMetrics(
             request_id=str(uuid4()),
-            timestamp=datetime.now(),
+            timestamp=datetime.now(timezone.utc),
             tokens_before=result.tokens_before,
             tokens_after=result.tokens_after,
             tokens_saved=result.tokens_before - result.tokens_after,
--- a/headroom/integrations/langchain/langsmith.py
+++ b/headroom/integrations/langchain/langsmith.py
@@ -39,7 +39,7 @@ from __future__ import annotations
 import logging
 import os
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any
 from uuid import UUID
 
@@ -86,7 +86,7 @@ class PendingMetrics:
     tokens_saved: int
     savings_percent: float
     transforms_applied: list[str]
-    timestamp: datetime = field(default_factory=datetime.now)
+    timestamp: datetime = field(default_factory=lambda: datetime.now(timezone.utc))
 
 
 class HeadroomLangSmithCallbackHandler(BaseCallbackHandler):
--- a/tests/test_integration_metric_timestamps_utc.py
+++ b/tests/test_integration_metric_timestamps_utc.py
@@ -1,0 +1,64 @@
+"""Integration metric timestamps must be recorded in UTC, not local time.
+
+The SDK's canonical recorder (`headroom/client.py`) and the agno/strands
+adapters stamp metric timestamps in UTC, and `langchain/langsmith.py` exports
+`OptimizationMetrics.timestamp.isoformat()` as an absolute telemetry field. The
+autogen/crewai/langchain tool wrappers used a naive local `datetime.now()`,
+which records the local wall-clock time with no offset — read downstream as UTC,
+it is wrong by the local UTC offset for every non-UTC user.
+
+These adapters import their frameworks lazily, so the recording seams are
+exercised directly here without installing autogen/crewai/langchain.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timedelta
+from types import SimpleNamespace
+
+from headroom.integrations.autogen.agents import ToolMetricsCollector as AutogenCollector
+from headroom.integrations.autogen.agents import _record_metrics as autogen_record
+from headroom.integrations.crewai.agents import HeadroomToolWrapper as CrewaiWrapper
+from headroom.integrations.crewai.agents import ToolMetricsCollector as CrewaiCollector
+from headroom.integrations.langchain.agents import HeadroomToolWrapper as LangchainWrapper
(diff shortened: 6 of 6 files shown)


JerrettDavis (changes requested):
The individual call-site replacements are correct only after #3116 is present, but this PR currently targets main without that formatter fix. On the current base, format_timestamp(datetime.now(timezone.utc)) emits malformed ...+00:00Z; this PR expands the number of integration paths feeding aware datetimes into that broken persistence seam, while its tests stop at the in-memory object and do not exercise serialization. Please make the dependency explicit and update/rebase this branch after #3116 lands (or include equivalent formatter normalization here), then add one persistence/format assertion so the combined contract is covered. Until then the latest source is not independently safe to merge.

abhay-codes07:
Thanks for the careful review. I traced the data flow and I think the format_timestamp concern does not actually reach this PR, but I have made the serialization contract explicit either way.

The metric classes this PR touches (OptimizationMetrics, ToolCompressionMetrics) are integration-local and are serialized only via metrics.timestamp.isoformat() in langchain/langsmith.py:241. None of them go through headroom.utils.format_timestamp: `grep -rn format_timestamp headroom/integrations/` is empty, and the only .timestamp consumer in integrations is that isoformat call. format_timestamp is called exclusively by storage/jsonl.py and storage/sqlite.py on RequestMetrics, which this PR does not touch.

For an aware datetime, .isoformat() emits a valid `...+00:00` string, not the malformed `...+00:00Z` that appending "Z" to an aware value produces. So this path is correct independently of #3116 (that PR fixes the separate RequestMetrics -> format_timestamp seam).

Changes since your review:

1. Rebased onto current main.
2. Added the persistence/format assertion you asked for. _assert_utc now also checks the serialized form the telemetry actually exports: `ts.isoformat().endswith("+00:00")` plus a `datetime.fromisoformat()` round-trip. A naive local timestamp fails this (no offset), so the combined source-plus-serialization contract is now covered, not just the in-memory object.

Happy to add the explicit dependency note on #3116 if you would still prefer it, but since these metrics never touch format_timestamp I believe the branch is independently safe now.

JerrettDavis (changes requested):
Re-reviewed the updated head. The call-site changes remain locally correct, but the merge-safety blocker is still present on current main: headroom.utils.format_timestamp() still appends Z to dt.isoformat(), so an aware UTC datetime becomes the malformed +00:00Z form. The new assertion only calls ts.isoformat() directly and explicitly says it does not exercise format_timestamp; it therefore proves a different path and does not cover the persistence seam from the prior review. Please either land and rebase onto the formatter-normalization dependency, or include the equivalent normalization here, and add a regression that sends an aware integration timestamp through the actual shared persistence/format path. Until that contract is covered, this PR increases the number of aware values that can reach a formatter which does not accept them safely.

abhay-codes07:
Thanks for holding the line on this. You were right that the previous test proved a different path than the one that actually serializes, so I found and fixed the real serialization seam.

Tracing it end to end: the exported telemetry field `headroom.optimization_timestamp` is written in `HeadroomLangSmithCallbackHandler._attach_metrics_to_run` (langsmith.py:241) as `metrics.timestamp.isoformat()`, where `metrics` is a `PendingMetrics`. `set_headroom_metrics` never passes a timestamp, so it fell back to `PendingMetrics`'s default, which was a naive `datetime.now()`. That is the genuine persistence/format seam, and it had the same naive-local bug: it serialized with no offset and a downstream reader silently treats it as UTC.

Changes in this push:

1. `PendingMetrics.timestamp` default is now `datetime.now(timezone.utc)`, so the value that reaches the exporter is UTC-aware.
2. New regression `test_langsmith_exported_optimization_timestamp_is_utc` drives the actual export path: it constructs a default `PendingMetrics`, runs it through `_attach_metrics_to_run`, reads back the exported `headroom.optimization_timestamp` string, and asserts it ends with `+00:00` and round-trips through `datetime.fromisoformat` at zero offset. So the serialization contract is now covered at the real seam, not just on the in-memory object.

On the `format_timestamp` concern specifically: I confirmed none of these integration metrics reach `headroom.utils.format_timestamp`. `grep -rn format_timestamp headroom/integrations/` is empty, and `.isoformat()` on an aware value emits a valid `+00:00`, never the `+00:00Z` that appending "Z" would produce. So this PR is independently safe on current main and does not need #3116 to land first. Branch is rebased onto latest main.

JerrettDavis (approved):
Re-reviewed current head f3aa0009. The newly added LangSmith default and real export-path regression close the integration serialization concern from my earlier reviews: that path emits a valid aware +00:00 ISO value and round-trips it. The adapter metric records are integration-local types and the changed call sites do not route through the core format_timestamp persistence helper; #3116 remains an independent fix for that helper rather than a merge-safety dependency here. All current-head checks are green, and I found no remaining blocker.

PR #3119, opened 2026-08-19
fix(transforms/kompress): preserve line boundaries and tabular output in Kompress

### LINKED ISSUE #3099: [BUG] Compression merges text across shell command boundaries, producing well-formed lines that encode a false fact
### Environment
- headroom 0.33.0, proxy `--mode token --backend anthropic`, resolved profile `coding`, `proxy_mode: cache`
- Client: Claude Code via `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`
- `HEADROOM_OUTPUT_SHAPER=1`, `HEADROOM_OUTPUT_SHAPING=1`, no other overrides

### Problem

When a single tool result contains the output of several shell commands, compression can merge text across
the command boundary. The result is a line that looks well-formed but attributes one command's output to a
different command.

This is materially worse than truncation: truncation is visible and an agent knows to re-run, whereas a
merge is invisible and the agent proceeds on a wrong fact.

### Reproduction

A Bash tool call containing:

```bash
command -v rg || echo "rg NOT INSTALLED"
command -v ast-grep difft scc 2>/dev/null
```

Real output (two separate lines, first command printed nothing resolvable, second printed a path):

```
rg
/opt/homebrew/bin/ast-grep
```

What arrived after compression:

```
rg /opt/homebrew/bin/ast-grep
```

Read literally, that says `rg` resolves to `/opt/homebrew/bin/ast-grep`. It does not. I could not tell from
the compressed result whether `rg` existed, resolved, or failed, and had to re-run the two commands
separately in isolation to establish the fact. (The ground truth: `rg` was a shell *function*, and no `rg`
binary existed — a distinction the merged line actively obscures.)

### Second instance, same session

A `ls -la` listing had per-row field loss, inconsistently across rows of one listing:

```
-rw-r--r--@ 1 <user> 4921 12 20:14 daemon.log
```

The group column is gone and the month (`Aug`) is gone, while other rows in the same output retained all
fields. A model reading this cannot reliably parse sizes or dates, and the inconsistency across rows means
it cannot even learn the shape.

### Impact

Both cases produce confident-looking text that encodes a false fact. In an agent loop this turns into a
wrong conclusion acted upon, or — in the better case — redundant re-running of commands to disambiguate,
which spends more tokens than the compression saved.

### Suggested fix

Treat command/output boundaries as hard barriers during compression of shell results: never join lines
across them, even when crushing lines *within* one command's output. Where per-line field elision is
applied to tabular output (`ls -l`, `ps`, etc.), apply it uniformly across all rows or not at all, so the
result stays parseable.

### Note

Hit under the stock `coding` profile with no tuning overrides.

slxng1758:
## Description

KompressCompressor flattened `tool_result` content into a space-joined word list, discarding line boundaries and column padding — so compression could glue words from unrelated shell commands onto one fabricated line, and drop `ls -l` table fields inconsistently row to row. Now tracks each word's source line and never joins across a line boundary, and protects fixed-width tabular rows as atomic units.

Closes #3099

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `KompressCompressor.compress()`/`compress_batch()`: track each word's source line, never join kept words from different lines onto one line.
- Add `_tabular_protected_line_ids()`: runs of ≥3 consistent fixed-width-column lines (`ls -l`, `ps`) are kept whole, not partially crushed. Reuses the existing `HEADROOM_KOMPRESS_MUST_KEEP` toggle.
- Add `tests/test_kompress_line_integrity.py` reproducing both bugs.

## Testing

### Test Output

```text
tests/test_kompress_line_integrity.py .... [4 passed]
66 passed, 1 warning in 18.15s
ruff check .: All checks passed!
```

## Real Behavior Proof

- Environment: macOS/Apple Silicon (MPS), Python 3.12, real cached `chopratejas/kompress-v2-base` model.
- Exact command / steps: ran `compress()`/`compress_batch()` on Bash-transcript content shaped like the issue, before and after the fix, including an aggressive `target_ratio=0.15` run to force lines to fully empty out.
- Observed result: unpatched code collapsed a 24-line transcript onto one line and dropped `ls -la` fields — bug reproduced live. Patched code: no merges under any condition, tabular rows stayed intact even at ~22-29% overall keep ratio.
- Not tested: full HTTP proxy path; ONNX backend with a real model.

## Runtime Rollout Safety

- Rollout-managed feature(s): None new — reuses `HEADROOM_KOMPRESS_MUST_KEEP`
- Stable/default behavior changed: Yes — compression output is now line/row-safe by default (correctness fix, not opt-in)
- Kill switch / disable path: `HEADROOM_KOMPRESS_MUST_KEEP=0` disables tabular-row protection; the line-join fix itself has no toggle (never valid to merge lines)
- Qualification impact: None known
- Rollback path: revert this commit/PR

## Review Readiness

## Checklist

### DIFF
--- a/headroom/transforms/kompress_compressor.py
+++ b/headroom/transforms/kompress_compressor.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import contextlib
 import gc
 import hashlib
+import itertools
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed both single and batch reconstruction paths and the focused model-stub tests. Source-line IDs prevent retained tokens from separate records being fabricated into one line, and table-run protection applies symmetrically after inference while retaining the existing operator opt-out. The change may normalize intra-row spacing, but it preserves fields and boundaries—the correctness contract at issue. The branch is mergeable and checks are green. No blocking findings.

PR #3126, opened 2026-08-19
fix(savings): write the output-savings ledger atomically

gglucass:
## Description

`SavingsLedger.save` used `Path.write_text`, which truncates the file to zero bytes before writing the new content. A crash, SIGKILL or ENOSPC at that moment leaves a partial document that `load` discards on the next read, taking the seeded verbosity baseline and every accumulated treatment/control observation with it.

The window recurs continuously under load: `SavingsRecorder` rewrites this file every `flush_every` records on a busy proxy, and readers (`estimate`, `_reload_baseline_locked`, anything inspecting the ledger out of process) can be part-way through a read when it opens.

`fsutil.write_text` already implements the fix used elsewhere in the codebase (temp file in the same directory, fsync, `os.replace`), so this is a two-line reuse rather than new machinery.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `SavingsLedger.save` writes via `fsutil.write_text` instead of `Path.write_text`.
- Regression test: a save that dies mid-write leaves the previous ledger byte-identical and no temp file behind.

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_output_savings.py -q
tests/test_output_savings.py ........................................    [100%]
40 passed in 1.40s

All checks passed!

```

## Real Behavior Proof

- Environment: macOS 24.6.0, Python 3.10.18, uv, upstream/main at 17522fb0
- Exact command / steps: reverted `headroom/proxy/output_savings.py` to `upstream/main`, kept the new test, ran `uv run --frozen --extra dev pytest tests/test_output_savings.py::TestLedgerPersistence::test_failed_save_leaves_the_previous_ledger_intact -q`
- Observed result: fails on main with the data loss it describes, `AssertionError: the previous ledger survived / assert b'' == b'{"baseline"..."control":{}}'` -- the ledger is empty after the interrupted write. Passes with the fix.
- Not tested: no power-loss or real ENOSPC test; the failure is injected by patching the write. The atomicity itself comes from `os.replace` and is already relied on by `settings_store`, `install/state` and `fsutil`.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. This is the persistence path of an existing feature, not a gated one.
- Stable/default behavior changed: No. The same bytes land at the same path; only the write is staged through a same-directory temp file and renamed over the target.
- Kill switch / disable path: N/A. No new flag, no new setting.
- Qualification impact: None. One test added, no existing test changed.
- Rollback path: Revert the commit. The on-disk format is untouched, so a ledger written either way loads under either version.

## Review Readiness

### DIFF
--- a/headroom/proxy/output_savings.py
+++ b/headroom/proxy/output_savings.py
@@ -326,11 +326,26 @@ class SavingsLedger:
         return ledger
 
     def save(self, path: Any) -> None:
+        """Persist the ledger, atomically.
+
+        ``Path.write_text`` truncates the file to zero bytes before writing the
+        new content, so a crash, SIGKILL or ENOSPC mid-write leaves a truncated
+        document that ``load`` discards -- taking the baseline and every
+        accumulated observation with it. The recorder rewrites this file every
+        ``flush_every`` requests on a busy proxy, and readers (``estimate``,
+        ``_reload_baseline_locked``, out-of-process consumers of the ledger)
+        can be part-way through a read at that moment. ``fsutil.write_text``
+        writes a temp file in the same directory and ``os.replace``s it in, so
+        both crashes and concurrent readers see either the old ledger or the
+        complete new one.
+        """
         from pathlib import Path
+
+        from ..fsutil import write_text
 
         p = Path(path)
         p.parent.mkdir(parents=True, exist_ok=True)
-        p.write_text(json.dumps(self.to_dict(), separators=(",", ":")))
+        write_text(p, json.dumps(self.to_dict(), separators=(",", ":")))
 
     @classmethod
     def load(cls, path: Any) -> SavingsLedger:
--- a/tests/test_output_savings.py
+++ b/tests/test_output_savings.py
@@ -2,6 +2,10 @@
 
 from __future__ import annotations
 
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest persistence change against fsutil.write_text. The helper writes and fsyncs a same-directory temporary file, preserves existing mode/symlink behavior, atomically replaces the target, and removes the temporary file on failure. The injected os.replace failure test proves the prior ledger survives and cleanup occurs. The branch is mergeable and checks are green. No blocking findings.

gglucass:
Closing: superseded by #3231, which merged on 2026-08-24 and lands the same fix - `SavingsLedger.save()` now writes through `fsutil.write_text` (temp file + fsync + atomic `os.replace`), `load()` logs a warning instead of silently swallowing a corrupt ledger, and it additionally moves the flush off the event loop. Its `TestFlushDurability` suite covers the same failure modes as the test here (failed-save intactness, no temp residue, corrupt-file warning), so nothing in this PR remains unmerged. Thanks for the review @JerrettDavis - the conflict this branch now shows against main is exactly that overlap.

PR #3127, opened 2026-08-19
fix(savings): don't let a thin holdout arm decide the headline number

gglucass:
## Description

`best_estimate` preferred the measured A/B number the moment a single stratum held one sample in both arms:

```python
measured = self.estimate_from_holdout()
return measured if measured is not None else self.estimate_from_baseline()
```

A holdout arm starts empty and fills slowly. At a 1-3% holdout, conversation-stable assignment means control samples arrive in clusters over weeks, so for most of its life the arm is a handful of requests deciding the reported reduction. Three stale control samples produced a **-1439.9%** reduction that way.

That number leaves the estimator as a plain `reduction_percent` in `/stats`, so every consumer has to either render it or invent its own sanity clamp. Deciding when a measurement is worth believing belongs in the estimator.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

Three bars now stand between a holdout arm and the headline number. The first is per stratum, the other two are aggregate:

- The current upstream admission rule requires `MEASURED_MIN_CLUSTERS` (5) distinct conversation clusters in each contributing arm. Repeated requests from one conversation do not provide independent controls. This PR preserves that rule and adds the aggregate promotion checks below.
- The measured estimate is promoted only when it covers at least `MEASURED_MIN_COVERAGE` (0.5) of the requests the baseline can also score. Below that the holdout describes a corner of the traffic, not the traffic.
- ...and only when its 95% band is no wider than `MEASURED_MAX_CI_HALF_WIDTH_PCT` (+/-10pp). An arm can clear the sample floor and still be too noisy to say anything.
- Below any bar the synthetic control stands, still labelled `estimated`. A holdout-only deployment (no `learn --verbosity` baseline) still reports its measurement, since there is nothing to fall back to.
- The admission, coverage, and interval-width thresholds are module constants.

The latest refresh preserves upstream conversation-cluster admission and adds coverage/interval-width promotion. Earlier sample-count-based validation below is historical, not a description of the current implementation.

## Testing

### Historical Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_output_savings.py tests/test_output_savings_cli.py -q
tests/test_output_savings.py ........................................... [ 89%]
tests/test_output_savings_cli.py ...                                     [100%]
48 passed in 0.39s

All checks passed!

```

New regressions: `test_a_single_control_observation_does_not_contribute`, `test_repeated_identical_control_observations_do_not_contribute`, and `test_a_wide_band_does_not_take_over_even_at_full_size` (which pins that the width gate still bites on an arm that clears the sample floor).

Three pre-existing tests used 10-sample arms and now use the floor instead. Their assertions are unchanged; only the loop bound moved.

## Historical Real Behavior Proof

- Environment: macOS 24.6.0, Python 3.10.18, uv, upstream/main at 17522fb0
- Exact command / steps: set `MEASURED_MIN_ARM_SAMPLES = 1`, which is exactly the previous admission rule (`c.n == 0 or t.n == 0`), then ran `uv run --frozen --extra dev pytest tests/test_output_savings.py -q -k "single_control_observation or repeated_identical_control or thin_control_arm"`
- Observed result: all three fail without the floor, each promoting a 200-request headline off an arm that cannot support one. One control observation: `pct=20.0, ci_low_pct=20.0, ci_high_pct=20.0, n_requests=200, kind='measured'` - a zero-width 95% band over 100% coverage, so both aggregate gates pass it. Five identical control observations: `pct=-300.0, ci_low_pct=-300.0, ci_high_pct=-300.0` - same zero-width band, arbitrarily far from the treatment mean. Three scattered observations: `pct=56.5, ci_low_pct=-84.3, ci_high_pct=197.3`. All three return `None` from `estimate_from_holdout` with the floor in place, and `best_estimate().kind` is `estimated`.
- Not tested: no live proxy run with a real holdout; the arms are constructed directly on the ledger. The -1439.9% figure quoted above is from a shipped desktop build, not reproduced here.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. Output-savings estimation runs wherever a ledger exists; this changes which of two already-computed estimates is reported, not whether either is computed.
- Stable/default behavior changed: Yes, deliberately. A deployment whose holdout arm is thin now reports `estimated` where it previously reported `measured`, and `reduction_percent` moves to the synthetic-control number with it. Deployments with a well-populated holdout are unaffected. No request accounting, no persisted state and no wire format changes.
- Kill switch / disable path: No runtime flag. Revert this PR to remove its coverage/interval-width promotion gates while retaining upstream conversation-cluster admission. The historical `MEASURED_MIN_ARM_SAMPLES` constant is no longer used.
- Qualification impact: Three existing tests re-sized to the sample floor with assertions unchanged; one existing test now asserts the thin arm is excluded rather than merely unpromoted. No test was relaxed.
- Rollback path: Revert the commits. Both estimates are recomputed from the ledger on every read, so the previous headline returns immediately with no migration or backfill.

## Review Readiness

### Current validation (2026-09-10)

At `e6ca2f8342977a240756899fa32cb9bcd72e8cac`, all 70 tests in `tests/test_output_savings.py` pass locally, including thin controls, repeated same-conversation controls, insufficient coverage, wide intervals, qualified legacy migration, and fallback ordering. No live provider billing or traffic run is claimed.


JerrettDavis (changes requested):
The direction is right, but the current uncertainty gate still allows the exact thin-control failure it is meant to stop. _Accum.var is zero when 
 <= 1 (and also for several identical observations), so one control request can produce a zero-width 95% band. If treatment coverage is at least 50%, _measured_supersedes then promotes that single-request control as perfectly bounded. Please require a defensible minimum control sample/effective sample size—ideally per contributing stratum, since the estimator is stratified—in addition to coverage and CI width, and add regressions for one control observation and repeated identical control observations. The normal-approximation interval cannot establish confidence when the variance estimate has no degrees of freedom.

gglucass:
@JerrettDavis you're right, and the failure was worse than "allows": one control observation passes *both* aggregate gates by construction. `_Accum.var` is 0 below two observations, so the band comes out zero-wide, and since the thin stratum is usually the whole of the traffic, coverage is 100%. Addressed in 886049f3.

- Per-stratum floor, at the source. A stratum enters the measured estimate only once **both** arms hold `MEASURED_MIN_ARM_SAMPLES` (30) observations. It lives in `estimate_from_holdout` rather than in `_measured_supersedes` because the estimator is stratified: excluding a stratum too thin to divide by keeps the aggregate honest, whereas gating on the weakest stratum would block a well-measured promotion on one rare stratum that may never fill.
- Why 30, not "more than one". Your last sentence is the reason. Two observations give the variance a degree of freedom but not a usable one, and identical observations report 0 at any n. 30 is the conventional floor at which a z-interval on an estimated variance holds; below it `_finalize` would need a t-quantile it does not compute. It is a module constant alongside the other two, so a deployment that samples differently can move it.
- Both regressions you asked for, plus one for the gate you didn't want weakened. `test_a_single_control_observation_does_not_contribute` and `test_repeated_identical_control_observations_do_not_contribute` cover the two zero-variance shapes. `test_a_wide_band_does_not_take_over_even_at_full_size` pins that the width gate still bites on an arm that clears the sample floor, so the floor didn't quietly become the only live check.

Proof, by setting the floor to 1 (exactly the previous admission rule, `c.n == 0 or t.n == 0`) and running the new tests:

```text
one control observation      pct=20.0    ci=[20.0, 20.0]      n_requests=200  kind='measured'
five identical, at 200 tok   pct=-300.0  ci=[-300.0, -300.0]  n_requests=200  kind='measured'
three scattered              pct=56.5    ci=[-84.3, 197.3]    n_requests=200  kind='measured'
```

All three return `None` from `estimate_from_holdout` with the floor in place, and `best_estimate().kind` is `estimated`.

Three pre-existing tests used 10-sample arms and now use the floor; their assertions are unchanged, only the loop bound moved. `test_a_thin_control_arm_does_not_take_over` now asserts the arm is excluded outright rather than merely unpromoted, since that is what changed. 48 passed, `ruff check` and `mypy` clean.

I also filled in the `Runtime Rollout Safety` and `Review Readiness` sections the governance bot flagged on this PR and on the two siblings (#3126, #3128); all three are green now.

gglucass:
Addressed in 886049f3 (pushed this morning, after your review) — point by point:

- Per-stratum minimum: `estimate_from_holdout` now skips any stratum where either arm holds fewer than `MEASURED_MIN_ARM_SAMPLES = 30` observations (`min(c.n, t.n) < 30`), so the floor is enforced per contributing stratum, not in aggregate. 30 is documented at the constant as the conventional floor for a z-interval on an estimated variance; below it the module would need t-quantiles it does not compute.
- One control observation: regression `test_a_single_control_observation_does_not_contribute` — the single-request arm passes both aggregate gates (zero-wide band, 100% coverage) and is excluded only by the per-stratum floor; `estimate_from_holdout()` returns `None` and `best_estimate()` stays `estimated`.
- Repeated identical control observations: regression `test_repeated_identical_control_observations_do_not_contribute` — five identical control values (variance 0, arbitrarily far off) likewise fail the floor and cannot promote.

One boundary I left deliberately open: a stratum where BOTH arms hold >= 30 genuinely identical values still contributes with zero variance. At that size sample-variance 0 means a deterministic stratum (fixed-size structured outputs), where the zero-width contribution is arguably truthful rather than degenerate — but if you'd rather treat var == 0 as "cannot establish confidence" regardless of n, that's a one-line guard and I'm happy to add it.

Full suite: `tests/test_output_savings.py` — 45 passed.

JerrettDavis (approved):
Re-reviewed at 886049f3e91e94f3bf4cd45910df7c51ee5d722f. The per-stratum minimum arm size prevents a thin holdout arm from driving the measured estimate, and the added coverage/interval-width promotion gates keep weak measurements from replacing the baseline headline. The regression coverage exercises the sparse-arm and uncertainty cases that motivated my earlier request. This resolves my blocker.

JerrettDavis (approved):
Re-reviewed the main refresh at 593ad4a8aed2662f05240e44dfaab2a09bd18f70. The minimum sample floor, coverage gate, and uncertainty gate remain intact after integrating the modelled fallback from main. All 58 output-savings tests pass, including sparse/identical control samples, wide intervals, insufficient coverage, and fallback ordering. My previous blocker remains resolved.

gglucass:
Rebased onto main (e6ca2f8). Main landed the conversation-cluster gate (#3460) in the meantime, which overlaps this branch's per-arm sample floor, so the merge is not mechanical:

**Dropped `MEASURED_MIN_ARM_SAMPLES` instead of stacking it.** Under conversation-stable assignment a REQUEST floor is the wrong axis - 30 turns of one session would clear it while five separate conversations failed. `MEASURED_MIN_CLUSTERS` now solely owns "is this arm a real sample", and it does the job the floor was written for: the -1439.9% three-sample case is excluded by clusters, not by count.

**What is left here, and main does not have it:** the supersede decision. `estimate_from_holdout` qualifying is not the same question as the measured number deserving the headline slot. A qualifying measurement displaces the synthetic control only once it covers `MEASURED_MIN_COVERAGE` (50%) of the requests the baseline can also score, and its 95% band is within `MEASURED_MAX_CI_HALF_WIDTH_PCT` (10pp). Both are still reachable after the cluster gate: `test_a_wide_band_does_not_take_over_even_at_full_size` builds a stratum of 30 conversations that qualifies cleanly and is still too noisy to believe, and `test_a_holdout_covering_a_corner_of_traffic_does_not_take_over` builds one that is clean and tight but describes 13% of the traffic.

Tests now label conversations throughout, so the cluster gate is satisfied and the coverage/width assertions exercise the gates they name. 80 pass across the output-savings suites, ruff clean.

JerrettDavis (approved):
Re-reviewed the current refresh. The conversation-cluster gate remains intact, and coverage and interval-width checks prevent weak holdouts from replacing the fallback estimate. All 70 output-savings tests pass; the prior blocker remains resolved. I am correcting the description to reflect the current cluster admission rule.

PR #3128, opened 2026-08-19
fix(savings): stop scoring unobserved strata against the global mean

gglucass:
## Description

`BaselineModel.lookup` fell back to the all-requests mean whenever prefix back-off found nothing. That fallback is not a control for anything.

The baseline is seeded once, by `learn --verbosity` reading session history that predates the shaper. It therefore only ever covers the model families the user ran *before* installing. It also cannot be relearned: every transcript written since is already shaped, so a re-learn would collapse the baseline onto the treatment mean. Each family a user adopts later resolves to a single number derived from a different population, permanently.

Measured on a real ledger (15,658 baseline samples over 16 strata, all `opus`; 74,243 treatment requests over 83 strata):

| | |
|---|---|
| Requests scored against the global mean | 48% (every fable, sonnet, haiku, gpt turn) |
| Global mean applied to them | 1,083 output tokens |
| `sonnet\|new_user_ask\|m\|notools` | 4,666 requests, replies average **73** tokens, each credited ~1,010 saved |
| Share of total reported savings from the fallback | **74%** |
| Reported reduction | 42.6% |
| Reduction over strata the baseline actually observed | 32.1% |

Scoring a short no-tool ask against a long tool-calling turn is not a synthetic control, it is a unit conversion.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

**Reported savings will drop** for any user whose traffic has moved beyond the model family their baseline was seeded from. That is the point of the change, but it is a visible number moving, so flagging it explicitly.

## Changes Made

- `lookup` returns `(0.0, 0.0, 0)` when neither the exact stratum nor prefix back-off finds evidence. Both estimators already treat `n == 0` as "no evidence" and skip the request, so what changes is which requests qualify, not how they are scored.
- The global fallback stays reachable via `lookup(key, fall_back_to_global=True)` for callers that want it; nothing in the tree passes it today.
- `glob` is still observed and still backs `total_samples`, so the ledger format and the "has a baseline been seeded" check are unaffected.
- Also in this commit: prefix back-off merges every matching neighbour instead of returning the first in dict order. The old loop made the answer depend on insertion order, so a round-tripped ledger could score the same request differently. Happy to split this out if you would rather review it separately.

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_output_savings.py tests/test_verbosity_learn.py tests/test_output_savings_cli.py -q
tests/test_output_savings.py ........................................... [ 66%]
tests/test_verbosity_learn.py ...................                        [ 95%]
tests/test_output_savings_cli.py ...                                     [100%]
65 passed in 1.12s

All checks passed!

```

`test_lookup_falls_back_to_global` is rewritten into two tests: one pinning the new default, one pinning that the fallback still works when asked for.

## Real Behavior Proof

- Environment: macOS 24.6.0, Python 3.10.18, uv, upstream/main at 17522fb0. Ledger from a shipped desktop build, `~/.headroom/output_savings.json`, 7.4KB.
- Exact command / steps: reimplemented both back-off policies against that ledger and recomputed `estimate_from_baseline` under each. Confirmed the current-policy recomputation reproduces what the running proxy reports on `/stats` before comparing.
- Observed result: current policy reproduces the live `/stats` figures exactly (`tokens_saved` 31,493,556, `baseline_tokens` 73,986,108, 42.6%, 74,297 requests), which validates the recomputation. Under the new policy the same ledger gives 11,540,868 / 35,961,142 / 32.1% over 38,729 requests. Splitting the current-policy total by how each stratum resolved: 8.29M from exact stratum hits, 2.76M from prefix back-off, 20.43M from the global mean -- 74% of the reported savings resting on the fallback this PR removes.
- Not tested: one user's ledger. A baseline seeded from a corpus spanning several model families would lose less; one seeded before a model switch would lose more.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. Baseline lookup is not behind a rollout gate.
- Stable/default behavior changed: Yes, and visibly. Reported savings drop for any user whose traffic has moved past the model family their baseline was seeded from: 42.6% -> 32.1% on the real ledger measured above. Requests with no baseline evidence are now skipped rather than scored against an unrelated mean, so `n_requests` in the estimate drops with it. That is the fix, but it is a number users can see moving.
- Kill switch / disable path: No runtime flag. `lookup(key, fall_back_to_global=True)` restores the previous behavior at the call site, and `glob` is still accumulated, so the old number remains computable from the same ledger.
- Qualification impact: `test_lookup_falls_back_to_global` is split into two tests, one pinning the new default and one pinning the opt-in fallback. No assertion was relaxed.
- Rollback path: Revert the commit. The ledger format is unchanged and every estimate is recomputed from it on read, so the previous figures return immediately with no migration or backfill.

## Review Readiness

### DIFF
--- a/headroom/proxy/output_savings.py
+++ b/headroom/proxy/output_savings.py
@@ -144,12 +144,23 @@ class BaselineModel:
             self.strata.setdefault(key, _Accum()).merge(acc)
         self.glob.merge(other.glob)
 
-    def lookup(self, key: str) -> tuple[float, float, int]:
+    def lookup(self, key: str, *, fall_back_to_global: bool = False) -> tuple[float, float, int]:
         """Return ``(mean, var, n)`` for *key* with hierarchical back-off.
 
-        Falls back by trimming trailing (least-specific) stratum fields, then
-        to the global mean. Back-off keeps the estimate defined for strata the
-        baseline never saw, at the cost of specificity.
+        Falls back by trimming trailing (least-specific) stratum fields, so a
+        stratum the baseline never saw is still scored against its nearest
+        observed neighbours. Returns ``(0.0, 0.0, 0)`` when even that finds
+        nothing -- callers already treat ``n == 0`` as "no evidence".
+
+        ``fall_back_to_global`` restores the old last resort of the
+        all-requests mean. It is off by default because that mean is not a
+        control for anything: the baseline is seeded once, from whatever the
+        user ran *before* installing, so every model family they adopt later
+        resolves to it. On a real ledger that meant 48% of requests -- sonnet
+        and fable turns whose replies average 43-770 tokens -- being scored
+        against one opus-derived mean of 1,083, which alone produced 74% of the
+        reported savings. Scoring a short no-tool ask against a long
+        tool-calling turn is not a synthetic control, it is a unit conversion.
         """
         acc = self.strata.get(key)
         if acc is not None and acc.n > 0:
@@ -157,11 +168,16 @@ class BaselineModel:
         parts = key.split("|")
         while len(parts) > 1:
             parts = parts[:-1]
-            prefix = "|".join(parts)
+            prefix = "|".join(parts) + "|"
+            neighbours = _Accum()
             for k, a in self.strata.items():
-                if k.startswith(prefix + "|") and a.n > 0:
-                    return a.mean, a.var, a.n
-        return self.glob.mean, self.glob.var, self.glob.n
+                if a.n > 0 and k.startswith(prefix):
+                    neighbours.merge(a)
+            if neighbours.n > 0:
+                return neighbours.mean, neighbours.var, neighbours.n
+        if fall_back_to_global:
+            return self.glob.mean, self.glob.var, self.glob.n
+        return 0.0, 0.0, 0
 
     def to_dict(self) -> dict[str, Any]:
         return {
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest estimator change and regression coverage. Removing the unrelated global fallback prevents unsupported model families from receiving fabricated counterfactuals, while hierarchical backoff remains available within an observed family and now aggregates all matching neighbors deterministically instead of depending on insertion order. Callers already exclude 
 == 0, the opt-in legacy fallback remains explicit, and checks are green. No blocking findings.

gglucass:
Fresh production evidence for this from a Windows machine running the 0.37.0 wheel (desktop app, all-Codex traffic on gpt-5.6, verbosity baseline seeded from pre-install Claude sessions only):

- Every treatment stratum is a model family the baseline never observed, so `best_estimate()` scored the whole session against the global mean.
- The Opus-derived global mean (~1.4k output tokens) against short GPT tool-loop replies (averaging ~4 tokens) produced a reported output reduction of ~98-100%. The dashboard rendered it as "Output -100%" and priced it into the day's savings.

This is the same failure measured in the PR description, but in its purest form: a machine where 100% of requests take the fallback, and the credited figure is the only number ever shown. The desktop app now suppresses the backend figure client-side when no stratum has evidence, but that is a workaround; this PR removes the source.

PR #3131, opened 2026-08-19
fix(proxy): stop handing a non-streaming caller an event-stream body

### LINKED ISSUE #3130: Non-streaming request answered with an event stream: buffered path forwards the upstream content-type verbatim
### What happens

With `ANTHROPIC_BASE_URL` pointed at a local Headroom proxy, Claude Code intermittently aborts a turn with:

```
API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or
gateway intercepting the request. Response: content-type event-stream, body is an event
stream (the non-streaming request was answered with a stream), 8756 bytes, request-id
req_011CeC1JTMS8egPL3FBteQay, server cloudflare, intermediary headers anthropic-*
cf-cache-status cf-ray. This was the non-streaming retry of streaming request (no Anthropic
request-id), which failed with: other; 0 stream events received.
```

The reply is **complete** — 8756 bytes, a valid upstream `request-id`. It is simply wearing the
wrong wire format, so the SDK cannot parse it and the turn is lost.

### Why

The buffered (non-streaming) Anthropic path returns the upstream reply with its headers copied
wholesale — `headroom/proxy/handlers/anthropic.py`:

```python
# Remove compression headers since httpx already decompressed the response
response_headers = dict(response.headers)
response_headers.pop("content-encoding", None)
response_headers.pop("content-length", None)   # Length changed after decompression
...
return Response(
    content=response.content,
    status_code=response.status_code,
    headers=response_headers,
)
```

`content-type` rides along untouched. When the upstream answers with `text/event-stream`, a caller
that asked for JSON receives SSE at HTTP 200. The same wholesale copy is what puts `server:
cloudflare`, `cf-ray` and `cf-cache-status` in front of the client — which is the header set the
error message flags as evidence of an intermediary.

Directly above it, the buffered-**stream** (CCR) path already refuses exactly this shape and logs
the offending content-type (#2952):

```python
if buffered_stream_ccr and response.status_code == 200 and not resp_json:
    logger.warning(
        f"[{request_id}] CCR: rejecting malformed buffered 200 reply "
        f"(content-type={response.headers.get('content-type')!r}, "
        f"body_bytes={len(response.content)})"
    )
    return Response(..., status_code=502)   # upstream_protocol_error
```

The plain non-streaming path never got the same treatment. Its sibling comment states the
assumption explicitly:

```python
# DEBUG is right for the buffered non-stream path, where
# an unparseable body is just "no CCR handling".
```

On that path an unparseable body is **not** benign — it is a reply the client cannot consume.

### Reproduction

At the handler seam, stubbing `proxy._retry_request` to return an SSE body for a `stream`-less
request:

```
assert "event-stream" not in response.headers["content-type"]
E   AssertionError: assert 'event-stream' not in 'text/event-stream; charset=utf-8'
```

### Impact

Intermittent and load-dependent — it needs the upstream to answer a buffered request with SSE, so
it surfaces during retries and overload. That makes it worst exactly when sessions can least
afford to lose a turn.

### Environment

- headroom-ai 0.35.0 (pip, venv), `headroom proxy --port 8787 --memory` under `systemd --user`
- Ubuntu 24.04, Linux 6.17, Python 3.12.3
- Client: Claude Code 2.1.235 via `ANTHROPIC_BASE_URL=http://localhost:8787`
- Upstream: `api.anthropic.com`, claude-opus-5 / claude-sonnet-5

### Not established

I could not pin down *why* the upstream answered a `stream`-less request with an event stream, and
I am not claiming Headroom caused that. The defect reported here is downstream of it and holds
regardless: whatever the upstream returns, a caller that did not ask for streaming should never be
handed an SSE body as a 200 it cannot parse.

A PR follows.

Joaovsales:
## Description

The buffered (non-streaming) Anthropic path returns the upstream reply with its headers copied
wholesale, `content-type` included. When the upstream answers a `stream`-less request with
`text/event-stream`, that body reaches a caller that asked for JSON as a `200` it cannot parse.
The reply is **complete** — the client just cannot read it, and the turn is lost.

Directly above it, the buffered-**stream** (CCR) path already refuses this exact shape and logs
the offending content-type (#2952). The plain non-streaming path never got the same treatment;
its sibling comment states the assumption that made it invisible:

```python
# DEBUG is right for the buffered non-stream path, where
# an unparseable body is just "no CCR handling".
```

On that path an unparseable body is not benign — it is a reply the client cannot consume.

Closes #3130

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- **`headroom/proxy/nonstream_sse_policy.py` (new)** — the contract as three pure functions:
  `is_event_stream` (parameter- and case-tolerant media-type test), `should_recover_sse_reply`
  (the gate), `json_reply_headers` (upstream headers with the stale streamed-framing headers
  replaced by JSON). No I/O, no logging, no config — the handler owns those.
- **`headroom/proxy/handlers/anthropic.py`** — on the `resp_json` parse failure, recover instead
  of passing the body through. `_recover_sse_reply` rebuilds the message via
  `StreamingMixin._parse_sse_to_response`, the reconstruction the streaming path already runs for
  usage accounting, so no new parsing surface is added. When the frames yield nothing usable the
  request is refused with `upstream_protocol_error`, matching the vocabulary the CCR guard above
  already uses.

### Design notes

**Recovery happens in place, not as an early return.** It runs before CCR, turn hooks, the
security scan and usage accounting, so the rest of the buffered path sees a normal reply rather
than being skipped. A turn the provider already produced and billed stays a real turn.

**Three deliberate no-ops**, each pinned by its own test:
- a streaming caller is untouched — it asked for SSE, and rewriting it would break the turn;
- a JSON content-type is untouched — already correct;
- **non-200 is untouched** — an error status is already actionable, and its payload should reach
  the client verbatim. This keeps the change off the error path entirely.

**Correlation headers survive.** `request-id`, the `anthropic-*` family and the rate-limit headers
are preserved through recovery; only `content-type`, `content-length`, `content-encoding` and
`transfer-encoding` are dropped, because after the body is rebuilt as one JSON document they
describe nothing and a stale value is worse than an absent one. Dropping `request-id` to fix a
content-type would trade one defect for another.

## Testing

- [ ] Type checking passes (`mypy headroom`) — not run, see Not tested

### Test Output

**Before the fix** — handler tests against an unmodified `anthropic.py`, reproducing the reported
symptom exactly:

```text
$ pytest tests/test_nonstream_sse_policy.py -q -k TestNonStreamingCallerNeverGetsAnEventStream
tests/test_nonstream_sse_policy.py F.F.                                  [100%]

_ TestNonStreamingCallerNeverGetsAnEventStream.test_sse_reply_is_recovered_as_json _
tests/test_nonstream_sse_policy.py:241: in test_sse_reply_is_recovered_as_json
    assert "event-stream" not in response.headers["content-type"]
E   AssertionError: assert 'event-stream' not in 'text/event-...harset=utf-8'
E     'event-stream' is contained here:
E       text/event-stream; charset=utf-8

_ TestNonStreamingCallerNeverGetsAnEventStream.test_unrecoverable_event_stream_is_refused_not_forwarded _
tests/test_nonstream_sse_policy.py:276: in test_unrecoverable_event_stream_is_refused_not_forwarded
    assert response.status_code == 502
E   assert 200 == 502
E    +  where 200 = <Response [200 OK]>.status_code

================== 2 failed, 2 passed, 18 deselected in 5.44s ==================
```

The two that pass in both runs are controls: an ordinary JSON reply, and the correlation-header
assertion.

**After the fix:**

```text
$ pytest tests/test_nonstream_sse_policy.py -q
tests/test_nonstream_sse_policy.py ......................                [100%]
============================= 22 passed in 15.73s ==============================
```

**Neighbouring suites:**

```text
$ pytest tests/test_nonstream_sse_policy.py tests/test_anthropic_compaction_transforms.py \
         tests/test_anthropic_ccr_workspace_unbound.py tests/test_5xx_accounting_all_providers.py \
         tests/test_sse_byte_buffer_policy.py -q
============================== 43 passed in 6.27s ==============================
```

**Lint:**

```text
$ ruff check headroom/proxy/nonstream_sse_policy.py headroom/proxy/handlers/anthropic.py \
             tests/test_nonstream_sse_policy.py
All checks passed!
$ ruff format --check <same three files>
3 files already formatted
```

A broader sweep (`-k "anthropic or sse or buffered or ccr"`) ran 1382 passed / 12 failed. All 12
are in `tests/test_transforms/test_code_compressor.py` and are pre-existing in my environment —
the same file fails 43 tests on a checkout with this commit reverted. Missing tree-sitter grammars
locally; untouched by this change.

## Real Behavior Proof

- **Environment:** Ubuntu 24.04, Linux 6.17, Python 3.12.3. headroom-ai 0.35.0 installed via pip
  into a venv, running as `headroom proxy --port 8787 --memory` under `systemd --user`. Client:
  Claude Code 2.1.235 with `ANTHROPIC_BASE_URL=http://localhost:8787`. Upstream: `api.anthropic.com`
  (claude-opus-5, claude-sonnet-5).

- **Before — real client failure, not synthetic:**

  ```
  API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or
  gateway intercepting the request. Response: content-type event-stream, body is an event
  stream (the non-streaming request was answered with a stream), 8756 bytes, request-id
  req_011CeC1JTMS8egPL3FBteQay, server cloudflare, intermediary headers anthropic-*
  cf-cache-status cf-ray. This was the non-streaming retry of streaming request (no Anthropic
  request-id), which failed with: other; 0 stream events received.
  ```

  8756 bytes and a valid upstream `request-id` — the reply arrived intact and was discarded on
  wire format alone. The `server: cloudflare` / `cf-ray` / `cf-cache-status` headers the client
  flags as intermediary evidence are the same wholesale `dict(response.headers)` copy.

- **Exact steps after the patch:** applied the equivalent hunks to the installed 0.35.0 (the
  release predates the #2952 branch, so the anchor differs but the change is the same), then:

  ```
  $ systemctl --user restart headroom.service
  listening again after ~8s
  $ systemctl --user is-active headroom.service
  active
  ```

- **Observed result:** the patched build has been serving this workstation's live Claude Code
  traffic since the restart with no protocol failures:

  ```
  $ awk '$0 >= "2026-08-19 08:42:34"' ~/.headroom/logs/proxy.log \
      | grep "path=/v1/messages" | grep -oE "status=[0-9]+" | sort | uniq -c
       17 status=200
  ```

  No startup errors or tracebacks in `proxy.log`.

- **Not tested:**
  - **I could not reproduce the upstream-returns-SSE condition on demand.** The originating client
    failure above is real and captured, but my after-fix reproduction is at the handler seam
    (stubbed `_retry_request`), not against a live upstream emitting SSE. So the live evidence
    shows the patched build is healthy on the normal path; it does not show a production recovery.
  - **Why the upstream answered a `stream`-less request with an event stream is unresolved,** and I
    am not claiming Headroom caused it. This PR fixes the downstream handling, which holds either
    way.
  - The streaming forwarder's own `status >= 400` header passthrough in `streaming.py` has the same
    shape. I left it alone: there the client *did* ask for SSE, and forwarding an upstream JSON
    error verbatim is correct, since SDKs check status before parsing frames. Flagging it only so a
    maintainer can confirm that reading.
  - Non-Anthropic providers (`openai.py`, `gemini.py`, `bedrock.py`) — not examined.
  - `mypy headroom` — not run; the dev extra was not installed in this environment.

## Runtime Rollout Safety

- **Rollout-managed feature(s):** none — correctness fix, always on.
- **Stable/default behavior changed:** only for a reply that is currently unusable. A non-streaming
  caller that would have received an SSE `200` now receives the recovered JSON, or a `502
  upstream_protocol_error` when the frames yield nothing. Every other reply is byte-identical.
- **Kill switch / disable path:** none added. The code is inert unless a non-streaming request
  returns `200` with `content-type: text/event-stream`; adding a flag would mean a switch whose
  "off" position restores an unparseable response.
- **Unsafe override required:** no.
- **Qualification impact:** none expected — the buffered JSON path is untouched (control test).
- **Rollback path:** revert this commit. No migration, no persisted state, no config.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — n/a, no user-facing surface

## Additional Notes

Happy to split the policy module into the handler if a separate file is more than this deserves —
I put it out of line to match `sse_byte_buffer_policy.py` and the other `*_policy.py` modules, and
because the gate's three negative arms are each a real risk worth a named test.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -51,6 +51,10 @@ from headroom.proxy.image_isolation import run_image_compression_isolated
 from headroom.proxy.memory_decision import MemoryDecision
 from headroom.proxy.memory_query import MemoryQuery
 from headroom.proxy.model_router import estimate_input_tokens
+from headroom.proxy.nonstream_sse_policy import (
+    json_reply_headers,
+    should_recover_sse_reply,
+)
 from headroom.proxy.outcome import RequestOutcome
 
 logger = logging.getLogger("headroom.proxy")
@@ -383,6 +387,37 @@ class AnthropicHandlerMixin:
         except Exception:  # pragma: no cover - defensive
             logger.debug("CCR: marker probe failed; keeping the buffered path", exc_info=True)
             return True
+
+    def _recover_sse_reply(
+        self, response: httpx.Response, request_id: str
+    ) -> dict[str, Any] | None:
+        """Rebuild a Messages JSON object from an SSE body sent to a JSON caller.
+
+        Reuses ``StreamingMixin._parse_sse_to_response`` — the reconstruction
+        the streaming path already runs for usage accounting — so no new
+        parsing surface is introduced. Returns ``None`` when the frames do not
+        yield a usable message, which the caller must treat as a refusal
+        rather than a reason to pass the SSE body through.
+
+        HeadroomProxy inherits StreamingMixin, but this handler is also tested
+        and embedded without it, so the helper is resolved defensively.
+        """
+        parse_sse = getattr(self, "_parse_sse_to_response", None)
+        if parse_sse is None:  # pragma: no cover - defensive
+            logger.debug(f"[{request_id}] SSE recovery: StreamingMixin unavailable")
+            return None
+        try:
+            recovered = parse_sse(response.text, "anthropic")
+        except Exception:  # pragma: no cover - defensive
+            logger.debug(f"[{request_id}] SSE recovery: parse raised", exc_info=True)
+            return None
+        # ``_parse_sse_to_response`` always returns a skeleton dict, so a
+        # truthy result proves nothing. ``id`` is only ever set from
+        # ``message_start``, which makes it the cheapest evidence that real
+        # frames were seen rather than a truncated or empty body.
+        if not recovered or not recovered.get("id"):
+            return None
+        return recovered
 
     @staticmethod
     def _extract_anthropic_cache_ttl_metrics(usage: dict[str, Any] | None) -> tuple[int, int]:
@@ -3877,6 +3912,57 @@ class AnthropicHandlerMixin:
                                     f"[{request_id}] CCR: buffered stream:false request got a "
                                     f"non-JSON {response.status_code} reply "
                                     f"(content-type={response.headers.get('content-type')!r}): {e}"
+                                )
+                            elif should_recover_sse_reply(
+                                client_requested_stream=stream,
+                                status_code=response.status_code,
+                                content_type=response.headers.get("content-type"),
+                            ):
+                                # Same wire-format mismatch as the buffered-stream
+                                # case above; only the direction differs. This
+                                # caller never asked for SSE, and the headers
+                                # further down are copied from upstream verbatim —
+                                # so passing this through hands it a 200 in a
+                                # format it cannot parse, and the turn is lost even
+                                # though the reply arrived complete. Rebuild the
+                                # message in place so CCR, turn hooks, the security
+                                # scan and usage accounting all see a normal reply.
+                                recovered = self._recover_sse_reply(response, request_id)
+                                if recovered is None:
+                                    logger.error(
+                                        f"[{request_id}] Upstream answered a non-streaming "
+                                        f"{response.status_code} request with an unrecoverable "
+                                        f"event stream "
+                                        f"(content-type={response.headers.get('content-type')!r}, "
+                                        f"body_bytes={len(response.content)}): {e}"
+                                    )
+                                    return Response(
+                                        content=json.dumps(
+                                            {
+                                                "type": "error",
+                                                "error": {
+                                                    "type": "upstream_protocol_error",
+                                                    "message": (
+                                                        "Upstream answered a non-streaming "
+                                                        "request with an event stream that "
+                                                        "could not be recovered."
+                                                    ),
+                                                },
+                                            }
+                                        ).encode(),
+                                        status_code=502,
+                                        media_type="application/json",
+                                    )
+                                logger.warning(
+                                    f"[{request_id}] Upstream answered a non-streaming request "
+                                    f"with an event stream; recovered the message from "
+                                    f"{len(response.content)} SSE bytes"
+                                )
+                                resp_json = recovered
+                                response = httpx.Response(
+                                    status_code=200,
+                                    content=json.dumps(recovered).encode(),
+                                    headers=json_reply_headers(dict(response.headers)),
                                 )
                             else:
                                 logger.debug(
--- a/headroom/proxy/nonstream_sse_policy.py
+++ b/headroom/proxy/nonstream_sse_policy.py
@@ -1,0 +1,135 @@
+"""Wire-format contract policy for the buffered (non-streaming) reply path.
+
+The problem
+-----------
+
+The buffered Anthropic path returns the upstream reply with its headers
+copied wholesale::
+
+    response_headers = dict(response.headers)
+    response_headers.pop("content-encoding", None)
+    response_headers.pop("content-length", None)
+    ...
+    return Response(content=..., status_code=..., headers=response_headers)
+
+``content-type`` rides along untouched. When the upstream answers a
+``stream``-less request with ``text/event-stream``, that body reaches a
+caller that asked for JSON, as a ``200`` it cannot parse. Clients report
+it as an empty or malformed response and the turn is lost — the reply is
+*present and complete*, just wearing the wrong wire format.
+
+The buffered-stream (CCR) path already refuses this shape, logging the
+offending ``content-type`` and returning ``upstream_protocol_error``
+(#2952). The plain non-streaming path never got the same treatment: an
+unparseable body there was assumed to mean "no CCR handling", so it was
+logged at DEBUG and passed through.
+
+The contract
+------------
+
+A caller that did not set ``stream: true`` must never receive an
+event-stream body. Headroom owns both ends of that boundary, so it can
+enforce it rather than let the mismatch reach the client.
+
+Behaviour matrix
+----------------
+
+============================  ==============  =========  ====================
+Client asked for streaming?   Upstream C-T    Status     Result
+============================  ==============  =========  ====================
+yes                           any             any        untouched
+no                            application/…   any        untouched
+no                            text/event-…    != 200     untouched (real error)
+no                            text/event-…    200        recover, else refuse
+============================  ==============  =========  ====================
+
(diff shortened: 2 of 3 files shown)


chopratejas:
Superseded by #3142, which is merged — but this PR is in it, and thanks for it.

**What came from here:** `nonstream_sse_policy.py` ships essentially as you wrote it. The behaviour matrix is the clearest statement of the contract anyone put on this bug, and framing it as one pure predicate — with the three negative arms argued explicitly (a streaming caller *wants* SSE; a JSON content-type is already right; a non-200 carries an error the client should see verbatim) — is what made the decision site readable instead of a stack of inline conditions. Your test suite ships unmodified and passes against the merged implementation. You're credited via `Co-authored-by:`.

**What was taken from #3132 instead, and why** — these are the honest reasons, since you deserve them rather than a wave at "we combined them":

1. **Truncation.** The recovery guard here is `recovered.get("id")`, and `id` is only ever set from `message_start`. That proves the stream *began*, not that it *finished* — so a stream cut off mid-way passes, and the client gets HTTP 200 with a silently short assistant turn. #3132's `require_complete` additionally demands a terminal `message_stop`, every opened block closed, no in-band `error` event, and no delta the reconstructor can't replay, else 502. For a reconstruction handed to the client *as* the turn, a visible failure beats invisible truncation.

2. **The `index` field.** `_parse_sse_to_response` writes `"index"` into every content block (`streaming.py:425`). Returning that parser output directly means the client persists the turn and echoes it back — and `anthropic.py:185` already documents what happens then: a 400, `content.0.text.index: Extra inputs are not permitted`, *"aborts multi-turn sessions once a client echoes a reconstructed assistant turn back."* In fairness the existing inbound stripper masks this while the proxy is in the path, so it pollutes stored history rather than breaking the next turn outright — but it's the kind of thing that surfaces later, somewhere else.

3. **`json_reply_headers`.** This is the one I'd flag even outside the merge: the hand-rolled framing list leaves `connection`, `keep-alive` and `server` on the response. `helpers.sanitize_forwarded_response_headers` already owns `FRAMING_RESPONSE_HEADERS` and strips all three — and `server: cloudflare` is one of the very headers the reporter's client cites as evidence of an intermediary. #3142 delegates to that helper, so your module keeps the predicate and drops the header function.

One more thing worth knowing, found while comparing the two: main's parser required `startswith("data: ")` **with the space that the SSE spec makes optional**, and `continue`d otherwise — so a spec-valid stream framed without it parsed as *zero events*, which is literally the "0 stream events received" in the report. #3132 fixed that too. It's plausible this PR alone wouldn't have resolved the bug in the reporter's CDN environment.

Closing as superseded, not as rejected. Genuinely good work on the contract — please keep filing.

Issue #3078, opened 2026-08-17
[BUG] Opencode + GitHub Copilot + claude-sonnet: first call succeeds, second fails (api_error) after request mutation / prefix change


mars-peng-lb:
## Description

When routing GitHub Copilot's Anthropic-compatible endpoint (`api.githubcopilot.com`, `/v1/messages`, model `claude-sonnet-4.6`) through the headroom proxy, the **first** LLM call in an opencode session succeeds, but the **second** (and often subsequent) call fails with `api_error: "An error occurred while processing the request."` from the Copilot gateway.

Failed requests correlate consistently with headroom mutating the request body between calls: every failed request shows `CACHE-BUST: reason=prefix_change` and `body_mutated=true` with `mutation_reasons=structural_diff_vs_original,ccr_streaming_retrieve_buffered_non_stream`, while the successful first call was sent unmutated with a cache write.

## To Reproduce

1. Install headroom (0.35.0) with the opencode plugin
2. Configure opencode's `github-copilot` provider to route through the local headroom proxy
3. Start a session with `claude-sonnet-4.6` and send a first message → succeeds (unmutated body, cache write)
4. Send a second message in the same session → fails with `api_error`

## Expected Behavior

The compressed/mutated request should be accepted by the Copilot upstream, or the prefix should be kept stable so the mutation doesn't invalidate the provider cache on every call.

## Actual Behavior

- First call: body sent essentially unmutated, succeeds, writes ~22k–28k tokens to Copilot's prompt cache
- Second call: content_router/compressor rewrites older messages → prefix changes → Copilot gateway returns a generic `api_error`; headroom logs an inbound `status=200` with `tok_out=0` while the client receives the upstream error
- Intermittent: some windows (e.g. 29 consecutive calls in one session) succeed, other windows every second call fails — suggesting the gateway is sensitive to the specific mutation pattern

## Error Output
Client log:
level=ERROR message="stream error" providerID=github-copilot modelID=claude-sonnet-4.6
error.error.type=api_error error.error.message="An error occurred while processing the request."

## Headroom proxy log, failed request:
event=outbound_request forwarder=anthropic_messages method=POST path=/v1/messages
  body_mutated=true
  mutation_reasons=structural_diff_vs_original,ccr_streaming_retrieve_buffered_non_stream
CCR: stream:true request has headroom_retrieve available; using buffered stream:false upstream request for server-side retrieval handling
[...]

PR #3067, opened 2026-08-17
chore: release 0.36.0

JerrettDavis:
:robot: I have created a release *beep* *boop*
---

### Features

* add deterministic runtime rollout controls ([#1490](https://github.com/headroomlabs-ai/headroom/issues/1490)) ([3077ac8](https://github.com/headroomlabs-ai/headroom/commit/3077ac81e8ef3ddefebbe308ea37a4e9bb2100e6))
* **proxy:** let extensions report cost savings and their own latency ([#3051](https://github.com/headroomlabs-ai/headroom/issues/3051)) ([f9807fd](https://github.com/headroomlabs-ai/headroom/commit/f9807fd69e220f43068ec168515ae886dd36166f))
* **proxy:** unify savings attribution across stats, perf, metrics, and dashboard ([1b0b0b8](https://github.com/headroomlabs-ai/headroom/commit/1b0b0b89a4bf751c8bd592890aef9c3b339e8e37)), closes [#2976](https://github.com/headroomlabs-ai/headroom/issues/2976)
* **wrap/claude:** make the --1m fallback model configurable via HEADROOM_1M_MODEL ([#2983](https://github.com/headroomlabs-ai/headroom/issues/2983)) ([2a84725](https://github.com/headroomlabs-ai/headroom/commit/2a8472525d3a027c95dc38a10c4b6707b482cabc))

### Bug Fixes

* **anthropic:** honor the [1m] 1M-context tier, and price it correctly ([#3073](https://github.com/headroomlabs-ai/headroom/issues/3073)) ([6d2254d](https://github.com/headroomlabs-ai/headroom/commit/6d2254dfb5eb97f92249e0ee7aa04b2697adfa69))
* **ccr:** make --no-ccr disable server-side response handling too ([#3101](https://github.com/headroomlabs-ai/headroom/issues/3101)) ([131b119](https://github.com/headroomlabs-ai/headroom/commit/131b119c053e66fe825dabb3c242f6dc5c6049d7)), closes [#3082](https://github.com/headroomlabs-ai/headroom/issues/3082)
* **ccr:** make StreamingCCRHandler work on OpenAI streams ([#3069](https://github.com/headroomlabs-ai/headroom/issues/3069)) ([7ef736f](https://github.com/headroomlabs-ai/headroom/commit/7ef736fb1a8852a3dee52a362043c47084628a2a))
[...]

PR #3132, opened 2026-08-19
fix(proxy/anthropic): stop answering a non-streaming turn with an event stream

### LINKED ISSUE #3130: Non-streaming request answered with an event stream: buffered path forwards the upstream content-type verbatim

taiseii:
## Description

Claude Code retries a failed streaming turn as `stream: false`. When the upstream answers that JSON request with `content-type: text/event-stream`, the buffered Anthropic path relays it verbatim, so the SDK gets a wire format it never asked for and a complete, already-billed turn is lost:

```
API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or
gateway intercepting the request. Response: content-type event-stream, body is an event
stream (the non-streaming request was answered with a stream), 8756 bytes, request-id
req_011CeC1JTMS8egPL3FBteQay, server cloudflare, intermediary headers anthropic-*
cf-cache-status cf-ray. This was the non-streaming retry of streaming request (no Anthropic
request-id), which failed with: other; 0 stream events received.
```

Two defects, one on each side of the hop.

**Request side.** #3078 normalized `Accept` only when Headroom itself flips `stream: true` to `false` for CCR (inside `if buffered_stream_ccr`). A *client-originated* non-stream turn never passed through that block, so it kept `Accept: text/event-stream` and put a self-contradicting request on the wire — "answer as JSON" in the body, "I only accept SSE" in the headers. The rewrite moves to the buffered boundary, which every non-streaming request reaches, so the CCR flip and the client's own turn are covered by one rule. Header-only: the body is untouched, so cache keys and breakpoints do not move.

**Response side.** `stream: false` should already select JSON, so an SSE reply is an upstream protocol violation whichever header provoked it — this half is what makes the turn recoverable regardless. The buffered path now adapts the reply to the JSON the caller asked for, reusing the existing `_looks_like_sse_response` detector so a gateway relaying SSE under a vaguer content-type is caught too.

Adaptation happens **before the first `response.json()`** — ahead of CCR retrieval, memory, turn hooks, usage and cost accounting, prefix tracking, the response cache, marker resolution and the security scan. Adapting at the final return would leave every one of those reading an unparseable body: zero usage recorded, markers unresolved, raw SSE in the cache.

The guard keys on what the **client** asked for, not what went upstream. A buffered CCR turn deliberately requests JSON on behalf of a streaming client and re-emits SSE further down; that path is untouched, with an existing test pinning it.

Closes #3130

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `handlers/anthropic.py`: force `Accept: application/json` at the buffered boundary whenever the outgoing body is `stream: false`; removed the narrower CCR-only rewrite it supersedes (the CCR flip reaches the same boundary, and `tests/test_buffered_ccr_accept_header.py` still passes unchanged).
- `handlers/anthropic.py`: new `_adapt_event_stream_to_json()`, called before the first `response.json()` when a non-streaming client gets a 200 that `_looks_like_sse_response()`. Faithful reconstruction is returned as JSON; anything else becomes `502 upstream_protocol_error` rather than a plausible-looking 200 carrying a silently short message.
- `handlers/anthropic.py`: the buffered return now uses `sanitize_forwarded_response_headers()` instead of popping two headers by hand. It was also replaying `transfer-encoding`, `connection` and `keep-alive` — the stale chunked framing behind the empty HTTP 200 in #3019 — plus `server`. `cf-*` is dropped locally rather than added to `FRAMING_RESPONSE_HEADERS`, since those are hop-by-hop by definition and CDN provenance is not.
- `handlers/streaming.py`: `_parse_sse_to_response()` gains `require_complete`. Strict mode demands `message_start`, a terminal `message_stop`, every opened block closed, no `error` event, and no delta type the reconstructor cannot replay; it also strips the streaming-only `index` that Anthropic rejects on the next request (`content.0.text.index: Extra inputs are not permitted`). Existing streaming callers keep the permissive default. Shared by both modes: CRLF normalization and accepting `data:` without the space.
- `tests/test_anthropic_buffered_sse.py`: new, 18 cases.

## Testing

### Test Output

```text
$ pytest tests/test_anthropic_buffered_sse.py -q
tests/test_anthropic_buffered_sse.py ..................                  [100%]
======================== 18 passed, 1 warning in 6.09s =========================

$ pytest -q tests/test_proxy tests/test_*{anthropic,sse,stream,ccr,cache,proxy}*.py
FAILED tests/test_cli_proxy_improvements.py::TestMissingProxyDepsError::test_proxy_command_exits_when_mcp_missing
FAILED tests/test_cli_proxy_improvements.py::TestMissingProxyDepsError::test_ensure_proxy_dependencies_exits_when_fastapi_missing
===== 2 failed, 2347 passed, 113 skipped, 9 warnings in 515.49s (0:08:35) ======

$ pytest -q          # whole suite
== 27 failed, 11641 passed, 273 skipped, 5838 warnings in 4491.02s (1:14:51) ===

# None of those failures come from this change. Re-running exactly the failing set
# with the two changed files reverted to origin/main gives an identical result, so
# they are pre-existing (kompress health, graceful shutdown, crewai/strands
# integrations, an evals benchmark, and the missing-proxy-deps CLI tests; the four
# that pass in isolation are order-dependent):

$ git checkout origin/main -- headroom/proxy/handlers/{anthropic,streaming}.py
$ pytest -q --tb=no $(cat failed.txt)
=================== 9 failed, 4 passed, 4 warnings in 32.57s ===================

$ git checkout HEAD -- headroom/proxy/handlers/{anthropic,streaming}.py
$ pytest -q --tb=no $(cat failed.txt)
=================== 9 failed, 4 passed, 4 warnings in 23.55s ===================

$ ruff check .
All checks passed!

$ mypy headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/streaming.py
```

The new tests were confirmed red before the change — with the two source files stashed, 15 of the then-17 failed; all pass after.

## Real Behavior Proof

- Environment: macOS 15 (arm64), Python 3.13.2, `headroom proxy` on `127.0.0.1:8787`, `rust_core: loaded`. A stand-in upstream on `127.0.0.1:9911` answers every request with the #3130 shape — HTTP 200, `content-type: text/event-stream`, plus `server: cloudflare`, `cf-ray`, `cf-cache-status`, `request-id` and `anthropic-ratelimit-requests-remaining` — and logs the `Accept` it received. The proxy is aimed at it with `x-headroom-base-url`.

- Exact command / steps: one `curl` at the running proxy with `stream: false` in the body and `accept: text/event-stream` in the headers, aimed at the stand-in upstream via `x-headroom-base-url`; run once on this branch, and once with `origin/main` checked out over the two changed files. Full command:

```bash
curl -s -D - http://127.0.0.1:8787/v1/messages \
  -H 'content-type: application/json' \
  -H 'accept: text/event-stream' \
  -H 'x-api-key: test-key' \
  -H 'anthropic-version: 2023-06-01' \
  -H 'x-headroom-base-url: http://127.0.0.1:9911' \
  -d '{"model":"claude-sonnet-4-6","max_tokens":64,"stream":false,
       "messages":[{"role":"user","content":"go"}]}'
```

- Observed result: before the change the client receives `content-type: text/event-stream` with a raw SSE body plus `cf-ray`, `cf-cache-status` and `server: cloudflare`, and the upstream logs `accept='text/event-stream'` — the reported failure, reproduced end to end. After the change the upstream logs `accept='application/json'` and the client receives a parseable JSON message with usage intact, `index` stripped, `request-id` and `anthropic-ratelimit-*` preserved, and the CDN headers gone. Full transcripts:

*Before* (`origin/main` checked out over the two files, same request, same upstream) — the reported failure, reproduced end to end:

```text
HTTP/1.1 200 OK
server: uvicorn
server: BaseHTTP/0.6 Python/3.14.3, cloudflare
content-type: text/event-stream
cf-ray: 9abc123def-AMS
cf-cache-status: DYNAMIC
request-id: req_011CeC1JTMS8egPL3FBteQay

event: message_start
data: {"type":"message_start","message":{"id":"msg_live", ...

# upstream log:
UPSTREAM SAW accept='text/event-stream' body_stream=false
```

*After* (this branch):

```text
HTTP/1.1 200 OK
server: uvicorn
request-id: req_011CeC1JTMS8egPL3FBteQay
anthropic-ratelimit-requests-remaining: 42
content-type: application/json
content-length: 281

{"content":[{"type":"text","text":"live proof"}],
 "usage":{"input_tokens":11,"output_tokens":7,"cache_read_input_tokens":0,
          "cache_creation_input_tokens":0},
 "id":"msg_live","type":"message","stop_sequence":null,
 "model":"claude-sonnet-4-6","role":"assistant","stop_reason":"end_turn"}

# upstream log:
UPSTREAM SAW accept='application/json' body_stream=false
```

Both halves are visible: the upstream now receives `Accept: application/json`, and the client gets JSON with `index` stripped, usage intact, `request-id` and `anthropic-ratelimit-*` preserved, and `server: cloudflare` / `cf-*` gone.

- Not tested: a live reproduction against the real Anthropic endpoint. The SSE-at-200 answer is intermittent and I could not provoke it on demand, so the upstream side of the proof is a stand-in that emits exactly the reported shape. The 502 path was exercised in tests, not live.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — no flag or channel gate is introduced.
- Stable/default behavior changed: yes, on the buffered Anthropic path. (1) A `stream: false` request now always leaves with `Accept: application/json`. (2) A 200 that looks like SSE for a non-streaming client is adapted to JSON, or 502s if it cannot be faithfully reconstructed — previously it was relayed as-is, which is the bug. (3) `transfer-encoding`, `connection`, `keep-alive`, `server` and `cf-*` are no longer replayed downstream. Streaming clients, the buffered-CCR SSE relay, translated backends and cache hits are unchanged.
- Kill switch / disable path: none. The prior behavior is the defect, so a flag would only preserve a broken path; `x-headroom-bypass` / `x-headroom-mode: passthrough` still converge on the same forwarder and are equally covered. Happy to add a gate if maintainers want one.
- Qualification impact: none expected. The 502 replaces a response the client already treats as a hard failure, so it converts a silent malformed 200 into a legible error.
- Rollback path: revert this commit; it is self-contained across two source files plus one new test file.

## Review Readiness

## Checklist

## Additional Notes

- Documentation unchecked: this is an internal wire-protocol correction with no user-facing surface, so there is nothing in the docs to amend. Screenshots n/a.
- The `Accept` mismatch is the most credible trigger for the SSE reply but is not proven from the code — `stream: false` alone should select JSON. That is why both halves are here: the request side removes the contradiction, the response side makes the turn recoverable no matter what provoked it.
- This does not explain the *preceding* streaming failure that delivered zero events and caused Claude Code to retry non-streaming in the first place. That remains a separate trigger, and I did not investigate it here.
- Open question for maintainers: strict reconstruction currently returns 502 on an incomplete stream. The alternative is returning the partial message with `stop_reason: null`. I chose the 502 because a silently truncated answer is harder to notice than an error, but I will switch it if you prefer salvage.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -226,6 +226,66 @@ def _looks_like_sse_response(response: httpx.Response) -> bool:
 class AnthropicHandlerMixin:
     """Mixin providing Anthropic API handler methods for HeadroomProxy."""
 
+    def _adapt_event_stream_to_json(
+        self,
+        response: httpx.Response,
+        request_id: str,
+    ) -> httpx.Response:
+        """Rebuild an SSE reply as the JSON a non-streaming caller asked for.
+
+        A caller that sent ``stream: false`` cannot parse ``text/event-stream``,
+        so relaying it verbatim loses a turn the upstream already charged for
+        (#3130). Reconstruction is strict: a truncated stream, or one carrying
+        an ``error`` event, becomes an explicit 502 rather than a successful
+        HTTP 200 whose message is silently short.
+        """
+        headers = {
+            k: v
+            for k, v in sanitize_forwarded_response_headers(
+                response.headers,
+                "content-type",
+            ).items()
+            if not k.lower().startswith("cf-")
+        }
+
+        parsed = None
+        try:
+            parsed = self._parse_sse_to_response(
+                response.content.decode("utf-8", "replace"),
+                "anthropic",
+                require_complete=True,
+            )
+        except Exception as exc:  # pragma: no cover - defensive
+            logger.warning(f"[{request_id}] SSE->JSON reconstruction raised: {exc}")
+
+        if parsed is None:
+            logger.error(
+                f"[{request_id}] Upstream answered a non-streaming request with an "
+                f"event stream that could not be faithfully reconstructed "
+                f"(body_bytes={len(response.content)}); returning 502 rather than "
+                f"a wire format the client cannot parse"
+            )
+            return httpx.Response(
+                502,
+                json={
+                    "type": "error",
+                    "error": {
+                        "type": "upstream_protocol_error",
+                        "message": (
+                            "Upstream answered a non-streaming request with an "
+                            "incomplete event stream."
+                        ),
+                    },
+                },
+                headers=headers,
+            )
+
+        logger.info(
+            f"[{request_id}] Upstream answered a non-streaming request with an "
+            f"event stream; adapted {len(response.content)} bytes of SSE to JSON"
+        )
+        return httpx.Response(200, json=parsed, headers=headers)
+
     async def _count_tokens_offloaded(self, model, messages):  # noqa: ANN001, ANN201
         from headroom.proxy.token_counting import count_tokens_offloaded
 
@@ -3509,20 +3569,9 @@ class AnthropicHandlerMixin:
                         body_mutation_tracker.mark_mutated(
                             "ccr_streaming_retrieve_buffered_non_stream"
                         )
-                    # The body now asks for a non-streaming reply, so the
-                    # client's ``Accept: text/event-stream`` no longer describes
-                    # the response being requested. Forwarding it unchanged
-                    # sends upstream a self-contradicting request: "answer as
-                    # JSON" in the body, "I only accept SSE" in the headers.
-                    #
-                    # Anthropic tolerates that. Stricter Anthropic-compatible
-                    # gateways do not: GitHub Copilot's returns a generic
-                    # ``api_error``, which is why a session's first call
-                    # succeeded and the next one — the first to carry a
-                    # redeemable marker, and so the first to be buffered —
-                    # failed (#3078).
-                    _accept_key = next((k for k in headers if k.lower() == "accept"), "accept")
-                    headers[_accept_key] = "application/json"
+                    # The ``Accept`` rewrite this flip used to do lives at the
+                    # buffered boundary below, which every non-streaming
+                    # request reaches — this one and the client's own (#3130).
                     logger.info(
                         f"[{request_id}] CCR: stream:true request has "
                         "headroom_retrieve available; using buffered stream:false "
@@ -3681,6 +3730,29 @@ class AnthropicHandlerMixin:
                         session_key=session_key,
                     )
                 else:
+                    # Whatever set it — the client's own ``stream: false`` or
+                    # the CCR flip above — this branch sends a non-streaming
+                    # request, so the client's ``Accept: text/event-stream`` no
+                    # longer describes what is being asked for. Forwarding it
+                    # unchanged puts a self-contradicting request on the wire:
+                    # "answer as JSON" in the body, "I only accept SSE" in the
+                    # headers.
+                    #
+                    # Anthropic tolerates the contradiction; stricter
+                    # Anthropic-compatible gateways do not — GitHub Copilot's
+                    # answers a generic ``api_error`` (#3078). On a
+                    # client-originated non-stream turn — Claude Code's retry
+                    # after a failed stream — an SSE answer to a JSON request
+                    # is the empty/malformed HTTP 200 of #3130.
+                    #
+                    # Mutated in place: ``headers`` is captured by the
+                    # closures defined below, and rebinding it here would
+                    # leave them holding the old mapping.
+                    if body.get("stream", False) is False:
+                        for _accept_key in [k for k in headers if k.lower() == "accept"]:
+                            headers.pop(_accept_key, None)
+                        headers["accept"] = "application/json"
+
                     # Populated once the upstream answers 200 with parseable
                     # JSON, so the guard below can fall back to it (#3088).
                     _salvageable_upstream: dict[str, Any] = {}
@@ -3850,6 +3922,28 @@ class AnthropicHandlerMixin:
                                     logger.error(
                                         f"[{request_id}] Failed to write debug dump: {dump_err}"
                                     )
+
+                        # A non-streaming request answered with an event
+                        # stream (#3130). The turn is complete and already
+                        # paid for — it is just wearing the wrong wire
+                        # format — so adapt it to the JSON this caller asked
+                        # for *here*, ahead of everything that reads the
+                        # body: CCR retrieval, memory, turn hooks, usage and
+                        # cost accounting, prefix tracking, the response
+                        # cache, marker resolution and the security scan.
+                        # Adapting at the final return instead would leave
+                        # every one of those looking at an unparseable body.
+                        #
+                        # ``stream`` is what the *client* asked for, not what
+                        # went upstream: a buffered CCR turn deliberately
+                        # requests JSON on behalf of a streaming client and
+                        # re-emits SSE further down, and must keep doing so.
+                        if (
+                            response.status_code == 200
+                            and not stream
+                            and _looks_like_sse_response(response)
+                        ):
+                            response = self._adapt_event_stream_to_json(response, request_id)
 
                         # Parse response for CCR handling
                         resp_json = None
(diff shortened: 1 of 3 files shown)


chopratejas:
Superseded by #3142, which is merged — and it's substantially this PR. Thanks for it.

**What came from here:** the whole engine. `require_complete` (demanding `message_start`, a terminal `message_stop`, every opened block closed, no in-band `error` event, and no unreplayable delta), the `index` strip, the CRLF normalization, the `data:` slice fix, the body-sniffing detection, and moving the `Accept: application/json` rewrite from the CCR-flip-only site to the buffered boundary every non-streaming request reaches. Your test suite ships unmodified and passes. You're credited via `Co-authored-by:`.

Three of your calls turned out to be the deciding ones when I compared the two PRs:

- **`index`.** You spotted that the reconstruction would poison the *next* request. `anthropic.py:185` documents exactly that 400, and #3131 — which returned the parser output directly — would have reintroduced it.
- **`require_complete`.** The alternative guard only proved `message_start` had been seen, which would have shipped truncated turns as successful 200s. Refusing with a 502 is right: for a reconstruction handed to the client *as* the turn, a visible failure beats invisible corruption.
- **The `data:` slice.** Main required `startswith("data: ")` with the space the SSE spec makes optional. A stream framed without it parsed as *zero events* — which is literally the "0 stream events received" in the report. That may well be a second contributing cause, and neither issue had spotted it.

**What changed on the way in**, both small:

- The decision site now calls `should_recover_sse_reply` from #3131's `nonstream_sse_policy.py` rather than the inline `status_code == 200 and not stream and _looks_like_sse_response(...)`. Same logic, but it puts the contract and its behaviour matrix in one pure, documented place. Your body sniff is preserved — it arrives as `body_is_event_stream`, keeping the policy module free of the response object.
- Everything else, including the `cf-*` filter on top of `sanitize_forwarded_response_headers`, is unchanged.

Closing as superseded, not as rejected — this was the stronger of the two PRs and most of it is now on `main`.

Worth flagging since it affects whether the reporters actually see relief: #3130 and #3140 are both on **0.35.0**, and `main` already carries #3102, #3092, #3091, #3094, #3101, #3069, #3084, #3124 and #3134 for this same symptom class — none of it shipped. It's all gated behind #3067 `chore: release 0.36.0`.

PR #3133, opened 2026-08-19
fix(kompress): degrade when a native dep is installed but unloadable

gglucass:
## Description

`_is_pytorch_available()` and `_is_onnx_available()` catch only `ImportError`. An
installed-but-unloadable native dependency raises `OSError` instead: on Windows
without the MSVC redistributable, `import torch` fails with `WinError 126` loading
`c10.dll`.

So a probe whose contract is to return a bool raises instead, and
`is_kompress_available()` raises with it. `content_router._get_kompress()` guards its
call with `except ImportError` too, so the `OSError` propagates past both guards into
the request path rather than degrading to "kompress unavailable".

Because the compressor is only cached on success, this repeats on every request, and
Python drops failed modules from `sys.modules`, so each one pays a full `torch` import
that ends in the same error.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `_is_pytorch_available()` / `_is_onnx_available()` now catch `(ImportError, OSError)`, so an installed-but-unloadable native dependency degrades to "unavailable" instead of raising out of a bool-returning probe.
- Both probes are `lru_cache`d. The answer cannot change within a process, and nothing else memoises a *failing* import, so without this the full failing `torch` import is retried on every request.
- Added `test_unloadable_native_dep_degrades_instead_of_raising`, which simulates the Windows condition (onnxruntime absent so the `or` cannot short-circuit, torch raising `OSError`) and asserts both the graceful `False` and the single import attempt.
- `test_module_import_without_torch` now calls `cache_clear()` around its assertion, which memoising makes necessary.

## Testing

### Test Output

```text
$ uv run --extra dev pytest tests/test_transforms/ -q
======================= 384 passed, 67 skipped in 18.13s =======================

$ uvx ruff check .
All checks passed!

$ uv run --extra dev mypy headroom/transforms/kompress_compressor.py
```

## Real Behavior Proof

- Environment: Reproduced from a Windows Server 2022 (10.0.20348) box where all three of vcruntime140.dll, vcruntime140_1.dll and msvcp140.dll are absent from System32, so the installed torch cannot load its DLLs. The behaviour was then isolated on macOS 26.5 / CPython 3.10.18 by simulating that condition, since the failure is an exception type and not platform-specific logic.
- Exact command / steps: On the Windows box, the packaged proxy logged `OSError: [WinError 126] ... Error loading "...\torch\lib\c10.dll"` escaping `_is_pytorch_available` at kompress_compressor.py:489 via `_load_kompress` at :1002. To isolate, patch `builtins.__import__` so `onnxruntime`/`optimum` raise `ImportError` (absent) and `torch` raises `OSError` (installed, unloadable), then call `is_kompress_available()` directly and through the same `try/except ImportError` guard `content_router._get_kompress()` uses, counting torch import attempts across 3 requests.
- Observed result: Before the change, `is_kompress_available()` raised `OSError` rather than returning a bool, the error escaped the caller's `except ImportError` on 3/3 requests, and torch was imported 4 times. After the change, it returns `False`, 0/3 requests raise, and torch is imported once. This is covered by the new test.
- Not tested: The fixed build has not been run end-to-end on the Windows box, because installing the MSVC redistributable there would remove the very condition under test. Whether kompress then loads successfully on Windows once the redistributable is present is out of scope for this PR, which only makes its absence degrade correctly.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. This is an availability probe on an existing optional dependency, not a gated feature.
- Stable/default behavior changed: No behaviour change where the dependencies import cleanly. The probe returns the same value; only the installed-but-unloadable case changes, from raising to returning False.
- Kill switch / disable path: N/A. Existing kompress controls (HEADROOM_KOMPRESS_ENDPOINT, the router's own gates) are untouched.
- Qualification impact: None expected. The one existing test that changed did so because memoising the probe makes an explicit cache_clear() necessary; its assertion is unchanged.
- Rollback path: Revert the commit. There is no persisted state or migration.

## Review Readiness

## Checklist


JerrettDavis (approved):
Reviewed the native dependency probes and tests at the updated head. Catching OSError matches installed-but-unloadable extension behavior while retaining successful import semantics, and memoizing the process-stable capability prevents repeated expensive failed imports. Tests explicitly clear both caches around patched import state. I updated the branch from main to pick up the dependency-security remediation that caused its prior pip-audit failure; refreshed checks are running. No blocking code findings.

PR #3137, opened 2026-08-19
fix(ccr): log the exception type when a CCR continuation call fails

### LINKED ISSUE #3129: CCR continuation failure log loses the cause when str(e) is empty
## Summary

`response_handler.py:512` logs a CCR continuation failure using only `str(e)`:

```python
logger.error(f"CCR: Continuation API call failed: {e}")
```

For exception types whose `str()` is empty — e.g. `httpx.TimeoutException('')`, a bare `Exception()`, `httpx.ConnectError('')` — this produces a log line with nothing after the colon:

```
CCR: Continuation API call failed:
```

The real cause (exception type, message) is lost entirely. We hit this in production: a CCR continuation call failed, and the only trace of it was this truncated line — no way to tell whether it was a timeout, a transport error, or something else.

Additionally, the exception appears to be swallowed after logging (the loop `break`s rather than re-raising), so calling code has no way to observe or react to the underlying error either — it only sees a degraded/incomplete response.

## Suggested fix

```python
logger.error("CCR: Continuation API call failed: %s: %r", type(e).__name__, e)
```

Using `%r` instead of implicit `str()` interpolation gives a non-empty representation for every exception type (falls back to `ClassName()` even when the message is empty).

Also worth considering: re-raising after logging (or invoking a callback/hook) so the caller can decide how to handle a failed continuation, rather than the caller only seeing a degraded response with no visibility into why.

## Environment

- headroom-ai 0.27.0
- Installed via `pip install headroom-ai`

## Repro

```python
import httpx
e = httpx.TimeoutException("")
print(str(e))   # ''
print(repr(e))  # "TimeoutException('')"
```

`f"...{e}"` on the first produces an empty suffix; `f"...{e!r}"` on the second does not.

abhay-codes07:
## Description

`response_handler.py` logs a failed CCR continuation call with only `str(e)`:

```python
logger.error(f"CCR: Continuation API call failed: {e}")
```

For exception types whose `str()` is empty — `httpx.TimeoutException('')`, a bare `Exception()`, `httpx.ConnectError('')` — this produces a log line with nothing after the colon:

```
CCR: Continuation API call failed:
```

The cause (type, message) is lost entirely. This was hit in production: a continuation call failed and the only trace was this truncated line, with no way to tell whether it was a timeout, a transport error, or something else.

The fix logs `type(e).__name__` plus `repr(e)`, both of which are non-empty for every exception type (a bare `Exception()` still reprs as `Exception()`), so the cause always survives.

Closes #3129.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/ccr/response_handler.py`: the continuation-failure log now uses `logger.error("CCR: Continuation API call failed: %s: %r", type(e).__name__, e)`.
- `tests/test_ccr_response_handler.py`: added `test_continuation_failure_logs_cause_for_empty_str_exception`, which drives the continuation path with an `api_call_fn` that raises `Exception("")` (empty `str`, non-empty `repr`) and asserts the logged line names the type and is never a bare `...failed:`.

## Testing

### Test Output

```text
tests/test_ccr_response_handler.py  ->  36 passed
uvx ruff@0.16.2 check headroom/ccr/response_handler.py tests/test_ccr_response_handler.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/ccr/response_handler.py  ->  Success: no issues found
```

## Real Behavior Proof

- Exact command / steps: reverted the log line to `f"...: {e}"` and ran the new test (`python -m pytest tests/test_ccr_response_handler.py -k continuation_failure`) -> failed with the captured log `ERROR ... CCR: Continuation API call failed:` (empty after the colon) and `assert "Exception" in ...` failing; restored the fix and re-ran -> 36 passed.
- Observed result: before the fix an empty-`str` exception logged `CCR: Continuation API call failed:` with no cause; after the fix it logs `CCR: Continuation API call failed: Exception: Exception('')`.
- Not tested: a live upstream timeout in a running proxy; the continuation loop is exercised directly with a raising `api_call_fn`, which is the exact except branch that logs.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This is a diagnostic log line, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: only the text of one error log line. Control flow is unchanged (the handler still degrades to the current response).
- Qualification impact: operators can now see the cause of a failed CCR continuation instead of a blank line.
- Rollback path: revert this PR; the log returns to interpolating `str(e)`.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A: internal log line)

## Additional Notes

The issue also raised, as a secondary "worth considering" point, re-raising the exception so the caller can react rather than only seeing a degraded response. I left the control flow as-is (the `break`/degrade path is a deliberate "degradation, not outage" design) and scoped this PR to the logging defect, which is the concrete data-loss the issue reports. Happy to follow up on the re-raise separately if maintainers want it.


JerrettDavis (approved):
Reviewed the latest diagnostic change and continuation control flow. Structured logger arguments preserve lazy formatting, and including both the exception type and repr guarantees a useful cause for empty-string exceptions without changing the deliberate degradation behavior. The regression test exercises the exact continuation failure branch; the branch is mergeable and checks are green. No blocking findings.

PR #3154, opened 2026-08-20
fix(docker): give :latest exactly one writer

### LINKED ISSUE #3150: [BUG] `:latest` points at the distroless `code-slim` variant, which segfaults on arm64 and blocks `headroom deploy`
## Description

`headroom deploy` fails on Apple Silicon because `ghcr.io/headroomlabs-ai/headroom:latest` is not the plain build. It resolves byte-for-byte to `:0.36.0-code-slim`, a distroless image in which `import onnxruntime` segfaults (SIGSEGV / exit 139). The container never binds its port, so the deploy's readiness gate times out and rolls back.

The plain versioned tag `:0.36.0` ships the same headroom release on a Debian base and starts fine. This is a release-tagging bug. The code did not regress.

> Please ignore the `onnxruntime cpuid_info warning: Unknown CPU vendor. cpuinfo_vendor value: 0` line. It is a red herring. Working containers print it too, including `:0.36.0` on this same host. I first filed this as a cpuinfo bug and got it wrong.

## To Reproduce

```bash
docker pull ghcr.io/headroomlabs-ai/headroom:latest
headroom deploy
```

## Actual Behavior

```
Selected persistent-docker (docker): Docker is available, so Headroom can run in a restartable container.
Updating existing deployment profile 'default'...
Restoring previous deployment 'default'...
Error: Deployment 'default' did not become ready after start.
```

The container crash-loops instead of starting:

```
$ docker ps -a --filter name=headroom
headroom-default   Restarting (139)     # 139 = SIGSEGV, 9 restarts
```

Running the image's own entrypoint under `faulthandler` puts the crash at the native extension import:

```
$ docker run --rm --entrypoint python3 ghcr.io/headroomlabs-ai/headroom:latest \
    -c "import faulthandler; faulthandler.enable(); import onnxruntime"

onnxruntime cpuid_info warning: Unknown CPU vendor. cpuinfo_vendor value: 0
Fatal Python error: Segmentation fault

Current thread 0x0000ffffbb6b9ba0 (most recent call first):
  File "/usr/local/lib/python3.13/site-packages/onnxruntime/capi/_pybind_state.py", line 32 in <module>
  File "/usr/local/lib/python3.13/site-packages/onnxruntime/__init__.py", line 26 in <module>

Extension modules: numpy._core._multiarray_umath, numpy.linalg._umath_linalg (total: 2)
```

Line 32 is `from .onnxruntime_pybind11_state import *`. The proxy imports onnxruntime at startup in `cache` mode, so it dies before it can listen.

## Expected Behavior

`headroom deploy` with no `--image` flag should land on an image that starts.

## Root cause

`:latest` and `:0.36.0` are different images of the same release:

| Tag | Manifest digest | Base | `import onnxruntime` |
|-----|-----------------|------|----------------------|
| `latest` | `sha256:6b34905489e3a5e4458cd95d2db2b1ac84fba64429c06fc653eb31e59376b079` | Distroless | **segfault** |
| `0.36.0` | `sha256:bb8e77d01b546a240765b059b145e8430b03d3cab52def44fdbf4a1614bf4568` | Debian 13 (trixie) | works |

`:latest` carries the same digest as `:0.36.0-code-slim`. Its arm64 sub-manifest is `sha256:b51e7527da7630d01c122bd7f8f0f080ec082d8d823e22ce74c80c40629e3bdd`.

Variant matrix at 0.36.0 on arm64:

| Variant | Base | Result |
|---------|------|--------|
| `0.36.0` | Debian 13 | pass |
| `0.36.0-nonroot` | Debian 13 | pass |
| `0.36.0-slim` | Distroless | **fail** |
| `0.36.0-code-slim` (= `latest`) | Distroless | **fail** |

This hits arm64 alone. The amd64 half of the same `:latest` manifest imports onnxruntime fine, which is why x86 CI stays green.

### Decisive evidence that the base image is at fault

Copy the broken image's own `site-packages` onto `debian:trixie-slim` and the import succeeds:

```dockerfile
FROM debian:trixie-slim
RUN apt-get update -qq && apt-get install -y -qq python3
COPY --from=ghcr.io/headroomlabs-ai/headroom:latest \
     /usr/local/lib/python3.13/site-packages /usr/local/lib/python3.13/site-packages
ENV PYTHONPATH=/usr/local/lib/python3.13/site-packages
# python3 -c "import onnxruntime" succeeds
```

Same wheel, same numpy 2.5.2, same Python 3.13.5 build, same host, same kernel. Only the base image differs.

### Ruled out

| Hypothesis | Test | Result |
|------------|------|--------|
| onnxruntime 1.29 broken on arm64 | ort 1.29.0 in `python:3.13-slim` arm64 | works |
| numpy version conflict | ort 1.29.0 with numpy pinned to 2.5.2 | works |
| Python build mismatch | Debian's system python 3.13.5, the identical build the image ships | works |
| CPU vendor detection | `/proc/cpuinfo` bind-mounted with a valid Apple part ID (`0x023`) | still crashes |
| Broken or missing shared library | `ctypes.CDLL` of `libonnxruntime.so` and of the pybind `.so` | both load fine |

The crash lands in `PyInit_onnxruntime_pybind11_state`, the pybind11 module init, rather than in library loading. All eight `DT_NEEDED` entries resolve. The wheel needs `GLIBCXX` <= 3.4.22 and `CXXABI` <= 1.3.11, and the distroless image provides 3.4.33 and 1.3.15, so no ABI version falls short. Both builds are GIL-enabled with `SOABI cpython-313-aarch64-linux-gnu`.

### Version bisect

Every plain versioned tag starts, including the current release:

| Version | Result | onnxruntime | Base |
|---------|--------|-------------|------|
| 0.28.0 | pass | 1.27.0 | Debian 13 |
| 0.29.0 | pass | 1.27.0 | Debian 13 |
| 0.30.0 | pass | 1.27.0 | Debian 13 |
| 0.31.0 | pass | 1.27.0 | Debian 13 |
| 0.32.0 | pass | 1.27.0 | Debian 13 |
| 0.33.0 | pass | 1.28.0 | Debian 13 |
| 0.34.0 | pass | 1.28.0 | Debian 13 |
| 0.35.0 | pass | 1.28.0 | Debian 13 |
| 0.36.0 | pass | 1.29.0 | Debian 13 |
| `latest` | **fail** | 1.29.0 | **Distroless** |

## Suggested fix

1. Point `:latest` at the plain Debian build instead of `code-slim`.
2. If the distroless variants should support arm64, they need an onnxruntime that initialises on that base. An arm64 smoke test of `python -c "import onnxruntime"` against each published variant would catch this, since amd64 CI passes while arm64 crashes.

## Workaround

```bash
headroom deploy --image ghcr.io/headroomlabs-ai/headroom:0.36.0
```

`headroom deploy --no-docker`, which runs the native Python runtime under launchd, also works.

## Environment

- **Headroom version**: 0.36.0
- **Python**: 3.13.5 in the container, 3.13.15 on the host (uv tool install)
- **OS**: macOS 26.5.2, Apple M3 (arm64)
- **Container runtime**: OrbStack 2.2.3, Docker API 29.4.0, kernel `7.0.14-orbstack`. I do not think the runtime is implicated: the same kernel runs onnxruntime fine in every non-distroless image I tested.
- **LLM Provider**: irrelevant

chopratejas:
## Description

Closes #3150. `ghcr.io/headroomlabs-ai/headroom:latest` resolved to the distroless `code-slim` build, whose `import onnxruntime` segfaults on arm64. The proxy imports onnxruntime at startup in cache mode, so the container never bound its port and `headroom deploy` crash-looped (exit 139) on Apple Silicon.

@ricwo's report is exceptionally good — it isolates the base image with a copy-`site-packages`-onto-`debian:trixie-slim` experiment, and explicitly retracts an earlier wrong theory about the `cpuid_info` line. I verified the tagging half independently against the live registry:

```
latest            sha256:6b34905489e3...   <- identical
0.36.0-code-slim  sha256:6b34905489e3...   <- identical
0.36.0            sha256:bb8e77d01b54...
```

**Root cause, proven from the job log rather than inferred.** `docker/metadata-action` defaults to `latest=auto`, which appends a bare `latest` for any semver release — and its own log line reads `suffixLatest=false`, meaning the per-tag `suffix=` that keeps every other tag variant-scoped never reaches it. All eight variant cells therefore emitted `:latest`, and the last to finish won. From the 0.36.0 `code-slim` cell:

```
latest=auto
suffixLatest=false
tags: [..."ghcr.io/headroomlabs-ai/headroom:code-slim",
           "ghcr.io/headroomlabs-ai/headroom:latest"]
pushing sha256:fbcbb68... to ghcr.io/headroomlabs-ai/headroom:latest
```

It landed on `code-slim` by scheduling luck. Any of the eight could have won on any release.

## Type of Change

## Changes Made

- **`flavor: latest=false`** on the `docker-manifest` metadata-action. Stops the tag being generated at all, leaving the root-cell promotion step as the single writer of `:latest`.
- **A runtime guard** in `Create multi-arch manifest`: if a suffixed variant reaches the push carrying a bare `latest`, the job fails instead of publishing. `VARIANT_NAME` is passed via `env:` rather than spliced inline.
- **A test that encodes the missing half of the contract.** `test_docker_latest_promotion_is_owned_by_root_manifest_cell` already existed and passed throughout — it asserted the *intended* writer was the root cell but never the *absence of unintended ones*. The new test asserts exclusivity: `latest=false` is set, no tag rule reintroduces `value=latest`, and the guard runs before anything is pushed.

## Testing

### Test Output

```text
tests/test_release_workflows.py   48 passed, 1 skipped, 1 failed

The failure is test_no_native_tls_in_wheel_build_tree:
  FileNotFoundError: [Errno 2] No such file or directory: 'cargo'
Pre-existing and environmental — cargo is not installed on this machine;
it fails identically on a clean main checkout.

ruff check: All checks passed
ruff format --check: 1 file already formatted
YAML parses; flavor='latest=false', env keys ['IMAGE','DIGEST_DIR','VARIANT_NAME'].
```

## Real Behavior Proof

- Environment: macOS (darwin 25.4.0), worktree off `main`. Live registry queried anonymously via the GHCR token endpoint.
- Exact command / steps: (1) resolved `latest`, `0.36.0` and all four variant tags to manifest digests directly from `ghcr.io/v2/.../manifests/*` to confirm the aliasing; (2) pulled the `docker-manifest (code-slim)` job log from the 0.36.0 release run to see which tags that cell actually pushed; (3) applied the fix and ran the workflow test suite; (4) **removed `latest=false` again and re-ran the new test** to confirm it reproduces the bug.
- Observed result: `:latest` and `:0.36.0-code-slim` share digest `sha256:6b34905489e3...` while `:0.36.0` is `sha256:bb8e77d01b54...`, exactly as reported. The code-slim job log shows `latest=auto` / `suffixLatest=false` and `pushing ... to ghcr.io/headroomlabs-ai/headroom:latest`. With the fix removed the new test fails on `assert 'latest=false' in ''`; with it restored, it passes.
- Not tested: I could not exercise the arm64 segfault or a real multi-arch push from here — no ghcr write credential and no arm64 runner. The tagging fix is verified at the config layer plus the registry evidence above; the end-to-end proof is the re-run described below.

## Runtime Rollout Safety

- Stable/default behavior changed: Yes, and that is the fix — `:latest` will track the plain Debian-based build instead of whichever variant cell happened to finish last.
- Kill switch / disable path: n/a (CI tagging policy).
- Qualification impact: A variant cell that would publish a bare `latest` now fails the Docker job loudly rather than silently repointing the default tag.

## Review Readiness

## Additional Notes

**The live `:latest` is still wrong until the images are re-tagged.** Merging this fixes future releases but does not touch the registry. Once merged, run `docker.yml` via `workflow_dispatch` with `version=0.36.0` to rebuild and repoint `:latest` at the plain build. I don't hold a `write:packages` credential, so that step needs a maintainer.

**Not fixed here, and it outlives this PR:** the distroless arm64 segfault itself. After this change `:latest` points at the Debian build that works, but `0.36.0-slim` and `0.36.0-code-slim` remain broken on arm64 for anyone selecting them explicitly. @ricwo's evidence points squarely at the distroless base — same wheel, same numpy 2.5.2, same Python 3.13.5, works on `debian:trixie-slim` and segfaults on distroless. That deserves its own issue; the two failures are independent and this one is a release-tagging bug, exactly as the report says.

Related but separate, from an earlier audit of this same file: the four bare variants set `RUNTIME_USER = "root"` in `docker-bake.hcl` while `Dockerfile:162` defaults to `nonroot`, and the `runtime-default` (nonroot) bake target is referenced by the docs but by no workflow. Worth its own change.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -293,6 +293,18 @@ jobs:
         uses: docker/metadata-action@v6
         with:
           images: ${{ env.REGISTRY }}/${{ steps.image-name.outputs.image_name }}
+          # `latest=false` is load-bearing (#3150). The action defaults to
+          # `latest=auto`, which appends a bare `latest` for any semver
+          # release — and it logs `suffixLatest=false`, so the per-tag
+          # `suffix=` below never reaches it. Every one of the 8 variant
+          # cells therefore pushed `ghcr.io/.../headroom:latest`, and the
+          # last cell to finish won. At 0.36.0 that was `code-slim`, so
+          # `:latest` resolved to the distroless build, whose
+          # `import onnxruntime` segfaults on arm64 — `headroom deploy`
+          # crash-looped on Apple Silicon. `:latest` has exactly one
+          # writer: the root-cell promotion step at the end of this job.
+          flavor: |
+            latest=false
           tags: |
             type=ref,event=branch,enable=${{ inputs.enable_ref_tags != 'false' && github.event_name != 'release' }},suffix=${{ matrix.variant.name != '' && format('-{0}', matrix.variant.name) || '' }}
             type=ref,event=pr,enable=${{ inputs.enable_ref_tags != 'false' && github.event_name != 'release' }},suffix=${{ matrix.variant.name != '' && format('-{0}', matrix.variant.name) || '' }}
@@ -311,6 +323,9 @@ jobs:
         env:
           IMAGE: ${{ env.REGISTRY }}/${{ steps.image-name.outputs.image_name }}
           DIGEST_DIR: ${{ runner.temp }}/digests
+          # Read by the bare-`latest` guard below. Via `env:` rather than
+          # inline `${{ }}` so the value is never spliced into the script.
+          VARIANT_NAME: ${{ matrix.variant.name }}
         run: |
           # Reconstruct full image references from the digest marker
           # filenames (each file is named after the bare hex digest
@@ -329,6 +344,16 @@ jobs:
             digest="$(basename "$f")"
             digest_refs+=("${IMAGE}@sha256:${digest}")
           done
+
+          # Belt-and-braces for #3150: only the root cell may ever carry a
+          # bare `latest`. A suffixed variant reaching this point with one
+          # means the tag rules regressed, and shipping it would repoint
+          # `:latest` at a non-default image. Fail instead of publishing.
+          if [ -n "${VARIANT_NAME}" ] && jq -e '.tags[]? | select(endswith(":latest"))' \
+               <<< "${DOCKER_METADATA_OUTPUT_JSON}" >/dev/null 2>&1; then
+            echo "::error::variant '${VARIANT_NAME}' would publish a bare :latest tag" >&2
+            exit 1
+          fi
 
           # Build `--tag` args from the metadata-action JSON output.
           # Empty tags array is valid (PR builds without ref-tags
--- a/tests/test_release_workflows.py
+++ b/tests/test_release_workflows.py
@@ -120,6 +120,51 @@ def test_docker_latest_promotion_is_owned_by_root_manifest_cell() -> None:
     assert guard_start < manifest_script.index("exit 1", guard_start) < create_start
 
 
+def test_only_the_root_cell_can_ever_publish_a_bare_latest_tag() -> None:
+    """`:latest` must have exactly one writer (#3150).
+
+    The sibling test above proves the *promotion step* is owned by the root
+    cell. That was never the whole contract: it checked the intended writer
+    and not the absence of unintended ones.
+
+    ``docker/metadata-action`` defaults to ``latest=auto``, which appends a
+    bare ``latest`` for any semver release. Its own log line reads
+    ``suffixLatest=false`` — the per-tag ``suffix=`` that keeps every other
+    tag variant-scoped does not reach it. So all eight variant cells emitted
+    ``ghcr.io/.../headroom:latest`` and the last to finish won the tag. At
+    0.36.0 that was ``code-slim``: ``:latest`` resolved to the distroless
+    build, whose ``import onnxruntime`` segfaults on arm64, and
+    ``headroom deploy`` crash-looped on Apple Silicon with SIGSEGV.
+
+    Two things hold the line, and both are asserted here: ``latest=false``
+    stops the tag being generated at all, and a runtime guard refuses to
+    publish if a suffixed variant ever carries one anyway.
+    """
+    workflow = yaml.safe_load((ROOT / ".github" / "workflows" / "docker.yml").read_text())
+    manifest = workflow["jobs"]["docker-manifest"]
+
+    meta = next(step for step in manifest["steps"] if step.get("id") == "meta")
+    flavor = meta["with"].get("flavor", "")
+    assert "latest=false" in flavor, (
+        "docker/metadata-action defaults to latest=auto; without latest=false "
+        "every variant cell publishes a bare :latest and the last one wins"
+    )
+
+    # No tag rule may reintroduce it explicitly either.
+    assert "value=latest" not in meta["with"]["tags"]
+
+    create = next(
+        step for step in manifest["steps"] if step["name"] == "Create multi-arch manifest"
+    )
+    script = create["run"]
+    # The guard reads the variant name from env, never spliced inline.
+    assert create["env"].get("VARIANT_NAME") == "${{ matrix.variant.name }}"
+    assert 'endswith(":latest")' in script
+    assert script.index('endswith(":latest")') < script.index("docker buildx imagetools create"), (
+        "the bare-latest guard must run before anything is pushed"
+    )
+
+
 def test_docker_manifest_downloads_exactly_one_artifact_per_architecture() -> None:
     """Each manifest cell must download exactly its two architecture digests.
 


JerrettDavis (approved):
Reviewed the latest workflow and regression test. Disabling metadata-action's implicit latest tag removes the competing writers, the pre-push guard is correctly scoped to non-root variants, and the existing explicit root promotion remains the sole writer. The PR is mergeable; the remaining docker-native-e2e check is still running, but I found no code-level blocker.

PR #3169, opened 2026-08-21
perf(compression/code): avoid a UTF-8 copy per code block in byte-to-char mapping

abhay-codes07:
## Description

`CodeStructureHandler._byte_spans_to_char_spans` converts tree-sitter's byte offsets to character offsets for each compressed code block. Its ASCII fast path was:

```python
n_bytes = len(content.encode("utf-8"))
if n_bytes == len(content):
    return spans
```

`content.encode("utf-8")` allocates a full UTF-8 byte copy of the whole code block, which is then discarded — just to compare its length against `len(content)`. For the common case (ASCII source, byte == char), `content.isascii()` answers the same question in C with no allocation.

The non-ASCII branch additionally re-encoded **every character** in the loop (`len(ch.encode("utf-8"))`) to get its UTF-8 width — one small allocation per character. A character's width (1-4 bytes) is fully determined by its code point, so it can be derived with `ord(ch)` and no encoding.

This runs per code block on the compression path, so for code-heavy traffic the discarded per-block allocation adds up. Output is byte-for-byte identical.

Benchmark (`_byte_spans_to_char_spans` on a ~20 KB ASCII block, mean of 20000 calls):

```
old (len(encode) == len) : 0.45 us/call  (+ a 20 KB bytes allocation)
new (isascii())          : 0.06 us/call  (~7x faster, no allocation)
```

Correctness was verified by comparing the old and new implementations on mixed content (`café 🚀 语言` — 2/4/3-byte characters): the remapped span offsets are identical.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `headroom/compression/handlers/code_handler.py` (`_byte_spans_to_char_spans`): use `content.isascii()` for the fast-path check (no UTF-8 allocation), and derive each character's UTF-8 width from its code point instead of re-encoding it per character.
- `tests/test_compression/test_code_handler.py`: added `TestByteSpanToCharSpan` — ASCII spans returned unchanged, and non-ASCII byte offsets (café / rocket / CJK) remapped to the correct character offsets including the exclusive-end boundary.

## Testing

### Test Output

```text
tests/test_compression/test_code_handler.py  ->  24 passed, 8 skipped (tree-sitter not installed)
uvx ruff@0.16.2 check headroom/compression/handlers/code_handler.py tests/test_compression/test_code_handler.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/compression/handlers/code_handler.py  ->  Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: ran the old and new implementations side by side on ASCII and on mixed non-ASCII content (`café 🚀 语言`) and confirmed identical remapped span offsets; benchmarked the ASCII path (0.45us -> 0.06us, ~7x, and the new path allocates nothing). Added the regression test and ran the file.
- Observed result: identical span mapping; the per-block UTF-8 allocation on the ASCII fast path is gone, and the non-ASCII loop no longer allocates per character.
- Not tested: an end-to-end code-compression request through `UniversalCompressor`; the static helper is exercised directly with byte-offset spans, which is exactly what tree-sitter feeds it.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This is an internal offset-mapping helper in the code structure handler, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: no. The returned spans are identical for ASCII and non-ASCII content; only the work done to compute them is cheaper.
- Qualification impact: less per-block allocation/CPU when mapping code spans; identical output.
- Rollback path: revert this PR; the helper returns to encoding the content for the length check and per-character.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A: internal helper, output unchanged)

## Additional Notes

Scope is deliberately small and behavior-preserving: `content.isascii()` is exactly equivalent to "every UTF-8 byte is one character," and the code-point width table (`<0x80 -> 1`, `<0x800 -> 2`, `<0x10000 -> 3`, else `4`) matches Python's UTF-8 encoding for every `str` character.

### DIFF
--- a/headroom/compression/handlers/code_handler.py
+++ b/headroom/compression/handlers/code_handler.py
@@ -605,16 +605,24 @@ class CodeStructureHandler(BaseStructureHandler):
         returned unchanged. Otherwise a byte->char table is built once
         and every span endpoint is remapped.
         """
-        n_bytes = len(content.encode("utf-8"))
-        if n_bytes == len(content):
+        # ASCII fast path: every byte is one character, so tree-sitter's byte
+        # offsets are already character offsets. ``content.isascii()`` checks
+        # this in C without allocating, where ``len(content.encode("utf-8"))``
+        # built (and discarded) a full UTF-8 copy of the content on every code
+        # block just to compare lengths.
+        if content.isascii():
             return spans
 
         # byte_to_char[b] = index of the character containing byte b;
-        # byte_to_char[n_bytes] = len(content) so exclusive ends map.
+        # byte_to_char[n_bytes] = len(content) so exclusive ends map. A
+        # character's UTF-8 width is derived from its code point (1-4 bytes)
+        # rather than re-encoding every character individually in the loop.
+        n_bytes = len(content.encode("utf-8"))
         byte_to_char = [0] * (n_bytes + 1)
         byte_pos = 0
         for char_idx, ch in enumerate(content):
-            ch_width = len(ch.encode("utf-8"))
+            cp = ord(ch)
+            ch_width = 1 if cp < 0x80 else 2 if cp < 0x800 else 3 if cp < 0x10000 else 4
             for b in range(byte_pos, byte_pos + ch_width):
                 byte_to_char[b] = char_idx
             byte_pos += ch_width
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed at eca1506235e9912479659724ada8f1e93f03db5e. isascii() is the right allocation-free fast path, and the code-point width mapping preserves UTF-8 byte-to-character semantics for BMP and supplementary characters. The mixed-width/exclusive-end regression covers the meaningful boundary behavior. No blocker found.

PR #3155, opened 2026-08-20
fix(metrics): attribute tool-schema savings per model, not just compression

chopratejas:
## Description

Reported against 0.36.0 (VS Code + Copilot + Claude Code): the per-model breakdown disagreed with the headline printed four lines above it.

```
Tokens saved: 625,277
  · messages       36,071
  · tool schemas  589,206
Per-Model Breakdown
  <a>: 35,907 tokens saved
  <b>:      0 tokens saved
  <c>:    164 tokens saved
  <d>:      0 tokens saved
```

The rows sum to **36,071** — the *messages* line exactly. All 589,206 tokens of tool-schema deferral, 94% of the headline, had no row to land in, so every tool-heavy model reported "0 tokens saved" while real dollars were credited to it.

Deferral is disjoint from message compression by construction: deferred schemas never enter the message token counts, so they move neither `tokens_saved` nor `tokens_sent`. The headline, the PERF line, and the savings ledger (#2795) all already fold the two together. Three per-model surfaces did not.

Closes #

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- **`perf/analyzer.py`** — the per-model loop summed `tokens_saved` while its own headline summed `tokens_saved + tool_saved`. Now uses the same all-layers construction (`headline_before = before + tool_saved`), and prints a `· messages / · tool schemas` split line only when there is a split to show.
- **`proxy/savings_tracker.py`** — added a `tool_tokens_saved` bucket to `_empty_by_model_entry()`, normalization, and `_record_by_model_locked()`; `record_request()` gained a `tool_search_saved` parameter. `_by_model_snapshot_locked()` ranks and computes `savings_percent` off the combined figure and exposes `headline_tokens_saved`.
- **`proxy/prometheus_metrics.py`** — **the seam.** `record_request` already accepted `tool_search_saved` and already folded it into the per-model *dollars*, but never passed it to `savings_tracker.record_request`. Tokens and money therefore disagreed on the same row.
[...]

PR #3170, opened 2026-08-21
fix(savings): record tool-schema dollars disjointly beside the folded headline

gglucass:
## Description

Since the attribution unification (#2976, refined by #3123), `record_request` folds the tool-schema layer's dollars into `compression_savings_usd` — on the lifetime block, the display session, and every history checkpoint — while `total_tokens_saved` on those same checkpoints stays message-only. The two halves of one checkpoint now disagree about what "saved" means, and any consumer that derives a $/token rate from a checkpoint reads a figure inflated by `1 + tool/message`.

That inflation is not hypothetical. On one real install, message compression was 3,299,618 tokens next to 18,435,490 tokens of tool-schema deferral (a 5.59x ratio — far from the 0.24x in the sample #2976 cites, this ratio varies wildly by workload). The blended savings rate a downstream consumer derived went from $4.99/M to $32.88/M — above the input list price of every model in the mix, a rate no genuine input-token saving can reach.

There is also a smaller correctness gap: a request whose only saving is deferral (`tokens_saved == 0`, common on tool-heavy turns) appended no checkpoint at all, so its folded dollars landed in whichever later bucket happened to checkpoint next.

#3155 fixed the token half of this per model, and stated the principle this PR extends to dollars: kept as its own bucket rather than widened in place, so the persisted meaning of an existing field never changes under a reader that predates it. This PR is deliberately **additive** — `compression_savings_usd` keeps its current folded semantics everywhere, so the dashboard, PERF headline, and every existing reader are unchanged. The layer is simply also recorded disjointly beside it, so message-only dollars are recoverable by subtraction and the checkpoint's token/dollar pair can be read consistently.

## Type of Change

## Changes Made

- Record `tool_tokens_saved` / `tool_schema_savings_usd` on the lifetime block and display session, wired through `_default_state`, `_sanitize_state` (including the history-recovery max), and both normalizers — the fields survive restarts rather than resetting the way the ad-hoc output fields do.
- Carry both fields on every history checkpoint, and extend the checkpoint append gate so a deferral-only request checkpoints too.
- Expose `tool_tokens_saved_delta` / `tool_schema_savings_usd_delta` (plus cumulatives) on rollup buckets, following the exact delta convention of the existing series, and add the delta columns to the rollup CSV export.
- State files written before these fields existed load with both at 0; legacy tuple-shaped history entries normalize the same way.
- New test module `tests/test_tool_schema_savings_split.py` (6 tests) pinning the disjoint recording, reload durability, legacy-state defaults, the deferral-only checkpoint, the rollup deltas, and the subtraction identity; existing shape assertions in `tests/test_proxy_savings_history.py` updated for the two new fields.

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_proxy_savings_history.py \
    tests/test_tool_schema_savings_split.py tests/test_per_model_tool_savings.py \
    tests/test_savings_tool_search_aggregation.py tests/test_savings_tracker_zero_price.py \
    tests/test_agent_savings.py
======================== 103 passed, 1 warning in 7.36s ========================

$ uv run --frozen --extra dev pytest tests/test_stats_new_input_savings_rate.py \
    tests/test_dashboard_token_savings.py tests/test_proxy_project_savings.py \
    tests/test_output_savings.py tests/test_savings_ledger.py
=================== 76 passed, 2 skipped, 1 warning in 8.15s ===================

$ uv run --frozen ruff check headroom/proxy/savings_tracker.py tests/test_tool_schema_savings_split.py tests/test_proxy_savings_history.py
All checks passed!
$ uv run --frozen ruff format --check headroom/proxy/savings_tracker.py tests/test_tool_schema_savings_split.py tests/test_proxy_savings_history.py
3 files already formatted
$ uv run --frozen mypy headroom/proxy/savings_tracker.py
```

## Real Behavior Proof

- Environment: macOS, Python 3.10 venv via `uv run --frozen`, this branch.
- Exact command / steps: one `record_request` shaped like the reporting install's traffic (100 message tokens saved, 559 deferred schema tokens, claude-opus-5 pricing at $5/M input), then derive $/token from the appended checkpoint both ways.
- Observed result: the naive checkpoint-derived rate reproduces the impossible $32.88/M seen in the field almost exactly, and subtracting the new disjoint field recovers the model's actual input rate:

```text
checkpoint: total_tokens_saved=100 compression_savings_usd=0.003295 tool_tokens_saved=559 tool_schema_savings_usd=0.002795
folded-dollars / message-tokens rate: $32.95/M  (input list price is $5/M)
after subtracting the disjoint field:  $5.00/M
```

- Not tested: a long-lived multi-process proxy soak, and the downstream consumer's subtraction path itself (that lands with the consumer release that reads the new fields).

## Runtime Rollout Safety

- Rollout-managed feature(s): None -- persisted savings accounting only; no rollout-gated runtime behavior is touched.
- Stable/default behavior changed: No. Additive fields beside existing ones; `compression_savings_usd` keeps its current folded semantics, so every existing reader sees identical values.
- Kill switch / disable path: None needed -- old readers ignore the new fields, and state files written before them load with both at 0 (covered by tests).
- Qualification impact: None beyond the unit suites run above; displayed figures are unchanged.
- Rollback path: `git revert` -- the fields stop being written and remain optional on read, so state files produced while this was live stay loadable.

## Review Readiness

Related: #3155 (token half of the same split), #3171 (companion PR on what the tool-schema bucket should be priced at — independent; either lands without the other).

### DIFF
--- a/headroom/proxy/savings_tracker.py
+++ b/headroom/proxy/savings_tracker.py
@@ -434,6 +434,8 @@ def _normalize_history_entry(entry: Any) -> dict[str, Any] | None:
     total_input_cost_usd = 0.0
     output_tokens_saved = 0
     output_savings_usd = 0.0
+    tool_tokens_saved = 0
+    tool_schema_savings_usd = 0.0
     provider = PROVIDER_UNKNOWN
     model = MODEL_UNKNOWN
 
@@ -450,6 +452,8 @@ def _normalize_history_entry(entry: Any) -> dict[str, Any] | None:
         total_input_cost_usd = _coerce_float(entry.get("total_input_cost_usd"))
         output_tokens_saved = _coerce_int(entry.get("output_tokens_saved"))
         output_savings_usd = _coerce_float(entry.get("output_savings_usd"))
+        tool_tokens_saved = _coerce_int(entry.get("tool_tokens_saved"))
+        tool_schema_savings_usd = _coerce_float(entry.get("tool_schema_savings_usd"))
         provider = _normalize_provider(entry.get("provider"))
         model = _normalize_model(entry.get("model"))
     elif isinstance(entry, list | tuple) and len(entry) >= 2:
@@ -479,6 +483,8 @@ def _normalize_history_entry(entry: Any) -> dict[str, Any] | None:
         "total_input_cost_usd": round(total_input_cost_usd, 6),
         "output_tokens_saved": output_tokens_saved,
         "output_savings_usd": round(output_savings_usd, 6),
+        "tool_tokens_saved": tool_tokens_saved,
+        "tool_schema_savings_usd": round(tool_schema_savings_usd, 6),
     }
 
 
@@ -487,6 +493,8 @@ def _empty_display_session() -> dict[str, Any]:
         "requests": 0,
         "tokens_saved": 0,
         "compression_savings_usd": 0.0,
+        "tool_tokens_saved": 0,
+        "tool_schema_savings_usd": 0.0,
         "cache_read_tokens": 0,
         "cache_savings_usd": 0.0,
         "total_input_tokens": 0,
@@ -607,6 +615,11 @@ def _normalize_display_session(entry: Any) -> dict[str, Any]:
         "tokens_saved": tokens_saved,
         "compression_savings_usd": round(
             _coerce_float(entry.get("compression_savings_usd")),
+            6,
+        ),
+        "tool_tokens_saved": _coerce_int(entry.get("tool_tokens_saved")),
+        "tool_schema_savings_usd": round(
+            _coerce_float(entry.get("tool_schema_savings_usd")),
             6,
         ),
         "cache_read_tokens": _coerce_int(entry.get("cache_read_tokens")),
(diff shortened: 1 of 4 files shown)


JerrettDavis (approved):
The additive split preserves every existing folded-dollar consumer while making message-only dollars recoverable alongside the already message-only token count. New cumulative fields are normalized across legacy state/history, recovered monotonically on restart, included in deferral-only checkpoints, and converted to rollup/CSV deltas using the tracker’s existing convention. The priced/unpriced paths remain internally consistent, and full current-head CI is green. I found no blocking issue.

gglucass:
Thanks @JerrettDavis - highly recommend merging this one for accurate $ savings metrics on the next release :-)

gglucass:
Bumping this one @chopratejas @JerrettDavis @DevanshiVyas - headroom is currently overreporting savings substantially.

gglucass:
Pushed a test fix; this is green again and ready for review.

`test_lifetime_and_session_record_tool_schema_disjointly` had started failing, but not because of anything in the diff. Every record carried the literal `2026-08-21T09:10:00Z`, and `_display_session_snapshot_locked` expires the session against `_utc_now()` rather than against the recorded timestamp. Once that literal aged past `DEFAULT_DISPLAY_SESSION_INACTIVITY_MINUTES` (60), the snapshot returned an empty `display_session` and the assertion read `0.0` instead of `6.0`. The suite was green in August CI and went red about three weeks later without a line changing. `lifetime` was correct throughout, so the tracker was never wrong here; the test was.

Records now derive their timestamps from now, and the rollup tests derive their expected hour-bucket keys from the same anchor instead of hardcoding them. That also removes a second rot that was about eleven days out: `max_history_age_days=30` would have begun pruning the same literal.

Verified on the MERGED result of `main` + this branch, not on the branch head alone:

- Merge into `e67b3c8a`: clean, no conflicts.
- `pytest tests/test_tool_schema_savings_split.py tests/test_proxy_savings_history.py tests/test_persistent_metrics_persistence.py` -> 47 passed.

JerrettDavis (approved):
Re-reviewed the latest head against the previously approved commit. Production code is unchanged; the update replaces expiring literal test timestamps with recent session timestamps and stable past-hour buckets. All 47 savings-split, persistence, and history tests pass locally; pinned Ruff check and format check pass. The time-dependent tests continue to exercise the original display-session and rollup assertions.

JerrettDavis (approved):
Re-reviewed exact head 3d65d97b5e1b064818d0c726ac9771ce9b78b2f6 after its merge from current main. The disjoint tool-schema savings fields and semantic rollup conflict resolution preserve the approved behavior alongside current cache-cost fields. The head is mergeable, contains current main, and all reported checks are terminal green.

PR #3171, opened 2026-08-21
fix(savings): price tool-schema deferral at the cache-read rate

gglucass:
## Description

`estimate_request_savings_usd` prices the tool-schema deferral layer with `_estimate_compression_savings_usd`, i.e. at the full `input_cost_per_token` rate, the same rate message compression uses. That is the wrong counterfactual. Deferred schemas are byte-stable prompt-prefix content, so had they been sent they would have ridden the SAME region of the request as the rest of the prefix: cold-written once, billed at the cache-read discount on every warm turn after, re-written when the TTL expired.

On one real install where deferral ran 5.59x message compression, the blended savings rate read $32.88/M against $4.99/M actual, which is above the input list price of every model in the mix. No genuine input-token saving can reach that: a saved input token cannot be worth more than sending it would have cost.

**This branch was rewritten on 2026-09-09.** It previously proposed a flat 0.1x-of-input fraction. Since then #3351 merged `_bucket_by_cache_mix` into `cost.py` to model exactly this counterfactual for the full-savings cost card. Keeping a separate rule here would leave two surfaces quoting different prices for the same deferred tokens, so this now reuses that helper instead.

## Type of Change

## Changes Made

- `headroom/proxy/savings_tracker.py`: new `_estimate_tool_schema_savings_usd`, which splits the layer with `cost.py`'s `_bucket_by_cache_mix` by the request's own observed read/write/uncached mix and prices each share at its own rate. A request with no billed input breakdown falls back to list price for the whole amount, matching that helper's documented behaviour, so the two surfaces cannot disagree about the same request.
- `headroom/proxy/savings_tracker.py`: new `_model_input_rates`, which resolves `(cache_read, cache_write, list)` per-token rates. Models litellm prices without cache rates derive them from the list price (0.1x read, 1.25x write) rather than silently pricing a cached token at list. Follows the module's existing fallback rules: an unavailable litellm or unpriced model falls back to `DEFAULT_FALLBACK_INPUT_COST_PER_TOKEN`, and a legitimately free model (a real `0.0`) prices as free rather than as "unknown".
- `headroom/proxy/savings_tracker.py`: `estimate_request_savings_usd` gains `cache_write_tokens` and `uncached_tokens`.
- `headroom/proxy/prometheus_metrics.py`: `record_request` already received both and forwarded only `cache_read_tokens`, so the mix could never reach the estimator. Both are now passed.
- `tests/test_tool_schema_pricing.py`: rewritten for the mix-based rule (11 tests) - per-mix pricing, the no-breakdown fallback, derived cache rates, free models, unpriced models, zero/negative counts, bucket separation, and the call-site wiring through the real `PrometheusMetrics.record_request` path.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest $(grep -rln "tool_search_saved\|tool_schema\|estimate_request_savings_usd" tests/) -q
309 passed, 2 warnings in 33.13s

uvx ruff check headroom/proxy/savings_tracker.py headroom/proxy/prometheus_metrics.py tests/test_tool_schema_pricing.py
All checks passed!

uv run --frozen --extra dev mypy headroom/proxy/savings_tracker.py headroom/proxy/prometheus_metrics.py
```

## Real Behavior Proof

- Environment: macOS 15, Python 3.11.13, branch rebased onto `e67b3c8a`.
- Exact command / steps: call `estimate_request_savings_usd(model, compression_tokens_saved=1_000_000, tool_schema_tokens_saved=1_000_000, **mix)` across four cache mixes and three models.
- Observed result: the deferral bucket now tracks the request's mix instead of always reading as the message rate.

```text
                                 message   deferral
claude-opus-5
  fully warm (100% cache read)    $5.00     $0.50
  mixed 50/30/20                  $5.00     $3.12
  cold (100% write)               $5.00     $6.25
  no breakdown                    $5.00     $5.00
gpt-5.2
  fully warm (100% cache read)    $1.75     $0.17
  mixed 50/30/20                  $1.75     $1.09
  cold (100% write)               $1.75     $2.19
  no breakdown                    $1.75     $1.75
gemini-3-pro
  fully warm (100% cache read)    $3.00     $0.30
  mixed 50/30/20                  $3.00     $1.88
  cold (100% write)               $3.00     $3.75
  no breakdown                    $3.00     $3.00
```

- Worth reading carefully: this is NOT a uniform reduction. A cold request prices deferral ABOVE the message rate ($6.25 vs $5.00 on opus), because cache writes bill at a 1.25x premium. The old flat-rate treatment was wrong in both directions; it happened to be wrong in the expensive direction on the cache-heavy traffic that prompted the report.
- Not tested: installs without litellm beyond the unit-tested fallback path; dashboard rendering of the re-priced bucket on a long-lived live proxy.

## Runtime Rollout Safety

- Rollout-managed feature(s): None - pricing estimator only, no rollout-gated runtime behavior touched.
- Stable/default behavior changed: Yes, deliberately. The `tool_schema` bucket now prices by cache mix, so reported tool-schema dollars fall on warm traffic and rise slightly on cold traffic. No control flow, wire format, or persistence shape changes.
- Kill switch / disable path: None - pure function change; reverting the commit restores input-rate pricing exactly.
- Qualification impact: Savings dollars shown on the dashboard and `/stats` drop on tool-heavy cache-warm traffic. A correction, not a regression. Token counts untouched.
- Rollback path: `git revert` - no persisted state depends on the rate that produced a recorded dollar figure.

## Review Readiness

- [x] PR is ready for human review

Related: #3170 records tool-schema dollars disjointly so consumers can tell the layers apart. Independent; either lands without the other. Together they make the headline both legible and honest.

### DIFF
--- a/headroom/proxy/savings_tracker.py
+++ b/headroom/proxy/savings_tracker.py
@@ -24,7 +24,7 @@ from typing import Any
 
 from headroom import paths as _paths
 from headroom.proxy import project_name_policy
-from headroom.proxy.cost import _bucket_by_cache_mix
+from headroom.proxy.cost import _bucket_by_cache_mix, _cache_input_rates
 from headroom.proxy.persistent_metrics import PersistentMetricsState
 
 PROJECT_NAME_MAX_LENGTH = project_name_policy.PROJECT_NAME_MAX_LENGTH
@@ -330,46 +330,31 @@ def _estimate_cache_savings_usd(model: str, cache_read_tokens: int) -> float:
         return 0.0
 
 
-# Multipliers used only when litellm knows a model's list price but not its
-# cache prices. Anthropic bills cache reads at 0.1x input and cold writes at
-# 1.25x; OpenAI and Gemini read between 0.1x and 0.25x. Reads take the
-# conservative (cheapest) end so an unknown cache price cannot inflate savings.
-_CACHE_READ_MULTIPLIER = 0.1
-_CACHE_WRITE_MULTIPLIER = 1.25
-
-
 def _model_input_rates(model: str) -> tuple[float, float, float]:
     """``(cache_read, cache_write, list)`` per-token input rates for ``model``.
 
-    Mirrors ``_estimate_compression_savings_usd``'s fallback rules: an
-    unavailable litellm or an unpriced model falls back to
-    ``DEFAULT_FALLBACK_INPUT_COST_PER_TOKEN``, and a legitimately free model
-    (a real ``0.0``) prices as free rather than as "unknown".
+    Rates and the missing-cache-price fallback come from ``cost.py``'s
+    ``_cache_input_rates``, the same policy the cost card prices real requests
+    with, so the two surfaces cannot disagree about what a deferred token would
+    have cost. Only the catalog lookup is local, because this module resolves
+    model names through its own alias-aware ``_resolve_litellm_model``.
+
+    Mirrors ``_estimate_compression_savings_usd``'s outer rules: an unavailable
+    litellm or an unpriced model falls back to
+    ``DEFAULT_FALLBACK_INPUT_COST_PER_TOKEN`` for all three rates, and a
+    legitimately free model (a real ``0.0``) prices as free rather than as
+    "unknown".
     """
     litellm = _get_litellm_module()
+    if litellm is not None:
+        try:
+            rates = _cache_input_rates(litellm.model_cost.get(_resolve_litellm_model(model), {}))
+        except Exception:
+            rates = None
+        if rates is not None:
+            return rates
     fallback = float(DEFAULT_FALLBACK_INPUT_COST_PER_TOKEN)
-    default = (
-        fallback * _CACHE_READ_MULTIPLIER,
-        fallback * _CACHE_WRITE_MULTIPLIER,
-        fallback,
-    )
-    if litellm is None:
-        return default
-    try:
-        info = litellm.model_cost.get(_resolve_litellm_model(model), {})
-        list_cost = info.get("input_cost_per_token")
-        if list_cost is None:
-            raise RuntimeError("input cost unavailable")
-        list_rate = float(list_cost)
-        read_cost = info.get("cache_read_input_token_cost")
-        write_cost = info.get("cache_creation_input_token_cost")
-        return (
-            float(read_cost) if read_cost is not None else list_rate * _CACHE_READ_MULTIPLIER,
-            float(write_cost) if write_cost is not None else list_rate * _CACHE_WRITE_MULTIPLIER,
-            list_rate,
-        )
-    except Exception:
-        return default
+    return (fallback, fallback, fallback)
 
 
 def _estimate_tool_schema_savings_usd(
--- a/headroom/proxy/savings_tracker.py
+++ b/headroom/proxy/savings_tracker.py
@@ -24,6 +24,7 @@ from typing import Any
 
 from headroom import paths as _paths
 from headroom.proxy import project_name_policy
+from headroom.proxy.cost import _bucket_by_cache_mix
 from headroom.proxy.persistent_metrics import PersistentMetricsState
 
 PROJECT_NAME_MAX_LENGTH = project_name_policy.PROJECT_NAME_MAX_LENGTH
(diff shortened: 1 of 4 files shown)


JerrettDavis (approved):
Reviewed at 46bbd5f9521bf01e9d2445e05397050da86a0376. The new estimator uses the explicit cache-read rate when available, preserves genuinely free pricing, and falls back conservatively and consistently when pricing metadata is absent. The request-level regression confirms that message compression and schema deferral are now priced through distinct buckets. No blocking issues found.

gglucass:
Rewritten on top of current `main` (`e67b3c8a`), because the approach this branch originally proposed was overtaken by a merge.

The original diff priced the deferral layer at a flat 0.1x-of-input fraction. Since then #3351 landed `_bucket_by_cache_mix` in `cost.py`, which models exactly this counterfactual for the full-savings cost card. Merging the flat fraction now would leave `cost.py` and `savings_tracker.py` quoting different prices for the same deferred tokens, so the branch reuses the merged helper instead of keeping a second rule.

The underlying bug is still live: `savings_tracker.py:351` on `main` still prices the layer with `_estimate_compression_savings_usd`, i.e. at the full input rate.

One finding worth flagging, since it contradicts the framing in the original PR body: this is **not** a uniform ~10x reduction. Priced by mix, a cold request values deferral ABOVE the message rate ($6.25 vs $5.00 per M on opus) because cache writes bill at a 1.25x premium. The flat-rate treatment was wrong in both directions; it was only ever wrong in the expensive direction on the cache-warm traffic that prompted the report. Full table is in the updated Real Behavior Proof section.

Also fixed a wiring gap the original missed: `record_request` already received `cache_write_tokens` and `uncached_input_tokens` and forwarded only `cache_read_tokens`, so no mix could have reached the estimator regardless of how it priced. There is now a test through the real `PrometheusMetrics.record_request` path asserting the mix arrives.

Verification: 309 passed across every suite touching `tool_schema` / `tool_search_saved` / `estimate_request_savings_usd`; ruff and mypy clean.

headroom/proxy/savings_tracker.py
```diff
@@ -329,13 +330,95 @@ def _estimate_cache_savings_usd(model: str, cache_read_tokens: int) -> float:
...
+    try:
+        info = litellm.model_cost.get(_resolve_litellm_model(model), {})
+        list_cost = info.get("input_cost_per_token")
+        if list_cost is None:
+            raise RuntimeError("input cost unavailable")
+        list_rate = float(list_cost)
+        read_cost = info.get("cache_read_input_token_cost")
+        write_cost = info.get("cache_creation_input_token_cost")
+        return (
+            float(read_cost) if read_cost is not None else list_rate * _CACHE_READ_MULTIPLIER,
```
JerrettDavis: [P2] Use a shared cache-price fallback policy with the cost card. For a model catalog entry containing only input_cost_per_token=5e-6, CostTracker._get_cache_prices returns (5e-6,5e-6,5e-6), while this helper returns (0.5e-6,6.25e-6,5e-6). Thus 1M deferred tokens on the same fully read-cached request quote $5 in the cost card and $0.50 here; a fully written request quotes $5 versus $6.25. This directly contradicts the stated same-request/same-price goal, and a universal Anthropic write multiplier can overstate another provider's savings. Please consolidate the rate/fallback policy used by both consumers (while preserving genuinely free models), and test both actual paths with missing cache-rate metadata.

JerrettDavis (changes requested):
The request cache-mix forwarding and the new 11-test suite pass locally. One pricing-consistency case remains: sharing the bucket helper does not yet make this estimator agree with the cost card when catalog cache rates are missing. Please add a regression comparing the two actual pricing paths, as detailed below.

gglucass:
Fixed in c1481cd. You were right, and the divergence was worse than a rate mismatch: the 0.1x/1.25x pair is Anthropic's, so it would have quoted a discount for providers that do not bill cache that way at all.

- Extracted the policy into `cost._cache_input_rates(info, long_context=...)`: an absent `cache_read_input_token_cost` / `cache_creation_input_token_cost` reads as `input_cost_per_token`, never as a multiplier. `CostTracker._get_cache_prices` and the deferral estimator both go through it, so they cannot drift again.
- Deleted `_CACHE_READ_MULTIPLIER` / `_CACHE_WRITE_MULTIPLIER`.
- Only the catalog lookup stays per-module: `savings_tracker` resolves model names through its own alias-aware `_resolve_litellm_model` (bare-prefix fall-through that `cost.py` deliberately does not have), so the helper takes the resolved `model_cost` entry rather than the model name.
- Free models preserved: `_cache_input_rates` returns `None` only when `input_cost_per_token` is missing; a real `0.0` prices as free. `_get_cache_prices` keeps its existing "free reads as no pricing" guard ahead of the call, so nothing downstream of the cost card changes.

On the regression: `test_missing_cache_prices_price_exactly_as_the_cost_card_does` now drives BOTH real paths against one cache-rate-less catalog entry - it calls `CostTracker._get_cache_prices` for real and asserts the estimator's read/write/uncached output equals `1M * card_rate` for each bucket, rather than asserting a constant. Your exact example is the fixture: `{"input_cost_per_token": 5e-6}` now yields `(5e-6, 5e-6, 5e-6)` on both sides, so 1M deferred tokens quote $5 either way, warm or cold.

148 tests pass across the cost/savings suites; ruff check + format clean.

JerrettDavis (approved):
The missing-cache-price regression now compares the two real consumers, and both use list price when catalog cache metadata is absent. Request cache-mix forwarding and free-model handling also pass. All 18 focused tests pass; this resolves the previous blocker, with no new blocking finding.

gglucass:
Superseded by #3699 (89a58fd1) - closing.

This branch added `_cache_input_rates` / `_model_input_rates` / `_estimate_tool_schema_savings_usd` to price tool-schema deferral at the cache-read rate. `counterfactual.py` now does that, and more carefully than this did:

- `savings_tracker` prices the tool-schema bucket with `Region.PREFIX`, and `split_tokens` fills the cache-read bucket FIRST for prefix content rather than apportioning it pro-rata across the request's mix. Pro-rata was the flaw this branch still had - on a 90/10 read/write request it hands prefix tokens a share of the 1.25x write bucket they never occupied.
- `resolve_rates` handles the long-context tiers, TTL ratios and provider fallbacks this branch's flat catalog read did not, and stamps a `basis` so an unpriceable model is visible rather than silently list-priced.

Verified on the rebased #3170 against the installed wheel: 4,000 deferred tokens on a 90k-read/10k-write request price at $0.0012 versus $0.012 at list - the 10x cache-read discount this PR was asking for, arriving through the shared entry point.

Closing in favour of that. #3170 (the disjoint tool-schema accumulator) is NOT superseded and I have rebased it onto v6 - it records the layer beside the folded headline, which #3699 does not do.

PR #3172, opened 2026-08-21
fix(proxy): drop a tool_reference naming the search tool itself

orty:
## Description

Anthropic can answer a match-all `tool_search_tool_regex` call (empty `input`) with a hit on **its own server tool**. Claude Code adds every hit to the session's loaded-tool set and replays it as a `tool_reference` in the `tools` array, so every later turn 400s with:

```
Tool reference 'tool_search_tool_regex' not found in available tools
```

A typed search tool is the search mechanism, never a reference target.

The history repair from #2805/#2971 cannot reach this: it walks `messages` only, and the poison is in `tools`. That is why `/compact` does not clear it — the session is dead from the first poisoned turn until the transcript is edited by hand.

This is consistent with a principle the codebase already holds. `strip_unsupported_tool_search_blocks` excludes typed search tools when building its `available` set, with the comment:

> Typed search tools (type starts with `tool_search_tool_`) are the search mechanism itself — they are never the target of a tool_reference lookup.

That reasoning is applied to `messages`. This PR extends the same rule to the `tools` array, which the existing function only reads and never filters.

## Evidence

Observed on a real Claude Code session (`claude-cli/2.1.237`), poisoned at the one search whose result self-referenced — 1 of 6 searches in that session, both of the others with empty `input` came back clean, so it is nondeterministic upstream behavior.

The failure splits by request class:

| | headroom session | outcome |
|---|---|---|
| Main-thread turns | `inject_sticky_replay` | 200, every time |
| Auto-compact side-requests | `inject_first_time`, `CCR: skipping proactive tracking for Claude Code compact summary` | 400, every time |

The repair fired on every failing request (`Tool search: dropped 2 unsupportable history block(s)`) and the request still 400'd.

That the rejected content is the client's own is visible on the wire: on all four 400s the outbound `body_bytes` equalled the inbound `content_length` exactly (658154, 523845, 525028, 511860), while across the same log only 10 of 701 `source=canonical` rows match any inbound size at all.

Switching model gave exactly one clean turn — the side-request state reset, its two fresh searches came back clean — then the next auto-compact side-request replayed the poisoned array and 400'd again.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/helpers.py` — new `strip_unsupported_tool_search_references(tools)` drops `tool_reference` entries whose name starts with `tool_search_tool_`. Returns the **original** object when nothing was removed, so the caller skips the write-back and the prefix cache is undisturbed on ~all requests (same identity contract as `strip_unsupported_tool_search_blocks`).
- `headroom/proxy/handlers/anthropic.py` — wired in immediately **before** the block repair. Ordering matters: stripping the tools array shrinks what a history `tool_reference` can resolve against, so the block repair must validate against the final view — the same constraint as #2888.
- `tests/test_issue_746_tool_search.py` — two cases: a poisoned array is cleaned, and a clean array is an identity no-op.

## Testing

Run against **v0.36.2** with this change cherry-picked (clean apply, no conflicts).

### Test Output

```text
$ python -m pytest tests/test_issue_746_tool_search.py -q
60 passed in 0.52s

$ python -m pytest tests/test_proxy/test_tool_search_repair_after_turn_hooks.py -q
4 passed in 1.34s

# broader tool-search regression
$ python -m pytest tests/test_issue_746_tool_search.py \
    tests/test_proxy/test_tool_search_repair_after_turn_hooks.py \
    tests/test_openai_tool_search_deferral.py \
    tests/test_savings_tool_search_aggregation.py -q
103 passed in 1.84s

$ python -m ruff check headroom/proxy/helpers.py headroom/proxy/handlers/anthropic.py tests/test_issue_746_tool_search.py
All checks passed!

$ python -m ruff format --check headroom/proxy/helpers.py headroom/proxy/handlers/anthropic.py tests/test_issue_746_tool_search.py
3 files already formatted

$ python -m mypy headroom/proxy/helpers.py headroom/proxy/handlers/anthropic.py
```

`tests/test_proxy/test_tool_search_repair_after_turn_hooks.py` is the file most adjacent to this change — it asserts what the repair sees in the tools array after the turn hooks run. An earlier revision of this description reported it as un-runnable. That was an environment gap, now resolved: the blocker is the optional **`h2`** dependency (`ImportError: Using http2=True, but the 'h2' package is not installed`), not the Rust extension. With `pip install 'httpx[http2]'` the file collects and all 4 tests pass.

Those 4 tests were also confirmed to fail identically on **clean v0.36.2 without this patch**, so the earlier failure was environmental and not a regression introduced here.

## Real Behavior Proof

- Environment: live Claude Code session (`claude-cli/2.1.237`) through a self-hosted headroom proxy; plus the unit-level runs above against v0.36.2 on Linux / Python 3.11.
- Exact command / steps: normal Claude Code usage until a match-all `tool_search_tool_regex` returned a hit naming `tool_search_tool_regex` itself; thereafter every auto-compact side-request replayed the poisoned `tools` array. Proxy logs inspected for `body_bytes` vs inbound `content_length` and for the repair's own log lines.
- Observed result: persistent `400 Tool reference 'tool_search_tool_regex' not found in available tools` on auto-compact side-requests while main-thread turns returned 200; the existing block repair logged `dropped 2 unsupportable history block(s)` and the request still 400'd, confirming the poison is in `tools` and out of that repair's reach. `/compact` did not clear it.
- Not tested: no live end-to-end rerun of a poisoned session against a proxy carrying this patch — the trigger is nondeterministic upstream behavior (1 of 6 searches) and has not been reproduced on demand. The fix is covered at unit level by the added cases plus the 4 after-turn-hook tests above.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — unconditional repair on an already-broken request shape, no flag.
- Minimum rollout channel: n/a — ships with the change.
- Stable/default behavior changed: only for requests that are currently rejected by Anthropic. A `tool_reference` naming a `tool_search_tool_*` entry is never resolvable, so removing it cannot change the outcome of a request that works today.
- Kill switch / disable path: none. The function is a no-op (returns the original object, identity-preserved) whenever no such entry is present, which is ~all traffic.
- Qualification impact: none. Sessions that were 400ing continue past the poisoned turn; unaffected sessions see a byte-identical `tools` array, so prefix caching is undisturbed.
- Rollback path: revert the commit; single and self-contained.

## Review Readiness

## Checklist

N/A — terminal and log output above.

## Additional Notes

- **Documentation:** marked N/A. No user-facing flag, endpoint, or configuration surface changes; the repair is internal to the Anthropic request path.
- **Ordering constraint:** the call site must stay ahead of the block repair. If a later change reorders them, `strip_unsupported_tool_search_blocks` will validate history against a stale tools array and the original 400 can return in a different shape.
- **Identity contract:** returning the original object on a no-op is load-bearing, not a micro-optimization — the caller keys its write-back on identity, and a fresh object on every request would perturb the prefix cache.


JerrettDavis (changes requested):
The repair is correctly ordered before history validation, and preserving object identity on a no-op is the right cache-safe contract. One correctness edge needs narrowing before merge: `strip_unsupported_tool_search_references()` currently drops any `tool_reference` whose referenced name merely starts with `tool_search_tool_`, even if the request has no typed search mechanism with that name. Tool-reference targets are user/provider tool names, and this patch should not establish an undocumented reserved name namespace or remove a legitimate deferred tool solely by prefix. Please first derive the names of actual search mechanisms in this same tools array (entries whose `type` starts with the typed-search prefix), then drop only `tool_reference` entries whose name exactly matches one of those mechanism names. Add regressions proving (1) the observed self-reference is removed, (2) a normal referenced tool with a `tool_search_tool_*` name remains when it is not the typed mechanism, and (3) a prefixed reference remains when no matching mechanism is present. The existing ordering and identity tests should remain.

JerrettDavis (approved):
Re-reviewed the scoped repair at exact head 06c1eee6. Mechanism names are derived only from typed search entries present in the same request; removal requires an exact name match; prefix-shaped client tools and references without a matching mechanism remain untouched; and the original list identity is preserved on every no-op path. Ordering before history repair remains correct. Current-main refresh clears the inherited Agno failure; newly triggered CI remains authoritative. Approval restored for final human review.

JerrettDavis (approved):
Reviewed exact head 19713cf6b1. The repair is correctly scoped to a tool_reference whose exact name matches a typed search mechanism present on the same request, preserving legitimate deferred references and clean-array identity. Current-main merge simulation was conflict-free; all 64 focused tests and Ruff passed locally, and hosted checks are green. Approving.

chopratejas (commented):
Reviewed at head `19713cf6`. **The change is sound — the scoping is exactly right.** One blocker that is about the base, not the code.

`mechanism_names` is derived from entries whose `type` starts with `tool_search_tool_` *on this request*, and a `tool_reference` is dropped only on an exact name match. So a typeless client tool legitimately named `tool_search_tool_custom` survives, and a prefix-shaped reference on a request that carries no mechanism survives. Both have tests. The identity return on the no-op path means the common request never rewrites `body["tools"]`, so the prefix cache is left alone — that detail is easy to get wrong and you got it right.

### Verified locally

- Merges onto current `main` with no conflicts.
- 64 tests in `tests/test_issue_746_tool_search.py` pass on the merged tree.
- Broad sweep (`-k "tool_search or anthropic or handler or deferral or tools"`) on the merged tree: **1396 passed** vs **1391** on plain `main` — exactly your 5 new tests, nothing else moved. (Three failures in `test_read_maturation_handler_nobust.py` appear identically on plain `main` under that selection; they are pre-existing order-dependent pollution, unrelated to this PR.)

### Blocker: the base has moved under this branch

The branch is **45 commits behind**, and `take_tool_search_hint_slot` was **removed from `main`** by #3745 (`fix(tool-search): make the core-tools override authoritative`) after this branch was cut. I hit that directly: the branch's test file cannot import against today's `main`.

The three-way merge resolves it correctly, so there is no conflict to fix by hand. But **this PR's green CI is from 09-22 and predates #3745**, so it is not evidence about current `main`. Please rebase (or push an empty commit) so CI re-runs against a base that includes #3745 before this is merged.

### Two minor notes, neither blocking

1. **Name-precedence inversion.** The new matcher reads `t.get("name") or t.get("tool_name")`. The existing `_tool_search_reference_names` reads `entry.get("tool_name") or entry.get("name")`, with the comment "Server-side blocks use `tool_name`; be liberal about `name`." The two disagree for an entry carrying both keys with different values. Worth aligning with the existing precedence so there is one rule in this file.

2. `mechanism_names` is built from `t["name"]` only, which is asymmetric with the deliberately liberal reference matching. Benign in practice — tool *definitions* use `name` — but if a mechanism ever arrives shaped like a server-side block it would not register.

Rebase for CI and I am happy with this.

PR #3173, opened 2026-08-21
fix: strip x-openai-internal-* headers when proxying to upstreams

GrabbyGrabby:
## Description

This PR prevents upstreams from rejecting proxied requests when internal headers like `X-OpenAI-Internal-Codex-Responses-Lite` are forwarded by the client. These headers are stripped case-insensitively before proxying.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `strip_openai_internal_headers(headers)` utility function to case-insensitively pop all `x-openai-internal-*` headers.
- Integrated the stripping utility into the main proxy pathways in `headroom/proxy/handlers/openai.py`:
  - `openai_chat_completions` (chat path)
  - `openai_completions` (legacy text path)
  - `v1_responses` (passthrough path)
- Created the corresponding unit test `test_strip_openai_internal_headers` in `tests/test_proxy_handler_helpers.py` to assert header scrubbing correctness.

## Testing

### Test Output

```text
pytest tests/test_proxy_handler_helpers.py
...
✔ test_strip_openai_internal_headers passes (1.02ms)

### DIFF
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -3112,6 +3112,8 @@ class OpenAIHandlerMixin:
         # Cloudflare Workers forward "br, zstd" which OpenAI may honor;
         # if httpx lacks brotli support the response body is undecipherable → 502.
         headers.pop("accept-encoding", None)
+        from headroom.proxy.helpers import strip_openai_internal_headers
+        headers = strip_openai_internal_headers(headers)
         tags = extract_tags(headers)
         from headroom.proxy.savings_attribution import bind_scope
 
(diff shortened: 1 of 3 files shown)


JerrettDavis (changes requested):
Current main already contains the narrowly scoped fix for the reported Codex failure: both Responses HTTP paths remove x-openai-internal-codex-responses-lite case-insensitively, with route-level coverage in test_handle_openai_responses_strips_codex_lite_header_upstream (and the WS metadata mirror is handled separately). This branch would now broaden that behavior to delete *every* x-openai-internal-* header from chat completions, Responses, and generic passthrough without evidence that every such header is invalid or client-only. That can silently disable future/first-party OpenAI protocol features and is a materially larger contract than the bug requires.

Please rebase and verify the remaining diff against current main. If there is a distinct still-reproducing header/upstream case, scope removal to the specific known header and affected route/target, with outbound-capture coverage proving the rejection and preserving unrelated internal headers. Otherwise this PR is functionally superseded and should be closed rather than landing a redundant, overbroad filter. The current patch also needs normal formatting cleanup, but the routing scope is the blocker.

PR #2346, opened 2026-07-17
feat(dashboard): reorganize settings GUI into feature-category pages

nangsontay:
## Description

The settings GUI (`/dashboard/settings`) split every knob across two **global**
tabs — "Settings" (basic) and "Advanced" — so a feature's basic and advanced
knobs landed on different tabs and you could never see one feature whole.
This reorganizes the GUI into **feature-category pages** with a **left sidebar**.
Each page shows its Normal knobs inline plus a **collapsible Advanced section**
(collapsed by default; auto-expanded on all-advanced pages). Two net-new pages:

- **Output Shaping** — the *live* `runtime_env` knobs (output shaper, verbosity,
  effort router, holdout…). These apply **without a restart**: Save persists to
  `settings.json` **and** hot-reloads via `runtime_env.set_overrides`, effective
  on the next request — the same in-memory path `headroom wrap` uses.
- **Observability** — new curated store knobs for OpenTelemetry metrics,
  Langfuse, the anonymous telemetry beacon, and periodic TOIN stats. Secret
  OTel/Langfuse keys/headers stay env-only.

The `SettingField` registry already carried `group` + `tier`; the data-model add
is small — a derived `page` (nav category, mapped from `group`) + a `live` marker.

### Follow-up work in this PR

- **Curated-knob expansion (56 → 95 fields).** Filled the biggest GUI gaps:
  `HEADROOM_MODE` (token/cache), Qdrant (URL/host/port/API-key), the
  Vertex/Bedrock/Gemini/Cloud Code upstream base URLs, CCR backend/Redis URL/TTL,
  networking & security (stateless / offline / strict-TLS / CORS + WS origins),
  the Kompress engine backend, dedup / tool-search, and the core CLI toggles
  (optimization / cache / rate-limit on-off, tool-result interception,
  embedding-server sidecar). Every env format is verified against its own reader.
- **Settings-first precedence.** A saved GUI setting now overrides a
  shell-exported env var: `CLI arg > settings.json > env > default`.
  `apply_to_environ` assigns instead of `setdefault`. `manifest_managed` knobs
  (`HEADROOM_HOST`/`HEADROOM_PORT` on supervised Docker/service installs) are the
  one exception — the install manifest keeps precedence, so a stale
  `settings.json` can never override a container's bind host/port.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
[...]

PR #3177, opened 2026-08-21
fix(settings): report live proxy configuration

### LINKED ISSUE #3090: `/settings` and `/settings/schema` report the wrong running config: CLI flags have no tier, `mode` has no field
headroom-ai 0.35.0, macOS 15 (darwin 25.5.0), Python 3.13.15 (uv tool venv).

No endpoint states the running proxy configuration truthfully. Two separate causes, one
of them worse than a plain omission — a wrong value flagged `env_override: true`.

## Repro

Start a proxy with a flag and one env var:

```
HEADROOM_MODE=cache HEADROOM_PROTECT_TOOL_RESULTS=Bash,Read,Glob,Grep,Write,Edit,WebFetch,TodoWrite \
  headroom proxy --port 8787 --lossless
```

Read it back:

```
$ curl -fsS http://127.0.0.1:8787/settings
{}
$ curl -fsS http://127.0.0.1:8787/settings/schema | jq '.fields[] | select(.key=="lossless")'
{ "key": "lossless", "env": "HEADROOM_LOSSLESS", "value": false, "stored": null, "env_override": true, ... }
$ curl -fsS http://127.0.0.1:8787/settings/schema | jq '[.fields[].key] | index("mode")'
null
```

Proxy runs `--lossless`. Schema says `lossless: false`, and marks it `env_override: true`, so
it reads as a deliberate setting rather than a gap. `mode` is honored (`HEADROOM_MODE=cache`
changes behavior) but absent from all 59 fields. `/settings` is `{}` because nothing was ever
stored.

Only `ps eww -o command= -p <pid>` tells the truth.

## Cause

1. **No CLI tier.** `headroom/settings_store.py:950` `effective_values` resolves
   `default ← file ← environ`. A `--flag` never enters that chain. It lands on the config object
   (`headroom/proxy/models.py:240` `lossless: bool = False  # CLI: --lossless`), which the
   endpoint does not read.
2. **`env_override: true` on a value nobody set.** Proxy entry calls `seed_proxy_env_defaults()`
   (`headroom/proxy/server.py:5137`, `:5205`), which `setdefault`s the savings profile into
   `os.environ`, including `"HEADROOM_LOSSLESS": "1" if self.lossless else "0"`
   (`headroom/agent_savings.py:91`, `:355`). Profile `coding` is not lossless, so `HEADROOM_LOSSLESS=0`
   is seeded. `env_override` is `bool(os.environ.get(field.env))` (`headroom/settings_store.py:1006`),
   and `"0"` is truthy as a string. Result: false value, authoritative-looking flag.
   `ps eww` on the proxy confirms the var is not in the environ it was started with.
3. **`mode` has no `SettingField`.** `grep -c 'SettingField(' headroom/settings_store.py` is 59,
   none keyed `mode`, though the same seeding writes `"HEADROOM_MODE": self.proxy_mode`
   (`headroom/agent_savings.py:70`).

## Expected

`/settings/schema` reports what the process runs. Either:

- add a CLI tier so `effective_values` becomes `default ← file ← environ ← cli`, resolved from the
  live config object, not re-read from env; and
- register `mode` as a field.

Minimum useful fix, if the tier is too big: stop reporting `env_override: true` for a value the
process seeded itself. Distinguish user env from profile seed — e.g. snapshot `os.environ` before
`seed_proxy_env_defaults()` and test `env_override` against that snapshot.

## Why it matters

`--lossless` is the workaround for a streaming bug (a `stream:true` request served as a buffered
non-stream call answers the client an empty HTTP 200). Verifying the workaround is on is a
per-session check, and the endpoint that should answer it states the opposite.

Related: #2812 — same registry, write path.

Sabeekhann:
## Description

The settings schema re-reads process environment values after the CLI has already resolved the live proxy configuration. Savings-profile defaults seeded later in startup can therefore make `/settings/schema` report values that are not actually active. For example, `--lossless` can produce `ProxyConfig.lossless=True` while the schema reports the seeded `HEADROOM_LOSSLESS=0`.

This change makes the live `ProxyConfig` the highest-precedence settings source, preserves provenance for profile-generated environment defaults, and adds the missing `HEADROOM_MODE` registry field.

Closes #3090

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `HEADROOM_MODE` with canonical `token` and `cache` choices.
- Made live `ProxyConfig` values take precedence in `/settings/schema`.
- Added `runtime_override` metadata for live values that differ from lower tiers.
- Prevented savings-profile defaults from being labeled as explicit environment overrides.
- Preserved provenance through multi-worker configuration serialization.
- Added regression tests for precedence, provenance, endpoint reporting, and worker round-tripping.

## Testing

- [ ] Unit tests pass (`pytest`) — focused settings tests pass; full suite not run
- [x] Linting passes (`ruff check .`) — changed files checked
- [ ] Type checking passes (`mypy headroom`) — not run

### Test Output

```text
$ focused settings-store test run
36 passed, 1 deselected in 0.09s

$ direct savings-profile provenance check
seed provenance check passed (18 generated keys)

$ ruff check <8 changed files>
All checks passed!

$ ruff format --check <8 changed files>
8 files already formatted

$ python -m compileall -q <8 changed files>
# No output

$ git diff --check
# No output
```

## Real Behavior Proof

- Environment: Linux x86_64, Python 3.12.13, source based on upstream `5e0ce242` (v0.36.2).
- Exact command / steps: Set profile-seeded `HEADROOM_MODE=cache` and `HEADROOM_LOSSLESS=0`, then generated the schema with live values `mode=token` and `lossless=True`, excluding those generated keys from explicit-environment provenance.
- Observed result: The schema reports `mode=token` and `lossless=true`; both generated variables report `env_override=false`, and both differences report `runtime_override=true`.
- Not tested: Complete repository and FastAPI endpoint suites. This environment lacks the compiled `headroom._core` extension and Rust toolchain, and dependency downloads are unavailable.

## Runtime Rollout Safety

- Stable/default behavior changed: Proxy compression behavior is unchanged; only settings registry/schema reporting changes.
- Qualification impact: The dashboard receives the actual live proxy values.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A; no user documentation change required
- [ ] My changes generate no new warnings — focused checks pass; full suite not run
- [ ] New and existing unit tests pass locally — focused tests pass; full suite not run
- [x] I did **not** edit `CHANGELOG.md` — generated by release-please

N/A — settings API behavior only.

## Additional Notes

This is intentionally narrow for #3090. PR #2346 overlaps by adding `HEADROOM_MODE`, but currently includes unrelated dashboard/runtime work; this PR isolates the live-config precedence and provenance correction.


JerrettDavis (approved):
The schema now models the actual precedence chain through the live ProxyConfig, while separately retaining whether lower-tier environment values were operator-provided or profile-generated. The alias/inverted-flag projection covers the curated fields, secret masking remains at serialization, and the seeded-key provenance survives the parent/worker payload round trip. Focused endpoint, store, profile, and worker tests cover the regression. I found no blocking issue in the current head.

Issue #3178, opened 2026-08-21
[BUG] Native GIL hold in single-cache-miss compression starves /readyz for 60s+; the in-process watchdog cannot fire by construction


gglucass:
## Description

A native-code stall inside the single-cache-miss compression path can hold the GIL long enough to freeze the entire proxy — `/readyz` stops answering while the TCP listener keeps accepting connections — and the in-process watchdog that is supposed to bound this exact path cannot fire, because waking from its timed `join` itself requires the GIL.

The watchdog's own code comment already names this ceiling ("daemon watchdog cannot stop native GIL holds; native layer owns that fix" — `content_router.py`, single-miss watchdog block). This report is field confirmation that the ceiling is being hit, with a thread dump from a production install.

Observed from the Headroom Desktop fleet (which supervises the proxy and probes `/readyz`): 24 events across 7 users since 2026-07-23, regressed and most recent on wheel 0.35.0 / macOS. When it happens the desktop marks the backend unreachable after 3 failed probes and auto-pauses routing, so users silently lose optimization until restart.

## To Reproduce

Not deterministically reproducible from here — captured in the field. The shape is:

1. Proxy serving normally, then a request enters the single-cache-miss compression path (`ContentRouter.apply`, pass 2, `len(pending_tasks) == 1`).
2. The spawned `headroom-single-compress-watchdog` worker enters native code that does not release the GIL.
3. Every Python thread freezes: uvicorn stops answering `/readyz`, the log goes silent, while process CPU keeps burning (native spin) and the kernel backlog keeps accepting TCP.

## Expected Behavior

The 20s deadline (`HEADROOM_COMPRESSION_DEADLINE_MS` default) fails the request open via PASSTHROUGH, and `/readyz` keeps answering throughout — health checks should not be starvable by one request's compression.

## Actual Behavior

61+ seconds of full-process silence, three times the deadline. The timed `join(20.0)` expires in the OS but the joining thread cannot resume — resuming from `_wait_for_tstate_lock` needs the GIL the runaway worker is holding — so the watchdog is inert precisely in the case it exists for.

Thread dump captured at the 61s mark (paths anonymized; note `log_silent_secs=61`, `process_cpu_secs=61`, `port_accepts_tcp=True`, `backend_readyz_outcome=timeout`):

```text

Thread 0x0000000170653000 (most recent call first):
[...]

PR #3180, opened 2026-08-21
fix(proxy): hard watchdog that dumps and exits when a native call seizes the GIL

gglucass:
## Description

First aid for #3178. Every watchdog the proxy currently has — the compression path's timed thread joins, asyncio timeouts, uvicorn itself — needs the GIL to act. A native call that computes without releasing the GIL therefore freezes the entire process beyond all of their reach: `/readyz` stops answering while the kernel backlog keeps accepting connections, logging goes silent, and the 20s compression deadline expires without its joiner being able to wake (resuming from a timed `join` needs the very GIL the runaway worker holds). The field report in #3178 shows exactly this: 61+ seconds of full-process silence at three times the deadline, diagnosed only because the desktop supervisor happens to inject a SIGUSR1 faulthandler dump before force-killing. Standalone and docker deployments have neither the dump nor the recovery — they zombie until a human notices every client hanging.

`faulthandler.dump_traceback_later` is the one escape hatch the interpreter offers: its timer runs on a C thread that never takes the GIL, and on expiry it writes every thread's stack with signal-safe code and hard-exits. This PR arms it as a last-resort liveness watchdog: a heartbeat thread re-arms the timer while the interpreter is healthy; if the GIL is seized past the deadline, the timer fires — writing the culprit's exact call site into stderr (the proxy log) and ending a process that was already unable to serve anything. Supervised deployments restart it automatically; unsupervised ones trade an eternal zombie that hangs every client for a fast-failing crash with a self-identifying diagnosis. Future occurrences of #3178 will name the native call we currently cannot.

The deadline is deliberately generous: 90 seconds against a 20s compression deadline and 15s stats timeouts, so it can only ever fire after every softer recovery path has already been starved out.

## Type of Change

## Changes Made

- New `headroom/proxy/hard_watchdog.py`: `start_hard_watchdog()` arms `faulthandler.dump_traceback_later(secs, exit=True)` synchronously (covered from the moment it returns, even if the very next call seizes the GIL), then a daemon heartbeat re-arms it every `secs/3`. `HEADROOM_HARD_WATCHDOG_SECS` tunes the deadline (default 90; `0` disables; floor 5s so scheduler jitter cannot shoot a healthy process). `atexit` cancels the timer for clean shutdowns; CPython additionally cancels it during finalization.
- Armed at the top of the app lifespan, per worker (the timer and the GIL are both process-local), before `proxy.startup()` so a startup-phase seizure is covered too.
- `tests/test_hard_watchdog.py` (4 tests): the GIL seizure is simulated with `ctypes.PyDLL` (unlike `CDLL`, calls through it do not release the GIL, so `libc sleep()` stands in perfectly for a GIL-holding native call). Pins: seized process is dumped and exited; healthy process running past the deadline is never shot (re-arm works); env parsing/clamping; disable path.

Deliberate semantics worth calling out:

- `exit=True` uses `os._exit` — no `atexit`, no `finally`. The savings tracker's throttled flush can lose at most one throttle window, identical to the supervisor's force-kill today; a frozen process was not going to flush anything either.
- The watchdog also arms inside any test process that runs the app lifespan. That is intentional: the timer only fires after 90 consecutive seconds in which no Python thread ran, which in CI is precisely the pathology worth killing loudly instead of timing out a runner six hours later.

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_hard_watchdog.py -v
tests/test_hard_watchdog.py::test_resolve_secs_defaults_and_clamps PASSED
tests/test_hard_watchdog.py::test_seized_gil_is_dumped_and_exited PASSED
tests/test_hard_watchdog.py::test_healthy_process_is_never_shot PASSED
tests/test_hard_watchdog.py::test_disabled_by_env PASSED
============================== 4 passed in 19.50s ==============================

$ uv run --frozen --extra dev pytest tests/test_proxy_favicon_route.py tests/test_proxy_debug_endpoints.py
======================== 30 passed, 1 warning in 10.32s ========================

$ uv run --frozen --extra dev ruff check headroom/proxy/hard_watchdog.py headroom/proxy/server.py tests/test_hard_watchdog.py
All checks passed!
$ uv run --frozen --extra dev mypy headroom/proxy/hard_watchdog.py
```

## Real Behavior Proof

- Environment: macOS, Python 3.10 via `uv run --frozen`, this branch.
- Exact command / steps: arm the watchdog at a 5s deadline, then seize the GIL for 60s via `ctypes.PyDLL(libc).sleep(60)`, which never releases it.
- Observed result: at the 5s mark the C timer fires despite the held GIL — `Timeout (0:00:05)!` followed by every thread's stack, ending with the culprit's exact call site (the seizing line itself), then exit code 1 instead of 55 more seconds of frozen process:

```text
$ HEADROOM_HARD_WATCHDOG_SECS=5 python -c '...start_hard_watchdog(); PyDLL(libc).sleep(60)...'
seizing the GIL via PyDLL libc.sleep(60)...
Timeout (0:00:05)!
Thread 0x000000016d57f000 (most recent call first):
  File ".../headroom/proxy/hard_watchdog.py", line 70 in _heartbeat
  ...
Thread 0x00000001efeda2c0 (most recent call first):
  File "<string>", line 6 in <module>        <- the GIL-seizing call site
$ echo $?
1
```

- Not tested: a live reproduction of the original native stall from #3178 (not deterministically reproducible — the point of this watchdog is that the next field occurrence documents itself), and Windows (the libc-based seizure harness is POSIX-only; the watchdog itself is platform-independent faulthandler).

## Runtime Rollout Safety

- Rollout-managed feature(s): None — no rollout-gated feature is touched; this is a process-liveness guard.
- Stable/default behavior changed: Yes, deliberately: a process whose GIL is seized for 90+ consecutive seconds now dumps all stacks and exits 1 instead of hanging forever. Healthy processes are unaffected (heartbeat re-arms at deadline/3; pinned by `test_healthy_process_is_never_shot`).
- Kill switch / disable path: `HEADROOM_HARD_WATCHDOG_SECS=0` disables it entirely; any larger value retunes the deadline.
- Qualification impact: None beyond the suites above; the watchdog writes nothing and takes no action unless the process is already unable to serve.
- Rollback path: `git revert` — no persisted state, no config migration; the env var simply stops being read.

## Review Readiness

### DIFF
--- a/headroom/proxy/hard_watchdog.py
+++ b/headroom/proxy/hard_watchdog.py
@@ -1,0 +1,103 @@
+"""Last-resort liveness watchdog that survives a held GIL.
+
+The in-process compression watchdogs bound slow work with timed thread joins,
+which quietly assumes the GIL keeps changing hands: resuming from a timed
+``join`` needs the GIL, so a native call that computes without releasing it
+freezes every Python thread at once -- uvicorn, ``/readyz``, logging, and the
+very watchdog meant to bound the offender. Field report: 61+ seconds of
+full-process silence at three times the compression deadline, TCP backlog
+still accepting connections nothing would ever answer (#3178).
+
+``faulthandler.dump_traceback_later`` is the one escape hatch the interpreter
+offers: its timer runs on a C thread that never takes the GIL, and on expiry
+it writes every thread's stack with signal-safe code and (optionally) hard
+exits. A Python heartbeat thread re-arms the timer while the interpreter is
+healthy; if the GIL is seized for longer than the deadline, the heartbeat
+cannot re-arm and the C timer fires -- naming the culprit stack in stderr
+(the proxy log) and ending a process that was already unable to serve
+anything. Supervised deployments restart it; unsupervised ones trade an
+eternal zombie that hangs every client for a visible crash with a diagnosis.
+
+The deadline is deliberately generous (90s against a 20s compression
+deadline and 15s stats timeouts): this must only ever fire when the process
+is beyond every softer recovery path.
+"""
+
+from __future__ import annotations
+
+import atexit
+import faulthandler
+import logging
+import os
+import sys
+import threading
+import time
+
+logger = logging.getLogger(__name__)
+
+ENV_VAR = "HEADROOM_HARD_WATCHDOG_SECS"
+DEFAULT_SECS = 90.0
+# Below this the heartbeat interval (a third of the deadline) gets close to
+# scheduler jitter and a busy-but-healthy interpreter could be shot.
+MIN_SECS = 5.0
+
+_started = threading.Event()
+
+
+def _resolve_secs() -> float:
+    raw = os.environ.get(ENV_VAR, "")
+    if not raw.strip():
+        return DEFAULT_SECS
+    try:
+        secs = float(raw)
+    except ValueError:
+        logger.warning("Ignoring non-numeric %s=%r; using %ss", ENV_VAR, raw, DEFAULT_SECS)
+        return DEFAULT_SECS
+    if secs <= 0:
+        return 0.0
+    return max(secs, MIN_SECS)
+
+
+def _arm(secs: float) -> None:
+    # Re-arming replaces the previous timer, so a healthy interpreter never
+    # lets it expire. ``file`` is the underlying stderr fd, which supervisors
+    # already redirect into the proxy log.
+    faulthandler.dump_traceback_later(secs, exit=True, file=sys.stderr)
+
+
+def _heartbeat(secs: float) -> None:
+    while True:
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
The C-level timer plus Python heartbeat correctly targets the failure class that ordinary asyncio/thread watchdogs cannot observe: a native call holding the GIL. It arms before startup native work, is per-process/idempotent, has a conservative default and explicit kill switch, and the subprocess tests prove both hard-exit stack capture under a seized GIL and continued survival while the interpreter makes progress. Current-head CI is fully green. I found no blocking correctness issue.

PR #3181, opened 2026-08-21
ci: verify documented environment variables

### LINKED ISSUE #3103: [CI] Check that every environment variable named in the docs is actually read by the code
### Problem

Documentation names environment variables that the code does not read, and nothing catches it.

The concrete case, found during review of #2684: both `SECURITY.md` and the security-model page told operators to relocate the plaintext Copilot OAuth refresh token with `$HEADROOM_COPILOT_AUTH_PATH`. The implementation reads `HEADROOM_COPILOT_AUTH_FILE` (`headroom/copilot_auth.py::headroom_copilot_auth_path()`), and has for a while. An operator following the security guidance would have set a variable that does nothing, believed the token had moved, and left it at the default path.

That is the failure mode that makes this worth automating rather than remembering: a documented knob that silently does not exist is worse than an undocumented one, because it produces false confidence. Security docs are where it costs the most, but the same risk applies to every configuration page.

### Proposed check

A CI step that extracts every environment-variable-shaped token from the docs and asserts each resolves to a read site in the source:

- **Scan**: `docs/content/docs/**/*.mdx`, `SECURITY.md`, `README.md`
- **Extract**: identifiers matching `HEADROOM_[A-Z0-9_]*` plus the small set of non-prefixed names the proxy reads (`ANTHROPIC_*`, `OPENAI_*`, `DO_NOT_TRACK`)
- **Assert**: each appears as a string literal in `headroom/` or `crates/`
- **Handle**: documented wildcards such as `HEADROOM_OTEL_METRICS_*` should match by prefix rather than fail

I ran this by hand against #2684 — 30 names referenced, all resolving — so the shape is known to work and the current docs are known to be clean. Roughly 30 lines of Python; it needs no import of the package, only a grep over the tree, so it can run in the docs workflow without installing dependencies.

### Possible extension

The inverse direction — env vars read in `headroom/` that appear in no documentation — is more useful still but much noisier, since plenty are internal. If it is wanted, it should probably start as a report rather than a gate.

### Context

Raised in review of #2684 by @JerrettDavis ("Please use the implemented variable and add a doc/source consistency check if practical") and agreed there as a follow-up rather than growing a docs-only PR into tooling.

Sabeekhann:
## Description

Add a dependency-free CI guard that verifies environment variables documented in `README.md`, `SECURITY.md`, and `docs/content/docs/**/*.mdx` still have either a repository implementation reference or an explicitly reviewed external-consumer policy entry.

This prevents misspelled or removed configuration names from remaining in the documentation and giving users false confidence that a setting is active.

Closes #3103

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `scripts/ci/verify_documented_env_vars.py` to compare documented environment variables with implementation source.
- Added generic syntax/context-aware environment-variable discovery rather than a fixed provider-prefix allowlist.
- Added reviewed exclusions for non-environment uppercase constants and an explicit external-consumer policy for variables consumed by libraries/tools outside direct repository read sites.
- Added scanning for Python, Rust, SDK, plugin, deployment, Docker, installer, and workflow sources while excluding tests, examples, and fixtures from implementation evidence.
- Added CI execution for both documentation and relevant source changes.
- Added repository-level representative-family assertions and focused regression coverage.
- Added an explicit regression in `01e1cb21` proving verifier tests cannot satisfy implementation evidence for themselves.

## Testing

- [x] Unit tests pass (`pytest`) — last verified before commit `01e1cb21`
- [x] Linting passes (`ruff check .`) — last verified before commit `01e1cb21`

### Test Output

Last verified results before the latest regression-only commit:

```text
$ python scripts/ci/verify_documented_env_vars.py
Verified 185 documented environment variables (167 against repository source; 18 against the approved external-consumer policy).

$ python -m pytest scripts/tests -q
77 passed

$ ruff check .
All checks passed!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,7 @@ jobs:
       dashboard: ${{ steps.filter.outputs.dashboard }}
       packaging: ${{ steps.filter.outputs.packaging }}
       workflows: ${{ steps.filter.outputs.workflows }}
+      env_docs: ${{ steps.filter.outputs.env_docs }}
     steps:
       - uses: actions/checkout@v7
       - uses: dorny/paths-filter@v4
@@ -99,6 +100,27 @@ jobs:
               - '.github/workflows/**'
             workflows:
               - '.github/workflows/**'
+            env_docs:
+              - 'headroom/**'
+              - 'crates/**'
+              - 'sdk/**'
+              - 'plugins/**'
+              - 'deploy/**'
+              - 'docker/**'
+              - 'scripts/**'
+
+  env-doc-consistency:
+    needs: changes
+    if: needs.changes.outputs.env_docs == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.PY_VERSION }}
+      - name: Verify documented environment variables
+        run: python scripts/ci/verify_documented_env_vars.py
 
   lint:
     needs: changes
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -22,6 +22,9 @@ on:
     branches: [main]
     paths:
       - 'docs/**'
+      - 'README.md'
+      - 'SECURITY.md'
+      - 'scripts/ci/verify_documented_env_vars.py'
       - '.github/workflows/docs.yml'
   workflow_dispatch:
 
@@ -35,6 +38,13 @@ jobs:
     timeout-minutes: 15
     steps:
       - uses: actions/checkout@v6
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+
+      - name: Verify documented environment variables
+        run: python scripts/ci/verify_documented_env_vars.py
 
       - name: Setup Node.js
         uses: actions/setup-node@v6
--- a/scripts/ci/verify_documented_env_vars.py
+++ b/scripts/ci/verify_documented_env_vars.py
@@ -16,9 +16,114 @@ from collections import defaultdict
 from dataclasses import dataclass
 from pathlib import Path
 
+# Match uppercase snake-case names without assuming who owns them.  Wildcard
+# families may be written as ``VENDOR_*`` or ``VENDOR_<FAMILY>``; both are
+# normalized to ``VENDOR_*`` before comparison with concrete source names.
+# Lowercase letters are part of the boundary so fragments such as ``DS_S`` in
+# ``.DS_Store`` are not mistaken for environment variables.
 ENV_VAR_PATTERN = re.compile(
-    r"(?<![A-Z0-9_])(?:(?:HEADROOM|ANTHROPIC|OPENAI)_[A-Z0-9_]+(?:\*)?|DO_NOT_TRACK)"
-    r"(?![A-Z0-9_])"
+    r"(?<![A-Za-z0-9_])"
+    r"(?P<name>"
+    r"(?:[A-Z][A-Z0-9]*(?:_[A-Z0-9]+)*_(?:\*|<[A-Z][A-Z0-9_]*>))"
+    r"|(?:[A-Z][A-Z0-9]*(?:_[A-Z0-9]+)+)"
+    r")"
+    r"(?![A-Za-z0-9_])"
+)
+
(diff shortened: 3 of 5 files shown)


JerrettDavis (changes requested):
The guard does not currently verify 'every environment variable documented'; its extractor only recognizes HEADROOM_*, ANTHROPIC_*, OPENAI_*, and DO_NOT_TRACK. The scanned docs also document many operational variables outside those prefixes, including AWS_ACCESS_KEY_ID, AWS_ENDPOINT_URL_BEDROCK_RUNTIME, BEDROCK_TARGET_API_URL, CLAUDE_CODE_USE_{BEDROCK,FOUNDRY,VERTEX}, GOOGLE_APPLICATION_CREDENTIALS, GITHUB_COPILOT_ENTERPRISE_URL, HTTP_PROXY/HTTPS_PROXY/NO_PROXY, OTEL_EXPORTER_OTLP_*, REQUESTS_CA_BUNDLE, SSL_CERT_FILE, VERTEXAI_PROJECT, XAI_API_KEY, and others. Misspelling or removing any of those remains invisible while CI prints that all documented variables were verified.

Please either implement a syntax/context-aware generic environment-variable extractor (with a reviewed exclusion list for non-variable constants), or maintain a comprehensive explicit prefix/name policy and test representative AWS, Google/Vertex, GitHub Copilot, proxy/TLS, OTEL, and provider-key families. Add a repository-level coverage assertion that known documented variables from each family are actually discovered, not merely that discovered names resolve. Also ensure the success message describes the true checked scope if any deliberate exclusions remain. Without that, the new green gate gives materially broader assurance than it enforces.

Sabeekhann:
@JerrettDavis Thank you for the detailed review. The requested changes have been addressed in [`f7e92e54`](https://github.com/Sabeekhann/headroom/commit/f7e92e54e243dbbe54c6d8af63e25329e9402d62): generic environment-variable discovery, reviewed exclusions and an explicit external-consumer policy, repository-level discovery assertions across the requested families, accurate success-scope reporting, and protection against self-validation.

Validation passes with 77 script tests, full Ruff lint and formatting checks, and 185 documented variables verified (167 against repository source and 18 against the reviewed external-consumer policy).

Could you please re-review the updated PR?

JerrettDavis (changes requested):
The generic documentation extractor, explicit external-consumer policy, scoped success message, and representative-family assertions address most of the prior review. One self-validation hole remains: source discovery scans scripts recursively, including scripts/tests/test_verify_documented_env_vars.py. That test now contains literal AWS_ACCESS_KEY_ID, BEDROCK_TARGET_API_URL, GOOGLE_APPLICATION_CREDENTIALS, GITHUB_COPILOT_ENTERPRISE_URL, proxy/TLS names, and the other representative variables. If a real implementation reference is removed, those test literals can keep source_variables green—the exact repository assertion intended to catch the removal becomes implementation evidence for itself. Excluding only scripts/ci/verify_documented_env_vars.py is insufficient. Please exclude test directories/files (at minimum scripts/tests, preferably a clear general rule) from implementation evidence while still running them as tests, and add a regression where a name appears only in verifier tests yet remains missing. Then rerun the repository guard; any newly exposed names should be resolved through real source or the reviewed external-consumer policy. Also update the PR body's stale 114/70 counts to the current 185/77 evidence already reported in the discussion.

Sabeekhann:
@JerrettDavis Thanks for the follow-up review. I’ve added an explicit regression in `01e1cb21` that creates a variable reference only inside `scripts/tests/test_verify_documented_env_vars.py` and verifies that it is excluded from `source_variables()` and still reported as missing. This directly proves verifier tests cannot serve as implementation evidence.

The PR’s GitHub Actions runs are currently sitting in `action_required` and are not starting, so I’m not claiming fresh CI pass counts until those workflows actually execute. Once they run, I’ll update the validation counts accordingly.

Could you please re-review the updated PR?

JerrettDavis (approved):
Reviewed exact head 01e1cb210c1762e54e5f727ee7b23c6f11c2e6f7. The new regression directly verifies that names appearing only under scripts/tests do not count as implementation references, resolving my prior concern. Exact-head validation: verifier succeeded with 185 documented variables (167 repository-source, 18 approved external-consumer policy); Ruff check and format check passed; the three focused policy/regression tests passed. The full verifier test file had 12 passing tests and one pre-existing Windows-only path-separator assertion mismatch (backslashes versus the Linux-oriented expected annotation path), unrelated to this change.

PR #3192, opened 2026-08-21
fix(proxy): astgrep disk verify truncation

### LINKED ISSUE #2421: Read outliner can present a client-truncated file prefix as a complete definition list

slxng1758:
## Description

`_detect_truncation()` currently relies on client-specific truncation banners in `tool_result` text. This works for clients that emit a recognizable line-count banner, but not for clients such as Cursor (byte-based, direction-reversed truncation) and OpenCode (placeholder + external file pointer).

Because `headroom proxy` runs alongside the client it wraps, it can optionally verify the returned content against the actual file on disk. If the `tool_result` is a strict prefix of the file, we can identify truncation without depending on client-specific output formatting.

This adds disk verification as an opt-in fallback: the existing banner-based detection always runs first, and disk verification is only attempted when no banner is found. All disk-verification failures or ambiguous cases return `UNKNOWN` rather than guessing.

Following review, the verifier is also constrained to a trusted, confined workspace root rather than trusting caller-supplied input directly — see the last four bullets under Changes Made.

**Current scope:** the session-token workspace binding this fallback requires is wired for `headroom wrap claude` and plugin-loaded `headroom wrap opencode` only. Cursor remains unsupported/not tested — it has no header-injection surface today to transmit an authoritative session token, so disk verification stays `UNKNOWN` for it; extending Cursor support needs a separate, safe transport design.

Related to #2430
Closes #2421

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made
- Added optional disk-based Read truncation verification behind `HEADROOM_VERIFY_TRUNCATION_ON_DISK`.
- Disk verification compares the visible `tool_result` against the file on disk and distinguishes `COMPLETE`, `TRUNCATED`, and `UNKNOWN`.
- Disk verification only runs when the existing `_detect_truncation()` banner check finds no truncation marker, avoiding unnecessary file I/O when the client already provides a signal.
- Added a `current_cwd` contextvar populated from `x-headroom-cwd` so relative Read paths can be resolved against the client's working directory.
- Registered `HEADROOM_VERIFY_TRUNCATION_ON_DISK` as a hot-reloadable runtime setting.
- Gated verification on a new `is_current_request_trusted` contextvar, set from `is_loopback_host(request.client.host)` — a server-observed fact, not the caller-supplied header — so `x-headroom-cwd` is never sufficient authority on its own.
- Confined all path resolution strictly under the canonicalized `cwd` (`.resolve()` + `relative_to()`), rejecting `..` traversal, absolute-path escapes, and post-resolution symlink escapes.
- Required the resolved target to be a regular file and under a byte cap (new `HEADROOM_VERIFY_TRUNCATION_MAX_BYTES` knob, default 5MB), both checked via `fstat` on the open fd before any read.
- Opened with `O_NONBLOCK` so a non-regular/special file can't park the read indefinitely; disk I/O already runs off the event loop via the existing compression thread pool.
- Added unit and integration tests covering successful verification, missing/unreadable paths, relative paths, content races, feature gating, banner precedence, and the adversarial cases below.

## Testing

### Test Output

```text
$ pytest tests/test_tool_result_interceptors.py -v
51 passed in 2.21s
(includes TestVerifyReadAgainstDisk::* [14 cases, incl. absolute-path escape, `..`
traversal, symlink escape, sibling-prefix boundary, non-regular file, oversized
file, and untrusted-request-touches-no-disk], plus the end-to-end disk-verification
and banner-precedence cases)

$ pytest tests/test_proxy_loopback_gating.py -k trusted
16 passed
(includes test_record_headroom_stack_binds_trusted_true_for_loopback_client and
..._false_for_non_loopback_client — proves a spoofed x-headroom-cwd from a
non-loopback peer does not bind trust)

$ ruff check headroom/proxy/interceptors/astgrep.py headroom/proxy/project_context.py \
    headroom/proxy/server.py headroom/proxy/runtime_env.py \
    tests/test_tool_result_interceptors.py tests/test_proxy_loopback_gating.py
All checks passed!

$ mypy headroom/proxy/interceptors/astgrep.py headroom/proxy/project_context.py \
    headroom/proxy/server.py headroom/proxy/runtime_env.py
```

## Real Behavior Proof

- Environment: Mac OS local dev machine, no live proxy/network involved. Exercised the real interceptor pipeline (`apply_to_messages`) directly against an actual file on disk, using the same fixture and cwd-binding path a live proxy request would use.
- Exact command / steps: wrote a 9-function Python fixture to a temp file, built a client-truncated `tool_result` (first 8 functions, cut mid-file, **no truncation banner text anywhere**), bound `x-headroom-cwd` to the temp directory via `set_current_cwd()`, then called `apply_to_messages()` once with `HEADROOM_VERIFY_TRUNCATION_ON_DISK` unset and once with it set to `1`.
- Observed result: With disk verification disabled, the existing ast-grep interceptor produced its normal header without indicating that the input was truncated. With HEADROOM_VERIFY_TRUNCATION_ON_DISK=1, disk verification detected that the 50-line tool_result was a strict prefix of the 59-line file and produced:
 ```text
=== HEADROOM_VERIFY_TRUNCATION_ON_DISK unset (default) ===
[headroom: outlined by ast-grep — 8 definition(s); bodies elided. Re-read the file with a line range to see a specific body.]

=== HEADROOM_VERIFY_TRUNCATION_ON_DISK=1, x-headroom-cwd bound to the real directory ===
[headroom: outlined by ast-grep — 8 definition(s) in the visible portion; input was truncated upstream (showing through line 50 of 59 total). Bodies elided. Re-read remaining lines to see more.]
```

**Trust-gate behavior:** same fixture and disk layout, `HEADROOM_VERIFY_TRUNCATION_ON_DISK=1`, but the request context's trust flag left at its default (`False`, simulating a non-loopback caller) instead of being marked trusted.

```text
=== trusted context (loopback) ===
[headroom: outlined by ast-grep — 8 definition(s) in the visible portion; input was truncated upstream ...]

=== untrusted context (non-loopback) — same file, same cwd header, same tool_result ===
[headroom: outlined by ast-grep — 8 definition(s); bodies elided. Re-read the file with a line range to see a specific body.]
(no disk read attempted — verified via a call-recording wrapper around os.open asserting zero invocations)
```
Also verified at the HTTP-middleware level with `TestClient(..., client=("100.90.0.5", 12345))`: a request carrying a spoofed `x-headroom-cwd` header from a non-loopback client IP does not bind trust, regardless of header content.

- Not tested: a live `headroom proxy` session with a real coding-agent client; the WebSocket/Codex responses-WS path.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. The feature is gated by `HEADROOM_VERIFY_TRUNCATION_ON_DISK`
- Minimum rollout channel: N/A. `--intercept-tool-results` must already be enabled for the ast-grep interceptor to run.
- Stable/default behavior changed: no
- Kill switch / disable path: Unset `HEADROOM_VERIFY_TRUNCATION_ON_DISK`; the setting is hot-reloadable and does not require a restart.
- Qualification impact: none
- Rollback path: revert the commit, or unset the env var on a running proxy (takes effect immediately, no restart).

## Review Readiness

## Checklist

## Additional Notes

this branch is based on `fix/astgrep-truncation-marker` #2430 not main


JerrettDavis (changes requested):
The opt-in verifier introduces a local-file oracle across an untrusted request boundary. Both x-headroom-cwd and the Read tool's file_path come from the HTTP caller, absolute paths are accepted, symlinks are followed, and Path.read_text() reads the target synchronously with no size bound. A caller who can reach the proxy can probe arbitrary proxy-host files with chosen prefixes (the response reveals complete/prefix status and, on a prefix, total line count), or point it at a very large/special file and block the event loop / exhaust memory. The feature flag does not make those inputs trustworthy.

Please constrain verification to a server-established/validated workspace root rather than accepting an arbitrary header as authority; resolve the candidate strictly under that root, reject escapes and symlinks (including post-resolution escapes), require a regular file, and enforce a conservative byte cap before reading. The disk I/O should also be off the async request loop or otherwise bounded. Add adversarial tests for absolute paths outside the workspace, .. traversal, symlink escape, missing/non-regular/oversized files, and an untrusted/non-loopback request. If the proxy cannot establish a trusted workspace for the request, return UNKNOWN without touching disk. This is important because otherwise enabling a correctness fallback materially expands the proxy's host-filesystem attack surface.

slxng1758:
@JerrettDavis thanks for the comment, I have made the changes - please let me know if theres anything else to be done

JerrettDavis (changes requested):
Thank you for closing several parts of the filesystem-oracle finding: remote peers now short-circuit before path handling, containment is segment-aware, file type/size are checked on the opened descriptor, and the adversarial coverage is much stronger. Three security/correctness gaps remain. First, for any loopback request, the caller-supplied x-headroom-cwd is still the trust root; a local/SSRF-capable caller can set it to any absolute directory and then probe files beneath that chosen root. The prior request was for a server-established/registered workspace root, not loopback plus an arbitrary header. If no authoritative workspace binding exists for this route, disk verification should fail closed rather than elevating the header. Second, resolve(strict=True) followed by os.open(resolved) still has a TOCTOU escape: the resolved file or an ancestor can be replaced with a symlink between those calls and open follows it. Use descriptor-relative/no-follow traversal (or a platform-safe equivalent) and verify the opened object remains under the authoritative root. Third, _read_disk_content_bounded performs the regular-file read synchronously; O_NONBLOCK prevents FIFO waits but does not move a multi-megabyte regular-file read off the async request path. Please execute the bounded verification off-loop at the actual interceptor call boundary, or establish with code/tests that this interceptor already runs on a worker. Add a regression proving a loopback request cannot choose an arbitrary root, plus a no-follow/race-resistant open test and an off-loop assertion. The current changes materially improve the patch, but loopback alone is not workspace authorization.

slxng1758:
@JerrettDavis Thanks for the input, I have addressed the remaining 3 gaps in this PR

JerrettDavis (changes requested):
The latest revision resolves the original filesystem-oracle/TOCTOU/off-loop concerns for the path it actually supports: the root is now wrap-registered behind a random live-session token, descriptor-relative O_NOFOLLOW traversal fails closed, and compression executor context is propagated. Two blocking correctness/readiness gaps remain:\n\n1. The committed suite is red on supported Windows. _open_regular_file_under_root() intentionally returns None when dir_fd traversal is unavailable, but the platform-independent positive tests still require COMPLETE/TRUNCATED and a successful read. Running the focused files on Windows produced 5 failures (	est_exact_match_is_complete, 	est_strict_prefix_is_truncated, 	est_relative_path_resolves_against_registered_cwd, 	est_open_regular_file_under_root_proceeds_from_plain_sync_context, and the end-to-end truncation test; 88 passed, 5 skipped). Please make the platform contract and tests agree—either provide a race-safe Windows implementation or explicitly gate positive cases and assert fail-closed behavior there. The PR currently has only template/label checks, so CI does not reveal this.\n\n2. The authoritative workspace binding is wired only into wrap claude: that path alone creates _session_token, records it, and injects X-Headroom-Session-Token. The description motivates this fallback specifically for Cursor and OpenCode, but those wrap/configuration paths never register or transmit a token, so 
esolve_registered_cwd() returns None and disk verification cannot activate for those clients. Please wire the binding through each supported client that is claimed to benefit (with end-to-end tests), or narrow the advertised behavior/scope so the feature does not claim to fix clients it cannot serve.\n\nThese are separate from the prior security concerns, which are materially and thoughtfully addressed in the new design.

JerrettDavis (changes requested):
The two latest commits resolve the prior core findings: Windows now explicitly fails closed to UNKNOWN with platform-aware tests, and OpenCode now mints/registers one session token and forwards that same token through both fetch and Node request transport paths. The authoritative-root/no-follow/off-loop model remains intact.

Two readiness corrections remain:

1. The expanded focused suite is still red on supported Windows: 	ests/test_providers_opencode_config.py::test_build_launch_env_with_project searches the JSON-serialized OPENCODE_CONFIG_CONTENT for raw str(plugin), but JSON correctly doubles Windows backslashes. My run on this exact head was 142 passed, 10 skipped, 1 failed. Please parse the JSON and assert the plugin entry semantically, as the Codex path tests recently had to do for the same reason, and include this OpenCode configuration test in the Windows job so the new cross-platform gate covers the path added here rather than only 	est_tool_result_interceptors.py.

2. The opening description still says the fallback addresses clients such as Cursor and OpenCode, while the new runtime docstring correctly states that Cursor has no header-capable binding and therefore remains UNKNOWN. Please narrow the user-facing claim to OpenCode (and any actually wired client), and list Cursor as unsupported/not tested until it can transmit an authoritative session token.

These are narrow test/description issues; the latest security and transport implementation itself looks coherent on this pass.

slxng1758:
Fixed both:

* `test_build_launch_env_with_project` now parses `OPENCODE_CONFIG_CONTENT` and compares the `plugin` entry semantically instead of substring-matching the raw path against its JSON serialization (which doubles backslashes on Windows), and `windows-tool-result-interceptors` now also runs `test_providers_opencode_config.py`, `test_cli/test_wrap_opencode.py`, and `test_proxy_workspace_registration_middleware.py`, so the gate covers the OpenCode session-token path end-to-end, not just the interceptor.
* PR description now has an explicit "Current scope" note: the session-token binding covers `wrap claude` and plugin-loaded `wrap opencode` only; Cursor is called out as unsupported/not tested pending a safe transport mechanism.

All four files pass locally on macOS (146 passed) — the Windows-specific skip/fail-closed behavior is what the new CI job will confirm on this push. `actionlint` is clean on the updated workflow.

JerrettDavis (approved):
The follow-up resolves both remaining readiness findings. The OpenCode config assertion now compares parsed JSON rather than serialized Windows path text; the Windows workflow covers the interceptor, OpenCode config/CLI, and workspace-registration middleware; and the PR description clearly scopes Cursor out until it has an authoritative header-injection path. Focused Windows validation on this exact head: 137 passed, 9 skipped. The session-token transport and fail-closed Windows design remain coherent.

PR #3201, opened 2026-08-22
fix(wrap): never reuse a proxy with incompatible routing config

### LINKED ISSUE #3198: [BUG] wrap reuses a running proxy with incompatible routing config (backend / upstream URLs are never compared)
## Description

When `headroom wrap <tool> --port P` finds something already listening on P, `_ensure_proxy_unlocked` (`headroom/cli/wrap.py`) decides "reuse vs restart" by comparing only feature flags and a subset of upstream URLs between the requested config and the running proxy's `/health` config:

- compared: `memory`, `learn`, `code_graph`, `savings_profile`, `openai_api_url` (only if requested), `vertex_api_url` (only if requested or cleared), plus `factory_api_url` on the droid topic branch
- NEVER compared: `backend`, `anthropic_api_url`, `cloudcode`/`gemini` URLs, and (on the auggie topic branch) `augment_api_url`

Two asymmetries make this actively dangerous:

1. `backend` is not checked at all in the shared path. `wrap copilot` does its own backend check (wrap.py, the `effective_backend != running_backend` guard) precisely because reusing a wrong-backend proxy breaks auth and routing, but codex/aider/claude/every other wrapped tool has no such guard.
2. The URL checks only fire when the CALLER requests the URL. A proxy started with a non-default upstream (e.g. `--augment-api-url`, `--openai-api-url https://xlb...`) is "compatible" with every plain wrap that requests nothing, so its captive routing is silently inherited by an unrelated session.

## Real-world impact (observed 2026-08-22, full evidence below)

Setup state on the machine:

| Port | Listener | Config (from /health) |
|---|---|---|
| 8787 | PID 57525, v0.37.0-dev, started 07:56 by `install agent run --profile init-auggie-40465f15` | backend=anthropic, mode=token |
| 8788 | PID 73974, v0.34.0-dev, started Aug 1 (uptime 1,844,201 s), zero live clients | backend=anthropic, augment_api_url=https://xlb.api.augmentcode.com |

What happened:

1. `headroom wrap codex --port 8787` reused the init-auggie deployment on 8787 (persistent manifest, version match, no missing feature flags). The session ran through the proxy (WS `/v1/responses`, `tokens_saved=1861`), so 8787 routing itself worked, but the dashboard at http://127.0.0.1:8787/dashboard showed the auggie deployment's identity, which read as "not the proper dashboard" for the codex wrap.
2. After Ctrl+C, `headroom wrap codex --port 8788` reused the 21-day-old AUGMENT proxy: the version check did not fire (separate bug, see related issue) and the config check found nothing "missing" because codex wrap requests no special flags. Codex was launched with `openai_base_url=http://127.0.0.1:8788/p/ai/v1` pointing at a proxy configured for a different product's upstream. The wrap process (PID 3169) spawned no proxy child; the old proxy logged zero requests afterwards (its log handle, proxy.log.5, untouched since the previous day), and codex ended up with a direct connection to chatgpt.com (104.18.32.47:443), i.e. completely outside Headroom while the user believed they were wrapped.

The failure signature is terrible: wrap prints "Proxy already running on port 8788" and everything looks fine, while the session is either misrouted or silently unproxied.

## Expected behavior

A wrap only reuses a running proxy when the routing-relevant config matches in BOTH directions:

- effective backend equal (requested `backend`/`HEADROOM_BACKEND`/default vs `running_config["backend"]`)
- every upstream URL field (`openai_api_url`, `anthropic_api_url`, `vertex_api_url`, `cloudcode_api_url`, `gemini_api_url`, and topic-branch fields `augment_api_url`, `factory_api_url`) equal after `_normalize_proxy_api_url`, where "running has it set, caller did not request it" is a mismatch, not a pass

On mismatch:

- zero live clients attached: kill via the `pid` exposed in `/health` config and restart with the requested config (same mechanics as the existing "missing flags" path)
- live clients attached: do NOT silently reuse (a routing mismatch is broken, not degraded; cf. the factory-api-url branch which already fails loud for exactly this reason). Preferred: fall through to the fresh-start path so `_find_available_port` auto-bumps to a free port, mirroring the precedent in `wrap auggie` ("Port P is serving a different session; using port N for this session"), and print WHY (which keys mismatched).

Implementation note: make the checked URL key list data-driven (a module-level tuple of config keys) and read the running side straight from the `/health` config dict, so a mixed-version fleet (older/newer proxy binaries with fields the local CLI does not know) is still caught. This matters in practice: the proxy that caused this incident was started by a different build than the CLI that reused it.

## Regression tests

`tests/test_cli/test_wrap_persistent.py` / `test_wrap_helpers.py` style, mocking `_check_proxy`/`_query_proxy_health`:

- running backend=anthropic + augment_api_url set, wrap codex (no requests) -> restart (no clients) or port bump (clients attached)
- requested backend=anyllm vs running backend=anthropic -> mismatch
- running openai_api_url set vs plain wrap -> mismatch (today: silently reused)
- exact match on all fields -> reuse (guard the happy path)
- mismatch with `pid` absent from config -> clear error, no silent reuse

## What I did NOT verify

- Why codex 0.149.0 in the second session bypassed the proxy entirely (direct chatgpt.com connection despite `--config openai_base_url=...`). Possibly a codex-side fallback when the base URL misbehaves; needs a separate codex-level repro. The headroom-side fix above removes the precondition regardless.
- Whether the first session's model-quality degradation (tool calls emitted as commentary after a truncated tool result, no fold marker) is caused by the proxy; that session DID flow through 8787 (0.37.0-dev, output shaper on, `HEADROOM_VERBOSITY_LEVEL=4` in runtime_env). Tracked separately under the dedup/fold work (see .overlay/issues/cross-turn-dedup-unrecoverable-pointers-openai-chat.md).

rNoz:
## Description

When `headroom wrap <tool> --port P` finds a proxy already listening, `_ensure_proxy_unlocked` decides reuse-vs-restart by comparing feature flags (`memory`, `learn`, `code_graph`, savings profile) and, only when the caller explicitly requests them, `openai_api_url`/`vertex_api_url`. Two gaps make this silently dangerous:

1. **The backend is never compared** in the shared path (`wrap copilot` has its own guard; every other tool does not).
2. **URL checks are asymmetric**: a proxy started with a non-default upstream (e.g. a Copilot tenant URL, or an augment/factory upstream from a feature-branch build) is "compatible" with every plain wrap that requests nothing, so unrelated sessions silently inherit captive routing.

Observed in production: `headroom wrap codex --port 8788` reused a 21-day-old augment-routed proxy (`backend=anthropic`, `augment_api_url` set, zero clients) because nothing in the reuse check objected; the session ended up outside Headroom while the wrap reported "Proxy already running".

This PR adds `_proxy_routing_mismatches` (symmetric comparison of effective backend, including `HEADROOM_BACKEND`, plus all upstream-URL keys, where "running has it set, caller did not request it" is a mismatch) and, on mismatch:

- **zero live clients**: kill the proxy via the `pid` exposed in `/health` config and restart with the requested config (existing missing-flags mechanics);
- **live clients attached**: never reuse and never disrupt — start a dedicated proxy on the next free port (the `wrap auggie` "different session" precedent);
- **persistent deployment with routing mismatch**: leave the deployment alone (its routing comes from the manifest; killing its proxy from the ephemeral path orphans the manifest) and start a dedicated proxy on a free port;
- **proxy exposes no config block at all**: never reused (it cannot be compatibility-checked and reports no PID to restart with) — dedicated port instead.

Closes #3198

## Type of Change

## Changes Made

- `headroom/cli/wrap.py`: new `_PROXY_ROUTING_URL_KEYS` + `_proxy_routing_mismatches` helper (module-level so tests and future callers share it; URL keys absent from a running proxy's config compare equal to unrequested/`None`, so they are inert, never wrong, against builds that do not expose them). Wired into the persistent-deployment reuse gate (mismatch now diverts to a dedicated port instead of falling through to the ephemeral kill path) and into the ephemeral/shared branch (mismatch restarts when idle, port-bumps when clients are attached; configless proxies are no longer reused).
- `tests/test_cli/test_wrap_persistent.py`: helper unit tests (backend via arg and `HEADROOM_BACKEND`, symmetric URL mismatch both directions, normalization tolerance, missing-key tolerance), restart-on-routing-mismatch, port-bump-with-clients, configless-no-reuse, persistent-divert, happy-path reuse. The pre-existing `..._restarts_persistent_deployment_for_feature_mismatch` test asserted the old kill-and-hijack behavior for a routing mismatch; updated to the divert semantic with an explanatory comment.
- `tests/test_cli/test_wrap_helpers.py`: the dashboard-URL regression test now mocks a config-bearing proxy (configless is covered by the new dedicated-port test).

## Testing

### Test Output

```text
$ uv run pytest tests/test_cli/test_wrap_persistent.py -q
38 passed in 0.45s
$ uv run pytest tests/test_cli/test_wrap_helpers.py tests/test_cli/test_wrap_persistent.py -q
96 passed in 0.56s
$ uv run pytest tests/test_cli -q -k 'wrap or unwrap or proxy'
525 passed, 1 skipped, 182 deselected in 12.27s
$ uv run ruff format headroom/cli/wrap.py tests/test_cli/test_wrap_persistent.py tests/test_cli/test_wrap_helpers.py
3 files left unchanged
$ uv run ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_persistent.py tests/test_cli/test_wrap_helpers.py
All checks passed!
$ uv run mypy headroom/cli/wrap.py
```

## Real Behavior Proof

- Environment: macOS 26.5 arm64, Python 3.13, local source build 0.37.0-dev (integration overlay with this PR applied), live ports: 8789 (plain 0.37.0-dev proxy), 8791 free.
- Exact command / steps: with the plain proxy healthy on 8789, `python -c "from headroom.cli import wrap as w; w._ensure_proxy(8789, False, agent_type='auggie', augment_api_url='https://xlb.api.augmentcode.com')"`, then `GET /health` on the resulting port.
- Observed result: `Proxy on port 8789 has incompatible routing config (augment_api_url); restarting with the requested config...` — the plain proxy was killed and an augment-routed proxy came up (on 8791: the just-freed 8789 lost the bind race and 8790 was held by another stale listener; the fallback messaging named the port change). `/health` on 8791: `version 0.37.0-dev, augment_api_url https://xlb.api.augmentcode.com`. Before this PR the identical call printed "Proxy already running on port 8789" and reused the plain proxy.
- Not tested: the attached-clients port-bump path against real live traffic (unit-tested with mocked markers); the persistent-deployment divert against a real `install agent` deployment (unit-tested); Windows.

## Runtime Rollout Safety

- Stable/default behavior changed: yes — wraps no longer silently reuse proxies whose backend/upstream routing differs, and no longer reuse configless proxies. Previously-correct reuse (identical config) is unchanged and still prints the dashboard URL.
- Kill switch / disable path: `headroom wrap <tool> --no-proxy` bypasses the check entirely (pre-existing).
- Qualification impact: none
- Rollback path: revert the commit; the previous reuse heuristic returns.

## Review Readiness

## Checklist

## Additional Notes

Adversarially reviewed by a second model (glm-5.2, high reasoning); initial verdict needs-work, all accepted findings applied (configless no-reuse, persistent-divert, honest docstring about which URL keys this build's own `/health` exposes) and re-tested green. Companion PRs (same incident): `-dev` version-restart fix, orphan-proxy watchdog. Follow-up for the auggie/droid feature branches: expose `augment_api_url`/`factory_api_url` in their `/health` config payloads so same-build detection works there too (this PR already catches older/foreign builds that expose them).

### DIFF
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -3609,6 +3609,58 @@ def _proxy_needs_version_restart(payload: dict[str, Any] | None) -> bool:
     )
 
 
+# Upstream-URL config keys that decide where a running proxy forwards
+# traffic. A running proxy is only reusable when every key matches the
+# requested value. This build's /health config payload only serializes the
+# first five keys; the augment/factory entries catch proxies built from
+# feature branches or other versions that DO expose them (exactly the
+# mixed-fleet shape of the 2026-08-22 incident, where a 0.34-era proxy
+# reported augment_api_url). Keys absent from the running config simply
+# compare equal to an unrequested (None) value, so they are inert, never
+# wrong, against proxies that do not expose them.
+_PROXY_ROUTING_URL_KEYS = (
+    "openai_api_url",
+    "anthropic_api_url",
+    "vertex_api_url",
+    "cloudcode_api_url",
+    "gemini_api_url",
+    "augment_api_url",
+    "factory_api_url",
+)
+
+
+def _proxy_routing_mismatches(
+    running_config: Mapping[str, Any],
+    *,
+    backend: str | None,
+    requested_api_urls: Mapping[str, str | None] | None = None,
+) -> list[str]:
+    """Return routing-level keys where a running proxy differs from the request.
+
+    Unlike the feature-flag check in ``_ensure_proxy_unlocked`` (memory,
+    learn, code graph), a mismatch here is never "degraded but usable": the
+    proxy would forward this session's traffic to a backend or upstream the
+    caller did not ask for. URL comparison is symmetric: a running proxy with
+    a non-default upstream the caller did not request is a mismatch too.
+    """
+    mismatches: list[str] = []
+    effective_backend = backend or os.environ.get("HEADROOM_BACKEND") or "anthropic"
+    running_backend = running_config.get("backend")
+    if (
+        isinstance(running_backend, str)
+        and running_backend
+        and running_backend != effective_backend
+    ):
+        mismatches.append("backend")
+    requested_api_urls = requested_api_urls or {}
+    for key in _PROXY_ROUTING_URL_KEYS:
+        if _normalize_proxy_api_url(running_config.get(key)) != _normalize_proxy_api_url(
+            requested_api_urls.get(key)
+        ):
+            mismatches.append(key)
+    return mismatches
+
+
 def _detect_running_proxy_backend(port: int) -> str | None:
     """Read the backend of an already-running proxy from its health endpoint."""
     return _copilot_detect_running_proxy_backend(port)
@@ -4070,6 +4122,11 @@ def _ensure_proxy_unlocked(
         or bool(copilot_refresh_oauth_token)
         or copilot_api_token_expires_at is not None
     )
+    # Set True when the proxy on the requested port belongs to a persistent
+    # deployment whose routing config does not match this wrap: the deployment
+    # owns the port and must not be reused, killed, or restarted from the
+    # shared ephemeral path; the wrap starts a dedicated proxy elsewhere.
+    persistent_routing_mismatch = False
     # --no-proxy reuses an already-running proxy, so backend/region/provider
     # flags (which only apply when we start one) would be silently dropped.
     if no_proxy and (backend or anyllm_provider or region):
@@ -4141,12 +4198,35 @@ def _ensure_proxy_unlocked(
                         requested_openai_url = _normalize_proxy_api_url(openai_api_url)
                         if running_openai_url != requested_openai_url:
                             missing.append("openai-api-url")
-                    if not missing:
+                    routing_mismatches = helpers._proxy_routing_mismatches(
+                        running_config,
+                        backend=backend,
+                        requested_api_urls={
+                            "openai_api_url": openai_api_url,
+                            "anthropic_api_url": anthropic_api_url,
+                            "vertex_api_url": (None if clear_vertex_api_url else vertex_api_url),
+                        },
+                    )
+                    if routing_mismatches:
+                        # A persistent deployment's routing comes from its
+                        # manifest; the shared ephemeral path must not kill it
+                        # (the manifest would point at a dead or hijacked
+                        # port). Start this wrap's proxy on a free port and
+                        # leave the deployment running.
+                        keys_str = ", ".join(routing_mismatches)
+                        click.echo(
+                            f"  Persistent deployment '{manifest.profile}' on port {port} "
+                            f"has incompatible routing config ({keys_str}); "
+                            "starting a dedicated proxy on a different port instead."
+                        )
+                        persistent_routing_mismatch = True
+                    elif not missing:
                         click.echo(f"  Proxy already running on port {port}")
                         click.echo(f"  Dashboard:    http://127.0.0.1:{port}/dashboard")
                         return None, port
-                # Features mismatch or config unavailable — fall through to
-                # the non-persistent path which handles proxy restart.
+                # Features mismatch or config unavailable — fall through to the
+                # non-persistent path which handles proxy restart. A routing
+                # mismatch skips that path entirely (see flag above).
             else:
                 if helpers._recover_persistent_proxy(port):
                     # If the caller requested feature-sensitive config (e.g.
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
The compatibility decision is now symmetric for backend and exposed upstream URLs, so a caller cannot silently inherit captive routing from an otherwise healthy shared proxy. The lifecycle branches are appropriately conservative: idle ephemeral owners are identity-stopped and replaced; active or persistent owners are left intact and diverted to a dedicated port; configless owners are never guessed compatible. Tests cover matching reuse, both mismatch directions, attached clients, persistent ownership, absent config/PID, and failed termination. The branch is clean and the full cross-platform CI matrix is green.

Non-blocking hardening: parse a health-reported PID defensively before `int(proxy_pid)` so a malformed/legacy value becomes the same actionable “no usable PID” Click error rather than leaking `ValueError`.

JerrettDavis (approved):
Re-reviewed the routing compatibility changes. Backend and upstream comparisons are symmetric, environment-derived effective routing is accounted for, persistent proxies are preserved on routing mismatch, active shared proxies are diverted rather than disrupted, and unverified config is not silently reused. Focused coverage and the full suite are green. No blocking issues found.

PR #3202, opened 2026-08-22
fix(proxy): reap wrap-spawned proxies once no wrap clients remain

### LINKED ISSUE #3199: [BUG] wrap-spawned proxies are never reaped once clientless: no reaper for unclean exits, last graceful exit leaks too
## Description

"Why are there stale proxies when I have no running sessions?" Because nothing in Headroom ever stops a wrap-spawned proxy whose wrappers are gone. There is no reaper, no idle timeout, no orphan watchdog. Observed on 2026-08-22: three `headroom proxy` processes from Aug 1 still alive (PIDs 73974 port 8788, 63454 port 8789, 79339 port 8790, all `--augment-api-url`), 21 days after their last session, holding ports and getting silently reused by later wraps (see related issue for the incident that caused).

Two distinct leak paths, both by-products of deliberate design choices that lack a counterpart:

1. Unclean wrapper death. `_start_proxy` launches the proxy with `start_new_session` (setsid) on POSIX (and breakaway flags on Windows) ON PURPOSE, so that a crashed/killed wrapper cannot tree-kill a shared proxy out from under other clients (`_make_cleanup` docstring). But when the wrapper dies without running cleanup (SIGKILL, crash, terminal close, laptop power loss), its marker in `~/.headroom/clients/<port>/<pid>.json` goes stale, `_live_proxy_clients` unlinks stale markers on the next visit, and the proxy itself is simply never stopped.
2. Graceful last-exit leak. `_make_cleanup.cleanup` terminates the proxy only when the exiting wrapper is the one that STARTED it (`proxy_holder[0] is not None`). When the starter exits while other clients remain (correctly leaving the proxy up), every later wrapper exits with `holder is None`, so the last graceful exit still leaves the proxy running. Even a perfectly behaved multi-wrap topology leaks the proxy at the end.

Confirmed live on 2026-08-22: `~/.headroom/clients/8788/` contained `73826.json` for a long-dead PID (the Aug 1 wrapper) next to the current wrap's `3169.json`; the proxy (PID 73974) was reparented to init and healthy after 21 days of doing nothing.

## Expected behavior

A wrap-spawned proxy with zero live clients should eventually stop itself, without endangering:

- standalone `headroom proxy` instances (no wrap markers exist by design; must never self-exit)
- persistent installs (`install agent run`, supervisor-owned deployments like the `init-*` profiles; must never self-exit)
- active sessions (a codex WS relay or a long anthropic stream must not be cut because markers briefly look empty)

## Suggested fix: orphan watchdog in wrap-spawned proxies

- `_start_proxy` (wrap.py) sets a new env marker, e.g. `HEADROOM_WRAP_OWNED=1` (plus optional `HEADROOM_WRAP_ORPHAN_GRACE_SECONDS`, default ~900 s). Only wrap-spawned proxies get it; standalone and persistent-run launchers do not.
- The proxy server, when the marker is set, spawns an asyncio watchdog task: every ~30 s, list `paths.proxy_clients_dir(port)`, keep markers whose PID is alive and not recycled (reuse the `pid_alive` + `_identity_mismatch` logic, currently in `cli/wrap.py`; move to a shared module, e.g. `headroom/utils/proc.py`, so proxy code does not import from the CLI layer), and count own active sessions via the in-process `ws_session_registry` (and any in-flight HTTP gauge, see below).
- When live clients == 0 and active sessions == 0 continuously for the grace period, trigger the server's normal shutdown event (`_get_shutdown_event`). Grace period absorbs the startup window (marker is registered before `_ensure_proxy` runs, so the dir is never empty at boot for well-behaved wraps) and transient crashes mid-request.
- Precedent for self-termination on orphaning already exists: `headroom/ccr/mcp_server.py` watches for reparenting to init and reaps itself.

Cheaper complementary fix in `_make_cleanup`: when the last live client unregisters and its own `proxy_holder[0]` is None (it reused the proxy), attempt the same stop path used for `unwrap` (`_stop_local_proxy_for_unwrap`) instead of doing nothing. This closes leak path 2 immediately but cannot help path 1 (nobody is alive to run cleanup), so the watchdog is still needed.

## Edge cases to test

- wrapper killed with SIGKILL, proxy exits within grace period
- two wrappers share a proxy, both exit gracefully (starter first, then reuser): proxy stops
- persistent deployment / standalone proxy: watchdog inactive, process survives with zero clients
- active WS session with zero live markers (crashed wrapper mid-session): no shutdown while `active_sessions > 0`
- marker dir missing entirely (never registered): treated as zero clients after grace
- recycled PID with matching liveness but different start time: marker treated as stale (covered by existing `_identity_mismatch`)

## What I did NOT verify

- Whether a global in-flight HTTP request gauge exists for the watchdog to consult; if not, the grace period (default 15 min) is the mitigation, since even long streaming requests finish well inside it. WS sessions are covered explicitly via the registry.
- Windows service semantics of the watchdog (the breakaway flags mean the proxy survives console close; the watchdog is what eventually stops it there too).

rNoz:
## Description

Nothing in Headroom ever stops a wrap-spawned proxy whose wrappers are gone. Two leak paths, both by-products of deliberate hardening without a counterpart:

1. **Unclean wrapper death.** `_start_proxy` detaches the proxy (`start_new_session` on POSIX, breakaway console flags on Windows) on purpose, so a crashed wrapper cannot tree-kill a shared proxy out from under other clients. But when the wrapper dies without running cleanup (SIGKILL, crash, terminal close), its client marker goes stale, `_live_proxy_clients` unlinks it, and the proxy itself is never stopped.
2. **Graceful last-exit leak.** `_make_cleanup` only terminates the proxy when the exiting wrapper *started* it (`proxy_holder[0] is not None`). When the starter exits while other clients remain (correct), every later wrapper exits with no handle, so even a perfectly behaved multi-wrap topology leaks the proxy at the end.

Observed in production: three `headroom proxy` processes still alive 21 days after their last session, holding ports and being silently reused by later wraps (which then inherited their stale version and captive routing; see the two companion PRs).

The fix: wrap-spawned proxies carry `HEADROOM_WRAP_OWNED=1` and run an orphan-watchdog task that polls the per-port client-marker dir every 30s and gracefully stops the process once no live markers, no active WebSocket sessions, and no served-request activity remain for a grace period (default 900s, floor 60s, `HEADROOM_WRAP_ORPHAN_GRACE_SECONDS`). Standalone `headroom proxy` instances and persistent installs never get the marker env var and are never affected; multi-worker runs are excluded (WS sessions are per-worker, and wrap never requests multiple workers).

Closes #3199

## Type of Change

## Changes Made

- `headroom/proxy/orphan_watchdog.py` (new): env config (`WRAP_OWNED_ENV`, grace parsing with floor), marker liveness (`pid_alive` + start-time identity, pruning stale markers), and `orphan_watchdog_loop` (idle clock reset by live markers, active WS sessions, or movement of the proxy's served-request counter — the last covers direct-HTTP and SSH-forwarded clients that write no marker). Stop is `signal.raise_signal(SIGTERM)`, which reaches uvicorn's handler on Unix and the emulated handler table on Windows (unlike `os.kill(pid, SIGTERM)`, which is `TerminateProcess` there).
- `headroom/proxy/server.py`: lifespan spawns the watchdog when `orphan_watchdog_enabled()` and not a multi-worker run (`HEADROOM_PROXY_CONFIG_JSON` absent), and cancels it in teardown like the other periodic tasks.
- `headroom/cli/wrap.py`: `_start_proxy` sets `HEADROOM_WRAP_OWNED=1` in the proxy subprocess env; `_proc_identity`/`_identity_mismatch` moved to `headroom/_subprocess.py` (`proc_identity`/`identity_mismatch`) so proxy code does not import the CLI layer — thin module-level delegates keep the existing monkeypatch-based tests working unchanged.
- `tests/test_proxy/test_orphan_watchdog.py` (new, 13 tests): env parsing (default/override/floor/garbage), marker pruning (live, dead PID, recycled PID, garbage JSON, missing dir), loop semantics (exit after grace, hold on live client, hold on active WS session, hold on served-request movement, grace reset on returning client, missing registry counts as idle).
- `tests/test_cli/test_wrap_claude_vertex_proxy_env.py`: `_start_proxy` marks the subprocess env wrap-owned.

## Testing

### Test Output

```text
$ uv run pytest tests/test_proxy/test_orphan_watchdog.py -q
13 passed in 2.77s
$ uv run pytest tests/test_proxy tests/test_cli -q
973 passed, 1 skipped in 49.85s   # 1 pre-existing unrelated failure
                                  # (test_litellm_vertex_backend_path_...) fails
                                  # identically on clean upstream/main
$ uv run ruff format / ruff check on touched files
All checks passed!
$ uv run mypy headroom/_subprocess.py headroom/cli/wrap.py headroom/proxy/orphan_watchdog.py
```

## Real Behavior Proof

- Environment: macOS 26.5 arm64, Python 3.13, local source build 0.37.0-dev (integration overlay with this PR applied).
- Exact command / steps: start wrap-spawned proxies via `wrap._ensure_proxy(...)` (which goes through `_start_proxy` and thus `HEADROOM_WRAP_OWNED=1`), with zero registered clients; wait; inspect `~/.headroom/logs/proxy.log`.
- Observed result: two such proxies (plain on 8789 PID 48365, augment-mode on 8791 PID 49729) self-exited at the 900s default grace: `event=proxy_shutdown reason=signal pid=49729` at 09:51:49 against a ~09:36 start, port released afterwards (`lsof` clean). A machine-wide sweep beforehand showed three 21-day-old orphans (PIDs 73974, 63454, 79339); with this build they can no longer accumulate.
- Not tested: Windows signal path (reasoned from CPython/uvicorn semantics; unit-injected `stop` callable covers the loop side); the multi-worker exclusion with a real `--workers 2` run (guarded by env presence only); marker pruning under active PID recycling on Linux without psutil (`/proc` fallback is unit-tested via mocks).

## Runtime Rollout Safety

- Stable/default behavior changed: only for wrap-spawned proxies, which now self-exit after 15 clientless minutes. Standalone and persistent proxies are untouched.
- Kill switch / disable path: omit `HEADROOM_WRAP_OWNED` (not settable by users of wrap; standalone proxies never have it), or set `HEADROOM_WRAP_ORPHAN_GRACE_SECONDS` very high.
- Qualification impact: none
- Rollback path: revert the commit; orphaned proxies leak again (status quo ante).

## Review Readiness

## Checklist

## Additional Notes

Adversarially reviewed by a second model (glm-5.2, high reasoning); initial verdict needs-work, all accepted findings applied (multi-worker exclusion, Windows-safe stop signal, served-request activity hold for marker-less clients, wider test timing margins) and re-tested green. The reviewer independently verified the boot-race ordering, teardown cancellation, and the kill-vs-self-exit race as already correct. Companion PRs (same incident): `-dev` version-restart fix, routing-config reuse check.

### DIFF
--- a/headroom/_subprocess.py
+++ b/headroom/_subprocess.py
@@ -43,6 +43,62 @@ def pid_alive(pid: int) -> bool:
     return True
 
 
+def proc_identity(pid: int) -> tuple[str, float] | None:
+    """Best-effort ``(source, start_time)`` identity for a PID.
+
+    Used to defeat PID reuse: a marker is only trusted while the live PID is
+    *the same process* that wrote it. Returns ``None`` when start time can't be
+    determined (e.g. macOS without psutil), in which case callers fall back to
+    existence-only liveness, no regression, just no reuse protection there.
+
+    The ``source`` tag ("psutil" vs "proc") guards against comparing values in
+    different units; we only compare like-for-like.
+    """
+    try:
+        import psutil  # type: ignore[import-untyped]  # optional dependency; portable when present
+
+        return ("psutil", psutil.Process(pid).create_time())
+    except Exception:
+        pass
+    # Linux fallback: field 22 of /proc/<pid>/stat is starttime in clock ticks
+    # since boot, a stable per-process value. `comm` (field 2) may contain
+    # spaces/parens, so split after the final ')'.
+    try:
+        with open(f"/proc/{pid}/stat", "rb") as fh:
+            fields = fh.read().rpartition(b")")[2].split()
+        return ("proc", float(fields[19]))
+    except (OSError, IndexError, ValueError):
+        return None
+
+
+def identity_mismatch(
+    src: Any,
+    recorded: Any,
+    pid: int,
+    *,
+    identity_fn: Any = None,
+) -> bool:
+    """True only if ``pid``'s current identity *provably* differs from the
+    recorded ``(src, recorded)`` identity (i.e. the PID was recycled).
+
+    Conservative by design: any uncertainty (unknown/legacy identity, unknown
+    start time, mismatched source) returns ``False``, never claim a mismatch
+    without proof, since the caller uses this to decide whether to trust or
+    discard state tied to a live PID. ``identity_fn`` defaults to
+    :func:`proc_identity` and exists so callers with their own (mockable)
+    identity source can share this comparison.
+    """
+    if identity_fn is None:
+        identity_fn = proc_identity
+    if not isinstance(src, str) or not isinstance(recorded, int | float):
+        return False  # legacy / identity-less record, can't tell
+    ident = identity_fn(pid)
+    if ident is None or ident[0] != src:
+        return False  # can't compare like-for-like, don't claim mismatch
+    # Start times are stable per process; >1s apart means a different process.
+    return bool(abs(ident[1] - float(recorded)) > 1.0)
+
+
 def run(*args: Any, **kwargs: Any) -> _sp.CompletedProcess:
     if kwargs.get("text") or kwargs.get("universal_newlines"):
         kwargs.setdefault("encoding", "utf-8")
--- a/headroom/_subprocess.py
+++ b/headroom/_subprocess.py
@@ -38,7 +38,7 @@ def pid_alive(pid: int) -> bool:
         os.kill(pid, 0)
     except PermissionError:
         return True
-    except (ProcessLookupError, OSError, SystemError):
+    except (ProcessLookupError, OSError, SystemError, OverflowError, ValueError):
         return False
     return True
 
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -703,6 +703,10 @@ def _start_proxy(
     # Ensure proxy subprocess uses UTF-8 (Windows defaults to cp1252)
     proxy_env = os.environ.copy()
     _scrub_copilot_proxy_seed_env(proxy_env)
+    # A wrap-owned proxy must be a single authoritative process so its orphan
+    # watchdog can observe all activity before deciding to self-terminate.
+    proxy_env.pop("HEADROOM_WORKERS", None)
+    proxy_env.pop("HEADROOM_PROXY_CONFIG_JSON", None)
     proxy_env["PYTHONIOENCODING"] = "utf-8"
     # `python -m headroom.cli` prepends the launch cwd to sys.path, so running
     # `wrap` from a directory that contains a `headroom/` folder (most commonly a
@@ -4547,6 +4551,8 @@ def _marker_pid_reused(marker: Path, pid: int) -> bool:
         rec = json.loads(_read_text(marker))
     except (OSError, ValueError):
         return False
+    if not isinstance(rec, dict):
+        return False
     return _identity_mismatch(rec.get("start_src"), rec.get("start_time"), pid)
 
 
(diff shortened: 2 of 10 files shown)


JerrettDavis (changes requested):
The detached-owner marker and PID-reuse handling are thoughtful, but the shutdown predicate is not yet safe for active traffic.

1. `_served_request_count` detects only counter *movement*. A direct HTTP streaming/slow request increments `_request_counter` when its request id is allocated, then the value remains stable for the lifetime of that request. If it runs longer than the grace window (large streamed generation, slow upstream, long upload/forward), the watchdog sees no marker, no WebSocket session, and no later movement, then sends SIGTERM mid-request. The comment assumes streams finish within 15 minutes, but shutdown correctness cannot depend on that latency assumption. Please gate on an actual active HTTP/in-flight request count (increment/decrement in a `finally` at the request boundary), while retaining counter movement as recent-activity evidence if useful. Add a test holding one HTTP request active beyond grace and assert no stop until it completes plus a full grace interval.

2. `_active_session_count` maps registry exceptions to `0`. For a task whose action is process termination, inability to determine session activity must fail closed, not be treated as proof of idleness. Return an unknown state and suppress/reset shutdown until the registry is observable again; add recovery coverage.

The watchdog should stop only when absence of every supported activity signal is positively established. Current CI is green, and the wrap-only/multi-worker scoping plus identity-aware marker pruning otherwise look solid.

JerrettDavis (approved):
Re-reviewed the watchdog hardening after the prior blockers. The outer raw ASGI middleware now tracks HTTP streaming bodies and WebSocket scopes through their complete lifetime, entry/completion generation changes restart the full grace period, and client/session/request/generation observation failures all fail closed by resetting the idle window. Ownership, rollback, PID-reuse, and single-worker wiring have focused coverage; the full suite is green. Prior concerns are resolved.

PR #3209, opened 2026-08-22
fix(output-savings): seed the holdout key on the whole first user message

gglucass:
## Description

`conversation_key_from_body` seeds the output-shaping holdout assignment. It hashed the first 512 characters of the **first** text block of the first user message.

Agent clients open every conversation with injected context - project instructions, IDE state, memory digests - that is byte-identical across conversations in a project and routinely kilobytes long. The user's own words, the only part that distinguishes one conversation from the next, sat past the 512-character cut, so whole populations of conversations shared a single key.

Assignment is deterministic, so one key means one arm permanently. On a real 78k-request ledger, a holdout nominally set to 3% had produced:

| family | treatment | control |
|---|---|---|
| fable | 22,222 | 1 |
| sonnet | 15,401 | 0 |
| haiku | 523 | 0 |
| opus | 38,891 | 3,532 |

Those clients were not being sampled at 3%. They were frozen into one arm, so their output-shaping savings could never be measured against a control no matter how long the holdout ran - and the aggregate measured estimate, had it ever passed its coverage gate, would have compared one frozen population against another rather than a randomized contrast.

No tracking issue; this was found from production ledger data.

## Type of Change

## Changes Made

- `conversation_key_from_body` now seeds on every text block of the first user message, joined and in full, instead of a 512-character prefix of the first block.
- Added `_MAX_SEED_CHARS = 65_536` as a cost guard only, documented as such and set far above any client prologue so it cannot recreate the bug it replaces.
- Non-text blocks (images) still never seed the key.
- Added two regression tests: two conversations differing only past a long identical prologue, and an image block leaving the key unchanged.

Stability is preserved, which the design depends on: user turns are never compressed (`protected_user_message` in `compression_units`), so the first user message is immutable for the life of a conversation, and identical openers still map to one key and one arm.

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/ -q -k "output_sav or shaper or holdout or telemetry_session"
158 passed, 8 skipped, 11927 deselected, 1 warning in 50.48s

$ uvx ruff check .
All checks passed!

$ uvx ruff format --check headroom/proxy/output_savings_policy.py tests/test_output_savings_policy.py
2 files already formatted

$ uv run --frozen --extra dev mypy headroom/proxy/output_savings_policy.py
```

## Real Behavior Proof

- Environment: macOS 15.6, Python 3.11, headroom at 91186b40 (0.36.4), replayed against 170 real Claude Code conversation openers taken from 7 days of local transcripts.
- Exact command / steps: loaded `output_savings_policy.py` from this branch alongside an inline copy of the previous implementation, built a body per opener as `[{"type": "text", "text": prologue}, {"type": "text", "text": opener}]` with a 2,113-character injected prologue, then counted distinct keys and `assign_arm(key, 0.03)` outcomes under both versions.
- Observed result: before, 170 conversations collapsed to 3 distinct keys (one per model family) and 0 of 170 landed in control; after, the same 170 produced 120 distinct keys with 4 in control (2.4%), and both arms were reachable for every model family. The 120-rather-than-170 figure is repeated openers, which correctly share a key. The before-figures reproduce the production ledger in the description, where whole families sit at 0 or 1 control samples.
- Not tested: no live traffic was routed through a proxy built from this branch; the OpenAI Responses path is unchanged and still keys off `_stable_response_identifier`; single-turn templated side-calls whose entire first message is a fixed template remain one key by construction, which this change does not address.

## Runtime Rollout Safety

- Rollout-managed feature(s): output shaper holdout arm assignment, reached only when `HEADROOM_OUTPUT_SHAPER` is enabled and `HEADROOM_OUTPUT_HOLDOUT` is above zero.
- Stable/default behavior changed: Yes. Conversations hash to different keys than before, so a given conversation may be assigned the opposite arm from the one it had under the previous version.
- Kill switch / disable path: set `HEADROOM_OUTPUT_HOLDOUT=0`, which routes every conversation to treatment and bypasses the key entirely; `HEADROOM_OUTPUT_SHAPER=0` disables the surrounding feature.
- Qualification impact: ledger accumulators recorded before this change stay in whatever arm the old key assigned, so a stratum can hold pre- and post-change samples from different populations; expect measured estimates to move as the newly reachable arms fill in.
- Rollback path: revert this commit. The key reverts with it, no persisted format or stored data changes, and existing ledger entries remain readable either way.

## Review Readiness

## Checklist

Not applicable; no user-visible surface changes.

## Additional Notes

Documentation checkbox is unchecked because no docs describe the holdout key; the rationale lives in the function's docstring, which this PR expands.

Two related gaps are deliberately out of scope. Single-turn templated side-calls, where the entire first message is a fixed template, still collapse to one key and stay unmeasurable. And a verbosity baseline seeded once at install cannot be relearned from shaped traffic, so the control arm this fix unblocks is the only route to a fresh comparator.


JerrettDavis (changes requested):
The change moves the collision boundary, but the new 65,536-character prefix retains the same structural failure mode: clients with a shared injected prologue at or above that size still collapse every conversation into one arm. That is realistic with large repository instructions/memory in modern long-context clients, and it conflicts with the docstring/PR claim that every text block feeds the seed 'in full.' The implementation also already receives the complete parsed request, so truncating before a cryptographic digest does not provide a meaningful parsing or memory bound; it only discards entropy.

Please derive the key from the complete first-user text (ideally incrementally hashing block boundaries and all block contents, so no large concatenated seed is needed). If an explicit CPU guard is required, preserve entropy from the entire message—for example, a documented sampled/prefix+suffix scheme—rather than recreating a deterministic all-prefix cutoff. Add a regression with an identical prologue longer than the chosen guard and distinct user text after it.

gglucass:
Addressed at the current head (`3ef5e20f`), which replaces the guard entirely rather than moving it:

- Change: `conversation_key_from_body` and `conversation_key_from_responses_body` now build the key with an incremental SHA-256 - the model, every block boundary, and every text block's full contents are fed to the digest field by field (`_absorb` / `_absorb_text_blocks`). There is no concatenated seed string and no character cutoff of any kind, so no prologue length can recreate the all-prefix collapse.
- Memory bound: constant - the digest consumes each already-parsed block in place, which is the point you made about truncation-before-hashing discarding entropy without bounding anything.
- Regression: `test_conversation_key_separates_conversations_behind_a_64kb_prologue` uses a 96,000-character identical prologue - deliberately larger than the 65,536-character guard from the previous revision - followed by distinct user text, and asserts the keys differ while identical openers still key together. `test_conversation_key_survives_a_long_identical_client_prologue` covers the realistic multi-block injected-context shape, and `test_responses_conversation_key_separates_past_a_long_prologue` covers the Responses path the same way.
- Verified: `pytest tests/test_output_savings_policy.py` - 11 passed; `ruff check` clean.

The red `test-agno` check is unrelated to this branch: agno 3.0.0 (2026-08-24) removed `agno.models.metrics` and the job installs latest agno, so it fails on main and every open PR. Fix in #3260; once that merges, a refresh here should go fully green.

gglucass:
Refreshed from main: `test-agno` is green again (#3260 merged), and the branch changes are untouched. The remaining red `lint` check is also inherited from main, not from this diff: #3260 and #3246 overlapped on the agno 3.0.0 compat fix, leaving a dead module-level import that fails F401 on every run including main's own. Cleanup in #3268 - one more refresh after that merges and this should be fully green.

JerrettDavis (approved):
Re-reviewed the post-feedback implementation at exact head 4c75bf3c. Both Messages and Responses paths now absorb every text block of the first user turn without a cutoff or large concatenated seed; the field separators preserve the previous digest for inputs below the former guard, image-only/non-user shapes remain deterministic, and the 96KB regressions directly cover the prior structural failure. Updating from current main clears the inherited lint issue; newly triggered CI remains the readiness gate. Approval restored for final human review.

JerrettDavis (approved):
Re-reviewed at 8f700f290e826fccdfb9130edb97930acf43b91c. The production policy is unchanged from my previous approval; the added regression verifies that plain-string Responses input distinguishes conversations and agrees with the equivalent user-item form. All 12 output-savings policy tests pass locally, and the current checks are green. Approval retained for human review.

PR #3211, opened 2026-08-22
fix(memory): tolerate None message content in inline memory parser

abhay-codes07:
## Description

`InlineMemoryWrapper.chat` / `.chat_with_response` read `response.choices[0].message.content` and hand it straight to `parse_response_with_memory`, which immediately does `re.search(pattern, response_text, ...)`:

```python
raw_content = response.choices[0].message.content   # str | None
parsed = parse_response_with_memory(raw_content)
...
match = re.search(memory_pattern, response_text, re.DOTALL | re.IGNORECASE)
```

A chat completion's `message.content` is `None` whenever the model answers with tool/function calls instead of text. `re.search` on `None` raises `TypeError: expected string or bytes-like object, got 'NoneType'`, crashing the whole request. So enabling tools on any request routed through the inline memory wrapper turns the first tool-call turn into an unhandled exception.

This is the same robustness class the function already guards for the `<memory>` block payload (a non-object / non-list JSON is tolerated at `inline_extractor.py:132`-`:143`); `None` content was the remaining gap, one level up.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/memory/inline_extractor.py`: `parse_response_with_memory` now returns early when `response_text` is not a `str`, preserving the payload unchanged (so a `None` tool-call turn is not coerced to text) with no memories extracted. Widened its parameter to `str | None` and `ParsedResponse.content` / `.raw` and the two `chat*` return types to `str | None` to match what a chat completion actually yields.
- `tests/test_memory_wrapper.py`: added `test_parse_response_with_memory_tolerates_none_content` next to the existing non-object-block regression.

## Testing

### Test Output

```text
tests/test_memory_wrapper.py  ->  7 passed in 0.78s
uvx ruff@0.16.2 check headroom/memory/inline_extractor.py tests/test_memory_wrapper.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/memory/inline_extractor.py  ->  Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: before the fix, `parse_response_with_memory(None)` raised `TypeError: expected string or bytes-like object, got 'NoneType'`. After the fix, the same call returns `ParsedResponse(content=None, memories=[], raw=None)`, and a normal text response with a `<memory>` block still parses (`content == "hello  world"`, memories extracted).
- Observed result: `None` content degrades gracefully and stays `None`; the string path is byte-for-byte unchanged.
- Not tested: no live `chat.completions.create` tool-call round trip (the wrapper takes an injected client); the `None` content a tool-call turn produces is reproduced directly.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. `InlineMemoryWrapper` is an SDK-side convenience wrapper (`headroom.memory`), not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: no. Non-`None` string inputs take exactly the previous path; only the previously-crashing `None` case changes, from `TypeError` to a graceful no-op parse.
- Qualification impact: none; text responses are handled identically.
- Rollback path: revert this PR; the parser returns to assuming a `str`.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/memory/inline_extractor.py
+++ b/headroom/memory/inline_extractor.py
@@ -56,9 +56,9 @@ Skip for greetings/small talk. If nothing: <memory>{"memories": []}</memory>"""
 class ParsedResponse:
     """Response with extracted memories."""
 
-    content: str  # The actual response (without memory block)
+    content: str | None  # The actual response (without memory block); None for tool-call turns
     memories: list[dict[str, Any]]  # Extracted memories
-    raw: str  # Original full response
+    raw: str | None  # Original full response
 
 
 def inject_memory_instruction(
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
The fix preserves the SDK's actual message.content contract: tool-call turns remain None, memory extraction becomes a no-op, and text responses retain the existing parser path. The widened result/return annotations are consistent with both call sites, the focused regression covers the prior exception, and the current head is fully green. I found no blocking issue.

PR #406, opened 2026-05-06
fix(proxy): hot-fix Codex /v1/responses compression via PyO3 inline call

chopratejas:
## Summary

Restores `/v1/responses` request-side compression that has been silently OFF since **v0.20.16** (PR-C5, 2026-05-03). Closes the regression Codex subscription users have been reporting.

PR-C5 retired the Python `/v1/responses` compression pipeline with the comment *"Rust handles item-aware compression natively (see `crates/headroom-proxy/src/handlers/responses.rs`)"* — but the standalone `headroom-proxy` Rust binary that was supposed to do that compression **is not deployed by the CLI**:

- `headroom proxy` runs only the Python proxy via `uvicorn`.
- `headroom wrap codex` spawns `python -m headroom.cli proxy` (`headroom/cli/wrap.py:139`).
- The `headroom-proxy` Rust binary is a workspace member but never built into the wheel; `which headroom-proxy` returns nothing.
- PR-A0's "rust core deployment smoke test" only verifies `headroom._core` (the PyO3 compression library), not the standalone proxy binary.

Result: every `/v1/responses` request between v0.20.16 and now has been forwarded to `api.openai.com` / `chatgpt.com` uncompressed.

## Why PyO3 inline (Layer 1) and not the originally-intended two-process chain (Layer 2)

The inline call requires zero deployment changes — the wheel already ships `headroom._core` and PR-A0 verifies it loads at proxy startup. The two-process chain (build + ship the standalone binary, teach CLI to spawn both processes, add an internal port for Python, configure Rust's `--upstream` to point at Python) is the right long-term move but not a hot-fix.

This PR exposes the existing `headroom_core::transforms::compress_openai_responses_live_zone` function as a PyO3 binding and calls it inline from the Python `handle_openai_responses` handler. **Same dispatcher code that the standalone binary would have run, just invoked in-process.**

## Commits

| # | Title |
|---|---|
| c1 | `fix(core): expose compress_openai_responses_live_zone via PyO3` — adds the binding to `crates/headroom-py/src/lib.rs` + 14-test contract in `tests/test_responses_pyo3_compression.py` |
| c2 | `fix(proxy): wire PyO3 compression into /v1/responses handler` — calls the binding from `headroom/proxy/handlers/openai.py::handle_openai_responses` right after upstream-URL resolution |

## Behavior

| Path | Pre-PR | Post-PR |
|---|---|---|
| `/v1/responses` HTTP non-stream | uncompressed | live-zone compressed |
[...]

PR #3212, opened 2026-08-22
fix(codex): detect ChatGPT auth from id_token claims so wrap/init emit requires_openai_auth

### LINKED ISSUE #3206: [BUG] wrap/init codex omit requires_openai_auth — Codex sends no Authorization, every request 401s "Missing bearer"
## Summary

`headroom wrap codex` / `headroom init codex` write a `[model_providers.headroom]` block into `~/.codex/config.toml` that omits **`requires_openai_auth = true`**. Codex therefore attaches **no `Authorization` header** to that custom provider, and every request through the proxy fails:

```
unexpected status 401 Unauthorized: Missing bearer or basic authentication in header,
url: http://127.0.0.1:8787/v1/responses, cf-ray: a2f18ee25e077b0b-EWR
```

The proxy is not at fault on the wire — it forwards whatever auth it receives. Codex never sends any. Net effect for a ChatGPT-subscription user: Headroom is 100% unusable with Codex, silently, and `headroom doctor` reports everything green.

This cost me ~15h of dead automation before I bisected it (two scheduled Codex loops, 100% failure, no visible signal other than the 401 buried in the rollout log).

## Environment

| | |
|---|---|
| headroom | 0.35.0 **and** 0.36.4 (reproduced on both) |
| codex-cli | 0.149.0 |
| auth mode | `chatgpt` (subscription OAuth, no `OPENAI_API_KEY`) |
| OS | macOS 15.6 (darwin 25.6.0), arm64 |
| proxy cmd | `headroom proxy --host 127.0.0.1 --port 8787 --mode cache --backend anthropic --no-telemetry` |

## Reproduce

```bash
headroom proxy --port 8798 --mode cache --backend anthropic --no-telemetry &

# FAILS — 401 "Missing bearer"
codex exec -c model_provider=hdbg \
  -c 'model_providers.hdbg.name=dbg' \
  -c 'model_providers.hdbg.base_url=http://127.0.0.1:8798/v1' \
  -c 'model_providers.hdbg.wire_api=responses' \
  --skip-git-repo-check "reply exactly: DBG"

# PASSES — identical except one key
codex exec -c model_provider=hdbg \
  -c 'model_providers.hdbg.name=dbg' \
  -c 'model_providers.hdbg.base_url=http://127.0.0.1:8798/v1' \
  -c 'model_providers.hdbg.wire_api=responses' \
  -c 'model_providers.hdbg.requires_openai_auth=true' \
  --skip-git-repo-check "reply exactly: REQAUTH"
```

## Fix

Add one line to the generated provider block:

```toml
[model_providers.headroom]
name = "OpenAI via Headroom proxy"
base_url = "http://127.0.0.1:8787/v1"
supports_websockets = true
requires_openai_auth = true          # <-- missing
env_http_headers = { "X-Headroom-Project" = "HEADROOM_PROJECT" }
```

Applied locally and confirmed: Codex now works **through** the proxy, so the compression/cache savings are retained. (Unwrapping Codex entirely is the other way to clear the 401, but it throws the savings away — that is a workaround, not a fix.)

Worth also emitting the key on the `headroom init codex` path and adding a migration for configs already written without it.

## Ruled out during the bisect

Recording these so nobody re-chases them:

- **Proxy uptime / staleness** — a freshly started proxy fails identically.
- **Headroom version** — reproduced on 0.35.0 and 0.36.4.
- **WebSocket transport** — `supports_websockets = false` still 401s; the HTTP fallback path is not the trigger.
- **Header stripping by the proxy** — `curl` through the proxy with the real ChatGPT JWT and a correctly shaped body (`store:false`, `stream:true`, `input` as a list) returns **200** and a real streamed response. The proxy forwards auth correctly.
- **Token expiry / routing detection** — `~/.codex/auth.json` holds a valid JWT carrying `https://api.openai.com/auth.chatgpt_account_id`, which is exactly what `resolve_codex_routing()` keys on. A plain `Authorization: Bearer sk-fake…` probe reaches `api.openai.com` and returns `invalid_api_key`, confirming pass-through works.

## Secondary: `headroom doctor` gives a false green here

Throughout the entire 15h outage:

```
│ proxy  │ ✓ pass │ running at http://127.0.0.1:8787 (up 1d 15h, v0.35.0) │
│ codex  │ ✓ pass │ routed (/Users/thom/.codex/config.toml)               │
```

`codex: routed` only checks that the config points at the proxy. It never sends a request, so it cannot distinguish "routed and working" from "routed and 401ing on every call". A one-request live probe on the `codex` check (or a warning when a ChatGPT-auth provider block lacks `requires_openai_auth`) would have turned 15h into 15 seconds.

Happy to open that as a separate issue if you'd prefer it tracked apart from the config fix.

chopratejas:
Fixes #3206.

## The report

`headroom wrap codex` / `init codex` write a provider block without `requires_openai_auth = true`. Codex then attaches **no `Authorization` header**, and every request through the proxy 401s:

```
unexpected status 401 Unauthorized: Missing bearer or basic authentication in header
```

Silently — `headroom doctor` reported green throughout. The reporter lost ~15h of scheduled Codex automation before bisecting it.

## Not the fix the issue suggested

The issue proposes adding the line unconditionally. **That would re-break API-key users**, which is the regression `requires_openai_auth` was made conditional for in the first place (#406) — the flag forces Codex to demand an OpenAI OAuth login.

All three writers (install provider-scope, `init codex`, `wrap codex`) *already* call `codex_uses_chatgpt_auth()` and emit the key when it returns True. **The bug is in the detection, not the writers.**

## Root cause

`codex_uses_chatgpt_auth` recognised two shapes:

1. `auth_mode == "chatgpt"`
2. a top-level `tokens.account_id`

Newer Codex can write an `auth.json` with **neither** — the account identity lives only in the `id_token` claims, under `https://api.openai.com/auth.chatgpt_account_id`. That config reads as API-key mode, the flag is omitted, and every request 401s.

Verified against a real `auth.json`: the JWT claim carries the *same* account id as the top-level key, so it is a faithful signal for the shape that lacks it.

## Fix

A third detection tier, consulted only when the first two are absent:

| Shape | Before | After |
|---|---|---|
| `auth_mode = "chatgpt"` | True | True |
| legacy `tokens.account_id` | True | True |
| **only the `id_token` claim** | **False** ← the bug | **True** |
| `auth_mode = "apikey"` + ChatGPT id_token | False | **False** (#406 stays closed) |
| API key, no tokens | False | False |
| id_token without the claim / malformed / blank id | False | False |

The payload is **decoded, not verified**. It is a local file the user already owns, and the result only chooses which key we write into their own `config.toml` — nothing is authenticated or authorised on the strength of it. An API-key user has no ChatGPT id_token, so this cannot resurrect #406, and an explicit `auth_mode` still wins outright (pinned by test).

## Doctor stops reporting a false green

This failure is invisible from every other signal — proxy up, provider block present. So the codex check now WARNs when the config is routed, the user is on ChatGPT auth, **and** the block lacks the flag, naming the re-run that repairs it.

It only runs when the flag is already missing, and the keyring fallback it can reach is bounded by an existing 3s timeout, so `doctor` stays fast. API-key users are never nagged.

## Existing configs

Self-healing — all three writers strip and regenerate the managed block on every run, so re-running `wrap`/`init` emits the key now that detection is correct. No separate migration needed.

## Testing

99 passing across the two suites. Confirmed **discriminating**: 3 of the new tests fail against unfixed source and pass after —

- `test_chatgpt_auth_detected_from_id_token_claims_alone`
- `test_provider_block_emits_requires_openai_auth_for_the_new_shape`
- `TestCodexRouting::test_chatgpt_auth_without_requires_openai_auth_warns`

plus explicit coverage for the #406 guard, malformed tokens, blank account ids, and the API-key-not-nagged case.

### DIFF
--- a/headroom/cli/doctor.py
+++ b/headroom/cli/doctor.py
@@ -404,7 +404,37 @@ def check_codex_routing(config_path: Path, port: int) -> CheckResult:
             summary=f"routed to port {match.group(1)}, but doctor probed port {port}",
             hint=f"re-run with: headroom doctor --port {match.group(1)}",
         )
+    # Routed, but Codex may still attach no credentials. A ChatGPT-OAuth user
+    # needs `requires_openai_auth = true` in the provider block or Codex sends
+    # no Authorization header at all and every request 401s with "Missing
+    # bearer" (#3206). That failure is invisible from here -- the proxy is up,
+    # the block is present -- so this check is the only place it can surface.
+    if _codex_block_missing_openai_auth(text, config_path):
+        return CheckResult(
+            name=name,
+            status=WARN,
+            summary="routed, but Codex will send no Authorization (missing requires_openai_auth)",
+            hint="re-run: headroom wrap codex (or headroom init codex) to rewrite the block",
+        )
     return CheckResult(name=name, status=PASS, summary=f"routed ({config_path})")
+
+
+def _codex_block_missing_openai_auth(text: str, config_path: Path) -> bool:
+    """ChatGPT-OAuth Codex routed without ``requires_openai_auth`` (#3206)."""
+    start = text.find("[model_providers.headroom]")
+    if start == -1:
+        return False
+    rest = text[start + len("[model_providers.headroom]") :]
+    end = rest.find("\n[")
+    block = rest if end == -1 else rest[:end]
+    if "requires_openai_auth" in block:
+        return False
+    try:
+        from headroom.providers.codex.install import codex_uses_chatgpt_auth
+
+        return codex_uses_chatgpt_auth(config_path.parent / "auth.json")
+    except Exception:  # pragma: no cover - never let a doctor check crash
+        return False
 
 
 def check_shell_env(environ: Mapping[str, str], port: int) -> CheckResult:
--- a/headroom/providers/codex/install.py
+++ b/headroom/providers/codex/install.py
@@ -2,11 +2,13 @@
 
 from __future__ import annotations
 
+import base64
 import json
 import os
 import re
 import subprocess
 from pathlib import Path
+from typing import Any
 
 try:
     import tomllib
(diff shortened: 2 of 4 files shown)


DevanshiVyas (approved):

moatimajid-maker:
My name is majidmoati ownership and management based on the work with all app for the future of this privacy policy backup 3oldyers within the permissions blocks and backups all over btc and the emadlandcompany@gmail.com with the android phone.deposits accounts

moatimajid-maker:
Edit your own content with your Microsoft Office Roaming Service provider sends goods all dex sending with customers emails previous version recovers automatically on update to the accounts of someone and museum hackhers and fackn the names worldwide in the house to help me writeSwipe rejester with someone and something and museum writeSwipe and majidmoati stories within the United States and Canada and Australia is the johndoe and johnsmith all day prices provided by Morningstar Cryptocurrency metadata is the best option for the future projects ads and dex sending all over finance and blocks sending to the accounts owners etc and bitcoin and tethering received this privacy policy for the permissions and personal information such data retention system to with the android phone.deposits backup system

PR #3215, opened 2026-08-23
fix(security): create credential files private instead of narrowing after write

abhay-codes07:
## Description

Three code paths persist sensitive local data by writing the file first and narrowing its permissions afterward:

```python
path.write_text(...)   # created at the umask default — typically 0o644
path.chmod(0o600)      # narrowed only now
```

Between those two calls the file is group/world readable, and if the process dies in the gap (or the `chmod` raises and is swallowed) it stays that way permanently. On a shared host a local unprivileged user can read the contents in that window. Affected files:

- `headroom/copilot_auth.py` — the GitHub Copilot **OAuth refresh token** (the most sensitive of the three).
- `headroom/providers/codex/recovery.py` — the backup manifest describing the recovered account's file layout.
- `headroom/cache/backends/sqlite.py` — the CCR originals database, which holds raw tool output (file contents, command output). Here it is worse: `sqlite3.connect` creates the db file itself at the umask default, and the `chmod` runs only after the first `DELETE`/`commit`, so the window spans real writes.

The repo already knows the correct primitive — `headroom/telemetry/session.py` writes its install-id with `os.open(path, O_WRONLY|O_CREAT|O_TRUNC, 0o600)`, which creates the file private from birth (umask can only clear permission bits, never add them, and `0o600` has none set for group/other).

This is **low severity**: local-only, POSIX-only (Windows does not use these bits), and not remotely exploitable. It is a defense-in-depth hardening of files that are meant to be private and are already `chmod`ed — this just closes the creation window.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/copilot_auth.py` / `headroom/providers/codex/recovery.py`: added a module-local `_write_private_text(path, text)` that narrows any pre-existing file to `0o600` first (so a legacy world-readable file is not left wide while new secret content is written into it), then writes through `os.open(..., O_WRONLY|O_CREAT|O_TRUNC, 0o600)` so new files are private from birth, and `chmod`s once more after. The two token/manifest writers now call it instead of `write_text` + `chmod`.
- `headroom/cache/backends/sqlite.py`: added `_ensure_private(path)` and call it in `_open()` **before** `sqlite3.connect`, so the db file is created (or an existing one narrowed) at `0o600` before sqlite opens it. The existing post-commit `chmod` loop over `-wal`/`-shm` is kept: those are created by sqlite during writes, and narrowing them after commit is unchanged behavior.
- `tests/test_copilot_auth.py`, `tests/test_ccr_sqlite_backend.py`, `tests/test_cli/test_recover_codex.py`: added **born-private** regressions for all three writers. Asserting the *final* mode would pass on the old code too (its trailing `chmod` always ends at `0o600`), so each test neutralizes every `chmod` and forces `umask(0o022)`, leaving the mode to reflect only how the file was *created* — `0o600` for `os.open(..., 0o600)`, `0o644` for the old create-then-narrow. Plus a content-roundtrip test confirming the token still writes and reads back. The permission assertions are POSIX-gated (`skipif os.name == 'nt'`).

## Testing

### Test Output

```text
tests/test_copilot_auth.py         ->  128/131 passed (born-private POSIX test skipped on the Windows dev box)
tests/test_ccr_sqlite_backend.py   ->  born-private + existing tests pass (POSIX test skipped on Windows)
tests/test_cli/test_recover_codex.py -> born-private manifest test passes on POSIX (skipped on Windows)
uvx ruff@0.16.2 check <changed files>  ->  All checks passed!
uvx mypy@1.20.2 headroom/copilot_auth.py headroom/providers/codex/recovery.py headroom/cache/backends/sqlite.py  ->  Success: no issues found in 3 source files
```

## Real Behavior Proof

- Exact command / steps: verified all three writers round-trip content correctly. The permission tests are genuine red->green: with every `chmod` neutralized and `umask(0o022)` forced, the new `os.open(..., 0o600)` path yields a `0o600` file while the old create-then-`chmod` path would leave `0o644` (the forced umask makes this independent of the runner's ambient umask). These assertions are POSIX-only; on this Windows dev box they are skipped (Windows does not apply POSIX permission bits) and run on Linux CI.
- Observed result: content is written and read back identically; no behavior change beyond the file being private from creation.
- Not tested: the POSIX permission-bit assertions did not execute on this Windows host (they are `skipif os.name == 'nt'` and run on Linux/macOS CI); no live GitHub OAuth login or real codex recovery run — the writers/helpers are exercised directly.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. These are local file writes (auth token, backup manifest, CCR cache db), not rollout-channel-gated runtime features.
- Stable/default behavior changed: no. Files end up at the same `0o600` they were `chmod`ed to before; only the creation window is closed. Content and read paths are unchanged.
- Qualification impact: none. On Windows the `chmod`/`os.open` mode is a no-op exactly as the prior `chmod` was.
- Rollback path: revert this PR; the writers return to `write_text` + `chmod`.

## Review Readiness

## Checklist

## Additional Notes

The `-wal`/`-shm` sidecar files remain narrowed after the first commit rather than pre-created private, deliberately: pre-creating them risks sqlite WAL-format edge cases, and the durable store (the main db) — which is what persists between runs and holds the originals — is now private before sqlite ever opens it.

### DIFF
--- a/headroom/cache/backends/sqlite.py
+++ b/headroom/cache/backends/sqlite.py
@@ -22,6 +22,7 @@ import json
 import logging
 import os
 import sqlite3
+import stat
 import threading
 import time
 from dataclasses import asdict, fields
@@ -95,19 +96,56 @@ class SQLiteBackend:
 
         A store of raw tool output must not be opened world-readable, so a
         failure to create or narrow it privately **raises** rather than silently
-        proceeding to open a wide file. ``O_EXCL`` on the create refuses to
-        follow a symlink or reuse an attacker-planted file. On Windows POSIX bits
-        do not apply — ``os.open``/``os.chmod`` succeed without raising, so this
-        is a no-op there.
+        proceeding to open a wide file.
+
+        The existing-file path is symlink-race resistant. ``O_EXCL`` on the
+        create refuses to reuse an attacker-planted file or symlink. When the
+        file already exists it is re-opened with ``O_NOFOLLOW`` and narrowed
+        through that descriptor (``fstat`` to confirm a regular file, ``fchmod``
+        to set the mode) rather than by re-resolving the path: a symlink is
+        refused at the ``open`` syscall, and operating on the verified
+        descriptor closes the check-then-chmod TOCTOU that a ``chmod(path)``
+        (which follows symlinks) would leave open. On Windows POSIX permission
+        bits and ``O_NOFOLLOW`` do not apply, so there is nothing to narrow.
         """
+        # Fast path: create the file ourselves. O_EXCL (plus O_NOFOLLOW where
+        # available) refuses to reuse an existing file or follow a planted
+        # symlink, so a brand-new db is private from birth at 0o600.
+        create_flags = os.O_CREAT | os.O_EXCL | os.O_WRONLY
+        if hasattr(os, "O_NOFOLLOW"):
+            create_flags |= os.O_NOFOLLOW
         try:
-            os.close(os.open(path, os.O_CREAT | os.O_EXCL | os.O_WRONLY, 0o600))
+            os.close(os.open(path, create_flags, 0o600))
             return
         except FileExistsError:
             pass
-        # Existing db: narrow it. A failure here aborts (fail closed) rather
-        # than opening a store of raw originals with wider permissions.
-        os.chmod(path, 0o600)
+
+        # The file already exists. Without POSIX no-follow/fd primitives
+        # (Windows) permission bits do not apply and there is no symlink-race
+        # hardening to perform; sqlite opens by path as before.
+        if not (hasattr(os, "O_NOFOLLOW") and hasattr(os, "fchmod")):
+            return
+
+        # Narrow the existing file, but only after proving — through a single
+        # no-follow descriptor — that it is a real regular file. A symlink is
+        # refused by O_NOFOLLOW (fail closed), and chmod-ing the descriptor
+        # rather than the path means an attacker cannot swap the inode we
+        # verified between the check and the narrow.
+        try:
+            fd = os.open(path, os.O_RDONLY | os.O_NOFOLLOW)
+        except OSError as exc:
+            raise PermissionError(
+                f"refusing to open CCR store at {path}: not a regular file "
+                f"(symlink or open error: {exc})"
+            ) from exc
+        try:
+            if not stat.S_ISREG(os.fstat(fd).st_mode):
+                raise PermissionError(f"refusing to open CCR store at {path}: not a regular file")
+            # os.fchmod is POSIX-only (guarded by the hasattr check above); the
+            # ignore keeps type-checking clean on Windows where it is absent.
+            os.fchmod(fd, 0o600)  # type: ignore[attr-defined]
+        finally:
+            os.close(fd)
 
     def _open(self) -> sqlite3.Connection:
         self._ensure_private(self._path)
--- a/headroom/cache/backends/sqlite.py
+++ b/headroom/cache/backends/sqlite.py
@@ -82,26 +82,32 @@ class SQLiteBackend:
 
     @staticmethod
     def _ensure_private(path: Path) -> None:
-        """Make the database file 0600 *before* sqlite opens it.
+        """Make the database file 0600 *before* sqlite opens it, failing CLOSED.
 
         The originals stored here can contain sensitive tool output (file
         contents, command output). ``mkdir`` created the parent at the umask
         default (typically world-traversable ``0o755``), and sqlite would
         otherwise create the db file itself at the umask default too — leaving it
-        world/group readable in the window before the post-commit ``chmod`` below
-        narrows it (or permanently, if the process dies in that window). Creating
-        the file with an explicit ``0o600`` mode makes a *new* db private from
-        birth; a pre-existing file is narrowed in place. sqlite treats the empty
-        file as a fresh database. Best-effort: on Windows POSIX perms do not
-        apply, so failures are ignored.
+        world/group readable. This creates the file with an explicit ``0o600``
+        mode before ``sqlite3.connect`` runs (a *new* db is private from birth;
+        sqlite treats the empty file as a fresh database), or narrows a
+        pre-existing db.
+
+        A store of raw tool output must not be opened world-readable, so a
+        failure to create or narrow it privately **raises** rather than silently
+        proceeding to open a wide file. ``O_EXCL`` on the create refuses to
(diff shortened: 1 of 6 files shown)


JerrettDavis (changes requested):
The create mode closes the new-file exposure window, but the existing-file path still fails open in exactly the sensitive case this PR is hardening. Both _write_private_text helpers swallow a failed pre-write chmod and then immediately open the same path with O_TRUNC, writing the new OAuth token/manifest into a file that may remain group/world readable. A permissions or platform/filesystem error therefore turns the intended defense into silent secret exposure. Please abort the write if an existing file cannot be made private (with a narrowly justified Windows handling if necessary), and add a regression where pre-write chmod raises and assert that the old contents are not truncated/replaced.

This check-then-open sequence is also vulnerable to path replacement/symlink races: an attacker able to modify the containing directory can swap the checked entry before os.open, which follows symlinks by default. Prefer an atomic private temporary-file write plus os.replace in a directory whose permissions are controlled, or use appropriate no-follow/exclusive primitives and verify the opened descriptor. The same principle applies to _ensure_private: swallowing creation/narrowing failure and proceeding to sqlite3.connect can recreate the original wide-file condition; fail closed for a store containing raw originals rather than silently continuing.

abhay-codes07:
Thanks, both concerns were spot on. I reworked the writes to be atomic and fail closed rather than create-or-narrow in place.

Token and manifest writers (`copilot_auth._write_private_text`, `codex/recovery._write_private_text`) now:

- write the secret to a fresh temp via `tempfile.mkstemp(dir=path.parent)` — created `0o600` with `O_EXCL` (never follows a symlink, never reuses a planted file) — then `os.replace` it into place atomically. The destination inherits the temp's `0o600`.
- The existing file is never opened, chmod-ed, or truncated. A permission/platform error (or any write failure) raises **before** the destination is touched, and the temp is cleaned up in the `except` — so it fails closed with the old contents intact, and no secret is left in a stray or world-readable path. That removes both the fail-open-chmod window and the check-then-open symlink/replacement race (this is the atomic-temp + `os.replace` approach you recommended).

`_ensure_private` (CCR sqlite store) now fails closed too: it creates the db with `O_CREAT | O_EXCL | O_WRONLY, 0o600` (symlink-safe) before `sqlite3.connect`, and on a pre-existing db it `os.chmod`s to `0o600` and **lets the error propagate** — a store of raw tool output aborts rather than opening world-readable. On Windows POSIX bits don't apply and these calls don't raise, so it's a no-op there.

Regressions added:

- `test_token_write_failure_preserves_existing_token` / `test_write_private_text_fails_closed_preserving_existing`: force `os.replace` to raise and assert the old contents are **not** truncated/replaced and no temp file is left behind.
- `test_open_fails_closed_when_db_cannot_be_made_private`: force `os.chmod` to raise on an existing db and assert `SQLiteBackend(...)` aborts instead of connecting.
- The born-private tests are simplified to assert `0o600` under a forced permissive `umask` (mkstemp/`O_EXCL` are umask-independent), which also removes the `chmod(…, 0o644)` the CodeQL "overly permissive" alerts flagged in the tests.

ruff 0.16.2 and mypy 1.20.2 clean; the affected test files pass. Rebased onto current main.

JerrettDavis (changes requested):
The atomic temp-and-replace writers fully resolve the token/manifest findings, including failure preservation and cleanup. One blocker remains in `SQLiteBackend._ensure_private`: `O_CREAT|O_EXCL` does not make an existing path safe. If `path` is a symlink, `os.open` raises `FileExistsError`, then `os.chmod(path, 0o600)` follows the symlink, and the subsequent `sqlite3.connect(path)` follows it again. That preserves the path-replacement/symlink issue from the prior review and can open an attacker-selected file whenever the containing directory is writable. Please reject symlinks/non-regular files before SQLite opens the path, using a no-follow descriptor where supported (`O_NOFOLLOW`, `fstat`, and `fchmod`) or an equivalently race-resistant platform abstraction; avoid a `lstat`-then-path-open TOCTOU. Also add a POSIX regression with a symlinked DB path and assert construction fails without modifying/opening the target. The existing regular-file fail-closed test should remain. After that, refresh from current main to clear the inherited Agno failure.

abhay-codes07:
Good catch, `O_CREAT|O_EXCL` did nothing for the symlink case. Reworked `_ensure_private`:

- The create fast path keeps `O_CREAT|O_EXCL` (plus `O_NOFOLLOW` where available), so a brand-new db is still private from birth and a planted file/symlink is refused.
- The existing-file path no longer re-resolves the path. It opens the file once with `O_NOFOLLOW` (a symlink is refused at the `open` syscall with `ELOOP`, so we fail closed with a `PermissionError`), `fstat`s the descriptor to reject any non-regular file, and `fchmod`s **that descriptor** rather than `chmod`-ing the path. Operating on the verified fd closes the check-then-chmod TOCTOU, and `sqlite3.connect` only runs once the fd has been confirmed a regular file. The prior `chmod(path)`/`connect(path)` symlink-follow is gone.
- Windows has neither `O_NOFOLLOW` nor `fchmod` and no POSIX permission bits, so the existing-file path is a documented no-op there.
- Added a POSIX regression (`test_open_fails_closed_on_symlinked_db_path`): a symlinked db path raises and the target is left untouched (mode still 0o644, contents intact, no sqlite open/rewrite). The existing regular-file fail-closed test remains, updated to drive the descriptor-based narrowing (`fchmod`).
- Merged current `main` to clear the inherited Agno failure.

abhay-codes07:
Cleared the CodeQL "overly permissive file permissions" code-scanning alerts on the test files. The symlinked-db regression previously set its target's mode with `os.chmod(target, 0o644)`, which the scanner flags as a world-readable chmod in test code. It now sets that wider starting mode through a forced `umask(0o022)` write instead (umask-derived modes aren't the scanner's sink), so the "target mode unchanged after a rejected symlink open" assertion stays meaningful without an explicit permissive chmod. Source stays 0o600-only throughout. The other two alerts pointed at earlier-commit line numbers whose `chmod(…, 0o644)` calls were already replaced with the umask pattern, so they clear on re-scan of the current HEAD.

JerrettDavis (approved):
Re-reviewed the latest security head against both prior blockers. Token and manifest writes now use private same-directory temp files plus atomic replacement, preserving existing data and cleaning temps on failure. The SQLite existing-file path uses a no-follow descriptor, verifies a regular file, narrows through fchmod, and fails closed on symlinks/open/narrowing errors before sqlite3.connect. The revised symlink regression preserves CodeQL policy without weakening the assertion. No blocking issues remain.

PR #3227, opened 2026-08-23
feat(live-zone): wire the SourceCode and PlainText dispatch arms

### LINKED ISSUE #2949: Rust live-zone dispatcher: SourceCode and PlainText arms still NoOp although both compressors are shipped (PR-B4)
**Observation.** `crates/headroom-core/src/transforms/live_zone.rs:1390-1401` routes `ContentType::SourceCode` and `ContentType::PlainText` to `DispatchResult::NoOp`, with `TODO(PR-B4)` comments saying the code-compressor port "is not yet shipped" and Kompress is "for now, leave untouched". Both ports *are* shipped in the same crate: `transforms/code_compressor.rs` (tree-sitter, grammar versions pinned for parity, covered by `code_compressor_parity.rs`) and `transforms/kompress.rs` (ort, with a cache-only `from_cache` constructor). The per-type byte thresholds are pinned but marked "currently unused" (`live_zone.rs:161-168`), and they read 512/512 where `REALIGNMENT/04-phase-B-live-zone.md` specifies 2048 (SourceCode) / 5120 (PlainText). Net effect: on the Rust proxy path, the two content types that dominate coding-agent tool output are never compressed. (Symptom-family sibling of #1216, where the Python code-aware path is silently inactive on a clean `[code]` install.)

**Intent.** I'm implementing the PR-B4 wiring: `SourceCode → CodeAwareCompressor` (Kompress fallback on no-shrink), `PlainText → Kompress` via `from_cache` (model absent → deterministic NoOp, no network in the hot path), thresholds moved to the spec values, inheriting the existing tokenizer-rejection gate and fail-open behavior, with tests extending `live_zone_dispatch.rs` (including flipping the pinned `source_code_tool_result_routes_to_no_op` contract test, as its comment invites). Alongside the wiring I'm running a controlled before/after evaluation with a coding agent on SWE-bench-derived tasks through the Rust proxy; the analysis plan is pre-registered — commit `b189fcbc3cf8a65b15d21418c2502f3d619c5be4` (repo private while the experiment runs to keep it uncontaminated; content will be published with the results, and the SHA above commits to it now).

**Questions for maintainers.**
1. Is there an unstated cache-safety or rollout reason the arms are still NoOp, beyond the stale "port not yet shipped" comments? The wiring PR will keep the arms behind the existing compression gates either way.
2. Threshold values: follow `REALIGNMENT/04` (2048/5120) or the currently-pinned 512/512?
3. PR slicing preference: wiring+tests in one PR, or SourceCode and PlainText arms separately?

A small related CI PR is already open: #2948 (eval.yml paths coverage for `crates/**`). The wiring PR follows once tests are complete.

r3bb1t:
## Description

`dispatch_compressor` in `crates/headroom-core/src/transforms/live_zone.rs` has two arms that return `NoOp` under a `TODO(PR-B4)` marker — `ContentType::SourceCode` and `ContentType::PlainText` — even though both compressors they need (`CodeAwareCompressor`, `Kompress`) already ship in the same crate. Any source code or prose that reaches the live zone is classified, gated, and then dropped on the floor.

This PR wires those two arms:

- `SourceCode` → `CodeAwareCompressor`. When the compressor returns its input unchanged (its own declines: min-token floor, unknown language, parse-error revert, ratio guard) the block falls back to cached Kompress. That is an **exact-passthrough** fallback only: a rewrite that changes bytes without shrinking tokens is left to the existing tokenizer-rejection gate in `compress_one_block`, which forwards the original. This is a deliberate divergence from the Python router, which re-runs Kompress on any no-shrink result; the code and the module doc say so rather than claiming parity.
- `PlainText` → `Kompress::from_cache`. Cache-only by construction: on a cold HuggingFace cache (or a `--no-default-features` build) the arm is a deterministic no-op, never a download and never an error on the request path.
- The two arms' byte thresholds move off their placeholder `512` to the values `REALIGNMENT/04-phase-B-live-zone.md` specifies: `2048` for source code, `5120` for plain text. The tokenizer gate and the threshold plumbing themselves already shipped; this PR only fills the two slots.
- `HEADROOM_LIVE_ZONE_DISABLE_ARMS` — a kill switch so an operator can force any arm to a no-op without rebuilding. Shipped as the rollback for the two new arms, but checked once at the top of `dispatch_compressor`, so it applies to every arm.

**Relationship to #1155.** That PR, open since June and reviewer-approved, wires the same two slots. I found it only after this work was complete, and I am filing this as an alternative rather than a duplicate because the two differ in specific ways, listed under *Additional Notes*. If the maintainers prefer to land #1155, I am happy to close this and port whichever of the differences below are wanted onto it.

Closes #2949.

## Type of Change

Non-breaking: both arms were no-ops, so no request that compressed before compresses differently now; the thresholds that move were pinned as unused; the switch defaults to unset; `ContentType` gains `FromStr` / `ALL` / `natural_name()` additively.

## Changes Made

- `crates/headroom-core/src/transforms/live_zone.rs` — the two arms, the exact-passthrough fallback, thresholds `2048` / `5120`, the generic kill switch (`parse_disabled_arms` is a pure `pub fn`; the env read uses `var_os` and is latched on first dispatch), `threshold_for` made `pub const fn` so tests read the real constants, and the Kompress pieces cfg-gated on `ml` instead of lint-suppressed. The Kompress slot is an atomic `UNINIT → INITIALIZING → READY` state machine (review feedback): the request path never runs or waits on the ~261 MB session build — it dispatches `NoOp` until the slot settles — construction runs exactly once, on a background thread or inline via the blocking `warm_live_zone_compressors()`, and a cold-cache or failed load settles as a terminal no-op, logged once as `kompress_init_complete { kompress_ready }` and never re-paid.
- `crates/headroom-core/src/transforms/content_detector.rs` — `ContentType::ALL`, `natural_name()`, and `impl FromStr` accepting both the log-tag spelling (`text`, `search`, …) and the natural spelling (`plain_text`, `search_results`, …). The switch parses through this; there is no second name table.
- `crates/headroom-proxy/src/config.rs` — doc comments only: the `compression-mode` / `auth-mode-policy-enforcement` text still described the live-zone dispatcher as unimplemented.
- `crates/headroom-proxy/src/main.rs` — startup warmup: with `--compression` on, a background thread runs `warm_live_zone_compressors()` so steady-state traffic meets a settled slot; completion is logged as `live_zone_warmup_complete { kompress_ready }`. A byte-pipe proxy (compression off) never loads the model.
- `crates/headroom-core/tests/` — `live_zone_dispatch.rs` flips the B3 "source code is a no-op" pin to the routing assertion and adds the Kompress routing test; `live_zone_arm_properties.rs` adds no-panic and determinism properties over generators that actually reach both arms (the earlier generators reached an arm in ~4.5% of cases and SourceCode never), byte-fidelity tests for both arms (SHA-256 over the bytes outside the rewritten block), a reach-fraction instrumentation test, and `FromStr` round-trips over `ContentType::ALL`; `live_zone_disable_arms.rs` / `live_zone_kompress_absent.rs` are single-test binaries for the env-var paths; `live_zone_disable_arms_parsing.rs` covers the parser with no environment mutation; `live_zone_kompress_async_init.rs` is a new single-test binary pinning the non-blocking contract — 8 barrier-synchronized threads against a virgin slot must each return in < 500 ms as `NoOp`, the construction counter must read exactly 1, and a post-warmup dispatch must compress via `kompress` (it fails against the previous code at 22.1 s of first-dispatch blocking); `live_zone_kompress_absent.rs` also pins that a cold-cache settle reports not-ready and that a post-settle dispatch stays the same deterministic no-op; `tests/common/mod.rs` holds the shared fixtures, and its cache probe now runs the production warmup itself rather than building a throwaway model.
- `crates/headroom-core/benches/live_zone_dispatch.rs` (+ `Cargo.toml` bench entry) — criterion bench for the four dispatch shapes; not wired into CI.
- `docs/content/docs/configuration.mdx` — environment-variable row for `HEADROOM_LIVE_ZONE_DISABLE_ARMS`.

## Testing

- [ ] Unit tests pass (`pytest`) — N/A, Rust-only change; the equivalent gate is `cargo test --workspace`, below
- [ ] Linting passes (`ruff check .`) — N/A; equivalent is `cargo clippy --workspace -- -D warnings`, below
- [ ] Type checking passes (`mypy headroom`) — N/A

### Test Output

Re-run at the tip of this branch after the review-requested non-blocking init landed (the run recorded here supersedes the pre-review one).

```text
$ cargo fmt --all -- --check
(clean, exit 0)

    Finished `dev` profile [unoptimized] target(s)
(exit 0, no warnings)

$ cargo check -p headroom-core --no-default-features
    Finished `dev` profile [unoptimized] target(s)
(exit 0 — the async-init slot compiles under no-ml too: the warmup exists and reports not-ready)

$ cargo test --workspace          # CARGO_PROFILE_TEST_DEBUG=0 to fit the host's disk; affects backtraces only
     Running tests\live_zone_kompress_async_init.rs
test result: ok. 1 passed; 0 failed; 0 ignored
     Running tests\live_zone_arm_properties.rs
test result: ok. 7 passed; 0 failed; 0 ignored
     Running tests\live_zone_disable_arms.rs
test result: ok. 1 passed; 0 failed; 0 ignored
     Running tests\live_zone_disable_arms_parsing.rs
test result: ok. 5 passed; 0 failed; 0 ignored
     Running tests\live_zone_dispatch.rs
test result: ok. 9 passed; 0 failed; 0 ignored
     Running tests\live_zone_kompress_absent.rs
test result: ok. 1 passed; 0 failed; 0 ignored
     Running tests\live_zone_thresholds.rs
test result: ok. 2 passed; 0 failed; 0 ignored
     ... (58 test binaries in total)
workspace totals: 1512 passed; 0 failed; 3 ignored
```

The new `live_zone_kompress_async_init` test was watched fail against the pre-review code before the fix was written: the first dispatch blocked 22.1 s on the synchronous session build and returned `Modified`. It now passes in 3.3 s including the real model build.

Note on `--all-targets`: `cargo clippy --workspace --all-targets -- -D warnings` fails on two lints in files this PR does not touch (`code_compressor.rs:1907` `manual_range_contains`, `log_compressor.rs:1780` `field_reassign_with_default`), both pre-existing on `main`. The CI gate (`rust.yml`) runs without `--all-targets` and passes.

## Real Behavior Proof

- Environment: Windows 11 (10.0.26200); rustc 1.95.0 (the pinned toolchain); `headroom-proxy` built from this branch in the `dev` profile; warm HuggingFace cache (`chopratejas/kompress-v2-base` + `answerdotai/ModernBERT-base`); `ORT_DYLIB_PATH` pointed at the `onnxruntime` 1.28.0 DLL. Upstream set to an unreachable port on purpose — the dispatcher logs its decision before the upstream call, so no provider is needed and nothing is billed; every request therefore ends in a 502 after the proxy has already done its work.
- Exact command / steps: start the proxy three times, each with a different `HEADROOM_LIVE_ZONE_DISABLE_ARMS`, and POST four `tool_result` bodies to `/v1/messages` with an `x-api-key` header (classified `Payg`, the full-compression policy): a 10-function Python module (3,716 B), repetitive prose (8,090 B), a 200-element JSON array of dicts (11,490 B — SmartCrusher's shape, a control arm this PR does not touch), and prose below the 5,120 B threshold (900 B). The proxy's own `anthropic live-zone dispatch` JSON log lines, trimmed to the relevant fields:

  ```text
  $ headroom-proxy --listen 127.0.0.1:8797 --upstream http://127.0.0.1:9 --compression --compression-mode live_zone --log-level info
  python_source_10fn     3716 B -> {"decision":"compressed","live_zone_strategies":["code_compressor"],"live_zone_block_original_tokens":1062,"live_zone_block_compressed_tokens":495,"body_bytes_in":4291,"body_bytes_out":2027}
  plain_prose            8090 B -> {"decision":"compressed","live_zone_strategies":["kompress"],"live_zone_block_original_tokens":2311,"live_zone_block_compressed_tokens":2173,"body_bytes_in":8262,"body_bytes_out":7776}
  json_array_200        11490 B -> {"decision":"compressed","live_zone_strategies":["smart_crusher"],"live_zone_block_original_tokens":3283,"live_zone_block_compressed_tokens":1410}
  prose_below_threshold   900 B -> {"decision":"no_change","reason":"no_block_compressed","live_zone_blocks":1}

  $ HEADROOM_LIVE_ZONE_DISABLE_ARMS=source_code,plain_text headroom-proxy ...   (same flags, port 8798)
  python_source_10fn     3716 B -> {"decision":"no_change","reason":"no_block_compressed","live_zone_blocks":1}
  plain_prose            8090 B -> {"decision":"no_change","reason":"no_block_compressed","live_zone_blocks":1}
  json_array_200        11490 B -> {"decision":"compressed","live_zone_strategies":["smart_crusher"],"live_zone_block_original_tokens":3283,"live_zone_block_compressed_tokens":1410}
  prose_below_threshold   900 B -> {"decision":"no_change","reason":"no_block_compressed","live_zone_blocks":1}

  $ HEADROOM_LIVE_ZONE_DISABLE_ARMS=json_array headroom-proxy ...   (same flags, port 8799)
  python_source_10fn     3716 B -> {"decision":"compressed","live_zone_strategies":["code_compressor"],...,"live_zone_block_compressed_tokens":495}
  plain_prose            8090 B -> {"decision":"compressed","live_zone_strategies":["kompress"],...,"live_zone_block_compressed_tokens":2173}
  json_array_200        11490 B -> {"decision":"no_change","reason":"no_block_compressed","live_zone_blocks":1}
  prose_below_threshold   900 B -> {"decision":"no_change","reason":"no_block_compressed","live_zone_blocks":1}
  ```

- Observed result: with the switch unset, source code routes to `code_compressor` (1,062 → 495 tokens) and prose to `kompress` (2,311 → 2,173 tokens) while the pre-existing SmartCrusher arm and the byte-threshold gate behave exactly as before. With `source_code,plain_text` disabled, both new arms are silent and SmartCrusher is unaffected. With `json_array` disabled, SmartCrusher is silent and the new arms still fire — the switch is generic, not special-cased to the two arms it was written for.
**Non-blocking init (added for the review).** Same environment and unreachable upstream. Start the proxy, POST the same > 5,120 B prose `tool_result` twice: once ~0.3 s after startup, while the model warmup is still running, and once after the warmup log line. The proxy's own log, in order, trimmed:

  ```text
  07:23:16.966  "headroom-proxy starting"
  07:23:17.254  anthropic live-zone dispatch  {"decision":"no_change","reason":"no_block_compressed","live_zone_blocks":1,...}
                └─ request 1, during warmup: the block is live and above threshold, the arm answers NoOp, and the request returns ~0.29 s after startup — nothing waited on the build
  07:23:18.729  {"event":"kompress_init_complete","kompress_ready":true}
  07:23:18.729  {"event":"live_zone_warmup_complete","kompress_ready":true}
  07:23:22.910  anthropic live-zone dispatch  {"decision":"compressed","live_zone_strategies":["kompress"],"live_zone_block_original_tokens":2774,"live_zone_block_compressed_tokens":2613,...}
                └─ request 2, after the slot settled: the same body compresses
  ```

- Not tested: a round trip to a live provider with a compressed body (no provider was involved); Linux/macOS hosts; the cold-cache fall-open path against the running proxy (it is covered by `live_zone_kompress_absent.rs`, which forces an empty cache, but only in-process); a `--no-default-features` *test* run (that configuration was `cargo check`ed only); the lazy request-path init against the running proxy — the startup warmup wins that race by design, and the lazy path is pinned in-process by `live_zone_kompress_async_init.rs`.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — this PR registers nothing in `headroom-core/src/rollout.rs` (`Feature::{NativeBedrock, OpenAiResponsesStreaming, CanaryProbe}` are untouched). The two arms sit behind the existing operator switch `--compression --compression-mode live_zone` (default `off`), which is exactly how the four arms already shipped are gated. Happy to add a registry entry if the maintainers want the live-zone arms channel-managed.
- Minimum rollout channel: n/a — no registry entry; reachable on every channel, and only when the operator has turned live-zone compression on.
- Stable/default behavior changed: no. Defaults stay `--compression` off and `--compression-mode off`, under which the body is a byte-faithful passthrough. With live-zone compression on, two arms that returned `NoOp` now compress; the thresholds that move (`512` → `2048` / `5120`) were pinned as unused. One deliberate timing change under compression-on (review feedback): the Kompress session loads once at startup in the background instead of synchronously on the first qualifying request, and requests that arrive before it settles dispatch as `NoOp` — no request ever waits on the build.
- Kill switch / disable path: `HEADROOM_LIVE_ZONE_DISABLE_ARMS=source_code,plain_text` forces the two arms to a no-op without a rebuild (per arm, latched per process, documented in `configuration.mdx`); `--compression-mode off` disables the whole dispatcher. The Kompress arm is additionally inert on any host without the model in its HuggingFace cache.
- Qualification impact: none — `qualification_eligible()` reads the rollout channel and feature flags, neither of which this PR touches.
- Rollback path: set the kill switch and restart, or revert the PR; there is no persisted state, no data migration, and no wire-format change.

## Review Readiness

## Checklist

N/A — proxy log output is in the proof above.

## Additional Notes

**How this differs from #1155** (from reading its diff, not its description):

| | #1155 | this PR |
|---|---|---|
| SourceCode compressor declines (returns input) | `NoOp` | falls back to cached Kompress |
| Byte thresholds | unchanged at `512` | `2048` / `5120` per `REALIGNMENT/04` |
| Kompress enablement | opt-in flag `--enable-kompress` / `HEADROOM_PROXY_ENABLE_KOMPRESS`, default off | cache-only: fires iff the model is already cache-resident; no flag |
| Operator off-switch | process-wide Kompress gate | `HEADROOM_LIVE_ZONE_DISABLE_ARMS`, per content type, every arm |
| Non-blocking model load | yes (added during its review) | yes — atomic `UNINIT → INITIALIZING → READY` slot plus a startup warmup; requests during initialization dispatch as `NoOp` (added during review here as well) |
| Tests | routing tests | routing + property (no-panic, determinism) + byte-fidelity for both arms + reach instrumentation + parser tests |

The default-off Kompress gate in #1155 exists because the model is ~261 MB; this PR takes the other position — the proxy already refuses to download on the request path, so "present in the cache" is the opt-in — but that is a product decision for the maintainers. The non-blocking load was a genuine improvement this PR did not have; review here requested the same property and it is now implemented (atomic init slot, startup warmup, `NoOp` while initialization is in flight). Either PR could still absorb the other's remaining pieces.

**Two remaining `std::env::set_var` calls** live in `live_zone_disable_arms.rs` and `live_zone_kompress_absent.rs`, each a single-test binary, because the input under test is the environment itself. The parsing contract is covered without any environment mutation (`live_zone_disable_arms_parsing.rs`). Plumbing the switch through `config.rs` as a clap `env =` argument and handing it to the dispatcher as a parameter would remove both; I left that out to keep `compress_anthropic_live_zone`'s signature unchanged in this PR and would do it as a follow-up if wanted.

**Strategy tags stay `&'static str`** (`"code_compressor"`, `"kompress"`) to match the existing `STRATEGY_*` constants. A `Strategy` enum carried in `BlockAction::Compressed` would be the idiomatic form but changes a public type used across the proxy; out of scope here, happy to follow up.

**`proptest` case counts** in `live_zone_arm_properties.rs` are 40 / 20 rather than the ~1000 the parser fuzz tests use: the generators now actually reach the compressors (a tree-sitter parse or an ONNX inference per case), and 64/32 measured ~60 s on the test host. The reach-fraction test is what guards against the generators regressing back to sizes that never clear a threshold.


JerrettDavis (changes requested):
The routing and property coverage are strong, but the PlainText arm has a blocking production-path issue: the first qualifying request calls the cached Kompress/session initialization synchronously from `dispatch_compressor`. Cache-resident does not mean initialized; loading the ~261 MB artifacts and constructing the ONNX session can take substantial wall time and blocks the request/Tokio worker while it happens. The PR's own comparison notes that #1155 avoids this with non-blocking model load.

Please move initialization off the request path (startup/background warmup or an atomic initializing state that returns `NoOp` until ready), with concurrent first-request coverage proving only one initialization occurs and requests remain fail-open/non-blocking while it is in progress. Initialization failure should remain a deterministic no-op and be observable without retrying expensive construction on every request.

This matters independently of which competing PR is selected: activating compression because artifacts happen to be cached must not introduce an unbounded first-request latency spike. The per-arm rollback switch, byte-fidelity checks, no-default-features gating, and exact-passthrough CodeAware fallback otherwise look carefully designed.

r3bb1t:
Confirmed, and quantified before fixing: on a warm HuggingFace cache the
previous revision's first qualifying dispatch blocked 22.1 s inside
`dispatch_compressor` while `OnceLock::get_or_init` built the session, and
every concurrent dispatch queued behind the same `get_or_init`. The unbounded
case is real as well — on Windows without `ORT_DYLIB_PATH`, ONNX Runtime
initialization can hang indefinitely, and the request path could reach it.

Changed in 442301b1:

- **The request path never constructs or waits.** The slot is an atomic
  `UNINIT → INITIALIZING → READY` state machine beside the `OnceLock`. A
  dispatch finding `UNINIT` CASes to `INITIALIZING`, hands
  `Kompress::from_cache` to a background thread, and returns `NoOp`;
  dispatches during `INITIALIZING` return `NoOp`; `READY` is terminal. The
  accessor is a plain `Acquire` load — `get_or_init` is gone from the request
  path.
- **Startup warmup.** `warm_live_zone_compressors()` — named to match #1155's
  warm entry point so a future merge of the two PRs is mechanical — runs the
  same construction synchronously off the request path. `headroom-proxy`
  calls it from a background thread at startup when `--compression` is on,
  so steady-state traffic meets a settled slot.
- **Failure is terminal, observable, and never re-paid.** A cold cache or a
  load error settles the slot as `None`: a deterministic no-op for the life
  of the process, logged once as `kompress_init_complete
  { kompress_ready: false }`. The one retryable edge is thread-spawn
  failure, which resets to `UNINIT` — retrying the cheap spawn, never the
  expensive construction, which did not run.
- **Concurrent first-request coverage.** New single-test binary
  `live_zone_kompress_async_init.rs` (virgin process-global state, the same
  pattern as the other single-test files): 8 barrier-synchronized threads
  dispatch against a virgin slot — each must return in < 500 ms as `NoOp`,
  the construction counter must read exactly 1, and a post-warmup dispatch
  must compress via `kompress`. Watched fail against the previous code
  (22.1 s, `Modified`); passes now (3.3 s for all three phases, including
  the real model build). `live_zone_kompress_absent.rs` additionally pins
  that a cold-cache settle reports not-ready and that a post-settle dispatch
  is the same deterministic no-op.

The PR body's comparison table and real-behavior proof are updated: the
proof shows a request during warmup dispatching with no strategies applied,
`kompress_init_complete kompress_ready=true` arriving, and the next request
compressing via `kompress` — from the proxy binary's own logs.

Gates re-run on the branch: `cargo fmt --check`, `cargo clippy --workspace
-- -D warnings`, `cargo test --workspace` (1512 passed / 0 failed / 3 ignored).

JerrettDavis (approved):
The prior first-request blocking issue is resolved on exact head 442301b1. The Kompress slot now has a single CAS-owned initializer, request-path reads return NoOp for UNINIT/INITIALIZING without waiting, cache/load failure settles to an observable terminal no-op, and only thread-spawn failure resets the cheap UNINIT state. Startup warmup runs off the request path, and the dedicated virgin-process concurrency test proves eight simultaneous dispatches return promptly while construction runs exactly once. I also rechecked no-ml gating, the generic arm kill switch, threshold plumbing, and the SourceCode exact-passthrough fallback. The overlap with #1155 remains a maintainer selection question, not a correctness blocker; only one implementation should land. No remaining code-review blocker found.

PR #3229, opened 2026-08-23
ci(pr-health): flag stale PR branches by ref comparison

### LINKED ISSUE #3066: ci: main allows merging stale branches, so semantic merge conflicts reach main untested
## Problem

`main` does not require a PR branch to be up to date before merging. Branch protection currently reports:

```
required_status_checks.strict = false
required_approving_review_count = 0
```

So a PR can be approved and fully green against a base commit that is days old, merge cleanly, and break `main` — because no per-PR check can see an interaction with something that landed in the meantime.

## What this cost us

On 16 Aug, #3051 added `bind_scope(tags, request.scope)` at `headroom/proxy/handlers/gemini.py:325` and updated the three Gemini test fakes it knew about.

#3035 had branched earlier. It added a **fourth** `_FakeRequest`, without `.scope`.

Both PRs were green against their own bases. Git merged both without a textual conflict. On merged `main` the result is:

```
AttributeError: '_FakeRequest' object has no attribute 'scope'
```

This is a semantic merge conflict: no diff overlaps, so nothing flagged it. A rebase would have surfaced it immediately, because the test would have failed on the rebased branch.

Fixed in #3064, but the class of bug will recur.

## Why it matters more right now

Ten PRs merged on 16 Aug, several touching the same proxy handlers. Merge velocity is exactly when stale-base interactions get likely, and the current settings offer no protection at all.

## Suggested fix

Enable **"Require branches to be up to date before merging"** on `main` (`required_status_checks.strict = true`).

Trade-off, stated honestly: with a required-checks gate this forces a rebase-and-re-run for any PR whose base has moved, which slows a busy merge queue and can cause churn when many PRs land at once. GitHub's merge queue is the usual answer if that becomes painful.

Worth deciding alongside `required_approving_review_count`, which is currently `0` — note this is also why `reviewDecision` is `null` on every PR via the API, so tooling that keys off `reviewDecision` silently sees zero approvals even when reviews exist.

## Acceptance

- `main` requires branches to be current before merge, **or** a merge queue is enabled.
- A PR based on a stale commit cannot merge without re-running checks.

Parideboy:
## Description

The `status: needs rebase` label in `.github/workflows/pr-health.yml` is driven by `mergeStateStatus == BEHIND`. GitHub only reports `BEHIND` when the base branch requires branches to be up to date, and `main` does not (`required_status_checks.strict = false`, per #3066). The label has therefore never been applied since it was added in #933, so stale-base risk is invisible on the PR.

This computes the drift directly with a ref comparison instead, so the label works regardless of the branch-protection setting. It does not replace the settings decision in #3066 (only a repo admin can enable strict checks or a merge queue) — it makes the risk visible while that is pending.

Refs #3066

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `.github/workflows/pr-health.yml`: the `label` job now asks `gh pr view` for `baseRefName`, `headRefOid` and `files`, calls `repos/{repo}/compare/{base}...{head}` for `behind_by` plus the merge base, and — only when the branch is behind — calls `compare/{merge_base}...{base}` for the files the base branch changed in the meantime. That second call has no `[]` fallback: `base_files` is set to an empty-string sentinel first and reset to it when the call fails, so partial output from a call that died midway is discarded and a missing merge base takes the same path.
- `.github/scripts/pr-health-labels.py`: added `drift_state()`, `parse_behind_by()` and `parse_base_files()`, plus a `--field checks|drift` switch with `--behind-by` / `--base-files`. A PR is `stale` when it is behind **and** the base moved on files the PR also changes, `current` otherwise, and `unknown` when either comparison call failed. `parse_base_files()` maps the sentinel, invalid JSON and any non-list payload to `None`, which is what makes a failed base-file comparison `unknown` rather than an empty set of moved files; a successful comparison that returns zero files still classifies as `current`. `unknown` leaves the label untouched, mirroring the existing `mergeStateStatus == UNKNOWN` guard, so a transient API error never clears a real label. An explicit `mergeStateStatus == BEHIND` is still honoured, so nothing regresses if strict checks are enabled later.
- The same `stale` verdict now holds back `status: ready for review`, replacing the old `BEHIND` condition.
- File overlap is the deliberate narrowing: plain "behind by at least one commit" matches 18 of 20 open PRs on a base branch this busy, which would make the label noise. Overlap targets the case that hides a semantic merge conflict from per-PR checks, which is what #3066 describes.
- Cost is 1-2 extra API calls per PR, so at most ~200 per scheduled run at the existing 100-PR cap.
- Added unit tests for the new logic and workflow-contract assertions, including a regression that runs the extracted `base_files` shell block under `bash` with `gh` stubbed to print partial output and exit 1, asserts the sentinel survives, feeds it to the script for an `unknown` verdict, and asserts the drift dispatch has no `else` so `unknown` neither adds nor removes the label.

## Testing

### Test Output

```text
$ python -m pytest scripts/tests/test_pr_health_labels.py scripts/tests/test_pr_health_workflow.py -q
scripts\tests\test_pr_health_labels.py .............                     [ 65%]
scripts\tests\test_pr_health_workflow.py .......                         [100%]
============================= 20 passed in 0.10s ==============================

$ ruff check .github/scripts/pr-health-labels.py scripts/tests/
All checks passed!

$ ruff format --check .github/scripts/pr-health-labels.py scripts/tests/test_pr_health_labels.py scripts/tests/test_pr_health_workflow.py
3 files already formatted
```

## Real Behavior Proof

- Environment: local checkout of the branch at 738f984d, Windows 11, Python 3.13, `gh` 2.x authenticated against `headroomlabs-ai/headroom`
- Exact command / steps: replayed the new workflow steps against the 25 most recent open PRs — for each one, `gh pr view --json baseRefName,files,headRefOid,isDraft,labels,mergeStateStatus,reviewDecision,statusCheckRollup`, then `gh api repos/headroomlabs-ai/headroom/compare/{base}...{head}` and `compare/{merge_base}...{base}`, feeding the results to `python .github/scripts/pr-health-labels.py --state-json ... --field drift --behind-by ... --base-files ...`
- Observed result: review follow-up on 2026-08-27, replayed #3192, #3180, #3177 and this PR with a successful base-file comparison and again with a failed one. The successful verdicts are unchanged (`stale`, `stale`, `stale`, `current`); the failed comparison now returns `unknown` on all four, where the pre-fix script returned `current` and the label step would have removed `status: needs rebase`. Reverting the workflow block to the old `|| echo '[]'` fallback fails both new workflow tests (`assert 'partia[]' == ''`). Original sweep: 3 of 25 PRs classified `stale` — #3192, #3180 and #3177, each 13 commits behind `main` and each touching `headroom/proxy/server.py`, which `main` also changed in that window. The other 22 came back `current`, including PRs that are behind by 1 to 13 commits on unrelated files. Before this change none of the 25 reported `mergeStateStatus == BEHIND` and none carried `status: needs rebase`. Workflow YAML still parses via `yaml.safe_load`.
- Not tested: no run of the workflow on GitHub Actions (`pull_request_target` and the cron schedule can only fire from the default branch after merge), and no `actionlint`/`act` run locally as neither tool is installed on this machine — CI covers both.

## Runtime Rollout Safety

- Rollout-managed feature(s): none, this is CI-only automation and touches no runtime code path
- Minimum rollout channel: not applicable
- Stable/default behavior changed: no product behavior changes; the only observable change is which PRs receive the `status: needs rebase` label
- Kill switch / disable path: revert this commit, or delete the label — the `label` job keeps working, and every `gh pr edit` failure is already tolerated
- Qualification impact: none
- Rollback path: `git revert` of this commit restores the previous `mergeStateStatus == BEHIND` condition

## Review Readiness

## Checklist

## Additional Notes

- Documentation checkbox is unchecked because the label semantics are described in the label description itself (updated by this PR) and no docs page covers the PR-health labels.
- `mypy headroom` is unchecked because nothing under `headroom/` changed; the touched script lives in `.github/scripts/`.
- The overlap comparison uses the files `gh pr view --json files` returns, so a PR with an unusually large file list could under-report overlap. That fails safe (no label) rather than mislabelling.
- This PR intentionally does not close #3066: the acceptance criteria there are strict status checks or a merge queue on `main`, which need admin rights on the repo.


JerrettDavis (changes requested):
The direct drift comparison is the right replacement for `mergeStateStatus == BEHIND`, but the second API failure currently clears a potentially valid label.

When `compare/$merge_base...$base_ref` fails, the workflow executes `|| echo '[]'`. `drift_state()` then sees a positive `behind_by` with no moved files and returns `current`, so the `elif` removes `status: needs rebase`. That contradicts the PR's stated fail-safe behavior (transient comparison failures leave the label untouched); only failure of the first compare currently becomes `unknown`.

Please preserve an explicit unknown/error sentinel for the base-files comparison and have drift classification return `unknown` when the branch is behind but that comparison was unavailable. Add a workflow/logic regression proving a failed second call neither adds nor removes the label. Empty files from a successful comparison can remain `current`. The rest of the overlap-based narrowing and ready-label gating looks coherent, and CI is green on this head.

Parideboy:
Fixed in 738f984d.

The base-file comparison now uses an empty-string sentinel instead of falling back to `[]`: the workflow sets `base_files=''` before the call and resets to it when the call fails, which also discards partial output from a call that died midway, and a missing `merge_base` takes the same path rather than silently yielding an empty list. `parse_base_files()` maps that sentinel, invalid JSON and any non-list payload to `None`, and `drift_state()` returns `unknown` when the branch is behind but that comparison never answered, so neither branch of the label dispatch runs. A successful comparison that returns zero files still classifies as `current`, as you said it could.

Two regressions pin it:

- `test_failed_base_file_comparison_does_not_clear_the_rebase_label` extracts the `base_files` shell block from the workflow and runs it under `bash` with `gh` stubbed as a function that prints partial output and exits 1, asserts `base_files` ends empty, feeds that to the real script and asserts `unknown`, then asserts the drift dispatch is `if`/`elif` with no `else`.
- `test_drift_state_is_unknown_when_the_base_file_comparison_failed` covers the classifier directly, with `test_drift_state_is_current_when_the_base_moved_no_files` pinning the successful-but-empty case.

Reverting the workflow block to the old `|| echo '[]'` fallback fails both new workflow tests (`assert 'partia[]' == ''`).

Replayed against #3192, #3180, #3177 and this PR: the successful-comparison verdicts are unchanged (`stale`, `stale`, `stale`, `current`), and a failed second comparison returns `unknown` on all four, where the previous script returned `current`.

JerrettDavis (approved):
Re-reviewed the latest head after the fail-safe follow-up. A failed or partial base-file comparison now yields unknown and leaves an existing rebase label untouched; a successful empty comparison remains distinguishable as current. The workflow dispatch has no destructive unknown branch, and the focused workflow/helper suite passes locally (20 tests). I also removed the single trailing-whitespace defect from the contributor branch without changing behavior. No blocking issues remain.

PR #3236, opened 2026-08-24
fix(memory/graph): apply relation_type filter to both sides of a BOTH subgraph query

abhay-codes07:
## Description

`SQLiteGraphStore.query_subgraph` builds the relationship query for a `direction=BOTH` traversal without parenthesizing the `OR`, then appends the `relation_types` filter:

```python
else:  # BOTH
    rel_query = "SELECT * FROM relationships WHERE source_id = ? OR target_id = ?"
    rel_params = [current_id, current_id]

if relation_types is not None and len(relation_types) > 0:
    rel_query += f" AND relation_type IN ({placeholders})"
```

The resulting SQL is `WHERE source_id = ? OR target_id = ? AND relation_type IN (...)`. In SQL, `AND` binds tighter than `OR`, so this parses as:

```sql
WHERE source_id = ?  OR  (target_id = ? AND relation_type IN (...))
```

The type filter therefore applies **only to the incoming (`target_id`) side**. Every **outgoing** edge from the current node is returned regardless of type, so a `direction=BOTH` subgraph traversal with a `relation_types` filter pulls in relationships — and the entities they point to — that should have been excluded. `get_relationships` already groups this correctly (`"(source_id = ? OR target_id = ?)"`); `query_subgraph` was the outlier.

Reproduction (X has `likes->Liked` and `dislikes->Disliked`):

```python
subgraph = await store.query_subgraph([X], max_hops=1,
                                       direction=RelationshipDirection.BOTH,
                                       relation_types=["likes"])
# BEFORE: subgraph.relationships includes the 'dislikes' edge; entities include 'Disliked'
# AFTER:  only the 'likes' edge and {'X', 'Liked'}
```

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/memory/adapters/sqlite_graph.py`: parenthesize the `BOTH` predicate — `WHERE (source_id = ? OR target_id = ?)` — so the appended `AND relation_type IN (...)` filters both directions. A comment records the precedence trap and points at the already-correct `get_relationships`.
- `tests/test_sqlite_graph_store.py`: added `test_query_subgraph_both_direction_respects_relation_type_filter` — a node with two outgoing edges of different types, queried `BOTH` + `relation_types=["likes"]`, must not return the other-typed edge or its entity. Red→green: it fails on the un-parenthesized query and passes with the fix.

## Testing

### Test Output

```text
tests/test_sqlite_graph_store.py::...both_direction_respects_relation_type_filter  ->  passes with fix, FAILS without it (verified via git stash)
uvx ruff@0.16.2 check headroom/memory/adapters/sqlite_graph.py tests/test_sqlite_graph_store.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/memory/adapters/sqlite_graph.py  ->  Success: no issues found in 1 source file
```

(Note: this test file has pre-existing, unrelated failures on `main` in `TestSQLiteGraphStoreMemoryTrackerIntegration` — present before and after this change; my new test and the traversal suite are unaffected.)

## Real Behavior Proof

- Exact command / steps: built a store with X --likes--> Liked and X --dislikes--> Disliked, then `query_subgraph([X], max_hops=1, direction=BOTH, relation_types=["likes"])`. Before the fix the returned relationships contained the `dislikes` edge and the entity set contained `Disliked`; after the fix only the `likes` edge and `{X, Liked}` are returned. Confirmed at the raw-SQL level too: `source_id = ? OR target_id = ? AND relation_type IN (?)` returns the wrong-typed row, `(source_id = ? OR target_id = ?) AND relation_type IN (?)` does not.
- Observed result: the type filter now applies to both directions; unfiltered and single-direction queries are unchanged.
- Not tested: no change to `OUTGOING`/`INCOMING` branches (single-condition, unaffected); their existing tests still pass.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This is a SQLite graph-store query in the memory subsystem, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: only the incorrect result is corrected — a `BOTH` subgraph query with a `relation_types` filter now excludes outgoing edges of other types, as intended. Queries without a type filter, and `OUTGOING`/`INCOMING` queries, are unchanged.
- Qualification impact: none.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/memory/adapters/sqlite_graph.py
+++ b/headroom/memory/adapters/sqlite_graph.py
@@ -455,8 +455,14 @@ class SQLiteGraphStore:
                         rel_query = "SELECT * FROM relationships WHERE target_id = ?"
                         rel_params = [current_id]
                     else:  # BOTH
+                        # The OR must be parenthesized: a later ``AND relation_type
+                        # IN (...)`` binds tighter than ``OR`` in SQL, so without
+                        # the parens the type filter would apply only to the
+                        # ``target_id`` (incoming) side, letting outgoing edges of
+                        # every type leak into the subgraph. (``get_relationships``
+                        # already groups this correctly.)
                         rel_query = (
-                            "SELECT * FROM relationships WHERE source_id = ? OR target_id = ?"
+                            "SELECT * FROM relationships WHERE (source_id = ? OR target_id = ?)"
                         )
                         rel_params = [current_id, current_id]
 
--- a/tests/test_sqlite_graph_store.py
+++ b/tests/test_sqlite_graph_store.py
@@ -413,6 +413,46 @@ class TestSQLiteGraphStoreTraversal:
         assert "A" in entity_names
 
     @pytest.mark.asyncio
+    async def test_query_subgraph_both_direction_respects_relation_type_filter(self, tmp_path):
+        """A relation_types filter must apply to BOTH the outgoing and incoming
+        sides of a ``direction=BOTH`` traversal.
+
+        Regression: the BOTH query was ``WHERE source_id = ? OR target_id = ?``
+        without parentheses, then ``AND relation_type IN (...)`` was appended.
+        SQL binds ``AND`` tighter than ``OR``, so the type filter applied only to
+        the incoming (target_id) side, letting outgoing edges of every type leak
+        into the subgraph.
+        """
+        store = SQLiteGraphStore(db_path=str(tmp_path / "graph.db"))
+        x = Entity(user_id="u", name="X", entity_type="node")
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Confirmed the precedence bug and fix: parenthesizing the bidirectional endpoint predicate makes the subsequently appended relation-type predicate apply to both incoming and outgoing edges. The regression test exercises the real store/query path and verifies both edge and entity exclusion, while the unfiltered and single-direction branches remain unchanged. Full CI is green on the current head.

PR #3240, opened 2026-08-24
fix(memory/sqlite): match numeric and boolean metadata filters, not just strings

abhay-codes07:
## Description

`SQLiteMemoryStore._build_query_conditions` binds a metadata filter value as JSON text for anything that isn't a `str`:

```python
conditions.append(f"json_extract(metadata, '$.{key}') = ?")
params.append(json.dumps(value) if not isinstance(value, str) else value)
```

But `json_extract` returns a **native** SQLite value (INTEGER / REAL / TEXT). Comparing that against `json.dumps(value)` — `"5"`, `"true"`, `"1.5"` — never matches, because SQLite does not equate `5 = '5'`. So a metadata filter on any numeric or boolean value silently returns **nothing**; only string filters work.

Reproduction (memory stored with `metadata={"priority": 5, "archived": True, "score": 1.5, "tag": "x"}`):

```python
await store.query(MemoryFilter(user_id="u", metadata_filters={"priority": 5}))    # -> [] (want 1)
await store.query(MemoryFilter(user_id="u", metadata_filters={"archived": True}))  # -> [] (want 1)
await store.query(MemoryFilter(user_id="u", metadata_filters={"score": 1.5}))      # -> [] (want 1)
await store.query(MemoryFilter(user_id="u", metadata_filters={"tag": "x"}))        # -> [1]  (only strings work)
```

This affects both `query()` and `count()` (they share the builder), on the public `MemoryFilter(metadata_filters=...)` API.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/memory/adapters/sqlite.py`: bind the **native** Python value for scalar metadata filters (`str` / `int` / `float`; `bool` is an `int` subclass, and a JSON `true` extracts to `1`, which `True` binds to). A non-scalar value (`dict`/`list`) — not a bindable SQLite type — keeps the JSON-text comparison so the query does not raise. The JSON-path key validation (injection guard) is unchanged.
- `tests/test_memory/test_query_conditions.py`: added `test_metadata_scalar_filters_bind_native_values` (int/bool/float/str bind natively, not as `"5"`/`"true"`) and `test_metadata_non_scalar_filter_keeps_json_text` (a list value keeps `json.dumps`).

## Testing

### Test Output

```text
tests/test_memory/test_query_conditions.py  ->  5 passed in 1.27s
(the scalar test FAILS on pre-fix code, verified via git stash)
uvx ruff@0.16.2 check headroom/memory/adapters/sqlite.py tests/test_memory/test_query_conditions.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/memory/adapters/sqlite.py  ->  Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: saved a memory with `metadata={"priority":5,"archived":True,"score":1.5,"tag":"x"}` and queried each field. Before the fix, `priority=5`, `archived=True`, and `score=1.5` each returned 0 rows while `tag="x"` returned 1; after the fix all four return 1, a non-matching `priority=6` returns 0, and a dict-valued filter runs without raising.
- Observed result: numeric/boolean/string metadata filters all match correctly; non-matching values still exclude; non-scalar values do not crash.
- Not tested: no change to the JSON-path key-injection validation (covered by existing security tests) — only the bound value is corrected.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This is a SQLite memory-store query builder, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: string filters are unchanged; numeric/boolean filters change from silently returning nothing to correctly matching. No schema or API change.
- Qualification impact: none.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/memory/adapters/sqlite.py
+++ b/headroom/memory/adapters/sqlite.py
@@ -550,9 +550,22 @@ class SQLiteMemoryStore:
                 # while blocking malicious attempts like "'] OR 1=1--"
                 if not _validate_metadata_key(key):
                     continue
-                # Use JSON extraction for metadata filtering
+                # Use JSON extraction for metadata filtering.
+                #
+                # ``json_extract`` returns a NATIVE SQLite value (INTEGER / REAL /
+                # TEXT), so a scalar filter must bind the native Python value.
+                # Binding ``json.dumps(value)`` instead compared the numeric/boolean
+                # column against its text form ("5", "true") — and SQLite never
+                # equates ``5 = '5'`` — so an int/float/bool metadata filter matched
+                # nothing. ``bool`` is a subclass of ``int``, so it is covered here
+                # (a JSON ``true`` extracts to 1, and Python ``True`` binds to 1).
+                # A non-scalar value (dict/list) is not a bindable SQLite type, so it
+                # keeps the JSON-text comparison rather than raising.
                 conditions.append(f"json_extract(metadata, '$.{key}') = ?")
-                params.append(json.dumps(value) if not isinstance(value, str) else value)
+                if isinstance(value, (str, int, float)):
+                    params.append(value)
+                else:
+                    params.append(json.dumps(value))
 
         return conditions, params
 
--- a/tests/test_memory/test_query_conditions.py
+++ b/tests/test_memory/test_query_conditions.py
@@ -38,3 +38,37 @@ def test_agent_id_only_still_applied():
 
     assert "agent_id = ?" in conditions
     assert "turn_id = ?" not in conditions
+
+
+def test_metadata_scalar_filters_bind_native_values():
+    """A metadata filter on a numeric/boolean value must bind the NATIVE value.
+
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Verified against SQLite's JSON1 execution semantics, not only the parameter-building assertions: JSON integers, booleans, and reals are extracted as native INTEGER/REAL values and now match the native bound parameters as intended. The change is narrowly scoped, preserves key validation, and the full CI matrix is green.

Non-blocking follow-up: `MemoryFilter.metadata_filters` is typed as `dict[str, Any]`, but arrays/objects (and JSON null) still do not reliably match with the existing equality strategy; for example, `json.dumps(["a", "b"])` includes whitespace while `json_extract` returns compact JSON. That is pre-existing behavior rather than a regression in this scalar fix, but the new non-scalar test only proves parameter construction, not query matching. A future change should either define metadata filters as scalar-only or add execution-level semantics/tests for compound values and null.

Issue #21, opened 2026-02-01
High Memory Usage (5-9GB RAM) - Unbounded In-Memory Stores & Large ML Models


prakersh:
# High Memory Usage (5-9GB RAM) - Unbounded In-Memory Stores & Large ML Models

## Summary

Users are experiencing high memory usage when running headroom proxy:
- **5GB RAM** without `--memory` flag
- **9GB RAM** with `--memory` flag enabled

Investigation reveals multiple contributing factors: large ML models loaded by default, and unbounded in-memory stores when memory features are enabled.

## Environment

- Headroom version: 0.3.0
- User A: 5GB RAM usage (without `--memory` flag)
- User B: 9GB RAM usage (with `--memory` flag)

### Memory Usage WITHOUT `--memory` Flag (~5GB)

The base memory usage comes from:

**1. LLMLingua-2 Model** (loaded by default)
- Model: `microsoft/llmlingua-2-xlm-roberta-large-meetingbank`
- Parameters: 560M
- Memory: ~1.2-2.4GB
- Location: `headroom/transforms/llmlingua_compressor.py:202`

**2. CompressionStore** (`headroom/cache/compression_store.py`)
- Stores original + compressed content strings for CCR
- Limit: 1000 entries, 5-minute TTL
- Tracks `bytes_used` in stats
- Status: **Bounded** ✅

**3. SemanticCache** (`headroom/proxy/server.py:315`)
- Caches full API response bodies
- Limit: 1000 entries, 1-hour TTL
- Status: **Bounded** ✅

**4. BatchContextStore** (`headroom/ccr/batch_store.py`)
- Stores full message history + tools for every batch request
- Limit: 10,000 contexts, 24-hour TTL
- Risk: 10K contexts × full conversation history = potentially very large
- Status: **Bounded but high limit** ⚠️

**5. RequestLogger** (`headroom/proxy/server.py`)
- Uses deque with `maxlen=10000`
- Status: **Bounded** ✅

### Additional Memory WITH `--memory` Flag (+4GB, total ~9GB)

When `--memory` flag is enabled, these additional stores are activated:

**6. InMemoryGraphStore** (`headroom/memory/adapters/graph.py`)
- Stores entities and relationships for the knowledge graph
- Limit: **None** - grows indefinitely
- Status: **Unbounded** ❌

**7. HNSWVectorIndex** (`headroom/memory/adapters/hnsw.py`)
- Stores embeddings, metadata, and content strings for vector search
[...]

PR #3252, opened 2026-08-25
fix(telemetry): release TOIN lock before auto-save disk write

inix-x:
## Description

TOIN's periodic auto-save held the instance `RLock` across the entire storage write. `_maybe_auto_save` called `self.save()` from inside `with self._lock:`, and because the lock is reentrant, `save()`'s own two lock acquisitions ("serialize under lock but write outside lock to prevent blocking other threads during slow file I/O", its own docstring) became no-ops. The backend write — full JSON serialization plus temp-file and atomic rename — therefore ran with the lock fully held, blocking every concurrent `record_compression` / `record_retrieval` call (the request-time API used by content routing and smart compression) for the whole file write.

Releasing the call from under the lock exposed two races the held-lock write had accidentally prevented, both fixed here:

1. `export_patterns()` nests live pattern containers (`query_pattern_frequency`, `strategy_success_rates`, `common_query_patterns`, …), so dumping them outside the lock raced concurrent mutators with `RuntimeError: dictionary changed size during iteration`, silently deferring saves via the broad failure log. The locked step now snapshots through `copy.deepcopy`.
2. A commit landing between snapshot and finalize was marked persisted without ever reaching disk (dirty cleared unconditionally; there is no shutdown flusher), stranding data until process exit. `save()` now reads a mutation counter alongside its snapshot and finalizes only when unchanged; otherwise it takes one bounded extra pass so late commits are persisted instead of lost.

Closes nothing on file — found by local code review of the telemetry lock paths (#21 in my contribution notes).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/telemetry/toin.py`:
  - `_maybe_auto_save`: eligibility check stays under the lock; `save()` is now called outside it.
  - `save()`: deep-copy snapshot under the lock, write outside, then conditional finalize gated on a `_mutations` counter bumped at every `_dirty = True` site (all mutators already hold the lock). One bounded re-snapshot pass (`_SAVE_PASSES = 2`) covers a commit racing the write; exhaustion leaves `_dirty` armed for the next trigger.
  - Docstrings corrected — they previously documented the reentrancy trap as intentional.
- `tests/test_toin_fixes.py`:
  - `TestAutoSaveLockScope::test_auto_save_writes_outside_the_instance_lock` — a fake backend probes whether the instance lock is acquirable from a **second thread** during the write (a same-thread acquire cannot detect an RLock held reentrantly).
  - `TestAutoSaveLockScope::test_mutation_during_write_is_persisted_by_the_next_pass` — a blocked first write, a real `record_compression` committed mid-write, asserting the convergence pass rewrites the snapshot including the late commit and clears the dirty flag only then.

## Testing

### Test Output

```text
$ python -m pytest tests/test_toin_fixes.py -q
15 passed, 8 skipped in 1.81s

$ ruff check headroom/telemetry/toin.py tests/test_toin_fixes.py
All checks passed!

$ ruff format --check headroom/telemetry/toin.py tests/test_toin_fixes.py
2 files already formatted

$ mypy headroom/telemetry/toin.py

Fail-before evidence (fix reverted to upstream/main, bytecode caches cleared):
$ python -m pytest "tests/test_toin_fixes.py::TestAutoSaveLockScope" -q
FAILED test_auto_save_writes_outside_the_instance_lock      # lock held during write
FAILED test_mutation_during_write_is_persisted_by_the_next_pass
    assert len(writes) == 1                                  # mid-write commit finalized away
```

The remaining failures in `tests/test_stateless_toin.py` (2) and `tests/test_toin_observation_only.py` (6) reproduce byte-for-byte on clean upstream/main `6262c28a` with this branch's changes stashed — pre-existing, unrelated to this patch.

## Real Behavior Proof

- Environment: macOS (arm64), CPython 3.13.13, project venv; branch based on upstream/main `6262c28a`.
- Exact command / steps: with the three touched files reverted to upstream/main and `__pycache__` cleared: `python -m pytest "tests/test_toin_fixes.py::TestAutoSaveLockScope" -q` (fail-before); re-applied the commits and re-ran the same command plus `ruff check`, `ruff format --check`, `mypy headroom/telemetry/toin.py`. Additional adversarial probes during review drove a commit mid-write through a blocking fake backend, injected a failure into a second pass, and raced two overlapping savers.
- Observed result: fail-before — the lock-scope test records `lock_free_during_write=False` (probe thread cannot acquire while the writer holds the RLock) and the convergence test records one write with `_dirty` cleared over an unwritten mid-write commit. After the fix — both pass (`17 passed`), the blocked-backend probe shows the late commit persisted by the second pass, a failing second pass leaves `dirty=True` with the save clock unadvanced and the structured `toin_save_failed` warning contained, and two overlapping savers converge on the newest state before returning.
- Not tested: Windows atomic-rename behavior of the backend (pre-existing coverage area); the Rust proxy path (this module is Python-only).

## Runtime Rollout Safety

- Rollout-managed feature(s): none changed — TOIN persistence behaves identically when uncontended.
- Stable/default behavior changed: timing only — request threads no longer block behind the periodic flush; on-disk format is untouched.
- Kill switch / disable path: unchanged (`auto_save_interval=0` disables auto-save).
- Qualification impact: none.
- Rollback path: revert the two commits; prior behavior restores (with the blocking write and the races it masked).

## Review Readiness

## Checklist

- [x] I have performed a self-review of my own code

### DIFF
--- a/headroom/telemetry/toin.py
+++ b/headroom/telemetry/toin.py
@@ -60,6 +60,7 @@ detection).
 
 from __future__ import annotations
 
+import copy
 import hashlib
 import json
 import logging
@@ -432,6 +433,10 @@ class ToolIntelligenceNetwork:
     Thread-safe for concurrent access.
     """
 
+    # Bounded re-snapshot passes when a mutation races a backend write; leftovers
+    # stay dirty and the next record_* trigger retries, so this only caps I/O.
+    _SAVE_PASSES = 2
+
     # ── Deprecation warning de-dupe (PR-B5) ───────────────────────────────
     # `get_recommendation` is retired as a per-request mutator. We emit
     # `DeprecationWarning` once per process; if every call warned, busy
@@ -482,6 +487,10 @@ class ToolIntelligenceNetwork:
         # Tracking
         self._last_save_time = time.time()
         self._dirty = False
+        # Bumped alongside every _dirty=True so save() can tell whether new
+        # data landed while its snapshot was being written (all mutators hold
+        # the lock, so the counter needs no lock of its own).
+        self._mutations = 0
 
         # Load existing data from backend
         if self._backend is not None:
@@ -708,6 +717,7 @@ class ToolIntelligenceNetwork:
             pattern.last_updated = time.time()
             pattern.confidence = self._calculate_confidence(pattern)
             self._dirty = True
+            self._mutations += 1
             self._prune_patterns_locked()
 
         # Auto-save if needed (outside lock)
@@ -1015,6 +1025,7 @@ class ToolIntelligenceNetwork:
 
             pattern.last_updated = time.time()
             self._dirty = True
+            self._mutations += 1
             self._prune_patterns_locked()
 
         self._maybe_auto_save()
@@ -1319,6 +1330,7 @@ class ToolIntelligenceNetwork:
 
             self._prune_patterns_locked()
             self._dirty = True
+            self._mutations += 1
 
     def _merge_patterns(self, existing: ToolPattern, imported: ToolPattern) -> None:
         """Merge imported pattern into existing."""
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest head adversarially, including the lock-scope change, mutation-counter coverage, bounded convergence behavior, and concurrent-save ordering. The snapshot/finalize protocol preserves newly recorded data while moving backend I/O out of the request-path lock. Focused TestAutoSaveLockScope coverage passes locally on Windows/Python 3.13 and Ruff is clean. The current Agno failure is in untouched integration code (agno.models.metrics is absent from the installed dependency), so it is unrelated to this change.

JerrettDavis (approved):
Re-reviewed exact head d03e8f209a3e0c69d1368308657dff3c29c868c6. The added failure and pass-exhaustion regressions confirm the two remaining safety properties: failed storage writes leave the dirty clock/state retryable, and bounded convergence never falsely marks continuously-mutated state clean. The unrelated Agno dependency failure remains outside this diff. No blocker found.

PR #3253, opened 2026-08-25
feat(beacon): schema v2 — routing/compression training signal, gzip transport

chopratejas:
## What

Nine new sections on the session heartbeat — `quality`, `hist`, `shapes`, `cache`, `trajectory`, `clients`, `errors`, `strata`, `config` — plus a gzip transport. Everything is a counter, a ratio, or a slug from a fixed list in the source. No prompts, code, file paths, tool names, project names, or error text can appear.

## No regression in v1

Every existing field keeps its name, type, and meaning. `tests/test_beacon_schema_v2.py` locks the v1 surface with literal key lists and a shape-stability invariant, so a v1 field that silently changes type fails there rather than in the corpus six weeks later.

The v2 section is built in `_payload_v2()` behind `try/except`, and `_fold_extra()` likewise. A bug in new code can cost the new fields and nothing else. The first version built one dict literal — any raise dropped `tokens.saved` along with it.

## The STRUCT/MAP flip

Four fields ship as lists of records rather than objects keyed by data (`clients`, `config`, `strata`, `waste`). DuckDB infers a JSON object as `STRUCT` when the keys are few and identifier-shaped, and as `MAP` otherwise — so a key set that grows with the fleet flips type mid-corpus and `BINDER ERROR`s the reader.

Caught against the real 369k-session corpus, not in theory: `clients` inferred `STRUCT(claude_code, codex)` at two harnesses and `MAP(VARCHAR, BIGINT)` at 22, and `clients.cursor` errors on one of them.

## Transport

Body is gzipped at level 6 (~6x). The receiver sniffs magic bytes rather than trusting `Content-Encoding`, enforces a decompression-bomb limit *during* the stream read, and leaves the plain path byte-identical for the existing fleet.

A `400`/`415` disables gzip for the process and retries once. A `413` does not — an uncompressed retry is 6x larger and guaranteed to fail again.

At current rates (~235k heartbeats/day): v1 today 892 MB/day, v2 uncompressed 1,802 MB/day, **v2 gzipped 295 MB/day** — a net reduction against what we send now.

## `headroom telemetry`

Prints the real payload, built by the same `_Session.payload()` the beacon uses. \"Go read worker.js\" answers what happens to the data; it does not answer what leaves your machine.

Uses `read_install_id()` and `resource_attributes(create_install_id=False)` — the default mints and persists the id, so building the report would itself be the act that identifies the machine. Filtering the id out of the *output* is not enough; by then the file exists.

## Receiver

Allowlist gains the nine keys. `MAX_BODY_BYTES` 64KB → 256KB — that figure dated from \"a beacon event is ~2KB\", and worst-case v2 is 81,965 B uncompressed.

## Verification

Against the production corpus (369,015 sessions): **14/14** `beacon.sh` reports pass, **18/18** dashboard queries pass, v1 totals byte-identical at `273,495,971,478` tokens saved.

`sample-event.json` regenerated as a real v2 payload. The old v1-only version would have passed the post-deploy smoke test whether or not the new allowlist landed; the new one distinguishes the cases 9/9 vs 0/9.

Suite: 11,655 passed, 1 failed — `test_mcp_reconcile.py::test_ordinary_install_does_not_adopt_serena`, which reproduces identically on clean `main`.

## Not verified here

`node test-rollup.mjs` and `npx wrangler dev` have not been run — no `node` in the authoring environment. `DecompressionStream` is the one path with no fallback and has never been executed. **Both are required before deploy.**

## Deploy order

**Worker first, client second.** The worker is backward-compatible with the v1 fleet (allowlist additive, plain path byte-identical, cap only raised). The reverse is not: the worker strips unknown keys at ingest, so v2 data arriving before the allowlist lands is gone, not recoverable by re-rolling. The corpus already carries 6 such rows from a dev proxy.

Merging this PR does not ship the client — releases are separate commits.

### DIFF
--- a/deploy/beacon/sample-event.json
+++ b/deploy/beacon/sample-event.json
@@ -12,7 +12,25 @@
           {
             "key": "service.version",
             "value": {
-              "stringValue": "0.36.6-dev"
+              "stringValue": "0.37.0"
+            }
+          },
+          {
+            "key": "headroom.install_id",
+            "value": {
+              "stringValue": "00000000000000000000000000000000"
+            }
+          },
+          {
+            "key": "headroom.install_mode",
+            "value": {
+              "stringValue": "pip"
+            }
+          },
+          {
+            "key": "headroom.stack",
+            "value": {
+              "stringValue": "proxy"
             }
           },
           {
@@ -26,18 +44,6 @@
             "value": {
               "stringValue": "arm64"
             }
-          },
-          {
-            "key": "headroom.install_id",
-            "value": {
-              "stringValue": "00000000000000000000000000000000"
-            }
-          },
-          {
-            "key": "headroom.stack",
-            "value": {
-              "stringValue": "proxy"
-            }
           }
         ]
       },
@@ -48,7 +54,7 @@
           },
           "logRecords": [
             {
-              "timeUnixNano": "1787677313932464896",
+              "timeUnixNano": "1787703693355874048",
               "body": {
                 "kvlistValue": {
                   "values": [
@@ -66,7 +72,7 @@
                             {
                               "key": "id",
                               "value": {
-                                "stringValue": "3d35a3b60609a37b"
+                                "stringValue": "4996e02d2b7cafe7"
                               }
                             },
                             {
@@ -78,13 +84,13 @@
                             {
                               "key": "duration_s",
                               "value": {
-                                "intValue": "225"
+                                "intValue": "135"
                               }
                             },
                             {
                               "key": "turns",
                               "value": {
-                                "intValue": "6"
+                                "intValue": "4"
                               }
                             },
                             {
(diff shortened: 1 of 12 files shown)


JerrettDavis (approved):
Reviewed the schema-v2 field provenance, bounded vocabularies/cardinality, configuration allowlist, structural tool-shape encoding, wire-size shedding, gzip fallback, and streaming decompression cap on the current implementation. I also ran the receiver suite directly with Node on the PR head (passes) and the two targeted Python modules: 56/60 passed locally; the remaining four were environment/setup failures (missing optional tomlkit and an unbuilt Windows native extension), while the repository CI test shards passed. The prior Agno failure was unrelated and fixed on main, so I updated this branch from main; exact-head CI is now rerunning. No blocking correctness or privacy issue found.

Issue #3205, opened 2026-08-22
`headroom wrap claude` sessions crash repeatedly (4x in one day); `headroom doctor`'s claude-routing check gives false negatives for `init`-based sessions


Alley007:
## Environment
- macOS (Darwin 25.5.0)
- Headroom v0.36.2, proxy running via `headroom init claude` (persistent-task preset, user scope, port 8787)
- Claude Code CLI v2.1.239, installed via `npm install -g @anthropic-ai/claude-code`
- Python 3.13

## Issue 1: `headroom wrap claude` sessions crash without warning

We're running several long-lived Claude Code CLI sessions in separate Terminal.app windows, each doing continuous multi-hour agentic work (content generation + REST API calls to an external site). Over one working session today:

- 3 separate sessions launched with `headroom wrap claude` crashed a total of **4 times**, each with no visible error message — the process just disappeared. In every case this happened mid-session, sometimes right after a task completed cleanly, sometimes with no apparent trigger.
- In contrast, **one session launched via the passive `headroom init claude` routing** (not `wrap`) has stayed up continuously for 14+ hours doing the same kind of work, with zero crashes.

I don't have stack traces or crash logs from the terminal at the moment of failure — from the outside, all I could observe was the process no longer running (no error dialog, no exit message captured). Happy to try to capture better diagnostics if there's a recommended way (e.g. wrapping with `2>&1 | tee crash.log`) — will do that on the next restart if useful for a report.

**Is this a known issue with `wrap` specifically?** Given `init`-based routing has been rock solid by comparison, it feels like something specific to how `wrap` spawns/supervises the child process, rather than proxy instability in general (the proxy itself has stayed up 14+ hours with no restarts per `headroom doctor`).

## Issue 2: `headroom doctor`'s `claude` check gives a false negative for `init`-routed sessions

For the stable long-running session mentioned above (routed via `headroom init claude`, not `wrap`), running `headroom doctor` reports:

But this is incorrect — direct process inspection shows the session genuinely is routed:
- `ps eww -p <pid>` on the actual running `claude` process shows `ANTHROPIC_BASE_URL=http://127.0.0.1:8787` set in its live environment.
- `mcp__headroom__*` tools (`headroom_compress`, `headroom_retrieve`, `headroom_stats`) are present and actively firing compression on that session's tool output.
- `headroom_stats` for that session shows real, ongoing savings (174+ requests, 164 compressed at time of check).

[...]

PR #3255, opened 2026-08-25
fix(doctor): recognize IPv6 loopback and case-insensitive proxy routing URLs

abhay-codes07:
## Description

`headroom doctor`'s routing check matches the configured `ANTHROPIC_BASE_URL` against:

```python
_LOOPBACK_URL_RE = re.compile(r"https?://(?:127\.0\.0\.1|localhost):(\d+)")
```

This misses two loopback URL forms that route correctly through the proxy, so `check_claude_routing` (and `check_codex_routing`, which uses the parallel `_CODEX_BASE_URL_RE`) reports a **correctly-routed** session as "not routed" and sends the user off to debug a working setup:

1. **IPv6 loopback literal** — `http://[::1]:8787` never matches, so an IPv6-configured route reads as unrouted.
2. **Case** — the regex is case-sensitive, but URL schemes and hostnames are not. `HTTP://Localhost:8787` or `http://LOCALHOST:8787` (valid) both fail to match.

```text
http://127.0.0.1:8787   -> MATCH
http://[::1]:8787       -> NO MATCH  -> doctor: "not routed" (false negative)
HTTP://Localhost:8787   -> NO MATCH  -> doctor: "not routed" (false negative)
```

This is the same false-negative class the routing check was already corrected for twice (#3205, #3213): a diagnostic that tells a routed user they are not routed is actively misleading.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/cli/doctor.py`: added the IPv6 loopback literal `[::1]` to the host alternation (shared `_LOOPBACK_HOST` pattern) and compiled both `_LOOPBACK_URL_RE` and `_CODEX_BASE_URL_RE` with `re.IGNORECASE`. Non-loopback URLs (e.g. `https://gateway.corp.example`) still correctly do not match, so the "points at X, not the proxy" WARN path is unchanged.
- `tests/test_cli_doctor.py`: added a parametrized `test_loopback_url_variants_pass` covering `localhost`, `[::1]`, `[::1]/v1`, and upper-cased scheme/host.

## Testing

### Test Output

```text
tests/test_cli_doctor.py  ->  95 passed in 1.19s
(4 of the 5 new variants FAIL on pre-fix code — [::1] and the upper-cased URLs — verified via git stash)
uvx ruff@0.16.2 check headroom/cli/doctor.py tests/test_cli_doctor.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/cli/doctor.py  ->  Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: with `ANTHROPIC_BASE_URL=http://[::1]:8787` (and the upper-cased variants) in a settings file, `check_claude_routing(path, 8787)` returned WARN "not routed" before the fix and PASS "routed" after; `http://127.0.0.1:8788` still WARNs on the port mismatch and `https://gateway.corp.example/v1` still WARNs as non-proxy.
- Observed result: IPv6 and case-variant loopback URLs now classify as routed; port-mismatch and non-loopback WARNs are unchanged.
- Not tested: no live `headroom doctor` invocation against a running proxy; the pure classifier `check_claude_routing` is exercised directly (it is what the command calls).

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This is the `headroom doctor` diagnostic classifier, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: only the false-negative is corrected — additional valid loopback forms now classify as routed. Port-mismatch, Ollama-collision, and non-loopback WARNs are unchanged, and no URL that was previously WARNed as non-proxy now passes except genuine loopback forms.
- Qualification impact: none.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/cli/doctor.py
+++ b/headroom/cli/doctor.py
@@ -48,8 +48,15 @@ WARN = "warn"
 FAIL = "fail"
 SKIP = "skip"
 
-_LOOPBACK_URL_RE = re.compile(r"https?://(?:127\.0\.0\.1|localhost):(\d+)")
-_CODEX_BASE_URL_RE = re.compile(r'base_url\s*=\s*"https?://(?:127\.0\.0\.1|localhost):(\d+)')
+# Loopback host forms a proxy URL can take: IPv4, the ``localhost`` alias, and
+# the IPv6 literal ``[::1]``. IGNORECASE because URL schemes and hostnames are
+# case-insensitive — a hand-set ``HTTP://Localhost`` or ``http://[::1]`` is a
+# valid loopback route, and failing to match it makes `doctor` report a
+# correctly-routed session as "not routed", sending the user to debug a
+# non-problem (the false-negative class of #3205 / #3213).
+_LOOPBACK_HOST = r"(?:127\.0\.0\.1|\[::1\]|localhost)"
+_LOOPBACK_URL_RE = re.compile(rf"https?://{_LOOPBACK_HOST}:(\d+)", re.IGNORECASE)
+_CODEX_BASE_URL_RE = re.compile(rf'base_url\s*=\s*"https?://{_LOOPBACK_HOST}:(\d+)', re.IGNORECASE)
 
 # Ollama's fixed default port. `ollama launch claude` writes
 # ``ANTHROPIC_BASE_URL=http://127.0.0.1:11434`` into the launched Claude Code
--- a/headroom/cli/doctor.py
+++ b/headroom/cli/doctor.py
@@ -20,6 +20,7 @@ from dataclasses import asdict, dataclass
 from datetime import datetime
 from pathlib import Path
 from typing import Any
+from urllib.parse import urlsplit
 
 import click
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The added legitimate variants are correct, but this is still a prefix match rather than a URL classification, so `doctor` can report a non-loopback destination as safely routed. For example, `ANTHROPIC_BASE_URL=http://localhost:8787@evil.example/v1` matches `_LOOPBACK_URL_RE`, captures 8787, and passes even though URL parsing makes `evil.example` the actual host. Similar suffix/userinfo lookalikes should remain WARNs. Since this PR is explicitly hardening routing classification, please parse the configured URL with `urllib.parse.urlsplit` and require an exact loopback hostname (`127.0.0.1`, `::1`, or case-normalized `localhost`) plus the expected port and acceptable scheme. For Codex TOML, extract the quoted `base_url` value, then feed that value through the same parser/helper rather than matching the URL prefix. Please add negative tests for userinfo and hostname suffix lookalikes alongside the new positive variants. The current `test-agno`/matrix reds appear inherited from the old base and can be cleared by updating from current main after the correctness fix.

abhay-codes07:
Thanks, this is a fair call, prefix matching was the wrong tool here. Reworked to classify by parsing:

- `_loopback_route_port(url)` now parses with `urllib.parse.urlsplit` and returns a port only when the scheme is http(s), there is **no userinfo**, the host is an **exact** loopback literal (`127.0.0.1`, `::1`, or case-normalized `localhost`), and an explicit valid port is present. `urlsplit` de-brackets and lowercases the host, so `[::1]` and `HTTP://Localhost` still classify. Your example `http://localhost:8787@evil.example/v1` parses to host `evil.example` and is refused (as is any URL carrying userinfo at all).
- Codex TOML: I extract the quoted `base_url` value from the `[model_providers.headroom]` block and feed it through the same helper instead of prefix-matching the URL. A non-loopback `base_url` now WARNs, and scoping to the Headroom block means an earlier provider's `base_url` can't be mistaken for the route.
- Added negative tests for userinfo and hostname-suffix lookalikes (`http://localhost:8787@evil.example/v1`, `http://localhost.evil.example:8787`, `http://127.0.0.1.evil.example:8787`) for both the Claude and Codex paths, alongside the existing positive variants; confirmed the userinfo cases fail against the old prefix-match code and pass now.
- Rebased/merged current `main` to clear the inherited `test-agno`/matrix reds.

JerrettDavis (changes requested):
The new parser correctly rejects lookalike and remote URLs when ase_url is present, but check_codex_routing() still returns PASS when the [model_providers.headroom] block has no matching ase_url at all. match is optional and the function falls through to the auth check and final PASS on None, so an incomplete/malformed block can still be reported as safely routed. Please make a missing or unparseable Headroom-block ase_url a WARN with the same rewrite guidance, and add a regression test for a marker block with no ase_url (plus a nonmatching quoting/malformed value if those are intended to be supported).

Issue #3251, opened 2026-08-25
[BUG] --memory on persistent-docker (macOS) 500s every request: sqlite3.OperationalError: unable to open database file


eldestar:
## Summary
`headroom deploy --providers auto --memory` on macOS (persistent-docker preset) starts and reports healthy, but every real request through the proxy returns a 500. The Anthropic handler's memory-context lookup fails with `sqlite3.OperationalError: unable to open database file` when opening the per-project SQLite memory store — even though the exact same file is trivially openable (read/write/WAL/CREATE TABLE) via a manual `sqlite3`/`python3` check run moments later, as the same container user, against the same path.

## Environment
- headroom-ai 0.36.5 (`uv tool install --python 3.13 "headroom-ai[all]"`)
- Docker Desktop 29.4.0 (client+server), macOS 15.7.7 (24G720), Apple Silicon (arm64)
- Image: `ghcr.io/headroomlabs-ai/headroom@sha256:071ca5343a473029336cbe6f262734e01d6698f584dcbf69d9fb8776775903e5`
- Deployment: `headroom deploy --providers auto --memory` → `persistent-docker` preset. `~/.headroom` bind-mounted to `/tmp/headroom-home/.headroom` in the container via Docker Desktop's default macOS bind-mount file sharing.

## Steps to reproduce
1. `headroom deploy --providers auto --memory` on macOS with Docker Desktop.
2. Confirm the deployment is `running`/`healthy` (`headroom install status --profile default`).
3. Send any real request through the proxy, e.g. `ANTHROPIC_BASE_URL=http://127.0.0.1:8787 claude -p "hi"`.
4. Request fails: `API Error: 500 Internal Server Error`.

## What I found
`docker logs` shows a full traceback ending in:
```
File ".../headroom/memory/adapters/sqlite.py", line 82, in _get_conn
    conn = sqlite3.connect(str(self.db_path))
sqlite3.OperationalError: unable to open database file
```
via `proxy/handlers/anthropic.py:handle_anthropic_messages` → `memory_handler.py` `_ensure_initialized`/`_do_init` → `LocalBackend._init_locked` → `HierarchicalMemory.create` → `create_memory_system` → `_create_store` → `SQLiteMemoryStore.__init__` → `_init_db`.

The target path (`.headroom/memories/projects/<hash>/memory.db`) already exists and is fully functional. From inside the running container, as the same uid:gid (502:20) the app runs as:
- `touch` a new file in that exact directory — succeeds
[...]

PR #3257, opened 2026-08-25
fix(proxy/memory): fail open when backend init errors instead of 500ing the request

abhay-codes07:
## Description

`MemoryHandler._ensure_initialized` documents a "fail-open contract — no exception propagates to request handlers", but only `asyncio.TimeoutError` and `asyncio.CancelledError` are caught:

```python
try:
    await asyncio.wait_for(_do_init(), timeout=STARTUP_INIT_TIMEOUT_SECONDS)
except asyncio.TimeoutError:
    ... # fail open
except asyncio.CancelledError:
    ... # reset + re-raise
```

Any *other* exception raised synchronously during backend init — e.g. `sqlite3.OperationalError: unable to open database file` — escapes both handlers, propagates up through `handle_anthropic_messages` → `memory_handler` → `LocalBackend._init_locked` → `SQLiteMemoryStore._init_db`, and **500s the whole request**.

This is exactly the failure in issue #3251: on the documented macOS Docker Desktop `persistent-docker --memory` deploy, the per-project SQLite memory store cannot be opened in WAL mode over the bind-mount (a known Docker Desktop/VirtioFS + SQLite interaction), and **every** request returns `500 Internal Server Error` — even though compression, caching, and tool-search savings don't need memory at all. An optional subsystem should never take down core proxy traffic; that is Ask #3 of the issue ("a memory-store open failure on the request path should degrade gracefully instead of 500ing").

The environmental root cause (bind-mount + WAL) is out of scope here; this fixes the graceful-degradation gap so `--memory` being unavailable disables memory context for the request rather than killing it.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/memory_handler.py`: added an `except Exception` branch to `_ensure_initialized` that mirrors the timeout branch — best-effort close the possibly-half-assigned backend, null it, leave `_initialized=False` so a later request retries, log an error, and **return** (fail open) instead of propagating. Ordering is preserved: `TimeoutError` (an `Exception`) is still caught first, and `CancelledError` (a `BaseException`, not an `Exception`) still falls through to its own re-raising branch.
- `tests/test_memory_handler_concurrent_init.py`: added `test_ensure_initialized_fails_open_on_backend_init_error` — a `LocalBackend` whose init raises `sqlite3.OperationalError`; asserts `_ensure_initialized` does not raise, cleans up the backend, stays uninitialized, and retries on the next call.

## Testing

### Test Output

```text
tests/test_memory_handler_concurrent_init.py  ->  9 passed in 1.80s
(the new test FAILS on pre-fix code — the OperationalError propagates — verified via git stash)
uvx ruff@0.16.2 check headroom/proxy/memory_handler.py tests/test_memory_handler_concurrent_init.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/proxy/memory_handler.py  ->  Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: monkeypatched `LocalBackend` so its `_ensure_initialized` raises `sqlite3.OperationalError("unable to open database file")` (the #3251 failure). Before the fix, `await handler._ensure_initialized()` re-raised the error (which on the real path becomes a request 500). After the fix, it returns cleanly with `_initialized=False`, `_backend=None`, the half-init backend `close()`d once, and a later call retries.
- Observed result: a backend init failure no longer propagates; memory is skipped for the request and retried later; timeout and cancellation behavior are unchanged.
- Not tested: no live macOS Docker Desktop repro of the bind-mount/WAL open failure (environment-specific); the fail-open path is exercised by injecting the same exception the deploy hits.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This is the memory-subsystem lazy-init on the proxy request path, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: only the failure path — a backend that cannot initialize now disables memory for the request instead of 500ing it. Successful init, the timeout branch, and the cancellation branch are unchanged.
- Kill switch / disable path: memory is already opt-in (`MemoryConfig.enabled`); this makes a broken backend behave like memory-off for affected requests.
- Qualification impact: none for healthy deployments; broken-memory deployments go from all-requests-500 to requests-served-without-memory.

## Review Readiness

## Checklist

## Additional Notes

Addresses Ask #3 of #3251 (graceful degradation). The environmental root cause (Docker Desktop macOS bind-mount + SQLite WAL) and Asks #1/#2 (repro / named-volume preset) are deployment concerns left out of this code fix.

### DIFF
--- a/headroom/proxy/memory_handler.py
+++ b/headroom/proxy/memory_handler.py
@@ -316,6 +316,27 @@ class MemoryHandler:
             self._initialized = False
             logger.info(f"Memory: backend initialization cancelled (backend={self.config.backend})")
             raise
+        except Exception as exc:
+            # Fail-open for ANY init failure, not just timeout. Memory is an
+            # optional subsystem: a backend that cannot open (e.g. a SQLite
+            # ``unable to open database file`` on a Docker Desktop macOS
+            # bind-mount, issue #3251) must NOT propagate and 500 the whole
+            # request — the docstring's fail-open contract has to hold here too.
+            # Null the possibly-half-assigned backend (same reasoning as the
+            # timeout branch) and leave ``_initialized=False`` so a later
+            # request can retry once the environment recovers.
+            existing_backend = self._backend
+            if existing_backend is not None:
+                await self._close_backend_instance(existing_backend, reason="init_error")
+            self._backend = None
+            self._initialized = False
+            logger.error(
+                "Memory: backend initialization failed (backend=%s); "
+                "serving requests without memory context. Subsequent requests will retry: %s",
+                self.config.backend,
+                exc,
+            )
+            return
 
     async def _init_backend_locked(self) -> None:
         """Actual backend-init body. Must be called with ``_init_lock`` held."""
--- a/tests/test_memory_handler_concurrent_init.py
+++ b/tests/test_memory_handler_concurrent_init.py
@@ -100,6 +100,53 @@ async def test_ensure_initialized_noop_when_disabled(tmp_path):
     assert handler._backend is None
 
 
+@pytest.mark.asyncio
+async def test_ensure_initialized_fails_open_on_backend_init_error(tmp_path, monkeypatch):
+    """A backend that cannot open must NOT propagate — memory is optional.
+
+    Regression (#3251): a SQLite ``unable to open database file`` on a Docker
+    Desktop macOS bind-mount escaped ``_ensure_initialized`` (which only caught
+    TimeoutError/CancelledError) and 500'd every request. It must fail open: log,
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed the fail-open path and refreshed head at 5c15b01c. Cancellation remains explicitly re-raised, timeouts retain their dedicated handling, partial backend state is closed and cleared, and _initialized remains false so recovery is possible without poisoning subsequent requests. The focused test verifies cleanup and retry, while current main removes the unrelated Agno failure. Newly triggered CI remains the readiness gate; no correctness blocker found.

Issue #3247, opened 2026-08-25
[BUG] VS Code Copilot Anthropic models fail through Headroom, GPT models work normally


ympaul:
## Description
 
When using `headroom wrap vscode` with GitHub Copilot in VS Code, GPT models work correctly but Anthropic models (Claude Sonnet / Claude Opus) fail with:
 
```text
Invalid bearer token
```
 
The issue only occurs when Headroom is enabled.
 
If Headroom is disabled, the same Claude models work normally through GitHub Copilot.
 
This appears to be specific to the Anthropic forwarding/authentication path.
 
<br>
 
## To Reproduce
 
 
1. Install Headroom:
 
```bash
pip install headroom-ai
```
 
2. Authenticate Copilot:
 
```bash
headroom copilot-auth login
```
 
3. Start the VS Code integration:
 
```bash
headroom wrap vscode
```
 
4. Open VS Code with GitHub Copilot Chat enabled.
 
5. Select a GPT model (GPT-5.x, GPT-4o, etc.) and send a message.
 
6. Observe that the GPT request succeeds.
 
7. Switch to Claude Sonnet or Claude Opus.
 
8. Send a message.
 
9. Observe the error:
 
```text
Invalid bearer token
```
 
<br>
 
## Expected Behavior
 
All GitHub Copilot models should function correctly through Headroom.
 
Claude Sonnet and Claude Opus should behave the same way as GPT models and return successful responses.
 
<br>
 
## Actual Behavior
 
GPT models succeed:
 
```text
200 OK
```
 
Claude Sonnet and Claude Opus fail:
 
```text
Sorry, your request failed. Please try again.
 
Reason: Invalid bearer token
```
 
VS Code displays the error immediately after the request is sent.
 
<br>
 
## Code Sample
 
```bash
headroom wrap vscode
```
 
Then use GitHub Copilot Chat in VS Code:
 
```text
GPT-5 ✅ Works
GPT-4o ✅ Works
Claude Sonnet ❌ Invalid bearer token
Claude Opus ❌ Invalid bearer token
```
 
<br>
 
## Error Output
 
VS Code error:
 
```text
Sorry, your request failed. Please try again.
 
Reason: Invalid bearer token
```
 
Relevant Headroom log excerpts:
 
```text
event=outbound_headers forwarder=anthropic_messages
```
 
```text
apply_copilot_api_auth: incoming token not suitable (kind=unknown), will replace
```
 
```text
event=outbound_request
[...]

Issue #3244, opened 2026-08-24
headroom wrap vscode routes to api.anthropic.com instead of api.githubcopilot.com on corporate SSL-inspection networks (Zscaler/Cisco Secure Access)


ekakuman:
## Environment
- macOS Apple Silicon
- headroom 0.36.2
- VS Code GitHub Copilot Business (corporate org)
- Corporate SSL inspection (Cisco Secure Access / Zscaler)

## Symptom
`headroom wrap vscode` fails immediately with 401. Every request goes to `api.anthropic.com` instead of `api.githubcopilot.com`.

From the proxy log:

event=outbound_request forwarder=streaming method=POST path=https://api.anthropic.com/v1/messages
Forwarding upstream streaming error status=401 url=https://api.anthropic.com/v1/messages

## Root cause identified
`apply_copilot_api_auth` logs "incoming token not suitable (kind=unknown), will replace" on every request. The token replacement fails because:

1. The Rust core's TLS stack (`_core.abi3.so`) cannot validate `api.github.com` through the corporate SSL inspection proxy (which presents a company-issued CA cert)
2. Python-layer env vars (`SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE`, `HEADROOM_TLS_STRICT=0`) do NOT reach the Rust binary
3. When token replacement fails, the proxy silently falls back to routing `/v1/messages` → `api.anthropic.com` instead of `api.githubcopilot.com`

## What was tried (all failed)
- `HEADROOM_TLS_STRICT=0` — documented fix, no effect on Rust TLS
- `SSL_CERT_FILE` + `REQUESTS_CA_BUNDLE` with full corporate CA bundle — works for Python httpx, not Rust core
- `GITHUB_COPILOT_TOKEN_EXCHANGE_URL=http://127.0.0.1:19500` pointing at a local proxy that successfully fetches tokens — Python layer reads this, Rust layer ignores it
- `GITHUB_COPILOT_API_URL=https://api.githubcopilot.com` — no effect on routing
- `sudo security add-trusted-cert -r trustRoot` — CA added to System keychain, no effect
- Confirmed: direct httpx calls to `api.github.com` and `api.githubcopilot.com` both succeed with the corporate CA bundle

## Evidence of value
Brief working windows (initial cached Copilot token before first refresh needed) saved **58.9k tokens / $2.90 equivalent** in a single day's sessions. The feature works — only the token refresh path is broken.

## Request
[...]

PR #3258, opened 2026-08-25
fix(providers): route Claude requests to Copilot when the OpenAI target is a Copilot host

abhay-codes07:
## Description

Through `headroom wrap vscode` / `wrap copilot --subscription`, GitHub Copilot **GPT** models work but **Claude** models fail with `Invalid bearer token` (issue #3247). The logs tell the story:

```text
# GPT — works:
event=outbound_request  path=https://api.githubcopilot.com/chat/completions  status=200

# Claude — fails:
event=outbound_request  path=https://api.anthropic.com/v1/messages           status=401
```

GitHub Copilot serves **both** surfaces from the same host: its OpenAI surface (`/chat/completions`, `/responses`) and its Anthropic surface for Claude models (`/v1/messages`) — `build_copilot_upstream_url` already documents and handles this. But `resolve_api_targets` resolves each provider target independently: when the Copilot flow points the **OpenAI** target at a Copilot host (so GPT works), the **Anthropic** target is left at its default `https://api.anthropic.com`. Claude-model requests are therefore forwarded to the real Anthropic API carrying the GitHub Copilot bearer, which Anthropic rejects with `Invalid bearer token`.

## Fix

In `resolve_api_targets`, when the resolved OpenAI target is a Copilot upstream host **and no explicit Anthropic target was configured**, default the Anthropic target to that same Copilot host. Claude requests then reach `https://api.githubcopilot.com/v1/messages` — the surface that serves them, where the Copilot bearer is valid. An explicit `ANTHROPIC_TARGET_API_URL` always wins (only a `None` override is filled in), and non-Copilot OpenAI targets are untouched, so direct-Anthropic setups are unaffected.

Reproduction:

```python
resolve_api_targets(ProviderApiOverrides(openai="https://api.githubcopilot.com", anthropic=None, ...))
# BEFORE: targets.anthropic == "https://api.anthropic.com"   -> Copilot bearer 401s there
# AFTER:  targets.anthropic == "https://api.githubcopilot.com"
```

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/providers/registry.py`: `resolve_api_targets` now fills a `None` Anthropic override with the OpenAI target when that target is a Copilot host (`is_copilot_upstream_url`). Explicit overrides and non-Copilot targets are unchanged.
- `tests/test_provider_registry.py`: added three tests — Copilot OpenAI target routes Anthropic to Copilot; an explicit Anthropic override wins; a non-Copilot OpenAI target leaves the Anthropic default alone.

## Testing

### Test Output

```text
tests/test_provider_registry.py tests/test_provider_registry_extended.py tests/test_banner_upstream_targets.py  ->  37 passed in 12.11s
(the new Copilot test FAILS on pre-fix code — verified via git stash)
uvx ruff@0.16.2 check headroom/providers/registry.py tests/test_provider_registry.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/providers/registry.py  ->  Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: `resolve_api_targets` with `openai="https://api.githubcopilot.com"` (and the `api.business.githubcopilot.com` variant) and `anthropic=None` returned `anthropic="https://api.anthropic.com"` before the fix and the Copilot host after; an explicit `anthropic="https://api.anthropic.com"` is preserved; `openai="https://api.openai.com"` leaves `anthropic` at the default.
- Observed result: Claude-model requests now resolve to the Copilot host that serves them; OpenAI/direct-Anthropic behavior is unchanged.
- Not tested: no live macOS/VS Code Copilot round trip (environment-specific); the target-resolution seam that decides the upstream host is exercised directly. `is_copilot_upstream_url` already recognizes the github.com Copilot hosts (verified).

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This is upstream target resolution in the provider registry, not a rollout-channel-gated runtime feature.
- Stable/default behavior changed: only the broken case changes — a Copilot OpenAI target with no Anthropic override now sends Claude to Copilot instead of 401ing against api.anthropic.com. Explicit Anthropic targets and non-Copilot OpenAI targets are byte-for-byte unchanged.
- Kill switch / disable path: set `ANTHROPIC_TARGET_API_URL` explicitly to opt out of the default.
- Qualification impact: none for non-Copilot deployments.

## Review Readiness

## Checklist

## Additional Notes

Fixes the routing/auth mismatch at the resolution layer so it applies uniformly across the Copilot config paths (`wrap vscode`, `wrap copilot --subscription`) that set the OpenAI target to a Copilot host. If a specific deploy sets neither target to a Copilot host (relying solely on path-based passthrough routing for OpenAI), configuring `ANTHROPIC_TARGET_API_URL` to the Copilot host remains the explicit escape hatch.

### DIFF
--- a/headroom/providers/registry.py
+++ b/headroom/providers/registry.py
@@ -162,9 +162,27 @@ def resolve_extra_headers(
 
 def resolve_api_targets(overrides: ProviderApiOverrides) -> ProviderApiTargets:
     """Resolve normalized upstream provider targets from configured overrides."""
+    from headroom.copilot_auth import is_copilot_upstream_url
+
+    openai = _normalize_api_url(overrides.openai, default=DEFAULT_OPENAI_API_URL)
+
+    # GitHub Copilot serves BOTH its OpenAI surface (``/chat/completions``,
+    # ``/responses``) and its Anthropic surface (``/v1/messages``, for Claude
+    # models) from the same host. When the OpenAI target is a Copilot host
+    # (``wrap copilot --subscription`` / ``wrap vscode`` both point it there so
+    # GPT models work) but no Anthropic target was set, Claude-model requests
+    # fell back to ``DEFAULT_ANTHROPIC_API_URL`` (api.anthropic.com) and 401'd
+    # with the Copilot bearer — "Invalid bearer token" (#3247). Default the
+    # Anthropic target to the same Copilot host so those requests reach the
+    # surface that actually serves them. An explicit ``ANTHROPIC_TARGET_API_URL``
+    # still wins (only a ``None`` override is filled in here).
+    anthropic_override = overrides.anthropic
+    if anthropic_override is None and is_copilot_upstream_url(openai):
+        anthropic_override = openai
+
     return ProviderApiTargets(
-        anthropic=_normalize_api_url(overrides.anthropic, default=DEFAULT_ANTHROPIC_API_URL),
-        openai=_normalize_api_url(overrides.openai, default=DEFAULT_OPENAI_API_URL),
+        anthropic=_normalize_api_url(anthropic_override, default=DEFAULT_ANTHROPIC_API_URL),
+        openai=openai,
         gemini=_normalize_api_url(overrides.gemini, default=DEFAULT_GEMINI_API_URL),
         cloudcode=_normalize_api_url(overrides.cloudcode, default=DEFAULT_CLOUDCODE_API_URL),
         vertex=_normalize_api_url(overrides.vertex, default=DEFAULT_VERTEX_API_URL),
--- a/tests/test_provider_registry.py
+++ b/tests/test_provider_registry.py
@@ -56,6 +56,55 @@ def test_resolve_api_targets_normalizes_trailing_v1() -> None:
     assert targets.vertex == "https://vertex.example"
 
 
+def test_copilot_openai_target_routes_anthropic_to_copilot() -> None:
+    """When the OpenAI target is a Copilot host and no Anthropic override is set,
+    the Anthropic target must default to the same Copilot host.
+
+    Copilot serves Claude models via its Anthropic surface (``/v1/messages``) on
+    the same host. Without this, Claude requests fell back to api.anthropic.com
+    and 401'd with the Copilot bearer ("Invalid bearer token", #3247).
+    """
+    targets = resolve_api_targets(
+        ProviderApiOverrides(
+            anthropic=None,
+            openai="https://api.githubcopilot.com",
+            gemini=None,
+            cloudcode=None,
+            vertex=None,
+        )
+    )
+    assert targets.openai == "https://api.githubcopilot.com"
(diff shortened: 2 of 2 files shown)


chopratejas:
Verified this end to end before merging — the diagnosis holds and the fix is correct.

**The bug reproduces on `main` (0.36.5).** The chain: `wrap vscode` (`cli/wrap.py:5904`) and `wrap copilot --subscription` (`:5837`) set **only** `openai_api_url`/`OPENAI_TARGET_API_URL` to the Copilot host — neither ever sets an Anthropic target. `resolve_api_targets` resolves each provider independently, so `anthropic` falls back to the default. `/v1/messages` then calls `handle_anthropic_messages(request)` with no base URL (`providers/proxy_routes.py:276`), landing on `api.anthropic.com`. Because `apply_copilot_api_auth` early-returns for non-Copilot URLs (`copilot_auth.py:1578`), the client's Copilot bearer is forwarded **verbatim** to Anthropic → `Invalid bearer token`. GPT models take the OpenAI handler, reach the right host, and work — exactly the reported asymmetry.

Running the real config chain with `OPENAI_TARGET_API_URL=https://api.githubcopilot.com`:

```
main:     anthropic=https://api.anthropic.com      copilot_auth_applies=False
this PR:  anthropic=https://api.githubcopilot.com  copilot_auth_applies=True
```

The premise that Copilot's Claude surface is `/v1/messages` on the same host isn't taken on faith — it's already encoded in `build_copilot_upstream_url`'s `keep_v1` branch (`copilot_auth.py:1289-1305`) and the #2409 regression it guards against. Independently corroborated by #3244, whose proxy log shows `forwarder=streaming ... https://api.anthropic.com/v1/messages` → 401.

Also checked: Business (`api.business.githubcopilot.com`) and GHE (`copilot-api.<tenant>.ghe.com`) hosts route correctly, an explicit `ANTHROPIC_TARGET_API_URL` still wins, and non-Copilot/unset controls are untouched. 367 passed / 8 skipped across the copilot + provider-registry + api-target suites.

One small correction to the PR description: Headroom doesn't attach the Copilot bearer here — it passes through the *client's* token, since `apply_copilot_api_auth` declines to touch a non-Copilot URL. Same outcome, but it's why the follow-up below matters.

**Two gaps this doesn't cover**, both on the Anthropic path and both now handled in #3277:
1. The buffered/non-stream arm applies no Copilot auth at all — `_retry_request` forwards headers untouched, so no minted/refreshed token and no `Copilot-Integration-Id`. Pre-existing, but this PR is what makes it reachable for Copilot.
2. Those turns are attributed to `anthropic` rather than `copilot`, because the buffered arm builds its URL by f-string and so skips the `build_copilot_upstream_url` chokepoint where the routing flag is set. Attribution only — the URL is identical and pricing is unaffected.

Neither blocks this. Merging.

PR #3261, opened 2026-08-26
feat(proxy): self-limiting session state for the compression-cache registry

chopratejas:
## Problem

The per-session `CompressionCache` registry (the map that replays previously-compressed messages byte-identically so the provider prefix cache stays warm) had no lifetime management:

- Idle/dead sessions lived forever until the hardcoded 500-session cap was hit.
- At capacity, eviction dropped the oldest-**created** quarter — which could wipe the busiest long-lived session (busting every one of its prefixes at once) while dead sessions survived.
- Neither the cap nor any TTL was tunable, which blocks gateway deployments (e.g. Kong sidecar/pool) fanning many concurrent sessions into one process.

## Changes

- **Idle-TTL sweep**: sessions idle longer than `HEADROOM_COMPRESSION_CACHE_TTL_SECONDS` (default 3900s) are evicted by a lazy sweep, at most once per 60s, piggybacked on `_get_compression_cache` — same pattern as `PrefixCacheTrackerRegistry._maybe_cleanup`, no background task. `last_seen` refreshes on **every** access, so an active session never expires.
- **LRU capacity eviction**: the registry is now an access-ordered `OrderedDict`; capacity pressure sheds the *idlest* quarter, never a busy session.
- **Tunable cap**: `HEADROOM_COMPRESSION_CACHE_MAX_SESSIONS` (default 500, floor 1).

## Why 3900s

Eviction is bust-free only once the provider's own prompt cache has lapsed. Providers don't expose their cache TTLs, and the risk is one-sided (late eviction costs a few MB; early eviction *causes* the bust this state exists to prevent), so the default is the upper bound of documented lifetimes across providers — Anthropic's 1h extended breakpoint, OpenAI's "up to an hour off-peak", Gemini's 60-min default — plus 5m grace. A parse-time floor of 600s keeps the TTL from ever dropping below the prefix tracker's session TTL: after the tracker expires, the byte-identical swap is the only remaining protection for a still-live provider prefix.

Read-hit signals are untouched: they govern the freeze boundary, never eviction — `read_hits == 0` usually means cold start or TTL lapse, where the map was just (re)written into the provider cache and deleting it would guarantee a second bust.

## Behavior impact

- Steady state (any session active within the TTL): zero change — same instances, same bytes, same freeze behavior.
- A session returning after >65 min idle now finds its map evicted — but every provider had already forgotten its prefix by then, so that turn was paying the cache-write price regardless (fail-open, no failed requests).
- Capacity eviction now protects busy sessions instead of punishing them.

## Testing

- New `tests/test_compression_cache_registry.py`: LRU-not-FIFO capacity eviction, small-cap edge case, TTL sweep eviction, access-refreshes-clock, sweep rate limiting.
- 386 tests pass across compression-cache, cache-stability (Anthropic + OpenAI), prefix-overlay, cold-start, cache-mode, and Bedrock-tracker suites; ruff check/format clean.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -2074,7 +2074,16 @@ class AnthropicHandlerMixin:
             # On a confirmed-cold turn we deliberately do NOT replay the previously
             # forwarded prefix: the cache is dead (nothing to keep byte-identical for)
             # and the replay would clobber the whole-prefix recompaction we just did.
-            if _decision.should_compress and not _skip_compression_for_backpressure:
+            #
+            # Backpressure skips the compression PIPELINE but must NOT skip this
+            # replay: on the saturated path `optimized_messages` is the raw
+            # originals, which mismatch the compressed prefix the provider cached
+            # — so every gated request busted its session's prompt cache exactly
+            # when traffic (and the re-write cost) peaked. The overlay itself is
+            # O(prefix) comparisons plus one token recount only when it actually
+            # replays, which is far cheaper than the whole-prefix cache re-write
+            # it prevents, so it stays on even under backpressure.
+            if _decision.should_compress:
                 if _cold_recompact_active:
                     _overlay_replayed = False
                 else:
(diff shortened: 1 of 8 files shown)


chopratejas:
Second commit adds org-scale hardening: (1) the cache-safety replay (`overlay_cached_prefix`) now stays ON under pre-upstream backpressure — previously the saturated path forwarded raw originals and busted every gated session's prompt cache at peak load (regression test proves both directions); (2) `HEADROOM_FROZEN_VERDICTS_MAX` env knob for the process-wide verdict store; (3) `HEADROOM_COMPRESSION_CACHE_MAX_ENTRIES` per-session entry cap knob. Plus tests/test_org_scale_limits.py including an active-sessions-survive-transient-flood churn test. 570+ tests green across backpressure, overlay, cache-stability, content-router, and registry suites.

PR #3262, opened 2026-08-26
feat(compress): session-aware /v1/compress (sidecar mode) + /v1/usage relay

chopratejas:
> **Stacked on #3261** (session-state lifecycle). Merge that first; this PR's diff is the last commit.

## Why

Gateways that own routing (e.g. Kong as the upstream caller) can't use Headroom's proxy path, and the stateless `/v1/compress` pushes all byte-replay bookkeeping onto the caller. This PR moves that state into the endpoint: **the caller sends the raw conversation + a session id every turn, forwards the returned bytes verbatim, and gets a byte-identical prefix — provider prompt cache preserved, no forwarding through Headroom.**

## Design

- **Session pre-work** mirrors the proxy's Zone 1: content-addressed swap of previously-computed compressed bytes, then freeze the **entire locally-replayable prefix** (`compute_frozen_count`).
- **Freeze posture deliberately differs from the proxy's `min(tracker, cache)`**: in sidecar mode, whatever this endpoint previously returned *is* the provider's cache contract — recompressing an already-returned message (even into a smaller form) is a bust. Over-freezing only forgoes tail compression; it can never bust. (A test caught exactly this: recompression drift produced a smaller form, and `overlay_cached_prefix`'s non-inflation guard then couldn't repair it.)
- **`PrefixCacheTracker.record_returned()`** — the sidecar equivalent of "last forwarded", captured at return time because whatever is returned is what the caller forwards.
- **`POST /v1/usage`** (same loopback exposure policy): the caller relays the provider's usage block; `update_from_response` makes freeze decisions provider-confirmed. Optional — skipping it degrades freeze precision, never correctness.
- Sessions are namespaced `compress:<id>`; the registry's TTL/LRU lifecycle from #3261 applies automatically. No session id ⇒ stateless contract byte-for-byte unchanged.

## Caller contract (Kong)

1. Send raw history + `config.session_id` (or `x-headroom-session-id`) every turn.
2. Forward the returned `messages` to the provider **verbatim**.
3. Optionally relay the provider's usage block to `/v1/usage`.

## Testing

11 cases in `tests/test_compress_session_mode.py`: stateless regression + no state leakage, invalid-id rejection, 2-turn and 3-turn byte-stability, header-based ids, cross-session isolation, usage-feedback freeze advancement, unknown-session 404, usage validation, TTL-eviction fail-open, explicit `frozen_message_count` precedence. Plus 435+ existing tests green across compress-API, CCR store bridge, tokenizer routing, cache-stability, tracker, overlay, backpressure, and registry suites.

## Phase 2 (follow-up, deliberately not here)

Migrate the proxy request path onto this same session engine so both modes share one compression/state codepath. This PR establishes the target abstraction; rewriting the battle-tested inline handler logic belongs in its own PR.

### DIFF
--- a/headroom/cache/compression_cache.py
+++ b/headroom/cache/compression_cache.py
@@ -123,6 +123,12 @@ class CompressionCache:
         # `RLock` (not `Lock`) so future code can call locked methods from
         # inside another locked method without self-deadlock.
         self._lock = threading.RLock()
+        # Serializes one sidecar-mode compress turn per session (pre-work,
+        # pipeline, post-work run as one block on an executor thread). The
+        # sidecar contract is sequential turns per conversation; this lock
+        # keeps a contract-violating concurrent pair from interleaving and
+        # tearing the tracker's prev-original/prev-returned snapshots.
+        self.session_turn_lock = threading.Lock()
         self._cache: OrderedDict[str, _CacheEntry] = OrderedDict()
         # `_stable_hashes` is CONTENT-KEYED, not positional. It records "we
         # have seen this content before and it is known not to compress
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -1269,6 +1269,16 @@ class SessionTrackerStore:
         self._lineage_affinities: dict[str, str | None] = {}
         self._lineage_counter = itertools.count(1)
 
+    def peek(self, session_id: str) -> PrefixCacheTracker | None:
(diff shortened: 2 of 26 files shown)


chopratejas:
Superseded by #3270 — identical changeset squashed to one conventional commit. After #3261's squash-merge broke this branch's ancestry, its commit history could not pass the commitlint gate against main (a pre-squash 120-char header plus restack merge commits), and rewriting this branch's history wasn't possible in this session. #3270 carries the full diff and the review hardening.

PR #3263, opened 2026-08-26
refactor(proxy): single session-turn brain shared by /v1/compress and the proxy paths

chopratejas:
> **Stacked on #3262** (which stacks on #3261). Merge order: #3261 → #3262 → this. This PR's diff is the last commit.

## Goal

One brain. The cache-management tier — freeze computation, Zone-1 byte swap, cached-prefix overlay — previously existed twice: inline in the proxy request handlers, and (as of #3262) in the `/v1/compress` sidecar path. This PR extracts it into **`headroom/proxy/session_engine.py`**, invoked by BOTH. Every future cache-management fix lands in both modes by construction.

## Design

**`prepare_turn(...)` → `TurnPrep`** — freeze + `mark_stable_from_messages` + `apply_cached`, with two *deliberately different, documented* freeze policies:
- `FREEZE_POLICY_PROXY`: `min(tracker_frozen, cache_count)` — never freeze past provider-confirmed (the #327 posture). The proxy passes its already-composed tracker/strict-override value, reproducing the previous `min()` byte-for-byte.
- `FREEZE_POLICY_SIDECAR`: `max(cache_count, explicit)` — freeze everything locally replayable, because in sidecar mode whatever was previously returned *is* the provider's cache contract; recompressing it (even "better") busts.

**`finalize_turn(...)` → `TurnFinal`** — the byte-identical cached-prefix replay (`overlay_cached_prefix`) + conditional token recount hook.

This was run as a **strictly behavior-preserving extraction**: the bar was every pre-existing test passing *unmodified*, and it held.

## What migrated / what didn't

| Path | Status |
|---|---|
| `/v1/compress` sidecar turn | ✅ engine (freeze/swap + overlay); lock, executor offload, savings accounting, record_returned unchanged |
| `anthropic.py` token-mode pre-block | ✅ engine; background compression, cold-start fast pass, sync pipeline all consume `prep.pipeline_input` unchanged |
| `anthropic.py` + `openai.py` overlay blocks | ✅ engine; `_cold_recompact_active` skip, inflation-guard wiring, `normalize_message_cache_control` preserved |
| `openai.py` proxy token-mode **pre**-block | ⏸ deliberately inline, documented at the site: adopting `prepare_turn` would change behavior three ways (no `mark_stable` call today; cache mode keeps the latest observation mutable; token mode freezes on `compute_frozen_count` alone). Behavior preservation beats completeness — unifying it is a candidate follow-up *with* a semantics decision, not a silent refactor. |

## One knowing behavioral delta (exceptional path only)

A tokenizer crash during the post-overlay recount now **fails open** to the pipeline's token count instead of failing the request (the hook is fail-open). No test covered the old propagation; a request now survives a tokenizer bug with slightly stale token accounting.

## Testing

- `tests/test_session_engine.py`: 11 direct unit tests — both freeze policies (proxy clamps by cache count; sidecar maxes with explicit; explicit-wins), overlay fires/doesn't, recount hook called only on replay, empty-snapshot fallbacks, unknown policy rejected.
- **Parity bar: 491 passed, 2 skipped**, all unmodified — cache-stability (Anthropic + OpenAI), `test_cache`, overlay, backpressure (including the replay-under-saturation regression test), cold-start fast pass, stage timings, cache-mode, Bedrock wiring, aligner, session-mode byte-stability (incl. tracker-loss), compress-API, CCR bridge, org-scale, and registry suites. These suites assert forwarded-byte behavior, so they are the no-bust proof.
- ruff check/format clean.

## Build process

Implemented by two sequential context-inheriting subagents (engine + sidecar migration, then proxy migration), each gated on the full green sweep, with the parent session independently re-running all 491 tests and spot-checking the freeze formulas and call sites before commit. Sequential because both steps touch `openai.py`.

### DIFF
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -1270,14 +1270,24 @@ class SessionTrackerStore:
         self._lineage_counter = itertools.count(1)
 
     def peek(self, session_id: str) -> PrefixCacheTracker | None:
-        """Return the tracker for ``session_id`` if one exists, else None.
+        """Return the live tracker for ``session_id``, else None.
 
         Never creates: lookup paths that must not leave a footprint (e.g. the
         ``/v1/usage`` unknown-session check, where ``get_or_create`` would let
         a flood of novel ids grow the store unboundedly within each TTL
         window) use this instead of :meth:`get_or_create`.
+
+        A TTL-expired-but-unswept tracker answers None too: the sweep runs
+        lazily from get_or_create at 60s granularity, so without this check an
+        expired session would keep answering with stale pre-expiry state — and
+        a caller that then touched it (``update_from_response`` stamps
+        ``_last_activity``) would resurrect the dead tracker indefinitely,
+        making the documented 404-on-expired contract nondeterministic.
         """
-        return self._trackers.get(session_id)
+        tracker = self._trackers.get(session_id)
+        if tracker is None or tracker.is_expired:
+            return None
+        return tracker
 
     def get_or_create(self, session_id: str, provider: str) -> PrefixCacheTracker:
         """Get existing tracker or create a new one for this session."""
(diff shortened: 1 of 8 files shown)


chopratejas:
Superseded by #3271 — identical changeset squashed to one conventional commit (same commitlint/ancestry situation as #3262 → #3270; verified `git diff` between this branch and the replacement commit is empty).

PR #3277, opened 2026-08-26
fix(proxy/anthropic): authenticate and attribute buffered Copilot turns

chopratejas:
## Description

Follow-up to #3258. That PR points the Anthropic target at the Copilot host so Claude models stop 401'ing. This PR fixes two things on the Anthropic path that were only ever correct on the **streaming** arm, and which #3258 makes reachable for real Copilot traffic.

Copilot serves Claude models from its Anthropic surface (`/v1/messages`) on the same host as its OpenAI surface, so the resolved Anthropic target can be a Copilot host with no per-request `upstream_base_url` involved. That is the case both arms below get wrong.

**1. The buffered arm sent no Copilot credential.** `apply_copilot_api_auth` is keyed on the upstream URL and was applied only by `_stream_response` (`handlers/streaming.py:1205`). The buffered/non-stream arm sends through `_retry_request` (`proxy/server.py:2132`), which forwards headers untouched — so the request carried whatever the client happened to send and none of Headroom's own credential handling: no minted or refreshed token (the one `wrap vscode` explicitly hands the proxy), no `Copilot-Integration-Id` default. A client token that went stale mid-session 401'd here while the streaming path recovered. That arm is not an edge case — it is the CCR `stream:true → buffered stream:false` flip, and Claude Code's non-stream retry.

**2. Copilot turns were attributed to "anthropic".** `build_copilot_upstream_url` is the only place `mark_request_routed_to_copilot` fires (`copilot_auth.py:1288`), and `emit_request_outcome` relabels the provider off that flag (`proxy/outcome.py:419`). The buffered arm built its URL by f-string, skipping the chokepoint, so those turns showed as `anthropic` on the dashboard. The URL produced is byte-identical either way — this is attribution only, not routing. `proxy/cost.py` has no Copilot-specific branch, so pricing is unaffected.

Both changes are inert off the Copilot path: `apply_copilot_api_auth` returns the headers unchanged for a non-Copilot URL, and `build_copilot_upstream_url` only joins base + path there.

Independent of #3258 and based on `main` — the gaps are reachable today by setting `ANTHROPIC_TARGET_API_URL` to a Copilot host.

## Type of Change

## Changes Made

- `handlers/anthropic.py`: build the default-target URL through `build_copilot_upstream_url` instead of an f-string, so the routed-to-Copilot flag is set for attribution.
- `handlers/anthropic.py`: apply `apply_copilot_api_auth` on the buffered arm before the upstream send. Mutated in place, matching the accept-header handling directly above — the closures below capture `headers`, and the CCR continuation rebuilds its own header set from it, so the continuation inherits the auth too.
- New test pinning both at the `_retry_request` seam: URL built, headers as they go on the wire, and the flag as it stands at send time.

## Testing

- [x] Linting passes (`ruff check`, CI-pinned 0.16.3)

### Test Output

Both new assertions fail on `main` with exactly the symptoms described, and pass with the fix:

```text
$ git stash && pytest tests/test_proxy/test_anthropic_copilot_upstream_auth.py
tests/.../test_buffered_turn_to_copilot_is_authenticated
E   KeyError: 'authorization'
tests/.../test_buffered_turn_to_copilot_is_flagged_for_attribution
E   assert False is True
==================== 2 failed, 2 passed, 1 warning in 3.38s ====================

$ git stash pop && pytest tests/test_proxy/test_anthropic_copilot_upstream_auth.py
========================= 4 passed, 1 warning in 2.88s =========================
```

The two that pass on `main` are the invariants this must not break (path `/v1` preserved per #2409, non-Copilot target untouched).

Regression run over the affected surface:

```text
$ pytest tests/ -k "copilot or anthropic or outcome or provider_registry or proxy_routes or upstream"
= 3 failed, 1111 passed, 33 skipped, 11112 deselected in 152.98s =
```

The 3 failures are `tests/test_proxy/test_openai_transport_path_prefix.py` and are **pre-existing on `main`** (verified by running that file on a clean checkout — same 3 fail). Untouched by this PR, which is Anthropic-path only.

```text
$ uvx ruff@0.16.3 check headroom/proxy/handlers/anthropic.py tests/test_proxy/test_anthropic_copilot_upstream_auth.py
All checks passed!
$ mypy headroom/proxy/handlers/anthropic.py
```

## Real Behavior Proof

- **Environment:** macOS arm64, Python 3.12.13, `main` @ 0.36.5.
- **Exact command / steps:** drive `POST /v1/messages` through the real app (`create_app` + `TestClient`, non-stream body) with the Anthropic target set to `https://api.githubcopilot.com`, intercepting `_retry_request` to capture what was about to go on the wire. Copilot token minting stubbed to a fixed value.
- **Observed result:** before — no `Authorization` header at all on the buffered arm, and `request_routed_to_copilot()` is `False` at send time. After — `Authorization: Bearer <minted>` plus `Copilot-Integration-Id` and `Editor-Version`, flag `True`, URL unchanged at `https://api.githubcopilot.com/v1/messages`. With a non-Copilot target, no credential is invented and the flag stays `False`.
- **Not tested:** against live `api.githubcopilot.com` — no Copilot subscription in this environment. Token minting is stubbed, so the refresh path itself is exercised only to the provider boundary. Anthropic **batch** endpoints (`/v1/messages/batches`, `handlers/anthropic.py:5066+`) still build against `self.ANTHROPIC_API_URL` and will point at Copilot, which does not serve them — pre-existing and out of scope here — filed as #3278.

## Runtime Rollout Safety

- **Rollout-managed feature(s):** none — no flag or channel involved.
- **Stable/default behavior changed:** no, for every non-Copilot upstream: the URL is byte-identical and `apply_copilot_api_auth` early-returns for non-Copilot URLs. Behavior changes only when the Anthropic target is a Copilot host, which is the broken case.
- **Kill switch / disable path:** set `ANTHROPIC_TARGET_API_URL` to a non-Copilot host; both paths go inert.
- **Unsafe override required:** none.
- **Qualification impact:** none.
- **Rollback path:** revert this commit — it is self-contained to one file plus a new test.

## Review Readiness

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -27,7 +27,7 @@ import httpx
 from headroom.agent_savings import proxy_pipeline_kwargs
 from headroom.ccr.context_tracker import looks_like_claude_code_compact_summary
 from headroom.ccr.marker_resolution import resolve_markers_in_response
-from headroom.copilot_auth import build_copilot_upstream_url
+from headroom.copilot_auth import apply_copilot_api_auth, build_copilot_upstream_url
 from headroom.pipeline import PipelineStage, summarize_routing_markers
 from headroom.proxy.auth_mode import (
     classify_auth_mode,
@@ -3501,10 +3501,20 @@ class AnthropicHandlerMixin:
 
             # Direct Anthropic API, or a provider-compatible Anthropic
             # Messages endpoint such as Vertex AI publisher rawPredict.
+            # Both arms build through `build_copilot_upstream_url` because that
+            # is the only place `mark_request_routed_to_copilot` fires, and the
+            # outcome funnel relabels `provider` to "copilot" off that flag (see
+            # proxy/outcome.py). The resolved target reaches Copilot without any
+            # per-request `upstream_base_url` — `wrap vscode` points the
+            # Anthropic target at the Copilot host — so this else arm carries the
+            # Claude-on-Copilot traffic, and building it by f-string attributed
+            # every one of those turns to "anthropic" on the dashboard.
+            # For a non-Copilot base the builder only joins base + path, so the
+            # URL itself is unchanged.
             url = (
                 build_copilot_upstream_url(upstream_base_url, request.url.path)
                 if upstream_base_url
-                else f"{self.ANTHROPIC_API_URL}/v1/messages"
+                else build_copilot_upstream_url(self.ANTHROPIC_API_URL, "/v1/messages")
             )
             if upstream_base_url and request.url.query:
                 url = f"{url}?{request.url.query}"
@@ -3761,6 +3771,27 @@ class AnthropicHandlerMixin:
                         for _accept_key in [k for k in headers if k.lower() == "accept"]:
                             headers.pop(_accept_key, None)
                         headers["accept"] = "application/json"
+
+                    # Copilot auth is applied per-URL, and until now only the
+                    # streaming forwarder did it (``_stream_response``). This
+                    # buffered arm sends via ``_retry_request``, which forwards
+                    # headers untouched — so a Claude turn routed to Copilot
+                    # arrived with whatever the client happened to send and none
+                    # of Headroom's own credential handling: no minted or
+                    # refreshed token (the one ``wrap vscode`` hands the proxy),
+                    # no ``Copilot-Integration-Id`` default. A client token that
+                    # went stale mid-session therefore 401'd here while the
+                    # streaming path recovered.
+                    #
+                    # A non-Copilot URL returns the headers unchanged, so this is
+                    # a no-op everywhere else.
+                    #
+                    # Mutated in place for the same reason as the accept header
+                    # above: the closures below capture ``headers``, and the CCR
+                    # continuation rebuilds its own header set from it.
+                    _copilot_authed_headers = await apply_copilot_api_auth(dict(headers), url=url)
+                    headers.clear()
+                    headers.update(_copilot_authed_headers)
 
                     # Populated once the upstream answers 200 with parseable
                     # JSON, so the guard below can fall back to it (#3088).
--- a/tests/test_proxy/test_anthropic_copilot_upstream_auth.py
+++ b/tests/test_proxy/test_anthropic_copilot_upstream_auth.py
@@ -113,6 +113,23 @@ def _headers_lower(send: _Send) -> dict[str, str]:
     return {k.lower(): v for k, v in send.headers.items()}
 
 
+def _emitted_providers(anthropic_api_url: str, monkeypatch: pytest.MonkeyPatch) -> list[str]:
+    """Provider labels on the outcomes this request emitted.
+
+    The relabel happens inside ``emit_request_outcome``, which runs in a task
+    created by ``asyncio.shield`` — so this also pins that the flag survives the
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed the current head adversarially. The buffered path now uses the same URL-based Copilot detection and credential refresh boundary as streaming, while preserving the non-Copilot no-op behavior and /v1/messages path. The tests exercise the actual _retry_request seam, including attribution state and negative-control headers. The PR description accurately captures scope and rollout behavior, and the current required checks are green. Ready for final human merge review.

chopratejas:
Self-review notes — three things I checked that the diff doesn't show, and one behavioural nuance a reviewer should weigh in on.

**The attribution fix is now pinned end to end, not at the flag.** My original test asserted `request_routed_to_copilot()` was true at send time, which stops one step short: `emit_request_outcome` runs inside a task created by `asyncio.shield` (`server.py:2186`), so the flag has to survive the context copy into that task for the relabel to reach the dashboard. It does, but the test wasn't proving it. Added a test that captures the emitted outcome and asserts the provider label. Verified against the parent commit — Copilot target emits `anthropic` without the fix, `copilot` with it; non-Copilot control unmoved both ways.

**Header handling on the Copilot path, observed rather than assumed.** Driving a request with realistic client headers:

```
client sends:  x-api-key, anthropic-version, anthropic-beta, copilot-integration-id, user-agent
-> Copilot:    Authorization: Bearer <minted>   (x-api-key REMOVED)
               anthropic-version / anthropic-beta preserved
               copilot-integration-id preserved as the client's value
               Editor-Version defaulted
-> Anthropic:  completely untouched, x-api-key intact, no Authorization invented
```

Worth calling out that the client's Anthropic `x-api-key` is stripped before the request reaches Copilot — a credential that must not cross hosts, and doesn't.

**Failure mode.** The new `await` sits inside the `try:` at `anthropic.py:3522`, so a token-provider failure is caught at `:4818` and returns a sanitized 502 with `record_failed`, and the `finally` still runs `_finalize_pre_upstream()`. Same class as any upstream failure, not an unhandled 500.

**The nuance worth a second opinion:** because real `wrap vscode` bearers are `kind=unknown` to `_is_forwardable_copilot_bearer_token` (they're `tid=...;exp=...`, not `tid_...` — see the log in #3244), this means buffered turns now get Headroom's **minted** token where they previously forwarded the client's verbatim. I believe that's right: it is exactly what the streaming path (`streaming.py:1205`) and the GPT path (`openai.py`) already do, and GPT models demonstrably work — so the minted token is entitled. Genuinely forwardable `gho_` CLI tokens are still passed through untouched, which is the case the `_is_forwardable_copilot_bearer_token` docstring warns about. But it is a real behavioural change on that arm, so flagging it rather than burying it.

Minor, not addressed: the token mint is a potentially blocking call made while holding the pre-upstream semaphore. The provider caches tokens so it's a per-refresh cost, not per-request, but it is new work inside that section.

JerrettDavis (approved):
Re-reviewed current head a29d0141 after the post-approval test commit. The only new source is stronger end-to-end attribution coverage: it captures record_outcome after the asyncio.shield task boundary and proves Copilot is emitted as copilot while the Anthropic control remains anthropic. This directly strengthens the behavior reviewed previously and introduces no production change. No new blocker found; the remaining two Python shards are still settling.

PR #3308, opened 2026-08-27
fix(memory): reload persisted HNSW index at vector-index creation

tatus9:
## Description

`_create_vector_index` builds `HNSWVectorIndex` with `auto_save=True` and a `save_path` derived from the db path, so every mutation persists the index next to the db — but **nothing ever loaded those artifacts back**. For any fresh process this means:

1. The in-memory vector index starts **empty**: vector/hybrid search returns nothing until every row is re-indexed (`HNSWVectorIndex.search` short-circuits on an empty index).
2. Worse, the **first write from the empty process overwrites the persisted artifacts** with the near-empty in-memory index (auto_save), discarding them for every later process. The cross-process persistence that the `save_path` derivation promises silently never happened.

This PR reloads the persisted index right after construction when the artifacts exist. Missing artifacts (first run) are a no-op; a dimension mismatch or corrupt artifact logs a warning and starts fresh instead of crashing, so upgrades that change `vector_dimension` cannot brick startup.

Closes #

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/memory/factory.py` (`_create_vector_index`, HNSW branch): construct the index into a local, call `index.load_index(hnsw_save_path)` when a save path is configured, log the restored entry count, and return it. `FileNotFoundError` → first run, pass; any other exception → warn and start empty.
- `tests/test_memory/test_factory_hnsw_persistence.py`: two regression tests (fail on `main`, pass with this change).

## Testing

### Test Output

Reproduction (fails on `main`, passes with this PR):

```
$ uv run --frozen --with pytest --with pytest-asyncio --with hnswlib --with numpy \
    -m pytest tests/test_memory/test_factory_hnsw_persistence.py -q

main:    2 failed   (fresh index starts empty; persisted artifacts get clobbered)
this PR: 2 passed
```

```
FAILED tests/test_memory/test_factory_hnsw_persistence.py::test_fresh_index_reloads_persisted_entries
FAILED tests/test_memory/test_factory_hnsw_persistence.py::test_fresh_save_does_not_clobber_persisted_index
```

No regressions in the surrounding suites (identical results before/after; the 5 remaining failures in `test_factory.py` are pre-existing optional-dependency errors — `ModuleNotFoundError: No module named 'openai'` — unrelated to this change):

```
$ -m pytest tests/test_memory/test_factory.py tests/test_memory/test_hnsw_batch_capacity.py tests/test_memory/test_hierarchical.py -q
before: 5 failed, 78 passed, 4 skipped
after:  5 failed, 78 passed, 4 skipped
```

`ruff check` / `ruff format --check` clean on both changed files.

Manual verification on a long-running deployment (global memory db, headroom proxy + per-call helper processes): fresh processes went from `0` entries in the vector index to the full persisted entry count immediately, restoring vector search after every restart.

### DIFF
--- a/headroom/memory/factory.py
+++ b/headroom/memory/factory.py
@@ -7,6 +7,7 @@ and proper wiring between components.
 
 from __future__ import annotations
 
+import logging
 import threading
 from importlib.metadata import entry_points
 from pathlib import Path
@@ -23,6 +24,8 @@ from headroom.memory.config import (
 if TYPE_CHECKING:
     from headroom.memory.ports import Embedder, MemoryCache, MemoryStore, TextIndex, VectorIndex
 
+
+logger = logging.getLogger(__name__)
 
 # Extension groups for memory backends registered via setuptools entry points.
 _MEMORY_STORE_GROUP = "headroom.memory_store"
@@ -307,7 +310,7 @@ def _create_vector_index(config: MemoryConfig) -> VectorIndex:
         if config.db_path:
             hnsw_save_path = config.db_path.parent / f"{config.db_path.stem}_hnsw"
 
-        return HNSWVectorIndex(
+        index = HNSWVectorIndex(
             dimension=config.vector_dimension,
             ef_construction=config.hnsw_ef_construction,
             m=config.hnsw_m,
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the latest head. The factory now restores the exact auto-save artifact path before exposing the index, treats first-run absence as normal, and fails open with an operator-visible warning for incompatible/corrupt artifacts. The regression tests cover both immediate visibility and the more damaging first-write clobber scenario. No blocking correctness issues found.

PR #3309, opened 2026-08-27
[security] Fix: path traversal in Bedrock model_id allows SSRF

failsafesecurity:
### Summary

The Bedrock `invoke` and `invoke-with-response-stream` handlers extract `model_id` from the URL path via Axum's `Path<String>` extractor, which URL-decodes the path segment. The decoded `model_id` was directly interpolated into the upstream URL path without any validation:

```rust
let path = format!("/model/{model_id}/{action}");
```

### Impact

**Severity: Critical**

An attacker can craft a request with path traversal sequences in the `model_id`:

```http
POST /model/..%2F..%2F..%2Ffoundation-models/invoke HTTP/1.1
Host: proxy:8787
Content-Type: application/json

{"anthropic_version": "bedrock-2023-05-31", "max_tokens": 1, "messages": [{"role": "user", "content": "test"}]}
```

Axum decodes `%2F` to `/`, so `model_id` becomes `../../foundation-models`. The resulting upstream URL:

```
https://bedrock-runtime.us-east-1.amazonaws.com/model/../../../foundation-models/invoke
```

The SigV4 signature is computed **after** the URL is constructed, so it signs the manipulated path — AWS will accept the signed request. This allows:
- Redirecting the upstream call to arbitrary paths on the Bedrock endpoint
- Accessing Bedrock management/listing endpoints not intended for the proxy
- Bypassing the intended `/model/{model_id}/invoke` path structure

### Affected Code

- `crates/headroom-proxy/src/bedrock/invoke.rs` — `build_bedrock_upstream()` function
- `crates/headroom-proxy/src/bedrock/invoke_streaming.rs` — `build_bedrock_streaming_upstream()` function

### Fix

Added `is_safe_model_id()` validation in `bedrock::vendor` that rejects `model_id` containing:
- Forward slashes (`/`) — the primary SSRF vector (decoded from `%2F` by Axum)
- Backslashes (`\`)
- Double-dot sequences (`..`) — path traversal segments
- Empty strings

The check is applied in both the non-streaming (`invoke.rs`) and streaming (`invoke_streaming.rs`) handlers **before** constructing the upstream URL. Invalid `model_id` values are rejected with `400 Bad Request` and a structured `bedrock_model_id_invalid` log event.

### Proof of Concept

```bash
# Before fix: this request would traverse the path
curl -X POST http://proxy:8787/model/..%2F..%2F..%2Fadmin/invoke \
  -d "{\"anthropic_version\":\"bedrock-2023-05-31\",\"max_tokens\":1,\"messages\":[{\"role\":\"user\",\"content\":\"test\"}]}"

# After fix: returns 400 with {"error":{"type":"bedrock_model_id_invalid",...}}
```

### Testing

Unit tests added in `vendor.rs` covering:
- Valid model IDs (anthropic, amazon, meta, geo-prefixed)
- Path traversal attempts (`../../admin`, `foo/bar`, `..\..\admin`, `..`, empty)

Signed-off-by: Failsafe Security <security@getfailsafe.com>

### DIFF
--- a/crates/headroom-proxy/src/bedrock/invoke.rs
+++ b/crates/headroom-proxy/src/bedrock/invoke.rs
@@ -65,7 +65,7 @@ use crate::proxy::AppState;
 // would risk drift from the middleware's resolution + WARN log.
 use headroom_core::auth_mode::AuthMode;
 
-use crate::bedrock::vendor::is_anthropic_model_id;
+use crate::bedrock::vendor::{is_anthropic_model_id, is_safe_model_id};
 
 /// RAII guard that observes the `bedrock_invoke_latency_seconds`
 /// histogram on drop. Created at handler entry; observed when the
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
Reviewed both native Bedrock paths and URL construction. Validation happens before compression/signing/upstream construction in streaming and non-streaming handlers; rejecting decoded separators and double-dot sequences closes the path-segment escape while preserving documented foundation and inference-profile IDs. Url::set_path handles other reserved characters as path data, and focused unit coverage exercises valid and malicious IDs. No blocking issues found.

PR #3310, opened 2026-08-27
[security] Fix: unrestricted file read in headroom_read MCP tool

failsafesecurity:
### Summary

The `headroom_read` MCP tool (gated behind `HEADROOM_MCP_READ=on`, default off) used `Path(file_path).expanduser().resolve()` without any path sandboxing. When enabled, an AI agent could read **any file on the filesystem** — including SSH keys, API credentials, and `.env` files.

### Impact

**Severity: High**

When `headroom_read` is enabled, a prompt-injected AI agent or malicious MCP client can exfiltrate secrets:

```json
{
  "method": "tools/call",
  "params": {
    "name": "headroom_read",
    "arguments": {"file_path": "~/.ssh/id_rsa"}
  }
}
```

Other sensitive paths that can be read:
- `~/.env` — API keys, database URLs
- `~/.aws/credentials` — AWS access keys
- `~/.kube/config` — Kubernetes cluster credentials
- `/etc/passwd` — system user enumeration
- `~/.gnupg/` — PGP private keys
- Any application source code or configuration

The `expanduser()` call expands `~` to the home directory, and `resolve()` follows symlinks — so even a symlink to a sensitive file would be read.

### Affected Code

- `headroom/ccr/mcp_server.py` — `_handle_read()` method (line ~944)

### Fix

**Workspace sandbox (default-on):**
- File reads are restricted to the current working directory and `HEADROOM_WORKSPACE_DIR`
- Paths that resolve outside these roots are rejected with an access denied error
- Symlinks are followed by `resolve()` but the resolved path must still be within the allowed roots

**Sensitive file pattern blocking:**
- Even in unrestricted mode, paths matching sensitive patterns are blocked:
  - `.ssh/`, `id_rsa`, `id_ecdsa`, `id_ed25519`, `.pem`
  - `.env`, `.aws/credentials`, `.gnupg/`, `.kube/config`
  - `.docker/config`, `.npmrc`, `.pypirc`, `.netrc`

**Opt-in unrestricted mode:**
- `HEADROOM_MCP_READ_UNRESTRICTED=1` restores the old behavior for operators who need it
- Defaults to off (restricted mode)

### Proof of Concept

```python
# Before fix (with HEADROOM_MCP_READ=on):
# An AI agent calls headroom_read with:
{"file_path": "~/.ssh/id_rsa"}
# → Returns the contents of the SSH private key

# After fix:
# → Returns {"error": "Access denied: ~/.ssh/id_rsa is outside the allowed workspace directory..."}
```

### Testing

- Python syntax validated with `ast.parse()`
- Sandbox check uses `Path.resolve()` which handles symlinks, `..`, and `~` expansion
- Sensitive pattern matching is substring-based on the resolved path

Signed-off-by: Failsafe Security <security@getfailsafe.com>

### DIFF
--- a/headroom/ccr/mcp_server.py
+++ b/headroom/ccr/mcp_server.py
@@ -87,6 +87,37 @@ _READ_ENABLED = os.environ.get("HEADROOM_MCP_READ", "off").lower().strip() in (
     "enabled",
 )
 
+# When headroom_read is enabled, restrict file reads to the workspace
+# directory and the current working directory by default. This prevents
+# prompt-injected AI agents from exfiltrating secrets (e.g. ~/.ssh/id_rsa,
+# .env files, /etc/passwd) via the MCP tool. Operators who need
+# unrestricted reads can opt in with HEADROOM_MCP_READ_UNRESTRICTED=1.
+_READ_UNRESTRICTED = os.environ.get("HEADROOM_MCP_READ_UNRESTRICTED", "off").lower().strip() in (
+    "on",
+    "true",
+    "1",
+    "yes",
+    "enabled",
+)
+
+# Sensitive file patterns that are always blocked regardless of unrestricted
+# mode, to prevent accidental secret exfiltration even by trusted agents.
+_SENSITIVE_PATTERNS = (
+    ".ssh/",
+    ".env",
+    ".aws/credentials",
+    ".gnupg/",
+    ".kube/config",
+    ".docker/config",
+    ".npmrc",
+    ".pypirc",
+    ".netrc",
+    "id_rsa",
+    "id_ecdsa",
+    "id_ed25519",
+    ".pem",
+)
+
 DEFAULT_PROXY_URL = os.environ.get("HEADROOM_PROXY_URL", "http://127.0.0.1:8787")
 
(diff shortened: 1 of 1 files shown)


JerrettDavis (changes requested):
The proposed sandbox still exposes sensitive Headroom state and does not implement the stated cross-platform guarantee. _paths.workspace_dir() defaults to ~/.headroom, which contains credentials/state such as copilot_auth.json and subscription/license data, yet that entire directory is added as an allowed read root and those files are not in _SENSITIVE_PATTERNS. In addition, sensitive matching uses case-sensitive POSIX substrings against str(path), so entries such as .aws/credentials, .kube/config, and .docker/config do not match resolved Windows paths with backslashes. Finally, the sensitive check is nested under if not _READ_UNRESTRICTED, contradicting the PR description/comment that these patterns are blocked regardless of unrestricted mode. Please define the intended trust boundary explicitly (project root versus Headroom's state directory), avoid granting the whole state directory, implement path-component/case-aware matching across platforms, decide and consistently implement unrestricted-mode semantics, and add executable tests for traversal, symlink escape, Windows-style/case variants, Headroom credential files, and unrestricted mode. A security fix at this boundary needs regression coverage, not only syntax validation.

PR #3325, opened 2026-08-28
fix(proxy): bound decompressed request bodies to close a zip-bomb DoS

### LINKED ISSUE #3284: [security] Unbounded decompression of request bodies enables zip-bomb denial of service
## Security Finding: Unbounded decompression of request bodies enables zip-bomb denial of service

**Severity**: HIGH (CVSS 8.7)
**CWE**: CWE-409
**Repository**: headroomlabs-ai/headroom

### Description

The proxy decompresses zstd/gzip/deflate/brotli request bodies with no cap on the decompressed output size. The only pre-decompression guard checks Content-Length (the compressed wire size) against 100 MB, so a small compressed payload that expands to gigabytes bypasses the check and exhausts memory.

### Impact

Unauthenticated remote denial of service: a single ~10 MB request kills the proxy process via OOM, disrupting all AI agent traffic

### Remediation

Cap the decompressed output size. For gzip/deflate, use zlib.decompressobj with a max_length loop that raises ValueError if the total exceeds MAX_REQUEST_BODY_SIZE. For zstd, construct ZstdDecompressor(max_window_size=MAX_REQUEST_BODY_SIZE) and read in chunks with a running total. For brotli, use brotli.Decompressor with chunked reads and a total cap. Apply the same MAX_REQUEST_BODY_SIZE to the decompressed bytes, not just the wire Content-Length. Also reject requests with no Content-Length when Content-Encoding is set, to close the chunked-encoding bypass.

### Proof of Concept

{"steps":["Craft a gzip bomb: a small compressed payload that decompresses to a very large size. Classic technique: compress a file filled with repeated null bytes (e.g., 10 GB of zeros compresses to ~10 MB with gzip).","Send POST to the proxy's /v1/messages endpoint with Content-Encoding: gzip and the bomb as the body. Content-Length reflects the ~10 MB compressed size, which passes the 100 MB gate at openai.py:3272.","The handler calls _read_request_json(request) at openai.py:3286, which calls _read_request_body_bytes at helpers.py:2662, which calls _gzip.decompress(raw) at helpers.py:2569.","_gzip.decompress allocates the full 10 GB decompressed result as a Python bytes object. The proxy process is OOM-killed or hangs."],"request_example":"python3 -c \"import gzip, http.client; body = gzip.compress(b'\\x00' * (10 * 1024**3)); conn = http.client.HTTPConnection('127.0.0.1', 8787); conn.request('POST', '/v1/messages', body, {'Content-Encoding': 'gzip', 'Content-Type': 'application/json', 'Content-Length': str(len(body)), 'x-api-key': 'sk-any'}); print(conn.getresponse().status)\"","vulnerable_response":"Process exits with signal 9 (OOM-killed) or hangs indefinitely with 100% memory usage. No HTTP response is returned.","secure_response":"413 Payload Too Large or 400 Bad Request returned immediately, before decompression, because the decompressed size exceeds a configured maximum.","reproduction":{"steps":[{"command":"cat > /tmp/pocwork-D-006-le6e93_i/poc/poc_oom2.py <<'PYEOF'\nimport gzip, resource, os, sys\nGB = 1024**3\n# Build the bomb with full memory available.\nbomb = gzip.compress(b\"\\x00\" * (2 * GB), compresslevel=9)\nprint(f\"wire bytes: {len(bomb)} ({len(bomb)/1024/1024:.1f} MB)  -> decompresses to 2 GB\", flush=True)\npid = os.fork()\nif pid == 0:\n    # Child: cap address space at 384 MB, then decompress (same call as helpers.py:2569).\n    soft, hard = resource.getrlimit(resource.RLIMIT_AS)\n    resource.setrlimit(resource.RLIMIT_AS, (384 * 1024 * 1024, hard))\n    try:\n        out = gzip.decompress(bomb)\n        print(f\"child: decompressed OK size={len(out)}\", flush=True)\n    except MemoryError as e:\n        print(f\"child: MemoryError while fully materializing -> process would be OOM-killed\", flush=True)\n    finally:\n        os._exit(0)\nelse:\n    _, status = os.waitpid(pid, 0)\n    print(f\"parent: child exited with status {status}\")\nPYEOF\npython3 poc/poc_oom2.py","output":"wire bytes: 2087280 (2.0 MB)  -> decompresses to 2 GB\nchild: MemoryError while fully materializing -> process would be OOM-killed\nparent: child exited with status 0","error":false}],"summary":"A gzip bomb sent to /v1/messages passes the 100 MB compressed-body gate; decompression materializes 2 GB and kills the process."},"trail":{"version":1,"synthesis":"graph","status":"confirmed","nodes":[{"id":"n1","kind":"discovery","entity":"origin","label":"Security agent","detail":"Discovery to confirmed proof","status":"observed","source":"system","source_index":0,"phase":"discovery"},{"id":"n2","kind":"discovery","entity":"application","label":"Audited headroom-main proxy codebase","status":"observed","source":"discovery_event","source_index":0,"phase":"discovery","detail":"All proxy server and provider files reviewed","offset_ms":247698,"timestamp_ms":1787774245152},{"id":"n3","kind":"discovery","entity":"service","label":"Found 100 MB request-body gate","status":"observed","source":"agent_event","source_index":4,"phase":"validation","detail":"Content-Length checked before decompress; compressed body below limit accepted","tool_call_id":"call_dc16300efe8e40388deae06b","offset_ms":7571,"timestamp_ms":1787777228041},{"id":"n4","kind":"action","entity":"endpoint","label":"POST gzip bomb to /v1/messages","status":"inferred","source":"reported_poc","source_index":1,"phase":"exploitation","detail":"Content-Encoding gzip; ~10 MB compressed body passes the gate"},{"id":"n5","kind":"action","entity":"application","label":"Handler decompresses via gzip.decompress","status":"inferred","source":"reported_poc","source_index":2,"phase":"exploitation","detail":"_read_request_json calls _read_request_body_bytes, which calls _gzip.decompress"},{"id":"n6","kind":"blocked","entity":"blocked","label":"Initial bomb build hit MemoryError first","status":"blocked","source":"agent_event","source_index":11,"phase":"validation","detail":"2 GB allocation failed under 384 MB cap before decompression","tool_call_id":"call_dfe258a3ef594937b2ba8714","offset_ms":52800,"timestamp_ms":1787777273270},{"id":"n7","kind":"proof","entity":"outcome","label":"2 GB bomb caused MemoryError","status":"confirmed","source":"reproducer","source_index":0,"phase":"validation","detail":"2 MB wire bomb decompresses to 2 GB; memory-capped child process dies","tool_call_id":"call_1d4c5b4d9e6b424a99538af3","offset_ms":57824,"timestamp_ms":1787777278294}],"edges":[{"from":"n2","to":"n3","relation":"led_to"},{"from":"n3","to":"n4","relation":"led_to"},{"from":"n4","to":"n5","relation":"led_to"},{"from":"n5","to":"n6","relation":"blocked_at"},{"from":"n6","to":"n7","relation":"led_to"},{"from":"n5","to":"n7","relation":"confirmed"},{"from":"n1","to":"n2","relation":"performed"}],"phases":["discovery","exploitation","validation"],"started_at_ms":1787774245152,"finished_at_ms":1787777278294,"duration_ms":3033142,"omitted_count":40}}

### Affected Code

```
headroom/proxy/helpers.py:2540: async def _read_request_body_bytes(request: Request) -> bytes:
headroom/proxy/helpers.py:2547:     encoding = (request.headers.get("content-encoding") or "").lower().strip()
headroom/proxy/helpers.py:2548:     raw = await request.body()
headroom/proxy/helpers.py:2550:     if encoding in ("zstd", "zstandard"):
headroom/proxy/helpers.py:2554:             dctx = zstandard.ZstdDecompressor()
headroom/proxy/helpers.py:2555:             reader = dctx.stream_reader(raw)
headroom/proxy/helpers.py:2556:             raw = reader.read()  # no max_output_size
headroom/proxy/helpers.py:2565:     elif encoding == "gzip":
headroom/proxy/helpers.py:2569:             raw = _gzip.decompress(raw)  # no size limit
headroom/proxy/helpers.py:2572:     elif encoding == "deflate":
headroom/proxy/helpers.py:2576:             raw = zlib.decompress(raw)  # no max_length
headroom/proxy/helpers.py:2579:     elif encoding == "br":
headroom/proxy/helpers.py:2583:             raw = brotli.decompress(raw)  # no size limit
headroom/proxy/helpers.py:1063: MAX_REQUEST_BODY_SIZE = 100 * 1024 * 1024  # checks COMPRESSED size only
headroom/proxy/handlers/openai.py:3272: if content_length and int(content_length) > MAX_REQUEST_BODY_SIZE:  # guard on wire size
```

### Verification

Adversarially verified (GLM): confidence 0.95. The code clearly shows decompression of gzip/deflate/zstd/brotli request bodies with no cap on decompressed output size, while the only size guard checks the compressed wire Content-Length against 100 MB. A zip bomb (e.g., 10 GB of zeros compressing to ~10 MB) would pass the wire-size check and cause OOM during decompression, enabling unauthenticated remote denial of service.

---

*Reported by [OpenClaw BountyBot](https://github.com/yajat-kapur-failsafe/openclaw-bountybot) via Failsafe Nexus (Pandora) automated security analysis. Please review carefully before acting.*

chopratejas:
## Description

Closes #3284.

The handlers gate on `Content-Length` — the **compressed** wire size — against `MAX_REQUEST_BODY_SIZE`, but every codec then decompressed with a one-shot call (`gzip.decompress`, `zlib.decompress`, `brotli.decompress`, `stream_reader.read()`) that materializes the whole expansion before anything can inspect it. A client could buy unlimited capacity with a compression ratio.

I reproduced it end to end. `zstd` is considerably worse than the reported gzip case:

| codec | wire size | expands to |
|---|---|---|
| gzip | 1.99 MB | 2 GB |
| zstd | **16 KB** | 512 MB |

At 16 KB the attack costs essentially no bandwidth, and one request is enough.

All four codecs converge on a single chokepoint, `_read_request_body_bytes`, which `_read_request_json` also funnels through — so one change covers every caller.

## Type of Change

## Changes Made

- Added `MAX_DECOMPRESSED_BODY_SIZE`, set to `MAX_REQUEST_BODY_SIZE` deliberately: nobody should get more room by arriving compressed.
- Decompress incrementally, re-checking the cap after every 64 KiB chunk, so peak allocation is the cap plus one chunk rather than the full expansion:
  - **gzip/deflate** — `zlib.decompressobj` with a `max_length`, which leaves unproduced output in `unconsumed_tail` instead of memory.
  - **zstd** — `stream_reader` read in chunks.
  - **brotli** — `Decompressor` fed in slices.
- `RequestBodyTooLarge` subclasses `ValueError`, so every existing `except ValueError` call site keeps answering 400 unchanged while a caller that would rather answer 413 has a type to branch on. Re-raised ahead of the generic handlers so the size refusal isn't reworded into a vague "failed to decompress".

Behaviour deliberately preserved:

- **Multi-member gzip** still concatenates, matching `gzip.decompress`. A plain `decompressobj` stops after the first member, which would have silently truncated such bodies.
- **Truncated streams still raise** rather than yielding a short body — losing the one-shot call must not turn a corrupt request into a silently empty one.
- An **old brotli** without `Decompressor` now fails closed with a clear message instead of falling back to the unbounded call. An outdated library is not a reason to reopen the hole.

## Testing

- [x] Linting passes (CI-pinned `ruff` 0.16.3)

### Test Output

```text
$ pytest tests/test_request_body_decompression_limits.py
16 passed, 2 skipped in 1.20s          # 2 skipped = brotli not installed locally

$ pytest tests/test_proxy/ tests/test_proxy_hardening.py tests/test_request_body_decompression_limits.py
311 passed, 2 skipped in 48.45s

$ pytest tests/ -k "body or compression or forwarding or passthrough or helpers or batch or gemini"
1538 passed, 201 skipped, 10589 deselected in 85.42s

$ uvx ruff@0.16.3 check headroom/proxy/helpers.py tests/test_request_body_decompression_limits.py
All checks passed!
$ mypy headroom/proxy/helpers.py
```

The tests assert the property that actually matters — **refused without being materialized** — not merely "an oversized body is rejected". A guard placed *after* a one-shot `decompress()` would satisfy the latter while still OOM-ing the process, so `test_bomb_never_materializes` asserts on peak allocation via `tracemalloc`.

Verified they catch the bug by swapping the old one-shot implementation back in behind the same signature:

```text
FAILED test_gzip_bomb_is_refused
FAILED test_deflate_bomb_is_refused
FAILED test_bomb_never_materializes[_gzip_bomb-<lambda>]
FAILED test_bomb_never_materializes[_deflate_bomb-<lambda>]
FAILED test_truncated_gzip_still_errors
FAILED test_reader_refuses_a_bomb
6 failed, 10 passed, 2 skipped
```

## Real Behavior Proof

- **Environment:** macOS arm64, Python 3.12.13, `main` @ 0.37.0.
- **Exact command / steps:** build real bombs by streaming a compressor (so the fixture itself never holds the expansion), then call `_read_request_body_bytes` through a stand-in Request, measuring peak allocation with `tracemalloc`.
- **Observed result**, 400 MB bomb from 398 KB of wire:

```text
OLD gzip.decompress()  peak = 829.8 MB   (materialized the whole bomb)
NEW bounded            peak = 105.0 MB   (refused at the cap)
```

  A 2 GB bomb is refused in 185 ms with peak 108.2 MB. End to end through the real entry point, both gzip and zstd bombs raise `RequestBodyTooLarge`, while ordinary compressed bodies, identity bodies, and unknown-encoding rejection all behave exactly as before.

- **Not tested:** brotli is not installed in this environment, so its two tests skip locally and are exercised in CI where the extra is present. The implementation is the same shape as the others, but I have not run it.

## Known gap, deliberately not in scope

The `Content-Length` gate is skipped entirely for **chunked** requests (no `Content-Length` header), so `await request.body()` will still read an arbitrarily large *compressed* body into memory. That is a real but much weaker vector — it needs bandwidth proportional to the damage, with no amplification, which is what made the bomb dangerous. Bounding the raw read affects every request path including legitimate large uploads, so it deserves its own change rather than riding along here. Filed as #3326.

## Runtime Rollout Safety

- **Rollout-managed feature(s):** none.
- **Stable/default behavior changed:** no for any body under the cap — bodies round-trip byte-identically, including multi-member gzip. The only behavior change is that a body decompressing past 100 MB is now refused instead of being allocated, which is the bug.
- **Kill switch / disable path:** none by design; raising `MAX_DECOMPRESSED_BODY_SIZE` widens the ceiling if an operator genuinely needs it.
- **Unsafe override required:** none.
- **Qualification impact:** none.
- **Rollback path:** revert this commit — one function plus three helpers in a single file.

## Review Readiness

### DIFF
--- a/headroom/proxy/helpers.py
+++ b/headroom/proxy/helpers.py
@@ -1062,6 +1062,18 @@ def append_text_to_latest_user_input_item(
 # Maximum request body size (100MB - increased to support image-heavy requests)
 MAX_REQUEST_BODY_SIZE = 100 * 1024 * 1024
 
+# A *decompressed* body obeys the same ceiling as an uncompressed one. The
+# Content-Length gate at the handlers only ever saw the compressed wire size, so
+# a client could buy unlimited extra capacity with a compression ratio: ~2MB of
+# zeros expands to 2GB and the process dies allocating it (#3284). Same number,
+# deliberately: nobody should get more room by arriving gzipped.
+MAX_DECOMPRESSED_BODY_SIZE = MAX_REQUEST_BODY_SIZE
+
+# How much decompressed output to pull per step. The cap is re-checked after
+# every chunk, so the peak allocation is the limit plus one chunk — never the
+# full expansion of a bomb.
+_DECOMPRESS_CHUNK_SIZE = 64 * 1024
+
 # Maximum SSE buffer size (10MB - prevents memory exhaustion from malformed streams)
 MAX_SSE_BUFFER_SIZE = 10 * 1024 * 1024
 
@@ -2537,24 +2549,130 @@ def apply_session_sticky_ccr_tool(
     return tools_out, True
 
 
+class RequestBodyTooLarge(ValueError):
+    """A decompressed request body exceeded :data:`MAX_DECOMPRESSED_BODY_SIZE`.
+
+    Subclasses ``ValueError`` so every existing ``except ValueError`` call site
+    keeps answering 400 unchanged, while giving a caller that would rather
+    answer 413 a type to branch on.
+    """
+
+
+def _inflate_bounded(raw: bytes, *, wbits: int, label: str, multi_member: bool = False) -> bytes:
+    """Incrementally inflate ``raw``, stopping the instant output passes the cap.
+
+    ``zlib.decompress``/``gzip.decompress`` materialize the whole expansion
+    before anything can inspect it, which is what makes a bomb fatal. Feeding
+    the stream through ``decompressobj`` with a ``max_length`` keeps unproduced
+    output in ``unconsumed_tail`` instead of memory, so the check below runs
+    before the allocation rather than after it.
+    """
+    import zlib
+
+    out = bytearray()
+    pending = raw
+    while True:
+        decompressor = zlib.decompressobj(wbits)
+        while True:
+            chunk = decompressor.decompress(pending, _DECOMPRESS_CHUNK_SIZE)
+            out += chunk
+            if len(out) > MAX_DECOMPRESSED_BODY_SIZE:
+                raise RequestBodyTooLarge(
+                    f"Decompressed {label} request body exceeds "
+                    f"{MAX_DECOMPRESSED_BODY_SIZE // (1024 * 1024)}MB"
+                )
+            pending = decompressor.unconsumed_tail
+            if decompressor.eof or not pending:
+                break
+            if not chunk:
+                # Input left over but nothing produced: the stream cannot
+                # advance, and looping again would spin forever.
+                raise ValueError(f"Failed to decompress {label} request body: stalled stream")
+        out += decompressor.flush()
+        if len(out) > MAX_DECOMPRESSED_BODY_SIZE:
+            raise RequestBodyTooLarge(
+                f"Decompressed {label} request body exceeds "
+                f"{MAX_DECOMPRESSED_BODY_SIZE // (1024 * 1024)}MB"
+            )
+        if not decompressor.eof:
+            raise ValueError(f"Failed to decompress {label} request body: truncated stream")
+        # gzip streams may carry several members; `gzip.decompress` concatenates
+        # them, so restart on the trailer to keep that behavior.
+        pending = decompressor.unused_data
+        if not (multi_member and pending):
+            break
+    return bytes(out)
+
+
+def _zstd_bounded(raw: bytes) -> bytes:
+    """Read a zstd frame in chunks so a high-ratio frame cannot outrun the cap."""
+    import zstandard
+
+    out = bytearray()
+    dctx = zstandard.ZstdDecompressor()
+    with dctx.stream_reader(raw) as reader:
+        while True:
+            chunk = reader.read(_DECOMPRESS_CHUNK_SIZE)
+            if not chunk:
+                break
+            out += chunk
+            if len(out) > MAX_DECOMPRESSED_BODY_SIZE:
+                raise RequestBodyTooLarge(
+                    f"Decompressed zstd request body exceeds "
+                    f"{MAX_DECOMPRESSED_BODY_SIZE // (1024 * 1024)}MB"
+                )
+    return bytes(out)
+
+
+def _brotli_bounded(raw: bytes) -> bytes:
+    """Feed brotli input in slices, checking the cap after each one."""
+    import brotli
+
+    decompressor_cls = getattr(brotli, "Decompressor", None)
+    if decompressor_cls is None:
+        # Fail closed rather than fall back to the unbounded one-shot call: an
+        # old library is not a reason to reopen the hole.
+        raise ValueError(
+            "Installed 'brotli' is too old for incremental decompression "
+            "(needs brotli.Decompressor); upgrade it to accept br request bodies."
+        )
+
+    decompressor = decompressor_cls()
+    out = bytearray()
+    for start in range(0, len(raw), _DECOMPRESS_CHUNK_SIZE):
+        out += decompressor.process(raw[start : start + _DECOMPRESS_CHUNK_SIZE])
+        if len(out) > MAX_DECOMPRESSED_BODY_SIZE:
+            raise RequestBodyTooLarge(
+                f"Decompressed brotli request body exceeds "
+                f"{MAX_DECOMPRESSED_BODY_SIZE // (1024 * 1024)}MB"
+            )
+    is_finished = getattr(decompressor, "is_finished", None)
+    if is_finished is not None and not is_finished():
+        raise ValueError("Failed to decompress brotli request body: truncated stream")
+    return bytes(out)
+
+
 async def _read_request_body_bytes(request: Request) -> bytes:
     """Read and (if needed) decompress the request body, returning raw UTF-8 bytes.
 
     Mirrors ``_read_request_json`` but returns the bytes pre-parse so
     forwarders can implement byte-faithful passthrough (PR-A3, fixes P0-2).
-    Raises ``ValueError`` on any decompression failure.
+    Raises ``ValueError`` on any decompression failure, and the
+    :class:`RequestBodyTooLarge` subclass when the *decompressed* body would
+    exceed :data:`MAX_DECOMPRESSED_BODY_SIZE`.
     """
     encoding = (request.headers.get("content-encoding") or "").lower().strip()
     raw = await request.body()
 
+    # Every branch below decompresses incrementally against
+    # MAX_DECOMPRESSED_BODY_SIZE. RequestBodyTooLarge is re-raised ahead of the
+    # generic handlers so the size refusal is not reworded into a vague
+    # "failed to decompress" (#3284).
     if encoding in ("zstd", "zstandard"):
         try:
-            import zstandard
-
-            dctx = zstandard.ZstdDecompressor()
-            reader = dctx.stream_reader(raw)
-            raw = reader.read()
-            reader.close()
+            raw = _zstd_bounded(raw)
+        except RequestBodyTooLarge:
+            raise
         except ImportError:
             raise ValueError(
                 "Request body is zstd-compressed but the 'zstandard' package is not installed. "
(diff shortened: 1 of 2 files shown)


chopratejas:
Self-review notes — differential testing found two real compatibility breaks in my own patch, both now fixed in the follow-up commit.

**Method:** rather than only testing that bombs are refused, I ran the bounded reader against the one-shot call it replaces across valid / empty / truncated / header-only / multi-member / padded / trailing-garbage / random / wrong-codec inputs, for all three codec families, and required every case to either return identical bytes or error on both sides.

Two diverged:

```
case                 OLD                     NEW (before fix)
trailing NULs        ok: payload             zlib.error       *** DIVERGES ***
empty input          ok: b''                 ValueError       *** DIVERGES ***
```

**Trailing NUL padding is the one that matters.** gzip is a *sequence* of members, and CPython's reader skips NULs between and after them — `gzip.decompress(member + b"\x00" * 16)` returns the payload. My member-restart loop tried to parse that padding as a fresh member and raised. Real clients emit this padding, so it would have rejected bodies that worked before, and no amount of bomb testing would have surfaced it. Fixed by skipping padding before each restart; a body that is *only* padding still fails, matching CPython's `BadGzipFile`. Pinned by tests that assert the old behavior alongside the new.

Deflate needed no change — old and new already agreed on every case, including raw-deflate and trailing garbage.

**Peak memory on legitimate traffic, which the original PR did not measure.** I was concerned the `bytearray` → `bytes` copy might have doubled peak for large valid bodies. It did not — an 80MB body now peaks *lower* than before:

```
OLD gzip.decompress    len=80MB  peak=189.7MB
NEW bounded            len=80MB  peak=162.3MB
```

**Brotli is no longer untested.** I installed it side-by-side and exercised the real function: round-trip, empty payload, truncated (ValueError), and a bomb — whose ratio is the most extreme of the three, **0.1 KB of wire expanding to 64 MB**. Differential against `brotli.decompress` agrees on valid/empty/random. The two brotli tests still skip where the extra is absent, and run in CI where it is present.

Also tightened two `pytest.raises(Exception)` assertions ruff flagged (B017) to the real types — `gzip.BadGzipFile` and `zlib.error` — since a blind `Exception` would have passed for the wrong reason.

22 passed with brotli, 20 passed + 2 skipped without, 295 across `tests/test_proxy/` and the hardening suite; ruff 0.16.3 and mypy clean.

PR #3338, opened 2026-08-28
fix(proxy): gate the raw request body, not just its Content-Length header

### LINKED ISSUE #3326: Chunked requests skip the Content-Length body-size gate entirely
Split out of #3284 / PR #3325, where this was found and deliberately left out of scope.

The request-body size gate is written as:

```python
if content_length and int(content_length) > MAX_REQUEST_BODY_SIZE:
    return 413
```

(`handlers/openai.py:3264`, `:5324`, `handlers/batch.py:64`, `handlers/gemini.py:280`.)

A request sent with `Transfer-Encoding: chunked` carries no `Content-Length`, so `content_length` is falsy and the gate is skipped. `await request.body()` then reads the whole body into memory with no ceiling.

**This is genuinely weaker than the zip bomb** and should be prioritised accordingly. There is no amplification: an attacker must actually send N bytes to make the proxy hold N bytes, so it is ordinary bandwidth-bound resource exhaustion rather than the 32,000:1 leverage that made #3284 severe. #3325 already bounds the *decompressed* size on this path, so a chunked **compressed** body is capped after inflation regardless — this issue is only about the raw read.

Left out of #3325 on purpose: bounding the raw read touches every request path, including legitimate large uploads (image-heavy requests are exactly why `MAX_REQUEST_BODY_SIZE` is 100MB), so it wants its own change and its own thought about streaming clients rather than riding along in a security fix.

Sketch of a fix, if someone picks it up: read the stream incrementally via `request.stream()` and abort past `MAX_REQUEST_BODY_SIZE` instead of relying on a header the client controls — which also closes the case of a **lying** `Content-Length` (a small declared value with a large body), not just a missing one.

Low priority.

AmirF194:
## Description

Closes #3326.

The four handlers named in the issue (`openai.py`, `batch.py`, `gemini.py`,
`anthropic.py`) all gate on the `Content-Length` header before calling
`_read_request_json`, but that header is only present when the client sends
one. A chunked-transfer-encoding request has no `Content-Length`, so it
walks straight past every one of those checks, and `_read_request_body_bytes`
then reads the whole body through the unbounded `request.body()`.

`_read_request_json` and `_read_request_body_bytes` are the single
chokepoint all six call sites route through (confirmed by reading each of
the six named lines, plus a repo-wide grep for other `_read_request_json`
callers), so one change at that chokepoint covers all of them.

## Type of Change

## Changes Made

- Added `_read_request_body_bounded` in `headroom/proxy/helpers.py`: reads
  the body via `Request.stream()` instead of `Request.body()`, refusing the
  request with `RequestBodyTooLarge` once the running total crosses
  `MAX_REQUEST_BODY_SIZE`, before the rest of the body arrives.
- `_read_request_body_bytes` now reads through it instead of calling
  `request.body()` directly.

## Testing

### Test Output

```text
$ pytest tests/test_request_body_decompression_limits.py -q
25 passed, 2 skipped in 4.25s   # 2 skipped = brotli not installed in this environment

$ pytest tests/ -k "helpers or request_body or decompress or proxy_body or size_limit or content_length" -q
246 passed, 12 skipped, 3 failed in 62.57s
# The 3 failures are in tests/test_cli/test_wrap_helpers.py::TestProxyClientRefCounting
# (process ref-counting / PID checks) and reproduce identically against
# unmodified main in this sandboxed container, unrelated to this change.

$ ruff check headroom/proxy/helpers.py tests/test_request_body_decompression_limits.py tests/test_proxy_handler_helpers.py
All checks passed!
$ ruff format --check headroom/proxy/helpers.py tests/test_request_body_decompression_limits.py tests/test_proxy_handler_helpers.py
3 files already formatted
```

`mypy headroom --ignore-missing-imports` did not run cleanly in this
environment: it fails at `numpy/__init__.pyi` with "Type statement is only
supported in Python 3.12 and greater" before reaching this file, a
pre-existing mismatch between the installed numpy stubs and this repo's
pinned `python_version = "3.10"` mypy target, not something this diff
introduces.

One existing test needed a fixture update: `test_proxy_handler_helpers.py`'s
`_VertexGeminiImageRequest` stand-in only implemented `.body()`, and the new
code path calls `.stream()` (a real Starlette `Request` always has both).
Added a `.stream()` method that yields the same body in one chunk.

## Real Behavior Proof

- Environment: Docker `python:3.12-slim`, headroom-ai 0.37.0's published
  wheel (for the compiled `_core` extension) with this branch's
  `headroom/proxy/helpers.py` overlaid on top.
- Exact command / steps: built a genuine Starlette `Request` over a hand-rolled
  chunked ASGI `receive()` channel (no `content-length` header, matching what
  a real chunked-transfer-encoding request looks like at the ASGI layer), and
  called the actual `_read_request_body_bytes` with `MAX_REQUEST_BODY_SIZE`
  turned down to 10KB so the demonstration does not need to stream tens of
  megabytes.
- Observed result: against unmodified `main`, a 50,000-byte chunked body with
  no `Content-Length` header is read in full with no error. Against this
  branch, the same request raises `RequestBodyTooLarge` before the stream
  finishes, and a 5,000-byte chunked body under the cap still round-trips
  byte for byte.
- Not tested: brotli-encoded bodies (package not installed in this sandbox,
  same as the two skipped unit tests above), and the repo's own CI matrix
  (Rust extension build, HuggingFace model prefetch) end to end.

## Runtime Rollout Safety

- Stable/default behavior changed: no for any request under
  `MAX_REQUEST_BODY_SIZE` (100MB by default); a chunked request that would
  previously have been read unbounded is now refused past that size, which
  is the fix.
- Kill switch / disable path: none by design; raising `MAX_REQUEST_BODY_SIZE`
  widens the ceiling if an operator needs to.
- Qualification impact: none.
- Rollback path: revert this commit, one function plus a two-line call-site
  change in a single file.

## Review Readiness

## Checklist

## Additional Notes

Not in scope: a handful of other call sites in these same handler files
(`batch.py`'s and `bedrock.py`'s passthrough forwarders, and two streaming
passthrough paths in `openai.py`/`anthropic.py`) also call
`request.body()` directly rather than through this chokepoint. None of them
are among the six lines the issue names, they exist to forward a request
byte-for-byte rather than to parse it as JSON, and fixing them is a larger,
separate change than this one. Flagging it here rather than leaving it
implicit.

### DIFF
--- a/headroom/proxy/handlers/bedrock.py
+++ b/headroom/proxy/handlers/bedrock.py
@@ -76,9 +76,12 @@ class BedrockHandlerMixin:
         from headroom.proxy.helpers import (
             COMPRESSION_TIMEOUT_SECONDS,
             MAX_MESSAGE_ARRAY_LENGTH,
+            MAX_REQUEST_BODY_SIZE,
+            RequestBodyTooLarge,
             _headroom_bypass_enabled,
             _strip_internal_headers,
             extract_tags,
+            read_cached_request_body,
             read_request_json_with_bytes,
         )
         from headroom.proxy.modes import is_cache_mode
@@ -134,13 +137,34 @@ class BedrockHandlerMixin:
             if isinstance(err, ClientDisconnect):
                 logger.debug("[%s] %s client disconnected during body read", request_id, LOG_TAG)
                 return Response(status_code=204)
+            if isinstance(err, RequestBodyTooLarge):
+                # The read aborted mid-stream once the size ceiling was
+                # crossed, so there is no raw body to fail open with, and
+                # forwarding one anyway would defeat the ceiling. Reject.
+                logger.warning("[%s] %s request body too large: %s", request_id, LOG_TAG, err)
+                return JSONResponse(
+                    status_code=413,
+                    content={
+                        "error": {
+                            "type": "request_too_large",
+                            "message": (
+                                f"Request body too large. Maximum size is "
+                                f"{MAX_REQUEST_BODY_SIZE // (1024 * 1024)}MB"
+                            ),
+                        }
+                    },
+                )
             logger.warning(
                 "[%s] %s could not parse body; forwarding verbatim: %s",
                 request_id,
                 LOG_TAG,
                 err,
             )
-            raw_only = await request.body()
+            # The raw bytes were already fully read (and cached) by
+            # read_request_json_with_bytes before it failed decoding them;
+            # request.body() here would re-drain an already-consumed stream
+            # and raise RuntimeError("Stream consumed").
+            raw_only = read_cached_request_body(request) or b""
             return await self._forward_bedrock(
                 url=url,
                 headers=verbatim_headers,
--- a/headroom/proxy/helpers.py
+++ b/headroom/proxy/helpers.py
@@ -2550,7 +2550,7 @@ def apply_session_sticky_ccr_tool(
 
 
 class RequestBodyTooLarge(ValueError):
-    """A decompressed request body exceeded :data:`MAX_DECOMPRESSED_BODY_SIZE`.
+    """A raw or decompressed request body exceeded its size ceiling.
 
     Subclasses ``ValueError`` so every existing ``except ValueError`` call site
     keeps answering 400 unchanged, while giving a caller that would rather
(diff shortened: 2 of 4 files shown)


AmirF194:
CI caught a real regression in my own diff: switching to `Request.stream()` for the bounded read means a second read on the same request raises `RuntimeError: Stream consumed`, which is exactly what the semaphore-leak backpressure tests do (they replay one request through the handler multiple times).

Pushed a fix: cache the bounds-checked bytes under our own attribute instead of touching the stream twice. Deliberately not Starlette's own `request._body` (that value comes from an unbounded `.body()` call, so trusting it would skip the size gate this function exists to add). Verified locally: the 4 previously-failing backpressure cases plus the full decompression-limits and proxy-handler-helpers suites all pass.

AmirF194:
The previous commit's fix for the double-read (Stream consumed) broke two things the local
verification didn't cover, since it ran a narrow test subset:

- Any caller whose request is a lightweight test double with no `.stream()` (there are
  several across the suite) hit an `AttributeError` instead of being read.
- The malformed-body fallback in `bedrock.py` called `request.body()` again after the
  bounded reader had already drained the stream, hitting the same `RuntimeError: Stream
  consumed` this PR exists to fix, just from a second call site.

Pushed 87efe2ec: `_read_request_body_bounded` now falls back to a single bound-checked
`.body()` read when `.stream` isn't present, and the bedrock fallback reads the
already-cached bytes instead of re-reading the stream (rejecting with 413 instead of
forwarding verbatim if the abort was actually a size violation, since nothing is cached in
that case). Ran the full suite this time, not just the two files from before: 159 passed, 2
pre-existing skips, ruff clean.

JerrettDavis (changes requested):
The current head is not green: the lint job's repository-wide mypy gate reports headroom/proxy/helpers.py:2706: Returning Any from function declared to return bytes [no-any-return]. This is in the new lightweight-test-double fallback of _read_request_body_bounded, where wait request.body() is inferred as Any. Please give that fallback a concrete bytes value/annotation (and preserve the explicit post-read size check) so the required type gate passes. The bounded streaming design and Bedrock cached-raw fail-open path otherwise look coherent on this pass.

AmirF194:
Fixed. The test-double fallback now casts the body to bytes before the size check, matching the cast() pattern already used a few functions up in this file for the same reason: Starlette isn't installed in the lint job, so mypy resolves request.body() as Any there.

Verified in a clean container with mypy 1.20.2 run exactly as the lint job does (mypy headroom --ignore-missing-imports, nothing but mypy installed): the no-any-return error at helpers.py:2706 is gone, no new errors, 528 source files checked clean. Confirmed the same setup reproduces the original error against the prior commit, so this is the right harness.

Also reran the full suite this PR touched (backpressure, decompression limits, proxy handler helpers, batch handlers, output-only blocks, bedrock passthrough) plus ruff check and format: 159 passed, 2 pre-existing skips, ruff clean. The post-read size check is untouched.

JerrettDavis (approved):
The follow-up fixes the requested type-gate issue by assigning the awaited test-double body to a concrete ytes value via cast, while retaining the explicit post-read size check. The bounded streaming path and cached-body behavior are unchanged. Focused validation on this head: 78 passed, 2 skipped. A standalone --follow-imports=skip mypy run reports unrelated existing Any/type-alias errors elsewhere in this large helper module, but the reviewed fallback return is no longer among them.

AmirF194:
Thanks for the thorough re-check! Good to hear the cast fix holds up under focused validation. Let me know if anything else comes up once the held checks run.

PR #3341, opened 2026-08-29
perf(memory): bound traffic-learner _persisted_ids with the dedup window

abhay-codes07:
## Description

`HeadroomTrafficLearner._persisted_ids` (a `dict[content_hash, memory_id]`) grew **one entry per distinct persisted pattern for the whole process lifetime** and was never trimmed, while its sibling dedup set `_saved_hashes` **is** trimmed to `dedup_window`:

```python
self._saved_hashes.add(h)
if len(self._saved_hashes) > self._dedup_window:
    self._saved_hashes.pop()          # trimmed
# ...
self._persisted_ids[pattern.content_hash] = memory_id   # never trimmed
```

`_persisted_ids` is only ever *read* behind an `if h in self._saved_hashes:` guard (the dedup shortcut at the top of `_accumulate`). So once a hash leaves the dedup window, its `_persisted_ids` entry can never be read again — it's pure dead weight on the long-lived learner (which owns background save/flush tasks). The asymmetry is the tell: `_saved_hashes` is trimmed and `_pattern_counts` is LRU-capped right next to this, but `_persisted_ids` was missed.

The fix keeps the two in lockstep:
- When `_saved_hashes` evicts a hash, drop its `_persisted_ids` entry in the same step.
- The async save worker records an id only while its hash is still tracked in `_saved_hashes` (it may have been evicted between enqueue and persist).

Neither changes the dedup/bump semantics — a hash evicted from `_saved_hashes` was already re-learnable and could already produce a duplicate row; these only stop `_persisted_ids` from leaking.

Reproduction (before): 200 distinct persisted patterns with `dedup_window=4` → `_persisted_ids` holds ~200 entries while `_saved_hashes` holds ≤4. After: `_persisted_ids` stays a subset of `_saved_hashes`, bounded to ≤4.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `headroom/memory/traffic_learner.py`: on `_saved_hashes` eviction in `_accumulate`, capture the popped hash and `_persisted_ids.pop(evicted, None)`; in `_save_worker`, guard the id record with `and pattern.content_hash in self._saved_hashes`.
- `tests/test_memory/test_traffic_learner.py`: added `test_persisted_ids_bounded_in_lockstep_with_saved_hashes` — feeding 200 distinct patterns through a `dedup_window=4` learner leaves `_persisted_ids` a subset of `_saved_hashes` and bounded to ≤4 (not 200).

## Testing

### Test Output

```text
tests/test_memory/test_traffic_learner.py  ->  153 passed, 2 failed
  (the 2 failures are pre-existing on main in TestProjectForPattern — project-root
   path matching, unrelated to this change; confirmed they fail without this patch)
uvx ruff@0.16.2 check headroom/memory/traffic_learner.py tests/test_memory/test_traffic_learner.py  ->  All checks passed!
uvx mypy@1.20.2 headroom/memory/traffic_learner.py  ->  Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: drove `_accumulate` with 200 distinct patterns (each corroborated to `min_evidence=1`) through a `dedup_window=4` learner, recording an id for each still-tracked hash. Without the fix `_persisted_ids` held ~200 entries and was not a subset of the 4-element `_saved_hashes` (test fails). With the fix it stays a subset and ≤4. Confirmed the two `TestProjectForPattern` failures pre-date this change by stashing the patch.
- Observed result: `_persisted_ids` no longer grows without bound; it tracks the dedup window. Dedup/bump behavior for in-window hashes is unchanged.
- Not tested: an end-to-end run against a live memory backend (the async `_save_worker` path needs a backend); the eviction/record invariant is exercised directly, and the worker guard mirrors it.

## Runtime Rollout Safety

- Stable/default behavior changed: no. Dedup shortcut and evidence-bump behavior are unchanged for any hash still in the dedup window; only stale `_persisted_ids` entries (already unreadable) are dropped.
- Qualification impact: none.
- Rollback path: revert this commit; `_persisted_ids` goes back to unbounded growth.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A: internal memory bound)

## Additional Notes

This mirrors the bounding already applied to `_saved_hashes` (dedup_window trim) and `_pattern_counts` (LRU cap) in the same class, closing the one companion structure that was left uncapped.

### DIFF
--- a/headroom/memory/traffic_learner.py
+++ b/headroom/memory/traffic_learner.py
@@ -1275,10 +1275,16 @@ class TrafficLearner:
             # Ready to save
             del self._pattern_counts[h]
             self._saved_hashes.add(h)
-            # Trim saved hashes to prevent unbounded growth
+            # Trim saved hashes to prevent unbounded growth, and drop the evicted
+            # hash's persisted-id entry in lockstep. ``_persisted_ids`` is only
+            # ever read behind an ``h in _saved_hashes`` guard (the dedup check
+            # above), so an id for a hash no longer tracked is dead weight;
+            # without this it grew one entry per distinct persisted pattern for
+            # the whole process lifetime while ``_saved_hashes`` stayed bounded.
             if len(self._saved_hashes) > self._dedup_window:
                 # Remove oldest (arbitrary, set is unordered, but prevents growth)
-                self._saved_hashes.pop()
+                evicted = self._saved_hashes.pop()
+                self._persisted_ids.pop(evicted, None)
 
             # Persist the real accumulated count, not the dataclass default.
             pattern.evidence_count = count
@@ -1310,9 +1316,13 @@ class TrafficLearner:
                     },
                 )
                 self._patterns_saved += 1
-                # Track id so future re-sightings bump this row.
+                # Track id so future re-sightings bump this row — but only while
+                # the hash is still within the dedup window. If it was evicted
+                # from ``_saved_hashes`` between enqueue and now, recording its id
+                # would re-leak an entry that can never be read again (the dedup
+                # read at the top of ``_accumulate`` is gated on ``_saved_hashes``).
                 memory_id = getattr(memory, "id", None)
-                if memory_id is not None:
+                if memory_id is not None and pattern.content_hash in self._saved_hashes:
                     self._persisted_ids[pattern.content_hash] = memory_id
                 logger.debug(f"Traffic learner saved pattern: {pattern.content[:80]}")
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The steady-state eviction path now removes _persisted_ids in lockstep, but startup hydration still repopulates both structures without applying dedup_window: _hydrate_persisted_state() iterates every matching database row and unconditionally inserts into _saved_hashes and _persisted_ids. A long-lived installation with many persisted traffic patterns therefore starts with both maps unbounded, and they remain oversized until enough new patterns happen to evict entries one at a time. Please cap hydration to the same window (with a deterministic retention policy if possible) and add a hydration regression test whose database contains more than dedup_window distinct rows; assert both collections are bounded and have matching keys after start()/hydration.

PR #3343, opened 2026-08-29
perf(compression/router): reuse apply()'s detection in compress(), not re-detect

abhay-codes07:
## Description

`ContentRouter.apply()` runs the native `_detect_content` on **every** routed message (in the routing loop, for its recent-code / analysis-intent protection checks), then queues cache-miss messages for the parallel compression pass — where `compress()` ran `_detect_content` **again** on the byte-identical content:

```python
# apply() routing loop (per message)
detection = _regex_detect_content_type(content) if force_kompress else _detect_content(content)
is_code = detection.content_type == ContentType.SOURCE_CODE
...
pending_tasks.append((i, content, context, msg_bias, content_key, enforce_reversibility))

# Pass 2 -> compress() (per cache-miss message)
detection = _detect_content(content)   # same bytes, detected a second time
```

`_detect_content` is the router's hottest per-message cost — the native Magika→unidiff→PlainText chain — and it is **not** memoized (there's no content-keyed cache; the sibling `_determine_strategy` already takes a precomputed `detection=` for exactly this reason). So every cache-miss message on the content-aware path was detected twice.

This threads the detection already computed in the routing loop through the pending-task tuple into `compress()` via a new `precomputed_detection` argument. Only the full detection is reused; under `force_kompress` the loop used the lightweight regex detector and `compress()` skips detection entirely, so `None` is passed there. A cache-miss message is now detected **once**. Output is byte-identical: detection is a pure function of the content, and `compress()` still detects itself when no result is passed (e.g. its direct callers).

This complements the existing `compress`↔`_determine_strategy` dedup (already guarded in `test_content_router_detection_dedup.py`) — that removed the second detection *inside* `compress`; this removes the duplicate detection *between* `apply` and `compress`.

Verified on a cache-miss tool message: `_detect_content` drops from **2 calls to 1**; compressed output is byte-identical across log / code / JSON payloads.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `headroom/transforms/content_router.py`:
  - `compress()` gains `precomputed_detection: DetectionResult | None = None`; when supplied it is reused instead of calling `_detect_content` (the `force_kompress` and empty-content branches are unchanged).
  - `_timed_compress()` (parallel executor entry) gains the same optional argument and forwards it.
  - The `_PendingTask` tuple carries a 7th field (the detection), populated in the routing loop (`None` under `force_kompress`) and passed to `compress()`/`_timed_compress()` in all three Pass-2 execution paths (single-inline, single-with-watchdog, thread pool). The Pass-3 merge unpack absorbs the extra field.
- Tests:
  - `tests/test_content_router_detection_dedup.py`: added `test_apply_runs_detection_once_per_cache_miss_message` (apply drives exactly one `_detect_content` for a cache-miss message) and `test_compress_precomputed_detection_matches_recomputed` (passing a precomputed detection yields identical strategy + output); extended the module docstring.
  - `tests/test_transforms_content_router.py`, `tests/test_content_router_single_item_deadline.py`, `tests/test_dataset_recall_runner.py`: updated the `router.compress` stub signatures to accept the new optional `precomputed_detection` keyword (they are invoked through `apply()`).

## Testing

### Test Output

```text
tests/test_transforms_content_router.py tests/test_content_router_*.py tests/test_dataset_recall_runner.py  ->  103 passed
uvx ruff@0.16.2 check headroom/transforms/content_router.py <touched test files>  ->  All checks passed!
uvx mypy@1.20.2 headroom/transforms/content_router.py  ->  Success: no issues found in 1 source file
```

## Real Behavior Proof

- Exact command / steps: wrapped `_detect_content` with a call counter and ran `apply()` on one cache-miss tool message — 2 calls before this change, 1 after. Hashed `apply()`'s output across mixed log/code/JSON tool payloads with and without the change: identical (`d730248d…`). Ran the full content-router test suite (103 passed) and confirmed the two new tests fail with the change reverted.
- Observed result: one native detection per cache-miss message instead of two; identical routing and compressed output.
- Not tested: a production end-to-end throughput measurement (the per-message detection count and output equivalence are demonstrated directly).

## Runtime Rollout Safety

- Stable/default behavior changed: no. Compressed output is byte-identical; only the redundant second detection per cache-miss message is removed. `compress()`'s existing callers (which pass no `precomputed_detection`) are unchanged.
- Qualification impact: none (byte-identical output).
- Rollback path: revert this commit; `compress()` goes back to detecting the content itself in the cache-miss pass.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A: internal hot-path optimization, output unchanged)

## Additional Notes

The detection is threaded explicitly through the task tuple (rather than a shared cache) so it stays correct under the parallel thread-pool compression path — no shared mutable state. Only the full `_detect_content` result is reused; the `force_kompress` regex path passes `None`, matching `compress()`'s own `force_kompress` short-circuit.

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -2161,11 +2161,17 @@ class ContentRouter(Transform):
             logger.debug("TOIN recording failed (non-fatal): %s", e)
 
     def _timed_compress(
-        self, content: str, context: str, bias: float
+        self,
+        content: str,
+        context: str,
+        bias: float,
+        precomputed_detection: DetectionResult | None = None,
     ) -> tuple[RouterCompressionResult, float]:
         """Compress with wall-clock timing.  Used by parallel executor."""
         t0 = time.perf_counter()
-        result = self.compress(content, context=context, bias=bias)
+        result = self.compress(
+            content, context=context, bias=bias, precomputed_detection=precomputed_detection
+        )
         return result, (time.perf_counter() - t0) * 1000
 
     def compress(
@@ -2174,6 +2180,7 @@ class ContentRouter(Transform):
         context: str = "",
         question: str | None = None,
         bias: float = 1.0,
+        precomputed_detection: DetectionResult | None = None,
     ) -> RouterCompressionResult:
         """Compress content using optimal strategy based on content detection.
 
@@ -2183,6 +2190,12 @@ class ContentRouter(Transform):
             question: Optional question for QA-aware compression. When provided,
                 tokens relevant to answering this question are preserved.
             bias: Compression bias multiplier (>1 = keep more, <1 = keep fewer).
+            precomputed_detection: A ``_detect_content(content)`` result the
+                caller already computed for the same content. When supplied, it
+                is reused instead of re-running the native detection chain — the
+                router's hottest per-message cost. ``apply`` computes detection
+                once per message (for its code-protection checks) and passes it
+                here so a cache-miss message is not detected twice.
 
         Returns:
             RouterCompressionResult with compressed content and routing metadata.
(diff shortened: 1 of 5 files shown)


JerrettDavis (approved):
Reviewed the latest head across the inline, watchdog, and thread-pool paths. The full detection is carried only for the byte-identical cache-miss content, while force-Kompress retains its lightweight/short-circuit behavior; direct compress() callers still recompute by default. CI is green and the equivalence/call-count coverage is appropriate. No blocking issues found.

PR #3365, opened 2026-08-31
fix(proxy): scope OpenAI rate limits by api-key

### LINKED ISSUE #3364: [BUG] OpenAI api-key clients share one rate-limit bucket
## Summary

OpenAI Chat Completions and Responses requests authenticated with an `api-key` header all consume the same rate-limit bucket. A request using one key can therefore cause a different key's valid request to receive Headroom's HTTP 429.

This is a functional multi-client isolation bug, not a security report.

## Affected versions

- Released `v0.37.0`
- Current `main` at `7b3d226598fec88b5766bcf171cdd7f2a5d1cd3b`

## Reproduction

Configuration:

```python
ProxyConfig(
    optimize=False,
    cache_enabled=False,
    rate_limit_enabled=True,
    rate_limit_requests_per_minute=1,
    rate_limit_tokens_per_minute=100_000,
)
```

Start a local OpenAI-compatible upstream, then send the same non-streaming `/v1/chat/completions` request twice through the real FastAPI proxy and real HTTP transport:

1. First request: `api-key: azure-key-A`
2. Second request: `api-key: azure-key-B`

Observed output:

```text
results=[
  ('azure-key-A', 200, None),
  ('azure-key-B', 429, 'Rate limited. Retry after 59.7s')
]
upstream_calls=1
```

## Expected behavior

Each API credential has an independent bucket. Both requests should return HTTP 200 and the upstream should receive two calls.

## Actual behavior

The first request consumes the shared `default` bucket. The second credential is rejected by Headroom before reaching the upstream.

## Root cause

Both OpenAI handlers derive the rate-limit key only from `Authorization`:

```python
rate_key = headers.get("authorization", "default")[:20]
```

When authentication is supplied through `api-key`, `Authorization` is absent and every credential resolves to `default`. The affected sites are the Chat Completions and Responses handlers.

## Impact

Deployments using header-based OpenAI-compatible gateways can see unrelated clients rate-limit one another. With the default limiter enabled, this produces incorrect HTTP 429 responses during otherwise valid traffic.

## Duplicate search

I searched open and closed issues for `api-key rate limit`, `Azure rate limit`, `shared rate limit bucket`, `rate_key authorization default`, and related terms and found no matching report.

C0sm0cats:
## Description

Scope OpenAI rate-limit buckets by either `Authorization` or `api-key`. OpenAI-compatible clients that authenticate only through `api-key` previously all consumed the shared `default` bucket and could incorrectly rate-limit one another.

Closes #3364

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add one OpenAI rate-limit identity helper with bearer-token precedence, `api-key` fallback, and an opaque SHA-256 identity derived from the complete credential.
- Use the helper in Chat Completions and Responses handlers.
- Add handler-level regressions proving distinct `api-key` credentials receive independent buckets on both endpoints.

## Testing

### Test Output

```text
$ .venv/bin/pytest -q tests/test_proxy_openai_rate_limit_key.py tests/test_proxy_openai.py tests/test_proxy_openai_responses_bypass.py
27 passed, 1 warning in 3.99s

$ .venv/bin/ruff check headroom/proxy/handlers/openai.py tests/test_proxy_openai_rate_limit_key.py
All checks passed!

$ .venv/bin/ruff format --check headroom/proxy/handlers/openai.py tests/test_proxy_openai_rate_limit_key.py
2 files already formatted

$ .venv/bin/mypy --follow-imports=skip headroom/proxy/handlers/openai.py
```

## Real Behavior Proof

- Environment: Linux, Python 3.14, current `main` plus this patch, real FastAPI proxy app and real local OpenAI-compatible HTTP upstream, RPM set to 1.
- Exact command / steps: started a local HTTP upstream with a call counter, then sent the same `/v1/chat/completions` body twice using `api-key: azure-key-A` and `api-key: azure-key-B`.
- Observed result: both credentials received HTTP 200 and independently reached the upstream.

```text
HTTP Request: POST http://127.0.0.1:38689/v1/chat/completions "HTTP/1.0 200 OK"
HTTP Request: POST http://127.0.0.1:38689/v1/chat/completions "HTTP/1.0 200 OK"
results=[('azure-key-A', 200, None), ('azure-key-B', 200, None)]
upstream_calls=2
```

- Not tested: external/cloud gateways; full repository test suite. A full-import mypy invocation currently reaches three pre-existing `no-any-return` errors in `headroom/perf/analyzer.py`; the changed handler passes isolated type checking.

## Runtime Rollout Safety

- Rollout-managed feature(s): built-in OpenAI proxy rate limiting.
- Stable/default behavior changed: only requests with `api-key` and no `Authorization` move from the shared `default` bucket to a credential-specific bucket.
- Kill switch / disable path: disable rate limiting with `--no-rate-limit` or `rate_limit_enabled=False`.
- Qualification impact: OpenAI-compatible `api-key` clients no longer consume one another's RPM allowance.

## Review Readiness

## Checklist

N/A.

## Additional Notes

Documentation is unchanged because this makes the existing per-key rate-limit contract work for the already-forwarded `api-key` authentication form. The warning above is the existing Starlette `TestClient` deprecation warning.

### DIFF
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -120,11 +120,20 @@ def _openai_rate_limit_key(headers: dict[str, str]) -> str:
     """Return the credential identity used by the OpenAI rate limiter.
 
     OpenAI-compatible gateways may authenticate with either a bearer token or
-    an ``api-key`` header. Requests without either retain the existing shared
-    fallback bucket.
+    an ``api-key`` header. Hash the complete value so common credential prefixes
+    do not merge buckets and raw credential material is not retained in memory.
+    Requests without either retain the existing shared fallback bucket.
     """
-    credential = headers.get("authorization") or headers.get("api-key") or "default"
-    return credential[:20]
+    authorization = headers.get("authorization")
+    api_key = headers.get("api-key")
+    if authorization:
+        kind, credential = "authorization", authorization
+    elif api_key:
+        kind, credential = "api-key", api_key
+    else:
+        return "default"
+    digest = hashlib.sha256(f"{kind}:{credential}".encode()).hexdigest()
+    return f"{kind}:{digest}"
 
 
 def _response_ccr_hashes(messages: list[dict[str, Any]], markers: list[str]) -> list[str]:
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -116,6 +116,17 @@ _CCR_HASH_RE = re.compile(
 _BARE_CCR_HASH_RE = re.compile(r"[a-fA-F0-9]{12,24}")
 
 
+def _openai_rate_limit_key(headers: dict[str, str]) -> str:
+    """Return the credential identity used by the OpenAI rate limiter.
+
+    OpenAI-compatible gateways may authenticate with either a bearer token or
+    an ``api-key`` header. Requests without either retain the existing shared
+    fallback bucket.
+    """
+    credential = headers.get("authorization") or headers.get("api-key") or "default"
+    return credential[:20]
+
+
 def _response_ccr_hashes(messages: list[dict[str, Any]], markers: list[str]) -> list[str]:
     """Return the distinct retrievable CCR hashes exposed by a response.
 
@@ -3520,7 +3531,7 @@ class OpenAIHandlerMixin:
 
         # Rate limiting
         if self.rate_limiter:
-            rate_key = headers.get("authorization", "default")[:20]
+            rate_key = _openai_rate_limit_key(headers)
             allowed, wait_seconds = await self.rate_limiter.check_request(rate_key)
             if not allowed:
                 await self.metrics.record_rate_limited(provider=openai_chat_outcome_provider)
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The shared helper is the right consolidation point, and both OpenAI handler paths now use it consistently. One isolation gap remains: returning `credential[:20]` means distinct credentials that share their first 20 characters still consume the same RPM/TPM bucket. Common provider-issued prefixes make this avoidable coupling, and the PR description currently promises credential-specific buckets.

Please derive a stable opaque identity from the full selected credential (for example, a SHA-256 digest, optionally with a non-secret scheme discriminator), while retaining Authorization precedence and the existing `default` fallback. Add a unit regression with two credentials whose first 20 characters are identical and whose suffixes differ, and keep the handler-level chat/Responses assertions. This also avoids retaining a credential prefix as the bucket dictionary key.

The rest of the focused source change and current CI look sound.

C0sm0cats:
Addressed in 6bac361: the helper now hashes the full selected credential with a header-kind discriminator, so credentials sharing the first 20 characters no longer collide and raw prefixes are not retained. Added the requested identical-prefix/different-suffix regression. Focused suite: 27 passed; ruff and isolated mypy pass.

JerrettDavis (approved):
The follow-up resolves the isolation issue: the bucket identity now hashes the complete selected credential, includes an auth-kind discriminator, preserves Authorization precedence, and retains the unauthenticated default bucket without keeping credential material in the limiter key. The shared-prefix regression covers the prior collision case, and both handler paths remain exercised. Focused validation on this head: 27 tests passed; targeted Ruff and mypy checks passed.

PR #3378, opened 2026-09-01
fix(compression): protect skill bodies and directive words from lossy rewrite

chopratejas:
Three fixes found by measuring the compressor against real skill files rather than reading it.

## 1. Skill bodies were taking the lossy path

`Skill` appeared in **none** of the exclude sets, and only 3 of 40 real `SKILL.md`
bodies tripped the existing code/markup guard. So skill instructions were being
lossily rewritten in production.

That matters more than ordinary text loss, because of *what* was being dropped.
Measured over 40 real skill bodies:

| | retained |
|---|---|
| negations (`not`, `never`, `unless`, …) | **73.5%** |
| modals (`must`, `only`, `always`, …) | **65.5%** |

Two skills kept 0 of 2 and 1 of 6 of their negations. A directive that loses its
negation is not degraded — it is **reversed**. "Do not guess at model names"
becomes "do guess at model names".

`Skill` is now in `DEFAULT_EXCLUDE_TOOLS` and `DEFAULT_BYTE_EXACT_EXCLUDE_TOOLS`,
deliberately **not** `DEFAULT_VERBATIM_EXCLUDE_TOOLS` — dedup stays safe and
valuable here.

## 2. Directive words added to the must-keep guard

Every must-keep class before this protected tokens the model could not
*reconstruct* (identifiers, paths, flags, numbers). None protected tokens whose
loss *inverts* the surrounding sentence. Compressed prompts carry instructions as
often as they carry tool output.

With the directive class added, over the same 40 bodies:

- negation retention **73.5% → 91.9%**
- modal retention **65.5% → 96.8%**
- cost: **0.2 percentage points** of compression (ratio 0.716 → 0.718)

These are short, common words, so the model was already keeping most of them;
pinning the rest is nearly free. The existing must-keep behaviour is unchanged —
a spot check confirms identifiers, hex addresses, paths, flags and numbers are
still retained.

## 3. Two comments that contradicted the code

- `kompress_compressor.py` recorded retention figures from a prototype pattern
  (94.8% / 90.1% / 0.1pp) rather than the one that shipped (91.9% / 96.8% / 0.2pp).
- `server.py` said an extension raising from `install()` "aborts startup". It does
  not: `proxy/extensions.py:169-183` logs it, prints a SKIPPED line to stderr and
  keeps serving, so one bad plugin cannot brick the proxy. The comment now states
  the consequence that actually matters for a licensed plugin — the feature fails
  closed, the process does not.

## Testing

`tests/test_kompress_must_keep.py`, `test_config.py`, `test_kompress_failsafe.py`,
`test_force_kompress_all.py`, `test_compressor_config_exposure.py` — **95 passed**.

https://claude.ai/code/session_01HRaEPjQcCSAwD2xbQMNENu

### DIFF
--- a/headroom/config.py
+++ b/headroom/config.py
@@ -240,10 +240,22 @@ DEFAULT_EXCLUDE_TOOLS: frozenset[str] = frozenset(
         # plugin that supplies only the first half cannot be relied on to be
         # installed.
         "read_file",
+        # Skill bodies (Claude Code's `Skill` tool) are INSTRUCTIONS, not tool
+        # output, and lossy compression on a directive does not degrade it --
+        # it inverts it. "do not guess at model names" losing its "not" is a
+        # correctness failure, not a quality tradeoff. Measured across 40 real
+        # SKILL.md bodies on the lossy path: only 73.5% of negations (not,
+        # never, avoid, unless, without) and 65.5% of modals (must, always,
+        # only, required) survived; two skills kept 0/2 and 1/6 of theirs.
+        # Bodies load on demand and are read once, so the tokens at stake are
+        # small and the downside is unbounded. Named below as well, because
+        # protecting a tool needs both halves.
+        "Skill",
         # Lowercase variants for case-insensitive matching
         "read",
         "glob",
         "grep",
+        "skill",
         "write",
         "edit",
         "web_search",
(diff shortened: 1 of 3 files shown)


JerrettDavis (changes requested):
The safety direction looks right, and I reproduced the stated focused result (`95 passed`; Ruff clean) on `a02873c7f`. I do think this needs one small blocking test addition before merge, because none of those 95 tests exercises either behavior introduced here.

Please add focused regression coverage that:

- asserts `Skill` is protected by both `DEFAULT_EXCLUDE_TOOLS` and `DEFAULT_BYTE_EXACT_EXCLUDE_TOOLS` (including the case-insensitive lookup path), and preferably proves a representative Skill result reaches the provider byte-for-byte; and
- parametrically asserts the new directive/negation vocabulary matches `_KOMPRESS_MUST_KEEP_RE`, with at least one single-item or batch compression test where the fake model drops a directive such as `not`/`must` and the override retains it.

Right now `tests/test_kompress_must_keep.py` only covers the older identifiers/paths/numbers, while the exclusion tests do not mention `Skill`. That means either half of this correctness fix could regress while the exact suite cited in the PR remains green. The production code itself otherwise looks consistent with the existing exclusion and must-keep mechanisms.

JerrettDavis:
Thanks for the work protecting skill instructions. I checked the entire patch against current main: the skill exclusions, directive-word protection, and accompanying changes are already present. Closing this as integrated so we don’t keep a second review open for code that has landed.

PR #3380, opened 2026-09-01
fix(cache): keep cached-prefix replay after background recompression

### LINKED ISSUE #3379: [BUG] Prompt-cache bust loop since 0.36: overlay_cached_prefix declines byte-larger replay after background recompression (Anthropic proxy path)
## Description

Since the non-inflation size bound added in #3052 (first shipped in 0.36.x; measured here on 0.37.0), the Anthropic proxy token path repeatedly busts the provider prompt cache on long Claude Code sessions.

Mechanism:

1. `FREEZE_POLICY_CONFIRMED_CLAMP` freezes only the provider-confirmed prefix, so background compression keeps improving messages that were already forwarded (by design - the #327 posture).
2. When background compression lands a smaller form of an already-forwarded message, the next request's optimized output differs from what the provider cached. `overlay_cached_prefix` exists to repair exactly this drift by replaying the previously-forwarded bytes (#1850) - the session-engine docstring relies on it ("its overlay inputs ... are refreshed on every response, so drift repair is reliable there", `session_engine.py`).
3. The replay is now declined precisely in that case: the previously-forwarded prefix is byte-larger than the freshly recompressed candidate, so the size bound fires ("overlay: replay inflated compact JSON - skipping") and the recompressed bytes go upstream -> the provider cache misses from the first changed byte -> full prefix re-write.
4. Every subsequent background-compression completion repeats this. Steady state on a busy session is a 160-210k-token cache re-write every 1-2 requests.

Economics: the declined "inflation" would have billed at the ~0.1x cache-read rate; the "smaller" recompressed request instead bills the whole divergent tail at the 1.0x input rate plus the cache-write premium.

## Measured impact

macOS, headroom-ai 0.37.0, Claude Code through the proxy, token mode. Same machine and workload; the boundary is the 0.35.0 -> 0.37.0 wheel swap at 12:30 UTC on 2026-09-01:

| window | dollar cache-hit rate | cache writes / forwarded input | billable $/M forwarded input |
|---|---|---|---|
| Aug 31 (0.35.0) | 91% | 2-5% | $1.98 |
| Sep 1 before swap (0.35.0) | 90% | 2-5% | $2.01 |
| Sep 1 after swap (0.37.0) | 52% | 12-17% | $4.40 |

Per-request PERF excerpts from one session (~260-340k `tok_before`):

On 0.35.0, `tok_saved` stays pinned (10450 -> 10449 -> 10448 ...) and every request reads 230-260k cached tokens while writing 0.3-6k.

On 0.37.0, `tok_saved` oscillates as background compression lands, and EVERY change coincides with `cache_read` collapsing to the ~7.7k tools+system head plus a 160-210k `cache_write`:

```
tb= 281217 rd= 178331 wr=   2711 saved= 6701
tb= 284593 rd=   7720 wr= 169820 saved= 7787   <- compression landed, replay declined, prefix re-written
tb= 287857 rd= 177540 wr=   1260 saved= 7787
tb= 289650 rd=   7720 wr= 171844 saved= 7969   <- again
tb= 293965 rd= 179564 wr=   4694 saved= 7969
tb= 298889 rd=   7720 wr= 180100 saved= 8013   <- again
```

## To Reproduce

1. Run a long Claude Code session (>200k context) through `headroom proxy` on 0.36.0+ in token mode with background compression active.
2. Watch PERF lines: after each background-compression completion (`tok_saved` changes), the next request logs `cache_read` of roughly tools+system only, plus a full-prefix `cache_write`. With DEBUG logging, "overlay: replay inflated compact JSON - skipping cached-prefix replay" fires on those turns.
3. Same session on 0.35.0: `tok_saved` stays pinned; reads stay incremental.

## Expected behavior

The confirmed-clamp proxy path keeps the forwarded prefix byte-identical to what the provider cached (#1850's guarantee), even when this turn's recompression produced a smaller form. The size bound still protects callers whose snapshots can be stale; on the proxy path the snapshots refresh from every response, the replay source is always last turn's forwarded bytes (so growth is bounded, not compounding), and the #3026 misalignment failure the bound arrived with remains fully covered by the alignment/append-only guards.

Happy to submit the fix - PR incoming that threads an `enforce_non_inflation` flag through `finalize_turn` and passes `False` from the confirmed-clamp call site only.

gglucass:
## Description

Since the non-inflation size bound added in #3052 (shipped 0.36.x), the Anthropic proxy token path repeatedly busts the provider prompt cache: the bound declines the cached-prefix replay exactly when background compression lands a SMALLER form of an already-forwarded message, so the forwarded bytes change mid-history and the provider cache misses from the first changed byte. Steady state on a busy session is a 160-210k-token cache re-write every 1-2 requests (measured across a 0.35.0 -> 0.37.0 swap on identical workload: dollar cache-hit rate 90% -> 52%, billable input $/M up 2.2x; full data in #3379).

Reworked per review (thanks @chopratejas - the direction is yours): rather than disabling the size bound wholesale, `overlay_cached_prefix` now takes `confirmed_frozen_count` and splits at that floor.

- Inside the provider-confirmed prefix: replay is unconditional. The replay source is exactly what the provider hashed, so changing those bytes can only bust the cache; an "inflating" replay there bills at the ~0.1x cache-read rate and is still the cheaper request.
- Beyond the floor: the size bound arbitrates each turn exactly as on main - a shrinking replay repairs freeze drift (#1850), an inflating one is declined so a fresh compression improvement reaches the wire.
- When the provider count collapses (cold cache, TTL lapse) the floor collapses with it and every accumulated improvement lands at once: the natural re-baselining that bounds long-session growth (the #3026 direction of travel) by the TTL-lapse cadence instead of pinning forever.

The #3052 alignment/append-only guards stay enforced unconditionally.

The OpenAI chat path now carries the same floor (reviewer question: is the root cause live there too?). It is: the tracker is provider-generic and is fed OpenAI's own `prompt_tokens_details.cached_tokens`, the forwarded-bytes feedback loop is identical, both handlers compress through the same ContentRouter, and OpenAI's automatic prompt caching busts on the first changed byte exactly like Anthropic's. Exposure is narrower - the token path's `FREEZE_POLICY_REPLAYABLE` already stops most recompression of already-forwarded messages - but not zero: an entry evicted from the local byte-replay cache drops out of that freeze, and cache mode freezes from the tracker only. Remaining callers with no provider-confirmed count in scope (sidecar, `openai.py:9940`) pass nothing and keep today's fully size-bounded behavior; `bedrock.py` / `gemini.py` never call `finalize_turn` at all, tracked in #3391.

Closes #3379

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `overlay_cached_prefix` gains keyword-only `confirmed_frozen_count: int | None = None`: an unconditional-replay floor for the provider-confirmed prefix; the size bound still governs everything beyond it, and alignment guards are untouched.
- `finalize_turn` forwards the floor (default `None`, so all existing callers are unchanged).
- The OpenAI chat call site passes the same floor, snapshotted before `prepare_turn`'s replayable clamp (that count runs over the LOCAL byte-replay cache and can exceed what the provider confirmed; a cold prefix zeroes it, so cold re-baselining matches Anthropic's).
- The Anthropic confirmed-clamp call site stashes the pre-clamp tracker count (the post-`prepare_turn` min under-floors when the local byte-replay cache evicted entries) and passes it; the stash defaults to 0 on paths that never compute a tracker count (backpressure).
- Tests: floor replays a recompressed confirmed prefix; improvement beyond the floor lands while the confirmed region replays; originals-drift beyond the floor is still repaired when the replay shrinks; alignment guards unaffected by the floor; block-append within the floor; `finalize_turn` pass-through; the reviewer-requested multi-turn harness test - a background improvement to an already-forwarded message never busts the warm cache across 8 turns and lands on the first cold turn; the reviewer-requested growth bound - forwarded size grows by only its own delta each warm turn and the first cold turn re-baselines to exactly the fresh pipeline's output; and an end-to-end OpenAI chat test asserting the provider-confirmed prefix is replayed even when the replay inflates (fails without the new argument).

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_cache_prefix_overlay.py tests/test_session_engine.py tests/test_cross_turn_cache_safety.py tests/test_cache_control_move_bust.py -q
============================== 58 passed in 0.44s ==============================
$ uvx ruff check headroom/cache/prefix_tracker.py headroom/proxy/session_engine.py headroom/proxy/handlers/anthropic.py tests/test_cache_prefix_overlay.py tests/test_session_engine.py tests/test_cross_turn_cache_safety.py
All checks passed!
$ uv run --frozen --extra dev mypy headroom/cache/prefix_tracker.py headroom/proxy/session_engine.py
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), Python 3.12; regression measured on the released headroom-ai 0.37.0 under real Claude Code traffic through `headroom proxy` (token mode); this branch validated with the repo test suite.
- Exact command / steps: diffed per-request PERF logs across a 0.35.0 -> 0.37.0 swap on the same machine and workload (sequences and aggregates in #3379); then ran the pytest/ruff/mypy commands above on this branch. A deployment carrying the unconditional-replay behavior on the confirmed prefix restored the healthy pattern live: the broken wheel capped cached reads at 157k across 32 large requests (prefix perpetually re-written) while the patched deployment accumulated to 305k cached reads within 15 minutes.
- Observed result: 58 passed including the new multi-turn harness test; on stock 0.37.0 every background-compression completion produced a cache_read collapse to the ~7.7k tools+system head plus a 160-210k cache_write.
- Not tested: a live OpenAI `cached_tokens` comparison for the OpenAI half (the Anthropic measurement is what #3379 carries; the OpenAI change is argued from shared mechanism plus an end-to-end handler test, not from its own dollar data). A live provider run of this exact branch (the wheel comparison plus the live patched-deployment measurement isolate the bound as the trigger; the unit and multi-turn regressions reproduce decline and fix deterministically). `mypy headroom` was run only on the two touched modules, not the whole package.

## Runtime Rollout Safety

- Stable/default behavior changed: On the Anthropic proxy confirmed-clamp path and the OpenAI chat path, both of which now replay the provider-confirmed prefix unconditionally (the pre-0.36 posture) while keeping the size bound beyond the confirmed floor. The overlay's default and every other caller keep the fully size-bounded behavior.
- Kill switch / disable path: remove the `confirmed_frozen_count=` argument at the two call sites (anthropic.py, openai.py chat); the parameter defaults to `None`.
- Qualification impact: None.
- Rollback path: Revert the commits; the parameter is additive and default-preserving.

## Review Readiness

## Checklist

Not applicable.

## Additional Notes

Lineage for review: #1850 made freeze forward the cached prefix byte-identical, #3052 added alignment guards plus the size bound after #3026's misalignment blowup, #3271 unified both modes on the session engine. The first cut of this PR disabled the bound on the confirmed-clamp path wholesale; the reviewed rework scopes unconditional replay to the provider-confirmed prefix, which additionally gives cold-lapse re-baselining (improvements land whenever the floor collapses), addressing the growth analysis raised in review. Economics discussion in the review thread.

### DIFF
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -448,6 +448,8 @@ def overlay_cached_prefix(
     current_original_messages: list[dict[str, Any]],
     previous_original_messages: list[dict[str, Any]] | None,
     previous_forwarded_messages: list[dict[str, Any]] | None,
+    *,
+    enforce_non_inflation: bool = True,
 ) -> list[dict[str, Any]]:
     """Replay a positional, non-inflating cached prefix when it is safe.
 
@@ -471,6 +473,17 @@ def overlay_cached_prefix(
     compact UTF-8 JSON for the replayed result must not exceed the optimized
     candidate. These bounds prefer a cache miss to corrupting or inflating a
     client's live history.
+
+    ``enforce_non_inflation`` gates only the compact-JSON size bound, never the
+    alignment/append-only guards. The proxy's confirmed-clamp path passes
+    False: its snapshots are refreshed from every provider response, so the
+    replayed prefix is exactly what the provider has cached, and re-forwarding
+    a freshly recompressed (smaller) history instead busts the cache from the
+    first changed byte - a full prefix re-write at the full input rate to save
+    bytes that would have billed at the ~0.1x cache-read rate. Growth stays
+    bounded with the bound off: the replay source is always last turn's
+    forwarded bytes, so the prefix is byte-stable turn over turn rather than
+    compounding.
     """
     prev_orig = previous_original_messages
     prev_fwd = previous_forwarded_messages
@@ -548,15 +561,16 @@ def overlay_cached_prefix(
                     + [merged]
                     + list(optimized_messages[message_index + 1 :])
                 )
-                replayed_bytes = _compact_json_bytes(replayed)
-                optimized_bytes = _compact_json_bytes(optimized_messages)
-                if (
-                    replayed_bytes is None
-                    or optimized_bytes is None
-                    or len(replayed_bytes) > len(optimized_bytes)
-                ):
-                    logger.debug("overlay: block replay inflated compact JSON — skipping")
-                    return optimized_messages
+                if enforce_non_inflation:
+                    replayed_bytes = _compact_json_bytes(replayed)
+                    optimized_bytes = _compact_json_bytes(optimized_messages)
+                    if (
+                        replayed_bytes is None
+                        or optimized_bytes is None
+                        or len(replayed_bytes) > len(optimized_bytes)
+                    ):
+                        logger.debug("overlay: block replay inflated compact JSON — skipping")
+                        return optimized_messages
                 return replayed
     # Append-only guard on CONTENT ONLY, message-by-message. Replay the
     # previously-forwarded (cached, compressed) bytes for the longest LEADING
@@ -604,15 +618,16 @@ def overlay_cached_prefix(
     # Replay the cached (compressed) prefix byte-identical up to the first
     # divergence; keep this turn's freshly-produced output for the rest.
     replayed = list(prev_fwd[:k]) + list(optimized_messages[k:])
-    replayed_bytes = _compact_json_bytes(replayed)
-    optimized_bytes = _compact_json_bytes(optimized_messages)
-    if (
-        replayed_bytes is None
-        or optimized_bytes is None
-        or len(replayed_bytes) > len(optimized_bytes)
-    ):
-        logger.debug("overlay: replay inflated compact JSON — skipping cached-prefix replay")
-        return optimized_messages
+    if enforce_non_inflation:
+        replayed_bytes = _compact_json_bytes(replayed)
+        optimized_bytes = _compact_json_bytes(optimized_messages)
+        if (
+            replayed_bytes is None
+            or optimized_bytes is None
+            or len(replayed_bytes) > len(optimized_bytes)
+        ):
+            logger.debug("overlay: replay inflated compact JSON — skipping cached-prefix replay")
+            return optimized_messages
     return replayed
 
 
--- a/headroom/cache/prefix_tracker.py
+++ b/headroom/cache/prefix_tracker.py
@@ -449,7 +449,7 @@ def overlay_cached_prefix(
     previous_original_messages: list[dict[str, Any]] | None,
     previous_forwarded_messages: list[dict[str, Any]] | None,
     *,
-    enforce_non_inflation: bool = True,
+    confirmed_frozen_count: int | None = None,
 ) -> list[dict[str, Any]]:
     """Replay a positional, non-inflating cached prefix when it is safe.
 
@@ -474,16 +474,19 @@ def overlay_cached_prefix(
     candidate. These bounds prefer a cache miss to corrupting or inflating a
     client's live history.
 
-    ``enforce_non_inflation`` gates only the compact-JSON size bound, never the
-    alignment/append-only guards. The proxy's confirmed-clamp path passes
-    False: its snapshots are refreshed from every provider response, so the
-    replayed prefix is exactly what the provider has cached, and re-forwarding
-    a freshly recompressed (smaller) history instead busts the cache from the
-    first changed byte - a full prefix re-write at the full input rate to save
-    bytes that would have billed at the ~0.1x cache-read rate. Growth stays
-    bounded with the bound off: the replay source is always last turn's
-    forwarded bytes, so the prefix is byte-stable turn over turn rather than
-    compounding.
+    ``confirmed_frozen_count`` bounds UNCONDITIONAL replay. Leading positions
+    the provider has already confirmed cached (a message count derived from
+    ``cache_read_input_tokens``) are always replayed byte-identical: the
+    replay source there is exactly what the provider hashed, so changing
+    those bytes can only bust the cache. Beyond the floor the size bound
+    still arbitrates each turn: a shrinking replay (the pipeline emitted
+    original bytes for a frozen message) is repaired, while an inflating one
+    (fresh compression improved on the forwarded form) is declined so the
+    improvement reaches the wire. When the provider count collapses (cold
+    cache, TTL lapse) the floor collapses with it and every accumulated
+    improvement lands at once - the natural re-baselining that keeps
+    long-session growth bounded (#3026). Callers with no provider-confirmed
+    count pass None and keep the fully size-bounded behavior.
     """
     prev_orig = previous_original_messages
     prev_fwd = previous_forwarded_messages
@@ -561,7 +564,7 @@ def overlay_cached_prefix(
                     + [merged]
                     + list(optimized_messages[message_index + 1 :])
                 )
-                if enforce_non_inflation:
+                if message_index >= max(confirmed_frozen_count or 0, 0):
                     replayed_bytes = _compact_json_bytes(replayed)
                     optimized_bytes = _compact_json_bytes(optimized_messages)
                     if (
@@ -618,16 +621,31 @@ def overlay_cached_prefix(
     # Replay the cached (compressed) prefix byte-identical up to the first
     # divergence; keep this turn's freshly-produced output for the rest.
     replayed = list(prev_fwd[:k]) + list(optimized_messages[k:])
-    if enforce_non_inflation:
-        replayed_bytes = _compact_json_bytes(replayed)
-        optimized_bytes = _compact_json_bytes(optimized_messages)
-        if (
-            replayed_bytes is None
-            or optimized_bytes is None
-            or len(replayed_bytes) > len(optimized_bytes)
-        ):
+    replayed_bytes = _compact_json_bytes(replayed)
+    optimized_bytes = _compact_json_bytes(optimized_messages)
+    if (
+        replayed_bytes is None
+        or optimized_bytes is None
+        or len(replayed_bytes) > len(optimized_bytes)
+    ):
+        # Something in the replay is byte-larger than this turn's fresh form:
+        # fresh compression improved on already-forwarded bytes. Landing the
+        # improvement is only safe OUTSIDE the provider-confirmed prefix -
+        # inside it, the improvement would change bytes the provider has
+        # already cached and bust the whole suffix. Split at the confirmed
+        # floor: replay the confirmed region unconditionally, forward
+        # everything beyond it fresh so the improvement reaches the wire.
+        floor = min(k, max(confirmed_frozen_count or 0, 0))
+        if floor <= 0:
             logger.debug("overlay: replay inflated compact JSON — skipping cached-prefix replay")
             return optimized_messages
+        logger.debug(
+            "overlay: replay inflated beyond the confirmed floor — replaying %d/%d "
+            "confirmed messages, forwarding the rest fresh",
+            floor,
+            k,
+        )
+        return list(prev_fwd[:floor]) + list(optimized_messages[floor:])
     return replayed
 
 
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -1691,6 +1691,14 @@ class AnthropicHandlerMixin:
                         compressor.close()
 
             _compression_failed = False
+            # Provider-confirmed frozen count, stashed BEFORE prepare_turn
+            # clamps it against the local byte-replay cache: the overlay's
+            # unconditional-replay floor must cover everything the provider
+            # has cached, not just what local state can byte-replay (the
+            # replay source is the positional prev_fwd snapshot, which does
+            # not depend on the CompressionCache). Stays 0 on paths that
+            # never compute a tracker count (backpressure, cache mode).
+            _confirmed_frozen = 0
             original_messages = messages  # Preserve for 400-retry fallback
             _decision = CompressionDecision.decide(
                 headers=request.headers,
@@ -1766,6 +1774,7 @@ class AnthropicHandlerMixin:
                             prepare_turn,
                         )
 
+                        _confirmed_frozen = max(int(frozen_message_count or 0), 0)
                         _prep = prepare_turn(
                             comp_cache,
                             messages,
@@ -2150,16 +2159,17 @@ class AnthropicHandlerMixin:
                         previous_original_messages,
                         previous_forwarded_messages,
                         count_tokens=tokenizer.count_messages,
-                        # This path's snapshots refresh from every provider
-                        # response, so prev_fwd is exactly what the provider
-                        # cached. Declining a byte-larger replay here would
-                        # re-forward freshly recompressed history at the full
-                        # input rate and bust the prompt cache from the first
-                        # changed byte every time background compression
-                        # improves on an already-forwarded message; the size
-                        # bound stays on for callers without refreshed
-                        # snapshots.
-                        enforce_non_inflation=False,
+                        # The provider-confirmed prefix is replayed
+                        # unconditionally: those bytes are exactly what the
+                        # provider cached, so declining a byte-larger replay
+                        # there re-forwards freshly recompressed history at
+                        # the full input rate and busts the cache every time
+                        # background compression improves on an
+                        # already-forwarded message. Beyond the confirmed
+                        # floor the size bound still lets improvements
+                        # through, and a collapsed floor (cold cache) lets
+                        # every accumulated improvement land at once.
+                        confirmed_frozen_count=_confirmed_frozen,
                     )
                     _overlay_replayed = _final.replayed
                     if _overlay_replayed:
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -2150,6 +2150,16 @@ class AnthropicHandlerMixin:
                         previous_original_messages,
                         previous_forwarded_messages,
                         count_tokens=tokenizer.count_messages,
+                        # This path's snapshots refresh from every provider
+                        # response, so prev_fwd is exactly what the provider
+                        # cached. Declining a byte-larger replay here would
+                        # re-forward freshly recompressed history at the full
+                        # input rate and bust the prompt cache from the first
+                        # changed byte every time background compression
+                        # improves on an already-forwarded message; the size
+                        # bound stays on for callers without refreshed
+                        # snapshots.
+                        enforce_non_inflation=False,
                     )
                     _overlay_replayed = _final.replayed
                     if _overlay_replayed:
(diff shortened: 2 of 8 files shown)


chopratejas:
Reviewed this closely. The diagnosis is right and the cost data in #3379 is a real regression worth fixing — but I think this particular fix has a structural problem, and I'd like to lay out the reasoning so it can be checked.

## The overlay's replay region is much wider than the frozen region

`prepare_turn` under `FREEZE_POLICY_CONFIRMED_CLAMP` computes `frozen = min(tracker_frozen, cache_count)` — deliberately narrow, because, as the module docstring puts it, the confirmed clamp "deliberately KEEPS not-yet-cached content compressible."

But `overlay_cached_prefix` never receives `frozen_message_count`. Its replay length `k` is derived purely from canonical equality of *client originals*:

```python
while k < limit and _canonicalize_for_prefix_compare(
    current_original_messages[k]
) == _canonicalize_for_prefix_compare(prev_orig[k]):
    k += 1
```

Clients don't rewrite history, so `k` covers **every previously-seen message** — typically `k == n`, far past `frozen`. Combined with `next_forwarded_messages = copy.deepcopy(optimized_messages)` (anthropic.py:3474) feeding the replayed bytes back as next turn's `prev_fwd`, each position becomes pinned to its first-forwarded form permanently.

The size bound is currently the only thing reconciling those two regions. With it enforced, an improved (smaller) compression declines the replay, the fresh output goes out, and the prefix re-baselines — one bust, but the improvement lands. With `enforce_non_inflation=False`, the replay region swallows the compressible region the clamp policy deliberately created, and improvements to already-forwarded messages can never reach the wire.

## Measured effect

Driving the real `overlay_cached_prefix` over 40 turns, with messages compressed on first sight and background compression later finding a better form (this PR's stated trigger), feeding forwarded bytes back each turn:

```
turn        RAW  ideal pipe     fwd ON    fwd OFF  OFF/ON
   1      4,079       1,283      1,283      1,283   1.00x
  10     40,783       5,127      5,127     12,827   2.50x
  20     81,583       6,967      6,967     25,667   3.68x
  40    163,183      10,647     10,647     51,347   4.82x
```

Bound on, forwarded output tracks the ideal pipeline exactly. Bound off, it settles ~4.8x above ideal at turn 40 and the ratio widens every turn. To be clear about what this is and isn't: compression still works — first-sight compression lands, and this is far better than raw. What is lost is every *later* improvement, and superseded-read elimination is inherently a later discovery. This is a simulation harness, not a live provider run, so treat the shape as the claim and the constants as illustrative.

That widening is what makes me uneasy about "Growth stays bounded... byte-stable turn over turn rather than compounding." The prefix doesn't compound — agreed. But frozen-prefix-plus-growing-tail isn't bounded, and this is the direction #3052 was fixing (#3026: 187,986 → 753,077 tokens, `400 prompt_too_long`). Three things that might otherwise catch it don't: `_cold_recompact_active` is gated behind `HEADROOM_COLD_RECOMPACT`, off by default (anthropic.py:1585); the inflation-revert guard is explicitly skipped when `_overlay_replayed` (anthropic.py:2204), which is exactly this path; and there's no `prompt_too_long` handling in `headroom/`.

## The docstring may be prescribing a different remedy

The PR cites the confirmed-clamp sentence, but the paragraph just above it reads:

> Recompressing an already-returned message — even into a *smaller* form — is a bust: ... `overlay_cached_prefix`'s non-inflation guard **cannot repair a shrunken form** ... **Freezing the entire locally-replayable prefix eliminates that recompression outright.**

So the codebase already knows about this exact failure and its prescribed remedy is to freeze so the recompression never happens — not to relax the guard that catches it. Worth reconciling with, since the PR reads that section as support.

## Suggested direction

Scope the relaxed bound to the provider-confirmed prefix rather than disabling it globally: replay `prev_fwd` unconditionally up to `min(k, tracker_frozen)`, and let freshly-improved compression through beyond it. That captures the cache-hit benefit where the provider actually has bytes cached, while leaving the compressible tail compressible. Changing content past the confirmed read point doesn't invalidate the cached region — though it does forfeit a pending write, so it isn't literally free and would be worth measuring. A savings-vs-rewrite-cost threshold instead of a strict byte comparison would work too.

## Test coverage

I was wrong in an earlier draft to say no multi-turn coverage exists — `tests/test_cross_turn_cache_safety.py` has a real harness that feeds `forwarded` back as `prev_forwarded`. But its `_apply_freeze` models *freeze-emits-originals*, where replay makes things smaller and the bound never fires. It doesn't cover recompression-shrinks, where replay inflates. That harness looks like the natural home for a test asserting forwarded size stays within some factor of pipeline output across ~20 turns — which I think would fail on this branch.

Happy to be shown wrong on the growth analysis, particularly if the confirmed clamp constrains `k` in production in a way I've missed.

gglucass:
Thanks for the close read @chopratejas  - the k-past-frozen mechanics, the `next_forwarded_messages` feed-back, and all three missing-safety-net observations check out against the code, and you were right that the docstring paragraph I cited cuts the other way. Reworked in ad66154d along the lines you suggested; body updated. Two places where I'd push back on the framing, because they change what the sim's columns mean, and then what the rework does with your direction.

**The ideal-pipe column is not reachable on a warm cache.** Landing an improvement to an already-forwarded message necessarily changes bytes the provider hashed, so each landing is a full-suffix re-write. "One bust and the improvement lands" is the right description of a single event - but background compression lands continuously on real traffic. Measured on 0.37.0 (data in #3379): a re-baseline every 1-2 requests, dollar cache-hit 90% -> 52%, billable input $/M 2.2x. The fwd-ON column pays that toll silently because the sim tracks tokens, not the 0.1x/1.0x split. At your own turn-40 constant, a pinned prefix at 4.8x-of-ideal read from cache costs ~0.48x of what an ideal-sized *fresh* prefix would - and the empirical version of this comparison already ran: 0.35.0 (exactly the bound-off behavior, for months) at $1.98/M vs 0.37.0 at $4.40/M on the same workload.

**The prescribed-remedy paragraph has the same pinning property.** `FREEZE_POLICY_REPLAYABLE` freezes everything locally replayable, so recompression is "eliminated outright" - first-returned forms are forwarded forever there too. The fwd-OFF column *is* the REPLAYABLE posture's output; the docstring prescribes the same tradeoff at prepare-time rather than repair-time. So I don't think the codebase's own remedy rescues the ideal-pipe column; nothing can, while the cache is warm.

That said, your `min(k, tracker_frozen)` direction is better than my boolean for a reason worth making explicit: **when the provider count collapses (cold cache, TTL lapse), the floor collapses with it, and every accumulated improvement lands at once.** That converts your growth concern from "unbounded pinning" into "bounded by the TTL-lapse cadence" structurally, with no env dependence - unlike `HEADROOM_COLD_RECOMPACT`, which does the same job but ships off by default.

What ad66154d does:

- `overlay_cached_prefix(..., confirmed_frozen_count=None)`: inside the floor, replay is unconditional; beyond it the size bound arbitrates exactly as on main (shrinking replay repairs #1850 freeze drift, inflating replay is declined so the improvement ships). When the whole-list bound declines, it now splits at the floor instead of abandoning the replay entirely.
- The anthropic call site stashes the pre-clamp tracker count (the post-`prepare_turn` min under-floors when the local byte-replay cache evicted entries - and that eviction gap is the drift-repair case the confirmed clamp most depends on), defaulting to 0 on paths that never compute one (backpressure), which degrades to today's behavior there.
- Sidecar/OpenAI callers pass nothing and are byte-for-byte unchanged.
- Your multi-turn harness test is in `test_cross_turn_cache_safety.py`: a background improvement to an already-forwarded message never busts the warm cache across 8 turns driving the real tracker, and lands on the first cold turn. You predicted a forwarded-size-factor assertion would fail on the v1 branch - agreed, it would have; on the rework the bound it fails against is the TTL-lapse cadence, which is the bound I'd defend as correct rather than a ceiling relative to the ideal pipe on warm turns.

The one cost your direction accepts (and I think rightly): an improvement arriving beyond the floor forfeits that region's pending cache write for one turn. That window is 1-2 turns of content on Claude Code traffic, so it is bounded and small; flagging it since you asked for it to be measurable - happy to add a counter if you want it observable before merging.

chopratejas:
Reviewed the mechanism end to end. The direction is right and the split-at-the-floor shape is a better fix than disabling the size bound: the alignment and append-only guards stay unconditional, the final splice is bounded by `min(k, floor)` where `k` comes from the verified-safe guard loop, and `_confirmed_frozen` defaults to 0 on the backpressure paths. I couldn't fault the correctness. Given the measured 90% → 52% cache-hit collapse, I'd like this in.

Two things before/after merge.

**1. This is Anthropic-only, and the same root cause is plausibly live on OpenAI.**

`finalize_turn` has three production call sites; only one gets the new floor:

| file | line | `confirmed_frozen_count` |
|---|---|---|
| `handlers/anthropic.py` | 2147 | yes |
| `handlers/openai.py` | 3915 | no |
| `handlers/openai.py` | 9940 | no |

`frozen_message_count` is derived only in `anthropic.py:1550`, and `streaming.py:2033` already carries a comment that it "doesn't propagate frozen_message_count either."

OpenAI's automatic prompt caching is prefix-based and busts on the first changed byte exactly like Anthropic's, with the same order of magnitude in play (gpt-5 $1.25/M vs $0.125/M cached). If the OpenAI path also recompresses already-forwarded history, #3379 has an untracked twin there. Could you confirm whether it does? If yes, worth an issue linked from here so the asymmetry is deliberate and visible rather than incidental. (`bedrock.py` and `gemini.py` don't call `finalize_turn` at all — separate question, but Bedrock-served Claude has the same caching.)

**2. Cache-optimal is not context-window-optimal.**

Pinning the confirmed prefix to its older, larger form means compression improvements never land inside it while the cache stays warm. The description argues TTL lapse re-baselines, but on a busy session every turn refreshes the cache, so the TTL may not lapse for a long time. The dollars are right (replay bills at ~0.1x), but the prefix stays fat and the session reaches the context window sooner, compacting more often.

That interacts directly with #2942, which exists to fix compact-every-turn thrash near the window. Could the multi-turn harness test be extended to assert something about prefix token growth across the 8 turns, not just that the cache stays warm? Otherwise we may trade a cache regression for a compaction regression and only find out in the field.

gglucass:
Both confirmed, both addressed.

**1. OpenAI: yes, it does. Filed #3391.**

I traced every ingredient rather than reasoning from symmetry, and the mechanism is fully present on the chat path:

- `openai_prefix_tracker.get_frozen_message_count()` (openai.py:3712) is the same provider-confirmed quantity as the Anthropic one - `PrefixCacheTracker` is provider-generic, and `update_from_response` is fed OpenAI's real `prompt_tokens_details.cached_tokens` at openai.py:4786 and openai.py:5164.
- The forwarded-bytes feedback loop is identical: `get_last_forwarded_messages()` is the replay source at the openai.py:3915 call site.
- The trigger is provider-agnostic too. The semantic compressor warms in a background thread (`ensure_background_load` in content_router.py), so a message compresses better on a later turn than on the turn it was first forwarded, and both handlers route through the same router (`openai_pipeline.apply` / `anthropic_pipeline.apply`).

So the count is already computed on that path and just never reaches the overlay. The fix is the same argument, with the same pre-strict-override snapshot caveat as anthropic.py:1553. I kept it out of this PR deliberately: #3379's dollar data is Anthropic-only, and an OpenAI change should ship with its own `cached_tokens` measurement rather than inheriting my constants. The issue also records that openai.py:9940 has no count in scope and that bedrock.py / gemini.py never call `finalize_turn` at all, which is the larger gap of the three.

**2. Growth assertion added, and your framing is right.**

`test_prefix_growth_bounded_and_rebaselines_on_the_cold_turn` in `test_cross_turn_cache_safety.py`, driving the worst case for the floor: background compression finds a better form for *every* already-forwarded message, improving each turn. Measured forwarded/fresh-pipeline ratio over turns 2-8:

```
1.35, 1.84, 2.33, 2.83, 3.35, 3.88, 4.35
```

That reproduces your shape and I am not going to argue with it - while the cache is warm the pinned prefix does widen against an ideal pipeline, and at your constants. What the test asserts is the two bounds that keep it from being #3026:

- each warm turn's forwarded growth is exactly its own delta message, never the pinned prefix re-inflating (rules out compounding);
- the first cold turn re-baselines to *precisely* the fresh pipeline's output, so the accumulated gap is bounded by the re-baseline cadence rather than by session length. This assertion is also what fails on the v1 of this PR, where replay was unconditional over the whole list.

On "a busy session refreshes the cache so the TTL may not lapse": the re-baselining channel is wider than TTL lapse alone, because the floor is the provider-*confirmed* count and that lags the head. Data point from running a floor-split replay against an older pin on real Claude Code traffic: it read about 22% fewer cached tokens than an unconditional full-prefix replay on the same workload. That is a blunt instrument, measured against a different code path, so treat it as an order of magnitude rather than a constant - but it does say a meaningful slice of the prefix sits beyond the confirmed floor on live traffic, and that slice takes improvements every turn under the size bound, not at TTL lapse. It also means this PR does not fully restore the pre-#3052 cache behaviour, which I would rather say out loud than have discovered later.

Happy to add the `overlay: replay inflated beyond the confirmed floor` path counter to `/stats` if you want the warm-pin width observable in the field before merge, but the growth question itself now has a test rather than an argument.

gglucass:
Revising my answer on the OpenAI half: rather than leave it to a follow-up, the fix is now in this PR (openai.py chat path), and one claim in my earlier comment and in #3391 needs correcting.

**Correction: the OpenAI token path is less exposed than I said.** It runs `FREEZE_POLICY_REPLAYABLE`, not the confirmed clamp - it freezes everything the local byte-replay cache can serve, which is the posture the `session_engine` docstring prescribes ("freezing the entire locally-replayable prefix eliminates that recompression outright"). So on that path a background improvement to an already-forwarded message mostly never reaches the pipeline output at all. My earlier comment presented the two paths as symmetric; they are not, and the table in #3391 reads as more alarming than the code warrants. I have corrected the issue.

**What is still exposed, and why the floor is worth carrying there:**

- The REPLAYABLE freeze is bounded by the LOCAL compression cache. An entry evicted by LRU turnover drops out of the freeze, so that message is recompressed while the provider still has the old bytes cached - and the size bound then declines the (inflating) replay, which is exactly the #3379 sequence. That eviction gap is the same one this PR's Anthropic stash exists for.
- Cache mode freezes from the tracker only ("the freeze comes from the tracker, never from the cache count"), so it has no replayable-prefix protection at all.

**The change:** the floor is snapshotted before `prepare_turn`'s replayable clamp, because that count runs over the local cache and can exceed what the provider actually confirmed - passing it would pin content OpenAI never cached. A cold prefix zeroes it above, so cold re-baselining behaves as on Anthropic. `openai.py:9940` still has no count in scope and is unchanged; `bedrock.py` / `gemini.py` still never call `finalize_turn`, which #3391 now tracks as its remaining scope.

**Test:** an end-to-end chat-completions test where last turn's forwarded bytes are LARGER than what the pipeline produces now, so the size bound alone declines the replay. It asserts the confirmed prefix goes out byte-identical anyway, and it fails without the new argument (verified by removing it).

Full run on the branch: 70 passed across the cache/overlay/session-engine/OpenAI suites, plus 98 in the adjacent cache suites, ruff and mypy clean on the touched modules.

Happy to split the OpenAI half back out if you would rather land the measured Anthropic fix on its own - it is one argument at one call site, so it separates cleanly either way.

PR #3381, opened 2026-09-01
fix(proxy): fall back when forced Kompress is cold

### LINKED ISSUE #3363: [BUG] No compression transforms ever apply on OpenAI chat.completions
Content compression never applies — only trivial tool_schema_compaction runs, savings ~0% across every content type and profile tested
Summary

After a clean install with the ML extras, correct proxy setup, and systematically testing every relevant configuration knob, no content-level compressor (Kompress / SmartCrusher / CodeCompressor) ever fires on real requests. The only transform that ever appears in headroom inspect is tool_schema_compaction, which trims tool-definition metadata, not message content. Reported savings stay at 0–1% regardless of payload size, content type (JSON logs, plain prose, literal repeated text), savings profile, or endpoint.

Environment
OS: Windows
Headroom version: 0.37.0 (proxy + CLI match, confirmed via headroom doctor)
Install method: pip install "headroom-ai[all,ml]" (first attempt with --force-reinstall --no-deps installed without the ML stack; a second pip install --upgrade --force-reinstall "headroom-ai[all,ml]" — no --no-deps — correctly pulled in transformers, sentence-transformers, onnxruntime, torch, tokenizers, safetensors, huggingface-hub, confirmed via pip show headroom-ai)
Deployment: headroom proxy --port 8787 run as a persistent Windows service via NSSM (auto-start, AppStdout/AppStderr logging enabled, later --log-messages added to AppParameters)
Client: a third-party agent harness ("DSH", OpenAI-compatible), configured with baseURL: http://localhost:8787/v1 for its openai provider, both with and without an explicit api: openai-completions override (see below)
Steps to reproduce
Install and run Headroom as above; confirm headroom doctor shows proxy: pass, version: pass.
Point an OpenAI-compatible client at http://127.0.0.1:8787/v1 with a valid API key.
Send a request containing a large, genuinely redundant payload — e.g. 220 near-identical log lines, only the timestamp differing per line (~48KB / ~13k raw tokens). Ask the model a simple question about it.
Check headroom inspect for the request.
Expected behavior

For content this repetitive, SmartCrusher/Kompress should show a substantial reduction (the project's own README claims 60–95% on JSON/log-type workloads, and cites a 92%-reduction "SRE incident debugging" example of comparable size).

Actual behavior

headroom inspect shows only:

tokens 20106 → 20092 (saved 14, 0.0696%)   transforms: openai:responses:tool_schema_compaction
(no per-message content changes — savings came from structural / transform-level edits)

No router:text, router:smart_crusher, or any content-compression transform ever appears.

Note on prior art: this matches a secondary symptom described in the now-closed issue #1701 (closed via #1738). That issue's primary/titled bug was a Windows-specific proxy hang ("becomes unresponsive after processing one request") on v0.28.0, which #1738 evidently fixed. But buried in that same report, the one request that did complete before the hang showed an identical pattern to ours: "No actual compression transforms applied — all 76 transforms are either router:protected:, router:excluded:tool, or read_lifecycle:stale:. No router:text, router:smart_crusher, or any actual compression transform appears." Since the hang was the headline bug and the fix (#1738) was scoped to that, it's plausible the "no real compression" symptom was never actually addressed — it just became harder to notice once the proxy stopped crashing before anyone could check. We're filing this as a new, separate issue (rather than commenting on the closed #1701) since the two symptoms are independent: our proxy stays fully responsive across dozens of requests over several hours, with no hang at all — only the missing-compression part reproduces.

Troubleshooting already attempted (all confirmed via pip show, headroom doctor, the proxy startup banner, or headroom inspect — not just "set and hoped")
#	Hypothesis	Action taken	Result
1	ML stack not installed, so Kompress silently falls back to a non-ML path	Reinstalled with pip install --upgrade --force-reinstall "headroom-ai[all,ml]"; confirmed transformers, onnxruntime, torch, sentence-transformers, etc. present in pip show headroom-ai	No change
2	Built-in compressors are opt-in (Enabled compressors field in the dashboard's Advanced settings is empty by default)	Investigated; deprioritized once #3–#5 below produced clearer evidence	Inconclusive — UI input for this field did not render properly in-browser, could not be set directly
3	Default savings profile (coding) protects the two most recent conversation turns verbatim to preserve provider prefix-cache	Set HEADROOM_SAVINGS_PROFILE=agent-90 (per docs: "Forces ML-based (Kompress) compression with a 10% keep-ratio... Compresses both user and system messages") as a service env var via nssm set Headroom AppEnvironmentExtra; confirmed via proxy startup log that Mode: token (not cache) took effect	No change
4	User/system-role messages are protected from compression by default (confirmed default behavior per LiteLLM's Headroom integration docs)	Set HEADROOM_COMPRESS_USER_MESSAGES=1 as a service env var	No change
5	Client was hitting /v1/responses instead of /v1/chat/completions, and Kompress/SmartCrusher may not be wired to the Responses-endpoint path	Added api: openai-completions to the client's provider config to force /v1/chat/completions; confirmed via headroom inspect that the transform prefix changed from openai:responses: to openai:chat:	Endpoint changed successfully; compression result unchanged (saved 14, 0.0656%, same tool_schema_compaction-only transform)

Content types tested across the above configurations, all with 0–1% savings regardless: a short JSON array (8 records), a 600-record JSON log (blocked by an unrelated org TPM rate limit before compression could be observed), a 170-record JSON log (25.6k tokens), a coherent ~1,850-word prose passage (~2.6k tokens, incoherent-content-detection task), an intentionally-incoherent short surreal narrative (~1.3k tokens), and a 220-line log with near-total literal repetition (~13k raw tokens, the case above).

Additional observations
The tokens X → Y header and the saved N (P%) figure in headroom inspect output appear to disagree with each other in some cases (e.g. tokens 21351 → 20305 is a delta of 1,046, but the same line reports saved 14). Possibly the header reflects total conversation-token growth across turns while saved/% reflects only the isolated transform's effect — but this wasn't clear from the CLI output alone and may be worth clarifying in headroom inspect's own docs/output formatting.
headroom doctor's savings check (16 tokens / $0.00 saved lifetime) did not move at all across dozens of real requests totaling well over 100k input tokens combined, across the whole troubleshooting session.
Request

This looks related to a secondary symptom noted in the now-closed #1701 (whose primary/titled bug — a Windows proxy hang — was fixed by #1738), but that symptom doesn't appear to have been specifically addressed, and reproduces cleanly here on the current release (v0.37.0) with no hang at all. Filing separately since the two are independent. Is this a known/duplicate issue elsewhere, or is there a configuration step we're missing that isn't yet documented? Happy to provide the full headroom inspect transcripts or test files (synthetic, non-sensitive) if useful.

UniversePeak:
## Description

Closes #3363

A savings profile can set `force_kompress=True`. When the Kompress model is installed but its weights are still cold or unavailable, the forced router selected `KOMPRESS`, which passed content through unchanged. That made structured payloads lose their normal SmartCrusher/log/search fallback and looked like a chat-completions compression outage.

## Type of Change

- [x] Bug fix (non-breaking change)

## Changes Made

- Check Kompress readiness before honoring forced routing.
- Trigger the existing background model warmup when the model is cold.
- Re-enter normal content detection while the model is unavailable, preserving structural compressors instead of silently forcing a no-op.
- Extend `headroom doctor` to query `/health` and report the optional Kompress component as ready, disabled, or not ready with an actionable hint.

## Testing

- [x] Focused doctor tests pass (`python -m pytest -q tests/test_cli_doctor.py`)
- [x] Ruff check and format check pass on all changed files
- [x] Python compile checks pass for all changed files
- [x] Added a regression test for cold forced-Kompress routing with a structural fallback

### Test Output

```text
python -m pytest -q tests/test_cli_doctor.py
96 passed in 31.66s

python -m ruff check headroom/cli/doctor.py headroom/transforms/content_router.py tests/test_cli_doctor.py tests/test_agent_savings.py
All checks passed!
```

## Real Behavior Proof

- Environment: Windows, Python 3.11.9; Headroom source checkout at the current `main` base.
- Exact command / steps: The regression test supplies a cold Kompress implementation (`is_ready() == False`) and a structural log compressor, then runs `ContentRouter.compress()` with forced routing enabled.
- Observed result: The router requests background warmup and chooses the structural `LOG` strategy, returning the compacted output rather than a forced-KOMPRESS passthrough. `headroom doctor` now surfaces a cold model as `not ready — content compression is passing through`.
- Not tested: Full native Rust-backed proxy integration on this host. The repository's supported editable setup reached the Rust build but failed because the Windows MSVC linker is not installed (`link: extra operand`; the environment has no Visual Studio C++ build tools).

## Runtime Rollout Safety

- Minimum rollout channel: Default behavior is unchanged unless forced Kompress routing is enabled.
- Stable/default behavior changed: Cold forced routing now uses the existing detector and safe structural compressor.
- Kill switch / disable path: Existing Kompress disable settings remain unchanged.
- Qualification impact: None.

## Review Readiness

## Checklist

- [x] One logical bug fix
- [x] Added regression coverage
- [x] Did not edit `CHANGELOG.md`

### DIFF
--- a/headroom/cli/doctor.py
+++ b/headroom/cli/doctor.py
@@ -565,6 +565,48 @@ def check_savings(stats: dict[str, Any] | None, savings_file: Path) -> CheckResu
     return CheckResult(name=name, status=PASS, summary=f"{summary} ({source})")
 
 
+def check_kompress_health(health: dict[str, Any] | None) -> CheckResult:
+    """Report whether the optional Kompress model can actually run.
+
+    ``/livez`` only proves that the proxy process is alive, and ``/stats``
+    cannot distinguish a healthy zero-savings workload from a cold Kompress
+    model. The health payload has the component-level readiness state needed
+    to explain the "ML extras installed, but every request passes through"
+    failure mode.
+    """
+    name = "kompress"
+    if health is None:
+        return CheckResult(name=name, status=SKIP, summary="proxy health endpoint not reachable")
+
+    checks = health.get("checks")
+    component = checks.get(name) if isinstance(checks, dict) else None
+    if not isinstance(component, dict):
+        return CheckResult(
+            name=name,
+            status=WARN,
+            summary="proxy does not report Kompress readiness (older version?)",
+            hint="restart the proxy on the current version",
+        )
+
+    if not bool(component.get("enabled", True)):
+        return CheckResult(name=name, status=PASS, summary="disabled")
+
+    backend = component.get("backend")
+    backend_text = f" ({backend})" if backend else ""
+    if bool(component.get("ready")):
+        return CheckResult(name=name, status=PASS, summary=f"ready{backend_text}")
+
+    return CheckResult(
+        name=name,
+        status=WARN,
+        summary=f"not ready{backend_text} — content compression is passing through",
+        hint=(
+            "pre-download the Kompress model and retry; inspect /debug/warmup "
+            "for the resolved load state"
+        ),
+    )
+
+
 def check_budget(stats: dict[str, Any] | None) -> CheckResult:
     """Is a spend budget configured on the proxy?"""
     name = "budget"
@@ -704,6 +746,7 @@ def doctor(port: int, emit_json: bool) -> None:
     """
     base_url = f"http://127.0.0.1:{port}"
     livez = probe_json(f"{base_url}/livez")
+    health = probe_json(f"{base_url}/health", timeout=5.0) if livez else None
     stats = probe_json(f"{base_url}/stats", timeout=5.0) if livez else None
     installed = get_version()
 
@@ -720,6 +763,7 @@ def doctor(port: int, emit_json: bool) -> None:
         check_wrap_marker_staleness(project_local_claude_settings),
         check_codex_routing(codex_config_path(), port),
         check_shell_env(os.environ, port),
+        check_kompress_health(health),
         check_savings(stats, savings_path()),
         check_budget(stats),
     ]
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -2048,6 +2048,36 @@ class ContentRouter(Transform):
             "on",
         }
 
+    def _force_kompress_ready(self) -> bool:
+        """Return whether forced Kompress routing can actually run.
+
+        ``force_kompress`` is an opt-in routing preference, not a reason to
+        disable every other compressor. A cold or unavailable ML model used to
+        turn otherwise compressible logs/JSON into a silent passthrough because
+        the forced KOMPRESS strategy returned its input and the normal
+        structural fallback was never considered. Keep the background warmup,
+        but let the regular detector choose a safe structural compressor until
+        the model is ready.
+        """
+        if not self.config.enable_kompress:
+            return False
+        try:
+            compressor = self._get_kompress()
+        except Exception:
+            return False
+        if compressor is None:
+            return False
+        try:
+            ready = bool(compressor.is_ready())
+        except Exception:
+            return False
+        if not ready:
+            try:
+                compressor.ensure_background_load()
+            except Exception:
+                logger.debug("Kompress background warmup failed", exc_info=True)
+        return ready
+
     def _kompress_model_ready(self) -> bool:
         """Whether the ML compressor is ready (or deliberately disabled).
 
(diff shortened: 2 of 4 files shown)


JerrettDavis (approved):
Reviewed a68a48d52. Forced Kompress routing now checks model readiness before bypassing content detection; cold or unavailable models can use the existing structural routes while warmup is scheduled. The doctor check reads the existing component health payload without treating optional Kompress as process liveness. Focused validation passes (134 agent-savings and doctor tests). No blocking code findings; ready for final human review subject to current CI status.

JerrettDavis (approved):
Reviewed exact head 58c588d14d. The cold forced-Kompress path now preserves structural routing while still initiating warmup, and doctor reports the existing component health contract correctly. A current-main merge simulation was conflict-free; 207 impacted tests passed locally (2 optional skips), Ruff passed, and hosted checks are green. Approving.

PR #3385, opened 2026-09-02
fix(savings): price compression savings at the request's realized input rate

gglucass:
## Description

The persisted lifetime ledger prices every compression-saved input token at the model's full list rate, while `_estimate_input_cost_usd`, in the same module, prices the request's real spend on the cache read/write/uncached split. The two sides of the savings fraction use different price sheets.

That asymmetry matters because compression re-applies the same removals on every warm-prefix turn: clients resend the original transcript, the proxy re-removes the same bytes to keep the wire form byte-identical to the cached prefix, and each re-application lands in the ledger again. On a 94%-cache-read mix (the dominant Claude Code profile) most of those saved-token instances would have billed at the provider's cache-read discount, not list, so flat list pricing overstates the layer roughly 6-7x. A user-run audit of a live install's `/stats` reached the same figure independently (12.5% headline vs 2.4% priced at the rate actually paid) and used it to dispute the product's honesty.

This PR prices saved tokens at the request's realized blended input rate whenever the caller supplies the cache breakdown, reusing `_estimate_input_cost_usd` unchanged. The rate self-adjusts: a cold request (uncached only) still prices at list, a warm-prefix recurrence prices near the cache-read rate, no session state needed. It is the lifetime twin of `cost.py`'s session-scoped `cache_aware_savings_usd`, and the compression-layer sibling of #3171 (tool-schema deferral at the cache-read rate). Callers without a breakdown are byte-for-byte unchanged.

## Type of Change

## Changes Made

- `_estimate_compression_savings_usd` gains optional `cache_read_tokens` / `cache_write_tokens` / `uncached_input_tokens` kwargs; with a breakdown present it prices saved tokens at `_estimate_input_cost_usd(split) / split_tokens`, otherwise the historical flat-list path runs unchanged.
- `estimate_request_savings_usd` accepts and threads the same split into the `compression` bucket; `tool_schema` is deliberately untouched (owned by #3171).
- `record_request`'s unpriced fallback and `record_lifetime_request`'s `compression_savings_usd` default pass the split they already receive.
- `record_compression_savings` (checkpoint path, no split available) keeps flat-list pricing, now with a comment saying so.
- `prometheus_metrics.record_request` threads `cache_write_tokens` / `uncached_input_tokens` into `estimate_request_savings_usd`.
- One-time restatement of the persisted ledger onto the realized basis at load (`_reprice_state_to_realized_basis`), so an upgraded install does not carry a mix of list-priced and realized-rate dollars. No new per-entry field: every aggregate (lifetime, history point, project, model, display session, `lifetime_metrics.cost`) already carries `tokens_saved` next to `total_input_tokens`/`total_input_cost_usd`, so the dollars are recomputed from them, each at its own realized rate. A persisted `pricing_basis` marker makes it run once; a ledger with no realized rate yet is left alone and retried on the next load rather than freezing list prices in under the new name.
- New `tests/test_compression_savings_cache_aware.py`; six stubs in `tests/test_proxy_savings_history.py` widened to accept the new kwargs. Two existing expectations restated with the ledger (`test_non_finite_state_values_coerce_to_defaults`, `test_savings_tracker_migrates_v4_lifetime_to_v5_metrics_and_preserves_legacy_state`) - both now assert the recomputed figure rather than the list-priced one.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_compression_savings_cache_aware.py tests/test_proxy_savings_history.py tests/test_savings_tracker_zero_price.py tests/test_stats_new_input_savings_rate.py -q
53 passed, 1 warning in 5.42s
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), Python 3.12 via uv, this branch's worktree, real litellm pricing catalog (no stubs).
- Exact command / steps: `uv run --frozen python` replaying one request through `SavingsTracker.record_request` with `model=claude-opus-4-1, tokens_saved=100000, cache_read=940000, cache_write=40000, uncached=20000` (the measured mix from a live install's `/stats`), then printing the lifetime ledger against the old flat-list estimate for the same tokens.
- Observed result: realized-rate ledger $0.2460 vs flat-list $1.5000 for the identical request, a 6.1x correction on catalog opus pricing.
- Not tested: Windows; installs without litellm (that fallback path is covered by unit tests only); the desktop app's consumption of the changed dollar slope.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. This is accounting, not a request-path behavior change.
- Stable/default behavior changed: Yes, reported `compression_savings_usd` accrues at a lower, cache-aware rate on cache-heavy traffic. Request bodies, forwarding, and token counts are untouched.
- Kill switch / disable path: Not applicable; callers that omit the breakdown keep the old pricing by construction.
- Qualification impact: Downstream consumers of `compression_savings_usd` (dashboard, desktop app) will see a lower accrual slope AND a one-time downward restatement of the existing lifetime figure on first load after upgrade. Derived series move with it: history points are restated at their own realized rate, which keeps `/stats-history` day/provider/model deltas non-negative. Consumers that accumulate their own totals from those deltas should expect the step change.
- Rollback path: Revert the commit; the ledger schema is unchanged.

## Pricing-basis coexistence with #3171

Until #3171 lands, `estimate_request_savings_usd` returns `compression` at the realized rate beside `tool_schema` at list rate: two price sheets in one payload, with `tool_schema` the inflated one. That is deliberate (the tool-schema layer is #3171's to reprice, and doing it here would collide), but if #3171 slips, the payload ships mixed bases and `tool_schema` keeps the ~6x overstatement on cache-heavy traffic. The persisted restatement above covers `compression_savings_usd` only, for the same reason.

## Review Readiness

### DIFF
--- a/headroom/proxy/prometheus_metrics.py
+++ b/headroom/proxy/prometheus_metrics.py
@@ -788,6 +788,8 @@ class PrometheusMetrics:
             tool_schema_tokens_saved=tool_search_saved,
             output_tokens_saved=output_tokens_saved,
             cache_read_tokens=cache_read_tokens,
+            cache_write_tokens=cache_write_tokens,
+            uncached_input_tokens=uncached_input_tokens,
         )
         async with self._lock:
             self.requests_total += 1
--- a/headroom/proxy/savings_tracker.py
+++ b/headroom/proxy/savings_tracker.py
@@ -243,8 +243,47 @@ def _resolve_litellm_model(model: str) -> str:
     return model
 
 
-def _estimate_compression_savings_usd(model: str, tokens_saved: int) -> float:
-    """Estimate compression savings in USD from saved input tokens."""
+def _estimate_compression_savings_usd(
+    model: str,
+    tokens_saved: int,
+    *,
+    cache_read_tokens: int = 0,
+    cache_write_tokens: int = 0,
+    uncached_input_tokens: int = 0,
+) -> float:
+    """Estimate compression savings in USD from saved input tokens.
+
+    When the caller supplies the request's cache read/write/uncached breakdown,
+    saved tokens price at that request's realized blended input rate — the same
+    split ``_estimate_input_cost_usd`` prices real spend with. Removed bytes
+    would have travelled in this request's mix: a removal on a cold request
+    would have billed near list (or the cache-write premium), while a removal
+    re-applied on a warm-prefix turn would have billed at the provider's
+    cache-read discount. Flat list pricing values every re-applied removal at
+    full rate every turn, which overstates cache-heavy traffic roughly 7x
+    (a 94%-cache-read mix bills near 10% of list). Lifetime twin of
+    ``cost.py``'s session-scoped ``cache_aware_savings_usd``.
+
+    Without a breakdown (legacy checkpoints, callers that never learned the
+    split) the historical flat-list behavior is unchanged.
+    """
+    saved = _coerce_int(tokens_saved)
+    if saved <= 0:
+        return 0.0
+    breakdown_tokens = (
+        max(_coerce_int(cache_read_tokens), 0)
+        + max(_coerce_int(cache_write_tokens), 0)
+        + max(_coerce_int(uncached_input_tokens), 0)
+    )
+    if breakdown_tokens > 0:
+        realized_input_cost = _estimate_input_cost_usd(
+            model,
+            0,
+            cache_read_tokens=max(_coerce_int(cache_read_tokens), 0),
+            cache_write_tokens=max(_coerce_int(cache_write_tokens), 0),
+            uncached_input_tokens=max(_coerce_int(uncached_input_tokens), 0),
+        )
+        return float(saved) * (realized_input_cost / float(breakdown_tokens))
     litellm = _get_litellm_module()
     if tokens_saved <= 0:
         return 0.0
@@ -336,6 +375,8 @@ def estimate_request_savings_usd(
     tool_schema_tokens_saved: int = 0,
     output_tokens_saved: int = 0,
     cache_read_tokens: int = 0,
+    cache_write_tokens: int = 0,
+    uncached_input_tokens: int = 0,
 ) -> dict[str, float]:
     """Price one request's distinct savings layers for external telemetry.
 
@@ -346,7 +387,11 @@ def estimate_request_savings_usd(
 
     return {
         "compression": _estimate_compression_savings_usd(
-            model, max(_coerce_int(compression_tokens_saved), 0)
+            model,
+            max(_coerce_int(compression_tokens_saved), 0),
+            cache_read_tokens=cache_read_tokens,
+            cache_write_tokens=cache_write_tokens,
+            uncached_input_tokens=uncached_input_tokens,
         ),
         "tool_schema": _estimate_compression_savings_usd(
             model, max(_coerce_int(tool_schema_tokens_saved), 0)
@@ -702,6 +747,8 @@ class SavingsTracker:
         if timestamp_dt is None:
             timestamp_dt = _utc_now()
 
+        # Checkpoint callers have no cache breakdown, so this path keeps the
+        # historical flat-list pricing (see _estimate_compression_savings_usd).
         delta_usd = _estimate_compression_savings_usd(model, delta_tokens)
 
         with self._lock:
@@ -810,7 +857,13 @@ class SavingsTracker:
         else:
             # No priced breakdown available: only message savings are known here,
             # so this path stays message-only exactly as before.
-            delta_savings_usd = _estimate_compression_savings_usd(model, delta_tokens_saved)
+            delta_savings_usd = _estimate_compression_savings_usd(
+                model,
+                delta_tokens_saved,
+                cache_read_tokens=cache_read_tokens,
+                cache_write_tokens=cache_write_tokens,
+                uncached_input_tokens=uncached_input_tokens,
+            )
         delta_output_savings_usd = (
             max(_coerce_float(priced.get("output_shaping")), 0.0)
             if priced is not None
@@ -987,7 +1040,13 @@ class SavingsTracker:
         )
         metrics.setdefault(
             "compression_savings_usd",
-            _estimate_compression_savings_usd(model, _coerce_int(metrics.get("tokens_saved"))),
+            _estimate_compression_savings_usd(
+                model,
+                _coerce_int(metrics.get("tokens_saved")),
+                cache_read_tokens=cache_read_tokens,
+                cache_write_tokens=cache_write_tokens,
+                uncached_input_tokens=uncached_input_tokens,
+            ),
         )
         metrics.setdefault(
             "cache_savings_usd", _estimate_cache_savings_usd(model, cache_read_tokens)
--- a/headroom/proxy/savings_tracker.py
+++ b/headroom/proxy/savings_tracker.py
@@ -35,6 +35,10 @@ HEADROOM_SAVINGS_PATH_ENV_VAR = _paths.HEADROOM_SAVINGS_PATH_ENV
 DEFAULT_SAVINGS_DIR = ".headroom"
 DEFAULT_SAVINGS_FILE = "proxy_savings.json"
 SCHEMA_VERSION = 5
+# Pricing era of the persisted dollar figures. State written before compression
+# savings moved to the request's realized input rate carries no marker and is
+# repriced once on load (_reprice_state_to_realized_basis).
+PRICING_BASIS_REALIZED = "realized"
 DEFAULT_MAX_HISTORY_POINTS = 5000
 DEFAULT_MAX_PROJECTS = 50
 DEFAULT_MAX_HISTORY_AGE_DAYS = 365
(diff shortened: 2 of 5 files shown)


chopratejas:
The core reasoning holds up: pricing saved tokens at list while `_estimate_input_cost_usd` prices real spend on the cache split was a genuine inconsistency, and self-adjusting via the request's own breakdown (cold prices at list, warm near cache-read, no session state) is the right mechanism.

I also checked the provider-genericity worry and it does **not** bite. I expected `info.get("cache_read_input_token_cost", input_cost_per_token)` to break on models where litellm stores the key as an explicit `None` — `.get` would return `None`, `float(None)` would raise, and the whole estimate would fall through to the flat $3/M fallback, which for gpt-5 at $1.25/M list would *raise* reported savings. Against the real catalog the keys are absent rather than null:

```
claude-opus-4-1   input 1.5e-05   cache_read 1.5e-06   cache_creation 1.875e-05
gpt-5             input 1.25e-06  cache_read 1.25e-07  cache_creation ABSENT
gemini-2.5-pro    input 1.25e-06  cache_read 1.25e-07  cache_creation ABSENT
deepseek-chat     input 2.8e-07   cache_read 2.8e-08   cache_creation 0.0
groq/llama-3.3    input 5.9e-07   cache_read ABSENT    cache_creation ABSENT
```

So the defaults resolve correctly and deepseek's explicit `0.0` is honored. Combined with the breakdown being populated in `anthropic.py`, `openai.py`, `gemini.py` and `streaming.py`, the fix is genuinely cross-provider. Good.

**The blocking concern is the persisted ledger, not the arithmetic.**

There's no pricing-basis version in `savings_tracker.py` — `_normalize_history_entry` normalizes schema *shapes*, not pricing *eras*. After this merges, the lifetime ledger holds list-priced dollars from before and realized-rate dollars from after, with no way to tell them apart, and the headline converges only as old entries age out.

That means the user who audited 12.5% vs 2.4% upgrades and still sees an inflated lifetime number — the exact thing this PR sets out to fix. Could we add a `pricing_basis` field and recompute-on-read for legacy entries, or do a one-time ledger reset with a release note? Either is fine; shipping without one leaves the reported bug substantially unfixed.

**Two smaller things:**

- The docstring calls the blended rate "the realized rate," but the PR's own rationale says removals are re-applied to keep the wire form byte-identical to the cached prefix — i.e. they concentrate in the *cached* region, whose marginal rate is cache-read, not blended. On the 940k/40k/20k mix that's $1.00/M vs the blended $1.64/M, so the ledger still overstates by ~64%. Much better than 6x, but worth saying it's an upper bound rather than the realized rate.
- The new `if saved <= 0: return 0.0` makes the later `if tokens_saved <= 0: return 0.0` unreachable.
- Until #3171 lands, `estimate_request_savings_usd` returns `compression` at realized rate beside `tool_schema` at list rate — two price sheets in one payload. Worth a note in the description in case #3171 slips.

gglucass:
All three taken. Thanks for checking the litellm catalog directly rather than taking the `.get` default on trust - that was the failure mode I was least sure about.

**The persisted ledger: restated, without a new per-entry field.**

You are right that this was the substantive gap: shipping as it stood left the audited user's lifetime number inflated until the old entries aged out, which is most of the reported bug.

I went with recompute-on-read rather than a `pricing_basis` per entry plus a reset, because the ledger already holds everything needed to tell the eras apart. Every persisted aggregate - lifetime, each history point, projects, by_model, display session, `lifetime_metrics.cost` - carries `tokens_saved` next to `total_input_tokens` and `total_input_cost_usd`, and that cost was already priced on the cache split. Their ratio *is* the realized rate for exactly the traffic that aggregate covers, so the dollars can simply be recomputed from tokens. No migration, no reset, no data loss, and it converges on the first load after upgrade rather than over weeks.

Details worth knowing when reviewing `_reprice_state_to_realized_basis`:

- Each aggregate uses its **own** rate where it has one, falling back to the lifetime rate. That matters for history: each point is restated at the rate that prevailed up to that point, so the cumulative series stays monotone and the `/stats-history` day/provider/model deltas (which `_build_rollup` derives from those cumulative points) stay non-negative. There is a test asserting the daily deltas specifically.
- A single persisted `pricing_basis` marker at the root makes it a one-time restatement, and it is written by `_save_locked` so it survives. The recomputation is idempotent anyway, since it derives from tokens rather than scaling the previous dollars.
- A ledger with no realized rate yet (state predating input-cost tracking) is **not** stamped: the dollars are left alone and the restatement retries on the next load, rather than freezing list prices in under the name "realized".
- `pricing_basis` is also exposed in `snapshot()` so a downstream consumer can tell what basis it is reading.

Two existing tests now assert the restated figure rather than the list-priced one (`test_non_finite_state_values_coerce_to_defaults`, `test_savings_tracker_migrates_v4_lifetime_to_v5_metrics_and_preserves_legacy_state`), both with a comment explaining why. Flagging that explicitly since changed expectations in existing tests deserve a second look rather than a rubber stamp.

One consequence I want visible rather than discovered: this moves the reported lifetime figure **down, once, on upgrade**, and the derived series move with it. That is the intent, but the desktop app and dashboard accumulate from those numbers, so the step change is now called out in Qualification Impact.

**"The realized rate" -> upper bound.** Corrected in the docstring, with your $1.00/M vs $1.64/M example: removals are re-applied to keep the wire form byte-identical to the cached prefix, so they concentrate in the cached region whose marginal rate is cache-read, not the blend. Bounding it exactly needs per-removal cache-region attribution that the savings path does not carry, so the blend stands as the tight cheap over-estimate - now labelled as one.

**Dead branch.** Removed.

**#3171.** Noted in the description under its own heading, including that the persisted restatement covers `compression_savings_usd` only, so if #3171 slips the `tool_schema` bucket keeps both the list basis and the ~6x overstatement.

JerrettDavis (changes requested):
The live request-rate change is useful, but the legacy migration does not preserve additive/monotonic history when the realized rate changes. _reprice_state_to_realized_basis multiplies cumulative saved tokens by each point's changing cumulative average rate. Reproduced via SavingsTracker load with day 1: 1M saved, 1M input, $10 input cost; day 2: still 1M saved, 10M cumulative input, $19 cumulative cost. The migrated history becomes [$10, $1.90], lifetime becomes $1.90, and history_response clamps the negative day-2 delta to zero, leaving daily totals of $10. This is a normal cold-to-cache-heavy mix, not corrupt input. The same weighting also need not reconcile lifetime with independently repriced model/project buckets. Please migrate from incremental evidence with consistent aggregation, or explicitly preserve/label legacy estimates where attribution cannot be recovered; do not mark mutually inconsistent dollar totals as one realized basis. Add a cold-to-warm rate-change regression asserting cumulative monotonicity and daily/lifetime reconciliation. All 50 focused savings/history tests pass locally after supplying the native module, but the existing migration fixture only covers rates that keep the rewritten cumulative series increasing.

gglucass:
Fixed in `1b6ea2dd`.

You are right, and the repro is exact: pricing each cumulative history point at its own cumulative average rate rewrites the series downward whenever the realized rate falls, and `_build_rollup` clamps the resulting negative delta to zero.

The migration now restates from increments. For each history point it takes the clamped deltas of `total_tokens_saved`, `total_input_tokens` and `total_input_cost_usd` - the same reading `_build_rollup` applies - prices the saved delta at that increment's own realized rate, and re-accumulates. Increments are non-negative by construction, so the restated series is monotonic and the day / provider / model deltas sum back to the lifetime. Lifetime is that final cumulative plus any savings outside the retained history (aged-out points, checkpoint writers) at the lifetime rate.

On per-bucket attribution: it is not recoverable from cumulative snapshots, so projects and models no longer invent their own weighting. They take the lifetime's effective compression rate (restated dollars over restated saved tokens), which keeps them reconciled with the headline rather than mutually inconsistent under one `pricing_basis` marker.

Regression added, parametrized over your exact case and over one where the warm day also saves: `test_cold_to_warm_rate_change_keeps_history_monotonic_and_reconciled`. Day 1 is 1M saved / 1M input / $10; day 2 carries 10M cumulative input / $19 cumulative cost, once with no new savings and once with 500k more. It asserts the cumulative series is non-decreasing, that the daily deltas sum to the lifetime, and that the by-model bucket equals the headline when one model covers every saved token. Under the old weighting the first case fails at [$10, $1.90], exactly as you reproduced.

```
uv run --frozen --extra dev pytest tests/ -k "savings or history or prometheus"
424 passed, 27 skipped, 11986 deselected, 3 warnings in 94.73s
```

JerrettDavis (approved):
The revised migration resolves my earlier decreasing-history finding: it prices nonnegative history increments and reaccumulates them, then applies one effective compression rate to aggregate buckets. The cold-to-warm regression now verifies monotonic cumulative dollars and daily/lifetime reconciliation. All 52 focused pricing, persistence, and history tests pass. The live blended rate and legacy restatement remain estimates from aggregate evidence, as described; no remaining blocking code finding on this head.

gglucass:
Superseded by #3699 (89a58fd1) - closing.

Both halves of this PR are answered there, and the second one is answered *against* the approach taken here:

**Forward pricing.** `counterfactual.price_savings` prices removals against the observed cache mix per region, and `savings_tracker` now calls it with `Region.LIVE_ZONE` for compression. That is strictly better than this branch's realized blended input rate - and this branch's own docstring said so: "The blended rate is an UPPER BOUND ... removals concentrate in the cached region, whose marginal rate is cache-read rather than the blend." #3699's `split_tokens` fills the read bucket first instead of apportioning pro-rata, which is exactly that correction.

**Historical restatement.** `_reprice_state_to_realized_basis` here rewrites already-persisted dollars by deriving a rate from `total_input_cost_usd / total_input_tokens`. #3699 considered and rejected that, for a reason I think is right: the cache mix was never persisted, so pre-v6 dollars cannot be re-derived - a ratio of two aggregates is not the mix those particular removals sat in. It labels instead (`savings_basis`, `savings_basis_migrated_at`, `compression_savings_list_usd`), so a reader can see which part of a lifetime total predates the flip rather than trusting a restatement built on a proxy.

Rewriting users' historical money figures on a proxy rate is the riskier of the two, so I am happy to defer. No residual from this branch worth carrying forward.

PR #3386, opened 2026-09-02
fix(stats): gate the output-shaping claim on the shaper being active

gglucass:
## Description

`/stats` builds its `savings.by_layer.output_shaping` entry from the persisted output-savings ledger (`get_recorder().estimate(...)`) without checking whether the shaper is actually running. The ledger survives restarts and can span periods or key schemas where the arms accumulated unshaped traffic (the pre-#3209 coarse conversation keys locked whole projects into one arm). Result, observed in the field: a stable-channel install with the shaper `blocked_by_channel` published a "measured" -147.9% output reduction over 19,644 requests — a claim about shaping that deployment never performed. A user audit quoted that line back at us as evidence the numbers can't be trusted, and they were right to.

This PR fixes it at both ends. The /stats claim is gated: when the shaper is disabled (rollout-blocked channel, kill switch, config), the layer reports `active: false, method: "inactive"` with a zero claim, while active deployments publish exactly what they did before (including honest negatives, which #3127's holdout gating owns). And the ledger no longer accumulates the data that made the claim bogus: a treatment observation is recorded only when the request was actually shaped.

That second half came out of review. The gate alone fixes the install that reported the -147.9% and leaves the next install exposed, because a ledger entry carries no shaper state - so unshaped traffic accumulated while blocked republishes the same number on the first request after the channel unblocks. The stratum label is attached whenever the shaper is *configured* on, but the shaping itself can still be a no-op, and a treatment arm holding unshaped output is precisely what manufactures a reduction the deployment never performed.

Same spirit as #3377: Headroom only claims what it measured on this deployment's traffic.

## Type of Change

## Changes Made

- New module-level `_output_reduction_payload(shaper_active, est)` in `headroom/proxy/server.py`; the /stats route now tracks `_shaper_active` from `OutputShaperSettings.from_env(...).enabled` and delegates payload construction to it.
- Inactive shaper: `{"available": true, "active": false, "method": "inactive", "tokens_saved": 0, ...}` so consumers can render "inactive" instead of mistaking absence for a zero measurement.
- Active shaper: unchanged payload plus an additive `active: true` key.
- `SavingsRecorder.record_from_labels` records a treatment observation only when the shaper's own `output_shaper:verbosity:` label is present on the same channel. That label is appended only when `shape_request` actually changed the request (`output_shaper.py` sets `changed` and the label together, and every handler extends labels only on `changed`), so it is the per-request proof that shaping happened. Control is unshaped by definition and records unconditionally.
- `SavingsLedger` persists a `shaped_only` marker; arms written before that rule are dropped on load with a warning, since they cannot be told apart entry by entry. The offline baseline is kept - it is learned from pre-shaper history, costs a `learn --verbosity` run to rebuild, and was never the poisoned part.
- Every payload branch emits the same key set, nulls where a value does not exist (`method`, `band_is_ci`, `ci_low_percent`, `ci_high_percent`), so a consumer reading the band cannot KeyError on the inactive or no-data paths. Numeric totals stay 0 rather than null because callers already sum them.
- New `tests/test_stats_output_reduction_gate.py` covering the blocked-install shape (the real -147.9% figures), the active passthrough, the no-data case, key-set stability across all four branches, the unshaped-treatment drop, and the legacy-ledger arm drop.
- Existing tests that recorded treatment from a bare stratum label now attach the shaper label (`tests/test_output_savings.py`, `tests/test_output_savings_cli.py`) - they encoded the old contract.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_stats_output_reduction_gate.py tests/test_routing_stats_seam.py tests/test_dashboard_token_savings.py -q
3 passed; 8 passed
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), Python 3.12 via uv, this branch's worktree.
- Exact command / steps: Reproduced the field payload by calling `_output_reduction_payload` with the exact `SavingsEstimate` a blocked stable install published (tokens_saved=-7,522,490, pct=-147.9, n_requests=19,644, kind="measured"), shaper_active=False, then shaper_active=True.
- Observed result: inactive path returns `method="inactive", tokens_saved=0, reduction_percent=0.0`; active path returns the estimate byte-identical to today's payload plus `active: true`.
- Not tested: an end-to-end /stats call on a live rollout-blocked install (no blocked-channel machine at hand); the desktop app renders from its own ledger recompute since 0.8.7 and does not consume this field for its headline.

## Runtime Rollout Safety

- Rollout-managed feature(s): Reads the output-shaper's rollout state; changes no rollout behavior itself.
- Stable/default behavior changed: Yes, for deployments where the shaper is disabled: the /stats output_shaping layer reports inactive/zero instead of the persisted ledger estimate. Deployments with the shaper active are unchanged.
- Kill switch / disable path: Not applicable; the gate follows the shaper's own enablement, which already has one.
- Qualification impact: Blocked-channel installs stop advertising phantom (often negative) output savings in /stats and telemetry built on it.
- Rollback path: Revert the commit; the payload shape addition (`active`) is additive and the ledger format is untouched.

## Review Readiness


chopratejas:
Agreed on the principle, and the seam extraction makes it testable — happy to see this land as a stopgap. Two things though, one of which I think matters more than the bug being fixed.

**The poisoned ledger comes back the moment the channel unblocks.**

The description says "the ledger is untouched, so nothing is lost for when the channel unblocks." Looking at `output_savings.py`, `OutputSavingsLedger.record(arm, key, output_tokens)` stores arm, key and tokens — there's no shaper-active tag on an entry. So the unshaped traffic accumulated during the blocked period is indistinguishable from shaped traffic once the gate opens, and `/stats` will publish the same -147.9% again on the first request after activation.

This fixes the install that reported it and leaves the next install — the one that just unblocked — fully exposed, with the added downside that the number will now appear *after* we told the user we'd fixed it. Could we tag each entry with the shaper state at record time and have `estimate()` count only shaped-era entries, or segment the ledger on activation transitions? If that's too big for this PR, an issue linked from here would be enough; I just don't want the gate to read as a complete fix.

**Payload shape now has three variants with different key sets.**

| branch | keys |
|---|---|
| inactive | `available, active, method, tokens_saved, baseline_tokens, reduction_percent, requests` |
| active + data | above plus `band_is_ci, ci_low_percent, ci_high_percent` |
| active, no data | `available, active` |

A consumer doing `payload["ci_low_percent"]` or `payload["band_is_ci"]` KeyErrors on the inactive path. The no-data case had this already, but the inactive path is new and looks populated while missing three keys, which is the easier one to get wrong. Emitting the full key set with nulls would keep the contract stable — worth it given the desktop app and dashboard both read this.

gglucass:
You were right that this mattered more than the bug being fixed, and "nothing is lost for when the channel unblocks" was wrong as written. Both fixed in this PR rather than deferred to an issue.

**Tagged at the source instead of at `estimate()`.**

Rather than add a shaper-state field to each entry, I used the state that is already on the same channel. `record_from_labels` now records a treatment observation only when the shaper's own `output_shaper:verbosity:` label is present. That label is not decorative: `output_shaper.py` sets `result.changed = True` and appends it in the same breath (four call sites, all four shaping functions), and every handler extends labels only under `if shape_result.changed`. So its presence is the per-request proof that the request was actually shaped.

That closes a hole slightly wider than the one you named. The stratum label is attached whenever the shaper is *configured* on, not when it shapes - so even on an install that never had a blocked channel, a treatment arm can fill with unshaped output whenever the shaping is a no-op (level resolves to none, nothing steerable in the body). A treatment arm holding unshaped output is exactly what manufactures a "measured" reduction from traffic nobody shaped. Control is unshaped by definition, so it records unconditionally, unchanged.

**The already-poisoned arms.**

Entries written before this rule cannot be told apart entry by entry, which is the whole problem, so `SavingsLedger` now persists a `shaped_only` marker and drops pre-marker arms on load with a warning. Arms re-accumulate from live traffic in hours. The offline `baseline` is deliberately kept: it is learned from pre-shaper history, costs a `learn --verbosity` run to rebuild, and was never the poisoned part.

I preferred this to segmenting on activation transitions because segmentation still has to decide what to do with the ambiguous segment, and the answer for that segment is "do not trust it" either way.

**Payload shape.**

All four branches (inactive, active-no-data, active-with-data, and the defensive fallback) now emit the same key set, with nulls for `method`, `band_is_ci`, `ci_low_percent`, `ci_high_percent` where no value exists. Numeric totals stay `0` rather than null, because `/stats` already sums `tokens_saved` and `reduction_percent` downstream and nulls would push a type check onto every consumer. There is a test asserting the key sets are identical across all four.

**Test-contract change worth a look:** six existing assertions in `tests/test_output_savings.py` / `tests/test_output_savings_cli.py` recorded treatment from a bare stratum label. They now attach the shaper label, since recording without it is the behaviour this PR removes. Flagging it rather than letting it pass as a mechanical edit.

201 tests pass across the output-shaping suites, plus ruff and mypy on the touched modules.

JerrettDavis (changes requested):
The 61 focused ledger/CLI/stats tests pass, but the /stats gate still publishes old shaping claims when effective steering is disabled. Reproduced through the actual /stats route with rollout enablement true, HEADROOM_VERBOSITY_LEVEL=0, and a prior measured estimate: shape_request(...).changed is False, yet the endpoint returns active=true, method=measured, tokens_saved=200 and reduction_percent=20. _shaper_active is set from settings.enabled before the resolved level is considered. The stats settings also omit steering_allowed_for(proxy.config), while the real handlers use it to force level zero in cache mode. Please derive this layer active state from the same effective steering configuration as the request handlers, including cache-mode restrictions and level zero, before publishing historical estimates. Add endpoint-level regressions for explicit zero, learned/controller zero, and cache mode with an existing ledger; testing _output_reduction_payload(False, ...) alone cannot catch the incorrect boolean passed by the endpoint.

gglucass:
Fixed in 3d62581e. The gate was reading `settings.enabled` before the level was resolved, and it built its settings without `steering_allowed_for`, so cache mode never reached it at all.

/stats now builds the same `OutputShaperSettings` the request handlers build - including `steering_enabled=steering_allowed_for(proxy.config)` - resolves the level through `resolve_verbosity_level`, and requires `level > 0`. That is the same test `shape_request` applies before it touches a body, so the published claim and the shaping now agree by construction rather than by coincidence.

Endpoint-level regressions through the real `/stats` route, each with an estimate already in the ledger: explicit `HEADROOM_VERBOSITY_LEVEL=0`, a learned `verbosity.json` zero, a controller zero under `HEADROOM_VERBOSITY_AUTOTUNE`, and `mode="cache"` with a level set - plus the positive case at level 3 to pin that a live shaper still publishes. The four gating tests fail against the previous commit.

The branch is also merged up to main; the conflict was an unrelated adjacent addition in server.py.

JerrettDavis (approved):
Re-reviewed the effective steering gate after the main refresh. The stats path now uses steering_allowed_for and the resolved verbosity level, matching request shaping for cache mode and explicit/controller/learned level zero. The original endpoint reproduction now correctly returns active=false, method=inactive, and zero savings when shaping changes nothing at level zero. All 66 focused stats, ledger, and CLI tests pass. The earlier blocking gate mismatch is resolved.

JerrettDavis (approved):
Re-reviewed after resolving the new upstream-main conflict. Preserved the actual-shaping gate together with main’s conversation labels and qualified cluster accounting. Updated inherited fixtures to distinguish shaped legacy traffic from traffic without conversation provenance. All 80 focused savings, stats, CLI, and conversation-label tests pass. Ruff lint/format and diff checks pass. Remote CI is pending on this refreshed head; no PR merge performed.

PR #3387, opened 2026-09-02
fix(compression): pause AST compression per language after repeated invalid-syntax discards

gglucass:
## Description

The AST pass validates its own output and returns the original on failure ("never serve broken code"). That guarantee is right, but it means a grammar that keeps mangling real code costs the full compression latency on every request and then throws the work away. Field measurement from one install's retained logs: 356 invalid-syntax discards (TypeScript 270, Python 47, JS 27, PHP 8, C++ 4), each preceded by ~1.2s in `compressor:mixed` - the slowest stage in that install's pipeline spending most of its time on work that was discarded, while the user files the added latency under "the proxy is slow".

This PR adds a dynamic per-language circuit breaker beside the existing static `_UNSAFE_TREE_SITTER_LANGUAGES` quarantine (which is for grammars unsafe to load; this is for grammars that load fine but fail validation on real traffic). Once 10 of the last 20 validations for a language fail, attempts for that language pause for 30 minutes and requests pass through at original size immediately. The trip logs one warning with the counts; the window clears on trip so a fresh run of failures must re-earn the next pause. Other languages are isolated, and compression output is unchanged in every case - the breaker only decides whether to attempt, never what to serve.

## Type of Change

## Changes Made

- `headroom/transforms/code_compressor.py`: module-level breaker state (`_SYNTAX_BREAKER_WINDOW=20`, `_SYNTAX_BREAKER_MIN_FAILURES=10`, `_SYNTAX_BREAKER_COOLDOWN_S=1800`), `_syntax_breaker_open()` checked after language detection and before the parse, `_record_syntax_outcome()` fed by the existing syntax-validation result (including the Python recovery retry's final verdict). Lock-guarded; compression already runs in a thread pool.
- Kill switch: `HEADROOM_CODE_SYNTAX_BREAKER=0|false|off` disables both recording and skipping.
- New `tests/test_code_compressor_syntax_breaker.py`: trip threshold, success dilution, cooldown expiry and re-earn, env kill switch, and a short-circuit test proving `_compress_with_ast` is not invoked while open.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_code_compressor_syntax_breaker.py tests/test_code_compressor_language_alias.py tests/test_code_compressor_thread_safety.py -q
33 passed, 8 skipped
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), Python 3.12 via uv, this branch's worktree.
- Exact command / steps: pytest run above; the short-circuit test pre-opens the breaker for python and monkeypatches `_compress_with_ast` to raise if reached, then calls `CodeAwareCompressor.compress()` on a 20x-repeated function body.
- Observed result: with the breaker open the compressor returns the original body byte-identical (ratio 1.0, syntax_valid) without entering the AST pass; with it closed behavior is unchanged.
- Not tested: a live proxy soak against the field TypeScript corpus that produced the 270 discards (not in the repo); the 10/20/30min constants are judgment values, tunable in one place at the top of the module.

## Runtime Rollout Safety

- Stable/default behavior changed: Yes, narrowly: a language failing validation on at least 10 of its last 20 attempts passes through uncompressed for 30 minutes instead of being attempted-and-discarded every request. Served bytes are identical in both cases (the discard path already returned the original).
- Kill switch / disable path: HEADROOM_CODE_SYNTAX_BREAKER=0 restores today's always-attempt behavior.
- Qualification impact: Installs with a failing grammar lose the latency of doomed attempts and a small amount of would-be compression on the failing language only; the warning names the language and counts when it trips.
- Rollback path: Revert the commit; no state or schema outlives the process (breaker state is in-memory only).

## Review Readiness

### DIFF
--- a/headroom/proxy/server.py
+++ b/headroom/proxy/server.py
@@ -476,6 +476,16 @@ logging.basicConfig(
 logger = logging.getLogger("headroom.proxy")
 
 
+def _code_syntax_breaker_status() -> dict[str, Any]:
+    """Per-language AST-compression breaker state, or {} if unavailable."""
+    try:
+        from headroom.transforms.code_compressor import syntax_breaker_status
+
+        return syntax_breaker_status()
+    except Exception:  # pragma: no cover - defensive; stats must not fail
+        return {}
+
+
 class _SuppressCancelledErrorFilter(logging.Filter):
     """Hide expected uvicorn CancelledError tracebacks during shutdown."""
 
@@ -4599,6 +4609,9 @@ def create_app(config: ProxyConfig | None = None) -> FastAPI:
                 "ccr_retrievals": compression_stats.get("total_retrievals", 0),
             },
             "compression_cache": compression_cache_stats,
+            # Per-language AST compression pauses. Empty on a healthy install;
+            # non-empty is the explanation for a savings drop in one language.
+            "code_syntax_breaker": _code_syntax_breaker_status(),
             # Always False: the anonymous telemetry beacon was removed, so no
             # telemetry is ever shipped externally (local collection only).
             "anon_telemetry_shipping": False,
--- a/headroom/transforms/code_compressor.py
+++ b/headroom/transforms/code_compressor.py
@@ -120,12 +120,50 @@ _UNSAFE_TREE_SITTER_LANGUAGES: frozenset[str] = frozenset({"perl"})
 # on every request. The static _UNSAFE_TREE_SITTER_LANGUAGES set above is for
 # grammars unsafe to LOAD; this breaker is for grammars that load fine but
 # mangle real code.
+#
+# SCOPE: process-global, deliberately. One pathological payload pauses that
+# language for every session served by this process, which is right for the
+# single-user local proxy this ships as (a failing grammar is a property of the
+# process's tree-sitter build, not of the caller) and wrong for a shared or
+# hosted deployment, where one tenant would pause a language for the others.
+# Keying by tenant needs a caller identity plumbed through the Transform
+# interface, which carries code and language only; until something needs that,
+# HEADROOM_CODE_SYNTAX_BREAKER=0 is the per-deployment opt-out.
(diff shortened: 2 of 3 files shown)


chopratejas:
Nice framing of the distinction against `_UNSAFE_TREE_SITTER_LANGUAGES` — load-unsafe vs loads-fine-but-mangles is exactly the right split, and the invariant that the breaker only decides *whether to attempt* and never *what to serve* makes this easy to reason about. Lock discipline looks correct, per-language isolation works, and clearing the window on trip so a cooldown expiry must re-earn the next pause is a good detail. The short-circuit test earns its keep.

Three things:

**1. The breaker state is process-global, so it's cross-tenant.** `_syntax_breaker_outcomes` and `_syntax_breaker_open_until` are module-level dicts. On a shared or hosted proxy, one user's pathological TypeScript payload pauses TypeScript compression for every other session on that instance for 30 minutes. That's fine for a single-user local proxy and probably not fine elsewhere. Either key the state by tenant/session, or state the per-process scope explicitly in the module comment so the limitation is deliberate.

**2. The savings drop is invisible.** While the breaker is open, that language compresses at ratio 1.0 with only a `logger.debug`. `/stats` shows reduced savings with nothing explaining why — which is the same class of unexplained number that #3386 and #3377 are busy fixing. A counter or a `/stats` field (breaker open, per language, with the trip count) would keep this consistent with that work and make the field diagnosis one query instead of a log hunt.

**3. Test flake risk.** `monkeypatch.setattr(cc.time, "monotonic", ...)` patches the real `time` module's attribute for the duration of the test, not a module-local indirection. It's undone correctly, but anything else running concurrently — pytest-xdist, a background thread in another fixture — sees the fake clock. A module-level `_now()` wrapper to patch instead would remove the sharp edge.

On the constants: 10-of-20 with a 30-minute cooldown seems reasonable to me as a starting point, and I like that they're in one place at the top. No objection to shipping them as judgment values given the kill switch exists.

PR #3398, opened 2026-09-02
fix(output-savings): tag ledger entries with the shaper's state and estimate only from active ones

### LINKED ISSUE #3395: Output-savings ledger has no shaper-active tag: the -147.9% claim returns when the channel unblocks
## Summary

#3386 gates the `/stats` output-shaping claim on the shaper being active, which fixes the reported case (a rollout-blocked stable install advertising a "measured" -147.9% output reduction over 19,644 requests). It does not fix the underlying data: the claim comes back the moment the channel unblocks.

## Why the gate is not enough

`OutputSavingsLedger.record(arm, key, output_tokens)` in `headroom/proxy/output_savings.py` stores the arm, the stratum key and the token count. There is no record of whether the shaper was active when the entry was written.

So traffic accumulated while the shaper was blocked sits in the treatment and control arms indistinguishable from genuinely shaped traffic. `estimate()` counts it either way. #3386's gate suppresses the claim only while `shaper_active` is false — on the first request after the channel unblocks, `/stats` publishes the same poisoned estimate again, now with the added problem that we will have told users it was fixed.

The PR description frames the untouched ledger as a feature ("nothing is lost for when the channel unblocks"). What is preserved is unshaped traffic attributed to shaping.

## Suggested fix

Tag entries at record time and filter at estimate time:

- add a shaper-active (and ideally shaper-version / verbosity-level) field to each ledger entry
- have `estimate()` count only entries written while the shaper was active
- treat an activation-state transition as a segment boundary so pre-activation arms cannot contribute

Legacy entries have no tag. Excluding them is the honest default and consistent with the direction of #3377 and #3386 — an install that upgrades gets a smaller `n` rather than an inherited claim.

## Related

Gate: #3386. Holdout-arm gating: #3127, #3128. Key seeding: #3209. Same theme: #3377.

chopratejas:
## Description

The output-savings estimate is a counterfactual: traffic splits into a shaped treatment arm and an unshaped control arm, and the difference between them is the shaper's effect. `OutputSavingsLedger.record(arm, key, output_tokens)` stored the arm, the stratum and the token count — and nothing about whether the shaper was actually running when the entry was written.

The stratum label is attached whenever the shaper is *configured* on, but the shaping itself can be inert: level 0 means no steering at all, which `mode="cache"` forces and a learned or env level of 0 selects. Both arms keep filling while nothing is being shaped, and differencing them reports noise as a shaping effect. That is how a rollout-blocked stable install published a "measured" -147.9% output reduction over 19,644 requests.

#3386 gates the /stats *display* on the shaper being active, which suppresses the claim while it is off. The entries themselves stay unattributable, so the number returns on the first request after the channel unblocks. This PR fixes the data.

Closes #3395

### Overlap with #3386 — please read before merging both

#3386 gained a second commit (781c27f) after this issue was filed, and it now edits `output_savings.py` too, so these two PRs conflict and one should supersede the other's ledger half. The two take different approaches and the difference is not cosmetic:

- **#3386** drops a treatment observation unless the shaper's `output_shaper:verbosity:` label proves *this request's body* was changed, and records control unconditionally.
- **This PR** tags both arms with the shaper's resolved level at record time and filters both on it.

The asymmetry in #3386 is a bias in the same direction as the bug. `changed` is false for a request whose body offered nothing to steer, so the treatment arm becomes "requests the shaper found something to change" while control stays "every control request" — conditioning one arm on a post-assignment variable, which makes the arms different populations and biases the difference of means. More importantly it does not close the reported hole: while the shaper is configured on but inert (cache mode, level 0), #3386 drops treatment and *keeps* control, so control accumulates through the inactive period and pairs with treatment recorded after activation. #3395's "treat an activation-state transition as a segment boundary" only holds if both arms are segmented.

I have not touched #3386's branch, and `server.py` is untouched here — the /stats gate and this are complementary once the ledger half is resolved one way.

## Type of Change

## Changes Made

- `output_savings_policy.stratum_label(arm, key, *, verbosity_level=None)` appends the resolved level as a `:L<n>` tail; `parse_stratum_label` returns `(arm, stratum, level | None)`. The level rides the existing label rather than a new one so `telemetry.session` (which counts one entry per label, keyed on the text before the first colon) sees no new transform, and existing `startswith(prefix + key)` matches keep working. A stratum key is a fixed vocabulary joined with `|` and cannot contain a colon, so the tail is unambiguous.
- `SavingsLedger.record(..., *, shaper_active: bool)` — required, no default. Active observations go to `treatment`/`control`; inactive ones to new `inactive_treatment`/`inactive_control`, which no estimator reads. Making the arms themselves clean is the point: every current estimator gets the filter for free, and a future one cannot reintroduce the bug by forgetting to apply it.
- The tag is symmetric across arms and describes the deployment's state at record time, not what the shaper did to a particular body — see the overlap note above for why that distinction matters.
- Inactive observations are kept rather than discarded, and surfaced: `SavingsLedger.inactive_requests`, and `headroom output-savings` now prints the excluded count so a small `n` reads as "the shaper was off" instead of history disappearing.
- Persistence: the file carries `schema_version: 2`. A schema-1 file (the first load after any upgrade) cannot have its arms split entry by entry — the tag they would need is exactly what is missing — so they are dropped with a one-time warning and re-accumulated from live traffic. An upgraded install gets a smaller `n` rather than an inherited claim. The offline baseline survives: it is learned from pre-shaper history, is unshaped by definition, and costs a `learn --verbosity` run to rebuild. The warning is once-per-process because the recorder re-reads the file on every flush and every `estimate`.
- `SavingsRecorder.estimate_request_savings` returns 0 when the level says the shaper was inert. That number is priced in USD downstream, so letting it through would move the claim the ledger refuses to make onto the savings rollup and Prometheus instead.
- Three handler sites (`handlers/anthropic.py`, and the chat + Responses paths in `handlers/openai.py`) resolve the verbosity level *before* the arm branch and pass it into the label for both arms. Treatment behaviour is unchanged — same level, resolved once. Control now costs one `resolve_verbosity_level` call it did not make before (a couple of small file reads, and only on holdout traffic).
- Tests: new `TestShaperActiveTagging`, `TestRecorderTagsFromTheLabel` and `TestLedgerSchemaMigration` in `tests/test_output_savings.py`; level round-trip and untagged-label cases in `tests/test_output_savings_policy.py`. Existing call sites updated for the two signature changes.

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_stateless_writers.py tests/test_proxy_byte_faithful_forwarding.py \
    tests/test_proxy_savings_history.py tests/test_savings_tracker_zero_price.py tests/test_output_shaping_rollup.py \
    tests/test_openai_codex_ws_lifecycle.py tests/test_openai_responses_output_shaper.py tests/test_rollout.py \
    tests/test_cli_learn.py tests/test_output_savings_policy.py tests/test_output_savings_cli.py \
    tests/test_verbosity_learn.py tests/test_output_savings.py tests/test_output_shaper.py tests/test_runtime_env.py \
    tests/test_openai_responses_additional_tools.py tests/test_output_shaper_responses.py -q
463 passed, 1 warning in 22.84s

$ uv run --frozen --extra dev ruff check <touched files>
All checks passed!

$ uv run --frozen --extra dev mypy headroom/proxy/output_savings.py headroom/proxy/output_savings_policy.py headroom/cli/output_savings.py
```

Full suite: `pytest tests --ignore=tests/test_memory_eval.py` gives 11673 passed, 15 failed. Eleven are reproduced verbatim on a clean `origin/main` worktree and are environmental or unrelated (litellm not installed here, a Windows-only event-loop assertion, model-registry pricing, route-advice). The other four are `test_proxy_health.py` kompress-readiness tests, which pass in isolation on both trees and fail only inside the full run — order-dependent global state, and nothing this PR touches goes near kompress or `/readyz`. `tests/test_memory_eval.py` fails to import on `origin/main` too (no `litellm`).

## Real Behavior Proof

- Environment: macOS 15 (arm64), Python 3.14 via uv, this branch's worktree.
- Exact command / steps: wrote a schema-1 ledger holding the field shape — treatment `n=19,644` at mean 641 tokens, control `n=1,000` at mean 258, plus a 200-sample baseline — then loaded it, drove 100 requests through `SavingsRecorder` with the shaper inert (`L0`, both arms), then 100 more with it live (`L3`, treatment 700 / control 1000).
- Observed result:

```text
# same arms, unfiltered (pre-fix arithmetic)
unfiltered arms: kind=measured pct=-148.6 n=19644

# after
WARNING output-savings ledger predates the shaper-active tag (schema 1); dropping 1 treatment
        and 1 control strata and re-accumulating from live traffic (baseline kept)
legacy load:               kind=estimated pct=0.0  n=0  baseline_samples=200
100 inert-shaper requests: kind=estimated pct=0.0  n=0  excluded=100
per-request rollup for an inert treatment request: 0 tokens
100 shaped requests:       kind=measured  pct=30.0 n=50 excluded=100
on-disk schema_version=2
```

  The -148.6% is this reconstruction's figure for the field's -147.9%, not the field file itself.
- Not tested: an end-to-end proxy run on a real rollout-blocked install (no blocked-channel machine at hand); the handler label change is covered by the existing shaper/WS/Responses suites rather than by a live request.

## Runtime Rollout Safety

- Rollout-managed feature(s): reads the output shaper's already-resolved settings; changes no rollout behaviour itself.
- Stable/default behavior changed: Yes, for the reported failure mode. A deployment whose shaper is inert (level 0 / cache mode) reports no output-reduction estimate instead of one built from unshaped traffic, and every install drops its pre-upgrade arms once on first load. Deployments with the shaper live at level > 0 are unchanged apart from the one-time drop.
- Kill switch / disable path: not applicable; the filter follows the shaper's own level, which already has one.
- Qualification impact: blocked or cache-mode installs stop advertising output savings they never performed, in `/stats`, in `headroom output-savings`, and in the per-request USD rollup.
- Rollback path: revert the commit. A schema-2 file read by the reverted code still loads — the extra keys are ignored and the arms it reads hold shaped-only data, i.e. the conservative side.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A: the ledger's on-disk shape is not documented outside the module.

## Additional Notes

The verbosity level is carried on the label but the ledger stores only the active/inactive split, not per-level arms. Segmenting strata by level would fragment an already sparse per-stratum baseline for a distinction the arm contrast does not use; the level's job here is to answer "was anything being shaped", and that is a boolean.

https://claude.ai/code/session_01J7Jx24NvP1xXZRSn3WG1vi

### DIFF
--- a/headroom/cli/output_savings.py
+++ b/headroom/cli/output_savings.py
@@ -40,8 +40,19 @@ def output_savings() -> None:
     click.echo(f"\n{'=' * 56}")
     click.echo("Output-token reduction")
     click.echo(f"{'=' * 56}")
+    # Requests recorded while the shaper was inert (level 0 — cache mode, or a
+    # learned/env level of 0). They are real traffic but carry no shaping
+    # signal, so no estimate counts them. Saying so beats leaving an operator to
+    # wonder why a busy proxy reports nothing.
+    inactive = ledger.inactive_requests
+
     if est.n_requests == 0:
         click.echo("  No shaped requests recorded yet.")
+        if inactive:
+            click.echo(
+                f"  {inactive:,} request(s) were recorded while the shaper was inactive (level 0);"
+            )
+            click.echo("  they are excluded — an unshaped request measures nothing about shaping.")
         click.echo(
             f"  Baseline: {ledger.baseline.total_samples} samples, "
             f"{len(ledger.baseline.strata)} strata."
@@ -56,6 +67,8 @@ def output_savings() -> None:
     click.echo(
         f"  Reduction: {est.pct:.1f}%   (95% CI {est.ci_low_pct:.1f}% … {est.ci_high_pct:.1f}%)"
     )
+    if inactive:
+        click.echo(f"  Excluded:  {inactive:,} request(s) recorded while the shaper was inactive")
     if est.kind == "estimated":
         click.echo(
             "\n  Note: estimated vs the learned baseline. For a measured number,"
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -3232,12 +3232,18 @@ class AnthropicHandlerMixin:
                         model=model,
                         has_tools=bool(body.get("tools")),
                     )
-                    # Carry (arm, stratum) on the existing label channel so the
-                    # outcome funnel can feed the savings ledger from any path.
-                    transforms_applied.append(stratum_label(_arm, _stratum))
+                    # Carry (arm, stratum, level) on the existing label channel
+                    # so the outcome funnel can feed the savings ledger from any
+                    # path. The level is resolved for BOTH arms because it is
+                    # what tells the ledger the shaper was live when this
+                    # observation landed — level 0 (cache mode, or a learned/env
+                    # 0) shapes nothing while both arms keep filling, and an
+                    # observation the ledger cannot attribute to shaping is
+                    # excluded from the estimate rather than credited to it.
+                    _level, _src = resolve_verbosity_level(_shaper_settings)
+                    transforms_applied.append(stratum_label(_arm, _stratum, verbosity_level=_level))
 
                     if _arm == "treatment":
-                        _level, _src = resolve_verbosity_level(_shaper_settings)
                         shape_result = shape_request(body, _shaper_settings, level_override=_level)
                         if shape_result.changed:
                             body_mutation_tracker.mark_mutated("output_shaper")
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -746,12 +746,15 @@ def _shape_openai_responses_payload(
             model=model or str(payload.get("model", "")),
             has_tools=bool(payload.get("tools")),
         )
-        labels = [stratum_label(arm, stratum)]
+        # Resolved for both arms: the level tags the ledger entry with whether
+        # the shaper was live (level 0 = inert), and an untagged observation is
+        # excluded from the estimate rather than credited to shaping.
+        level, src = resolve_verbosity_level(settings)
+        labels = [stratum_label(arm, stratum, verbosity_level=level)]
 
         if arm != "treatment":
             return labels, False
 
-        level, src = resolve_verbosity_level(settings)
         shape_result = shape_responses_request(payload, settings, level_override=level)
         if shape_result.changed:
             labels.extend(shape_result.labels or [])
@@ -1459,11 +1462,14 @@ def _shape_openai_responses_for_output(
         model=model or str(payload.get("model") or ""),
         has_tools=bool(payload.get("tools")),
     )
-    result.labels.append(stratum_label(arm, stratum))
+    # Resolve the level BEFORE the arm branch: it tags the ledger entry for
+    # both arms with whether the shaper was live (level 0 = inert), and an
+    # observation the ledger cannot attribute is excluded from the estimate.
+    level, _source = resolve_verbosity_level(settings)
+    result.labels.append(stratum_label(arm, stratum, verbosity_level=level))
     if arm == "control":
         return result
 
-    level, _source = resolve_verbosity_level(settings)
     shaped = shape_openai_responses_request(
         payload,
         settings=settings,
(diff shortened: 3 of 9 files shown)


chopratejas:
Closing as superseded by #3386, and apologies for the noise — this was opened from issue #3395 without spotting that @gglucass had already implemented the same fix in #3386 (commit `fix(stats): record only shaped observations in the output-savings ledger`, 10:10Z) about an hour earlier.

#3386's approach is also the better one. It keys the treatment arm on the shaper's own per-request `output_shaper:verbosity:` label — proof that shaping actually happened on that request — rather than on whether the shaper was configured active. That additionally catches the case this PR would have missed: shaper enabled, but the shaping itself a no-op for that request (level resolves to none, or the body offers nothing to steer). A treatment arm holding unshaped output is exactly the thing being fixed, so the finer-grained signal is the right one.

Its `shaped_only` schema marker plus dropping pre-rule arms in `from_dict()` while keeping the offline baseline is also a cleaner migration than what was being built here — the baseline is learned from pre-shaper history and was never the poisoned part, so retaining it saves a `learn --verbosity` rebuild.

No review action needed. #3395 will be closed against #3386.

PR #3402, opened 2026-09-02
fix(copilot): close the CLI native-lane gaps and route Copilot bearers to Copilot

chopratejas:
## Summary

Closes the six gaps that made `headroom wrap copilot --native` (and `headroom install` for Copilot) look routed while traffic bypassed the proxy or never reached it, and adds tests pinning the Auto-mode passthrough contract. A second commit hardens the change after a deep review (details below).

The Copilot CLI resolves its API host as `settings.copilotUrl` → `COPILOT_API_URL` → token `endpoints.api`, honours `HTTP(S)_PROXY` for the loopback hop, and since 1.0.8x ships as a native binary inside a per-platform npm package. Each of those defeated the native lane silently.

| Gap | What it allowed before | Fix |
|---|---|---|
| Probe scanned only the legacy `pkg` dir | npm installs launched blind (probe answered "unknown") | Follow the launched binary / npm shim (incl. Windows `.cmd`) to the platform package's `app.js`; the launched build's verdict wins over a stale installer copy |
| `copilotUrl` setting outranks the env var | A pre-existing pin bypassed Headroom while the launch banner claimed routing | Read `settings.json` and legacy `config.json`; refuse a foreign origin, note a same-origin pin |
| Loopback hop went through the corporate proxy | Opaque connection errors inside the agent | All CLI lanes add `127.0.0.1`/`localhost`/`::1` to `NO_PROXY`, seeding both spellings from their union |
| `headroom install` used the BYOK lane | Lost Auto and Claude-on-Copilot | Native lane (`COPILOT_API_URL`) for the default and `anthropic` backends; BYOK kept for translated backends or explicit keys |
| Segmented plan hosts always normalized | Networks on GitHub's subscription-based routing blocked the generic host with no hint | Normalization stays the default (#2441, #610); `GITHUB_COPILOT_USE_ADVERTISED_HOST=1` opts in, `--subscription`/`--native`/`wrap vscode` print the advertised host, GHE.com hosts are always honoured. (#3403 proposes flipping the default.) |
| Copilot bearers on a stock-target shared proxy | GitHub tokens were forwarded to `api.openai.com` / `api.anthropic.com` | Chat, Responses, `/v1/messages`, `/v1/models` and the generic passthrough redirect Copilot credentials to Copilot; operator gateway extra headers are withheld on the redirect |

Auto mode already worked through the proxy (concrete model id in the body, `Copilot-Session-Token` survives header stripping, `/models/session` and `/auto` fall to passthrough). `tests/test_copilot_auto_passthrough.py` pins that so it cannot regress unnoticed.

## Review follow-ups (second commit)

Three reviewers went over the diff; every finding below has a test that fails without the fix.

- **Real token shape.** Copilot API tokens are `tid=<hex>;exp=…;sku=…`, not a `tid_` prefix. The redirect and `_is_forwardable_copilot_bearer_token` now accept that shape, so an installed proxy with no Copilot seed forwards the CLI's own token.
- **No seat widening.** `ghu_` was routed but not forwardable, which would have made `apply_copilot_api_auth` substitute the operator's token for any caller presenting a garbage `ghu_` bearer. Routing is now a strict subset of the forwardable set.
- **Anthropic secrets.** `/v1/messages` redirects flag themselves and the handler withholds `anthropic_extra_headers` outright, instead of running them through the trust check where a trusted Copilot host would have leaked them.
- **Live code, real tests.** `_resolve_openai_upstream` had gone dead while its tests kept passing. The chat handler calls it again, both handlers share one redirect helper, and `tests/test_copilot_redirect_e2e.py` drives the real app (chat, responses, messages) asserting destination, token pass-through, `Copilot-Session-Token` survival, withheld secrets and model-aware system relocation.
- **Probe and env hygiene.** Stale installer bundles cannot vouch for a launched build; legacy roots come from the given environment; chunk-boundary matches are not missed; `copilot_home` follows Node's `os.homedir()`; BYOK and `--subscription` lanes drop a durable install's `COPILOT_API_URL`.

## Deliberate non-changes

- The install manifest omits `NO_PROXY` so it never overwrites a user's shell setting; the wrapper covers interactive launches.
- The VS Code WebSocket `/responses` path is not registered; Auto over that transport still falls to the generic passthrough.
- `headroom init copilot -g` still writes the BYOK variables; only `headroom install` moved to the native lane.

## Test plan

- [x] `tests/test_copilot_cli_gaps.py` (45), `tests/test_copilot_auto_passthrough.py` (55), `tests/test_copilot_redirect_e2e.py` (6): new, all passing
- [x] Copilot wrap/auth/install CLI suites, routing/handler suites: 543 passed, 5 skipped
- [x] ~2250 related handler/route/CLI tests across four chunks pass; the only failures (`tests/test_pr208_changes.py`) are a pre-existing websockets-stub ordering issue that passes when the file runs alone
- [x] `uvx ruff@0.16.3 format --check` and `check` clean (CI's pinned version)
- [x] `tests/conftest.py` isolates `COPILOT_HOME` so wrap tests never read a developer's real `~/.copilot/settings.json`

https://claude.ai/code/session_018rSdvTck9uJAzNJ8Y2M5mx

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,8 +279,9 @@ export GITHUB_COPILOT_ENTERPRISE_URL=https://ghe.example.com
 
 Business and Enterprise accounts are advertised a per-plan host
 (`api.business.githubcopilot.com` / `api.enterprise.githubcopilot.com`). Headroom
-routes them through the generic host by default and prints the advertised host
-at launch. If your network uses GitHub's subscription-based network routing,
+routes them through the generic host by default; `--subscription`, `--native`
+and `wrap vscode` print the advertised host at launch when they override it. If
+your network uses GitHub's subscription-based network routing,
 which blocks the generic host, set `GITHUB_COPILOT_USE_ADVERTISED_HOST=1` or pin
 `GITHUB_COPILOT_API_URL` to the advertised host. Data-residency hosts
 (`copilot-api.<tenant>.ghe.com`) are always honored.
--- a/README.md
+++ b/README.md
@@ -277,6 +277,19 @@ export GITHUB_COPILOT_ENTERPRISE_DOMAIN=ghe.example.com
 export GITHUB_COPILOT_ENTERPRISE_URL=https://ghe.example.com
 ```
 
+Business and Enterprise accounts are advertised a per-plan host
+(`api.business.githubcopilot.com` / `api.enterprise.githubcopilot.com`). Headroom
+routes them through the generic host by default and prints the advertised host
+at launch. If your network uses GitHub's subscription-based network routing,
+which blocks the generic host, set `GITHUB_COPILOT_USE_ADVERTISED_HOST=1` or pin
+`GITHUB_COPILOT_API_URL` to the advertised host. Data-residency hosts
+(`copilot-api.<tenant>.ghe.com`) are always honored.
+
+`headroom wrap copilot` refuses to launch when `~/.copilot/settings.json` (or
+`$COPILOT_HOME`) pins a different `copilotUrl`: the CLI ranks that setting above
+`COPILOT_API_URL`, so the session would silently bypass the proxy. The wrapper
+also exempts `127.0.0.1` from any `HTTP_PROXY`/`HTTPS_PROXY` in effect.
+
 For GitHub.com Enterprise Cloud URLs such as
 `github.com/enterprises/your-enterprise`, set neither — Headroom uses GitHub's
 normal token-exchange endpoint and the Copilot API endpoint advertised for the
--- a/docs/content/docs/persistent-installs.mdx
+++ b/docs/content/docs/persistent-installs.mdx
@@ -107,6 +107,14 @@ Provider scope is intentionally conservative. The current direct adapters are:
 
 For Copilot, Aider, Cursor, and broader env-driven setups, prefer `--scope user` or `--scope system`.
 
+Copilot installs use the CLI's native lane by default: the manifest sets only
+`COPILOT_API_URL`, so the CLI keeps its model picker and Auto mode and the proxy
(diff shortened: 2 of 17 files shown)


JerrettDavis:
Live Windows validation found two blocking issues in addition to the already-visible failing test.

1. `tests/test_copilot_cli_gaps.py::test_copilot_home_honours_override_then_default` fails on Windows because the new assertion expects `HOME` to control the default Copilot config path, but `copilot_home()` correctly prefers `USERPROFILE` on Windows.

2. `wrap vscode` still is not reliable on real Windows VS Code hardware. During live validation with the installed Copilot Chat extension on VS Code 1.135.0, the Copilot Chat log recorded:
   - `[CopilotCLISession] Proxy URL configured (authType=hmac), skipping client-side token validation`
   - `[CopilotCLISession] Failed to fetch models: Error: 401 "unauthorized: unable to validate HMAC for the given Copilot-Integration-ID"`
   - later `Error: net::ERR_CONNECTION_REFUSED`
   - user-visible request failures for `gpt-5.4` (`networkError | [retry-error-panel/unknown]` / `[panel/unknown]`)

Concrete evidence from `C:\Users\jd\AppData\Roaming\Code\logs\20260903T091340\window1\exthost\GitHub.copilot-chat\GitHub Copilot Chat.log`:
- line 45: `Proxy URL configured (authType=hmac)`
- line 52: `Failed to fetch models: Error: 401 "unauthorized: unable to validate HMAC for the given Copilot-Integration-ID"`
- lines 84-85: `gpt-5.4 | networkError`

The branch does successfully route some traffic through Headroom, but the current E2E only proves request-count movement; it does not prove that the actual VS Code panel request succeeds. On my machine the wrapped VS Code flow was still capable of producing the user-facing network error.

PR #3403, opened 2026-09-02
feat(copilot): honour GitHub's advertised Business/Enterprise host by default

chopratejas:
> **Parked.** Stacked on #3402. Do not merge until a real Business or Enterprise seat confirms the segmented host serves the models it should (see "Verification needed").

## What this changes

GitHub's token exchange advertises a per-plan chat host in `endpoints.api`: `api.business.githubcopilot.com` for Business seats and `api.enterprise.githubcopilot.com` for Enterprise seats. GitHub's own clients route with that value. Since #2441 (#2455) Headroom has folded both into the hardcoded generic `api.githubcopilot.com`, because at the time the segmented hosts returned "model not supported" for newer Claude models.

That hardcoding breaks enterprises on GitHub's subscription-based network routing, whose firewall allows only the per-plan host. They cannot reach the generic host at all, and the failure looks like a random connection error.

This PR flips the default:

- Business and Enterprise advertised hosts are honoured.
- `GITHUB_COPILOT_USE_ADVERTISED_HOST=0` (also `false`/`no`/`off`) restores the generic-host fold for seats where the segmented host still lags.
- An explicit `GITHUB_COPILOT_API_URL` or enterprise-domain pin still wins over both.
- The individual host stays normalized (#610 is unchanged).
- The launch notice from #3402 now fires only when the opt-out is active.
- GHE.com data-residency hosts were already always honoured; unchanged.

## Verification needed before un-parking

On a Business or Enterprise seat, with this branch:

```bash
headroom copilot-auth login
headroom wrap copilot --native -- --model claude-sonnet-4.6 -s -p "Reply with exactly: OK"
headroom wrap copilot --subscription -- --model claude-sonnet-4.6 -s -p "Reply with exactly: OK"
```

Both should print `COPILOT_PROVIDER_API_URL=https://api.business...` (or `api.enterprise...`) and return `OK`. If either fails with "the requested model is not supported", the #2441 condition still holds and this PR should stay parked (or become opt-in again).

## Test plan

- [x] `tests/test_copilot_auth.py`, `tests/test_copilot_cli_gaps.py`, `tests/test_copilot_auto_passthrough.py`, `tests/test_cli/test_wrap_copilot.py`, `tests/test_cli/test_copilot_auth.py`, `tests/test_cli/test_wrap_persistent.py`: pass
- [x] `tests/test_cli/test_wrap_opencode.py` subscription test updated and passing
- [x] `uvx ruff@0.16.3 format --check` and `check` clean
- [ ] Live Business or Enterprise seat run (see above)

https://claude.ai/code/session_018rSdvTck9uJAzNJ8Y2M5mx

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,13 +278,12 @@ export GITHUB_COPILOT_ENTERPRISE_URL=https://ghe.example.com
 ```
 
 Business and Enterprise accounts are advertised a per-plan host
-(`api.business.githubcopilot.com` / `api.enterprise.githubcopilot.com`). Headroom
-routes them through the generic host by default; `--subscription`, `--native`
-and `wrap vscode` print the advertised host at launch when they override it. If
-your network uses GitHub's subscription-based network routing,
-which blocks the generic host, set `GITHUB_COPILOT_USE_ADVERTISED_HOST=1` or pin
-`GITHUB_COPILOT_API_URL` to the advertised host. Data-residency hosts
-(`copilot-api.<tenant>.ghe.com`) are always honored.
+(`api.business.githubcopilot.com` / `api.enterprise.githubcopilot.com`), and
+Headroom routes through it, the same as GitHub's own clients. If that host
+rejects a model your seat should have, set
+`GITHUB_COPILOT_USE_ADVERTISED_HOST=0` to route through the generic
+`api.githubcopilot.com` instead, or pin `GITHUB_COPILOT_API_URL` explicitly.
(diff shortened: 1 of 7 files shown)


JerrettDavis (approved):
Reviewed the default advertised-host policy, explicit generic-host opt-out, explicit URL override precedence, individual-host compatibility, and data-residency routing. The README and launch notice reflect the new default. All 228 focused auth/Copilot/OpenCode wrapper tests pass. No blocking finding on this head; these tests validate routing behavior and do not establish live seat/model entitlement.

PR #3016, opened 2026-08-14
fix(learn): surface Codex analysis failures

JerrettDavis:
## Description

`headroom learn` could invoke Codex CLI from a non-Git working directory without Codex’s required bypass flag. The resulting backend error was then swallowed by the analyzer and rendered as “No actionable patterns found” with exit code 0. This fixes both coupled defects so Codex can run from discovered project locations and genuine analysis failures remain visible and machine-detectable.

Closes #3008

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Added `--skip-git-repo-check` to the Codex CLI analysis backend command.
- Added an explicit `analysis_error` result field instead of conflating backend failure with an empty recommendation set.
- Kept multi-project analysis best-effort, while returning exit code 1 after any project analysis fails.
- Prevented failed analysis from printing a misleading no-pattern success message.
- Added analyzer and CLI regression coverage for the command and failure-propagation contracts.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
uv run pytest -q tests/test_learn/test_analyzer.py tests/test_cli_learn.py
102 passed in 2.34s

uv run pytest -q tests/test_learn tests/test_cli_learn.py
257 passed, 7 skipped in 3.11s

uv run mypy headroom
Success: no issues found in 520 source files

uv run ruff check <changed files>
All checks passed!
uv run ruff format --check <changed files>
5 files already formatted

uv run pytest tests scripts/tests --splits 4 --group N --tb=short -q
shard 1: 2766 passed, 140 skipped in 174.08s
shard 2: 2699 passed, 207 skipped in 60.00s
shard 3: 2822 passed, 84 skipped in 76.10s
shard 4: 2734 passed, 172 skipped in 80.29s
```

## Real Behavior Proof

- Environment: macOS arm64, Python 3.13, Codex CLI 0.147.0-compatible command surface, current `main` including #2996.
- Exact command / steps: verified `codex exec --help`; exercised `_call_cli_llm` with a captured subprocess command; invoked the Click command with a simulated Codex nonzero backend result.
[...]

PR #3404, opened 2026-09-02
fix(learn): retry LLM analysis with a smaller digest when the prompt overflows the context window

gglucass:
## Description

`headroom learn` dies with zero recommendations when the analysis prompt overflows the model's context window. Field report (desktop telemetry): a 4-session / 443-call scan failed with ``LLM analysis failed: `claude -p --output-format stream-json --verbose` failed (exit 1): Prompt is too long``, and the whole project's analysis was lost.

The root cause is that `_build_digest` budgets `_MAX_DIGEST_TOKENS` (80k) for the digest as if the prompt had the window to itself. CLI backends do not work that way: `claude -p` loads the user's own context on top - CLAUDE.md files, MCP tool definitions, skills. The digest budget cannot know how big that share is, and the 4-chars-per-token estimate also runs hot on dense tool-call lines. On installs with heavy context the combined prompt overflows a 200k window and the analysis fails outright, every scan, until the user's sessions age out.

The fix makes the budget adaptive instead of guessing a universally safe constant: when the backend reports a context overflow, rebuild the digest at half the budget and retry, down to a 10k-token floor (80k -> 40k -> 20k -> 10k, then give up). A truncated digest that produces recommendations beats a full digest that produces an error. Non-overflow failures are re-raised on the first attempt exactly as before.

## Type of Change

## Changes Made

- `SessionAnalyzer.analyze` wraps the LLM call in a shrinking-budget retry loop; overflow detection via new `_is_prompt_too_long(exc)`, which matches the known phrasings across backends: claude CLI ("Prompt is too long"), Anthropic via LiteLLM ("prompt is too long: N tokens > M maximum"), OpenAI ("context_length_exceeded" / "maximum context length"), Gemini ("input token count exceeds the maximum").
- `_build_digest` gains a `max_tokens` parameter (default `_MAX_DIGEST_TOKENS`, so all other callers are unchanged); the char budget derives from it.
- New `_MIN_DIGEST_TOKENS = 10_000` floor so a persistent overflow terminates after 4 attempts instead of looping.
- Tests: retry produces a strictly shorter second digest and succeeds; persistent overflow stops after the 10k floor (4 calls) and returns stats-only; non-overflow failures still fail on the first call with no retry.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_learn/ -q
216 passed, 3 skipped in 8.18s
```

## Real Behavior Proof

- Environment: macOS (arm64), Python 3.12 via uv, this branch's worktree, real `claude` CLI installed.
- Exact command / steps: Drove the real CLI through `_call_cli_llm` with oversized digests to verify the trigger end-to-end.
- Observed result: Could not force the overflow on this machine - the account resolves to a 1M-context model, so even a ~950k-token prompt succeeds (confirmed via the stream-json `result` event's `contextWindow: 1000000`). The failure is real on 200k-window installs: the field telemetry's stderr carries `Prompt is too long` verbatim, which is the exact string `_is_prompt_too_long` matches and the exact `RuntimeError` message the retry test injects through the real `analyze` -> `_call_llm` path.
- Not tested: a live end-to-end overflow against a 200k-window account (none at hand); the gemini/codex CLI phrasings are covered by the pattern list, not live runs.

## Runtime Rollout Safety

- Rollout-managed feature(s): None; `learn` is not rollout-gated.
- Stable/default behavior changed: Only on the previously-failing path: an overflow now retries with a smaller digest instead of aborting the analysis. Successful analyses are byte-identical (same budget, same digest).
- Kill switch / disable path: Not needed; a persistent overflow degrades to today's behavior (stats-only result, warning logged) after 4 attempts.
- Qualification impact: Heavy-context installs (large CLAUDE.md, many MCP servers) go from 100% learn failure to a truncated-digest analysis.
- Rollback path: Revert the commit; no persisted formats touched.

## Review Readiness

### DIFF
--- a/headroom/learn/analyzer.py
+++ b/headroom/learn/analyzer.py
@@ -48,6 +48,10 @@ _MODEL_DEFAULTS: list[tuple[str, str]] = [
 ]
 
 _MAX_DIGEST_TOKENS = 80_000  # Budget for the digest (leave room for prompt + output)
+# Smallest digest budget worth retrying with after a context overflow. The
+# budget halves on each "prompt is too long" failure; below this the digest
+# carries too little signal to produce useful recommendations.
+_MIN_DIGEST_TOKENS = 10_000
 
 # CLI tools to try when no API key is set (checked in order).
 # Each entry: (binary_name, model_identifier, command_prefix). The claude-cli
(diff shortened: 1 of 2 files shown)


chopratejas:
Hi @gglucass  - thanks for this fix - can you resolve the conflict?

gglucass:
Rebased on main and resolved the conflict. The retry loop now also sets `result.analysis_error` on the give-up path, so it keeps the #3016 behaviour (empty result is never reported as success). `tests/test_learn/` passes locally (250 passed).

JerrettDavis (approved):
The retry loop is bounded to four attempts, rebuilds the digest at each smaller budget, and retains terminal analysis errors and existing recommendation weighting. Non-overflow errors still stop immediately. I reviewed the digest truncation path and reran all 92 analyzer tests successfully; current reported CI is passing or skipped. No blocking finding.

gglucass:
Still open and still reproducible on current `main` (`e67b3c8a`).

Re-checked today:

- `headroom/learn/analyzer.py:50` still hardcodes `_MAX_DIGEST_TOKENS = 80_000` and `:309` still spends it as `char_budget = _MAX_DIGEST_TOKENS * 4`, with no retry on a context overflow. A `claude -p` backend loads the user's own CLAUDE.md, MCP tool definitions and skills on top of that budget, so on heavy-context installs the combined prompt overflows and `headroom learn` returns zero recommendations for the whole project, every scan.
- No drift at all: `main` has not touched `analyzer.py` since this branch forked, and the branch is 0 commits behind.
- `pytest tests/test_learn/test_analyzer.py` on the PR head: 92 passed.

The failure mode is silent from the user's side (the scan just yields nothing), which is why it went unreported for a while. Ready to merge or to rebase on request.

PR #3414, opened 2026-09-03
fix(read-maturation): book matured Read savings once, not on every replay

gglucass:
## Description

Read maturation books its savings on every request instead of once. The client re-sends the raw conversation each turn, the transform replays the recorded marker each turn, and the handler's `original_tokens - optimized_tokens` diff therefore re-counts the same matured Read's removal on every request until end of session.

Measured live on one machine the day the feature was enabled (wheel 0.37.0, Claude Code traffic): `tokens.new_input_savings_percent` went 31.8% -> 78.15% same day, same workload, with no new removal behind the jump - pure replay recounting. Chart bars fed by per-request `tokens_saved` inflated the same way (1.2M "saved" in a single hour against 2.9M sent).

This PR makes the accounting first-appearance: a matured Read's savings book exactly once, on the request that matures it. The debt is measured from the request's own endpoints - the raw client snapshot vs what is actually forwarded - and charged once, at the handler's final accounting step, tokenized with the same tokenizer as the diff so the subtraction is on the booked scale. Measuring the endpoints rather than the maturation pass's own replacements matters: after a Read matures, its marker normally reaches the wire through the cached-prefix replay, so the pass replaces nothing on exactly the turns that re-book the removal. Newly-matured replacements and fresh-tail compression are untouched. Forwarded bytes are untouched - this changes what is counted, never what is sent.

## Type of Change

## Changes Made

- `ReadMaturationManager.replayed_token_debt(original_messages, outbound_messages, count_text)`: for every matured Read that the client re-sent verbatim and that is forwarded as its marker, books the removal on first appearance (`_booked`) and charges the token delta on every request after that. Deltas are tokenized once per tool call and cached (`_replay_token_deltas`). Marker echoes and unchanged wire forms are not replays.
- `_iter_tool_results`: module-level walk over tool results in both the Anthropic block and OpenAI `role="tool"` shapes, used on both endpoints.
- Anthropic handler: charges the debt exactly once, in the final consistency recount that produces the recorded `tokens_saved` - the pre-send hook and the maturation block both recompute the plain diff before it, so any adjustment made earlier is overwritten. `tok_before`/`tok_after` stay the honest wire counts; only the booked saving is first-appearance. The call is advisory (guarded), so a failure cannot skip the recount around it.
- Tests: `TestFirstAppearanceAccounting` - first appearance books in full, later requests charged without a replacement (the cached-prefix replay path), marker echo and unchanged wire form excluded, tokenize-once caching, replay request nets zero, nothing-matured zero debt. Handler-level `test_replayed_marker_is_not_rebooked_in_the_emitted_savings` drives a 5-turn session through the real handler and asserts on the emitted `x-headroom-tokens-saved`.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_read_maturation.py tests/test_read_maturation_handler_nobust.py tests/test_cache_ttl_preserved.py
39 passed, 1 warning in 4.88s

uv run --frozen --extra dev pytest tests/ -k "maturation or anthropic"
580 passed, 34 skipped, 11868 deselected, 2 warnings in 90.43s
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), Python 3.11, uv --frozen dev env; live measurement on a desktop deployment of wheel 0.37.0 carrying this patch's arithmetic as a sitecustomize shim, Claude Code traffic.
- Exact command / steps: enabled read_maturation fleet-side for one day, sampled `GET /stats` `tokens.new_input_savings_percent` before and after on the same machine and workload. Then, in-tree, drove a 5-turn session (quiesce_turns=2) through the real Anthropic handler with a mocked upstream that reports honest cache usage, and read the emitted `x-headroom-tokens-saved` and `x-headroom-transforms` per turn, before and after the fix.
- Observed result: rate inflated 31.8% -> 78.15% the day maturation turned on with no new removal. Handler-level, before the fix: saved = [0, 0, 11863, 11863, 11863] - the Read matures on turn 2 and turns 3 and 4 book its removal again with `transforms=router:noop`, i.e. the marker reached the wire through the prefix replay and the maturation pass reported no replay at all. After the fix: saved = [0, 0, 11863, 0, 0], with `x-headroom-tokens-before` unchanged (11974 -> 12018 across the same turns). Reverting only the subtraction turns the new handler test red.
- Not tested: multi-worker deployments (session-state stickiness constraint unchanged); OpenAI-format request paths (maturation is only wired into the Anthropic handler).

## Runtime Rollout Safety

- Rollout-managed feature(s): read_maturation (accounting correction inside the existing feature; no new feature).
- Minimum rollout channel: Beta (unchanged - inherits the feature's existing gate).
- Stable/default behavior changed: No. The feature is flag-gated default-off; with it off this code never runs, and with it on only the reported `tokens_saved` changes - forwarded bytes are identical.
- Kill switch / disable path: disabling read_maturation disables this with it; the accounting has no independent switch because it must never diverge from the transform it describes.
- Qualification impact: None.
- Rollback path: revert the commit; no persisted format or API surface changes.

## Review Readiness

## Additional Notes

Headroom Desktop currently ships this arithmetic as a version-pinned sitecustomize shim against 0.37.0 (estimator-scale, with a lock-guarded pending bridge because it cannot reach the handler seam). Once a wheel ships this PR, that shim is deleted - the in-tree form is token-exact and needs no bridge, which is why the fix belongs here.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -2307,7 +2307,17 @@ class AnthropicHandlerMixin:
                             maturation.holding_msg_indices,
                         )
                         optimized_tokens = tokenizer.count_messages(optimized_messages)
-                        tokens_saved = max(0, original_tokens - optimized_tokens)
+                        # First-appearance accounting: the client re-sends
+                        # the raw conversation every turn, so this diff
+                        # re-books every replayed marker's removal on every
+                        # request until end of session. Subtract the
+                        # replayed share, tokenized on the same scale as
+                        # the diff; matured content books exactly once, on
+                        # the turn it matures.
+                        replay_debt = maturation_mgr.replayed_token_debt(
+                            maturation, tokenizer.count_text
+                        )
+                        tokens_saved = max(0, original_tokens - optimized_tokens - replay_debt)
                         if maturation.newly_matured:
                             transforms_applied.append(f"read_maturation:{maturation.newly_matured}")
                         logger.debug(
@@ -2315,6 +2325,7 @@ class AnthropicHandlerMixin:
                             f"holding={len(maturation.holding_msg_indices)} "
                             f"matured={maturation.newly_matured} "
                             f"replayed={maturation.replacements_applied} "
+                            f"replay_debt={replay_debt} "
                             f"bytes_saved={maturation.bytes_saved}"
                         )
                 except Exception as e:
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -2277,6 +2277,11 @@ class AnthropicHandlerMixin:
             # request. Session state (matured markers) rides on the
             # prefix tracker — same affinity and TTL cleanup as the
             # freeze state. Advisory: must never fail the request.
+            # Bound when maturation runs, so the final accounting step below
+            # can charge this request's replayed-marker debt. Every earlier
+            # `tokens_saved` assignment is overwritten by that recount, so the
+            # adjustment belongs there and nowhere else.
+            _maturation_mgr = None
             if self.config.read_maturation and not _bypass:
                 try:
                     from headroom.config import ReadMaturationConfig
@@ -2297,6 +2302,7 @@ class AnthropicHandlerMixin:
                             compression_store=get_compression_store(),
                         )
                         prefix_tracker.read_maturation_manager = maturation_mgr
+                    _maturation_mgr = maturation_mgr
                     maturation = maturation_mgr.apply(
                         optimized_messages,
                         frozen_message_count=frozen_message_count,
@@ -2307,17 +2313,7 @@ class AnthropicHandlerMixin:
                             maturation.holding_msg_indices,
                         )
                         optimized_tokens = tokenizer.count_messages(optimized_messages)
-                        # First-appearance accounting: the client re-sends
-                        # the raw conversation every turn, so this diff
-                        # re-books every replayed marker's removal on every
-                        # request until end of session. Subtract the
-                        # replayed share, tokenized on the same scale as
-                        # the diff; matured content books exactly once, on
-                        # the turn it matures.
-                        replay_debt = maturation_mgr.replayed_token_debt(
-                            maturation, tokenizer.count_text
-                        )
-                        tokens_saved = max(0, original_tokens - optimized_tokens - replay_debt)
+                        tokens_saved = max(0, original_tokens - optimized_tokens)
                         if maturation.newly_matured:
                             transforms_applied.append(f"read_maturation:{maturation.newly_matured}")
                         logger.debug(
@@ -2325,7 +2321,6 @@ class AnthropicHandlerMixin:
                             f"holding={len(maturation.holding_msg_indices)} "
                             f"matured={maturation.newly_matured} "
                             f"replayed={maturation.replacements_applied} "
-                            f"replay_debt={replay_debt} "
                             f"bytes_saved={maturation.bytes_saved}"
                         )
                 except Exception as e:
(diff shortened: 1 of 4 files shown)


JerrettDavis (changes requested):
[P2] Carry the replay adjustment through the final accounting step. The new subtraction sets `tokens_saved` near line 2320, but the same handler unconditionally recounts the original snapshot and final outbound messages near lines 3170–3191 and resets `tokens_saved = max(0, original_tokens - optimized_tokens)`. That discards `replay_debt` before the request outcome is recorded, so ordinary successful requests still book the replayed removal. A presend-hook recount can also reset it earlier. Please retain the first-appearance adjustment through the final accounting path, keeping wire-token totals distinct from the first-appearance metric as needed, and add a handler-level two-request regression that inspects the emitted outcome/PERF savings. The 28 focused maturation tests pass, but the new tests only exercise the manager and reproduce the subtraction outside the handler; they do not catch this overwrite.

gglucass:
Fixed in `ab810213`. You were right that the adjustment never survived to the recorded outcome - and chasing it through the handler turned up something worse: the mechanism was inert on exactly the turns it targets.

Driving a 5-turn session through the real handler (quiesce_turns=2) and reading the emitted `x-headroom-tokens-saved` per turn: **[0, 0, 11863, 11863, 11863]**. The Read matures on turn 2, and turns 3 and 4 book the same removal again with `transforms=router:noop`. On those turns `apply` returned `replacements_applied=0` and no replayed pairs at all: once a Read matures, its marker reaches the wire through the cached-prefix replay, not through the maturation pass. So carrying `replayed_pairs` through the recounts would have changed nothing in production.

The debt is now measured from the request's own endpoints - the raw client snapshot vs the final outbound messages - so it covers every path that re-removes an already-booked matured Read. `replayed_token_debt(original_messages, outbound_messages, count_text)` books each matured `tool_call_id` on its first appearance and charges its cached token delta on every request after that; marker echoes and unchanged wire forms are not replays. `MaturationResult.replayed_pairs` is gone.

It is charged once, in the final consistency recount that produces the recorded `tokens_saved`, since both the maturation block and the pre-send hook recompute the plain diff before it. `tok_before`/`tok_after` stay the honest wire counts, so only the booked figure is first-appearance. The call is guarded, so a failure in it cannot skip the recount around it.

After the fix, same session: **[0, 0, 11863, 0, 0]**, with `x-headroom-tokens-before` unchanged.

Handler-level regression added: `test_replayed_marker_is_not_rebooked_in_the_emitted_savings` drives the 5-turn session through the real handler and asserts on the emitted header, not on the manager. Reverting only the subtraction turns it red.

```
uv run --frozen --extra dev pytest tests/ -k "maturation or anthropic"
580 passed, 34 skipped, 11868 deselected, 2 warnings in 90.43s
```

JerrettDavis (approved):
The replay adjustment now runs at the final accounting recount, resolving my earlier overwrite finding while leaving wire token counts separate from booked first-appearance savings. The new handler-level regression exercises cumulative requests and checks the emitted tokens-saved header after maturation and replay. All 34 focused maturation/handler tests pass. No remaining blocking finding on this head.

gglucass:
Still open and still relevant on current `main` (`e67b3c8a`).

Re-checked today:

- `main` has no first-appearance booking: no such logic in `headroom/transforms/read_maturation.py` or `headroom/proxy/handlers/anthropic.py`, so a matured Read's removal is still re-counted on every subsequent request in the session.
- No drift: `main` has not touched either file in the 9 commits since this branch forked.
- `pytest tests/test_read_maturation.py tests/test_read_maturation_handler_nobust.py` on the PR head: 34 passed.

Extra context for prioritising it: we are carrying this fix as a downstream patch in Headroom Desktop (shipped in v0.9.11) because the inflated per-request `tokens_saved` makes the savings chart and `new_input_savings_percent` unusable without it. That patch exists only to bridge until this lands, and it is pinned to a specific wheel version, so it goes inert on the next wheel bump. Merging this upstream lets us delete it.

PR #3419, opened 2026-09-03
Fix lossless compaction folding timestamped logs, misleading savings headline, and silent HEADROOM_MODEL_LIMITS no-op

Morteza-Rastgoo:
Three independent fixes found while evaluating Headroom as a compaction layer in front of a self-hosted vLLM gateway serving coding-agent traffic.

### 1. `lossless_compaction`: never fold timestamped log lines as grep rows

The grep-row folder matched `journalctl`/syslog-style lines (`Aug 16 11:03:22 host unit[pid]: msg`) because they contain colons and look like `file:line:match`. Folding them is **not lossless**: the timestamps are the information, and the model has to reconstruct the shape itself, which costs output tokens. Added a guard plus `tests/test_lossless_timestamp_guard.py`.

### 2. `proxy/cost.py`: report size-weighted savings, not mean-of-ratios

The headline averaged per-request ratios, so one tiny highly-compressible request (6,070 -> 83 tokens, 98.6%) could dominate the number. Users read that figure as "share of total tokens removed". Now weighted by total tokens.

### 3. `providers/anthropic.py`: warn when `HEADROOM_MODEL_LIMITS` has no effect

A `HEADROOM_MODEL_LIMITS` value that is valid JSON but has neither a `context_limits` nor a `pricing` key was silently ignored -- the config appeared applied but every model kept its default window. Now logs a warning naming the accepted keys.

### Testing

`pytest tests/test_lossless_timestamp_guard.py tests/test_provider_model_fallback.py -q` -> **54 passed**.

Happy to split these into separate PRs if you prefer.

### DIFF
--- a/docs/content/docs/configuration.mdx
+++ b/docs/content/docs/configuration.mdx
@@ -334,6 +334,7 @@ headroom proxy --learn --min-evidence 3
 | `HEADROOM_TELEMETRY` | Set to `on` for **local-only** usage stats (powers your own `/stats` and dashboard; nothing is sent externally) | `off` |
 | `HEADROOM_STATELESS` | Set to `true` to disable filesystem writes | `false` |
 | `HEADROOM_MODEL_LIMITS` | Custom model config (JSON string or file path) | -- |
+| `HEADROOM_LOSSLESS_COMPACTION` | Set to `0` to disable the byte-reversible lossless folds (grep heading, log run-collapse, diff/config folds) while leaving the rest of the compression pipeline active. Read per request, so the proxy's `/admin/runtime-env` hot-sync applies without a restart. | `1` |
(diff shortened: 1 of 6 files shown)


chopratejas:
Reviewed and merged, thanks for the careful writeups. All three fixes reproduced on main; the timestamp guard rejected zero rows across 5,921 real grep lines from this repo.

Two follow-ups landed on top of your commits before merging:
- The no-effect warning fired for `{"openai": {"context_limits": ...}}`, which is the documented shape for the OpenAI loader. It now stays quiet when another provider's section carries the consumed keys, with tests for both cases.
- ruff 0.16.3 (CI's pinned version) reformatted two lines, and `HEADROOM_LOSSLESS_COMPACTION` is now in the env-var table in the docs.

The two failing docker e2e jobs are unrelated: npm 10.9.4 crashes with `Cannot read properties of null (reading 'edgesOut')` installing the OpenClaw plugin on unmodified main too. We'll fix that separately.

PR #3424, opened 2026-09-04
fix(sdk): enforce dormant public type tests

dex0shubham:
## Description

The TypeScript SDK's public type assertions were not enforced: Vitest treated `expectTypeOf` calls as runtime no-ops, and the source tsconfig excluded the tests. This enables Vitest type checking with a dedicated tsconfig, renames the type test to `types.test-d.ts`, and corrects the stale `compress` options assertion to include `tokenBudget`.

`CompressResult.messages` remains `any[]`: the universal API returns the input format, which can be Anthropic, Gemini, or Vercel rather than OpenAI. The final change does not narrow that public result contract.

## Type of Change

- [x] Test infrastructure / bug fix

## Testing

Latest local validation on 159a646fb: 294 tests passed, 33 skipped, no type-test errors. `npm run typecheck` and `npm run build` pass.

## Real Behavior Proof

- Environment: macOS, Node, isolated branch dependency installation in `sdk/typescript`.
- Exact command / steps: `npm ci --ignore-scripts`; `npm test`; `npm run typecheck`; `npm run build`.
- Observed result: Vitest executes the type-test file and reports no type errors; runtime tests pass; TypeScript and ESM/CJS/declaration builds succeed. The author also reported a failing negative assertion probe in the prior description.
- Not tested: live provider requests; this change affects test enforcement. GitHub CI is pending on the current head.

## Runtime Rollout Safety

- Rollout-managed feature(s): None; test enforcement only.
- Stable/default behavior changed: No runtime behavior or public result-type narrowing.
- Kill switch / disable path: Revert the test configuration change.
- Qualification impact: Type assertions now participate in `npm test`.
- Rollback path: Revert this PR; no persisted-state changes.

## Review Readiness

- [x] Source review finding resolved.
- [x] Local SDK checks pass.
- Final human review is still required.

## Checklist

- [x] Tests cover the enforced public type assertions.
- [x] No manual CHANGELOG edit.


JerrettDavis (changes requested):
Enabling the dormant type checks is useful, but the new shared result type misrepresents the public universal compress() API. src/compress.ts promises to return the input format and calls fromOpenAI(result.messages, inputFormat) before returning CompressResult. For Gemini input such as [{role: "model", parts: [{text: "hello"}]}], the returned messages still have role "model" and parts, yet this change declares them OpenAIMessage[] (whose roles/content contract is different). Anthropic tool_use/tool_result and Vercel parts have the same mismatch. The any[] return from fromOpenAI hides this from tsc. Please separate the direct OpenAI client result from the universal result, or make the universal result format-aware, and add a type regression for at least a non-OpenAI input/result. Tightening the direct client contract should not silently promise OpenAI output from the format-preserving function. Local type checks report no errors; the SDK run had 293 passing runtime tests and one Vercel specification-version failure with the shared local dependency installation, which I am not attributing to this diff.

dex0shubham:
You're right — the universal `compress()` returns messages in the *input* format (`src/compress.ts` calls `fromOpenAI(result.messages, inputFormat)`), so for Gemini `model`/`parts`, Anthropic `tool_use`/`tool_result`, and Vercel parts the result is not OpenAI-shaped. Declaring `CompressResult.messages` as `OpenAIMessage[]` misrepresented that, and the `any[]` return from `fromOpenAI` hid it from tsc.

Reverted that tightening: `CompressResult.messages` stays `any[]`, with a comment (on the field and the type-test assertion) citing the format-preserving return so the reason is recorded rather than re-discovered. The PR is now scoped to what it should be — enabling the dormant `expectTypeOf` checks (via `typecheck.enabled` plus a dedicated `tsconfig.typecheck.json`, without which the checks silently no-op against the `test/`-excluded root tsconfig) and fixing the one genuinely-stale assertion (`compress` params was missing `tokenBudget`).

Separating the direct-OpenAI-client result from the format-preserving universal result (so the direct path can promise `OpenAIMessage[]` and the universal path stays format-aware) is a real improvement, but it's a public-API design change — happy to take it as a follow-up rather than fold it into this enforcement PR.

`npm test` 294 passed, `Type Errors: no errors`, `tsc --noEmit` and build clean.

JerrettDavis (approved):
The incorrect universal CompressResult.messages narrowing has been reverted while retaining enforcement of the previously dormant type tests. That resolves the format-preserving API finding and removes the downstream OpenClaw type incompatibility. Reviewed the final source change; SDK validation passes with 294 tests, 33 skips, no type-test errors, a clean tsc check, and a successful ESM/CJS/declaration build. CI is pending.

JerrettDavis (approved):
Refreshed from current main and resolved the shared Dockerfile drift to retain the locked CLI installer and npm pin. Compared the resulting feature patch with the previously approved version: it is unchanged. The approved SDK implementation and tests are byte-equivalent as a feature patch; the prior SDK test/typecheck evidence remains applicable, and I have not relabelled it as a new run. No new blocking source finding; hosted checks on this maintenance head are tracked separately.

dex0shubham:
Checking in: approved on 2026-09-08 and re-approved on 2026-09-10 after the merge-from-main refresh. The governance label had gone stale after the description was condensed; I restored the template sections on 2026-09-24 and the bot has re-marked it ready for review. No conflicts with `main`, all checks on the current head (`4fbfc2f`) are green. Anything else needed before merge?

PR #3333, opened 2026-08-28
chore: release 0.38.0

JerrettDavis:
:robot: I have created a release *beep* *boop*
---

### Features

* **beacon:** schema v2 — routing/compression training signal, gzip transport ([#3253](https://github.com/headroomlabs-ai/headroom/issues/3253)) ([67eb910](https://github.com/headroomlabs-ai/headroom/commit/67eb910e38b9879bb0bcbacdc36db69e36901075))
* **cache:** add the headroom-cache-ttl offline TTL estimator the docs reference ([#2670](https://github.com/headroomlabs-ai/headroom/issues/2670)) ([52c7aa2](https://github.com/headroomlabs-ai/headroom/commit/52c7aa2e6faa9ff5010c84d73fa35936288bbe8a))
* **dashboard:** Cost Saved headline card; label per-row savings as message-only ([#3353](https://github.com/headroomlabs-ai/headroom/issues/3353)) ([17b0412](https://github.com/headroomlabs-ai/headroom/commit/17b041235480654ee69d775e38f2f09ffe468e1d))
* **dashboard:** show routing savings for this session and all time ([#3362](https://github.com/headroomlabs-ai/headroom/issues/3362)) ([a811984](https://github.com/headroomlabs-ai/headroom/commit/a811984ffb62a866274dda2928f78922101c824b))
* **proxy/model-router:** add require_tools route condition ([#2362](https://github.com/headroomlabs-ai/headroom/issues/2362)) ([ae75ca4](https://github.com/headroomlabs-ai/headroom/commit/ae75ca49a0af04ae03903a4136483f289962dc22))
* **proxy:** accept the OpenAI Responses shape on the gateway turn ([#3661](https://github.com/headroomlabs-ai/headroom/issues/3661)) ([be00a79](https://github.com/headroomlabs-ai/headroom/commit/be00a798c34495e9f84cfbe7c76c02ad20ede978))
* **proxy:** full-savings cost card, cache-aware tool-schema pricing, routing-stats seam ([#3351](https://github.com/headroomlabs-ai/headroom/issues/3351)) ([b8fa3ad](https://github.com/headroomlabs-ai/headroom/commit/b8fa3ad9bdf77061fe808578fba302941ab4e51e))
[...]

PR #3438, opened 2026-09-05
fix(init): raise the ensure hook's timeout above cold-start latency

### LINKED ISSUE #3417: [BUG]  The ensure hook's 15s timeout is shorter than a cold start, so it can never succeed from cold, and `unwrap` doesn't remove it
| | |
|---|---|
| Headroom | 0.37.0 (pip, user install) |
| OS | Windows 11 Enterprise 10.0.26200 |
| Python | 3.12 |
| Privileges | standard user, **no admin**, corporate policy applied |
| Docker | not installed |

Repo note: the plugin marketplace registered by `headroom init` points at
`chopratejas/headroom`, while the public issue tracker referenced in the docs is
`headroomlabs-ai/headroom`. File against whichever is canonical.

---

### Summary

`headroom init <agent>` installs a Claude Code plugin whose hooks run
`headroom init hook ensure` with a **15-second timeout**. Starting the proxy from
cold takes **longer than 15 seconds**. So when the proxy is down, the hook is
killed before it can finish starting the proxy — which means the next invocation
is also cold, and is also killed.

The hook is registered on `PreToolUse` with matcher `Bash|PowerShell`, so this
becomes a **permanent 15-second stall on every single shell tool call, in every
project, in every session**, with no path to self-recovery.

`headroom unwrap claude` does **not** disable the plugin, so the stall persists
after running the documented undo command.

### Measured timings

Same command, same machine, `headroom init hook ensure`:

| Proxy state | Elapsed | Result under a 15s timeout |
|---|---|---|
| Cold, first ever start | **36,882 ms** | killed at 15s |
| Cold, subsequent | **15,900 ms** | killed at 15s |
| Warm | 558 / 558 / 530 ms | fine |

Even the best-case cold start (15.9s) exceeds the timeout. The hook is
**unwinnable by construction** whenever the proxy is not already running.

### Hook definitions installed

From `~/.claude/plugins/cache/headroom-marketplace/headroom/0.37.0/hooks/hooks.json`:

```json
{
  "hooks": {
    "SessionStart": [
      { "matcher": "startup|resume",
        "hooks": [{ "type": "command", "command": "headroom init hook ensure", "timeout": 15 }] }
    ],
    "PreToolUse": [
      { "matcher": "Bash|PowerShell",
        "hooks": [{ "type": "command", "command": "headroom init hook ensure", "timeout": 15 }] }
    ]
  }
}
```

### Three aggravating factors

1. **`unwrap` doesn't clear the plugin.** `headroom unwrap claude` removes the
   `settings.json` hooks, the `ANTHROPIC_BASE_URL` env override, and the MCP
   entries — but leaves `"headroom@headroom-marketplace": true` in
   `enabledPlugins` and leaves the plugin cache on disk. The plugin's own copies
   of both hooks keep firing. A user following the documented undo path is still
   stalled and has no indication why.

2. **`init` writes the hooks twice.** After `headroom init --global claude`, the
   identical `SessionStart` and `PreToolUse` hooks exist **both** in the plugin
   **and** written directly into `~/.claude/settings.json` (also `timeout: 15`).
   So the ensure command runs twice per session start and twice per shell call.
   Re-running `init` also silently flips the plugin flag back to `true`, undoing
   a user's manual disable.

3. **Routing has no fallback.** `cli/init.py:179` sets
   `env.ANTHROPIC_BASE_URL = http://127.0.0.1:<port>` in `settings.json`. If the
   proxy is not running, Claude Code fails outright with connection-refused —
   there is no direct-to-Anthropic fallback. Combined with (1), a user whose
   proxy died has a fully non-functional Claude Code and a documented undo
   command that doesn't fix it.

### Suggested fixes

- Raise the ensure-hook timeout above the real cold-start time (60s), or make
  `hook ensure` **return immediately** after spawning the proxy detached rather
  than blocking until healthy. The hook only needs to *trigger* a start.
- Drop the `PreToolUse` hook, or gate it behind a cheap liveness probe (a TCP
  connect to the port is sub-millisecond) so the expensive path runs only when
  the proxy is actually down.
- Make `unwrap` disable the plugin and remove the marketplace entry — it is
  currently the primary source of the hooks it claims to remove.
- Have `init` write the hooks in exactly one place, and not silently re-enable a
  plugin the user disabled.
- Consider failing open: if the proxy can't be reached, let the request go
  direct rather than hard-failing the session.

connectsudhindra-gif:
## Latest branch refresh

Refreshed cleanly from upstream main to incorporate its Bash classification fix, which addresses the two inherited failing Bash tests. All 86 Bash/init/plugin-manifest tests pass after the refresh. The tested npm installer pin also addresses the Docker OpenClaw installation crash. New-head CI is running.

## Description

The `headroom init <agent>` Claude Code plugin (and the equivalent hooks Headroom writes directly for Copilot and Codex) registers `headroom init hook ensure` on `SessionStart` and `PreToolUse` with a **15-second** timeout. Starting the proxy from cold measures **15.9-36.9 seconds**, and the hook's own internal `wait_ready()` call after spawning the proxy already allows up to **45 seconds**. The external 15s timeout kills the hook process well before either of those can complete.

Because `PreToolUse` matches `Bash|PowerShell`, this becomes a **permanent 15-second stall on every single shell tool call, in every project, in every session**, with no path to self-recovery — the hook can never win from a cold start.

Closes #3417

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `plugins/headroom-agent-hooks/hooks/hooks.json`: raised both `SessionStart` and `PreToolUse` hook timeouts from `15` to `60` seconds. This is the file installed into `~/.claude/plugins/cache/headroom-marketplace/...` that the issue's reproduction quotes directly.
- `headroom/cli/init.py`: added `_HOOK_ENSURE_TIMEOUT_SECONDS = 60` and used it in `_ensure_claude_hooks`, `_ensure_copilot_hooks`, and `_ensure_codex_hooks` instead of the hardcoded `15`, so the hooks Headroom writes into a user's own `settings.json` / `hooks.json` carry the same fixed timeout as the plugin-provided one.
- 60s was chosen to sit comfortably above the existing internal `wait_ready(timeout_seconds=45)` ceiling in `_ensure_profile_running`, so the host never kills the hook before that internal wait could have succeeded on its own.
- `tests/test_cli/test_init_cli.py`: added regression tests asserting each of the three hook-writing functions emits a timeout greater than the internal 45s wait.
- `tests/test_plugin_manifests.py`: added a regression test asserting every hook in the shipped `plugins/headroom-agent-hooks/hooks/hooks.json` carries a timeout greater than 45s.

Out of scope for this PR (tracked separately as other aggravating factors in #3417): `unwrap` not disabling the plugin, hooks being written in two places at once, and adding a direct-to-upstream fallback when the proxy is unreachable. Those are independent fixes with their own risk surface; this PR addresses only the timeout mismatch that makes the hook unwinnable by construction.

## Testing

- [ ] Type checking passes (`mypy headroom`) — not run; local `mypy` hits an unrelated numpy-stub/Python-version incompatibility in this environment, independent of this change

### Test Output

```text
$ python3 -m pytest tests/test_cli/test_init_cli.py tests/test_plugin_manifests.py -q
81 passed in 0.47s

$ python3 -m pytest tests/test_cli/ -q
740 passed, 1 skipped in 10.93s

$ ruff check headroom/cli/init.py tests/test_cli/test_init_cli.py tests/test_plugin_manifests.py
All checks passed!

# Manual run against a sandboxed HOME (real `claude` CLI + real headroom init claude,
# no side effects on the actual machine config):
$ HOME=/tmp/fakehome HEADROOM_MARKETPLACE_SOURCE=$(pwd) python3 -m headroom.cli init --global claude
Configured Claude Code (user scope).
$ cat /tmp/fakehome/.claude/settings.json
{
  "hooks": {
    "SessionStart": [{"hooks": [{"command": "... headroom init hook ensure ...", "timeout": 60}]}],
    "PreToolUse":   [{"hooks": [{"command": "... headroom init hook ensure ...", "timeout": 60}]}]
  },
  ...
}
```

## Real Behavior Proof

- Environment: macOS (Darwin 24.6.0), Python 3.12.3, real `claude` CLI on PATH, `HOME` sandboxed to a throwaway directory so the run could not touch the actual machine's `~/.claude` config.
- Exact command / steps: `HOME=<scratch> HEADROOM_MARKETPLACE_SOURCE=<repo> python3 -m headroom.cli init --global claude`, then inspected the written `<scratch>/.claude/settings.json`.
- Observed result: the command ran end-to-end (added the marketplace, installed the plugin, wrote hooks), and the written `settings.json` carries `"timeout": 60` on both the `SessionStart` and `PreToolUse` `headroom init hook ensure` entries, confirming the fix takes effect through the real `headroom init claude` code path, not just the unit tests. Verified afterward that the real `~/.claude/settings.json` on this machine was untouched (unchanged mtime, no `headroom-marketplace` entry).
- Not tested: an actual cold proxy start under a live Claude Code session timing out (or not) at the old vs. new limit — that requires a real multi-second proxy cold-start reproduction end-to-end inside Claude Code itself, which wasn't run. The 15.9-36.9s cold-start figures are taken from the issue's own measurements; this PR does not independently re-measure them, it only widens the external timeout to comfortably exceed the code's own internal 45s wait ceiling.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. The hook timeout is a static value written into user-owned config files (`settings.json` / `hooks.json`) by `headroom init`, and in the plugin's bundled `hooks.json` — it is not behind a feature flag or rollout channel.
- Stable/default behavior changed: Yes, for new/re-run `headroom init` invocations and for anyone picking up the updated bundled plugin: the ensure-hook timeout goes from 15s to 60s. Users who already ran `init` keep their existing `15` in their local config until they re-run `headroom init` (this is not auto-migrated).
- Kill switch / disable path: None needed / a user can hand-edit the `timeout` value in their own `settings.json` or `hooks.json` at any time.
- Qualification impact: None — this changes only how long the host waits before killing the `ensure` subprocess; it does not change the proxy's request/response behavior, pricing, or compression paths.
- Rollback path: Revert this PR; existing users' local config is unaffected until they re-run `headroom init`, and anyone who already has `60` written can hand-edit it back to `15` if needed.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A, no existing docs reference the 15s hook timeout value

## Additional Notes

The issue (#3417) also flags three other problems with the same hook system (`unwrap` not disabling the plugin, hooks being registered in two places at once, and no direct-to-upstream fallback when the proxy is unreachable). Those are left for separate PRs since each has its own risk surface and review scope; this PR fixes only the timeout mismatch that makes the hook unwinnable from a cold start.


JerrettDavis (approved):
Reviewed aecb433ce, including all three generated-hook paths and the bundled plugin hooks. The 60-second external timeout gives the normal cold-start path room for its existing 45-second readiness wait. Local validation: 81 tests passed in test_init_cli.py and test_plugin_manifests.py; changed Python files pass Ruff and formatting. No blocking code findings. CI is not fully green: shard 1 fails two Bash-language tests with tree-sitter unavailable, and Docker native fails in OpenClaw npm installation with the same edgesOut crash seen on release PR #3333. These failure paths are outside this diff; approval covers the hook change, while CI readiness remains unresolved. Existing generated hooks still require re-running init, as the description explains.

JerrettDavis (approved):
Re-reviewed the refreshed current head after the clean merge from current main. The hook-timeout diff remains scoped and semantically unchanged; no manual conflict resolution or contributor-logic edits were made. Local validation on this head: 81 init/plugin-manifest tests passed with development dependencies, Ruff passed, and formatting passed. Approved at SHA 6f1835b81; final merge readiness still depends on the new hosted matrix completing green.

PR #3442, opened 2026-09-06
fix(openclaw): declare transcriptSemantics and implement commitTurn

### LINKED ISSUE #3441: [BUG] OpenClaw plugin never declares transcriptSemantics, so OpenClaw degrades headroom to the legacy engine on every durable-commit turn
## Description

The OpenClaw plugin (`headroom-openclaw`) never declares the context-engine transcript-semantics contract that OpenClaw requires, so OpenClaw falls back to its built-in `legacy` engine for every turn that requires a durable commit. Headroom stays `status=loaded` and stays selected in `plugins.slots.contextEngine` the whole time, so every health surface reads green while compression is doing nothing.

We ran this in production on two independent gateway instances for a week. One saved ~210K tokens in a 24h window and then saved **nothing at all for the following 55 hours**, with no configuration change of any kind. The other has never compressed once in 7 days.

## Environment

- OpenClaw `2026.8.1 (ea80657)`, HTTP mode, Linux/arm64
- `headroom-openclaw` `0.37.0` (`.plugin-ref` = `v0.37.0`), loaded via `plugins.load.paths`
- `plugins.slots.contextEngine: "headroom"`
- Two independent gateway instances, separate volumes and configs

## Actual behavior

Identical warning on both instances, once per affected logical turn:

```
[context-engine] Context engine "headroom" degraded to "legacy" for this logical turn:
current-turn transcript fencing is not declared. The "legacy" engine will handle only
this turn; configuration is unchanged, and "headroom" will be retried next turn.
```

Prometheus, 24h increase, both instances:

| metric | instance A | instance B |
| --- | --- | --- |
| `headroom_tokens_saved_total` | 0 | 0 |
| `headroom_tokens_input_total` | 0 | 0 |
| `headroom_requests_total` | 0 | 0 |
| `headroom_inbound_requests_total` | 11,488 | 11,522 |

The inbound counter is health probes only — the proxy is up and reachable, it simply never receives a request, because the plugin never gets far enough to route one.

## Root cause

OpenClaw gates context-engine selection in `context-engine-turn-attempt-*.js`:

```js
const resolveSelectionIssue = (selection) => {
  const support = evaluateContextEngineHostSupport({ ... });
  if (!support.ok) return `host "${selection.host.id}" is missing ${support.missingCapabilities.join(", ")}`;
  if (isBaselineEngineSelection) return;
  if (selection.requiresDurableCommit
      && effective.engine.info.transcriptSemantics?.currentTurnFence !== "before-current-turn-entry-v1")
    return "current-turn transcript fencing is not declared";
  if (selection.requiresDurableCommit
      && !supportsContextEngineDurableTurnAdvancement(effective.engine))
    return "atomic idempotent turn advancement is not declared";
};
```

and in `host-compat-*.js`:

```js
function supportsContextEngineDurableTurnAdvancement(engine) {
  return engine.info.transcriptSemantics?.turnAdvancementIdempotency === "atomic-idempotent-v1"
      && typeof engine.commitTurn === "function";
}
```

So a context engine must declare three things to survive a durable-commit turn:

1. `info.transcriptSemantics.currentTurnFence === "before-current-turn-entry-v1"`
2. `info.transcriptSemantics.turnAdvancementIdempotency === "atomic-idempotent-v1"`
3. an `engine.commitTurn` function

`headroom-openclaw@0.37.0` declares none of them. Grepping the shipped bundle (`dist/index.js`, 40,377 bytes) for each string:

```
currentTurnFence                 0
transcriptSemantics              0
turnAdvancementIdempotency       0
commitTurn                       0
before-current-turn-entry-v1     0
```

Zero hits for all five, and zero hits across the repository. The engine is registered with `api.registerContextEngine("headroom", () => engine)` and its `info` carries no `transcriptSemantics` at all.

## Why this is easy to miss

Three things conspire to make it look healthy:

- The plugin loads, the slot resolves, and `plugins list` reports `status=loaded`. Every configuration check passes.
- The gate only applies when `selection.requiresDurableCommit` is true, so compression **still happens on other turns**. The result is not a clean failure — it is a silent, workload-dependent partial outage. Our 210K-then-zero swing was exactly this.
- The warning is emitted once per logical turn at warn level, so on a busy gateway it is easily lost, and no metric reflects it. `headroom_tokens_saved_total` simply stops increasing.

## Expected behavior

Headroom should either declare the contract and participate in durable-commit turns, or — if the durable-commit semantics genuinely cannot be honored — document that it will be bypassed for those turns, and ideally surface it as a metric so operators can see the degrade without reading gateway logs.

## Suggested fix

Declare the semantics on the engine's `info` and implement the commit hook:

```js
const engine = {
  info: {
    transcriptSemantics: {
      currentTurnFence: "before-current-turn-entry-v1",
      turnAdvancementIdempotency: "atomic-idempotent-v1",
    },
    // ...existing fields
  },
  async commitTurn(/* ... */) { /* ... */ },
  // ...existing methods
};
```

gingeekrishna:
## Description

Adds OpenClaw transcript-semantics declarations and an accepted-turn log for admitted-turn advancement. Fixes #3441.

Messages and advancement keys are written together in `HEADROOM_WORKSPACE_DIR/openclaw/commit-log.json` (default workspace: `~/.headroom`). Each commit re-reads the log inside an exclusive sibling-file lock and replaces the log atomically. Keys are retained without eviction; prior key-only logs are recognized for migration. Failed writes do not create an in-memory phantom duplicate.

## Locking and recovery

The lock serializes processes sharing the workspace and is never reclaimed merely because it is old. A waiting caller times out after 60 seconds instead of committing without ownership. Release checks the acquisition token.

A crashed holder can leave an orphan `.lock` file. Recovery is operational: stop writers using that workspace, establish that the holder is no longer running, remove the orphan lock, then resume. Do not delete a live writer's lock. Automatic crash recovery and power-loss durability are not claimed.

## Testing

- 99 plugin tests pass, including distinct-key and same-key child-process commits and an aged live-lock regression.
- TypeScript checking passes.
- The aged-lock regression fails against the prior age-reclaim implementation and passes against the current source. Test cleanup releases the child even on assertion failure.
- Current-head CI is pending.

## Real Behavior Proof

- Environment: macOS, two real Node child processes importing `DurableAdvancementKeyStore`.
- Exact command / steps: A acquires the lock and reads the empty log; pause A while alive for 31 seconds. Start B committing another key and verify it remains pending, then resume A.
- Observed result: B does not finish while A holds the lock. After release, both return `committed` and both message payloads remain in the log. The prior implementation acknowledged both but lost one turn.
- Not tested: automatic recovery from crashed holders (requires the operator procedure above); power-loss durability.

## Runtime Rollout Safety

- Rollout-managed feature(s): OpenClaw transcript-semantics declaration.
- Minimum rollout channel: Host capability negotiation.
- Stable/default behavior changed: Accepted messages and keys persist in a workspace-shared log. The log grows without eviction, and an orphan lock requires operator recovery.
- Kill switch / disable path: Revert the plugin change or use the host's legacy context-engine path.
- Qualification impact: Admission uses atomic/idempotent commit behavior; contention fails closed.
- Rollback path: Revert the plugin change. Preserve the log if retry history must remain available.

## Review Readiness

- [x] Prior message persistence, retry, cross-process, and stale-lock findings resolved.
- [x] Local tests and type checking pass.
- Final human review is required.


JerrettDavis (changes requested):
[P1] Implement the durable commit before advertising atomic-idempotent transcript support. `commitTurn` currently ignores `messages` and only inserts the key into an in-memory Set. A fresh engine loses every committed key, and after 513 commits the oldest key is deliberately accepted again; the new eviction test confirms this. That does not meet the host contract this declaration opts into: accepted messages and the advancement key must be committed atomically, and retries must return duplicate after a committed write (including restart/recovery). See the pinned upstream contract: https://github.com/openclaw/openclaw/blob/acbff80d68fce81afc695684b0a957661b9ee05a/docs/concepts/context-engine.md . Please implement durable accepted-turn/key storage, or use a host-supported stateless integration without claiming this guarantee. Add restart and replay coverage, including a previously committed key beyond the current 512-entry window, before enabling these semantics. The existing tests verify Set behavior rather than durable turn advancement.

gingeekrishna:
@JerrettDavis Fair — an in-memory Set (plus the 512-entry FIFO cap on top) doesn't meet the contract, and my own eviction test was proving the violation rather than covering it. Fixed properly:

\`commitTurn\` now goes through \`DurableAdvancementKeyStore\`, a small on-disk JSON store that:
- Persists to \`\$HEADROOM_WORKSPACE_DIR/openclaw/commit-log.json\` (or \`~/.headroom/openclaw/commit-log.json\`), overridable via a new \`HeadroomEngineConfig.commitLogPath\`.
- Writes atomically (temp file + rename) so a crash mid-write can't corrupt the log.
- Serializes concurrent commits of the same key through an in-process queue so a has-then-record race can't double-commit.
- **Never evicts a key** — no cap, no age-based pruning. A retry can arrive at any point after a restart or a long stall, so any eviction policy risks the exact failure this store exists to prevent.
- Fails loudly on a corrupt log file instead of silently treating it as empty — swallowing that would just be a different way to silently forget committed keys.

New tests cover exactly what you asked for: restart persistence (two separate engine instances constructed against the same commit-log path, simulating a process restart), a 600-key retention case proving key 0 is still reported as duplicate (no cap-driven eviction), concurrent same-key commits, and missing/corrupt log files — at both the engine level (\`engine.test.ts\`) and the store level directly (\`advancement-key-store.test.ts\`). 92/92 tests pass.

JerrettDavis (changes requested):
The durable file fixes ordinary restart/eviction behavior, and all 27 focused engine/store tests pass. The atomic commit contract still needs these corrections:

1. [P1] commitTurn still discards params.messages. The pinned host contract requires the accepted turn and key to advance together; persisting only keys does not resolve that part of the earlier finding.
2. [P1] Publish the in-memory key only after persistence succeeds (or roll it back). I loaded an empty store, made its parent unwritable by replacing the parent path with a file, then committed: the write failed with EEXIST, but retrying the same key returned duplicate despite nothing reaching disk.
3. [P1] Serialize across store instances sharing the default file, not only within one object. After two stores both load an empty path, committing a through one and b through the other leaves only ["b"] on disk. A restart then forgets the previously acknowledged a.

Please use a transactional shared store (or an equivalent cross-instance atomic protocol) for accepted messages and their key, with tests for failed-write retry and interleaved instances in addition to clean restarts. Host contract: https://github.com/openclaw/openclaw/blob/acbff80d68fce81afc695684b0a957661b9ee05a/docs/concepts/context-engine.md .

gingeekrishna:
@JerrettDavis All three fixed, and I reproduced each one on the prior commit before changing anything (matching your repro steps as closely as I could):

1. **\`messages\` discarded** — \`commitTurn\` now stores \`{messages, committedAt}\` per key instead of just the key, so the accepted turn and its key advance together in the same atomic write.

2. **Phantom duplicate after a failed write** — confirmed your exact repro: block the parent directory with a file, commit, get an error, retry the same key, get \`"duplicate"\` back despite nothing on disk. Root cause was mutating the in-memory \`Set\` before the write attempted. Fixed by removing the in-memory cache entirely — every \`tryCommit\` now re-reads the file fresh and only "sees" a key once the write recording it has actually succeeded. A failed write now rejects (not a false committed/duplicate), and retrying re-attempts the commit.

3. **Cross-instance race** — confirmed this one too: two store instances against the same file, commit \`"a"\` through one and \`"b"\` through the other, and only \`"b"\` survived on disk. The per-instance queue only serialized calls on the *same object*; each instance held its own stale in-memory copy. Fixed with a module-level queue keyed by the resolved absolute path, shared by every instance pointed at that file — combined with the no-cache design from #2, which also removes the stale-copy problem at its root.

New tests reproduce all three failures against the pre-fix code before verifying the fix passes, at both the store level and the engine level (messages persisted end-to-end). One tradeoff worth flagging: every commit now re-persists the full log rather than appending in memory, so it's O(n) per write instead of O(1) — the real cost of paying for durability + cross-instance correctness on every commit, not something I've optimized away. Given commits happen once per accepted turn (not a hot loop), this should be fine in practice, but flagging it in case it matters for very long-lived sessions. 96/96 tests pass.

JerrettDavis (changes requested):
The update resolves the discarded-message and failed-write/phantom-duplicate findings, and now serializes separate store instances within one module instance. The shared-file transaction still needs an inter-process boundary. pathQueues is a JavaScript Map, but the default commit-log path is shared by every gateway process using the same Headroom workspace. Two processes can each read the same log, add a different key, rename their snapshots, and both acknowledge success while the last rename discards the first accepted turn.

I reproduced this at a42c8cb827d0a671d5ffa0c39951a231cd2af8fd using two actual Node child processes and the real DurableAdvancementKeyStore. A barrier after each process read the initial {} file makes the ordinary read/write interleaving deterministic: both a and b returned committed, but the final log contained only b and its messages. Restart/replay therefore forgets the acknowledged a.

Please serialize the read/check/write transaction across processes sharing the file (for example, a transactional store or an inter-process file lock with defined recovery), or enforce exclusive ownership of that storage path before accepting commits. Add a child-process regression for distinct keys and for two commits of the same key. A module-local promise queue cannot establish the advertised atomic-idempotent contract for this shared default location.

gingeekrishna:
@JerrettDavis Confirmed and fixed — reproduced your exact repro (two real Node child processes racing the real store) before changing anything: distinct keys, one silently lost on disk; same key from both, both wrongly returned "committed" instead of one committed + one duplicate.

Replaced the in-process promise queue with a real inter-process lock: an exclusive lock file (\`<path>.lock\`) acquired via atomic exclusive creation (\`open(path, "wx")\`, atomic on both POSIX and Windows), held for the full read/check/write transaction. A lock older than 30s is treated as belonging to a crashed holder and broken, so a dead process can't wedge every future commit to the file.

While adapting your \`has()\`-rejects test to the new code, I also caught a related platform gap: on Windows, \`readFile\` through a path blocked by a non-directory component reports \`ENOENT\` — indistinguishable by error code alone from "no commit log yet" — so the store was silently treating a genuinely broken storage path as an empty one. Fixed by checking the parent directly rather than trusting the error code alone.

New tests spawn the real \`DurableAdvancementKeyStore\` in real child processes (via Node's native TS support, not a build step, and not Worker threads, which share memory and wouldn't exercise this at all) synchronized through a shared "go" file so they contend for the lock as close to simultaneously as possible: distinct keys from two processes (both must survive), and the same key from two processes (exactly one must win). Both confirmed to reproduce your exact symptoms against the pre-fix code before verifying the fix resolves them. 98/98 tests pass.

JerrettDavis (changes requested):
The exclusive file lock fixes ordinary cross-process contention, and all 98 plugin tests plus TypeScript checking pass. There is still a data-loss path in stale-lock recovery: age alone does not establish that the holder died. I reproduced this with two real Node processes against the current source: A acquires the lock and reads the log, then remains alive but pauses for 31 seconds; B deletes the aged lock, commits turn b, and returns committed; A resumes, writes its earlier snapshot, and also returns committed. The final log contains only a, so an acknowledged turn is lost.

Please use locking/recovery that cannot admit a second writer while the first can still commit (for example an OS-managed lock or transactional store, or fail closed until exclusive ownership is established). Also ensure an old holder cannot unlink a replacement holder’s lock in finally. Add a deterministic regression with a live holder paused past the recovery threshold, alongside the new distinct-key/same-key child-process tests. The current test coverage is a useful improvement; it needs to cover this recovery path before claiming atomic-idempotent-v1.

gingeekrishna:
@JerrettDavis Confirmed — your repro is exactly right, and I could see it by inspection once you named the pattern: age is not liveness, and my staleness guess let a genuinely-live holder get its lock stolen mid-transaction, plus the subsequent \`finally\` could unlink whatever lock file happened to be there by then, not necessarily its own.

Taking your explicit fallback since Node has no built-in binding for a real OS-mediated lock (flock/LockFileEx, where the kernel — not a timestamp — is the source of truth on holder liveness): **fails closed now**. The age-based takeover is removed entirely. A lock is held until its owner releases it, however long that takes; a caller that can't acquire it within the timeout gets a clear error instead of proceeding unsafely. Release also now verifies a unique per-acquisition token before unlinking, so a process can only ever remove the lock file it created — never a later holder's.

The trade is explicit and operational: a lock orphaned by a truly crashed process now needs a human to remove \`<path>.lock\` before commits resume, rather than self-healing. That's the cost of the guarantee you're holding this to.

New deterministic regression (no real 30+s sleep — it doesn't depend on the threshold's value at all, since there isn't one anymore): a worker occupies the real lock file and only releases it on signal; a concurrent \`tryCommit\` against the real store must not settle while it's held, and must succeed immediately once released. 99/99 tests pass.

JerrettDavis (approved):
The stale-lock data-loss finding is resolved: contenders no longer reclaim an aged lock and fail closed on timeout. I repeated the real two-process probe with A paused for 31 seconds: B remained pending until A resumed, then both commits and payloads persisted. All 99 plugin tests and TypeScript checking pass. I strengthened the live-holder regression to age its lock past the former threshold and release its child in finally; verified that it fails against the old implementation. The remaining operational limitation is explicit: a crashed holder can leave an orphan lock that must be removed only after writers are stopped and ownership is checked. CI is pending; final human review remains required.

gingeekrishna:
> The stale-lock data-loss finding is resolved: contenders no longer reclaim an aged lock and fail closed on timeout. I repeated the real two-process probe with A paused for 31 seconds: B remained pending until A resumed, then both commits and payloads persisted. All 99 plugin tests and TypeScript checking pass. I strengthened the live-holder regression to age its lock past the former threshold and release its child in finally; verified that it fails against the old implementation. The remaining operational limitation is explicit: a crashed holder can leave an orphan lock that must be removed only after writers are stopped and ownership is checked. CI is pending; final human review remains required.

Thank you @JerrettDavis

JerrettDavis (approved):
Reviewed the main-conflict resolution. Kept transcript semantics and commitTurn while preserving main’s delegated-compaction ownership and imports. All 37 engine/key-store tests and TypeScript checking pass. The prior feature intent is preserved; hosted checks for the maintenance head are tracked separately.

jedi-tal:
Hi @gingeekrishna - I filed #3441. From the outside this looks ready: CI green, approved on the 10th, no conflicts. Anything left before it can merge, or is it just waiting for a slot?

Also, is a tag likely to follow soon after? We build the plugin from a release tag, so merge alone doesn't reach us.

Happy to help either way - we run headroom on the two production OpenClaw gateways from the issue, so once there's a tag carrying this I can report back real before/after numbers on whether the degrade line actually stops.

Thanks for the fast turnaround on this one.
Tal

gingeekrishna:
@jedi-tal Thanks for checking in, and for filing #3441 with such a clear repro in the first place — that's what made this fixable at all.

Status is accurate on your end: approved (Sept 10), CI clean, no conflicts. From my side there's nothing outstanding on the code — it went through several rounds with @JerrettDavis (thank you again for the thoroughness there, especially catching the cross-process and stale-lock data-loss paths) and the current head has all of that addressed.

I'm an external contributor here (fork + PR), not a maintainer, so I don't have merge or release/tag permissions on this repo — merging it and deciding when it ships in a tag is genuinely up to the headroomlabs-ai maintainers, not something I can act on or give you a timeline for. Tagging you @JerrettDavis in case you have visibility into the merge queue.

@jedi-tal once it's tagged, I'd genuinely appreciate you reporting back real before/after numbers from your two gateways — that's exactly the kind of production signal that's hard to get from unit tests alone.

jedi-tal:
Hi @JerrettDavis,
Any plans to release this in the near future?

PR #3443, opened 2026-09-06
fix(proxy): repair client-side tool_search tool_reference blocks

gglucass:
## Summary

`strip_unsupported_tool_search_blocks` (added in #2805/#2807 to stop poisoned
Claude Code transcripts from 400ing) only scans the **server-side** tool-search
shape: a `tool_search_tool_result` block with nested `tool_references`. It never
scans the **client-side** shape documented under [Custom tool search
implementation](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool#custom-tool-search-implementation):
a plain `tool_result` whose `content` is a list of `tool_reference` blocks. That
is the shape Claude Code persists when it discovers MCP tools client-side
(e.g. `mcp__<server>__<tool>`).

Anthropic validates references in **both** shapes. When a referenced tool goes
absent between turns (an MCP server that dropped mid-session, or a side-request
carrying a smaller tools array), the request 400s with `Tool reference 'X' not
found in available tools`. The existing repair fixes the server-side block but
leaves the client-side one, so the transcript stays poisoned and every replay
400s even though the repair "fired".

## Root cause

The per-message loop keys entirely on `block.get("type") == "tool_search_tool_result"`.
A `tool_result` carrying `tool_reference` entries is never inspected.

## Fix

Extend the same absence test to client-side `tool_result` blocks: drop the
unsupportable `tool_reference` entries (a deferred-but-present tool stays valid),
and when that empties a search result, replace its content with a short
placeholder so the paired `tool_use` is not orphaned (an orphaned `tool_use`
400s on its own). Server-side handling, the paired `server_tool_use` cleanup,
and the identity-return fast path are unchanged.

## Testing

- Added: `uv run --extra dev pytest tests/test_issue_746_tool_search.py` (61 passing)

## Proof

- Repro: a client-side `tool_result` referencing an absent tool 400s on replay; the pre-fix repair leaves it untouched
- Fix: absence-keyed drop across both block shapes; present (incl. `defer_loading`) references are kept
- Tests: 3 new cases (absent dropped + tool_use preserved, present-deferred kept by identity, mixed keeps present / drops absent)
- Scope: only `tool_result` blocks that already carry `tool_reference` entries are touched; ordinary tool results are untouched

### DIFF
--- a/headroom/proxy/helpers.py
+++ b/headroom/proxy/helpers.py
@@ -3237,6 +3237,11 @@ def inject_tool_search_deferral(
 # ---------------------------------------------------------------------------
 
 _TOOL_SEARCH_RESULT_TYPE = "tool_search_tool_result"
+# A client-side tool-search result (a plain ``tool_result`` carrying
+# ``tool_reference`` blocks) that is left with no resolvable references keeps
+# this text as its content, so its paired ``tool_use`` stays valid -- dropping
+# the block outright would orphan the tool_use (which 400s on its own).
+_CLIENT_TOOL_REF_PLACEHOLDER = "[tool reference no longer available]"
 
 
 def _tool_search_reference_names(content: Any) -> list[str]:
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
[P2] Preserve valid custom-search references when no built-in server search tool is present. The new client-side branch uses `not has_search_tool` to delete references, but that flag only recognizes `tool_search_tool_*` server tools. Anthropic explicitly supports custom client-side search via a normal `tool_use` / `tool_result` and references to definitions in the top-level tools array: https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool#custom-tool-search-implementation . A built-in search tool is not required for that flow. I reproduced this with custom `ToolSearch` plus a present deferred `Calendar` tool: its valid Calendar reference is replaced by `[tool reference no longer available]`. Please validate client-side references against the available definitions without imposing the server-tool requirement, and add the no-built-in-tool case. The existing 61 tests pass, but the new positive cases all include `_SEARCH_TOOL` and miss this valid configuration.

gglucass:
Fixed in b4d9db89. Correct - `not has_search_tool` imposed a server-tool requirement that the custom client-side flow does not have, so a valid reference to a present deferred tool was replaced with the placeholder whenever no `tool_search_tool_*` was in the tools array.

The client-side branch now validates references against the available definitions only: a reference is dropped when its name is absent from the top-level tools array, and nothing else. The server-side `tool_search_tool_result` branch still requires the server tool, since that block type only exists as its output.

Added the missing configuration both ways, with no built-in search tool present: `test_repair_keeps_custom_client_side_references_without_a_builtin_search_tool` (custom `ToolSearch` plus a deferred `Calendar`, reference preserved and the transcript returned by identity) and `test_repair_still_drops_an_absent_reference_without_a_builtin_search_tool` (mixed result, `Calendar` kept, the absent one dropped).

JerrettDavis (approved):
Re-reviewed the client-side tool-reference validation. It now checks the available tool definitions without requiring a built-in server-search tool, preserving valid custom-search results while still removing absent references. All 63 tool-search tests pass, including the no-built-in-tool configuration from the earlier finding. The blocking issue is resolved.

JerrettDavis (approved):
Reviewed the refreshed tool-reference repair against current main. The conflict resolution retains client-side validation against actual tool definitions while using main's in-place server-block placeholders, preserving signed-thinking coordinates and avoiding orphaned tool results. Both client-reference regressions and the signed-thinking fingerprint regression are retained. All 64 focused tool-search tests pass; Ruff, formatting and whitespace checks pass. No new blocking finding; hosted checks are pending.

PR #3204, opened 2026-08-22
fix(proxy): write per-port proxy logs so concurrent instances stop truncating each other

rNoz:
## Description

Every proxy instance wrote the same two fixed files: `~/.headroom/logs/proxy.log` (runtime log, via the `RotatingFileHandler` in `_setup_file_logging`) and `~/.headroom/logs/proxy-stdio.log` (subprocess stdout/stderr from `_start_proxy`). The name is a constant (`paths.py` `_PROXY_LOG_FILE = "proxy.log"`), so N concurrent proxies on different ports (wrapped copilot, droid, claude; plus any persistent `install agent` deployment) run N independent rotating handlers over the *same* file and truncate/rename each other's history.

This fix keys both files by port (`proxy-<port>.log`, `proxy-stdio-<port>.log`) so each proxy rotates only its own file. `headroom perf` aggregates every per-port file and still reads a lone legacy `proxy.log` when that is all that exists (backward compat); the `proxy-stdio-*` captures (raw stdout/stderr, not PERF records) are excluded from the parser.

This matters more now that wrap-spawned proxies self-exit after an idle grace (orphan-proxy watchdog, #3202): the process is gone, so its per-port log is the only forensic evidence and must survive an unrelated proxy's rotation.

Closes #3203

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `headroom/paths.py`: `proxy_log_path(port=None)` and new `proxy_stdio_log_path(port=None)` are the single source of truth for the filenames (`proxy-<port>.log` / `proxy-stdio-<port>.log` with a port, legacy names without). The `_PROXY_LOG_FILE` constant stays only as the readers' fallback name.
- `headroom/proxy/helpers.py`: `_setup_file_logging(port=None)` targets `proxy_log_path(port)`. The duplicate-handler guard runs BEFORE the `RotatingFileHandler` is constructed (construction opens the file), so a second `create_app` on a different port in one process cannot leave an empty stray `proxy-<port>.log` or leak an fd; a handler pointing at a different port is replaced and closed so per-port keying holds.
- `headroom/proxy/server.py`: `create_app` resolves `config` first, then calls `_setup_file_logging(config.port)` (the proxy always knows its own port).
[...]

PR #3444, opened 2026-09-06
fix(proxy): give each non-default port its own proxy log file

### LINKED ISSUE #3421: [BUG] All deployment profiles share one proxy.log and rotate it under each other
**Version**: headroom-ai 0.37.0 (pipx), macOS 27, Python 3.11.

**Setup**: five persistent-service profiles created with `headroom install apply`
(default/8788 anthropic, plus openrouter/8791, ollama/8792 (anyllm), gemini/8793,
and one plain `headroom proxy` LaunchAgent on 8795).

**What happens**: every proxy process opens the same `~/.headroom/logs/proxy.log`
with its own RotatingFileHandler (`paths.py`: `proxy_log_path()` = `workspace_dir()/logs/proxy.log`,
no profile suffix). When one process rotates, the others keep writing into the renamed
file. Observed with `lsof`: the 8788 proxy was writing to `proxy.log.4` (2.4 MB, far past
the rotation size) while `proxy.log` belonged to the openrouter profile. Banners, per-day
counts, CCR lines and error signatures from all profiles end up interleaved, and any
"read proxy.log" diagnosis looks at the wrong service.

**Expected**: one log per profile (e.g. `logs/<profile>/proxy.log` or `proxy.<profile>.log`),
or `install apply` giving each profile its own workspace by default.

**Workaround that works**: run each secondary profile with
`HEADROOM_WORKSPACE_DIR=~/.headroom/profiles/<name>`, symlink
`profiles/<name>/deploy -> ~/.headroom/deploy` (otherwise `install agent run --profile`
cannot find its manifest), and point `HEADROOM_CONFIG_DIR`, `HEADROOM_SAVINGS_PATH`,
`HEADROOM_SAVINGS_EVENTS_PATH` back to the shared files so `headroom savings` stays unified.
Verified: each process now writes its own file, nothing leaks into the main log.

gingeekrishna:
## Summary
- Every proxy process attached its own `RotatingFileHandler` to the same shared `~/.headroom/logs/proxy.log`, with no scoping by port. Anyone running multiple persistent `install apply` profiles in one workspace (default/anthropic, openrouter, ollama, gemini, etc., each bound to a different port) ends up with every process rotating the same file out from under the others — banners, per-day counts, CCR lines, and error signatures from all profiles get interleaved, and any "read proxy.log" diagnosis looks at the wrong service.
- `paths.proxy_log_path()` was the one path helper in `paths.py` without port-scoping — its siblings (`proxy_clients_dir`, `beacon_lock_path`, `proxy_start_lock_path`) are all already keyed by port. Extends it with an optional `port` argument: the default port keeps the original unsuffixed `proxy.log` (no behavior change for the common single-instance case, or for existing tooling that reads that fixed path), while any other port gets its own `proxy.<port>.log`.
- Wires `config.port` through `create_app` into `_setup_file_logging` (which now delegates to `proxy_log_path` instead of re-deriving the same path independently), and threads `port` into `headroom wrap`'s subprocess log-path display helper so the path shown to the user always matches where that instance actually writes.

Fixes #3421

Known gap, left for a follow-up: `headroom perf`'s log aggregation (`parse_log_files`) still only globs the default `proxy.log*`, so per-profile perf analysis on non-default ports isn't wired up yet — this PR fixes the file-corruption bug, not full multi-profile `perf` support.

## Release note

Persistent `install apply` profiles on different ports (e.g. multiple backends sharing one workspace) each get their own `logs/proxy.<port>.log` instead of sharing and rotating the default profile's `proxy.log`.

---

Disclosure: investigated and fixed with AI-agent assistance (Claude Code).

### DIFF
--- a/e2e/wrap/Dockerfile
+++ b/e2e/wrap/Dockerfile
@@ -84,6 +84,11 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends nodejs && \
     rm -rf /var/lib/apt/lists/*
 
+# npm 10.9.8 crashes in Arborist's peer-set resolver when installing the local
+# OpenClaw plugin (Cannot read properties of null, reading 'edgesOut'). Pin the
+# tested npm version so both wrap-e2e callers use the same working installer.
+RUN npm install -g --no-fund --no-audit npm@11.16.0
+
 WORKDIR /workspace
 
 # Bring in the prebuilt manylinux wheel from stage 1.
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -611,13 +611,18 @@ def _find_available_port(start_port: int, max_attempts: int = 100) -> int:
     raise RuntimeError(f"No available port found in range {start_port}-{end_port - 1}")
 
 
-def _get_log_path() -> Path:
-    """Get path for proxy log file."""
+def _get_log_path(port: int | None = None) -> Path:
+    """Get path for proxy log file.
+
+    Delegates the filename to `paths.proxy_log_path` so this stays in sync
+    with what the proxy subprocess (started with `--port port`) actually
+    writes to — see `_setup_file_logging` in `headroom.proxy.helpers`.
+    """
     from headroom import paths as _paths
 
-    log_dir = _paths.log_dir()
-    log_dir.mkdir(parents=True, exist_ok=True)
-    return log_dir / "proxy.log"
+    log_path = _paths.proxy_log_path(port=port)
+    log_path.parent.mkdir(parents=True, exist_ok=True)
+    return log_path
 
 
 def _get_proxy_stdio_log_path() -> Path:
@@ -695,7 +700,7 @@ def _start_proxy(
         cmd.extend(["--vertex-api-url", vertex_api_url])
 
     timeout_seconds = _resolve_wrap_proxy_timeout_seconds()
-    log_path = _get_log_path()
+    log_path = _get_log_path(port)
     stdio_log_path = _get_proxy_stdio_log_path()
     stdio_log_file = open(stdio_log_path, "a", encoding="utf-8")  # noqa: SIM115
 
--- a/headroom/paths.py
+++ b/headroom/paths.py
@@ -74,6 +74,9 @@ _SYNC_STATE_FILE = "sync_state.json"
 _BRIDGE_STATE_FILE = "bridge_state.json"
 _LOGS_DIR = "logs"
 _PROXY_LOG_FILE = "proxy.log"
+# Mirrors `ProxyConfig.port`'s default (headroom/proxy/models.py). Kept as a
+# plain literal here to avoid this low-level module importing proxy code.
+_DEFAULT_PROXY_PORT = 8787
 _DEBUG_400_DIR = "debug_400"
 _CODEX_WIRE_DEBUG_DIR = "codex_wire"
 _BIN_DIR = "bin"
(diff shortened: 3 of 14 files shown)


JerrettDavis (changes requested):
[P2] Keep performance readers working when changing the log filename. The per-port isolation is useful, but `paths.proxy_log_path(port=8791)` now writes `proxy.8791.log` while `headroom.perf.analyzer.parse_log_files()` still only reads `proxy.log*`. I reproduced this at fb68289d6: a valid PERF record in `proxy.8791.log` produces zero records; renaming the same file to `proxy.log` produces one. This affects both `headroom perf` and the `/stats` throughput calculation in `server.py`, so existing non-default-port deployments lose reporting after upgrading. I saw the known-gap note in the description; this is a regression of an existing consumer and should be addressed alongside the writer change. Please include per-port logs and their rotations in the reader (with explicit aggregation/scoping behavior) and add a regression covering both legacy and per-port filenames. No other blocking finding from the changed code reviewed.

gingeekrishna:
@JerrettDavis Good catch, thanks — that was a real regression, not just the "known gap" I'd flagged. Fixed in the latest commit:

- \`parse_log_files()\` now globs both \`proxy.log*\` and \`proxy.[0-9]*.log*\`, so \`headroom perf\` and \`/stats\` throughput aggregate every port's log again (including its rotations), with regression tests covering plain aggregation and the mtime-based windowing optimization.
- Also caught and fixed 21 CI test failures from the same root cause on the writer side: several test files monkeypatched \`_get_log_path\` with a zero-arg lambda, which broke once it started being called with a positional \`port\` argument. Updated every mock accordingly.

The two remaining red checks (\`docker-native-e2e\`, \`docker-wrap-e2e\`) are an unrelated pre-existing \`npm install\` failure (\`Cannot read properties of null (reading 'edgesOut')\`) in the OpenClaw plugin build step — it fails before the proxy process is even started, so it's not something this PR's changes touch.

gingeekrishna:
Dug into the two remaining red checks (\`docker-native-e2e\`, \`docker-wrap-e2e\`) to make sure they're not actually caused by this PR.

They aren't. The same \`Wrap E2E\` workflow, same \`npm error Cannot read properties of null (reading 'edgesOut')\` during the OpenClaw plugin's \`npm install\`, is currently failing on several unrelated branches from today (\`refactor/codex-provider-slice\`, \`feat/enterprise-governance-sessions-telemetry\`, \`pr/3283-secure-public-bind\`, \`pr/3360-vscode-claude-1m-context\`, \`fix/docker-mcp-stdin\`), and it also failed on #3442 (an unrelated OpenClaw fix merged separately). It's a repo-wide npm-version flake, not specific to this branch.

There's already a fix in flight: #3454 ("chore(ci): pin npm for OpenClaw wrap installs") pins the npm version, and its branch is the only one in that history where the Wrap E2E workflow actually passed. Once that lands (or this branch is rebased onto it), these two checks should go green without any change needed here.

JerrettDavis (approved):
The updated reader resolves my earlier blocking finding: perf now includes legacy proxy.log rotations and per-port proxy.<port>.log rotations, with explicit workspace-wide aggregation and the existing time-window behavior. Reviewed the writer/reader wiring and refreshed mocks; 381 focused tests pass with one skip, including mixed-log aggregation and expiry-window coverage. No remaining blocking code finding on this head. The two Docker jobs still fail during OpenClaw npm installation with edgesOut, matching the separate npm maintenance issue; CI is not yet fully green.

gingeekrishna:
> The updated reader resolves my earlier blocking finding: perf now includes legacy proxy.log rotations and per-port proxy..log rotations, with explicit workspace-wide aggregation and the existing time-window behavior. Reviewed the writer/reader wiring and refreshed mocks; 381 focused tests pass with one skip, including mixed-log aggregation and expiry-window coverage. No remaining blocking code finding on this head. The two Docker jobs still fail during OpenClaw npm installation with edgesOut, matching the separate npm maintenance issue; CI is not yet fully green.

Thank you @JerrettDavis

JerrettDavis:
Thanks for working on the shared-log collision. I’m consolidating that work in #3204, which covers per-port runtime logs, per-port stdio capture, and separate worker log files, along with the readers. Closing this parallel implementation so we settle on one filename and logging policy.

gingeekrishna:
Sure @JerrettDavis

PR #3450, opened 2026-09-06
fix(sdk): preserve proxy authentication alongside provider credentials

trs-80:
## Description

The TypeScript SDK omits its proxy `apiKey` when an OpenAI or Anthropic provider credential is present, causing token-protected remote proxies to reject those requests with 401. Send `X-Headroom-Proxy-Token` from shared `rawFetch()` while preserving provider authentication and the existing bearer fallback. The server already supports this header via #3422.

## Type of Change

## Changes Made

- Add the dedicated proxy-token header in `sdk/typescript/src/client.ts` whenever an SDK proxy key is configured; provider `Authorization`/`x-api-key` headers and the bearer fallback are untouched.
- Add eight regression cases covering both providers with explicit keys, environment keys, proxy-only authentication, and provider-only authentication.

## Testing

- [x] Unit tests pass (`pytest`) — N/A, TypeScript-only change; `npm test` run instead

### Test Output

```text
cd sdk/typescript
npm test
Test Files  18 passed | 5 skipped (23)
Tests       302 passed | 33 skipped (335)
npm run typecheck  # passed
npm run build      # passed
git diff --check   # passed
```

## Real Behavior Proof

- Environment: macOS, Node.js 26.5.1; locked SDK dependencies installed with `npm ci`.
- Exact command / steps: `cd sdk/typescript && npm ci && npm test -- test/client-expanded.test.ts`, then `npm test`, `npm run typecheck`, `npm run build`.
- Observed result: before the fix, six of the eight new cases failed because `X-Headroom-Proxy-Token` was absent from the request headers. After the fix all eight pass and still assert the expected provider headers (`Authorization: Bearer <provider-key>` for OpenAI, `x-api-key` for Anthropic, bearer fallback when no provider key is set). Tests drive the real SDK clients and inspect requests at a mocked fetch boundary.
- Not tested: live provider calls, a deployed remote proxy, and the Python test suite. The SDK suite reports 33 skipped tests.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. The header is sent unconditionally from `rawFetch()` when a proxy key is configured; no rollout flag or channel gate is involved.
- Minimum rollout channel: N/A — not rollout-managed; the fix ships to every channel with the next SDK release.
- Stable/default behavior changed: Yes. When an SDK proxy key is configured, chat/message requests now also carry `X-Headroom-Proxy-Token`. Provider auth headers and the existing bearer fallback are unchanged, and no header is added when no proxy key is set.
- Kill switch / disable path: Do not configure a proxy key (`apiKey` option / `HEADROOM_API_KEY`) — the header is only added when one exists. Otherwise pin the previous SDK version.
- Qualification impact: None. Server-side support for the header landed in #3422, and proxies that do not read it ignore the extra header.
- Rollback path: Revert this commit, or pin the preceding SDK release.

## Review Readiness

## Checklist

## Additional Notes

Python-specific template checks (`pytest`, `ruff`, `mypy`) are not applicable to this TypeScript-only change; the equivalent SDK commands are recorded above. Documentation is unchanged — no new dependencies or public configuration options are introduced.

### DIFF
--- a/sdk/typescript/src/client.ts
+++ b/sdk/typescript/src/client.ts
@@ -510,6 +510,7 @@ export class HeadroomClient implements HeadroomClientInterface {
       ...options.headers,
     };
     if (this.apiKey) {
+      headers["X-Headroom-Proxy-Token"] = this.apiKey;
       // Don't override provider auth headers
       if (!headers["Authorization"] && !headers["x-api-key"]) {
         headers["Authorization"] = `Bearer ${this.apiKey}`;
--- a/sdk/typescript/test/client-expanded.test.ts
+++ b/sdk/typescript/test/client-expanded.test.ts
@@ -2,7 +2,7 @@
  * Tests for expanded HeadroomClient — chat.completions, messages, metrics, CCR, etc.
  * Uses mocked proxy (no real server needed).
  */
-import { describe, it, expect, vi, beforeEach } from "vitest";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 import { HeadroomClient } from "../src/client.js";
 
 const mockFetch = vi.fn();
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed SDK credential construction against the proxy token reader and upstream header filtering. The dedicated proxy-token header preserves explicit/environment provider credentials while retaining proxy-only behavior. Full SDK validation: 302 tests pass, 33 optional tests skip, and TypeScript checking passes. No blocking finding on this head.

JerrettDavis (approved):
Re-reviewed the refreshed head after its rebase onto current main. The patch remains the same focused credential fix: the dedicated proxy token is sent without displacing OpenAI or Anthropic provider authentication, and the eight header-matrix regressions cover explicit, environment, proxy-only, and provider-only cases. The head is current, mergeable, and its governance checks pass. No new blocking finding.

PR #3455, opened 2026-09-07
fix(learn): distinguish Grep and Glob searches by path

### LINKED ISSUE #3453: [BUG] learn loop detection drops `path` from Grep/Glob signatures, merging distinct searches
## Description

`headroom learn`'s loop signature for `Grep` and `Glob` is built from `pattern` alone. `ToolCall.input_summary` ([`models.py:62-76`](https://github.com/headroomlabs-ai/headroom/blob/e59cf1012a2b277689555833fd221a92279f777d/headroom/learn/models.py#L62-L76)) reads only that one field and drops `path`:

```python
if self.name in ("Grep", "grep"):
    return str(self.input_data.get("pattern", "?"))
if self.name in ("Glob", "glob"):
    return str(self.input_data.get("pattern", "?"))
```

So one pattern searched in three different directories produces one signature — `grep::timeouterror` for all three — and the three merge into a single reported loop with attributed waste. `Read` keys on the full `file_path` and `Bash` on the full command, so `Grep`/`Glob` are the outliers: they are the only tools whose signature deliberately discards a field that changes what the call actually did.

This is not the truncation defect in #3440 — nothing is being cut short here; the field is never read. It is also not the schema defect in that issue, where a normalized custom tool carries no `pattern` at all. It is a third, independent mechanism, and it reproduces on released 0.36.5.

`sample_input` has the same blind spot, so the digest line a maintainer reads shows `sample='TimeoutError'` with no indication that three different trees were searched.

### Note: the two schema paths report this differently

Secondary to the above, and recorded so it is not mistaken for a bug in the fix for #3440. That fix adds a fallback: when the per-tool fields yield nothing, identity comes from key-sorted JSON of the whole input. A normalized `codebase_search` arriving as `Grep{query, path}` therefore takes the fallback and **does** distinguish by `path`, while a native `Grep{pattern, path}` does not — visible as `[c]` below.

This is not specific to `path`. Every entry in the per-tool field table is narrower than the input it keys, so the fallback is uniformly the more discriminating of the two: `Edit{file_path, old_string, new_string}` merges three distinct edits to one file, while those same three arriving under a foreign schema stay distinct. That is inherent to a fallback that cannot know which fields carry meaning, and erring toward distinctness is the right default there.

What makes `Grep` the sharpest instance is that both paths are reachable for the same logical tool in one session — `codebase_search` and native `Grep` both normalize to `grep` — so two people doing identical work get different `learn` output depending on which CLI produced the transcript. On its own it does not argue that `path` belongs in the identity key; the fallback asserts only "when the fields are unknown, do not guess."

## To Reproduce

1. `pip install headroom-ai` (reproduced on 0.36.5; the code path is unchanged on `main`)
2. Save the script below as `repro_path.py`
3. `python repro_path.py`

## Code Sample

```python
"""Repro: Grep/Glob signatures drop `path`, so one pattern in N directories is one loop."""
import headroom
from headroom.learn.loops import _canonical_signature, detect_loops

dirs = ["/srv/app/services/ingest/", "/srv/app/services/billing/", "/srv/app/web/api/"]

greps = [tc("Grep", f"call_g{i}", {"pattern": "TimeoutError", "path": d}, msg_index=i)
         for i, d in enumerate(dirs)]
d = detect_loops([SessionData(session_id="s1", tool_calls=greps)])
print(f"\n[a] Grep 'TimeoutError' in {len(dirs)} DISTINCT dirs ->", len(d), "loop(s)")
for lp in d:
    print(f"     count={lp.count} wasted={lp.wasted_tokens:,} sample={lp.sample_input!r}")
print("     signatures:", {_canonical_signature(c) for c in greps})

globs = [tc("Glob", f"call_gl{i}", {"pattern": "**/*.py", "path": d}, msg_index=i)
         for i, d in enumerate(dirs)]
dg = detect_loops([SessionData(session_id="s2", tool_calls=globs)])
print(f"\n[b] Glob '**/*.py' in {len(dirs)} DISTINCT dirs ->", len(dg), "loop(s)")
print("     signatures:", {_canonical_signature(c) for c in globs})

# Same intent, but arriving as a normalized custom tool that carries `query`, not `pattern`.
qs = [tc("Grep", f"call_q{i}", {"query": "TimeoutError", "path": d}, msg_index=i)
      for i, d in enumerate(dirs)]
dq = detect_loops([SessionData(session_id="s3", tool_calls=qs)])
print(f"\n[c] normalized Grep carrying `query`+`path`, {len(dirs)} DISTINCT dirs ->", len(dq), "loop(s)")
print("     signatures:", {_canonical_signature(c) for c in qs})
```

## Expected Behavior

Depends on the design question below, which is why I have not sent a fix.

| Input | If `path` is part of identity | If `path` is deliberately excluded |
|---|---|---|
| `Grep{pattern, path}` × 3 distinct dirs | 0 loops — 3 distinct searches | 1 loop — intentional, and documented as such |
| `Glob{pattern, path}` × 3 distinct dirs | 0 loops | 1 loop |

The normalized `Grep{query, path}` case in `[c]` follows whichever rule is chosen only once input normalization lands as well — see #3440's open question on where that belongs. Until then it takes the generic fallback either way, so I have left it out of the table rather than state an expectation this issue cannot deliver.

## Actual Behavior

Run against `main` @ `e67b3c8a`:

```
[a] Grep 'TimeoutError' in 3 DISTINCT dirs -> 1 loop(s)
     count=3 wasted=20,000 sample='TimeoutError'
     signatures: {'grep::timeouterror'}

[b] Glob '**/*.py' in 3 DISTINCT dirs -> 1 loop(s)
     signatures: {'glob::**/*.py'}

[c] normalized Grep carrying `query`+`path`, 3 DISTINCT dirs -> 1 loop(s)
     signatures: {'grep::?'}
```

Same script against the fix for #3440 — `[a]` and `[b]` are byte-identical, confirming that change leaves this defect exactly as found, while `[c]` diverges:

```
[c] normalized Grep carrying `query`+`path`, 3 DISTINCT dirs -> 0 loop(s)
     signatures: {'grep::{"path": "/srv/app/services/billing/", "query": "timeouterror"}',
                  'grep::{"path": "/srv/app/services/ingest/", "query": "timeouterror"}',
                  'grep::{"path": "/srv/app/web/api/", "query": "timeouterror"}'}
```

## Environment

- **Headroom version**: originally reproduced on 0.36.5 (pipx); output above is from `main` @ `e67b3c8a` (reports `0.38.0-dev`). `input_summary` is unchanged across all three.
- **Python version**: 3.13
- **OS**: Ubuntu (WSL2)
- **LLM Provider**: Anthropic (Claude Code CLI)

## The design question

Should `path` be part of loop identity for `Grep`/`Glob`?

- **Include it.** Three directories are three searches; merging them reports a loop that did not happen, and inflates measured waste the same way #3440's defects do. Consistent with `Read`, with `Bash`, and with the post-PR custom-tool path.
- **Exclude it deliberately.** An agent hunting one pattern across the whole tree arguably *is* flailing, and surfacing that is the point of loop detection. If this is the intent, it is currently accidental — it falls out of a display helper rather than a decision — and it should be written down and applied to both schemas.

I lean toward including it, on the grounds that `wasted_tokens` is presented as a measured lower bound and this makes it an over-estimate. But it changes what `learn` reports on real histories, so I would rather not decide it inside a bug-fix PR. Happy to send the fix in whichever direction you prefer — say the word and I will pick it up.

## Related

- #3440 — grouping-key truncation and the foreign-schema placeholder. Explicitly leaves this defect as found; the correction note in that issue is what this one promised.
- #3452 — replay double-counting via missing `tool_call_id` dedup, split out of #3440.
- #1159 — the feature this module came from (loop weighting).

atirna:
## Description

Fixes #3453.

`headroom learn` grouped native Grep and Glob calls by pattern alone, so the same pattern in different directories appeared as one loop. This includes the search path in the existing summary used for loop signatures and samples, so separate searches remain separate while repeated searches in one directory still form a loop.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Include `path` alongside `pattern` for Grep and Glob input summaries.
- Add regression coverage for separate paths, repeated same-path searches, and the generic summary fallback.

## Testing

### Test Output

```text
uv run pytest tests/test_learn/test_loop_weighting.py -q
16 passed in 0.12s

uv run ruff check headroom/learn/models.py tests/test_learn/test_loop_weighting.py
All checks passed!

uv run ruff format --check headroom/learn/models.py tests/test_learn/test_loop_weighting.py
2 files already formatted
```

## Real Behavior Proof

- Environment: macOS on Apple Silicon, Python 3.13.14.
- Exact command / steps: created one `SessionData` session with three native Grep or Glob calls using one pattern and `/srv/ingest`, `/srv/billing`, and `/srv/web`, then ran `detect_loops`.
- Observed result: before, each tool produced one signature and one three-call loop with 20,000 attributed tokens. After, each produces three path-qualified signatures and no loop; three calls with the same path still produce one loop.
- Not tested: full suite and scanning a retained production transcript.

## Runtime Rollout Safety

- Minimum rollout channel: Standard release.
- Stable/default behavior changed: Grep and Glob searches with the same pattern in different directories no longer merge into one loop.
- Kill switch / disable path: Revert this change.
- Qualification impact: Existing `headroom learn` loop reports become path-specific for native Grep and Glob calls.
- Rollback path: Revert this change; pathless grouping resumes.

## Review Readiness

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas (N/A: the summary format is direct)
- [ ] I have made corresponding changes to the documentation (N/A: no public API or documentation change)

### DIFF
--- a/headroom/learn/models.py
+++ b/headroom/learn/models.py
@@ -67,10 +67,10 @@ class ToolCall:
             return cmd[:100] + "..." if len(cmd) > 100 else cmd
         if self.name in ("Read", "read"):
             return str(self.input_data.get("file_path", "?"))
-        if self.name in ("Grep", "grep"):
-            return str(self.input_data.get("pattern", "?"))
-        if self.name in ("Glob", "glob"):
-            return str(self.input_data.get("pattern", "?"))
+        if self.name in ("Grep", "grep", "Glob", "glob"):
+            pattern = self.input_data.get("pattern", "?")
+            path = self.input_data.get("path", "?")
+            return f"{pattern} in {path}"
         if self.name in ("Edit", "edit", "Write", "write"):
             return str(self.input_data.get("file_path", "?"))
         return str(self.input_data)[:80]
--- a/tests/test_learn/test_loop_weighting.py
+++ b/tests/test_learn/test_loop_weighting.py
@@ -16,6 +16,8 @@ surfaced, were ranked no higher than a one-off rule. These tests pin:
 from pathlib import Path
 from unittest.mock import MagicMock, patch
 
+import pytest
+
 from headroom.learn.analyzer import SessionAnalyzer, _build_digest
 from headroom.learn.fixtures import (
     error_loop_session,
@@ -31,6 +33,8 @@ from headroom.learn.models import (
     ProjectInfo,
     Recommendation,
     RecommendationTarget,
+    SessionData,
+    ToolCall,
 )
 
 
@@ -48,6 +52,46 @@ def _project() -> ProjectInfo:
 
 
 class TestDetectLoops:
+    @pytest.mark.parametrize(
+        ("name", "pattern"),
+        [("Grep", "TimeoutError"), ("Glob", "**/*.py")],
+    )
+    def test_search_signatures_include_path(self, name: str, pattern: str):
+        def call(path: str, index: int):
+            return ToolCall(
+                name=name,
+                tool_call_id=f"{name}-{index}",
+                input_data={"pattern": pattern, "path": path},
+                output="x" * 40000,
+                is_error=False,
+                output_bytes=40000,
+                msg_index=index,
+            )
+
+        distinct_paths = [
+            call("/srv/ingest", 0),
+            call("/srv/billing", 1),
+            call("/srv/web", 2),
+        ]
+        assert len({_canonical_signature(call) for call in distinct_paths}) == 3
+        assert detect_loops([SessionData(session_id=name, tool_calls=distinct_paths)]) == []
+
+        same_path = [call("/srv/ingest", index) for index in range(3)]
+        loop = detect_loops([SessionData(session_id=name, tool_calls=same_path)])[0]
+        assert loop.count == 3
+        assert loop.sample_input == f"{pattern} in /srv/ingest"
+
+    def test_unknown_tool_summary_still_uses_generic_input(self):
+        input_data = {"query": "TimeoutError", "path": "/srv/ingest"}
+        tool_call = ToolCall(
+            name="Search",
+            tool_call_id="search-1",
+            input_data=input_data,
+            output="",
+            is_error=False,
+        )
+        assert tool_call.input_summary == str(input_data)[:80]
+
     def test_refetch_loop_detected_despite_no_errors(self):
         loops = detect_loops([refetch_loop_session(repetitions=5)])
         assert len(loops) == 1


JerrettDavis (approved):
Reviewed the search summary and its use as loop identity. All 16 loop-weighting tests pass: distinct Grep/Glob paths remain separate and repeated searches of the same path still form a loop. No blocking finding on this head. Integration note: #3461 replaces summary-based identity, so its explicit Grep/Glob identity fields must preserve path before both changes land.

PR #3457, opened 2026-09-07
fix(proxy): repair tool-search history in place so signed thinking blocks keep their coordinates

### LINKED ISSUE #3456: tool-search history repair is discarded when the poisoned block shares a message with signed thinking blocks (permanent upstream 400)
## Summary

The tool-search history repair added in #2807 / #2889 correctly identifies the unsupportable
`tool_search_tool_result`, but on a body that carries **signed thinking blocks** the repair is
thrown away before the request goes on the wire. Upstream then 400s on the exact
`tool_reference` the repair had already found, on every turn, so the session is permanently
wedged — a resume replays the same transcript and fails again.

Observed continuously on a Claude Code session routed through the proxy:

```
API Error: 400 Tool reference 'mcp__claude_ai_Instacart__cart' not found in available tools
```

## Root cause

`body_forwarding.thinking_block_fingerprint()` keys each thinking block **positionally**:

```python
out.append((message_index, block_index, json.dumps(block, sort_keys=True, ensure_ascii=False)))
```

`strip_unsupported_tool_search_blocks()` repairs by *removing* the offending
`tool_search_tool_result` and its paired `server_tool_use`. In real Claude Code traffic that pair
sits inside an assistant message that also contains signed thinking blocks:

```
messages[413].content = [thinking(sig), text, server_tool_use, tool_search_tool_result, thinking(sig), tool_use]
                         ^0                 ^2               ^3                          ^4
```

Removing blocks 2 and 3 moves the thinking block at index 4 to index 2 →
`thinking_blocks_survived_mutation()` returns False → `select_outbound_body()` returns
`OutboundBody(content=original_body_bytes, source="passthrough", dropped_mutations=True)` and the
client's original bytes go upstream, discarding **every** mutation including the repair.

So the requests that need the repair are exactly the requests that cannot keep it.

Dropping a whole message (`if not kept: continue`, "the whole turn was tool-search bookkeeping")
has the same effect one level up: it shifts `message_index` for every later message, so any signed
thinking block anywhere after it moves too.

## Evidence

Proxy log for one failing request — repair runs, body is marked mutated, upstream still 400s:

```
01:26:19,795 [hr_1788758740_000006] Tool search: dropped 2 unsupportable history block(s) (tools array cannot resolve their tool_reference entries)
01:26:22,270 event=outbound_request forwarder=streaming body_bytes=19196423 body_mutated=true mutation_reasons=image_compression,tool_search_history_repair
01:26:28,427 [hr_1788758740_000006] Forwarding upstream streaming error status=400 url=https://api.anthropic.com/v1/messages
```

Instrumenting the repair call site on the same session:

```
TSDIAG tools=2437 search_tools=1 ts_blocks=18 unsat_blocks=1 msgs=533
```

i.e. exactly one unsupportable block, correctly detected, and still 400 on the wire.

## Repro sketch

1. Session with tool search on for some turns, so the transcript stores `server_tool_use` +
   `tool_search_tool_result` pairs in assistant messages that also carry signed thinking blocks
   (extended thinking on — the default for Claude Code on Opus).
2. Make one referenced tool disappear from `tools` (remove/disconnect an MCP server, or re-auth a
   server that changes its advertised tool set, or restart the client with `--tool-search false`
   so every tool ships inline and no `tool_search_tool_*` tool is present).
3. Every subsequent request 400s with `Tool reference 'X' not found in available tools`, despite
   the proxy logging that it repaired the block.

## Proposed fix

Repair **in place**: swap each unsupportable block for a short constant text block instead of
removing it, and never remove a message. Every thinking block keeps its coordinates and its bytes,
the fingerprint matches, and the repair survives to the wire. This is the same reasoning already
documented for the CCR sibling `strip_unsupported_ccr_retrieve_blocks` ("Neutralize rather than
drop"), which does not have this bug.

Secondary: `HEADROOM_DEBUG_DUMP` only dumps on the buffered path, so a streaming 400 — every
Claude Code turn — writes no diagnostic at all. That is what made this hard to see.

PR follows.

Eyalm321:
## Description

`strip_unsupported_tool_search_blocks()` repairs by removing the unsupportable
`tool_search_tool_result` and its paired `server_tool_use`. When that pair sits in an assistant
message that also carries **signed thinking blocks** — the normal shape of Claude Code traffic with
extended thinking on — the removal moves every later block in that message.
`thinking_block_fingerprint()` keys a thinking block by `(message_index, block_index, json)`, so
`thinking_blocks_survived_mutation()` returns False and `select_outbound_body()` forwards the
client's ORIGINAL bytes, discarding every mutation including the repair. Upstream then 400s on the
exact reference the repair had already found, and because a resume replays the same transcript the
session is wedged permanently.

Dropping a whole message (`if not kept: continue`) has the same effect one level up: it shifts
`message_index` for every message after it.

Fix: repair in place. Swap each unsupportable block for a short constant text block and never
remove a message, so every thinking block keeps its coordinates and its bytes. Same reasoning and
same shape as the CCR sibling `strip_unsupported_ccr_retrieve_blocks` right below it ("Neutralize
rather than drop"), which does not have this bug.

Also: dump the request on a **streaming** upstream error, not only on the buffered path. Claude
Code streams every turn, so `HEADROOM_DEBUG_DUMP` previously captured nothing for the most common
400s — that blindness is why this took three separate investigations to pin down. Same gating as
the existing dumps (off by default, never in stateless mode, redacted unless `=full`).

Closes #3456

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/helpers.py`: `strip_unsupported_tool_search_blocks()` replaces unsupportable
  blocks with a constant `_TOOL_SEARCH_PLACEHOLDER_BLOCK` text block instead of deleting them; no
  message is ever dropped. Return contract unchanged (`(messages, count)`, original object by
  identity when nothing changed), so the call site and its cache/prefix behaviour are unchanged.
  The placeholder is constant so the repaired prefix stays byte-stable across turns.
- `headroom/proxy/handlers/anthropic.py`: log line now says "repaired ... (replaced with text in
  place)" to match the behaviour.
- `headroom/proxy/handlers/streaming.py`: write the same diagnostic dump the non-streaming handlers
  write when an upstream response is >= 400.
- `tests/test_issue_746_tool_search.py`: three tests updated from drop- to in-place semantics, plus
  a new regression test asserting `thinking_block_fingerprint()` is byte-identical across a repair.

## Testing

### Test Output

```text
$ pytest tests/test_issue_746_tool_search.py tests/test_proxy/test_tool_search_repair_after_turn_hooks.py -q
63 passed, 1 warning in 3.09s

$ pytest tests -k "tool_search or thinking or body_forward or outbound" -q
208 passed, 26 skipped, 11637 deselected, 1 warning in 16.08s

$ ruff check .
All checks passed!

$ mypy headroom
Success: no issues found in 530 source files
```

## Real Behavior Proof

- Environment: proxy 0.38.0-dev (this branch on `main` @ e67b3c8a), Python 3.14, Fedora 44,
  `HEADROOM_MODE=token HEADROOM_BACKEND=anthropic`, upstream `api.anthropic.com`, live Claude Code
  2.1.263 session (claude-opus-5, extended thinking on, ~533 messages / 19 MB body / 2436 client
  tools).
- Exact command / steps: reproduce the wedged session, capture the real outbound body with `HEADROOM_DEBUG_DUMP=full`, then replay that body through the helper before and after this change.
  1. The session had looped on `API Error: 400 Tool reference 'mcp__claude_ai_Instacart__cart' not
     found in available tools` for every turn — a claude.ai connector that no longer advertised
     that tool, while 18 `tool_search_tool_result` blocks in the transcript still referenced it.
  2. Captured the real outbound body with `HEADROOM_DEBUG_DUMP=full` plus the new streaming dump.
  3. Replayed the captured body through the helper, before and after this change.
- Observed result: before, the repair was computed and then discarded and upstream still returned 400; after, the same body repairs with every thinking-block fingerprint byte-identical and the wedged live session answers normally again.
  - Before: proxy logged `Tool search: dropped 2 unsupportable history block(s)` and
    `body_mutated=true mutation_reasons=...,tool_search_history_repair`, upstream still returned
    400. Instrumented at the call site: `TSDIAG tools=2437 search_tools=1 ts_blocks=18
    unsat_blocks=1 msgs=533` — detected, repaired, discarded.
  - After, replaying the captured 19 MB body through the helper:
    ```
    no search tool (client sent all): repaired=36 unsatisfied_refs_left=0 msgs=533 thinking_fingerprint_identical=True
    with injected search tool:        repaired=2  unsatisfied_refs_left=0 msgs=533 thinking_fingerprint_identical=True
    ```
  - Live: restarted the proxy on this branch and prompted the same wedged session — it answered
    normally on the first turn, full history preserved, and has kept answering with the repair
    firing each turn and no further 400s.
- Not tested: the OpenAI/Codex handlers (untouched by the helper change; they do not carry
  Anthropic signed thinking blocks), and Copilot/third-party Anthropic-compatible upstreams, where
  client-originated `tool_search_tool_*` entries are stripped before this stage.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — this is a correctness fix inside the existing, unconditional
  tool-search history repair (#2807 / #2889).
- Minimum rollout channel: n/a (no flag).
- Stable/default behavior changed: yes, narrowly — an unsupportable tool-search block is now
  forwarded as a short text block instead of being removed, and an assistant turn that consisted
  only of tool-search bookkeeping keeps its message slot as text blocks instead of disappearing.
  Both shapes are valid to the API; the previous shapes were valid too, they just could not survive
  the signed-thinking passthrough.
- Kill switch / disable path: `HEADROOM_TOOL_SEARCH=0` still disables Headroom's own deferral
  injection; the repair itself is deliberately unconditional (transcripts poisoned before the flag
  was turned off must still recover), unchanged by this PR.
- Qualification impact: none expected; token accounting is unaffected beyond a few tokens of
  placeholder text per repaired block (repairs are rare and the text is ~10 tokens).
- Rollback path: revert this commit — behaviour returns to removal, i.e. to the bug.

## Review Readiness

## Checklist

### DIFF
--- a/headroom/proxy/handlers/_debug_dump.py
+++ b/headroom/proxy/handlers/_debug_dump.py
@@ -8,8 +8,14 @@ OpenAI handlers so the gating stays consistent across providers.
 
 from __future__ import annotations
 
+import json
+import logging
 import os
+from datetime import datetime
+from pathlib import Path
 from typing import Any
+
+logger = logging.getLogger(__name__)
 
 
 def _debug_dump_mode(config: Any) -> str:
@@ -45,3 +51,52 @@ def _redact_debug_value(value: Any, _max_len: int = 80) -> Any:
     if isinstance(value, list):
         return [_redact_debug_value(v, _max_len) for v in value]
     return value
+
+
+def write_upstream_error_dump(
+    config: Any,
+    *,
+    request_id: str,
+    url: str,
+    status: int,
+    provider: str | None = None,
+    model: str | None = None,
+    body: Any = None,
+    transforms: Any = None,
+    stream: bool = False,
+) -> Path | None:
+    """Write a diagnostic dump of an erroring (>=400) upstream request.
+
+    Returns the path written, or ``None`` when the dump is disabled (the
+    default, and always in stateless mode) or could not be written. Never
+    raises: a diagnostic must not turn an upstream error into a proxy error.
+    """
+    mode = _debug_dump_mode(config)
+    if mode == "off":
+        return None
+    try:
+        from headroom import paths as _hr_paths
+
+        dump_dir = _hr_paths.debug_400_dir()
+        dump_dir.mkdir(parents=True, exist_ok=True)
+        dump_path = dump_dir / f"{datetime.now().strftime('%Y%m%d_%H%M%S')}_{request_id}.json"
+        dump_path.write_text(
+            json.dumps(
+                {
+                    "request_id": request_id,
+                    "url": url,
+                    "status": status,
+                    "provider": provider,
+                    "model": model,
+                    "stream": stream,
+                    "transforms": transforms,
+                    "body": body if mode == "full" else _redact_debug_value(body),
+                },
+                indent=2,
+                default=str,
+            )
+        )
+        return dump_path
+    except Exception:
+        logger.debug("upstream error debug dump skipped", exc_info=True)
+        return None
--- a/headroom/proxy/handlers/_debug_dump.py
+++ b/headroom/proxy/handlers/_debug_dump.py
@@ -14,6 +14,7 @@ import os
 from datetime import datetime
 from pathlib import Path
 from typing import Any
+from urllib.parse import urlsplit, urlunsplit
 
 logger = logging.getLogger(__name__)
 
@@ -53,6 +54,33 @@ def _redact_debug_value(value: Any, _max_len: int = 80) -> Any:
     return value
 
 
+def _redact_url(url: str) -> str:
+    """Drop the query string and fragment from ``url``.
+
+    Some upstreams carry credentials in the query (Gemini's ``?key=``), and the
+    query is never needed to debug a rejected request, so it is elided in every
+    mode — ``full`` opts in to prompt content, not to credentials.
+    """
+    try:
+        parts = urlsplit(url)
+    except ValueError:
+        return "<unparseable url>"
+    if not parts.query and not parts.fragment:
+        return url
+    redacted = urlunsplit((parts.scheme, parts.netloc, parts.path, "", ""))
+    return f"{redacted}?<redacted>" if parts.query else redacted
+
+
+def _decode_sent_body(body: Any) -> Any:
+    """Decode the exact bytes sent upstream back to JSON for the dump."""
+    if not isinstance(body, (bytes, bytearray)):
+        return body
+    try:
+        return json.loads(body)
+    except ValueError:
+        return f"<{len(body)} bytes, not JSON>"
+
+
 def write_upstream_error_dump(
     config: Any,
     *,
@@ -62,10 +90,17 @@ def write_upstream_error_dump(
     provider: str | None = None,
     model: str | None = None,
     body: Any = None,
+    body_source: str | None = None,
     transforms: Any = None,
     stream: bool = False,
 ) -> Path | None:
     """Write a diagnostic dump of an erroring (>=400) upstream request.
+
+    ``body`` may be the raw bytes that went on the wire; they are decoded here,
+    after the off-by-default gate, so a disabled dump costs nothing.
+    ``body_source`` records which body was forwarded (for example
+    ``"passthrough"`` when the proxy's mutations were dropped), since that is
+    often the whole explanation for a 400.
 
     Returns the path written, or ``None`` when the dump is disabled (the
     default, and always in stateless mode) or could not be written. Never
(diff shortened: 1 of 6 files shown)


JerrettDavis (approved):
Reviewed in-place tool-search repair, object identity on no-op paths, paired-block handling, signed-thinking coordinates, and the streaming diagnostic writer. Validation: 81 tool-search/debug-dump tests plus 78 byte-forwarding/signed-thinking tests pass. The dump remains opt-in and disabled in stateless mode; it uses the existing best-effort redaction semantics. No blocking finding on this head.

PR #3458, opened 2026-09-07
fix: preserve retrieval results through the Pi MCP wrapper

willgao-dev:
## Problem

Pi exposes MCP calls through `mcp_daemon` with an `action` and a `server.tool` selector.
Headroom sees the wrapper name instead of the selected tool.
Retrieval results can therefore enter compression again instead of keeping their original content.

## Change

- Resolve valid `mcp_daemon` calls to the existing canonical MCP name.
- Reuse the existing exclusion and retrieval protection paths.
- Leave search, describe, malformed arguments, and Hermes behavior unchanged.
- Test string and object arguments, invalid selectors, and Responses output preservation.

## Local verification

- `python -m pytest tests/test_hermes_tool_call_unwrap.py tests/test_openai_responses_compression_units.py tests/test_openai_responses_read_protection.py -q`: 74 passed.
- `ruff check headroom/config.py tests/test_hermes_tool_call_unwrap.py tests/test_openai_responses_compression_units.py`: passed.
- `ruff format --check headroom/config.py tests/test_hermes_tool_call_unwrap.py tests/test_openai_responses_compression_units.py`: passed.
- `git diff --check`: passed.
- Independent review found no required changes.

Tests use Python 3.13 and the installed Headroom 0.37 native extension with the current Python source.
A native rebuild did not complete because the machine lacks Rust. This change does not modify native code.
The full suite and live proxy rollout remain unverified.

### DIFF
--- a/headroom/config.py
+++ b/headroom/config.py
@@ -387,12 +387,12 @@ _HERMES_TOOL_CALL_WRAPPER = "tool_call"
 
 
 def unwrap_tool_call_name(name: str, arguments: Any) -> str:
-    """Extract the real tool name from a Hermes deferred ``tool_call`` wrapper.
+    """Extract the real tool name from a Hermes or Pi wrapper.
 
     Non-wrapper names pass through unchanged. Malformed/unparseable wrappers
     fail open and return the wrapper name (caller decides what that means).
     """
-    if name != _HERMES_TOOL_CALL_WRAPPER:
+    if name not in {_HERMES_TOOL_CALL_WRAPPER, "mcp_daemon"}:
         return name
     raw = arguments
     if isinstance(raw, str):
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
Reviewed the Pi MCP retrieval wrapper handling and fixed the minor malformed-name regression by using tuple membership, which tolerates dict/list tool names as the previous comparison did. All 74 focused tests, malformed-name checks, and Ruff validation passed. No blocking issue found in the resulting change.

JerrettDavis:
Thanks for tracking down the Pi MCP-wrapper retrieval issue. #3506 includes this entire patch and its regression tests, along with the broader Pi evidence and image fixes. I’m closing this duplicate and keeping #3506 as the place to finish that work; it still needs the final human merge review.

PR #3460, opened 2026-09-07
fix(output-savings): count conversations, not requests, in the holdout gate

gglucass:
## Description

Holdout assignment is conversation-stable by design: `assign_arm` hashes a conversation key, so a whole conversation is treatment or control. Every gate downstream then counts **requests**.

Those are not the same unit. One long agent session assigned to control lands as thousands of requests in a single stratum, and `estimate_from_holdout` divides by exactly those counts:

```python
var += (n * n) * (c.var / c.n + t.var / t.n)
```

so one session reads as a precise measurement of an entire stratum, with a tight band to match.

On a real 91,462-request ledger the two largest terms in a `-1.6%` "measured" reduction came from strata holding **four** control requests each:

```
  -1,116,322 tok   nT=5046  nC=4   opus|mechanical_continuation|m|tools   c=324.5  t=545.7
    -354,245 tok   nT=1459  nC=4   opus|unknown|m|tools                   c=467.2  t=710.1
```

Excluding strata with fewer than five control samples flips the same ledger to `+3.6%`. The arm as a whole covers 15-23% of that machine's opus traffic and 0% of its fable and sonnet traffic (38,000 requests), against a configured 3% holdout: a lottery over a handful of long sessions, not a sample of traffic.

A request-count floor does not catch this. 2,588 control requests from one conversation clears any sample floor comfortably; it is still one draw answering one question.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `_Accum` tracks the distinct conversations behind its observations, saturating at `_CLUSTER_CAP` (32). The count is only ever compared against a threshold, so an exact tally of a busy stratum buys nothing and would grow a set entry per conversation in a file that is rewritten every 25 requests.
- `estimate_from_holdout` admits a stratum only once **both** arms hold `MEASURED_MIN_CLUSTERS` (5) distinct conversations. Excluded rather than down-weighted: the arm describes one session's worth of work, and no weighting recovers a comparison that was never made.
- The conversation rides the existing `transforms_applied` label channel next to the stratum (`output_shaper:conv:<12 hex>`), so `RequestOutcome` and its construction sites are untouched, exactly as the stratum label already does. All three shaper paths emit it: Anthropic messages, OpenAI chat, OpenAI responses.
- The conversation key is already a sha256 digest, so the label is a truncation of one. No request content reaches the label or the ledger.
- `record()` takes an optional conversation; a request without one still records its output tokens and simply does not advance the cluster count.
- `_Accum.to_dict` omits the field when empty, so a baseline model serializes byte-identically to before.

## Testing

`mypy headroom` is left unchecked for one pre-existing error in `headroom/integrations/langchain/chat_model.py:394`, a file this PR does not touch. Type checking is clean on every changed file (see below).

Manual testing is the production-ledger replay in Real Behavior Proof, not a live proxy run.

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_output_savings.py tests/test_output_savings_policy.py tests/test_output_shaper_responses.py tests/test_output_shaper_conversation_label.py -q
103 passed in 4.72s

$ uv run --frozen --extra dev pytest tests/ -k "output_shap or shaper or stratum or savings" -q
399 passed, 24 skipped, 12089 deselected, 2 warnings in 21.36s

$ uvx ruff check .
All checks passed!

$ uv run --frozen --extra dev mypy headroom/proxy/output_savings.py headroom/proxy/output_savings_policy.py

$ uv run --frozen --extra dev mypy headroom
headroom/integrations/langchain/chat_model.py:394: error: Function "RunnableBinding" could always be true in boolean context  [truthy-function]
```

New regressions: `test_one_conversation_per_arm_does_not_qualify` (2,500 requests an arm, one session each side), `test_thin_control_arm_does_not_ride_on_a_thick_treatment_one`, `test_best_estimate_falls_back_when_the_holdout_is_one_conversation`, `test_a_ledger_written_before_conversations_were_tracked_does_not_qualify`, `test_cluster_tracking_saturates`, `test_conversation_survives_a_save_load_cycle`, `test_recorder_reads_the_conversation_off_the_label_channel`, `test_a_request_without_a_conversation_label_still_records`.

Codecov flagged the label emission in the two handlers (3 lines each) as uncovered, which was fair: nothing drove those paths. `tests/test_output_shaper_conversation_label.py` now drives both end to end against a mocked upstream and asserts the stratum and conversation labels arrive together on `x-headroom-transforms`. Revert-checked: deleting either `transforms_applied.append(conversation_label(...))` fails the matching test, and restoring it passes. That is the regression the gate needs guarded, since a handler that emits the stratum alone feeds strata that can never qualify and the measured estimate is withheld forever with nothing in the logs to say why.

Four existing tests recorded arms with no conversations and now pass distinct ones; their assertions are unchanged. One responses test pinned `len(labels) == 1` for the control arm and now pins both labels.

## Real Behavior Proof

- Environment: macOS 24.6.0, Python 3.11.13, uv, upstream/main at e67b3c8a
- Exact command / steps: loaded a production ledger (`~/.headroom/output_savings.json`, 91,462 treatment and 6,356 control requests across 89 strata, written by a 0.37.0 proxy at a 3% holdout) through `SavingsLedger.from_dict` and called `estimate_from_holdout` / `best_estimate`, then re-ran it with conversation counts backfilled at 1, 4 and 5 per arm/stratum
- Observed result: unmodified, `estimate_from_holdout()` returns `None` and `best_estimate().kind` is `estimated` - the file carries no conversation data, so nothing in it can distinguish one session from a hundred. Backfilled at 1 and at 4 conversations per arm: still withheld. At 5: `measured pct=-1.86 over 36,445 requests`, i.e. the gate admits the number as soon as the evidence exists rather than suppressing it categorically.
- Not tested: no live proxy run was made against a real holdout with the new label in flight; the label round-trip is covered by `test_recorder_reads_the_conversation_off_the_label_channel` rather than an end-to-end request. The `+3.6%` figure quoted above is a control-n sensitivity sweep over the same ledger, not an output of this code.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. This changes which of two already-computed estimates is reported, not whether either is computed.
- Stable/default behavior changed: Yes, deliberately. A deployment whose holdout is thin in conversations now reports `estimated` where it previously reported `measured`. Every existing ledger is in that state on upgrade, since none carries conversation data, and returns to `measured` once its holdout accrues five conversations per stratum in both arms. No request accounting, no wire format change; the ledger gains one optional key per accumulator.
- Kill switch / disable path: No runtime flag. Setting `MEASURED_MIN_CLUSTERS = 0` restores the previous admission rule exactly.
- Qualification impact: Four existing tests re-recorded with conversations, assertions unchanged; one responses test now pins two labels instead of one. No test was relaxed.
- Rollback path: Revert the commit. The estimate is recomputed from the ledger on every read, so the previous headline returns immediately. `clusters` keys left in a ledger by this build are ignored by older code (`_Accum.from_dict` reads only n/sum/sumsq).

## Review Readiness

## Checklist

Documentation is unchecked because there is none to change: the ledger format is internal, the label is internal, and the two constants are documented at their definitions.

## Additional Notes

Composes with #3127 (request-count floors on the same estimator) rather than replacing it: that PR keeps a thin arm from deciding the headline, this one keeps a thick arm that is one conversation from doing the same. Both touch `estimate_from_holdout`, so whichever lands second needs a trivial rebase.

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -3200,6 +3200,7 @@ class AnthropicHandlerMixin:
                 from headroom.proxy.output_savings import (
                     assign_arm,
                     conversation_key_from_body,
+                    conversation_label,
                     stratum_key,
                     stratum_label,
                 )
@@ -3226,7 +3227,8 @@ class AnthropicHandlerMixin:
                         _holdout = float(runtime_env.getenv("HEADROOM_OUTPUT_HOLDOUT", "0") or "0")
                     except ValueError:
                         _holdout = 0.0
-                    _arm = assign_arm(conversation_key_from_body(body), _holdout)
+                    _conversation = conversation_key_from_body(body)
+                    _arm = assign_arm(_conversation, _holdout)
 
                     # Stratum from request features observable now (mirrors the
                     # offline baseline so live and learned strata line up).
@@ -3239,7 +3241,10 @@ class AnthropicHandlerMixin:
                     )
                     # Carry (arm, stratum) on the existing label channel so the
                     # outcome funnel can feed the savings ledger from any path.
+                    # The conversation rides with it: it is the unit the arm was
+                    # assigned to, so it is the unit the estimator has to count.
                     transforms_applied.append(stratum_label(_arm, _stratum))
+                    transforms_applied.append(conversation_label(_conversation))
 
                     if _arm == "treatment":
                         _level, _src = resolve_verbosity_level(_shaper_settings)
--- a/headroom/proxy/handlers/openai.py
+++ b/headroom/proxy/handlers/openai.py
@@ -718,6 +718,7 @@ def _shape_openai_responses_payload(
         from headroom.proxy.output_savings import (
             assign_arm,
             conversation_key_from_responses_body,
+            conversation_label,
             stratum_key,
             stratum_label,
         )
@@ -736,7 +737,8 @@ def _shape_openai_responses_payload(
             holdout = float(runtime_env.getenv("HEADROOM_OUTPUT_HOLDOUT", "0") or "0")
         except ValueError:
             holdout = 0.0
-        arm = assign_arm(conversation_key_from_responses_body(payload), holdout)
+        conversation = conversation_key_from_responses_body(payload)
+        arm = assign_arm(conversation, holdout)
 
         turn_kind = classify_responses_turn(payload.get("input")).value
         approx_input_tokens = len(json.dumps(payload)) // 4
@@ -746,7 +748,9 @@ def _shape_openai_responses_payload(
             model=model or str(payload.get("model", "")),
             has_tools=bool(payload.get("tools")),
         )
-        labels = [stratum_label(arm, stratum)]
+        # The conversation is the unit the arm was assigned to, so the ledger
+        # needs it to count distinct conversations rather than requests.
+        labels = [stratum_label(arm, stratum), conversation_label(conversation)]
 
         if arm != "treatment":
             return labels, False
@@ -4490,6 +4494,7 @@ class OpenAIHandlerMixin:
             from headroom.proxy.output_savings import (
                 assign_arm,
                 conversation_key_from_body,
+                conversation_label,
                 stratum_key,
                 stratum_label,
             )
@@ -4514,7 +4519,8 @@ class OpenAIHandlerMixin:
                     _holdout = float(runtime_env.getenv("HEADROOM_OUTPUT_HOLDOUT", "0") or "0")
                 except ValueError:
                     _holdout = 0.0
-                _arm = assign_arm(conversation_key_from_body(body), _holdout)
+                _conversation = conversation_key_from_body(body)
+                _arm = assign_arm(_conversation, _holdout)
                 _turn_kind = classify_turn(body.get("messages", [])).value
                 _stratum = stratum_key(
                     turn_kind=_turn_kind,
@@ -4523,8 +4529,10 @@ class OpenAIHandlerMixin:
                     has_tools=bool(body.get("tools")),
                 )
                 # Carry (arm, stratum) on the transforms channel so the outcome
-                # funnel feeds the output-savings ledger from the chat path too.
+                # funnel feeds the output-savings ledger from the chat path too,
+                # plus the conversation the arm was actually assigned to.
                 transforms_applied.append(stratum_label(_arm, _stratum))
+                transforms_applied.append(conversation_label(_conversation))
                 if _arm == "treatment":
                     _level, _src = resolve_verbosity_level(_shaper_settings)
                     _shape_result = shape_openai_chat_request(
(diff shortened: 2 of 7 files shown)


JerrettDavis (changes requested):
The conversation gate and all three handler label paths pass the focused tests (61 ledger tests plus 37 handler tests). Please keep observations without conversation provenance out of the measured accumulator after upgrade. _Accum.from_dict retains the old n/sum/sumsq, and new labeled observations add their clusters to that same accumulator. Once five fresh conversations arrive in each arm, all old unknown-conversation traffic becomes measured too. Reproduction: load an old stratum with 2,500 unlabeled requests per arm (control=1000 tokens, treatment=2000), then add five distinct labeled conversations per arm with identical 1000-token outputs. The estimate switches from None to measured=-99.80% over 2,505 treatment requests, despite the known-conversation sample showing no difference. Those old requests could all be one conversation, the exact case this gate is intended to exclude. Preserve legacy totals for historical/estimated reporting, but use a separate provenance-qualified accumulator or measurement epoch for the measured result; likewise, later unlabeled requests must not enter an already-qualified measured stratum. Add save/load/upgrade coverage that reaches the five-conversation threshold and confirms unknown-provenance traffic stays excluded.

gglucass:
Fixed in f377a45f. You were right that the clusters vouched for the wrong thing: `_Accum` carried one set of totals, so upgrading a ledger kept its legacy `n/sum/sumsq` and the five fresh conversations only had to arrive alongside them.

Each accumulator now holds a conversation-qualified subset (`qn/qsum/qsumsq`) fed only when an observation arrives with a conversation label, and `estimate_from_holdout` reads that subset exclusively - means, weights and variance. The full totals are untouched and still feed the estimated and modelled tiers and historical reporting, per your suggestion. The same split is what keeps later unlabelled requests out of an already-qualified stratum: they never reach `qn`. `to_dict` omits the q-fields when empty, so a pre-upgrade file and a baseline model serialize exactly as before.

Three tests on your reproduction: `test_upgraded_legacy_traffic_never_joins_the_measured_arm` loads a legacy dict of 2,500 unlabelled requests per arm, adds five distinct labelled conversations each side at identical output, and asserts the measured estimate is the 5-request labelled sample at 0% rather than the 2,505-request number; `test_the_qualified_subset_survives_a_save_load_cycle` runs that through save/reload so the split cannot be lost on restart; `test_later_unlabelled_requests_stay_out_of_a_qualified_stratum` adds 2,000 unlabelled requests after the gate opens and asserts the estimate does not move. All three fail against the previous commit.

JerrettDavis (approved):
Re-reviewed the provenance-qualified accumulators and their merge/save/load paths. The measured estimate now uses only labelled observations, so fresh conversations no longer admit old unknown-provenance totals or later unlabelled traffic into the measured arm. Historical totals remain available for the other reporting tiers. All 64 ledger tests pass, including the original 2,500-request upgrade case and persistence regression. The earlier blocking finding is resolved.

Issue #3452, opened 2026-09-06
[BUG] learn loop detection double-counts replayed transcripts (no tool_call_id dedup)


njchae:
## Description

`headroom learn`'s loop detection counts the same tool call once per transcript file it appears in. `detect_loops()` ([`loops.py:121-130`](https://github.com/headroomlabs-ai/headroom/blob/e59cf1012a2b277689555833fd221a92279f777d/headroom/learn/loops.py#L121-L130)) merges each session's qualifying groups into a global view keyed by signature, and never dedups by `tool_call_id`:

```python
for sig, calls in per_session.items():
    if len(calls) >= min_occurrences:
        groups.setdefault(sig, []).extend(calls)   # <- no id dedup
```

`ToolCall.tool_call_id` carries the provider-assigned id and is already parsed by the scanners, so the information needed to tell a replay from a repetition is present and unused.

When a provider resumes a conversation by writing a **new** transcript that replays prior history, every replayed call is scanned again and counted again. Three reads become six; the measured waste rises with them. Nothing about the agent's behavior changed — only the number of files on disk describing it.

This is not the truncation defect in #3440. That one merges *distinct* calls because the grouping key is lossy; this one over-counts calls that are correctly grouped and genuinely identical, because the same call is presented to the scanner more than once. The two compound: an already-inflated loop is then multiplied by the number of replays.

### Consequence: `min_occurrences` is evaluated before dedup

The threshold check (`len(calls) >= min_occurrences`) runs on the per-session group *before* any deduplication. So a group whose repetitions are entirely replays of a single call clears the bar on replay count alone, and is reported as a loop with a real signature and attributed waste — despite the underlying session containing one call. Any fix that dedups after the threshold has to re-check it, or it converts double-counted loops into phantom ones rather than removing them.

### Consequence: `tool_call_id` uniqueness becomes a scanner contract

Deduping by id means an id must be unique *across* sessions, not merely within one. Where a scanner synthesizes ids rather than reading them from the transcript, two unrelated sessions running the same loop would otherwise look like one session replayed twice — the same defect in the opposite direction. This is worth stating explicitly wherever the dedup lands, since it constrains every current and future plugin.

## To Reproduce

1. `pip install headroom-ai` (originally reproduced on 0.36.5; the code path is unchanged on `main`)
2. Save the script below as `repro_dedup.py`
3. `python repro_dedup.py`

## Code Sample

```python
[...]

PR #3461, opened 2026-09-07
fix(learn): derive loop signature from tool input identity, not display summary

### LINKED ISSUE #3440: [BUG] learn loop detection uses the display summary as its grouping key → inflated loop counts
## Description

`headroom learn`'s loop detection groups tool calls by `ToolCall.input_summary` — a property whose own docstring says *"Short summary of tool input **for display**"* ([`models.py:62-76`](https://github.com/headroomlabs-ai/headroom/blob/e59cf1012a2b277689555833fd221a92279f777d/headroom/learn/models.py#L62-L76)). `_canonical_signature()` ([`loops.py:88-100`](https://github.com/headroomlabs-ai/headroom/blob/e59cf1012a2b277689555833fd221a92279f777d/headroom/learn/loops.py#L88-L100)) then uses that deliberately-lossy display string as the **identity** key for grouping:

```python
def _canonical_signature(tc: ToolCall) -> str:
    raw = tc.input_summary.strip()      # <- display helper used as identity key
    ...
    return f"{tc.name.lower()}::{raw}"
```

**Scope:** `Read` returns the full `file_path`, so its signatures are faithful. Everything else is truncated — `Bash` at 100 characters (`cmd[:100] + "..."`, `models.py:67`) and every non-builtin tool at 80 (`str(input_data)[:80]`, `models.py:76`). Calls that diverge only *after* the cutoff compare equal and merge into one loop, so this reaches the **default Claude Code path** through ordinary long shell pipelines; it isn't confined to custom-tool providers.

> **Correction (2026-09-06).** This paragraph originally read "`Read`/`Grep`/`Glob` return
> full paths and patterns, so those signatures are faithful." That is wrong for
> `Grep`/`Glob`: `input_summary` returns only `pattern` and drops `path`, so one pattern
> searched in two different directories produces the same signature — `grep::timeouterror`
> for both — and the two merge into one loop. I reproduced this on released `0.36.5`, so it
> is a *separate, pre-existing* defect rather than the truncation bug described here. The
> PR for this issue leaves that behavior exactly as it found it; I'll report it separately
> rather than widen this issue, since the fix turns on a design question this issue doesn't
> raise.

The arithmetic downstream is sound — `_tokens()` sums real per-call `output_bytes` and `wasted_tokens` is a measured lower bound. The defect is upstream of it, in the grouping that decides *which calls are the same loop*: correct arithmetic over a wrong partition. Inflated counts then propagate into `format_loops_for_digest()` (the "HIGHEST PRIORITY" LLM section) and `apply_loop_weighting()`, so a phantom loop can outrank a real one.

> **Split (2026-09-06).** This issue originally also covered a third defect — `detect_loops()`
> never dedups by `tool_call_id`, so a resume transcript that replays prior history counts every
> replayed call once per file. That is a different mechanism from the two below and is now tracked
> separately in #3452. The original numbering is kept so the earlier discussion and the linked PR
> still line up — there is deliberately no "Bug 2" here.

### Bug 1 — display truncation collides distinct calls

Two cutoffs, one failure. A long `rg` invocation over different paths, or patches under a common worktree, become byte-identical strings once truncated, so N unrelated calls collapse into a single fake "repeated N×" loop. Shell pipelines routinely exceed 100 characters, which is why the Bash branch matters as much as the custom-tool one.

### Bug 3 — a builtin tool name carrying a foreign input schema collapses to a placeholder key

`normalize_tool_name()` ([`_shared.py:171-177`](https://github.com/headroomlabs-ai/headroom/blob/e59cf1012a2b277689555833fd221a92279f777d/headroom/learn/_shared.py#L171-L177)) maps provider tool names onto the builtin ones — `codebase_search`, `search_text` and `search_code` all become `Grep`; `read_many_files` becomes `Read`; `write_file` becomes `Write` — but it renames the tool without normalizing its *input*. `input_summary` then looks for `pattern` or `file_path`, finds neither, and returns its literal `"?"` placeholder for every one of them. Every such call in a session therefore shares one signature, `grep::?`, and N unrelated searches collapse into a single fake loop.

Same symptom as Bug 1, different mechanism, and it needs no long input: these collide at any length. Unlike the `Grep`/`path` defect corrected above, the fix here carries no design question — the generic key-sorted-JSON path already exists and is unambiguously the right key when the guess misses — so this one is folded into the PR.

## To Reproduce

1. `pip install headroom-ai` (reproduced on 0.36.5; the code path is identical on `main`)
2. Save the script below as `repro_loops.py`
3. `python repro_loops.py`

## Code Sample

```python
"""Repro: headroom learn loop detection groups by a display summary."""
import headroom
from headroom.learn.loops import _canonical_signature, detect_loops

# --- Bug 1a: Bash cmd[:100] collides commands that diverge after the cutoff ---
base = ("rg --no-heading --line-number --color never "
        "'TimeoutError|ValueError|KeyError' /srv/app/services/ingest/pipeline/")
bash = [tc("Bash", f"call_b{i}", {"command": f"{base}module_{i}/handlers.py"}, msg_index=i)
        for i in range(5)]
d1a = detect_loops([SessionData(session_id="s1", tool_calls=bash)])
print("\n[1a] 5 DISTINCT long Bash commands ->", len(d1a), "loop(s)")
for lp in d1a:
    print(f"     count={lp.count} wasted={lp.wasted_tokens:,}")
print("     command len:", len(bash[0].input_data["command"]),
      "-> input_summary len:", len(bash[0].input_summary))
print("     summaries identical?", bash[0].input_summary == bash[1].input_summary)

# --- Bug 1b: str(input_data)[:80] collides distinct custom-tool calls ---
prefix = "/home/user/worktrees/feature-branch-alpha/services/api/internal/handlers"
patches = [tc("apply_patch", f"call_p{i}",
              {"input": f"*** Update File: {prefix}/handler_{i}.py\n@@\n-old\n+new\n"},
              msg_index=i) for i in range(5)]
d1b = detect_loops([SessionData(session_id="s2", tool_calls=patches)])
print("\n[1b] 5 DISTINCT apply_patch files ->", len(d1b), "loop(s)")
for lp in d1b:
    print(f"     count={lp.count} wasted={lp.wasted_tokens:,}")

# --- Bug 3: a builtin name with a foreign schema -> placeholder key ---
# `codebase_search`/`search_text`/`search_code` normalize to Grep but carry
# `query`, not `pattern`; `read_many_files` normalizes to Read but carries `paths`.
searches = [tc("Grep", f"call_s{i}", {"query": f"distinct query number {i}"}, msg_index=i)
            for i in range(5)]
d3 = detect_loops([SessionData(session_id="s3", tool_calls=searches)])
print("\n[3] 5 DISTINCT normalized-Grep calls ->", len(d3), "loop(s)")
for lp in d3:
    print(f"     count={lp.count} wasted={lp.wasted_tokens:,}")
print("     signatures:", {_canonical_signature(c) for c in searches})
```

## Expected Behavior

| Input | Expected |
|---|---|
| 5 `Bash` commands, 133 chars each, differing only after char 100 | 0 loops — 5 distinct commands |
| 5 `apply_patch` calls to 5 different files sharing a long path prefix | 0 loops — 5 distinct calls |
| 5 `Grep` calls normalized from `codebase_search`, each a different `query` | 0 loops — 5 distinct searches |

Output-limit variants of *one* command must of course still collapse — that collapse is the feature, and it has to survive any fix.

## Actual Behavior

```
headroom 0.36.5

[1a] 5 DISTINCT long Bash commands -> 1 loop(s)
     count=5 wasted=40,000
     command len: 133 -> input_summary len: 103
     summaries identical? True

[1b] 5 DISTINCT apply_patch files -> 1 loop(s)
     count=5 wasted=40,000

[3] 5 DISTINCT normalized-Grep calls -> 1 loop(s)
     count=5 wasted=40,000
     signatures: {'grep::?'}
```

This also reproduces end-to-end through the real `ClaudeCodePlugin` scanner on constructed transcripts, not just at the unit level: a session holding one genuine 5× re-fetch loop plus 5 distinct long-Bash calls and 5 distinct `apply_patch` calls, replayed once by a resume transcript, is reported as **3 loops / ~290,000 wasted tokens**, where only one loop is real and even that one is inflated (12× rather than 7×). That figure combines the truncation defects in this
issue with the replay double-counting now tracked in #3452 — the two compound, which is why I
measured them together. Full before/after transcript will be in the PR.

## Environment

- **Headroom version**: 0.36.5 (pipx). Also verified against `main` @ `e59cf10`: `headroom/learn/loops.py` is byte-identical to 0.36.5, and `models.py` differs only by an unrelated `analysis_error` field — `input_summary` is unchanged.
- **Python version**: 3.13
- **OS**: Ubuntu (WSL2)
- **LLM Provider**: Anthropic (Claude Code CLI); custom-tool transcripts from Codex CLI

## Proposed fix — PR incoming

Separate identity from display. `input_summary` stays the display helper it documents itself to be, and the loop signature gets its own identity function.

- **Bug 1** — derive the signature from the untruncated input: the identifying field per builtin tool (the full `command`, `file_path` or `pattern`), and key-sorted JSON of the whole `input_data` for anything else. The existing pagination/bare-int collapsing still runs on top, so re-fetch variants keep collapsing. `sample_input` keeps using `input_summary`, so digest lines are unchanged.
- **Bug 3** — fall back to that key-sorted JSON whenever the per-tool fields yield *nothing*, not only when the tool is absent from the table. `Grep` keyed on a present `pattern` still ignores `path`, so the defect corrected above is left exactly as found.

Both are behavior-preserving for the builtin path pinned by `tests/test_learn/test_loop_weighting.py`. I'm opening a PR with these plus regression tests (failing before, passing after); the `tool_call_id` dedup ships as its own PR against #3452. I'm filing this issue alongside the PR only because of the open questions below, which I don't think a PR should decide on its own.

## Open question — should repeated `Edit`/`Write` count as a loop?

Not included in the PR, because it's a judgment call rather than a defect.

`models.py:74-75` returns the bare `file_path` for `Edit`/`Write`, and the signature scheme has no notion of read-like vs mutating. So 12 distinct edits to one file — different `old_string`/`new_string` each time, ordinary iterative work — currently surface as one `refetch-loop` with attributed wasted tokens:

```
[4] 12 distinct edits to one file -> 1 loop(s)
    kind=refetch-loop count=12 wasted=5,500 sample='/repo/docs/plan.md'
```

Repeatedly *reading* one path is plausibly waste. Repeatedly *editing* one file is often just work. But I can see the argument that hammering one file is a signal worth surfacing, so: is this intended? If you'd like it changed, I'd restrict `refetch-loop` to read-like tools while leaving `error-loop` applicable to any tool — happy to fold that into the same PR or send it separately, whichever you prefer.

## Open question — where should input-schema normalization live?

Bug 3's fix is a fallback inside the loop signature: when the per-tool fields are missing, key on the whole input. That is correct locally, but it treats a symptom. `normalize_tool_name()` renames a provider's tool onto the builtin schema while leaving its input untouched, so *every* consumer that reads `input_data["file_path"]` or `["pattern"]` — `input_summary` included — may be reading a field that was never there. The alternative is to normalize the input alongside the name in `_shared.py`, so `read_many_files`' `paths` becomes `file_path` once, centrally, and the fallback becomes unnecessary.

There is a concrete payoff, which I only noticed while checking what the PR left alone. Because each builtin's field list is narrower than the input it keys, the generic fallback is uniformly *more* discriminating than the named-field path — `Grep{pattern, path}` ignores `path` while `Grep{query, path}` keys on it, and `Edit{file_path, old_string, new_string}` merges three distinct edits to one file while those same three under a foreign schema stay distinct. Every tool in the table behaves this way. Normalizing inputs alongside names removes that split at a stroke: every provider's call reaches the same field list, so each tool has exactly one identity rule instead of one per schema. The questions that remain — does `path` belong in `Grep`'s key, do repeated `Edit`s count as a loop — then get decided once, deliberately, rather than falling out of which CLI wrote the transcript.

I haven't done that here: it changes a contract every plugin depends on, and it's the kind of call I don't think a PR should make unilaterally. Happy to send it as a follow-up if you'd like it.

## Related

- #3452 — the replay double-counting defect, split out of this issue.
- #3453 — the `Grep`/`Glob` `path` defect described in the correction above, reported separately as promised there.
- #1159 — the feature this module came from (loop weighting). This is an accuracy follow-up, not a regression report against it.
- #1094 — adjacent, roughly the mirror image at runtime: over-*retrieval* when the kept summary suffices, versus over-*counting* in loop detection here.

njchae:
## Description

`headroom learn`'s loop detector groups tool calls by `ToolCall.input_summary`, a helper
whose own docstring says it is a *"Short summary of tool input for display."* It truncates
Bash commands at 100 characters (`models.py:67`) and renders any non-builtin tool as
`str(input_data)[:80]` (`models.py:76`). Using a deliberately lossy display value as an
*identity* key means calls that differ only after the cutoff compare equal and merge into
one phantom loop, so `learn` reports loops that never happened and attributes real tokens
to them.

A second, narrower form of the same collapse: `normalize_tool_name()` renames provider
tools onto the builtin names — `codebase_search` becomes `Grep`, `read_many_files` becomes
`Read` — without normalizing their *input*. `input_summary` then looks for `pattern` or
`file_path`, finds neither, and returns its `"?"` placeholder for all of them, so N
unrelated searches share the signature `grep::?` and merge into one loop. This one needs no
long input; the calls collide at any length.

The estimation math is not at fault — it sums real `output_bytes` correctly. The fault is in
the partition it sums over.

This is one of two PRs splitting #3440. The replay double-counting originally filed with it
is a different mechanism and ships separately against #3452; the two are independent and
either can land first. They conflict textually in `detect_loops` — this PR renames the merge
loop's variable, that one rewrites its body — so whichever lands second needs a ~6-line
rebase. Happy to do it either way round.

Closes #3440

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/learn/loops.py`: added `_IDENTITY_FIELDS` and `_identity_input()`. The loop key
  now comes from the untruncated input — the identifying field per builtin tool
  (`command` / `file_path` / `pattern`), and key-sorted `json.dumps` for everything else —
  instead of `input_summary`.
- `headroom/learn/loops.py`: `_identity_input()` falls back to the key-sorted JSON
  whenever the per-tool fields yield *nothing*, not only when the tool is absent from
  `_IDENTITY_FIELDS`. Without this, a builtin name carrying a foreign schema keys on an
  empty string and every such call merges into one loop — the same defect, reintroduced in
  the new code. `Grep` keyed on a *present* `pattern` still ignores `path`, so the separate
  defect now filed as #3453 is left exactly as found. The same branch means an unexpected
  input shape returns a signature rather than raising: identity is derived from files the
  user did not write.
- `headroom/learn/loops.py`: identity and the *stored* signature are kept separate.
  Groups are keyed on a `blake2b` digest of the full signature, so the grouping tables
  stay a fixed size per loop no matter how large the input is. `LoopPattern.signature`
  keeps its body capped at the width `input_summary` previously imposed. That cap is
  load-bearing: `apply_loop_weighting` requires a *majority* of the signature's tokens to
  appear in a recommendation, so an untruncated signature puts the threshold out of reach
  for long inputs and silently costs a real loop its measured-waste boost. I hit this and
  added regression tests for it — see Additional Notes.
- `headroom/learn/loops.py`: whitespace collapsing switched from `re.sub(r"\s+", " ", raw)`
  to `" ".join(raw.split())`. Same output, roughly 4x cheaper on the now-untruncated input.
  See Performance for where the remaining cost actually sits — it is not here.
- The existing pagination / bare-int collapsing still runs on top of the new key, so
  re-fetch variants of one command keep collapsing exactly as before.
- `tests/test_learn/test_loop_weighting.py`: 11 new tests, 17 cases with parametrization.
  I ran them against `upstream/main` rather than assume the split: **7 fail there and pass
  here**; the other 10 pin behavior that must survive — pagination and >100-char command
  variants must still collapse, whitespace variants must still fold together, detection
  must not depend on call order, and weighting must hold at four input sizes. The weighting
  and signature-size ones guard against a mistake I made building this (see Additional
  Notes) rather than against the reported bug, so they pass on `main` too.

`ToolCall.input_summary` itself is untouched and still used for display (`sample_input`,
digest lines), so digest output formatting does not change.

## Testing

### Test Output

```text
$ pytest -q tests/test_learn/
264 passed, 4 skipped

$ pytest -q tests/test_learn/test_loop_weighting.py
30 passed          # 13 pre-existing + 17 added by this PR

$ ruff check .
All checks passed!

$ ruff format --check headroom/learn/ tests/test_learn/
30 files already formatted

$ mypy headroom
# The one error is pre-existing and in a file this PR does not touch:
# headroom/integrations/langchain/chat_model.py:394 [truthy-function]
# Zero errors in headroom/learn/loops.py.

$ pytest -q tests/test_learn/ \
    --cov=headroom.learn.loops --cov=headroom.learn.fixtures --cov-report=term-missing
Name                         Stmts   Miss Branch BrPart  Cover   Missing
headroom/learn/fixtures.py      27      3     10      1    84%   33-36
headroom/learn/loops.py        104      0     36      1    99%   296->288
# fixtures.py is untouched by this PR; its uncovered lines predate it.
# Every statement in loops.py is covered.

$ pytest -q          # full suite, this branch
21 failed, 11916 passed, 583 skipped, 6044 warnings in 791.19s (0:13:11)

$ pytest -q          # same suite, unpatched upstream/main @ e67b3c8a
21 failed, 11899 passed, 583 skipped, 6044 warnings in 651.63s (0:10:51)
```

**The 21 failures are pre-existing and identical on both sides.** I ran the entire suite on
this branch and on `upstream/main` @ `e67b3c8a`, and diffed the sorted `FAILED` lists:
empty diff. The only difference between the two runs is the passing count — `11916` here
versus `11899` on the base — which is exactly the 17 test cases this PR adds. No
existing test changes outcome.

The pre-existing failures are `test_integrations/langchain` (7 — wants a live local
Ollama), `test_proxy_health` (4), `test_cli/test_wrap_copilot` (3),
`test_mcp_registry/test_claude_registrar` (2), `test_install/test_providers` (2),
`test_cli/test_wrap_bridge` (1), `test_cli/test_unwrap_claude` (1) and
`test_scripts/test_repro_codex_replay_smoke` (1). None touch `headroom/learn/`.
The 4 `test_proxy_health` ones pass when that file is run on its own (`17 passed`) and fail
only in a full-suite run, on the base as well as here, so they look order-dependent —
flagging it in case it is news, but it is not something this PR introduces or fixes.

## Real Behavior Proof

- Environment: Ubuntu 22.04.5 LTS on WSL2 (Linux 6.18.33.2), Python 3.13.11,
  ruff 0.16.3. Branch `8f3a58b2` on `upstream/main` `e67b3c8a` (`headroom 0.38.0-dev`).
  The "before" side is `upstream/main` @ `e67b3c8a` itself. I also ran it against the
  released `headroom-ai 0.36.5` from pipx as a cross-check and got identical output, which
  is expected: `learn/loops.py` is byte-identical between the two. No LLM/provider is
  involved in the code path under test.

- Exact command / steps: two constructed Claude transcripts are written to disk
  (a main session plus a resume that replays it), then scanned through the *real*
  production path — `ClaudeCodePlugin.scan_project()` → `detect_loops()` →
  `_build_digest()`. The same two files are scanned twice, once by each interpreter:

  ```bash
  # before (unpatched upstream/main)
  PYTHONPATH=<main worktree> python run_proof.py data
  # after (this branch)
  PYTHONPATH=<this branch>   python run_proof.py data
  ```

  Ground truth planted in the fixtures: **one** genuine re-fetch loop — the same file read
  5 times with a growing `head -N` limit, which the existing pagination-collapsing rule is
  supposed to merge — extended by 2 further repeats in the resume, so **7**. Also present
  and all mutually distinct: 5 long `rg` commands that diverge only after character 100,
  and 5 `apply_patch` calls sharing an 80-character path prefix. Neither of those is a loop.

- Observed result: both phantom loops disappear, leaving only the real (still replay-inflated) loop; full before/after `_build_digest()` output below.

  ```text
  before (upstream/main):  loops detected: 3
    [refetch-loop] Bash:        count=12 wasted=110,011   <- 1 real loop, replay-inflated
    [refetch-loop] Bash:        count=10 wasted=90,009    <- 5 distinct rg cmds, phantom
    [refetch-loop] apply_patch: count=10 wasted=90,009    <- 5 distinct patches, phantom
                                total wasted ~290,029

  after (this branch):     loops detected: 1
    [refetch-loop] Bash:        count=12 wasted=110,011
  ```

  Both phantom loops disappear — that is this PR's entire job. The surviving loop is the
  real one, and its count is **still inflated**: the fixtures include a resume transcript
  that replays every call under its original `tool_use_id`, and de-duplicating that is
  #3452's fix, not this one. With both PRs applied the same scenario reports
  `loops detected: 1, count=7 wasted=60,006`, which is the ground truth. I am reporting
  this PR's effect in isolation rather than quoting the combined figure.

  `_build_digest()` output is quoted as-is — verbatim what the LLM step would receive.

- **Aggregate check against real history.** The constructed transcripts prove the mechanism;
  to confirm the shape actually occurs, I ran the real `ClaudeCodePlugin` over my own 187
  Claude Code transcripts — 166 sessions, 19,188 tool calls. Counts only, no paths, commands
  or session content, for the reason given under Not tested.

  The finding that matters here is a property of the corpus rather than of any patch:
  **80% of the 8,623 real Bash commands exceed the 100-character cutoff.** So the collapse
  this PR fixes is the common case on the default Claude Code path, not a tail case
  reachable only through custom-tool providers.

  Caveat on the summary table: my aggregate A/B was run with both this change and #3452's
  applied together (749 loops / 5,691,304 wasted tokens on `main`, versus 547 / 3,446,209
  with both), so I am not attributing that delta to this PR alone. Happy to re-run the
  A/B against this branch in isolation if the split number is wanted for review.

- **Performance.** `_identity_input` serializes the whole input instead of a truncated
  slice, so I measured it rather than guess. Worst case is the non-builtin path, which
  JSON-serializes the entire input; `detect_loops` over 2000 calls, one payload each:

  ```text
  payload      main      this PR
    1 KB     5.7 ms      7.8 ms     <- 1.4x
   10 KB    30.5 ms     43.7 ms     <- 1.4x
   40 KB   116.7 ms    175.3 ms     <- 1.5x

  peak allocation @40 KB      0.17 MB -> 2.38 MB
  signature bytes retained  0.0046 MB -> 0.0056 MB
  ```

  The residual is inherent to a lossless key: `json.dumps` plus the hash account for
  nearly all of it. Bounding the stored signature keeps retained bytes flat; without the
  cap, retained signatures grew to 82 MB on this benchmark.

  Two notes on where the cost sits, since it is easy to draw the wrong conclusion from the
  table above. Switching whitespace collapsing to `" ".join(str.split())` is worth roughly
  4x on that step, but the step is only dominant on the **custom-tool** path. On the
  **Bash** path the pagination substitution — which now runs over the untruncated command
  rather than a 100-character slice — is the larger term, and this PR does not reduce it.
  Per call on a 40 KB Bash command: `_PAGINATION_RE` 0.912 ms, `_INT_RE` 0.312 ms,
  whitespace 0.119 ms (was 0.495 ms). If you benchmark the Bash path you will see that
  shape, not the custom-tool one.

  Second, worst case is not the common case. Across 166 real transcripts on my machine
  (19,188 calls, 8,623 of them Bash) the median Bash command is 203 characters and p95 is
  1,223 — three orders of magnitude below the synthetic 40 KB payload — and end-to-end
  `detect_loops` over all of them goes 64.5 ms to 180 ms. This is `learn`, an offline
  command that reads history files, not a proxy transform, so it is not on the `<50ms`
  P99 path.

- Not tested: the LLM step itself, any quoted output from my own history, non-Claude scanners, and the perf table's scope — each detailed below.
  - The LLM step itself. `headroom learn` has no flag that halts before the model call, so
    the proof drives scanner → `detect_loops` → `_build_digest` directly. That is the whole
    production path minus the model, and `_build_digest` output is the model's literal input.
  - Any *quoted* output from my own history. The before/after transcript above is
    constructed, deliberately: loop signatures embed absolute repo paths and full shell
    commands, and I am not willing to publish mine. The code path exercised is production
    code; only the input files are synthetic. Real history is used in the aggregate check
    above, but only as counts — no signature, path, command or session content.
  - The non-Claude scanners (`codex`, `gemini`, `grok`, `opencode`) beyond their existing
    test suites. The change is in provider-agnostic `learn/loops.py`, so they share it, but
    I did not build per-provider fixtures.
  - The performance table is synthetic payloads on one machine (WSL2), chosen as a worst
    case; the 64.5 ms → 180 ms figure beside it is a real-history run on the same machine.
    I did not measure the scanner or the LLM step around `detect_loops`, so both bound the
    cost of the change, not the cost of `learn` overall.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. `learn` loop detection is not behind a rollout
  channel or feature flag.
- Stable/default behavior changed: yes, and intentionally — loop grouping and therefore
  reported loop counts and `wasted_tokens` change. This is the fix. No public API, CLI flag,
  config key, or digest format changes; `ToolCall.input_summary` keeps its current behavior
  for display.
- Kill switch / disable path: none added. Reverting the single commit fully restores
  prior behavior.
- Qualification impact: `learn` runs will report fewer, more accurate loops, so the
  generated guardrail text changes. Nothing persisted by earlier runs is read back or
  migrated, so old and new outputs do not need to agree.
- Rollback path: `git revert` of one commit. No schema, cache, or on-disk state is
  touched.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A, see Additional Notes
- [x] I did **not** edit `CHANGELOG.md` — it is generated by release-please from my
      Conventional Commit PR title (a CI guard enforces this)

N/A — terminal output is included above.

## Additional Notes

- **Documentation:** no user-facing surface changed (no new flag, config, or output format),
  and `docs/failure-learning` does not document the grouping key, so there is nothing to
  update. Happy to add a note if you would like the identity-vs-display distinction written
  down.

- **The subtlest part of this change, called out for review.** Making the grouping key
  lossless is only half of it. `LoopPattern.signature` is consumed in exactly one place —
  `_signature_tokens` in `apply_loop_weighting` — and that consumer does *fuzzy* matching:
  it needs a majority of the signature's word tokens to appear in a recommendation's text.
  Truncation used to bound that token set as a side effect. My first cut removed the
  truncation everywhere, which quietly broke weighting: a repeated long command that
  previously had its guessed 100 tokens raised to its measured 40,000 stopped matching at
  all and fell back to the guess, ranking below one-off rules. Nothing failed; the number
  just got worse. So identity now comes from the full input while the stored signature
  stays bounded, and two of the tests pin exactly that split. If you would rather see
  `apply_loop_weighting` made length-insensitive instead, that is a reasonable alternative
  — I kept to the bound because it is bounded the same way so majority-match stays
  reachable. (Not bit-identical to the old path: truncation used to happen via
  `input_summary` *before* pagination/int collapse and appended `...`; the stored
  signature now collapses the full command first, then takes `body[:100]`. Fuzzy
  token sets can differ at the margin; directionally that is better.)

- **The fallback is deliberately more discriminating, called out for review.**
  `_identity_input` keys on a per-tool field list when it recognises the tool, and on
  key-sorted JSON of the whole input when it does not. Every field list in the table is
  narrower than the input it keys, so the fallback always distinguishes calls that the
  named-field path merges — `Grep{query, path}` separates by `path` where
  `Grep{pattern, path}` does not, and the same holds for `Bash`, `Edit` and `Write` against
  a foreign schema. This is accepted, not overlooked: a fallback cannot know which fields
  carry meaning, and erring toward distinctness is the conservative direction (it under-
  reports loops rather than inventing them, matching the "prefer false negatives" principle
  in CONTRIBUTING). Making the two paths agree means either deciding per-tool which sibling
  fields belong in identity — see #3453 and the `Edit`/`Write` question below — or
  normalizing inputs in `_shared.py` alongside the name mapping, which is #3440's second
  open question. Both are out of scope here; I did not want the asymmetry to read as an
  oversight in review.

- **A related defect deliberately *not* fixed here.** `Grep`/`Glob` identity drops `path`,
  so one pattern searched in two different directories merges — `grep::timeouterror` on
  both sides, before and after. It predates this PR (reproduced on released `0.36.5`) and
  this PR leaves it exactly as found: `_identity_input` only falls back to the generic key
  when the per-tool fields yield *nothing*, and `pattern` is present here. Fixing it means
  deciding whether two searches of one pattern in different trees are the same call — a
  design call, not a contract violation. Now filed as #3453 with a repro, rather than
  widened into this change.

  Its sibling — identity read from an *absent* field, where `normalize_tool_name()` renames
  a foreign tool onto a builtin whose key it does not carry — I did fix here, because the
  distinction matters: there is no design question to settle. The generic key-sorted-JSON
  path already exists and is unambiguously the right key when the per-tool guess misses,
  and leaving it would have meant shipping a brand-new `_identity_input()` with a known
  collapse-to-empty hole in it. #3440 carries this as Bug 3, with a second open question
  on whether input-schema normalization belongs in `_shared.py` alongside the name mapping,
  which would make the fallback unnecessary. That part I have not done — it changes a
  contract every plugin depends on.

- **One thing I deliberately did not fix.** `input_summary` returns the full `file_path`
  for `Edit`/`Write`, so repeated *legitimate* edits to one file group as a "loop"
  (I have seen 179 edits to a single file reported this way). For reads, a same-path repeat
  is plausibly waste; for mutations it is just normal work. Fixing it means teaching the
  signature scheme which tools are read-like versus mutating, which is a design call rather
  than a contract violation — so I have left it alone and raised it as an open question in
  #3440 rather than smuggling it into this PR. Happy to follow up if you want it changed.

- **Scope, stated plainly.** The collapse bites hardest on providers whose tools fall to
  the `[:80]` branch, and on any tool `normalize_tool_name()` renames onto a builtin whose
  schema it does not share. On the default Claude builtin path `Read`, `Edit` and `Write`
  already returned faithful identities, so only >100-character Bash commands were exposed
  there — which, per the aggregate check above, is 80% of real Bash calls.

- **`mypy headroom`** reports 1 pre-existing error in `integrations/langchain/chat_model.py`,
  untouched by this PR. I left it alone rather than mix an unrelated fix into this change.


JerrettDavis (commented):
All 30 loop-weighting tests pass, including long inputs, provider-schema fallback, and weighting behavior. I found an integration hazard with #3455: this PR no longer reads input_summary for identity, while #3455 fixes search-path identity solely in input_summary. Combining those behaviors groups Grep(pattern=TimeoutError) in /srv/ingest, /srv/billing, and /srv/web into one count=3 loop with 20,000 wasted tokens again. Please coordinate the two changes so the explicit Grep/Glob identity includes path when #3455 lands, and retain its different-path regression in the combined tests. Neither PR should silently undo the other during the main refresh. No separate blocking finding against this PR current base.

njchae:
Good catch. Easier to do the merge than argue about it, and it does break.

Take #3455, merge it into where this branch was before, and #3455's own `test_search_signatures_include_path` fails. Identity wasn't reading `input_summary` anymore, so `/srv/ingest`, `/srv/billing` and `/srv/web` all came out as `grep::timeouterror`. One loop, count=3, 20,000 tokens, same as you described.

`40f2be11` puts `path` into `_IDENTITY_FIELDS` for `grep` and `glob` so the two sides agree. Merge is clean after that and your test passes untouched, 274 total.

I kept a regression on the identity side as well (`TestSearchIdentityIncludesPath`), including the case where `Grep` has no path at all and shouldn't collapse into a rooted one. Your display assertion isn't affected either way, `sample_input` still reads `input_summary`.

Something for next time, since identity doesn't go through `input_summary` anymore: a new distinguishing field added to the summary won't split the groups on its own. There's a comment on `_IDENTITY_FIELDS` about it now.

JerrettDavis (changes requested):
The explicit search path resolves the earlier #3455 integration concern, and all 37 loop-weighting tests pass. One identity ambiguity remains in the new two-field representation:

[P2] Preserve field boundaries when rendering pattern and path. `_identity_input()` joins them with a space, so distinct valid searches `(pattern="error in src", path="logs")`, `(pattern="error in", path="src logs")`, and `(pattern="error", path="in src logs")` all become `grep::error in src logs`. I reproduced one count=3 loop with 20,000 reported wasted tokens for those three different searches. Before adding path to this concatenation, their different patterns remained separate. Please encode the selected fields unambiguously (for example, a structured tuple/JSON representation) and add a regression for patterns and paths containing spaces, while retaining the shell-pagination normalization.

JerrettDavis (approved):
The field-boundary collision is resolved. Re-running the original three-search probe now produces three distinct structured signatures and no false loop. Multi-field inputs retain JSON boundaries, while single-field shell commands still use the existing pagination normalization. Unicode text is preserved for recommendation matching. All 51 loop-weighting tests pass, including repeated spaced paths, separator-like inputs, and shell pagination variants.

JerrettDavis (approved):
Re-reviewed after the main merge/conflict resolution. The full-input identity, structured multi-field encoding, bounded fuzzy signature, and replay de-duplication integration remain coherent; shell pagination normalization is preserved, and the merge adds no unrelated contributor logic. The head is current, mergeable, and all 38 checks pass. No new blocking finding.

PR #3464, opened 2026-09-07
fix(proxy): support Codex Live HTTP call creation

### LINKED ISSUE #3463: [BUG] Codex Desktop Live voice POST /v1/live loses Location through proxy
## Description

Codex Desktop Live voice cannot start when ChatGPT subscription traffic is routed through Headroom 0.37.0.

The Codex client sends an HTTP `POST /v1/live` request with a multipart body containing `sdp` and `session`. Headroom handles `/v1/live` only as a WebSocket route, so the HTTP request falls through to the generic OpenAI passthrough. In the reproduced request, the passthrough returns `302`, the client follows the redirect to `/`, and the final response has no `Location` header. Codex then reports:

```text
stream disconnected before completion: realtime call response missing Location
```

Normal Codex Responses traffic continues to work through the same proxy.

## To Reproduce

1. Install Headroom 0.37.0 as a uv tool with proxy support.
2. Run the persistent proxy on `127.0.0.1:8787`:

   ```text
   python -m headroom.cli proxy --host 127.0.0.1 --port 8787 --mode cache --backend anthropic --no-telemetry
   ```

3. Configure Codex Desktop to use Headroom:

   ```toml
   model_provider = "headroom"
   openai_base_url = "http://127.0.0.1:8787/v1"

   [model_providers.headroom]
   name = "OpenAI via Headroom proxy"
   base_url = "http://127.0.0.1:8787/v1"
   supports_websockets = true
   requires_openai_auth = true
   ```

4. Sign in to Codex Desktop with ChatGPT subscription OAuth.
5. Confirm that an ordinary Codex conversation works through Headroom.
6. Start a new Live voice conversation.
7. Observe the voice overlay fail before the session starts.

## Expected Behavior

Headroom should transparently proxy the Codex WebRTC call-creation request to the ChatGPT Codex backend and return the upstream SDP response together with its `Location` header.

## Actual Behavior

The HTTP `POST /v1/live` request enters the generic passthrough and returns `302`. The client follows the redirect to `/`, receives a response without `Location`, and terminates the realtime session.

## Code Sample

This reproduces in Codex Desktop rather than through the Headroom Python client. The minimal Codex configuration is included in the reproduction steps above.

The relevant request shape observed at the proxy is:

```http
POST /v1/live HTTP/1.1
Content-Type: multipart/form-data; boundary=codex-realtime-call-boundary

<sdp and session form fields omitted>
```

## Error Output

Codex Desktop:

```text
stream disconnected before completion: realtime call response missing Location
```

Headroom proxy log, with headers and identifiers omitted:

```text
event=proxy_inbound_request method=POST path=/v1/live content_type=multipart/form-data
event=proxy_inbound_response method=POST path=/v1/live status=302
event=proxy_inbound_request method=GET path=/ referer=http://127.0.0.1:8787/v1/live
```

## Environment

- **Headroom version**: 0.37.0
- **Python version**: 3.14.3
- **OS**: Ubuntu 24.04.4 LTS, x86_64
- **LLM Provider**: OpenAI Codex Desktop with ChatGPT subscription OAuth
- **Codex Desktop version**: 26.901.51231
- **Codex client version reported in the proxy log**: 0.153.4
- **Installation**: uv tool, persistent user service

## Additional Context

- Related issue: #2653. A follow-up comment there reports the same missing-`Location` error on Headroom 0.34.0.
- Related fix: #2709 added a dedicated `/v1/live` WebSocket transport. The current failure is the HTTP WebRTC call-creation request on the same path, not the WebSocket relay.
- The current `proxy_routes.py` registers `CODEX_LIVE_ROUTE_PATHS` with `app.websocket(...)`; there is no corresponding HTTP `POST` handler.
- Codex uses the API-style multipart request when the configured base URL does not contain `/backend-api`. The ChatGPT backend uses a JSON body at `/backend-api/codex/realtime/calls?intent=quicksilver&architecture=avas`.
- A local prototype intercepts `POST /v1/live`, parses the `sdp` and `session` fields, forwards them as JSON to the ChatGPT backend route, and preserves the upstream response headers. Its isolated conversion/header-preservation test passes, and the patched module passes `py_compile`.
- A live Codex Desktop round trip after that prototype patch has not yet been verified.
- No authorization token, account identifier, raw SDP, session body, or unredacted request headers are included in this report.

gingeekrishna:
## Description

Codex Desktop sends WebRTC call creation as a multipart `POST /v1/live`. Headroom only registered the path as a WebSocket route, so the HTTP request fell through to the generic passthrough and the client lost the upstream `Location` header.

Closes #3463

## Type of Change

## Changes Made

- Register HTTP POST handlers alongside the existing Codex Live WebSocket aliases.
- Convert the `sdp` and JSON-encoded `session` multipart fields to the ChatGPT Codex realtime call payload.
- Preserve the upstream status, body, and headers, including `Location`.
- Keep non-ChatGPT-authenticated requests on the existing OpenAI passthrough path.
- Add regression coverage for the forwarding contract.

## Testing

- [x] Linting passes for touched files (`python -m ruff check headroom/providers/codex/live.py headroom/providers/proxy_routes.py tests/test_codex_live.py`)

### Test Output

```text
python -m py_compile headroom/providers/codex/live.py headroom/providers/proxy_routes.py tests/test_codex_live.py
passed
python -m ruff check headroom/providers/codex/live.py headroom/providers/proxy_routes.py tests/test_codex_live.py
All checks passed
python -m ruff format --check headroom/providers/codex/live.py headroom/providers/proxy_routes.py tests/test_codex_live.py
3 files already formatted
python -m pytest tests/test_codex_live.py tests/test_provider_proxy_routes.py -q
blocked during collection: ModuleNotFoundError: No module named 'headroom._core'
```

## Real Behavior Proof

- Environment: Windows checkout; Python environment available, but the compiled `headroom._core` module is not installed in this worktree.
- Exact command / steps: Ran the focused pytest command, syntax compilation, Ruff lint, and Ruff format checks listed above.
- Observed result: Syntax, lint, and format checks passed. Pytest could not collect tests because the native extension was unavailable.
- Not tested: Live Codex Desktop round trip and full pytest suite, because the native extension was unavailable and no live credentials were used.

## Runtime Rollout Safety

- Minimum rollout channel: Default behavior for the existing `/v1/live` path.
- Stable/default behavior changed: Only HTTP POST requests with ChatGPT Codex authentication use the new dedicated handler.
- Kill switch / disable path: Revert the commit.
- Qualification impact: None.

## Review Readiness

## Checklist

## Additional Notes

The full test suite still needs to run in CI or an environment with the compiled `headroom._core` extension.


JerrettDavis (changes requested):
The 31 focused tests pass, but real Starlette request/HTTPX response reproductions expose three forwarding problems that should be addressed before this route is ready.

1. Strip internal headers before the upstream POST. Unlike the WebSocket path, handle_codex_live_http forwards decision.headers without _strip_internal_headers; a synthetic X-Headroom-Proxy-Token reaches the captured ChatGPT-bound request unchanged. Please apply the established upstream header filtering and add a regression asserting that the proxy credential and other x-headroom headers never leave the proxy while provider authorization/account routing survive.

2. Resolve whether this is ChatGPT traffic before parsing/validating multipart content. This confirms the existing Copilot fallback finding: a real non-ChatGPT multipart request reaches the fallback with its stream consumed and await request.body() raises RuntimeError: Stream consumed; a JSON request receives the new Missing sdp or session 400 instead of passthrough. Leave the body untouched when declining this handler, and exercise the actual registered route and fallback with Request objects rather than only FakeRequest.form.

3. Normalize upstream response headers to match the bytes being returned. HTTPX response.content is decoded; copying content-encoding/content-length from a gzip response onto those decoded bytes makes the downstream client decompress again (reproduced as httpx.DecodingError) and can leave the wrong length. Use the shared response-header sanitation pattern, retain Location, and let Response compute the decoded content length. Add a compressed upstream response regression.

The current Docker CI failures are the separate OpenClaw npm edgesOut installation failure, not evidence for or against these HTTP behaviors.

gingeekrishna:
@JerrettDavis All three fixed, and I reproduced each one on the pre-fix code first to confirm the diagnosis before changing anything:

1. **Internal header leak** — \`upstream_headers\` now goes through \`_strip_internal_headers()\` before \`resolve_codex_routing()\`, matching the pattern already used in \`_handle_chatgpt_codex_alpha_search\`. Verified a synthetic \`X-Headroom-Proxy-Token\`/\`X-Headroom-Bypass\` no longer reach the outbound call, while \`Authorization\`/\`ChatGPT-Account-ID\` still do.

2. **Body consumed before the routing decision** — routing is now resolved from headers alone; the handler only calls \`request.form()\` after confirming ChatGPT auth. Reproduced both symptoms pre-fix: a real multipart request without ChatGPT auth hit \`RuntimeError: Stream consumed\` in the passthrough fallback, and a JSON body got the handler's own 400 instead of falling through. Both now return 200 via the real upstream (tested against the actual registered \`/v1/live\` route + fallback through a loopback HTTP server, not just the handler function directly).

3. **Stale compression headers on decoded content** — switched to \`sanitize_forwarded_response_headers\`, the same helper the Anthropic/OpenAI handlers already use for exactly this reason. \`Content-Encoding\`/\`Content-Length\`/\`Transfer-Encoding\` are dropped so Starlette computes the correct length for the decoded body; \`Location\` and everything else still comes through.

New tests: a header-stripping/response-sanitation regression on the handler directly, plus two end-to-end tests that exercise the real registered route and fallback through a loopback HTTP upstream (one multipart, one JSON) — both confirmed to fail with the exact errors you described before the fix and pass after it.

JerrettDavis (approved):
Re-reviewed the latest forwarding changes and reran the original adversarial reproduction. All three earlier findings are resolved: internal proxy credentials are stripped, non-ChatGPT multipart/JSON requests retain their bodies for passthrough, and decoded upstream responses no longer carry stale encoding/length headers. The 10 Codex Live tests pass, including real HTTP/WebSocket integration paths. No remaining blocking findings in this change.

PR #3466, opened 2026-09-07
perf(memory): reindex FTS5 a page at a time instead of a record at a time

trs-80:
## Description

`headroom memory reindex` clears `memory_fts` and rebuilds it with one `index_memory()` call per record. `FTS5TextIndex.index_raw()` commits per call, so rebuilding a 1,000-memory store costs 1,000 transactions and 1,000 fsyncs. The adapter already ships `index_batch_memories()`; this uses it.

Closes # N/A

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `reindex_memories()` now calls `index_batch_memories()` once per page, reusing the existing `_REINDEX_PAGE_SIZE` constant rather than introducing a new knob. No adapter changes.
- On a failed page, the previous per-record loop runs **for that page only**, so valid records still land and the failing record is still named in the warning.
- A page that fails and then fully recovers does not change the exit status. `ok` stays tied to whether records actually got indexed, exactly as before.
- Added `benchmarks/memory_fts_reindex_benchmark.py` for reproduction.

**Why per page rather than one batch over the whole store.** A single transaction would be marginally faster, but the reindex deliberately clears the index before rebuilding (`DELETE FROM memory_fts`, `headroom/cli/memory.py:1208-1210`). One failed statement would then roll the whole rebuild back to an *empty* index — silently destroying search coverage while reporting a clean-looking failure. A page bounds both journal growth and blast radius, and keeps essentially all of the speedup.

## Testing

Six new tests in `tests/test_cli_memory_index_sync.py`, reusing that file's existing helpers and its `_REINDEX_PAGE_SIZE` monkeypatch pattern:

1. `..._batches_pages_instead_of_indexing_record_by_record` — asserts the observed page sizes are `[2, 2, 1]` and fails outright if the per-record path is touched.
2. `..._batched_output_matches_per_record_output` — every stored column (`memory_id`, `content`, `user_id`, `session_id`, `category`) and the FTS query result ordering are identical either way.
3. `..._empty_store_indexes_nothing_and_succeeds`.
4. `..._indexes_more_records_than_one_page`.
5. `..._failed_batch_still_indexes_valid_records_in_that_page` — the failure mode the design is built around.
6. `..._failed_record_is_named_and_exits_nonzero` — per-record diagnosability and exit status survive batching.

Tests 1 and 5 fail on the base commit and pass with this change.

### Test Output

```text
$ pytest tests/test_cli_memory_index_sync.py tests/test_memory/ -q
477 passed, 145 skipped in 10.38s

$ pytest tests/test_memory/ tests/test_cli_memory_index_sync.py tests/test_cli/ -q
1212 passed, 146 skipped, 2 failed
# Both failures (test_unwrap_claude_stops_claude_owned_persistent_deployment,
# test_wrap_codex_prepare_only_updates_config) also fail on the base commit in
# this environment: they read ambient CODEX_HOME / ANTHROPIC_BASE_URL from the
# developer's shell rather than isolating it. Unrelated to this change.

# The two behavioral tests against the base commit:
$ git stash push headroom/cli/memory.py
$ pytest tests/test_cli_memory_index_sync.py -q -k "reindex_batches or failed_batch"
2 failed

$ pre-commit run   # via commit hooks
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), Python 3.13.11, stdlib `sqlite3` with FTS5, worktree off `upstream/main` @ `e67b3c8a`.
- Exact command / steps: `python benchmarks/memory_fts_reindex_benchmark.py --records 1000 --runs 3` for the indexing operation; separately, a `CliRunner` harness timing the whole `memory reindex` command over a 1,000-memory store, run on this branch and then with `headroom/cli/memory.py` stashed. Each run uses a fresh temporary database.
- Observed result: the indexing operation alone goes from **759.7 ms to 8.1 ms** (median of 3, ~94x). The **whole `memory reindex` command** goes from **733.5 ms to 15.8 ms** (median of 3, ~46x) — reported separately because the command also pages the primary store and does vector-orphan cleanup, which this change does not touch. The audit that motivated this measured 778.69 ms vs 5.60 ms for the operation; reproduced within noise.
- Not tested: real-world stores larger than 1,000 records, concurrent access during reindex, and behavior on network filesystems where fsync costs differ. Synthetic records (~40 tokens each) rather than production memory content. No end-to-end proxy or agent-latency claim is made.

## Runtime Rollout Safety

- Stable/default behavior changed: no. Same records indexed, same fields, same FTS results, same output lines, same exit status. Only the transaction granularity changes.
- Kill switch / disable path: none needed. If the batch path raises for any reason, the command falls back to the previous per-record behavior automatically, per page.
- Qualification impact: none.
- Rollback path: revert the commit. `reindex` is idempotent — it clears and rebuilds the index from the primary store — so re-running it on any build produces the same result regardless of which version wrote the index.

## Review Readiness

## Checklist

## Additional Notes

- No user-facing docs describe this internal rebuild path, so "documentation update" is satisfied by the comment explaining why the batch is bounded to a page rather than the whole store — that constraint is the non-obvious part of this diff.
- Follow-up, deliberately not in this PR: the two failing tests noted above depend on the developer's ambient `CODEX_HOME` and `ANTHROPIC_BASE_URL`, so they pass or fail depending on whose shell runs them. That is a test-isolation bug worth its own change.

### DIFF
--- a/benchmarks/memory_fts_reindex_benchmark.py
+++ b/benchmarks/memory_fts_reindex_benchmark.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python3
+"""Measure FTS5 reindex throughput: per-record commits vs batched transactions."""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import statistics
+import sys
+import tempfile
+import time
+from pathlib import Path
+
+# Run as a script sys.path[0] is benchmarks/, so an editable install of another
+# checkout would win and we would silently measure the wrong tree.
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+from headroom.memory.adapters.fts5 import FTS5TextIndex
+from headroom.memory.models import Memory
+
+
+def make_memories(count: int, content_words: int) -> list[Memory]:
+    body = " ".join(f"token{n}" for n in range(content_words))
+    return [
+        Memory(
+            id=f"mem-{index:06d}",
+            content=f"{body} unique{index}",
+            user_id="bench-user",
+            session_id=f"session-{index % 7}",
+        )
+        for index in range(count)
+    ]
+
+
+def benchmark(count: int, content_words: int, page_size: int, runs: int) -> dict[str, object]:
+    memories = make_memories(count, content_words)
+    per_record: list[float] = []
+    batched: list[float] = []
+
+    for _ in range(runs):
+        # Fresh temporary database per mode per run so neither inherits the
+        # other's page cache or index state.
+        with tempfile.TemporaryDirectory(prefix="headroom-fts-bench-") as tmp:
+            fts = FTS5TextIndex(db_path=str(Path(tmp) / "memory.db"))
+            start = time.perf_counter()
+            for memory in memories:
+                asyncio.run(fts.index_memory(memory))
+            per_record.append((time.perf_counter() - start) * 1000)
+
+        with tempfile.TemporaryDirectory(prefix="headroom-fts-bench-") as tmp:
+            fts = FTS5TextIndex(db_path=str(Path(tmp) / "memory.db"))
+            start = time.perf_counter()
+            for offset in range(0, len(memories), page_size):
+                asyncio.run(fts.index_batch_memories(memories[offset : offset + page_size]))
+            batched.append((time.perf_counter() - start) * 1000)
+
+    return {
+        "records": count,
+        "content_words": content_words,
+        "page_size": page_size,
+        "runs": runs,
+        "per_record_ms": statistics.median(per_record),
+        "batched_ms": statistics.median(batched),
+        "speedup": round(statistics.median(per_record) / statistics.median(batched), 1),
+    }
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--records", type=int, nargs="+", default=[1000])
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
Reviewed page-level FTS5 indexing, transaction rollback behavior, and per-record fallback after a failed batch. All 30 CLI index-sync tests pass, including batch/per-record parity, multi-page coverage, empty stores, and partial failure status. The supplied 1000-record benchmark measured median 546.7 ms per-record versus 5.5 ms batched over three runs locally (99.2x). No blocking correctness issue found.

JerrettDavis (approved):
Re-reviewed the rebased head on current main. The patch remains the previously validated page-transaction reindex implementation: successful batches count exact rows, failed pages roll back and retry per record, and partial failures remain visible and nonzero. The branch is current and mergeable; governance checks pass. No new blocking finding.

JerrettDavis (approved):
Re-reviewed ddb6851a, including the changes since the previous approval and the complete current reindex delta. Page transactions retain rollback and per-record recovery, successful retries preserve a zero exit status, and unrecoverable records remain named with a nonzero exit. The benchmark now uses the production page-size default. All 30 CLI index-sync tests pass locally; current hosted checks report no failures. No blocking finding.

PR #3468, opened 2026-09-07
perf(relevance): order BM25 query terms once, not once per document

trs-80:
## Description

`bm25_score` collected `query_freq.keys()` into a `Vec` and sorted it on every call. That sort exists only to make matched-term output deterministic, and the query is fixed for a whole batch — so `score_batch` repeated the same allocation and sort once per document.

Closes # N/A

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `query_freq` is now a `BTreeMap<String, usize>` instead of a `HashMap`. This **removes** the work rather than hoisting it: the map already iterates in key order, so the collect-and-sort disappears entirely and both `score` and `score_batch` benefit. Net effect on `bm25_score` is fewer lines, not more.
- Added `benches/bm25_scoring.rs` (criterion, following the existing three benches) varying document count and query length.
- Added `examples/bm25_query_prep_parity.rs`, the harness backing the bit-identical claim below.
- New test `matched_terms_are_lexicographic_regardless_of_context_order`. To be clear, this does **not** fail on the base commit — the base sorts too. It is a forward guard: swapping back to a `HashMap` without restoring the sort would make matched-term order nondeterministic, and this catches that.

**Why `BTreeMap` rather than a hoisted sorted `Vec`.** A `Vec` passed down from `score_batch` would keep iteration contiguous, but it needs a signature change, a second construction site in `score`, and leaves the sort in the code. The `BTreeMap` deletes the sort outright, and it measured faster in every shape tested, so the extra machinery was not worth it.

## Testing

### Test Output

```text
$ cargo test --release --lib
925 passed; 0 failed; 1 ignored

$ cargo test --release --lib relevance
49 passed; 0 failed

$ cargo clippy --release --all-targets     # no errors
$ cargo fmt --check                        # clean

# Python side, against a maturin build of THIS branch (see note below):
$ pytest tests/test_relevance.py tests/test_relevance_extra.py \
         tests/test_relevance_split.py tests/parity/ \
         tests/test_acceptance.py tests/test_issue_746_tool_search.py -q
124 passed, 4 skipped
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), Rust release profile, criterion 0.8, Python 3.13.11, worktree off `upstream/main` @ `e67b3c8a`.
- Exact command / steps: `cargo run --release --example bm25_query_prep_parity > out.txt` on this branch and with `src/relevance/bm25.rs` stashed, then `diff`. For timings, `cargo bench --bench bm25_scoring` on each side, run twice in opposite order to control for thermal drift. Python tests were run against a `maturin build --release` wheel from this branch, not a copied `.so`.
- Observed result: **bit-identical output.** The parity harness prints raw IEEE-754 score bits (not decimals), rankings, reasons and matched terms for empty batches/contexts/documents, no-overlap, repeated query terms, exact ties, the long-token bonus, and 3 configs x 5 document counts x 3 query lengths. Baseline and candidate produce the same 3,739 lines, SHA `4e48098593e36aa1937e1629c4048e1280cfab47`. Timings, candidate faster in all nine shapes:

  | shape | baseline | candidate | delta |
  |---|---|---|---|
  | 10 docs, 4-word query | 51.55 us | 50.22 us | -2.6% |
  | 10 docs, 16-word query | 60.72 us | 56.49 us | -7.0% |
  | 10 docs, 64-word query | 65.94 us | 61.39 us | -6.9% |
  | 100 docs, 4-word query | 515.4 us | 504.7 us | -2.1% |
  | 100 docs, 16-word query | 590.6 us | 553.2 us | -6.3% |
  | 100 docs, 64-word query | 597.6 us | 562.4 us | -5.9% |
  | 1000 docs, 4-word query | 5.184 ms | 5.133 ms | -1.0% |
  | 1000 docs, 16-word query | 5.926 ms | 5.639 ms | -4.8% |
  | 1000 docs, 64-word query | 5.975 ms | 5.631 ms | -5.8% |

  The reverse-order confirmation run reproduced the 64-word column at -5.5%, -4.4% and -5.9%.
- Not tested: real proxy or agent workloads — these are synthetic documents from a fixed 16-word vocabulary, and no end-to-end latency claim is made. Only x86-free Apple-silicon measurements. CJK content takes a separate scorer in `text_crusher` and is unaffected.

## Runtime Rollout Safety

- Stable/default behavior changed: no. Scores are bit-identical, and rankings, reasons and matched-term ordering are unchanged.
- Kill switch / disable path: none needed; there is no behavioral switch to disable.
- Qualification impact: none. Scoring output is unchanged, so no re-qualification of compression decisions is required.
- Rollback path: revert the commit. `bm25_score` is a pure function over its inputs with no persisted state, so nothing needs migrating in either direction.

## Review Readiness

## Checklist

## Additional Notes

- The comment above the loop was rewritten rather than deleted: it explains why terms are visited alphabetically at all, which is still the load-bearing reason the ordering is not an implementation detail.
- Documentation: no user-facing docs describe this internal scoring path; the behavioral contract lives in that comment and in the new ordering test.

### DIFF
--- a/crates/headroom-core/Cargo.toml
+++ b/crates/headroom-core/Cargo.toml
@@ -210,3 +210,7 @@ harness = false
 [[bench]]
 name = "auth_mode"
 harness = false
+
+[[bench]]
+name = "bm25_scoring"
+harness = false
--- a/crates/headroom-core/benches/bm25_scoring.rs
+++ b/crates/headroom-core/benches/bm25_scoring.rs
@@ -1,0 +1,63 @@
+//! Criterion benchmark for `BM25Scorer::score_batch`.
+//!
+//! Query preparation used to happen once per document: every `bm25_score`
+//! call collected and sorted the same query keys. Cost therefore scales with
+//! documents x query length, so both are varied here.
+
+use std::hint::black_box;
+
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+use headroom_core::relevance::{BM25Scorer, RelevanceScorer};
+
+const VOCAB: [&str; 16] = [
+    "handler",
+    "parse",
+    "timeout",
+    "cache",
+    "retry",
+    "index",
+    "error",
+    "550e8400-e29b-41d4-a716-446655440000",
+    "9f8e7d6c-1234-4321-abcd-0123456789ab",
+    "12345",
+    "987654",
+    "resolve",
+    "connection",
+    "serialize",
+    "upstream",
+    "budget",
+];
+
+fn text(seed: usize, words: usize) -> String {
+    let mut parts: Vec<&str> = Vec::with_capacity(words);
+    for i in 0..words {
+        parts.push(VOCAB[(seed * 7 + i * 3) % VOCAB.len()]);
+    }
+    parts.join(" ")
+}
+
+fn bench_score_batch(c: &mut Criterion) {
+    let scorer = BM25Scorer::default();
+    let mut group = c.benchmark_group("bm25/score_batch");
+
+    for docs in [10usize, 100, 1000] {
+        for query_words in [4usize, 16, 64] {
+            let items: Vec<String> = (0..docs).map(|d| text(d, 60)).collect();
+            let refs: Vec<&str> = items.iter().map(|s| s.as_str()).collect();
+            let context = text(9_999, query_words);
+
+            group.throughput(Throughput::Elements(docs as u64));
+            group.bench_with_input(
+                BenchmarkId::from_parameter(format!("{docs}d_q{query_words}w")),
+                &(refs, context),
+                |b, (refs, context)| {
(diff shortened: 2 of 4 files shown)


JerrettDavis (approved):
Reviewed the query-frequency construction and both scoring paths. BTreeMap iteration preserves the old sorted String-key order and term frequencies, so floating-point accumulation and matched-term ordering remain unchanged while preparation is shared across the batch. All 15 BM25 tests pass after rebuilding headroom-core from this head; the new benchmark and parity example compile. No blocking findings.

JerrettDavis (approved):
Re-reviewed the rebased head on current main. The BTreeMap keeps the former lexicographic iteration order while eliminating per-document sorting; scoring accumulation, matched-term ordering, and rankings remain unchanged. The branch is current and mergeable; governance checks pass. No new blocking finding.

PR #3471, opened 2026-09-08
perf(ccr): bound retrieval history with a deque instead of reslicing

trs-80:
## Description

Once the CCR retrieval history reaches its 1,000-event cap, every further retrieval appended to the list and then rebuilt it from a 1,000-element slice, under the store lock. `collections.deque(maxlen=1000)` drops the oldest event as the new one lands, so the trim is deleted rather than moved.

This is a maintenance change that also removes an O(cap) copy. It is **not** an end-to-end speedup, and the measurements below say so: the history append itself gets ~3.9× cheaper, while the full `retrieve()` path is unchanged within noise.

`_pending_feedback_events` deliberately stays an unbounded list. It is a separate drain-by-swap queue that still owes a notification for every event, including ones the capped history has already discarded, so both containers now carry a comment saying which is which.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `headroom/cache/compression_store.py`: `_retrieval_events` becomes `deque(maxlen=self._max_events)`; the append-then-reslice in `_log_retrieval` is removed. `_max_events` stays as the single readable name for the cap — it is read at construction, by the new tests and by the benchmark, and is never assigned after `__init__`.
- `tests/test_compression_store.py`: two tests — retention and ordering at the cap, and one feedback notification per retrieval past the cap.
- `benchmarks/ccr_retrieval_history_benchmark.py`: standalone script measuring the history append at steady state, following the convention the other benchmarks in that directory use.

Consumer audit before changing anything: the container is only ever hit with `len`, `append`, `clear`, iteration, `list()` and `sys.getsizeof` — all deque-safe. The one slice, `events_copy[-limit:]` in `get_retrieval_events`, operates on the list copy taken under the lock, so deque's lack of slice support is never reached. Repo-wide there is no non-Python consumer, no subclass, no pickling or deepcopy path and no `isinstance(list)` check, and the deleted line was the only assignment to the attribute — so nothing can rebind an unbounded list back onto it. One `threading.Lock` guards every access.

## Testing

Both new tests pass before and after this change, deliberately: for a maintenance change, parity is the claim. They exist because the pre-existing `test_retrieval_events_list_bounded` asserts only `len(...) <= 1000`, which cannot tell which end was dropped. Mutation-checked: an unbounded `deque()` fails the length assertion, and a head-trim keeping the oldest 1,000 passes the length check but fails on order.

### Test Output

```text
$ pytest tests/test_compression_store.py tests/test_ccr.py tests/test_ccr_feedback.py \
         tests/test_critical_gaps.py tests/test_critical_fixes.py tests/test_proxy_ccr.py \
         tests/test_ccr_sqlite_backend.py tests/test_ccr_batch_store.py \
         tests/test_memory_tracker_integration.py -q
209 passed, 3 skipped in 16.66s

$ pytest tests/ -q -k "ccr or compression_store or compression_feedback or telemetry \
                       or critical_gaps or critical_fixes or memory_tracker" \
         --ignore=tests/test_integrations
1 failed, 921 passed, 11 skipped in 77.38s
# the one failure is test_ccr_marker_round_trip_live, a live-API test that fails
# identically on the unchanged base

$ python benchmarks/ccr_retrieval_history_benchmark.py
container: deque
cap: 1000 events
appends per run: 20000, runs: 7
median: 0.481 us/append

$ pre-commit run --files headroom/cache/compression_store.py \
      tests/test_compression_store.py benchmarks/ccr_retrieval_history_benchmark.py
```

## Real Behavior Proof

- Environment: macOS 15.6, Python 3.13.11, branch based on `main` @ `e67b3c8a`. Single process, single thread, no contention.
- Exact command / steps: filled a store's history to its cap so every subsequent append must evict, then timed 20,000 `_log_retrieval` calls, 7 runs, median reported. Ran the unchanged tree and this change in both orderings to rule out ordering effects. Separately timed the full public `retrieve()` path, 2,000 calls × 5 runs, also in both orderings.
- Observed result: history append **1.81 to 0.47 us** and, with the order reversed, **1.85 to 0.48 us** per append. Full `retrieve()`: 7.67 vs 7.89 us, and reversed 7.79 vs 7.82 us — unchanged within noise in both directions. Behavior parity checked value by value: `get_stats()["event_count"]`, `get_retrieval_events()` contents/order/length across limits from negative through 1e9, and `clear()` are all identical; `deque.clear()` preserves `maxlen`, so the container stays bounded afterwards.
- Not tested: no multi-threaded or contended measurement, no multi-worker or production workload, and no memory-over-time profile. The benchmark times a private method (`_log_retrieval`) because the public `retrieve()` buries the difference in backend I/O and the feedback drain — that is stated in the benchmark's own docstring. The result is a microbenchmark of one operation and must not be read as a proxy for retrieval latency.

One value does change numerically, and it is not hidden: `get_memory_stats().size_bytes` shifts by a few hundred bytes, because `sys.getsizeof` reports a deque's 64-slot blocks rather than a list's contiguous array (56 to 760 bytes empty; 8,056 to 8,680 at 1,000 entries). It is a shallow size that already excluded the events themselves, `budget_bytes` is `None`, and no test or dashboard compares it against anything.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. CCR retrieval-event history is not rollout-gated.
- Minimum rollout channel: N/A — no rollout-managed feature is touched.
- Stable/default behavior changed: No. The cap, its size, retention order, retrieval results, lock guarantees and feedback delivery are all unchanged; only the container enforcing the cap differs, plus the `size_bytes` gauge noted above.
- Kill switch / disable path: N/A — no flag. Reverting the commit restores the list and its trim.
- Qualification impact: None. No compression decision, retrieval result, CCR marker, TTL or persistence path is touched — this is the in-memory display history only.
- Rollback path: `git revert` of this commit, single step. No schema, data or config migration is involved.

## Review Readiness

## Checklist

Not applicable — no user-visible surface changes.

## Additional Notes

- No separate documentation file needed updating; the two comments in `__init__` are the documentation this needs, because the one thing a future reader can get wrong here is bounding `_pending_feedback_events` to match.
- Overlaps nothing in flight: PR #3465 touches this same module, but its entire footprint is inside `_clean_expired`, which never reads the retrieval history, and this change never touches `_stale_heap_entries`.
- Unrelated, left alone, noticed while running the suites: `tests/test_realignment_live_multi_turn.py::test_ccr_marker_round_trip_live` and `tests/test_memory_usage_integration.py::TestProxyMemoryIntegration::test_real_api_calls_memory_tracking` both need live API access and fail on a clean checkout without it.

### DIFF
--- a/benchmarks/ccr_retrieval_history_benchmark.py
+++ b/benchmarks/ccr_retrieval_history_benchmark.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python3
+"""Measure CCR retrieval-history maintenance once the event cap is reached.
+
+Past 1,000 events the history has to discard the oldest event on every single
+retrieval. Bounding the container itself makes that O(1); the previous
+append-then-reslice rebuilt a 1,000-pointer list each time.
+
+This times ``_log_retrieval`` directly — the store's public ``retrieve()`` also
+does backend I/O and a feedback drain, which would bury the difference. Read
+the result as "cost of one history append at steady state", not as a proxy for
+end-to-end retrieval latency.
+
+Usage:
+    python benchmarks/ccr_retrieval_history_benchmark.py
+    python benchmarks/ccr_retrieval_history_benchmark.py --events 5000 --runs 7
+"""
+
+from __future__ import annotations
+
+import argparse
+import statistics
+import sys
+import time
+from pathlib import Path
+
+# Run as a script, sys.path[0] is benchmarks/, so an editable install of another
+# checkout would win for `import headroom` and this would silently measure the
+# wrong tree.
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+from headroom.cache.compression_store import CompressionStore  # noqa: E402
+
+
+def _fill_to_cap(store: CompressionStore) -> None:
+    """Bring the history to its cap so every later append must evict."""
+    for i in range(store._max_events):
+        store._log_retrieval(
+            hash_key="h",
+            query=f"warm{i}",
+            items_retrieved=1,
+            total_items=1,
+            tool_name="bench",
+            retrieval_type="full",
+        )
+    store._pending_feedback_events.clear()
+
+
+def measure(events: int) -> float:
+    """Return microseconds per history append at steady state."""
+    store = CompressionStore(enable_feedback=True)
+    _fill_to_cap(store)
+
+    start = time.perf_counter()
+    for i in range(events):
+        store._log_retrieval(
+            hash_key="h",
+            query=f"q{i}",
+            items_retrieved=1,
+            total_items=1,
+            tool_name="bench",
+            retrieval_type="full",
+        )
+    elapsed = time.perf_counter() - start
+
+    # Each run uses a fresh store, so the unbounded feedback queue this also
+    # filled goes away with it.
+    return elapsed / events * 1e6
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--events", type=int, default=20_000)
+    parser.add_argument("--runs", type=int, default=7)
+    args = parser.parse_args()
+
+    store = CompressionStore()
+    print(f"container: {type(store._retrieval_events).__name__}")
+    print(f"cap: {store._max_events} events")
+
+    samples = [measure(args.events) for _ in range(args.runs)]
+
+    print(f"appends per run: {args.events}, runs: {args.runs}")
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
Reviewed all retrieval-history readers, retention, locking, clear, and feedback-drain behavior. The deque preserves the newest 1,000 events and public list snapshots; the separate feedback queue remains unbounded until drained, so eviction from display history does not lose notifications. All 76 compression-store tests pass, including overflow ordering and feedback delivery. The supplied benchmark also runs successfully. No blocking findings.

PR #3480, opened 2026-09-08
fix(savings): count compression savings once per conversation, not once per turn

gglucass:
## Description

On the OpenAI `/v1/responses` path, a request's `tokens_saved` is the running total for the whole **conversation**, not for that turn. Codex re-sends the entire transcript every turn and the router recompresses all of it, so the same removed token is booked again on every remaining turn. Anthropic's frozen cached prefix means its `tokens_saved` is already novel-only. The two providers have therefore never meant the same thing by the same field, and everything that sums it inherits the difference.

Measured across one machine's retained proxy logs on 2026-09-08: gpt/codex models reported 8.10M tokens saved against 3.90M tokens of new input (2.1x), where Claude reported 4.18M against 43.3M. Two requests from the same machine and the same minute:

```text
gpt-6-astra   tok_before=156260 tok_after=80240 tok_saved=76020 cache_read=66304
claude-opus-5 tok_before=114024 tok_after=113566 tok_saved=458   cache_read=98503
```

What inherits the inflation: lifetime savings totals, `by_model.savings_percent`, the dollar figure (repeat removals priced at the full uncached input rate when the alternative was a cache read at roughly 0.1x), and any rate dividing savings by new input, which then has no 100% ceiling.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- New `headroom/proxy/conversation_savings.py`: a bounded, thread-safe per-conversation ledger that converts the cumulative-per-turn series into the incremental one. 512 conversations, LRU; a forgotten conversation re-counts its transcript once rather than leaking.
- `RequestOutcome` gains `conversation_key` and `conversation_tokens_saved`, both defaulting to `None`. `None` means "this path does not distinguish", and the funnel falls back to `tokens_saved` unchanged.
- `emit_request_outcome` derives the novel figure once and uses it for the two cumulative consumers (`metrics.record_request`, `cost_tracker.record_tokens`). The request log and the transformations feed keep `outcome.tokens_saved`: those tokens really did leave this request's payload, so per-request surfaces still describe the wire.
- The PERF line gains `tok_novel=` only when it differs from `tok_saved`.
- The three `/v1/responses` emit sites populate the pair: HTTP non-streaming, HTTP streaming (via `_stream_response` -> `_finalize_stream_response` -> `RequestOutcome.from_stream`), and the WebSocket per-turn and session-residual records. On WS the conversation total is the last frame's figure, not the session sum, because the session sum adds cumulative values together.
- The key is derived with the existing `conversation_key_from_body` and taken **before** compression, so a rewritten first user message cannot move it mid-conversation. It is left unset when compression did not run, so a bypassed turn cannot reset the running total.

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_proxy/test_conversation_savings.py -q
...............                                                          [100%]
15 passed in 0.30s

$ uv run --frozen --extra dev pytest tests/test_request_outcome.py tests/test_openai_codex_ws_lifecycle.py \
    tests/test_openai_codex_routing.py tests/test_openai_responses_output_shaper.py \
    tests/test_proxy_project_savings.py tests/test_handler_outcome_tag_invariant.py tests/test_proxy/ -q
250 passed, 1 warning in 89.55s

$ uv run --frozen --extra dev pytest tests/ -q
5 failed, 7538 passed, 514 skipped, 6 warnings in 178.47s

$ uv run --frozen --extra dev mypy headroom/proxy/outcome.py headroom/proxy/conversation_savings.py headroom/proxy/handlers/streaming.py

$ uvx ruff check headroom/proxy/
All checks passed!
```

The 5 full-suite failures are pre-existing on `origin/main`. Verified by stashing this branch and re-running the full suite on the base commit, which fails the same `tests/test_providers/test_deepseek.py` cases (stale pricing registry) plus the same order-dependent `tests/test_lossless_mode.py::test_router_lossless_search_no_marker_and_recoverable`. Both files pass in isolation with this change applied.

## Real Behavior Proof

- Environment: macOS 24.6.0, Python 3.11, `uv run --frozen --extra dev`, base commit 88f935a1. Field evidence from a separate machine running the released 0.37.0 wheel with live Codex and Claude Code traffic.
- Exact command / steps: aggregated every `PERF` line in `~/.headroom/logs/proxy.log*` by model family, summing `tok_saved` against per-request new input (`tok_after - cache_read + cache_write`); cross-checked against `/stats` `prefix_cache.by_provider` and `persistent_savings.by_model`; then ran the suites above against this branch.
- Observed result: gpt/codex 443 requests, 8,098,522 `tok_saved` against 3,896,164 new input (2.08x, i.e. a 67.5% "share of new input"); claude 2,163 requests, 4,180,023 against 43,296,720 (8.8%). With this change the same twenty-turn conversation books its removals once: `test_a_long_conversation_counts_each_removed_token_once` sums twenty turns reporting 10,000 each and asserts 10,000, which fails on `main` with 200,000.
- Not tested: no live end-to-end replay against the real ChatGPT WebSocket transport; the WS path is covered by unit tests and the existing `tests/test_openai_codex_ws_lifecycle.py` suite only. Bedrock and Gemini are untouched by construction (they never set the pair) and were not exercised beyond the full suite.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. This is unconditional accounting on the OpenAI `/v1/responses` path; it is not gated by `rollout.py` FEATURES and adds no new feature name.
- Minimum rollout channel: N/A (not rollout-managed). Behaves identically on stable, beta and dev.
- Stable/default behavior changed: Yes, and this is the point. Cumulative savings totals on `/v1/responses` stop counting the same removed token once per remaining turn, so reported Codex savings fall by roughly an order of magnitude on long conversations. Affected surfaces: `SavingsTracker` / Prometheus totals, `by_model.savings_percent`, the cost dashboard's savings dollars, and the `headroom savings` ledger. Per-request surfaces (`tok_before`/`tok_after`, request log, transformations feed) are unchanged. No provider other than OpenAI `/v1/responses` is affected: every other handler leaves the new fields `None` and takes the unchanged `tokens_saved` path, which is covered by `test_paths_without_conversation_identity_are_unchanged`.
- Kill switch / disable path: None added. Disabling would mean restoring a known-wrong number, so a flag would only preserve the bug. Reverting the commit is the disable path; the ledger is a leaf module with no other callers.
- Unsafe override required: No. No new env vars, no config keys, no unsafe overrides.
- Qualification impact: Anyone tracking Codex savings across this version boundary will see a step change that is a correction, not a regression. A same-window before/after comparison of Codex savings totals is not valid across this commit; compare per-conversation unique removals instead. Suggest calling it out in the release notes rather than shipping it silently.
- Rollback path: Revert the single commit. State is in-process only (a bounded in-memory `OrderedDict`, no file, no schema, nothing persisted), so a rollback needs no migration and leaves nothing behind. Records already written under either accounting are unaffected either way.

## Review Readiness

## Checklist

## Additional Notes

Deliberate scope limits, happy to extend if you would rather have them here:

- **Per-request surfaces keep the wire figure on purpose.** `tok_before`/`tok_after`, the request log and the transformations feed still say what left this process for this request. Only quantities that accumulate across turns de-duplicate. If you would rather the feed also show the novel figure, it is one line.
- **This lowers reported Codex savings, substantially.** That is the point, but it is a visible drop for anyone whose traffic is mostly Codex, so it may deserve a release note rather than a silent bump.
- **The 512-conversation cap is a memory bound, not a correctness one.** A conversation evicted while still live re-counts its transcript once. Sized well above what any one client keeps warm; happy to make it configurable.
- Documentation and CHANGELOG are unchecked above because I did not want to guess your release framing for a user-visible metric change. Say the word and I will add both.


JerrettDavis (changes requested):
The accounting needs two distinctions that the current integration does not preserve.

1. [P1] Independent Responses conversations share the deduplication bucket. The handlers call conversation_key_from_body, whose Responses fallback uses an explicit metadata identifier, otherwise the first 512 instruction characters, otherwise the literal responses. It does not seed from the input text. Two ordinary payloads with the same model and different user input therefore generate the same key; passing their independent 100-token savings through ConversationSavings attributes [100, 0]. Because the ledger is process-wide, unrelated sessions/projects can suppress each other's savings. Please use an actual scoped conversation identity, and avoid deduplicating when one cannot be established. A holdout-assignment fallback is not sufficient identity for accounting. Add an integration regression deriving keys from real Responses bodies with distinct input and shared instructions, rather than injecting conv-a in the outcome test.

2. [P2] A Responses request is not necessarily a full-transcript snapshot. This handler also supports previous_response_id, and clients can submit only fresh input while the provider retains prior context. Compression savings on those payloads are per-request increments, not a cumulative total. With the same explicit conversation id and two different previous_response_id/fresh-input requests that each remove 100 tokens, the new attribution records [100, 0]. Please distinguish full-history replay from incremental/server-state input before applying the difference, and retain ordinary per-request accounting when the cumulative premise is not established. Cover both wire shapes through the handler/funnel.

All 15 new conversation-savings tests pass. They exercise the difference arithmetic and constructor plumbing but assume valid identity and cumulative input, so they miss both cases above.

gglucass:
Both points taken, and they were both real: the holdout key was never identity, and I had assumed every Responses payload is a transcript.

**1. Identity.** New `savings_conversation_key` in `conversation_savings.py` replaces `conversation_key_from_body` at all three emit sites. It returns `None` (per-request accounting, no ledger entry) unless the body names one conversation: Codex's `prompt_cache_key`, an explicit top-level `conversation_id`/`session_id`/`thread_id`, the same inside `metadata`/`client_metadata`, or a transport session the caller vouches for (the WebSocket socket id, or a `conversation_id`/`session_id`/`x-headroom-session-id` request header on HTTP). The instructions prefix and the literal `responses` are deliberately not accepted. Regression: two real Responses bodies with shared instructions and distinct input get the same holdout key and `None` from the savings key; the same pair through the handler books `[100, 100]`.

**2. Cumulative premise.** With `previous_response_id` or a server-side `conversation` (string or object) the key is `None` regardless of identity. Handler regression: full-transcript replay under one `prompt_cache_key` books `[100, 0]`; two fresh-input requests under the same key with different `previous_response_id` book `[100, 100]`. A session-header case covers `[100, 0, 100]` across two sessions.

Suites: conversation savings 19, codex routing 40, WS lifecycle 52, request outcome 39, handler tag invariant 3, project savings 16, all green; mypy clean on the ledger and funnel.

JerrettDavis (changes requested):
The missing-identity fallback and previous_response_id/server-conversation guards resolve the two reproduced wire shapes. The new prompt_cache_key identity still aliases independent conversations, however. This parameter groups prompt-cache routing; it is not a unique conversation ID. OpenAI explicitly documents multi-turn-agent cache keys shared across a user's sessions/forks, as well as a shared key for independent single-turn judging requests: https://developers.openai.com/api/docs/guides/prompt-caching .

At 1a5a349eacd334c3057d2bed977b23791b62da67, I ran the new actual-handler test harness with two different user inputs, prompt_cache_key="shared-support-prefix", and distinct session_id headers (session-0/session-1). Each compression removed 100 tokens, but metrics recorded [100, 0]. The helper prioritizes the shared cache key over the true session identity, so even clients supplying a session header lose isolation.

Please prefer and scope an actual conversation/session identifier, and do not enable global accounting deduplication from prompt_cache_key alone unless a specific adapter establishes its per-conversation uniqueness. Keep per-request accounting otherwise. Add a handler regression with a shared cache key plus different session IDs, and one with only a shared cache key. All 59 conversation/Codex-routing tests pass; the new identity tests currently assume prompt_cache_key is a session ID.

gglucass:
Agreed, and thanks for the docs pointer: a cache-routing group is not a conversation. `prompt_cache_key` is out of the identity list entirely. Only an explicit `conversation_id`/`session_id`/`thread_id` (top level or in `metadata`/`client_metadata`) or a caller-vouched transport session (the WebSocket socket, or a `conversation_id`/`session_id`/`x-headroom-session-id` header) de-duplicates; everything else keeps per-request accounting. The docstring records why, and that an adapter which establishes per-conversation uniqueness for a client may pass the key as `session_id`, which nothing does today.

Handler regressions added: shared `prompt_cache_key` with `session-0`/`session-1` headers books `[100, 100]`; the same shared key with no session books `[100, 100]`; a repeat under `session-0` then books `0`, so the session header is what de-duplicates. The earlier full-transcript test now uses `metadata.conversation_id`. Conversation savings 19, Codex routing 41, WS lifecycle 52 green; mypy clean.

JerrettDavis (approved):
Verified the follow-up: prompt_cache_key is no longer treated as conversation identity, distinct session headers retain independent savings, and requests with only a shared cache key use ordinary per-request accounting. The handler regression reproduces the previously failing case and now books [100, 100]. All 60 routing and conversation-savings tests pass; the earlier incremental-input and missing-identity guards remain intact.

PR #3482, opened 2026-09-08
fix(tokenizers): estimate replayed thinking from text and signature size

gglucass:
## Description

Estimate replayed Anthropic thinking blocks explicitly instead of pricing their serialized JSON as ordinary text. The shared content walker counts visible thinking text plus `len(signature) * 3 // 16`, an approximation based on decoded signature size. This keeps omitted-display reasoning from counting as free while reducing the old base64-as-prose estimate.

Encrypted reasoning cannot be counted exactly from its opaque signature. This is a local heuristic, informed by the author's traffic sample, not a replacement for provider-reported usage or provider token counting.

## Changes Made

- Add the thinking branch to the shared tokenizer walker.
- Cover signed/unsigned blocks, omitted visible thinking, and provider-walker consistency.

## Testing

Latest reviewer validation on macOS/Python 3.13 at `127a039bd`: `pytest tests/test_tokenizers.py -q` — **55 passed, 14 skipped**. Optional tokenizer dependencies account for the skips. Earlier author validation reported Ruff, formatting, and focused mypy passing.

## Real Behavior Proof

The prior empty-thinking/4,000-character-signature case now adds a 750-token estimate instead of zero. No new live provider calibration was performed during review; accuracy across models, languages, signature formats, and thinking-retention policies remains approximate.

## Runtime Rollout Safety

- Changes local token estimates for replayed thinking; request content is unchanged.
- These estimates feed sizing, compression decisions, and local savings displays.
- No persistent schema migration or experimental opt-in is required.
- Rollback: revert the thinking-count branch and its tests.

## Review Readiness

- [x] The reported omitted-thinking zero-count issue is addressed.
- [x] Focused tests pass.
- [ ] Final human review and merge.

### DIFF
--- a/headroom/tokenizers/base.py
+++ b/headroom/tokenizers/base.py
@@ -328,6 +328,16 @@ class BaseTokenizer(ABC):
                         total += frames * 1000
                     else:
                         total += 3200
+                elif part_type == "thinking":
+                    # Anthropic extended thinking replayed by the client. Only
+                    # the thinking text is model input; ``signature`` is an
+                    # opaque server-side verification blob (~1 KB of base64 per
+                    # block) that the provider does not tokenize or bill. The
+                    # JSON catch-all below priced it as text: on a 369-message
+                    # Claude Code session, 99 signatures counted 256K of 414K
+                    # tokens against a provider-reported ~313K, inflating every
+                    # tokens_before-derived figure and the size stratum.
+                    total += self.count_text(part.get("thinking", "") or "")
                 else:
                     # Unknown type - estimate from JSON
                     total += self._count_serialized(part)
--- a/headroom/tokenizers/base.py
+++ b/headroom/tokenizers/base.py
@@ -329,15 +329,22 @@ class BaseTokenizer(ABC):
                     else:
                         total += 3200
                 elif part_type == "thinking":
-                    # Anthropic extended thinking replayed by the client. Only
-                    # the thinking text is model input; ``signature`` is an
-                    # opaque server-side verification blob (~1 KB of base64 per
-                    # block) that the provider does not tokenize or bill. The
-                    # JSON catch-all below priced it as text: on a 369-message
-                    # Claude Code session, 99 signatures counted 256K of 414K
-                    # tokens against a provider-reported ~313K, inflating every
-                    # tokens_before-derived figure and the size stratum.
+                    # Anthropic extended thinking replayed by the client. The
+                    # ``signature`` is the encrypted full reasoning: on the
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
[P1] Do not equate visible thinking text with the replayed model input. Anthropic documents that `signature` carries encrypted full reasoning, the server decrypts it for prompt construction, and `display="omitted"` returns an empty `thinking` field without removing that reasoning. With this patch, a block containing empty thinking plus a 4,000-character signature counts exactly like empty content. Retained thinking can therefore disappear from context-size estimates; summarized thinking also need not represent the full reasoning. See https://platform.claude.com/docs/en/build-with-claude/thinking (Controlling thinking display / Thinking encryption).

Avoid tokenizing opaque base64 as ordinary prose, but retain an explicit estimate or provider-derived accounting for the hidden reasoning instead of assigning it zero based on display text. Please cover omitted and summarized thinking on a model that preserves prior blocks, and revise the comment that says only the visible text is model input. The local tokenizer suite passes (54 passed, 14 skipped), but its new equality tests only encode the disputed assumption.

gglucass:
Agreed, and thanks for the docs pointer. Re-read it: on the keep-all-turns models (Opus 4.5+, Sonnet 4.6+, all 5.x) prior-turn thinking stays in context and is billed as input, and under the 5.x default `display: "omitted"` the `thinking` text is empty with the whole reasoning in the signature. Every thinking block in my local Claude Code sessions is that shape (237/237, 158/158, 105/105 empty text), so text-only priced all of it at zero. Wrong in the other direction.

Amended: price the text plus the decoded signature bytes at ~4 bytes/token (`len(signature) * 3 // 16`). On the session in the description that puts the 99 signatures at roughly 1.5K tokens each, which is about the 313K - 158K gap between provider-reported and text-only, so it is an estimate anchored on one measurement rather than a claim about the encoding. The equality tests are gone; the new ones check an omitted block prices above zero, a signed block prices 750 above its unsigned twin for a 4000-char signature, and that stays below the JSON catch-all. Comment rewritten to say what the signature is.

JerrettDavis (approved):
Re-reviewed at 127a039bd. Omitted/empty visible thinking no longer contributes zero: the shared walker counts visible text plus a signature-size estimate, and the provider walker follows the same rule. All 55 tokenizer tests pass, with 14 optional skips. This resolves my zero-count blocker. The signature contribution remains a heuristic, not an exact provider token count; the PR description/title have been aligned with that limitation and the final implementation.

PR #3483, opened 2026-09-08
fix(compression): compress cache_control blocks in the request's final message

gglucass:
## Description

`ContentRouter._process_content_blocks` hard-skips any block carrying `cache_control` so a key the provider already holds is never busted. Claude Code stamps `cache_control` on its newest `tool_use` and newest `tool_result` every turn. That newest tool result is the one block Headroom gets a clean shot at before the frozen-prefix guard covers it, and the rule skipped it on exactly that turn. Measured on one day of Claude Code traffic (650 requests): 1,229 `cache_control_protected` visits, about two per request, on a session where content compression totalled ~5K tokens against 53M forwarded.

The request's final user/tool message has never been forwarded, so no provider key exists for it yet. The marker there is the client staking out next turn's breakpoint; whatever we forward is what gets cached, and the frozen prefix replays those bytes from the next turn on. This releases `cache_control` blocks in that one position only. Every earlier message keeps the hard skip, and assistant blocks keep it in every position.

## Type of Change

## Changes Made

- `headroom/transforms/content_router.py`: `fresh_turn = messages_from_end == 1 and role in ("user", "tool")`; the `cache_control` skip applies `and not fresh_turn`. Contract 1 in the docstring names the exception. Every rewrite branch already spreads the source block, so the marker rides on the compressed block.
- `tests/test_transforms_content_router.py`: the existing tool_result test now passes `messages_from_end=3` and asserts the protected count; a new test proves the final-message block is compressed with its marker intact, and that an assistant block in the final position keeps the skip.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_transforms_content_router.py tests/test_transforms -q -x
(see CI; router suite green locally)
uvx ruff check headroom/transforms/content_router.py tests/test_transforms_content_router.py
All checks passed!
uvx ruff format --check headroom/transforms/content_router.py tests/test_transforms_content_router.py
2 files already formatted
```

## Real Behavior Proof

- Environment: macOS 15, Python 3.12, headroom-ai 0.37.0 as installed by Headroom Desktop, Claude Code 2.1.x as the client.
- Exact command / steps: Summed the router's `route_counts=` INFO lines for 2026-09-08 across the proxy log; inspected `cache_control` placement in the transformations feed's request bodies; ran the fix as a sitecustomize wrapper against the installed wheel with a probe that feeds a final-message tool_result carrying `cache_control` through `_process_content_blocks`.
- Observed result: 650 requests, 6,930 unfrozen messages, 311 blocks compressed, 1,229 `cache_control_protected`; every feed request had the marker on exactly the last `tool_use` and last `tool_result`. With the fix the final-message block is compressed, the output block carries `cache_control: {type: ephemeral}`, the client's original dict is untouched, and a `messages_from_end=3` block is still protected and counted.
- Not tested: fleet-scale cache-hit effect over days (the desktop's staging soak and difference-in-differences gate cover that before it reaches stable).

## Runtime Rollout Safety

- Stable/default behavior changed: Yes. A `cache_control` block in the request's final user/tool message is now eligible for compression; all other positions are unchanged.
- Kill switch / disable path: None beyond revert; the existing tool exclusions and read protection still apply to the released block.
- Qualification impact: None.

## Review Readiness

## Checklist

N/A

## Additional Notes

Cache-safety reasoning for the maintainer: the provider caches what it receives. On the block's first sighting nothing is cached under its key, so compressing it costs nothing; from the next turn on it is inside the frozen prefix and replays byte-exact. The old rule made the block compressible only during the few turns between its first sighting and the freeze, when compressing it DID bust a key. This flips that: compress when it is free, never when it is not. Headroom Desktop carries this as a sitecustomize vendor pinned to 0.37.0 until a wheel ships it.

## Replay guarantee (after review)

The exception is gated on `prefix_replay_guaranteed=True`, an `apply()` kwarg passed only by the proxy paths that replay last turn's forwarded prefix via `finalize_turn`. A caller without that overlay (SDK client, evals, batch, a bare router) keeps the hard skip, because on the next turn the block is no longer final and the router would otherwise forward the client's original bytes. Apply-level tests cover both callers, a retried final message, and the overlay restoring the compressed bytes on turn 2.


JerrettDavis (changes requested):
[P2] The final-message exception needs a replay guarantee, not position alone. ContentRouter is also used without the proxy’s frozen-prefix state. I exercised two successive apply() calls on the same router, stubbing only the compressor with a deterministic accepted result: a cache_control tool_result was compressed on turn 1, then restored to the original bytes when the client replayed its transcript plus the next assistant/user turn. Its marked prefix changes from 15 bytes back to 14,290 bytes, defeating the cache key this rule is meant to preserve. The router’s own compression cache cannot help because the older marked block is hard-skipped before consulting it.

Please retain the hard skip for callers without a verified replay guarantee, or carry the original-to-forwarded mapping through subsequent turns before allowing this exception. Add an apply-level multi-turn regression, including a caller without proxy prefix state and a retried final message. All 60 router tests pass, but the new test invokes the block helper only once and cannot establish the claimed next-turn replay behavior.

gglucass:
Good catch, and the replay you exercised is real: a bare router hard-skips the block again once it is no longer final, so without an overlay turn 2 forwards the original bytes.

Took the first option. The exception now sits behind an apply() kwarg, `prefix_replay_guaranteed=True`, that only the proxy paths which run `finalize_turn` pass (Anthropic messages, OpenAI chat, and the session-tracked branch of the compress endpoint; the batch handler, Gemini, the SDK client and evals do not, so they keep the hard skip). The router reads it as a strict `is True`, default off.

Apply-level tests added: (1) a caller with no proxy prefix state runs two turns and the marked block is never touched on either, with zero compress calls; (2) under the guarantee, turn 1 compresses and keeps the marker, a retried final message is byte-identical, turn 2 shows the router alone reverting to the original bytes, and `overlay_cached_prefix` with last turn's original/forwarded pair restores the compressed block. Router suite 72 passed; the handler suites that drive `pipeline.apply` are green too.

JerrettDavis (approved):
Re-reviewed at 5b01738cf. The fresh cache_control exception now defaults off and requires an explicit replay guarantee; bare ContentRouter callers preserve the marked bytes on both turns. The proxy call sites that enable it participate in the existing forwarded-prefix replay path. My original two-call repro now keeps the 14,290-byte block identical across turns, and all 62 content-router tests pass. This resolves the SDK/bare-router regression I reported.

PR #3484, opened 2026-09-08
fix(kompress): append the CCR marker whenever the saving pays for it

gglucass:
## Description

Kompress previously added a CCR retrieval marker only below a fixed compression ratio. Smaller reductions could therefore produce lossy output without a retrieval marker, which the router discarded.

The single, batch, and remote paths now compare the complete original payload with the complete candidate plus retrieval marker. A marked candidate ships only when it is smaller under the same `cl100k_base` estimate, and its result accounting uses those measured counts. Otherwise the original passes through. CCR-disabled behavior is unchanged.

This is a fixed tokenizer estimate, not an exact billing count for every OpenAI model or other provider. If the encoder is unavailable, the fallback compares character counts rather than claiming a provider-token bound.

## Type of Change

- [x] Bug fix

## Testing

Latest local validation: 81 tests passed across marker, remote, deadline, compressor, size-gate, and CCR policy/resolution suites. Ruff check and format check pass for the five changed Python files.

## Real Behavior Proof

- Environment: macOS, Python 3.13, deterministic compressor stubs with real source-derived CCR hashes and `cl100k_base` encoding.
- Exact command / steps: Run the marker/remote/deadline suites and related compressor/CCR suites. Re-run the single/batch boundary probe using `" ".join(["alpha"] * 36 + ["bureaucratic"] + ["alpha"] * 63)` with the supplied drop-first-36 model stub.
- Observed result: The formerly expanding candidate (100 original tokens, 103 with marker) now passes through unchanged, reporting 100 to 100 with no retrieval key. Tests also verify that candidates with net savings ship with marker-inclusive accounting, and partial deadline completion preserves the unprocessed tail.
- Not tested: exact billing equivalence for other model tokenizers, live model/provider traffic, or long-term retrieval-rate effects. Current-head CI is pending.

## Runtime Rollout Safety

- Stable/default behavior changed: CCR-enabled Kompress uses full-payload net savings rather than a fixed ratio threshold for marker admission. Marker-inclusive successful result counts use the same tokenizer estimate as the gate.
- Kill switch / disable path: `KompressConfig(enable_ccr=False)` disables CCR markers as before.
- Qualification impact: Compression accounting includes the marker on admitted CCR results; provider billing remains an estimate.

## Review Readiness

- [x] Prior marker-accounting findings resolved.
- [x] Focused and related local checks pass.
- Final human review is required.

## Checklist

- [x] Single, batch, remote, and deadline regressions covered.
- [x] No manual CHANGELOG edit.

### DIFF
--- a/headroom/transforms/kompress_compressor.py
+++ b/headroom/transforms/kompress_compressor.py
@@ -1294,41 +1294,41 @@ class KompressResult:
         return (self.tokens_saved / self.original_tokens) * 100
 
 
-_marker_encoder: Any = None
+_payload_encoder: Any = None
 
 
-def ccr_marker_cost(marker: str) -> int:
-    """Token cost of a retrieval marker, in the tokens the provider will bill.
+def payload_tokens(text: str) -> int:
+    """Token count of a complete payload, in one consistent unit.
 
-    The marker is 12 words but 36-45 cl100k_base tokens: the 24-hex hash alone
-    is ~16 and the word counts add a token per few digits, so any fixed word
-    allowance is wrong for some hash or some count (a 40-word allowance let a
-    43-token marker ship on a payload of 100 single-token words: 100 -> 102).
-    Price the actual marker text with cl100k_base (a hard dependency) and fall
-    back to one token per character, the most any BPE can produce, if the
-    encoder is unavailable.
+    The unit is cl100k_base (tiktoken, a hard dependency): exact for OpenAI
+    models and an ESTIMATE for every other provider, whose tokenizers differ
+    in detail but not in the shape that matters here (a hex hash is many
+    tokens, a common word is one, a word's tokenization depends on what
+    precedes it). Without an encoder, one token per character, the most any
+    BPE can produce.
 
-    The compressor accounts in words, and the words it drops are the cheap
-    ones, at least one token each. So comparing saved WORDS against marker
-    TOKENS is conservative in the direction that matters: a marker is added
-    only when the dropped words are worth at least its tokens under the
-    tightest possible tokenization, and the reported ``compressed_tokens``
-    (kept words plus marker tokens) errs high, never low.
+    The CCR gate measures the whole original and the whole candidate-plus-
+    marker with this, never a marker-only cost against a word count: the
+    marker is 36-45 tokens for 12 words, the words Kompress drops can be one
+    token each, and the word left at the head of the candidate can tokenize
+    differently from its space-prefixed form in the source. Only a comparison
+    of the two complete texts in one unit establishes that the shipped
+    payload is smaller.
     """
-    global _marker_encoder
-    if _marker_encoder is None:
+    global _payload_encoder
+    if _payload_encoder is None:
         try:
             import tiktoken
 
-            _marker_encoder = tiktoken.get_encoding("cl100k_base")
+            _payload_encoder = tiktoken.get_encoding("cl100k_base")
         except Exception:
-            _marker_encoder = False
-    if _marker_encoder:
+            _payload_encoder = False
+    if _payload_encoder:
         try:
-            return len(_marker_encoder.encode(marker))
+            return len(_payload_encoder.encode(text, disallowed_special=()))
         except Exception:
             pass
-    return len(marker)
+    return len(text)
 
 
 def ccr_retrieval_marker(
@@ -1763,12 +1763,15 @@ class KompressCompressor(Transform):
             compressed = " ".join(compressed_words)
             compressed_count = len(compressed_words)
             cache_key: str | None = None
-            marker_cost = 0
+            original_tokens = n_words
+            compressed_tokens = compressed_count
 
             # CCR marker: anything the lossy pass shrank must stay retrievable,
-            # and the complete marked payload must still be a saving. The
-            # marker is built first and priced (ccr_marker_cost); a result
-            # whose saved words cannot pay for it is passed through.
+            # and the complete marked payload must be smaller than the
+            # original. Both are measured whole, in one unit (payload_tokens);
+            # a candidate that is not is passed through. The accounting then
+            # reports that same measurement, so ``tokens_saved`` describes
+            # the shipped payload rather than a word count.
             if self.config.enable_ccr:
                 ccr_source = ccr_original if ccr_original is not None else content
                 ccr_source_tokens = len(ccr_source.split())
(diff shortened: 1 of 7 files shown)


JerrettDavis (changes requested):
[P2] Check the complete marked result in the accounting used for acceptance. The new 13-word threshold does not establish that the retrieval marker pays for itself in tokens, and compressed_tokens/compression_ratio are still computed before appending it. Using this PR’s _DropFirst fixture with drop=16 on _prose(100), and a normal-length CCR hash c00eb437e5e5c00eb437e5e5, I get 100→96 words including the marker but 196→200 cl100k_base tokens; the result still reports ratio=0.84. The router consumes the pre-marker compressed_tokens, so the newly admitted weak-shrink case can appear beneficial while expanding the payload.

Please include marker overhead in the returned accounting and reject or pass through a candidate when the complete marked payload does not save tokens (or use a documented conservative bound when a provider tokenizer is unavailable). Cover single and batch paths at this boundary with a real-length hash; the current marker-word test and abc123 fixture do not catch it.

gglucass:
Confirmed your numbers: a real-length marker is 36 cl100k tokens (38 with large counts), the hash alone is 16, and the 13-word gate let the drop-16-of-100 case ship at reported ratio 0.84 while the marked payload went 196 -> 200 tokens.

Amended in three places:

- **Gate is a documented conservative bound in the module's own unit.** `CCR_MARKER_COST_WORDS = 40`: a marker is appended only when the pass saves MORE than 40 words. Every dropped word is at least one token, so that covers the marker even when Kompress drops nothing but single-token filler, which is the common case. A result that cannot pay passes through untouched (under CCR an unmarked lossy result is discarded by the router's #1307 guard anyway, so this just stops paying for the store).
- **Returned accounting includes the marker.** `compressed_tokens` and `compression_ratio` are now computed after the marker decision as kept words plus the marker cost, so what the router consumes describes the payload as shipped. Same on the batch path.
- **Remote client aligned.** `kompress_remote.py` carried its own `ratio < 0.8` copy with the same pre-marker accounting; it now uses the shared bound, passthrough and marker function.

Tests use the real-length hash `c00eb437e5e5c00eb437e5e5`: the bound covers a real marker's cl100k count, your drop-16 case passes through and the tiktoken assertion shows the marked payload would have grown, a 300 -> 240 shrink is marked and reports `240 + 40`, batch mirrors both, and no-CCR mode still ships bare lossy output unchanged. Three existing fixtures that compressed 40-word blocks moved past the bound or dropped CCR where it was irrelevant.

Separate observation, not in this PR: `_try_ml_compressor` returns Kompress's word count as the compressed token count while the original is priced by the estimator (~2 tokens/word on code and logs), so the router's view of any Kompress result is roughly twice as good as reality. Happy to open a follow-up that recounts the final payload with `_estimate_tokens` if you want that fixed.

JerrettDavis (changes requested):
The fixed 40-word allowance narrows the failure but does not bound the actual marker cost. At 110def43a, use source " ".join(["alpha"] * 99 + ["nfs"]) with the existing _DropFirst(41) model stub. Its actual SHA-256 prefix is 7dbb8f8de9f3e1d7c6f3a6e1 (no artificial hash required). The emitted marker costs 43 cl100k_base tokens: the original is 100 tokens, the shipped marked payload is 102, while KompressResult reports original_tokens=100, compressed_tokens=99 and ratio=0.99. I reproduced this through compress() with the actual source-derived key.

Please gate on the complete marked payload using the caller's tokenizer or another defensible bound that covers variable hash/numeric marker text, and keep reported counts on the same unit as the input. Simply increasing a fixed allowance based on one sample repeats the same gap. Add this source/hash case to the single, batch, and remote parity coverage. The 25 focused marker/placeholder/nonblocking/remote tests pass, but the current marker-bound test checks only one chosen hash.

gglucass:
Reproduced exactly: `alpha` x99 + `nfs`, drop 41, real key `7dbb8f8de9f3e1d7c6f3a6e1`, marker 43 cl100k tokens, payload 100 -> 102. A fixed allowance was the wrong shape for a variable-cost marker; you are right that raising it just moves the gap.

Now the marker is built first and priced as itself: `ccr_marker_cost(marker)` encodes the actual marker text with cl100k_base (a hard dependency of the package), falling back to one token per character if the encoder is unavailable, which is the most any BPE can produce. The gate is `saved_words > marker_cost`, else passthrough, and a marked result reports `kept_words + marker_cost`. Units: the compressor accounts in words, and the words it drops are at least one token each, so comparing saved words against marker tokens is conservative in the direction that matters and the reported `compressed_tokens` errs high, never low; the docstring says so. Same code on the batch path and in `kompress_remote.py`.

Coverage for your case: single path passes through (`100, 100, 1.0`), batch passes through, remote parity passes through, each with the real source-derived key rather than a stub hash. A 300-word single-token source dropping 60 is marked on all three paths and reports `240 + ccr_marker_cost(marker)`, where the test recomputes the cost from the emitted marker. Also pinned: the cost equals the tiktoken count of the real marker and varies with digits and hash, and the no-encoder fallback equals `len(marker)`. Twelve suites green, mypy clean.

JerrettDavis (changes requested):
Measuring the actual marker fixes the previous hash-length example, but the gate still mixes saved words with marker tokens and does not measure the complete retained payload. Tokenization can change at the new leading word. A concrete current-head single/batch repro using the supplied _DropFirst(36) stub: source = " ".join(["alpha"] * 36 + ["bureaucratic"] + ["alpha"] * 63). The real source hash is 0bb5ef170dc0b9664fe1ab09. Both paths ship a marked result and report 100 -> 99, but cl100k_base measures 100 -> 103. The retained leading bureaucratic tokenizes differently from space-prefixed bureaucratic in the source.

Please compare the full original text and full candidate-plus-marker in one consistent token unit, and populate the result accounting from that same measurement. Applying a marker-only cost to word counts cannot establish the net-saving invariant. Add this boundary regression for single and batch, and apply the same policy to the remote path. If the chosen encoder is an estimate for the actual provider, describe it as such rather than provider-billed tokens.

Validation: the marker and remote suites pass (14 tests); the related deadline suite has one pass and one failure in test_compress_partial_run_keeps_processed_head_plus_verbatim_tail, because the new gate returns the original payload. Please update that regression setup to test partial-run behavior with sufficient net savings while retaining coverage of the pass-through gate.

gglucass:
Reproduced: `alpha` x36 + `bureaucratic` + `alpha` x63, drop 36, key `0bb5ef170dc0b9664fe1ab09`, cl100k 100 -> 103 while the result said 99. You are right that no marker-only cost against a word count can establish the invariant; the head word retokenizes.

The gate now measures both complete texts in one unit. `payload_tokens(text)` encodes with cl100k_base (a hard dependency; exact for OpenAI models, documented as an estimate for every other provider) and falls back to one token per character without an encoder. On each path the marked candidate is built, `payload_tokens(original)` and `payload_tokens(candidate + marker)` are compared, the candidate passes through unless it is strictly smaller, and `original_tokens`/`compressed_tokens`/`compression_ratio` are populated from that same measurement. Same code on the batch path and in `kompress_remote.py`. No-CCR results are untouched (word counts, no marker, as before).

Coverage: both boundary sources (yours and the earlier single-token one) pass through on single, batch and remote, each asserting first that cl100k measures the marked candidate larger than the source; a 300-word source dropping 60 is marked on all three paths and reports `(tok(source), tok(marked))`; the unit helper is pinned to tiktoken and to `len(text)` without an encoder. The deadline partial-run regression is parametrized: 200 words keeps the head-plus-verbatim-tail assertions with a real net saving, and the original 20-word shape now asserts the pass-through. Fourteen suites green, mypy clean.

JerrettDavis (approved):
The full-payload gate resolves the marker-accounting findings on single, batch, and remote paths. I re-ran the leading-word boundary repro: the former 100-to-103-token expansion now passes through unchanged with no key. Successful candidates report marker-inclusive counts from the same tokenizer estimate. Deadline coverage now exercises both net-saving partial output and no-saving pass-through. All 81 focused and related tests pass, and Ruff checks pass. I made a documentation-only clarification that cl100k_base is a fixed estimate, not exact billing for every OpenAI model/provider, and the character fallback is not a token bound. CI remains pending on the current head.

Issue #455, opened 2026-05-11
Dashboard bugs: headline savings can show 0%, Compression Quality can exceed 100%, and recent request summary is misleading for backend-routed traffic


SvenMeyer:
Summary
In `headroom 0.21.22`, the dashboard can display misleading or incorrect metrics for backend-routed traffic (in our case `litellm-azure`). I will attach a screenshot later, but the API payload already shows the problem.

Problems observed
1. Headline savings can show `0%` even when compression is happening
2. `Compression Quality` can exceed `100%`
3. `Recent Requests`/summary can look broken or misleading even when recent request logs are present

Root causes we found
1. The headline uses `stats.tokens.active_savings_percent`
   - In our backend-routed OpenAI/LiteLLM Azure path, this can become `0` because `proxy_attempted_tokens` / `attempted_input_tokens_total` is not populated consistently.
   - Meanwhile `proxy_savings_percent` and the underlying request logs clearly show non-zero compression.

2. `Compression Quality` is mislabeled / computed from:
   - `totalWaste / saved`
   - This can legitimately exceed `100%`, so presenting it as a quality/confidence percentage is misleading.
   - In the dashboard template, the detail string literally reports e.g. `X% of removed tokens were identified waste`, which is not a bounded “quality” metric.

3. Backend-routed traffic previously needed extra patching for `recent_requests`
   - We locally use a patch derived from PR #391 so backend-routed OpenAI/LiteLLM traffic appears in `/stats.recent_requests` and streaming output tokens are tracked.
   - Even with logs present, the dashboard-level summary still makes the request activity look much worse than it is.

Evidence
Example live `/stats` behavior we observed:
- `tokens.saved`: non-zero
- `proxy_savings_percent`: non-zero
- `recent_requests`: populated
- yet the dashboard headline can still show `0%`

Relevant code paths
- `headroom/proxy/server.py`
  - exposes both `active_savings_percent` and `proxy_savings_percent`
- `headroom/dashboard/templates/dashboard.html`
  - headline currently renders `active_savings_percent`
  - Compression Quality widget uses waste/saved-derived confidence text

Suggested fixes
- Use `proxy_savings_percent` as fallback when `active_savings_percent` is zero/invalid/missing
- Or better: only show the “active” metric when `proxy_attempted_tokens > 0`
- Rename or redesign `Compression Quality` so it is not presented as a bounded quality percentage if it can exceed 100
[...]

PR #1653, opened 2026-07-01
fix(dashboard): align token savings headline denominator

chris-yyau:
## Description
Fixes a dashboard denominator mismatch in the Token Savings card.

The headline was showing the active attempted-token ratio, while the same card's sublabel reports total-wire savings. This made sessions show values like about 17% in the headline and about 1.2% in the total-wire line for the same saved-token count.

This changes the headline to use `stats.tokens.savings_percent`, with `proxy_savings_percent` as a fallback, so the headline and card copy use the same denominator.

Closes #

## Type of Change
- [x] Bug fix (non-breaking change fixes issue)
- [ ] New feature (non-breaking change adds functionality)
- [ ] Breaking change (fix or feature cause existing functionality change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made
- Updated `headlineSavingsPercent` to prefer the total-wire `savings_percent` metric.
- Updated the headline tooltip to say `Of total wire input tokens`.
- Added a focused dashboard regression test that prevents the headline getter from using `active_savings_percent` or `proxy_attempted_tokens`.

## Testing
- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added new functionality
- [x] Manual testing performed

### Test Output

```text
$ python3 - <<'PY'
from pathlib import Path
html = Path('headroom/dashboard/templates/dashboard.html').read_text(encoding='utf-8')
assert 'stats.tokens?.savings_percent' in html
assert 'Of total wire input tokens' in html
start = html.index('get headlineSavingsPercent()')
end = html.index('get headlineSavingsTitle()', start)
headline = html[start:end]
assert 'active_savings_percent' not in headline
assert 'proxy_attempted_tokens' not in headline
print('dashboard headline denominator check passed')
PY
dashboard headline denominator check passed

$ uv run --extra dev pytest tests/test_dashboard_token_savings.py
============================= test session starts ==============================
platform darwin -- Python 3.13.3, pytest-9.0.3, pluggy-1.6.0
collected 1 item
tests/test_dashboard_token_savings.py::test_token_savings_headline_uses_total_wire_denominator PASSED [100%]
============================== 1 passed in 0.10s ===============================

$ uv run --extra dev ruff check tests/test_dashboard_token_savings.py
All checks passed!
```

## Real Behavior Proof
[...]

PR #3485, opened 2026-09-08
feat(savings): show the new-input savings basis beside the whole-wire figure

gglucass:
## Description

The proxy already computes two savings bases in `/stats`: `savings_percent` (saved over everything forwarded, which recounts a session's cached history every turn and drifts toward 0% on long sessions) and `new_input_savings_percent` (compression-only saved over the tokens that newly entered context: provider-billed uncached + cache-write). Only the first one is shown anywhere: the dashboard headline and `headroom savings` both render the whole-wire figure. A user on a 300-turn Claude Code session sees 0.4% while compression is working on every new turn, and a user comparing the CLI with a client that reads the new-input field sees two unexplained numbers.

This surfaces the existing new-input figure as a second labelled view next to the whole-wire one. No calculation changes.

- Dashboard: the headline "Tokens Saved" percentage leads with `new_input_savings_percent` when the provider reported a cache breakdown (`new_input_tokens > 0`), with a title explaining the basis; it falls back to whole-wire otherwise (Bedrock, no cache data) so it never collapses to 0% while compression is happening (#455). The "Of total wire" line beneath it is unchanged.
- CLI `headroom savings`: each window line gains `· of new input NN.N%` when that window has cache data, plus a one-line legend. Windows without cache data print exactly what they print today.
- Ledger: `record_savings_event` takes optional `new_input_tokens` and `deferred_tokens`; the proxy call site passes uncached + cache-write and the tool-search deferral share. Events without a cache breakdown write the same line as before and do not contribute to the new-input ratio on either side, so the ratio never pairs one event's savings with another's denominator. Numerator is compression-only (saved minus deferral), matching `/stats`.

## Type of Change

## Changes Made

- `headroom/savings_ledger.py`: optional `new_input` / `deferred` event fields; `_Bucket.new_input_tokens`, `new_input_saved`, `new_input_savings_percent`; both exported in `to_dict()`.
- `headroom/proxy/prometheus_metrics.py`: ledger append passes `new_input_tokens` (None when no cache breakdown) and `deferred_tokens`.
- `headroom/cli/savings.py`: `_window_line` appends the new-input figure when present; legend line.
- `headroom/dashboard/templates/dashboard.html`: `headlineUsesNewInput` gate, headline getter and title; stale "attempted denominator" comment replaced.
- Tests: ledger pairing and no-cache-data isolation, CLI output with and without cache data, proxy call-site kwargs with and without a cache breakdown, dashboard headline order and titles.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_savings_ledger.py tests/test_dashboard_token_savings.py tests/test_savings_ledger_before_forwarded.py tests/test_savings_ledger_offload.py tests/test_proxy_project_savings.py tests/test_dashboard -q
46 passed, 3 skipped
uvx ruff check headroom/savings_ledger.py headroom/proxy/prometheus_metrics.py headroom/cli/savings.py tests/test_savings_ledger.py tests/test_dashboard_token_savings.py tests/test_savings_ledger_before_forwarded.py
All checks passed!
uv run --frozen --extra dev mypy headroom/savings_ledger.py headroom/cli/savings.py headroom/proxy/prometheus_metrics.py
```

## Real Behavior Proof

- Environment: macOS 15, Python 3.12, headroom-ai 0.37.0 as installed by Headroom Desktop, real Claude Code traffic on 2026-09-08.
- Exact command / steps: Summed PERF lines for one opus-5 session (354 requests, 53.7M tokens forwarded, 45.0M cache reads, 1.17M cache writes) and compared the two bases; ran `headroom savings` through `CliRunner` against a temp ledger with one event carrying `new_input_tokens=900` and one without.
- Observed result: whole-wire 0.42% versus new-input 2.9% on the same session, which is the gap this view explains. CLI output: the window line without cache data is byte-identical to today; with cache data it ends in `· of new input 30.8%` and the legend prints once.
- Not tested: dashboard rendering in a browser (the template getters are covered by the existing regex tests; no visual change beyond the number and its tooltip).

## Runtime Rollout Safety

- Stable/default behavior changed: Yes. The dashboard headline percentage and `headroom savings` lines show the new-input basis when cache data exists; the whole-wire figure stays visible beside it. Ledger lines gain two optional fields; old lines still parse.
- Kill switch / disable path: None; both figures are shown, nothing is hidden.
- Qualification impact: None.
- Rollback path: Revert the commit; ledger lines with the extra fields are ignored by the previous reader.

## Review Readiness

## Checklist

N/A

## Additional Notes

Why a second view and not a toggle: the two numbers answer different questions ("how much of what I sent was cut" versus "how much of what I newly added was cut") and hiding either one is what caused the confusion. Documentation: the CLI legend and the dashboard tooltip carry the definition; happy to add a docs line if you point me at the page.


JerrettDavis (changes requested):
The new-input basis needs consistent request coverage before becoming a headline. I reproduced two mismatches using the actual PrometheusMetrics.record_request → savings ledger path.

1. [P2] Zero-saving requests disappear from the new ledger denominator. A request saving 100 tokens with 100 newly billed input tokens, followed by a request saving 0 with 10,000 newly billed input tokens (both with cache-read activity), produces CLI/ledger new-input savings of 50.0%, while the matching dashboard ratio is 0.98%. Both the ledger_saved > 0 call-site guard and record_savings_event's saved <= 0 early return drop the second denominator. Please retain denominator-only observations for this basis (without changing historical saved-event semantics accidentally) and test the real metrics-to-ledger path with a zero-saving request.

2. [P2] The dashboard getter now promotes a ratio whose numerator includes unqualified traffic. Starting with that first qualified request, adding a request saving 10,000 tokens with no cache/uncached breakdown raises the dashboard headline from 50% to 99.02%; the correctly paired ledger basis stays at 50%. /stats currently uses all tokens_saved_total but only cache-qualified new_input_tokens. The new headlineUsesNewInput gate sees one qualified request and exposes the mixed-cohort result for the whole session. Please pair savings and denominator from the same qualified requests before promoting this field, and add a mixed-provider/missing-breakdown regression.

The direction and labeling are useful, and the ledger already handles mixed events better than /stats in the second case. Focused validation: 25 tests passed and 2 skipped. The added string-based dashboard test does not exercise either aggregate mismatch.

gglucass:
Both reproduced and fixed at the metrics-to-ledger and metrics-to-/stats paths, not in the dashboard string.

**1. Zero-saving requests in the denominator.** `record_request` now writes the ledger event when the request has a cache breakdown even if `ledger_saved` is 0, and `record_savings_event` only early-returns on `saved <= 0` when there is no `new_input_tokens`. The aggregate treats such an event as denominator-only: it adds to `new_input_tokens` and returns before touching calls, before/saved or cost, so historical saved-event semantics are unchanged (`savings_percent` in your scenario stays 50 percent while `new_input_savings_percent` reads 0.98). Tests: the real `PrometheusMetrics.record_request` path with a zero-saving cache-read request reaching the ledger, a no-breakdown zero-saving request still skipped, and the 100/100 then 0/10,000 aggregate.

**2. Paired numerator for /stats.** `cache_by_provider` gains `new_input_saved_tokens`, accumulated in the same branch that fills `uncached_input_tokens`/`cache_write_tokens`, summed into the prefix-cache totals, and `new_input_savings_percent` now divides that by the same cohort's denominator instead of `tokens_saved_total`. Regression: a qualified request at 50 percent followed by a Bedrock request saving 10,000 with no breakdown stays at 50.0 in /stats while the whole-wire figures still count it. Since the dashboard reads that field, the `headlineUsesNewInput` gate now promotes a paired number.

Ledger 21 passed, before-forwarded 5, stats new-input 3, dashboard 1, plus cache TTL metrics, cost card attribution, dual-ruler funnel and cache telemetry green; mypy clean on the ledger, metrics and cost modules.

chopratejas:
Strong agreement with the problem statement: `new_input_savings_percent` has been computed and dark, and "0.4% on a 300-turn session while compression is working every turn" is a real and bad experience. Surfacing it is right.

One concern about which of the two figures leads, and a smaller note.

### Leading the headline with the new-input basis

`646e7055` (#1653) moved the headline *off* a compressible-only denominator for what looks like the same reason in reverse: the headline and the card copy disagreed on the same saved-token count — "about 17% in the headline and about 1.2% in the total-wire line". This PR restores that split. On my local ledger (13,328 requests) it's a wider gap than #1653's:

| basis | value |
|---|---|
| whole wire | 5.15% |
| new input | 26.81% |

The gap is entirely cache reads — 158,719,203 of 186,324,159 denominator tokens (85.2%) are cache-read.

The mitigation here is labelling both, which is a real improvement over the pre-#1653 state. My worry is that the label doesn't travel. The headline is the figure that gets screenshotted, quoted in a thread, and pasted into a comparison; the card is titled "Tokens Saved", so a leading 26.8% reads as "Headroom saved 26.8% of my tokens." That isn't what it means — it's efficacy on the content compression could touch. The whole-wire figure is the honest answer to "what fraction of tokens did I save", because without Headroom the cached history would have been billed too (as cache reads) and so belongs in the denominator.

Concretely, I'd suggest the same two numbers in the opposite order — whole wire stays the headline, new input goes on the line below with its basis spelled out. That keeps #1653's invariant, keeps the conservative number in the screenshot-able slot, and still fixes the 0.4% problem for anyone who reads the card.

Entirely your call, and I may be over-weighting the screenshot risk.

### The "Of total wire" line is currently a duplicate

Worth fixing whichever direction this goes: since #1653 aligned the headline onto `savings_percent`, that sublabel renders the *same field* as the headline — 1dp in one, 2dp in the other. Post-merge here it stops being a duplicate, so this resolves incidentally, but it's the reason that slot is free.

### Question on the ledger

Not a blocker, and possibly my misreading — since this PR touches `savings_ledger.py`: on my machine `savings_events.jsonl` sums to 64.7M tokens saved while `proxy_savings.json` `lifetime.tokens_saved` reads 10.1M, ~6.4x apart. I assume they're scoped to different layers (deferral in one, not the other), but if the new-input numerator is derived from whichever of those is narrower, the published ratio may not mean what the label says. Might be worth a sanity check while you're in there.

Happy to be wrong on the headline ordering — the underlying change is clearly needed either way.

https://claude.ai/code/session_01SkV95XewUxjYQzbZ7Z3Vjk

gglucass:
Taking your ordering - pushed as c80abc08. You are right, and the PR was
internally inconsistent about it besides: the CLI hunk already led with whole
wire and appended `· of new input X%`, so the dashboard was the only surface
inverting it. Both agree now.

What the commit does:
- Headline reverts to `savings_percent` (whole wire), title back to naming that
  basis, and `headlineUsesNewInput` is gone. Net -2 lines in the dashboard.
- The new-input figure takes the sublabel slot, gated on
  `new_input_tokens > 0`. That resolves your second point directly rather than
  incidentally: the old `Of total wire` line was rendering the headline's own
  field at 2dp against the headline's 1dp, so the slot was free.
- The card shows the number and its basis only. The explanation of what the
  basis means ("what compression can touch", and why the headline reads low on
  a long session) is in the tooltip, not on the card.

One thing worth putting in front of you, since it is the same failure #1653
fixed and I would rather you weigh it than find it: Extra Headroom's desktop
app leads with the new-input basis on its own dashboard, deliberately. With
this ordering a user running both sees roughly 5% upstream and roughly 27%
downstream from the same machine on the same day. I am not asking to reopen
the call - the screenshot argument is the stronger one for a card titled
"Tokens Saved", and I would rather be conservative in the slot that travels.
But that divergence is now real and downstream's to reconcile, not yours.

## On the ledger gap

Not misreading, and it has an answer. The ledger numerator is
`tokens_saved + deferral_saved` (`PrometheusMetrics.record_request`, just above
the `record_savings_event` call), while `proxy_savings.json`
`lifetime.tokens_saved` is compression only. Tool-schema deferral rides the
cached prefix and is large on MCP-heavy setups, which is the direction of your
6.4x. It is setup-dependent rather than systematic: on my ledger the two are
1.222B and 1.322B, the ledger slightly *smaller*, so no deferral pressure here
at all. (It is also not `--reset`, which deletes the jsonl but leaves
`proxy_savings.json`, so that skews the other way.)

For the published ratio specifically it does not contaminate, and the PR is
explicit about it:
- `_Bucket.add` does `new_input_saved += max(saved - deferred, 0)`, so deferral
  is subtracted from the new-input numerator. Same definition /stats uses.
- Pre-PR events carry neither `new_input` nor `deferred`, so they contribute to
  neither side of the new-input basis. No backfill, no half-counted history.
- The whole-wire `savings_percent` is unchanged and still counts deferral in
  both numerator and `before`, exactly as it did before this PR.

So the narrower figure is the one the new-input label names. Good catch to make
me check it though - if `deferred` had not been threaded through, the ratio
would have divided prefix-riding savings by a new-input denominator they never
entered.

JerrettDavis (changes requested):
The previous denominator examples are addressed, and all 33 focused tests pass. One remaining eligibility mismatch still makes the two advertised new-input figures disagree.

[P2] headroom/proxy/prometheus_metrics.py:992 admits uncached-only usage to the ledger, while the cache_by_provider path around lines 851-863 requires a cache read or write. Through actual record_request calls and /stats, a request with 100 saved / 100 new / 1000 cache-read tokens followed by one with 0 saved / 10000 new / no cache activity leaves the dashboard at 50%, while the ledger reports about 1%. Please share the same eligible request cohort and paired numerator/denominator across both paths, including valid uncached-only and cache-read-only usage, and cover this variation with an integrated regression.

gglucass:
Fixed in 795a191e.

You were right about the cohort split. The ledger writes when a request newly BILLED input (`uncached_input_tokens or cache_write_tokens`); /stats accumulated its pair inside `cache_by_provider`, gated on cache ACTIVITY (`cache_read_tokens > 0 or cache_write_tokens > 0`). So the dashboard dropped uncached-only requests entirely and credited cache-read-only requests' savings to a numerator with no matching denominator.

The new-input pair now lives on the metrics object as `new_input_tokens_total` / `new_input_saved_tokens_total`, incremented on exactly the predicate the ledger call site uses, and /stats reads that total instead of re-deriving it from `uncached_input_tokens + cache_write_tokens` off the cache rows.

I deliberately did not widen `cache_by_provider`: its `uncached_input_tokens` is also the cache hit-rate denominator, so widening it would move a cache statistic to fix a savings one, and `cost.py` skips providers with `requests == 0` anyway. `new_input_saved_tokens` was per-provider only in this PR and `totals` was its only reader, so nothing lost a consumer.

Regression: `test_stats_and_ledger_share_one_new_input_cohort` records your exact case through real `record_request` calls plus a third cache-read-only turn, then compares `/stats` against `aggregate_savings()` over the ledger the same requests wrote. Both report `new_input_tokens == 10_100` and ~0.98%, and the cache-read-only turn's 500 saved tokens still count on the whole-wire figure. It fails on the pre-fix tree with `assert 100 == 10100`.

gglucass:
Change request addressed and pushed — ready for review again.

JerrettDavis (approved):
Re-reviewed the current head after its main update. The revised calculation now keeps numerator and denominator within the same cohort, addressing the prior correctness concern, and the focused regression suite passes locally (4 passed). The current head is conflict-free and all attached checks are complete with no failures. Approved.

JerrettDavis (approved):
Reviewed exact head 1afa22e442. The new-input numerator and denominator now share the same billed-input cohort across ledger and /stats, zero-saving denominator events are retained, deferral is excluded from the new-input numerator, and whole-wire remains the conservative headline. Current-main merge simulation was conflict-free; 35 focused tests passed (2 optional skips), Ruff and mypy passed, and hosted checks are green. Approving.

PR #3487, opened 2026-09-08
fix(router): scope per-request options so concurrent requests cannot collide

### LINKED ISSUE #3486: [BUG] Concurrent requests cross-contaminate compression options — one shared `ContentRouter` stores per-request state on `self`
## Description

`HeadroomProxy` builds a single `ContentRouter` instance and reuses it for every request, but
`ContentRouter.apply()` stores that request's options on the instance (`self._runtime_*`) with no
lock. Compression runs on a multi-threaded executor, so two overlapping requests interleave inside
`apply()` and the second one's options silently overwrite the first one's.

The result is that a request can be compressed under a different `target_ratio`,
`force_kompress`, `skip_kompress`, `kompress_model` or `compression_policy` than it asked for.
Nothing raises — the request succeeds, the response is just compressed under the wrong policy and
accounted under the wrong numbers.

The three pieces:

1. **One shared instance.** `headroom/proxy/server.py:965` builds `anthropic_router`, and
   `openai_router` is *the same object* whenever the two kompress flags resolve equal. The comment
   at `server.py:945` states the reuse is deliberate ("we reuse ONE ContentRouter instance and the
   Kompress model still loads once").

2. **Per-request state on `self`.** `headroom/transforms/content_router.py:4817-4832` assigns
   `self._runtime_target_ratio`, `self._runtime_force_kompress`, `self._runtime_skip_kompress`,
   `self._runtime_kompress_model` and `self._runtime_compression_policy` from `kwargs` on every
   `apply()` call. The in-code comment says "Store runtime options on self for access by
   `_route_and_compress_block`". There is no lock around any of it — the only `threading.Lock` in
   the module guards the compression cache.

3. **Genuinely concurrent.** `server.py:1125-1133` sizes `_compression_executor` at
   `os.cpu_count()` by default, and `_run_compression_in_executor` does no serialization (its
   `_compression_metrics_lock` usage is gauge bookkeeping only). The dedicated background
   compression thread (`server.py:1153`) can also be inside `apply()` on the same instance at the
   same time.

Two further consequences worth flagging:

- `_runtime_target_ratio` is part of the **compression cache key** (`content_router.py:5364`,
  `6320`, `6391`). A block compressed under request B's ratio can be cached under request A's,
  so the mis-attribution can outlive the race.
- `_runtime_compression_policy` drives TOIN read-only behaviour (`content_router.py:2113-2115`),
  so a PAYG request can pick up a subscription request's policy and vice versa.

## To Reproduce

1. Check out `main` (reproduced at `71068639`; `server.py:965` and `content_router.py:4817` are
   identical on `upstream/main`).
2. Save the script under **Code Sample** as `repro_router_race.py`.
3. `python repro_router_race.py`

The script drives `ContentRouter` directly rather than through the proxy, because the proxy path
needs two concurrent in-flight requests to line up. The probe on `router.compress` only makes the
interleaving deterministic — it does not create the sharing, which is what `server.py:965` already
does in production.

## Expected Behavior

Request A, which passed `target_ratio=0.10, force_kompress=False`, is compressed with
`target_ratio=0.10, force_kompress=False`. Per-request options should not be visible to any other
request.

## Actual Behavior

Request A is compressed with request B's `target_ratio=0.90, force_kompress=True`.

## Code Sample

```python
"""Two concurrent ContentRouter.apply() calls on ONE router instance.

Request A asks for target_ratio=0.10 / force_kompress=False.
Request B asks for target_ratio=0.90 / force_kompress=True.
A observes B's options because apply() stores them on `self`.
"""
import json
import threading

class Tok:
    def count_text(self, t):
        return max(1, len(str(t)) // 4)


# One router, as HeadroomProxy.__init__ builds it (server.py:965).
router = ContentRouter(ContentRouterConfig(lossless=True, enable_kompress=False))

payload = json.dumps(
    [{"id": i, "name": f"row-{i}", "status": "ok", "value": i * 3} for i in range(200)]
)
messages = lambda: [
    {"role": "user", "content": "go"},
    {
        "role": "user",
        "content": [{"type": "tool_result", "tool_use_id": "t1", "content": payload}],
    },
]

a_entered = threading.Event()
b_done = threading.Event()
observed = {}
orig_compress = router.compress


def probe(*args, **kwargs):
    if threading.current_thread().name == "A" and "A" not in observed:
        a_entered.set()
        b_done.wait(10)  # let request B run its own apply() to completion
        observed["A"] = (router._runtime_target_ratio, router._runtime_force_kompress)
    return orig_compress(*args, **kwargs)


router.compress = probe


def request_a():
    router.apply(messages(), Tok(), target_ratio=0.10, force_kompress=False)


def request_b():
    a_entered.wait(10)
    router.apply(messages(), Tok(), target_ratio=0.90, force_kompress=True)
    b_done.set()


ta = threading.Thread(target=request_a, name="A")
tb = threading.Thread(target=request_b, name="B")
ta.start(), tb.start()
ta.join(15), tb.join(15)

print("request A passed:          target_ratio=0.10, force_kompress=False")
print(f"request A compressed with: target_ratio={observed['A'][0]}, force_kompress={observed['A'][1]}")
assert observed["A"] == (0.10, False), "request A used request B's compression options"
```

## Error Output

```
request A passed:          target_ratio=0.10, force_kompress=False
request A compressed with: target_ratio=0.9, force_kompress=True
  File "repro_router_race.py", line 70, in <module>
    assert observed["A"] == (0.10, False), "request A used request B's compression options"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: request A used request B's compression options
```

## Environment

- **Headroom version**: 0.38.0-dev (commit `71068639`; verified identical on `upstream/main`)
- **Python version**: 3.13.11
- **OS**: macOS 26.6.2
- **LLM Provider**: provider-independent — the shared router serves both the Anthropic and OpenAI
  pipelines
- **`os.cpu_count()`**: 16 (default `compression_max_workers`)

## Additional Context

Possible fixes, roughly in order of size:

- Pass the per-request options down the call chain (`apply` → `compress` →
  `_apply_strategy_to_content`) instead of via `self`. Largest diff, removes the class of bug.
- Bundle them into a small per-call context object threaded through the same path — same idea,
  one parameter instead of six.
- Give each pipeline its own router and hold a lock across `apply()`. Smallest diff, but it
  serialises compression and gives up the executor's parallelism, so probably not what you want.

Not addressed here, but adjacent: the cached entries keyed on a mis-attributed
`_runtime_target_ratio` would need invalidating for any fix to be complete.

I have not opened a PR — happy to if you have a preference among the above.

Searched open and closed issues before filing; #1171 and #3178 touch the same executor but not
this. Found while reviewing the largest symbols in the codebase, not from a production incident,
so I have no field report of the resulting compression-quality or savings-accounting drift.

trs-80:
## Description

The proxy builds ONE `ContentRouter` and shares it across every request, while compression runs on an `os.cpu_count()`-wide executor with no serialization. `apply()` stored that request's options as plain instance attributes, so two overlapping requests silently overwrote each other's compression settings — the request succeeded, it was just compressed under another request's policy.

Closes #3486

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added a `_PerRequestOption` descriptor in `headroom/transforms/content_router.py` backing each per-request option with a `ContextVar`, and declared the five options (`_runtime_target_ratio`, `_runtime_force_kompress`, `_runtime_skip_kompress`, `_runtime_kompress_model`, `_runtime_compression_policy`) on `ContentRouter`.
- Built each option's `ContextVar` eagerly in `__init__` so two concurrent first writes cannot race to create two vars for the same option.
- Started `apply()`'s own compression fan-out through `contextvars.copy_context()` — the single-task watchdog thread and the parallel `ThreadPoolExecutor` — so those workers still inherit the options of the request that spawned them. One context per task, since a `Context` cannot be entered by two threads at once.
- Added two regression tests to `tests/test_transforms_content_router.py`.

No call site changed. Every existing `self._runtime_x` read and write keeps working, including the save/override/restore around a call in `transforms/compression_batches.py` and the four test files that read these attributes directly.

## Testing

### Test Output

New regression test fails on `main` and passes with the fix:

```text
$ git stash && python -m pytest tests/test_transforms_content_router.py \
    -k "per_request_options or fanout_inherits" -q
tests/test_transforms_content_router.py F.                               [100%]
=================================== FAILURES ===================================
___________ test_concurrent_apply_keeps_per_request_options_isolated ___________
tests/test_transforms_content_router.py:1883: in test_concurrent_apply_keeps_per_request_options_isolated
    assert observed["a"] == (0.10, False), (
E   AssertionError: request A compressed with request B's options: (0.9, True)
E   assert (0.9, True) == (0.1, False)
================== 1 failed, 1 passed, 59 deselected in 3.20s ==================

$ git stash pop && python -m pytest tests/test_transforms_content_router.py \
    -k "per_request_options or fanout_inherits" -q
tests/test_transforms_content_router.py ..                               [100%]
======================= 2 passed, 59 deselected in 0.47s =======================
```

Tests that read these attributes directly, all passing:

```text
$ python -m pytest tests/test_transforms_content_router.py tests/test_force_kompress_all.py \
    tests/test_compression_policy_toin_gate.py tests/test_agent_savings.py -q
======================= 110 passed, 2 warnings in 4.54s ========================
```

Broader compression selection — identical failure set before and after, so no regressions:

```text
$ python -m pytest tests/ -k "router or compress or kompress or batch" \
    --ignore=tests/test_integrations/langchain -q
= 19 failed, 2218 passed, 63 skipped, 9953 deselected in 272.71s (0:04:32) =

$ diff before.txt after.txt && echo IDENTICAL
IDENTICAL
```

The 19 are pre-existing on `upstream/main` in my environment (verified by running the same selection with the patch stashed); `tests/test_integrations/langchain/test_langchain_1x_compat.py` also fails to collect on a clean tree with `NameError: name 'tool' is not defined`, unrelated to this change.

```text
$ python -m ruff check headroom/transforms/content_router.py tests/test_transforms_content_router.py
All checks passed!
$ python -m mypy headroom/transforms/content_router.py
```

## Real Behavior Proof

- Environment: macOS 26.6.2, Python 3.13.11, `os.cpu_count()` = 16, branched from `upstream/main` at e67b3c8a.
- Exact command / steps: `python -m pytest tests/test_transforms_content_router.py -k "per_request_options or fanout_inherits" -q`, run once with the patch stashed and once with it applied. Two threads call `apply()` on one shared router, A with `target_ratio=0.10, force_kompress=False` and B with `target_ratio=0.90, force_kompress=True`; A is held inside `compress` until B's `apply()` returns, then reads back the options it is compressing under.
- Observed result: on `main`, A compresses with `(0.9, True)` — request B's options. With this patch, A compresses with `(0.10, False)`. Full output in Test Output above.
- Not tested: no end-to-end proxy run under concurrent live traffic, and no measurement of the resulting compression-quality or savings-accounting drift in production — the bug was found by code review, not from a field report. Stale compression-cache entries keyed on a mis-attributed `_runtime_target_ratio` are not invalidated by this patch. `headroom/transforms/smart_crusher.py` carries the same `_runtime_compression_policy` pattern on a separate class and is left alone here.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — this is a correctness fix to existing behavior, not behind a rollout channel.
- Stable/default behavior changed: yes, on the default path, and only in the concurrent case — a request now sees the options it was called with instead of possibly another request's. Single-request and sequential behavior is unchanged, including the fact that values persist after `apply()` returns.
- Kill switch / disable path: none added; reverting the commit restores the previous behavior.
- Qualification impact: none expected — the change is confined to how five existing options are stored; no option's meaning, default, or public signature changed.
- Rollback path: revert this commit. No migration, no persisted state, no config.

## Review Readiness

## Checklist


JerrettDavis (approved):
Reviewed the context-scoped descriptors, per-router isolation, batch override/restore path, and both internal worker paths. All 71 router and compression-batch tests pass. An additional probe using string tool messages exercised the single-task watchdog and multi-task ThreadPoolExecutor and verified all five request options survive into those workers. No blocking code findings.

One test improvement: test_compression_fanout_inherits_per_request_options currently uses Anthropic content blocks, which compress inline. Please use role=tool string messages there and assert that the probes run off the calling thread, so the regression actually fails if the executor copy_context propagation is removed.

trs-80:
@JerrettDavis Confirmed and fixed in d01cf182. The Anthropic `tool_result` blocks compressed inline on the calling thread, so the test passed even with `copy_context()` deleted from the `executor.submit` — I verified that by deleting it and watching the test stay green.

Rewritten with `role="tool"` string messages (the only content that reaches Pass 2's `pending_tasks`) plus an assertion that the probes ran off the calling thread. Mutation check now: with `copy_context()` the test passes; without it, it fails with `[(6138834944, None, None), ...]` — four pool threads, options lost. Full file: 61 passed.

JerrettDavis (approved):
Re-reviewed after refreshing from main. The ContextVar descriptors isolate all five runtime overrides, and both the watchdog and fan-out workers enter independent copied contexts, so request settings survive internal thread boundaries without cross-request leakage. The post-update concurrency regressions pass locally (2/2); the head is current, mergeable, and attached checks pass. No blocking finding.

trs-80:
Superseded by #3556, which fixed #3486 with the same ContextVar + copy_context approach and additionally scopes the tool-call maps and read-protection sets. Rebasing this onto main leaves no source diff, and the two tests here pass against #3556's implementation while duplicating tests/test_content_router_concurrency.py. Closing.

PR #2304, opened 2026-07-16
fix(openclaw): delegate native compaction

GautamSharma99:
## Description

Fix the OpenClaw context engine claiming ownership of persistent compaction while its `compact()` implementation performed no compaction.

Previously, `ownsCompaction` was `true`, which disabled OpenClaw's built-in automatic compaction. However, `compact()` did not read or rewrite
the session transcript and still returned `{ ok: true, compacted: true }`.

Headroom now declares that it does not own persistent compaction and delegates `/compact` and overflow-recovery requests through OpenClaw's
official `delegateCompactionToRuntime(...)` API. Headroom continues to perform zero-cost compression while assembling model input.

OpenClaw's context-engine contract requires non-owning engines to delegate `compact()` explicitly; setting `ownsCompaction: false` alone does
not restore manual or overflow-recovery compaction:

https://docs.openclaw.ai/concepts/context-engine#ownscompaction

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Changed `HeadroomContextEngine.info.ownsCompaction` from `true` to `false`.
- Replaced the successful no-op `compact()` implementation with OpenClaw native-runtime delegation.
- Added a small runtime adapter that dynamically loads `openclaw/plugin-sdk/core` without bundling the optional OpenClaw peer dependency.
- Return the real delegated compaction result instead of manufacturing a successful result.
- Increment compaction statistics only when OpenClaw reports that compaction occurred.
- Propagate native compaction failures so OpenClaw can apply its normal failure-isolation behavior.
- Removed stale claims that RollingWindow drops messages during Headroom compression.
- Updated the OpenClaw README to distinguish input compression from persistent compaction.
- Added an Unreleased changelog entry.
- Added regression coverage for delegated success, delegated no-op, and delegated failure behavior.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
$ cd plugins/openclaw
$ npx vitest run test/engine.test.ts

Test Files  1 passed (1)
Tests       13 passed (13)

$ npm run typecheck
> tsc --noEmit
Passed

[...]

PR #3488, opened 2026-09-08
feat(openclaw): 2026.9.x turn contract, multi-upstream routing, tool-payload-safe compression, opt-in durable compaction

GenorTG:
## Description

**Goal:** make the Headroom OpenClaw plugin work correctly on OpenClaw 2026.9.x and be safe for tool-heavy agent sessions, **without changing any default that upstream `main` has chosen**. Everything in this PR is either (a) a fix for behavior that is currently broken on the 2026.9.x runtime, or (b) opt-in and off by default.

The plugin on `main` is a good design; this PR is the result of running it in production against a multi-provider OpenClaw gateway and fixing what broke, then hardening those fixes through review rounds (including a post-approval production fix for frozen `api.config`). I have tried to keep each change small, tested, and traceable to a concrete failure.

### What differs from `main` today, and why

| Area | `main` today | This PR | Why |
|------|--------------|---------|-----|
| **Turn contract** | Engine does not declare `transcriptSemantics` or implement `commitTurn()` | Declares `currentTurnFence` + `turnAdvancementIdempotency`; `commitTurn()` durably records the accepted turn (key + SHA-256 digest of `messages`, no message bodies) keyed by `advancementKey`, returns `duplicate` on retry / after restart | On 2026.9.x the runtime logs `Context engine "headroom" degraded to "legacy" for this logical turn` every turn and never calls `assemble()`. Without this, the plugin does nothing. |
| **Turn-store locking** | n/a | `store-lock.ts`: ownership published atomically (hard-link publish, `O_EXCL` fallback); a lock is stale only when *old* **and** its owner is provably gone (EPERM counts as alive); recovery via atomic `rename`; release never unlinks someone else's lock | Round-4 review reproduced a live lock being unlinked during the `open(wx)` → PID-write window. |
| **Tool payloads through the proxy** | `extractText()` flattens tool content; images dropped; `toolName` falls back to `"headroom"` | Non-text blocks never leave the process — a short placeholder goes on the wire — and are restored from the local originals by `tool_call_id` / position. OpenAI `tool.name` set from `toolName`. | Vision/browser/`view_image` results were silently lost; the proxy's protect list keys on tool names. An intermediate design that sent blocks as base64 text inflated a 60 KB image to ~50k tokens and depended on the proxy echoing `_headroomMeta`, which it does not do reliably — replaced. |
| **Deferred tool wrappers** | OpenClaw Tool Search sends every MCP call as `tool_call`; the proxy's protect/exclude lists key on that name | `agentToOpenAI` resolves `{id: "mcp:<server>:<server>__<tool>"}` / Hermes `{name}` to `mcp__<server>__<tool>` on the wire (wrapper restored from originals); opt-in `protectToolResults` restores named tools' results verbatim after every compress round trip | Per-tool protection was impossible for MCP tools; prose outputs (video/vision analysis) came back paraphrased |
| **Durable compaction** | `ownsCompaction: false`, `compact()` → `delegateCompactionToRuntime()` (#2304) | **Same default** (`persistentCompaction: "openclaw"`). Opt-in `"hybrid"` (Headroom replace-only pre-pass + turn-end hygiene, OpenClaw still summarizes) and `"headroom"` (zero-LLM `/v1/compress` rewrite, `ownsCompaction: true`) | Upstream's delegation is correct and stays the default. The opt-in modes exist for very large tool-heavy transcripts where LLM `/compact` is slow or fails. |
| **Truncation boundary** | n/a | `truncate-boundary.ts`: the durable tail always starts at a turn start (OpenClaw's own `isTurnStartMessage` rule), never at a `toolResult` or assistant continuation; orphan tool results dropped; `resetLeaf()` used so the prefix is actually removed | Round-3/4 review: `branch(parentId)` retained the prefix, and a raw suffix cut could start at an orphan `toolResult`. |
| **Multi-upstream routing** | One `*_TARGET_API_URL` per proxy | Opt-in `providerUpstreams` → `x-headroom-base-url` (the proxy's own documented mechanism); protocol-aware path prefix (`/v1` vs `/v1beta` for Gemini) | Real deployments route several OpenAI-compatible APIs through one proxy; Gemini must keep `/v1beta` to reach `handle_gemini_generate_content` (round-1 review). |
| **Session-gated providers** | n/a | Opt-in `providerSessionHeaders` (per-process UUID under an operator-chosen header) | e.g. opencode-go returns `MissingSessionID` without one. No provider names hardcoded. |
| **Frozen gateway config** | In-place rewrite assumes `api.config` is mutable | Detect frozen/sealed providers; no-op when disk already routes `baseUrl` + `x-headroom-base-url`; clear error if routing still needed | OpenClaw 2026.9.x freezes plugin config; session-header refresh threw `Cannot assign to read only property 'opencode-go'` even when routing was already correct on disk. |
| **Assemble cost** | Proxy called every turn | Skip when rough estimate < `(tokenBudget − assembleReserveTokens) × assembleSkipBudgetRatio` (defaults 20000 / 0.7). OpenClaw hands `assemble()` the full window as `tokenBudget` while its overflow precheck also counts the system prompt and a ≥ 20k reserve, so a flat 85 % rule never fired before native compaction on ~200k windows; opt-in, provider-aware `skipAssembleWhenGatewayRouted`; images estimated at a fixed ~1.5k tokens (not `base64.length / 4`) | Sub-budget turns on 1M-window models paid multi-minute compress for 0 savings; base64 length was defeating the budget check. |
| **Durable compress defaults** | `protect_recent: 0` when hygiene runs | `protect_recent: 2`; replace-mode skips messages with protected tool/image payloads | Avoid rewriting the turn the model is about to answer, and never lossy-rewrite multimodal payloads on disk. |

### Review history

| Round | Finding | Resolution |
|-------|---------|------------|
| 1 | `commitTurn` was a no-op; Gemini `/v1beta` collapsed to `/v1` | `TurnAdvancementStore`; `resolveProxyPathPrefix()` |
| 2 | Failed persist returned `duplicate` on retry; cross-instance clobbering; truncate via `branch()` kept the prefix | Reload-under-lock + atomic write; exclusive lock; `resetLeaf()` |
| 3 | Upstream #2304 flipped to delegation | Adopted as the default; Headroom modes made opt-in |
| 4 | Empty new lock treated as stale; raw suffix cut at orphan `toolResult`; Windows test failures; stale PR body | `store-lock.ts`; `truncate-boundary.ts`; portable paths/URLs in tests; this rewrite |
| 5 | Backward boundary alignment understated `tokensAfter` | `adjustTokensAfterForBoundaryAlignment` + regression (approved 2026-09-14) |
| 6 | Frozen `api.config` TypeError on session-header refresh (`opencode-go`) | Freeze-safe in-place rewrite; no-op when already routed |

Closes # (no upstream issue; surfaced from production use of the plugin on OpenClaw 2026.9.x)

## Type of Change

- [x] Documentation update
- [x] Performance improvement

## Changes Made

All changes are confined to `plugins/openclaw/`. No Python runtime or proxy changes.

**Always-on fixes (no config needed)**
- `engine.ts` — declare `transcriptSemantics`; `commitTurn()` → `TurnAdvancementStore`; budget short-circuit in `assemble()`; CCR retrieve hint only when `ccrHashes` is non-empty; pass local originals to `openAIToAgent`.
- `turn-advancement-store.ts` + `store-lock.ts` — durable, idempotent advancement keyed by `advancementKey`; records hold only a digest of the accepted messages (OpenClaw owns the transcript) and are pruned after 14 days / 5 000 records so the file stays small; v1 files that embedded message bodies are migrated on first commit; reload-under-lock; atomic temp+rename persist; safe cross-process lock protocol (see table).
- `convert.ts`, `content-blocks.ts`, `original-lookup.ts` — placeholder wire format for non-text blocks; originals-based restore (`tool_call_id` → `hrIndex` hint → position); OpenAI `tool.name`; `isError` preserved; fixed per-image token estimate; assistant text no longer duplicated across blocks.
- `compaction.ts` + `truncate-boundary.ts` — turn-aligned truncation, orphan `toolResult` removal, `resetLeaf()`; replace-mode skips protected payloads; `protect_recent: 2` (`compress-request-config.ts`).

**Opt-in (default off / matches upstream)**
- `compaction-mode.ts`, `openclaw-compaction.ts`, `transcript-hygiene.ts`, `hygiene-debounce.ts`, `transcript-projection.ts` — `persistentCompaction: "openclaw" | "hybrid" | "headroom"` (default `"openclaw"` = upstream delegation); `transcriptHygiene` (on only in `hybrid`).
- `gateway-config.ts`, `proxy-routing.ts`, `session-headers.ts` — `providerUpstreams`, `/v1` vs `/v1beta`, `providerSessionHeaders`; freeze-safe in-place rewrite when OpenClaw seals `api.config`.
- `assemble-skip.ts` — provider-aware `skipAssembleWhenGatewayRouted` (reads `runtimeSettings.model.provider`).
- `openclaw.plugin.json` — schema for the new keys; `README.md`, `docs/PR_OVERVIEW.md`, `docs/tool-call-preservation.md`, `docs/TEST_MATRIX.md`, `test/README.md`.

## Testing

The plugin is TypeScript, so the Python commands in the template do not apply; the equivalents are:

- [x] Unit tests pass (`npm test` — vitest; equivalent of `pytest`)
- [x] Linting passes (`npm run typecheck` — `tsc --noEmit`; the plugin has no separate lint config)
- [x] Type checking passes (`npm run typecheck`)

### Test Output

```text
$ cd plugins/openclaw && npm test
 Test Files  22 passed (22)
      Tests  324 passed (324)
$ npm run typecheck        # clean
$ npm run build            # dist/index.js 90.65 KB
$ node test/live-compress-smoke.mjs   # against a running Headroom proxy
Request tool.name: view_image
Response has image block: true
PASS: live compress smoke test — image payload preserved
```

New/updated tests for this round:
- `test/gateway-config.test.ts` — frozen/sealed OpenClaw `api.config` no-ops when already routed; clear error when routing still needed; mutable path still injects session headers.
- `test/store-lock.test.ts` — fresh empty lock is never stolen (**deterministic two-process test**: a child creates the lock with `open(wx)` and holds it before writing its PID; the parent waits ≥ hold time, child observes `STILL_HELD`, both commits land); stale recovery requires age **and** dead owner; EPERM = alive; hard ceiling for PID reuse; release never unlinks a foreign lock.
- `test/truncate-boundary.test.ts` + `test/compaction.test.ts` — the 41-message fixture (assistant `toolCall` + `toolResult` + 39 users) plans and applies to 39 messages starting at `user-0`; multi-call turns are kept whole; a proxy cut landing on a `toolResult` is re-aligned; a transcript with no turn boundary is not truncated.
- Windows portability: `tmpdir()`/`join()` instead of `/tmp` literals, `file://` URL for the child-process dist import, `npm` via shell on win32, `homedir()` fallback for the state dir. I don't have a Windows host — would appreciate a re-run there.

Full coverage map: `docs/TEST_MATRIX.md`.

### Environment.

- OpenClaw 2026.9.x gateway, `plugins.slots.contextEngine: "headroom"`, plugin built from this branch and loaded from `dist`
- Headroom proxy 0.37.x on `127.0.0.1:8787`, `HEADROOM_MODE=token`
- Providers routed through the proxy: an Anthropic-shape portal API, OpenRouter, an OpenCode-compatible API (`HEADROOM_UPSTREAM_ALLOWED_HOSTS` set accordingly)
- Linux, Node 26

### Exact command / steps.

1. `cd plugins/openclaw && npm install && npm test && npm run typecheck && npm run build`
2. Truncation against the **real** OpenClaw `SessionManager` (SQLite store, not the test double): open a temp session, append the 41-message fixture, `planHeadroomCompaction({ force: true })`, `applyCompactionPlan()`, reopen the store and read the branch.
3. Lock boundary: `npx vitest run test/store-lock.test.ts` (spawns the second process).
4. Tool payloads: `node test/live-compress-smoke.mjs` and `node test/live-compress-stress.mjs` against the live proxy; a raw `/v1/compress` probe with a 60 KB PNG `view_image` result before/after the wire-format change.
5. Gateway: load the built plugin, restart, run tool-heavy sessions (browser / vision / file tools), watch proxy and gateway logs.

### Observed result.

- Real `SessionManager`: persisted branch 41 → **39** after apply (`bytesFreed: 507`), first kept message `user-0`, last `user-38`, no `toolResult` / assistant in the reloaded transcript.
- Lock boundary test: parent `commit()` blocked ~400 ms until the child released; child printed `STILL_HELD`; lock file gone afterwards; store contains the parent key. Old protocol would have unlinked the child's lock immediately.
- Wire format: for the same 60 KB image transcript, `tokens_before` dropped from 49,780 (base64 sent twice) to 9,765; `view_image` content on the wire 80,110 → 62 chars; image block restored locally. Live stress (80-message mixed-tool transcript): 23,518 → 5,025 tokens, `isError` and image blocks preserved with and without `_headroomMeta` echo.
- Gateway: the per-turn `degraded to "legacy"` log line disappears once `transcriptSemantics` is declared; `assemble()` fires each turn. Earlier production measurement with this branch: ~7.7 % overall input-token savings, ~9 % per compressed turn, best single turn ~50k tokens saved. With opt-in `hybrid` compaction, a ~232k-token session compacted to ~24k.

### Not tested.

- **Windows.** The three failures from the last review were addressed by reading the tests (hardcoded `/tmp`, non-URL ESM import path, `npm` spawn); not executed on a Windows host.
- OpenClaw < 2026.9.x — older runtimes ignore `transcriptSemantics` and never call `commitTurn()`, so behavior should be unchanged, but I only run 2026.9.x.
- `persistentCompaction: "headroom"` as the *sole* durable compactor in production. Under sustained load I saw `Pending input ownership ended` mid-turn with it, which is why the default stays `"openclaw"` and my own deployment uses `"hybrid"`.
- Remote (TLS) proxy and multi-proxy fleets — single local proxy per gateway only.

## Runtime Rollout Safety

- Rollout-managed feature(s): `providerUpstreams`, `providerSessionHeaders`, `persistentCompaction` (`hybrid`/`headroom`), `transcriptHygiene`, `skipAssembleWhenGatewayRouted`, `assembleCompressConfig` — all opt-in, default off / upstream-equivalent.
- Minimum rollout channel: any OpenClaw 2026.9.x install; no canary needed since defaults match `main` and the always-on parts are fixes for a path that currently does not run.
- Stable/default behavior changed: **No** for compaction (`"openclaw"` = current `main` delegation). **Yes, intentionally**, for: `assemble()` now actually runs on 2026.9.x; tool/image payloads are preserved instead of flattened; sub-budget turns skip the proxy. For operators who already set `gatewayProviderIds` with non-`/v1` providers, the rewritten URL path now normalizes (`/v1`, or `/v1beta` for Gemini) instead of 404-ing.
- Kill switch / disable path: unset the opt-in keys to return to upstream behavior for that feature; `plugins.slots.contextEngine` back to `legacy` disables the engine entirely; reverting the branch restores `main`.
- Qualification impact: compression metrics (CCR, content_router) for 2026.9.x operators become real rather than zero, since `assemble()` now fires.
- Rollback path: `git revert` the merge; the turn-advancement store file (`headroom-turn-advancements.json` next to the session store) is additive and can be deleted safely.

## Review Readiness

## Checklist

- [x] I have made corresponding changes to the documentation (`README.md`, `docs/*`, `openclaw.plugin.json` schema)
- [x] My changes generate no new warnings (`npm run typecheck`, `npm run build` clean)
- [x] I have added tests that prove my fix is effective or that my feature works (324 vitest cases; reviewer fixtures locked as regressions)

## Additional Notes

- **On the compaction fork in the road:** I initially took `ownsCompaction: true` with a real implementation; upstream landed #2304 (delegate to the runtime) in the meantime. I think upstream's default is the right one for stock installs and adopted it. The Headroom-owned modes are kept only as explicit operator opt-ins for cases where LLM `/compact` is impractical. If maintainers would rather not carry those modes, they are isolated (`compaction-mode.ts`, `transcript-hygiene.ts`, `hygiene-debounce.ts`, `transcript-projection.ts`) and can be dropped without touching the always-on fixes.
- **Tradeoffs:** per-process (not per-request) session UUIDs; `/v1` path normalization drops deeper upstream path segments (proxy matches those by query string); the lock waits up to 5 s and then throws rather than stealing — a crashed holder is recovered after 30 s (dead PID) or 10 min (hard ceiling).
- **Scope:** only `plugins/openclaw/`. Nothing in `headroom/` (Python) or the OpenClaw runtime is modified.
- Operator configuration examples for the opt-in features are in `plugins/openclaw/README.md`.


JerrettDavis (changes requested):
Two blockers remain at 656ea3fae44ed6688af4c2ec2f7f5267dc318664.

1. [P1] The advertised atomic-idempotent advancement contract is still a no-op. commitTurn ignores every parameter and always returns committed, including when the same advancementKey is retried in the same engine or a new engine instance. I reproduced all three acknowledgments from the built distribution. The host contract requires atomically storing the accepted messages with the advancementKey and returning duplicate for a committed key; its real payload field is messages, not acceptedTurn. Returning a recognized status clears the host outbox without fulfilling that durability obligation. Please implement the promised durable advancement (with restart/retry and failed-write tests), or use a host-supported compression-only integration that does not claim this contract. Merely changing the acknowledgment shape is insufficient. Contract: https://github.com/openclaw/openclaw/blob/acbff80d68fce81afc695684b0a957661b9ee05a/docs/concepts/context-engine.md

2. [P2] Unconditionally normalizing every provider path to /v1 bypasses the Gemini compression handlers. Routing google currently rewrites https://generativelanguage.googleapis.com/v1beta to http://127.0.0.1:8787/v1. With actual proxy route registration and mocked terminal handlers, POST /v1beta/models/gemini-2.5-flash:generateContent reaches handle_gemini_generate_content, while the rewritten /v1/models/... reaches generic handle_passthrough. This loses the native Gemini compression path even though the new unit expectation passes. Please normalize according to provider protocol and preserve the proxy-supported /v1beta routes for native Gemini; add a routing-to-handler regression rather than only asserting the rewritten string.

Validation: all 85 plugin tests pass, typecheck and build pass; the two probes above use the built plugin and actual proxy routing respectively. The custom OpenAI-compatible routing and session-header features do not remove these default/native-provider and durability requirements.

GenorTG:
Hi @JerrettDavis — thank you for the thorough review on `656ea3f`. Both blockers are addressed in **`9b0f778`** (pushed to `pr-prep`):

**[P1] Durable `commitTurn`**
- `commitTurn()` now uses the real **`messages`** payload (not `acceptedTurn`).
- New `TurnAdvancementStore` atomically persists accepted messages keyed by `advancementKey` to disk.
- Returns `{ status: "duplicate" }` on retry and after gateway restart (new engine instance).
- Tests cover retry, restart, key conflict, and failed persist.

**[P2] Gemini `/v1beta` routing**
- Replaced unconditional `/v1` normalization with protocol-aware `resolveProxyPathPrefix()`.
- `google`/`gemini` rewrite to `http://<proxy>/v1beta` so OpenClaw's `${baseUrl}/models/{model}:generateContent` hits `handle_gemini_generate_content` (not generic passthrough).
- OpenAI-compatible providers still use `/v1`.
- Routing regression test asserts the full generateContent URL matches the Gemini handler path.

All **97** plugin tests pass; typecheck and build clean. Happy to adjust anything else — appreciate the clear contract references.

plugins/openclaw/src/turn-advancement-store.ts
```diff
@@ -0,0 +1,130 @@
...
+    const record: TurnAdvancementRecord = {
+      advancementKey: params.advancementKey,
+      messages: params.messages,
+      messagesDigest: digest,
+      sessionId: params.sessionId,
+      committedAtMs: Date.now(),
+    };
+
+    this.options.injectBeforePersist?.();
+    this.records.set(params.advancementKey, record);
```
JerrettDavis: [P1] Do not acknowledge a failed persistence attempt on retry. records.set happens before persist. Creating a directory at `${storePath}.tmp` makes the actual writeFileSync fail with EISDIR; after removing that obstruction, retrying the same turn returns duplicate even though storePath still does not exist. The host can then clear its outbox for an unpersisted turn. Publish the in-memory record only after successful persistence, or roll it back on every failure, and add a real filesystem-failure/retry test.

plugins/openclaw/src/turn-advancement-store.ts
```diff
@@ -0,0 +1,130 @@
...
+    return "committed";
+  }
+
+  has(advancementKey: string): boolean {
+    this.ensureLoaded();
+    return this.records.has(advancementKey);
+  }
+
+  private ensureLoaded(): void {
+    if (this.loaded) {
```
JerrettDavis, line 243: [P1] Serialize read-modify-write across stores sharing this file. Each instance loads once and later replaces the entire file from its private Map. With two instances at the same path, call has() on both, commit key a through the first, then key b through the second: both return committed, but disk contains only b; retrying a still returns duplicate from stale memory. This does not require simultaneous writes. The default path is shared across sessions and restarts, so use durable transactional storage or a cross-instance/process lock covering reload, duplicate check, and persistence. Add two-instance and two-process preservation tests.

plugins/openclaw/src/compaction.ts
```diff
@@ -0,0 +1,397 @@
...
+  const estimatedAfter = Math.max(
+    Math.floor(tokensBefore * (keepCount / branchMessages.length)),
+    Math.floor(tokensBefore * 0.15),
+  );
+
+  return {
+    mode: "truncate",
+    tokensBefore,
+    tokensAfter: estimatedAfter,
+    truncateParentId: firstKept.parentId,
```
JerrettDavis: [P1] Forced truncation retains the prefix it intends to remove. For the 100-message/900k-token fixture, the actual planner selects parent entry-59 and appends the last 40 messages while reporting 360k tokens. applyCompactionPlan calls branch(parent), which retains that parent's ancestors, so the resulting branch contains the original first 60 messages plus the re-appended last 40: still all 100 messages. Please create the intended compacted branch with appropriate preserved instructions/tool-call boundaries, and verify its reloaded contents and actual reduction through the host session manager. Host branch implementation: https://github.com/openclaw/openclaw/blob/main/src/agents/sessions/session-manager-entries.ts

JerrettDavis (changes requested):
The native Gemini prefix correction addresses the earlier routing issue, and all 97 plugin tests plus typecheck/build pass locally. The new persistence and compaction code still has three reproducible blockers. The filesystem probes below use the actual compiled store; the compaction probe uses the actual planner and the host's branch semantics. Please exercise real failure and persisted-state outcomes rather than only the before-write test hook or mocked call counts.

GenorTG:
Hi @JerrettDavis — thanks again for the careful second pass. All three new P1 items are addressed in **`2706e58`** (pushed to `pr-prep`).

**1. Failed persist must not return `duplicate` on retry**

`commit()` now reloads from disk under lock, writes atomically, and only then updates in-memory state. If persist fails (reproduced with a directory at `${storePath}.tmp` → `EISDIR`), the key is not cached as committed and a retry returns `committed`, not `duplicate`.

**2. Cross-instance / cross-process serialization**

Each `commit()` acquires an exclusive lock (`${storePath}.lock`), reloads the full record set from disk inside the lock, then read-check-writes. Two instances or two processes on the same path now preserve all keys — no last-write-wins clobbering.

**3. Truncate must actually drop the prefix**

`applyCompactionPlan()` now always uses `resetLeaf()` for truncate mode instead of `branch(parentId)`. Per OpenClaw's branch semantics, branching from an ancestor retains the full prefix; `resetLeaf()` starts a fresh tail with only `appendMessages`. Tests verify reloaded branch length and bytes are actually reduced (not just plan/mocked call counts).

**Tests:** 120 plugin tests pass (typecheck + build clean). New coverage includes real filesystem failure/retry, two-instance + two-process preservation, compress-result truncate apply, and a regression guard showing the old `branch(parentId)` path would have retained all 100 messages.

Happy to adjust if anything still looks off on your side.

GenorTG:
Hi @JerrettDavis — quick note on the new upstream change in **`eb4da64`** (`fix(openclaw): delegate native compaction`, #2304), now on `main`.

I understand why upstream flipped `ownsCompaction` to `false` and wired `compact()` → `delegateCompactionToRuntime(...)`: the old `ownsCompaction: true` stub disabled OpenClaw auto-compaction while doing no durable rewrite — a real bug.

**This PR intentionally takes the other fork.** With `ownsCompaction: true` *and* a real implementation (`compaction.ts` + `maintain()`), Headroom handles **both**:

1. **`assemble()`** — zero-LLM, per-turn input compression (SmartCrusher / proxy `/v1/compress` on the in-flight prompt)
2. **`compact()` / `/compact`** — durable SQLite transcript rewrite via the same Headroom proxy (algorithmic, not OpenClaw LLM summarization)

Upstream `main` keeps Headroom on the **assemble path only** and delegates persistent compaction back to OpenClaw native (LLM summarization / safeguard). **This PR would enable Headroom for all compaction**, not just assembly — same proxy, same cost model (no extra LLM calls for compaction), but actually shrinking the stored transcript instead of only compressing what gets sent each turn.

Trade-off is explicit: upstream optimizes for contract correctness + native OpenClaw compaction semantics; this PR optimizes for tool-heavy agents with huge transcripts where LLM `/compact` is slow/expensive and the original no-op was blocking auto-compaction entirely.

Latest `pr-prep` (`6c43e8f`, merge with `main`) keeps the durable Headroom compaction path + the P1 fixes from earlier review rounds. Happy to discuss whether the project wants both modes (delegate vs own) as a config flag if that's preferable long-term.

GenorTG:
Follow-up: added **`persistentCompaction`** config so both designs coexist without forking behavior.

```json
"headroom": {
  "config": {
    "persistentCompaction": "headroom"   // default — durable /compact via Headroom /v1/compress (zero LLM)
    // "persistentCompaction": "openclaw"  // upstream #2304 behavior — delegate to OpenClaw native compaction
  }
}
```

- **`"headroom"`** (default): `ownsCompaction: true`, durable SQLite rewrite + turn-advancement contract (this PR's path)
- **`"openclaw"`**: `ownsCompaction: false`, `compact()` → `delegateCompactionToRuntime()` (matches `main` after eb4da64)
- **`assemble()`** per-turn compression: Headroom in **both** modes

Pushed in latest `pr-prep` commit; 128 plugin tests pass. Deprecated boolean alias `durableCompaction` also accepted for migration.

GenorTG:
## Update — hybrid compaction + upstream-safe defaults (live on my OpenClaw now)

I pushed `2e33da4` on `GenorTG:pr-prep` to consolidate the compaction work and avoid the confusing split between the earlier `persistentCompaction` commit (which defaulted to `"headroom"`) and what we actually want in production.

### Why I changed this

Running Headroom as the sole durable compactor (`ownsCompaction: true`) caused real failures in production — mid-turn `Pending input ownership ended` errors and weak savings vs OpenClaw's safeguard path on large sessions. I also hit OpenClaw degrading Headroom → legacy every turn when `transcriptSemantics` was missing in delegated modes.

The fix is **hybrid** for my deployment: Headroom does replace-only pre-passes (tool blob shrink) and OpenClaw still owns the LLM summarization / history prune. For the **plugin default**, I switched to **`"openclaw"`** so stock installs behave like upstream (`ownsCompaction: false`) unless an operator explicitly opts into `"hybrid"` or `"headroom"`.

### What shipped in `2e33da4`

- `persistentCompaction: "hybrid" | "headroom" | "openclaw"` (default **`openclaw`**)
- `transcriptHygiene` — turn-end replace-only maintenance (enabled by default only in hybrid mode)
- `transcriptSemantics` always declared (2026.9.x turn contract fix)
- Hygiene skips proxy `/v1/compress` when rough token estimate is below soft threshold
- README + `PR_BODY.md` updated with merge-safety / defaults section

### Tests

- `npm test` in `plugins/openclaw`: **136/136 passing**
- `npm run typecheck` + `npm run build` clean

### My live config (operator opt-in, not plugin defaults)

I'm running this on my gateway now with explicit hybrid tuning:

```json
"persistentCompaction": "hybrid",
"transcriptHygiene": { "enabled": true, "softThresholdTokens": 400000 }
```

Plus `HEADROOM_MODE=token` on the proxy and `plugins.slots.contextEngine: "headroom"`. Context size drops and compaction behavior match what I wanted — VNC session compact succeeded (~232k → ~24k tokens), and the main agent DB is back to a sane size after archive/hygiene cleanup.

Happy to adjust defaults or docs if maintainers prefer hybrid as the documented example while keeping `openclaw` as the code default.

JerrettDavis (changes requested):
The failed-write/stale-instance persistence fixes and resetLeaf correction address the earlier concrete examples. Two boundaries still need work before the durability claim is reliable:

1. [P1] plugins/openclaw/src/turn-advancement-store.ts:77-81,99-113 treats a newly created empty lock as stale. openSync(wx) creates the file before its PID is written; a second writer can unlink that live lock and commit concurrently. Holding the first lock descriptor open before the PID write reproduced a successful second commit and deletion of the first owner’s lock against the compiled store. Please publish ownership atomically or use a locking protocol with safe stale recovery; an empty/new lock and EPERM from process probing are not proof that its owner is dead. Add a deterministic two-process regression at that boundary.

2. [P1] plugins/openclaw/src/compaction.ts:328-352,375-382 still selects a raw suffix before resetLeaf. A 41-message fixture beginning with assistant toolCall cut-call and its toolResult, followed by 39 user messages, retains 40 messages starting at the orphan toolResult when forced trimming runs. Please choose complete turn/tool groups and preserve required instructions, then verify the reloaded native transcript. The new all-user fixtures do not exercise the boundary requested in the previous review.

Typecheck/build pass. The plugin suite had 133 passes and three Windows portability/setup failures; the two additional reproductions above independently demonstrate the correctness gaps. The PR body should also be refreshed to describe the current OpenClaw default and actual compaction behavior.

GenorTG:
Hi @JerrettDavis — thanks for the two reproductions; both were real. Addressed in **`8998cea`** (code) and **`5b740af`** (PR body), pushed to `pr-prep`.

**1. [P1] Lock: empty new lock treated as stale**

You're right that `open(wx)` → `write(pid)` left a window where the file exists but has no owner, and that "unparseable ⇒ stale" plus "EPERM ⇒ dead" turned that window into a steal. The lock is now its own module, `src/store-lock.ts`:

- **Ownership is published atomically.** The owner record (`{pid, token, createdAtMs}`) is written to a private temp file and `link(2)`-ed to the lock path; `EEXIST` means held. The lock is never observable empty. (`O_EXCL` fallback for filesystems without hard links — safe because of the next rule.)
- **Staleness needs age *and* a provably dead owner.** A lock younger than `staleAfterMs` (30 s) is never stale, regardless of content. Past that, only `ESRCH` or an unreadable record qualifies; **EPERM counts as alive**. A hard ceiling (10 min) covers PID reuse.
- **Recovery is claimed with an atomic `rename`** to a unique path before unlinking, so two waiters can't both "recover" and one delete a lock a third writer just took. Release only unlinks if the token on disk is ours.
- Waiting uses `Atomics.wait` instead of a spin loop; timeout 5 s → throws (never steals).

Deterministic two-process regression at exactly your boundary (`test/store-lock.test.ts`): a child does `openSync(lock, "wx")`, signals, **holds the fd without writing a PID** for 400 ms, then writes/releases. The parent's `commit()` blocks ≥ 400 ms, the child observes its lock `STILL_HELD`, both finish, store is intact. Plus unit cases for: old+alive not stolen, old+dead recovered, old+empty recovered, hard ceiling, foreign-token release.

**2. [P1] Truncate selected a raw suffix**

New `src/truncate-boundary.ts`, used by both the forced path and the compress-result path:

- The cut is moved to the nearest **turn start** using the same rule as the runtime's `isTurnStartMessage` (`user`, non-runtime-carrier `custom`, `bashExecution`, `branchSummary`, `compactionSummary`) — forward first (keep whole turns, slightly fewer messages), backward if nothing follows, and **no truncation at all** if the branch has no turn boundary.
- `dropOrphanToolResults` as a defensive post-pass; whole tool groups stay together.
- Your 41-message fixture is locked as a test: forced trimming now yields **39** messages starting at `user-0`, no `toolResult`.

Verified against the **real** SQLite-backed `SessionManager` (not the test double): append the fixture → plan → `applyCompactionPlan` → reopen: persisted branch 41 → 39, `bytesFreed: 507`, first `user-0`, last `user-38`, no orphan `toolResult`/assistant.

**Windows**

I don't have a Windows host, but the three failures looked like: hardcoded `/tmp` in path expectations/`cwd`, a bare filesystem path in the child's ESM `import` (needs `file://`), and `spawnSync("npm")` without a shell. All three are fixed (`tmpdir()`/`join()`, `pathToFileURL`, `shell: process.platform === "win32"`, `homedir()` fallback for the state dir). Would appreciate a re-run on your side.

**PR body**

Rewritten from scratch: describes the current default (`persistentCompaction: "openclaw"` = your #2304 delegation, `ownsCompaction: false`), a side-by-side table of what differs from `main` and why, what each review round changed, and updated Real Behavior Proof. The compaction modes are isolated so they can be dropped if the project prefers not to carry them.

`npm test`: 292 passing across 21 files; typecheck + build clean.

JerrettDavis (changes requested):
The new lock protocol closes the empty-lock race, and turn-boundary selection fixes the orphan-result case. One accounting error remains in the new backward-alignment path. In uildCompactionPlanFromCompressResult, when no safe boundary exists at/after the proxy cut, startIndex < dropped prepends originals.slice(startIndex, dropped) to the compressed tail, but the returned plan still reports the proxy's original 	okensAfter. Those restored verbatim messages are not included in that value, so compact()/hygiene reports and downstream stimatedTokens understate the transcript after the fix. Please either choose only a forward boundary or recompute a conservative tokensAfter that includes the restored prefix, and add a regression where selection falls back backward from the proxy cut. The lock ownership, bounded store, protected tool payload restoration, and requested forced-truncation boundary now look sound.

JerrettDavis (changes requested):
The latest head still does not address the accounting blocker from ae63fd02. In buildCompactionPlanFromCompressResult, the backward-alignment branch still prepends originals.slice(startIndex, dropped) to the proxy result but returns result.tokensAfter unchanged. Those restored verbatim messages were never counted by the proxy's tokens_after, so the persisted transcript and compaction/hygiene accounting can still diverge and downstream estimatedTokens can be understated. The new deferred-tool/protectToolResults work is orthogonal to that path. Please recompute a conservative tokensAfter that includes the restored prefix (or avoid backward alignment), and add the requested regression where the safe boundary is before the proxy cut and asserts accounting for the actual retained transcript.

GenorTG:
Hi @JerrettDavis — the accounting blocker from `ae63fd02` / your Sep 12 review is addressed in **`ace9eed1`** (also merged current `main` into `pr-prep`).

**What was wrong**
In `buildCompactionPlanFromCompressResult`, backward turn-boundary alignment prepended `originals.slice(startIndex, dropped)` onto the proxy tail but still reported the proxy's `tokens_after`. Those restored messages were never in the proxy count, so compact/hygiene and downstream `estimatedTokens` could understate the retained transcript.

**Fix**
- New `adjustTokensAfterForBoundaryAlignment`: add a rough token estimate for the restored prefix; subtract an estimate when forward alignment drops extra compressed messages.
- Clamp to `[1, tokensBefore]`.
- Backward alignment kept on purpose — without it, a mid-turn proxy cut with no forward turn-start would refuse to truncate even when a safe earlier boundary exists.

**Regression**
`accounts for messages restored when the safe boundary is before the proxy cut` — proxy window is only the trailing assistant+toolResult; selection walks back to the preceding user; asserts `tokensAfter === proxyTokensAfter + estimateRoughTokens([restoredUser])`.

**Validation**
- `plugins/openclaw`: **321** vitest tests passing
- typecheck + build clean
- `pr-prep` is up to date with `origin/main` (merge `3a06c0df`)

JerrettDavis (approved):
Re-reviewed the new commits against the prior compaction-accounting finding. Boundary realignment now adds restored pre-cut messages, subtracts forward-dropped compressed messages, clamps the estimate conservatively, and has a regression covering the backward-alignment case. The default persistent-compaction documentation now matches code/schema. Current main is included, the branch is mergeable, and checks are green.

GenorTG:
Hi @JerrettDavis — thanks again for the Sep 14 approval and for sticking through the durability / Gemini / lock / truncate / tokensAfter rounds.

**This update (post-approval, production follow-up):**

1. **Merged current `main` into `pr-prep`** so the branch is current with upstream (deps, proxy, beacon, etc.).
2. **Freeze-safe gateway routing** — OpenClaw 2026.9.x hands plugins a frozen/sealed `api.config`. `applyGatewayProviderBaseUrlsInPlace` was throwing `TypeError: Cannot assign to read only property 'opencode-go'` when refreshing session headers, even when disk config already had the correct proxy `baseUrl` + `x-headroom-base-url`.
   - If providers are frozen **and already routed** → no-op (safe; proxy path still works).
   - If frozen **and not routed** → clear operator error asking to bake routing into `openclaw.json`.
   - Mutable configs still get session-header injection as before.
3. **Tests / validation** — `plugins/openclaw`: **324** vitest cases passing (22 files), typecheck + build clean. New regressions in `test/gateway-config.test.ts` for frozen already-routed / frozen-not-routed / mutable session-header paths.

All prior review blockers remain addressed; defaults still match upstream (`persistentCompaction: "openclaw"`). Happy to tweak anything else before merge.

JerrettDavis (changes requested):
[P1] Frozen-config fallback silently removes the configured session contract — `plugins/openclaw/src/gateway-config.ts:205-208` treats matching proxy `baseUrl`/upstream as sufficient and skips the write even when `providerSessionHeaders[providerId]` caused `mutated`. The new regression test makes this concrete: `x-opencode-session` is configured but the resulting provider headers intentionally leave it `undefined`. `session-headers.ts` documents that this option exists for upstreams which return `MissingSessionID` when the header is absent, so routing to Headroom does not make omission safe; requests still reach that upstream without its required session identifier. Please include the requested session header/value in the already-routed predicate (or fail clearly when a frozen config cannot satisfy it) and make the frozen test assert that configured provider requirements are never silently dropped. Re-reviewed exact head `33984079`.

GenorTG:
Hi @JerrettDavis — addressed the Sep 22 [P1] frozen-config session-header drop in the latest `pr-prep` push.

**Fix:** `providerAlreadyRouted` now requires any configured `providerSessionHeaders[providerId]` to already be present (non-empty) on the frozen provider. Matching proxy `baseUrl` + `x-headroom-base-url` alone is no longer enough. If the session header is missing on a frozen config, we throw the clear bake-into-`openclaw.json` error (mentions `providerSessionHeaders`) instead of silently continuing.

**Tests:** frozen already-routed fixture now includes `x-opencode-session`; new case asserts throw when routed but session header absent. `gateway-config` suite: 32 passed.

Also tightened `store-lock` create modes (`0o600`) and replaced the readonly-mutation regex with substring checks (CodeQL notes on that file / ReDoS).

JerrettDavis (changes requested):
The exact-head code change addresses the frozen-session-header blocker, and local plugin validation is otherwise clean (324/325 passed in the full run; the one Windows two-process timeout passed immediately in isolation; typecheck and build passed). The newly authorized hosted CI exposed a deterministic repository gate, though: commitlint rejects several PR commits, including Add providerUpstreams config for per-provider upstream overrides and Normalize rewritten proxy pathname to /v1, because they have no conventional type or subject. Please reword or squash the non-conventional commits so the required commitlint job passes. No CI rerun is needed until the commit history changes.

GenorTG:
Hi @JerrettDavis — the commit history is now conventional. Reworded the 7 flagged subjects; the rewrite is **message-only**:

- `git diff 58a254f0..6f18c0a1` is empty; both heads share tree hash `b618e27d7a367506fc5bd1c69bbd832ab743e91b` — zero file changes, exactly the code you validated at `58a254f0`.
- New head: `6f18c0a1` (force-pushed to `GenorTG:pr-prep`).

| was | now |
|-----|-----|
| `Declare transcriptSemantics and implement commitTurn` | `feat(openclaw): declare transcriptSemantics and implement commitTurn` |
| `Add providerUpstreams config for per-provider upstream overrides` | `feat(openclaw): add providerUpstreams config for per-provider upstream overrides` |
| `Normalize rewritten proxy pathname to /v1` | `fix(openclaw): normalize rewritten proxy pathname to /v1` |
| `Add providerSessionHeaders config for per-provider session injection` | `feat(openclaw): add providerSessionHeaders config for per-provider session injection` |
| `Add tests for new behavior + fix no-upstream proxy pathname` | `fix(openclaw): add tests for new behavior and fix no-upstream proxy pathname` |
| `openclaw plugin: bound the turn-advancement store` | `fix(openclaw): bound the turn-advancement store` |
| `openclaw plugin: make the assemble() budget skip aware of OpenClaw's reserve` | `perf(openclaw): make the assemble() budget skip aware of OpenClaw's reserve` |

The `merge:` commits were left untouched on the assumption commitlint ignores merge commits by default — if the action still counts them, say the word and I will give them conventional subjects too.

The push currently shows all workflow runs as `action_required` — could you approve the hosted CI run so `commitlint` (and the rest) executes? Expected all-green since nothing else changed. Also: the three remaining unresolved review threads (`turn-advancement-store.ts` x2, `compaction.ts` x1) are round-2 findings fixed and validated in the Sep-14 approval — happy to leave them for you to resolve.

GenorTG:
Follow-up on the history cleanup: I also normalized the four `merge:` subjects to `chore(merge): …` — belt-and-braces in case the linter ever counts merge commits. Final head: `1476be87`, and the tree is still exactly the code you validated — tree hash `b618e27d7a367506fc5bd1c69bbd832ab743e91b`, `git diff 58a254f0..1476be87` empty (zero file changes; the only edits are commit messages).

Pre-checked deterministically here: `@commitlint/cli@19` run locally against **all 35 PR commits** with this repo's `.commitlintrc.json` — 0 failures.

Heads-up: this was another history-rewrite push, so the fork-PR run-approval gate will show for the new head again — could you approve the workflow run when convenient? Once it executes, `commitlint` and the rest should pass unchanged.

PR #3490, opened 2026-09-08
fix(cli): strip unmarked headroom_memory TOML table before injecting

bernardopg:
## Description

`_inject_memory_mcp_config` (used by `headroom wrap codex`) only guards
against a duplicate `[mcp_servers.headroom_memory]` table via the
`_MEMORY_MCP_MARKER`/`_MEMORY_MCP_END` comment pair it writes. If that
table already exists in a user's `~/.codex/config.toml` **without**
those markers — e.g. hand-copied from the usage example in
`headroom/memory/mcp_server.py`, or left over from before this table
was marker-guarded — the injector appends a second
`[mcp_servers.headroom_memory]` table instead of replacing the
existing one. Two tables sharing the same key is invalid TOML, and
silently breaks Codex's MCP config parsing on startup.

This PR adds a helper that strips any unmarked
`[mcp_servers.headroom_memory]` table before the marker-wrapped
section is appended, mirroring the pattern already used for
`[model_providers.headroom]`.

No issue number to close (found while reviewing the Codex wrap flow).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `_strip_existing_codex_memory_mcp_table()` in `headroom/cli/wrap.py`, mirroring the existing `_strip_existing_codex_headroom_provider_table()` helper: it regex-matches an unmarked `[mcp_servers.headroom_memory]` table and removes it.
- Called this new helper from `_inject_memory_mcp_config()` right before appending the marker-wrapped `headroom_memory` section, in the branch that runs when no `_MEMORY_MCP_MARKER` is found in the existing config.
- Added a regression test, `test_inject_replaces_unmarked_memory_block`, in `tests/test_cli/test_wrap_codex.py`.

## Testing

### Test Output

```text
$ .venv/bin/python -m pytest tests/test_cli/test_wrap_codex.py -q
platform linux -- Python 3.13.15, pytest-9.0.3, pluggy-1.6.0
rootdir: /home/bitter/development/oss/headroom
configfile: pyproject.toml
plugins: anyio-4.12.1, cov-7.0.0, langsmith-0.9.3, respx-0.23.1, asyncio-1.3.0
collected 93 items

tests/test_cli/test_wrap_codex.py ...................................... [ 40%]
.......................................................                  [100%]

======================== 93 passed in 66.82s (0:01:06) =========================

$ .venv/bin/python -m pytest tests/test_cli/ -q
[...]
734 passed, 4 skipped, 1 failed in 92.84s (0:01:32)

FAILED tests/test_cli/test_unwrap_claude.py::test_unwrap_claude_stops_claude_owned_persistent_deployment
```
This one failure is pre-existing on a clean `main` checkout (verified
via `git stash` + rerun before making any change here) and is
unrelated to this PR — it's an assertion on unwrap-claude output
messaging, not touched by this diff.

```text
$ .venv/bin/python -m ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_codex.py
All checks passed!

$ .venv/bin/python -m mypy headroom/cli/wrap.py
```

## Real Behavior Proof

- Environment: local dev checkout, Python 3.13.15, `.venv` from repo (`pytest 9.0.3`, `ruff`, `mypy`).
- Exact command / steps: Ran `.venv/bin/python -m pytest tests/test_cli/test_wrap_codex.py::TestCodexMemoryMcpConfig::test_inject_replaces_unmarked_memory_block -q`, which (1) seeds `~/.codex/config.toml` with an unmarked `[mcp_servers.headroom_memory]` table plus `[profiles.default]` and an unrelated `[mcp_servers.other]` table, (2) calls `_inject_memory_mcp_config("codex-user")` — the same function `headroom wrap codex` invokes — and (3) reads the resulting `config.toml` back and parses it with `tomllib.loads`.
- Observed result: With the fix applied, the test passes — exactly one `[mcp_servers.headroom_memory]` table remains in the file, wrapped in the `_MEMORY_MCP_MARKER`/`_MEMORY_MCP_END` comments, with the updated `--user codex-user` value; `[profiles.default]` and `[mcp_servers.other]` are byte-for-byte untouched; `tomllib.loads(content)` succeeds and `parsed["mcp_servers"]["headroom_memory"]["args"][-1] == "codex-user"`. I additionally reverted just `headroom/cli/wrap.py` to its pre-fix state (`git checkout e67b3c8a -- headroom/cli/wrap.py`, keeping the new test) and reran the same test to confirm it fails without the fix: `assert content.count("[mcp_servers.headroom_memory]") == 1` fails with `assert 2 == 1`, because the injector left the old unmarked table in place and appended a second, marker-wrapped `[mcp_servers.headroom_memory]` table below it. Separately, I fed that exact two-table content through `tomllib.loads()` directly and confirmed it raises `tomllib.TOMLDecodeError: Cannot declare ('mcp_servers', 'headroom_memory') twice (at line 14, column 29)` — reproducing the exact failure mode Codex's TOML parser hits on startup. I then restored the fixed `wrap.py` and reran the test to confirm it passes again.
- Not tested: real `codex` binary startup against the patched config (no local Codex CLI install in this environment) — covered instead by the `tomllib.loads` validity assertion, which is what actually breaks Codex's parser today.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — this only affects the `headroom wrap codex` CLI subcommand's local config-file writer, not any server-side or rollout-gated runtime behavior.
- Minimum rollout channel: N/A (CLI-only, no feature flag).
- Stable/default behavior changed: no. Behavior only changes for the edge case where an unmarked `[mcp_servers.headroom_memory]` table already exists (previously produced broken/invalid TOML); the normal marker-found and no-table-at-all paths are unchanged.
- Kill switch / disable path: N/A — no runtime toggle; users can simply not run `headroom wrap codex` or manually edit `~/.codex/config.toml`.
- Qualification impact: none.
- Rollback path: revert this commit; the prior (buggy) behavior only manifested for a narrow, previously-broken edge case, so reverting is safe.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A — internal helper, no user-facing docs describe this edge case)

N/A — CLI/config-file fix, no UI.

## Additional Notes

- "Manual testing performed" is unchecked because I did not run the real `codex` binary against the patched config in this environment (no Codex CLI installed here); the regression test's `tomllib.loads` round-trip is the direct proof that the generated config is valid TOML, which is the actual failure mode this PR fixes.
- The pre-existing `test_unwrap_claude_stops_claude_owned_persistent_deployment` failure on `tests/test_cli/` is unrelated to this change and reproduces identically on a clean `main` checkout.


headroom/cli/wrap.py
```diff
@@ -2794,6 +2794,26 @@ def _strip_existing_codex_headroom_provider_table(content: str) -> str:
...
+    comments that make the marker-based idempotency check work (e.g. hand-
+    added from the usage example in ``headroom/memory/mcp_server.py``).
+    """
+    if "[mcp_servers.headroom_memory]" not in content:
+        return content
+
+    import re  # local import to match surrounding helper convention
+
+    memory_table = re.compile(
+        r"(?ms)^[ \t]*\[mcp_servers\.headroom_memory\][^\n]*\n.*?(?=^[ \t]*\[|\Z)"
```
JerrettDavis: [P1] Avoid treating table-looking text inside multiline strings as a real table. With valid input `instructions = """\n[mcp_servers.headroom_memory]\nkeep this instruction\n"""\nmodel="gpt-4o"\n`, the actual injector deletes the string contents, closing delimiter, and model setting, then writes an unterminated string to config.toml. This path previously appended a valid table without changing the string. Please use TOML-aware table boundaries and validate the candidate document before replacing the file; preserve unrelated strings/settings. Add a regression through `_inject_memory_mcp_config` for this input.

headroom/cli/wrap.py
```diff
@@ -2794,6 +2794,26 @@ def _strip_existing_codex_headroom_provider_table(content: str) -> str:
...
 
+def _strip_existing_codex_memory_mcp_table(content: str) -> str:
+    """Remove a pre-existing, unmarked ``[mcp_servers.headroom_memory]`` table.
+
+    Guards ``_inject_memory_mcp_config`` against duplicate TOML tables when
+    the section exists without the ``_MEMORY_MCP_MARKER``/``_MEMORY_MCP_END``
+    comments that make the marker-based idempotency check work (e.g. hand-
+    added from the usage example in ``headroom/memory/mcp_server.py``).
+    """
+    if "[mcp_servers.headroom_memory]" not in content:
```
JerrettDavis: [P2] Handle equivalent TOML key spellings. A valid existing `[mcp_servers."headroom_memory"]` table bypasses this exact-substring guard, so injection still writes a second semantic table and `tomllib.loads` raises `Cannot declare ('mcp_servers', 'headroom_memory') twice`. Quoted keys and whitespace around the dotted key should resolve to the same target table. Please cover these spellings in the replacement logic and regression tests.

JerrettDavis (changes requested):
Thanks for adding the parseable-TOML regression. All 93 focused tests pass. The basic unmarked-table case works, but the new text-level removal can corrupt valid input before writing it back. Please address the two cases below with end-to-end injection tests that parse the result and verify unrelated settings are preserved. No CI action is needed from you.

headroom/cli/wrap.py
bernardopg: Fixed in d565b5a7: `_strip_existing_codex_memory_mcp_table` no longer uses a DOTALL `.*?` to find the table body. It now scans line-by-line while tracking whether each line starts inside an open triple-quoted string (`_multiline_string_start_flags`), so a `[mcp_servers.headroom_memory]`-looking line inside an `instructions = """..."""` value is never mistaken for a real header.

Added `test_inject_preserves_table_looking_text_in_multiline_string`, an end-to-end regression through `_inject_memory_mcp_config` using exactly the input you described, asserting via `tomllib.loads` that the string and the `model` setting after it survive untouched and the real memory MCP block is still injected.

headroom/cli/wrap.py
bernardopg: Fixed in d565b5a7: the header matcher now accepts quoted-key and whitespace variants (`[mcp_servers."headroom_memory"]`, `[ mcp_servers . headroom_memory ]`, etc.) instead of an exact-substring guard, via a bare-or-quoted key pattern applied to both segments of the dotted key.

Added `test_inject_replaces_quoted_key_memory_table`, an end-to-end regression through `_inject_memory_mcp_config` covering the `[mcp_servers."headroom_memory"]` spelling, asserting the result round-trips through `tomllib.loads` with exactly one `mcp_servers.headroom_memory` table and unrelated settings preserved.

JerrettDavis (changes requested):
Thanks for addressing the original multiline-string and quoted-key examples. All 96 focused tests pass, but two valid TOML inputs still fail through _inject_memory_mcp_config on the latest head:

1. [P1] _TOML_GENERIC_HEADER_RE explicitly excludes array-of-table headers. Given `[mcp_servers.headroom_memory]` followed by `command="old"`, then `[[profiles.custom.entries]]` and `name="preserve"`, the cleanup deletes the entire unrelated profiles table. Please recognize arrays of tables as boundaries and add a parse-and-preserve injection regression.

2. [P2] _multiline_string_start_flags counts triple quotes inside comments as string delimiters. A first line `# Example triple quote: """` followed by the existing memory table makes the cleanup miss the table; injection then writes duplicate tables and tomllib raises. Please tokenize comment/string boundaries correctly, and validate the complete candidate TOML before replacing the config so malformed output cannot overwrite a working configuration.

These reproductions use temporary config directories and the real injector. The same validation should cover the expanded provider-table cleanup, which still uses text-level boundaries.

bernardopg:
Thanks — addressed the follow-up review in 16314146.

### Structural TOML cleanup
Replaced the text-level table-boundary logic with `tomlkit` structural edits for unmarked `mcp_servers.headroom_memory` tables, existing `model_providers.headroom` tables, and orphaned local Headroom provider tables during managed-block cleanup. This preserves multiline strings, comments (including `"""` inside a comment), quoted keys, nested tables, and `[[array-of-tables]]` boundaries without inferring TOML structure from lines or regexes.

### Safe writes
Both `_inject_memory_mcp_config()` and `_inject_codex_provider_config()` now parse the complete generated candidate with `tomllib.loads()` immediately before writing. An invalid candidate follows the existing warning path and cannot overwrite the current configuration.

### Regression coverage
Added end-to-end temporary-Codex-config tests for:
1. an unmarked memory table followed by `[[profiles.custom.entries]]`;
2. a preceding comment containing `"""`, followed by the old memory table;
3. provider cleanup followed by an unrelated array table; and
4. invalid candidates for both injection paths, verifying the original config remains byte-for-byte unchanged.

Validation:
- `pytest tests/test_cli/test_wrap_codex.py -q` — 101 passed
- `ruff check headroom/cli/wrap.py tests/test_cli/test_wrap_codex.py` — passed
- `mypy headroom/cli/wrap.py` — passed
- `git diff --check` — clean

JerrettDavis (approved):
Re-reviewed after the structural TOML fix and main refresh. TOMLKit now removes only parsed provider/MCP tables, preserving multiline strings, comments, quoted keys, and unrelated array tables; malformed source/candidate configs are never written. All 101 Codex wrap tests pass locally on the updated head. The branch is current, mergeable, and attached checks pass. The prior parsing blocker is resolved.

bernardopg:
Hi @JerrettDavis @chopratejas @DevanshiVyas 👋

Quick status update on this PR:

### CI failures on the previous head (`2129a111`)
The failing `test (2)`, `test (3)` and `test (4)` jobs were unrelated to this change. All failures came from pricing/budget/savings tests that exist on `main` (`test_savings_provider_matrix.py`, `test_pricing_from_litellm.py`, `test_models.py::test_claude_info`, and budget tests raising `KeyError: 'input_cost_per_token'`). This PR only touches `headroom/cli/wrap.py` and `tests/test_cli/test_wrap_codex.py`. Those tests have since been fixed upstream (#3699, #3722, #3575).

### What I did
Merged the latest `main` into the branch (`9301feab`, no conflicts). Local validation on the new head:
- Codex wrap tests plus the previously failing pricing/savings/budget/cost test files: 307 passed, 2 skipped
- `ruff check` and `mypy headroom/cli/wrap.py`: clean

### Action needed
Since this PR comes from a fork, the workflows for `9301feab` are waiting in `action_required`. Could one of you approve the workflow runs when you have a moment? After CI is green, the PR only needs an approving review. The earlier parsing blocker was already confirmed resolved in the last review.

Thanks a lot for the thorough reviews on this one!

PR #3506, opened 2026-09-09
fix: preserve Pi evidence and native images during compression

willgao-dev:
Pi can lose evidence when compression removes JSON rows without a recovery reference or replaces a recovered original with cached text. Cached mixed content can also lose native images.

This change preserves recovery results across the Pi MCP wrapper, WebSocket deltas, the content router, and the session cache. Mixed compression stores the complete original and includes its reference. Storage failure preserves the original text. Cached references require available originals.

`HEADROOM_IMAGE_OPTIMIZE=0` disables image optimization in the CLI and content router. Mixed native content bypasses text-only cache replacement.
`HEADROOM_COMPRESS_USER_MESSAGES=0` also overrides the aggressive profile default.

Validation:

- 366 regression tests pass. One optional Anthropic API test skips because `ANTHROPIC_API_KEY` is absent.
- Ruff and `git diff --check` pass.
- An independent review found the cache replacement and image deletion defects. New session preparation tests reproduce and fix both.
- The synthetic Pi test compresses evidence, retrieves its original once, and returns the same required values as the uncompressed baseline.
- All 32 concurrent synthetic compression requests recover exactly. Token savings range from 73.6% to 73.8%.
- Native image bytes remain unchanged in OpenAI and Anthropic payloads.
- The separate Pi MCP suite passes 22 tests, including MP4, MOV, and WebM frame conversion.

The local test command was:

```bash
uv run --no-project --python 3.13 --with pytest --with pytest-asyncio python /tmp/run-headroom-final-tests.py
```

The runner imports this checkout and the installed Headroom 0.37 native wheel. It calls `pytest.main` with the listed files and `-q -ra --import-mode=importlib`.

```text
tests/test_agent_savings.py
tests/test_retrieval_delta_fidelity.py
tests/test_mixed_content_recovery.py
tests/test_mixed_content_sections.py
tests/test_tag_protection_integration.py
tests/test_transforms_content_router.py
tests/test_session_cache_fidelity.py
tests/test_proxy_image_preservation.py
tests/test_ccr_cache_lifetime.py
tests/test_hermes_tool_call_unwrap.py
tests/test_compression_cache.py
tests/test_compression_cache_registry.py
tests/test_cross_turn_cache_safety.py
tests/test_ccr_sqlite_backend.py
tests/test_ccr_row_drop_store_bridge.py
tests/test_ccr_rust_marker_hash_bridge.py
tests/test_ccr_retrieve_history_repair.py
tests/test_ccr_response_handler_openai_responses.py
tests/test_ccr_tool_injection.py
tests/test_ccr_mcp_http.py
tests/test_openai_responses_compression_units.py
tests/test_openai_responses_read_protection.py
tests/test_provider_codex_images.py
tests/test_provider_openai_images.py
tests/test_image_compression_isolation.py
```

The exact-recovery fixture uses 14% more uncached input tokens than bypass because it needs another model call. Payload savings are not total-task savings. These tests do not establish unchanged quality for every lossy result.

The local runtime pins the tested fork revision. Local rollout uses a separate restart command that preserves launchd registration and verifies a new listener. Upstream review and merge remain pending.

### DIFF
--- a/headroom/agent_savings.py
+++ b/headroom/agent_savings.py
@@ -338,6 +338,8 @@ def proxy_pipeline_kwargs(config: object) -> dict[str, object]:
 
     if getattr(config, "compress_user_messages", False):
         kwargs["compress_user_messages"] = True
+    if os.environ.get("HEADROOM_COMPRESS_USER_MESSAGES") == "0":
+        kwargs["compress_user_messages"] = False
 
     compress_system_messages = getattr(config, "compress_system_messages", None)
     if compress_system_messages is not None:
--- a/headroom/cache/compression_cache.py
+++ b/headroom/cache/compression_cache.py
@@ -15,6 +15,8 @@ import time
 from collections import OrderedDict
 from dataclasses import dataclass
 
+from .compression_store import cached_references_available
+
 logger = logging.getLogger(__name__)
 
 
@@ -153,6 +155,10 @@ class CompressionCache:
         """Retrieve compressed content by hash, refreshing LRU position on hit."""
         with self._lock:
             entry = self._cache.get(hash)
+            if entry is not None and not cached_references_available(entry.compressed):
+                self._total_tokens_saved -= entry.tokens_saved
+                del self._cache[hash]
+                entry = None
             if entry is None:
                 self._misses += 1
                 return None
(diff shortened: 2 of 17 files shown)


JerrettDavis (approved):
Reviewed the evidence-preservation paths, including Pi wrapper unwrapping, recovered-original guards, mixed-content recovery storage, expired/evicted-reference invalidation in both caches, and native image preservation. Ran all 25 listed local test files against this checkout and the existing native extension: 365 passed, 2 skipped. Ruff 0.16.3 check and format check pass on all 17 changed Python files. The failure-path tests preserve original evidence when recovery storage is unavailable, and the session preparation tests preserve protected recovery results and image blocks. The description correctly distinguishes payload reduction from total-task savings. No blocking source issue found; full remote CI has not run.

PR #3513, opened 2026-09-09
docs(security): refresh security policy and add security.txt

Spoint:
Rolling supported-versions policy (repo is 0.37, doc said 0.27), GitHub Private Vulnerability Reporting, 90-day coordinated disclosure + SLA table, Safe Harbor section, broadened scope (npm TS SDK, install runners), softened the absolute never-log-API-keys claim, and RFC 9116 .well-known/security.txt.


JerrettDavis (changes requested):
I refreshed this branch from main and retained the optional-dependency advisory section while resolving SECURITY.md. The rolling-version policy improves the stale release list, but two factual reporting details still need correction:

1. [P2] SECURITY.md promises a published PGP key/fingerprint at security.txt, while the new file contains only a commented-out Encryption placeholder. Please remove the unavailable encryption instruction until a real public key is published, or supply and link the actual key.

2. [P2] The revised credential guidance still says Headroom does not persist credentials / write them to disk. save_headroom_copilot_oauth_token explicitly writes an OAuth refresh credential to the Headroom auth file, and install/wrap paths can persist credential-bearing configuration. Please describe those storage paths and protections accurately rather than retaining an absolute guarantee. Likewise, qualify the passthrough-by-default statement against the proxy’s actual default compression behavior.

The proposed disclosure targets and Safe Harbor commitments also need the final human maintainer review before adoption.

PR #3514, opened 2026-09-10
fix(tabular): keep each table's schema when a document holds several

no1fret-spec:
## The bug

A document holding more than one table collapses into a single table, and the
text between the tables is dropped.

`parse_markdown_table` keeps every line containing a `|` and discards the rest,
and `parse_csv` treats only the first row as headers. So the second and later
header rows are filed as **data**, and every row beneath them is relabelled
under the first table's columns.

A three-table brief compressed to this — note `Audience,Tone` and `Type,Format`
sitting in the roster as though they were people, under a schema claiming all
25 rows are `{Name, Role}`:

```
[25]{Name:string,Role:string}
Beth Prine,COO
...
Audience,Tone
Patients,"Warm, simple, reassuring"
...
Type,Format
Email,Subject + 3 short paragraphs + CTA
```

It round-trips through a CSV parser cleanly and every value survives, so
nothing downstream notices — but the schema is wrong, and a model reading it
back will state that Audience is a member of staff.

## Reproduction

```python
from headroom.transforms.tabular_ingest import TabularCompressor

t1 = "\n".join(f"| Person {i} | Role number {i} |" for i in range(30))
t2 = "\n".join(f"| Audience {i} | Tone descriptor {i} |" for i in range(30))
doc = ("## Team\n\n| Name | Role |\n| --- | --- |\n" + t1 +
       "\n\n## Voice\n\n| Audience | Tone |\n| --- | --- |\n" + t2)

r = TabularCompressor().compress(doc)
assert "Audience,Tone" not in r.compressed   # fails on main
assert "## Voice" in r.compressed            # fails on main
```

Both assertions fail before this change and pass after.

The loss only shows above the `min_savings_chars` threshold: at small sizes the
result is rejected and passed through, so the corruption appears exactly when
the compressor is doing its job.

## The change

Markdown is split into ordered table/text segments, and delimited text into
blank-line-separated blocks, so each table keeps its own header and renders as
its own CSV+schema block while the prose around it is carried through.

Anything ambiguous passes through verbatim rather than being guessed at,
following the #1652 invariant that a compressed table must never state facts
the original didn't:

- a pipe run that isn't a clean rectangle (a `|` inside a sentence)
- a header row with no rows under it
- a ragged block

CSV block splitting goes through the `csv` reader rather than splitting raw
lines, so a blank line inside a quoted field stays part of its cell.

Single-table documents with no surrounding prose keep the original
record-array path, so their output is byte-identical.

### Known limitation

CSV tables run together with **no** blank line between them are
indistinguishable from one table that happens to contain a row of column
names, so they stay merged. There's no signal to separate them; this is noted
in `parse_csv_tables`.

## Tests

`tests/test_transforms_tabular.py`: 39 → 59. New coverage for segment
splitting, quoted-field blank lines, header-only and ragged block rejection,
per-table schemas for both markdown and CSV, and a parametrized property test
asserting no cell or heading is ever dropped across table counts and sizes.

## Verified on

- macOS 15 (Darwin 25.6.0), Python 3.13.15, headroom-ai 0.37.0
- 59 passing in `test_transforms_tabular.py`; 134 across `test_compress_api`,
  `test_compression_decision`, `test_builtin_compressor_adapters`,
  `test_compression_batches`
- `ruff check`, `ruff format`, and `mypy --disallow-untyped-defs
  --warn-return-any` clean on the changed module
- Beyond unit tests: ran the patched module in the installed uv tool venv and
  confirmed the real `TabularCompressor` output through a fresh interpreter —
  two schema blocks, headings preserved, no header-row leak

## Not tested

- **The full suite.** The Rust `headroom._core` extension isn't built in my
  checkout, so I ran against a prebuilt `_core.abi3.so` with the checkout ahead
  of it on `PYTHONPATH`. In that hybrid setup `tests/test_compress_passthrough.py`
  fails 5 tests **on unmodified `main` as well** — baseline noise from my
  environment, not this change. Worth a maintainer confirming under real CI.
- The proxy/server path end-to-end; I exercised `TabularCompressor` directly.
- Multi-table **fixed-width** input, which `detect_content_type` classifies as
  `PLAIN_TEXT` and so never reaches this code. Left as a single table.
- Any platform other than macOS.

### DIFF
--- a/headroom/transforms/tabular_ingest.py
+++ b/headroom/transforms/tabular_ingest.py
@@ -7,6 +7,11 @@ array of records and routes it through the existing, battle-tested
 `SmartCrusher`, which already does lossless ``csv-schema`` compaction first and
 lossy row-drop with reversible ``<<ccr:HASH>>`` markers as a fallback.
 
+A document may hold several tables with prose between them. Each table
+carries its own header row, so they are parsed and rendered separately and
+the text around them is carried through — folding them together filed the
+later header rows as data and dropped the prose entirely.
+
 No new compression algorithm and no new CCR plumbing live here — only the
 text→records bridge.
 """
@@ -24,6 +29,9 @@ from .content_detector import ContentType, detect_content_type
 # Mirrors content_detector's separator-cell pattern (e.g. ``| --- | :--: |``).
 _MD_SEP_CELL = re.compile(r"^:?-{2,}:?$")
 
+# A parsed table: its header row, then its data rows.
+Table = tuple[list[str], list[list[str]]]
+
 
 # ─── Public dataclasses (mirror SearchCompressor / LogCompressor surface) ────
 
@@ -39,6 +47,17 @@ class TabularCompressorConfig:
     min_savings_chars: int = 1
 
 
+@dataclass(frozen=True)
+class Segment:
+    """One piece of a parsed document: a table, or the text around it.
+
+    Exactly one of ``table`` / ``text`` is set.
+    """
+
+    table: Table | None = None
+    text: str | None = None
+
+
 @dataclass
 class TabularCompressionResult:
     """Result of tabular-text compression."""
@@ -50,6 +69,8 @@ class TabularCompressionResult:
     rows: int
     columns: int
     strategy: str = "tabular"
+    # Number of distinct tables found. > 1 means each kept its own schema.
+    tables: int = 1
 
     @property
     def compression_ratio(self) -> float:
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
Keeping each table’s schema is a useful improvement, and all 59 existing focused tests pass. Two public-compressor reproductions still silently alter data:

1. [P1] headroom/transforms/tabular_ingest.py:295 unconditionally assigns generated prose keys. A 40-row Markdown table headed `_text1 | Value` followed by prose is replaced entirely by that prose under `_text1`; compress reports success and no table rows survive. Please allocate collision-free keys for both text and tables and assert that the rows survive through TabularCompressor.compress.

2. [P2] split_csv_blocks at lines 104-108 treats `,` (a valid two-empty-cell record) as a table separator. In a quoted Name/Role CSV, inserting that row between records 29 and 30 creates a second table with Person30/Role30 as its schema, relabelling all later rows. Please distinguish physical blank lines from all-empty CSV records, preserving the record or passing through ambiguity, and cover the public compression path.

These cases need correction to support the lossless/schema-preservation claim.

PR #3517, opened 2026-09-10
fix(learn): show both ends of CLI output on a parse failure

gglucass:
## Description

`_call_cli_llm` and `_call_claude_cli_streaming` attach a HEAD-only excerpt of
the CLI's output to their parse failures ("returned unparseable output. First
2000 chars:", "did not emit a final `result` event"). For those two failures
specifically, the head is the one part that is never diagnostic: a model's JSON
answer that does not parse looks like valid JSON for its first 2000 chars in
every case. What separates a payload truncated mid-value from one with prose
after the closing fence is the END of it, and that is exactly what gets cut.

This is not hypothetical. Downstream (the Headroom desktop app) these
RuntimeErrors are reported with the CLI's stderr attached, and a real report
arrived with 2000 characters of well-formed ```json ... and no way to tell which
of the two shapes it was, so the defect could not be acted on at all.

`_failure_detail` already makes this argument for the non-zero-exit path ("stdout
is tailed rather than headed because CLI backends emit the error last"). This
extends the same reasoning to the parse failures, which need both ends: the head
is where a wrapper (fence, preamble) shows up, the tail is where the break is.

## Type of Change

## Changes Made

- Added `_output_snippet(text, limit=_MAX_SNIPPET_LEN)`: returns short output
  unchanged, otherwise `head + "...[N chars omitted]..." + tail`, still bounded
  by the same `_MAX_SNIPPET_LEN` budget.
- Used it at the three sites that excerpt CLI OUTPUT for a parse failure: the
  non-streaming unparseable stdout, the missing final `result` event, and the
  unparseable `result` text.
- Reworded those messages from "First 2000 chars:" to "Head and tail of the
  output:". Deliberately no counts in the marker line: downstream groups these
  errors by that line, so a length in it would open a new group per failure.
- Left the stderr excerpts and `_failure_detail` alone.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_learn/ -q
260 passed, 4 skipped in 11.02s
```

## Real Behavior Proof

- Environment: macOS 15, Python 3.12, uv --frozen, headroom main @ 4e1f7769
- Exact command / steps: Added `test_claude_cli_unparseable_result_shows_both_ends`, which feeds `_call_cli_llm` a claude-cli `result` event carrying a ~7 KB fenced JSON payload followed by trailing prose (the shape that cannot be diagnosed today), then ran `uv run --frozen --extra dev pytest tests/test_learn/test_analyzer.py -q`.
- Observed result: 97 passed. The RuntimeError now contains both the opening ```json fence and the trailing "another shape" prose, plus a "chars omitted" marker, and stays shorter than the payload. Before the change the same assertion on the tail fails, since the message stopped at 2000 chars.
- Not tested: A live `claude -p` run against a real API failure; the CLI is mocked, as in the surrounding tests.

## Runtime Rollout Safety

- Stable/default behavior changed: No. Only the text of three error messages; short output is byte-identical to before.
- Kill switch / disable path: Not applicable; no flag, no new dependency, no I/O.
- Qualification impact: None. `headroom learn` success paths are untouched.

## Review Readiness

### DIFF
--- a/headroom/learn/analyzer.py
+++ b/headroom/learn/analyzer.py
@@ -602,6 +602,28 @@ def _strip_fenced_json(raw: str) -> dict:
     return result
 
 
+def _output_snippet(text: str, limit: int = _MAX_SNIPPET_LEN) -> str:
+    """Bounded excerpt of CLI output showing BOTH ends.
+
+    A head-only excerpt cannot diagnose the failure it is attached to. When a
+    model's JSON answer does not parse, the head is valid-looking JSON in every
+    case: what tells a payload truncated mid-value from one with trailing prose
+    after the closing fence is the END of it. Same reasoning as
+    :func:`_failure_detail` tailing stdout -- CLI backends put the interesting
+    part last -- except a parse failure needs the head too, because that is
+    where a wrapper (a fence, a preamble) shows up.
+
+    Returns *text* unchanged when it already fits, so short output -- the common
+    case -- reads exactly as before.
+    """
+    if len(text) <= limit:
+        return text
+    head = limit // 2
+    tail = limit - head
+    omitted = len(text) - limit
+    return f"{text[:head]}\n...[{omitted} chars omitted]...\n{text[-tail:]}"
+
+
 def _failure_detail(
     stderr: str | None, stdout: str | None, *, result_text: str | None = None
 ) -> str:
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Reviewed the bounded head/tail excerpts across all three parse-failure paths and their callers. The new regression and 95 other analyzer tests pass, with Ruff lint/format clean. One existing Windows path assertion in test_includes_project_info fails because Path renders backslashes; that code is unchanged by this PR. No blocking finding in the proposed change.

PR #3520, opened 2026-09-10
feat(savings): track lifetime output spend so a bill-share rate has a denominator

gglucass:
## Description

`SavingsTracker` has accumulated `total_input_cost_usd` since it existed, and
`output_savings_usd` was added later for the shaper (what it stopped the model
emitting). There is no counterpart for what the model DID emit, so the lifetime
block describes the whole input bill and none of the output bill.

That gap has a consequence beyond a missing field. Any figure answering "what
share of my bill did Headroom remove" has to put compression, output shaping
and tool-schema deferral in the numerator, but the only denominator available
is input spend. Output-shaping dollars then sit over an input-only denominator
and the rate reads high. I hit this concretely in a downstream dashboard, where
an all-time rate built that way read 17.0% next to 11.6% for the same month on
the same data, with 2.2pp of the gap being nothing but the extra layers.

This adds the missing denominator half. No existing number changes.

- `lifetime["total_output_cost_usd"]`: emitted output tokens priced at the
  model's output rate.
- Rollup series points gain `total_output_cost_usd_delta` and the cumulative
  `total_output_cost_usd`, matching how `total_input_cost_usd` already appears,
  so per-bucket rates are computable and not just an all-time one.
- Priced by `_estimate_output_cost_usd`, a named passthrough to
  `_estimate_output_savings_usd`. Same rate table, same zero-price handling,
  same fallback, so output spend and output saving can never disagree about a
  model's price. Split out so the call sites read as spend and saving rather
  than one function doing double duty under a "savings" name.
- Fed by `output_tokens`, which `PrometheusMetrics.record_request` already
  accepts and simply never passed down.

## Type of Change

## Changes Made

- `headroom/proxy/savings_tracker.py`: `_estimate_output_cost_usd`;
  `record_request(output_tokens=...)`; `lifetime["total_output_cost_usd"]`
  accumulated and checkpointed into history; `_normalize_history_entry`
  defaults it for legacy entries; `_build_rollup` carries the cumulative and
  the per-bucket delta; CSV export gains `total_output_cost_usd_delta`.
- `headroom/proxy/prometheus_metrics.py`: pass `output_tokens` through to the
  tracker.
- `tests/test_savings_tracker_output_cost.py`: new.
- `tests/test_proxy_savings_history.py`: three exact-shape assertions gain the
  new key (normalized entry, persisted checkpoint, CSV header).

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_proxy_savings_history.py \
  tests/test_savings_tracker_zero_price.py tests/test_cost_tracker_totals.py \
  tests/test_proxy_project_savings.py tests/test_per_model_tool_savings.py \
  tests/test_savings_ledger.py tests/test_output_savings.py \
  tests/test_output_shaping_rollup.py tests/test_savings_tracker_output_cost.py \
  tests/test_savings_tracker_litellm_resolution_cache.py -q
174 passed, 2 skipped, 1 warning in 31.98s

uvx ruff check headroom/proxy/savings_tracker.py headroom/proxy/prometheus_metrics.py \
  tests/test_savings_tracker_output_cost.py tests/test_proxy_savings_history.py
All checks passed!
uvx ruff format --check <same four files>
4 files already formatted

uv run --frozen --extra dev mypy headroom/proxy/savings_tracker.py headroom/proxy/prometheus_metrics.py
```

## Real Behavior Proof

- Environment: macOS 15, Python 3.11, this branch off `main` at 4e1f7769. Live
  figures pulled from a running headroom-ai 0.37.0 proxy on port 6767 carrying
  real Claude Code traffic, 2026-09-10.
- Exact command / steps: confirmed the running proxy's `/stats`
  `persistent_savings.lifetime` block has `total_input_cost_usd` and no output
  counterpart (`"total_output_cost_usd" in lifetime` -> `False`). Then drove
  the patched `SavingsTracker` with that same session's real shape:
  `record_request(model="claude-opus-5", input_tokens=12_809_792,
  tokens_saved=1_609_216, output_tokens=70_442)`.
- Observed result: input 12,809,792 tok -> `total_input_cost_usd` $64.05;
  output 70,442 tok -> `total_output_cost_usd` $1.76, the number that did not
  exist before. The bill-share rate on that session moves from 11.16% with an
  input-only denominator to 10.89% with input + output. The daily series point
  gained `total_output_cost_usd` and `total_output_cost_usd_delta` beside the
  existing `output_savings_usd_delta`.
- Not tested: the overstatement's size is workload-dependent and small here
  because this session is input-heavy (output is 0.5% of tokens, though at ~5x
  the input rate). An output-heavy workload separates the two rates much
  further. I also did not add a `/stats` field beyond what the lifetime block
  already passes through verbatim.

## Runtime Rollout Safety

- Stable/default behavior changed: No. Every existing field keeps its value;
  this only adds keys.
- Kill switch / disable path: None needed, nothing reads the field yet.
- Qualification impact: None.
- Rollback path: Revert. Persisted state with the extra key is read by the
  previous version through `_normalize_history_entry`, which ignores unknown
  keys.

## Review Readiness

## Checklist

N/A

## Additional Notes

One known limitation, called out because a consumer will trip on it otherwise:
the field accumulates from the first request after this ships, while
`total_input_cost_usd` carries the machine's full history. A combined rate
computed on the day of the bump divides a full-history numerator by a
partial-history output denominator. That is the same ramp-up `output_savings_usd`
had when it landed, so I followed the precedent rather than inventing a
backfill, but it means consumers should gate on `total_output_cost_usd > 0` and
treat a 0 as "no coverage" rather than "no output spend". Happy to add an
explicit first-recorded-at marker instead if you would rather it be
self-describing.


JerrettDavis (changes requested):
The additional output-spend accounting fits the intended bill-share denominator. One persistence gap needs correction before it can be treated as a lifetime value.

[P2] In headroom/proxy/savings_tracker.py:911-914, total_output_cost_usd is accumulated and saved, but _sanitize_state does not restore it from lifetime_raw. Using a real tracker and temporary state file, a $0.01 output request followed by reload loses the field; a second identical request leaves lifetime and daily rollup at $0.01 instead of $0.02. Please add it to default state and lifetime recovery/sanitization, with a save/reload/append regression that verifies both lifetime and rollup remain monotonic. The five new tests pass but do not exercise a restart.

gglucass:
Fixed in 3decbe1e.

`total_output_cost_usd` accumulated onto `lifetime` and was checkpointed into history, but `_sanitize_state` never read it back, so a restart restarted the counter at 0 while history kept the old total. The rollup's `max(delta, 0)` clamp then reported that as a flat bucket instead of a negative one, which is why the five tests passed.

`output_tokens_saved` and `output_savings_usd` had the same hole (both accumulate via `lifetime.get(key, 0)` and were recovered nowhere), so all three are now in `_default_state` and recovered in `_sanitize_state` from the lifetime block and the last checkpoint, exactly as the input side already was.

Regression: `test_lifetime_output_spend_survives_a_restart` does record / reload / record and asserts lifetime is $0.02 (not $0.01), `output_savings_usd` 0.004, `output_tokens_saved` 800, and the daily rollup bucket at 0.02 total / 0.02 delta. It fails on the pre-fix tree with `KeyError: 'total_output_cost_usd'` at the reload. Two existing exact-dict assertions on the lifetime block were widened for the three additive keys.

gglucass:
Change request addressed and pushed — ready for review again.

JerrettDavis (approved):
Re-reviewed the current head after its main update. The reload path correctly reconstructs lifetime output cost, savings, and token totals without changing the persistence contract. Focused local validation produced 46 passing tests; the one Windows-only fsync assertion also fails unchanged on pristine main, so it is not introduced by this PR. The current head is conflict-free and all attached checks are complete with no failures. Approved.

JerrettDavis (approved):
Re-reviewed exact head 0fcb9a2cd60b243af380caffefca9abbb87adb7f after its merge from current main. The lifetime output-cost feature and semantic rollup conflict resolution preserve the approved behavior alongside current cache-cost fields. The head is mergeable, contains current main, and all reported checks are terminal green.

PR #3543, opened 2026-09-10
fix(proxy): emit native Anthropic buffered SSE

danseerintel:
## Summary

- omit thinking signatures from content_block_start and emit them only as signature_delta
- stream tool_use and server_tool_use inputs through input_json_delta from an empty input object
- preserve stop_sequence in the message envelope and terminal delta
- catch non-ValueError reconstruction failures so an open stream returns the existing structured 502 SSE error

## Root cause

The buffered CCR path turns an upstream non-stream Message back into SSE. Its hand-built stream differed from Anthropic's contract: thinking signatures appeared in both the start block and a signature delta, server_tool_use started with a fully populated input and emitted no input_json_delta, and stop_sequence was dropped. A TypeError or AttributeError during reconstruction escaped because the call site caught only ValueError, leaving clients with an already-open HTTP 200 stream and no usable message.

This complements the already-merged signed-thinking byte-lock fix in #3124 and buffered Accept fix in #3102.

## Tests

- `python -m pytest tests/test_sse_thinking_blocks.py -q` (16 passed)
- `python -m ruff check headroom/proxy/handlers/streaming.py headroom/proxy/handlers/anthropic.py tests/test_sse_thinking_blocks.py`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/headroom/proxy/handlers/anthropic.py
+++ b/headroom/proxy/handlers/anthropic.py
@@ -4837,10 +4837,11 @@ class AnthropicHandlerMixin:
 
                             try:
                                 sse_events = self._response_to_sse(resp_json, "anthropic")
-                            except ValueError as sse_err:
+                            except Exception as sse_err:
                                 logger.warning(
                                     f"[{request_id}] CCR: Failed to convert buffered response "
-                                    f"to SSE: {sse_err}"
+                                    f"to SSE: {sse_err}",
+                                    exc_info=True,
                                 )
 
                                 async def _conversion_error_sse():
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
Reviewed the buffered Anthropic SSE reconstruction after refreshing from main. Thinking signatures are emitted in signature_delta, server-tool inputs use input_json_delta, stop_sequence is present in the native envelope/delta positions, and conversion failures stay in the existing fail-safe path with diagnostics. All 16 SSE regression tests pass locally; the head is current, mergeable, and attached checks pass. No blocking finding.

PR #3548, opened 2026-09-11
fix: aggregate Anthropic usage across CCR continuations

gagecane:
## Description

Anthropic CCR continuations return final-call-only usage, omitting completed earlier calls. Sum the four principal token counters across completed calls in the shared response handler; preserve unknown required/malformed counters and separate nested cache-creation durations. Content and other final-response metadata remain unchanged.

## Type of Change

- [x] Bug fix

## Changes Made

- Aggregate completed Anthropic CCR calls in the shared handler, preserving unknown counters and separate cache durations.
- Add regression tests and standalone synthetic HTTP reproduction with before/after evidence.

## Testing

- [x] New regression tests and standalone failing-before/passing-after HTTP reproduction
- [x] Affected suite: 106 passed (two upstream deprecation warnings)
- [x] Ruff check and format-check for changed Python files
- [x] Whole-project mypy, installed commit hooks, Rust formatting and Clippy
- [ ] Full upstream suite and all CI gates passing

The ten new unit cases cover multiple calls, non-mutation, unknown counters, nested cache durations, failed continuation and retrieval-round limits. A legacy fixture omitted usage entirely; it now expects unknown counters while retaining the content assertions.

### Test Output

Affected-suite terminal output:

```text
======================= 106 passed, 2 warnings in 12.16s =======================
```

Recorded synthetic HTTP results: [before](https://github.com/gagecane/headroom/blob/d45550bff6867e023399f76a45633b20bc0f59dd/experiments/ccr_usage/usage-main.json) (2 pass, 4 fail; exit 1), [after](https://github.com/gagecane/headroom/blob/d45550bff6867e023399f76a45633b20bc0f59dd/experiments/ccr_usage/usage-main-fixed.json) (6 pass; exit 0).

## Real Behavior Proof

- Environment: Ubuntu, Python 3.13.12; native extension built from main `04cdf79ab0a8423d88148ba63e960ac6b4007b9c`. Synthetic Anthropic-shaped HTTP provider and a real loopback Uvicorn proxy; no paid model or provider credentials. Optimization disabled to isolate CCR; a real seeded in-memory retrieval must succeed before scoring.
- Exact command / steps: follow the [pinned setup and test commands](https://github.com/gagecane/headroom/blob/d45550bff6867e023399f76a45633b20bc0f59dd/experiments/ccr_usage/README.md). Run `python experiments/ccr_usage/reproduce_usage.py --output actual.json` in separate source-built baseline and patched environments with prepared tokenizer cache. Fresh runs against both were repeated for this contribution.
- Before: JSON/SSE × 1/2/3 upstream calls: **2 pass, 4 fail, exit 1**. Multi-call responses contain only final-call usage.
- Observed result: after the fix, **6 pass, exit 0**. Expected input/output/cache-read/cache-create totals for three calls are **66/42/78/102**; all four match. Saved synthetic JSON is included beside the reproduction. Exit 2 distinguishes invalid execution from regression.
- Not tested: live provider invoices, live agent clients, concurrency, retries, every provider, unknown-counter SDK compatibility, or universal billing correctness. On this main revision, the seeded-marker SSE path buffers upstream JSON; this is not proof of an event-level upstream SSE fix.

## Runtime Rollout Safety

- Rollout-managed feature(s): No new rollout-managed feature; this changes existing Anthropic CCR response accounting.
- Minimum rollout channel: Existing CCR availability is unchanged; this patch introduces no channel requirement.
- Stable/default behavior changed: Yes, when Anthropic CCR continuations occur, response usage aggregates completed calls instead of reporting only the final call. Missing required or malformed counters remain unknown; other providers are unchanged.
- Kill switch / disable path: No patch-specific switch. At the response-handler level, `ResponseHandlerConfig(enabled=False)` bypasses CCR response handling, including this aggregation; that also disables automatic retrieval handling. Revert the patch to retain CCR with the previous accounting behavior.
- Qualification impact: Focused tests and synthetic HTTP cases pass. Full upstream verification, live billing and client compatibility remain incomplete as described below. No extra provider calls, retries, persistence or process-death recovery behavior are introduced.
- Rollback path: Revert commit `d45550bff6867e023399f76a45633b20bc0f59dd`; no schema or persistent-state migration is involved.

## Review Readiness

## Additional Notes

Related open PR #3013 changes event-level CCR stream splicing. Its pinned revision `b59ded54e51f81593a17f9433fdac808c9db0e4e` also fails four cases with this unchanged reproduction. Coordination is needed: with this shared-handler fix, that branch must avoid adding initial usage twice and publish cumulative input/cache usage in its terminal SSE delta. This PR targets current main and does not modify the unmerged PR.

No changelog or dependency configuration changed. `dependencies.lock` under experiments is a reproduction constraint snapshot, not a project dependency update.

### Broader verification

Whole-repository Ruff/format and mypy checks pass. Rust formatting/Clippy pass; Rust tests: 1,493 passed, 3 ignored. The native rebuild/import check and designated Python precheck pass: 175 passed, 4 skipped. Commitlint passes. Pushes avoided duplicate pre-push runs because these component checks were executed separately.

### Isolated full-suite results

[Versioned result counts and failed cases](https://github.com/gagecane/headroom/blob/dcfa5043de88276410e97ecd55c02719bf72ebbe/experiments/ccr_usage/isolated-results.json) and [environment, commands and qualifications](https://github.com/gagecane/headroom/blob/dcfa5043de88276410e97ecd55c02719bf72ebbe/experiments/ccr_usage/README.md#complete-isolated-suite-attempt) are preserved with a final dependency snapshot.

The suite reached completion in a Docker container with its own writable home: Ubuntu 26.04, Python 3.13.12, four CPUs, 8 GiB memory, CPU-only Torch/dev dependencies and verified offline all-MiniLM-L6-v2 cache. No provider credentials or host Docker socket were supplied. A 120-second per-test timeout was used after an exploratory run hung in an existing CLI test.

```text
pytest tests scripts/tests --timeout=120 --timeout-method=signal --tb=short -q
7 failed, 12091 passed, 594 skipped, 6138 warnings in 842.80s
```

The previous host-write failures are cleared. Of these seven new failures, five were container setup effects: direct Docker exec gives Python parent PID zero, a mocked-UID supervisor test needs a username environment variable, and an image test needs Pillow. Rerunning all three affected files through a shell parent with `LOGNAME=root` and Pillow installed gives **111 passed, 7 skipped**. Four originally failing cases now pass; the image case skips because its optional model is unavailable offline. It remains unvalidated. Counts overlap; do not sum the rerun into the full-suite total.

The remaining two failures reproduce on verified, unmodified main (`04cdf79`):

- `tests/test_cli_proxy_improvements.py::TestMissingProxyDepsError::test_proxy_command_exits_when_mcp_missing` hangs/times out.
- `tests/test_cli_proxy_improvements.py::TestMissingProxyDepsError::test_ensure_proxy_dependencies_exits_when_fastapi_missing` does not raise the expected SystemExit.

These tests mock `builtins.__import__`; the CLI uses `importlib.import_module`, whose cached-module path bypasses that mock. The baseline file run yields **2 failed, 57 passed** with a 15-second limit. No unrelated test was changed or excluded to obtain a green result. The accounting fix's tests passed in the full run.

**This is not a clean full-suite pass.** The PR remains draft with the baseline failures and optional-model gap visible. No production logic changed during this verification.


gagecane:
Verification update for `dcfa5043de88276410e97ecd55c02719bf72ebbe`: the accounting regression suite and synthetic HTTP before/after reproduction pass as documented in the PR body. The isolated full suite completed; the two remaining CLI failures also reproduce on unmodified main. I am keeping unrelated baseline test repairs out of this PR, per CONTRIBUTING.md.

The upstream **CI, Security, rust, Merge Conflicts, and Changelog Guard** workflows currently report `action_required`; the [CI run](https://github.com/headroomlabs-ai/headroom/actions/runs/34552141776) has zero jobs. Only PR Governance has executed successfully. My account has read-only upstream permissions.

Could a maintainer approve the pending fork workflow runs? I will address relevant CI/review findings in this contribution. Keeping the PR draft pending this verification; no clean full-suite or upstream CI pass is claimed.

JerrettDavis (changes requested):
The additive-counter handling is careful, but the missing-usage branch currently discards valid metadata from the final completed response. When either round lacks a usage dict, _combine_anthropic_usage replaces current['usage'] with dict.fromkeys(...); any non-additive/current-round fields (for example service_tier or newer provider usage metadata) disappear even though they do not depend on the missing prior counters. Please start from a copy of the current usage when it is a dict, set only the aggregate counters/cache details whose totals are unknowable to None, and preserve unrelated current fields. Add a regression with a missing prior usage and a final usage object containing an extra metadata field. The known-counter aggregation and partial-exit behavior otherwise look sound.

PR #3550, opened 2026-09-11
fix(transforms): scope ContentRouter per-request state to the request (#3549)

### LINKED ISSUE #3549: [BUG] Shared ContentRouter request state can be overwritten by concurrent compression requests
### Summary

`ContentRouter` appears to keep several values that belong to an individual compression request on the long-lived router instance.

Examples include runtime compression options and temporary tool/read-protection state.

This is risky when the same `ContentRouter` instance is used concurrently by multiple requests.

A valid interleaving is:

```text
Request A writes request-local router state
Request B overwrites the same instance fields
Request A continues and reads Request B's state
```

This can affect more than metrics. Some of this state participates in routing and exact-read protection decisions.

### Example failure mode

Suppose Request A contains:

```text
tool_call_id = call_A
command      = cat src/a.py
```

The router builds request-local command metadata indicating that `call_A` is a read operation whose result should remain protected.

Before Request A consumes that information, Request B executes on the same router and builds:

```text
tool_call_id = call_B
command      = echo ok
```

If the second request replaces shared command/protection state, Request A can subsequently observe a map that no longer contains `call_A`.

The read-protection decision can then become:

```text
expected:
call_A → protected

observed after concurrent overwrite:
call_A → not found → not protected
```

That can route an exact source-file read into a lossy compression path.

### Suggested fix

I would not solve this with a coarse lock around the router.

That would serialize compression and hide the ownership problem rather than remove it.

Instead, move all per-request mutable state into an immutable or request-owned object, for example:

```python
@dataclass(frozen=True)
class CompressionRequestContext:
    target_ratio: float | None
    force_kompress: bool
    skip_kompress: bool
    kompress_model: str | None
    compression_policy: object | None

    tool_call_args: Mapping[str, object]
    tool_call_commands: Mapping[str, str]

    protected_tool_ids: frozenset[str]
    protected_message_indices: frozenset[int]

    relevance_context: str
```

Then pass it explicitly:

```python
router.apply(
    messages,
    request_context=request_context,
)
```

and through downstream helpers:

```text
apply(request_context)
    ↓
route(request_context)
    ↓
read-protection(request_context)
    ↓
compress(request_context)
```

The long-lived `ContentRouter` should ideally contain only things that are genuinely safe to share across requests:

```text
immutable configuration
compressor instances
thread-safe shared caches
long-lived services/resources
```

It should not need temporary `self._runtime_*`, request tool maps, or request protection sets.

### Alternative incremental fix

If a full request-context refactor is too large for one PR, an incremental migration could be:

1. Create `CompressionRequestContext`.
2. Move tool-call/read-protection state first.
3. Pass it explicitly to helpers that need it.
4. Move runtime compression overrides next.
5. Remove the corresponding `self._runtime_*` and request-temporary instance fields.
6. Keep shared caches independently synchronized.

This allows the refactor to be split without requiring a large one-shot rewrite.

### Deterministic regression test

The concurrency test does not need probabilistic stress or thousands of iterations.

Use synchronization primitives to force the problematic interleaving:

```text
Thread A:
  build request A state
  signal A_READY
  wait for B_WRITTEN

Thread B:
  wait for A_READY
  build request B state
  signal B_WRITTEN

Thread A:
  continue read-protection/compression
```

For Request A:

```text
command = cat src/a.py
```

For Request B:

```text
command = echo ok
```

Assertions should verify that Request A still sees its own read metadata and that its source-file result remains byte-identical.

The stronger regression would assert both:

```text
without concurrent B:
A output == original source bytes

with forced A/B interleaving:
A output == original source bytes
```

After the fix, both executions should produce the same protected result.

### Why this matters

Thread-safe caches do not by themselves make the surrounding router request-safe.

A shared router can safely process concurrent requests only if state that belongs to one request cannot be overwritten by another request.

Keeping request state outside the shared router also makes the ownership model clearer and should reduce the chance of similar races when new runtime options or protection mechanisms are added.

dajiaohuang:
## Description

A `ContentRouter` is long-lived and shared. `headroom/proxy/server.py` builds one instance per provider pipeline and every request runs `apply()` on it, from the request-path `ThreadPoolExecutor` — and, nested inside that, from a second pool for the parallel compression pass.

`apply()` stored two pieces of per-request state on `self`:

- `self._tool_call_args` / `self._tool_call_commands`, written by `_build_tool_name_map()`
- `self._protect_read_tool_ids` / `self._protect_read_msg_indices`, the read-protection sets derived from them

A concurrent request could therefore overwrite them mid-`apply`. The harmful case is read protection: with `HEADROOM_PROTECT_READS=1`, if Request B's scan landed between A's scan and A's read-protection classification, A's `cat src/a.py` observation was no longer recognized as a file read and was handed to lossy compression — destroying exactly the bytes the agent needs for a line-precise edit.

Closes #3549

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add a frozen `CompressionRequestContext` holding the per-request tool-call maps and read-protection sets.
- Split the old `_build_tool_name_map()` into `_scan_tool_calls()` (returns the three maps instead of storing them on `self`) and `_build_request_context()` (the scan plus read-protection classification). `_build_tool_name_map()` is kept as a thin wrapper over the scan, so its existing caller is unaffected.
- `apply()` builds one context per request and threads it through the string path (`_bash_search_fold`, the read-observation branch) and the content-block path (`_process_content_blocks`).
- Delete the `self._tool_call_args` / `self._tool_call_commands` instance fields and the inline `self._protect_read_*` writes in `apply()`.
- Widen `_process_content_blocks`'s `tool_name_map` parameter to `Mapping[str, str] | None` — the block path only reads it, so no per-request dict copy is needed.
- Add `tests/test_content_router_request_context_isolation.py`, a deterministic regression test. Repoint three white-box assertions in `tests/test_observed_wire_shapes.py` that read the removed private attributes at the new per-request builder, preserving their original semantics.

`self._runtime_*` (target ratio, force/skip kompress, model, policy) is deliberately **not** moved in this PR — see Additional Notes.

## Testing

### Test Output

New regression test, unpatched router (RED):

```text
$ git checkout -- headroom/transforms/content_router.py
$ uv run pytest tests/test_content_router_request_context_isolation.py -p no:randomly -q
tests\test_content_router_request_context_isolation.py ..F               [100%]

E   AssertionError: request B overwrote request A's read-protection state:
    A's `cat src/a.py` result was handed to lossy compression
E   assert 'LOSSY <<ccr:deadbeef>>' == 'def f():\n  ...\n    x = 1\n'
========================= 1 failed, 2 passed in 0.54s =========================
```

Same test, patched router (GREEN) — only the source file changed:

```text
$ uv run pytest tests/test_content_router_request_context_isolation.py -p no:randomly -q
============================== 3 passed in 0.42s ==============================
```

Existing suite:

```text
$ uv run pytest tests/test_observed_wire_shapes.py \
      tests/test_content_router_request_context_isolation.py -p no:randomly -q
22 passed

$ uv run pytest tests/test_content_router_*.py tests/test_transforms_content_router.py \
      tests/test_observed_wire_shapes.py tests/test_compression_batches.py \
      tests/test_compression_units.py tests/test_compression_determinism.py \
      tests/test_compression_cache.py tests/test_compression_safety_rails.py \
      tests/test_code_compressor_thread_safety.py tests/test_compression_store.py \
      tests/test_compression_observability.py -p no:randomly -q
286 passed, 8 skipped

$ uv run ruff check <changed files>      -> All checks passed!
$ uv run ruff format --check <changed files> -> 3 files already formatted
$ uv run mypy headroom/transforms/content_router.py -> Success: no issues found
```

The 8 skips are pre-existing in `test_code_compressor_thread_safety.py` (native/Rust-gated), not introduced here.

## Real Behavior Proof

- Environment: Windows 10 Pro N (10.0.19045), Python 3.13.8, pytest 9.0.3, `uv sync --extra dev` (native `headroom._core` built), base `origin/main` @ `04cdf79a`.

- Exact command / steps: Run the RED and GREEN pytest commands shown below, then the focused pytest, ruff, and mypy commands in the Test Output section.\n
```text
git checkout -- headroom/transforms/content_router.py      # unpatched
uv run pytest tests/test_content_router_request_context_isolation.py -p no:randomly -q

cp content_router_fixed.py headroom/transforms/content_router.py   # patched
uv run pytest tests/test_content_router_request_context_isolation.py -p no:randomly -q
```

- Observed result: unpatched → `1 failed, 2 passed`, with Request A's `cat src/a.py` bytes replaced by the lossy sentinel. Patched → `3 passed`, A's bytes byte-identical. Instrumentation on the unpatched router shows the shared map being overwritten inside A's `apply()` (`commands before B: {'call_A': 'cat src/a.py'}` → `commands after B: {'call_B': 'echo ok'}`), A's `_protect_read_tool_ids` ending up empty, `router:read_protected` absent from `transforms_applied`, and `compress()` being called on A's file bytes.

- Not tested: not run through a live proxy process with real concurrent HTTP traffic — the interleaving is forced with `threading.Event` synchronization rather than reproduced from a loaded server. No provider calls; no Modal. The Rust/native compression path is unchanged by this patch but was not separately exercised. `_runtime_*` request state is untouched here and still carries the same class of hazard (follow-up, see below). Full `tests/` run in progress at time of opening; results will be posted as a comment.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. No feature flag gates this change.
- Minimum rollout channel: n/a — pure bug fix, no new behavior.
- Stable/default behavior changed: No. Single-request behavior is identical: the same maps are computed from the same messages and the same classification runs. The only intended difference is under concurrency, where a request now reads its own state. `HEADROOM_PROTECT_READS` semantics are unchanged.
- Kill switch / disable path: existing `HEADROOM_PROTECT_READS=0` disables read protection entirely, as before.
- Qualification impact: none.
- Rollback path: revert the commit (single self-contained module change plus tests).

## Review Readiness

- [ ] I have performed a self-review

## Checklist

## Additional Notes

**Scope: why `self._runtime_*` is not in this PR.** `apply()` also stores `_runtime_target_ratio`, `_runtime_force_kompress`, `_runtime_skip_kompress`, `_runtime_kompress_model` and `_runtime_compression_policy` on `self` and comments them "Store runtime options on self for access by `_route_and_compress_block`". They are the same class of hazard, but they are *mutated from outside the router*: `headroom/transforms/compression_units.py` and `compression_batches.py` save and restore them around their own calls into the router. Moving them changes a cross-module contract and is a materially larger change than this one, so it is left as a follow-up to keep this PR to one logical change.

**Why a context object rather than a `contextvar`.** The parallel compression pass runs on a nested `concurrent.futures.ThreadPoolExecutor`. `asyncio.run_in_executor` copies context, but a bare `ThreadPoolExecutor.submit()` does not — a contextvar would silently lose read protection in exactly the pass this issue is about. Passing the context explicitly keeps it correct without serializing anything, so compression stays parallel (no lock).

**Test-adaptation note.** Three assertions in `tests/test_observed_wire_shapes.py` read the private attributes this patch removes. They now call `_build_request_context()` and assert on its fields; their original assertion values and intent (which message index / tool id is classified as a read) are unchanged, and both tests still assert the public verbatim-output outcome.

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -44,7 +44,7 @@ import re
 import sys
 import threading
 import time
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field, replace
 from enum import Enum
@@ -1706,6 +1706,29 @@ class ContentRouterConfig:
     search_group_by_file: bool = False
 
 
+@dataclass(frozen=True)
+class CompressionRequestContext:
+    """Immutable per-request state for one ``ContentRouter.apply`` call.
+
+    A ``ContentRouter`` is long-lived and shared: the proxy builds one per provider
+    pipeline and runs ``apply`` on a thread pool, so a request-scoped value stored on
+    the instance can be overwritten mid-``apply`` by a concurrent request (#3549).
+    Building this once per request and passing it down keeps that state owned by the
+    request instead of the router.
+    """
+
+    tool_names: Mapping[str, str] = field(default_factory=dict)
+    tool_call_args: Mapping[str, str] = field(default_factory=dict)
+    tool_call_commands: Mapping[str, str] = field(default_factory=dict)
+    protected_tool_ids: frozenset[str] = frozenset()
+    protected_message_indices: frozenset[int] = frozenset()
+
+
+# Stand-in for helpers reached without going through ``apply``: no tool metadata,
+# no read protection.
+_EMPTY_REQUEST_CONTEXT = CompressionRequestContext()
+
+
 class ContentRouter(Transform):
     """Intelligent router that selects optimal compression strategy.
 
@@ -1850,11 +1873,6 @@ class ContentRouter(Transform):
         # different strategies for the same block, share one provider
         # invocation. See `_lossless_provider_result`.
         self._lossless_provider_memo: dict[tuple[int, int, int], tuple[str, str] | None] = {}
-
-        # tool_call_id → compact args text, populated by _build_tool_name_map.
-        self._tool_call_args: dict[str, str] = {}
-        # tool_call_id → raw shell command (bash-search fold), same population.
-        self._tool_call_commands: dict[str, str] = {}
 
         # Phase 0 (#1171): cap the input size handed to kompress (ModernBERT
         # ONNX). Its inference scales O(tokens) and runs synchronously on the
@@ -4529,16 +4547,14 @@ class ContentRouter(Transform):
 
     # Transform interface
 
-    def _build_tool_name_map(self, messages: list[dict[str, Any]]) -> dict[str, str]:
-        """Build mapping from tool_call_id to tool_name.
+    def _scan_tool_calls(
+        self, messages: list[dict[str, Any]]
+    ) -> tuple[dict[str, str], dict[str, str], dict[str, str]]:
+        """Collect tool-call metadata in one scan: names, compact args, shell commands.
 
-        Scans assistant messages to find tool calls and extract their names.
-        Supports both OpenAI and Anthropic message formats. Also populates
-        ``self._tool_call_args`` (id → compact args text) in the same scan, so
-        the relevance split can score a tool output against the *precise* ask
-        that triggered it (grep pattern, read path, …), not just the user
-        prompt. Read-only after build → safe to read from the parallel
-        compression pass.
+        Supports both OpenAI and Anthropic message formats. Returns the values
+        instead of storing them on the router, so a concurrent request cannot
+        overwrite them mid-``apply`` (#3549).
         """
         mapping: dict[str, str] = {}
         args_map: dict[str, str] = {}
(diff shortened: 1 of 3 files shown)


JerrettDavis (approved):
Reviewed the request-context refactor after refreshing from main. Tool names, arguments, shell commands, and read-protection sets are immutable per apply call and are passed explicitly through both string and block paths, eliminating the shared-router overwrite without relying on thread-local propagation. All 22 targeted concurrency/wire-shape tests pass locally; the head is current, mergeable, and attached checks pass. No blocking finding.

JerrettDavis:
This previously approved branch is now conflicted because its core fix has landed on `main` as #3556 (`0bef39811`, per-request ContentRouter runtime-state isolation with current concurrency coverage). Rebasing this older competing implementation would reintroduce overlapping state/context machinery and duplicate tests. Recommend closing #3550 as superseded by #3556 rather than resolving the conflict.

PR #3553, opened 2026-09-11
fix(proxy): keep non text blocks in place when relocating system sections

### LINKED ISSUE #3552: [BUG] Mid-conversation system hoist uses a Claude-only allow-list, so third-party Anthropic-compatible upstreams get image blocks moved into `system` → HTTP 422
## Description

#2991 narrowed the #765 relocation guard with a model gate. That gate is a hardcoded
**Claude-family allow-list**, so it never matches a third-party Anthropic-compatible
upstream reached via the documented per-request `X-Headroom-Base-Url` override:

```python
# headroom/proxy/helpers.py:926  (relocate_system_messages_to_top_level)
model_id = str(model or "").lower()
supports_mid_conversation = any(
    family in model_id
    for family in (
        "claude-fable-5", "claude-mythos-5", "claude-opus-4-8",
        "claude-opus-5", "claude-sonnet-5",
    )
)
```

The call site makes this strictly worse for non-Anthropic upstreams — it passes `model=None`,
which yields `model_id = ""` and therefore `supports_mid_conversation = False`:

```python
# headroom/proxy/handlers/anthropic.py:3310
relocate_system_messages_to_top_level(
    body["messages"], body.get("system"),
    (str(model) if (not upstream_base_url
                    or getattr(self, "anthropic_backend", None) is not None
                    or _is_googleapis_endpoint(upstream_base_url))
     else None),
)
```

So for a custom upstream the guard hoists **every** mid-conversation system section, on every
request. Claude Code 2.1.268 sends `mid-conversation-system-2026-04-07` unconditionally, so
this fires constantly.

The hoist is not content-preserving. `_system_message_to_blocks` appends every block dict
verbatim:

```python
# headroom/proxy/helpers.py:889
for block in content:
    if isinstance(block, dict):
        blocks.append(block)          # image/document blocks included
```

When a user pastes a screenshot, Claude Code attaches the image block to a system section.
The hoist therefore moves an **image block into the top-level `system` parameter**. z.ai's
Anthropic-compatible endpoint accepts text blocks only in `system`, and rejects it:

```
HTTP 422
{"detail":[{"type":"string_type","loc":["body","system","str"],
            "msg":"Input should be a valid string","input":[...]}]}
```

Claude Code surfaces this as `API Error: 422` and the session is dead on its first message.

## Reproduction

Identical body, two paths. Only the proxy hop differs.

```bash
BODY='{"model":"glm-5.3","max_tokens":16,
 "system":[{"type":"text","text":"You are Claude Code."}],
 "messages":[
  {"role":"user","content":[{"type":"text","text":"look at this"}]},
  {"role":"system","content":[
     {"type":"text","text":"<system-reminder>image attached</system-reminder>"},
     {"type":"image","source":{"type":"base64","media_type":"image/png","data":"<1x1 png>"}}]},
  {"role":"assistant","content":[{"type":"text","text":"ok"}]},
  {"role":"user","content":[{"type":"text","text":"hi"}]}]}'

# through headroom (X-Headroom-Base-Url -> z.ai)
curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8787/v1/messages \
  -H "x-api-key: $ZAI" -H 'anthropic-version: 2023-06-01' \
  -H 'X-Headroom-Base-Url: https://api.z.ai/api/anthropic' \
  -H 'content-type: application/json' -d "$BODY"
# -> 422

# direct to the same upstream
curl -s -o /dev/null -w '%{http_code}\n' https://api.z.ai/api/anthropic/v1/messages \
  -H "x-api-key: $ZAI" -H 'anthropic-version: 2023-06-01' \
  -H 'content-type: application/json' -d "$BODY"
# -> 200
```

Proxy log for the failing request:

```
WARNING [hr_...] Relocated role=system message(s) out of messages[] into the
        top-level system parameter (Anthropic wire-contract guard, issue #765)
WARNING event=cache_breakpoints in_system=2 out_system=3 in_messages=1 out_messages=0
        in_msg_count=2 out_msg_count=1 out_last_tail=-1 tail_grew=true
INFO    event=outbound_request mutation_reasons=image_compression,system_role_relocated,...
WARNING [hr_...] UPSTREAM_ERROR status=422 model=glm-5.3 compressed=yes
        transforms=['router:text_block:mixed', 'anthropic:tool_schema_compaction'] stream=True
```

## Expected

The request that succeeds when sent directly should also succeed through the proxy.

## Why it matters

1. **It breaks working requests.** The upstream accepts `role:"system"` inside `messages[]`
   natively; the guard converts a 200 into a 422. Headroom is the only thing that fails here.
2. **The allow-list can't be right for upstreams Headroom doesn't own.** GLM, and any other
   Anthropic-compatible gateway, will never appear in a list of Claude model families, so the
   conservative branch is permanent for them — this isn't a stale-list problem that adding
   `glm-*` fixes.
3. **Hoisting into `system` isn't a safe fallback anyway.** `system` is more constrained than
   `messages[]` on at least one real upstream (text blocks only), so the "safe" direction moves
   content into a *narrower* slot. Non-text blocks should never be hoisted regardless of model.
4. It also re-introduces #2991's cache and semantics costs for every custom-upstream user.

## Suggested fix

Any one of these resolves it; (1) and (2) are complementary:

1. **Don't hoist non-text blocks.** If a system section contains a block whose `type` is not
   `text`, leave the section in `messages[]`. `system` is text-only on Anthropic too, so this
   is a correctness fix independent of upstream.
2. **Skip the guard for foreign upstreams.** When a request carries `X-Headroom-Base-Url`
   (or otherwise targets a non-Anthropic, non-Vertex upstream), Headroom doesn't know the
   wire contract and shouldn't rewrite the client's body to match a guessed one. Passing
   `model=None` currently means "hoist everything", which is the opposite of conservative.
3. **Make the allow-list configurable** (env or config), so operators of Anthropic-compatible
   gateways can opt out without patching `site-packages`.

## Environment

- **Headroom version**: 0.37.0 (`headroom-ai`); `helpers.py` and `handlers/anthropic.py` on
  `main` are identical to the installed copy at time of filing
- **Python**: 3.11.5
- **OS**: macOS (Darwin 27.0.0)
- **Client**: Claude Code CLI 2.1.268, betas include `mid-conversation-system-2026-04-07`
- **Upstream**: `https://api.z.ai/api/anthropic`, model `glm-5.3`, via `X-Headroom-Base-Url`

## Related

- #2991 — added the model gate this issue is about
- #765 — the guard's original motivation

hktitof:
## Description

Closes #3552

when a payload carries a mid conversation system message holding non text blocks, `relocate_system_messages_to_top_level` hoisted the whole thing into the top level `system` parameter, image and document blocks included
the top level `system` parameter only takes text, so anthropic compatible upstreams that type `system` as a string reject the request, the reporter hit `Input should be a valid string` with `loc body system str` on a z.ai style endpoint
the fix keeps the hoist text only: text blocks and bare strings move up, non text blocks stay in a system message at the original position, nothing is dropped and the message order is untouched

### Steps to reproduce
1. run the new tests on untouched main: `python -m pytest -q tests/test_proxy_handler_helpers.py::test_relocate_system_messages_keeps_image_blocks_out_of_top_level_system`
2. Expected (after this fix): text moves to top level `system`, the image block stays in a mid conversation system message
3. Actual (raw output on untouched main 04cdf79a):

```text
FAILED tests/test_proxy_handler_helpers.py::test_relocate_system_messages_keeps_image_blocks_out_of_top_level_system
FAILED tests/test_proxy_handler_helpers.py::test_relocate_system_messages_hoists_only_text_from_mixed_sections
FAILED tests/test_proxy_handler_helpers.py::test_relocate_system_messages_image_only_sections_pass_through_unchanged
========================= 3 failed, 53 passed in 1.95s =========================
```

an image only system section was also needlessly rewritten into a top level system list with an image block in it, which is exactly the shape upstreams choke on

## Type of Change

## Changes Made

- `headroom/proxy/helpers.py`: the hoist now splits each relocated system section, text blocks and bare strings move to the top level `system` parameter, non text blocks stay behind in a system message at the original spot, sections that hold nothing text shaped pass through unchanged, existing behavior for text only and string content is byte identical
- `tests/test_proxy_handler_helpers.py`: 3 regression tests, image block kept out of top level system, mixed section hoists text only and retains the image, image only section passes through unchanged

## Testing

### Test Output

```text
python -m pytest -q tests/test_proxy_handler_helpers.py
56 passed in 1.93s

without the fix (git restore --source main -- headroom/proxy/helpers.py):
3 failed, 53 passed
(the 3 new tests fail, every pre existing test still passes)

ruff check .
All checks passed!

ruff format --check .
1577 files already formatted

mypy headroom
Success: no issues found in 532 source files
```

## Real Behavior Proof

- Environment: linux, python 3.12.3, headroom main 04cdf79a plus the fix (4f15cc02) in a venv, no live provider call involved
- Exact command / steps: the pytest commands in the test output block, plus a restore dance, restoring main `helpers.py` turns the 3 new tests red, restoring the fix turns them green, so the tests fail without the change and pass with it
- Observed result: after the fix the top level `system` list only ever contains text blocks and the image block survives in a mid conversation system message, which is the wire shape upstreams typing `system` as a string accept
- Not tested: a live call against a z.ai or similar endpoint, i verified the wire shape at the helper level, the reporter's exact upstream config is not available to me

## Runtime Rollout Safety

- Stable/default behavior changed: yes, mid conversation system sections with non text blocks keep those blocks in place instead of moving them into the top level `system` parameter, text only and string content payloads are byte identical, that is the fix
- Kill switch / disable path: none needed, revert the commit
- Qualification impact: none
- Rollback path: revert the one commit, nothing else to unwind

## Review Readiness

### DIFF
--- a/headroom/proxy/helpers.py
+++ b/headroom/proxy/helpers.py
@@ -917,7 +917,9 @@ def relocate_system_messages_to_top_level(
 
     The relocated content is appended after any existing top-level ``system``
     so wire order (system prompt, then conversation) is preserved and no content
-    is dropped.
+    is dropped. Only text-shaped content moves: non-text blocks (images,
+    documents) stay in a mid-conversation system section at their original
+    position, because top-level `system` accepts text blocks only (issue #3552).
 
     Returns ``(clean_messages, new_system, changed)``. When no system-role
     message is present the inputs pass through unchanged (``changed=False``) so
@@ -979,14 +981,48 @@ def relocate_system_messages_to_top_level(
         return messages, system, False
 
     relocated_blocks: list[Any] = []
+    retained: dict[int, dict[str, Any]] = {}
     for i in sorted(system_indices):
-        relocated_blocks.extend(_system_message_to_blocks(messages[i]))
-
-    clean_messages = [m for i, m in enumerate(messages) if i not in system_indices]
+        message = messages[i]
+        content = message.get("content") if isinstance(message, dict) else None
+        if isinstance(content, list):
+            # Only text-shaped content may move into the top-level ``system``
+            # parameter (text blocks and bare strings). Non-text blocks such as
+            # images or documents stay in place so nothing is dropped and
+            # upstreams that reject non-text system blocks keep working
+            # (issue #3552).
+            hoisted_from_list: list[Any] = []
+            leftovers: list[Any] = []
+            for block in content:
+                if isinstance(block, dict) and block.get("type") == _TEXT_BLOCK_TYPE:
+                    hoisted_from_list.append(block)
+                elif isinstance(block, str) and block:
+                    hoisted_from_list.append({"type": _TEXT_BLOCK_TYPE, "text": block})
+                else:
+                    leftovers.append(block)
+            relocated_blocks.extend(hoisted_from_list)
+            if leftovers:
+                retained[i] = {**message, "content": leftovers}
+        else:
+            # String (and other) content converts losslessly to text blocks.
+            relocated_blocks.extend(_system_message_to_blocks(message))
 
     if not relocated_blocks:
+        if retained:
+            # Nothing text-shaped to relocate: the sections stay as they are.
+            return messages, system, False
         # System message(s) carried no content — drop the empty entries only.
+        clean_messages = [m for i, m in enumerate(messages) if i not in system_indices]
         return clean_messages, system, True
+
+    clean_messages = []
+    for i, message in enumerate(messages):
+        if i in system_indices:
+            trimmed = retained.get(i)
+            if trimmed is not None:
+                clean_messages.append(trimmed)
+            continue
+        clean_messages.append(message)
 
     if system is None or system == "" or system == []:
         new_system: Any = relocated_blocks
--- a/tests/test_proxy_handler_helpers.py
+++ b/tests/test_proxy_handler_helpers.py
@@ -404,6 +404,91 @@ def test_relocate_system_messages_moves_valid_shape_for_unsupported_model() -> N
     assert system == [{"type": "text", "text": "mid-turn instruction"}]
 
 
+def test_relocate_system_messages_keeps_image_blocks_out_of_top_level_system() -> None:
+    image_block = {
+        "type": "image",
+        "source": {"type": "base64", "media_type": "image/png", "data": "aGk="},
+    }
+    messages = [
+        {"role": "user", "content": [{"type": "text", "text": "look at this"}]},
+        {
+            "role": "system",
+            "content": [
+                {"type": "text", "text": "<system-reminder>image attached</system-reminder>"},
+                image_block,
+            ],
+        },
+        {"role": "assistant", "content": [{"type": "text", "text": "ok"}]},
+        {"role": "user", "content": [{"type": "text", "text": "hi"}]},
+    ]
+    system = [{"type": "text", "text": "You are Claude Code."}]
+
+    clean, new_system, changed = relocate_system_messages_to_top_level(messages, system, None)
+
+    assert changed is True
+    assert isinstance(new_system, list)
+    assert all(not isinstance(block, dict) or block.get("type") == "text" for block in new_system)
+    retained = [
+        message.get("content")
+        for message in clean
+        if isinstance(message, dict) and message.get("role") == "system"
+    ]
+    assert any(image_block in (content or []) for content in retained)
+
+
+def test_relocate_system_messages_hoists_only_text_from_mixed_sections() -> None:
+    messages = [
+        {"role": "user", "content": "question"},
+        {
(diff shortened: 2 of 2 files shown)


JerrettDavis (approved):
Reviewed the mixed system-section handling after refreshing from main. Text is hoisted in order, image/document blocks are preserved byte-for-byte in place rather than being placed in Anthropic's text-only top-level system field, and image-only sections remain unchanged. All 56 helper tests pass locally on the updated head; it is current, mergeable, and attached checks pass. No blocking finding.

hktitof:
all 56 helper tests still pass on the refreshed branch, ready whenever you have time @JerrettDavis!

PR #3556, opened 2026-09-11
fix(transforms): isolate ContentRouter per-request runtime state across concurrent requests

### LINKED ISSUE #3486: [BUG] Concurrent requests cross-contaminate compression options — one shared `ContentRouter` stores per-request state on `self`

liquidweavergit:
## Summary

Fixes #3486.

`ContentRouter` is instantiated once at proxy startup and shared, as a single object, across every concurrent request — `apply()` calls are dispatched onto a real `ThreadPoolExecutor` (`headroom/proxy/server.py`). `apply()` stored per-call state (the F2.2 `CompressionPolicy`, `target_ratio`, `force_kompress`/`skip_kompress`, `kompress_model`, and the tool-call-id maps built per request) as plain instance attributes with **no locking and no per-request scoping**. A second, concurrent `apply()` call on a different worker thread could overwrite a first request's values mid-flight.

Concretely this is a cross-tenant data leak: a Subscription-mode request (`toin_read_only=True` — must never write to the shared cross-user TOIN learning pool) could have its `self._runtime_compression_policy` clobbered by a concurrent PAYG request, causing its content to be incorrectly written into the shared pool.

- **Reproduction / regression test** (first commit): forces the race deterministically with a `threading.Event`-paused `apply()` call and a concurrent interloper `apply()` call on the same shared router instance. Fails reliably (5/5 runs) against the current code, asserting `1 == 0` extra TOIN writes.
- **Fix** (second commit): bundles the affected fields into a `_PerRequestRuntimeState` dataclass stored in a per-instance `ContextVar`, bound fresh at the top of every `apply()` call. `ContextVar` values are Context-local per OS thread by default (no cross-thread propagation), so concurrent `apply()` calls on different `ThreadPoolExecutor` workers are now fully isolated. All existing internal read/write call sites (`self._runtime_x = value`, `getattr(self, "_runtime_x", default)`) are unchanged — properties proxy them onto the `ContextVar` transparently, so no other call site in the ~1000-line `apply()` method needed to change. Direct callers that never go through `apply()` (`compress()`, tests calling `_record_to_toin()` directly, etc.) keep their pre-fix behavior via a lazily-created per-thread default.

## Test plan

- [x] New regression test (`tests/test_content_router_concurrency.py`) fails reliably against the pre-fix code (5/5 runs) and passes deterministically against the fix (5/5 runs).
- [x] `tests/test_transforms/test_content_router.py`, `tests/test_compression_policy_toin_gate.py`, `tests/test_code_compressor_thread_safety.py`, `tests/test_hermes_tool_call_unwrap.py`, `tests/test_observed_wire_shapes.py`, and other `ContentRouter`-touching suites pass unchanged.
- [x] `ruff check` and `mypy` clean on the changed file.
- [x] Full suite (`pytest tests/`): 12021 passed, 592 skipped — identical to the pre-fix baseline aside from the new test flipping fail→pass. The 7 failures in `test_proxy_health.py` / `test_read_maturation_handler_nobust.py` are pre-existing full-suite ordering flakes, independently reproduced on the unfixed baseline (unrelated to this change).

### DIFF
--- a/headroom/transforms/content_router.py
+++ b/headroom/transforms/content_router.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 
 import asyncio
 import hashlib
+import itertools
 import json
 import logging
 import math
@@ -46,6 +47,7 @@ import threading
 import time
 from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor
+from contextvars import ContextVar
 from dataclasses import dataclass, field, replace
 from enum import Enum
 from typing import Any
@@ -1706,6 +1708,47 @@ class ContentRouterConfig:
     search_group_by_file: bool = False
 
 
+@dataclass
+class _PerRequestRuntimeState:
+    """Per-``apply()``-call state, isolated across concurrent requests (#3486).
+
+    ``ContentRouter`` is instantiated once at proxy startup and shared, as a
+    single Python object, across every concurrent request — ``apply()`` calls
+    are dispatched onto a real ``ThreadPoolExecutor`` (see
+    ``headroom/proxy/server.py``). Before this fix, ``apply()`` stashed this
+    exact state (compression policy, runtime overrides, tool-call maps) as
+    plain, unsynchronized attributes directly on ``self``, so one request's
+    ``apply()`` call could read a *different, concurrently-running* request's
+    values — a cross-tenant data leak (e.g. a Subscription-mode request
+    incorrectly writing its content into the shared, cross-user TOIN
+    learning pool because a concurrent PAYG request overwrote the policy
+    field mid-flight).
+
+    Every field here previously lived directly on ``ContentRouter`` as
+    ``self._runtime_*`` / ``self._tool_call_*``. They're bundled into one
+    dataclass so ``apply()`` can bind a brand-new, request-exclusive instance
+    of it into a ``ContextVar`` (see ``ContentRouter._runtime_state_var``)
+    instead of mutating shared instance state. ``ContextVar`` values are
+    Context-local per OS thread by default (no propagation across threads
+    unless explicitly copied), which is exactly the isolation two concurrent
+    ``ThreadPoolExecutor`` workers need.
+    """
+
+    compression_policy: Any = None
+    target_ratio: float | None = None
+    force_kompress: bool = False
+    skip_kompress: bool = False
+    kompress_model: str | None = None
+    tool_call_args: dict[str, str] = field(default_factory=dict)
+    tool_call_commands: dict[str, str] = field(default_factory=dict)
+
+
+# Monotonic counter so each ContentRouter instance gets a uniquely named
+# ContextVar (names only matter for repr/debugging; uniqueness avoids any
+# confusion when introspecting multiple router instances in one process).
+_runtime_state_var_ids = itertools.count()
+
+
 class ContentRouter(Transform):
     """Intelligent router that selects optimal compression strategy.
 
@@ -1851,10 +1894,21 @@ class ContentRouter(Transform):
         # invocation. See `_lossless_provider_result`.
         self._lossless_provider_memo: dict[tuple[int, int, int], tuple[str, str] | None] = {}
 
-        # tool_call_id → compact args text, populated by _build_tool_name_map.
-        self._tool_call_args: dict[str, str] = {}
-        # tool_call_id → raw shell command (bash-search fold), same population.
-        self._tool_call_commands: dict[str, str] = {}
+        # #3486: per-request runtime state (compression policy, runtime
+        # overrides, tool-call maps — see `_PerRequestRuntimeState`) lives in
+        # a ContextVar, not on `self`, so concurrent `apply()` calls on this
+        # shared singleton can't observe or clobber each other's values.
+        # `apply()` binds a fresh `_PerRequestRuntimeState()` at the top of
+        # every call. The default below only matters for callers that use
+        # `self._tool_call_args` / `self.compress()` / `_record_to_toin()`
+        # directly, without going through `apply()` — same pre-#3486
+        # behaviour those callers already relied on (a single, instance-
+        # scoped default they can read and mutate).
+        self._default_runtime_state = _PerRequestRuntimeState()
+        self._runtime_state_var: ContextVar[_PerRequestRuntimeState] = ContextVar(
+            f"headroom_content_router_runtime_state_{next(_runtime_state_var_ids)}",
+            default=self._default_runtime_state,
+        )
 
         # Phase 0 (#1171): cap the input size handed to kompress (ModernBERT
         # ONNX). Its inference scales O(tokens) and runs synchronously on the
@@ -1922,12 +1976,11 @@ class ContentRouter(Transform):
         # ``kwargs["compression_policy"]`` at the start of ``apply()``
         # and read by ``_record_to_toin`` to gate TOIN writes when
         # ``policy.toin_read_only`` is true (Subscription mode).
-        # Defaults to ``None`` so direct ``compress()`` callers (e.g.
-        # tests, hand-written pipelines that don't go through the
-        # proxy) keep pre-F2.2 behaviour: TOIN writes are not gated.
-        # Same pattern the existing ``_runtime_target_ratio`` /
-        # ``_runtime_kompress_model`` fields below use.
-        self._runtime_compression_policy: Any = None
+        # Defaults to ``None`` (see ``_PerRequestRuntimeState``) so direct
+        # ``compress()`` callers (e.g. tests, hand-written pipelines that
+        # don't go through the proxy) keep pre-F2.2 behaviour: TOIN writes
+        # are not gated. Now backed by the #3486 ContextVar state — see
+        # the ``_runtime_compression_policy`` property below.
 
         self._cache = CompressionCache()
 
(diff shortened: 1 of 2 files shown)


JerrettDavis (changes requested):
The outer-request isolation is directionally right, but the ContextVar state is not propagated into the worker threads that apply() itself creates. Python starts ThreadPoolExecutor/threading.Thread workers with a fresh context; this patch imports only ContextVar and leaves the existing executor.submit(self._timed_compress, ...) and watchdog Thread targets unchanged. Those workers therefore read _default_runtime_state, losing this request's target_ratio, force/skip flags, model, policy, and tool maps. In particular, a forced or subscription-policy request can execute fan-out compression under defaults. Please run each spawned compression task under its own copy_context() (one copied Context per task, since a Context cannot be entered concurrently), and add a regression that observes request options inside the internal fan-out—not only two outer apply threads.

liquidweavergit:
## Follow-up: closed two more isolation gaps found in review

After the initial fix landed, I did an adversarial review of the `ContextVar` approach specifically looking for places where the "isolation is automatic" assumption might not actually hold. Found two more instances of the same #3486-class bug and fixed both.

### What was missed

1. **Internal fan-out threads don't inherit the calling thread's `Context`.** `apply()`'s Pass 2 (compression) sometimes spawns its own threads internally — a watchdog `threading.Thread` for the single-task case, and a `ThreadPoolExecutor` when more than one message needs compressing. Neither `threading.Thread` nor `ThreadPoolExecutor.submit` copies the calling thread's `contextvars.Context` to the worker. So even though the *top-level* `apply()` calls were correctly isolated from each other, a worker thread spawned *inside* one `apply()` call would read the `ContextVar`'s global default (an empty `_PerRequestRuntimeState()`) instead of the state that call just bound — silently dropping `force_kompress` / `target_ratio` / `kompress_model` / `compression_policy` for that compressed block.

   Fixed by snapshotting `copy_context()` at both spawn sites and running the worker via `ctx.run(...)`. The thread-pool branch takes a **fresh snapshot per submitted task**, not one shared/hoisted snapshot — a single `Context` object raises `RuntimeError` if `.run()`'d from two threads concurrently, so each fanned-out task needs its own.

2. **Read-protection state wasn't migrated at all.** `_protect_read_tool_ids` / `_protect_read_msg_indices` (the `HEADROOM_PROTECT_READS` bookkeeping for which tool results are genuine file reads that must stay byte-exact) were still plain, unguarded instance attributes — the exact same race as the fields already moved into `_PerRequestRuntimeState`, just with a different consequence (a real file read losing its byte-exact protection, rather than a compression policy misapplication). Moved both into the same `ContextVar`-backed dataclass/property pattern as the rest of the per-request fields.

Also tightened the `_PerRequestRuntimeState` docstring, which previously implied the `ContextVar` gives blanket isolation across all of `apply()`'s concurrency. That's true for top-level `apply()` calls on distinct threads, but not for threads `apply()` spawns internally — added a CAUTION section naming both spawn sites so this doesn't get reintroduced later.

### How it's tested

Four new tests in `tests/test_content_router_concurrency.py`, each written first and confirmed to fail against the pre-fix code before the fix was applied (RED), then confirmed passing after (GREEN):

- `test_parallel_fanout_workers_lose_this_requests_runtime_state` — reproduces gap #1 for the `ThreadPoolExecutor` fan-out path.
- `test_single_pending_task_watchdog_thread_loses_this_requests_runtime_state` — reproduces gap #1 for the single-task watchdog thread.
- `test_concurrent_apply_calls_leak_read_protection_state_across_requests` — reproduces gap #2.
- `test_concurrent_apply_calls_with_internal_fanout_do_not_cross_contaminate` — a combined scenario: two `apply()` calls running concurrently, each *also* internally fanning out via its own `ThreadPoolExecutor` (different `target_ratio`/`kompress_model` per request), asserting every fanned-out worker only ever observes its own request's settings. This is the case that most directly matches production proxy traffic (concurrent requests, each potentially multi-message enough to trigger internal fan-out).

Verification:
- Targeted run — `tests/test_content_router_concurrency.py`, `tests/test_transforms/test_content_router.py`, `tests/test_openai_responses_read_protection.py`, `tests/test_observed_wire_shapes.py`: **103 passed**.
- Full suite: unchanged from the baseline reported in the PR description aside from these new tests (12021+4 passed); the previously-noted `test_proxy_health.py` / `test_read_maturation_handler_nobust.py` flakes are still pre-existing/order-dependent and reproduce identically without this change.

JerrettDavis (approved):
Re-reviewed the requested isolation follow-up. Each internal worker now enters its own copied Context, the watchdog path propagates the request context, and read-protection IDs/indices are part of the same per-request state. The new regressions cover fan-out, watchdog, concurrent requests, and combined concurrent fan-out. The clean main refresh preserves the fix. No remaining code blocker; the current matrix is still running against the shared rustls audit issue addressed by #3584.

JerrettDavis (changes requested):
The current-head lint run found two B007 violations in the new combined fan-out regression: `tag` is unpacked but unused in the `reqA_obs` and `reqB_obs` loops (`tests/test_content_router_concurrency.py:659` and `:670`). Please rename those two loop variables to `_tag` (or otherwise use them) and rerun the focused lint/test checks. This is the only branch-specific failure I found; the Context propagation fix itself remains sound.

JerrettDavis (approved):
Approved on head 5d6b24c60. I applied the minimal maintainer-side CI cleanup only: renamed the two unused tuple variables and ran Ruff's formatter on the same new test file. No contributor implementation logic changed. Fresh verification on the exact commit: 
uff check . passed, 
uff format --check . reported all 1,583 files formatted, 67 ContentRouter/concurrency tests passed, and git diff --check passed.

PR #3557, opened 2026-09-11
fix(wrap): stop blocking the launch on the Serena pre-index

### LINKED ISSUE #3436: [FEATURE] Do not block the wrap launch on the Serena pre-index
## Problem Statement

`headroom wrap claude` blocks the launch until `serena project index` finishes or hits the stall budget. In a directory that holds several repositories the index cannot finish inside any sensible budget, so every launch waits the full budget and then throws the work away. #3183 made the budget configurable, but the wait itself is still on the launch path.

Serena indexes lazily on demand, and its MCP server activates the project in the background while Claude Code is already usable. The synchronous pre-index is the only step in the wrap that makes the user wait for indexing.

Where the wait happens on `main`:

- the call site, run on every wrap: https://github.com/headroomlabs-ai/headroom/blob/e59cf1012a2b277689555833fd221a92279f777d/headroom/cli/wrap.py#L2395
- the blocking `communicate(timeout=...)`: https://github.com/headroomlabs-ai/headroom/blob/e59cf1012a2b277689555833fd221a92279f777d/headroom/cli/wrap.py#L2287

<details>
<summary>Measurements from one parent-of-repos directory</summary>

The directory holds all the cert-manager repositories with several worktree checkouts each: 6.5 GB, 34,639 Go files as seen by Serena.

| Run | Result |
|---|---|
| `serena project index` at the parent, unpruned | 7% (2,375 of 34,639 files) after 11 minutes, about 10 s per file; killed |
| `headroom wrap claude` at the parent, unpruned | `Serena: pre-index timed out (will index on demand)` after 300 s, every launch |
| After pruning to one checkout per repo (1,605 files), cold | 13 m 41 s once |
| After pruning, warm cache | 14.6 s per launch (4 s of that is `uvx` start-up, 3 s is the index itself) |

The per-file cost comes from `gopls` creating and discarding a workspace view for each file when the project root has no `go.mod`, so it is not something a larger budget fixes.

</details>

## Proposed Solution

Do not wait for the pre-index on the launch path. Two options, either is fine:

1. Start `serena project index` detached (the child already gets its own process group) and launch the agent straight away. Print one line saying the index is warming in the background.
2. Only run the pre-index when there is no `.serena/cache` directory yet, so a warm project pays nothing at launch and a cold one warms once.

Option 1 keeps the current behaviour of always warming the cache. Option 2 is smaller and removes the remaining ~15 s from every warm launch.

## Use Case

I launch Claude Code from a parent directory so it has context across all the cert-manager repositories. The Serena tools are still useful there, on demand. The pre-index is not: it either times out or, once the tree is pruned, costs ~15 s per launch for a cache that Serena would fill lazily anyway.

## Alternatives Considered

- `HEADROOM_SERENA_INDEX_TIMEOUT=60`: shortens the wasted wait but still wastes it.
- `--code-memory none`: removes Serena entirely, which loses the on-demand symbol tools.
- Pruning the tree with a root `.gitignore` for Serena: needed anyway, and what I did, but it only makes the wait short rather than removing it.

## Additional Context

- Follows on from #3093 and #3183. The comment on #3093 asking whether to drop `--project-from-cwd` and rely on `activate_project` points the same way: Serena copes without a pre-index.
- Happy to test a PR.

[Claude Fable 5.1]

Parideboy:
## Description

`headroom wrap` ran `serena project index` synchronously, so the agent did not start until the index finished or hit the stall budget (300s by default, configurable since #3183). As reported in #3436, that is a wait the user always pays and often gets nothing for: in a directory holding several repositories the index cannot finish inside any sensible budget, so every launch waited the whole budget out and then threw the partial work away. Even a small, already-warm project paid about 15s per launch for a cache Serena fills lazily anyway.

This moves the index off the launch path. It is now spawned and left to run alongside the agent, so the launch is immediate and the cache still warms. Serena's MCP server answers symbol queries from whatever is already cached and indexes the rest on demand, so nothing is lost by not waiting.

Closes #3436

## Type of Change

- [ ] Documentation update
- [x] Performance improvement

## Changes Made

- `_index_serena_project` now spawns `serena project index` and returns immediately instead of calling `communicate(timeout=...)`. It prints one line saying the index is warming in the background.
- The child's `stdout`/`stderr` are now `DEVNULL` rather than pipes. This is load-bearing, not tidying: nothing drains those pipes once the wrap stops waiting, and the indexer writes a progress line per file, so a full 64 KiB pipe buffer would block the child permanently. Serena already records its own failures in `.serena/logs/indexing.txt`.
- New `_stop_background_serena_index`, registered with `atexit`, stops the child when the wrap exits so a long index does not outlive the session that started it. It reuses the existing `_kill_serena_index_tree`, which already handles the `uvx` → `serena` grandchild on both platforms (#2938).
- `HEADROOM_SERENA_INDEX_TIMEOUT` keeps its name and its meaning — how long the launch may wait — with a new default of 0. A positive integer opts back into the previous blocking behaviour with that budget; an unusable value warns and falls back to the background, as it did before. `_resolve_serena_index_timeout_seconds` is renamed `_resolve_serena_index_wait_seconds` and the `_SERENA_INDEX_TIMEOUT = 300` constant is gone, since nothing defaults to it any more.
- Tests in `tests/test_cli/test_wrap_serena_boost.py` are updated for the new default and extended: the launch never waits, output is redirected rather than piped, the exit hook kills a running index and leaves a finished one alone, and the blocking path still works when the env var asks for it.

### Two things a reviewer may want to check

**Killing a half-finished index is cheap.** Upstream `ProjectCommands._index_project` calls `ls_mgr.save_all_caches()` every 30 seconds during the loop, and `request_document_symbols` skips files whose content hash it has already cached. So a killed index loses at most the last 30 seconds of work, and the next wrap resumes roughly where this one stopped rather than starting over. A hard kill of the wrap process itself (SIGKILL, an ungraceful terminal close) skips `atexit` and does leave the indexer running — that is the one gap.

**The indexer and Serena's MCP server now write `.serena/cache/**` concurrently**, where previously the pre-index always finished first. Upstream `solidlsp.util.cache.save_cache` writes the pickle in place with no atomic rename, so a torn write is possible. Serena tolerates this by design: `_load_document_symbols_cache` wraps the load in `try/except` with the comment *"cache can become corrupt, so just skip loading it"*, and `_save_document_symbols_cache` catches and logs rather than raising. The worst case is a cache rebuild, not a failure.

Issue #3436 offered a second option — skip the pre-index whenever a `.serena/cache` directory already exists. I did not take it: it leaves the first and most expensive launch blocking, and a cache that exists but is stale or partial would then never be refreshed. Backgrounding removes the wait in both cases.

## Testing

### Test Output

```text
$ python -m pytest tests/test_cli/test_wrap_serena_boost.py -q -p no:randomly
======================== 49 passed, 1 skipped in 0.55s ========================

$ python -m pytest tests/test_cli/ -q -p no:randomly --ignore=tests/test_cli/test_mcp.py
============ 5 failed, 723 passed, 2 skipped, 2 warnings in 39.52s ============
# The 5 failures and the test_mcp.py collection error reproduce identically on
# an unmodified upstream/main checkout on this machine: Windows has no
# signal.SIGHUP, the symlink tests need privileges, and the local pydantic-core
# is newer than pydantic expects. None involve the changed code.

$ python -m ruff check . && python -m ruff format --check .
All checks passed!
1577 files already formatted

Found 12 errors in 3 files (checked 532 source files)
# All 12 pre-exist in headroom/ccr/mcp_server.py, headroom/memory/mcp_server.py
# and headroom/release_version.py (local MCP SDK version drift). Zero in
# headroom/cli/wrap.py. Local mypy here is 2.3.1, not CI's 1.20.2.
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.13.11, uvx 0.11.28, serena-agent from PyPI, real `serena project index` child processes (no mocks). A scratch project holding 533 real `.py` files with a `.serena/project.yml` and a cold cache.
- Exact command / steps: imported the patched `headroom.cli.wrap` from this branch, `chdir` into the scratch project, then (1) called `_index_serena_project(verbose=True)` and timed the return, checked the child with `tasklist`, slept, and inspected `.serena/cache`; (2) repeated it and called `_stop_background_serena_index()` while the child was still running; (3) repeated with `HEADROOM_SERENA_INDEX_TIMEOUT=5` and with `HEADROOM_SERENA_INDEX_TIMEOUT=60s`.
- Observed result: (1) `_index_serena_project() returned in 0.01s`, printing `Serena: indexing project in the background (symbol tools work meanwhile)`; the child kept running afterwards and produced `document_symbols.pkl` (31,492,659 bytes) and `raw_document_symbols.pkl` (6,048,168 bytes) on its own. (2) `2s later: child pid 29672 alive=True (poll=None)` then `after the exit hook: child pid 29672 alive=False (poll=1)`, handle cleared; the cache was empty at that point, which matches Serena's 30-second flush cadence — a kill inside the first window loses that window. (3) with `=5` the call blocked and printed `Serena: pre-indexing project (waiting up to 5s)…` then `Serena: project pre-indexed (symbol cache warmed)`, returning in 2.46s with nothing left for the exit hook; with `=60s` it printed `Serena: ignoring HEADROOM_SERENA_INDEX_TIMEOUT='60s' (want a positive integer number of seconds) — indexing in the background instead` and returned in 0.01s.
- Not tested: a full `headroom wrap claude` end-to-end launch (the proof exercises the changed function directly against real Serena rather than starting an agent); POSIX behaviour — `start_new_session` and the `os.killpg` branch of `_kill_serena_index_tree` are covered only by unit tests on this Windows machine; the reporter's own multi-repository Go tree; and the concurrent cache-write case is reasoned from upstream Serena source, not reproduced.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — this is not behind a rollout channel. It is the Serena pre-index step of `headroom wrap`, which runs whenever `--code-memory serena` (the default) is active and the project already has `.serena/project.yml`.
- Minimum rollout channel: n/a (stable; ships with the next release).
- Stable/default behavior changed: yes. The default pre-index goes from blocking the launch for up to 300s to running in the background. No config or flag changes name or meaning; `HEADROOM_SERENA_INDEX_TIMEOUT` changes only its default.
- Kill switch / disable path: set `HEADROOM_SERENA_INDEX_TIMEOUT` to a positive integer to restore the previous blocking pre-index with that budget (`HEADROOM_SERENA_INDEX_TIMEOUT=300` reproduces today's default exactly). `--code-memory none` still disables Serena entirely.
- Qualification impact: none beyond the updated unit tests. The pre-index has always been best-effort and non-fatal, and it still is; nothing downstream reads its result.
- Rollback path: revert this commit. It touches one function, one new helper and one env-var resolver in `headroom/cli/wrap.py`, with no persisted state, no config migration and no on-disk format change — a revert restores the old behaviour immediately.

## Review Readiness

## Checklist

## Additional Notes

"I have made corresponding changes to the documentation" means the in-file documentation: `HEADROOM_SERENA_INDEX_TIMEOUT` is documented only in `headroom/cli/wrap.py` and its test module, both updated here, so there is no user-facing doc page to change. No screenshots — this is CLI timing behaviour.

@wallrj offered to test a PR in #3436; this is ready for that if you would like to try it against the cert-manager tree.

### DIFF
--- a/headroom/cli/wrap.py
+++ b/headroom/cli/wrap.py
@@ -19,6 +19,7 @@ Usage:
 
 from __future__ import annotations
 
+import atexit
 import errno
 import importlib.util
 import io
@@ -2142,43 +2143,49 @@ def _serena_project_skip_reason(root: Path) -> str | None:
     return None
 
 
-#: Upper bound on the synchronous pre-index. The agent does not launch until
-#: this call returns, so the number is a stall budget, not just a safety net.
-_SERENA_INDEX_TIMEOUT = 300
+#: Env knob for the launch-blocking pre-index. Unset (the default) means the
+#: wrap does not wait at all: ``serena project index`` runs in the background
+#: while the agent starts (#3436). A positive integer restores the pre-#3436
+#: behaviour of blocking the launch for at most that many seconds.
 _SERENA_INDEX_TIMEOUT_ENV = "HEADROOM_SERENA_INDEX_TIMEOUT"
 
 
-def _resolve_serena_index_timeout_seconds() -> int:
-    """Resolve the Serena pre-index stall budget from env, else the default.
+def _resolve_serena_index_wait_seconds() -> int:
+    """Seconds the launch may block on the pre-index (0 = do not wait).
 
-    A wrap launched from a directory Serena has already claimed re-indexes the
-    whole tree on every run, and 300s of that is time the agent is not running
-    (#3093). The budget is therefore tunable per environment, which also keeps
-    it reachable from ``wrap ... -- agents`` sessions that take no flags.
+    The pre-index used to be synchronous, so this knob sized a stall budget
+    (#3093). On a directory holding several repositories no budget is big
+    enough — the index cannot finish, so every launch waited the whole budget
+    out and then threw the partial work away (#3436). The default is therefore
+    now 0: the index runs in the background and the agent starts immediately.
 
-    Unlike :func:`_resolve_wrap_proxy_timeout_seconds`, a bad value is not
-    fatal here: the pre-index is best-effort, so an unusable setting falls back
-    to the default rather than aborting a launch that would otherwise succeed.
-    It is reported unconditionally, because a knob that looks applied but is
-    not is the failure this issue is about.
+    Setting the variable to a positive integer opts back into a blocking
+    pre-index with that budget, which is the escape hatch for anyone who wants
+    a warm cache before the first prompt (and the way to get the old behaviour
+    back).
+
+    A bad value is not fatal: the pre-index is best-effort, so an unusable
+    setting falls back to the background default rather than aborting a launch
+    that would otherwise succeed. It is reported unconditionally, because a
+    knob that looks applied but is not is the failure #3093 was about.
     """
     raw = os.environ.get(_SERENA_INDEX_TIMEOUT_ENV, "").strip()
     if not raw:
-        return _SERENA_INDEX_TIMEOUT
+        return 0
 
-    timeout_seconds: int | None
+    wait_seconds: int | None
     try:
-        timeout_seconds = int(raw)
+        wait_seconds = int(raw)
     except ValueError:
-        timeout_seconds = None
-    if timeout_seconds is None or timeout_seconds <= 0:
+        wait_seconds = None
+    if wait_seconds is None or wait_seconds <= 0:
         click.echo(
             f"  Serena: ignoring {_SERENA_INDEX_TIMEOUT_ENV}={raw!r} "
             f"(want a positive integer number of seconds) "
-            f"— using {_SERENA_INDEX_TIMEOUT}s"
+            "— indexing in the background instead"
         )
-        return _SERENA_INDEX_TIMEOUT
-    return timeout_seconds
+        return 0
+    return wait_seconds
 
 
 def _kill_serena_index_tree(proc: subprocess.Popen) -> None:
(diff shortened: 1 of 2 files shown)


JerrettDavis (approved):
Re-reviewed the current head after its main update. The background preindexing lifecycle, cache handoff, and failure isolation remain sound, and the focused Serena boost suite passes locally (49 passed, 1 skipped). The current head is conflict-free and all attached checks are now complete with no failures. Approved.

PR #1926, opened 2026-07-09
feat(transforms): first-class C# support in CodeAwareCompressor

jecacs:
Refs #1664

## Description

First-class C# support in `CodeAwareCompressor` via the tree-sitter `csharp` grammar, at parity with Java/C++/Rust: `using` directives, namespace headers, and type/member signatures preserved verbatim; method/constructor/destructor/operator/local-function bodies compressed; malformed input passes through unchanged. **No new dependencies** — the grammar ships inside the already-pinned `tree-sitter-language-pack==0.13.0` (resolves only as `"csharp"`; `c_sharp`/`cs` raise `LookupError`). Spec and maintainer go-ahead in the issue.

Closes #1664

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- `CodeLanguage.CSHARP` + full `_LANG_CONFIGS` entry; `_LANGUAGE_PREFILTER` and `content_detector` patterns chosen to be C#-distinctive (so Java doesn't mis-tag).
- New data-driven `LangConfig` fields (pattern of #1334's `class_body_node_types`): `container_node_types` — block-scoped `namespace { }` routed through class compression so members compress without the wrapper being re-emitted verbatim; `opaque_node_types` — `#if`…`#endif` wrappers preserved verbatim without recursion (recursing + wrapper re-emit duplicated whole files, up to ~1.9x input on real repos); `#if` blocks wrapping only usings are emitted with the imports so they stay ahead of type declarations.
- Shared-path fixes surfaced by real C# repos, each guarded and covered by a fail-before test: keep an Allman `{` on its own line in class reconstruction (K&R path byte-for-byte unchanged; Allman Java now compresses instead of falling back); line-based child extraction no longer swallows the following line for nodes ending at column 0 (C# `#region`/`#endregion` span their trailing newline — the over-slice duplicated the next member's signature or the closing brace); uncaptured top-level nodes preceding the first captured node (license banners, `#region License`) are emitted first instead of relocated below the code (tree-sitter-c-sharp rejects top-level `#region` after a type declaration, so relocation forfeited compression for the whole file).
- `TestCSharpSupport` (8 tests) + a C# case in the parametrized member-container test; CHANGELOG entry.

## Testing

[...]

PR #3559, opened 2026-09-11
feat(transforms): first-class Elixir support in CodeAwareCompressor

### LINKED ISSUE #3495: [FEATURE] Elixir support in CodeAwareCompressor
## Problem Statement

`CodeAwareCompressor` covers 11 languages. Elixir is not one of them, and the
gap is invisible rather than graceful: `headroom/compression/detector.py`
already classifies `elixir` as code, and `_try_detect_code` then scores Elixir
source as **`python`** — Elixir's `def foo(...)` heads match Python's pattern
exactly. The file is handed to a parser that cannot read it, the output fails
the syntax gate, and the original is returned.

Measured on the routed path (`_try_compress_text`, default config, no language
hint) across three real repos — `phoenixframework/phoenix @1e6183e`,
`elixir-ecto/ecto @97437d5`, `phoenixframework/phoenix_live_view @441dc69`:

```text
562 .ex/.exs files      0 tokens saved      0.0%
```

Elixir/Phoenix agent sessions are exactly the workload Headroom is for — large
context modules, LiveViews, ExUnit files — and today they pay full price.

## Proposed Solution

Add Elixir to `CodeAwareCompressor` via the `elixir` tree-sitter grammar, and
fix the detector so Elixir stops being routed to Python.

**Elixir has no definition node types.** `def`, `defp`, `defmodule`
and `alias` are macros, so tree-sitter-elixir emits a generic `call` for every
one of them. The current data-driven `LangConfig` dispatches on `node.type`, so
it would either match nothing or match every function call in the file.

Proposal, in the pattern of #1334's `class_body_node_types` and #1926's
`container_node_types`:

- `LangConfig.call_node_type` plus four `call_*_targets` sets. A new
  `_effective_type(node, lang_config)` resolves such a node to a `_CALL_ROLE_*`
  sentinel by looking up the macro's target identifier. Those sentinels are what
  Elixir's `function_nodes`/`class_nodes`/`import_nodes`/`type_nodes` contain,
  so every existing dispatch site keeps working unchanged and **no other
  language's behavior moves**. Identity function for every other grammar.
- `body_open_token`/`body_close_token` generalise the existing same-line
  delimiter branch so `do`/`end` reuses the K&R brace path.
- `_get_definition_name` learns to read the defined name from the macro's first
  argument rather than from the macro identifier.

The role table is taken from the grammar's own classification 
`queries/highlights.scm` in `elixir-lang/tree-sitter-elixir` - so this is mainline implementation not invented by me.

## API surface

No public API change. `CodeLanguage.ELIXIR`, `ex`/`exs`/`iex` aliases, a
`_LANG_CONFIGS` entry, the `LangConfig` fields above, and Elixir patterns in
`_LANGUAGE_PREFILTER` / `content_detector`.

## Changes to existing behavior

`_try_detect_code` no longer tags Elixir
source as `python`. Two Elixir-exclusive markers (`defmodule`, `defp`,
`@moduledoc`, `|>` — none of which have a Python spelling) settle the language.
Guarded: 0 of headroom's own 1,478 `.py` files are affected.

## User stories

- **Golden path.** *Given* an agent reads `lib/my_app/accounts.ex`, *when* it
  passes through the proxy, *then* function bodies are elided and
  `defmodule` / `alias` / `@spec` / heads (including `when` guards) survive
  verbatim.
- **Edge case.** *Given* `for encoder <- [...] do defimpl ... end end`, *when*
  compressed, *then* the wrapper is emitted once. Declarations nested inside a
  macro wrapper stay verbatim rather than being captured and re-emitted with the
  wrapper — that duplication reached 1.9x input on `ecto/lib/ecto/json.ex`
  during development and is now a fail-before test.

## Failure modes and recovery

A grammar change that renames `call`/`do_block` degrades to pass-through, not
corruption: the existing validation gate discards any output that does not
re-parse, and the per-language syntax breaker pauses Elixir after repeated
discards. Every risky construct prefers the false negative — one-line
`def foo, do: expr`, `~H` sigils, `defstruct` and Ecto `schema` blocks are all
kept verbatim by design.

## Security considerations

No new dependency, no new network or filesystem surface, no new imports at
module load. The `elixir` grammar ships inside the already-pinned
`tree-sitter-language-pack==0.13.0` (`heex` and `erlang` resolve too; `eex`
raises `LookupError`). `uv.lock` untouched.

## Use Case

Elixir/Phoenix codebases: context modules, LiveViews, Ecto schemas, ExUnit
suites. Measured savings on the routed path with the change in place:

| repo | before | after |
|---|---|---|
| phoenix @1e6183e | 0 (0.0%) | 71,257 tokens (16.7%) |
| ecto @97437d5 | 0 (0.0%) | 173,394 tokens (26.9%) |
| phoenix_live_view @441dc69 | 0 (0.0%) | 135,722 tokens (22.7%) |

## Alternatives Considered

**Elixir's own `Code.string_to_quoted/2`, or Sourceror.** Better ASTs —
Sourceror especially, since it preserves comments and formatting that this
compressor reconstructs by hand. Both are BEAM-hosted, which is disqualifying
here: compression runs in-process at a P50 of ~2ms, and shelling out to an
`elixir` binary costs ~200ms cold, makes an Elixir/OTP install a hard runtime
dependency of a Python package, and has no path forward for the eventual
`headroom-core` Rust port. tree-sitter is the only option that is already a
dependency, runs in-process, and has a Rust crate.

The compiler AST is still the right tool for *verification*, and it is used that
way below.

## Additional Context

**Willing to contribute:** yes — implemented and verified on
`cortfritz:feat/elixir-code-compression`, awaiting a maintainer 👍 before I open
the PR, per CONTRIBUTING.

Verification already done on that branch:

- **562/562** compressed files re-parse under the real **Elixir 1.20 / OTP 28**
  `Code.string_to_quoted/2` — the compiler front end, not tree-sitter.
- **Zero content duplication** across all three repos; max inflation 28 chars
  (an omitted-line marker longer than the one line it replaced).
- Full suite vs `main` in the same env: `1436 failed / 9572 passed` on main,
  `1436 failed / 9592 passed` on the branch — identical failures and errors,
  +20 passed being exactly the 20 new tests. (The pre-existing failures are my
  sandbox: no network, optional extras absent.)
- All **30** recorded `code_aware_compressor` parity fixtures produce
  byte-identical output on branch and main — no existing language moved.
- `mypy headroom/` clean across 530 files; ruff check + format clean at the
  pinned 0.16.3; patch coverage 98.6%, project coverage of both changed modules
  held.

**Prior art:** #1926 (C# support) is the template I followed for scope, tests
and proof; #1664 is its issue.

**Known v1 limitations, deliberate, happy to file as follow-ups:** a body that
is a single `|>` pipeline is one AST statement and is kept whole, so
pipeline-heavy modules compress less; `@doc`/`@moduledoc` heredocs are preserved
rather than subject to `docstring_mode`, because in Elixir they are siblings of
the definition rather than its first statement; `.heex` templates are out of
scope.

cortfritz:
## Description

Elixir has no dedicated definition nodes: `def`, `defmodule` and `alias` are
macros, so tree-sitter-elixir emits a generic `call` for all of them.
Dispatching on `node.type` would either match nothing or match every function
call in the file — which is why the data-driven `LangConfig` could not express
Elixir as-is.

`LangConfig` gains `call_node_type` plus per-role target sets. `_effective_type`
resolves such a node to a `_CALL_ROLE_*` sentinel from the macro's target
identifier, and those sentinels are what Elixir's
`function_nodes`/`class_nodes`/`import_nodes`/`type_nodes` contain — so every
existing dispatch site keeps working unchanged and **no other language's
behavior moves**. The same-line body delimiters are generalised
(`body_open_token`/`body_close_token`) so `do`/`end` reuses the existing K&R
brace branch, and `_get_definition_name` reads the defined name from the macro's
first argument instead of the macro identifier.

**No new dependencies** — the `elixir` grammar ships inside the already-pinned
`tree-sitter-language-pack==0.13.0`.

Without this, Elixir source is scored as `python` by `_try_detect_code` (Elixir
`def foo(...)` matches Python's pattern), handed to a parser that cannot read
it, and compressed by **0%**.

Closes #3495

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `CodeLanguage.ELIXIR` + full `_LANG_CONFIGS` entry; `ex`/`exs`/`iex` aliases;
  `_LANGUAGE_PREFILTER` and `content_detector` patterns chosen to be
  Elixir-exclusive (bare `def` deliberately **not** matched, so Python cannot
  score on them).
- New `LangConfig` fields, in the pattern of #1334's `class_body_node_types` and
  #1926's `container_node_types`: `call_node_type` + `call_function_targets` /
  `call_class_targets` / `call_import_targets` / `call_type_targets`, and
  `body_open_token` / `body_close_token`.
- Role table taken from the grammar's own classification —
  `queries/highlights.scm` in `elixir-lang/tree-sitter-elixir`. The function
  role is that file's function-definition subset verbatim (9/9), and a test pins
  it so a grammar bump surfaces drift.
- `defmodule`/`defprotocol`/`defimpl` and ExUnit `describe` are containers whose
  members compress individually; `def`/`defp`/`defmacro`/`defguard`/
  `defdelegate` and `test`/`setup` bodies compress with heads (including `when`
  guards) preserved; `alias`/`import`/`require`/`use` are imports;
  `defstruct`/`defexception`/`defoverridable` and Ecto `schema`/`embedded_schema`
  stay verbatim.
- Non-definition top-level nodes are opaque. Elixir wraps declarations in
  ordinary macros; capturing an inner `defimpl` while the uncaptured `for`/`if`
  wrapper was re-emitted verbatim duplicated the whole file — observed at **1.9x
  input** on `ecto/lib/ecto/json.ex`, now a fail-before test.
- **Behavior change to call out:** `_try_detect_code` no longer scores Elixir as
  `python`. Two Elixir-exclusive markers settle the language, and Python's score
  is folded in rather than dropped because those matches *are* Elixir `def`
  lines.
- **From @JerrettDavis's review:** the container-head reconstruction now goes
  through `LangConfig.body_open_token` instead of a hard-coded `{`. Detail under
  *Real Behavior Proof → Observed result*.
- Docs: `code-compression.mdx` gains Elixir + an Elixir section;
  `limitations.mdx` language count 11 → 12.
- `TestElixirSupport` (23 tests) + an Elixir/Python routing test in
  `test_transforms_content_detection.py`.

## Testing

### Test Output

```text
$ pytest tests/ -q --continue-on-collection-errors
main @04cdf79a : 1456 failed, 9668 passed, 891 skipped, 191 errors in 232.47s
this branch    : 1456 failed, 9692 passed, 891 skipped, 191 errors in 199.88s
```

Identical failure and error counts; **+24 passed is exactly the 24 tests this PR
adds** (20 plus the four regressions from @JerrettDavis's review). The
pre-existing failures are my sandbox — no network, optional extras absent — and
reproduce byte-for-byte on `main`.

```text
$ pytest <10 code-compressor / detector / router suites> -q
255 passed

$ mypy headroom/
Success: no issues found in 532 source files
$ ruff check headroom/ tests/...   &&   ruff format --check <changed files>
All checks passed! / already formatted        (ruff pinned 0.16.4)
```

Fail-before, Elixir-scoped selection with both sources reverted to `main`:
10 failed / 4 passed (the 4 are negative guards that pass trivially when Elixir
is unsupported). On the branch: 23 passed. The four review regressions fail
before their fix commit and pass after.

Rebased onto `main` @04cdf79a and fully re-verified there, including #3531 which
moved the Rust `tree-sitter` pin to `=0.26.12`. None of the 32 intervening
commits touch the files this PR changes; the rebase was conflict-free.

## Real Behavior Proof

- Environment: macOS 15 (Darwin 25.6.0, arm64), Python 3.14 venv, the real
  `CodeAwareCompressor` with `CodeCompressorConfig(enable_ccr=False)`, no mocks,
  `tree-sitter-language-pack==0.13.0`. Independently validated with **Elixir
  1.20.0 / Erlang OTP 28** `Code.string_to_quoted/2` — the compiler front end,
  not tree-sitter. Separately, the branch is installed from my fork and running
  as a live local proxy serving real Claude Code traffic.

- Exact command / steps: cloned 8 public Elixir repos at depth 1 and ran
  every `.ex`/`.exs` file through the **routed** entry point
  (`_try_compress_text`, default config, **no** language hint — what a tool
  result actually hits at the proxy), plus 4 private Elixir/Phoenix codebases of
  our own. Every rewritten file was written to disk and fed to
  `Code.string_to_quoted/2`. A separate pass counted every non-generated line in
  input vs output to detect duplication, and replayed the 30 recorded
  `code_aware_compressor` parity fixtures through Python on both `main` and this
  branch.

- Observed result: 12 repos, 9,211 files, **18,506,612 → 14,607,815 tokens,
  3,898,797 saved (21.1%)**. `main` saves **0** on the same corpus.

  | repo | ★ | files | before | after | saved |
  |---|---|---|---|---|---|
  | `plausible/analytics` @e74d6fb | 29.0k | 1,256 | 1,794,081 | 1,445,298 | 19.4% |
  | `elixir-lang/elixir` @36351ae | 26.6k | 559 | 2,438,899 | 2,052,748 | 15.8% |
  | `firezone/firezone` @c15d759 | 9.1k | 1,515 | 2,438,521 | 1,701,101 | 30.2% |
  | `livebook-dev/livebook` @9714e05 | 5.9k | 443 | 843,756 | 625,858 | 25.8% |
  | `ash-project/ash` @ca57d67 | 2.5k | 1,019 | 1,766,510 | 1,386,042 | 21.5% |
  | `phoenixframework/phoenix` @1e6183e | | 204 | 427,478 | 356,221 | 16.7% |
  | `elixir-ecto/ecto` @97437d5 | | 126 | 643,798 | 470,404 | 26.9% |
  | `phoenix_live_view` @441dc69 | | 232 | 597,992 | 462,270 | 22.7% |
  | **public total** | | **5,354** | **10,951,035** | **8,499,942** | **22.4%** |
  | 4 private codebases (web app, distributed Phoenix, 9-app umbrella, LiveView Native mobile) | | 3,857 | 7,555,577 | 6,107,873 | 19.2% |

  **4,360 / 4,360** rewritten files re-parse under real Elixir 1.20 / OTP 28.
  **Zero** content duplication in any repo. Worst-case inflation +28 chars, an
  omitted-line marker longer than the line it replaced. Latency P50 0.6–3.1 ms,
  P99 ≤ 75 ms (worst is `elixir-lang/elixir`, whose 1,800+-line modules hit the
  pre-existing symbol-analysis cost shared with every language). All 30 parity
  fixtures byte-identical to `main`.

  Detection across the corpus, by token: 74.1% `source_code/elixir` (8,392
  files) → the AST path; 22.7% `text` (852 files, mostly generated seed data);
  2.5% `build`; 0.03% still `source_code/python` (18 short modules carrying only
  one Elixir-exclusive marker, 8,378 tokens).

  Savings track the **body-to-signature ratio**, not repo size — only bodies
  compress. The two lowest rows are the two most documentation-dense repos,
  measured as share of tokens inside preserved `@doc`/`@moduledoc` heredocs:
  `elixir` 16.2%, `phoenix` 12.2%, `ash` 6.6%, `livebook` 3.9%, `firezone` 2.8%.
  `ecto` is the honest counter-example — 14.2% docs but 26.9% saved, because its
  planner and changeset modules have unusually long bodies. Libraries compress
  least, applications most.

  One of the private codebases measures 5.6% across its whole tree but 20.2% on
  `lib/` + `test/`: 73.5% of its tokens are four generated seed `.exs` files of
  map and aircraft data, the largest 8.0M tokens alone. Those are data literals,
  the detector classifies them `text`, and they never reach the AST compressor.
  Correct — but it means whole-tree percentages are meaningless on repos that
  vendor generated data, so the table uses `lib/` + `test/` for that codebase.

  **In production:** across 632 completed turns from 8 distinct Claude Code
  sessions on contexts up to 463k tokens, the AST path fired 35 times at ratio
  0.09 (`router:code_aware:0.09`) with no syntax failures and no user-visible
  breakage.

  **On @JerrettDavis's review finding:** confirmed, and it is more severe than
  reported. For `defmodule(...)\n) do` it does fall back to uncompressed input
  as described. For `defimpl Foo,\n  for: Bar do` and a wrapped `describe`
  head, the mutilated output contains no ERROR or MISSING node, so
  `_verify_syntax` passes it and the compressor **ships it** — the `for:` target
  is silently gone, and `Code.string_to_quoted/2` rejects the result with
  `unexpected reserved word: end`. tree-sitter-elixir is more permissive than
  the compiler for this shape, so the never-serve-broken-code guarantee leaked.
  Fixed by routing the head check through `LangConfig.body_open_token`;
  punctuation delimiters keep the original start-of-line rule so brace languages
  are byte-for-byte unchanged (all 30 parity fixtures confirm), and a word
  delimiter must be the final whitespace-separated token so neither `weirdo` nor
  the atom `:do` qualifies. All three shapes now compress, stay shorter, and
  parse under the real compiler.

- Not tested: the proxy over HTTP as a *test* (it is running in production,
  but the measurements call the transform and its routed entry point directly);
  CCR retrieval round-trips (`enable_ccr=False` throughout); the Rust port in
  `headroom-core` — Elixir is Python-only, as C#/PHP/Perl/Bash already are, so
  no parity fixtures are affected; Windows and Linux (macOS only); `.heex`
  templates (the `heex` grammar ships in the same wheel but is out of scope).

## Runtime Rollout Safety

- Rollout-managed feature(s): none. This adds a language to an existing
  transform; `headroom/rollout.py` has no `code_aware` gate and this PR adds no
  rollout channel, flag or qualification.
- Minimum rollout channel: n/a — not rollout-managed.
- Stable/default behavior changed: yes, two, both scoped to Elixir input.
  (1) `.ex`/`.exs` content now takes the AST path instead of passing through
  unchanged. (2) `_try_detect_code` no longer labels Elixir source `python`;
  guarded by a test, and 0 of headroom's own 1,478 `.py` files change label. No
  other language's output moves — all 30 parity fixtures byte-identical.
- Kill switch / disable path: `HEADROOM_CODE_AWARE_ENABLED=0` disables
  code-aware compression wholesale, as today. Installing without the `code`
  extra leaves tree-sitter absent and the compressor inert. The existing
  per-language syntax breaker (`HEADROOM_CODE_SYNTAX_BREAKER`) auto-pauses
  Elixir after repeated invalid-output discards without touching other
  languages.
- Qualification impact: none. No new gates, no change to existing
  qualification criteria.
- Rollback path: revert the commits. No migration, no persisted state, no
  config-schema change, no wire-format change; `uv.lock` untouched.

## Review Readiness

## Checklist

N/A — terminal evidence above.

## Additional Notes

**On the Codecov patch-coverage report (26.38%, 53 lines).** That number is a CI
artifact, not a gap in the tests. No CI job installs the `code` extra — the
shards install `"${WHEEL}[dev]"`, and `dev` does not carry `tree-sitter` or
`tree-sitter-language-pack`. Every class in `test_code_compressor.py` gated on
`@pytest.mark.skipif(not TREE_SITTER_INSTALLED)` is therefore skipped in
coverage, which is seven classes including `TestCSharpSupport`,
`TestPhpSupport`, `TestRealASTRuns` and this PR's `TestElixirSupport`. Measured
locally with the extra installed, patch coverage is **70/71 executable added
lines (98.6%)** on `code_compressor.py` and **2/2** on `content_detector.py`;
the single miss is a `current is None` guard I do not believe is reachable from
a real parse. Project coverage of both changed modules holds (86% → 86%,
48% → 48%).

This is pre-existing and repo-wide rather than something this PR introduces, so
I have not changed CI here. Adding `tree-sitter`/`tree-sitter-language-pack` to
the `dev` extra would make those seven classes actually run — no new package,
both already pinned under `code` — but it is a dependency change and squarely a
maintainer call. Happy to do it here or in a separate PR, whichever you prefer.

**Dependency justification:** none added, none bumped; `uv.lock` untouched.

**Architecture:** malformed input passes through byte-identical; every risky
construct prefers the false negative (verbatim) over corruption; invalid
reassembly falls back via the existing validation gate; no new imports at module
load; P99 < 75 ms across all 12 proof repos.

**Known v1 limitations, deliberate, happy to file as follow-ups:**
- A body that is a single `|>` pipeline is one AST statement and is kept whole,
  so pipeline-heavy modules compress less.
- `@doc`/`@moduledoc` heredocs are preserved rather than subject to
  `docstring_mode` — in Elixir they are siblings of the definition, not its
  first statement, so the Python docstring path does not reach them.
- Declarations nested inside a `for`/`if` macro wrapper stay verbatim.

**Two incidental findings, neither blocking, happy to file separately:**
- The router's transform label is `code_aware:<ratio>` with no language, while
  `_try_compress_text` emits `code_aware:<lang>:<ratio>`. The language is
  dropped at the router, so live traffic cannot be attributed per language —
  which is why I can report that the AST path fired 35 times in production but
  not how often it was Elixir.
- `tree-sitter` resolves to `0.26.0` on the Python side under the current pin
  (`>=0.25.2,<0.27`) while earlier testing here was on 0.25.2. Both work; the
  numbers above are from 0.26.0.

### DIFF
--- a/docs/content/docs/code-compression.mdx
+++ b/docs/content/docs/code-compression.mdx
@@ -18,7 +18,23 @@ Naive truncation breaks code. Cutting a function in half leaves invalid syntax t
 | Tier | Languages | Support Level |
 |---|---|---|
 | Tier 1 | Python, JavaScript, TypeScript | Full AST analysis |
-| Tier 2 | Go, Rust, Java, C, C++, C#, PHP | Function body compression |
+| Tier 2 | Go, Rust, Java, C, C++, C#, PHP, Elixir | Function body compression |
+
+### Elixir
+
+Elixir has no dedicated definition nodes -- `def`, `defmodule` and `alias` are
+macros, so every one of them parses as a generic `call`. Headroom dispatches on
+the macro name instead of the node type, which means:
+
+- `defmodule`, `defprotocol`, `defimpl` and ExUnit `describe` blocks are
+  containers: their members compress individually.
+- `def`, `defp`, `defmacro`, `defguard`, `defdelegate`, and ExUnit
+  `test`/`setup` bodies are compressed; heads (including `when` guards) are kept.
+- `alias`, `import`, `require`, `use` are treated as imports and preserved.
+- `defstruct`, `defexception`, and Ecto `schema`/`embedded_schema` blocks are
+  preserved verbatim -- they declare the module's shape, not its behaviour.
+- One-line `def foo(x), do: expr` clauses have no block to compress and pass
+  through unchanged, as do `~H`/`~s` heredoc sigils, which are single statements.
 
 ## What Gets Preserved vs Compressed
 
--- a/docs/content/docs/limitations.mdx
+++ b/docs/content/docs/limitations.mdx
@@ -55,7 +55,7 @@ Headroom is designed to compress LLM context without losing accuracy. This page 
 
 ## Code Compression
 
-Headroom includes an AST-aware CodeCompressor (tree-sitter, 11 languages: Python, JavaScript, TypeScript, Go, Rust, Java, C, C++, Perl, C#, PHP) but it is gated behind safety protections that prevent it from firing in most real-world scenarios. This is intentional.
+Headroom includes an AST-aware CodeCompressor (tree-sitter, 12 languages: Python, JavaScript, TypeScript, Go, Rust, Java, C, C++, Perl, C#, PHP, Elixir) but it is gated behind safety protections that prevent it from firing in most real-world scenarios. This is intentional.
 
 **Why code mostly passes through:**
 
--- a/headroom/transforms/code_compressor.py
+++ b/headroom/transforms/code_compressor.py
@@ -650,6 +650,13 @@ _LANG_CONFIGS: dict[CodeLanguage, LangConfig] = {
         uses_colon_after_signature=False,
         detection_hints=("defmodule ", "defp ", "@moduledoc", "|>", " do\n"),
         call_node_type="call",
+        # The three target sets below partition the 15 definition macros the
+        # grammar itself recognises — `queries/highlights.scm` in
+        # elixir-lang/tree-sitter-elixir, `#any-of? @keyword "def" ...`. The
+        # function set is that file's own function-definition subset verbatim;
+        # the rest are split by whether the macro owns a member list
+        # (container) or declares shape (verbatim). Re-check against upstream
+        # when the pinned grammar version moves.
         call_function_targets=frozenset(
             {
                 "def",
@@ -677,7 +684,17 @@ _LANG_CONFIGS: dict[CodeLanguage, LangConfig] = {
         # Structure declarations, kept verbatim: they are the module's shape,
         # not its behaviour, and `schema`/`embedded_schema` field lists are
         # exactly what a reader of compressed Ecto code needs.
-        call_type_targets=frozenset({"defstruct", "defexception", "schema", "embedded_schema"}),
+        call_type_targets=frozenset(
+            {
+                "defstruct",
+                "defexception",
+                # Declares which generated functions may be replaced; carries no
+                # body of its own.
+                "defoverridable",
+                "schema",
+                "embedded_schema",
+            }
+        ),
         # Anything at the top level that is not itself a definition is opaque:
         # emitted verbatim, never recursed into. Elixir wraps declarations in
         # ordinary macros (`for encoder <- [...] do defimpl ... end end`), and
--- a/headroom/transforms/code_compressor.py
+++ b/headroom/transforms/code_compressor.py
@@ -16,7 +16,8 @@ Supported Languages (Tier 1):
 - Python, JavaScript, TypeScript
 
 Supported Languages (Tier 2):
-- Go, Rust, Java, C, C++
+- Go, Rust, Java, C, C++, C#, PHP
+- Elixir (macro-call dispatch; see LangConfig.call_node_type)
 - Bash (validated and preserved losslessly)
 
 Compression Strategy:
@@ -274,7 +275,8 @@ def _get_parser(language: str) -> Any:
         except Exception as e:
             raise ValueError(
                 f"Language '{language}' is not supported by tree-sitter. "
-                f"Supported: python, javascript, typescript, go, rust, java, c, cpp, csharp, php, bash. "
+                f"Supported: python, javascript, typescript, go, rust, java, c, cpp, "
+                f"csharp, php, elixir, bash. "
                 f"Error: {e}"
             ) from e
 
@@ -331,6 +333,7 @@ class CodeLanguage(Enum):
     PERL = "perl"
     CSHARP = "csharp"
     PHP = "php"
+    ELIXIR = "elixir"
     BASH = "bash"
     SHELL = "bash"  # Alias: tree-sitter-language-pack exposes the Bash grammar.
     UNKNOWN = "unknown"
@@ -361,6 +364,9 @@ _LANGUAGE_ALIASES: dict[str, CodeLanguage] = {
     "php5": CodeLanguage.PHP,
     "php7": CodeLanguage.PHP,
     "php8": CodeLanguage.PHP,
+    "ex": CodeLanguage.ELIXIR,
+    "exs": CodeLanguage.ELIXIR,
+    "iex": CodeLanguage.ELIXIR,
     "shell": CodeLanguage.BASH,
     "sh": CodeLanguage.BASH,
     "zsh": CodeLanguage.BASH,
@@ -401,6 +407,15 @@ class DocstringMode(Enum):
 # =========================================================================
 
 
+# Dispatch sentinels for macro-call languages (see LangConfig below). They
+# stand in for grammar node types that Elixir does not have, so they must not
+# collide with any real tree-sitter node type.
+_CALL_ROLE_FUNCTION = "__call_function__"
+_CALL_ROLE_CLASS = "__call_class__"
+_CALL_ROLE_IMPORT = "__call_import__"
+_CALL_ROLE_TYPE = "__call_type__"
+
+
 @dataclass(frozen=True)
 class LangConfig:
     """Data-driven configuration for a programming language.
@@ -436,6 +451,28 @@ class LangConfig:
     # descendants individually while the top-level pass re-emits the whole
     # wrapper verbatim, duplicating content. Prefer the false negative.
     opaque_node_types: frozenset[str] | None = None
+
+    # --- Macro-call languages (Elixir) -----------------------------------
+    #
+    # Elixir has no `function_definition` node: `def`, `defp`, `defmodule` and
+    # `alias` are macros, so every one of them parses as a generic `call` whose
+    # first child is the macro's `identifier`. Dispatching on `node.type` alone
+    # would either match nothing or match every function call in the file.
+    # When `call_node_type` is set, `_effective_type` resolves such a node to
+    # one of the `_CALL_ROLE_*` sentinels by looking up that identifier in the
+    # target sets below; the sentinels are what the language's
+    # `function_nodes`/`class_nodes`/`import_nodes`/`type_nodes` contain, so
+    # every existing dispatch site keeps working unchanged.
+    call_node_type: str | None = None
+    call_function_targets: frozenset[str] = frozenset()
+    call_class_targets: frozenset[str] = frozenset()
+    call_import_targets: frozenset[str] = frozenset()
+    call_type_targets: frozenset[str] = frozenset()
+
+    # Block delimiters for languages whose signature and body share a line.
+    # Brace languages use `{`/`}`; Elixir uses the `do`/`end` keywords.
+    body_open_token: str = "{"
+    body_close_token: str = "}"
 
 
 _LANG_CONFIGS: dict[CodeLanguage, LangConfig] = {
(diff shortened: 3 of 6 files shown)


JerrettDavis (changes requested):
The macro-role dispatch is conservative and the single-line Elixir fixtures are strong, but multiline container heads are reconstructed incorrectly. _compress_class_ast takes race_line = node_lines[sig_end] and only retains it when race_line.strip().startswith('{'). For an Elixir defmodule/defprotocol/describe written across lines, the body begins on a line such as ) do; that line is excluded from header_lines and fails the hard-coded brace check, so compression drops the closing head plus do and syntax validation falls back to the uncompressed input. Please make class-header body-token handling use LangConfig.body_open_token (while preserving existing brace languages) and add a compressible multiline defmodule(...) do regression that proves the result is shorter and syntactically valid.

cortfritz:
thanks @JerrettDavis !  will dig into that.

PR #3576, opened 2026-09-13
fix(mcp): surface a plugin-provided Serena that reconcile cannot see

### LINKED ISSUE #3570: [BUG] Serena can be registered twice: mcp reconcile cannot see a plugin-provided Serena MCP server
Headroom registers Serena by default (`--code-memory serena`) via `_serena_server_spec()` in `cli/wrap.py`, called through `_setup_coding_compressor` → `_setup_serena_mcp`. Independently, the `claude-plugins-official` marketplace ships a `serena` plugin. If a user has both, Claude Code runs **two** Serena MCP servers side by side, and Headroom never notices.

### Headroom only looks at one of the two possible locations

`cli/mcp.py:240` does `observed = registrar.get_server("serena")`, and `ClaudeRegistrar` reads only the `mcpServers` object (`mcp_registry/claude.py:200-241`). A plugin-provided server lives in the plugin's own `.mcp.json` instead. Nothing in the tree reads it — `git grep installed_plugins` and `git grep '.claude/plugins'` both return zero hits.

### The plugin's spec is the worse of the two

Which is what makes this more than cosmetic:

```
plugin:   uvx --from git+https://github.com/oraios/serena serena start-mcp-server
headroom: uvx --from serena-agent serena start-mcp-server \
            --project-from-cwd --context claude-code --open-web-dashboard False
```

The plugin variant has no `--project-from-cwd` (does not bind to the project), no `--context claude-code`, and no `--open-web-dashboard False` — so the browser popup that #1003 works around comes back. It also tracks `main` rather than a PyPI release, so it can break on any upstream commit.

`doctor` does not surface it either; it has no MCP checks at all (`git grep -n "mcp\|serena" headroom/cli/doctor.py` returns nothing).

### Reproduce

1. Install the `serena` plugin from the `claude-plugins-official` marketplace.
2. Run any Headroom Claude setup with the default `--code-memory serena`.
3. `claude mcp list | grep -i serena` returns two lines — one of them `plugin:serena:serena`.
4. `headroom mcp reconcile` reports the entry as consistent.

### Suggested fix

A warning in `headroom mcp reconcile` and/or `headroom mcp status` when a plugin-provided Serena is detected. Please **not** an automatic removal — Headroom should not mutate a third-party plugin. A pointer to `claude plugin disable serena@claude-plugins-official` would be enough.

Related but distinct: #3054 dealt with drift in the Headroom-managed entry and produced `mcp reconcile`. This one is about a second Serena that `reconcile` structurally cannot see.

---

*Found while cleaning up an installation; investigated together with Claude Code (Opus 5). Line references are from `main` at `f302a38e`.*

L4XB:
## Description

Closes #3570.

Headroom registers Serena itself (`--code-memory serena`), and the `claude-plugins-official` marketplace ships a `serena` plugin. A user with both runs two Serena MCP servers side by side, and Headroom never noticed: a plugin's server lives in the plugin's own `.mcp.json`, never in `mcpServers`, so `registrar.get_server("serena")` — and therefore `mcp reconcile` — reported the configuration as consistent.

As the issue notes, the plugin's spec is the worse of the two: no `--project-from-cwd`, no `--context claude-code`, no `--open-web-dashboard False`, and it tracks `main` rather than a PyPI release.

## Type of Change

## Changes Made

- `ClaudeRegistrar.get_plugin_servers(name)` reads `~/.claude/plugins/installed_plugins.json` and each install's own `.mcp.json`, returning a `PluginServer` (plugin key, spec, source path, `disable_command`).
- `mcp reconcile` and `mcp status` report what it finds. `status` had no MCP-server checks at all, which is why the second Serena was invisible there too.
- **Reporting only**, exactly as the issue asks: the plugin belongs to whoever installed it, so the output names the plugin, the file it came from, and the `claude plugin disable` command — and changes nothing. A missing or malformed registry degrades to "nothing found" rather than failing the command it was only annotating.

## Testing

- [x] Type checking passes (`mypy headroom`) — see the note below

### Test Output

```text
$ python -m pytest -q tests/test_mcp_registry/ tests/cli/
191 passed in 1.37s

$ python -m ruff check .
All checks passed!

$ python -m mypy headroom
Found 9 errors in 2 files (checked 531 source files)
```

`mypy` reports the same 9 errors on this branch and on its merge base `aa4739e`, all in `headroom/memory/backends/direct_mem0.py` and `headroom/integrations/langchain/chat_model.py`. This branch touches neither file and adds no error.

13 new cells: 9 in `tests/test_mcp_registry/test_claude_registrar.py` (including `test_plugin_provided_server_is_found`, which asserts `get_server("serena") is None` in the same breath — that is the defect), 4 in `tests/cli/test_mcp_plugin_servers.py` driving the real commands through `CliRunner`.

Each was checked against a mutant that breaks the thing it names:

| mutation | killed by |
| --- | --- |
| emit no warning | the reconcile + status cells |
| match any plugin rather than the named server | 3 cells, incl. `a_plugin_shipping_a_different_server_is_not_reported` |
| report each install scope separately | `one_plugin_installed_at_two_scopes_is_reported_once` |
| drop the `claude plugin disable` pointer | 2 cells |
| warn unconditionally | `reconcile_stays_quiet_without_a_plugin_serena` (the accept control) |
| let a malformed registry raise | 3 cells |
| remove the `status` call | `status_reports_a_plugin_provided_serena` |

Two cells assert the non-mutation directly: the plugin's `.mcp.json` and the registry are byte-identical after the command runs.

## Real Behavior Proof

- Environment: macOS 15 (Darwin 25.6.0), Python 3.11, this branch run from a venv against an isolated `HOME` holding one installed `serena@claude-plugins-official` plugin, so no real Claude or Headroom configuration was read or written.
- Exact command / steps: `HOME=<isolated home> headroom mcp reconcile` and `HOME=<isolated home> headroom mcp status`, each run once on the merge base `aa4739e` and once on this branch.
- Observed result: `reconcile` gains four warning lines naming the plugin, its `.mcp.json`, and the `claude plugin disable` command; `status`, which had no MCP-server checks at all, gains the same four lines under `Agent Config:`. With a `HOME` that has no such plugin, `diff` reports both commands byte-identical before and after.
- Not tested: I do not have the `serena` plugin installed for real, so the registry layout is taken from a real Claude install on this machine rather than from a real `serena` install; and I did not run the two servers side by side against a live Claude Code session, because the change reports the condition rather than acting on it.

The fixture is the registry shape a real install uses:

```text
$HOME/.claude/plugins/installed_plugins.json
  {"version": 2, "plugins": {"serena@claude-plugins-official": [
    {"scope": "user", "installPath": ".../cache/claude-plugins-official/serena/1.0.0", "version": "1.0.0"}]}}

$HOME/.claude/plugins/cache/claude-plugins-official/serena/1.0.0/.mcp.json
  {"serena": {"command": "uvx", "args": ["--from", "git+https://github.com/oraios/serena", "serena", "start-mcp-server"]}}
```

```text
════ mcp reconcile — BEFORE (aa4739e) ════
Serena reconciliation for Claude
  observed: absent
  recommendation: uvx --from serena-agent serena start-mcp-server --project-from-cwd --context claude-code --open-web-dashboard False

════ mcp reconcile — AFTER (this branch) ════
Serena reconciliation for Claude
  observed: absent
  recommendation: uvx --from serena-agent serena start-mcp-server --project-from-cwd --context claude-code --open-web-dashboard False
  warning: plugin serena@claude-plugins-official also provides a Serena MCP server
           from .../.claude/plugins/cache/claude-plugins-official/serena/1.0.0/.mcp.json
           Claude Code runs both; Headroom manages only its own entry.
           to run one: claude plugin disable serena@claude-plugins-official
```

On the layout point: `installed_plugins.json` has that `{"plugins": {"<plugin>@<marketplace>": [{"installPath": ...}]}}` shape, and the marketplace checkout carries `external_plugins/serena/.mcp.json` with exactly the spec the issue quotes. The code reads `<installPath>/.mcp.json`, which is the copy Claude actually runs; a plugin that ships no MCP server simply has no such file (verified against `gopls-lsp`, whose install directory holds only LICENSE and README). If your installs put the manifest elsewhere, say so and I will follow it.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — this is a CLI diagnostic in `headroom mcp reconcile` / `headroom mcp status`, it adds no runtime behavior to the proxy or the transform pipelines, so there is no registry entry in `headroom/rollout.py` to gate.
- Minimum rollout channel: n/a, not a rollout-managed feature; it is available on `stable` like the rest of the `mcp` CLI.
- Stable/default behavior changed: output only, and only when a plugin-provided server of that name exists; with no such plugin both commands are byte-identical to `main` (measured above), exit codes are unchanged, and the warning is not an error.
- Kill switch / disable path: none needed, because nothing is enabled, installed, disabled or written — the user's own `claude plugin disable <plugin>` is what the message points at, and that is theirs to run.
- Qualification impact: none beyond the 13 new cells — no new dependency, no new binary, no network or subprocess call; the code reads two JSON files under `~/.claude/plugins/` and tolerates both being absent or malformed.
- Rollback path: revert the commit; no state is created, so there is nothing to clean up afterwards.

## Review Readiness

## Checklist

## Additional Notes

Documentation is unticked because the change adds no user-facing concept or flag — it annotates two existing commands. Say the word if you want a line in the MCP docs describing the warning and I will add it.


JerrettDavis (changes requested):
The read-only diagnostic direction is right, but the detector currently reports configurations Claude will not actually launch and the printed remedy does not reliably clear the warning:

- `get_plugin_servers` never reads `settings.json` / `enabledPlugins`. `claude plugin disable <id>` leaves the install record in place and writes `enabledPlugins[id] = false`, so this implementation continues warning immediately after the user runs the command it recommends.
- It reads only the flat `.mcp.json` shape; plugin manifests using `{"mcpServers": {...}}` are silently missed.
- It treats project-scoped install records as globally active, which can warn in unrelated projects.

Please gate on enabled state, support both manifest shapes, and scope records to the active context (or deliberately limit this command to user-scope installs with tests). PR #3578 addresses the same issue with those cases covered, so coordinating around one implementation would also keep the backlog low-noise.

L4XB:
All three are fixed in bfbb989, and all three were real — I checked each against the Claude install on this machine rather than taking them on trust, and the `enabledPlugins` one in particular made the warning self-defeating.

**1. `enabledPlugins` — you are right, and this was the worst of the three.** `~/.claude/settings.json` on this machine holds

```json
"enabledPlugins": {
  "dx@ykdojo": true,
  "gopls-lsp@claude-plugins-official": true,
  "ui-ux-pro-max@ui-ux-pro-max-skill": true
}
```

so `claude plugin disable` flips the value and the install record stays exactly where it was — the detector would have gone on warning after the user did the one thing it told them to do. End to end against an isolated `HOME`, before and after writing what that command writes:

```
$ headroom mcp reconcile
Serena reconciliation for Claude
  observed: absent
  recommendation: uvx --from serena-agent serena start-mcp-server ...
  warning: plugin serena@claude-plugins-official also provides a Serena MCP server
           ...
           to run one: claude plugin disable serena@claude-plugins-official

# settings.json now has enabledPlugins["serena@claude-plugins-official"] = false

$ headroom mcp reconcile
Serena reconciliation for Claude
  observed: absent
  recommendation: uvx --from serena-agent serena start-mcp-server ...
```

One judgement call worth your eye: only an explicit `false` silences it. A plugin that is installed but has no `enabledPlugins` entry is still reported, because for a detector whose whole point is "you are running two of these and cannot see it", a false negative costs more than a false positive. Say the word if you would rather require an explicit `true`; it is a one-line change and `test_a_plugin_with_no_enabled_entry_is_still_reported` is where the decision lives.

**2. Manifest shapes — also right, and worse than "some".** Both shapes are in the marketplace already:

```
9  flat      {"<server>": {...}}       e.g. external_plugins/serena/.mcp.json
5  wrapped   {"mcpServers": {...}}     e.g. external_plugins/discord/.mcp.json
```

That is 14 manifests in `claude-plugins-official`, so the flat-only reader was missing over a third of them. `_plugin_manifest_servers` now takes `mcpServers` when it is an object and the manifest itself otherwise.

**3. Scope — taken as your second option: user-scope only, with a test.** A project-scope record is active in one project and `mcp reconcile` is told nothing about which project the user means, so guessing is worse than saying nothing. `test_a_project_scope_install_is_not_reported` pins it.

**Verification.** 198 passed (was 191); 7 new cells. Each was checked against a mutant that breaks the thing it names:

| mutation | result |
| --- | --- |
| ignore `enabledPlugins` entirely (your point 1) | **2 failed** / 196 passed |
| treat a missing entry as disabled | **3 failed** / 195 passed |
| flat manifests only (your point 2) | **1 failed** / 197 passed |
| wrapped manifests only | **8 failed** / 190 passed |
| accept every scope (your point 3) | **1 failed** / 197 passed |
| key the disable check on the wrong plugin | **1 failed** / 197 passed |
| *(restored)* | **198 passed** |

`ruff check`, `ruff format` and `mypy headroom` are clean — mypy reports the same 9 errors as the merge base, in `direct_mem0.py` and `chat_model.py`, neither of which this branch touches.

**On #3578:** no argument from me about which lands. You asked to coordinate around one implementation and that is entirely your call — if #3578 is the one you want, close this and I will not re-open the ground. I brought this one up to your bar because you asked for the changes; I would rather hand you a finished branch than a half one, whichever you end up taking.

JerrettDavis (changes requested):
The new commit fixes disabled-plugin handling and wrapped mcpServers, but project-scoped detection is still a blocking false negative. Real v2 registry entries use scope local with a projectPath; get_plugin_servers now drops every non-user record, even when headroom mcp status runs from that exact project, where Claude would launch the duplicate Serena. Please compare the resolved current working directory against projectPath with appropriate path normalization and report an active matching local/project install, while remaining quiet outside its project. Add positive in-project and negative out-of-project regressions.

L4XB:
Fixed in 00c22c5, and you are right that this was the wrong reading. "The command knows nothing about the project" was true of the *user's intent* but not of the *record* — a `local` entry names its project, and inside that tree Claude launches the server. Refusing to speak there is the false negative that matters, because that is the only place the duplicate actually runs.

A record is active when:

- its scope is `user` — everywhere, as before; or
- it carries a `projectPath` and the resolved working directory **is** that directory or lies **below** it.

Containment rather than a string prefix: `Path.cwd().resolve()` compared against `Path(projectPath).resolve()` with `project in here.parents`, so `/work/api-v2` is not inside `/work/api`. A project record with no usable `projectPath` cannot be placed and is still skipped — a detector that cannot place a record should not speak for it. `OSError` from either `resolve()` (a deleted cwd) degrades to "not active" rather than failing the command it was only annotating.

**Both regressions you asked for, plus the edges around them.** 203 passed, was 198; 6 new cells:

- `test_a_project_scope_install_is_reported_inside_its_project` — positive, cwd is the project
- `test_a_project_scope_install_is_reported_below_its_project` — positive, cwd is `project/src/deep`
- `test_a_project_scope_install_is_quiet_outside_its_project` — negative
- `test_a_sibling_directory_that_shares_a_prefix_is_not_the_project` — `/work/api-v2` vs `/work/api`
- `test_a_project_scope_install_without_a_path_is_not_reported`
- `test_a_disabled_project_scope_install_is_not_reported` — `enabledPlugins` still wins inside the project

Each checked against a mutant that breaks the thing it names:

| mutation | result |
| --- | --- |
| user scope only (this PR's previous behaviour) | **2 failed** / 201 passed |
| every scope active everywhere | **3 failed** / 200 passed |
| string prefix instead of path containment | **1 failed** / 202 passed |
| the project directory itself does not count | **1 failed** / 202 passed |
| report a record with no `projectPath` | **1 failed** / 202 passed |
| *(restored)* | **203 passed** |

`ruff check`, `ruff format` and `mypy headroom` are clean; mypy reports the same 9 pre-existing errors as the merge base, in two files this branch does not touch.

**One thing I could not measure:** none of the three plugins installed on this machine is project-scope, so `installed_plugins.json` here has only `scope: "user"` records and no `projectPath` key at all. The shape I coded against is the one you describe — `scope: "local"` plus `projectPath` — rather than one I observed. If the real field is spelled differently, or if Claude scopes it to the project root only and not to subdirectories, tell me which and I will follow it; the containment rule is one line and the cells are written around it.

JerrettDavis (changes requested):
The new commit correctly recognizes a projectPath inside that project, but enabled state is still resolved only from the user `~/.claude/settings.json`. Claude writes project-scope enable/disable state to `<project>/.claude/settings.json` and local-scope state to `<project>/.claude/settings.local.json`; those more-specific values can override the user value. As a result this head can still warn for a project-local `false`, or miss a project-local `true`, and the printed disable remedy may not clear the warning because `_disabled_plugins()` never reads the file the command changed.

Please resolve `enabledPlugins` for the active directory across user, project, and local settings with Claude's precedence, require the effective value Claude uses, and add both override directions (`user true` + project/local false, and user false/absent + project/local true). The updated #3578 head now covers these cases and may be useful for consolidating this duplicate effort.

L4XB:
Fixed in 53a1777, and you are right — reading only `~/.claude/settings.json` was wrong in both directions, and the disable remedy was the sharper half of it: `claude plugin disable` run inside a project writes the file `_disabled_plugins()` never opened, so the warning survived the command it printed.

**What changed.** `_disabled_plugins() -> set[str]` becomes `_enabled_plugins_for(project_path) -> dict[str, bool]`, resolved most-specific-last:

1. `~/.claude/settings.json`
2. `<project>/.claude/settings.json`
3. `<project>/.claude/settings.local.json`

and the working directory's own `.claude/` applied last of all. Within one directory `settings.local.json` beats `settings.json`. Only an explicit `false` still means off — a plugin absent from all three layers stays reported, unchanged from the previous round.

The project layers come from **the install record's own `projectPath`**, not from the working directory, and the resolution moved inside the per-record loop. That matters for the case the cwd-only reading gets wrong: run `headroom mcp status` from `project/src/deep` and the layer that has to be consulted is `project/.claude/`, which is nowhere near the cwd. `_install_project_path()` is factored out of `_install_is_active_here()` so both use the same resolved path.

Non-boolean values are dropped rather than coerced. `bool(0)` is `False`, and a disable nobody wrote is the wrong direction to fail in.

**Both override directions, plus the edges.** 212 passed, was 203; 9 new cells:

- `test_a_project_false_overrides_a_user_true` — user `true`, project `false`, quiet
- `test_a_local_false_overrides_a_user_true` — same via `settings.local.json`
- `test_a_project_true_overrides_a_user_false` — off globally, back on here, reported
- `test_a_local_true_overrides_a_user_false` — same via `settings.local.json`
- `test_local_settings_override_project_settings` — `settings.local.json` is the more specific of the two
- `test_a_project_disable_applies_below_the_project_root` — cwd is `project/src/deep`
- `test_a_project_setting_does_not_apply_outside_that_project` — a user-scope install stays reported elsewhere
- `test_a_cwd_disable_silences_a_user_scope_install`
- `test_a_non_boolean_enabled_value_is_ignored`

Each checked against a mutant that breaks the thing it names:

| mutation | result |
| --- | --- |
| user settings only (the reviewed head) | **7 failed** / 205 passed |
| no `settings.local.json` layer | **3 failed** / 209 passed |
| `settings.json` wins over `settings.local.json` | **1 failed** / 211 passed |
| no per-install project layer (cwd only) | **1 failed** / 211 passed |
| user settings win over the project layers | **4 failed** / 208 passed |
| coerce non-boolean enabled values | **1 failed** / 211 passed |
| ignore the install's project, use cwd for the layer | **1 failed** / 211 passed |
| *(restored)* | **212 passed** |

`ruff check` and `ruff format --check` are clean; `mypy headroom` reports the same 9 pre-existing errors as the merge base `aa4739e`, in `direct_mem0.py` and `chat_model.py`, neither of which this branch touches.

**One case I did not close.** A **user**-scope install has no `projectPath` to anchor the project layer to, so for those records only the working directory's own `.claude/` is read — run from `project/src/deep` and a `false` written at `project/.claude/settings.json` is missed. Closing it means walking up from the cwd to find the project root, and I did not want to guess at the bound: stopping at the first ancestor with a `.claude/` directory, at `$HOME`, or at a VCS root are three different behaviours and they disagree on real layouts. Tell me which one Claude actually uses and it is a small change.

**On #3578:** it is approved and it covers these cases. If that is the one you are taking, close this — I said last round I would not re-open the ground and I meant it. I brought this head up to what you asked for because you asked for it, not to argue for it.

Symbol references are to this branch; measurements are at `53a1777`, against merge base `aa4739e`.

L4XB:
Closing this in favour of #3578, and I think that is the right call rather than
a third round here.

You asked twice for the precedence fix and #3578 has it: `_enabled_plugins`
merges `user settings.json` -> `<project>/.claude/settings.json` ->
`<project>/.claude/settings.local.json` in that order, which is the effective
value Claude actually uses and the file `claude plugin disable` writes, so the
remedy the warning prints clears the warning. It has the two override
regressions in both directions, it reports each plugin once, and it carries the
`docs/content/docs/mcp.mdx` change and the `headroom-issue-3570.json` fixture
that this PR never had.

Rewriting the resolution here would land the same function twice and hand you a
conflict, which is worse for you than one of the two going away.

Two small things from this branch that are worth carrying over if they are not
already there, so nothing is lost with it:

- `PluginServer` is exported from `headroom/mcp_registry/__init__.py` here and
  added to `__all__`. If #3578 leaves the type reachable only as
  `headroom.mcp_registry.claude.PluginServer`, that export is a one-line add.
- The reconcile output in this branch names the plugin id alongside the server
  name, so `plugin:serena:serena` is identifiable in a list where two Serenas
  appear. Worth a look at whether #3578's message does the same.

Thanks for the three rounds of review; they were specific and each one was
right.

PR #3577, opened 2026-09-14
docs(proxy): document cache-mode memory behavior

### LINKED ISSUE #3573: [DOCS] --memory context injection is silently skipped in --mode cache
The README advertises cross-agent memory prominently — *"one shared store across Claude, Codex, Gemini and Grok, with automatic dedup"* (`README.md:51`), and lists `--memory` as a supported flag for Claude Code (`README.md:230`). What it does not say is that in **`--mode cache`, context injection never happens at all**.

`proxy/handlers/anthropic.py:2670`:

```python
elif is_cache_mode(self.config.mode):
    # Cache mode: skip injection entirely so the next-turn
    # prefix bytes remain byte-equal to this turn's bytes.
```

The behaviour itself is correct and the comment explains it well — keeping the prefix byte-equal is worth more than the memory block. This is purely a documentation gap, and an expensive one: a user who enables `--memory` on a cache-mode deployment gets a feature that reports as `ENABLED` in the startup banner and does nothing.

**Measured on a `--mode cache --backend anthropic --memory` deployment**, over 135 memory-active requests:

```
memory_project_resolved  : 135
"Memory: Injected tools" : 135
memory_context stage     : 137
memory_inject            :   0
memory_save              :   0
```

Both per-project databases stayed empty.

**Suggested fix:** one sentence in the README and in `docs/content/docs/proxy.mdx` next to `--memory`, e.g. *"Context injection is skipped in `--mode cache`, where prefix stability takes precedence. Use `--mode token` if you need memory recall."*

### Related, found while investigating — happy to split these out

**1. The lookup runs before the check that discards it.** `anthropic.py:2632` performs `search_and_format_context(...)`; the cache-mode branch at `:2670` is only reached afterwards. In cache mode the result is guaranteed to be thrown away. Measured 92 ms median / 143 ms p95 / 12.6 s total across 137 requests. Hoisting the `is_cache_mode` check above the search would remove the cost entirely.

**2. The memory tools are deferred by Headroom's own tool-search layer.** `memory_*` is not in `_TOOL_SEARCH_CORE_TOOLS` (`proxy/helpers.py:3099`), so `inject_tool_search_deferral` sets `defer_loading = True` on them. Since the only write path in the proxy is an explicit `memory_save` tool call by the model, and the schema is hidden behind tool search with nothing prompting the model to look for it, nothing is ever written. Together with the point above, `--memory` on a cache-mode deployment is inert in both directions.

**3. The container deployment's `/health` has no `config` block.** A task-preset proxy returns `config.memory` / `config.learn` / `config.code_graph`; the `persistent-docker` one returns `deployment` instead and no `config`. `_query_proxy_config` (`cli/wrap.py:3469`) therefore returns `None`, and `wrap` treats that as a feature mismatch — *"did not expose config; restarting with requested features"* — restarting the container mid-session. It also makes `--memory` and `--learn` unverifiable from outside the process.

*Investigated together with Claude Code (Opus 5); line references from `main` at `f302a38e`.*

axelray-dev:
## Description

Document the existing cache-mode behavior for users who enable persistent memory. This makes clear that cache mode prioritizes provider prefix stability over memory context injection.

Closes #3573

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Changes Made

- Add a cache-mode memory note to the root README compatibility guidance.
- Explain the cache-mode memory limitation beside the `--memory` option in the proxy documentation.

## Testing

### Test Output

```text
$ npm exec -- fumadocs-mdx
[MDX] generated files in 10.36819700000001ms
fumadocs-mdx: passed

$ git diff --check
git diff --check: passed

The file-aware documentation checks found the cache-mode memory note in both README.md and docs/content/docs/proxy.mdx.
```

## Real Behavior Proof

- Environment: Linux, Node.js v22.23.2, Headroom checkout based on upstream main commit `aa4739e54ad46d48dd829f730af359c4824c04c4`.
- Exact command / steps: Ran `npm exec -- fumadocs-mdx`, checked each documentation file for its cache-mode memory explanation, and ran `git diff --check`.
- Observed result: MDX generation completed successfully, both documentation files contain the cache-mode memory limitation, and the diff has no whitespace errors.
- Not tested: The Next.js production build did not finish during the local run; deployed documentation rendering and live proxy requests were not tested.

## Runtime Rollout Safety

- Rollout-managed feature(s): None; this change only updates documentation.
- Minimum rollout channel: Not applicable.
- Stable/default behavior changed: Documentation now states the existing cache-mode memory behavior.
- Kill switch / disable path: Not applicable; no runtime behavior changed.
- Qualification impact: None.
- Rollback path: Revert this documentation-only change.

## Review Readiness

## Checklist


JerrettDavis (approved):
Reviewed the current head against the cache-mode memory path and CLI mode precedence. The documentation accurately calls out that cache mode skips memory context injection and gives users the correct token-mode escape hatch. Docs validation is green and the branch is clean and mergeable.

chopratejas:
Thanks for documenting this — the behavior is real, but it's Anthropic-only and the text states it unconditionally, so I'd like it scoped before merging.

`grep -rn "skipped_cache_mode"` across the package returns exactly one hit: `headroom/proxy/handlers/anthropic.py:2676`. Checking all four injection sites:

| route | injection site | cache-mode skip? |
|---|---|---|
| Anthropic `/v1/messages` | `anthropic.py:2630-2695` | **yes** (`elif is_cache_mode(...)` -> `decision="skipped_cache_mode"`) |
| OpenAI `/v1/chat/completions` | `openai.py:4130-4188` | **no** — injects via `append_text_to_latest_user_chat_message` |
| OpenAI `/v1/responses` | `openai.py:5686-5780` | **no** |
| Gemini `generateContent` | `gemini.py:592-635` | **no** (`injected_live_zone_tail_gemini`) |

The OpenAI and Gemini paths inject into the live-zone tail specifically so they *don't* bust the prefix (`openai.py:4122-4127`, `memory_handler.py:978`).

So the sentence is wrong for three of four providers — and it lands at `README.md:255`, immediately after the paragraph about OpenAI-compatible and MCP-native clients, which is the most misleading possible placement for an Anthropic-only caveat. Same issue in the `--memory` row at `docs/content/docs/proxy.mdx:128`. As written, Codex / Cursor / Copilot / Gemini users will believe memory is off when it's actually on.

Scoping it to "on the Anthropic `/v1/messages` path" fixes it.

**One secondary thing while you're in there:** the new note's remedy ("use `--mode token`") reads as contradicted by `proxy.mdx:178-183`, which says the default `coding` profile's `proxy_mode="cache"` overrides `--mode`. The code actually lets the flag win (`cli/proxy.py:1265`: `mode or HEADROOM_MODE or cache`, and `agent_savings.py:100-104` uses `setdefault`), so `proxy.mdx:178` is the stale one — but landing this as-is puts two conflicting statements in the same document. Worth fixing both in one pass.

PR #3581, opened 2026-09-14
fix(sdk): emit the final SSE event when the stream has no trailing newline

dex0shubham:
## Description

The TypeScript SDK's `parseSSE()` (used by `HeadroomClient` streaming at `client.ts:106`/`:158`, and exported from the package root) drops the **final** event of a stream when that event arrives without a trailing newline.

The read loop accumulates bytes, splits on `\n`, and holds the last (possibly incomplete) segment back in `buffer` for the next chunk:

```ts
buffer += decoder.decode(value, { stream: true });
const lines = buffer.split("\n");
buffer = lines.pop()!;   // last segment held back
```

But when the read completes it `break`s immediately, and the `finally` releases the reader — so whatever is still in `buffer` is discarded. If the stream closes right after its last `data: {...}` line with no trailing `\n`, that event is silently lost. The streaming `TextDecoder` is also never flushed, so a multi-byte character split across the final chunk boundary can be lost too.

This restructures the loop so the terminating read flushes the decoder and processes every remaining line (including the last held-back one) before exiting. `[DONE]` termination, non-JSON `data:` skipping, and `reader.releaseLock()` in `finally` are unchanged.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- **`sdk/typescript/src/utils/stream.ts`** — `parseSSE`: on the final read (`done`), flush the decoder (`decoder.decode()` with no args) and stop popping the last line so every remaining line is processed; then `break`. A final unterminated `data:` line is now emitted instead of dropped.
- **`sdk/typescript/test/utils/stream.test.ts`** (new) — first test coverage for this module (there was none): the no-trailing-newline regression, newline-terminated multi-event, event split across chunk boundaries, `[DONE]` termination, non-JSON line skipping, and a multi-byte char split across chunks.

## Testing

- [x] Unit tests pass (`npm test` — vitest)
- [x] Type checking passes (`npm run typecheck` — tsc --noEmit)
- [x] Build passes (`npm run build` — tsup)

### Test Output

```text
$ cd sdk/typescript && npm test
 Test Files  19 passed | 5 skipped (24)
      Tests  300 passed | 33 skipped (333)

$ npm run typecheck   # tsc --noEmit
(clean, no errors)

$ npm run build       # tsup
(ESM + CJS + DTS built successfully)
```

## Real Behavior Proof

- Environment: Node v22.16.0, Windows 11; TypeScript SDK from this branch (`cd sdk/typescript && npm install`).
- Exact command / steps: `cd sdk/typescript && npm test`, `npm run typecheck`, `npm run build`. The new tests drive the real exported `parseSSE` against `Response` objects backed by a `ReadableStream` with controlled chunk/byte boundaries.
- Observed result: The regression test streams `['data: {"a":1}\n', 'data: {"b":2}']` (no trailing newline on the last line). Before the fix `parseSSE` yields only `[{a:1}]` — `data: {"b":2}` is popped into `buffer` and discarded when the loop exits. After the fix it yields `[{a:1},{b:2}]`. All 6 new tests pass; full suite 300 passed | 33 skipped; typecheck and build clean.
- Not tested: a live proxy stream. `parseSSE` is pure (input `Response` → parsed events) and fully covered by the unit tests over the byte-level behavior that mattered.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. Client-side TypeScript SDK parser fix; no server/runtime feature flag or rollout channel involved.
- Stable/default behavior changed: Only the buggy path changes — a final event that was dropped is now delivered. Streams that already ended with a trailing newline parse identically to before (no regression).
- Kill switch / disable path: N/A — correctness fix, not a gated feature.
- Qualification impact: None — additive coverage; the fix strictly delivers events that were previously lost.
- Rollback path: Revert this PR; no persisted state, migrations, or config to unwind.

## Review Readiness

## Checklist

N/A

## Additional Notes

- Scope kept minimal: only `parseSSE`'s loop termination changed, plus a new test file. `collectStream` and the callers are untouched.
- CHANGELOG left untouched — the `no-manual-changelog` CI guard owns it; the Conventional Commit title drives the entry.


JerrettDavis (approved):
Reviewed the current post-update head. The decoder flush and terminal-buffer handling correctly preserve the final unterminated SSE event without changing normal newline-delimited or [DONE] behavior, and the focused tests cover chunk and multibyte boundaries. The branch is current, cleanly mergeable, and all applicable checks are green.

PR #3582, opened 2026-09-14
fix(transforms): classify grep -A/-B/-C context lines as search results

### LINKED ISSUE #3580: [BUG] grep -A/-B/-C context lines detected as PLAIN_TEXT — source code routed to Kompress and silently corrupted
## Description

`grep -A/-B/-C` **context** lines are classified as `PLAIN_TEXT` and routed to the Kompress prose compressor, which drops words by design. When those context lines contain source code, the code is silently corrupted — operators, keywords and whole statements are removed — and the result is presented to the model as the file's real contents.

The trigger is a single character: the separator GNU grep uses for context lines. Match lines use `:` and are detected as `search` (safe). Context lines produced by `-A`/`-B`/`-C` use `-`, and that alone flips detection to `text`.

This matters for coding agents specifically: `grep -rn <symbol> -A 30 <file>` is an extremely common way to read code, and the corruption is invisible — no error, no truncation notice, and the compressed text stays syntactically plausible.

Related but not the same: #3423 (closed) was also Kompress corrupting code, but its root cause was a missing BASH entry in `CodeLanguage`. This one is detection/routing, and the payload is Python.

## To Reproduce

1. `uv tool install --python 3.13 "headroom-ai[all]"` (v0.37.0)
2. Run the script below.

## Code Sample

```python
from headroom.transforms.content_router import _detect_content

SRC = [
    'env = payload.get("env")',
    'env_map = dict(env) if isinstance(env, dict) else {}',
    'previous = {name: env_map.get(name) for name in values}',
    'payload["env"] = env_map',
    'path.write_text(json.dumps(payload, indent=2), encoding="utf-8")',
]
F = "./headroom/providers/claude/install.py"

def build(sep):
    # 14 repeats -> ~9 KB, comfortably over the compression size floor
    return "\n".join(
        f"{F}:{40 + i}{sep}{line}"
        for _ in range(14)
        for i, line in enumerate(SRC)
    )

for label, sep in [("grep match line   'path:NN:code'", ":"),
                   ("grep -A context   'path-NN-code'", "-")]:
    print(f"{label} -> {_detect_content(build(sep)).content_type.value}")

print(f"plain source code (no prefix)    -> "
      f"{_detect_content(chr(10).join(SRC * 14)).content_type.value}")
```

## Expected Behavior

Source code in grep context lines is detected as `source_code` (or at least `search`, as the colon-separated match lines already are) and routed to the AST-aware compressor. Code is never handed to a compressor that drops words.

## Actual Behavior

```
grep match line   'path:NN:code' -> search
grep -A context   'path-NN-code' -> text
plain source code (no prefix)    -> source_code
```

The dash-separated variant is `text`, so it reaches Kompress and gets word-dropped.

## Real behavior proof

Observed live, not synthetically: a Claude Code CLI session routed through `headroom proxy` on `127.0.0.1:8787` (durable `install apply --scope provider`, backend `anthropic`). The command was:

```
grep -rn "def apply_provider_scope" -A 30 headroom/providers/claude/install.py
```

What the model received (note the missing `=`):

```
    env_map dict(env) if isinstance(env, dict) else {}
```

The same slice re-read raw with `sed -n '40,47p'`:

```
    env_map = dict(env) if isinstance(env, dict) else {}
```

Across that one payload, the compressed copy dropped `=` from an assignment, dropped `for` from a dict comprehension, dropped `+ "\n"` from a `write_text` call, and removed the line `payload["env"] = env_map` entirely. The trailing marker read `[129 words compressed to 102 (from 36 source lines). Retrieve more: hash=...]`, i.e. the prose path had been taken.

Confirmed on the `sed`-based re-read that the file on disk was correct and only the compressed copy was wrong.

## Environment

- Headroom: 0.37.0 (PyPI, `uv tool install --python 3.13 "headroom-ai[all]"`)
- Python: 3.13
- OS: macOS 15 (Darwin 25.5.0, arm64)
- Client: Claude Code CLI 2.1.270, routed via `ANTHROPIC_BASE_URL=http://127.0.0.1:8787`
- Proxy: `headroom install apply --preset persistent-service --scope provider --providers manual --target claude`, mode `cache`, backend `anthropic`

## What I did not test

- Whether `main` already fixes this. The repro is against released 0.37.0. Detection delegates to the Rust `headroom._core` magika chain, which I could not run from a source checkout without building the extension, and a source diff cannot settle a model-driven classification.
- Other context-line producers (`ripgrep -A`, `git grep -A`), which use the same `-` convention and would likely behave the same way.
- Whether languages other than Python are affected — presumably any, since the misroute happens before language selection.

lorenzozanee:
## Description

Closes #3580

`grep -A`/`-B`/`-C` context lines (`path-NN-content`) were classified as `PLAIN_TEXT` and routed to the Kompress prose compressor, which drops words by design. When those lines contained source code, operators, keywords, and whole statements were silently removed before the content reached the model. The detector only recognized the `path:NN:content` match-line shape, so the single-character separator difference flipped context lines into the lossy path.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/content_detector.py`: new `_is_grep_context_line` predicate accepting the dash-dash `path-NN-content` shape with the existing `<`/`>`/`=` prefix exclusions plus a path-shape guard (prefix must contain `/` or `.`, which keeps dates like `2026-09-14` and dashed prose like `version-2-release` out). `_is_search_result_line` tries the unchanged colon branch first, then the context branch. Thresholds (two-line floor, 30% ratio, confidence, metadata) are untouched.
- `headroom/transforms/mixed_content.py`: the section splitter carves context lines into `SEARCH_RESULTS` sections via the same predicate, so mixed prose + grep output no longer strands code in `PLAIN_TEXT` sections.
- `docs/content/docs/text-and-logs.mdx`: one-row addition documenting the context-line detection signal.
- Regression tests for both shapes, the mixed `-A` block with `--` separators, dashed file names, and the date/prose/markup/single-line guards.

## Testing

### Test Output

```text
$ python -m pytest tests/test_transforms/ tests/test_transforms_content_detection.py \
    tests/test_mixed_content_sections.py tests/test_mixed_content_scan_cache.py \
    tests/test_content_router_detection_dedup.py tests/test_cache_aligner_detector_only.py \
    tests/test_transforms/test_content_router.py -q
709 passed, 78 skipped

$ ruff check headroom/transforms/content_detector.py headroom/transforms/mixed_content.py \
    tests/test_transforms_content_detection.py tests/test_mixed_content_sections.py
All checks passed!
$ ruff format --check (same files)
4 files already formatted
$ git diff --check
CLEAN
```

`mypy` is unticked: the two touched files are clean, but a full `mypy headroom` run reports 10 pre-existing errors in untouched files (`kompress_compressor.py`, `proxy/helpers.py`, `image/trained_router.py`, …), so the box cannot honestly be ticked for the whole tree.

## Real Behavior Proof

- Environment: repo `main` @ `aa4739e`, Python 3.13 venv, native `headroom._core` built via `maturin develop` (the verdicts below come from the pure-Python detector, which is the same function the router's `PLAIN_TEXT` override invokes on both backends; the native magika chain returns `PlainText` for grep output by design).
- Exact command / steps:
  1. `grep -rn "def apply_provider_scope" -A 6 headroom/providers/claude/ --include='*.py'` → `detect_content_type` on the captured output.
  2. `rg -n "def apply_provider_scope" -A 6 headroom/providers/` → same check.
  3. The issue's reproducer payload rebuilt with the real dash-dash emission shape (70 lines).
- Observed result:
  1. `search` (confidence 1.0, 7/7 lines), one `SEARCH_RESULTS` splitter section — was `text` before the fix.
  2. `search` (confidence 0.94, 35/39 lines; remainder is `--` separators) — was `text` before the fix.
  3. `search` (70/70 lines) — was `text` before the fix.
- Not tested: end-to-end proxy run through the Kompress ML model (ML extras not installed in this environment); upstream CI (Rust fmt/clippy, parity harness, full suite).

## Runtime Rollout Safety

- Rollout-managed feature(s): none — plain detection-routing change, no feature flag.
- Minimum rollout channel: N/A, effective on upgrade.
- Stable/default behavior changed: yes — `grep`/`rg`/`git grep` `-A`/`-B`/`-C` context output with path prefixes now routes to the search fold instead of Kompress. Match-line, prose, log, diff, and code verdicts are unchanged (colon code path byte-identical, all thresholds kept).
- Kill switch / disable path: none specific to this predicate; rollback is a plain revert of a 5-file diff.
- Qualification impact: token-saving profiles shift for grep-context payloads (previously lossy-compressed as prose, now folded losslessly as search output); no API, config, or schema change.

## Review Readiness

## Checklist

N/A.

## Additional Notes

- The `path:NN-content` colon-dash mix from the issue's reproducer script is intentionally still `PLAIN_TEXT`: no real grep emission mixes separators within one line, and accepting it would newly claim changelog/log shorthand such as `notes.txt:3-updated`. A regression test pins the rejection with this rationale.
- Line-number-only single-file output (`30:code` / `31-code`, no path prefix) still routes as before — claiming it needs hunk-run structure analysis (bare `2026-09-14` date lines share the shape), which deserves its own reproducer and review rather than riding along here.
- The Rust regex oracle (`crates/headroom-core/.../content_detector.rs`) is untouched: it is off the production path (magika chain + Python re-check), consistent with how the #3221 detector guards shipped Python-only.


JerrettDavis (approved):
Reviewed the current head. The path-shape-guarded dash/dash predicate correctly captures grep/ripgrep context lines while preserving the existing thresholds and excluding date/prose/markup lookalikes; the mixed-content splitter uses the same predicate, and regressions cover realistic grouped output. The branch is current, cleanly mergeable, and all applicable checks are green.

PR #3558, opened 2026-09-11
test(litellm): cover standalone max_completion_tokens forwarding

axelray-dev:
## Description

Current main already forwards `max_completion_tokens` as a standard LiteLLM parameter. After refreshing this branch, the remaining change adds buffered and streaming regression coverage for requests that provide `max_completion_tokens` without `max_tokens`, ensuring the value stays top-level and does not leak into `extra_body`.

Addresses #3554

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Add buffered and streaming regression tests that verify the parameter is preserved and not duplicated in `extra_body`.

## Testing

- [x] Unit tests pass (`pytest`)
- [ ] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [ ] Manual testing performed

### Test Output

```text
$ .venv/bin/python -m pytest tests/test_litellm_openai_passthrough.py -q
6 passed in 2.57s

$ .venv/bin/ruff format --check headroom/backends/litellm.py tests/test_litellm_openai_passthrough.py
2 files already formatted

$ .venv/bin/ruff check headroom/backends/litellm.py tests/test_litellm_openai_passthrough.py
All checks passed!
```

## Real Behavior Proof

- Environment: Linux, Python 3.11.16, headroom checkout at commit `77136a57d41bbd720272d5f7d2f8ea1b93278bbe`, LiteLLM 1.88.1.
- Exact command / steps: Ran `LiteLLMBackend.send_openai_message` with a request containing `max_completion_tokens=64` and captured the arguments passed to the LiteLLM completion call.
- Observed result: The backend passed `max_completion_tokens=64` as a top-level LiteLLM argument and did not create `extra_body`.
- Not tested: Live Bedrock or Bifrost calls, and the separate Bedrock prompt-caching gate described in #3554.

## Review Readiness

- [x] I have performed a self-review
- [x] This PR is ready for human review

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
[...]

PR #3583, opened 2026-09-14
fix(litellm): enable Bedrock prompt caching and forward max_completion_tokens on OpenAI path

### LINKED ISSUE #3554: [BUG] Prompt caching gate uses wrong LiteLLM API
**Describe the bug**
Prompt caching (`cache_control`) is never injected for Bedrock requests, even for models that support it. Root cause: the gate in `backends/litellm.py` checks `litellm.get_supported_openai_params(model=..., custom_llm_provider="bedrock")` for `"cache_control"`. In litellm 1.96.2, that function never returns `cache_control` for any Bedrock model — including ones litellm's own `supports_prompt_caching()` correctly reports as `True`. The correct check is `litellm.utils.supports_prompt_caching(model=..., custom_llm_provider="bedrock")`.

Separately, on the OpenAI-compatible proxy path, `_OPENAI_STANDARD_PARAMS` does not include `max_completion_tokens` (only `max_tokens`). Gateways that translate `max_tokens` → `max_completion_tokens` (e.g. Bifrost) cause that field to fall into `extra_body`, which Bedrock rejects with `"extra_body: Extra inputs are not permitted"`.

**To Reproduce**

1. Run headroom proxy with `--backend bedrock`, any region.
2. Send a chat completion through the OpenAI-compatible proxy path with a model like `global.anthropic.claude-sonnet-5` or `us-gov.anthropic.claude-sonnet-5`, a system message over 1024 tokens, and (if going through a gateway that renames the param) `max_completion_tokens` instead of `max_tokens`.
3. Response either 500s (`extra_body: Extra inputs are not permitted`) or succeeds with `cache_creation_input_tokens`/`cache_read_input_tokens` absent from usage.

Confirmed directly:

```python
import litellm
litellm.get_supported_openai_params(model="global.anthropic.claude-sonnet-5", custom_llm_provider="bedrock")
# -> no "cache_control" in the list

from litellm.utils import supports_prompt_caching
supports_prompt_caching(model="global.anthropic.claude-sonnet-5", custom_llm_provider="bedrock")
# -> True
```

**Expected behavior**
Cache control should be injected whenever `litellm.utils.supports_prompt_caching()` is `True` for the resolved model, and `max_completion_tokens` should be treated the same as `max_tokens` in `_OPENAI_STANDARD_PARAMS`.

**Environment**

- headroom version: (fill in `headroom --version` or image tag)
- litellm version: 1.96.2
- Backend: bedrock
- Deployment: proxy mode, behind Bifrost gateway (OpenAI-compat provider path)

**Relevant log output**

```
litellm.BadRequestError: BedrockException - {"message":"extra_body: Extra inputs are not permitted"}
```

lorenzozanee:
## Description

On `--backend bedrock`, the OpenAI-compatible route (`send_openai_message` / `stream_openai_message`) forwarded the client's `messages` verbatim. Two gaps followed:

1. `max_completion_tokens` was missing from `_OPENAI_STANDARD_PARAMS`, so a gateway that renames `max_tokens` (e.g. Bifrost) saw the field fall into `extra_body`, which Bedrock rejects with `extra_body: Extra inputs are not permitted`.
2. A client that never sends `cache_control` (OpenAI-compat gateways strip it) never got a Bedrock `cachePoint`, so `cache_read_input_tokens` never appeared in usage even for models litellm reports as caching-capable.

This change forwards `max_completion_tokens` as a first-class LiteLLM parameter and marks the first system message with an ephemeral cache breakpoint when `litellm.utils.supports_prompt_caching()` is true for the *mapped* litellm model (`bedrock/...`, not the client alias).

Closes #3554

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `max_completion_tokens` to the LiteLLM backend's standard OpenAI parameter set so it is forwarded directly instead of landing in `extra_body` (both buffered and streaming paths share the allowlist).
- Gate Bedrock prompt caching on `litellm.utils.supports_prompt_caching(model=<mapped litellm id>)` at both OpenAI call sites; `get_supported_openai_params` never reports `cache_control` for Bedrock models so it cannot be the gate.
- Inject the breakpoint on the first system message (last non-empty text block for list content), respect client-placed `cache_control` anywhere, never mutate the caller's `body["messages"]`, and forward verbatim when the support check is undecidable.
- Add `tests/test_litellm_bedrock_openai_caching.py` (14 tests): allowlist behavior on both paths, injection for caching models via litellm's real registry, mapped-model gate check, non-supporting models / non-Bedrock providers unchanged, client markers respected, probe-failure fallback, and marker-to-`cachePoint` proof through litellm's real `AmazonConverseConfig` transform.

## Testing

### Test Output

```text
$ python3 -m pytest tests/test_litellm_bedrock_openai_caching.py tests/test_litellm_openai_passthrough.py tests/test_litellm_nonstream_cache_usage.py tests/test_litellm_caller_key.py tests/test_backends/ -q
43 passed in 1.31s

$ ruff check headroom/backends/litellm.py tests/test_litellm_bedrock_openai_caching.py
All checks passed!

$ ruff format --check headroom/backends/litellm.py tests/test_litellm_bedrock_openai_caching.py
2 files already formatted

$ git diff --check
(clean)
```

`mypy headroom/backends/litellm.py` was run but is blocked by a pre-existing environment failure unrelated to this change (numpy `.pyi` stub syntax error under the container's toolchain; identical on the unmodified base commit).

## Real Behavior Proof

- Environment: Linux, Python 3.14.6, checkout at `aa4739e5`, litellm 1.100.1.
- Exact command / steps: ran `LiteLLMBackend(provider="bedrock").send_openai_message` with a system+user message and `max_completion_tokens=64`, capturing the `acompletion` kwargs; ran litellm's real `AmazonConverseConfig().transform_request` on the resulting messages.
- Observed result: `max_completion_tokens=64` passed as a top-level LiteLLM argument with no `extra_body`; the system message carried `cache_control: {"type": "ephemeral"}` and transformed to `[{"text": ...}, {"cachePoint": {"type": "default"}}]`. Against the base commit the same probe produced `extra_body: {"max_completion_tokens": 64}` and no marker.
- Not tested: live AWS Bedrock / Bifrost round trip (requires deployment credentials and network).

## Runtime Rollout Safety

- Stable/default behavior changed: yes, for `--backend bedrock` OpenAI-path requests to caching-capable models (system prompt now carries one ephemeral breakpoint) and for any request sending `max_completion_tokens` (now top-level instead of `extra_body`). Non-Bedrock providers and non-supporting models are byte-identical to before.
- Kill switch / disable path: revert this commit
- Qualification impact: Bedrock prompt-cache billing applies (first write bills the prefix at the cache-write rate until read back)
- Rollback path: revert this commit

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (n/a — behavior fix, no documented contract changes)

N/A

## Additional Notes

- Open PRs #3558 (allowlist one-liner) and #3565 (opt-in caching behind `bedrock_openai_prompt_caching`) overlap this fix. This PR deliberately implements the issue's stated expectation directly (inject whenever the mapped model supports caching) in the smallest diff without rollout machinery; if reviewers prefer opt-in for billing reasons, the injection sites are the two marked call sites.
- Pre-existing, unrelated failures on the base commit (identical with and without this change): 3 `test_backend_bugs.py::TestOpenAIURLNormalization` tests and `tests/test_backend_nonstreaming_cache_metrics.py` collection (`headroom._core` native module not built in this environment).


JerrettDavis (changes requested):
This duplicates the Bedrock prompt-caching work in #3565 but enables cache writes unconditionally for every caching-capable mapped model. Prompt-cache writes have a distinct billing impact, so stable/default behavior must remain opt-in; #3565 is already reviewed and implements the capability gate plus edrock_openai_prompt_caching control. Please do not introduce a competing unconditional path. The max_completion_tokens forwarding fix can be separated if it is still missing after the accepted implementation lands.

PR #3586, opened 2026-09-15
fix(litellm): preserve Anthropic thinking blocks; fix OpenAI-in reasoning passthrough

chopratejas:
## Description
Preserve replay-safe Anthropic extended-thinking blocks through the LiteLLM backend and correctly pass OpenAI reasoning controls to Anthropic-family targets.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation-only change

## Changes Made
- Preserve signed `thinking` and valid `redacted_thinking` history for Anthropic-family routes while stripping it cross-vendor.
- Reconstruct leading thinking blocks in Anthropic responses, buffering streaming thinking until a signature is available.
- Forward `reasoning_effort` and `max_completion_tokens` as LiteLLM-mapped top-level parameters.
- Preserve reasoning metadata on rebuilt OpenAI responses.

## Testing
- [x] Added or extended automated tests for request, response, streaming, truncation, late-signature, and cross-vendor cases.
- [x] Full hosted backend matrix is green on the reviewed pre-update head.
- [x] Live Claude Sonnet 4.5 continuation and OpenAI-in reasoning behavior were exercised by the contributor.

## Real Behavior Proof
The contributor reports a successful two-turn Claude Sonnet 4.5 thinking continuation, an OpenAI-in request returning reasoning plus a signed block, and validation through LiteLLM's real Anthropic request transform that the signature reaches the provider payload.

## Runtime Rollout Safety
The behavior is gated to resolved Anthropic-family model IDs. Cross-vendor routes strip Anthropic thinking metadata. Unsigned Anthropic-shaped response blocks must remain filtered before rollout so replay cannot poison a subsequent tool continuation.

## Review Readiness

## Known Limitations
Flattened LiteLLM streaming cannot faithfully reconstruct redacted or multiple independently signed thinking blocks. Those cases require a future native `anthropic_messages` passthrough.

## Relationship
Independent of #3585; #3585 selects per-protocol native providers, while this PR implements translation fidelity.


JerrettDavis (changes requested):
The Anthropic request/response paths now correctly refuse replay-unsafe unsigned thinking, but the rebuilt OpenAI response bypasses that guard: `send_openai_message` serializes every entry in `c.message.thinking_blocks` with `_thinking_block_to_dict`. LiteLLM can surface unsigned reasoning blocks there, and a client that replays the returned message to a Claude-family route will send that invalid block back and can still get the 400 this PR is intended to prevent.

Please apply the same replay-safe filtering here (retain signed `thinking` and valid `redacted_thinking`, drop unsigned/malformed entries) and add a regression test where `thinking_blocks` contains unsigned reasoning. The rest of the thinking ordering/buffering and reasoning-parameter mapping looks thoughtfully covered.

PR #3592, opened 2026-09-15
fix(install): preserve host file ownership with rootless Podman

### LINKED ISSUE #3569: [BUG] Image v0.37.0 switches to `USER root`; with `--userns=keep-id` all newly created files in bind mounts are owned by a subordinate UID
## Summary

The `:latest` image for **v0.37.0** declares `USER root`, whereas v0.35.0 ran as `nonroot`.
Combined with the `--userns=keep-id` flag that `persistent-docker` correctly applies on
rootless Podman (added for #2804), every **newly created** file in the bind-mounted
`~/.headroom` ends up owned by a subordinate UID that the host user cannot read or write.

Existing files keep working, which makes the regression silent: namespace-root holds
`CAP_DAC_OVERRIDE` for mapped owners, so appends to pre-existing files succeed and preserve
ownership. Only files the proxy creates from scratch flip to the subordinate UID.
`headroom doctor` reports **0 failures** throughout.

## Environment

- Headroom CLI 0.37.0 (uv tool install)
- Image `ghcr.io/headroomlabs-ai/headroom:latest` → `org.opencontainers.image.version=v0.37.0`
- Rootless Podman on Arch-based Linux, `podman-docker` installed
- Profile: `persistent-docker`, `HEADROOM_CONTAINER_RUNTIME=podman` set explicitly
  (`/usr/bin/docker` is a shell-script shim here, so the symlink heuristic in
  `_container_runtime_is_podman()` cannot detect Podman on its own — see note below)

## What happens

`podman inspect` on the image:

```
org.opencontainers.image.version=v0.37.0
Config.User='root'
```

The previous container, created from the v0.35.0 image, reported `Config.User=nonroot`.

The container is started with `--userns=keep-id` (confirmed in `.Config.CreateCommand`),
which produces this mapping:

```
$ cat /proc/<pid>/uid_map
         0     100000       1000     # container root  -> host subuid 100000
      1000       1000          1     # host user       -> container 1000
      1001     101000      64536
```

Because the image process runs as container-`root`, it writes to the bind mount as host UID
100000. After a single restart:

```
$ find ~/.headroom ! -user "$(id -un)" -printf '%u:%g  %M  %p\n'
100000:100000  -rw-r--r--  ~/.headroom/.beacon_lock_8787
100000:100000  -rw-------  ~/.headroom/subscription_state.json
100000:100000  drwx------  ~/.headroom/cache/Microsoft

$ head -c 50 ~/.headroom/subscription_state.json
head: cannot open '~/.headroom/subscription_state.json' for reading: Permission denied
```

`subscription_state.json` is recreated at mode `600` under the subordinate UID, so the host
CLI — which runs as the regular user — can no longer read it. `headroom doctor` then reports
`budget: no budget configured` rather than surfacing the permission error.

## Steps to reproduce

1. Rootless Podman, `persistent-docker` profile, `HEADROOM_CONTAINER_RUNTIME=podman`
2. `headroom install restart --profile default` to pull the v0.37.0 image
3. `find ~/.headroom ! -user "$(id -un)"`

## Expected

Files created by the proxy in bind-mounted host directories stay owned by the invoking user.

## Suggested fix

Either of:

- Restore a non-root `USER` in the image so that, under `keep-id`, the process runs as
  container UID 1000 and therefore as the host user; or
- have `runtime.py` pass `--user 1000:1000` **alongside** `--userns=keep-id`. Note this is
  safe and is *not* a revert of #2804 — that bug was `--user` **without** `keep-id`, which
  selects an unmapped subordinate UID. With `keep-id` in place the same flag maps back to
  the host user.

## Side note on `_container_runtime_is_podman()`

Unrelated to the ownership bug, but in the same area: the heuristic resolves `docker` via
`os.path.realpath()` and compares the basename. On distributions where `podman-docker` ships
`/usr/bin/docker` as a **shell script** (`exec /usr/bin/podman "$@"`) rather than a symlink,
detection returns `False` and the container silently falls back to the Docker code path.
The `HEADROOM_CONTAINER_RUNTIME` override documented in the docstring works and is what I
use — flagging it only in case script shims are worth covering in the heuristic itself.

---

*Filed with assistance from Claude Code.*

qinghuanandejiangshi:
## Description

Rootless Podman versions that honor an image's `USER root` despite `--userns=keep-id` create bind-mounted files as a subordinate host UID. A newly created mode-600 state file then becomes unreadable to the host CLI. Reproduced with Podman 5.4.2 and host UID 1007.

Keep `--userns=keep-id` and also select the invoking user's UID/GID with `--user`. This is not the `--user`-without-keep-id mapping that caused #2804. The override is limited to Linux. Docker retains its existing user arguments, macOS Podman keeps its existing keep-id-only behavior, Windows remains unchanged, and the configured image is preserved.

Fixes #3569 (the file-ownership defect). The wrapper-script autodetection side note is not changed; its existing environment override remains necessary.

## Type of Change

- [x] Bug fix
- [x] Documentation update

## Changes Made

- Apply the existing host UID/GID selection to Linux Podman as well as Docker, after selecting Podman's keep-id namespace.
- Replace the old keep-id-only assertion with a 12-case platform/UID/GID/image matrix, including macOS and Windows regression coverage.
- Document the Podman environment override and the limits of recovery for files already owned by another UID.

## Testing

- [x] New tests fail before the fix and pass after it (4 cases).
- [x] Final focused run (installation subsystem plus the PATH-sensitive MCP case): 162 passed, 6 skipped.
- [x] `ruff check .` and `ruff format --check .`: passed.
- [x] Repository pre-commit checks, including mypy: passed.
- [x] `make ci-precheck-rust` with Rust 1.95.0: passed.
- [x] Manual real-behavior testing performed.

Full Python run: 12,148 passed, 595 skipped, one failure in `test_ordinary_install_does_not_adopt_serena` because the invocation did not put the venv CLI on PATH. The original checkout fails identically; adding `.venv/bin` to PATH makes it pass. The final focused run above includes this recheck and the platform guard added after the full run. This is a full run plus scoped revalidation, not a claim of one entirely green full-suite invocation.

`git push` completed with the installed pre-push hook: the full `make ci-precheck` passed, including a Rust 1.95.0 extension rebuild, the required Python regression set, and commitlint. No hooks were bypassed.

### Test Output

```text
# Full Python run (PATH-sensitive failure explained above):
1 failed, 12148 passed, 595 skipped, 6143 warnings in 931.68s
# Final installation subsystem and corrected-PATH MCP recheck:
162 passed, 6 skipped in 0.52s
# Installed pre-push hook's Python regression set:
175 passed, 4 skipped, 1 warning in 4.09s
✅ ci-precheck PASSED — safe to push.
```

## Real Behavior Proof

- Environment: Debian 13.4, Python 3.13.5, rootless Podman 5.4.2, ordinary user UID/GID 1007. This ran inside a disposable privileged Docker Desktop Linux container to permit nested user namespaces, with no host data directories mounted. Podman used VFS, disabled cgroups and no network. `podman info` reported rootless=true. No provider/model was used.

- Exact command / steps: Import the Debian rootfs with `USER root`, run the standalone command below, repeat with `--user 1007:1007`, then execute the real `build_runtime_command` output twice with root and nonroot fixture images.

The initial standalone reproduction was:

```sh
podman --storage-driver=vfs run --rm --network=none --cgroups=disabled \
  --userns=keep-id -v /home/tester/proof:/proof localhost/root-fixture \
  sh -c 'id; umask 077; echo written > /proof/state; cat /proc/self/uid_map'
stat -c '%u:%g %a %n' /home/tester/proof/state
cat /home/tester/proof/state
```

- Observed result: Before the fix, process UID 0, file owner `100000:100000`, mode 600, host read failed with Permission denied. Repeating with `--user 1007:1007` produced a host-readable file owned by `1007:1007`.

The end-to-end follow-up called the real `build_runtime_command(DeploymentManifest(...))` and executed its returned command through a docker→podman symlink. The fixture image's entrypoint created a private file/directory and appended to a pre-existing private file. The original source failed the ownership assertion. After the patch, two consecutive runs produced:

```text
uid=1007(tester) gid=1007(tester) groups=1007(tester)
0 probe-state 1007 1007 0o600
0 existing 1007 1007 0o600
0 probe-cache 1007 1007 0o700
uid=1007(tester) gid=1007(tester) groups=1007(tester)
1 probe-state 1007 1007 0o600
1 existing 1007 1007 0o600
1 probe-cache 1007 1007 0o700
PASS new files, private directories, append and restart
```

The same end-to-end probe passed with a custom image declaring `USER 1000:1000`, while the host remained UID 1007. Test files left with incorrect ownership by the red run were removed before the green run; this patch does not repair existing ownership.

- Not tested: the production Headroom image/proxy startup, Arch/SELinux, rootful Podman, newer Podman releases, or Windows runtime behavior. Python unit/lint checks ran on macOS with Python 3.12.14. Current Podman documentation describes different default USER behavior; the explicit process user is needed for the reproduced 5.4.2 behavior.

## Runtime Rollout Safety

- Rollout-managed feature(s): None; this is an installer command-generation bug fix.
- Minimum rollout channel: No feature rollout channel applies; use a build containing this patch.
- Stable/default behavior changed: Linux Podman receives an explicit invoking UID/GID alongside keep-id. Image names and configuration schemas are unchanged.
- Kill switch / disable path: Stop the persistent deployment with `headroom install stop --profile <profile>`; there is no new feature flag.
- Unsafe override required: None for this fix. The existing HEADROOM_CONTAINER_RUNTIME=podman override is still needed when a wrapper script cannot be detected.
- Qualification impact: Real rootless Podman 5.4.2 file-ownership proof and platform regressions passed; other runtime environments listed above remain untested.
- Rollback path: Revert this patch and recreate/restart the deployment. No data migration is involved; the original ownership problem can recur. Files already owned by a subordinate UID require separate recovery.

## Review Readiness

## Checklist

- [x] Style guidelines followed; hard-to-understand mapping behavior documented
- [x] Corresponding documentation updated
- [x] Regression tests added
- [x] No `CHANGELOG.md` edits


JerrettDavis (approved):
Reviewed the Linux/rootless-Podman command construction and its platform matrix. Combining --userns=keep-id with the explicit host UID/GID addresses the reproduced image-USER root behavior while retaining the existing macOS and Windows paths. The head is current, mergeable, and green.

PR #3593, opened 2026-09-15
fix: fix security issue in server.py

anupamme:
## Description
Add an explicit 64 KiB request-body limit to `POST /admin/runtime-env` to prevent
unbounded request-body buffering and JSON parsing.

## Type of Change
- [x] Security fix
- [x] Bug fix

## Changes Made
- Enforce the 64 KiB request-body limit against bytes actually consumed from the
  ASGI request stream (`request.stream()`), aborting with 413 as soon as the
  cumulative size exceeds the cap.
- Parse JSON only after the bounded body has been fully read; `request.json()`
  (which buffers unconditionally) is no longer called.
- Do not rely on the client-controlled `Content-Length` header as the
  enforcement boundary.
- Added endpoint-level regression tests for: a body at/under the limit with no
  `Content-Length`, an oversized chunked/streamed body, and a body whose
  declared `Content-Length` understates the actual bytes delivered.
- Removed an unrelated `min_tokens` → `min_ntoks` log-label rename that had been
  bundled into the original commit.

## Testing
- New: 64 KiB-or-smaller request without `Content-Length` → 200
- New: oversized streamed/chunked request → 413, JSON parser never invoked
- New: `Content-Length` smaller than actual streamed body → 413, JSON parser never invoked
- Existing `tests/test_runtime_env.py` suite (23/23 passing)

## Real Behavior Proof
`POST /admin/runtime-env` rejects bodies exceeding 64 KiB before any JSON
parsing occurs, regardless of what `Content-Length` claims or whether the
request is chunked.

## Runtime Rollout Safety
The change is limited to request-body handling for `/admin/runtime-env`
(loopback-only, same-origin-guarded admin endpoint). Requests within the
64 KiB limit behave exactly as before.

## Review Readiness
- [x] Tests cover the security boundary


JerrettDavis (changes requested):
The proposed guard does not enforce the stated 64 KiB security boundary. `Content-Length` is client-controlled: a request can advertise (for example) `Content-Length: 1` while the ASGI server delivers a larger body, and `request.json()` will still buffer and parse all of it. Conversely, valid HTTP/ASGI requests that use chunked transfer encoding (or otherwise omit the header) are rejected with 413 even when the JSON is tiny.

Please enforce the limit against the bytes actually consumed from the request stream (aborting once the cumulative size exceeds the cap), then decode/parse only that bounded buffer. Add endpoint-level regressions for: a body exactly at/under the limit without `Content-Length`, an oversized chunked body, and a body whose declared length is smaller than the bytes delivered. The limit should reject the latter two without invoking unbounded JSON parsing.

Also, the `min_tokens` → `min_ntoks` log-label rename is unrelated to this security fix and appears to make the label less clear; please drop it unless there is a separate demonstrated defect it addresses.

anupamme:
Thanks for the detailed review — agreed, the previous implementation incorrectly treated `Content-Length` as the security boundary.

I've reworked the fix (d10135b1) so the 64 KiB limit is enforced against bytes actually consumed from `request.stream()`, before any JSON parsing. This handles requests with no `Content-Length`, and prevents a mismatched/understated declared length from letting an oversized body through — `request.json()` is no longer called at all.

Added endpoint-level regressions covering:
- a body at/under 64 KiB with no `Content-Length` → 200
- an oversized streamed/chunked body → 413
- a body with `Content-Length: 1` but far more bytes actually delivered → 413

The two oversized cases assert the JSON parser is never invoked, not just that the response is 413.

Also removed the unrelated `min_tokens` → `min_ntoks` rename, and updated the PR description to fill in the required sections.

JerrettDavis (approved):
Reviewed d10135b1 against the earlier finding. The endpoint now counts the actual streamed bytes before JSON parsing, accepts a body exactly at the cap without Content-Length, and rejects oversized streamed or understated-length bodies without invoking the parser. The unrelated log-label change is gone. All 23 runtime-env tests pass locally; current hosted checks report no failures. No remaining blocking finding in this change.

PR #3599, opened 2026-09-15
fix(transforms): classify grep -A/-B/-C context lines as search results

### LINKED ISSUE #3580: [BUG] grep -A/-B/-C context lines detected as PLAIN_TEXT — source code routed to Kompress and silently corrupted

lorenzozanee:
## Description

`grep -A`/`-B`/`-C` context lines (both the real GNU `path-NN-content` shape and the reported `path:NN-content` shape) were classified as `PLAIN_TEXT`, so code inside them was routed to the word-dropping Kompress prose compressor and silently corrupted (`=` dropped from assignments, whole lines removed) behind a plausible `[N words compressed to M]` marker.

`content_detector._is_search_result_line` only accepted the `path:NN:content` match shape, and the mixed-content splitter had the same gap. This change recognizes both context-line shapes as `SEARCH_RESULTS` with path-shape guards (prefix must contain `/` or `.`, same `</>=` exclusions as the colon branch) so dates and dashed prose stay out.

Closes #3580

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Added `_is_grep_context_line` predicate plus a `_GREP_COLON_DASH_PATTERN` branch in `headroom/transforms/content_detector.py`, sharing a `_prefix_looks_like_path` guard so the three grep-line shapes cannot drift.
- Taught the mixed-content splitter (`headroom/transforms/mixed_content.py`) to carve the same shapes into `SEARCH_RESULTS` sections via the shared predicate.
- Added one detection-signal table row in `docs/content/docs/text-and-logs.mdx`.
- Added 6 regression tests (both separator shapes, mixed match/context blocks with `--` separators, date/dashed-prose/markup guard cases).
- Thresholds, confidences, and the Rust production chain are untouched; the SearchCompressor already parses `-` context lines and keeps them verbatim.

## Testing

### Test Output

```text
python -m pytest tests/test_transforms_content_detection.py tests/test_mixed_content_sections.py \
  tests/test_mixed_content_scan_cache.py tests/test_transforms_content_router.py \
  tests/test_transforms/test_content_router.py tests/test_content_router_detection_dedup.py \
  tests/test_transforms/test_detect_fallback_1123.py tests/test_search_compressor.py \
  tests/test_transforms_search_compressor.py tests/test_search_compressor_cjk.py \
  tests/test_bash_search_lossless_fold.py -q
527 passed in ~19s
ruff check: clean. ruff format --check: clean. git diff --check: clean.
```

`mypy` reports no errors in the two changed source files (10 pre-existing errors elsewhere, e.g. proxy/server.py), so the type-check box above stays unchecked.

## Real Behavior Proof

- Environment: Linux, submitter worktree `.venv` (native `headroom._core` built via maturin).
- Exact command / steps: fed real `grep -A` output through the content router and the SearchCompressor; the exact `if manifest.scope != ...` line from the issue was checked verbatim after compression.
- Observed result: match lines and context lines both route to `search`; the SearchCompressor keeps the issue's line byte-identical (before the fix, context lines routed to Kompress and were corrupted).
- Not tested: full `pytest tests/` suite, `cargo test --workspace`, parity suite, and the live native (Magika/ONNX) classification path — upstream CI still required.

## Runtime Rollout Safety

- Minimum rollout channel: N/A (not rollout-gated).
- Stable/default behavior changed: Yes — grep `-A`/`-B`/`-C` context lines now classify as search results instead of prose (the intended fix for #3580); all other inputs unchanged.
- Kill switch / disable path: Not applicable; no flag, no new dependency, no I/O.
- Qualification impact: None.

## Review Readiness

## Checklist

N/A.

## Additional Notes

- Prior attempt #3582 (closed unmerged) diagnosed the same area; this fix additionally covers the reported `path:NN-content` shape, which a dash-dash-only predicate would miss.
- Follow-up (not this PR): port the predicate to the Rust oracle `crates/headroom-core/src/transforms/content_detector.rs`; the locked production chain in `detection.rs` is intentionally untouched. No new parity fixture was added for the same reason.
- Out of scope, unchanged pre-existing behavior: pathless single-file `grep` lines (`30:def...`) and drive-letter paths (`C:\...`) still classify as before.


JerrettDavis (approved):
Reviewed both detection and mixed-content sectioning paths. The shared predicate keeps grep match/context lines classified consistently, preserves the existing splitter behavior, and includes useful false-positive guards and mixed-group regressions. The head is current, mergeable, and green.

PR #3603, opened 2026-09-15
fix(memory): align project routing with wrap headers

### LINKED ISSUE #3597: headroom wrap sets X-Headroom-Project but the CCR resolver reads x-headroom-project-id/x-headroom-cwd — the supported launcher doesn't satisfy its own resolver
## Summary

Two different, non-overlapping header namespaces are used for two related but distinct
features, and the one documented/supported launcher only sets one of them.

**Per-project savings tracking** (`rtk gain`'s per-project breakdown, `proxy/project_context.py:31-37`
+ `proxy/outcome.py:350` + `savings_tracker.py:663-665`) reads:
```
X-Headroom-Project
```
which is exactly what `headroom wrap` sets, via `cli/wrap.py:1480-1518`:
```python
_PROJECT_HEADER_NAME = "X-Headroom-Project"
def _project_name_from_cwd(): ... Path.cwd().name ...
def _apply_project_header_env(env):
    env["ANTHROPIC_CUSTOM_HEADERS"] = f"{_PROJECT_HEADER_NAME}: {project}"
```

**CCR workspace resolution** (`proxy/handlers/anthropic.py`, see also issues #3595/#3596 filed
alongside this one) reads a *different* pair of headers, per
`memory/storage_router.py:149-185`:
```
x-headroom-project-id
x-headroom-cwd
```

So a user who runs `headroom wrap claude` — the officially supported entry point — still
gets `CCR: workspace unresolved` on every request and an empty `projects` map from the
per-workspace compression tracker, because `wrap` never sets `x-headroom-project-id` or
`x-headroom-cwd`, only `X-Headroom-Project`.

Also worth noting: `headroom init hook ensure` (`cli/init.py:1058-1082`) and the
`tool_envs` written by `headroom init` manifests (`install/models.py`) set neither header
namespace at all — they only restart the proxy process if it's down (and set
`ANTHROPIC_BASE_URL`/`ENABLE_TOOL_SEARCH`). So the gap isn't specific to non-`wrap` setups
either; no first-party code path currently sets `x-headroom-cwd` / `x-headroom-project-id`.

## Suggested fix

Either:
- have `wrap` additionally set `X-Headroom-Cwd` (absolute cwd) and/or `X-Headroom-Project-Id`
  alongside `X-Headroom-Project`, or
- unify the two features to read the same header(s), since they both ultimately want "which
  project/workspace is this request for".

Env: headroom-ai 0.28.0 (pipx), macOS.

axelray-dev:
## Description

When `headroom wrap claude` runs from two directories named `api`, both requests currently send `X-Headroom-Project: api`. Using that label as a memory or CCR identity can make the two workspaces share history.

This keeps `X-Headroom-Project` as the human-readable savings label and adds `X-Headroom-Cwd`, derived from the canonical launch directory, for routing. The resolver now ignores the label when it is the only project signal.

Closes #3597

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Keep `X-Headroom-Project` for savings attribution only.
- Add a percent-encoded `X-Headroom-Cwd` header containing the canonical launch directory.
- Hash the decoded cwd for memory and CCR routing, while retaining the basename as the display label.
- Preserve user-supplied project and cwd headers without duplicating or overwriting them.
- Add regressions for same-basename directories, label-only fail-closed behavior, and an explicit user-supplied label collision.

## Testing

### Test Output

```text
216 passed, 1 warning in 4.80s
7 files already formatted
All checks passed!
git diff --check: clean
```

## Real Behavior Proof

- Environment: Ubuntu 24.04 VPS, Python 3.12.3, Rust extension built in the test environment.
- Exact command / steps: Wrapped two temporary directories with the same basename (`api`), using the same explicit `X-Headroom-Project` value, then resolved both memory and CCR workspaces.
- Observed result: Both requests retained the same display label but produced different cwd headers and different memory and CCR keys. A request containing only the display label returned the fail-closed result.
- Not tested: A live request to an external Claude provider.

## Runtime Rollout Safety

- Stable/default behavior changed: `headroom wrap claude` now sends a second `X-Headroom-Cwd` routing header; `X-Headroom-Project` remains a savings label.
- Kill switch / disable path: Remove the generated routing header or disable project-scoped memory/CCR using the existing configuration.
- Qualification impact: Requests with only the savings label remain unscoped and fail closed; requests with a cwd identity retain project-scoped memory and CCR.
- Rollback path: Revert commit `101b332`.

## Review Readiness

## Checklist


JerrettDavis (changes requested):
`X-Headroom-Project` is not a safe project-identity tier for memory or CCR. `_apply_project_header_env()` intentionally emits only `Path.cwd().name`, so two unrelated working trees with the same basename (for example `/work/acme/api` and `/work/other/api`) both send `api`. Hashing that display value does not restore the discarded path information; both requests resolve to the same key and can share stored memories/CCR history across projects.

Please keep the existing fail-closed behavior when only this analytics label is available, or introduce a separate header containing a stable, collision-resistant project identity derived from the full canonical working directory (while retaining `X-Headroom-Project` as the human-readable savings label). Add an end-to-end regression that launches/wraps two distinct directories with the same basename and proves their memory and CCR workspace keys differ. Also cover an explicit user-supplied `X-Headroom-Project` collision, since that header is overrideable by design.

axelray-dev:
Addressed the routing collision:

- `X-Headroom-Project` remains the human-readable savings label only.
- Wrapped Claude requests now also carry a canonical `X-Headroom-Cwd`; memory and CCR routing hash that full path.
- Added regressions for same-basename worktrees and an explicit user-supplied label collision.

Verification on the Ubuntu VPS: 216 focused tests passed, Ruff format/check passed, and the GitHub CI run passed, including the Docker-native end-to-end job.

JerrettDavis (changes requested):
The separate cwd header fixes the original same-basename issue, and all 154 focused wrapper/routing tests pass. There is one remaining isolation regression in `_identity_from_cwd`: it now percent-decodes every input, including literal `project_root_override` and system-prompt paths that were never URL-encoded.

Reproduced on 101b3325: both `/work/acme%2Fapi` and `/work/acme/api` resolve to `api-6eefc419fbace911` through either the CLI override or the `cwd:` prompt tier. Those are distinct filesystem paths, so their memories and CCR history can collide; existing percent-containing projects can also move to a different store after upgrade.

Please decode only at the encoded-header boundary and keep literal CLI/prompt paths unchanged. Add regressions proving that literal `%2F` (and `%20`) paths stay distinct from decoded paths, while a wrapper-generated encoded header still resolves to the same identity as its original literal cwd. This should preserve the new wrapper behavior without changing the existing path contract.

chopratejas:
Pushed `ef22855b` to address the round-2 isolation regression — the analysis in that review was correct.

**What changed:** `_identity_from_cwd` now takes a keyword-only `percent_encoded` flag and decodes **only** when it is set. The single caller that sets it is Tier 2, the `x-headroom-cwd` header — the one boundary that actually encodes. Tier 3 (`project_root_override`) and Tier 4 (system-prompt `cwd:`) pass their literal path through untouched, so a real `/work/acme%2Fapi` directory no longer collapses onto `/work/acme/api`.

**Regressions added** (`tests/test_memory_storage_router.py`):
- `test_resolver_literal_percent_path_distinct_via_cli_override` — literal `%2F` vs `/` stay distinct through the CLI override
- `test_resolver_literal_percent_path_distinct_via_system_prompt` — same through the `cwd:` tier
- `test_resolver_literal_percent_space_path_distinct_from_space` — `%20` vs a real space
- `test_resolver_encoded_cwd_header_matches_literal_cwd_identity` — the wrapper-encoded header still resolves to the same key as the literal cwd, i.e. the behaviour the encoding exists for

158 tests pass locally across `test_memory_storage_router.py`, `test_memory_handler_project_isolation.py`, `test_proxy_handler_helpers.py`, `test_cli/test_wrap_helpers.py`.

One pre-existing nit left alone as out of scope: `_apply_project_header_env` still early-returns on `if not project`, so neither header is set when the cwd basename is empty (filesystem root).

chopratejas:
Pushed `69e4c0b7`: main moved under this PR since the last push — #3606 merged and converted `_resolve_ccr_workspace` from a `@staticmethod` to an instance method (to thread `project_root_override` through `self.config`). This PR's own tests still called it the old static way (`AnthropicHandlerMixin._resolve_ccr_workspace(request, body)`), which is what's failing in `test (1)` / `test (4)` above. Updated all 13 call sites in `test_proxy_handler_helpers.py` and `test_cli/test_wrap_helpers.py` to `AnthropicHandlerMixin()._resolve_ccr_workspace(...)`. 158 tests pass locally across both files plus `test_memory_storage_router.py` / `test_memory_handler_project_isolation.py`.

JerrettDavis (approved):
Re-verified the conflict-resolved head against current main: mergeable and conflict-free. The merge retains main's instance-based handler usage; contributor test code is Ruff-clean; 162 focused tests passed locally; and the full refreshed CI/security matrix is green. Ready for maintainer merge review.

chopratejas:
Thanks — the core fix is right and I verified the header naming is consistent end to end: `wrap.py:2909` emits canonical `X-Headroom-Cwd`, `ProjectResolver._first_nonempty_header` sweeps case-insensitively, Claude Code forwards `ANTHROPIC_CUSTOM_HEADERS` verbatim, and the proxy strips `x-headroom-*` before upstream so it doesn't leak to providers. 162 tests pass on the branch.

Two things I'd like addressed before merging.

**1. Blocking: the unconditional `unquote` creates a cross-project collision.** `storage_router.py:237` decodes every `x-headroom-cwd` value, not just wrap-generated ones:

```
header x-headroom-cwd: /work/a%2Fb   main -> ('a-2Fb-f4d672a360015ef4')   PR -> ('b-ddaf707935d9b95f')
header x-headroom-cwd: /work/a/b     main -> ('b-ddaf707935d9b95f')       PR -> ('b-ddaf707935d9b95f')
```

Two distinct directories now share one memory DB. This PR adds three tests proving that *can't* happen on the CLI-override and system-prompt tiers (`test_resolver_literal_percent_path_distinct_via_*`) while introducing it on the tier it calls "the collision-resistant project identity."

`x-headroom-cwd` is a public, documented header (`cli/proxy.py:776,809,1537`) — the Kong plugin, LiteLLM, or any curl-based client can set a raw path, and they were never told to encode. A distinct header name (`x-headroom-cwd-enc`) or an explicit encoding marker would avoid this.

**2. Worth calling out in the rollout-safety section: `--memory-project-root` becomes dead for wrapped Claude Code sessions.** Tier 2 (`x-headroom-cwd`) outranks tier 3 (`project_root_override`) — `storage_router.py:173-183`. That precedence is pre-existing and documented, but until now the header was only present if someone set it by hand. An operator who set `--memory-project-root /shared/monorepo` to pool memory across a monorepo will now silently get per-launch-directory stores instead.

**Two smaller notes:**

- "The resolver now ignores the label when it is the only project signal" describes pre-existing behavior, not a change — `ProjectResolver.resolve` never read `x-headroom-project` on `main` (`PROJECT_HEADER` lives in `proxy/project_policy.py:11` and feeds savings attribution only). I verified label-only resolves to `None` identically on both. The four new label tests are fine to keep as characterization tests, but the PR overstates the delta.
- `wrap.py:5415-5416` still says "tag every request with the launch directory's name via X-Headroom-Project", which is now stale. Also `_project_name_from_cwd` uses `Path.cwd().name` (unresolved) while the new `_project_cwd_from_cwd` uses `.resolve()`, so under a symlinked launch dir the savings label and the memory display label disagree.

**One scope note for #3597:** this is Claude-Code-only. `_apply_project_header_env` is called from exactly one site (`wrap.py:5416`). Codex, OpenCode, Cursor, Copilot and Kimi wraps route project identity through `with_project_prefix` -> `/p/<name>`, which is savings attribution only — they still depend entirely on the system-prompt cwd tier. Worth saying so in the PR body so the issue isn't marked resolved for harnesses this doesn't cover.

Issue #3608, opened 2026-09-16
[BUG] Mixed-content routing can lose the parent HTML type


arx48:
## Description

Some complete HTML documents are routed as mixed content even though their top-level structure is clearly HTML.

After the document enters the mixed-content path, it is split into smaller sections. These sections are then detected independently, and some of them are classified as plain text or other types instead of HTML.

As a result, the original parent HTML structure is lost and HTMLExtractor does not get a proper chance to process the document.

## Observed behavior

The affected inputs were complete HTML/XHTML documents with strong structural indicators such as:

- `DOCTYPE`
- `<html>` root element
- normal HTML markup structure

However, because the content also contained text, scripts, JSON, or other embedded structures, it entered the mixed-content path.

After splitting, several child sections were no longer treated as HTML.

In one experiment, forcing the split sections to retain the parent HTML type reduced the content from about 157,827 tokens to 810 tokens, roughly a 99.5% reduction.

This suggests that HTMLExtractor itself works well. The problem appears to happen earlier, during content classification and mixed-content routing.

## Expected behavior

A clearly identifiable container format should take precedence over generic mixed-content detection.

For example, if the document has a valid HTML root structure, HTML should remain the dominant parent type even if it contains:

- JavaScript
- embedded JSON
- text
- tables
- styles

Mixed content should describe the internal composition of the document, not necessarily replace the top-level document type.

PR #3612, opened 2026-09-16
fix(router): stop embedded JSON from preempting the HTML extractor

### LINKED ISSUE #3609: [BUG] Embedded JSON compression can short-circuit parent HTML compression
## Description

The embedded JSON pre-pass can terminate routing too early when a larger structured document contains a small JSON fragment.

I observed this with HTML content that contains embedded JSON. The embedded JSON was successfully compressed, but the router returned immediately afterward, so the full document never reached the HTML-specific compressor.

## Observed behavior

In one case:

- The embedded JSON fragment was reduced from about 814 tokens to 277 tokens.
- The overall document reduction was only around 2%.
- When the embedded JSON path was disabled, the same content reached HTMLExtractor.
- HTMLExtractor reduced the document from about 24,758 tokens to 1,102 tokens, roughly a 95.5% reduction.

So a small local optimization prevented a much larger document-level optimization.

## Expected behavior

Embedded JSON compression should not automatically terminate processing of the entire parent document.

For content with a clear parent format such as HTML, I would expect one of the following behaviors:

- Detect and preserve the parent format first.
- Compress embedded JSON as a nested optimization, then continue processing the parent HTML.
- Only allow embedded JSON compression to short-circuit routing when there is no stronger parent-level strategy available.

## Environment
Headroom 0.37.0
Proxy mode
Token optimization mode
WSL / Linux

SulimanAbdulrazzaq:
## Description

The embedded-JSON pre-pass at the top of `ContentRouter._apply_strategy_to_content`
returns the moment it splices a routed span back in:

```python
routed = route_embedded_json(content, _dispatch_span, tok=_estimate_tokens)
if routed is not None:
    return routed, _estimate_tokens(routed), ["embedded_json"]
```

That return is unconditional, so it fires before the requested strategy runs. For a
document routed to `HTML`, any compressible JSON anywhere in the page means
`HTMLExtractor` is never reached — the router banks the span-level win and stops.
That is the wrong trade: the extractor drops navigation, scripts and styling from the
whole document, which is normally much larger than one embedded object.

This PR defers **only** for `HTML` in lossy mode, and defers the *call*, not just the
return:

- every other strategy keeps the existing early return, byte for byte;
- lossless-only mode (`HEADROOM_LOSSLESS` / `ContentRouterConfig(lossless=True)`) keeps
  the existing early return too: it stops at STAGE 0 and never reaches extraction;
- for `HTML`, `recursive_json.has_embedded_json()` — the span scan alone, with no
  dispatch — decides only whether a fallback is *available*. The route itself is held
  as a thunk and invoked at most once, after extraction;
- neither the STAGE 0 lossless-fold return nor anything else before the HTML dispatch
  can discard that fallback;
- if extraction yields nothing — extractor disabled or unavailable, extractor raising,
  or an empty/whitespace-only extraction (the registry adapter passes trafilatura's
  `""` through as a result) — the route runs then, and its result is returned exactly
  as the pre-pass returned it, with chain `["html", "embedded_json"]`;
- if the route also comes back empty, the block returns the STAGE 0 fold it skipped on
  the way in, so deferring can never cost it the byte-exact win `main` returns.

Because the skip now keys on a route being *available* rather than on its result, one
shape changes outcome: a folding HTML page whose span is routable but comes back no
smaller reaches the extractor, where the pre-pass returned the fold. That is the
document-level extraction this issue asks for, and it is the only way to skip the fold
without first paying for the result. It is pinned by
`test_extraction_wins_the_fold_when_a_span_is_routable_but_unprofitable`.

An operator-selected external compressor still runs before the built-in dispatch, as it
does for every other HTML block — and with the route deferred, such a request no longer
pays for an embedded result it was always going to discard.

Closes #3609

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/recursive_json.py`: the span scan `route_embedded_json` already
  ran is extracted as `_routable_spans()`, and `has_embedded_json()` exposes it. It
  answers "is there a span this would dispatch?" without running any compressor, so no
  `<<ccr:…>>` marker is registered and no TOIN row is written to find out.
- `headroom/transforms/content_router.py`:
  - the embedded-JSON pre-pass no longer short-circuits `CompressionStrategy.HTML` in
    lossy mode, and no longer *runs* there either: `embedded_fallback` holds the
    deferred route, not its result;
  - the STAGE 0 fold return is skipped while such a route is pending, so the HTML
    dispatch runs;
  - any extraction miss, including an empty or whitespace-only extraction, invokes the
    route then and returns it;
  - the lossless-then-lossy layer around the STAGE 0 return is extracted as
    `_lossless_fold_result()`, so the deferral path can return that same fold when the
    route comes back empty.
- `tests/test_transforms_content_router.py`: regression tests for the deferral, the
  STAGE 0 fold path, every extraction-miss shape with and without a fold, the
  no-win-from-the-route path, the widened shape above, lossless-only mode, and the
  unchanged early return for non-HTML strategies. The deferral tests assert the route
  is never called behind a successful extraction, not merely that its result is unused.
- `tests/test_recursive_json.py`: `has_embedded_json` agrees with `route_embedded_json`
  on exactly which blocks hold a routable span, and is availability, not benefit.
- Merged `main` (including #3556); see Additional Notes.

## Testing

- [ ] Unit tests pass (`pytest`) — full suite not run; the two changed test files,
      #3556's concurrency tests and the related selection below pass
- [ ] Type checking passes (`mypy headroom`) — ran `mypy` on the two changed modules only, see below

### Test Output

Environment: GitHub Codespace (Ubuntu 24.04, 4 cores), `uv sync --frozen --extra dev --extra html`
(uv selected CPython 3.13.15; maturin-built `headroom._core`; trafilatura installed and
`_get_html_extractor()` returns a live extractor). Every run below is on the same tree,
`git rev-parse HEAD^{tree}` = `bd7fc1f67a2aae248b708c25d7bd44cb5b55b3c4`.

```text
$ uv run pytest tests/test_transforms_content_router.py tests/test_recursive_json.py -q
98 passed in 4.45s

$ uv run pytest tests/test_content_router_concurrency.py -q
5 passed in 3.09s

$ uv run pytest tests -q -k "router or embedded or recursive_json or html or lossless" --ignore=tests/test_memory_eval.py
633 passed, 24 skipped, 12542 deselected, 1 warning in 29.38s
# tests/test_memory_eval.py is ignored: it imports litellm, which the dev extra does not install.

$ uv run ruff check .
All checks passed!

$ uv run ruff format --check headroom/transforms/content_router.py headroom/transforms/recursive_json.py tests/test_transforms_content_router.py tests/test_recursive_json.py
4 files already formatted

$ uv run mypy headroom/transforms/content_router.py headroom/transforms/recursive_json.py
```

Red/green, same tree, only the two source modules swapped:

```text
# this branch's tests against main's two modules
$ git checkout origin/main -- headroom/transforms/content_router.py headroom/transforms/recursive_json.py
$ uv run pytest tests/test_transforms_content_router.py -q -k "test_embedded_json or test_lossless_fold_does_not_preempt or test_lossless_only_mode_keeps or test_deferring_costs_nothing or test_plain_html_with_a_fold or test_extraction_wins_the_fold"
14 failed, 5 passed, 69 deselected in 0.94s
$ uv run pytest tests/test_recursive_json.py -q
ImportError: cannot import name 'has_embedded_json' ... (1 error in 0.24s)

# and against this branch's previous head ac4e95d7, which held the result eagerly
$ git checkout ac4e95d7 -- headroom/transforms/content_router.py headroom/transforms/recursive_json.py
$ uv run pytest tests/test_transforms_content_router.py -q -k "<same selection>"
3 failed, 16 passed, 69 deselected in 0.74s
```

The three that fail only against `ac4e95d7` are the ones this round adds:
`test_embedded_json_does_not_preempt_the_html_extractor` and
`test_lossless_fold_does_not_preempt_html_extraction_for_embedded_json` now assert the
route is never *called* behind a successful extraction, and
`test_extraction_wins_the_fold_when_a_span_is_routable_but_unprofitable` pins the
widened shape.

## Real Behavior Proof

- Environment: GitHub Codespace (Ubuntu 24.04, 4 cores), CPython 3.13.15 from `uv sync --frozen --extra dev --extra html`, maturin-built `headroom._core`, live trafilatura extractor; tree `bd7fc1f67a2aae248b708c25d7bd44cb5b55b3c4`, this branch merged with `main` at `964671d8`.
- Exact command / steps: `uv run python cost3612.py <variant>` builds one HTML page (120-record `<script type="application/ld+json">` island, 8 prose paragraphs, `nav`/`footer` boilerplate, a 100-newline run) and calls `ContentRouter()._apply_strategy_to_content(page, CompressionStrategy.HTML, context="")` once, with `recursive_json.route_embedded_json`, `ContentRouter._apply_strategy_to_content` and `ContentRouter._record_to_toin` wrapped in counting spies; nothing is stubbed. The same command is run with only `headroom/transforms/content_router.py` swapped between this head, `ac4e95d7` and `origin/main`.
- Observed result: input 4,179 tokens. This head returns 549 tokens `['html']` with route_calls=0, nested span dispatches=0, TOIN rows=1, 0.032 s. `ac4e95d7` returns the same 549 tokens `['html']` but with route_calls=1, nested span dispatches=1, TOIN rows=2 and 0.121 s — one nested JSON compression, and one TOIN row, for output it then discards. `origin/main` returns 2,134 tokens `['embedded_json']`, extractor never called. So the deferral removes the discarded work without changing what a successful HTML request returns.
- Not tested: the full test suite (the two changed test files, #3556's concurrency tests and the related selection above were run), `mypy headroom` beyond the two changed modules, a live proxy session, and CCR-marker registration under a marker-enabled config (the probe observed 0 `<<ccr:` markers in every variant, so the decisive counters are the route/dispatch/TOIN ones above).

| `content_router.py` | Output on that page | Chain | Embedded route called | Nested span dispatches | TOIN rows |
|---|---|---|---|---|---|
| `origin/main` | 2,134 tokens | `['embedded_json']` | 1 | 1 | 1 |
| Previous head `ac4e95d7` | **549 tokens** | `['html']` | 1 (discarded) | 1 (discarded) | 2 |
| This head | **549 tokens** | `['html']` | **0** | **0** | **1** |

## Runtime Rollout Safety

- Rollout-managed feature(s): none — this is inside the router's strategy dispatch, not behind a rollout channel.
- Stable/default behavior changed: yes, for `CompressionStrategy.HTML` content in lossy mode. Content with embedded JSON now reaches `HTMLExtractor` instead of returning the span-spliced text. Content whose span is routable but unprofitable *and* which also has a STAGE 0 lossless fold now reaches the extractor too, where it previously returned the fold; that is the one shape whose outcome the deferral widens, and it is pinned by a test. Non-HTML strategies, lossless-only mode, and HTML with no routable span are unchanged.
- Kill switch / disable path: `ContentRouterConfig.enable_html_extractor = False` turns the HTML branch off; the deferred route then runs in the fallback and returns what the pre-pass returned for that content.
- Qualification impact: none expected — non-HTML strategies and lossless-only mode take the identical code path they take today.
- Rollback path: revert these commits; the pre-pass returns unconditionally again.

## Review Readiness

## Checklist

- [ ] New and existing unit tests pass locally with my changes — the two changed test files,
      #3556's concurrency tests and the related selection pass; the full suite was not run

N/A.

## Additional Notes

- Merged `main` on 2026-09-18 to resolve the conflict with #3556, which moved the router's
  per-request state into a `ContextVar`. `content_router.py` merged cleanly: this change
  reads only `self.config.lossless`, `self._lossless_then_lossy` and locals, none of which
  #3556 moved. The only conflict was both PRs appending tests to
  `tests/test_transforms_content_router.py`; both sets are kept.
- `main` has moved on since that merge (`67eb910e`). The three commits since touch a
  different part of `content_router.py` (the beacon shape table from #3253) and do not
  conflict with this branch; say the word and I will merge `main` again.
- Docs unchanged: the routing behaviour described in the docs (embedded JSON is a
  pre-pass, HTML goes to the extractor) is what this makes true; nothing documented
  changes wording.
- Scope note: #3608 from the same reporter is about classification putting complete
  HTML documents on the mixed-content path. That is a different call site and is not
  touched here. Plain HTML without embedded JSON keeps today's STAGE 0 behaviour.
- Developed with AI assistance (Claude Code) and validated with the commands recorded
  above.


JerrettDavis (changes requested):
The main deferral case works, and the 72 available router/recursive-JSON tests pass locally (the HTML dependency suite is skipped in this environment). One intervening return still loses the held-back result:

**Preserve the embedded candidate through the lossless-first return.** After assigning `embedded_fallback`, the method runs `_lossless_first(content, strategy)`. When that finds any fold, the `_ll_label is not None` branch returns before both the HTML extractor and the new fallback. HTML with repeated whitespace therefore still never reaches extraction, and can now return much more content than before this PR.

Reproduced on 8ceb8108 with a page containing 120 JSON records and a 100-newline run: the real embedded-JSON/SmartCrusher candidate is 3,097 estimated tokens, but the new HTML path returns 4,473 tokens with `['lossless_log']`; the HTML dispatch is never called. The input is 4,492 tokens. This reproduction uses the real span scanner and lossless fold, a passthrough ML stub, and an HTML-dispatch spy. A smaller regression can extend your existing embedded-router stub test with a long newline run and assert that the HTML extractor still executes.

Please reconcile this earlier return with the HTML deferral: in normal lossy mode, allow the intended document-level extraction and retain the embedded result as a fallback; preserve the explicit lossless-only contract separately. Cover the successful-extraction and extractor-disabled/unavailable cases when the input also has a lossless fold, so the change cannot discard the already-banked embedded savings.

SulimanAbdulrazzaq:
@JerrettDavis thanks, reproduced and fixed in 7e84a640.

**What was fixed**
- The STAGE 0 fold return is skipped while an embedded-JSON result is held, so HTML extraction runs and that result stays the fallback.
- Lossless-only mode keeps the immediate embedded-JSON return; the deferral is lossy-mode only.
- Every extraction miss returns the held result directly, exactly as the pre-pass did: extractor disabled, unavailable, or raising, and an empty or whitespace-only extraction (the registry adapter passes trafilatura's `""` through as a result).

**Tests added** in `tests/test_transforms_content_router.py`
- `test_lossless_fold_does_not_preempt_html_extraction_for_embedded_json`: a 100-line blank run gives a real STAGE 0 fold; asserts the HTML dispatch runs and wins.
- `test_embedded_json_is_returned_when_html_extraction_yields_nothing`: disabled, unavailable, raising, empty, and whitespace extraction, each with and without a real fold (10 cases).
- `test_lossless_only_mode_keeps_the_embedded_json_early_return`.

Against the previous head `8ceb8108`, 9 of the 13 embedded-JSON/HTML tests fail.

**Validation** (Codespace, `uv sync --frozen --extra dev --extra html`, CPython 3.14.2, maturin-built `_core`, trafilatura 2.0.0)

```text
pytest tests/test_transforms_content_router.py -q: 77 passed
pytest tests -q -k "router or embedded or recursive_json or html or lossless" --ignore=tests/test_memory_eval.py: 613 passed, 35 skipped
ruff check .: All checks passed!
ruff format --check (both changed files): 2 files already formatted
mypy headroom/transforms/content_router.py: Success
```

Your scenario (120 records plus a 100-newline run; real span scanner, fold, and trafilatura; passthrough ML stub; dispatch spy), 4,560 input tokens:
- branch base: 2,176 tokens, `['embedded_json']`
- previous head: 4,542 tokens, `['lossless_log']`, HTML dispatch never called
- this head: 372 tokens, `['html']`; with the extractor disabled, 2,176 tokens, `['html', 'embedded_json']`

Not run: the full suite and whole-package `mypy headroom`. `tests/test_memory_eval.py` was skipped because it needs `litellm`, and native detection used the pure-Python fallback in this environment.

SulimanAbdulrazzaq:
@JerrettDavis I merged `main` into the branch to clear the conflict with #3556. `content_router.py` merged cleanly: this change only reads `self.config.lossless`, `self._lossless_then_lossy` and locals, none of which #3556 moved into the `ContextVar`. The only conflict was both PRs appending tests to `tests/test_transforms_content_router.py`, and both sets are kept.

Re-run on the merged tree:

- the router test file passes (84), #3556's concurrency tests pass (5), and the related selection passes (626 passed, 24 skipped);
- with `main`'s router and this branch's tests, exactly the 13 new tests fail;
- the real-page numbers are unchanged: 372 tokens through the extractor instead of `main`'s 2,176, the held embedded result when extraction yields nothing, and lossless-only mode identical to `main`.

The STAGE 0 fold fix from 7e84a640 is unchanged, and the Real Behavior Proof fields are now in the format the governance check reads. Could you take another look when you have a moment?

JerrettDavis (changes requested):
The new head fixes the prior blank-extraction and STAGE 0 fallback cases, but one enterprise-path issue remains: `route_embedded_json(...)` still runs before HTML extraction and stores its result in `embedded_fallback`. On the normal path where HTML extraction succeeds, that result is discarded, yet the nested compressor may already have written CCR/TOIN state and incurred the full JSON-compression cost. Please make the embedded fallback lazy—run it only after HTML extraction has produced no usable content—so successful HTML requests do not create unreachable recovery state or pay for compression whose output is never returned. The lossless early-return behavior can remain unchanged.

SulimanAbdulrazzaq:
@JerrettDavis done — the route is deferred, not just its result.

**What changed**

`recursive_json` grows `has_embedded_json()`: the span scan `route_embedded_json` already
ran, exposed on its own. It answers whether a routable span exists without calling the
dispatch, so nothing is compressed, no `<<ccr:…>>` marker is registered and no TOIN row is
written just to find out. `embedded_fallback` now holds the route as a thunk, invoked at
most once and only after extraction has come up empty.

A routable span is not a promise of a smaller result, so the thunk can return `None` where
the eager version had a concrete result. The STAGE 0 fold that such a block skipped on the
way in is returned instead — the lossless-then-lossy logic is extracted as
`_lossless_fold_result()` so both call sites share it — and deferring can never cost a
block the byte-exact win `main` returns for it.

One shape does change outcome, and I would rather point at it than have you find it: the
skip of the STAGE 0 return now keys on a route being *available* rather than on its result,
because the result is exactly what you asked us not to buy up front. So a folding HTML page
whose span is routable but comes back no smaller now reaches the extractor, where the
pre-pass returned the fold. That is the document-level extraction #3609 asks for, and
`test_extraction_wins_the_fold_when_a_span_is_routable_but_unprofitable` pins it. A page
with no routable span still takes the STAGE 0 return unchanged
(`test_plain_html_with_a_fold_keeps_the_stage_0_return`).

Every return between the pre-pass and the HTML dispatch was walked: the lossless-only
return and the non-HTML early return are unreachable with a fallback pending (it is only
set for HTML in lossy mode); the relevance-split return is LOG/SEARCH only; the STAGE 0
fold return is skipped while one is pending and is re-offered by the fallback; the external
compressor return discarded the eager result before and now costs nothing to reach.

**Evidence** (Codespace, `uv sync --frozen --extra dev --extra html`, CPython 3.13.15, live
trafilatura; one tree, `git rev-parse HEAD^{tree}` = `bd7fc1f67a2aae248b708c25d7bd44cb5b55b3c4`)

One real page — 120-record `ld+json` island, 8 prose paragraphs, a 100-newline run, 4,179
input tokens — through `_apply_strategy_to_content` with counting spies on
`route_embedded_json`, the nested `_apply_strategy_to_content` and `_record_to_toin`, and
nothing stubbed:

| `content_router.py` | Output | Chain | Route called | Nested span dispatches | TOIN rows | Wall |
|---|---|---|---|---|---|---|
| `main` | 2,134 tokens | `['embedded_json']` | 1 | 1 | 1 | 0.100 s |
| previous head `ac4e95d7` | 549 tokens | `['html']` | 1 (discarded) | 1 (discarded) | 2 | 0.121 s |
| this head | 549 tokens | `['html']` | **0** | **0** | **1** | 0.032 s |

Same answer as the previous head, without the compression it threw away or the second TOIN
row behind it.

```text
$ uv run pytest tests/test_transforms_content_router.py tests/test_recursive_json.py -q
98 passed in 4.45s
$ uv run pytest tests/test_content_router_concurrency.py -q
5 passed in 3.09s
$ uv run pytest tests -q -k "router or embedded or recursive_json or html or lossless" --ignore=tests/test_memory_eval.py
633 passed, 24 skipped, 12542 deselected in 29.38s
$ uv run ruff check .                                   All checks passed!
$ uv run ruff format --check <the 4 changed files>      4 files already formatted
$ uv run mypy headroom/transforms/{content_router,recursive_json}.py
```

Red/green on that same tree, swapping only the two source modules: against `main`'s modules
14 of the 19 selected router tests fail and `tests/test_recursive_json.py` cannot import
`has_embedded_json`; against `ac4e95d7` exactly 3 fail — the two that now assert the route
is never *called* behind a successful extraction, plus the new widened-shape test.

The PR body is updated with the same numbers.

JerrettDavis:
The latest deferral direction addresses the prior ordering finding, but this branch is now superseded by the narrower, current-main implementation in #3620. That PR preserves HTML-first extraction, falls through on empty extraction, retries embedded JSON only in passthrough fallback, retains lossless behavior and strategy accounting, and is green/clean after conflict-adapted verification. #3612 remains conflicted and carries a much larger recursive-JSON/test delta for the same outcome. Recommend closing it in favor of #3620 rather than resolving a competing implementation.

SulimanAbdulrazzaq:
Thanks for reviewing this through several rounds, @JerrettDavis. Agreed: #3620 covers the same fix on current main with a smaller change, so I'm closing this in its favor.

PR #3613, opened 2026-09-16
fix(proxy): stop get_recent deep-copying dropped message payloads

gglucass:
## Description

`RequestLogger.get_recent()` built each returned row with `dataclasses.asdict(entry)` and only then dropped `request_messages`, `compressed_messages` and `response_content`. `asdict` deep-copies every field first, so every call walked the full request and compressed transcripts of every entry it returned, for fields it was about to discard.

`/stats` calls `get_recent(10_000)` synchronously on the event loop on every build (both the uncached path and the `?cached=1` snapshot rebuild). Measured against 0.37.0 with a 300-message Claude Code transcript shape (~400 KB per entry): 3.4 ms per entry, so a full 10k deque costs ~34 s per `/stats` call, during which the loop serves nothing else. A desktop dashboard polling `/stats` reproduced this in production: 4,769 `/stats` calls in 55 min with a mean latency of 517 s, `/v1/messages` mean 135 s (max 1,839 s), 61k `Exceeded concurrency limit` warnings and a watchdog restart storm (Headroom Desktop user report, 2026-09-16).

The fix builds the row from `fields()` while skipping the heavy fields, so the payloads are never touched. Kept fields are scalars plus flat `dict`/`list` containers; those get the same shallow copy `asdict` produced for them, so callers still get copies rather than aliases into the log entry.

Fixes the backend half of the desktop report; the desktop side stops polling uncached `/stats` for connector verification separately.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Made

- `headroom/proxy/request_logger.py`: `get_recent` iterates `fields()` and skips `_HEAVY_FIELDS` instead of calling `asdict()` and filtering afterwards. Shallow-copies the kept `dict`/`list` fields.
- `tests/test_proxy/test_request_logger.py`: new test plants a leaf whose `__deepcopy__` raises inside both message payloads and asserts `get_recent` still returns the row with copied (not aliased) containers. Fails on the previous implementation, passes on this one.

## Testing

- [x] Unit tests pass locally (`uv run --frozen --extra dev pytest tests/test_proxy/test_request_logger.py`, 5 passed; `pytest tests/test_proxy -k "stats and not history"`, 6 passed)
- [x] `uvx ruff check` and `uvx ruff format --check` clean on both files
- [x] New test verified to fail on the old code (`git stash` the source change: 1 failed, 4 passed)

### Test Output

```
$ uv run --frozen --extra dev pytest tests/test_proxy/test_request_logger.py -q
tests/test_proxy/test_request_logger.py .....                            [100%]

============================== 5 passed in 0.23s ===============================

$ uvx ruff check headroom/proxy/request_logger.py tests/test_proxy/test_request_logger.py
All checks passed!
```

## Real Behavior Proof

- Environment: macOS 15, Python 3.12, headroom-ai 0.37.0 venv (the desktop app's pinned runtime), plus this branch.
- Exact command / steps: Built a `RequestLog` with `request_messages`/`compressed_messages` set to a 300-message transcript (3 content blocks each, ~416 KB serialized) and timed `dataclasses.asdict(entry)` in a loop; then ran the new unit test against the old and new `get_recent`.
- Observed result: 3.4 ms per `asdict` call on 0.37.0, i.e. ~34 s for the 10,000-entry `get_recent` that `/stats` performs on the event loop. With this change the per-row cost is a `fields()` walk plus shallow copies of two small containers (microseconds), and the new test passes only on the new code.
- Not tested: an end-to-end `/stats` latency comparison on a long-lived proxy with a full 10k deque; the production numbers above come from a user's `proxy.log`, not from a controlled run.

## Runtime Rollout Safety

- Stable/default behavior changed: no; `/stats` returns the same rows and keys, they are just built without walking the dropped payloads
- Kill switch / disable path: none needed; revert the commit
- Qualification impact: none
- Rollback path: revert the commit

## Review Readiness

## Checklist

## Screenshots

n/a

## Additional Notes

`get_recent_with_messages` keeps `asdict` on purpose: it returns the payloads, so the walk is the point there.


JerrettDavis (changes requested):
The performance direction is right, but this changes the isolation contract of get_recent(). sdict() recursively copied every retained field; the replacement only shallow-copies the top-level dict/list. RequestLog.tags is dict[str, Any] and savings_breakdown is list[dict[str, Any]], so a caller can now mutate nested objects that remain owned by the in-memory log (for example, 
ecent[0]['savings_breakdown'][0]['tokens'] = ... changes the next /stats result). Please deep-copy only the retained fields (avoiding the three heavy payload fields), or otherwise recursively isolate their nested mutable values, and add a nested-container regression test. That preserves the speedup without weakening the endpoint's previous copy semantics.

gglucass:
Addressed in 541aed34: the retained fields are now deep-copied (`deepcopy` per kept field), so nested `tags` / `savings_breakdown` values stay isolated from the in-memory log while the three heavy payload fields are still never walked. The regression test now mutates a nested tag and a `savings_breakdown` entry and asserts the next `get_recent` is unchanged; the `_NoCopy` leaf still proves the payload walk is gone.

JerrettDavis (approved):
Re-verified the revised and refreshed head: heavy message fields are never traversed, retained fields remain deeply copied, and the prior nested-alias concern is resolved. The branch is current with main, mergeable/conflict-free, and the full current CI/security matrix is green. Ready for maintainer merge review.

PR #3616, opened 2026-09-16
fix(spreadsheet): render an .xls cell the way the .xlsx loader renders it

L4XB:
## Description

`_load_xls` reads a legacy workbook with `sheet.row_values(i)`, which hands back xlrd's **raw storage** rather than the value. The `.xlsx` loader goes through openpyxl and hands back the value. So the two loaders disagree about the same workbook.

Measured on `main` (35b1156), the same three cells written to both formats:

| cell | `.xls` today | `.xlsx` |
|---|---|---|
| `2024-01-01` | `45292.0` | `2024-01-01 00:00:00` |
| `TRUE` | `1` | `True` |
| `12` | `12.0` | `12` |

The date is the one that matters: a serial number is not recoverable from the compressed text, so whatever reads the table downstream sees `45292.0` and has no way to know it was a date. The boolean and the whole number are the same class of drift — `1` and `12.0` are not what the sheet says.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/spreadsheet_ingest.py`: new `_xls_cell(cell, datemode)` converts each cell by its xlrd type, matching what openpyxl already delivers — `XL_CELL_DATE` → `datetime` via `xldate_as_tuple` with the workbook's `datemode` (the 1900/1904 epoch matters and only the book knows which it is); `XL_CELL_BOOLEAN` → `bool`; a whole `XL_CELL_NUMBER` → `int`, since xlrd stores every number as a double; `XL_CELL_ERROR` → the text Excel shows (`#DIV/0!`), which is what openpyxl with `data_only=True` gives.
- A cell whose `xldate_as_tuple` has no date part — a time such as `12:00` is `(0, 0, 0, 12, 0, 0)` — becomes a `datetime.time`, again matching openpyxl. Building a `datetime` from it raises `ValueError: year 0 is out of range`.
- A float that is not whole, a string, and an empty cell are untouched: `1.5` stays `1.5`, and a blank still renders as an empty field through `_rows_to_csv`.
- `sheet.row_values(i)` becomes `sheet.row(i)`, because the conversion needs the cell type that `row_values` throws away.
- The `# pragma: no cover - legacy .xls; needs optional xlrd + binary fixture` comes off `_load_xls`: it is covered now.
- `pyproject.toml`: `xlrd>=2.0.1` (the `spreadsheet` floor) and `xlwt>=1.3.0` join the dev extra beside `openpyxl`, with the same style of comment. The CI test job installs only `[dev]`, and every new test `importorskip`s both: `xlwt` is the only way to write a legacy `.xls` fixture, and `xlrd` is what `_load_xls` reads it with. `uv.lock` is regenerated to match (13 added lines, nothing else moves).

## Testing

Three cases in `tests/test_transforms_tabular.py`, each building a real `.xls` with `xlwt`:

- the row renders as `2024-01-01 00:00:00,True,12,1.5,ok` — it was `45292.0,1,12.0,1.5,ok`;
- `load_spreadsheet` of the `.xls` and of the equivalent `.xlsx` return **equal** dictionaries, which is the property the first test is really about;
- a time-only cell renders as `12:00:00` in both formats, where it rendered as `0.5` (and raised with only the first commit applied).

`mypy` is unchecked because it was not run locally; CI's lint job runs it.

### Test Output

```text
$ ruff check headroom/transforms/spreadsheet_ingest.py tests/test_transforms_tabular.py
All checks passed!

$ ruff format --check headroom/transforms/spreadsheet_ingest.py tests/test_transforms_tabular.py
2 files already formatted

$ python -m pytest tests/test_transforms_tabular.py -rs
43 passed in 4.03s
```

Before `xlrd` joined the dev extra, the same command in the CI test environment gave 40 passed, 3 skipped (`could not import 'xlrd'`).

## Real Behavior Proof

- Environment: macOS 26.6 (arm64), CPython 3.12.11, `xlrd 2.0.2`, `openpyxl`, `xlwt 1.3.0`; the wheel built with `maturin build --profile ci` (Rust 1.98.1) and installed as `"${WHEEL}[dev]"`, as the CI test job does.
- Exact command / steps: build a one-cell `.xls` with `xlwt` and the equivalent `.xlsx` with `openpyxl`, then call `headroom.transforms.spreadsheet_ingest.load_spreadsheet` on each. That module imports cleanly on its own; only the test file's other imports pull in `headroom._core`.
- Observed result: on `main` the `.xls` row reads `45292.0,1,12.0,1.5,ok` and a time-only cell reads `0.5`; with this PR both match the `.xlsx` output:

```text
# before (main)
xls  -> ['Starts', '0.5']                           # the time-only cell
xls  -> ['When,Active,Units,Rate,Text',
         '45292.0,1,12.0,1.5,ok']                   # the value cells

# after
xls  -> ['Starts', '12:00:00']
xlsx -> ['Starts', '12:00:00']
xls  -> ['When,Active,Units,Rate,Text',
         '2024-01-01 00:00:00,True,12,1.5,ok']
```

- Not tested: `mypy headroom` locally (CI's lint job runs `mypy headroom --ignore-missing-imports`; green at 466683c). Note on CI: its test job installs only `"${WHEEL}[dev]"`, and `xlrd` was only in the `spreadsheet` extra, so the three new cases were skipped there (shard 4: `s`/`ss`) until `xlrd` joined `dev` in this PR.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. `spreadsheet_ingest` is not behind a rollout channel.
- Stable/default behavior changed: yes, for `.xls` input only — a date, a boolean, a whole number and an error cell now render as their value instead of as xlrd's storage, which is what the `.xlsx` path has always done. `.xlsx`, `.csv` and every other input are byte-identical.
- Kill switch / disable path: none needed; revert the commit.
- Qualification impact: any golden output that pinned `45292.0`, `1` or `12.0` for a `.xls` would change. There is none in the repo — `_load_xls` was `# pragma: no cover` before this PR.
- Rollback path: `git revert`; the change is confined to `_xls_cell` / `_load_xls`, two dev-extra lines and their `uv.lock` entries.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (no documented behaviour changes: the `.xls` path now matches the `.xlsx` path the docs already describe)
- [ ] New and existing unit tests pass locally with my changes (full suite locally: 12550 passed, 599 skipped, 1 failed; the failure, `tests/test_cli/test_mcp_reconcile.py::test_ordinary_install_does_not_adopt_serena`, needs the venv's `bin` on `PATH` and passes with it)

## Additional Notes

The time-only case was found after the first commit: `xldate_as_tuple` returns `(0, 0, 0, 12, 0, 0)` for `0.5`, and `datetime(0, 0, 0, 12, 0, 0)` raises, so the whole workbook failed to load. Commit 466683c fixes it and adds the third test.


JerrettDavis (approved):

L4XB:
@JerrettDavis a follow-up after your approval: the three new `.xls` tests were being skipped in CI. The test job installs `"${WHEEL}[dev]"`, but `xlrd` was only in the `spreadsheet` extra, so `pytest.importorskip("xlrd")` skipped all three. That shows as the `s` / `ss` after `test_transforms_tabular.py` in shard 4, and it accounts for 16 of the 19 lines Codecov lists as uncovered. The three left are the `xldate_as_tuple` fallback and the error-cell branch.

The new push adds one dev-extra line, `"xlrd>=2.0.1"` (the same floor `spreadsheet` uses), plus the matching `uv.lock` entries. `uv lock --check` passes on main but had been failing on this branch since `xlwt` joined `dev`; no other lock entry changes. No source or test code changed, so could you renew the approval on the new head?

I checked it with a wheel built with `maturin build --profile ci` and installed as `"${WHEEL}[dev]"`, as the test job does:
- `tests/test_transforms_tabular.py`: 40 passed, 3 skipped before; 43 passed after. Coverage of `spreadsheet_ingest.py` goes from 55% to 93%.
- With `xlrd` installed, the three tests fail against the pre-PR `spreadsheet_ingest.py` (`45292.0,1,12.0,1.5,ok`, `0.5`) and pass with this PR.
- I ran `pytest tests scripts/tests` with and without `xlrd`. The only outcome change is those three tests going from skipped to passed.

I also corrected the description: the time-only cell renders as `0.5` on `main`; the `ValueError` only happened with the first commit of this PR applied.

JerrettDavis (approved):
Verified the refreshed head against current main: mergeable and conflict-free. The .xls normalization is reasonable and its tests cover dates, time-only values, booleans, integers, and .xlsx parity; the full current CI/security/native matrix is green. Ready for maintainer merge review.

PR #2693, opened 2026-07-31
fix(grok): route CLI traffic to api.x.ai on shared proxies

p-clements:
## Description

`headroom wrap grok` points the Grok CLI at the local proxy via `GROK_MODELS_BASE_URL` and, when it starts the proxy, sets `openai_api_url=https://api.x.ai`. When Claude/Codex already own port 8787, wrap reuses that process without restart (other wrappers attached), so process-wide `OPENAI_API_URL` stays `https://api.openai.com`.

Grok CLI cannot stamp `x-headroom-base-url` (no custom attribution headers). Inference then goes to OpenAI with xAI session tokens and returns 401. Observed on the wire:

- `x-xai-token-auth: xai-grok-cli`
- `user-agent: grok-pager/... grok-shell/...`
- outbound `POST https://api.openai.com/v1/chat/completions` status 401

This PR adds per-request Grok CLI detection and routes chat / models / responses to `https://api.x.ai` when those signals are present and the OpenAI target is still its default, without restarting the shared proxy or changing Claude/Codex paths. An operator-configured gateway is never bypassed.

Not covered by #2626 (attribution / UA map only when custom base is already `api.x.ai`).

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Add `is_grok_cli_request()` in `providers/grok/runtime.py` using narrow wire signals: exact `x-xai-token-auth: xai-grok-cli` and the `grok-pager/` / `grok-shell/` UA product tokens, compared per whitespace-split token so wrappers like `litellm-grok/1.0` cannot match
- Resolve OpenAI chat and responses final upstream as: `x-headroom-base-url` then Grok CLI then process `OPENAI_API_URL`
- Route `GET /v1/models` and OpenAI-classified passthrough through the same Grok to xAI base
- Add unit tests for detector, resolver priority, and passthrough/models base selection
- Add `tests/test_grok_shared_proxy_call_sites.py`, which drives `handle_openai_chat` and `handle_openai_responses` against a stubbed transport and asserts the outbound URL — reverting either call site to the old `custom_only or OPENAI_API_URL` formula turns these red
[...]

PR #3617, opened 2026-09-16
fix(grok): route grok login session tokens to cli-chat-proxy.grok.com

gglucass:
## Description

`headroom wrap grok` (and `wrap grok-build`) start the proxy with the OpenAI target set to `https://api.x.ai`. That is the right host for an `xai-` API key and the wrong host for a `grok login` session: in that mode the Grok CLI sends its auth.x.ai OIDC session token as the bearer (with `x-xai-token-auth: xai-grok-cli`), api.x.ai rejects it, and every message fails with "Authentication required: your session has expired or your credentials were rejected" even though `grok login` just succeeded. Unsetting the proxy env vars makes Grok work again immediately, so users read it as a Headroom bug (paid-user report to the desktop app, 2026-09-16, with a `RUST_LOG=debug` capture showing login succeeding against auth.x.ai and only the 127.0.0.1 proxy calls failing).

Session tokens are only valid at `https://cli-chat-proxy.grok.com`, which is the CLI's own default base for session auth (its embedded docs: "When you set `models_base_url`, Grok uses API key auth (`Authorization: Bearer`) instead of session auth"). This PR routes Grok CLI requests whose bearer is not an `xai-` key to that host, at the three places the OpenAI-compatible upstream is resolved. An `xai-` key keeps `api.x.ai`, and an explicit `x-headroom-base-url` still wins.

Related: #2693 makes a shared proxy recognise the Grok CLI and send it to `api.x.ai` instead of `api.openai.com`; this PR is about which xAI host accepts the credential and composes with it (the session check runs before the target fallback).

## Type of Change

## Changes Made

- `providers/grok/runtime.py`: add `SESSION_API_URL`, `is_grok_cli_request()` (exact `x-xai-token-auth: xai-grok-cli` marker or `grok-shell/` / `grok/` UA product token, matched per whitespace-split token) and `session_upstream()` (returns the session host for a Grok CLI request whose bearer is not an `xai-` key, else `None`).
- `proxy/handlers/openai.py`: `_resolve_openai_upstream_base()` falls back to `session_upstream()` when no `x-headroom-base-url` header is present, so `/v1/chat/completions` and `/v1/responses` route a session token correctly.
- `providers/proxy_targets.py`: `select_passthrough_base_url()` returns the session host for the same requests.
- `providers/proxy_routes.py`: `GET /v1/models` (the CLI's model catalog fetch) uses the session host for the same requests.
- Tests for the detector, the credential split, the handler resolver (session host, `xai-` key stays, explicit header wins) and passthrough selection.

## Testing

### Test Output

```text
uv run --frozen --extra dev pytest tests/test_provider_grok.py tests/test_proxy/test_openai_upstream_header.py tests/test_provider_proxy_targets.py tests/test_openai_codex_routing.py -q
59 passed in 1.02s

uv run --frozen --extra dev pytest tests -k 'grok or passthrough or upstream or models or openai_chat or responses_routing' -q
653 passed, 37 skipped, 12399 deselected in 106.80s

uvx ruff check headroom tests      -> All checks passed!
uvx ruff format --check headroom tests -> 1420 files already formatted
uv run --frozen --extra dev mypy headroom/proxy/handlers/openai.py headroom/providers/grok headroom/providers/proxy_targets.py headroom/providers/proxy_routes.py --ignore-missing-imports
Success: no issues found
```

## Real Behavior Proof

- Environment: macOS 15 (arm64), Grok CLI 0.2.112 (`grok-shell/0.2.112`), Headroom 0.37.0 wheel behind the desktop app, Python 3.12.
- Exact command / steps: (1) Ran `grok -p "Reply with OK"` through the proxy with the api.x.ai target: CLI reports not signed in. (2) `curl https://cli-chat-proxy.grok.com/v1/chat/completions` with a Grok-CLI-shaped request (`Authorization: Bearer <jwt>`, `x-xai-token-auth: xai-grok-cli`, `User-Agent: grok-shell/0.2.112`) both directly and through the Headroom proxy with the upstream pointed at that host. (3) Same request to `https://api.x.ai`.
- Observed result: cli-chat-proxy answers `auth_kind=bearer, x_xai_token_auth=xai-grok-cli` for the session-shaped request (credential recognised, then rejected because the token was a placeholder) and `x_xai_token_auth=none, Unauthenticated` without the marker header; the proxy forwards the marker header intact (identical response direct vs. via proxy). api.x.ai answers the same request `unauthenticated:bad-credentials`. Confirms the host split and that nothing in the proxy strips the header the session host keys on.
- Not tested: a full `grok -p` round trip with a live `grok login` session through this branch (my session had expired and `grok login` needs a browser); the `xai-` API-key path against api.x.ai (unchanged behaviour, covered by unit tests only).

## Runtime Rollout Safety

- Stable/default behavior changed: Yes, only for requests that identify as the Grok CLI and carry a non-`xai-` bearer: they now go to `cli-chat-proxy.grok.com` instead of the configured OpenAI target. Every other request, and any Grok CLI request with an `xai-` key, is unchanged.
- Kill switch / disable path: Not applicable as an env flag; an explicit `x-headroom-base-url` header still overrides the session host, and reverting the PR restores the previous routing.
- Qualification impact: None.

## Review Readiness

## Checklist

Not applicable.

## Additional Notes

- Documentation: no docs change; `render_setup_lines()` still prints `api.x.ai` as the proxy upstream, which stays true for API-key users. Happy to add a line about session routing if wanted.
- `tests/test_provider_grok.py::test_grok_build_launch_env_sets_models_base_url` fails on `main` in any shell that exports `GROK_CLI_CHAT_PROXY_BASE_URL`, because `build_launch_env(environ={})` treats the empty mapping as falsy and reads `os.environ`. Pre-existing, not touched here.


gglucass:
Pushed a follow-up that raises patch coverage: added `test_list_models_routes_grok_session_login_to_session_host`, which drives `GET /v1/models` (the Grok CLI's catalog fetch) through a `TestClient` and asserts a session-login request resolves to `cli-chat-proxy.grok.com` while an `xai-` key keeps the configured target.

Local coverage of the changed code after that test:

- `providers/grok/runtime.py` (the actual new logic): 100%.
- `providers/proxy_targets.py`: the new Grok hook is covered both ways; the 3 lines Codecov still lists (`57->66`, `65`, `101`) are pre-existing — line 65 is the existing Azure `x-headroom-base-url` warning log, not touched here.
- The chat/responses fallback in `handlers/openai.py` and the `/v1/models` route hook are both exercised.

Codecov's original 30% was inflated by a coverage-upload/shard artifact that made `runtime.py` look uncovered (it is 100% locally). Happy to add more if you'd like a specific path pinned.

JerrettDavis (approved):

chopratejas:
Thanks for this — the underlying bug is real and worth fixing: a `grok login` session token genuinely shouldn't be sent to `api.x.ai`. I'd like to tighten the routing predicate before this lands, because it decides where to send a **bearer credential** and it's driven entirely by client-controlled headers.

**The predicate fires when the credential is foreign.** `session_upstream()` returns the grok host whenever the bearer is *not* `xai-`, and the client match (`headroom/providers/grok/runtime.py:53-61`) accepts either `x-xai-token-auth: xai-grok-cli` **or** any whitespace-delimited User-Agent token starting with `grok-shell/` or `grok/`. Run against this branch:

```
{'authorization':'Bearer sk-proj-REALKEY','user-agent':'grok/1.0'}           -> https://cli-chat-proxy.grok.com
{'authorization':'Bearer sk-litellm-abc','x-xai-token-auth':'XAI-GROK-CLI'}  -> https://cli-chat-proxy.grok.com
{'user-agent':'grok/1.0'}   (no auth at all)                                 -> https://cli-chat-proxy.grok.com
```

So any client that sets one header gets its `Authorization` forwarded to a public third-party host.

**The deployment case I'm most concerned about:** this branch runs on *every* install, including Headroom fronting Kong or LiteLLM with `OPENAI_TARGET_API_URL` pinned to an internal gateway (`proxy_targets.py:52-54`, `handlers/openai.py:381-383`). A request carrying the marker header silently egresses past the gateway that was the whole point of that deployment. Note the adjacent code at `proxy_targets.py:55-62` hardens specifically against client-chosen destinations (with a CVE reference) — this moves the other way.

**Two cheap changes that keep the fix:**

1. Drop the bare `grok/` UA arm and require the `x-xai-token-auth` marker — that's the strong, product-specific signal. (`grok/` is also already claimed for a different meaning at `headroom/proxy/auth_policy.py:22,37`, so reusing it here overloads an existing loose prefix.)
2. Require the bearer to be JWT-shaped (`eyJ`-prefixed, three dot-separated segments). A real `grok login` session token always is; `sk-…`, `ghu_…` and gateway keys never are.

I'd also want an env kill switch (`HEADROOM_GROK_SESSION_ROUTING=0`) before shipping, since there's currently no way for an operator to turn this off.

**Everything else checks out** — I verified `_bearer_is_api_key` is case-folded and scheme-correct, explicit `x-headroom-base-url` still wins, the host is a constant so there's no SSRF surface, and path composition matches the `api.x.ai` shape it replaces.

One side effect worth knowing about, benign but surprising: `_resolve_openai_upstream_base()` becoming non-`None` means `merge_extra_headers` (`proxy/helpers.py:1841-1846`) will now **drop** operator-configured `openai_extra_headers` for these requests unless grok.com is a trusted upstream. Correct fail-safe, but an operator who configured a header for api.x.ai loses it silently.

gglucass:
You are right, and the threat model is the part I had wrong: I was treating this as "identify the Grok CLI", which is attribution, when it is really "may this credential be re-pointed at a third-party host". Those deserve different predicates. Reproduced all three of your cases verbatim on the old branch before changing anything:

```text
{'authorization':'Bearer sk-proj-REALKEY','user-agent':'grok/1.0'}          -> https://cli-chat-proxy.grok.com
{'authorization':'Bearer sk-litellm-abc','x-xai-token-auth':'XAI-GROK-CLI'} -> https://cli-chat-proxy.grok.com
{'user-agent':'grok/1.0'}  (no auth at all)                                 -> https://cli-chat-proxy.grok.com
```

Pushed e24cf4a with all three changes you asked for:

**1. Bare `grok/` arm dropped from session routing.** `grok-shell/` and the `x-xai-token-auth` marker remain. I kept `is_grok_cli_request` loose for attribution and documented that it must never decide where a credential is sent, rather than tightening it globally — it is exported, and the two questions genuinely differ.

**2. The credential has the final say.** `_bearer_is_session_jwt` requires `eyJ` + three non-empty dot-separated segments. Deliberately not case-folded, since base64url is case-sensitive.

**3. `HEADROOM_GROK_SESSION_ROUTING=0`** turns it off outright.

I did *not* make the marker header mandatory on top of the JWT check, because your case 2 shows the marker alone is not sufficient anyway, while requiring it would break a real `grok-shell` that does not send it. The JWT shape is what actually separates a session token from a gateway key. Say the word if you would rather have both.

After:

```text
all three of your cases                          -> None
Bearer ghu_abcdef  + marker                      -> None
Bearer sk-ant-xyz  + grok-shell/ UA              -> None
Bearer eyJabc      + grok-shell/ UA (not 3 segs) -> None
Bearer eyJa..y     + grok-shell/ UA (empty seg)  -> None
real JWT + marker / + grok-shell/ UA             -> https://cli-chat-proxy.grok.com
Bearer xai-abc123  + grok-shell/ UA              -> None   (keeps api.x.ai)
HEADROOM_GROK_SESSION_ROUTING=0, real JWT        -> None
```

Those are now a parametrized test plus a kill-switch test. 48 passed across the grok, passthrough, auth_policy, proxy-targets, banner, MCP-registry and litellm-passthrough suites.

On the `merge_extra_headers` side effect — agreed it is benign and fail-safe, and its blast radius shrinks a long way here, since `_resolve_openai_upstream_base()` now only goes non-`None` for a genuine session JWT rather than any request with a borrowed header. Happy to handle dropping `openai_extra_headers` on the grok.com path separately if you want it explicit.

JerrettDavis (changes requested):
The API-key cases are fixed on e24cf4a, but the credential-routing blocker remains for foreign JWTs. _bearer_is_session_jwt() proves only syntax (yJ + three nonempty segments), not that the token was issued for Grok. Internal gateways/SSO commonly use JWT bearer credentials, and both accepted client signals (x-xai-token-auth and grok-shell/ UA) are caller-controlled. Thus a request such as Authorization: Bearer eyJ...<an internal gateway JWT> plus the marker still bypasses an operator's configured OPENAI_TARGET_API_URL and forwards that credential to cli-chat-proxy.grok.com; routing is on by default, so the kill switch does not prevent this default exfiltration path. Please bind the override to trusted deployment context—for example only when the configured target is already the canonical https://api.x.ai, or make session rerouting explicit opt-in—and add a regression proving a configured internal target cannot be bypassed by a JWT-shaped bearer plus spoofed Grok headers.

PR #3618, opened 2026-09-16
fix(tabular): pass a table through when one cell is past the csv field limit

L4XB:
## Description

`csv.field_size_limit` is **128 KB per cell** by default. One pasted document, log excerpt or base64 blob in a column goes past it, and the stdlib reader then raises for the **whole text**, not for the row:

```python
>>> csv.field_size_limit()
131072
>>> list(csv.reader(io.StringIO("id,body\nl," + "x" * 200_000)))
_csv.Error: field larger than field limit (131072)
```

`parse_csv` does not catch it, so `TabularCompressor.compress` raises out of a content type it advertises support for:

```
>>> TabularCompressor().compress("id,title,body\nl,short,ok\n2,long," + "x" * 200_000)
Error: field larger than field limit (131072)
```

The module already has a policy for a table it cannot use — `parse_tabular` returns `None` for a ragged one (#1652) and the text passes through verbatim to the plain-text route. An oversized cell is the same situation and should take the same exit.

This is not hypothetical for a context compressor: the content it sees is tool output and pasted material, which is exactly where a single 128 KB+ cell comes from. Two other projects that ingest CSVs hit it and say so in their own code — `onyx/backend/onyx/utils/csv_utils.py`: *"real-world data (long descriptions, pasted docs, base64 blobs) routinely exceeds"*, and `skyvern/forge/sdk/workflow/models/block.py`. Both raise the limit process-globally, which is a reasonable call for an application and the wrong call for a library running inside someone else's process.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/transforms/tabular_ingest.py`: `parse_csv` catches `csv.Error` and returns `([], [])`, which makes `parse_tabular` return `None` and the compressor pass the content through unchanged. The docstring records why the limit is not raised instead.
- No process-global state is touched, so a host application's own `csv` behaviour is unaffected.
- Nothing else changes: a table that parses today parses identically, and the `[]` return reuses the empty-input path that was already there.

## Testing

Two cases in `tests/test_transforms_tabular.py`, next to the ragged-table ones they mirror:

- `test_parse_csv_gives_up_on_a_cell_past_the_field_size_limit`
- `test_compress_passes_through_a_table_with_an_oversized_cell`

`pytest` is unchecked because the suite does not run in my checkout, not because it failed — see **Not tested**.

### Test Output

```text
$ ruff check headroom/transforms/tabular_ingest.py tests/test_transforms_tabular.py
All checks passed!

$ ruff format --check headroom/transforms/tabular_ingest.py tests/test_transforms_tabular.py
2 files already formatted

$ python -m pytest tests/test_transforms_tabular.py
tests/test_transforms_tabular.py:24: in <module>
    from headroom.transforms.content_router import (
headroom/transforms/error_detection.py:41: in <module>
    from headroom._core import (
E   ModuleNotFoundError: No module named 'headroom._core'
```

The collection error is identical with my changes stashed: the file imports the compiled Rust extension and there is no Rust toolchain on this machine.

## Real Behavior Proof

- Environment: macOS 15 (arm64), CPython 3.12, no Rust toolchain.
- Exact command / steps: call `headroom.transforms.tabular_ingest.parse_csv` and `TabularCompressor().compress` on `"id,title,body\nl,short,ok\n2,long," + "x" * 200_000`. That module imports cleanly on its own; only the test file's other imports pull in `headroom._core`.
- Observed result: `parse_csv` raised `field larger than field limit` before the change and returns `([], [])` after it, and `compress` passes the text through byte-identical (output below).

```text
# before (source change stashed)
parse_csv -> Error : field larger than field limit (131072)

# after
parse_csv -> ([], [])
compress -> passthrough, byte-identical
```

- Not tested: `pytest tests/test_transforms_tabular.py` and `mypy headroom`, both of which need `headroom._core`; CI builds it, so the two new cases run there. Also **not** tested end to end through `ContentRouter`, for the same reason — I verified the public `TabularCompressor` / `parse_tabular` / `parse_csv` surface, and I do not claim to know whether the router's dispatch would have swallowed the exception or propagated it.

## Runtime Rollout Safety

- Rollout-managed feature(s): none. `tabular_ingest` is not behind a rollout channel.
- Stable/default behavior changed: only for text that raises today. A table that parses now parses identically; a table that raised now passes through verbatim and is still compressed by the plain-text route.
- Kill switch / disable path: none needed; revert the commit.
- Qualification impact: none — no golden output can have pinned an exception from this path.
- Rollback path: `git revert`; the change is four lines inside `parse_csv`.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (no documented behaviour changes)
- [ ] New and existing unit tests pass locally with my changes (cannot run the suite here — see **Not tested**)

## Additional Notes

If you would rather compress these tables than pass them through, the other option is a `csv.field_size_limit` bump. I did not take it because it is process-global and headroom is imported into someone else's process: raising it there silently changes how that application's own `csv` code behaves. If you want it anyway, a set/restore around the parse keeps the blast radius to this module, at the cost of a race between concurrent parses on separate threads — which is the trade-off skyvern's comment describes when it chose to set the limit once at import instead.


JerrettDavis (approved):

PR #3605, opened 2026-09-15
fix(proxy): gateway turn contract for /v1/compress behind a compress-turn seam

chopratejas:
## Description

Adds a gateway turn contract to `POST /v1/compress`, allowing an API gateway to submit a complete provider request and later report the provider response. Requests without a top-level `gateway` object remain unchanged.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Adds the request half to `/v1/compress` and the response half at `/v1/compress/response`, including bounded pending turns, idempotent rounds, usage relay, and redrive obligations.
- Adds a reusable six-stage compress-turn extension seam and turn-hook priority/provider-header support.
- Keeps native tool-schema deferral in the licensed extension while sharing protocol-shape helpers with core.
- Accepts Anthropic, OpenAI chat, OpenAI Responses, and Kong-shaped usage payloads.
- Documents the gateway contract in `docs/content/docs/proxy.mdx` and `wiki/proxy.md`.

## Testing

- [x] Unit tests pass (`pytest`)
- [x] Linting passes (`ruff check .`)
- [x] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [x] Manual testing performed

### Test Output

```text
tests/gateway + gateway unit + turn-hook seam + test_issue_746_tool_search:
358 passed

ruff 0.16.3 lint/format and mypy on touched modules:
clean

Kong 3.9 Docker e2e (HEADROOM_KONG_E2E=1):
4 passed
```

## Real Behavior Proof

- Environment: Kong 3.9 Docker e2e and real-provider Claude Code through Kong on SWE-bench-mini.
- Exact command / steps: `HEADROOM_KONG_E2E=1` gateway e2e; real-provider traffic through the gateway contract with usage relay and licensed native deferral.
- Observed result: four Kong e2e tests passed; prefix cache remained intact with zero busts; relayed usage matched Anthropic token-for-token; licensed tool-search deferral reduced cache-read tokens per turn by 27–30%.
- Not tested: production rollout at scale.

## Runtime Rollout Safety

- Rollout-managed feature(s): gateway contract and in-process pending-turn registry.
- Minimum rollout channel: controlled gateway deployment before broader enablement.
[...]

PR #3619, opened 2026-09-16
fix(proxy): gateway turn fixes found by benchmarking behind a gateway

chopratejas:
## Description

Fixes five defects found by running real Claude Code sessions through a gateway that uses the `/v1/compress` turn contract (#3605).

## Type of Change

## Changes Made

- **Prompt text is never compressed on replaying callers.** The coding profile compresses user messages, and on turn 1 that rewrote a plain agent's task ("Fix this bug in the library source" became "Fix bug in source", and test ids lost their `tests/test_` prefix). When `prefix_replay_guaranteed` is set, `content_router` protects the opening user messages and the newest user message, for string and list content alike. The one exception is a newest user message that answers a fenced shell command in the assistant turn before it: that is a text harness's tool observation and still compresses. Tool results still compress, and cache freeze is unchanged.
- **Anthropic's four cache breakpoint limit.** One benchmark request failed with `max 4 cache_control blocks, found 5`. Session turns in the OpenAI handler now re-place markers with `normalize_message_cache_control`. A last-stop guard, `enforce_cache_breakpoint_budget`, runs in the gateway extension for Anthropic bodies: it re-places markers, then drops the oldest extras, and logs `event=cache_breakpoint_budget` when it repairs.
- **Failure relay without a body.** `POST /v1/compress/response` rejected an upstream 4xx/5xx without a response body as `missing_response`. A body is now required only on 2xx.
- **Usage totals after a re-drive.** The done payload carries `billed_usage` summed across rounds. After one or more re-drives, `response` is never null: it is the hook's replacement or the latest provider response (including when the hook hands back that same object), with `usage` set to the totals in its own key shape (chat or Anthropic). Turns with no re-drive keep `response: null`. Before, the client saw the last round only.
- **Cold Kompress stall.** When native Kompress init is deferred at startup, a daemon thread preloads it after a short delay. `HEADROOM_KOMPRESS_WARMUP=0` disables it. It is off under pytest and skipped on glibc < 2.28 (the #1908 segfault hosts).
- Docs: `proxy.mdx` and `wiki/proxy.md` describe the 2xx-only body rule and `billed_usage`.

## Testing

- [x] Type checking passes (`mypy headroom`) on the changed modules

### Test Output

```text
$ pytest tests/test_transforms_content_router.py tests/test_compress_session_mode.py \
    tests/test_compress_route_tokenizer_by_model.py tests/gateway tests/test_proxy \
    tests/test_kompress_preload_deferral.py
672 passed, 4 skipped in 228.49s

$ uvx ruff@0.16.3 format --check <changed files> && uvx ruff@0.16.3 check <changed files>
12 files already formatted
All checks passed!

$ mypy <changed modules>
```

New tests: 5 prompt-protection tests in `test_transforms_content_router.py` (3 fail without the fix), `tests/gateway/test_cache_breakpoint_budget.py`, failure-relay and billed-usage tests in `test_redrive_loop.py`, `billed_usage` in the response-half contract test, `tests/test_proxy/test_kompress_background_warmup.py`.

## Real Behavior Proof

- Environment: headroom proxy on this branch behind a gateway, Claude Code with claude-sonnet-5, real Anthropic calls.
- Exact command / steps: 8 smoke sessions (32 calls) on SWE-bench-style tasks, with a spend logger recording every provider round.
- Observed result: 0 errors. Task statements reached the provider verbatim. On every re-driven call (2 and 3 rounds) the client's usage equalled the sum of the per-round provider usage. Kompress warmed in the background in about 4 s at startup; in a replay of the slow session the first code-aware turn went from 5.4 s to 1.3 s.
- Not tested: glibc < 2.28 hosts (warm-up is skipped there by design). The original body behind the 5-breakpoint error was not captured, so the guard is covered by unit tests rather than a live repro.

## Runtime Rollout Safety

- Stable/default behavior changed: yes. Prompt text on replaying callers is no longer compressed, the done payload adds `billed_usage`, and Kompress preloads in the background when deferred.
- Kill switch / disable path: `HEADROOM_KOMPRESS_WARMUP=0` for the warm-up. The other changes are correctness fixes with no switch.
- Qualification impact: none expected; the breakpoint guard only acts on bodies that would otherwise get a 400.

## Review Readiness

## Checklist

## Additional Notes

Open follow-up, not in this PR: steady-state code-aware compression falls back to Kompress at 0.3 to 1.3 s per tool result even when it saves nothing. A gate for that is worth a separate change.


JerrettDavis (changes requested):
Two current-head blockers:\n\n1. The new prompt protection is inconsistent between list and string content. prompt_turn correctly includes the newest user message, and list-form content passes that directly as protect_prompt_text; however, the string-content gate checks (prompt_turn and i < first_assistant_index). That second condition excludes every newest user turn after the first assistant message, so the plain-string follow-up prompt can still be compressed—the exact behavior this PR says it fixes. Please protect prompt_turn directly for string content and add a regression with an assistant turn followed by a plain-string user prompt.\n\n2. The billed usage rewrite does not reach the normal re-drive completion path. inal_response is step.response only when a hook returns a distinct replacement; SuspendedHookRunner deliberately returns None when the final object is the latest provider response. In that common case, the response in the done payload remains 
ull, the gateway forwards its held last-round response, and that response still reports last-round rather than summed usage. illed_usage is present beside it, but existing gateway behavior is not shown applying that field and the advertised invariant (‘final response usage is set to those totals’) is false. Please return a rewritten latest response for multi-round completion (without changing the no-replacement single-round contract), and add a test where the hook completes with the same latest response object and verifies the client-visible usage is cumulative.

chopratejas:
Thanks, both were real. Fixed in df902334.

**1. String follow-up prompts.** `prompt_turn` is now the single decision for both content shapes, and the string path uses it directly. The newest user turn is protected unless it answers a fenced shell command in the assistant turn before it, because then it is a text harness's tool observation (the reason for the old `i < first_assistant_index` guard). New tests:
- `test_follow_up_prompt_string_after_assistant_turn_is_verbatim_under_replay`: assistant reply, then a plain-string prompt, stays verbatim. Fails on the previous head.
- `test_text_harness_observation_text_block_still_compresses_under_replay`: the list-content twin of the existing string observation test, so both shapes decide the same way.

**2. Summed usage when the hook returns the latest response.** After one or more re-drives, `done` now returns the hook's replacement or, when the hook handed back the latest provider response (`replaced is None`), the response posted on that call, with `usage` rewritten to the totals. Turns with no re-drive keep `response: null`, so `test_no_search_call_means_done_with_null_response` is unchanged.
- `test_one_redrive_round_end_to_end` now uses the same-object hook path and asserts a non-null `response` and cumulative client-visible usage (720 prompt, 14 completion, 300 cached), where it previously skipped the check when `response` was null. Fails on the previous head.
- `test_registry_max_redrives_env_caps_the_loop` pinned `response: null` on the cap path. That is also a multi-round completion, so it now asserts the latest provider answer with usage summed over all three calls.
- `proxy.mdx` and `wiki/proxy.md` describe when `response` is null.

Local: 690 passed across the router, session, gateway, test_proxy and read-protection suites; ruff 0.16.3 and mypy clean.

PR #3620, opened 2026-09-16
fix(transforms): stop embedded JSON pre-pass short-circuiting HTML extraction

### LINKED ISSUE #3609: [BUG] Embedded JSON compression can short-circuit parent HTML compression

xiechimon:
## Description

The embedded-JSON pre-pass in `_apply_strategy_to_content` runs before any strategy dispatch and returns immediately on a successful splice. For a whole HTML document containing a JSON fragment, a small local minify pre-empts whole-document extraction: the report measured a 2% splice win blocking a 95.5% extraction (my reproduction: 12.4% blocking 45.9%). The parent format's stronger compressor never runs.

This defers the embedded-JSON route when the block's strategy is HTML: extraction gets first refusal, and the splice runs in the passthrough fallback only when extraction found nothing. This is option 3 from the issue ("only allow embedded JSON compression to short-circuit routing when there is no stronger parent-level strategy available"), scoped to the HTML case the report measured. Non-HTML strategies keep the immediate pre-pass untouched. Lossless mode never reaches the strategy dispatch, so it keeps the immediate-return behavior (JSON minification is itself byte-lossless).

While tracing the fallback I also fixed an adjacent hole in the same branch: `trafilatura` returns `''` (not `None`) when nothing extracts, and the HTML branch treated `''` as a successful compression, so an unextractable page blanked the whole block and only the empty-output guard in `compress()` caught it. Empty extraction now collapses to the passthrough path like `None` does, which is also what makes the deferred splice reachable for shell-only documents.

Closes #3609

## Type of Change

## Changes Made

- `headroom/transforms/content_router.py`: the embedded-JSON pre-pass body moved unchanged into `_embedded_json_route()`; HTML blocks skip the immediate splice and retry it in the passthrough fallback, appending `embedded_json` to the existing chain (`["html", "embedded_json"]`) so log readers can see the extraction miss. Empty HTML extraction (`''`) now falls through to passthrough instead of being accepted as a compression result.
- `tests/test_transforms_content_router.py`: two regression tests (extractor runs on HTML with embedded JSON; deferred splice still fires when extraction misses).
- `route_embedded_json` itself is unchanged; its unit tests pass untouched.

## Testing

### Test Output

```text
$ uv run pytest tests/test_transforms_content_router.py tests/test_recursive_json.py -q
71 passed in 5.78s

$ uv run ruff check .
All checks passed!

$ uv run mypy headroom/transforms/content_router.py
```

Both new tests fail on base `35b1156` (chain `["embedded_json"]`, `<html>` retained) and pass with the fix. I also ran every test file that imports `ContentRouter` (86 files, `grep -rl`): `1462 passed, 41 skipped, 0 failed`.

## Real Behavior Proof

- Environment: WSL2 Ubuntu, Python 3.11 via uv, headroom 0.37.1-dev from source (`uv sync --frozen --extra dev --extra html`, trafilatura 2.0.0), pure-Python detection (no ONNX runtime installed).
- Exact command / steps: a real HTML document (30 varied prose paragraphs x2 + one 60-row embedded JSON array in a single-line `<script>`, 19,297 chars) through `ContentRouter().compress()`, run with the patch and then with the patched file stashed (same interpreter):

```text
# WITH FIX
detection : ContentType.HTML
doc tokens: 4824 -> compressed tokens: 2610
strategy  : html  chain=['html']
reduction : 45.9%
<html> tag still present: False

# BASE 35b1156
detection : ContentType.HTML
doc tokens: 4824 -> compressed tokens: 4226
strategy  : html  chain=['embedded_json']
reduction : 12.4%
<html> tag still present: True
```

- Observed result: with the fix the extraction runs and the markup is gone; on base the JSON splice short-circuits and the full HTML document (markup included) survives.
- Not tested: proxy mode end-to-end (`/v1/compress` through a live LLM provider); Windows/macOS; ONNX native detection (the pure-Python fallback classifies these documents as HTML identically).

## Runtime Rollout Safety

- Rollout-managed feature(s): none — no flag/rollout system gates this path.
- Stable/default behavior changed: yes, for HTML-classified blocks containing routable embedded JSON: the strategy chain changes from `["embedded_json"]` to `["html"]` (extraction wins) or `["html", "embedded_json"]` (extraction misses). Everything else is byte-identical.
- Kill switch / disable path: `enable_html_extractor=false` routes HTML blocks to passthrough; the deferred splice then fires exactly as the pre-fix pre-pass did.
- Qualification impact: none.
- Rollback path: revert the commit.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (n/a — no user-facing docs cover route ordering)
- [x] I did **not** edit `CHANGELOG.md` — it is generated by release-please from my Conventional Commit PR title

## Additional Notes

A tiny local JSON win should never block the parent document's real compressor. Extraction is lossy and the splice is lossless, so deferring is a behavior trade (extraction output replaces markup+JSON entirely); the fallback keeps the splice whenever extraction has nothing to offer.

AI use: an AI agent located the short-circuit, ran the before/after reproductions and the test suite, and drafted this description. The fix approach follows option 3 suggested in the issue. I reviewed every line of the diff and the reproduced numbers.


JerrettDavis (approved):

JerrettDavis (approved):
Re-verified the revised head against current main: mergeable and conflict-free. HTML extraction now runs first, embedded JSON is computed lazily only on an HTML miss, real-content tests cover both the extraction and fallback paths, and the full current CI/security matrix is green. Ready for maintainer merge review.

JerrettDavis (approved):
Re-reviewed exact head 3e332694 after the clean rebase. The conflict-adapted router flow preserves HTML-first extraction, passthrough JSON fallback, strategy accounting, and lossless behavior; targeted regressions and Ruff pass locally, and all required checks are green.

PR #3627, opened 2026-09-17
fix(content-router): trust confident HTML detection over mixed-content heuristics

### LINKED ISSUE #3608: [BUG] Mixed-content routing can lose the parent HTML type

SyN-droMe:
## Description

A complete HTML document (clear DOCTYPE/`<html>` root) that also contains embedded JSON (e.g. a `<script type="application/ld+json">` block) or enough body prose trips `is_mixed_content`'s regex heuristics (`has_json_blocks`, `has_prose`), so it gets routed through the generic MIXED path (split section-by-section, each section compressed independently) instead of reaching the dedicated HTML extractor. The split JSON section can compress well on its own, but the surrounding markup never gets HTML-specific extraction, and the split sections lose the parent HTML type entirely.

`_determine_strategy` already has this exact override for `SOURCE_CODE` (a Python file with dict literals and docstrings trips the same heuristics, and the code trusts the native detector's confident SOURCE_CODE verdict over them). HTML never got the same override.

Closes #3608

## Type of Change

## Changes Made

- `headroom/transforms/content_router.py`: `_determine_strategy` now trusts a confident `ContentType.HTML` detection (`confidence >= 0.7`, matching the threshold `detect_content_type` already uses for its own internal HTML branch) over the `is_mixed_content` regex heuristics, same pattern as the existing `SOURCE_CODE` override immediately above it.
- `tests/test_transforms_content_router.py`: two new tests mirroring the existing `SOURCE_CODE` override test pair. One confirms the override fires and routes to `CompressionStrategy.HTML`; the other confirms a low-confidence HTML detection still falls back to `MIXED`.

## Testing

### Test Output

```text
$ uv run pytest tests/test_transforms_content_router.py -k "html_overridden or html_confidence_below or source_code_overridden or plain_text or confidence_below_threshold" -v
tests/test_transforms_content_router.py::test_mixed_false_positive_on_source_code_overridden_by_detector PASSED
tests/test_transforms_content_router.py::test_mixed_still_used_when_detector_says_plain_text PASSED
tests/test_transforms_content_router.py::test_mixed_still_used_when_detector_confidence_below_threshold PASSED
tests/test_transforms_content_router.py::test_mixed_false_positive_on_html_overridden_by_detector PASSED
tests/test_transforms_content_router.py::test_mixed_still_used_when_html_confidence_below_threshold PASSED
======================= 5 passed, 59 deselected in 2.66s =======================

$ uv run pytest tests/test_transforms_content_router.py tests/test_transforms/test_content_router.py \
    tests/test_mixed_content_sections.py tests/test_transforms/test_html_extractor.py \
    tests/test_compressor_selection.py tests/test_router_registry_dispatch.py \
    tests/test_lossless_first_dispatch.py tests/test_router_external_dispatch.py \
    tests/test_content_router_detection_dedup.py tests/test_content_router_token_units.py \
    tests/test_content_router_compact_json.py tests/test_content_router_exclude_tools.py \
    tests/test_content_router_user_blocks.py tests/test_content_router_single_item_deadline.py \
    tests/test_content_router_tool_role_reversibility.py tests/test_mixed_content_scan_cache.py \
    tests/test_evals/test_html_extraction_eval.py -q
562 passed, 3 skipped (network-dependent eval cases)

$ uv run ruff check headroom/transforms/content_router.py tests/test_transforms_content_router.py
All checks passed!

$ uv run ruff format --check headroom/transforms/content_router.py tests/test_transforms_content_router.py
2 files already formatted

$ uv run mypy headroom/transforms/content_router.py
```

Also confirmed the new tests fail without the fix and pass with it. I reverted just the source change (kept the new tests) and reran: `test_mixed_false_positive_on_html_overridden_by_detector` failed with `AssertionError: assert <CompressionStrategy.MIXED> is <CompressionStrategy.HTML>`.

## Real Behavior Proof

- Environment: WSL2/Ubuntu, Python 3.12.14, `headroom` installed editable via `uv sync --extra dev` + `uv pip install trafilatura` (the `html` extra), `HEADROOM_DETECT_BACKEND=python` (pure-Python detection backend, no native Magika/ONNX dependency needed here since `is_mixed_content` and the regex HTML detector are pure Python regardless of backend).
- Exact command / steps: ran a script that builds `ContentRouter()` with default config and calls `.compress()` on a synthetic HTML document (DOCTYPE, `<html>`/`<head>`/`<body>`, an embedded JSON-LD `<script>` block, and three paragraphs of body prose, matching the shape both issues describe), once against unfixed `content_router.py` and once against the fix.
- Observed result: before the fix, is_mixed_content is True, detect_content_type returns ContentType.HTML at confidence 1.00, strategy_used is CompressionStrategy.MIXED, and the document goes from 1623 to 1577 characters (2.8% reduction). After the fix, same input and same detection result, but strategy_used is CompressionStrategy.HTML and the document goes from 1623 to 703 characters (56.7% reduction). Only the routing decision changes between the two runs, which is exactly the bug both issues describe.
- Not tested: I don't have access to either reporter's actual page content, so I can't reproduce their exact 95.5%/99.5% numbers. My synthetic test document is smaller and plainer than a real scraped page, so 56.7% is a real but more modest number for this specific input. I also did not test the native Rust/Magika detection backend directly. I used the pure-Python backend, which `_determine_strategy`'s override logic treats identically since it only branches on `detection.content_type` and `detection.confidence`, independent of backend.

## Runtime Rollout Safety

- Rollout-managed feature(s): None. This is a routing-logic fix inside the existing `ContentRouter`, not behind a new flag.
- Stable/default behavior changed: Yes. An HTML document that previously routed through MIXED when it also triggered the mixed-content heuristics now routes through HTML strategy when detection confidence is >= 0.7. Same shape of behavior change as the existing SOURCE_CODE override.
- Kill switch / disable path: The existing `enable_html_extractor` config flag (default `True`) already gates whether the HTML strategy actually extracts once routed there. Setting it `False` falls through to passthrough for HTML-strategy content. The routing override itself has no separate toggle, matching the SOURCE_CODE override it mirrors, which also has none.
- Unsafe override required: None.
- Qualification impact: N/A
- Rollback path: Standard revert. The change is a single additive `if` branch with no state, schema, or migration involved.

## Review Readiness

## Checklist

- [x] I did **not** edit `CHANGELOG.md`, it is generated by release-please from my Conventional Commit PR title (a CI guard enforces this)

## Additional Notes

This is a distinct bug from #3609/#3612, but a prerequisite for that fix to matter on the documents both issues actually describe. #3612 patches `_apply_strategy_to_content`'s embedded-JSON pre-pass, but only defers it when `strategy == CompressionStrategy.HTML` is already selected, which never happens for a document that trips `is_mixed_content` (the exact shape #3608 and #3609 both describe) until this PR's routing fix lands, since such a document currently gets `CompressionStrategy.MIXED` instead and never reaches that branch. I left a comment on #3609 with the detail. Happy to rebase either PR on the other, whichever order review lands in.


SyN-droMe:
Hey @chopratejas @JerrettDavis @DevanshiVyas, just flagging this is ready for review whenever you get a chance, no rush. Happy to make any changes if something's off.

JerrettDavis (approved):
Verified the refreshed head against current main: mergeable and conflict-free. The HTML-confidence override is correctly scoped, and the full current CI/security matrix is green. Ready for maintainer merge review.

chopratejas:
Thanks — the routing bug you found is real, and I reproduced it: a DOCTYPE page with a JSON-LD block detects as `HTML @ confidence 1.0`, `is_mixed_content` says `True`, and `main` routes MIXED and the document **grows** (961 -> 963 bytes). Your fix routes it to HTML and gets 961 -> 517.

The problem is what happens after the reroute.

**1. Blocking: the structured data is silently deleted with no retrieval marker.** With this patch applied, that same page's JSON-LD payload is gone from the output — product ID, price, the `aggregateRating` block — and `CCR_RETRIEVAL_MARKER_RE` finds nothing. Under MIXED today that data survives verbatim. trafilatura keeps article prose and drops everything else, so this routes the exact document class it targets (a page whose payload is in a `<script>` tag) into a destructive path.

`ContentRouter.LOSSY_UNMARKED_STRATEGIES` (`content_router.py:1914-1920`) is `{KOMPRESS, TEXT, CODE_AWARE}` — **`HTML` is not in it** — so the #1307 tool-ground-truth reversibility guard at `:6023-6028` and `:7065-7070` never fires. An MCP or WebFetch tool result loses its structured data with no way to recover it.

HTML extraction is *already* unmarked-lossy on `main` for non-mixed HTML (I measured 700 -> 539, no marker), so this amplifies an existing gap rather than creating a new class — but it amplifies it precisely onto structured-data pages.

**Ask: add `CompressionStrategy.HTML` to `LOSSY_UNMARKED_STRATEGIES` in this PR**, plus a test pinning the JSON-LD case. That makes the guard skip the block when it would destroy unrecoverable tool ground truth, and I think the PR is then good to land.

**2. The `>= 0.7` confidence guard is vacuous.** `detect_content_type` already refuses to return HTML below 0.7 (`content_detector.py:195-197`; `_try_detect_html` returns `None` below 0.5 at `:444`), and the native backend hardcodes `confidence=1.0` for every result (`content_router.py:983-986`). So `detection.confidence >= 0.7` can never be `False` at this call site under either backend — the override is unconditional in practice. `test_mixed_still_used_when_html_confidence_below_threshold` only passes because it monkeypatches `_detect_content` into a state the real detector can't produce. Worth either removing the guard or making it real, because as written a reviewer reads it as "high-confidence only" when it isn't.

**3. Native-backend blast radius is wider than the PR describes.** With `HEADROOM_DETECT_BACKEND=rust`, a plain XML tool envelope (`<results><item>…`) detects as `ContentType.HTML @ 1.0`. So any XML-shaped MCP/tool payload that also trips `has_json_blocks`/`has_prose` now goes to trafilatura instead of MIXED. The "Real Behavior Proof" section used the pure-Python backend — the claim that the override treats both backends identically is true as written, but the native backend classifies a *larger* set of content as HTML.

**4. Minor:** without the `[html]` extra (trafilatura is optional, `pyproject.toml:280-281`), `_get_html_extractor()` returns `None` and this falls through to passthrough (`content_router.py:3773-3774`). Those installs go from MIXED split compression to no compression. Small in practice, but untested.

For what it's worth I ran a broad sweep on this branch (`pytest -k "html or mixed" tests/`): 412 passed, 8 skipped. Nothing fails — the gap is an *absent* test, not a failing one. Nothing in the suite asserts what the HTML extractor does to embedded structured data once routing can reach it, which is exactly where the payload disappears.

Also flagging that #3620 and #3612 are addressing overlapping ground here — might be worth the three of you settling on one approach before we land any of them.

PR #3628, opened 2026-09-17
fix(smart-crusher): keep non-dict arrays and object keys whole under lossless_only

### LINKED ISSUE #3625: [BUG] lossless_only is accepted but inert in 0.37.0 — SmartCrusher drops array items to max_items_after_crush with no CCR marker
## Summary

In 0.37.0, `SmartCrusherConfig.lossless_only` is accepted by both the Python wrapper and the PyO3 config, but has **no effect**: output is byte-identical with the flag on and off. On a JSON object whose value is an array, the crusher drops items down to `max_items_after_crush` (default 15) and emits **no CCR marker**, so a consumer has no way to detect that content was removed — and no configuration produces marker-free *and* complete output.

This is the guarantee #1091 was closed by introducing, and it is a different failure from #1478 (there, the Rust struct rejected the kwarg with a `TypeError`; here it accepts it and ignores it).

## Reproduce

```python
import json
from headroom._core import SmartCrusher, SmartCrusherConfig

data = json.dumps({"slugs": [f"r{i}" for i in range(53)]})
for cfg in (dict(lossless_only=True),
            dict(lossless_only=False),
            dict(lossless_only=True, enable_ccr_marker=True),
            dict(lossless_only=True, enable_ccr_marker=False)):
    out = SmartCrusher(SmartCrusherConfig(**cfg)).crush(data, "", 1.0).compressed
    print(f"{str(cfg):58} -> slugs {len(json.loads(out)['slugs']):2}/53  "
          f"ccr_marker={'<<ccr:' in out}")
```

```
{'lossless_only': True}                                    -> slugs 15/53  ccr_marker=False
{'lossless_only': False}                                   -> slugs 15/53  ccr_marker=False
{'lossless_only': True, 'enable_ccr_marker': True}         -> slugs 15/53  ccr_marker=False
{'lossless_only': True, 'enable_ccr_marker': False}        -> slugs 15/53  ccr_marker=False
```

The package's own Python wrapper behaves identically, so this is not a matter of bypassing it:

```
python wrapper lossless_only=True  slugs 53 -> 15  ccr_marker=False  bytes 372 -> 91
python wrapper lossless_only=False slugs 53 -> 15  ccr_marker=False  bytes 372 -> 91
python wrapper output identical for True/False: True

_core (PyO3)   lossless_only=True  slugs 53 -> 15  ccr_marker=False  bytes 372 -> 91
_core (PyO3)   lossless_only=False slugs 53 -> 15  ccr_marker=False  bytes 372 -> 91
_core (PyO3)   output identical for True/False: True
```

## Expected

Per the `crush()` docstring in `headroom/transforms/smart_crusher.py`, with `lossless_only=True` "rows are never dropped and opaque cells render inline", and the output is byte-recoverable. Either that, or the row-drop path should emit the CCR marker that `enable_ccr_marker=True` promises — the current behaviour offers neither.

## Impact (measured)

We run an LLM job scheduler that compresses large command stdout before it reaches the model, using the native primitives directly (`lossless_only=True`, CCR disabled — we never want a retrieval round-trip). Because we did not trust the flag, every candidate is validated by an independent JSON-equivalence check before it is allowed to replace the original, and rejected results fall back to the uncompressed output.

Over one week across ~30 agents: **260 compressions were rejected by that check. Of the 121 whose original was still recoverable, 121 were genuine data loss** — dropped array elements (33 → 15, i.e. `max_items_after_crush`), dropped object keys (8, 15 and 35 at a time), and in some cases a list replaced by a string. None carried a marker. Zero were benign reformatting (whitespace, key order, equivalent escaping).

Without that external check, those would have been silently truncated tool results presented to a model as complete — the failure mode #1091 describes, minus the marker that at least made it visible.

For calibration on the same payloads: `json.dumps(obj, separators=(",", ":"))` was faithful on 335/335 and saved roughly 5x more estimated tokens than the crusher's accepted results, at ~0.5 ms per payload versus ~250 ms.

## Environment

- headroom-ai 0.37.0 (wheel), Python 3.12.14, Linux x86_64 (glibc 2.36), Docker
- `SmartCrusherConfig.__text_signature__` contains `lossless_only=False`, so the field is accepted, not rejected

## Suggested direction

Either gate the row-drop path on `lossless_only` in the Rust crusher (and gate the opaque-blob path as #1091 asked), or — if the guarantee cannot be honoured — reject the config at construction rather than accepting a flag that does nothing. A silently inert safety flag is worse than an absent one: it reads as a guarantee in every call site that sets it.

SulimanAbdulrazzaq:
## Description

`lossless_only` is documented as producing marker-free, byte-recoverable output where rows are never dropped. Only the dict-array row-drop path (`crush_array_with_source`) checked it, though. The string, number and mixed-array crushers and the object-key crusher kept sampling in strict mode, and they drop items without emitting any CCR marker. So `{"slugs": [53 strings]}` came back with 15 slugs whether `lossless_only` was on or off, and nothing in the output showed that items were removed.

Each of those four crushers now returns its input unchanged under `lossless_only`, as `crush_array_with_source` already does for dict arrays. Behavior with `lossless_only` off is unchanged.

Closes #3625

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `crushers.rs`: `crush_string_array`, `crush_number_array` and `crush_object` return their input under `lossless_only`, with strategies `string:lossless_only`, `number:lossless_only` and `object:lossless_only`. Each check comes after the crusher's existing passthrough checks.
- `crusher.rs`: `crush_mixed_array` returns its items under `lossless_only` (`mixed:lossless_only`).
- `config.rs`: the `lossless_only` doc comment now names these crushers.
- Tests:
  - three unit tests in `crushers.rs`;
  - two in `crusher.rs`, one of them end to end through `crush()` using the #3625 payload;
  - one Python test through the public `SmartCrusher` wrapper.

  Each Rust test first asserts that the default config really drops items on the same input, so the strict assertion cannot pass vacuously.

## Testing

### Test Output

```text
# Before the fix (origin/main 2c56a1b3 plus these tests only)
$ cargo test -p headroom-core --lib lossless_only
test result: FAILED. 3 passed; 5 failed; 0 ignored; 0 measured; 922 filtered out
$ pytest tests/test_transforms/test_smart_crusher_bugs.py -q
1 failed, 6 passed

# With the fix
$ cargo fmt --all -- --check                                  # exit 0
$ cargo clippy -p headroom-core --all-targets -- -D warnings   # exit 0
$ cargo test -p headroom-core --lib lossless_only
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 922 filtered out
$ cargo test -p headroom-core --lib smart_crusher
test result: ok. 334 passed; 0 failed; 0 ignored; 0 measured; 596 filtered out
$ cargo test -p headroom-core
test result: ok. 929 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out   (lib; every integration test binary also ok)
$ cargo run -p headroom-parity -- run --fixtures tests/parity/fixtures
[smart_crusher   ] total=17 matched=17 skipped=0 diffed=0
$ pytest (the file list from make ci-precheck-python) -q
176 passed, 4 skipped, 1 warning
$ pytest tests/test_lossless_mode.py tests/test_lossless_then_lossy.py tests/test_agent_savings.py tests/test_transforms_config_compressor.py -q
121 passed
$ ruff check .              # All checks passed!
$ ruff format --check .     # 1605 files already formatted
```

## Real Behavior Proof

- Environment: Linux x86_64, Python 3.14.2, Rust 1.95.0, headroom-ai 0.37.0 editable install built with `uv sync --frozen --extra dev`, which builds the Rust core via maturin.
- Exact command / steps: ran the #3625 reproduction script unchanged, then the same payload through the `SmartCrusher` wrapper and through `ContentRouter`, first on `origin/main` and then on this branch after `uv sync --frozen --extra dev --reinstall-package headroom-ai`. In full:
  1. Ran the reproduction script from #3625 unchanged.
  2. Sent the payload below through `SmartCrusher(config=SmartCrusherConfig(lossless_only=True)).crush(json.dumps(doc))`.
  3. Sent it through `ContentRouter(ContentRouterConfig(smart_crusher_lossless_only=True)).compress(json.dumps(doc), context="")`, which is the config `HEADROOM_LOSSLESS_ONLY=1` sets.

  ```python
  doc = {
      "slugs": [f"r{i}" for i in range(53)],
      "sizes": list(range(1, 41)),
      "mixed": [f"entry-{i}" if i % 2 == 0 else i for i in range(40)],
      "meta": {f"k{i:02d}": f"long description for entry {i}, above the small-value floor" for i in range(40)},
  }
  ```
- Observed result: on `origin/main`, strict mode returned 15 of 53 slugs and no CCR marker; on this branch it returns all 53 and the output decodes back to the input. Raw output:
  ```text
  # origin/main
  {'lossless_only': True}                                    -> slugs 15/53  ccr_marker=False
  {'lossless_only': False}                                   -> slugs 15/53  ccr_marker=False
  {'lossless_only': True, 'enable_ccr_marker': True}         -> slugs 15/53  ccr_marker=False
  {'lossless_only': True, 'enable_ccr_marker': False}        -> slugs 15/53  ccr_marker=False
  SmartCrusher lossless_only=True:                  slugs 15/53, sizes 15/40, mixed 21/40, meta keys 15/40 | equal to input: False | marker: False
  ContentRouter(smart_crusher_lossless_only=True):  slugs 15/53, sizes 15/40, mixed 21/40, meta keys 15/40 | equal to input: False | marker: False

  # this branch
  {'lossless_only': True}                                    -> slugs 53/53  ccr_marker=False
  {'lossless_only': False}                                   -> slugs 15/53  ccr_marker=False
  {'lossless_only': True, 'enable_ccr_marker': True}         -> slugs 53/53  ccr_marker=False
  {'lossless_only': True, 'enable_ccr_marker': False}        -> slugs 53/53  ccr_marker=False
  SmartCrusher lossless_only=True:                  slugs 53/53, sizes 40/40, mixed 40/40, meta keys 40/40 | equal to input: True | marker: False
  ContentRouter(smart_crusher_lossless_only=True):  slugs 53/53, sizes 40/40, mixed 40/40, meta keys 40/40 | equal to input: True | marker: False
  ```
  `ContentRouter(ContentRouterConfig(lossless=True))`, which is what `--lossless` builds, returned the input intact both before and after, because that mode stops at the byte-exact fold.
- Not tested: a running `headroom proxy` process (the router ran in-process with the same config), macOS and Windows, the full `pytest` suite, and `cargo test` / `cargo clippy` for the whole workspace. Only `headroom-core` changed, and the parity harness was run.

## Runtime Rollout Safety

- Stable/default behavior changed: no. Only `lossless_only=True` (for example `HEADROOM_LOSSLESS_ONLY=1`) changes: string, number and mixed arrays and wide objects are no longer truncated.
- Kill switch / disable path: leave `lossless_only` off, or unset `HEADROOM_LOSSLESS_ONLY`.
- Qualification impact: strict lossless mode compresses these JSON shapes less, because they now pass through intact.

## Review Readiness

## Checklist

N/A

## Additional Notes

- The unit-test and "existing tests" boxes are unchecked because only the file sets above ran, not the full suite. mypy was not run because no module under `headroom/` changed. No user docs mention `lossless_only`, so only the Rust doc comment was updated.
- With `lossless_only` off, behavior is unchanged: these crushers still drop items without a CCR marker. Emitting a marker there is the other option #3625 raises, and it needs a separate decision.
- Under `lossless_only`, dict arrays can still be rendered as the lossless CSV-plus-schema table string when compaction wins. That is the existing documented compaction, not a drop.
- This doesn't overlap with #3401. The two PRs edit the shared test file far apart, and the new test does not use `max_items_after_crush`.


SulimanAbdulrazzaq:
The red `audit` job is pre-existing on `main`, not from this PR: `cargo audit` reports RUSTSEC-2026-0285 (rustls 0.23.41, advisory published 2026-09-14, fixed in >= 0.23.45). `Cargo.lock` still pins 0.23.41 on `main`, and this branch does not touch it.

Today's scheduled `rust` run on `main` failed the same way, with every other job green: https://github.com/headroomlabs-ai/headroom/actions/runs/35194899475. Recent `main` pushes skip the job entirely, because no Rust files changed in them.

Everything else on this PR is green: 41 checks pass, including `test (ubuntu)`, `parity`, all three `wheels` targets and all three `simulator e2e` runners.

PR #3630, opened 2026-09-17
fix(kompress): bound download retry backoff before overflow

### LINKED ISSUE #3629: [BUG] Kompress retry overflow causes /v1/compress to return 503
We're seeing `/v1/compress` return **503** after repeated background Kompress loading failures on **Headroom 0.35.0, Python 3.13, Linux/Kubernetes**.

The service traceback points to:

```text
kompress_compressor.py:1089, _download_retry_blocked
    _DOWNLOAD_RETRY_BASE_SECONDS * (2 ** (failures - 1))
OverflowError: int too large to convert to float
```

The delay is meant to stop growing at 300 seconds, but the exponential calculation happens before `min()` applies that cap. It overflows when the failure count reaches **1,025**.

### Minimal reproduction

```python
for failures in (1024, 1025):
    try:
        print(failures, min(300.0, 5.0 * (2 ** (failures - 1))))
    except OverflowError as exc:
        print(failures, type(exc).__name__, str(exc))
```

```text
1024 300.0
1025 OverflowError int too large to convert to float
```

In our setup, offline mode is enabled and the ModernBERT tokenizer is missing from the cache. That is a separate deployment problem, but repeated loading failures should keep a capped retry delay rather than crash compression. Other compression paths continued working, so this only affected some requests.

Could the backoff be capped before calculating the exponent? [Affected code](https://github.com/headroomlabs-ai/headroom/blob/v0.35.0/headroom/transforms/kompress_compressor.py#L1078-L1091).

prakx1:
## Description

Fixes #3629.

After 1,025 consecutive Kompress loading failures, the retry calculation raises `OverflowError` before its five-minute cap can apply. Double the delay only until it reaches the cap, preserving the existing 5, 10, 20, 40, 80, 160, 300-second schedule.

## Type of Change

- [x] Bug fix (non-breaking)

## Changes Made

- Bound the calculation without changing retry timing or adding dependencies.
- Cover the normal schedule, 1,024/1,025 failures, a very large failure count, retry at expiry, and clearing the backoff after success.

## Testing

- [x] Unit tests pass (`pytest`) — full suite not completed; passing subsets are listed below.
- [x] Type checking passes (`mypy headroom`) — baseline error described below.

The regression failed before the fix with `OverflowError: int too large to convert to float` and passes afterward.

### Test Output

```sh
HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 HEADROOM_BEACON=off uv run --no-sync pytest -q \
  tests/test_kompress_download_backoff.py \
  tests/test_kompress_request_nonblocking.py \
  tests/test_kompress_failsafe.py \
  tests/test_kompress_preload_deferral.py \
  tests/test_transforms/test_kompress_compressor.py \
  tests/test_proxy/test_kompress_background_warmup.py
uv run --no-sync ruff check .
uv run --no-sync ruff format --check .
```

```text
113 passed in 58.30s
All checks passed!
1607 files already formatted
```

Pre-commit and commit-message hooks passed. All `make ci-precheck` gates passed after installing `pip` in the isolated virtual environment: Rust formatting/Clippy/workspace tests, the selected Python suite (175 passed, 4 skipped), and commitlint.

The installed mypy 1.19.1 reports one unrelated error at `headroom/integrations/langchain/chat_model.py:394`; the same error reproduces on unmodified main (`aceff2e`). The hook's pinned mypy 1.14.1 passes.

I started the full Python suite (13,110 collected tests) with `uv run --no-sync pytest --maxfail=1 -q`, then stopped it during unrelated CLI tests invoking external Serena indexing. The interruption surfaced as exit 143 in `test_wrap_codex_prepare_only_updates_stale_mcp_proxy_url` (2,041 passed, 111 skipped). This is **not** a full-suite pass; leaving the PR draft pending broader CI verification.

## Real Behavior Proof

The live failure traceback and 503 response are documented in #3629.

- Environment: macOS arm64, Python 3.13.13, source checkout installed with `uv sync --extra dev`; offline Kompress (`chopratejas/kompress-v2-base`), no model loaded or upstream LLM call.
- Exact command / steps: Run the complete offline probe below from the checkout. It seeds 1,025 failures, calls the real `ContentRouter.compress()` and `KompressCompressor.ensure_background_load()`, and checks passthrough and download-thread state.
- Observed result: Content was preserved, the compression strategy was `kompress`, and the background-load hook returned without overflow or starting a download thread. Output is included below.
- Not tested: Production rollout, full HTTP/LiteLLM end-to-end traffic, successful model inference, other operating systems, and completion of the full Python test suite.

The failure count is seeded to reproduce a long-running failure state without waiting days; the router and compressor are real objects, not mocks.

<details>
<summary>Exact runtime probe</summary>

```sh
env -u HEADROOM_KOMPRESS_ENDPOINT HEADROOM_OFFLINE=1 HF_HUB_OFFLINE=1 \
  TRANSFORMERS_OFFLINE=1 HEADROOM_BEACON=off .venv/bin/python - <<'PY'
import time
from headroom.transforms import kompress_compressor as kc

model = kc.HF_MODEL_ID
assert model not in kc._kompress_cache
with kc._download_threads_lock:
    kc._download_failures[model] = (1025, time.monotonic())
router = ContentRouter(ContentRouterConfig(enable_kompress=True, relevance_split=False))
router._kompress = kc.KompressCompressor()
router._runtime_force_kompress = True
payload = ' '.join(f'Synthetic sentence {i} about network retries and safe fallback behavior.' for i in range(40))
result = router.compress(payload)
assert result.compressed == payload
print('router_strategy:', result.strategy_used.value)
print('content_preserved:', result.compressed == payload)
router._kompress.ensure_background_load()
assert not kc._download_threads.get(model)
print('background_load_returned_without_overflow: True')
print('download_thread_started: False')
PY
```

</details>

Observed after the fix:

```text
router_strategy: kompress
content_preserved: True
background_load_returned_without_overflow: True
download_thread_started: False
```

Calling the real `ensure_background_load()` with the same seeded failure state raised the reported overflow before the fix. The expected model-not-ready warning remains; this fixes the retry crash, not underlying model-loading failures.

## Runtime Rollout Safety

- Rollout-managed feature(s): Existing Kompress background model loading only; no new rollout-managed feature or flag.
- Minimum rollout channel: Unchanged; this fix adds no new channel requirement and does not change existing Kompress availability.
- Stable/default behavior changed: Failed downloads no longer overflow while calculating backoff. The existing 5-to-300-second retry schedule and default configuration are unchanged.
- Kill switch / disable path: Existing `HEADROOM_DISABLE_KOMPRESS=1` or `--disable-kompress` disables Kompress ML loading while retaining structural compression; per-provider force-enable overrides must be unset. No new switch is introduced.
- Qualification impact: Regression coverage for retry timing, saturation, and recovery; no change to compression algorithms or model selection. No new model-quality qualification is claimed.
- Rollback path: Revert commit `59d174b` and rebuild/redeploy the previously used release; no database or configuration migration is involved.

## Review Readiness

## Checklist

- [x] Added a failing-before/passing-after regression test.
- [x] Followed existing style and commented the overflow guard.
- [x] No `CHANGELOG.md` edits.
- [x] No public API/configuration documentation changes needed.

AI-assisted implementation and verification.


JerrettDavis (approved):
Verified the refreshed head against current main: mergeable and conflict-free. The bounded retry loop avoids exponent overflow even for enormous failure counts, and the full current CI/security matrix is green. Ready for maintainer merge review.

PR #3635, opened 2026-09-17
feat(compression): describe what CCR compression dropped, not just how much

SANGITA-D:
## Description

When the log compressor drops lines, it leaves behind a CCR marker: `[N lines compressed to M. Retrieve more: hash=...]`. Today that marker says how much was cut but nothing about *what*, so an agent deciding whether to call `headroom_retrieve` is guessing. This PR adds a short descriptor of the dropped content's error types and source files, when they're extractable:

```
Before: [135 lines compressed to 50. Retrieve more: hash=abc123]
After:  [135 lines compressed to 50: 1 file, 3 exception types (E0308, E0425, E0599). Retrieve more: hash=abc123]
```

An agent chasing a `KeyError` that sees the marker only mentions Rust compiler codes now has a real reason to skip retrieval, and vice versa when its target error is named.

I don't have a linked issue for this — I couldn't find a prior discussion of the marker's information content, so I'm opening this as a concrete proposal with data behind it rather than a spec-first change. Happy to split this into an issue first if that's preferred, or to adjust the approach based on maintainer feedback.

## Type of Change

## Changes Made

- Add `extract_error_label` to `crates/headroom-core/src/transforms/log_compressor.rs`: pulls a rustc code (`E0308`) or Python exception name (`KeyError`) out of a line that looks like an error.
- Add `extract_source_file`: same idea for file paths (Python `File "..."` lines, rustc `--> file:line:col`).
- Add `summarize_omitted`: runs both extractors over every line about to be dropped and builds the descriptor appended to the marker, capped at 5 names shown before falling back to `...`.
- Descriptor only covers what's absent from the surviving text — an error type that also appears in a kept line isn't repeated, since the descriptor's job is to justify retrieval, not restate what's already visible.
- Additive and backward-compatible: if nothing is extractable, the descriptor is empty and the marker is byte-identical to today's output.
- 2 new unit tests covering descriptor presence and the empty-descriptor fallback.

## Testing

- [x] Unit tests pass (`cargo test -p headroom-core log_compressor`)
- [x] Linting passes (`cargo clippy -p headroom-core --all-targets -- -D warnings`)
- [x] Manual testing performed (real captured cargo failure, not just unit fixtures)

### Test Output

```text
$ cargo test -p headroom-core log_compressor
running 26 tests
test transforms::log_compressor::tests::ccr_marker_emitted_when_thresholds_clear ... ok
test transforms::log_compressor::tests::ccr_marker_descriptor_empty_when_nothing_extractable ... ok
test transforms::log_compressor::tests::ccr_marker_includes_error_types_and_files_when_available ... ok
... (23 more, all pre-existing)
test result: ok. 26 passed; 0 failed; 0 ignored; 0 measured; 901 filtered out

$ cargo fmt -p headroom-core -- --check   # clean, no diff
$ cargo clippy -p headroom-core --all-targets -- -D warnings   # clean, no warnings
```

## Real Behavior Proof

- Environment: macOS, local Rust workspace build, cargo test/fmt/clippy against current `main`.
- Exact command / steps: `cargo test -p headroom-core log_compressor`; `cargo fmt -p headroom-core -- --check`; `cargo clippy -p headroom-core --all-targets -- -D warnings`; also ran the compressor against a real captured `cargo build` failure log to confirm the new marker text renders correctly outside the unit tests.
- Observed result: 26/26 tests pass, fmt clean, clippy clean, real-log marker output matches the documented format.
- Separately, to check whether the descriptor actually changes an agent's behavior (not just that it compiles), I ran a blinded evaluation described below — not part of this repo's test suite, offered as evidence for the design decision.
- Not tested: log formats other than Python test output and Rust/cargo output at this evaluation's scale; non-Claude judge models; end-to-end multi-turn task success (this measures one isolated retrieval decision).

### Evaluation (design justification, not a repo test)

62 real Python test-failure logs (genuine bugs, actually run, not hand-crafted) produced 124 log/question pairs — for each log: how many distinct files, and how many distinct error types, can you be certain are present? Each pair was shown twice, once per marker version, to Haiku, Sonnet, and Opus independently — 744 blinded runs, no memory across calls. A deterministic rubric scored answers: 0 for wrong/overclaimed, 0.5 for an unnecessary hedge, 1 for a defensible correct answer.

```
model     old marker   new marker   gap
Haiku     0.690        0.952        +0.262
Sonnet    0.786        0.980        +0.194
Opus      0.911        1.000        +0.089
```

Every model improves; the gain is largest for the weakest model. Haiku hedged on 77/124 questions with the old marker, dropping to 12/124 with the new one — this matters because agent harnesses increasingly default to cheaper, smaller models to control cost and latency, so the marker hands the most help to the model that needs it most.

Limits: one log format at this scale (others the log compressor handles weren't checked here); all three judges are Claude-family models; measures one retrieval decision, not full task success. Full harness, corpus, and per-call transcripts are available on request — left out of this diff to keep it focused on the code change.

## Runtime Rollout Safety

- Rollout-managed feature(s): none — no new flag, always active wherever the CCR marker is already emitted.
- Stable/default behavior changed: marker text gains an optional descriptor clause; unchanged when nothing is extractable (byte-identical to the current marker in that case).
- Kill switch / disable path: none needed since there's no behavior change when extraction finds nothing; reverting this commit fully removes the feature.
- Qualification impact: none — purely additive text in an existing marker, no change to compression decisions, byte counts, or the retrieval mechanism itself.

## Review Readiness

## Checklist

- [ ] I have made corresponding changes to the documentation (no user-facing docs currently describe the marker's format in detail; happy to add if maintainers point me at the right place)


JerrettDavis (approved):
Verified the refreshed head against current main: mergeable and conflict-free. The CCR descriptor remains bounded and deterministic, exposes counts rather than paths, and its marker remains retrievable through the generic 24-hex fallback; the full current CI/security matrix is green. Ready for maintainer merge review.

crates/headroom-core/src/transforms/log_compressor.rs
```diff
@@ -896,6 +896,123 @@ fn is_summary_line(line: &str) -> bool {
...
+    let kept: BTreeSet<usize> = selected.iter().map(|l| l.line_number).collect();
+
+    let mut error_types: BTreeSet<String> = BTreeSet::new();
+    let mut files: BTreeSet<String> = BTreeSet::new();
+    for line in all_lines {
+        if kept.contains(&line.line_number) {
+            continue;
+        }
+        if let Some(label) = extract_error_label(&line.content) {
+            error_types.insert(label);
```
upramod, line 973: The description says error types already visible in kept lines are left out of the descriptor, but this only filters by line number. A KeyError in both kept and dropped lines would still appear in the marker. Should we exclude the kept types too, or is the marker meant to describe all dropped content? A test covering that overlap would help.

SANGITA-D: Good catch, thanks. You're right — `kept` was only used to skip lines by `line_number`, so it never checked whether the same error type/file was also extractable from a *different* line that survived. A `KeyError` in both a kept and a dropped line would indeed show up in the descriptor, which contradicts what I wrote it was doing.

Fixed in `e7c81b97`: I now build the set of labels/files extractable from the *kept* lines first, and exclude those from what gets collected off the dropped lines. Also added a direct unit test for `summarize_omitted` (`summarize_omitted_excludes_error_type_and_file_visible_in_kept_lines`) covering exactly this overlap — a value present in both a kept and a dropped line is now omitted from the descriptor, while a value present only in dropped lines still shows up.

upramod (approved):

JerrettDavis (approved):
Re-reviewed exact head b1fd0b83 after the clean rebase. The cumulative contributor patch is unchanged, focused log-compressor tests, formatting, and clippy pass locally, and all required checks are green.

PR #3606, opened 2026-09-15
fix(memory): honor project root override for Anthropic CCR

axelray-dev:
## Description

The documented `--memory-project-root` override is ignored by Anthropic CCR workspace resolution, so clients without cwd metadata remain unresolved. This change makes the configured project root available to that resolver.

Closes #3596

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Read `memory_project_root_override` from the proxy configuration when resolving an Anthropic CCR workspace.
- Add regression coverage for a configured project root and update resolver tests for the instance method.

## Testing

- [x] Unit tests pass (`pytest`)
- [ ] Linting passes (`ruff check .`)
- [ ] Type checking passes (`mypy headroom`)
- [x] New tests added for new functionality
- [ ] Manual testing performed

### Test Output

```text
UV_CACHE_DIR=$PWD/.uv-cache uv run --no-sync pytest tests/test_proxy_handler_helpers.py -q
54 passed in 6.70s

UV_CACHE_DIR=$PWD/.uv-cache uv run --no-sync pytest tests/test_memory_storage_router.py -q
28 passed in 0.30s

UV_CACHE_DIR=$PWD/.uv-cache uv run --no-sync ruff format --check headroom/proxy/handlers/anthropic.py tests/test_proxy_handler_helpers.py
2 files already formatted

UV_CACHE_DIR=$PWD/.uv-cache uv run --no-sync ruff check headroom/proxy/handlers/anthropic.py tests/test_proxy_handler_helpers.py
All checks passed!
```

## Real Behavior Proof

- Environment: Linux, CPython 3.11.16, checkout commit 8df4b38aa14c213ae617d9b31da2ca4b0dd5dc5a.
- Exact command / steps: `UV_CACHE_DIR=$PWD/.uv-cache uv run --no-sync pytest tests/test_proxy_handler_helpers.py -k configured_project_root_override -q`; the test invokes the resolver with an override and no project headers or cwd system-prompt line.
- Observed result: The configured override resolved to a stable `project-c-` key prefix and the `project-c` workspace label. A separate direct invocation returned `headroom-project-2a55b0efd88cd77d` and `headroom-project`.
[...]

Issue #3596, opened 2026-09-15
_resolve_ccr_workspace hardcodes project_root_override=None, making --memory-project-root / HEADROOM_MEMORY_PROJECT_ROOT a dead workaround


DominiciLuca:
## Summary

The CLI documents `--memory-project-root` / `HEADROOM_MEMORY_PROJECT_ROOT`
(`cli/proxy.py:617-624`) as the escape hatch for clients that don't put a `cwd:`-style
line in the system prompt. But `_resolve_ccr_workspace` (`proxy/handlers/anthropic.py:80-86`)
hardcodes the value it's supposed to read from:

```python
ctx = _CtxFor(
    headers=dict(request.headers),
    system_prompt=_extract_sys_prompt(body),
    base_user_id=request.headers.get("x-headroom-user-id", ""),
    project_root_override=None,          # <-- hardcoded
)
ident = ProjectResolver().resolve(ctx)
```

`storage_router.py:149-185`'s resolution order puts `ctx.project_root_override` as tier 3
(after the two headers, before the system-prompt scan) — but since it's always `None` here,
setting `--memory-project-root` / `HEADROOM_MEMORY_PROJECT_ROOT` has zero effect on CCR
workspace resolution. It presumably still affects the memory subsystem elsewhere, but not
the code path this option is documented to fix.

## Effect

Users following the documented workaround for "my client doesn't send a cwd in the system
prompt" get no relief: `CCR: workspace unresolved` keeps firing regardless of the env var
or flag.

## Suggested fix

Thread the actual configured project-root value into `_CtxFor(project_root_override=...)`
instead of the literal `None`.

Env: headroom-ai 0.28.0 (pipx), macOS.

PR #3644, opened 2026-09-17
fix(memory): make extract_system_prompt concatenative, not first-match

### LINKED ISSUE #3595: CCR workspace resolver misses Claude Code 2.x env block (now sent as isMeta user message, not body["system"])
## Summary

`extract_system_prompt` in `headroom/memory/storage_router.py:408-451` only scans
`body["system"]` (string or content-block list) and OpenAI-style `messages[].role == "system"`
for the `cwd:`-family prefixes defined in `_CWD_PREFIXES` (`storage_router.py:46-50`):

```python
_CWD_PREFIXES: tuple[str, ...] = (
    "Primary working directory:",  # Claude Code (current)
    "Working directory:",          # Claude Code (older) / Codex
    "cwd:",                        # Generic / debug format
)
```

Claude Code 2.1.272 no longer places this block in `system`. It now ships the
`# Environment` block (which contains `Primary working directory: ...`) as a **user-turn
meta message** (`isMeta: true`), consistent with the `mid-conversation-system-2026-04-07`
beta the client advertises via `anthropic-beta`. The string headroom greps for is present
in the request — just in `messages[]`, not in `system`.

## Effect

`_resolve_ccr_workspace` (`proxy/handlers/anthropic.py:1533-1563`) always fails to resolve
a workspace for Claude Code clients, so `proxy.log` floods with:

```
CCR: workspace unresolved; skipping track_compression (fail-closed — no x-headroom-cwd /
x-headroom-project-id header and no cwd: in system prompt)
```

on every single request, and per-workspace `track_compression` / proactive expansion never
fires for any Claude Code user who hasn't manually set `x-headroom-cwd` /
`x-headroom-project-id` headers.

## Suggested fix

`extract_system_prompt` should also scan `messages[].content[].text` (or wherever
`isMeta: true` meta blocks land in the parsed body) for the same `_CWD_PREFIXES`, not just
`body["system"]`.

## Workaround (client-side, in case it helps others)

Setting `x-headroom-cwd` / `x-headroom-project-id` explicitly via
`ANTHROPIC_CUSTOM_HEADERS` works around this, but shouldn't be necessary given the prefix
table already anticipates Claude Code's current wording.

Env: headroom-ai 0.28.0 (pipx), macOS, Claude Code 2.1.272.

chopratejas:
Closes #3595

## Problem

`extract_system_prompt` was first-match-wins: it returned as soon as the top-level `system` field had content, before ever looking at `messages[]`. Claude Code always sends a non-empty `system`, so the `messages[]` scan — including the cwd-in-user-message tier added for Trae in #1737 — was unreachable for every Claude Code request.

Claude Code 2.x moved its `<env>` block (which carries `Primary working directory:`) out of `system` and into an `isMeta` user message. With the scan below `system` dead, that block is never seen, and CCR/memory workspace resolution falls through to the fail-closed empty scope — for the default configuration, with no header workaround needed.

Reproduced against the exact reported shape:

```python
body = {
    "system": "You are Claude Code.",
    "messages": [{"role": "user", "content": [{"type": "text", "text": "<env>\nPrimary working directory: /work/myproj\n</env>"}]}],
}
extract_system_prompt(body)   # before: "You are Claude Code."  (cwd invisible)
                               # after:  "You are Claude Code.\n<env>\n...\n</env>"
```

## Fix

Every source (`system` field, `role="system"` messages, and the user-message cwd fallback) is now scanned and **concatenated** instead of short-circuiting.

The existing anti-spoofing precedence is preserved exactly: a `cwd:`-family line already present in a trusted source (`system` field or a `role="system"` message) still wins over one appearing in a user message, so a user turn can't redirect resolution at another project's memory store by pasting its own `cwd:` line. Covered by `test_extract_system_prompt_user_cwd_cannot_override_system_field_cwd`.

## Testing

- 3 new regressions in `tests/test_memory_storage_router.py`, covering: the exact CC 2.x isMeta-user-message shape (`system` as string), the same with `system` as a block list, and the anti-spoofing precedence guard.
- Full existing suite in that file (31 → 36 tests) passes unchanged.
- `ruff check` / `ruff format --check` clean.

Scope note: this fixes the one tier that needs no client cooperation. #3597 (header namespace mismatch) and #3596 (`project_root_override` hardcoded to `None` for CCR) are separate, already-tracked breaks in the same resolution chain — see #3603 and #3606.


JerrettDavis (approved):
Verified the updated head against current main: mergeable and conflict-free, the implementation review found no blocking issues, and the full current CI/security matrix is green. Ready for maintainer merge review.

chopratejas:
Holding this one — review found a prompt-injection path into memory routing, plus a hot-path cost on the non-Anthropic harnesses.

**The premise is sound.** Confirmed on `main`: the `system` field early-return at `storage_router.py:424-488` does make the `messages[]` scan dead for any client that sends a system prompt. And the parent concern about a longer concatenated string perturbing things downstream does **not** materialize — `RequestContext.system_prompt` has exactly one consumer, `_extract_cwd_from_system_prompt` (`storage_router.py:187`). Every construction site (`anthropic.py:384`, `anthropic.py:1294`, `openai.py:3499`, `openai.py:5627`, `gemini.py:373`) was checked: no cache key, no hash, no token budget, no prompt-prefix matcher reads it.

**1. Blocking: the anti-spoofing guard doesn't cover the case this PR creates.** The guard at `storage_router.py:511-513` only suppresses user content when the trusted text *already contains* a cwd prefix. In the case this PR targets — trusted prompt has no cwd line — **any** user text containing `cwd:` wins:

```python
body = {"system": "You are a helpful assistant.",
        "messages": [{"role":"user","content":"here is a log I pasted:\n  cwd: /tmp/attacker-project\nplease debug"}]}

main    -> ProjectResolver.resolve(...) == None          # fail-closed
PR 3644 -> ('attacker-project-7b20313994a2ae60', 'attacker-project')
```

A pasted shell transcript silently re-keys the memory store and the CCR workspace. `unresolved_project_fallback="empty"` no longer protects this path. The tier existed since #1737 but was unreachable for every client sending a system prompt; this makes it reachable for all of them. `_first_user_text_with_cwd` (`:459`) also scans *all* user messages, not just the first, so a mid-conversation paste qualifies.

Fix: require the `isMeta` flag or an enclosing `<env>`/`<environment>` block, and restrict to the first user message.

**2. The trust guard's substring test makes the fix no-op on some clients.** `any(prefix in trusted ...)` at `:512` matches the 4-char `"cwd:"` anywhere in the system prompt — and tool descriptions do mention it. Verified: `system="Tool: bash. Reports cwd: at start."` plus a real env block resolves to `('at-start-f35c33cc787c6d15', 'at start.')` on both `main` and this branch, discarding the real `/work/myproj` block.

**3. Hot-path cost, ~2000x, worst on the harnesses we care about.** OpenAI-shaped body (1 system message without a cwd, 800 messages, 4KB blocks), 100 iterations:

```
main     0.0003 ms/call
PR 3644  0.73   ms/call
```

Claude Code exits early (env block is in the first user message). Codex / OpenCode / Cursor / Copilot / LiteLLM via `/v1/chat/completions` and `/v1/responses` have a `role=system` message and never a cwd prefix in user turns, so they pay the full scan on every request — synchronously on the event loop, and twice per Anthropic request (`anthropic.py:384` + `:1294`). Cheap fix: scan the raw blocks for the prefixes before joining, or bail after the first user message.

**Minor:** PR body says "31 -> 36 tests"; the branch file has 31 total, 3 of them new.

**Note:** conflicts textually with #3603 at the tail of `tests/test_memory_storage_router.py` (both append after line 411) — source hunks don't overlap. Also worth noting that bodies carrying both a `system` field and a `role="system"` message — exactly what the Kong gateway and LiteLLM translation layers produce — now get both concatenated. Strictly more resolution, never different, but it's a routing-behavior change for gateway-fronted traffic that isn't tested here.

PR #3646, opened 2026-09-18
fix(install): use safe model backends for persistent services

### LINKED ISSUE #3604: Persistent-service crash loop on macOS 26 (Darwin 27): PyTorch MPS backend aborts on IOGPUMetalCommandBuffer during embedder/kompress model load
## Description

The `persistent-service` deployment enters a hard crash loop on macOS 26 (Darwin 27). It restarts roughly every 60 seconds continuously (200+ crashes observed in one session). Each crash is a native Metal assertion abort during PyTorch MPS model load, not a Python exception, so nothing in headroom can catch or retry around it.

Root cause: the generated deployment config sets both `HEADROOM_EMBEDDER_RUNTIME=pytorch_mps` (memory embedder, `headroom/memory/adapters/embedders.py`) and `HEADROOM_KOMPRESS_BACKEND=pytorch_mps` (compression model, `headroom/transforms/kompress_compressor.py`), forcing PyTorch onto the Apple Silicon MPS/Metal backend. `HEADROOM_KOMPRESS_BACKEND=pytorch_mps` has a try/except around load-time validation that falls back to ONNX/CPU on failure, but this crash is a native Metal abort that kills the whole process before Python gets a chance to catch it, so the fallback path never runs. `HEADROOM_EMBEDDER_RUNTIME=pytorch_mps` has no fallback at all.

This looks tied to a macOS 26 upgrade (crashes started the day after upgrading) — PyTorch's MPS backend commonly lags behind new macOS majors before Apple/PyTorch re-validate compatibility.

## To Reproduce

1. Install headroom with the `persistent-service` preset on Apple Silicon running macOS 26 (Darwin 27), with `HEADROOM_EMBEDDER_RUNTIME=pytorch_mps` and/or `HEADROOM_KOMPRESS_BACKEND=pytorch_mps` set (this is the default in the generated `run-headroom.sh`/`ensure-headroom.sh` for a persistent-service deployment)
2. Start/restart the deployment: `headroom install restart`
3. Watch `~/.headroom/deploy/default/runner.log` — the proxy starts, then aborts within seconds during embedder/kompress model load; launchd respawns it and it crashes again on a ~60s cycle

## Expected Behavior

The proxy should either load the model successfully, or fail gracefully by falling back to a CPU/ONNX backend, without entering an unrecoverable crash loop that leaves the deployment permanently unhealthy.

## Actual Behavior

The process aborts with a native Metal assertion immediately after `Starting proxy server...`, before the proxy becomes healthy. `headroom install status` alternates between `running`/`healthy: yes` and `healthy: no` as launchd keeps respawning the process. Every client routed through the proxy (e.g. Claude Code via `ANTHROPIC_BASE_URL`) intermittently sees `API Error: Connection lost mid-response` when a request lands mid-crash.

## Code Sample

```python
# Not a client-code issue — the crash happens inside the headroom proxy's
# own model-loading path (memory embedder / kompress compressor), triggered
# simply by running the persistent-service deployment with the pytorch_mps
# backend selected. No user code is involved.
```

## Error Output

```
2026-09-15 21:08:18.176 python3[52375:4668384] failed assertion _status < MTLCommandBufferStatusCommitted at line 323 in -[IOGPUMetalCommandBuffer setCurrentCommandEncoder:]
Starting proxy server...
2026-09-15 21:09:23.790 python3[55198:4677048] failed assertion _status < MTLCommandBufferStatusCommitted at line 323 in -[IOGPUMetalCommandBuffer setCurrentCommandEncoder:]
Starting proxy server...
2026-09-15 21:10:27.087 python3[60362:4687421] failed assertion _status < MTLCommandBufferStatusCommitted at line 323 in -[IOGPUMetalCommandBuffer setCurrentCommandEncoder:]
Starting proxy server...
```
(repeats on a ~60s cycle, 200+ occurrences in the log)

## Environment

- **Headroom version**: 0.37.0
- **Python version**: 3.13.13
- **OS**: macOS 26.0 (Darwin 27.0.0, build 26A428), Apple Silicon (arm64)
- **LLM Provider**: Anthropic

## Additional Context

- torch version: 2.14.0
- Deployment: `persistent-service` preset, launchd-managed (`com.headroom.default`)
- Suggested fix: detect repeated rapid restarts and auto-downgrade `HEADROOM_EMBEDDER_RUNTIME`/`HEADROOM_KOMPRESS_BACKEND` to `cpu`/`onnx` for that deployment, or at least surface an actionable error via `headroom doctor` instead of a silent restart cycle. Also worth reconsidering defaulting persistent-service deployments to `pytorch_mps` on macOS given how brittle Metal/MPS compatibility is across OS versions.
- Workaround: manually setting `HEADROOM_EMBEDDER_RUNTIME=cpu` and `HEADROOM_KOMPRESS_BACKEND=onnx` (or unset, since `auto` prefers ONNX CPU) in the deployment's `run-headroom.sh`/`ensure-headroom.sh`, then `headroom install restart`, resolves it.

lorenzozanee:
## Description

Native persistent-service manifests now explicitly select CPU for the memory embedder and ONNX for Kompress by default. This avoids selecting Metal-specific model backends during supervised startup, while preserving explicit backend overrides.

Closes #3604

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `HEADROOM_EMBEDDER_RUNTIME=cpu` and `HEADROOM_KOMPRESS_BACKEND=onnx` to native `persistent-service` manifests only.
- Keep explicit `extra_env` values last so users can intentionally opt into another backend.
- Add planner regression coverage for service defaults, overrides, and non-service presets.
- Document the independent backend settings and the explicit MPS opt-in boundary.

## Testing

### Test Output

```text
python -m pytest -q tests/test_install/test_planner.py tests/test_install/test_supervisors.py tests/test_install/test_runtime.py tests/test_cli/test_install_cli.py
133 passed

ruff check headroom/install/planner.py tests/test_install/test_planner.py
All checks passed

ruff format --check headroom/install/planner.py tests/test_install/test_planner.py
All checks passed

git diff --check
passed
```

## Real Behavior Proof

- Environment: Ubuntu/Linux development host; native macOS launchd and Apple Silicon Metal were not available.
- Exact command / steps: Ran the focused planner, supervisor, runtime, and install CLI test suite plus Ruff and diff checks listed above.
- Observed result: The planner emits CPU/ONNX defaults for native persistent services, preserves explicit backend overrides, and leaves persistent-task and Docker manifests unchanged; the focused suite passed.
- Not tested: Real Apple Silicon macOS 26 behavior, launchd restart recovery, and ambient exported backend variables. The current base source also does not prove that a clean persistent-service install injected MPS; this PR is defensive safe-default hardening rather than a claim of reproducing the reported crash.

## Runtime Rollout Safety

- Minimum rollout channel: Standard release.
- Stable/default behavior changed: Native `persistent-service` manifests use CPU/ONNX model backends by default.
- Kill switch / disable path: Explicit `--env HEADROOM_EMBEDDER_RUNTIME=pytorch_mps` and/or `--env HEADROOM_KOMPRESS_BACKEND=pytorch_mps`, subject to platform and dependency support.
- Qualification impact: Persistent services no longer select Metal-specific model backends by default.

## Review Readiness

## Checklist

N/A — this is a manifest, test, and documentation change.

## Additional Notes

`mypy` was not locally runnable because the maturin editable build failed: the installed Rust toolchain lacked an applicable cargo component. The PR does not claim CI success or real macOS runtime validation.


JerrettDavis (approved):
Reviewed the preset boundary and override ordering. The safe CPU/ONNX defaults apply only to native persistent-service manifests, Windows correctly remains on the task fallback, Docker is unchanged, and explicit --env values still win. The head is current with main, conflict-free, and the full install/init/native/security matrix is green.

chopratejas:
Thanks for looking at #3604 — the crash-loop is a real problem. I don't think this change closes it, though, and I'd like to rescope it before merging.

**1. The premise doesn't hold on current `main`.** Nothing in the installer ever injected `pytorch_mps`: there's no match for it anywhere under `headroom/install/`, and `_capture_passthrough_env()` (`headroom/cli/install.py:398-421`) carries exactly seven `*_TARGET_API_URL` vars — not the model-backend vars. So the generated `run-headroom.sh` never set MPS. The real effect of this change is to *shield* the service from an **ambient** value (e.g. a `launchctl setenv`), since `install/runtime.py:105-107` does `os.environ.copy()` then `update(manifest.base_env)`. That's worth something, but it's a different fix than the one described.

**2. It doesn't help the reporter.** `extra_env` is applied last (`planner.py:191-193`), so a user who opted in with `--env HEADROOM_EMBEDDER_RUNTIME=pytorch_mps` still crash-loops — which your own test `test_persistent_service_explicit_model_backends_override_safe_defaults` pins. For #3604 to actually close it needs the crash-loop detection / doctor surfacing the reporter asked for.

**3. Forcing `onnx` removes a fallback and can silently disable compression.** `_load_kompress()` (`transforms/kompress_compressor.py:975-1021`): `auto` already prefers ONNX CPU — so on macOS it never picks MPS anyway — **and** falls back to PyTorch when onnxruntime is missing or the ONNX session fails to load. Forced `onnx` goes straight to `_load_kompress_onnx()` with no fallback; the raise is caught at `:1999-2004` and degrades to **passthrough, i.e. compression silently off**. That's a real quality regression for any Linux/Windows/container service lacking onnxruntime — none of which had the MPS problem.

`HEADROOM_KOMPRESS_BACKEND=auto` blocks an ambient `pytorch_mps` just as effectively while keeping the fallback chain.

**4. It's gated on preset, not platform.** Running the planner on this branch, `--preset persistent-service --runtime docker` injects `{'HEADROOM_EMBEDDER_RUNTIME': 'cpu', 'HEADROOM_KOMPRESS_BACKEND': 'onnx'}` into the **container** env, which contradicts the "native manifests only" claim. Linux and Windows service installs get it too. Conversely `persistent-task` — which on macOS is a real launchd watchdog running the same proxy via the same exported `base_env` — is excluded by design, so the identical crash stays reachable there, as does the Windows service→task downgrade path (`planner.py:152-155`). It also interacts with `HEADROOM_DOCKER_GPUS` (`install/runtime.py:161`): a GPU-enabled container service gets pinned to ONNX CPU.

**Suggested shape:** `auto` instead of `onnx`, plus a `sys.platform == "darwin"` guard, and apply it to the task preset too.

**5. Doc conflict left in place.** `wiki/macos-deployment.md:676-681` still tells users to enable MPS by adding `HEADROOM_EMBEDDER_RUNTIME` to the LaunchAgent plist `EnvironmentVariables`. The generated plist has no `EnvironmentVariables` key (`supervisors.py:212-231`), and the runner script's `export …=cpu` now overrides anything launchd hands down — so that documented opt-in silently stops working with no warning. This PR only edited `wiki/persistent-installs.md`. Separately, `wiki/configuration.md:241` says `pytorch_mps` is the only accepted value for that var, while the new text presents `cpu` as a value (`cpu` is in fact a no-op sentinel — `proxy/memory_handler.py:357` only ever compares against `pytorch_mps`).

Tests all pass (164 install + 118 CLI/opencode), so this is about scope and the forced backend, not correctness of the code as written.

PR #3655, opened 2026-09-18
fix(memory): fail closed for unresolved tools

### LINKED ISSUE #3651: [BUG] Explicit memory tools bypass the fail-closed workspace check and use the global DB
## Summary

When workspace resolution fails, `BackendRouter._resolve_scope` returns a sentinel `ResolvedScope` with `project_key=None` but a **live backend pointed at the legacy global `memory.db`**. The comment on that line says the path is safe because the caller checks `project_key`:

`headroom/memory/storage_router.py:343`
```python
db_path=self._config.global_db_path,  # Unused — caller checks project_key.
```

Exactly one caller checks. The auto-injection path does (`headroom/proxy/memory_handler.py:779-788`, `event=memory_inject_skipped reason=project_unresolved`). The five explicit memory-tool paths do not:

| method | line | `project_key is None` guard |
|---|---|---|
| `_execute_save` | 1221 | none |
| `_execute_search` | 1319 | none |
| `_execute_update` | 1364 | none |
| `_execute_delete` | 1439 | none |
| `_execute_list` | 1460 | none |

All five call `_resolve_for_request(...)`, discard the scope, and use the returned backend.

## Reproduce

Default `unresolved_project_fallback` (i.e. fail-closed `empty`, **not** the opt-in `global`):

```python
import tempfile, pathlib
from headroom.memory.storage_router import BackendRouter, BackendRouterConfig, MemoryStorageMode, RequestContext

d = pathlib.Path(tempfile.mkdtemp())
cfg = BackendRouterConfig(mode=MemoryStorageMode.PROJECT, root_dir=d, global_db_path=d / "memory.db")
backend, scope = BackendRouter(cfg).backend_for(
    RequestContext(headers={}, system_prompt="", base_user_id="alice", project_root_override=None)
)
print(scope.project_key)                      # None
print(str(scope.db_path).endswith("memory.db"))  # True
print(backend is not None)                    # True
```

Output:
```
event=memory_project_unresolved behavior=empty user_id=alice hint='... set memory.unresolved_project_fallback=global to opt-in to legacy cross-project GLOBAL pooling (cross-project leak risk).'
project_key  : None
display_name : unresolved (no memory)
db_path      : .../memory.db
is global db : True
backend live : True
```

The warning itself describes global pooling as an opt-in with "cross-project leak risk" — and then the fail-closed branch hands back a global-DB backend anyway.

## Impact

Any request whose workspace does not resolve, from any project, that reaches an explicit `memory_save` / `memory_search` / `memory_update` / `memory_delete` / `memory_list` call reads and writes **one shared database**, with an un-partitioned `user_id` (composition at `memory_handler.py:669-673` only applies to non-local backends). That is the cross-project pooling class the module docstring says is structurally prevented, and it is reachable without the operator ever setting `unresolved_project_fallback=global`.

This is currently **latent rather than exploited**, for two independent reasons:
- the `memory_*` tools sit behind tool-search deferral, so the model rarely calls them (see the sub-finding in #3573)
- it requires an unresolved workspace — which, however, is the *default* state for Claude Code 2.x today (#3595, #3597, #3596)

The last point is why this is worth fixing now rather than filing as theoretical: as those resolution bugs get fixed the exposure shrinks, but until then the default population is exactly the unresolved population.

## Suggested fix

Two layers, ideally both:

1. Add the `project_key is None` guard to the five `_execute_*` methods so they fail closed the way injection does — refuse the operation with a clear error rather than silently using the global store.
2. Stop handing out a usable backend for a scope that is explicitly "no memory". Returning `db_path=None` (or a scope that raises on `backend_for`) would make the invariant structural instead of dependent on every caller remembering to check — the comment at `storage_router.py:343` is a load-bearing assumption that five call sites already violate.

Found while triaging the #3595 / #3597 / #3596 cluster; not reported by a user.

TayfurYldz:
## Description

Explicit memory tools did not honor the existing fail-closed behavior for unresolved PROJECT scope. They resolved the sentinel scope but still used its backend, allowing save/search/update/delete/list to proceed without a resolved workspace.

Closes #3651

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Reject explicit memory operations when PROJECT routing resolves to the fail-closed unresolved sentinel.
- Keep GLOBAL fallback behavior unchanged when explicitly configured.
- Add regression coverage for save, search, update, delete, and list.

## Testing

### Test Output

```text
$ PYTHONPATH=. /tmp/headroom3651-venv/bin/python -m pytest -q tests/test_memory_handler_project_isolation.py
10 passed in 1.32s

$ /tmp/headroom3651-venv/bin/ruff check headroom/proxy/memory_handler.py tests/test_memory_handler_project_isolation.py
All checks passed!

$ git diff --check
# no output
```

## Real Behavior Proof

- Environment: current repository main at base commit `ff23cfbc178424f3c085493fa9e5b1809461c189` was not used here; this branch is based on Headroom commit `7639bca`'s parent. Regression was run before the production-code change in this worktree.
- Exact command / steps: added a parameterized regression invoking `memory_save`, `memory_search`, `memory_update`, `memory_delete`, and `memory_list` with PROJECT storage and no resolvable project identity.
- Observed result: before the fix all 5 cases failed the fail-closed assertion (`saved`, `found`, `updated`, `deleted`, `ok`); after the fix the complete isolation test file passes 10/10.
- Not tested: full repository test suite and mypy were not run.

## Runtime Rollout Safety

- Rollout-managed feature(s): memory PROJECT storage mode.
- Minimum rollout channel: N/A; correctness fix within existing PROJECT fail-closed semantics.
- Stable/default behavior changed: only explicit memory tool calls with unresolved PROJECT scope now fail closed instead of reaching the fallback backend.
- Kill switch / disable path: existing `unresolved_project_fallback=global` remains the explicit opt-in to legacy global fallback.
- Qualification impact: prevents unresolved workspace operations from crossing the intended project boundary.
- Rollback path: revert this commit; no schema or persisted-data migration is involved.

## Review Readiness

## Checklist

- [x] New and existing targeted unit tests pass locally with my changes

N/A.

## Additional Notes

Documentation is unchanged because this restores the already-documented fail-closed PROJECT behavior rather than introducing a new configuration or API. No new complex logic required inline comments.


JerrettDavis (changes requested):
The fail-closed behavior is correctly applied to all five explicit memory tools, and the focused isolation suite passes locally (10 tests). One required quality gate still blocks this revision: 	ests/test_memory_handler_project_isolation.py is not Ruff-formatted (the formatter inserts the required separation before the new parametrized test at line 320). Please format that file; no logic changes are needed for this finding.

JerrettDavis (changes requested):
The Ruff formatting blocker is fixed, and I verified the revised head locally (
uff format --check, 
uff check, and all 10 focused project-isolation tests pass). The new revision is still not merge-ready for two current-head CI blockers:

1. commitlint fails because the new commit title Format memory isolation regression test is not a Conventional Commit. Please amend it to a valid title such as 	est(memory): format isolation regression test.
2. The security gate now fails on newly disclosed AnyIO CVEs CVE-2026-63374 and CVE-2026-64847 (fixed in AnyIO 4.14.2). This appears repository-wide rather than caused by this PR, but the current head cannot be approved while the required security matrix is red.

No logic changes are requested.

JerrettDavis (approved):
Reviewed exact head 93c782570bbc347a16f3cfeb23489bc7b3e1e2f4 against current main. The change is correct and appropriately tested, the head is mergeable and contains current main, and all reported checks are terminal green.

Issue #3657, opened 2026-09-18
[FEATURE] Allow excluding providers/routes from OpenCode transport plugin routing


alvarocavalcanti:
## Problem Statement

The OpenCode transport plugin reroutes **every** non-loopback HTTP(S) request through the local Headroom proxy. When a provider's endpoint has a client-attribution check that the proxy hop or forwarder breaks, there is no supported way to opt that provider — or a specific route — out of routing. The only workaround is to unload the plugin entirely, which also removes Headroom for every other provider.

Concrete case: OpenCode Zen free-tier models served on `/zen/v1/responses` return `403 FreeTierError` when the plugin is loaded, while the same request succeeds with the plugin removed. The bug itself is tracked in #3656 and should be fixed on its own — this request is the general escape hatch for the same class of problem, not a prerequisite for that fix.

## Proposed Solution

A supported allow/deny mechanism for the OpenCode transport plugin so selected provider hosts or request paths bypass the proxy and go straight upstream (no compression, no telemetry for those requests). Possible shapes, maintainer's call:

- plugin options, e.g. `{ "exclude": ["opencode.ai"] }` / `{ "include": [...] }`
- env vars, e.g. `HEADROOM_OPENCODE_EXCLUDE_HOSTS` / `HEADROOM_OPENCODE_INCLUDE_HOSTS`
- path-level exclusion, e.g. skip `**/zen/v1/responses`
- or documented support for the existing per-request bypass (`x-headroom-mode: passthrough`), if it can be configured per-provider from OpenCode config

## Use Case

Anyone running OpenCode against a gateway that validates the client (OpenCode Zen free tier, forks, enterprise gateways) wants to keep Headroom for Claude/Codex traffic but not have it break a specific provider. Today the choice is all-or-nothing: keep the plugin and lose the provider, or drop the plugin and lose Headroom everywhere it was relying on the plugin for routing.

## Alternatives Considered

- **Unload the transport plugin** (current workaround) — loses Headroom for every provider that depended on plugin routing (Gemini, Copilot, opencode, custom gateways).
- **Stop the proxy** — loses Headroom entirely.
- **Per-request `x-headroom-mode: passthrough`** — still routes through the proxy, so it does not help when the gate is about the proxy hop itself; and OpenCode has no per-provider header config exposing it.

## Example API (Optional)

```jsonc
// ~/.config/opencode/opencode.json
{
  "plugin": [
[...]

Issue #1602, opened 2026-06-30
[BUG] headroom wrap opencode: Zen provider requests are not recorded in dashboard telemetry


Krimson4397:
I'm not entirely sure whether this is a bug or an expected limitation, but after spending quite a bit of time debugging it, I believe there is an issue with how Headroom handles OpenCode's Zen provider when using headroom wrap opencode.

I'm running the latest source build from main (including the recent OpenCode fixes), building the OpenCode transport plugin from source, and installing Headroom directly from the repository. The wrapper itself appears to work correctly: headroom wrap opencode starts the proxy, launches OpenCode, loads the Headroom transport plugin, and both the Headroom MCP server and Serena MCP work without any problems. Requests to the Zen provider are also successful, so the model itself is functioning normally.

The issue is that none of these requests appear in the Headroom dashboard. The dashboard always shows zero providers, zero agent usage, zero tokens, and zero savings, even though OpenCode is actively using the Zen model through the proxy.

I added some temporary logging while trying to understand what was happening. From what I observed, requests such as POST /zen/v1/chat/completions do reach the proxy correctly and include the x-headroom-base-url: https://opencode.ai header. However, when those requests enter the generic passthrough handler, both provider and endpoint_name appear to be None, so the telemetry recording logic never runs. From reading the code, it looks like these provider-prefixed routes don't match any of the dedicated API handlers and instead fall through to the generic passthrough path, where there doesn't seem to be enough information to classify them as LLM requests.

I also experimented by temporarily forcing provider="zen" and endpoint_name="chat.completions" inside the passthrough handler just to see what would happen. That wasn't a real fix—it incorrectly classified unrelated passthrough traffic such as MCP, npm registry, and Context7 requests—but it did confirm that the telemetry pipeline itself works once provider and endpoint information are available. That makes me think the issue occurs before telemetry, during request classification or routing.

I don't know enough about Headroom's internal architecture to suggest the correct implementation. It may require extracting provider/endpoint information for provider-prefixed routes like /zen/v1/chat/completions, or there may already be another mechanism intended for this that isn't being triggered. I just wanted to report everything I observed in case it helps narrow down the problem. If there's additional logging or testing that would be useful, I'm happy to help.

PR #3659, opened 2026-09-18
fix(proxy): keep OpenCode Zen /responses requests streaming

### LINKED ISSUE #3656: [BUG] OpenCode Zen free-tier /responses models fail with 403 FreeTierError when routed through the OpenCode plugin
## Summary

When the Headroom OpenCode transport plugin is loaded, OpenCode Zen free-tier models served on the **`/zen/v1/responses`** route fail with HTTP 403 `FreeTierError`:

```
Error from provider (Console): OpenCode's free tier can only be used from within OpenCode
```

Removing the Headroom plugin from the OpenCode config makes the exact same command succeed. Other free models (on `/zen/v1/chat/completions`) and paid `/responses` models work fine through Headroom.

## Environment

- headroom-ai **0.37.0** (also reproduced on 0.35.0), installed via `uv tool`
- macOS (darwin), Python 3.13
- OpenCode CLI **1.18.31** (also reproduced on 1.18.27)
- Provider: `opencode` (OpenCode Zen pay-as-you-go), model: `muse-spark-1.3-contributor-free`
- Plugin loaded from `.../site-packages/headroom/providers/opencode/_dist/entry.opencode.js`
- Proxy: `headroom install` persistent service, `127.0.0.1:8787`, `--mode cache --backend anthropic --no-telemetry`

## Steps to reproduce

1. Load the Headroom OpenCode plugin in `~/.config/opencode/opencode.json`.
2. Run:

```bash
opencode run -m opencode/muse-spark-1.3-contributor-free "Reply with exactly: OK"
```

Result:

```
Error: Error from provider (Console): OpenCode's free tier can only be used from within OpenCode
```

Raw upstream body: `{"type":"error","error":{"type":"FreeTierError","message":"Error from provider (Console): OpenCode's free tier can only be used from within OpenCode"}}` with HTTP 403.

Expected: `OK`.

## Evidence

Proxy log for a failing request:

```
event=proxy_inbound_request path=/v1/responses
  headers: x-headroom-base-url=https://opencode.ai, x-headroom-original-path=/zen/v1/responses,
           x-opencode-client=cli, x-opencode-session=..., x-opencode-project=...
event=outbound_request forwarder=openai_responses method=POST
  path=https://opencode.ai/zen/v1/responses
  body_mutated=true mutation_reasons=responses_compression,ccr_streaming_retrieve_buffered_non_stream source=canonical
event=proxy_inbound_response path=/v1/responses status=403
```

A/B results (all with the same OpenCode version and credentials):

| setup | result |
|---|---|
| Headroom plugin loaded, `muse-spark-1.3-contributor-free` | **403 FreeTierError** |
| Headroom plugin removed from config, same model | **OK** |
| Second proxy `--no-optimize` (passthrough, no compression) | **403** |
| Second proxy `--no-http2` | **403** |
| Headroom plugin loaded, `nemotron-3.5-lightning-free` (`/chat/completions`) | **OK** |
| Headroom plugin loaded, paid `/responses` model (`gpt-5.4-nano`, title gen) | **OK** |

## Analysis

- The failure is specific to the **free-tier `/zen/v1/responses` route**. Free `/chat/completions` models and paid `/responses` models are unaffected.
- Not caused by compression/body mutation: a `--no-optimize` proxy still returns 403.
- Not caused by HTTP/2 upstream: `--no-http2` still returns 403.
- Zen's gate is client-attribution based. A raw `curl` to `https://opencode.ai/zen/v1/responses` gets the same 403 even **without** Headroom (so curl is not a useful control; the A/B above is the valid one). This suggests the proxy hop changes something Zen relies on to attribute the request to the OpenCode client — potentially the outbound TLS/ALPN fingerprint, or a request attribute that the `openai_responses` forwarder does not preserve (the chat path evidently does preserve it).
- `headroom/proxy/passthrough.py` currently special-cases Zen for exactly one path (`zen/v1/chat/completions`, added for #1602). There is no equivalent handling for `zen/v1/responses`.

## Requested fix

Make `POST https://opencode.ai/zen/v1/responses` transparent in the same way as the `zen/v1/chat/completions` route, so the free-tier gate still sees the request as coming from OpenCode.

## Scope

This bug should be fixable on its own. The general mechanism for opting a provider or route out of plugin routing — useful as an escape hatch for this class of problem, but not required for this fix — is tracked separately in #3657.

Happy to test builds, enable debug logging, or capture outbound headers if that helps isolate what Zen is checking.

alvarocavalcanti:
## Description

Fixes #3656.

When the OpenCode transport plugin is loaded, OpenCode Zen free-tier models on `/zen/v1/responses` (e.g. `muse-spark-1.3-contributor-free`) fail with `403 FreeTierError` ("OpenCode's free tier can only be used from within OpenCode"), while the same request succeeds with the plugin removed.

Root cause: the Responses path's buffered-CCR mode rewrites a streaming request to `stream:false` and `accept: application/json` before it goes upstream. Zen validates OpenCode-client attribution on the wire and rejects the reshaped request. The chat-completions path is unaffected because `_should_inject_openai_chat_ccr_tool` already leaves streaming chat requests alone.

The general "opt a provider/route out of plugin routing" escape hatch is tracked separately in #3657 and is not needed for this fix.

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/proxy/passthrough.py`: add `is_opencode_zen_base()`, reusing the existing `OPENCODE_ZEN_HOSTS`.
- `headroom/proxy/handlers/openai.py`: `_should_buffer_openai_responses_stream_ccr()` takes `upstream_base_url` and returns `False` for OpenCode Zen, so the client's streaming request is forwarded unchanged. `x-headroom-base-url` is now resolved once and reused for the secret-header gate, the routing decision, and this decision.
- Tests for both helpers, including that other custom bases still buffer.

## Testing

- [ ] Type checking passes (`mypy headroom`) — could not run cleanly in this local env (numpy stub/Python-version mismatch aborts mypy); `mypy --ignore-missing-imports headroom/proxy/passthrough.py` passes. CI will run the full check.

### Test Output

```text
$ ruff check headroom/proxy/handlers/openai.py headroom/proxy/passthrough.py tests/test_proxy_openai_responses_stream_ccr.py tests/test_proxy_passthrough.py
All checks passed!
$ ruff format --check <same files>
4 files already formatted

$ pytest tests/test_proxy_openai_responses_stream_ccr.py tests/test_proxy_passthrough.py -q
9 passed

$ pytest tests/test_proxy_openai_responses_stream_ccr.py tests/test_proxy_passthrough.py \
    tests/test_ccr_response_handler_extra.py tests/test_proxy_byte_faithful_forwarding.py \
    tests/test_proxy_openai_responses_integration.py tests/test_openai_codex_routing.py -q
140 passed, 14 skipped
```

## Real Behavior Proof

- Environment: macOS (darwin arm64), Python 3.13.7, OpenCode CLI 1.18.31, provider `opencode` (OpenCode Zen pay-as-you-go), model `muse-spark-1.3-contributor-free`, Headroom OpenCode transport plugin loaded, OpenCode pointed at the proxy under test.
- Exact command / steps: `opencode run -m opencode/muse-spark-1.3-contributor-free "Reply with exactly: OK"`.
- Observed result: before the fix the command fails with `Error from provider (Console): OpenCode's free tier can only be used from within OpenCode` (installed 0.37.0 proxy, unpatched); after the fix the same command returns `OK` (this branch, `python -m headroom.cli proxy`, 0.38.0-dev). Control: the unpatched 0.37.0 proxy with `--no-ccr` also returns `OK`, which isolates the buffered-CCR rewrite as the trigger rather than the proxy hop or headers.
- Not tested: Codex/ChatGPT auth flow, other custom gateways, and any other Zen model. The "before" run used released 0.37.0 while the "after" run used `main` (0.38.0-dev); the unit tests pin the predicate behaviour on this branch.

## Runtime Rollout Safety

- Stable/default behavior changed: buffered CCR on `/v1/responses` is skipped **only** when the client-supplied upstream is `opencode.ai` / `www.opencode.ai`. Every other upstream keeps the existing buffered path.
- Kill switch / disable path: `--no-ccr` still disables the path entirely; omitting `x-headroom-base-url` falls back to the configured upstream.
- Qualification impact: none.

## Review Readiness

Happy to adjust the carve-out (host-level vs path-level) if you'd prefer a different surface.

## Checklist

N/A (CLI proof in Real Behavior Proof above).

## Additional Notes

- Kept the carve-out host-scoped to match the existing OpenCode Zen special-casing in `headroom/proxy/passthrough.py` (#1602). A broader "never reshape custom-base `/responses` requests" rule is possible but would affect Codex-style custom gateways, so it is left out intentionally.
- Trade-off: on Zen `/responses`, CCR retrieval is no longer resolved server-side. The precondition for this path is that the client advertises the `headroom_retrieve` tool, so the model can still call it and have the client resolve it.


alvarocavalcanti:
Fixed — thanks. The `Observed result` value is now on the same `- Observed result: …` line instead of a nested bullet, which is why `proof_field_values()` read it as empty.

Re-ran the checker locally against the updated body:
```text
$ python3 scripts/pr-governance.py --event <event> --body-file <body> --report <report>
valid=true
needs_author_action=false
problems: []
```

All other proof/rollout fields were already inline; no code changes were needed for this.

JerrettDavis (approved):
Verified the revised head against current main: mergeable and conflict-free, focused tests pass locally, Ruff check/format are clean, and the full current CI/security matrix is green. Ready for maintainer merge review.

PR #3660, opened 2026-09-18
fix(smart-crusher): disclose dropped scalar array items

### LINKED ISSUE #3650: [BUG] SmartCrusher drops array items with no marker on default config — valid JSON that contradicts its own count field
## Summary

On stock defaults, `SmartCrusher` drops array items down to `max_items_after_crush` and emits **no marker of any kind**. The result is well-formed JSON that is silently incomplete — and when the document carries its own count field, that field is left untouched, so the output actively contradicts itself.

This is the default-configuration half of #3625. #3628 (merged) fixed the `lossless_only=True` case by making the four crushers return input unchanged under that flag. It did not add a marker to the default path, so the silent drop below is unchanged on `main`.

## Reproduce

No flags, no allowlist, no lossless mode — the public router on default config:

```python
import json

data = json.dumps({"slugs": [f"r{i}" for i in range(120)], "total": 120})
out = ContentRouter(ContentRouterConfig()).compress(data).compressed
p = json.loads(out)

print(len(p["slugs"]))        # 16   (of 120)
print(p["total"])             # 120  <- still says 120
print("<<ccr:" in out)        # False
print("compressed" in out.lower())  # False
```

Observed: `valid JSON: True`, `slugs kept: 16 of 120`, `total field still says: 120`, `ccr marker present: False`, `any warning text: False`.

## Why this is the severe one

Ordinary lossy compression is a tradeoff the consumer can reason about *because it is disclosed*. Here there is no disclosure at any layer:

- no CCR marker, so nothing to retrieve and nothing to detect
- no truncation sentinel in the array
- the sibling `total` field is not adjusted, so the payload is **self-inconsistent** rather than merely short
- output is valid JSON, so every downstream parse and schema check passes

A model reading this cannot distinguish it from complete data. It will reason over 16 records while being told there are 120, and any count, sum, or "did I see them all" judgement it makes is wrong with no signal that anything happened. That is a different failure class from "the response was summarized."

`config.rs` `opaque_markers_enabled()` = `enable_ccr_marker && !lossless_only` governs only opaque-blob offload, never item drops, so **no configuration produces marker-free *and* complete output** — the point #3625 made, and it still holds for the default path.

## Suggested fix

Item drops on the default path should be disclosed. Options, roughly in order of preference:

1. Emit a CCR marker (or a retrievable truncation sentinel) whenever items are dropped, so the content is recoverable via `headroom_retrieve` like other offloaded content.
2. Failing that, append an in-band sentinel item (`"… 104 more items"`) so the truncation is at least visible in-band.
3. Adjust or drop sibling count fields when the array they describe is truncated — leaving `total: 120` next to 16 items is worse than either a marker or nothing.

Option 1 is the one that matches the rest of the system's contract.

## Environment

`main` @ the merge of #3628. Reproduced with the editable install + native `_core.abi3.so` in `.venv`.

Related: #3625 (the `lossless_only` half, closed by #3628), #1091, #1478.

lorenzozanee:
## Description

SmartCrusher could shorten string and mixed arrays on the default path without exposing the dropped rows, leaving valid JSON whose contents contradicted its count fields. This change stores the original scalar array in the existing CCR store and appends a retrievable disclosure marker for recoverable string and mixed arrays. Homogeneous numeric arrays remain numeric when a string marker would change their type.

Fixes #3650

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add a shared SmartCrusher array-boundary helper that canonicalizes and stores the original array, then appends `<<ccr:HASH N_rows_offloaded>>` when string or mixed array items are dropped.
- Preserve numeric arrays unchanged when marker insertion would violate their homogeneous numeric shape.
- Add CCR round-trip coverage for string, mixed, numeric, and existing marker paths.

## Testing

### Test Output

```text
rustup run 1.98.0 cargo test -p headroom-core --test ccr_roundtrip
18 passed

rustup run 1.98.0 cargo test -p headroom-core --lib smart_crusher
334 passed

rustup run 1.98.0 cargo fmt --all -- --check
passed

git diff --check
passed
```

## Real Behavior Proof

- Environment: Linux checkout, Rust toolchain `1.98.0`, default SmartCrusher configuration with the local CCR store.
- Exact command / steps: Ran the focused CCR round-trip test, the SmartCrusher library tests, `cargo fmt --all -- --check`, and `git diff --check`. The new regression tests exercise string and mixed arrays that lose rows, and numeric arrays that must retain their homogeneous shape.
- Observed result: The focused tests pass; dropped string and mixed rows produce retrievable CCR markers, while numeric arrays remain numeric.
- Not tested: Upstream CI, a live provider/runtime request, and the reporter's external deployment configuration.

## Runtime Rollout Safety

- Minimum rollout channel: default behavior.
- Stable/default behavior changed: dropped string and mixed array rows are now disclosed through the existing CCR marker contract; numeric array shape is preserved.
- Kill switch / disable path: revert this commit.
- Qualification impact: none known.

## Review Readiness

## Checklist

Not applicable.

## Additional Notes

The numeric-array behavior is an explicit compatibility boundary: adding a string marker would change a homogeneous numeric array into a mixed array. Upstream CI and real provider/runtime behavior remain to be checked by maintainers.


JerrettDavis:
This is now superseded by #3692, which implements the same retrievable scalar-array disclosure across both Rust and Python, updates parity fixtures, and covers string, mixed, and numeric-array behavior. To avoid maintaining a competing Rust-only implementation and duplicate conflict surface, I recommend closing this PR in favor of #3692. Thank you for identifying and addressing the original correctness gap.

PR #3663, opened 2026-09-19
fix: exclude Bash from DEFAULT_EXCLUDE_TOOLS so columnar output keeps fields (#3652)

### LINKED ISSUE #3652: [BUG] Columnar command output (ls -la, git status) classifies as PLAIN_TEXT and loses fields to Kompress — Bash missing from DEFAULT_EXCLUDE_TOOLS
## Summary

Two compounding gaps let columnar command output (`ls -la`, `git status`, `docker ps`, ...) reach the prose word-dropping compressor on default config:

1. `Bash` is not in `DEFAULT_EXCLUDE_TOOLS` (`headroom/config.py:221-266`) — every other raw-output tool (`Read`, `Glob`, `Grep`, `view`, `read_file`, ...) is excluded from lossy compression by name; `Bash` isn't, even though its output is exactly the same category of "bytes the agent will act on verbatim."
2. Fixed-width columnar rows classify as `PLAIN_TEXT` (`headroom/transforms/content_detector.py`), not as structured/tabular content, so they route to the same word-dropping compressor as prose — the mechanism #3580/#3590 identified for grep. #3599 (merged) fixed the `grep -A/-B/-C` dash-context shape specifically; it did not touch `ls -la` / `git status` style output, which carries no line-number prefix at all.

This is #3590 with the wrong root-cause framing corrected, plus the residual reach of #3580 that #3599 didn't cover. Filing separately since #3580 is closed by #3599 and shouldn't be reopened for a different input shape.

## Reproduce

Stock `ContentRouterConfig()`, no flags:

```python
from headroom.transforms.content_detector import detect_content_type
import random
random.seed(1)

rows = [f"-rw-r--r--  1 tejas staff {random.randint(1000,99999)} Sep {d} 09:{d:02d} file_{d}.py" for d in range(1, 60)]
payload = "total 480\n" + "\n".join(rows)

print(detect_content_type(payload))          # PLAIN_TEXT, confidence=0.5
out = ContentRouter(ContentRouterConfig()).compress(payload).compressed
```

Observed on `main`:
- input: 60 lines
- output: 2 lines (every newline dropped; all rows joined with spaces)
- all 59 filenames survive, but the **owner field is dropped from 17 of 59 rows** — a row that reads `-rw-r--r-- 1 staff 39848 ... file_59.py` with no username is indistinguishable from a row that never had one
- trailer: `[533 words compressed to 482 (from 60 source lines). Retrieve more: hash=...]`

`git status` output (`On branch main\nChanges not staged for commit:\n\tmodified: ...` × 30) also classifies `PLAIN_TEXT` under the same call.

## Why this matters

The marker at the end names a word count and a line count, but not *which* lines lost *which* fields, and every surviving row keeps its original position — so an agent parsing `git status`/`ls -la` for a subsequent command (`git add <the 12 modified files>`, `rm <files over 50KB>`) is working from data with unmarked, un-locatable holes. This is the same "the marker is present but doesn't disclose the event" class as #3580/#3545/#3590, on inputs that don't carry line numbers at all, so the retrieval-by-hash escape hatch is the only recovery path and the agent has no signal it needs it.

## Suggested fix

Either half closes most of the reach on its own; both together is more robust:

1. Add `Bash`/`bash` (and any other raw-shell-exec tool name) to `DEFAULT_EXCLUDE_TOOLS`, mirroring the existing entries for `Read`/`view`/`read_file`.
2. Extend content detection to recognize fixed-width columnar/tabular text (repeated-whitespace-delimited rows with a consistent column count) as a distinct type routed to a lossless or tabular-aware compressor instead of the prose path — the same shift #3599 made for the line-numbered grep shape, generalized to inputs without line-number prefixes.

## Related

- #3580 / #3545 (closed by #3599) — same underlying mechanism, different input shape (line-numbered grep context vs. columnar command output).
- #3590 — reported this exact symptom (`ls -la` fusion) but framed it as a min-token-floor problem; the min-token floor is a mitigation, not the cause, and remains a live gap since `Bash` is still unlisted after #3599.

Frank-zhu0404:
## Description

Exclude `Bash` from `DEFAULT_EXCLUDE_TOOLS` so columnar command output (`ls -la`, `git status`, …) is not lossy word-dropped on stock config. Owner/fields that agents act on verbatim stay intact via the lossless fold path.

Fixes #3652

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `Bash` to `DEFAULT_EXCLUDE_TOOLS` in `headroom/config.py`
- Clarify content-router comments that Bash is excluded but `_bash_search_fold` still applies when relevant
- Add/extend tests for Bash columnar owner preservation and exclude-tools routing

## Testing

### Test Output

```text
platform linux -- Python 3.12.14, pytest-9.0.3, pluggy-1.6.0
rootdir: /workspace/pr-work/headroom-3652
configfile: pyproject.toml
plugins: asyncio-1.3.0, cov-7.0.0, respx-0.23.1, anyio-4.14.2, langsmith-0.9.3
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 40 items

tests/test_bash_tool_result_protection.py ....                           [ 10%]
tests/test_bash_search_lossless_fold.py ..............................   [ 85%]
tests/test_content_router_exclude_tools.py ......                        [100%]

============================== 40 passed in 4.24s ==============================
```

## Real Behavior Proof

- Environment: Linux box, Python 3.12.14, stock `ContentRouterConfig()`, worktree `/workspace/pr-work/headroom-3652` @ `cdd69160`
- Exact command / steps: `uv run pytest tests/test_bash_tool_result_protection.py tests/test_bash_search_lossless_fold.py tests/test_content_router_exclude_tools.py -q` (40 passed); new tests assert default Bash exclusion, lossy compress bypass, `ls -la` owner preserved (tejas ×59), `git status` fields preserved
- Observed result: Bash columnar rows keep owner/fields under default config; prior grep lossless-fold behavior still covered
- Not tested: live multi-provider production traffic; ruff/mypy full-tree this round

## Runtime Rollout Safety

- Rollout-managed feature(s): none (default exclude-tools list only)
- Stable/default behavior changed: yes — Bash tool results now take lossless/excluded path instead of prose word-drop (intended fix for #3652)
- Kill switch / disable path: remove `Bash` from exclude_tools / override ContentRouterConfig.exclude_tools
- Qualification impact: build/test logs via Bash may retain more verbatim bytes (intended)
- Rollback path: revert commit / omit Bash from DEFAULT_EXCLUDE_TOOLS

## Review Readiness

## Checklist

- [x] Muse review `muse-spark-1.3-contributor --thinking xhigh` → **VERDICT=PASS**


Frank-zhu0404:
Fixed: put `Exact command / steps` on the same line after the colon (required by `scripts/pr-governance.py`). Local governance check now `valid=true`. PR remains draft until ready.

JerrettDavis (changes requested):
[P1] The same lossy path remains enabled for two shell tool names the router already recognizes. `ContentRouterConfig.bash_tool_names` is `{"bash", "shell", "local_shell"}`, but this change adds only `Bash`/`bash` to `DEFAULT_EXCLUDE_TOOLS` and explicitly rewrites several controls to `local_shell` so they continue through lossy Kompress. A `local_shell` or `shell` result containing the issue's `ls -la`/`git status` fixture therefore still classifies as `PLAIN_TEXT` and can silently lose owner/path fields—the exact correctness failure this PR is meant to close. Please protect every supported raw-shell tool name by default (or add a general columnar lossless route that makes their output safe) and cover at least `shell` and `local_shell` with the field-survival regression. Re-reviewed exact head `cdd69160`.

JerrettDavis (approved):
Re-reviewed exact head aa1e9e7e. The follow-up shares the router raw-shell name set across bash, shell, and local_shell, covers case-folded spellings and field survival, 124 focused tests pass locally with Ruff/format clean, and the full matrix is green.

PR #3665, opened 2026-09-19
fix: preserve null and empty strings in lossless CSV compaction

upramod:
## Description

Fix strict-lossless CSV-schema compaction so nullable string columns preserve the distinction between missing values, JSON `null`, and empty strings.

Before this change, swapping `""` and `null` between two rows could produce byte-identical `lossless:table` output.

Closes #

## Type of Change

## Changes Made

- Render JSON `null` explicitly as `null`.
- Quote empty strings as `""`.
- Quote the literal string `"null"` so it remains distinct from JSON `null`.
- Add regression tests for the ambiguous cases and swapped-row collision.

## Testing

### Test Output

```
cargo test -p headroom-core csv_formatter_distinguishes_missing_null_empty_and_literal_null
cargo test -p headroom-core csv_formatter_nullable_swap_is_not_identical
```

Both passed.

## Validation run:
https://github.com/upramod/invariant-compression-gate/actions/runs/35455082632

## Real Behavior Proof

- Environment: Headroom 0.36.0, 0.37.0, and current upstream `main`
- Exact command / steps: run strict-lossless compaction on two inputs that swap `""` and `null` between rows
- Observed result: before the fix, outputs were byte-identical; after the fix, they differ
- Not tested: full repository Python/lint/type-check suite

## Runtime Rollout Safety

- Rollout-managed feature(s): None
- Stable/default behavior changed: Only CSV-schema rendering for affected nullable string values
- Kill switch / disable path: Use another formatter or disable the lossless CSV-schema path
- Qualification impact: Localized to preserving scalar distinctions in lossless CSV output
- Rollback path: Revert this commit

## Review Readiness

## Checklist


JerrettDavis (changes requested):
The null/empty/missing encoding is byte-distinct and the added semantic tests are useful, but this exact head fails the required Rust formatting gate in formatter.rs (the multiline assert_eq! block is what cargo fmt --check reports). Please apply the repository's cargo formatting to the Rust change so the committed source matches rustfmt; no logic change is needed.

upramod:
Done

JerrettDavis (approved):
Re-reviewed exact head a5235858 after the clean rebase. The contributor patch is unchanged, the malformed self-merge is gone, focused Rust tests and formatting pass locally, and all required checks are green.

Issue #3633, opened 2026-09-17
[BUG] opencode plugin routes ALL HTTP traffic through the proxy — breaks built-in WebFetch and every Node child process (npx/npm/Node MCPs)


christestet:
## Summary

The OpenCode plugin's transport layer routes **all** outbound HTTP traffic — not just LLM API traffic — through the Headroom proxy. Two propagation paths:

1. **In-process:** the plugin patches `globalThis.fetch` in the opencode (Bun) process with a `shouldRoute()` filter that returns `true` for *every* non-loopback http(s) URL. opencode's built-in `WebFetch` tool does client-side `fetch()` in that same process, so every webfetch request is rewritten to the proxy and rejected.
2. **Child processes:** the plugin wraps `child_process.spawn/exec/execFile/fork` and injects `NODE_OPTIONS=--import=<...>/hook-shim/handler.js` plus `HEADROOM_OPENCODE_TRANSPORT_PROXY_URL` into **every spawned child**. The hook-shim installs the same transport patch (it contains its own copy of `shouldRoute()` + `globalThis.fetch =`) into every Node child process. That breaks `npx`, `npm`, Node-based CLI tools (e.g. the context7 CLI) and Node MCP servers.

Non-Node children (curl, gh, python, ...) are unaffected, which makes the failure look inconsistent and is painful to debug.

## Environment

- headroom-ai 0.37.0 (pipx), proxy on `127.0.0.1:8787`
- OpenCode 1.18.31 (bun 1.3.14), plugin registered via `plugin` array in `opencode.json`
- Upstream: OpenAI-compatible vLLM endpoint

## Repro A — built-in WebFetch

1. Start `headroom proxy`, register the opencode plugin.
2. Ask the agent to fetch any page:

   ```
   > use webfetch on https://example.com
   StatusCode: non 2xx status code (400 GET https://example.com)
   ```

3. `curl https://example.com` from the same session's bash tool works fine (200) — curl is a non-Node child.

Manual proof at the proxy level:

```console
$ curl -s -H "x-headroom-base-url: https://example.com" http://127.0.0.1:8787/
{"detail":"Rejected unsafe upstream base URL"}   # status 400
```

The plugin rewrites `fetch("https://example.com")` to `http://127.0.0.1:8787/` with `x-headroom-base-url: https://example.com`; the proxy (correctly, SSRF guard) refuses non-LLM upstreams.

## Repro B — npx / npm / Node CLIs (context7 as example)

[...]

PR #3668, opened 2026-09-20
feat(opencode): exclude hosts from transport plugin routing

### LINKED ISSUE #3657: [FEATURE] Allow excluding providers/routes from OpenCode transport plugin routing

Ayushraj06-bit:
## Description

The OpenCode transport plugin rewrites every non-loopback HTTP(S) request to the Headroom proxy, both in the OpenCode process and in every spawned Node child via the hook-shim. When a provider validates the client on the wire and rejects the proxy hop (OpenCode Zen free tier, some enterprise gateways), the only way to keep that provider working is to unload the plugin, which drops Headroom for every other provider too.

This adds an exclude list to `shouldRoute()`. Hosts come from the `excludeHosts` plugin option, or from `HEADROOM_OPENCODE_EXCLUDE_HOSTS` (comma-separated) when the option is absent, with the same option-over-env precedence `proxyUrl` already uses. An entry matches the host and all of its subdomains (`example.com`, `.example.com` and `*.example.com` are equivalent). Excluded requests go straight upstream with their original URL, init and headers: no `x-headroom-base-url`, no compression, no telemetry. The effective list is exported to `process.env` and to spawned child env so the hook-shim copy of the transport makes the same decision. Default is empty, so nothing changes for existing installs.

Root cause: `shouldRoute()` (unchanged since #1105) only checked scheme, loopback and the proxy origin; it had no configuration input at all, and the child-process shim configures itself purely from env, so any opt-out has to be env-readable and exported by `withShimEnv`/`installProcessEnv` or children silently diverge from the parent (#3633 points this out).

Considered and not done here: path-level globs (both reported cases are host-level; can be layered on later as a separate variable), proxy-side passthrough (still a proxy hop, which is what the Zen gate rejects), and a default exclusion for `registry.npmjs.org` (a separate decision; this keeps zero behavior change by default). The Zen `/responses` bug itself is #3656 / #3659.

Closes #3657
Refs #3633

Note: #3657 is a feature request and did not yet have a maintainer 👍 when I opened this, so I am keeping it small and happy to change the surface: drop the plugin option and ship env-only, rename the variable to #3633's `HEADROOM_OPENCODE_ROUTE_EXCLUDE`, or hold until the surface is agreed.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `plugins/opencode/src/transport.ts`: `normalizeExcludeHosts` / `isExcludedHost`; `shouldRoute` takes the list and returns `false` on a match (fetch, `http`/`https.request`/`get`, and `http2.connect`, which is now allowed for excluded hosts instead of blocked); `installHeadroomTransport` resolves `options.excludeHosts ?? process.env.HEADROOM_OPENCODE_EXCLUDE_HOSTS`; `installProcessEnv` and `withShimEnv` export the effective list.
- `plugins/opencode/src/plugin.ts`: `excludeHosts?: string[]` on `HeadroomOpenCodePluginOptions`, passed through.
- `plugins/opencode/src/transport.test.ts`: four tests — env var with subdomain/case/`.`/`*.` forms and non-matching hosts still routed; option precedence over env; Node `https.request` and `http2.connect` on excluded hosts; child env propagation (and no variable set when nothing is excluded).
- `headroom/providers/opencode/_dist/entry.opencode.js`, `headroom/providers/opencode/hook-shim/handler.js`: rebuilt with `npm run build:standalone`.
- `plugins/opencode/README.md`, `docs/content/docs/opencode.mdx`: document the option, the variable and the matching rule.

## Testing

- [ ] Unit tests pass (`pytest`) — not run in full locally; `tests/test_providers_opencode_plugin_path.py` and `tests/test_providers_opencode_config.py` give the same result as `main` on my Windows checkout (44 passed, 1 pre-existing Windows-only path-escaping failure in `test_build_launch_env_with_project`). No Python source changed.
- [ ] Linting passes (`ruff check .`) — no Python changed.
- [ ] Type checking passes (`mypy headroom`) — no Python changed.

Plugin CI sequence from `.github/workflows/opencode-plugin.yml`, run after rebasing onto `main` (`67eb910e`): typecheck, build, test, standalone build, bundle parity, plus commitlint and `npm audit`.

### Test Output

```text
$ cd plugins/opencode && npm ci && npm run typecheck && npm run build && npm test
found 0 vulnerabilities
typecheck exit=0
 Test Files  2 passed (2)
      Tests  21 passed (21)

$ npm run build:standalone && cmp dist-standalone/entry.opencode.js ../../headroom/providers/opencode/_dist/entry.opencode.js && cmp dist-standalone/hook-shim/handler.js ../../headroom/providers/opencode/hook-shim/handler.js
entry.opencode.js: IDENTICAL
handler.js: IDENTICAL

# the four new tests against unpatched transport.ts (fix stashed, tests kept):
 ❯ src/transport.test.ts (19 tests | 4 failed)
     × keeps HEADROOM_OPENCODE_EXCLUDE_HOSTS hosts and their subdomains off the proxy (#3657)
     × prefers the excludeHosts option over HEADROOM_OPENCODE_EXCLUDE_HOSTS
     × leaves excluded hosts alone for Node https.request and http2.connect
     × propagates excluded hosts into child process env so the hook-shim matches
AssertionError: expected URL { href: 'http://127.0.0.1:8787/v1/responses', ... } to be 'https://opencode.ai/zen/v1/responses'

$ npx commitlint --config .commitlintrc.json --edit <msg>   # exit 0
$ npm audit --package-lock-only --audit-level=low            # found 0 vulnerabilities
```

## Real Behavior Proof

- Environment: Windows 11, Node v24.13.0, plugin built from this branch (`npm run build` for the in-process check, the committed wheel `hook-shim/handler.js` from this branch for the child-process check), fake proxy on `127.0.0.1` recording `x-headroom-base-url`. `HEADROOM_OPENCODE_EXCLUDE_HOSTS=opencode.ai`.
- Exact command / steps: (1) import `dist/index.js`, install the transport with `excludeHosts: ["opencode.ai"]`, `fetch("https://opencode.ai/zen/v1/responses")` then `fetch("https://api.anthropic.com/v1/messages")`, record what the original `fetch` received; (2) spawn `node --import=<hook-shim/handler.js>` with `HEADROOM_OPENCODE_TRANSPORT_PROXY_URL=<fake proxy>` and the variable above, `fetch("https://opencode.ai/zen/v1/responses")` from the child, watch the fake proxy; (3) same two scripts on unpatched `main` as the control.
- Observed result: (1) `{"url":"https://opencode.ai/zen/v1/responses","base":null}` then `{"url":"http://127.0.0.1:8787/v1/messages","base":"https://api.anthropic.com"}` — excluded host untouched, Anthropic still routed; (2) `child: fetch resolved status 401 from https://opencode.ai/zen/v1/responses` / `fake proxy saw: []` — the child went direct to the real host (401 is opencode.ai answering an unauthenticated request) and the proxy never saw it; (3) on `main` both scripts route `opencode.ai` to the proxy with `base: https://opencode.ai`, and on this branch the same happens when the variable is spelled wrong (`HEADROOM_OPENCODE_ROUTE_EXCLUDE`), i.e. an unknown variable changes nothing.
- Not tested: a live OpenCode session against the real Zen free-tier model (I don't have OpenCode Zen credentials); macOS/Linux; Bun-spawned children (they inherit `process.env`, which `installProcessEnv` now sets, same path as `NODE_OPTIONS`); the full Python suite (CI).

## Runtime Rollout Safety

- Stable/default behavior changed: no. With no `excludeHosts` option and no `HEADROOM_OPENCODE_EXCLUDE_HOSTS`, `shouldRoute()` is unchanged and the variable is never written. Only configured hosts bypass the proxy; for them `http2.connect` is now allowed instead of throwing.
- Kill switch / disable path: unset the variable / remove the option; excluded traffic goes back through the proxy on next launch.
- Qualification impact: none.
- Rollback path: revert this commit (source and the two committed bundles change together).

## Review Readiness

## Checklist

## Additional Notes

Start reading at `shouldRoute()` and `normalizeExcludeHosts()` in `plugins/opencode/src/transport.ts`, then `installHeadroomTransport` for the option/env resolution and the two env exports. The bundle diffs are the same hunks compiled. The one decision I'd most like challenged is shipping both a plugin option and an env var: the env var is required regardless (it is the only surface that reaches `headroom wrap opencode`, which registers the plugin as a bare path, and the hook-shim in child processes), and the option mirrors the reporter's proposed API; dropping the option is a six-line change if you'd rather keep one surface.

Known limits: entries are bare hosts (no scheme, port or path), IDN entries must be given in punycode, and like `HEADROOM_OPENCODE_TRANSPORT_PROXY_URL` the exported variable is not removed from `process.env` on dispose.


JerrettDavis (changes requested):
[P1] An explicit empty `excludeHosts` override does not propagate to child hook shims. `installHeadroomTransport()` correctly resolves `options.excludeHosts ?? env`, so `excludeHosts: []` is supposed to override a pre-existing `HEADROOM_OPENCODE_EXCLUDE_HOSTS`; however, both `installProcessEnv()` and `withShimEnv()` only set the variable when the resolved list is non-empty and never delete it when empty. Reproduction: start with `process.env.HEADROOM_OPENCODE_EXCLUDE_HOSTS="api.anthropic.com"`, install with `{ excludeHosts: [] }`, then spawn a child. The in-process transport routes Anthropic through Headroom, but the spawned child inherits the stale environment value and bypasses it, violating the documented option precedence and making parent/child routing disagree. Please clear the exclusion variable from the generated/process env when an explicit/updated resolved list is empty, preserve any intended lifecycle semantics, and add the empty-override regression. Re-reviewed exact head `b9f5d12b`.

Ayushraj06-bit:
Thanks — confirmed and fixed in `fec9dbdf` (second commit, so the delta is reviewable on its own; the branch is also rebased onto current `main`).

Reproduced your scenario first: `process.env.HEADROOM_OPENCODE_EXCLUDE_HOSTS="api.anthropic.com"`, install with `{ excludeHosts: [] }`, spawn a child — the parent routed Anthropic through the proxy while the child inherited `"api.anthropic.com"` and would have bypassed it.

Fix: the exported variable now mirrors the *effective* list rather than merely a non-empty one. `installProcessEnv()` and `withShimEnv()` share `withExcludeHostsEnv()`, which sets the variable when the resolved list is non-empty and deletes it when empty — so an explicit empty override clears a pre-existing value for the process env, for the default child env, for a caller-supplied env, and for a caller-supplied env that itself carries the stale value. Re-installs (refs > 1) go through the same export, so an updated resolution stays in step. Dispose semantics are unchanged (the variable is left as-is on the final disposer, same lifecycle as `HEADROOM_OPENCODE_TRANSPORT_PROXY_URL`).

Regressions added: `clears a pre-existing HEADROOM_OPENCODE_EXCLUDE_HOSTS when excludeHosts is explicitly empty` (in-process routing, `process.env`, and the three child-env shapes) and `keeps the exported variable in step with each re-install's resolved list`. Both fail on `b9f5d12b` with `expected 'api.anthropic.com' to be undefined` and pass now; 23/23. End to end with the committed hook-shim: parent with the stale variable installs `excludeHosts: []`, a child spawned through the wrapped `spawn` fetches `api.anthropic.com`, and the fake proxy sees `x-headroom-base-url` from both parent and child. Bundles rebuilt with `npm run build:standalone` and verified byte-identical against a fresh `npm ci` on the rebased lockfile.

JerrettDavis (approved):
Reviewed exact head 32b75ac8a9aae6edb1ed7c2f93fbb71de7c28e34 against current main. The change is correct and appropriately tested, the head is mergeable and contains current main, and all reported checks are terminal green.

PR #3670, opened 2026-09-20
fix(install): wait for the old process to stop before restarting

### LINKED ISSUE #3658: install restart leaves the deployment stopped (reports success)
`headroom install restart --profile <name>` prints `Restarted deployment '<name>'.` but the service is left in `stopped` state. `install start` afterwards works fine, so only the restart path is affected.

## Reproduce

```
$ headroom install status --profile copilot | grep '^Status'
Status:     running

$ headroom install restart --profile copilot
Restarted deployment 'copilot'.

$ headroom install status --profile copilot | grep -E '^Status|^Healthy'
Status:     stopped
Healthy:    no
```

Reproduced three times on the same profile. `headroom install start --profile copilot` brings it back every time.

## Impact

The command reports success, so there is no signal that the proxy is gone. Clients routed at that port fail with connection errors until someone notices and runs `install start`. In my case VS Code Copilot reported `Language model unavailable` for several minutes before I checked the deployment status.

## Environment

- headroom 0.37.0 (`pip`, native Python runtime)
- Python 3.12.10
- macOS 26.7 (Apple Silicon)
- preset `persistent-service`, supervisor `service` (launchd), scope `user`
- profile created with: `headroom install apply --profile copilot --port 8788 --providers manual --env ...`

Not tested on the `default` profile or on the docker/task presets.

gingeekrishna:
## Description

`headroom install restart` reports success but leaves the deployment stopped. Stopping the supervisor returns before the old proxy finishes its graceful shutdown, so `/readyz` still answers. `_start_deployment` treats a ready endpoint as "already running" and returns without starting anything. The old process then exits and nothing replaces it. `install start` works afterwards because by then nothing is answering.

Restart now waits (bounded, 15s) for the health endpoint to go down before starting, and fails instead of reporting success if it never does.

Closes #3658

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add `_wait_until_stopped` in `headroom/cli/install.py` and call it between `_stop_deployment` and `_start_deployment` in `install restart`.
- Add two regression tests: old process lingers after stop (restart must wait, then start), and old process never stops (restart must fail without starting or reporting success).

## Testing

### Test Output

```text
python -m pytest tests/test_cli/test_install_cli.py -q
42 passed

python -m ruff check headroom/cli/install.py tests/test_cli/test_install_cli.py -> All checks passed
python -m ruff format --check (same files) -> already formatted

Without the source change, test_install_restart_waits_for_old_process_before_starting fails (1 failed).
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.14.3. The issue was reported on macOS (launchd), which I could not run.
- Exact command / steps: the unit tests above simulate the old process still answering `/readyz` after stop.
- Observed result: restart waits for the endpoint to go down, then starts; it errors if the endpoint never goes down.
- Not tested: a real launchd / systemd restart; mypy; the wider test suite beyond `test_install_cli.py`; the docker and task presets. The root cause is inferred from the code path and the issue's symptoms, not reproduced on macOS.

## Runtime Rollout Safety

- Stable/default behavior changed: `install restart` now waits up to 15s for the old process to stop, and errors if it does not.
- Kill switch / disable path: n/a
- Qualification impact: none
- Rollback path: revert this commit

## Review Readiness

## Checklist

## Additional Notes

Docs: not applicable (no user-facing docs describe restart timing). Full unit suite not run locally; only `test_install_cli.py`.


Ayushraj06-bit:
Confirming the root cause with measurements, since you asked for a before/after: `launchctl bootout` returns as soon as the job is told to go (see openclaw/openclaw#43960 and this repo's own `_bootstrap_with_retry` comment about the EIO window "while launchd releases it"), and uvicorn keeps accepting connections until its next 100 ms `main_loop` tick. With the proxy's own server stack (uvicorn 0.32 / `timeout_graceful_shutdown=10`), a `probe_ready` call 15 ms after the shutdown signal returned `ready: true` in 5/5 runs.

Driving the real `install restart` command against a live uvicorn server, with only the supervisor primitives swapped for "send the server its shutdown signal" / "respawn it" (what `bootout` and `bootstrap`+`RunAtLoad` do), reproduces the issue verbatim on `main`:

```
probes after the shutdown signal (ms, ready): [(19, True)]
restart: exit=0 output="Restarted deployment 'copilot'."
after:   Healthy: no | start_supervisor called: 0
```

and passes with a bounded wait in place (5/5). Scripts and full output are in #3679's description.

One suggestion: `install stop` followed by `install start`, and `apply` on a running profile (`_remove_deployment` → `_start_deployment`), take the same path. Putting the wait at the end of `_stop_deployment` — as `wait_stopped()` next to `wait_ready()` in `runtime.py` — covers those too and keeps "Stopped deployment" honest as well. I had opened #3679 with that placement before seeing this PR; closing it in favour of this one.

gingeekrishna:
Thanks @Ayushraj06-bit, the measurements are very helpful, and good call on the placement. `stop` then `start` and `apply` on a running profile go through the same skip-start path.

Pushed a follow-up commit: the wait now lives in `_stop_deployment` as `wait_stopped()` next to `wait_ready()` in `runtime.py`. It fails with a ClickException if the endpoint is still answering after 15s, so `install stop` no longer reports 'Stopped' while the proxy is still up. `_remove_deployment` already swallows stop errors, so removal is unaffected.

Tests: unit tests for `wait_stopped` (waits, then times out), plus CLI tests that restart waits between stop and start and that both `stop` and `restart` fail without starting when the endpoint never goes down. The three new CLI tests fail without the change. `test_install_cli.py` + `tests/test_install/test_runtime.py`: 82 passed; ruff and mypy clean on the touched files. `test_native_installers.py::test_powershell_native_installer_supports_persistent_docker_lifecycle` fails locally on Windows with and without my change.

CI: the only failure on the earlier run was `test (3)` → `test_deepseek.py::test_cost_per_token_resolves_deepseek_v4_flash` (`0.15 == 0.3`). It is unrelated to this change: the same test fails on main (run 35631667410), and #3684 is open to fix it. I'll re-check CI on the new commit.

JerrettDavis (approved):
Reviewed exact head e372c757d4b85fa82918665cc7c9aa5100e763c4 after the history-only rebase. The cumulative stable patch ID is unchanged; 82 focused install/runtime tests pass, Ruff is clean, all fresh required checks are green, and GitHub reports the PR CLEAN.

PR #3671, opened 2026-09-20
fix(proxy): keep leading Anthropic system payloads valid

### LINKED ISSUE #3667: Wire-contract guard relocates non-text blocks into top-level `system`, causing 400 on every affected request
# Bug: `relocate_system_messages_to_top_level` emits invalid `system` blocks, causing a 400 on every affected request

**Version:** headroom-ai 0.37.0
**Component:** `headroom/proxy/helpers.py` — `_system_message_to_blocks`
**Backend:** `anthropic` (direct), proxy mode `token`
**Impact:** every request carrying a `role="system"` message whose content is a
list of non-text blocks fails upstream with HTTP 400. The client sees only
Anthropic's validation error, with no indication the proxy caused it.

## Symptom

```
API Error: 400 system.4.type: Input should be 'text'
```

From the proxy log, immediately before the failure:

```
WARNING - [hr_...] Relocated role=system message(s) out of messages[] into the
          top-level system parameter (Anthropic wire-contract guard, issue 765)
WARNING - event=cache_breakpoints ... in_system=2 out_system=3
WARNING - [hr_...] Forwarding upstream streaming error status=400
          url=https://api.anthropic.com/v1/messages
```

## Root cause

The wire-contract guard correctly moves `role="system"` messages out of
`messages[]` into the top-level `system` field. But when it converts the
message content into system blocks, it appends any `dict` verbatim:

```python
if isinstance(content, list):
    blocks: list[Any] = []
    for block in content:
        if isinstance(block, dict):
            blocks.append(block)          # <-- appended unchanged
        elif isinstance(block, str) and block:
            blocks.append({"type": _TEXT_BLOCK_TYPE, "text": block})
    return blocks
```

Anthropic's top-level `system` field accepts **text blocks only**. A block that
is perfectly valid inside `messages[].content` is not necessarily valid there,
so the relocation can produce an illegal request:

| Relocated block | Anthropic response |
|---|---|
| `{"type": "image", ...}` | `system.N.type: Input should be 'text'` |
| `{"role": "system", "content": "..."}` | `system.N.type: Field required` |

The guard is protecting against one wire-contract violation and introducing
another. Because the failure happens upstream, it presents as an opaque client
error rather than a proxy bug.

## Reproduction

```python
from headroom.proxy.helpers import relocate_system_messages_to_top_level

msgs = [
    {"role": "system", "content": [
        {"type": "image",
         "source": {"type": "base64", "media_type": "image/png", "data": "iVB"}}]},
    {"role": "user", "content": "hi"},
]
_, system, _ = relocate_system_messages_to_top_level(
    msgs, [{"type": "text", "text": "base"}], model="claude-opus-5")
print(system)
# [{'type': 'text', 'text': 'base'},
#  {'type': 'image', 'source': {...}}]   <-- illegal in top-level system
```

Forwarding that to `api.anthropic.com/v1/messages` returns 400.

## Proposed fix

Coerce each block to a valid text block before it reaches `system`, and drop
what cannot be represented as text rather than failing the whole request. Valid
text blocks pass through untouched so modifiers such as `cache_control`
survive. See `headroom-system-block-fix.patch`.

Dropping is a deliberate tradeoff against the "never lose content" principle in
the surrounding code: a non-text block is *unrepresentable* in `system`, so the
choice is between losing one block and losing the entire request. The drop is
logged:

```
event=system_relocation_block_dropped block_type=image reason=not_representable_as_text
```

An alternative worth considering: skip relocation entirely for system messages
containing non-text blocks, leaving them in `messages[]`. That preserves all
content but reintroduces the original wire-contract problem for those messages.

## Test cases

Verified against the patched build:

| Input block | Result |
|---|---|
| `{"role":"system","content":"..."}` (raw message) | dropped + warned |
| `{"type":"image",...}` | dropped + warned |
| `{"type":"text","text":"x","cache_control":{...}}` | passed through, `cache_control` intact |
| `{"text":"x"}` (no type) | normalized to `{"type":"text","text":"x"}` |
| `"plain string"` | wrapped as text block |

All five now produce only `type: "text"` blocks.

---

*Note:* the patch is diffed against the installed package
(`site-packages/headroom/proxy/helpers.py`) with paths rewritten to
`src/headroom/...`. Confirm the repository layout before applying.

## Patch

```diff
--- a/src/headroom/proxy/helpers.py
+++ b/src/headroom/proxy/helpers.py
@@ -884,8 +884,36 @@
 # the load-bearing values are named here.
 _ROLE_SYSTEM = "system"
 _TEXT_BLOCK_TYPE = "text"
+
+
+def _coerce_system_block(block: dict[str, Any]) -> dict[str, Any] | None:
+    """Coerce a content block into a block valid for top-level ``system``.
+
+    Anthropic's top-level ``system`` field accepts text blocks only. A block
+    that is valid inside ``messages[].content`` is not necessarily valid here,
+    so forwarding one verbatim makes Anthropic reject the entire request:
+
+        system.N.type: Input should be 'text'   (wrong type, e.g. "image")
+        system.N.type: Field required           (no type, e.g. a raw message)
+
+    Returns ``None`` for a block that cannot be represented as text, so the
+    caller can drop it rather than poison the whole request.
+    """
+    block_type = block.get("type")
+    text = block.get("text")
+
+    # Already a valid text block — pass through untouched so modifiers such as
+    # cache_control survive.
+    if block_type == _TEXT_BLOCK_TYPE:
+        return block if isinstance(text, str) and text else None
 
+    # Missing or wrong type but carries text: normalize instead of dropping.
+    if isinstance(text, str):
+        return {**block, "type": _TEXT_BLOCK_TYPE} if text else None
 
+    return None
+
+
 def _system_message_to_blocks(message: dict[str, Any]) -> list[Any]:
     """Convert a ``role="system"`` message into Anthropic system content blocks."""
     content = message.get("content")
@@ -895,7 +923,15 @@
         blocks: list[Any] = []
         for block in content:
             if isinstance(block, dict):
-                blocks.append(block)
+                coerced = _coerce_system_block(block)
+                if coerced is not None:
+                    blocks.append(coerced)
+                else:
+                    logger.warning(
+                        "event=system_relocation_block_dropped block_type=%s "
+                        "reason=not_representable_as_text",
+                        block.get("type"),
+                    )
             elif isinstance(block, str) and block:
                 blocks.append({"type": _TEXT_BLOCK_TYPE, "text": block})
         return blocks
```

lorenzozanee:
## Description

Fixes #3667

A leading Anthropic `system` message containing image or document blocks could remain in `messages[0]` after relocation, producing an invalid request shape. The fix keeps text blocks, including `cache_control`, in top-level `system`, warns and drops leading blocks without a valid text representation, and removes the invalid leading system message. Mid-conversation non-text retention from #3553 remains unchanged.

## Type of Change

- [x] Bug fix (non-breaking change fixes an issue)
- [ ] New feature (non-breaking change adds functionality)
- [ ] Breaking change (fix or feature cause existing functionality change)
- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Handle leading system content in `headroom/proxy/helpers.py` by preserving representable text blocks and removing invalid leading system sections.
- Preserve existing `cache_control` metadata and warn when a leading block cannot be represented as top-level text.
- Add regression coverage for mixed and all-non-text leading sections, warning behavior, and mid-conversation retention.

## Testing

- [x] New tests added new functionality

### Test Output

```text
./.venv/bin/python -m pytest -q tests/test_proxy_handler_helpers.py --disable-warnings
59 passed, 2 warnings

./.venv/bin/python -m pytest -q tests/test_proxy_anthropic_cache_stability.py tests/test_proxy_system_prompt_immutable.py tests/test_proxy_byte_faithful_forwarding.py --disable-warnings
102 passed, 1 skipped, 6 warnings

./.venv/bin/ruff check headroom/proxy/helpers.py tests/test_proxy_handler_helpers.py
All checks passed

git diff --check
passed
```

## Real Behavior Proof

- Environment: Linux, Python 3.14.6 host with the repository virtual environment; helper-level verification only, with no live Anthropic provider call.
- Exact command / steps: Ran the focused helper tests above and called `relocate_system_messages_to_top_level` with mixed and all-non-text leading `system` sections, including a text block with `cache_control` and an image block.
- Observed result: `changed=True`, the cleaned messages no longer began with a system-role message, top-level `system` contained valid text blocks, cached text metadata survived, and the unrepresentable image block produced `event=system_relocation_block_dropped`.
- Not tested: live Anthropic or other upstream requests, hosted CI, the full repository lint command, and full mypy.

## Runtime Rollout Safety

- Stable/default behavior changed: yes; invalid leading system sections are normalized, and unrepresentable leading blocks are dropped with a warning
- Kill switch / disable path: revert the commit
- Qualification impact: no live-provider qualification was performed
- Rollback path: revert the commit; no migration is required

## Review Readiness

- [x] I performed self-review
- [x] This PR ready human review

## Checklist

- [x] My code follows project's style guidelines
- [x] I performed self-review my code
- [ ] I commented my code, particularly in hard-to-understand areas
- [ ] I made corresponding changes documentation
- [ ] changes generate no new warnings
- [x] I added tests prove my fix effective feature works
- [x] New and existing unit tests pass locally changes
- [x] I did **not** edit `CHANGELOG.md` — generated by release-please from Conventional Commit PR title (a CI guard enforces this)

Not applicable.

## Additional Notes

The leading-section content-loss behavior is intentional: image/document blocks cannot be represented in Anthropic's top-level `system` field, while retaining them at `messages[0]` would preserve the invalid wire shape. Mid-conversation non-text blocks remain in place. Full lint, mypy, hosted CI, live provider behavior, and maintainer approval remain unknown.


JerrettDavis (changes requested):
The new _coerce_system_block() can still emit an invalid top-level Anthropic text block. Any dict with a non-empty text value is copied wholesale and only its type is overwritten. For example, a leading block shaped like {"type":"image","text":"caption","source":{...}} becomes {"type":"text","text":"caption","source":{...}}; similarly a tool-shaped block can retain id/name/input. Anthropic's top-level system field accepts text-block fields, not those incompatible payload fields, so this can preserve the same 400-class wire failure under a different type.

Please construct a valid text block from an explicit allowlist (at minimum type, text, and supported cache metadata) instead of spreading the original dict, and add a regression test with a wrong-type block that has both text and incompatible fields.

PR #3672, opened 2026-09-20
feat(proxy): let /transformations/feed omit message bodies for number-only pollers

gglucass:
## Description

`/transformations/feed` always returns `request_messages`, `compressed_messages` and `response_content` for every entry, and `RequestLogger.get_recent_with_messages` builds them with `asdict()`, which deep-copies each transcript first. A poller that only reads the per-request numbers (tokens, model, transforms) still pays for a full copy and a multi-MB JSON serialization on the event loop for each pull. Measured on a desktop client polling every 20 s: a `limit=100` pull is ~44 MB, of which ~400 bytes per entry are read, and three concurrent pulls take `/stats?cached=1` from 45 ms to 1.3 s.

This adds `?include_messages=0` (default `1`, byte-identical output to today). When off, the three body fields are omitted from the response and never walked server-side.

Closes #

## Type of Change

- [x] Performance improvement

## Changes Made

- `RequestLogger.get_recent_with_messages(n, include_messages=True)`: when `False`, builds each entry from the dataclass fields minus the three payload fields, deep-copying only what it returns (same idiom as #3613 for `get_recent`).
- `/transformations/feed` accepts `include_messages: bool = True` and omits the three body keys when it is off. Everything else in the item, including `log_full_messages`, is unchanged.
- Each item also carries `uncached_input_tokens`, `cache_write_tokens` and `cache_read_tokens` (already stored on `RequestLog`, never exposed by the feed). A number-only poller can then rate `tokens_saved` against new input (uncached + cache_write) the way `/stats` `new_input_savings_percent` does; the item's full-transcript `savings_percent` is unchanged. Present in both the full and the slim response.
- Tests: logger test proves the payloads are not traversed (a leaf whose `__deepcopy__` raises); endpoint test proves the slim item equals the full item minus the three keys and carries the cache split.

## Testing

### Test Output

```text
$ uv run --frozen --extra dev pytest tests/test_proxy/test_transformations_feed.py tests/test_proxy/test_request_logger.py -q
9 passed in 4.53s
$ uvx ruff check headroom/proxy/request_logger.py headroom/proxy/server.py tests/test_proxy/test_transformations_feed.py tests/test_proxy/test_request_logger.py
All checks passed!
$ uv run --frozen --extra dev mypy headroom/proxy/request_logger.py headroom/proxy/server.py
```

## Real Behavior Proof

- Environment: macOS 15, Python 3.12, this branch, in-process `RequestLogger` seeded with 100 entries carrying ~160 KB of messages each (the size of a mid-session Claude Code turn).
- Exact command / steps: `get_recent_with_messages(100)` and `get_recent_with_messages(100, include_messages=False)`, timed, then `json.dumps` of each result, timed and sized.
- Observed result: include_messages=True: copy 29.4 ms, json 56.5 ms, 24.45 MB. include_messages=False: copy 1.3 ms, json 1.9 ms, 0.05 MB. Same 100 entries, same numeric fields.
- Not tested: the dashboard UI (it does not pass the parameter, so it gets the default, unchanged response).

## Runtime Rollout Safety

- Minimum rollout channel: n/a (opt-in query parameter)
- Stable/default behavior changed: additive only; without the parameter every existing key is byte-identical, and three integer keys are added to each item
- Kill switch / disable path: callers stop passing `include_messages=0`
- Qualification impact: none
- Rollback path: revert the commit; no persisted state or schema involved

## Review Readiness

## Checklist

N/A

## Additional Notes

- Documentation: the endpoint has no reference page; the parameter is documented in the handler docstring.
- Overlaps #3613 on the `_HEAVY_FIELDS` constant and the import lines; both hunks are identical so either merge order applies cleanly.


JerrettDavis (approved):
Reviewed exact head 791c48194d57d88e66a35e5bc0f841c8ff1c8d6a after rebuilding the two contributor commits onto current main. The slim feed path avoids traversing transcript payloads while preserving the default response; 10 focused request-logger/feed tests pass, Ruff and Ruff format checks are clean, all fresh checks are green, and GitHub reports CLEAN.

PR #3675, opened 2026-09-20
fix(proxy): preserve empty transport error context

### LINKED ISSUE #3674: Transport errors surface as api_error with an empty message (str(httpx.ReadError()) == '')
## Summary

Any transport-layer exception during an Anthropic request is flattened into `{"type":"error","error":{"type":"api_error","message":""}}` — with an **empty message** — because `str()` of several `httpx` exceptions is the empty string. The client sees an unactionable, untyped error and cannot distinguish "the upstream API rejected this" from "the socket dropped."

This is worse than a cosmetic logging problem: clients that treat `api_error` as terminal will discard completed work on what is actually a retryable transport blip.

## Environment

- `headroom-ai` **0.37.0**, Python 3.13.14, macOS (arm64)
- Proxy mode, `127.0.0.1:8787`, client is an agent harness routing all Anthropic traffic through the proxy
- `HEADROOM_REQUEST_TIMEOUT=1800` set (so this is **not** a too-low-timeout case)

## Evidence

`~/.headroom/proxy.stderr.log`:

```
  File ".../httpx/_transports/default.py", line 118, in map_httpcore_exceptions
    raise mapped_exc(message) from exc
httpx.ReadError
```

`httpx.ReadError()` and `httpx.ReadTimeout()` both stringify to `""`:

```python
>>> str(httpx.ReadError())
''
>>> str(httpx.ReadTimeout())
''
```

The handler catches everything and stringifies it — `headroom/proxy/handlers/anthropic.py:3490-3500`:

```python
except Exception as e:
    logger.error(f"[{request_id}] Bedrock backend error: {e}")
    await _finalize_pre_upstream()
    return JSONResponse(
        status_code=500,
        content={
            "type": "error",
            "error": {"type": "api_error", "message": str(e)},
        },
    )
```

`grep -nE "ReadError|ReadTimeout|TimeoutException|RemoteProtocolError" headroom/proxy/handlers/anthropic.py` returns **no matches** — no transport exception is distinguished anywhere in that handler.

## Impact observed

Over one working day, **seven** long-running agent jobs died with this exact empty-message error. Every one had completed its work and failed at the final response. Durations ranged from 3 minutes to 1 hour 42 minutes, so total elapsed time does not predict it. Failures arrived in **bursts across independent sessions simultaneously**, which points at the shared upstream connection rather than any one request.

Because the error is typed `api_error` with no message, the client cannot retry intelligently — it looks like a model/API rejection rather than a dropped socket.

## Suggested fix

1. **Never emit an empty message.** Fall back to the exception class name when `str(e)` is empty:
   ```python
   message = str(e) or f"{type(e).__name__} (no message)"
   ```
2. **Classify transport errors distinctly.** `httpx.ReadError`, `ReadTimeout`, `ConnectError`, `RemoteProtocolError` are not `api_error`. Anthropic's own error taxonomy has `overloaded_error` / `api_error`; a dropped connection is closer to a 502/504 and should be typed so clients can retry. Emitting `api_error` for a socket drop is actively misleading.
3. **Consider retrying idempotent transport failures** inside the proxy before surfacing them, or at minimum document that `api_error` with an empty message means "transport failure, safe to retry."

Point 1 alone would have saved a day of debugging here — the empty string sent us looking at model config and API keys rather than at the proxy's transport.

## Related

The same empty-message pattern was previously hit in this environment via `httpx.ReadTimeout`, which is how the `str(exc) == ""` mechanism was first identified. The generic `except Exception` means every future httpx exception class inherits the same failure mode.

axelray-dev:
## Description

Transport exceptions with empty string representations were returned as Anthropic `api_error` responses with an empty message. Preserve the exception type in the response so clients and operators can identify dropped connections and timeouts.

Closes #3674

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add a shared exception-message formatter with a type-name fallback for empty messages.
- Use the fallback in LiteLLM Anthropic and Bedrock streaming error responses and add regression coverage for read errors and timeouts.

## Testing

### Test Output

```text
$ .venv/bin/pytest tests/test_litellm_transport_errors.py tests/test_proxy/test_bedrock_sse_ping.py -q
7 passed in 8.48s

$ .venv/bin/pytest tests/test_litellm_transport_errors.py tests/test_proxy/test_bedrock_sse_ping.py tests/test_route_advice.py tests/test_bedrock_streaming_input_tokens.py tests/test_backend_streaming_cache_metrics.py -q
33 passed in 9.14s

$ .venv/bin/ruff check tests/test_litellm_transport_errors.py headroom/utils.py headroom/backends/litellm.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/streaming.py
All checks passed!

$ .venv/bin/ruff format --check tests/test_litellm_transport_errors.py headroom/utils.py headroom/backends/litellm.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/streaming.py
5 files already formatted

$ .venv/bin/mypy headroom/utils.py headroom/backends/litellm.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/streaming.py --ignore-missing-imports
```

## Real Behavior Proof

- Environment: Linux x86_64, Python 3.11.16, Headroom checkout at commit `78575deb`
- Exact command / steps: Run the focused LiteLLM Bedrock regression suite with an empty-message `httpx.ReadError` and `httpx.ReadTimeout`.
- Observed result: Non-stream and stream error envelopes now contain `ReadError (no message)` and `ReadTimeout (no message)` instead of an empty message; the handler-level paths are covered as well as the backend formatter.
- Not tested: Live AWS Bedrock credentials or an external upstream transport failure. The full repository shard still has an unrelated DeepSeek pricing assertion failure on the base test suite.

## Runtime Rollout Safety

- Rollout-managed feature(s): None; this is an error-message fallback only.
- Minimum rollout channel: Stable; no new rollout configuration is required.
- Stable/default behavior changed: Empty transport exceptions now identify their exception type instead of returning an empty error message.
- Kill switch / disable path: Revert this PR; no persisted state or migration is involved.
- Qualification impact: None; the fallback is covered by focused backend and proxy tests.

## Review Readiness

## Checklist

N/A


JerrettDavis (changes requested):
The shared fallback is useful, but the production AnyLLM path still returns blank transport errors before these handler fallbacks can run. AnyLLMBackend catches exceptions internally and continues to build both non-streaming and streaming error payloads with str(e); the new proxy tests replace AnyLLMBackend with a fake that raises directly, so they bypass the real catch/serialization layer and pass while an actual empty-message httpx error still reaches the client as an empty message. The LiteLLM OpenAI-format send/stream methods retain the same direct str(e) behavior as well.

Please apply the normalization at every backend boundary that serializes these transport exceptions, and exercise the real backend catch paths (including the AnyLLM path) rather than only a handler-level fake. That will make the regression tests fail against the current production behavior and verify the issue is fixed across the supported response formats.

axelray-dev:
Addressed the review request in commit 4047a9f. Transport exception normalization now covers AnyLLM generic and streaming paths plus LiteLLM/OpenAI send and stream serialization, and the focused transport regression tests pass in CI. The remaining red test (2)-(4) jobs are unrelated existing cost-budget/counterfactual-pricing failures; the changed transport test module passes. Please re-review the current head.

JerrettDavis (dismissed):
Reviewed exact head e2c19baf1b46531414615cdc5d228cae8d0dab3c against current main. The change is correct and appropriately tested, the head is mergeable and contains current main, and all reported checks are terminal green.

chopratejas (commented):
Reviewed at head `e2c19baf`. **The logging fix is right and worth having. One defect blocks it: the same string is also fed to the error classifier, and that reclassifies errors on the wire.** Inline suggestions below; they are two lines and keep the whole benefit of the PR.

`format_exception_message` itself is good — `str(exc) or f"{type(exc).__name__} (no message)"`. Turning `{"type":"api_error","message":""}` and `LiteLLM streaming error: ` into something an operator can act on is a real improvement, and the 205-line test file is thorough.

### The defect

At `litellm.py:1189` and `:1798` the diff changes `error_str = str(e).lower()` to `error_str = error_message.lower()`. That variable is not a log string — it is the input to the status-code classifier, which substring-matches for `"rate"`, `"limit"`, `"authentication"`, `"credentials"` and `"not found"`. Those keywords are written for provider *messages*. It now receives a synthesized string containing the exception *class name*:

| exception (empty message) | before | after | |
|---|---|---|---|
| `RateLimitError` | 500 | 429 | arguably an improvement |
| `AuthenticationError` | 500 | 401 | arguably an improvement |
| `NotFoundError` | 500 | 500 | inconsistent — `notfounderror` has no space, so `"not found"` never matches |
| **`GenerateContentError`** | 500 | **429** | false positive: "gene**rate**" contains "rate" |
| **`TokenLimitError`** | 500 | **429** | a context-length error reported as a rate limit |

The 429s are the ones that hurt. Clients treat 429 as retry-with-backoff, so a permanently-failing, non-retryable error turns into a retry loop instead of surfacing to the user. And `GenerateContentError` is not a strawman — that shape of class name is common in the provider SDKs litellm wraps, and the matcher cannot tell it from a genuine rate limit.

The fix is to keep classification on the raw `str(e)`, exactly as today, and use `format_exception_message` only for the log line and the human-readable `message` field. That preserves every bit of what this PR is for.

### Rebase

This branch conflicted with `main` because #3678 landed and removed six now-dead `_finalize_pre_upstream()` calls from `handlers/anthropic.py`; your branch still carried one in the Bedrock error path.

**I have rebased it locally and resolved that** — keeping your `format_exception_message` change and dropping the line #3678 deleted. The rebased branch is 0 commits behind `main`, and `tests/test_litellm_transport_errors.py` plus `tests/test_anthropic_pre_upstream_backpressure.py` give 33 passed / 4 skipped on it.

I am deliberately **not** force-pushing that yet, so your branch is not rewritten twice — it makes sense to land the classifier fix first and rebase once. Tell me if you would rather have the rebase now and fix on top, and I will push it.

### One scope note

The title says `fix(proxy)`, but the diff also rewrites `backends/anyllm.py` at five sites. Accurate scope in the title would help whoever reads this in the changelog.

headroom/backends/litellm.py
```diff
@@ -1177,13 +1179,14 @@ async def send_message(
...
-            logger.error(f"LiteLLM error: {e}")
+            error_message = format_exception_message(e)
+            logger.error(f"LiteLLM error: {error_message}")
 
             # Map to Anthropic error format
             error_type = "api_error"
             status_code = 500
 
-            error_str = str(e).lower()
+            error_str = error_message.lower()
```
chopratejas: `error_str` is the status-code classifier's input, not a log string. Feeding it the synthesized `"{ClassName} (no message)"` makes the keyword matcher see the exception *class name*, so an empty-message `GenerateContentError` matches "rate" and is returned to the client as **429** — which clients retry with backoff. Classify on the raw message, as today:

```suggestion
            error_str = str(e).lower()
```

`error_message` stays exactly as you have it for the `logger.error` call and for the user-facing `message` / `error` fields below — that is the part of this PR that matters, and it is unaffected.

headroom/backends/litellm.py
```diff
@@ -1784,13 +1788,14 @@ async def send_openai_message(
...
-            logger.error(f"LiteLLM OpenAI error: {e}")
+            error_message = format_exception_message(e)
+            logger.error(f"LiteLLM OpenAI error: {error_message}")
 
             # Map to OpenAI error format
             error_type = "api_error"
             status_code = 500
 
-            error_str = str(e).lower()
+            error_str = error_message.lower()
```
chopratejas: `error_str` is the status-code classifier's input, not a log string. Feeding it the synthesized `"{ClassName} (no message)"` makes the keyword matcher see the exception *class name*, so an empty-message `GenerateContentError` matches "rate" and is returned to the client as **429** — which clients retry with backoff. Classify on the raw message, as today:

```suggestion
            error_str = str(e).lower()
```

`error_message` stays exactly as you have it for the `logger.error` call and for the user-facing `message` / `error` fields below — that is the part of this PR that matters, and it is unaffected.

PR #3677, opened 2026-09-21
fix(memory): avoid injecting tools into tool-free requests

### LINKED ISSUE #3676: [BUG] Memory tools are injected into requests that declare no tools, turning text completions into unserviceable tool calls
**Version:** headroom-ai 0.30.0 (first observed), still present in the 0.37.0 wheel
**Client:** claude-agent-sdk (Python) driving the bundled Claude CLI with `--tools ''`, `ANTHROPIC_BASE_URL` pointed at the proxy, `--memory` on. Any API client that sends no `tools` is exposed the same way.

## Summary

With memory enabled, the proxy appends the five memory tool definitions (`memory_save`, `memory_search`, `memory_update`, `memory_delete`, `memory_list`) to the upstream request even when the inbound request carries no `tools` at all (absent or `[]`). Under the Anthropic API an absent or empty `tools` list means a plain text completion. The proxy turns it into a tool-capable request advertising five tools the caller cannot service. The model calls one, the client rejects it, and the turn is spent with no text and a full bill.

## Observed

- Client-side error: `No such tool available: memory_search`. The definition is not in the client's registry, so it can only have arrived on the wire from the proxy.
- claude-agent-sdk with `max_turns=1`: `error_max_turns` on every attempt with the proxy in the path (two attempts, 136 s and 145 s, roughly 0.15 USD each, empty result). Same process, same prompt, same credential with `ANTHROPIC_BASE_URL` unset: a full 15 KB response on the first attempt. One variable.
- The trigger is prompt-dependent: a 30-char "Reply with the single word OK." with `tools=[]` came back clean on turn 1; a 6.5 KB extraction prompt tripped it every time. So the injection is gated by the memory decision upstream of the append, not by the inbound tools list.

## Where (0.37.0 wheel)

- `proxy/handlers/anthropic.py:2638-2652` calls `apply_session_sticky_memory_tools`, gated only on `memory_handler.config.inject_tools`.
- `proxy/helpers.py:2134` does `list(existing_tools) if existing_tools else []` and appends unconditionally.

In 0.30.0 the same chain was `handlers/anthropic.py:1719` (memory block, `memory_decision.inject`), `:1815` `compute_memory_tool_definitions('anthropic')`, `:1820` `apply_session_sticky_memory_tools`, `helpers.py:2248`, `memory_handler.py:485,530`, `memory_tool_adapter.py:611` `inject_tools`. Nothing in either version inspects whether the inbound `tools` list was absent or empty.

## Existing knobs do not cover it

- `HEADROOM_NO_MEMORY_TOOLS=1`: global, drops the tools for every client, memory writes stop working for clients that do want them.
- `HEADROOM_MEMORY_INJECTION_MODE=disabled`: global, kills context injection as well. The `tool` value also yields `inject=False` despite its name.
- `x-headroom-bypass: true`: per request, but client-side, and disables compression for that request too.
- `HEADROOM_EXCLUDE_TOOLS`: compression exclusion, not injection.

## Proposed fix

Early return in the memory tool injection chain when the inbound `tools` list is absent or empty: a request that declares no tools is a text completion and should stay one. Memory context injection (system prompt text) can still apply; only the tool definitions should be withheld. A tools-aware per-request knob would also work, but the early return matches what the API contract already says.

## Repro

1. Run the proxy with `--memory`.
2. Send a `/v1/messages` request with no `tools` key and a prompt long enough to trip the memory decision (a few KB of instructions asking for structured JSON output reproduced it reliably; a one-line prompt did not).
3. Observe a `tool_use` block for `memory_search` (or another `memory_*` tool) in the response.
4. Repeat with `ANTHROPIC_BASE_URL` unset: text completion, no `tool_use`.

axelray-dev:
## Description

Memory-enabled requests that omit tools or send an empty list can receive memory tool definitions through sticky replay. Clients that do not register tools can then receive unserviceable tool calls. Gate memory tool injection on the client declaring at least one tool while preserving existing client tools.

Closes #3676

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- Add a client-tool declaration gate before fresh or sticky memory tool injection.
- Apply the gate to Anthropic and OpenAI handler paths and preserve proxy-owned tools.
- Add regression coverage for fresh requests and sticky sessions without client tools.

## Testing

### Test Output

```text
$ .venv/bin/pytest tests/test_memory_tool_session_sticky.py tests/test_issue_728_empty_tools_injection.py -q
41 passed in 1.43s

$ .venv/bin/ruff format --check headroom/proxy/tool_injection_logging.py headroom/proxy/helpers.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py tests/test_memory_tool_session_sticky.py
5 files already formatted

$ .venv/bin/ruff check headroom/proxy/tool_injection_logging.py headroom/proxy/helpers.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py tests/test_memory_tool_session_sticky.py
All checks passed!

$ .venv/bin/mypy headroom/proxy/helpers.py headroom/proxy/tool_injection_logging.py headroom/proxy/handlers/anthropic.py headroom/proxy/handlers/openai.py --ignore-missing-imports
```

## Real Behavior Proof

- Environment: Linux, Python 3.11.16, Headroom checkout at the current PR head.
- Exact command / steps: Seed a sticky session with a client-declared tool, then call the memory helper for the same session with `client_declared_tools=False` and an empty tool list.
- Observed result: The seeded request injected memory tools; the later tool-free request returned `([], False)`, so no memory tools were added or replayed.
- Not tested: Live external Anthropic or OpenAI upstream calls. The full repository shard still has an unrelated DeepSeek pricing assertion failure on the base test suite.

## Runtime Rollout Safety

- Rollout-managed feature(s): None; this restores the existing tool declaration boundary for memory injection.
- Minimum rollout channel: Stable; no new rollout configuration is required.
- Stable/default behavior changed: Requests without client-declared tools no longer receive memory tool definitions; tool-capable requests retain existing behavior.
- Kill switch / disable path: Existing memory injection configuration remains available; no new override is introduced.
- Qualification impact: None; the behavior is covered by targeted regression tests.
- Rollback path: Revert this commit; no schema or persisted-data migration is involved.

## Review Readiness

## Checklist

- [x] New and existing targeted unit tests pass locally with my changes

N/A


JerrettDavis (changes requested):
The new guard fixes Anthropic and OpenAI chat, but the OpenAI Responses paths still call apply_session_sticky_memory_tools without client_declared_tools, so its default True preserves the bug there. This includes the HTTP /v1/responses injection path and the WebSocket Responses path: both derive an empty tools list when the client omitted tools or sent an empty list, and can still add memory tools.

Please propagate the original client-tools decision through every production call site, and add endpoint-level coverage for tool-free Responses requests, including WebSocket if that path supports memory injection. The current helper-only test passes even when a handler forgets the new argument, which is exactly what happened here.

axelray-dev:
Addressed the review request in commits 28208c0 and e78469b. Both HTTP /v1/responses and WebSocket Responses paths now propagate whether the client actually declared tools, with omitted/empty-tool endpoint coverage. The focused WebSocket suite passes in CI. The remaining red test (2)-(4) jobs are unrelated existing cost-budget/counterfactual-pricing failures; the changed Responses/WS test module passes. Please re-review the current head.

JerrettDavis (changes requested):
The new `client_declared_tools=bool(...)` gate is too broad for Codex traffic. An omitted or empty `tools` array does not establish that a known tool-capable client cannot service Headroom memory tools; Codex capability is established out-of-band. On this exact head, CI shard `test (3)` now fails 19 established routing/WebSocket cases in `test_openai_codex_routing.py` and `test_openai_codex_ws_lifecycle.py` (missing injected `tools`/`include`, including sticky replay and frame-shape cases). Please preserve the existing Codex memory-tool contract while suppressing injection only for requests that are actually known to be tool-incapable, and add coverage for that distinction. The full suite must return green.

axelray-dev:
Updated the memory-tool capability check in 54e272a9. Omitted or empty tools now preserve Codex and other tool-capable clients; suppression is limited to the explicitly known tool-incapable client. Fresh CI passed every relevant check except the existing test_install_restart_waits_for_old_process_before_starting failure, which raises Manifest.port and is outside this PR; the shard reported 3308 passed and 148 skipped. Please re-review the new head.

PR #3224, opened 2026-08-23
fix(install): replace the launchd wrapper with the proxy listener

rodboev:
## Description

The persistent service path can leave a wrapper, listener, native supervisor, or Docker container with conflicting ownership. Rebuild runtime ownership so launchd directly owns Python, Docker owns Docker across service/task presets, status and stop require deployment identity, readiness cannot be health-only, and failed cleanup preserves rollback state. Closes #3208

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (feature that would cause existing behavior to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)

## Changes Made

- Make the launchd-owned Python service exec the listener in place.
- Classify Docker ownership from both runtime and preset, with no competing native supervisor.
- Route Docker service/task start, restart, recovery, and teardown through the Docker owner.
- Verify exact process command, port, deployment profile, and runtime before status or stop.
- Add POSIX, macOS, and Windows process identity fallbacks without requiring psutil.
- Require identity on readiness and fail closed when cleanup cannot prove the new owner stopped.
- Persist an inactive previous-deployment snapshot, keep active state durable before owner start, and make environment/provider mutation activation transactional.
- Surface cleanup, persistence, supervisor, Docker, and recovery-snapshot deletion failures with actionable recovery guidance.

## Testing

- [x] Unit tests pass (`uv run --locked pytest tests/test_provider_package_init.py::test_install_registry_apply_provider_scope_mutations_skips_missing_and_none tests/test_install/test_providers.py tests/test_cli/test_install_cli.py tests/test_install/test_runtime.py tests/test_install/test_supervisors.py tests/test_install/test_paths.py tests/test_install/test_state.py -q`)
- [x] Linting passes (`uv run --locked ruff check` on changed install, provider, supervisor, and focused test files)
- [x] Type checking passes (`uv run --locked mypy headroom --ignore-missing-imports`)
- [x] New tests added for the regression
- [ ] Manual testing performed through live process/supervisor environments

### Test Output

```text
uv run --locked pytest tests/test_provider_package_init.py::test_install_registry_apply_provider_scope_mutations_skips_missing_and_none tests/test_install/test_providers.py tests/test_cli/test_install_cli.py tests/test_install/test_runtime.py tests/test_install/test_supervisors.py tests/test_install/test_paths.py tests/test_install/test_state.py -q
[...]

Issue #3208, opened 2026-08-22
KeepAlive does not recover a crashed launchd proxy: run_foreground Popen()s the listener instead of exec


thunter009:
## Summary

`headroom install apply --preset persistent-service` writes a launchd plist with `KeepAlive=true`, but a crash of the proxy is not reliably recovered. The launchd job PID is `headroom install agent run`, which **Popen()s** the listener instead of `exec`'ing it (`headroom/install/runtime.py` `run_foreground`).

```
launchd com.headroom.default  KeepAlive=true
  run-headroom.sh
    exec → python … `headroom install agent run`     ← job PID (parent)
             Popen → python -m headroom.cli proxy     ← :8787 listener (child)
```

## What happens on SIGKILL of the job PID

1. `_cleanup` (the SIGTERM handler that `terminate()`s the child) never runs — SIGKILL is uncatchable.
2. The child is reparented to PID 1 and can keep holding `:8787`.
3. KeepAlive restarts the parent.
4. The new `Popen` fails to bind (or races with the orphan).
5. Rapid exit → launchd throttles. Observed 2026-08-19 on macOS 26.6.1: `state = not running`, `runs = 2`, no respawn for 5+ minutes.

Reproduced against **headroom-ai 0.36.4** on 2026-08-22. Killing the *child* (the listener) does let the parent `wait()` and KeepAlive can restart; killing the *parent* is the broken path. A brief window with **two** proxy processes (orphan at PPID 1 + new child) was observed after parent SIGKILL.

## Why this is a Headroom bug, not a launchd one

`KeepAlive` is doing what it is supposed to: it is watching the process it spawned. That process is a wrapper whose death does not take the listener with it.

The generated runner already `exec`s `headroom install agent run`. If `run_foreground` itself `os.execvpe`'d `build_runtime_command(...)` (after installing the SIGTERM handler is no longer needed, or with stdout already redirected), the job PID would *be* the listener and KeepAlive would actually recover a crash.

`stop_runtime` / `runtime_status` already key off `runner.pid`, which today is the **child**. An exec would make that PID the job PID as well — `echo $$ > runner.pid` before exec, same file.

## Suggested fix

In `run_foreground`:

- Redirect stdout/stderr to the log file (already done).
- Write `os.getpid()` to the pid file (not `proc.pid`).
- `os.execvpe(command[0], command, env)` instead of `subprocess.Popen` + `wait()`.

[...]

PR #3679, opened 2026-09-21
fix(install): wait for the old runtime to stop before restart starts it

### LINKED ISSUE #3658: install restart leaves the deployment stopped (reports success)

Ayushraj06-bit:
## Description

`headroom install restart` printed `Restarted deployment '<name>'.` and left the service stopped; `install start` afterwards worked. The reporter hit it 3/3 on macOS with the `persistent-service` preset (launchd, user scope).

Root cause: `_stop_deployment` returns as soon as `launchctl bootout` and the SIGTERM have been *sent*. `bootout` is asynchronous (it returns once the job has been told to go — the same window `_bootstrap_with_retry` already rides out for EIO), and a SIGTERM'd uvicorn keeps accepting connections until its next event-loop tick. The very next thing `restart` does is `_start_deployment`, whose first line is the "already healthy? then no-op" probe added for `install start` idempotency in #1851. That probe lands ~15–20 ms after the signal, gets `ready: true` from the dying proxy, and the start is skipped. The old proxy then exits, the job is no longer bootstrapped, and KeepAlive has nothing to restart — hence "several minutes" of `Language model unavailable`.

Fix: add `wait_stopped()` in `headroom/install/runtime.py` (the mirror of `wait_ready()`) and call it at the end of `_stop_deployment`, so `stop`, `restart`, `apply` on a running profile and the restore-after-failed-apply path only continue once the ready endpoint has actually gone away. If it is still answering after the bound (15 s, one poll per second), `stop`/`restart` fail with a `ClickException` naming the URL instead of reporting success. Default behavior when the runtime goes down promptly is unchanged apart from one 1 s poll.

Not changed: the early return in `_start_deployment` (correct for `start`; only its precondition was broken), KeepAlive recovery of a SIGKILLed wrapper (#3208), and the ownership rework in #3224 — that PR also touches `_stop_deployment` and would rebase over this in a few lines; its pid-file gate on the early return does not close this race on its own (the new proxy would then bind-race the draining one).

Closes #3658

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

- `headroom/install/runtime.py`: `wait_stopped(manifest, timeout_seconds=15)` — polls `probe_ready` until it is false, same shape as `wait_ready`.
- `headroom/cli/install.py`: `_stop_deployment` calls `wait_stopped` after `stop_supervisor`/`stop_runtime` and raises `ClickException("Deployment '<profile>' is still answering <health_url> after stop.")` on timeout.
- `tests/test_cli/test_install_cli.py`: regression test `test_install_restart_waits_for_the_old_runtime_before_starting` (probe answers ready once after stop, then not; asserts `start_supervisor` runs and one poll happened); `test_install_restart_fails_when_the_old_runtime_keeps_answering` and `test_install_stop_fails_when_the_runtime_keeps_answering`; the existing restart test now asserts `wait_stopped` sits between `stop_runtime` and `start_supervisor`; the apply helpers/tests that reach `_stop_deployment` stub `wait_stopped` the same way they already stub `wait_ready`.
- `tests/test_install/test_runtime.py`: `wait_stopped` polling/timeout assertions next to the `wait_ready` ones.

## Testing

- [x] Unit tests pass (`pytest`) — `tests/test_cli/test_install_cli.py tests/test_install/` (209 passed, 1 skipped) and the wider set that imports the install package (280 passed; the one failure, `tests/test_providers_opencode_config.py::test_build_launch_env_with_project`, is a pre-existing Windows-only path-escaping failure identical on `main`). Full suite not run locally; CI.
- [x] Linting passes (`ruff check .`) — `All checks passed!`, `ruff format --check .` → `1614 files already formatted` (ruff 0.16.7, the pinned version).
- [x] Type checking passes (`mypy headroom`) — scoped to the two changed modules: `Success: no issues found in 2 source files`. The repo-wide run was not attempted locally; CI.

### Test Output

```text
$ ruff check . && ruff format --check .
All checks passed!
1614 files already formatted

$ mypy headroom/cli/install.py headroom/install/runtime.py --ignore-missing-imports

$ pytest tests/test_cli/test_install_cli.py tests/test_install/ -q
209 passed, 1 skipped in 41.64s

# the three new/changed restart tests against unpatched main (fix stashed, tests kept):
E   AssertionError: assert [] == ['start_supervisor']        # restart never started anything
FAILED test_install_restart_waits_for_the_old_runtime_before_starting
FAILED test_install_restart_uses_internal_helpers              # AttributeError: no wait_stopped
FAILED test_install_restart_fails_when_the_old_runtime_keeps_answering
3 failed in 0.72s

# mutation check: each of these breaks the fix and is caught
M1 result of wait_stopped ignored           -> 4 failed (all four new/changed tests)
M2 wait_stopped does not poll               -> 2 failed (restart regression test + runtime unit test)
M3 timeout swallowed, success printed anyway -> 2 failed (restart-fails + stop-fails)
M4 wait moved before stop_runtime           -> 1 failed (call-order test)
```

## Real Behavior Proof

- Environment: Windows 11, Python 3.11.9, uvicorn 0.32.1 / fastapi 0.115.5 (the proxy's server stack), headroom from this branch and from `main` (`67eb910e`). No launchd here, so the supervisor primitives were swapped for their effect: `stop_supervisor` → send the server its graceful shutdown signal and return immediately (what `launchctl bootout` does), `start_supervisor` → spawn the server again (what `bootstrap` + `RunAtLoad` does). Everything else in the command is the real code: `install_restart`, `_stop_deployment`, `_start_deployment`, `probe_ready`, `wait_ready`, `wait_stopped`, against a real uvicorn server serving `/readyz` with `timeout_graceful_shutdown=10`.
- Exact command / steps: (1) `python repro_window.py` — start the server, send the shutdown signal, call headroom's `probe_ready` immediately, five runs; (2) `python repro_restart_e2e.py` — with the server running, invoke `headroom install restart --profile copilot` through `CliRunner`, log every `probe_ready` call relative to the shutdown signal, sleep 3 s, then probe `/readyz` and count `start_supervisor` calls; run on `main` (3×) and on this branch (5×).
- Observed result: (1) `probe 15 ms after shutdown signal -> probe_ready=True` in 5/5 runs — the server answers ready after it has been told to exit. (2) `main`: 2 of 3 runs `probes after the shutdown signal (ms, ready): [(19, True)]` → `restart: exit=0 output="Restarted deployment 'copilot'."` → `after: Healthy: no | start_supervisor called: 0` — the issue verbatim; the third run's probe hung past the window and restart worked. This branch: 5/5 `RESULT: OK — restart actually restarted it`, `start_supervisor called: 1`, `Healthy: yes`; restart took 2.6–5.6 s here because a connect to a closed loopback port times out (2 s) on this machine instead of being refused — on macOS/Linux the refused connect is immediate, so the added wait is one 1 s poll.
- Not tested: a real launchd deployment on macOS (I don't have one; @szulawski, if you can run `headroom install restart --profile copilot` from this branch, that is the proof that matters), systemd, and the docker preset (`docker stop` is synchronous, so `wait_stopped` returns on its first probe there — covered by reasoning, not by a run).

## Runtime Rollout Safety

- Stable/default behavior changed: `install stop`/`install restart` (and `apply` on a running profile) now wait until `/readyz` stops answering before continuing — typically one 1 s poll — and exit 1 with `Deployment '<profile>' is still answering <health_url> after stop.` if it is still up after 15 s, where they previously printed success. Success output is unchanged.
- Kill switch / disable path: none needed; on a timeout the deployment is in the same state the old code left it in, and `install start` still works.
- Qualification impact: none.

## Review Readiness

## Checklist

## Additional Notes

Start at `_stop_deployment` in `headroom/cli/install.py`, then `wait_stopped` next to `wait_ready` in `headroom/install/runtime.py`; the tests read top-down from `test_install_restart_uses_internal_helpers`. The decision I'd most like challenged: failing `install stop` loudly on timeout. The alternative is a warning-only `stop` with the hard failure kept for `restart`; I went with the failure because "Stopped deployment" printed while the port still answers is the same false success the issue is about.

Docs: `docs/content/docs/persistent-installs.mdx` only lists the commands, so nothing to update.


Ayushraj06-bit:
Superseded by #3670, which was opened first with the same root cause and fix; I've left the launchd timing measurements and the end-to-end run there. The only difference here is placing the wait in `_stop_deployment` so `stop`+`start` and `apply` on a running profile are covered too — raised as a suggestion on #3670.

